PR #304, opened 2019-11-19
Timeout ceph must gather

crombus:


crombus:
yes no need take any special measures

PR #305, opened 2019-11-19
fix(ocs-operator): Fixes panic of ocs-operator when running vmware

ashishranjan738:
This commit fixes the panic of the ocs-operator when running in an vmware based ocp cluster. The operator crashes because its not able to assign a valid rack to the nodes as nodes in vmware based ocp cluster doesn't have any valid topology keys (https://github.com/openshift/ocs-operator/blob/master/pkg/controller/storagecluster/reconcile.go#L46-L50). So the current work around is to assign a rack with min number of nodes to the new node if the operator is not able to find a valid rack based on the topology key checks(https://github.com/openshift/ocs-operator/blob/master/pkg/controller/storagecluster/reconcile.go#L498-L530).

### DIFF
--- a/pkg/controller/storagecluster/reconcile.go
+++ b/pkg/controller/storagecluster/reconcile.go
@@ -495,36 +495,42 @@ func determinePlacementRack(nodes *corev1.NodeList, node corev1.Node, minRacks i
 		}
 	}
 
-	for rack := range nodeRacks.Labels {
-		nodeNames := nodeRacks.Labels[rack]
-		if len(nodeNames) == 0 {
-			rackList = append(rackList, rack)
-			continue
-		}
+	if len(targetAZ) > 0 {
+		for rack := range nodeRacks.Labels {
+			nodeNames := nodeRacks.Labels[rack]
+			if len(nodeNames) == 0 {
+				rackList = append(rackList, rack)
+				continue
+			}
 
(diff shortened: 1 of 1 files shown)


pkg/controller/storagecluster/reconcile.go
```diff
@@ -530,6 +530,11 @@ func determinePlacementRack(nodes *corev1.NodeList, node corev1.Node, minRacks i
 	}
 
 	sort.Strings(rackList)
+	if len(rackList) == 0 {
```
rohantmp: instead of `len(rackList) == 0`, I recommend `if len(tartgetAZ) == 0`, then:

```
if len(targetA) > 0 {
    # old method for pupulating rackList
} else {
    # your new method
}

Comment (rohantmp) [inline pkg/controller/storagecluster/reconcile.go]: instead of `len(rackList) == 0`, I recommend ` len(targetAZ) > 0`:

```
if len(targetAZ) > 0 {
    # fill rackList with nodes from targetAZ like before
} else {
    # just fill rackList with all the nodes like you're doing in the PR
}
```

ashishranjan738: done

obnoxxx:
(for the record, @rohantmp 's requests have been impemented)

PR #313, opened 2019-11-21
Remove "internal" prefix from BackingStore and BucketClass

umangachapagain:
Remove "internal" prefix from BackingStore and BucketClass as it is supposed to be used by the users.


obnoxxx:
@umangachapagain :

```
uncommitted CSV changes. run 'make gen-latest-csv' and commit results.
```

PR #318, opened 2019-11-22
Disable CephObjectStore when the cloud provider is AWS.

rohantmp:
When the cloud provider is AWS, AWS S3 will provide object storage instead of RGW.


functests/storageclusterinitialization_test.go
```diff
@@ -143,53 +143,53 @@ var _ = Describe("StorageClusterInitialization", func() {
...
-				Expect(ok).To(BeTrue())
-			}
-
-			return nil
-		}, 15*time.Second, 1*time.Second).ShouldNot(gomega.HaveOccurred())
-	}
+	// cephObjectStoreUserModify := func(shouldDelete bool) {
+	// 	objStoreUser := &cephv1.CephObjectStoreUser{}
+	// 	key := crclient.ObjectKey{Namespace: namespace, Name: fmt.Sprintf("%s-cephobjectstoreuser", name)}
+	// 	err := client.Get(context.TODO(),
```
obnoxxx: Not sure that commenting out massive blocks of code is a clean way to disable the tests... I think we want a better way.

kshlm:
In the latest e2e run, all the tests actually passed. The after test cleanup failed, which was reported as a failure. @rohantmp Can you please address @obnoxxx comments before we trigger another run.

functests/storageclusterinitialization_test.go
rohantmp: Agreed. Changed to include platform detection in the functests

kshlm:
@rohantmp There seems to be a valid failure with the StorageClusterInitialization functest. The same failure has happened twice. The test most likely requires an update to fix it for this PR.

rohantmp:
I meant to assert `Expect(err).To(BeNil())`, but I accidentally asserted `Expect(err).ToNot(BeNil())`.

A little weirded out by Gomega's syntax, but I can see a lot of useful stuff in it.

functests/storageclusterinitialization_test.go
```diff
@@ -345,12 +352,15 @@ var _ = Describe("StorageClusterInitialization", func() {
 		})
 		Context("after", func() {
 			It("resources have been modified", func() {
+				if currentCloudPlatform == scController.PlatformAWS {
```
obnoxxx: I think you want `!=` here, not `==`. This leads to test failure:
```
 STEP: Modifying CephObjectStore
• Failure [3.559 seconds]
StorageClusterInitialization
/go/src/github.com/openshift/ocs-operator/functests/storageclusterinitialization_test.go:27
  verify re-initialization
  /go/src/github.com/openshift/ocs-operator/functests/storageclusterinitialization_test.go:343
    after
     [...]
      to be nil
      /go/src/github.com/openshift/ocs-operator/functests/storageclusterinitialization_test.go:256

```

functests/storageclusterinitialization_test.go
```diff
@@ -373,12 +383,14 @@ var _ = Describe("StorageClusterInitialization", func() {
 			})
 
 			It("resources have been deleted", func() {
+				if currentCloudPlatform == scController.PlatformAWS {
```
obnoxxx: `!=` instead of `==`

obnoxxx:
I just update the PR with the requested trivial change.

obnoxxx:
updated with more test case fixes to (hopefully) let tests pass now ... :-)

functests/storageclusterinitialization_test.go
rohantmp: Wow, that's really embarrassing..

rohantmp: Thank you :)

PR #322, opened 2019-11-25
enhance(must-gather): Removes collection of empty files

ashishranjan738:
This commit modifies must-gather to remove collection of empty files and directories from the dump collected.

### DIFF
--- a/must-gather/collection-scripts/gather
+++ b/must-gather/collection-scripts/gather
@@ -32,4 +32,5 @@ done
 gather_noobaa_resources ${BASE_COLLECTION_PATH}
 gather_ceph_resources ${BASE_COLLECTION_PATH}
 
+find "${BASE_COLLECTION_PATH}" -empty -delete
 exit 0


must-gather/collection-scripts/gather
```diff
@@ -6,10 +6,12 @@ echo "Collecting operator pod logs"
...
-oc get pods --all-namespaces -l 'name in (local-storage-operator,ocs-operator)' -o jsonpath="${operatorPodLogCollectionJSONPath}" >> collector.sh
-oc get pods --all-namespaces -l 'name in (local-storage-operator,ocs-operator)' -o jsonpath="${operatorPodPreviousLogCollectionJSONPath}" >> collector.sh
-oc get pods --all-namespaces -l 'app in (noobaa,rook-ceph-operator)' -o jsonpath="${operatorPodLogCollectionJSONPath}" >> collector.sh
-oc get pods --all-namespaces -l 'app in (noobaa,rook-ceph-operator)' -o jsonpath="${operatorPodPreviousLogCollectionJSONPath}" >> collector.sh
+{ 
+  oc get pods --all-namespaces -l 'name in (local-storage-operator,ocs-operator)' -o jsonpath="${operatorPodLogCollectionJSONPath}";
+  oc get pods --all-namespaces -l 'name in (local-storage-operator,ocs-operator)' -o jsonpath="${operatorPodPreviousLogCollectionJSONPath}";
+  oc get pods --all-namespaces -l 'app in (noobaa,rook-ceph-operator)' -o jsonpath="${operatorPodLogCollectionJSONPath}";
+  oc get pods --all-namespaces -l 'app in (noobaa,rook-ceph-operator)' -o jsonpath="${operatorPodPreviousLogCollectionJSONPath}"; 
+} >> collector.sh
```
obnoxxx: This is a code cleanup/improvement that should not sneak into the actual change.
I suggest you drop it or put it into a separate commit, maybe even separate PR.

must-gather/collection-scripts/gather
```diff
@@ -24,12 +26,14 @@ resources+=(objectbuckets)
 # Add general resources to list if necessary 
 
 # Run the Collection of Resources using must-gather
-for resource in ${resources[@]}; do
+for resource in "${resources[@]}"; do
```
obnoxxx: Same here, a fix/improvement that is not related to the actual topic of this commit or even PR.

must-gather/collection-scripts/gather
```diff
@@ -24,12 +26,14 @@ resources+=(objectbuckets)
...
     oc adm --dest-dir=${BASE_COLLECTION_PATH} inspect ${resource} --all-namespaces
 done
 
 # Call other gather scripts
 gather_noobaa_resources ${BASE_COLLECTION_PATH}
 gather_ceph_resources ${BASE_COLLECTION_PATH}
 
+find ${BASE_COLLECTION_PATH} -type f -empty -delete
+find ${BASE_COLLECTION_PATH} -type d -empty -delete
```
obnoxxx: This is actually the only hunk of this patch that relates to the commit or even PR. This should be a separate commit, and probably the only commit in this PR.

must-gather/collection-scripts/gather
ashishranjan738: done.

must-gather/collection-scripts/gather
ashishranjan738: yup! done. I will take care of this from next time.

must-gather/collection-scripts/gather
obnoxxx: One possible small improvement: the linux manpage of find says this about `-empty`:
```
-empty File is empty and is either a regular file or a directory.
```
So it should be enough to just have
```
find ${BASE_COLLECTION_PATH} -empty -delete
```

Also please enclose the `${BASE_COLLECTION_PATH}` in double quotes (`"`). The script is not shellcheck proof, but we should at least not add more cases that would need to be fixed later.

ashishranjan738: yes removing the `-type` and keeping only a single statement of the find will do the work. But I think it will miss deleting the directories which contain only a single empty file right ?

ashishranjan738: i think i m wrong! tested it, it is working as expected. Updated the PR.

PR #324, opened 2019-11-25
fix(must-gather): updates collection script to collect CLBO pod logs

ashishranjan738:
This commit fixes must-gather collection scripts to collect CLBO pod logs which were earlier getting missed.

##### Note: Shifted to inverted grep for now as CLBO pods also have `status.Phase` as `Running` because of which `--field-selector` was not able to grab the crashLoopBackOff pods.


ashishranjan738:
Needs backporting for BZ https://bugzilla.redhat.com/show_bug.cgi?id=1770941

egafford:
Discussed this with @ashishranjan738 ; there is an unlikely case here of a pod being excluded if its name contains or is the word "Running", but this can be addressed as a separate and more minor issue.

PR #337, opened 2019-11-26
Fix Scripts according to shellcheck

crombus:


obnoxxx:
Created an issue #341 for this.
If you're interested, @crombus , please take your time to implement this. It would be appreciated!

crombus:
@obnoxxx I'll take up this and start on it. Please assign it to me

obnoxxx:
@crombus I think i can't add you yet, as you have to be a member of the org or something. ==> @jarrpa ?

jarrpa:
Yup, you'd need to be part of the openshift org. I left a comment on the Issue, at least.

ashishranjan738:
This PR needs a rebase. Also i was thinking if we can wait for this PR https://github.com/openshift/ocs-operator/pull/320 as it involves a revamp of must-gather.

obnoxxx:
/hold

holding this until test cases are added

ashishranjan738:
@crombus can you also add the description in the PR.

hack/shellcheck-test.sh
```diff
@@ -0,0 +1,51 @@
...
+                if grep -q '^#.*OCS-OPERATOR-SKIP-SHELLCHECK' "${1}"; then
+                        return 0
+                fi
+                shellcheck -x -e SC2181,SC2029,SC1091,SC1090,SC2012 "${1}"
+        else
+                return 0
+        fi
+}
+
+SHELLCHECK="$(command -v shellcheck 2>/dev/null)"
```
ashishranjan738: @crombus this needs to be changed. As discussed we will be using the shellcheck from this directory `$GOPATH/src/github.com/openshift/ocs-operator/tools/` and not rely on the available OS's shellcheck.

hack/verify-operator-sdk-gen.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+# HEKETI-SKIP-SHELLCHECK
```
jarrpa: Why is this file here?

must-gather/collection-scripts/gather
```diff
@@ -1,40 +1,4 @@
...
-
-# Call other gather scripts
-gather_noobaa_resources ${BASE_COLLECTION_PATH}
-gather_ceph_resources ${BASE_COLLECTION_PATH}
-
-echo "deleting empty files" >> ${BASE_COLLECTION_PATH}/gather-debug.log
-find "${BASE_COLLECTION_PATH}" -empty -delete >> ${BASE_COLLECTION_PATH}/gather-debug.log 2>&1
-exit 0
\ No newline at end of file
+# calling gather.sh from this file as must-gather basically looks for gather file in the path
```
jarrpa: Is this change just so we have a file ending in `.sh` so the shellcheck test can find it? If so, please expand the comment to explain that as well.

hack/common.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+# OCS-OPERATOR-SKIP-SHELLCHECK
```
jarrpa: Please comment as to why we're skipping shellcheck here.

Makefile
```diff
@@ -135,6 +135,10 @@ govet:
 	@echo "Running go vet"
 	go vet ./...
 
+shellcheck-test:
```
jarrpa: Please also add `shellcheck-test` to the list of targets for `ocs-operator-ci` on line 159. It should go first in the list.

hack/shellcheck-test.sh
```diff
@@ -0,0 +1,51 @@
...
+failed=0
+for script in ${SCRIPTS}; do
+        err=0
+        test_syntax "${script}"
+        [[ $? -ne 0 ]] && err=1
+        test_shellcheck "${script}"
+        [[ $? -ne 0 ]] && err=1
+        ((failed+=err))
+        if [[ ${err} -ne 0 ]]; then
+                echo "detected issues in ${script}" >&2
```
jarrpa: Instead of just saying "detected issues", have two echo statements, one above for "syntax test failed" and one here for "shellcheck failed".

hack/shellcheck-test.sh
```diff
@@ -0,0 +1,51 @@
...
+
+
+failed=0
+for script in ${SCRIPTS}; do
+        err=0
+        test_syntax "${script}"
+        [[ $? -ne 0 ]] && err=1
+        test_shellcheck "${script}"
+        [[ $? -ne 0 ]] && err=1
+        ((failed+=err))
```
egafford: Why are we exiting with a code equal to the number of scripts that failed their checks? I can see that it's more information, but it could send someone down an odd rabbit hole if they just notice the code and look for what it means canonically. There might well be a reason, but it seems nonstandard.

hack/shellcheck-test.sh
```diff
@@ -0,0 +1,51 @@
...
+
+SHELLCHECK="$(command -v shellcheck 2>/dev/null)"
+
+SCRIPT_DIR="$(cd "$(dirname "${0}")" && pwd)"
+
+BASE_DIR="$(cd "${SCRIPT_DIR}/.." && pwd)"
+
+if [[ -z "${SHELLCHECK}" ]]; then
+        echo "warning: could not find shellcheck ... installing shellcheck" >&2
+        scversion="stable" # or "v0.4.7", or "latest"
```
egafford: Nit, but as long as we're cleaning up other concerns, the comment here could be misleading. It could read to someone who doesn't know shellcheck as "we're pulling version stable, which is the same as v0.4.7". It's unlikely to cause confusion, but it could.

It's fine to leave the comment in: we don't likely need it, but it might be useful to someone. Still, the shellcheck repo has the comment "Or v0.4.7 for that version, or latest for daily", which is a bit clearer.

must-gather/collection-scripts/gather
obnoxxx: Yeah, that is a little weird.
If we need to have scripts not ending in ".sh", we should probably do something else than searching for `.sh` files, like grepping for `#!/bin/bash` or similar. E.g.:

```
find . \( -path ./vendor -o -path ./.git -o -path ./build \) -prune -o -name *~ -prune -o -type f -exec grep -l -e '^#!/bin/bash$' {} \;
```

That would avoid all the renaming...

Makefile
obnoxxx: Addition to the `ocs-operator-ci` could even be its own commit...

hack/verify-operator-sdk-gen.sh
crombus: @davidvossel created I don't know the exact purpose

must-gather/collection-scripts/gather
crombus: added the complete comment.

hack/shellcheck-test.sh
obnoxxx: Was this explained? Why was it resolved?

crombus: changed the exit to 1

Makefile
```diff
@@ -40,6 +40,7 @@ all: ocs-operator ocs-must-gather ocs-registry
 	cluster-clean \
 	ocs-operator-openshift-ci-build \
 	functest \
+        shellcheck-test \
```
obnoxxx: Now the splitting of the commits in the makefile is slightly off: the commit "Add shellcheck test target" only moves the `.PHONY` entry from a separate entry above the "shellcheck-test" target (which is legitmate), to the global `.PHONY` list (which is possibly more aligned with the current style of the Makefile.

It does not make a lot of sense to add a phony entry in one place and one commit and then move it to another place in a later commit. I assume you meant to have the addition of the shellcheck-test target in this commit as well?

hack/shellcheck-test.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
+  
+# Check for shell syntax & style.
+
```
obnoxxx: You need to set SHELLCHECK somewhere initially, like so:

```suggestion

SHELLCHECK="$(command -v shellcheck 2>/dev/null)"
```

Otherwise no shellcheck will actually be run...

hack/shellcheck-test.sh
```diff
@@ -0,0 +1,57 @@
...
+if [[ -z "${SHELLCHECK}" ]]; then
+        echo "warning: could not find shellcheck ... installing shellcheck" >&2
+        scversion="stable"
+        wget -qO- "https://storage.googleapis.com/shellcheck/shellcheck-${scversion?}.linux.x86_64.tar.xz" | tar -xJv
+        cp "shellcheck-${scversion}/shellcheck" $GOPATH/src/github.com/openshift/ocs-operator/tools/
+        FILE=$GOPATH/src/github.com/openshift/ocs-operator/tools/
+        if [ -f "$FILE" ]; then
+        echo "$FILE exist"
+        fi
```
obnoxxx: What's the purpose of this block? It seems to be here for debugging purposes?

Also, if we succeeded in installing shellcheck, we should set `SHELLCHECK`...

e.g. 

```suggestion
DEST="$GOPATH"/src/github.com/openshift/ocs-operator/tools/
cp "shellcheck-${scversion}/shellcheck" "${DEST}"
if [ -f "${DEST}" ]; then
        SHELLCHECK="${FILE}"
fi
```

hack/shellcheck-test.sh
```diff
@@ -0,0 +1,59 @@
...
+        scversion="stable"
+        wget -qO- "https://storage.googleapis.com/shellcheck/shellcheck-${scversion?}.linux.x86_64.tar.xz" | tar -xJv
+        cp "shellcheck-${scversion}/shellcheck" $GOPATH/src/github.com/openshift/ocs-operator/tools/
+        FILE=$GOPATH/src/github.com/openshift/ocs-operator/tools/
+        if [ -f "$FILE" ]; then
+        SHELLCHECK="${FILE}"
+        fi
+fi
+
+SHELLCHECK="$(command -v shellcheck 2>/dev/null)"
```
obnoxxx: This is too late. Needs to go further above, at least above the block with `if [[  -z "${SHELLCHECK}" ]]`... (see the previous suggestion) Otherwise it will always install, and then still not have shellcheck  if it is not installed in the system. I think in the CI, it did not run shellcheck, as locally, it finds errors in scripts for me.

Makefile
obnoxxx: tab instead of spaces here

hack/shellcheck-test.sh
```diff
@@ -0,0 +1,59 @@
...
+BASE_DIR="$(cd "${SCRIPT_DIR}/.." && pwd)"
+
+if [[ -z "${SHELLCHECK}" ]]; then
+        echo "warning: could not find shellcheck ... installing shellcheck" >&2
+        scversion="stable"
+        wget -qO- "https://storage.googleapis.com/shellcheck/shellcheck-${scversion?}.linux.x86_64.tar.xz" | tar -xJv
+        cp "shellcheck-${scversion}/shellcheck" $GOPATH/src/github.com/openshift/ocs-operator/tools/
+        FILE=$GOPATH/src/github.com/openshift/ocs-operator/tools/
+        if [ -f "$FILE" ]; then
+        SHELLCHECK="${FILE}"
```
obnoxxx: indentation

```suggestion
                SHELLCHECK="${FILE}"
```

hack/shellcheck-test.sh
```diff
@@ -0,0 +1,59 @@
...
+
+
+failed=0
+for script in ${SCRIPTS}; do
+        err=0
+        test_syntax "${script}"
+        [[ $? -ne 0 ]] && err=1
+        if [[ ${err} -ne 0 ]]; then
+                (((failed=err)))
+                echo "detected syntax issues in ${script}}" >$2
```
obnoxxx: typo

```suggestion
                echo "detected syntax issues in ${script}}" >&2
```

hack/shellcheck-test.sh
```diff
@@ -0,0 +1,59 @@
...
+SCRIPT_DIR="$(cd "$(dirname "${0}")" && pwd)"
+
+BASE_DIR="$(cd "${SCRIPT_DIR}/.." && pwd)"
+
+if [[ -z "${SHELLCHECK}" ]]; then
+        echo "warning: could not find shellcheck ... installing shellcheck" >&2
+        scversion="stable"
+        wget -qO- "https://storage.googleapis.com/shellcheck/shellcheck-${scversion?}.linux.x86_64.tar.xz" | tar -xJv
+        cp "shellcheck-${scversion}/shellcheck" $GOPATH/src/github.com/openshift/ocs-operator/tools/
+        FILE=$GOPATH/src/github.com/openshift/ocs-operator/tools/
```
obnoxxx: This FILE is not correct, it points to the directory. You need to set to the actual file in order to run shellcheck at all if it is not already installed in the system. The PR CI did not actually run shellcheck. This is why the ci passed. There are shellcheck errors in this shellcheck-test script.

```suggestion
        FILE="${GOPATH}"/src/github.com/openshift/ocs-operator/tools/shellcheck
```

hack/shellcheck-test.sh
```diff
@@ -0,0 +1,59 @@
...
+
+SCRIPT_DIR="$(cd "$(dirname "${0}")" && pwd)"
+
+BASE_DIR="$(cd "${SCRIPT_DIR}/.." && pwd)"
+
+if [[ -z "${SHELLCHECK}" ]]; then
+        echo "warning: could not find shellcheck ... installing shellcheck" >&2
+        scversion="stable"
+        wget -qO- "https://storage.googleapis.com/shellcheck/shellcheck-${scversion?}.linux.x86_64.tar.xz" | tar -xJv
+        cp "shellcheck-${scversion}/shellcheck" $GOPATH/src/github.com/openshift/ocs-operator/tools/
```
obnoxxx:
```suggestion
        cp "shellcheck-${scversion}/shellcheck" "${GOPATH}"/src/github.com/openshift/ocs-operator/tools/
```
This is complained about by shellcheck

hack/shellcheck-test.sh
```diff
@@ -0,0 +1,59 @@
...
+        FILE=$GOPATH/src/github.com/openshift/ocs-operator/tools/
+        if [ -f "$FILE" ]; then
+                SHELLCHECK="${FILE}"
+        fi
+fi
+
+cd "${BASE_DIR}" || exit 2
+SCRIPTS=$(find $GOPATH/src/github.com/openshift/ocs-operator \( -path "*/vendor/*" -o -path "*/build/*" \) -prune -o -name *~ -prune -o -type f -exec grep -l -e '^#!/bin/bash$' {} \;)
```
obnoxxx:
```suggestion
SCRIPTS=$(find "${GOPATH}"/src/github.com/openshift/ocs-operator \( -path "*/vendor/*" -o -path "*/build/*" \) -prune -o -name "*~" -prune -o -type f -exec grep -l -e '^#!/bin/bash$' {} \;)
```

to fix shellcheck for this

hack/shellcheck-test.sh
```diff
@@ -0,0 +1,59 @@
...
+for script in ${SCRIPTS}; do
+        err=0
+        test_syntax "${script}"
+        [[ $? -ne 0 ]] && err=1
+        if [[ ${err} -ne 0 ]]; then
+                (((failed=err)))
+                echo "detected syntax issues in ${script}}" >&2
+        fi
+        test_shellcheck "${script}"
+        [[ $? -ne 0 ]] && err=1
```
obnoxxx: The use of `$failed`, `$err`, and `$?` here is kind of inconsistent. For instance if (only very theoretical) you get an syntax error, but not a shellcheck error, it will still list that there is a shellcheck error. And since you're not counting errors any more, the use of both failed and error is kind of superfluous.

suggestion:

```
failed=0
for script in ${SCRIPTS}; do
        test_syntax "${script}"
        if [[ $? -ne 0 ]]; then
                failed=1
                echo "${script}: detected syntax issues" >&2
        fi
        test_shellcheck "${script}"
        if [[ $? -ne 0 ]]; then
                failed=1
                echo "${script}: detected shellcheck issues in" >&2
        fi
done
```

If you want to issue a positive message for each script that has no errors, you could use the two vars, like this:

```
failed=0
for script in ${SCRIPTS}; do
        err=0
        test_syntax "${script}"
        if [[ $? -ne 0 ]]; then
                err=1
                echo "${script}: detected syntax issues" >&2
        fi
        test_shellcheck "${script}"
        if [[ $? -ne 0 ]]; then
                err=1
                echo "${script}: detected shellcheck issues in" >&2
        fi
        if [[ $err -ne 0 ]]; then
                ((failed=err))
        else
                echo "${script}: ok" >&2
        fi
done
```

hack/shellcheck-test.sh
```diff
@@ -0,0 +1,62 @@
...
+SCRIPT_DIR="$(cd "$(dirname "${0}")" && pwd)"
+
+BASE_DIR="$(cd "${SCRIPT_DIR}/.." && pwd)"
+
+if [[ -z "${SHELLCHECK}" ]]; then
+        echo "warning: could not find shellcheck ... installing shellcheck" >&2
+        scversion="stable"
+        wget -qO- "https://storage.googleapis.com/shellcheck/shellcheck-${scversion?}.linux.x86_64.tar.xz" | tar -xJv
+        cp "shellcheck-${scversion}/shellcheck" "$GOPATH"/src/github.com/openshift/ocs-operator/tools/
+        FILE="$GOPATH"/src/github.com/openshift/ocs-operator/tools/
```
obnoxxx: You still need this in order to run shellcheck if it is not installed, otherwise it will not run shellcheck (e.g. in the CI)

```suggestion
        FILE="$GOPATH"/src/github.com/openshift/ocs-operator/tools/shellcheck
```

hack/shellcheck-test.sh
```diff
@@ -0,0 +1,62 @@
...
+        if [ -f "$FILE" ]; then
+                SHELLCHECK="${FILE}"
+        fi
+fi
+
+cd "${BASE_DIR}" || exit 2
+SCRIPTS=$(find "$GOPATH"/src/github.com/openshift/ocs-operator \( -path "*/vendor/*" -o -path "*/build/*" \) -prune -o -name *~ -prune -o -type f -exec grep -l -e '^#!/bin/bash$' {} \;)
```
obnoxxx:
```suggestion
SCRIPTS=$(find "$GOPATH"/src/github.com/openshift/ocs-operator \( -path "*/vendor/*" -o -path "*/build/*" \) -prune -o -name "*~" -prune -o -type f -exec grep -l -e '^#!/bin/bash$' {} \;)
```

If you run shellcheck, it will complain about this. :-)

hack/shellcheck-test.sh
obnoxxx: this still needs to be done, not sure why it was resolved. :-)

hack/shellcheck-test.sh
obnoxxx: not sure why this was marked resolved - still needed :-)

obnoxxx:
/hold cancel

since the this now contains the test case for shellcheck

PR #345, opened 2019-11-28
Adding Testcase to check accuracy of the data

xenolinux:
Changed common.go file to add the pod Busybox specs into the file. 
~ ~ ~
[ The main purpose of adding pod specs in common.go  : 
Writing a test case in which a busybox pod will be deployed, PVC will be created. Once pod and pvc are created, the file will be generated in the busybox pod, after copying the generated file to the mount point, md5 of both the files will be calculated. Cross checking md5 will be helpful to know if the data populated is corrupted or not.]

### DIFF
--- a/functests/common.go
+++ b/functests/common.go
@@ -41,11 +41,8 @@ func WaitForJobSucceeded(k8sClient *kubernetes.Clientset, jobName string, jobNam
 			return nil
 		}
 		return fmt.Errorf("Waiting on job %s/%s to succeed when it is currently %d", jobName, jobNamespace, job.Status.Succeeded)
-<<<<<<< HEAD
-	}, 200*time.Second, 1*time.Second).Should(gomega.Succeed())
-=======
-	}, 200*time.Second, 1*time.Second).ShouldNot(gomega.HaveOccurred())
->>>>>>> 8e9e938f57bfc8e455a85e8df90b1f4e6da2c12b
+	},
+		200*time.Second, 1*time.Second).Should(gomega.Succeed())
 }
 
 // GetRandomPVC returns a pvc with a randomized name
@@ -100,15 +97,12 @@ func GetDataValidatorJob(pvc string) *k8sbatchv1.Job {
 								},
 							},
 							Command: []string{"/bin/sh", "-c"},
-<<<<<<< HEAD
 							Args: []string{"dd if=/dev/zero of=/tmp/random.img bs=512 count=1", //This command creates new file named random.img
 								"md5VAR1=$(md5sum /tmp/random.img | awk '{ print $1 }')",  //calculates md5sum of random.img in pod and stores it in a variable
 								"cp /tmp/random.img /data/random.img",                     //copies random.img file to pvc's mountpoint
 								"md5VAR2=$(md5sum /data/random.img | awk '{ print $1 }')", //calculates md5sum of file random.img
-								"if [[ \"$md5VAR1\" != \"$md5VAR2\" ]];then exit 1; fi"},  //compares the md5sum of random.img file
-=======
-							Args:    []string{"dd if=/dev/zero of=/tmp/random.img bs=512 count=1;md5VAR1=$(md5sum /tmp/random.img | awk '{ print $1 }');cp /tmp/random.img /data/random.img;md5VAR2=$(md5sum /data/random.img | awk '{ print $1 }');if [[ \"$md5VAR1\" != \"$md5VAR2\" ]];then exit 1; fi"},
->>>>>>> 8e9e938f57bfc8e455a85e8df90b1f4e6da2c12b
+								"if [[ \"$md5VAR1\" != \"$md5VAR2\" ]];then exit 1; fi",   //compares the md5sum of random.img file
+							},
 						},
 					},
 					Volumes: []k8sv1.Volume{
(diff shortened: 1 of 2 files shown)


functests/common.go
```diff
@@ -57,3 +73,44 @@ func GetRandomPVC(storageClass string, quantity string) *k8sv1.PersistentVolumeC
 
 	return pvc
 }
+
+func GetDataValidatorJob(pvc string) *k8sbatchv1.Job {
```
ashishranjan738: @xenolinux this function needs a comment starting with its name since this is an exported function.

xenolinux: Added the comment.

functests/common.go
```diff
@@ -57,3 +73,45 @@ func GetRandomPVC(storageClass string, quantity string) *k8sv1.PersistentVolumeC
...
+							Name:  randomName,
+							Image: "busybox",
+							VolumeMounts: []k8sv1.VolumeMount{
+								k8sv1.VolumeMount{
+									MountPath: "/data",
+									Name:      "volume-to-debug",
+								},
+							},
+							Command: []string{"/bin/sh", "-c"},
+							Args:    []string{"dd if=/dev/zero of=/tmp/random.img bs=512 count=1;md5VAR1=$(md5sum /tmp/random.img | awk '{ print $1 }');cp /tmp/random.img /data/random.img;md5VAR2=$(md5sum /data/random.img | awk '{ print $1 }');if [[ \"$md5VAR1\" != \"$md5VAR2\" ]];then exit 1; fi"},
```
rohantmp: Have you tested whether the `md5VAR=$(..)` expressions are evaluated before or after passing the string to `/bin/sh`?
If it is evaluated before, both those files won't exist and you'll get an empty md5sum for both `md5VAR1` and `md5VAR2` and the job will succeed regardless of outcome.

jarrpa: Subshell ordering should guarantee that they are executed first, yes. My main concern, however, is that this is a little hard to read.

@xenolinux could you define the script in a variable earlier in this function, using `; \` to separate the commands with newlines, and providing a comment with a short paragraph describing what this script is trying to accomplish?

functests/common.go
```diff
@@ -28,6 +29,21 @@ func WaitForPVCBound(k8sClient *kubernetes.Clientset, pvcName string, pvcNamespa
...
+		job, err := k8sClient.BatchV1().Jobs(jobNamespace).Get(jobName, metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+
+		if job.Status.Succeeded > 0 {
+			return nil
+		}
+		return fmt.Errorf("Waiting on job %s/%s to succeed when it is currently %d", jobName, jobNamespace, job.Status.Succeeded)
+	}, 200*time.Second, 1*time.Second).ShouldNot(gomega.HaveOccurred())
```
egafford: Should(Succeed()) would be a bit more readable here: ShouldNot(HaveOccurred()) scans better when you're referencing the error itself as a variable. ShouldNot(HaveOccurred() is definitely correct though.

functests/common.go
egafford: Agreed with @jarrpa: this is a bit dense.

xenolinux: Ack. I will work on it.

functests/common.go
xenolinux: Acknowledged!

functests/common.go
xenolinux: @rohantmp Those are evaluated both times - before and after.

xenolinux: @jarrpa Added new lines in the slice itself to make it readable with the comments saying what does the respective command do.

functests/common.go
```diff
@@ -96,7 +100,15 @@ func GetDataValidatorJob(pvc string) *k8sbatchv1.Job {
...
 							Command: []string{"/bin/sh", "-c"},
+<<<<<<< HEAD
+							Args: []string{"dd if=/dev/zero of=/tmp/random.img bs=512 count=1", //This command creates new file named random.img
+								"md5VAR1=$(md5sum /tmp/random.img | awk '{ print $1 }')",  //calculates md5sum of random.img in pod and stores it in a variable
+								"cp /tmp/random.img /data/random.img",                     //copies random.img file to pvc's mountpoint
+								"md5VAR2=$(md5sum /data/random.img | awk '{ print $1 }')", //calculates md5sum of file random.img
+								"if [[ \"$md5VAR1\" != \"$md5VAR2\" ]];then exit 1; fi"},  //compares the md5sum of random.img file
+=======
 							Args:    []string{"dd if=/dev/zero of=/tmp/random.img bs=512 count=1;md5VAR1=$(md5sum /tmp/random.img | awk '{ print $1 }');cp /tmp/random.img /data/random.img;md5VAR2=$(md5sum /data/random.img | awk '{ print $1 }');if [[ \"$md5VAR1\" != \"$md5VAR2\" ]];then exit 1; fi"},
+>>>>>>> 8e9e938f57bfc8e455a85e8df90b1f4e6da2c12b
```
rohantmp: Some leftover merge artifacts got commited

xenolinux: yep. let me correct it.

jarrpa:
Putting this on hold to reduce load on the CI cluster while #431 merges.

/hold

PR #348, opened 2019-11-28
Bump noobaa operator v2.0.9 + core v5.2.11

guymguym:
Thanks @kshlm for the help!

Due to #347 we couldn't update go.mod but there is no change in the API so will update once that issue is resolved.


kshlm:
/ok-to-test
/lgtm

Thanks @guymguym!

kshlm:
Keeping fingers crossed that #347 does not cause problems in ci!

PR #349, opened 2019-11-28
Bug 1778805: Backport 4.2 - Bump noobaa operator v2.0.9 + core v5.2.11

guymguym:
Backport of #348

Thanks @kshlm for the help!

Due to #347 we couldn't update go.mod but there is no change in the API so will update once that issue is resolved.


kshlm:
@guymguym is there a bz for this?
/ok-to-test

guymguym:
@nimrod-becker bz?

nimrod-becker:
@kshlm @obnoxxx 
Since the BZ was fixed with the merge into the noobaa-operator branch and the DS build already takes the new version. How do we handle this? We keep on hitting the same problem :)
The images and versions are mainly for consuming this repo and not for DS builds, can we remove the BZ requirement?

kshlm:
We don't want to remove the BZ restriction on this branch. Especially not now so close to GA. If we can't get a bz for this right now, then my preference is to hold this till we have another bz that needs to update the images and do it then.

kshlm:
If this is affecting any testers, then we should really encourage them to use the downstream builds instead. It's just a matter of changing one word in the registry image (`s/ocs-dev/rhceph-dev/`) for the catalogsource.

obnoxxx:
we can "misuse" an existing BZ to get this into ocs-operator... ;-)

@kshlm - what do you think?

PR #352, opened 2019-12-03
deploy: Disable csi-snapshotter in Rook deployment

kshlm:


deploy/olm-catalog/ocs-operator/0.0.2/ocs-operator.v0.0.2.clusterserviceversion.yaml
```diff
@@ -1016,6 +1016,8 @@ spec:
                   valueFrom:
                     fieldRef:
                       fieldPath: metadata.namespace
+                - name: CSI_ENABLE_SNAPSHOTTER
+                  value: "false"
                 image: rook/ceph:v1.1.4-27.gf20c056
```
Madhu-1: question: dont we need to change the image tag to pull rook changes?

kshlm: Yep. But there isn't a new rook image yet. So I let it be. The main requirement for this is downstream, where we have a different image with the fix.

travisn: The latest image from the release branch as seen [here](https://hub.docker.com/r/rook/ceph/tags/) is `rook/ceph:v1.1.7-14.g102543d`. So does this value only impact devs who might be testing on upstream?

kshlm: This value is essentially NOP, when testing upstream rook image currently used. When we update to a newer image, snapshotter will be disabled. This is mainly needed downstream, where there isn't a valid snapshotter container image for the release.

PR #359, opened 2019-12-05
Adding version string into OCS storage cluster

aruniiird:
This PR is to add version string for storage cluster, which UI can use and display...

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1777114


aruniiird:
@umangachapagain , can you please take a look into it...

pkg/apis/ocs/v1/storagecluster_types.go
```diff
@@ -87,6 +87,9 @@ type StorageClusterStatus struct {
 	// +optional
 	FailureDomain string `json:"failureDomain,omitempty"`
 
+	// StorageClusterVersion provides the version of the storage cluster
+	StorageClusterVersion string `json:"storageClusterVersion,omitempty"`
```
umangachapagain:
```suggestion
	Version string `json:"version,omitempty"`
```

aruniiird: Updated all the files... The CRDs should be updated as we run `generate-k8s` and `generate-openapi` commands, right?

pkg/controller/storagecluster/reconcile.go
```diff
@@ -95,6 +96,9 @@ func (r *ReconcileStorageCluster) Reconcile(request reconcile.Request) (reconcil
 	// If Update request is made and StorageCluster is PhaseIgnored, no need to
 	// proceed further
 	if instance.Status.Phase == "" {
+		if instance.Spec.Version == "" {
+			instance.Spec.Version = version.Version
```
guymguym: Hi, why is it under `Spec` and not under `Status`?

aruniiird: @guymguym , according to the following doc, status is the current state of the object and specification is the complete configuration of the desired state...

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status

Couple of projects which I referred also had the same patern,
https://github.com/nats-io/nats-operator/blob/master/pkg/apis/nats/v1alpha2/cluster.go
https://github.com/libopenstorage/operator/blob/master/pkg/apis/core/v1alpha1/storagecluster.go

aruniiird:
Screen shot...
image

Correcting the conflicts and check failures...

aruniiird:
> Can you add an `additionalPrinterColumn` [here](https://github.com/openshift/ocs-operator/blob/master/deploy/crds/ocs_v1_storagecluster_crd.yaml#L6) for version?
> 
> Everything else looks good.

@umangachapagain , added the required changes, please take a look...

jarrpa:
None of my concerns were addressed, so I won't review it again quite yet.

pkg/apis/ocs/v1/storagecluster_types.go
```diff
@@ -126,9 +128,11 @@ const (
 // StorageCluster is the Schema for the storageclusters API
 // +k8s:openapi-gen=true
 // +kubebuilder:subresource:status
+// +kubebuilder:subresource:spec
```
umangachapagain: `spec` is not a subresource. remove this line.

pkg/apis/ocs/v1/storagecluster_types.go
```diff
@@ -126,9 +128,11 @@ const (
...
 // +kubebuilder:printcolumn:name="Age",type=date,JSONPath=.metadata.creationTimestamp
 // +kubebuilder:printcolumn:name="Phase",type=string,JSONPath=.status.phase,description="Current Phase"
 // +kubebuilder:printcolumn:name="Created At",type=string,JSONPath=.metadata.creationTimestamp
+// +kubebuilder:printcolumn:name="Version",type=string,JSONPath=.spec.version,description="StorageCluster Version"
```
umangachapagain:
```suggestion
// +kubebuilder:printcolumn:name="Version",type=string,JSONPath=.spec.version,description="Storage Cluster Version"
```

umangachapagain:
> Looking at this again, I second Guy's concern.
>
Even I don't like having `Version` in `Spec`, but that is what OCP Console expects us to do. This is to populate the Version data in the UI (regarding which multiple bugs had been raised).
>
> In this case we are not SETTING the version of the StorageCluster, we are REPORTING the version of the StorageCluster. More accurately, we are reporting the version of the ocs-operator that last reconciled the StorageCluster. 
>
Yes. Storage Cluster version is same as Operator Version.
>
>As such, you shouldn't only check for an empty version, you should check to see if the current value of Version is less than the current version of the operator and update if so. Additionally, there should be a version check to make sure we don't try to reconcile a StorageCluster with a version greater than our operator's.
> 
This seems to be taken care of.

@jarrpa What other unaddressed concerns do you have?

Just so we are clear, the initial idea of this PR was to add version spec for UI to correctly populate data (even if it means empty data).
How this version is initialized by the operator or how we block reconciliation based on version wasn't the scope of this PR. If required we can discuss more on these issues and work on it in subsequent PRs. Let's keep changes to this PR minimal. WDYT?

pkg/apis/ocs/v1/storagecluster_types.go
aruniiird: Made the needed changes...

rohantmp:
> that is what OCP Console expects us to do
Can you explain what exactly requires this?

umangachapagain:
> > Even I don't like having Version in Spec, that is what OCP Console expects us to do
> 
> Can you explain how exactly they require this?

The Console code strictly looks for `.Spec.Version` to populate the version column in UI. If you don't have it, it'll show empty (which leads to multiple bugs raised).

PR #360, opened 2019-12-08
Fix #351 - exclude non-owned (required) CRDs manifests

### LINKED ISSUE #351: OB/OBC CRDs are not owned by OCS downstream (although it bundles them)
CC @nimrod-becker @jarrpa @davidvossel 

Here is where it uses `required:` for these CRDs in the OCS CSV instead of using `owned:`
https://github.com/openshift/ocs-operator/blob/34106927774f51584ca6fa3b6ef4e9dc559db39e/deploy/olm-catalog/ocs-operator/0.0.2/ocs-operator.v0.0.2.clusterserviceversion.yaml#L263-L302

This is because OCS CSV is copying the same `required` that NooBaa CSV is using upstream.

However for downstream these CRDs should be `owned` by OCS operator.

The reason it works is that the operator bundle includes those CRDs, which seems like a fishy situation, as OLM is not aware that OCS operator is the one owning those CRDs.

guymguym:
Fix #351 - exclude non-owned (required) CRDs manifests
For OB/OBC to be properly required from community-operators
See https://github.com/operator-framework/community-operators/pull/658


guymguym:
BTW I see this new warning that I printed:
```
WARNING: CRD is not owned and not required cephfilesystems.ceph.rook.io
```
So the CSV does not include this CRD but it is included in the installed manifests.
Is that on purpose?

guymguym:
Thanks @kshlm - I suspect that we might need to wait for https://github.com/operator-framework/community-operators/pull/658 to be published to the community-operators catalog so that OLM deployments will work...

PR #371, opened 2019-12-17
 Add support for 5 mons when there are sufficient nodes. 

rohantmp:


rohantmp:
Added unit-tests

pkg/controller/storagecluster/reconcile.go
```diff
@@ -37,10 +37,9 @@ const (
 	rookConfigData    = `[osd]
 osd_memory_target_cgroup_limit_ratio = 0.5
 `
+	monOverrideEnvVar = "MON_COUNT_OVERRIDE"
```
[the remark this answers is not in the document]
rohantmp: can change

pkg/controller/storagecluster/reconcile.go
```diff
@@ -931,3 +915,28 @@ func remove(slice []string, s string) (result []string) {
...
+
+func getMonCountOverrideEnvVar() string {
+	return os.Getenv(monOverrideEnvVar)
+}
+
+func getMonCount(nodeCount int) int {
+	count := defaults.MonCount
+
+	// return static value if overriden
+	override := getMonCountOverrideEnvVar()
```
jarrpa: I don't think it's worth it having a separate function for this. Just call os.Getenv here.

pkg/controller/storagecluster/reconcile.go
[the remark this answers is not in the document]
jarrpa: I disagree with the suggestion, adding `Key` seems redundant. I would rather see `monCountOverrideEnvVar`.

pkg/controller/defaults/defaults.go
```diff
@@ -16,6 +16,8 @@ const (
 var (
 	// MonCount is the default number of monitors to be configured for the CephCluster
 	MonCount = 3
+	// MonCountLarge is the number of monitors to be configured for the CephCluster whenever enough nodes are available
+	MonCountLarge = 5
```
jarrpa: I would recommend changing this one and above to be a more complementary pair, `MonCountMin` and `MonCountMax`.

rohantmp:
I've made the requested changes, but I want to hold on pushing them until I've sanity checked them in a cluster. My openshift-installer is failing now. Working to unblock.

rohantmp:
ci blocked by #378

rohantmp:
@jarrpa @ashishranjan738 PTAL again

rohantmp:
rebasing

pkg/controller/storagecluster/reconcile.go
```diff
@@ -931,3 +915,24 @@ func remove(slice []string, s string) (result []string) {
...
+		count, err := strconv.Atoi(override)
+		if err != nil {
+			log.Error(err, "could not decode env var %s", monOverrideEnvVar)
+		} else {
+			return count
+		}
+	}
+
+	if nodeCount >= defaults.MonCountMax {
+		count = defaults.MonCountMax
```
egafford: If I read this correctly, it will result in four mons for a 4-node Ceph cluster, which will be more error-prone than 3 mons for a 4-node Ceph cluster. (With 4 mons, if any two nodes go down the cluster can't quorate. With 3, if two nodes go down but one of them is the node without a mon, which it will be 50% of the time, you've still got quorum.) 

Am I missing something that prevents this case? If not, can we add some logic with a set of workable mon counts (in our case, 3 and 5) and pick the largest one of those for which we have room? If for some reason the user really *wants* 4 mons, we should let the monCountOverrideEnvVar be authoritative, but most users will want an odd mon count.

rohantmp: Yeah, the vars are bit of a misnomer actually. We only set odd counts, so currently only 3 and 5 are used.

rohantmp:
Investigating a potential rook issue which blocks this: mon count doesn't effectively shrink when mon quroum isn't full

rohantmp:
@egafford PTAL

pkg/controller/storagecluster/reconcile.go
egafford: Ah, I see. The logic is a bit odd, but it works so long as we're only allowing min and max. Until that changes, no need to complicate things. Thanks Rohan.

PR #378, opened 2019-12-22
Fixing the source-manifests.sh script

aruniiird:
This fixes the broken make file commands, like 'gen-latest-csv',
'source-manifests' etc

Fixes: https://github.com/openshift/ocs-operator/issues/377

### DIFF
--- a/hack/source-manifests.sh
+++ b/hack/source-manifests.sh
@@ -80,7 +80,7 @@ gen_args="olm-catalog gen-csv --csv-version=$TMP_CSV_VERSION"
 if [ -n "$CSV_REPLACES_VERSION" ]; then
 	gen_args="$gen_args --from-version=$CSV_REPLACES_VERSION"
 fi
-$OPERATOR_SDK "$gen_args" 2>/dev/null
+$OPERATOR_SDK $gen_args 2>/dev/null
 OCS_TMP_CSV="deploy/olm-catalog/ocs-operator/${TMP_CSV_VERSION}/ocs-operator.v${TMP_CSV_VERSION}.clusterserviceversion.yaml"
 mv $OCS_TMP_CSV $OCS_CSV
 # Make variables templated for csv-merger tool


hack/source-manifests.sh
```diff
@@ -80,7 +80,7 @@ gen_args="olm-catalog gen-csv --csv-version=$TMP_CSV_VERSION"
 if [ -n "$CSV_REPLACES_VERSION" ]; then
 	gen_args="$gen_args --from-version=$CSV_REPLACES_VERSION"
 fi
-$OPERATOR_SDK "$gen_args" 2>/dev/null
+$OPERATOR_SDK $gen_arg" 2>/dev/null
```
kshlm: Sorry, but what exactly is this supposed to fix?  From what I see, this change actually breaks the operator-sdk command being run here.

kshlm: If you suspect this line is failing, first try removing the output redirection to null `2>/dev/null` and see what the actual error is.

aruniiird: Corrected the variable name (accidentally deleted 's' instead of the double quotes)...

Here is the error which is printed (ran with `bash -x` ),
```
+ rm -f /tmp/tmp.A8eUg7yz7s
+ TMP_CSV_VERSION=9999.9999.9999
+ gen_args='olm-catalog gen-csv --csv-version=9999.9999.9999'
+ '[' -n '' ']'
+ build/_output/tools/operator-sdk-v0.8.2-x86_64-linux-gnu 'olm-catalog gen-csv --csv-version=9999.9999.9999'
Error: unknown command "olm-catalog gen-csv --csv-version=9999.9999.9999" for "operator-sdk"
Run 'operator-sdk --help' for usage.
```

the `operator-sdk` is taking the whole `$gen_args` string as a single argument...

hack/source-manifests.sh
```diff
@@ -80,7 +80,7 @@ gen_args="olm-catalog gen-csv --csv-version=$TMP_CSV_VERSION"
 if [ -n "$CSV_REPLACES_VERSION" ]; then
 	gen_args="$gen_args --from-version=$CSV_REPLACES_VERSION"
 fi
-$OPERATOR_SDK "$gen_args" 2>/dev/null
+$OPERATOR_SDK $gen_args 2>/dev/null
```
rohantmp: Can we also remove the `2>/dev/null` and put `echo "Generating intermediate csv: Ignore warnings about missing fields spec.keywords, spec.maintainers, spec.provider"` before the line so it's not necessary?

aruniiird: Agree with you @rohantmp ... We could even exit (rather than continuing) if the command fails...
@kshlm thoughts?

rohantmp: `set -e` at the beginning of the script already ensures that any failed commands will cause the script to exit :)

PR #379, opened 2019-12-31
Changing the 'Description' of publically available storage providers

aruniiird:
Adding a more meaningful description to the publically available storage
providers

bug: https://bugzilla.redhat.com/show_bug.cgi?id=1760387


mykaul:
Not sure if 'MCG' shouldn't be replaced with 'Multi-Cloud Gateway'.
(Also, upstream, I would just use Noobaa..)

@nimrod-becker  - thoughts?

tools/csv-merger/csv-merger.go
```diff
@@ -456,7 +456,9 @@ func generateUnifiedCSV() *csvv1.ClusterServiceVersion {
 			ocsCSV.Spec.CustomResourceDefinitions.Owned[i].Description = internalCRDDescription + "StorageCluster Initialization represents a set of tasks the OCS operator wants to implement for every StorageCluster it encounters."
 		// backingstore and bucketclass can be used by the admin, so avoid adding internal prefix to these resources
 		case "backingstores.noobaa.io":
+			ocsCSV.Spec.CustomResourceDefinitions.Owned[i].Description = "Backing Stores are storage targets that are used to store chunks of data on MCG buckets."
```
nimrod-becker: Got an input from PM/Marketing, we should not use acronyms 
```suggestion
			ocsCSV.Spec.CustomResourceDefinitions.Owned[i].Description = "Backing Stores are storage targets that are used to store chunks of data on object buckets."
```

tools/csv-merger/csv-merger.go
```diff
@@ -456,7 +456,9 @@ func generateUnifiedCSV() *csvv1.ClusterServiceVersion {
...
 		case "bucketclasses.noobaa.io":
+			ocsCSV.Spec.CustomResourceDefinitions.Owned[i].Description = "BucketClass is a CRD representing a class for buckets that defines tiering policies and data placements for an OBC on MCG buckets."
```
nimrod-becker: Got an input from PM/Marketing, we should not use acronyms

```suggestion
			ocsCSV.Spec.CustomResourceDefinitions.Owned[i].Description = "BucketClass is a CRD representing a class for buckets that defines tiering policies and data placements for object buckets."
```

tools/csv-merger/csv-merger.go
aruniiird: Changed the description according to the above comment...

tools/csv-merger/csv-merger.go
aruniiird: Changed the description according to the above comment...

aruniiird:
Here is the screen shot,
image

mykaul:
> Here is the screen shot,
> image

Ganesha, not ganesha (capitalize G)

aruniiird:
Changed according to the latest comments,

image

aruniiird:
Made the required changes,

image

aruniiird:
@obnoxxx ,
> There are conflicts. This at least needs a run of `make gen-latest-csv`.

I have made the changes needed... Please take a look

aruniiird:
@umangachapagain , @obnoxxx , let me know any further changes required...

aruniiird:
> I don't like this PR. Changing the StorageCluster description to read like a product description is disingenuous and technically incorrect. Overriding the Noobaa field descriptions shouldn't be done a tthe ocs operator level, it should be done in the NooBaa CSV. Similarly, the change to the Ceph NFS CRD should happen in Rook, not here.

Hi @jarrpa , changing description is already placed in the OCS Operator code. This PR merely changes the description and doesn't add any new feature.

What I understood from the existing code is, we want a different descriptions to be placed (instead of their default ones) which is much more meaningful in an OCP-OCS environment (rather than the individual component level). This way we could change it (description) into a more customer friendly descriptions for any of our components.

jarrpa:
Just revisiting this to put it to a close: My issue wasn't that the description was being changed, it's the content of the proposed change that was not appropriate. We should not be editing the CRD descriptions to look like a product overview. As such, I'll be closing this PR, as I believe the NooBaa-related changes were already taken care of elsewhere.

PR #385, opened 2020-01-08
Annotating the internal CRDs

aruniiird:
Backend change for adding 'operators.operatorframework.io/internal-objects' annotation,
which holds an array of CRD names which are internal to the operator. This will be later
used by the UI code for hiding the internal tiles/cards.


umangachapagain:
/hold

There is no UI support as of now to validate this fix. Let's wait till we have that.

aruniiird:
Bipul has put a PR in the console front, which works with this PR as expected... Taken a screen shot,

Bipuls PR: https://github.com/openshift/console/pull/3877

image

umangachapagain:
/lgtm

Can you check why the test is failing?

Issue #386, opened 2020-01-09
Negative Path Testing: Functional tests for ocs-operator pod killed during StorageCluster creation


brainfunked:
I'm working on creating a suite of negative path or failure scenario tests for ocs-operator. The first implementation is for testing out whether the StorageCluster creation operation resumes properly when the ocs-operator pod is killed and is respawned during the operation. I'm starting off with tests that are 'deterministic'. To start with, I'll be monitoring the various stages of the StorageCluster creation operation: ceph cluster, nooba etc. For each stage, I'll kill the ocs-operator pod and see if the operation resumes and completes properly as a new pod is spawned.

PR #387, opened 2020-01-09
Negative path test: kill ocs-operator pod during cluster creation

brainfunked:
Working on #386


brainfunked:
This and other tests to follow require that the storage cluster not be created during the BeforeSuite and AfterSuite blocks. Since the test suite as a whole intends to break the cluster in various ways, I figured it made sense to make a separate negativetests package, rather than reusing the functests.

brainfunked:
I've pushed the commands (https://github.com/openshift/ocs-operator/pull/387/commits/679b96edd9ebb0904caa3d18216b880ef6e6dc75) I run to simulate the scenario manually.

To make the scenario fully automated, the code in those commands needs to be integrated into the ginkgo tests. I'll now work on the following:

1. Change the informer/watcher to monitor the ceph cluster status instead of the ocs-operator deployment. The idea is to expand the test such that the operator pod is killed at each of the stages of the StorageCluster creation. I need to learn how to monitor custom resources.
2. Merge the killpod code into the informer/watcher code to be run as a goroutine in response to the necessary condition being detected by the informer/watcher.
3. Integrate the combined code into the ginkgo test. At this point, it would probably make sense to use `wait.PollImmediate()` or equivalent to check for the appropriate status on the ceph cluster in a goroutine that's launched just before `deploymanager.StartDefaultStorageCluster()` is invoked. `killpod` functionality can be invoked as soon as the `wait.PollImmediate()` function returns.

brainfunked:
Changes pushed in dc1da3ec34fc4d8ec58de7eaf3a997f094ced405 run the failure scenario in a sequential manner. I tried out multiple approaches; including but not limited to
* running the informer code as a goroutine,
* moving the killpod functionality as a library function into the negativetests package and invoking it from the informer goroutine,
* using channels to communicate with the goroutines for signaling and termination etc.,
* using `wait.Poll` to monitor whether the operator deployment comes up.

Ran into various issues with these approaches; the biggest being that the goroutines didn't run properly or at all. The log messages written to `GinkgoWriter` from within the goroutine were not visible. The ginkgo documentation wasn't much of a help except to point to using gomega's `Eventually`. So, eventually, I modified all the code to use.. `Eventually`. As it turns out, the serialised approach that I've taken in the commit dc1da3ec34fc4d8ec58de7eaf3a997f094ced405 works out because the cluster setup happens in the beginning with the deploymanager.

Long term, to be able to add more negative scenarios which induce failures during cluster creation, it might be necessary to split up the `DeployManager.StartDefaultStorageCluster()` functionality such the the `DeployManager.WaitOnStorageCluster()` can be invoked manually in the tests. Perhaps, a fire-and-forget version for creating the storage cluster, without waiting for it to be created could be added. I think it would be necessary to add this functionality before adding more negative scenarios.

Anyway, for now, the code can be reviewed.

jarrpa:
I think this needs a complete rework given recent changes to the functional tests. And given the age, I'm going to close this PR for now. I still like the idea a lot, so please either update this PR and reopen it or open a new one.

PR #390, opened 2020-01-14
[WIP] Design doc for CRD Validation in OCS Operator Project

aruniiird:
This is an initial design draft to bring CRD Validations into
OCS Operator project.

This design and future implementation try to address issue: https://github.com/openshift/ocs-operator/issues/388


mhrivnak:
It would be useful to include a discussion of what specific validation you need to do, how much of it could be done with openapiv3, and how much of it you could mitigate from a controller by either 1) reporting an invalid state through Status or warning events, or 2) just fixing resource specs so they have the correct values. It's possible you could make a substantial usability improvement while waiting for the opportunity to implement a full validation webhook.

jarrpa:
@mhrivnak could you go into a bit more detail? I myself am ignorant as to what openapiv3 is, and would like to know the who and when for #2 above.

mhrivnak:
> @mhrivnak could you go into a bit more detail? I myself am ignorant as to what openapiv3 is, and would like to know the who and when for #2 above.

This explains it well: https://blog.openshift.com/openshift-4-2-declarative-dynamic-ui-for-your-operator/

In terms of option 2, sometimes controllers enforce state just through reconciliation, if they know what the state is supposed to be. For example, perhaps you require that if a CR has a particular value in its spec, then it must also have a label with the same value. If someone tries to create such a resource without the label, rather than call that invalid, your could have your controller just add the label. Obviously this approach can't be used in all cases and is not a substitute for validation, but it can be a useful tool to deal with certain kinds of situations that might be thought of as validation issues.

docs/design/crd-validation/README.md
```diff
@@ -0,0 +1,109 @@
+## Background
+
+This proposal is to provide CRD level validation for OCS Operator project. Currently, when the user (manually) changes CRD yaml file, the cluster will fail at the creation time if there is an error in the user input. It would have been nice to have a layer of validation when user submit the changes.
```
egafford: Validation is always good, but what sorts of validations do you think we could reasonably provide here? If the user changes a CRD, they're already clearly doing something we don't anticipate. Assuming that the semantics of a change they make is incorrect seems difficult, especially before we receive their actual CR (which is your intention if I read this doc correctly).

I definitely see that modifying CRDs is error-prone and that's unfortunate. Still, this document spends a lot of time on the "I want" clause of the "As a... I want... so that" user story. Who wants this feature? What concrete benefit (with examples) can we give them?

docs/design/crd-validation/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+#### Goals
+
+Our main goals are,
+
+- Minimal changes to the existing OCS Operator project
+    - there should not be any (no or very minimal) change to the existing code
+- Easy to add / remove / modify validations
+    - other developers should be able to modify validations
+
```
egafford: Some examples here of validations we think might help the user would be great.

docs/design/crd-validation/README.md
```diff
@@ -0,0 +1,109 @@
...
+6. The name of the front-end service.
+7. The webhook URL used for admission requests.
+8. A PEM-encoded CA certificate that signs the server certificate used by the webhook server.
+9. Rules that define when the API server should use this controller.
+10. The operation that triggers the API server to call this controller are,
+
+    - create
+    - update
+    - delete
+    - connect
```
egafford: This is a moderate amount of complexity. Why do we think creating this feature with this mechanism will be less error-prone than allowing the user to make their own changes?

docs/design/crd-validation/README.md
umangachapagain:
```suggestion
This proposal is to provide validation for OCS Operator CRs. Currently, when the user Creates/Updates the CR, the reconciliation might fail if there is an error in the user input. It will be a better user experience if we validate the request right when user submits the changes and before the CR gets written into etcd.
```

umangachapagain:
```suggestion
This proposal is to provide validation for OCS Operator CRs. Currently, when the user Creates/Updates the CR, the reconciliation might fail if there is an error in the user input. It will be a better user experience if we validate the request right when user submits the changes and before the CR gets written into etcd.
```

aruniiird:
Hi @mhrivnak ,

> It would be useful to include a discussion of what specific validation you need to do, 

Listing a few bugs which we discussed (thanks @umangachapagain),

a. [Implement OCP Version Check and Block OCS installation and upgrade if that break compatibility with OCP](https://issues.redhat.com/browse/RHSTOR-705)

b. [Block upgrade if there's a risk of running out of space (Ceph pool is nearfull)](https://bugzilla.redhat.com/show_bug.cgi?id=1783753)

c. Changing CR to use Ceph Based Storage Class for expansion should not be allowed

> how much of it could be done with openapiv3, and how much of it you could mitigate from a controller by either 1) reporting an invalid state through Status or warning events, or 2) just fixing resource specs so they have the correct values. It's possible you could make a substantial usability improvement while waiting for the opportunity to implement a full validation webhook.

Going through the openapiv3 link (you posted), some of the simple and direct integer / string checks can be done directly through it. So in the above list, 'a.'  can be done through v3 CRD checks, but 'b.' needs some additional logic (to check max pool size etc). So in order to answer your questions, currently we don't have an exhaustive list of validations to be performed and I am quite sure we have couple of validations which can done through OpenAPI v3 mode. Like you said, we can definitely get an advantage of using them until a full fledged validation webhooks are in place.

aruniiird:
After today's meeting with @jarrpa and team team, need some changes to the initial validation design doc... Will be adding those discussion details into doc... So changing the title back to [WIP] mode.

docs/design/crd-validation/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+This proposal is to provide validation for OCS Operator CRs. Currently, the user Creates/Updates the CR, the reconciliation might fail when there is an error in the user input. It will be a better user experience if we validate the request right after user submits the changes and before the CR gets written into etcd.
+
+## User Senarios
+
+Listing out some user cases, where validators will be helpful,
+
+a. **Expanding the cluster / Adding more capacity**
+
+From backend, validator could check the existing resources and accordingly validate the request. In future, it could check for different storage platforms (like AWS, Azure, VMWares', BareMetal etc) and further advance the validation checks.
```
jarrpa: If I'm understanding this correctly, this doesn't sound like a valid thing to validate, and would be almost impossible to guarantee across multiple platforms. The Kubernetes model of eventually consistency would be to let the request for additional devices go through and retry until it succeeds.

jarrpa:
I hate to say it, but this feels like a feature looking for a problem to solve. I admit it sounds like it'd be cool to implement, but I'm still not seeing a valid use case for it. For example, I could see an argument for preventing a *reduction* in storage (since Rook-Ceph doesn't currently support this) but we should still allow for a reduction as a potential rollback to a bad expansion.

jarrpa:
@tnielsen Do you have any thoughts as to what we could use such a feature for? Note that we could leverage an admission controller in the ocs-operator for CRs other than StorageClusters, like a CephCluster, for restrictions that are specific to the OCS use case.

docs/design/crd-validation/README.md
aruniiird: @egafford , 
Many of the initial bugs, according to which we started the validation project, has become invalid... One such use case we see is depicted in the following bug, can you please take a look,

[bug#1777073](https://bugzilla.redhat.com/show_bug.cgi?id=1777073)

travisn:
> @travisn Do you have any thoughts as to what we could use such a feature for? Note that we could leverage an admission controller in the ocs-operator for CRs other than StorageClusters, like a CephCluster, for restrictions that are specific to the OCS use case.

Implementing an admission controller is going to provide a much better error-handling experience for invalid CR settings. The static checking by openapiv3 can only get so far. In upstream rook, implementing an admission controller has come up on several occasions, but has not been completed yet. See https://github.com/rook/rook/issues/2363

As with any feature, the complexity of implementation needs to be reasonable compared to the benefits of the feature. In the case of OCS it may not make sense to add an admission controller because the UI can perform some checks for important error conditions before updating the CRs. OCS doesn't really need to worry about invalid CRs being generated, right? Or at least it's not directly supported. 

While for upstream Rook, it's fully supported to modify the CRs directly and the CRs can be quite complex, so it would make a lot of sense to implement an admission controller in Rook.

travisn:
Opened https://github.com/rook/rook/issues/4819 for ideas for a Rook admission controller

a2batic:
Thanks for working on this feature. 
For Snapshot Scheduling(https://issues.redhat.com/browse/RHSTOR-874) feature, this validation can be used to validate CR (https://backube.github.io/snapscheduler/usage.html) in the following places:-
1. Verifying the schedule string is valid 
2. Verifying the parts of schedule string are in limits ie. `minutes hours day-of-month month day-of-week` should have valid values

docs/design/crd-validation/README.md
egafford: All right; I see the case and I appreciate the added detail on the benefit. It seems as though there is ongoing discussion / disagreement about why this feature is needed, but you've addressed the request for changes.

aruniiird:
Closing this PR as a fresh design approach will be started soon.

PR #393, opened 2020-01-21
OCS Independent Mode

umangachapagain:
+  adds spec to enable External Storage Cluster
    ```
    externalStorage:
        enable: true
    ``` 
    External Storage Cluster is optional and defaults to false.
    When set to true, OCS will connect to an external OCS Storage Cluster instead of provisioning one locally.

+  adds an annotation to CSV indicating support for externalStorage 
    + `external.cluster.ocs.openshift.io/supported:"true"`

+  `$ oc get storagecluster`
```bash
NAME                      AGE   PHASE     EXTERNAL   CREATED AT
empty-storagecluster      40m   Ignored              2020-02-17T11:36:26Z
external-storagecluster   62m             true       2020-02-17T11:14:25Z
```


guymguym:
Don’t you need a network address to connect to and perhaps also a secret from that system?

You can refer to how noobaa spec added support for it using a ’joinSecret’ in https://github.com/noobaa/noobaa-operator/pull/203/files

umangachapagain:
> Don’t you need a network address to connect to and perhaps also a secret from that system?
> 
> You can refer to how noobaa spec added support for it using a ’joinSecret’ in https://github.com/noobaa/noobaa-operator/pull/203/files

Thanks for the pointer.

We are currently looking at different options. Will update here once we have something solid.

umangachapagain:
ci/prow/ocs-operator-e2e-aws needs to be updated to accommodate external clusters. 
Until then we need to bypass this.

aruniiird:
@jarrpa , @leseb , @travisn , please take look...

PR #400, opened 2020-01-29
fix: tear down failure in CI

ashishranjan738:
This commit fixes teardown failure in CI by adding an additional step to clean up storageCluster before deleting the namespace.

Signed-off-by: Ashish Ranjan <aranjan@redhat.com>

### DIFF
--- a/pkg/deploy-manager/common.go
+++ b/pkg/deploy-manager/common.go
@@ -6,6 +6,7 @@ import (
 	k8sv1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	types "k8s.io/apimachinery/pkg/types"
 	utilwait "k8s.io/apimachinery/pkg/util/wait"
 )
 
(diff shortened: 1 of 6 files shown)


ashishranjan738:
> Overall looks good, but why is there a rook client being added as part of this PR?

I will split it into a different commit. We need the rook client for watching the cephcluster state.

pkg/deploy-manager/common.go
```diff
@@ -27,6 +27,35 @@ func (t *DeployManager) CreateNamespace(namespace string) error {
...
+	err := t.deleteStorageCluster()
+	if err != nil && !errors.IsNotFound(err) {
+		return err
+	}
+
+	timeout := 300 * time.Second
+	interval := 10 * time.Second
+
+	// Wait for storagecluster and cephCluster to terminate
+	err = utilwait.PollImmediate(interval, timeout, func() (done bool, err error) {
```
umangachapagain: Technically, CephCluster will/should be deleted before StorageCluster due to OwnerRef. So, it makes more sense to check for deletion of CephCluster first and then for StorageCluster.

pkg/deploy-manager/storagecluster.go
```diff
@@ -181,6 +181,21 @@ func (t *DeployManager) createStorageCluster() (*ocsv1.StorageCluster, error) {
 	return newSc, nil
 }
 
+// deleteStorageCluster is used to delete the test suite storage cluster
+func (t *DeployManager) deleteStorageCluster() error {
```
umangachapagain: Is deleting the StorageCluster enough here? I think we should take care of the CephCluster finalizer so that deletion is faster.

pkg/deploy-manager/config.go
```diff
@@ -100,6 +107,23 @@ func NewDeployManager() (*DeployManager, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	rookConfig.GroupVersion = &ocsv1.SchemeGroupVersion
+	rookConfig.NegotiatedSerializer = serializer.DirectCodecFactory{CodecFactory: codecs}
+	rookConfig.APIPath = "/apis"
+	rookConfig.ContentType = runtime.ContentTypeJSON
+	if rookConfig.UserAgent == "" {
+		rookConfig.UserAgent = rest.DefaultKubernetesUserAgent()
+	}
```
umangachapagain: I don't think this will be necessary as rook/ceph client pkg takes care of this.

pkg/deploy-manager/common.go
ashishranjan738: since its running inside polling function. I don't think the sequence matter. The idea was to just check for any remaining of cephCluster or StorageCluster. Anyway i moved cephCluster check before storageCluster.

pkg/deploy-manager/storagecluster.go
ashishranjan738: hmm, when I tested it on my local cluster the tests were passing without the removal of finalizers. But removing the finalizer will be a good idea. Added that.

pkg/deploy-manager/config.go
ashishranjan738: yeah makes sense. Removed this part.

umangachapagain:
Can we rerun the test once again, just to be sure? 


/shrug

ashishranjan738:
> https://github.com/openshift/ocs-operator/blob/master/pkg/deploy-manager/subscription.go#L349:25 already takes care of deleting the OCS Cluster.
> 
> I'd suggest making the required changes inside that function.
> Also, if we just do` t.UninstallOCS()` before `t.DeleteNamespaceAndWait()`, we should be good.



> Can we rerun the test once again, just to be sure?
yeah

umangachapagain:
Wouldn't it be better to separate StorageCluster deletion from Namespace deletion?

/unshrug

ashishranjan738:
> Can we rerun the test once again, just to be sure?

yeah /retest

PR #401, opened 2020-01-30
Bug 1795532 - throttle background ceph activity on gp2 devices

umangachapagain:
- adds `TuneSlowDeviceClass` config to `StorageDeviceSet.Config`
This config is enabled when Cluster is backed by `gp2` or `io1`
StorageClass. In case of error reading the StorageClass, it
is disabled.


umangachapagain:
depends on https://github.com/rook/rook/pull/4779

/hold

umangachapagain:
/assign @leseb @travisn @jarrpa 

Do we need to expose the setting to the user? Would they want to disable this on `gp2`?
Currently, StorageCluster internally passes this value to the CephCluster. User will not see it.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -863,6 +881,22 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster) []rook.StorageClassDevi
 	return storageClassDeviceSets
 }
 
+func (r *ReconcileStorageCluster) isStorageClassProvisionerType(storageClassname, provisioner string, parameterTypes []string) (bool, error) {
+	storageClass := &storagev1.StorageClass{}
+	err := r.client.Get(context.TODO(), types.NamespacedName{Namespace: "", Name: storageClassname}, storageClass)
+	if err != nil {
+		return false, fmt.Errorf("failed to retrieve Storage Class. %+v", err)
```
leseb:
```suggestion
		return false, fmt.Errorf("failed to retrieve Storage Class %q. %+v", storageClassname, err)
```

pkg/controller/storagecluster/reconcile.go
```diff
@@ -863,6 +881,22 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster) []rook.StorageClassDevi
 	return storageClassDeviceSets
 }
 
+func (r *ReconcileStorageCluster) isStorageClassProvisionerType(storageClassname, provisioner string, parameterTypes []string) (bool, error) {
+	storageClass := &storagev1.StorageClass{}
+	err := r.client.Get(context.TODO(), types.NamespacedName{Namespace: "", Name: storageClassname}, storageClass)
+	if err != nil {
+		return false, fmt.Errorf("failed to retrieve Storage Class %q. %+v", storageClassname, err)
+	}
+	if storageClass.Provisioner == provisioner {
```
jarrpa: Instead of an `if`, use a `select` statement with one `case` for aws-ebs. Specify the throttled disk types in a const and reference them there.

obnoxxx:
@umangachapagain writes:
> /hold depends on #414 for version upgrade

Do you mean that this PR does now not upgrade the rook version any more and you would like to merge #414 first?

umangachapagain:
> @umangachapagain writes:
> 
> > /hold depends on #414 for version upgrade
> 
> Do you mean that this PR does now not upgrade the rook version any more and you would like to merge #414 first?

Yes.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -590,6 +597,21 @@ func (r *ReconcileStorageCluster) ensureCephConfig(sc *ocsv1.StorageCluster, req
 // ensureCephCluster ensures that a CephCluster resource exists with its Spec in
 // the desired state.
 func (r *ReconcileStorageCluster) ensureCephCluster(sc *ocsv1.StorageCluster, reqLogger logr.Logger) error {
+	// if StorageClass is "gp2" or "io1" based, set tuneSlowDeviceClass to true
+	// this is for performance optimization of slow device class
+	//TODO: If for a StorageDeviceSet there is a separate metadata pvc template, check for StorageClass of data pvc template only
+	for i, ds := range sc.Spec.StorageDeviceSets {
+		throttle, err := r.throttleStorageDevices(*ds.DataPVCTemplate.Spec.StorageClassName)
```
obnoxxx: I think for correctness, this should be done on a per-OSD-PV basis, not for the DataPVCTemplate: In theory (even though it would not be recommended), we could add completely different types of disks when we expand the storage, right? (This is what I have also been criticizing - I think we should offer a choice of the storage class upon initial creation, and then lock the storage class for future expansion of the created pool...)

pkg/controller/storagecluster/reconcile.go
```diff
@@ -51,6 +56,8 @@ var validTopologyLabelKeys = []string{
 	"topology.rook.io",
 }
 
+var throttleDiskTypes = []string{"gp2", "io1"}
```
obnoxxx: I can't say that I like hard coding two device types of one provisioner for throttling. This is not very robust, and not expandable without code changes... Can we have a configuration input (with some defaults) that specifies provisioners and disk types that we want to be throttled?

obnoxxx:
@umangachapagain btw you can rebase a PR on top of a dependent PR if you want to see better test results...

pkg/controller/storagecluster/reconcile.go
umangachapagain: As per the discussion on BZ, it seemed like throttling these two were good enough. I could add a `slowDeviceClasses` config that could used to determine which devices to throttle in combination with `tuneSlowDeviceClass`. But, is it really required/worth?

@jarrpa @leseb Thoughts?

pkg/controller/storagecluster/reconcile.go
umangachapagain: I not sure I understand the whole `StorageClassDeviceSet` design completely.
But, based on current implementation, `DataPVCTemplate` can have only one `PersistentVolumeClaimTemplate` which means a Device Set can have only similar devices (based on mentioned `Storage Class`). And this information is enough to decide on throttling (which is eventually handled by rook implementation).

We could have a mix of devices (currently possible only by having different DeviceSets) and separation of Metadata and Data PVCs which is why leaving a TODO comment to pick up once we have all required things based on new implementation.

@jarrpa and I had a long conversation on this and he approved of these changes.

PR #402, opened 2020-02-03
functest: increases the timeout for cluster teardown

ashishranjan738:
This commit increases the timeout for cluster teardown because pods are taking more than 200 seconds to terminate which causes teardown to get a timeout.

Signed-off-by: Ashish Ranjan <aranjan@redhat.com>

### DIFF
--- a/pkg/deploy-manager/common.go
+++ b/pkg/deploy-manager/common.go
@@ -45,7 +45,7 @@ func (t *DeployManager) DeleteStorageClusterAndWait(namespace string) error {
 		}
 	}
 
-	timeout := 300 * time.Second
+	timeout := 600 * time.Second
 	interval := 10 * time.Second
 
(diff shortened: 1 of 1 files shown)


jarrpa:
Why are we doing this? Please provide an explanation in the commit message. Do we really want customers to be kept waiting for 5 minutes before we abort?

jarrpa:
Please provide an explanation in the commit message as to why we're doing this.

ashishranjan738:
> Please provide an explanation in the commit message as to why we're doing this.

updated the commit message. The reason is that the pods are taking more than 200 seconds to get terminated, specifically the noobaa pod. This can be observed from the following logs https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_ocs-operator/398/pull-ci-openshift-ocs-operator-master-ocs-operator-e2e-aws/792/build-log.txt

PR #404, opened 2020-02-04
hack: Follow up of Cleanup and Consolidating of common variables

nizamial09:
Current common.sh script won't detect the editings done in the Makefile variable like REGISTRY_NAMESPACE,IMAGE_TAG,IMAGE_REGISTRY when we run `make ocs-registry`.
This commit will help fetch the Makefile variable value and use it as DEFAULT for the above mentioned variables.

Signed-off-by: Nizamudeen <nia@redhat.com>


hack/common.sh
```diff
@@ -36,9 +37,9 @@ NOOBAA_CSV="$OUTDIR_TEMPLATES/noobaa-csv.yaml"
 ROOK_CSV="$OUTDIR_TEMPLATES/rook-csv.yaml.in"
 OCS_CSV="$OUTDIR_TEMPLATES/ocs-operator.csv.yaml.in"
 
-DEFAULT_IMAGE_REGISTRY="quay.io"
-DEFAULT_REGISTRY_NAMESPACE="ocs-dev"
-DEFAULT_IMAGE_TAG="latest"
+DEFAULT_IMAGE_REGISTRY=$(grep -m 1 IMAGE_REGISTRY $MAKEFILE | sed -e 's/^.*= "//g;s/"//g')
```
ashishranjan738: Instead of doing a grep, I guess passing this information via env variable would be better.

nizamial09: I am getting this info from the Makefile by reading through the lines and providing necessary regex to eliminate some garbages. How can I do the same via an environment variable? @ashishranjan738

nizamial09:
> The code changes look solid, but the commit message needs some love. Please change the subject of the commit to something like `hack: Consolidate Makefile variables`.

Sure. I can do that.

nizamial09:
There was a concern that this will completely eliminate the use of Makefile in the sense that there is no enivronmental variables in that file. Like earlier one could do like
`make REGISTRY_NAMESPACE=myregistry IMAGE_TAG=latest ocs-operator` and it would acknowledges the changes right. But that will be not the case after this @jarrpa

jarrpa:
> There was a concern that this will completely eliminate the use of Makefile in the sense that there is no enivronmental variables in that file. Like earlier one could do like
> `make REGISTRY_NAMESPACE=myregistry IMAGE_TAG=latest ocs-operator` and it would acknowledges the changes right. But that will be not the case after this @jarrpa

How so? I have a script that currently runs `make gen-latest-csv REGISTRY_NAMESPACE=${REGISTRY_NAMESPACE}` and that passes the values just fine.

nizamial09:
> > There was a concern that this will completely eliminate the use of Makefile in the sense that there is no enivronmental variables in that file. Like earlier one could do like
> > `make REGISTRY_NAMESPACE=myregistry IMAGE_TAG=latest ocs-operator` and it would acknowledges the changes right. But that will be not the case after this @jarrpa
> 
> How so? I have a script that currently runs `make gen-latest-csv REGISTRY_NAMESPACE=${REGISTRY_NAMESPACE}` and that passes the values just fine.

Yep. Its working. Sorry for the confusion

PR #409, opened 2020-02-11
changes the sequence of Provided APIs

umangachapagain:
Brings OCS APIs to the front.

image
Screenshot of OCS 4.3 on OCP 4.4 Console


go.mod
```diff
@@ -11,6 +11,7 @@ require (
 	github.com/go-logr/logr v0.1.0
 	github.com/go-logr/zapr v0.1.1
 	github.com/go-openapi/spec v0.19.2
+	github.com/go-openapi/validate v0.18.0 // indirect
```
egafford: The rest of this patch appears to be a simple copy/paste move with no substantive change. Why are we adding a module here?

umangachapagain: Removed. Was autogenerated for some reason.

egafford: Thanks! And thanks for the second PR to add the required dep.

PR #410, opened 2020-02-11
Updates the reconcile logic to use storageDevicePVC for mons

ashishranjan738:
This commit updates the reconcile logic to use storageDeviceSet PVC template for mons if no monPVCTemplate is provided. The reconcile flow after this commit is as follows.
1. If the `monPVCTemplate` is provided, the mons will be provisioned on the provided `monPVCTemplate`.
2. If the `monDataDirHostPath` is provided without the `monPVCTemplate`, the mons will be provisioned on the provided `monDataDirHostPath`.
3. If no `monPVCTemplate` and no `monDataDirHostPath` is provided, the mons will be provisioned using the PVC template of first StorageDeviceSets if present.

### DIFF
--- a/pkg/controller/storagecluster/reconcile.go
+++ b/pkg/controller/storagecluster/reconcile.go
@@ -23,6 +23,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	storagev1 "k8s.io/api/storage/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/strategicpatch"
@@ -640,7 +641,7 @@ func (r *ReconcileStorageCluster) ensureCephCluster(sc *ocsv1.StorageCluster, re
 	}
 
 	// Define a new CephCluster object
-	cephCluster := newCephCluster(sc, r.cephImage, r.nodeCount)
+	cephCluster := newCephCluster(sc, r.cephImage, r.nodeCount, reqLogger)
 
 	// Set StorageCluster instance as the owner and controller
 	if err := controllerutil.SetControllerReference(sc, cephCluster, r.scheme); err != nil {
@@ -729,7 +730,7 @@ func determineFailureDomain(sc *ocsv1.StorageCluster) string {
 }
 
 // newCephCluster returns a CephCluster object.
-func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, nodeCount int) *cephv1.CephCluster {
+func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, nodeCount int, reqLogger logr.Logger) *cephv1.CephCluster {
 	labels := map[string]string{
 		"app": sc.Name,
 	}
@@ -782,12 +783,32 @@ func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, nodeCount int) *
 			ContinueUpgradeAfterChecksEvenIfNotHealthy: true,
 		},
 	}
-
-	if len(sc.Spec.MonDataDirHostPath) != 0 {
-		cephCluster.Spec.DataDirHostPath = sc.Spec.MonDataDirHostPath
-	}
-	if sc.Spec.MonPVCTemplate != nil {
-		cephCluster.Spec.Mon.VolumeClaimTemplate = sc.Spec.MonPVCTemplate
+	monPVCTemplate := sc.Spec.MonPVCTemplate
+	monDataDirHostPath := sc.Spec.MonDataDirHostPath
+	// If the `monPVCTemplate` is provided, the mons will provisioned on the
+	// provided `monPVCTemplate`.
+	if monPVCTemplate != nil {
+		cephCluster.Spec.Mon.VolumeClaimTemplate = monPVCTemplate
+		// If the `monDataDirHostPath` is provided without the `monPVCTemplate`,
+		// the mons will be provisioned on the provided `monDataDirHostPath`.
+	} else if len(monDataDirHostPath) > 0 {
+		cephCluster.Spec.DataDirHostPath = monDataDirHostPath
+		// If no `monPVCTemplate` and `monDataDirHostPath` is provided, the mons will
+		// be provisioned using the PVC template of first StorageDeviceSets if present.
+	} else if len(sc.Spec.StorageDeviceSets) > 0 {
+		ds := sc.Spec.StorageDeviceSets[0]
+		cephCluster.Spec.Mon.VolumeClaimTemplate = &corev1.PersistentVolumeClaim{
+			Spec: corev1.PersistentVolumeClaimSpec{
+				StorageClassName: ds.DataPVCTemplate.Spec.StorageClassName,
+				Resources: corev1.ResourceRequirements{
+					Requests: corev1.ResourceList{
+						corev1.ResourceStorage: resource.MustParse("10Gi"),
+					},
+				},
+			},
+		}
+	} else {
+		reqLogger.Info(fmt.Sprintf("No monDataDirHostPath, monPVCTemplate or storageDeviceSets configured for storageCluster %s", sc.GetName()))
 	}
 	return cephCluster
 }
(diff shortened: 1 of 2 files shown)


deploy/crds/ocs_v1_storagecluster_crd.yaml
```diff
@@ -107,6 +107,8 @@ spec:
                 - dataPVCTemplate
                 type: object
               type: array
+            useHostPathForMons:
```
travisn: Do we allow the host path to be specified? For example, `/var/lib/rook` might not always be valid. Instead of a bool, should this just be a string for the host path? Or is it possible to specify somewhere else?

ashishranjan738: yup the hostPath is still there. This PR just adds a boolean flag to disable use of pvcTemplate when enabled.

ashishranjan738: yup the hostPath is still there. This PR just adds a boolean flag to disable use of pvcTemplate when enabled. https://github.com/openshift/ocs-operator/pull/410/files#diff-24bbbd5f2f20ffccf914757966c7401aR765

jarrpa:
I don't like the proliferation of booleans in the API. We should just be able to use a mon hostPath if:

1. A hostPath is configured
2. No PVC template is specified

Why do we need this?

ashishranjan738:
> 2\. No PVC template is specified

we need this because the changes required for enabling the hostPath for mons is not present in 4.3's UI :(

rohantmp:
- Leave the default behaviour as-is so UI is not affected (before the mon's on hostPath patch)
- UI does not specify `monDataDirHostPath`

so `monsOnHostPath = monDataDirHostPathExists && ! monPVCTemplateExists`

ashishranjan738:
> * Leave the default behaviour as-is so UI is not affected (before the mon's on hostPath patch)
> * UI does not specify `monDataDirHostPath`
> 
> so `monsOnHostPath = monDataDirHostPathExists && !monPVCTemplateExists`

updated the PR. @rohantmp can have a relook on this.

ashishranjan738:
> > * Leave the default behaviour as-is so UI is not affected (before the mon's on hostPath patch)
> > * UI does not specify `monDataDirHostPath`
> > 
> > so `monsOnHostPath = monDataDirHostPathExists && !monPVCTemplateExists`
> 
> updated the PR. @rohantmp can have a relook on this.

cc @jarrpa @travisn

jarrpa:
So if there's no UI change, why is there a Boolean the UI won't use?

ashishranjan738:
> So if there's no UI change, why is there a Boolean the UI won't use?

I removed the boolean variable

pkg/controller/storagecluster/reconcile.go
```diff
@@ -744,11 +745,33 @@ func newCephCluster(sc *ocsv1.StorageCluster, cephImage string) *cephv1.CephClus
 		},
 	}
 
-	if len(sc.Spec.MonDataDirHostPath) != 0 {
+	// If both `MonDataDirHostPath` and `monPVCTemplate` is provided, the mons will
+	// be provisioned on the provided `monPVCTemplate` and `monDataDirHostPath`.
+	if len(sc.Spec.MonDataDirHostPath) != 0 && sc.Spec.MonPVCTemplate != nil {
```
rohantmp: can we extract `len(sc.Spec.MonDataDirHostPath) != 0` and `sc.Spec.MonPVCTemplate != nil` into vars like `monDataDirHostPathExists` and `monPVCTemplate` instead of evaluating them every time? It'll be easier to read the code as english too.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -744,11 +745,33 @@ func newCephCluster(sc *ocsv1.StorageCluster, cephImage string) *cephv1.CephClus
...
+		// on the provided `monDataDirHostPath`.
+	} else if len(sc.Spec.MonDataDirHostPath) > 0 {
+		cephCluster.Spec.DataDirHostPath = sc.Spec.MonDataDirHostPath
+		// If only the `monPVCTemplate` is provided, the mons will provisioned on the
+		// provided `monPVCTemplate`.
+	} else if sc.Spec.MonPVCTemplate != nil {
+		cephCluster.Spec.Mon.VolumeClaimTemplate = sc.Spec.MonPVCTemplate
+		// If no `monPVCTemplate` and `monDataDirHostPath` is provided, the mons will
+		// be provisioned using the PVC template of first StorageDeviceSets if present.
+	} else if len(sc.Spec.StorageDeviceSets) > 0 {
```
rohantmp: I think we should log a warning if none of these are matched.

pkg/controller/storagecluster/reconcile.go
jarrpa: I don't think this case makes sense. In the case of both of these being configured we should probably just take the PVC template and not set the data dir hostpath. This would also be grounds for an admission controller to prevent this configuration, but that's in the future. I think the cases should just be:

1. If PVC template not nil (give it preference in the ordering, so data dir hostpath is ignored if this is set)
2. If data dir hostpath not empty
3. If StorageDeviceSets exist
4. Else, warning (as Rohan said), and set nothing, letting Rook-Ceph take its default behavior

pkg/controller/storagecluster/reconcile.go
ashishranjan738: added it in the ensure function.

pkg/controller/storagecluster/reconcile.go
ashishranjan738: done

ashishranjan738: yup updated the PR

pkg/controller/storagecluster/reconcile.go
```diff
@@ -727,12 +732,30 @@ func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, nodeCount int) *
...
+		cephCluster.Spec.Mon.VolumeClaimTemplate = &corev1.PersistentVolumeClaim{
+			Spec: corev1.PersistentVolumeClaimSpec{
+				StorageClassName: ds.DataPVCTemplate.Spec.StorageClassName,
+				Resources: corev1.ResourceRequirements{
+					Requests: corev1.ResourceList{
+						corev1.ResourceStorage: resource.MustParse("10Gi"),
+					},
+				},
+			},
+		}
```
jarrpa: You did not add the warning that was requested.

ashishranjan738: its there https://github.com/openshift/ocs-operator/pull/410/files#diff-24bbbd5f2f20ffccf914757966c7401aR590-R592

pkg/controller/storagecluster/reconcile.go
```diff
@@ -586,6 +587,10 @@ func (r *ReconcileStorageCluster) ensureCephConfig(sc *ocsv1.StorageCluster, req
 // ensureCephCluster ensures that a CephCluster resource exists with its Spec in
 // the desired state.
 func (r *ReconcileStorageCluster) ensureCephCluster(sc *ocsv1.StorageCluster, reqLogger logr.Logger) error {
+	if len(sc.Spec.MonDataDirHostPath) == 0 && sc.Spec.MonPVCTemplate == nil && len(sc.Spec.StorageDeviceSets) == 0 {
+		reqLogger.Info(fmt.Sprintf("No monDataDirHostPath, monPVCTemplate or storageDeviceSets configures for storageCluster %s", sc.GetName()))
```
rohantmp: s/configures/configured

pkg/controller/storagecluster/reconcile.go
rohantmp: @jarrpa why shouldn't we allow setting the dataDirHostPath even if mons are not using hostPath? Rook still ends up using the default `/var/lib/rook` for OSD.

 Seems like we're getting it for free with this new field.

pkg/controller/storagecluster/reconcile.go
ashishranjan738: done

ashishranjan738:
@jarrpa can you rereview this ?

pkg/controller/storagecluster/reconcile.go
```diff
@@ -590,6 +591,10 @@ func (r *ReconcileStorageCluster) ensureCephConfig(sc *ocsv1.StorageCluster, req
 // ensureCephCluster ensures that a CephCluster resource exists with its Spec in
 // the desired state.
 func (r *ReconcileStorageCluster) ensureCephCluster(sc *ocsv1.StorageCluster, reqLogger logr.Logger) error {
+	if len(sc.Spec.MonDataDirHostPath) == 0 && sc.Spec.MonPVCTemplate == nil && len(sc.Spec.StorageDeviceSets) == 0 {
```
jarrpa: There's no reason to have this this early and separate from the latter if/else statements. Move this down as the final `else` around line 762

ashishranjan738: because we don't have the logger variable in the scope of the newCephCluster function. And I don't think we should be changing the signature of the newCephCluster function.

PR #413, opened 2020-02-15
Bug 1775432- sets mon_osd_full_ratio=.85 & mon_osd_nearfull_ratio =.75

umangachapagain:
"rook-config-override" ConfigMap after changes,
```yaml
data:
  config: |

    [global]
    mon_osd_full_ratio = .85
    mon_osd_nearfull_ratio = .75
    [osd]
```

```bash
sh-4.2# ceph osd dump | grep full
full_ratio 0.85
nearfull_ratio 0.75
```

### DIFF
--- a/pkg/controller/storagecluster/reconcile.go
+++ b/pkg/controller/storagecluster/reconcile.go
@@ -33,7 +33,11 @@ import (
 
 const (
 	rookConfigMapName = "rook-config-override"
-	rookConfigData    = `[osd]
+	rookConfigData    = `
+[global]
+mon_osd_full_ratio = .85
+mon_osd_nearfull_ratio = .75
+[osd]
 osd_memory_target_cgroup_limit_ratio = 0.5
 `
 	monCountOverrideEnvVar = "MON_COUNT_OVERRIDE"


umangachapagain:
> It would be nice to add a short explanation of why these changes are needed to the commit message.
> 
Done.

> Also, be careful with whitespace changes on config changes that propagate to Ceph. We saw an issue before in which the lack of a newline at the end of the string caused a failure to read the config. I haven't personally tested whether the presence of a newline at the beginning will cause any issues, but do be sure it's being loaded successfully.
>
I have tested and it works just fine. I have added the output in the description for clarity.

egafford:
> > It would be nice to add a short explanation of why these changes are needed to the commit message.
> 
> Done.
> 
> > Also, be careful with whitespace changes on config changes that propagate to Ceph. We saw an issue before in which the lack of a newline at the end of the string caused a failure to read the config. I haven't personally tested whether the presence of a newline at the beginning will cause any issues, but do be sure it's being loaded successfully.
> 
> I have tested and it works just fine. I have added the output in the description for clarity.

Thanks, Umanga.

PR #414, opened 2020-02-15
Bug 1797918 - Enables balancer module

umangachapagain:
- enable "balancer" module

### DIFF
--- a/pkg/controller/storagecluster/initialization_reconciler.go
+++ b/pkg/controller/storagecluster/initialization_reconciler.go
@@ -251,7 +251,8 @@ func (r *ReconcileStorageCluster) newCephBlockPoolInstances(initData *ocsv1.Stor
 			Spec: cephv1.PoolSpec{
 				FailureDomain: initData.Status.FailureDomain,
 				Replicated: cephv1.ReplicatedSpec{
-					Size: 3,
+					Size:            3,
+					TargetSizeRatio: .5,
 				},
 			},
 		},
@@ -401,7 +402,8 @@ func (r *ReconcileStorageCluster) newCephFilesystemInstances(initData *ocsv1.Sto
 				DataPools: []cephv1.PoolSpec{
 					cephv1.PoolSpec{
 						Replicated: cephv1.ReplicatedSpec{
-							Size: 3,
+							Size:            3,
+							TargetSizeRatio: .5,
 						},
 						FailureDomain: initData.Status.FailureDomain,
 					},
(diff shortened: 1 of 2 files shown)


obnoxxx:
@umangachapagain the back and forth on #401 and this PR confirms my preference to separate updates of dependent components (like rook) and changes that do semantic changes (that possibly depend on the updated component) into separate PRs...

obnoxxx:
(not sure why it didn't have the ok-to-test label)

umangachapagain:
> @umangachapagain the back and forth on #401 and this PR confirms my preference to separate updates of dependent components (like rook) and changes that do semantic changes (that possibly depend on the updated component) into separate PRs...

Agree that it is messy which is why I am working on a separate PR for handling this matter.
(something that will not require us to upgrade version every time we add a feature)

For now, this PR gets the latest from rook and all other PRs can be built on top of this.

umangachapagain:
> @umangachapagain a couple of questions:
> 
>     * Is the setting of `TargetSizeRatio` independent of the addition of the `balancer` module? If yes, then it should be two commits instead of one.
> 
They are dependent.

>     * Does the addition of these parameters require rook to be updated to 1.2.4?
> 
Yes. `TargetSizeRatio` and `balancer` module support was recently added.

>     * The setting of `TargetSizeRatio` seems to be hard coded. Can this be made configurable? (This could be a topic of a later PR that might also make other settings like the Size of the ReplicatedSpec configurable, etc).
From what I understand, this seems to be a suitable value for OCS usecase and could be fixed.
@leseb Thoughts?

> In general, I think patches bumping versions of dependent components like rook should be in a separate PR.
>
I'll go ahead and separate the commits then.

obnoxxx:
@umangachapagain you can leave it in this one PR for now. Just remember next time, please, it makes stuff eaiser.

And with respect to the dependency of changes (ratio depends on balancer module and balancer module depends on the update of rook: please just explain and mention these in the commit messages / pr description and/or code comments. This is important to understand depndencies and will prevent PR review questions and future questions by someone looking at the code / git history.

obnoxxx:
@umangachapagain regarding the configurability of values - I have just been bitten too many times by hard coded values that were good defaults, but then someone still needed to change them.

This PR is ok to have them hard coded. But we may want to make them configurable in a follow-up commit.

obnoxxx:
> /test ocs-operator-e2e-a

incomplete ;-)

/test ocs-operator-e2e-aws

obnoxxx:
> >  Is the setting of `TargetSizeRatio` independent of the addition of the `balancer` module? If yes, then it should be two commits instead of one.
> 
> They are dependent.

It seems that they are actually independent:
* The balancer support was added in ceph with https://github.com/rook/rook/pull/4784.
* The TargetRatio which is for the pg_autoscaler, not for balancer, was added in https://github.com/rook/rook/pull/4803.

So I would argue that these should be two different patches.  One could even argue that they should be two differnt PRs... At least the PR description needs to mention both.

PR #417, opened 2020-02-19
deps-update: updates rook to v1.2.4

umangachapagain:


umangachapagain:
> /ok-to-test
> 
> This looks good in principle, but it has the update to the csv generation and the update to the deps in one patch. Is that required? Or good practice?

I prefer single commit because it addresses the same concern. Having 2 separate commits will make it look like it's two different things but it is not.
Same commit should bring the code changes and updated images IMO.

obnoxxx:
> I prefer single commit because it addresses the same concern. Having 2 separate commits will make it look like it's two different things but it is not.
> Same commit should bring the code changes and updated images IMO.

Well, it's not really code changes but dep updates.
But I think what you mean is that the image update for the csv generation and the vendored dep update should be consistent / in sync. That is true! (As much as I dislike committing the deps as code. It is horrible.) It still doesn't mean that they need to be in one commit since csv generation is in theory an independent thing from the actual ocs-operator code (which the deps are for). In that regard, I still think it is conceptually cleaner to separate the patches (in one PR), unless we are forced to squash because test or build would fail. All that said, I won't block this PR for it, just wanted to make the point for future cases.

/lgtm

PR #418, opened 2020-02-19
[WIP] BZ- 1780002: OCS operator should proceed with rook upgrade even if ceph health is not ok

sp98:
-The readiness probe fails for OCS opeartor (Readiness probe failed: stat: cannot stat '/tmp/operator-sdk-ready)when the Ceph cluster is in "Creating" phase. This causes ocs not to send upgrade request to rook. 
-Since rook is already designed to delay the updating the cluster if its not health, we can have ocs ignore the ceph cluster health.
-This change involves not adding 'conditions' to the ocs when cephCluster 'status.State' has value 'Creating', so that '/tmp/operator-sdk-ready' would be created.

PS: This is not ready for review yet. 

### DIFF
--- a/pkg/controller/util/status.go
+++ b/pkg/controller/util/status.go
@@ -115,19 +115,6 @@ func SetCompleteCondition(conditions *[]conditionsv1.Condition, reason string, m
 // This will only look for negative conditions: !Available, Degraded, Progressing
 func MapCephClusterNegativeConditions(conditions *[]conditionsv1.Condition, found *cephv1.CephCluster) {
 	switch found.Status.State {
-	case cephv1.ClusterStateCreating:
-		conditionsv1.SetStatusCondition(conditions, conditionsv1.Condition{
-			Type:    conditionsv1.ConditionProgressing,
-			Status:  corev1.ConditionTrue,
-			Reason:  "ClusterStateCreating",
-			Message: fmt.Sprintf("CephCluster is creating: %v", string(found.Status.Message)),
-		})
-		conditionsv1.SetStatusCondition(conditions, conditionsv1.Condition{
-			Type:    conditionsv1.ConditionUpgradeable,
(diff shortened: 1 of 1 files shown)


umangachapagain:
I do not like the idea of removing Conditions. How about we just set Operator to ready after carefully analyzing the conditions?

I have another issue raised asking for OCS to accurately track Ceph Cluster health. This has effects all the way to OCS UI E2E tests.

jarrpa:
I think rather than removing those two conditions entirely, we should be a little more nuanced about how we set and check them. Ultimately, our readiness probe is to tell the OLM whether we are in a stable state for upgrade or not, so instead of checking for ANY Condition we should check to see if ConditionUpgradable is true. We would then set it to true for CephClusterCreating, false for the other two, and then default to true if HEALTH_OK.

@travisn Does that make sense?

travisn:
The original issue was that the upgrade was blocked if the Ceph health was not HEALTH_OK. The discussion on BZ 1780002 was to not block the OCS upgrades based on Ceph health.

The change I see in this PR is related to the Rook CR status and I'm not sure why we are changing it. If an orchestration is in progress we probably don't want to trigger an OCS upgrade. 

Where is the check for Ceph status that was blocking the OCS upgrade? Or am I misinterpreting this change?

sp98:
> The original issue was that the upgrade was blocked if the Ceph health was not HEALTH_OK. The discussion on BZ 1780002 was to not block the OCS upgrades based on Ceph health.
> 
> The change I see in this PR is related to the Rook CR status and I'm not sure why we are changing it. If an orchestration is in progress we probably don't want to trigger an OCS upgrade.
> 
> Where is the check for Ceph status that was blocking the OCS upgrade? Or am I misinterpreting this change?

The change was made long back before we had a detailed discussion in the BZ. Please ignore these changes.

PR #424, opened 2020-02-24
ob and obc will move to owned instead of required

jackyalbo:
As part of not being dependent on community operators in downstream
fix to BZ1798571
upstream NooBaa will still require lib_bucket_provisionser  as a dependency


guymguym:
@umangachapagain That can't come from NooBaa itself.
Upstream noobaa will still require those CRDs and get them through OLM dependency on the lib-bucket-provisioner operator (CRD only operator).
However downstream we don't want OCS to depend on a community operator and therefore we decide to own those CRDs by OCS.

guymguym:
Yes it does. We still get it the same way, just now we patch it from required to owned.

guymguym:
But @umangachapagain you’re right that @jackyalbo should make sure that those crd files themselves are really updated from noobaa operator during CSV merge, and not statically stored in OCS git.

jackyalbo:
they are updated by the operator - was added by running "make gen-latest-csv"

jackyalbo:
fixed to 2 separate commits

tools/csv-merger/csv-merger.go
```diff
@@ -439,7 +439,15 @@ func generateUnifiedCSV() *csvv1.ClusterServiceVersion {
 			templateStrategySpec.Permissions = append(templateStrategySpec.Permissions, permissions...)
 
 			ocsCSV.Spec.CustomResourceDefinitions.Owned = append(ocsCSV.Spec.CustomResourceDefinitions.Owned, csvStruct.Spec.CustomResourceDefinitions.Owned...)
-			ocsCSV.Spec.CustomResourceDefinitions.Required = append(ocsCSV.Spec.CustomResourceDefinitions.Required, csvStruct.Spec.CustomResourceDefinitions.Required...)
+
+			for _, definition := range csvStruct.Spec.CustomResourceDefinitions.Required {
```
jarrpa: Is this in place because the NooBaa CSV we're importing isn't updated to reflect this change? If so, shouldn't the change be in NooBaa?

umangachapagain:
> @umangachapagain That can't come from NooBaa itself.
> Upstream noobaa will still require those CRDs and get them through OLM dependency on the lib-bucket-provisioner operator (CRD only operator).
>
If lib-bucket-provisioner operator is gone, how will NooBaa operator (older versions) function upstream?

guymguym:
@umangachapagain lib-bucket-provisioner is only removed from cumminity-operators for Openshift, not from upstream-community-operators for Kubernetes.

PR #429, opened 2020-02-28
Independent mode initialization and cluster status

vbnrh:
This PR is a followup for https://github.com/openshift/ocs-operator/pull/393 and completes the addition of 
1. Initialization of the external ceph cluster
2. Adds support for phases such as connecting, connected


pkg/controller/storagecluster/externalstoragecluster_controller.go
```diff
@@ -63,6 +63,27 @@ func (r *ReconcileStorageCluster) ReconcileExternalStorageCluster(sc *ocsv1.Stor
 			return reconcile.Result{}, err
 		}
 	}
+
+	// Handle CephCluster resource status
+	if found.Status.State == "" {
```
nizamial09: found.Status.Phase == ""

State will be changed to Phase on rook 1.3 and future OCS versions will be using Phase variable.

vbnrh: @nizamial09 could you please provide some reference for this change ?

nizamial09: https://github.com/rook/rook/blob/master/pkg/apis/ceph.rook.io/v1/types.go#L145

So for rook:master the `Phase` is used. Any build starting from rook v1.3 will starts to use `Phase` instead of `State`. So I guess its better to start using `Phase` here also.

nizamial09: @umangachapagain ^

umangachapagain: Yes. Use rook:master.

travisn: @nizamial09 Don't we set both Phase and State in master? I thought we needed to set State as well to avoid breaking changes. For example, there may by others than OCS that are using the State and would be surprised by the upgrade to 1.3.
OCS likely doesn't want to use the `rook/ceph:master` tag as it will be a bit unstable. Perhaps a particular master tag could be used and updated periodically if we need to move to master though.

nizamial09: I see your concern. My PR will completely eliminate the use of State from the master. But you are correct. We need to be doing it slowly. I will make another PR which will bring back the State for now and in future we can slowly deprecate it's use. @travisn

vbnrh: Made the changes from State to Phase

aruniiird:
@jarrpa , @leseb , @travisn , please take look...

pkg/controller/util/status.go
```diff
@@ -157,6 +161,47 @@ func MapCephClusterNegativeConditions(conditions *[]conditionsv1.Condition, foun
 	}
 }
 
+
+// MapExternalCephClusterNegativeConditions maps the status states from CephCluster resource into ocs status conditions.
+// This will only look for negative conditions: !Available, Degraded, Progressing
+func MapExternalCephClusterNegativeConditions(conditions *[]conditionsv1.Condition, found *cephv1.CephCluster) {
```
umangachapagain: Use variable for all the `Reason`s and `Message`s.

umangachapagain: Either use `ExternalCephCluster` or `External CephCluster`, not both.

pkg/controller/storagecluster/externalstoragecluster_controller.go
```diff
@@ -0,0 +1,317 @@
...
+		r.phase = statusutil.PhaseClusterExpanding
+	}
+
+	if found.Status.State == cephv1.ClusterStateConnecting {
+		sc.Status.Phase = statusutil.PhaseConnecting
+	}
+
+	if found.Status.State == cephv1.ClusterStateConnected {
+		sc.Status.Phase = statusutil.PhaseConnected
+	}
```
umangachapagain: Is this actually getting reflected on the CR? I don't see it being updated anywhere.

vbnrh: It was reflecting in the CR when we tested it. Could you elaborate on the updated part ?

pkg/controller/util/status.go
vbnrh: > Either use ExternalCephCluster or External CephCluster, not both.

Are you talking about the argument we are passing to this function or the function name ?

umangachapagain: I am talking about the `Message`s.

pkg/controller/storagecluster/externalstoragecluster_controller.go
umangachapagain: These lines are just assigning values to variable. The actual status update is happening somewhere else. I am asking if you are updating it yourself or depending on some other function to update it? (where is the API call to write this to etcd ?)

vbnrh: I was thinking that the reconcile logic would be taking care of writing/updating all the corresponding values. If you could point me on how we could do it manually, I'll take of it

pkg/controller/util/status.go
vbnrh: Made everything External CephCluster !
As for using variable in Reason and Message, We decided to use the existing MapCephClusterNegativeConditions and modified it to suit our needs. In that function, there were no variables for reason and message. Let me know if this fix is required ?

umangachapagain: I'd prefer using variables at least for `Reason`s like we do with `Phase`.

vbnrh: Added variables for reason

vbnrh:
@jarrpa , @leseb , @travisn , please take look

pkg/controller/storagecluster/generate.go
```diff
@@ -6,8 +6,19 @@ import (
 	ocsv1 "github.com/openshift/ocs-operator/pkg/apis/ocs/v1"
 )
 
+func generateNameForExternalCephCluster(initData *ocsv1.StorageCluster) string {
+	return generateNameForCephClusterFromString(fmt.Sprintf("%s-external", initData.Name))
+}
+
 func generateNameForCephCluster(initData *ocsv1.StorageCluster) string {
-	return generateNameForCephClusterFromString(initData.Name)
+	clusterName := ""
```
leseb: Why would we need to change the name of the cluster? IIRC OCS-Op only supports a single CephCluster anyways.

pkg/apis/ocs/v1/storagecluster_types.go
```diff
@@ -127,14 +127,26 @@ type NodeTopologyMap struct {
 
 // ConditionReconcileComplete communicates the status of the StorageCluster resource's
 // reconcile functionality. Basically, is the Reconcile function running to completion.
-const ConditionReconcileComplete conditionsv1.ConditionType = "ReconcileComplete"
+const (
```
umangachapagain: Move comments from Line 129, 129 here.

pkg/controller/storagecluster/generate.go
```diff
@@ -6,8 +6,19 @@ import (
 	ocsv1 "github.com/openshift/ocs-operator/pkg/apis/ocs/v1"
 )
 
+func generateNameForExternalCephCluster(initData *ocsv1.StorageCluster) string {
+	return generateNameForCephClusterFromString(fmt.Sprintf("%s-external", initData.Name))
+}
+
```
umangachapagain:
```suggestion
```

pkg/controller/storagecluster/generate.go
```diff
@@ -6,8 +6,19 @@ import (
...
 func generateNameForCephCluster(initData *ocsv1.StorageCluster) string {
-	return generateNameForCephClusterFromString(initData.Name)
+	clusterName := ""
+	// if 'ExternalStorage' is enabled, change the cluster name accordingly
+	if initData.Spec.ExternalStorage.Enable {
+		clusterName = generateNameForExternalCephCluster(initData)
```
umangachapagain:
```suggestion
		clusterName = generateNameForCephClusterFromString(fmt.Sprintf("%s-external", initData.Name))
```

pkg/controller/storagecluster/reconcile.go
```diff
@@ -262,19 +254,28 @@ func (r *ReconcileStorageCluster) Reconcile(request reconcile.Request) (reconcil
...
+	if instance.Spec.ExternalStorage.Enable {
+		ensureFs = []ensureFunc{
+			r.ensureCephConfig,
+			r.ensureCephCluster,
+		}
+	}
+	for _, f := range ensureFs {
 		err = f(instance, reqLogger)
-		if r.phase == statusutil.PhaseClusterExpanding {
+		if r.phase == statusutil.PhaseClusterExpanding && !instance.Spec.ExternalStorage.Enable {
```
umangachapagain:
```suggestion
		if r.phase == statusutil.PhaseClusterExpanding {
```

If it is PhaseClusterExpanding, it is  definitely not an External StorageCluster.
Instead of doing this, we should ensure we don't set the wrong phase to the wrong StorageCluster and infer the rest.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -654,29 +696,37 @@ func (r *ReconcileStorageCluster) ensureCephCluster(sc *ocsv1.StorageCluster, re
...
 			return r.client.Create(context.TODO(), cephCluster)
 		}
 		return err
 	}
 
 	// Update the CephCluster if it is not in the desired state
 	if !reflect.DeepEqual(cephCluster.Spec, found.Spec) {
-		reqLogger.Info("Updating spec for CephCluster")
+		reqLogger.Info(fmt.Sprintf("Updating spec for %s", cephClusterTypeStr))
 		// Check if Cluster is Expanding
```
umangachapagain: Move to else block.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -814,6 +888,26 @@ func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, nodeCount int, r
 	return cephCluster
 }
 
+func newExternalCephCluster(sc *ocsv1.StorageCluster, cephImage string) *cephv1.CephCluster {
+	labels := map[string]string{
+		"app": sc.Name,
+	}
+	externalCephCluster := &cephv1.CephCluster{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      generateNameForExternalCephCluster(sc),
```
umangachapagain: Either use `generateNameForExternalCephCluster` or `generateNameForCephCluster`.
Currently, you've mixed both the implementations.

pkg/controller/util/status.go
```diff
@@ -27,6 +27,16 @@ var (
 	PhaseClusterExpanding = "Expanding Capacity"
 	// PhaseDeleting is used when cluster is deleting
 	PhaseDeleting = "Deleting"
+	// PhaseConnecting is used when cluster is connecting to external cluster
+	PhaseConnecting = "Connecting"
+	//PhaseConnected is used when cluster has connected to external cluster
+	PhaseConnected = "Connected"
+)
+
+const (
```
umangachapagain: Exported Constants should have comments. 
Check lint errors.

pkg/controller/storagecluster/generate.go
umangachapagain: Idea was to make cluster identifiable by name, no other reason.

pkg/controller/util/status.go
aruniiird: Added the comments

pkg/controller/storagecluster/reconcile.go
aruniiird: Removed `generateNameForExternalCephCluster`, now using only `generateNameForCephCluster` method...

pkg/controller/storagecluster/reconcile.go
aruniiird: Moved the comment to the `else` side

pkg/controller/storagecluster/reconcile.go
aruniiird: Please take a look at the above change in the code, we tried to handle wrong phase issue.

pkg/controller/storagecluster/generate.go
aruniiird: Removed the external name genaration function, `generateNameForExternalCephCluster`...

pkg/controller/storagecluster/generate.go
aruniiird: Done...

pkg/apis/ocs/v1/storagecluster_types.go
aruniiird: Addressed...

pkg/controller/storagecluster/reconcile.go
```diff
@@ -31,12 +32,14 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
-	"github.com/openshift/ocs-operator/version"
 )
 
 // StorageClassProvisionerType is a string representing StorageClass Provisioner. E.g: aws-ebs
 type StorageClassProvisionerType string
 
+// ensureFunc s
```
umangachapagain: comment makes no sense to me.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -196,52 +186,54 @@ func (r *ReconcileStorageCluster) Reconcile(request reconcile.Request) (reconcil
...
+				case errors.IsAlreadyExists(err):
+					log.Info("StorageClusterInitialization resource already exists")
+				default:
+					log.Error(err, "Failed to create StorageClusterInitialization resource")
+					return reconcile.Result{}, err
+				}
+			} else {
+				// Error reading the object - requeue the request.
 				return reconcile.Result{}, err
 			}
```
umangachapagain: Let's move this out to a separate function (`ensureStorageClusterInitialization`) for sake of readability.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -262,19 +254,32 @@ func (r *ReconcileStorageCluster) Reconcile(request reconcile.Request) (reconcil
 		r.ensureCephConfig,
 		r.ensureCephCluster,
 		r.ensureNoobaaSystem,
-	} {
+	}
+	// for external cluster, we have a different set of ensure functions
+	if instance.Spec.ExternalStorage.Enable {
+		ensureFs = []ensureFunc{
+			r.ensureCephConfig,
```
umangachapagain: We don't need this.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -250,8 +242,8 @@ func (r *ReconcileStorageCluster) Reconcile(request reconcile.Request) (reconcil
 	r.conditions = nil
 	// Start with empty r.phase
 	r.phase = ""
-
-	for _, f := range []func(*ocsv1.StorageCluster, logr.Logger) error{
+	// list of default enusure functions
```
umangachapagain: Instead of mixing and complicating things, can we separate concerns? This looks like a code maintenance nightmare to me.

Let's have `ensureConvergedCluster` and `ensureExternalCluster` functions that will just do what it is supposed to do? We can use if/else to diverge the flow wherever needed.

@jarrpa WDYT?

pkg/controller/storagecluster/reconcile.go
```diff
@@ -262,19 +254,32 @@ func (r *ReconcileStorageCluster) Reconcile(request reconcile.Request) (reconcil
...
+			r.ensureCephCluster,
+		}
+	}
+	for _, f := range ensureFs {
 		err = f(instance, reqLogger)
 		if r.phase == statusutil.PhaseClusterExpanding {
+			if instance.Spec.ExternalStorage.Enable {
+				wrongPhaseErr := fmt.Errorf("External cluster should not be an expanding phase")
+				return reconcile.Result{}, wrongPhaseErr
+			}
```
umangachapagain: WHY?

pkg/controller/storagecluster/reconcile.go
aruniiird: made changes

pkg/controller/storagecluster/reconcile.go
aruniiird: Moved it to a separate function as suggested (`ensureStorageClusterInit`)...

pkg/controller/storagecluster/reconcile.go
aruniiird: Removed...

pkg/controller/storagecluster/reconcile.go
aruniiird: Removed...

pkg/controller/storagecluster/reconcile.go
aruniiird: @umangachapagain , may be I am not seeing the big picture here, but until you are doing different actions in the `for` loop for each type of storage clusters (external or converged), you don't have to make separate functions (as the actions taken on each iterations are same for any type storagecluster).

aruniiird:

pkg/controller/storagecluster/reconcile.go
jarrpa: At the moment I don't want separate reconciliation functions between converged and external clusters. I think the overhead of maintenance is worth it to make sure that we don't diverge greatly in execution between the two and ensure that we don't break one when modifying the other.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -271,10 +231,11 @@ func (r *ReconcileStorageCluster) Reconcile(request reconcile.Request) (reconcil
...
 			}
 		} else {
-			if instance.Status.Phase != statusutil.PhaseReady {
+			if instance.Status.Phase != statusutil.PhaseReady &&
+				instance.Status.Phase != statusutil.PhaseConnecting &&
+				instance.Status.Phase != statusutil.PhaseConnected {
 				instance.Status.Phase = statusutil.PhaseProgressing
-				phaseErr := r.client.Status().Update(context.TODO(), instance)
-				if phaseErr != nil {
+				if phaseErr := r.client.Status().Update(context.TODO(), instance); phaseErr != nil {
```
jarrpa, line 238: I see what you're going for here, but I prefer the way it was before; it's a little easier to parse.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -351,11 +313,89 @@ func (r *ReconcileStorageCluster) Reconcile(request reconcile.Request) (reconcil
 			}
 		}
 	}
-	phaseErr := r.client.Status().Update(context.TODO(), instance)
-	if phaseErr != nil {
+	if phaseErr := r.client.Status().Update(context.TODO(), instance); phaseErr != nil {
```
jarrpa, line 316: Same as before, leave the lines separated.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -626,23 +666,32 @@ func (r *ReconcileStorageCluster) ensureCephConfig(sc *ocsv1.StorageCluster, req
...
-	for i, ds := range sc.Spec.StorageDeviceSets {
-		throttle, err := r.throttleStorageDevices(*ds.DataPVCTemplate.Spec.StorageClassName)
-		if err != nil {
-			return fmt.Errorf("Failed to verify StorageClass provisioner. %+v", err)
-		}
-		if throttle {
-			sc.Spec.StorageDeviceSets[i].Config.TuneSlowDeviceClass = true
-		} else {
-			sc.Spec.StorageDeviceSets[i].Config.TuneSlowDeviceClass = false
+	if !sc.Spec.ExternalStorage.Enable {
```
jarrpa, line 669: Actually, for an external CephCluster I think we shouldn't be doing anything with StorageDeviceSets at all! What would you think about instead putting in an `if sc.Spec.ExternalStorage.Enable && sc.Spec.StorageDeviceSets != nil` clause that would throw an error? That way we require that there be no StorageDevices defined.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -626,23 +666,32 @@ func (r *ReconcileStorageCluster) ensureCephConfig(sc *ocsv1.StorageCluster, req
...
+			if throttle {
+				sc.Spec.StorageDeviceSets[i].Config.TuneSlowDeviceClass = true
+			} else {
+				sc.Spec.StorageDeviceSets[i].Config.TuneSlowDeviceClass = false
+			}
 		}
 	}
 
+	var cephCluster *cephv1.CephCluster
+	var cephClusterTypeStr = "CephCluster"
```
jarrpa, line 687: Similar to StorageClusters, there is no such thing as a new external CephCluster CR. I think it's fine to have at least one message saying that we're creating an external cluster, but we shouldn't need to keep repeating it for future log messages.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -654,29 +703,37 @@ func (r *ReconcileStorageCluster) ensureCephCluster(sc *ocsv1.StorageCluster, re
...
-		if len(found.Spec.Storage.StorageClassDeviceSets) < len(cephCluster.Spec.Storage.StorageClassDeviceSets) {
-			r.phase = statusutil.PhaseClusterExpanding
-		} else if len(found.Spec.Storage.StorageClassDeviceSets) == len(cephCluster.Spec.Storage.StorageClassDeviceSets) {
-			for _, countInFoundSpec := range found.Spec.Storage.StorageClassDeviceSets {
-				for _, countInCephClusterSpec := range cephCluster.Spec.Storage.StorageClassDeviceSets {
-					if countInFoundSpec.Name == countInCephClusterSpec.Name && countInCephClusterSpec.Count > countInFoundSpec.Count {
-						r.phase = statusutil.PhaseClusterExpanding
+		reqLogger.Info(fmt.Sprintf("Updating spec for %s", cephClusterTypeStr))
+		if sc.Spec.ExternalStorage.Enable {
+			sc.Status.Phase = string(found.Status.State)
```
jarrpa, line 716: Can you explain what this does? I think I have an idea, but I'm not quite sure. A comment in the code would be helpful.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -690,6 +747,13 @@ func (r *ReconcileStorageCluster) ensureCephCluster(sc *ocsv1.StorageCluster, re
 	}
 	objectreferencesv1.SetObjectReference(&sc.Status.RelatedObjects, *objectRef)
 
+	type mapNegativeConditionsFunc func(*[]conditionsv1.Condition, *cephv1.CephCluster)
```
jarrpa, line 750: Rather than doing this up here, it seems like it would make more sense to just do an `if/else` at like 766 to pick between mapping functions.

pkg/controller/storagecluster/reconcile.go
aruniiird: My main intention was, as we are not using `phaseErr` later, initializing it within if statement make it local and the variable will be marked immediately for Garbage Collection , once the if scope is done...
Reverted back to the previous way... (=

pkg/controller/storagecluster/reconcile.go
aruniiird: Got your point... 

I have made the change as you suggested, but has put `len(sc.Spec.StorageDeviceSets) != 0` approach and the logic behind was `len(<array>)` function will return `zero` for both nil and non-nil empty array.

pkg/controller/storagecluster/reconcile.go
aruniiird: Done...

pkg/controller/storagecluster/reconcile.go
aruniiird: Reasoning was: felt much more clutter if you put a if/else inside another else and since `Go` provides more functional style programming... (=
Changed...

pkg/controller/storagecluster/reconcile.go
jarrpa: In general I'm in favor of that, honestly. But in this case, the call `r.client.Status().Update(context.TODO(), instance)` is a bit too long and complex in punctuation. If it were something like `r.updateStorageCluster()` that would be more reasonable to me. I'm not suggesting it as a change here, mind you, just giving examples of when I would and would not advocate for this kind of optimization.

jarrpa: Also, maybe I'm missing something, but I still see the commit as being `if phaseErr := FOO; phaseErr != nil`, rather than the previous split.

pkg/controller/storagecluster/reconcile.go
jarrpa: Still missing a comment here.

aruniiird: In an external cluster mode, (and when ideal cephcluster.Spec doesn't match with the found cephCluster.Spec) we don't have much to do. Update the `Status.Phase` with the `State` of the found cluster.

@umangachapagain , will be able to explain better...

aruniiird:
@jarrpa ,
> I appreciate the work done to encapsulate some functionality of the Reconcile() function into their own functions. However, I'd like to have that as it's own commit. Could you split out the creation of versionCheck() and ensureStorageClusterInit() into their own commit?

Moved the function creation into their own commits..

pkg/controller/storagecluster/reconcile.go
aruniiird: Sorry, missed the change... Added it...

pkg/controller/storagecluster/reconcile.go
aruniiird: Added the comments...

pkg/controller/storagecluster/reconcile.go
```diff
@@ -644,6 +686,9 @@ func (r *ReconcileStorageCluster) ensureCephConfig(sc *ocsv1.StorageCluster, req
 // ensureCephCluster ensures that a CephCluster resource exists with its Spec in
 // the desired state.
 func (r *ReconcileStorageCluster) ensureCephCluster(sc *ocsv1.StorageCluster, reqLogger logr.Logger) error {
+	if sc.Spec.ExternalStorage.Enable && len(sc.Spec.StorageDeviceSets) != 0 {
+		return fmt.Errorf("'StorageDeviceSets' should not be initialized in an ExternalCluster mode")
```
jarrpa, line 690: `s/ExternalCluster mode/external CephCluster/`

pkg/controller/storagecluster/reconcile.go
jarrpa: You didn't mention why, though. Why are you copying State into Phase?

umangachapagain: This is very confusing at first look. I think what you want to do here is,
1. If it is an Update request for External Ceph Cluster just proceed with the update
2. If it is an Update request for Ceph Cluster,
       2.1 Check if Cluster is expanding and set PhaseExpanding accordingly. Then continue normal update
       2.2 If it is not expanding, proceed with normal update to reach desired state

pkg/controller/storagecluster/reconcile.go
```diff
@@ -681,20 +734,29 @@ func (r *ReconcileStorageCluster) ensureCephCluster(sc *ocsv1.StorageCluster, re
...
-					if countInFoundSpec.Name == countInCephClusterSpec.Name && countInCephClusterSpec.Count > countInFoundSpec.Count {
-						r.phase = statusutil.PhaseClusterExpanding
+		if sc.Spec.ExternalStorage.Enable {
+			// for an external cluster, put the current 'State' into Phase and update the status
+			sc.Status.Phase = string(found.Status.State)
+			if err := r.client.Status().Update(context.TODO(), sc); err != nil {
+				reqLogger.Error(err, "Failed to add conditions to status")
+				return err
+			}
+		} else {
```
umangachapagain, line 744:
```suggestion
		if !sc.Spec.ExternalStorage.Enable {
```

pkg/controller/storagecluster/reconcile.go
```diff
@@ -717,16 +779,37 @@ func (r *ReconcileStorageCluster) ensureCephCluster(sc *ocsv1.StorageCluster, re
...
+			statusutil.MapCephClusterNegativeConditions(&r.conditions, found)
+		}
 	}
 
 	// When phase is expanding, wait for CephCluster state to be updating
 	// this means expansion is in progress and overall system is progressing
 	// else expansion is not yet triggered
 	if sc.Status.Phase == statusutil.PhaseClusterExpanding &&
+		!sc.Spec.ExternalStorage.Enable &&
```
umangachapagain, line 793:
```suggestion
```
If `sc.Status.Phase == statusutil.PhaseClusterExpanding` is `true`,  `!sc.Spec.ExternalStorage.Enable` is also `true` because we never do expansion for External Cluster.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -717,16 +779,37 @@ func (r *ReconcileStorageCluster) ensureCephCluster(sc *ocsv1.StorageCluster, re
...
+		if found.Status.State == cephv1.ClusterStateConnecting {
+			sc.Status.Phase = statusutil.PhaseConnecting
+		} else if found.Status.State == cephv1.ClusterStateConnected {
+			sc.Status.Phase = statusutil.PhaseConnected
+		} else {
+			sc.Status.Phase = statusutil.PhaseNotReady
+		}
+
+		if err = r.client.Status().Update(context.TODO(), sc); err != nil {
+			reqLogger.Error(err, "failed to update ignored resource")
```
umangachapagain, line 808: Error message is incorrect. There is no Ignored resource here.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -86,31 +89,14 @@ func (r *ReconcileStorageCluster) Reconcile(request reconcile.Request) (reconcil
...
-			err = fmt.Errorf("Storage cluster version (%s) is higher than the OCS Operator version (%s)",
-				instance.Spec.Version, version.Version)
-			reqLogger.Error(err, "Incompatible Storage cluster version")
-			return reconcile.Result{}, err
-		}
-		// if the storage cluster version is less than the OCS Operator version,
-		// just update.
-		instance.Spec.Version = version.Version
+	if instance.Spec.ExternalStorage.Enable {
+		reqLogger.Info("External StorageCluster detected")
```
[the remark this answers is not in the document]
jarrpa: `s/External/external/`

pkg/controller/storagecluster/reconcile.go
```diff
@@ -673,6 +723,9 @@ func (r *ReconcileStorageCluster) ensureCephCluster(sc *ocsv1.StorageCluster, re
 	if err != nil {
 		if errors.IsNotFound(err) {
 			reqLogger.Info("Creating CephCluster")
+			if sc.Spec.ExternalStorage.Enable {
+				reqLogger.Info("External CephCluster detected...")
```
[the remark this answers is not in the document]
jarrpa: `s/External/external/`

pkg/controller/storagecluster/reconcile.go
jarrpa: Good point.

pkg/controller/storagecluster/reconcile.go
jarrpa:
```suggestion
			reqLogger.Error(err, "failed to update external cluster status")
```

pkg/controller/storagecluster/reconcile.go
```diff
@@ -268,7 +219,14 @@ func (r *ReconcileStorageCluster) Reconcile(request reconcile.Request) (reconcil
 		r.ensureCephConfig,
 		r.ensureCephCluster,
 		r.ensureNoobaaSystem,
-	} {
+	}
+	// for external cluster, we have a different set of ensure functions
+	if instance.Spec.ExternalStorage.Enable {
```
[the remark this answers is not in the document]
aruniiird: > Why define it and then override, when we can define once as needed.

This is generally the style/approach to avoid an unnecessary if/else. We could have added an if/else (as you suggested) if we were calling the functions, but here we are merely assigning the function literals (not the return values after calling them).
An if/else may usually clutters the code and even that case we are **not** avoiding any overriding (it will look something like this),
```go
var ensureFs []ensureFunc // here it is by-default initialized to nil
      ensureFs = []ensureFunc{  // we are re-defining either here or
            // Add support for additional resources here
            r.ensureStorageClasses,
            r.ensureCephObjectStores,
            r.ensureCephObjectStoreUsers,
            r.ensureCephBlockPools,
            r.ensureCephFilesystems,

            r.ensureCephConfig,
            r.ensureCephCluster,
            r.ensureNoobaaSystem,
      }
} else {
      ensureFs = []ensureFunc{  // re-defining here
            r.ensureCephCluster,
      }
}
```

I have no reservations. So if you feel that code readability is better by having if/else way, will change it accordingly...

pkg/controller/storagecluster/reconcile.go
aruniiird: Done...

pkg/controller/storagecluster/reconcile.go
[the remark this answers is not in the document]
aruniiird: Changed...

pkg/controller/storagecluster/reconcile.go
```diff
@@ -673,6 +723,9 @@ func (r *ReconcileStorageCluster) ensureCephCluster(sc *ocsv1.StorageCluster, re
 	if err != nil {
 		if errors.IsNotFound(err) {
 			reqLogger.Info("Creating CephCluster")
+			if sc.Spec.ExternalStorage.Enable {
```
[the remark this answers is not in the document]
aruniiird: Changed it as below,
```go
    reqLogger.Info("Creating external CephCluster")
} else {
    reqLogger.Info("Creating CephCluster")
}
```

pkg/controller/storagecluster/reconcile.go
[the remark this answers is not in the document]
aruniiird: Done...

pkg/controller/storagecluster/reconcile.go
aruniiird: Removed the lines...

pkg/controller/storagecluster/reconcile.go
aruniiird: Yup, removed the line...

pkg/controller/storagecluster/reconcile.go
aruniiird: Yup, made the changes accordingly...

pkg/controller/storagecluster/reconcile.go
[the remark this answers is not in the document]
aruniiird: As most of us feel if/else preferred way to go, made the changes... Please take a look

PR #430, opened 2020-02-28
Reduce TargetSizeRatio from .5 to .49 to avoid HEALTH_WARN

obnoxxx:
Currently, due to a floating point imprecision in ceph, we get a
"HEALTH_WARN 1 subtrees have overcommitted pool target_size_ratio"
after deplyoing with the TargetSizeRatio of .5 (0.5 + 0.5 > 1.0..).

This will be fixed in future Ceph versions
(https://github.com/ceph/ceph/pull/33035).

In order to avoid the warning until we have the updated Ceph,
this patch lowers the ratio to 0.49.

https://bugzilla.redhat.com/show_bug.cgi?id=1807950


jarrpa:
Putting this on hold to reduce load on the CI cluster while #431 merges.

/hold

PR #431, opened 2020-02-28
Unblock functional tests CI

jarrpa:
This PR contains two commits that are separate but necessary fixes to work around current blockers in the CI.

First, we are skipping the upgrade tests because something seems to be consistently wrong with them lately. I am marking them as "pending" so we can debug them later.

Second, there was a change recently in the OCP CI infrastructure that caused our CI to break because we were hard-coding the image registry URL. This commit updates our CI scripts following this PR as a guideline: https://github.com/openshift/knative-serving/pull/400/files


jarrpa:
This is now outdated, closing.

PR #433, opened 2020-03-02
Adds new OLM annotations to CSV

umangachapagain:
https://bugzilla.redhat.com/show_bug.cgi?id=1799056

Adding 2 new recommended annotations to OCS CSV.

"operatorframework.io/suggested-namespace" is used to
suggest namespace for the OCS Operator to be installed in.
Admin is still allowed to select other namespace.

"operatorframework.io/cluster-monitoring=true" is used to
inform admin that monitoring is supported by the operator
and provide them with a checkbox to opt-in, following
which the namespace will be labeled appropriately.
Admin can opt-out but will be unable to use Storage
Dashboard & other OCS monitoring & alerting features.


umangachapagain:
> My only comment here is that it could be broken into two commits to facilitate any potential future backports. Still, this seems perfectly clean and reasonable as is.

Given that the commit is very simple & small, and will not require a backport, I'd keep it as it is.

PR #434, opened 2020-03-04
Add output of oc commands in must-gather

ashishkumsingh:
Signed-off-by: Ashish Singh <assingh@redhat.com>


ashishkumsingh:
> I like the changes themselves, but the commit message is not good enough. You're not necessarily adding anything new, you are updating/improving the current commands we run. Please update the commit message to both update the subject and add a brief explanation why you're doing this.

Thanks for reviewing Jose.
Have updated the commit message.

PR #438, opened 2020-03-09
Hide ObjectBuckets & ObjectBucketClaims APIs

umangachapagain:
Follow Up for https://github.com/openshift/ocs-operator/pull/424

This is required to hide Internal Objects in UI.


guymguym:
Hey @umangachapagain Is it now needed because before as `required` it wasn't shown?
And now as `owned` it needs to be explicitly hidden?

umangachapagain:
> Hey @umangachapagain
> Is it now needed because before as `required` it wasn't shown?
> And now as `owned` it needs to be explicitly hidden?

Yes.
When "required", UI does not list it under `Provided APIs`.
When "owned", UI will list it under `Provided APIs` and expose to the user. We don't want user to directly use this (at least for now).

PR #441, opened 2020-03-10
Bug 1798571: ob and obc will move to owned instead of required

jackyalbo:
Cherry pick commit from PR #424
As part of not being dependent on community operators BZ1798571


obnoxxx:
We need this against 4.4.
Closing for 4.3

PR #442, opened 2020-03-12
Expose noobaa endpoint resource management to OCS operator

### LINKED ISSUE #403: Support override noobaa-endpoint resources in StorageCluster.spec.resources map
NooBaa endpoint pods have a spec to override the default resources in the noobaa CR spec:
https://github.com/noobaa/noobaa-operator/blob/8fb75596137a9359e4c00cc7c6f84eb3c5e5d8a9/pkg/apis/noobaa/v1alpha1/noobaa_types.go#L153

Need to add this to the StorageCluster.spec.resources map and reconcile it to the NooBaa CR - 
https://github.com/openshift/ocs-operator/blob/b3bf10e90e5cb14f5509c9c049f1eeba0e16855b/pkg/controller/storagecluster/noobaa_system_reconciler.go#L103-L105

CC @nimrod-becker @nb-ohad

nb-ohad:
A NooBaa CR can describe memory and CPU resource management (requests/limits) for deployed NooBaa endpoints.

This PR exposes these options in the StorageCluster CR and add reconcile code to update the NooBaa CR accordingly  

Fixes #403

### DIFF
--- a/pkg/controller/defaults/resources.go
+++ b/pkg/controller/defaults/resources.go
@@ -9,7 +9,7 @@ var (
 	// DaemonResources map contains the default resource requirements for the
 	// various OCS daemons
 	DaemonResources = map[string]corev1.ResourceRequirements{
-		"osd": corev1.ResourceRequirements{
+		"osd": {
 			Requests: corev1.ResourceList{
 				corev1.ResourceCPU:    resource.MustParse("1"),
 				corev1.ResourceMemory: resource.MustParse("4Gi"),
@@ -19,7 +19,7 @@ var (
 				corev1.ResourceMemory: resource.MustParse("8Gi"),
 			},
 		},
-		"mon": corev1.ResourceRequirements{
+		"mon": {
 			Requests: corev1.ResourceList{
 				corev1.ResourceCPU:    resource.MustParse("1"),
 				corev1.ResourceMemory: resource.MustParse("2Gi"),
@@ -29,7 +29,7 @@ var (
 				corev1.ResourceMemory: resource.MustParse("2Gi"),
 			},
 		},
-		"mds": corev1.ResourceRequirements{
+		"mds": {
 			Requests: corev1.ResourceList{
 				corev1.ResourceCPU:    resource.MustParse("3"),
 				corev1.ResourceMemory: resource.MustParse("8Gi"),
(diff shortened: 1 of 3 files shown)


pkg/controller/defaults/resources.go
```diff
@@ -79,10 +79,20 @@ var (
 				corev1.ResourceMemory: resource.MustParse("4Gi"),
 			},
 		},
-		"noobaa-db-vol": corev1.ResourceRequirements{
+		"noobaa-db-vol": {
 			Requests: corev1.ResourceList{
 				corev1.ResourceStorage: resource.MustParse("50Gi"),
 			},
 		},
+		"noobaa-endpoint": {
```
guymguym: @nb-ohad should we re-adjust the noobaa-core resources now that endpoints have split from it?

nb-ohad: @guymguym I was thinking this is the readjusted resources, Please correct me If I am wrong.

guymguym: Better check it

guymguym:
@nb-ohad @nimrod-becker bump - what is missing here?

pkg/controller/defaults/resources.go
```diff
@@ -9,7 +9,7 @@ var (
 	// DaemonResources map contains the default resource requirements for the
 	// various OCS daemons
 	DaemonResources = map[string]corev1.ResourceRequirements{
-		"osd": corev1.ResourceRequirements{
```
jarrpa, line 12: Why are we making these changes?

pkg/controller/defaults/resources.go
```diff
@@ -79,10 +79,20 @@ var (
...
+		"noobaa-endpoint": {
+			Requests: corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("1"),
+				corev1.ResourceMemory: resource.MustParse("2Gi"),
+			},
+			Limits: corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("1"),
+				corev1.ResourceMemory: resource.MustParse("2Gi"),
+			},
+		},
```
jarrpa, line 96: If the above changes are valid, move this to a separate commit.

nb-ohad: It is in a separate commit, Do you mean a separate PR?

pkg/controller/defaults/resources.go
nb-ohad: Its a change in `gofmt` that remove unnecessary type declarations

jarrpa: Ah, okay.

pkg/controller/defaults/resources.go
jarrpa: The issue is that the gofmt changes are unrelated to the NooBaa endpoint changes. The gofmt changes should go into their own commit, with a message explaining why they're happening.

PR #443, opened 2020-03-12
 Update OCS image registry and reenable the upgrade test

rohantmp:

Seeing if this will fix the upgrade test for now while a better solution is reached.


rohantmp:
Just sent the PR to test on prow based on my fuzzy understanding of a slack conversation. More research will follow.

rohantmp:
Ok, just realized image registry format is wrong. Will get back to this

rohantmp:
/hold
it's breaking againg

umangachapagain:
> /hold
> it's breaking again. Pushing with debug messages to see what the difference is.

This should be working now.
Also, the master is tracking 4.5, so upgrade from 4.3 directly to 4.5 might not be a good idea. Use 4.4. Maybe export a variable `PreviousRelease` and upgrade from it to the PR on top of `CurrentRelease`

hack/common.sh
```diff
@@ -56,8 +56,8 @@ MUST_GATHER_FULL_IMAGE_NAME="${IMAGE_REGISTRY}/${REGISTRY_NAMESPACE}/${MUST_GATH
 OCS_CLUSTER_UNINSTALL="${OCS_CLUSTER_UNINSTALL:-true}"
 OCS_SUBSCRIPTION_CHANNEL=${OCS_SUBSCRIPTION_CHANNEL:-alpha}
 
-UPGRADE_FROM_OCS_REGISTRY_IMAGE="${UPGRADE_FROM_OCS_REGISTRY_IMAGE:-quay.io/ocs-dev/ocs-registry:4.2.0}"
-UPGRADE_FROM_OCS_SUBSCRIPTION_CHANNEL="${UPGRADE_FROM_OCS_SUBSCRIPTION_CHANNEL:-$OCS_SUBSCRIPTION_CHANNEL}"
+UPGRADE_FROM_OCS_REGISTRY_IMAGE="${UPGRADE_FROM_OCS_REGISTRY_IMAGE:-$CATALOG_FULL_IMAGE_NAME}"
+UPGRADE_FROM_OCS_SUBSCRIPTION_CHANNEL="${UPGRADE_FROM_OCS_SUBSCRIPTION_CHANNEL:-4.4.0}"
```
umangachapagain, line 60: Does this mean, we'll start from `4.4.0` channel of latest `ocs-registry` ?
Will this upgrade to the latest image built by CI ?

umangachapagain: Since we are using hardcoded external image registry, this is still prone to error.
We should follow best practices given by OpenShift team.

But, this change will be bigger.

umangachapagain:
Re-running to see if it's stable.

/test ocs-operator-e2e-aws

hack/common.sh
rohantmp: Since we are using hardcoded external image registry, this is still prone to error.

Oops!

rohantmp: The old image was also produced and mirrored by openshift CI automation, but it'll be much better using the same registry.

hack/common.sh
```diff
@@ -56,12 +56,12 @@ MUST_GATHER_FULL_IMAGE_NAME="${IMAGE_REGISTRY}/${REGISTRY_NAMESPACE}/${MUST_GATH
...
 OCS_SUBSCRIPTION_CHANNEL=${OCS_SUBSCRIPTION_CHANNEL:-alpha}
 
-UPGRADE_FROM_OCS_REGISTRY_IMAGE="${UPGRADE_FROM_OCS_REGISTRY_IMAGE:-quay.io/ocs-dev/ocs-registry:4.2.0}"
-UPGRADE_FROM_OCS_SUBSCRIPTION_CHANNEL="${UPGRADE_FROM_OCS_SUBSCRIPTION_CHANNEL:-$OCS_SUBSCRIPTION_CHANNEL}"
-
 # Override the image name when this is invoked from openshift ci
 if [ -n "$OPENSHIFT_BUILD_NAMESPACE" ]; then
 	CATALOG_FULL_IMAGE_NAME="registry.svc.ci.openshift.org/${OPENSHIFT_BUILD_NAMESPACE}/stable:${CATALOG_IMAGE_NAME}"
 	echo "Openshift CI detected, deploying using image $CATALOG_FULL_IMAGE_NAME"
 	MUST_GATHER_FULL_IMAGE_NAME="registry.svc.ci.openshift.org/${OPENSHIFT_BUILD_NAMESPACE}/stable:ocs-must-gather-quay"
```
umangachapagain, line 63: We are using `registry.svc.ci.openshift.org/` here.
@jarrpa Isn't this what the OpenShift guys asked us not to use?

PR #446, opened 2020-03-16
Adds skipRange to CSV

umangachapagain:
Adding skipRange to allow Upgrades without upgrading to intermediate versions.


umangachapagain:
> To make things a bit more explicit, should we provide a variable to encode the minimum of the range to upgrade from, like so:
> 
> ```
> SKIP_MINIMUM="0.0.1"
> SKIP_RANGE=">=${SKIP_MINIMUM} <${CSV_VERSION}"
> ```

Good suggestion. This allows custom skipRange.
Fixed it.

PR #468, opened 2020-03-30
Bug: 1818124  - csi-resizer container image fails to deploy in disconnected environment behind Nexus proxy

b-ranto:
This allows us to override the image and point it to its downstream counterpart while fixing the issue with disconnected installs in the same time.


jarrpa:
Closing as dup of #467

PR #475, opened 2020-04-20
Annotation feature gate for Console

umangachapagain:
An annotation containing an array of unique identifiers for
console features is added to enable/disable them as needed.


bipuladh:
@umangachapagain looks like we might have some issues in UI implementing these values.

cloudbehl:
@bipuladh can you describe what is the issues you are seeing? https://github.com/openshift/console/pull/5079

bipuladh:
@cloudbehl I have described the issue at https://github.com/openshift/console/pull/5079/files#r411286609

PR #477, opened 2020-04-20
creates default rook-ceph-operator-config ConfigMap

umangachapagain:
Creating a rook-ceph-operator-config ConfigMap with default
operator configs which is picked up by rook. We do not
update this ConfigMap and is left upto the admin. To reset
to defaults, configmap must be deleted and reconcile must
be retriggered by deleting ocsinit object.

### DIFF
--- a/pkg/controller/ocsinitialization/ocsinitialization_controller.go
+++ b/pkg/controller/ocsinitialization/ocsinitialization_controller.go
@@ -6,6 +6,8 @@ import (
 	"os"
 	"reflect"
 
+	"github.com/openshift/ocs-operator/pkg/controller/defaults"
+
 	secv1client "github.com/openshift/client-go/security/clientset/versioned/typed/security/v1"
 	ocsv1 "github.com/openshift/ocs-operator/pkg/apis/ocs/v1"
 	statusutil "github.com/openshift/ocs-operator/pkg/controller/util"
@@ -32,7 +34,11 @@ var watchNamespace string
 
 const wrongNamespacedName = "Ignoring this resource. Only one should exist, and this one has the wrong name and/or namespace."
 
-const rookCephToolDeploymentName = "rook-ceph-tools"
+const (
+	rookCephToolDeploymentName = "rook-ceph-tools"
+	// This name is predefined by Rook
+	rookCephOperatorConfigName = "rook-ceph-operator-config"
+)
 
 // InitNamespacedName returns a NamespacedName for the singleton instance that
 // should exist.
@@ -306,6 +312,17 @@ func (r *ReconcileOCSInitialization) Reconcile(request reconcile.Request) (recon
 		return reconcile.Result{}, err
 	}
 
+	if !instance.Status.RookCephOperatorConfigCreated {
+		// if true, no need to ensure presence of ConfigMap
+		// if false, ensure ConfigMap and update the status
+		err = r.ensureRookCephOperatorConfig(instance)
+		if err != nil {
+			reqLogger.Error(err, "Failed to process %s Configmap", rookCephOperatorConfigName)
+			return reconcile.Result{}, err
+		}
+		instance.Status.RookCephOperatorConfigCreated = true
+	}
+
 	reason := ocsv1.ReconcileCompleted
 	message := ocsv1.ReconcileCompletedMessage
 	statusutil.SetCompleteCondition(&instance.Status.Conditions, reason, message)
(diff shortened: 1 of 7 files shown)


pkg/controller/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -317,3 +321,44 @@ func (r *ReconcileOCSInitialization) Reconcile(request reconcile.Request) (recon
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      rookCephOperatorConfigName,
+			Namespace: namespace,
+		},
+	}
+	data := make(map[string]string)
+	data["CSI_PROVISIONER_NODE_AFFINITY"] = "cluster.ocs.openshift.io/openshift-storage="
+	data["CSI_PROVISIONER_TOLERATIONS"] = defaultCSIToleration
+	data["CSI_PLUGIN_TOLERATIONS"] = defaultCSIToleration
+	data["CSI_ENABLE_SNAPSHOTTER"] = "false"
```
travisn, line 343: If these settings are never going to be reconciled again, we can only add settings that truly should be configured by the user. Specifically, whether to enable the snapshotter should not be set here. We should keep that as an env variable in the operator spec. Otherwise, the snapshotter will be disabled and will not be enabled in the future when we expect it to be enabled. It is still possible for the user to override the default, but we can't set a default where OCS needs to change the default in the future.

jarrpa: Just to verify: I imagine ConfigMap settings override env vars in the Deployment for Rook-Ceph?

travisn: correct, the CM settings override the env vars

umangachapagain:
> This seems like it belongs in the StorageClusterInit, since we will need a unique one for each StorageCluster.
>
It's actually one per rook operator.
For example: CSI drivers will be installed once and needs to be configured only once. Also, since the Env Var is attached to rook operator deployment, we can't have more than one CM per namespace.

pkg/controller/ocsinitialization/ocsinitialization_controller.go
umangachapagain: Makes sense. Will make the change.

Just to clear it out, even though we are creating this ConfigMap, I have retained the Env Vars in the deployment in CSV as a fall-back.

Meaning, if user wants to override anything, that can be done via CM. But, if user removes any of the defaults, the Env Var will be picked and applied to ensure that our recommended defaults are honoured.

If that is not desired, the setting key's value can be set to `nil` and we'll not fall-back.

jarrpa:
> > This seems like it belongs in the StorageClusterInit, since we will need a unique one for each StorageCluster.
> 
> It's actually one per rook operator.
> For example: CSI drivers will be installed once and needs to be configured only once. Also, since the Env Var is attached to rook operator deployment, we can't have more than one CM per namespace.

....right, duh. :laughing:

pkg/controller/ocsinitialization/ocsinitialization_controller.go
umangachapagain: Fixed this.

pkg/controller/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -317,3 +321,43 @@ func (r *ReconcileOCSInitialization) Reconcile(request reconcile.Request) (recon
...
+	rookCephOperatorConfig := &corev1.ConfigMap{}
+	err := r.client.Get(context.TODO(), types.NamespacedName{Name: rookCephOperatorConfigName, Namespace: initialData.Namespace}, rookCephOperatorConfig)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			// If it does not exist, create a ConfigMap with default settings
+			return r.client.Create(context.TODO(), newRookCephOperatorConfig(initialData.Namespace))
+		}
+		return err
+	}
+	// If it already exists, do not update. It is upto the user to
```
travisn, line 358:
```suggestion
	// If it already exists, do not update. It is up to the user to
```

pkg/controller/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -317,3 +321,43 @@ func (r *ReconcileOCSInitialization) Reconcile(request reconcile.Request) (recon
...
+	data := make(map[string]string)
+	data["CSI_PROVISIONER_NODE_AFFINITY"] = "cluster.ocs.openshift.io/openshift-storage="
+	data["CSI_PROVISIONER_TOLERATIONS"] = defaultCSIToleration
+	data["CSI_PLUGIN_TOLERATIONS"] = defaultCSIToleration
+	config.Data = data
+
+	return config
+}
+
+func (r *ReconcileOCSInitialization) ensureRookCephOperatorConfig(initialData *ocsv1.OCSInitialization) error {
```
travisn, line 348: Is this method called somewhere? I don't see where the new code is triggered.

mykaul:
@umangachapagain - what's the status of this PR?

umangachapagain:
> @umangachapagain - what's the status of this PR?
>
#520 was blocking this PR. Since, it is merged now, I'll be updating this PR today and should be merged by EOD.

pkg/controller/ocsinitialization/ocsinitialization_controller.go
umangachapagain: Forgot to add that to the commit. Fixed it.

pkg/controller/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -315,3 +325,43 @@ func (r *ReconcileOCSInitialization) Reconcile(request reconcile.Request) (recon
...
+  effect: NoSchedule`
+
+	config := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      rookCephOperatorConfigName,
+			Namespace: namespace,
+		},
+	}
+	data := make(map[string]string)
+	data["CSI_PROVISIONER_NODE_AFFINITY"] = "cluster.ocs.openshift.io/openshift-storage="
```
jarrpa, line 344: Shouldn't this be generated from `NodeAffinityKey`?

pkg/controller/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -315,3 +325,43 @@ func (r *ReconcileOCSInitialization) Reconcile(request reconcile.Request) (recon
 
 	return reconcile.Result{}, err
 }
+
+// returns a ConfigMap with default settings for rook-ceph operator
+func newRookCephOperatorConfig(namespace string) *corev1.ConfigMap {
+	var defaultCSIToleration string = `
```
jarrpa, line 331: Shouldn't this be generated from `NodeTolerationKey`?

pkg/controller/ocsinitialization/ocsinitialization_controller.go
umangachapagain: fixed.

pkg/controller/ocsinitialization/ocsinitialization_controller.go
umangachapagain: fixed.

PR #480, opened 2020-04-21
Removed hardcoded portable=true

rohan47:
`portable` was hardcoded to true in case placement is none which resulted in `portable=true` even if in the storageCluster CR it was specified `false`


pkg/controller/storagecluster/reconcile.go
```diff
@@ -1000,7 +1000,6 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster) []rook.StorageClassDevi
 				(&in).DeepCopyInto(&placement)
 
 				if len(topologyKeyValues) >= replica {
-					portable = true
```
travisn, line 1003: How about we also remove the variable on line 996 and then on line 1026 we use `ds.Portable` directly.

jarrpa: I suppose we can, yeah. We can just introduce the variable again later if we have need.

jarrpa:
An improved commit message, but there is a typo and could use a bit of rewording. Something like:
```
The OCP console used to require us to hardcode portable=true. Now it explicitly
sets portable=true so there is no longer need to hardcode it in the operator
code. Removing the hardcoded value from the operator code also makes it possible
to set the value of portable without setting a placement as a workaround.
```

umangachapagain:
@rohan47 You need to run `make gen-latest-csv`. Seems like it is not accepted by CI.
Also, [tests](https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_ocs-operator/480/pull-ci-openshift-ocs-operator-master-ocs-operator-ci/824#1:build-log.txt%3A1005) might need to be fixed.

rohan47:
@umangachapagain looks lik even after `make gen-latest-csv` its not accepted by CI

obnoxxx:
@rohan47  wrote:
>  looks like even after make gen-latest-csv it's not accepted by CI

Did you also commit the `hack/latest-csv-checksum.md5` file (or so) and force-push to this pr's branch?

rohan47:
> @rohan47 wrote:
> 
> > looks like even after make gen-latest-csv it's not accepted by CI
> 
> Did you also commit the `hack/latest-csv-checksum.md5` file (or so) and force-push to this pr's branch?

Yes

obnoxxx:
Functional test failing here: https://github.com/openshift/ocs-operator/pull/480/files#diff-3ab4add16e230d7d70aa431f83ce9551R491

obnoxxx:
> >    Did you also commit the hack/latest-csv-checksum.md5 file (or so) and force-push to this pr's branch?

> Yes

Strange... I get a diff when i run `make gen-latest-csv` locally and  the error to run ocs-operator-ci goes away once I commit it.

rohan47:
> > > Did you also commit the hack/latest-csv-checksum.md5 file (or so) and force-push to this pr's branch?
> 
> > Yes
> 
> Strange... I get a diff when i run `make gen-latest-csv` locally and the error to run ocs-operator-ci goes away once I commit it.

Its the timestamp in  `deploy/olm-catalog/ocs-operator/4.5.0/ocs-operator.v4.5.0.clusterserviceversion.yaml` and `hack/latest-csv-checksum.md5` files. Not sure what is wrong here.

obnoxxx:
@rohan47 FYI I just updated the PR with my results of running `make gen-latest-csv` (squashed into your patch) to see if that fixes the issue. It passes `make ocs-operator-ci` locally for me.

I had a similar issue with another PR recently, and IIRC, the problem went away after I had run `git clean -xdf`... Let's see how this goes.

PR #481, opened 2020-04-22
A job template to remove failed disk

xenolinux:
This job template will manage to avoid the usage of the toolbox in order to use complex ceph commands for removing failed osd. A script can be executed with Ceph commands as if running in the toolbox, but will instead run in a job where the logs can be collected and analyzed separately.


pkg/controller/storagecluster/reconcile.go
```diff
@@ -1037,3 +1038,96 @@ func getMonCount(nodeCount int) int {
...
+							VolumeMounts: []corev1.VolumeMount{
+								corev1.VolumeMount{
+									MountPath: "/etc/ceph",
+									Name:      "ceph-config",
+									ReadOnly:  true,
+								},
+							},
+							Command: []string{"/bin/bash", "-c"},
+							Args: []string{
+								"ceph osd out osd.FAILED_OSD_ID",   //this command marks osd out
```
travisn, line 1092: Does it work with the two commands as separate arguments in the string slice? In my testing the commands were a single multi-line parameter as seen [here](https://github.com/rook/rook/blob/master/cluster/examples/kubernetes/ceph/toolbox-job.yaml#L38-L43), but maybe this way works fine too.

must-gather/collection-scripts/gather
```diff
@@ -32,19 +32,20 @@ mkdir -p ${BASE_COLLECTION_PATH}/oc_output/
 commands=()
 
 # collect oc output of OC commands
-commands+=("version")
```
travisn, line 35: These changes aren't related to the new template, right? Seems like we should revert these changes.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -1037,3 +1038,96 @@ func getMonCount(nodeCount int) int {
 
 	return count
 }
+
+func newCleanupJob(sc *ocsv1.StorageCluster, RookImage string, OsdID int, reqLogger logr.Logger) *batchv1.Job {
```
travisn, line 1042: Remaining work items:
- Wrap the Job in an OpenShift template.
- Call this method from the reconcile loop so it is kept up-to-date. 
- When there is an upgrade we need to ensure the template is updated. For example, in the 4.5 release this template will likely change once we have a simplified rook job available.

must-gather/collection-scripts/gather
obnoxxx: Indeed, there seems to be a wird openshift-bot merge commit https://github.com/openshift/ocs-operator/pull/481/commits/5bb89d20ad5a8092582adb8ebea182fca8ad691d called "aa" from Mar 05 in this PR.

pkg/controller/storagecluster/reconcile.go
ashishranjan738: I think we changed it to string slice.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -1152,3 +1156,129 @@ func getMonCount(nodeCount int) int {
 
 	return count
 }
+
+//ensureJobTemplates returns spec of a job template
```
umangachapagain, line 1160: I see it only returns `error`??

pkg/controller/storagecluster/reconcile.go
```diff
@@ -1152,3 +1156,129 @@ func getMonCount(nodeCount int) int {
...
+func (r *ReconcileStorageCluster) ensureJobTemplates(sc *ocsv1.StorageCluster, reqLogger logr.Logger) error {
```
umangachapagain, line 1161: This func looks way too complicated than it needs to be.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -1152,3 +1156,129 @@ func getMonCount(nodeCount int) int {
...
+	osdCleanUpTemplate := &openshiftv1.Template{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      sc.Name + "-osd-removal",
+			Namespace: sc.Namespace,
+		},
+	}
+	_, err := controllerutil.CreateOrUpdate(context.TODO(), r.client, osdCleanUpTemplate, func() error {
```
umangachapagain, line 1168: Can we use `r.client.Create(context.TODO(), types.NamespacedName{...}, osdCleanUpTemplate)` for simplicity?

pkg/controller/storagecluster/reconcile.go
```diff
@@ -1152,3 +1156,129 @@ func getMonCount(nodeCount int) int {
...
+		osdCleanUpTemplate.Parameters = []openshiftv1.Parameter{
+			{
+				Name:     "ROOK_IMAGE",
+				Required: true,
+			},
+			{
+				Name:     "FAILED_OSD_ID",
+				Required: true,
+			},
+		}
```
umangachapagain, line 1183: Can we move this to part where you initialize `osdCleanUpTemplate` ?
This doesn't look like it needs to be separate.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -1152,3 +1156,129 @@ func getMonCount(nodeCount int) int {
...
+		return controllerutil.SetControllerReference(sc, osdCleanUpTemplate, r.scheme)
+	})
+	if err != nil && !errors.IsAlreadyExists(err) {
+		return err
```
umangachapagain, line 1187: Add more context to the error message. Saves our debugging time later on.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -1152,3 +1156,129 @@ func getMonCount(nodeCount int) int {
...
+	}
+	return nil
+
```
umangachapagain, line 1190: remove empty line.

pkg/controller/storagecluster/reconcile.go
umangachapagain, line 1160:
```suggestion
// ensureJobTemplates returns spec of a job template
```

cmd/manager/main.go
```diff
@@ -120,6 +122,11 @@ func main() {
 		os.Exit(1)
 	}
 
+	if err := openshiftv1.AddToScheme(mgrScheme); err != nil {
+		log.Error(err, "Failed adding openshift/v1 to scheme")
+		os.Exit(1)
+	}
```
umangachapagain, line 128: Why is this required?
I see it is also being used in multiple tests.

I never understood the purpose of this.

jarrpa: If I remember correctly, this is because we need to be able to understand and manipulate the specs for all the API objects. Since Templates are an OpenShift construct, we need to add the entire openshift scheme.

pkg/controller/storagecluster/reconcile.go
jarrpa: That is the ensureFunc template, so yes that's all i can return.

pkg/controller/storagecluster/reconcile.go
jarrpa: Specifically, something like `failed to create Template: %v` where you pass `err` as the variable.

pkg/controller/storagecluster/reconcile.go
jarrpa: We should do proper reconciliation and make sure we either Create or Update as appropriate. This seems like an appropriate use of this helper function, and we should probably consider using it elsewhere... unless there's a reason not to? The alternative is certainly more complicated: https://github.com/openshift/ocs-operator/blob/master/pkg/controller/storagecluster/reconcile.go#L727-L763

pkg/controller/storagecluster/reconcile.go
jarrpa: Well, this is a mutating function... my guess would be that the fetched object gets injected into `osdCleanupTemplate`, so then we have to modify whatever is set to make sure it is reconciled properly. @xenolinux is that correct?

xenolinux: Yes. That's right

pkg/controller/storagecluster/reconcile.go
xenolinux: ack

pkg/controller/storagecluster/reconcile.go
```diff
@@ -1152,3 +1156,129 @@ func getMonCount(nodeCount int) int {
...
+	}
+	_, err := controllerutil.CreateOrUpdate(context.TODO(), r.client, osdCleanUpTemplate, func() error {
+		osdCleanUpTemplate.Objects = []runtime.RawExtension{
+			{
+				Object: newCleanupJob(sc),
+			},
+		}
+		osdCleanUpTemplate.Parameters = []openshiftv1.Parameter{
+			{
+				Name:     "ROOK_IMAGE",
```
travisn, line 1176: We should be able to use the same image that is already being used for the rook operator. Can we lookup the rook operator image instead of requiring it as a parameter? The only input we should need from the admin is the osd id. The reconcile loop would also update the template whenever there is an upgrade to a new image.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -1152,3 +1156,129 @@ func getMonCount(nodeCount int) int {
...
+							Name:  "script",
+							Image: "${ROOK_IMAGE}",
+							VolumeMounts: []corev1.VolumeMount{
+								{
+									MountPath: "/etc/ceph",
+									Name:      "ceph-config",
+									ReadOnly:  true,
+								},
+							},
+							Command: []string{"/bin/bash", "-c", "# ensure the osd is marked out\nceph osd out osd.3\n# purge the osd\nceph osd purge osd.3\n"},
```
travisn, line 1245: We need to replace the failed id `3` with the param. This is probably leftover from a test.

travisn: For code readability I'd suggest using a golang multiline string

pkg/controller/storagecluster/reconcile.go
ashishranjan738: @xenolinux you need to wrap this around a user-friendly method

ashishranjan738: reframe this statement as the implementation has been changed since it started

pkg/controller/storagecluster/reconcile.go
ashishranjan738: Umm I would still prefer this createOrUpdate method as this has been a recommended way of creating/updating the resources. https://godoc.org/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#CreateOrUpdate

pkg/controller/storagecluster/reconcile.go
ashishranjan738: yup can be done

pkg/controller/storagecluster/reconcile.go
ashishranjan738: right can be done

pkg/controller/storagecluster/reconcile.go
ashishranjan738: yup changed to string literal

pkg/controller/storagecluster/reconcile.go
```diff
@@ -1152,3 +1156,134 @@ func getMonCount(nodeCount int) int {
...
+		osdCleanUpTemplate.Objects = []runtime.RawExtension{
+			{
+				Object: newCleanupJob(sc),
+			},
+		}
+		osdCleanUpTemplate.Parameters = []openshiftv1.Parameter{
+			{
+				Name:     "ROOK_IMAGE",
+				Required: true,
+				Value:    os.Getenv("ROOK_CEPH_IMAGE"),
```
travisn, line 1178: - This is an env var to the ocs operator? 
- Instead of the parameter, let's just set the image directly in the template spec. For example, on lines 1237 and 1256 just set it directly instead of setting it in the param here.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -1152,3 +1156,134 @@ func getMonCount(nodeCount int) int {
...
+							VolumeMounts: []corev1.VolumeMount{
+								{
+									MountPath: "/etc/ceph",
+									Name:      "ceph-config",
+									ReadOnly:  true,
+								},
+							},
+							Command: []string{"/bin/bash", "-c",
+								`# ensure the osd is marked out
+								 ceph osd out osd.FAILED_OSD_ID
```
travisn, line 1247: Do we need `${FAILED_OSD_ID}` similar to other places in the template, or why don't we need the brackets here?

pkg/controller/storagecluster/reconcile.go
```diff
@@ -1152,3 +1156,134 @@ func getMonCount(nodeCount int) int {
...
+								 # purge the osd
```
travisn, line 1248: Note that the script will have lots of spaces before these lines. It shouldn't hurt the script, but it would be better if we could avoid that whitespace.

xenolinux: ack

pkg/controller/storagecluster/reconcile.go
xenolinux: yes.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -1152,3 +1156,128 @@ func getMonCount(nodeCount int) int {
 
 	return count
 }
+
+// ensureJobTemplates ensures if the osd removal job template exists
+func (r *ReconcileStorageCluster) ensureJobTemplates(sc *ocsv1.StorageCluster, reqLogger logr.Logger) error {
+	osdCleanUpTemplate := &openshiftv1.Template{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      sc.Name + "-osd-removal",
```
travisn, line 1164: Will the template be called `openshift-storage-osd-removal`? Or what is `sc.Name`? How about if we just use a constant for the template name? Then the documentation can use the well-known template name.

jarrpa: This makes sense, it would simplify the documentation. I propose something simple like `ocs-osd-removal`.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -1152,3 +1156,128 @@ func getMonCount(nodeCount int) int {
...
+		Spec: batchv1.JobSpec{
+			Template: corev1.PodTemplateSpec{
+				Spec: corev1.PodSpec{
+					Volumes: []corev1.Volume{
+						{
+							Name: "mon-endpoint-volume",
+							VolumeSource: corev1.VolumeSource{
+								ConfigMap: &corev1.ConfigMapVolumeSource{
+									LocalObjectReference: corev1.LocalObjectReference{
+										Name: "mon-endpoint-volume",
```
ashishranjan738, line 1216: this name should be `rook-ceph-mon-endpoints`

travisn:
> this name should be rook-ceph-mon-endpoints

@xenolinux @ashishranjan738 With all the changes could you also confirm what testing has been done? For example, that the template is created by the OCS operator and the job can be run successfully with the `oc process` command. I'm assuming this issue with the mon endpoints name was found through testing, just wanted to confirm once it's all working end to end. Thanks!

jarrpa:
@xenolinux Okay, I found the problem I was having: you swashed the wrong commits. Currently, there are three commits:

1. 8ee281f8 -    5 hours ago Changed the name — Original changes
2. 19e88cc5 -    5 hours ago Adds vendor files — vendor files
3. 38963f46 -    5 hours ago Adds reconcile logic for removing osd — CSV and PR edits

At this point, I propose squashing commit 3 into commit 1 as a "fix" (e.g. without the commit message) that way we get all the code and CSV changes in one commit but still leave the vendor files separate. If you feel like it you can split the CSV changes into their own commit, but I'm willing to forego that in the interest of time.

ashishranjan738:
> > this name should be rook-ceph-mon-endpoints
> 
> @xenolinux @ashishranjan738 With all the changes could you also confirm what testing has been done? For example, that the template is created by the OCS operator and the job can be run successfully with the `oc process` command. I'm assuming this issue with the mon endpoints name was found through testing, just wanted to confirm once it's all working end to end. Thanks!

umm I tested it and seems like its working fine

PR #487, opened 2020-04-29
Propagate the cleanup policy

raghavendra-talur:
At the time of deletion of the StorageCluster, we want to check the
cleanup policy and propagate the same to the objects for which
StorageCluster has owner reference.

This commit applies the cleanup policy for rook.

Signed-off-by: Raghavendra Talur <rtalur@redhat.com>


raghavendra-talur:
@egafford Please take a look.

This commit depends on https://github.com/rook/rook/pull/5004 . When I bump up the required version of rook to v.1.3.0 in go.mod to bring the update of that PR, it leads to the following error:

```
$ make deps-update 
go mod tidy && go mod vendor
go: github.com/rook/rook@v1.3.0 requires
        k8s.io/kube-controller-manager@v0.0.0: reading https://proxy.golang.org/k8s.io/kube-controller-manager/@v/v0.0.0.mod: 410 Gone
make: *** [Makefile:39: deps-update] Error 1
```
I think I need some help with the dependency resolution here.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -55,7 +55,9 @@ osd_memory_target_cgroup_limit_ratio = 0.5
 `
 	monCountOverrideEnvVar = "MON_COUNT_OVERRIDE"
 	// EBS represents AWS EBS provisioner for StorageClass
-	EBS StorageClassProvisionerType = "kubernetes.io/aws-ebs"
+	EBS                 StorageClassProvisionerType = "kubernetes.io/aws-ebs"
+	CleanupPolicy                                   = "cleanupPolicy.ocs.openshift.io"
```
jarrpa, line 59: Rename this to `CleanupPolicyLabel`, change the value to have no uppercase letters, and insert a comment explaining what this and the following values are used for.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -55,7 +55,9 @@ osd_memory_target_cgroup_limit_ratio = 0.5
...
+	CleanupPolicyDelete                             = "yes-really-destroy-data"
```
leseb, line 60: Please also add:

```
CleanupDiskPolicyDelete                             = "yes-really-sanitize-disks"
```

pkg/controller/storagecluster/reconcile.go
```diff
@@ -173,6 +175,29 @@ func (r *ReconcileStorageCluster) Reconcile(request reconcile.Request) (reconcil
...
+				cephCluster := &cephv1.CephCluster{}
+				err = r.client.Get(context.TODO(), types.NamespacedName{Name: generateNameForCephCluster(instance), Namespace: instance.Namespace}, cephCluster)
+				if err != nil {
+					if errors.IsNotFound(err) {
+						reqLogger.Info("CephCluster not found")
+					} else {
+						return reconcile.Result{}, fmt.Errorf("Unable to retrive cephCluster: %v", err)
+					}
+				} else {
+					cephCluster.Spec.CleanupPolicy.DeleteDataDirOnHosts = "yes-really-destroy-data"
```
leseb, line 192:
```suggestion
					cephCluster.Spec.CleanupPolicy.DeleteDataDirOnHosts = CleanupPolicyDelete
```

leseb: Please add:

```
cephCluster.Spec.CleanupPolicy.SanitizeDiskSpec.SanitizeDiskConfirmation = CleanupDiskPolicyDelete
```

leseb:
@raghavendra-talur It looks like disk sanitizing is out of the picture, I think @jarrpa can confirm that. So you might have to ignore my previous review. Thanks.

obnoxxx:
@leseb, @raghavendra-talur, I don't think disk sanitizing is completely out of the picture.
But we need to clearly separate between two aspects:
1. Removing the tree of all objects that we (our operators) created under the storagecluster CR. We are required to implement two modes:
   * A *graceful* mode where we wait for all consumers to go away / dependent objects to be removed before continuing with the current object. This is to be implemented via appropriately crafted finalizers that ensure the correct ordering and waiting. This graceful mode should be the default.
   * A *forced mode* that should not wait but remove our objects right away, possibly leaving consumers in a bad state. This forced mode should be triggered by a certain parameter (label or annotation or whatever).
2. Wiping the data from the backend disks used for OSDs and MONs. Needs to be controllable independently of the graceful vs forceful mode. We need a separate control (label or annotation or whatever) to control it. There are also (at least two) extents to which we wipe/sanitize:
   * The absolute minimum requirement is to be able to re-install on the same set of disks.  This would be achieved by just wiping the metadata from the disks.
   * Wip all data from the disks.

I think /1 needs to be implemented: Currently, not all of our created objects are removed. This needs to be fixed. The /2 is also important, but somewhat independent, and maybe not quite as MVP. But since rook can already do it, or almost do it, by virtue of https://github.com/rook/rook/pull/5452, we should take it into account!

leseb:
> @leseb, @raghavendra-talur, I don't think disk sanitizing is completely out of the picture.
> But we need to clearly separate between two aspects:
> 
> 1. Removing the tree of all objects that we (our operators) created under the storagecluster CR. We are required to implement two modes:
>    
>    * A _graceful_ mode where we wait for all consumers to go away / dependent objects to be removed before continuing with the current object. This is to be implemented via appropriately crafted finalizers that ensure the correct ordering and waiting. This graceful mode should be the default.
>    * A _forced mode_ that should not wait but remove our objects right away, possibly leaving consumers in a bad state. This forced mode should be triggered by a certain parameter (label or annotation or whatever).
> 2. Wiping the data from the backend disks used for OSDs and MONs. Needs to be controllable independently of the graceful vs forceful mode. We need a separate control (label or annotation or whatever) to control it. There are also (at least two) extents to which we wipe/sanitize:
>    
>    * The absolute minimum requirement is to be able to re-install on the same set of disks.  This would be achieved by just wiping the metadata from the disks.
>    * Wip all data from the disks.
> 
> I think /1 needs to be implemented: Currently, not all of our created objects are removed. This needs to be fixed. The /2 is also important, but somewhat independent, and maybe not quite as MVP. But since rook can already do it, or almost do it, by virtue of [rook/rook#5452](https://github.com/rook/rook/pull/5452), we should take it into account!

I think we need @jarrpa to chime in because his last comment (on another communication channel) makes it clear that disk sanitizing is the admin responsibility...

travisn:
The primary goal of all of this is to be able to re-install without additional steps. If additional steps are needed, it is essentially a bug in the uninstall. 

With LSO, the interesting question is if Rook should attempt to wipe the devices if it can find them. Since the admin created the LSO PVs, the argument could be made that they should wipe them, or at least that LSO should own wiping them when the PV is requested for deletion. But unless there is a clear story for LSO to wipe the PVs, let's test if we can just have Rook attempt the cleanup that is already being implemented with https://github.com/rook/rook/pull/5452. But IMO we should delay this feature for 4.6 since the upstream Rook PR needs more discussion.

obnoxxx:
@leseb writes:
> I think we need @jarrpa to chime in because his last comment (on another communication channel) makes it clear that disk sanitizing is the admin responsibility...

I don't agree. Along the lines what @travisn argued as well:

* Whoever creates a PV is responsible for deleting its content before deleting the PV, if they want the data to be gone.
* ==> So for the PVs that *we* created (rook created OSD backing PVs, MON PVs, noobaa backing store PVs) we are also responsible for removing the content that we put into them, if that's our policy.

According to the original requirement, wiping should not be done by default but triggered by a switch. And we'd have two levels of deletion:
* deleting metadata, without caring if the data is really gone from the disks ==> this is sufficient for re-install on the same disks
* deleting data, so that the data can't be read any more from the disks ==> Might be required by people putting sensitive data on the disks.

If this needs more discussion, it's fine (imho) to delay the wiping to a later release. But we can't push that responsibility away. It's ours. :smiley:

pkg/controller/storagecluster/reconcile.go
leseb: @jarrpa does the ocs-op have the Rook types for cephv1?

pkg/controller/storagecluster/reconcile.go
leseb: Forget this.

pkg/controller/storagecluster/reconcile.go
leseb: Forget this.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -173,6 +175,29 @@ func (r *ReconcileStorageCluster) Reconcile(request reconcile.Request) (reconcil
 		if phaseErr != nil {
 			reqLogger.Error(phaseErr, "Failed to set PhaseDeleting")
 		}
+
+		// Check if the storagecluster is labelled for data deletion, if yes
+		// label the child resources for data deletion too before issuing delete
+		if v, found := instance.ObjectMeta.Labels[CleanupPolicy]; found {
+			if strings.Compare(v, CleanupPolicyDelete) == 0 {
```
travisn, line 182:
```suggestion
			if v == CleanupPolicyDelete {
```

pkg/controller/storagecluster/reconcile.go
travisn: To be clear, we still need the following set:
```
cephCluster.Spec.CleanupPolicy.Confirmation = CleanupPolicyDelete
```
The `Confirmation` field is new in rook, so you'll need to update the reference for the latest rook to pick it up.

pkg/controller/storagecluster/reconcile.go
travisn, line 59:
```suggestion
	CleanupPolicyLabel                              = "cleanup.ocs.openshift.io"
```

pkg/controller/storagecluster/reconcile.go
```diff
@@ -172,6 +176,29 @@ func (r *ReconcileStorageCluster) Reconcile(request reconcile.Request) (reconcil
 		if phaseErr != nil {
 			reqLogger.Error(phaseErr, "Failed to set PhaseDeleting")
 		}
+
+		// Check if the storagecluster is labelled for data deletion, if yes
+		// label the child resources for data deletion too before issuing delete
+		if v, found := instance.ObjectMeta.Labels[CleanupPolicyLabel]; found {
+			if v == CleanupPolicyDelete {
```
leseb, line 183: Why not using the type instead of the new constant?

pkg/controller/storagecluster/reconcile.go
```diff
@@ -56,6 +56,10 @@ osd_memory_target_cgroup_limit_ratio = 0.5
 	monCountOverrideEnvVar = "MON_COUNT_OVERRIDE"
 	// EBS represents AWS EBS provisioner for StorageClass
 	EBS StorageClassProvisionerType = "kubernetes.io/aws-ebs"
+	// If the CleanupPolicyLabel is set to the delete policy, ocs-operator will
+	// pass the information to rook to destroy the data.
+	CleanupPolicyLabel  = "cleanup.ocs.openshift.io"
+	CleanupPolicyDelete = "yes-really-destroy-data"
```
leseb, line 62: I don't think it's needed if you use the type

pkg/controller/storagecluster/reconcile.go
```diff
@@ -172,6 +176,29 @@ func (r *ReconcileStorageCluster) Reconcile(request reconcile.Request) (reconcil
...
+				cephCluster := &cephv1.CephCluster{}
+				err = r.client.Get(context.TODO(), types.NamespacedName{Name: generateNameForCephCluster(instance), Namespace: instance.Namespace}, cephCluster)
+				if err != nil {
+					if errors.IsNotFound(err) {
+						reqLogger.Info("CephCluster not found")
+					} else {
+						return reconcile.Result{}, fmt.Errorf("Unable to retrive cephCluster: %v", err)
+					}
+				} else {
+					cephCluster.Spec.CleanupPolicy.Confirmation = CleanupPolicyDelete
```
leseb, line 193:
```suggestion
					cephCluster.Spec.CleanupPolicy.Confirmation = cephv1.DeleteDataDirOnHostsConfirmation
```

raghavendra-talur: I have update the code to use the type when referring to rook spec. Done.

pkg/controller/storagecluster/reconcile.go
raghavendra-talur: In this line, we are referring to the ocs-operator cleanup policy and not rook's. I am not sure if it is a good idea to use the type from cephv1 although I am using the same string. 

Let me know if you would still prefer to use the type from ceph instead of a new constant local to ocs-operator.

pkg/controller/storagecluster/reconcile.go
raghavendra-talur: Same as above.

raghavendra-talur:
@travisn @jarrpa @leseb I have one more question

Should the logic be inside the finalizer https://github.com/openshift/ocs-operator/blob/master/pkg/controller/storagecluster/reconcile.go#L175 . It is currently outside the finalizer condition.

raghavendra-talur:
@leseb I missed your other comment:

> Rook-Ceph runs a cleanup job(s) on every host a Ceph daemon was running. So ocs-op should wait for that job cleanup to complete before removing its own finalizer.
> 
> The job labels are:
> 
> ```
> app=rook-ceph-cleanup
> rook-ceph-cleanup=true
> rook_cluster=rook-ceph
> ```
> 
> So you can use that to find the job and loops for each to be completed.

IIUC, the rook-ceph deletion starts only after the ocs-operator returns a nil for the reconcile request in the deletion mode. There is no finalizer waiting on rook as of today. 
Should I explicitly call the delete on rook-ceph and wait for it in the finalizer?

pkg/controller/storagecluster/reconcile.go
leseb: OCS should probably have its own type for this. It's just annoying to have to replicate all the things Rook does this way but it is what it is :).

At this point, I don't have a strong preference but I think a new type for OCS-op would make sense.

leseb:
> @travisn @jarrpa @leseb I have one more question
> 
> Should the logic be inside the finalizer https://github.com/openshift/ocs-operator/blob/master/pkg/controller/storagecluster/reconcile.go#L175 . It is currently outside the finalizer condition.

Because we are looking at doing cleanup before terminating the object then yes it looks like the right place to trigger that code.

leseb:
> @leseb I missed your other comment:
> 
> > Rook-Ceph runs a cleanup job(s) on every host a Ceph daemon was running. So ocs-op should wait for that job cleanup to complete before removing its own finalizer.
> > The job labels are:
> > ```
> > app=rook-ceph-cleanup
> > rook-ceph-cleanup=true
> > rook_cluster=rook-ceph
> > ```
> > 
> > 
> > So you can use that to find the job and loops for each to be completed.
> 
> IIUC, the rook-ceph deletion starts only after the ocs-operator returns a nil for the reconcile request in the deletion mode. There is no finalizer waiting on rook as of today.
> Should I explicitly call the delete on rook-ceph and wait for it in the finalizer?

It'd say yes, we should wait for the cleanup jobs to finish, this is in the ideal scenario. However, I've seen we have a graceful and un-graceful option so others may have a different opinion on this...

pkg/controller/storagecluster/reconcile.go
travisn: Wait, why can't OCS reference the consts or types in Rook? OCS is already referencing the `cephv1` package.

raghavendra-talur: It can, just that it probably not a good idea to. The logic reads like this

```
if ocsCleanupLabel is set to ocsCleanupDeleteValue
then
   edit the CleanupSpec of Rook to cephv1.CleanupPolicyDelete
   edit the CleanupSpec of Noobaa //future
fi

```

It so happens that the ocsCleanupDeleteValue and the cephv1.CleanupPolicyDelete value both are "yes-really-destroy-data"

travisn: Oh right, this really is an OCS setting so agreed we shouldn't reference rook's.

raghavendra-talur:
@jarrpa @leseb @travisn I have updated the patch. There are more changes required in ocs-operator for uninstall and I will bring in the changes in the next PR for whatever is marked TODO below. I have limited this PR to only passing the cleanup policy to Rook. Here is the current state.

1. There are no types on ocs-operator yet. We are using label cleanup.ocs.openshift.io with string value "yes-really-destroy-data".  TODO: make it a type.
2. There is no deletion ordering between StorageCluster and CephCluster today. Once the finalizer for the ocs operator is removed all the resources that it owns get deleted. Only the NooBaa resources are deleted in-order by the use of the finalizer. This patch brings the cleanup policy propagation under the finalizer and therefore ensures that it passed but the deletion of the CephCluster happens *after* the StorageCluster is deleted. TODO: delete the CephCluster before removing the finalizer.


I think I have addressed all the comments on this PR.

obnoxxx:
@raghavendra-talur  thanks for the update!

tests are failing.

```
...
Found 2 lint suggestions; failing.
make: *** [Makefile:136: golint] Error 1 
```

please run `make ocs-operator-ci` locally and check.

leseb:
> @jarrpa @leseb @travisn I have updated the patch. There are more changes required in ocs-operator for uninstall and I will bring in the changes in the next PR for whatever is marked TODO below. I have limited this PR to only passing the cleanup policy to Rook. Here is the current state.
> 
> 1. There are no types on ocs-operator yet. We are using label cleanup.ocs.openshift.io with string value "yes-really-destroy-data".  TODO: make it a type.
> 2. There is no deletion ordering between StorageCluster and CephCluster today. Once the finalizer for the ocs operator is removed all the resources that it owns get deleted. Only the NooBaa resources are deleted in-order by the use of the finalizer. This patch brings the cleanup policy propagation under the finalizer and therefore ensures that it passed but the deletion of the CephCluster happens _after_ the StorageCluster is deleted. TODO: delete the CephCluster before removing the finalizer.
> 
> I think I have addressed all the comments on this PR.

If we get ride of the finalized then there is a risk for an incomplete cleanup. If everything goes away on uninstall (namespace included) then it is really likely that Rook-Ceph metadata cleanup will not enough time to run. So this will leave a bunch of leftovers and will make re-install impossible.

Are we willing to accept that?

obnoxxx:
@leseb writes:
> If we get ride of the finalized then there is a risk for an incomplete cleanup. If everything goes away on uninstall (namespace included) then it is really likely that Rook-Ceph metadata cleanup will not enough time to run. So this will leave a bunch of leftovers and will make re-install impossible.
> 
> Are we willing to accept that?

We currently don't have a finalizer waiting for ceph-cluster to vanish. So this is about adding one to avoid incomplete cleanups.

You are right that we should have the finalizer. But as Talur suggested, this can be the topic of a follow-up PR to establish the finalizer. This PR has the topic to propagate the policy.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -172,7 +179,12 @@ func (r *ReconcileStorageCluster) Reconcile(request reconcile.Request) (reconcil
 		if phaseErr != nil {
 			reqLogger.Error(phaseErr, "Failed to set PhaseDeleting")
 		}
+
 		if contains(instance.GetFinalizers(), storageClusterFinalizer) {
+			err = r.setRookCleanupPolicy(instance, reqLogger)
```
obnoxxx, line 184: This propagates the cleanup policy in the delete part of the reconcile loop.
As discussed earlier today, at least as long as we don't have a finalizer for the CephCluster, this is kind of racy, and we should probably consider (additionall) propagating the policy as part of the normal reconcile code as well, to make it safer. I'd argue that we want the propagation always anyway, not only in delete as it's part of the storagecluster cr.

raghavendra-talur: @leseb The rook documentation says that the CleanupPolicy is to be set only when the deletion is imminent. https://github.com/rook/rook/blob/master/pkg/apis/ceph.rook.io/v1/types.go#L108

@obnoxxx and I were thinking of updating the rook CleanupPolicy in the main reconcile loop, even when the deletion is not imminent.

What do you suggest?

jarrpa: I think it's fine to leave it only in the delete path for now. In particular, we have to consider the scenario where the StorageCluster is labeled and then deleted before ocs-operator reconciles it. This means that you would anyway have a CephCluster without the appropriate cleanup policy that you would need to amend. The cleanup policy itself means nothing prior to the CephCluster being marked for deletion, so updating it just prior to that deletion seems fine. I agree that it's a little racy, but that raciness won't be meaningfully mitigated by calling this logic in the standard reconciliation path.

leseb: Just an FYI and it's also mention in Rook's doc, once the policy is in place all the future reconciles will be skipped.

PR #488, opened 2020-04-29
Add controller-expand-secret and namespace to SC

humblec:
These params in SC are required when we want to expand RBD or CephFS volumes.

Signed-off-by: Humble Chirammal <hchiramm@redhat.com>

### DIFF
--- a/pkg/controller/storagecluster/initialization_reconciler.go
+++ b/pkg/controller/storagecluster/initialization_reconciler.go
@@ -65,7 +65,7 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
 // on first run.
 func (r *ReconcileStorageCluster) newStorageClasses(initData *ocsv1.StorageCluster) ([]*storagev1.StorageClass, error) {
 	persistentVolumeReclaimDelete := corev1.PersistentVolumeReclaimDelete
-	allowVolumeExpansion := false
+	allowVolumeExpansion := true
 	ret := []*storagev1.StorageClass{
 		&storagev1.StorageClass{
 			ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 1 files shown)


humblec:
`AllowVolumeExpansion` is the real knob to turn this feature on/off, which I see as `disabled`. Actually speaking we don't need to "explicitly" disable it as by default its "disabled" though.

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -100,10 +102,12 @@ func (r *ReconcileStorageCluster) newStorageClasses(initData *ocsv1.StorageClust
 				"imageFeatures":             "layering",
 				"csi.storage.k8s.io/fstype": "ext4",
 				"imageFormat":               "2",
-				"csi.storage.k8s.io/provisioner-secret-name":      "rook-csi-rbd-provisioner",
-				"csi.storage.k8s.io/provisioner-secret-namespace": initData.Namespace,
-				"csi.storage.k8s.io/node-stage-secret-name":       "rook-csi-rbd-node",
-				"csi.storage.k8s.io/node-stage-secret-namespace":  initData.Namespace,
+				"csi.storage.k8s.io/provisioner-secret-name":            "rook-csi-rbd-provisioner",
```
Madhu-1, line 105: @humblec in current storageclass `AllowExpansion` is set to false, are you planning to change it to true?

humblec: As mentioned here https://github.com/openshift/ocs-operator/pull/488#issuecomment-621108180 depends on some decisions we can make it `true`, waiting for the same.

humblec: Enabling allowVolumeExpansion .

humblec:
> I don't see the point of adding this if the feature is disabled. Do the CSI drivers complain if they're not set?

Enabling it as we now have a confirmation on feature support.

humblec:
@jarrpa can you please review this ?

humblec:
@jarrpa as discussed yesterday can we backport this to release-4.5 branch too?

humblec:
Let me know if I need to file a backport PR or is the backport handled automatically with a label or so in this repo?

PR #491, opened 2020-04-30
If Docker is not found, use podman to build the containers.

mykaul:
You can still override the command using IMAGE_BUILD_CMD environment variable.
If it's not set, it tries to look for the docker command.
If it's not available, it'll try to use podman.

Tested 'make' on Fedora 32, without Docker.

Signed-off-by: Yaniv Kaul <ykaul@redhat.com>

### DIFF
--- a/hack/common.sh
+++ b/hack/common.sh
@@ -7,7 +7,14 @@ GO111MODULE="on"
 GOPROXY="https://proxy.golang.org"
 GOROOT="${GOROOT:-go env GOROOT}"
 
-IMAGE_BUILD_CMD="${IMAGE_BUILD_CMD:-docker}"
+IMAGE_BUILD_CMD="${IMAGE_BUILD_CMD}"
+if [ -z "$IMAGE_BUILD_CMD" ]; then
+    IMAGE_BUILD_CMD=$(command -v docker || echo "")
+fi
+if [ -z "$IMAGE_BUILD_CMD" ]; then
+    IMAGE_BUILD_CMD=$(command -v podman || echo "")
+fi
+
 IMAGE_RUN_CMD="${IMAGE_RUN_CMD:-${IMAGE_BUILD_CMD} run --rm -it}"
 
 OUTDIR="build/_output"


mykaul:
> While the code changes look good, this should be done in the scripts, not the Makefile, so all the functionality is available when running the scripts outside the Makefile. It shoul dbe fine to put it in `hack/common.sh`.

OK. In other projects[1][2][3] it was on the Makefile, but hack/common.sh works too.

[1] https://github.com/noobaa/noobaa-core/pull/5996/files
[2] https://github.com/ceph/ceph-container/pull/1666/files
[3] https://github.com/rook/rook/pull/5336/files

hack/common.sh
```diff
@@ -7,7 +7,17 @@ GO111MODULE="on"
 GOPROXY="https://proxy.golang.org"
 GOROOT="${GOROOT:-go env GOROOT}"
 
-IMAGE_BUILD_CMD="${IMAGE_BUILD_CMD:-docker}"
```
jarrpa, line 10: I think the issue is that `IMAGE_BUILD_CMD` doesn't exist yet, so setting it in the line below isn't being affcting the subsequent `if` stanzas. Rather than removing this like entirely, only remove the `:-docker` portion so that we inherit any environment variable value without giving it a default.

mykaul: @jarrpa  - but I was under the impression we do inherit the environment variable value even with my code - my code addition is only relevant if IMAGE_BUILD_CMD is not already set?

jarrpa: Not without explicitly marking it. The line above is what says "we will always define this value, and it will inherit from the env var *if it exists*". Without it, the assignments you make below in the `if` stanzas only apply within the scope of those stanzas, since if the env var isn't set it doesn't exist in the broader scope.

mykaul: OK, pushed.

jarrpa:
> The tests are failing because test environment does not have docker or podman.
> This is expected because test pods do not build images, it is done separately.
> 
> We need to skip exiting if we are in CI environment.
> 
> ```
> if [ -n "$OPENSHIFT_BUILD_NAMESPACE" ]; then
> 	echo "Openshift CI detected"
> fi
> ```

From that perspective, I think that if we're not using the docker or podman commands we shouldn't require it to be present. It would make sense to just remove the last `if` stanza.

mykaul:
> > The tests are failing because test environment does not have docker or podman.
> > This is expected because test pods do not build images, it is done separately.
> > We need to skip exiting if we are in CI environment.
> > ```
> > if [ -n "$OPENSHIFT_BUILD_NAMESPACE" ]; then
> > 	echo "Openshift CI detected"
> > fi
> > ```
> 
> From that perspective, I think that if we're not using the docker or podman commands we shouldn't require it to be present. It would make sense to just remove the last `if` stanza.

Removed.

hack/common.sh
```diff
@@ -7,7 +7,14 @@ GO111MODULE="on"
...
+IMAGE_BUILD_CMD="${IMAGE_BUILD_CMD}"
+if [ -z "$IMAGE_BUILD_CMD" ]; then
+    IMAGE_BUILD_CMD=$(command -v docker || echo "")
```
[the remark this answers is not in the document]
jarrpa: This is fine. If there's an error, we want it set to empty string.

hack/common.sh
```diff
@@ -7,7 +7,14 @@ GO111MODULE="on"
...
+fi
+if [ -z "$IMAGE_BUILD_CMD" ]; then
+    IMAGE_BUILD_CMD=$(command -v podman || echo "")
```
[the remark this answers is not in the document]
jarrpa: This is fine. If there's an error, we want it set to empty string.

PR #492, opened 2020-04-30
Add new platform GCP and AZURE

crombus:
https://bugzilla.redhat.com/show_bug.cgi?id=1825994

### DIFF
--- a/pkg/controller/storagecluster/initialization_reconciler.go
+++ b/pkg/controller/storagecluster/initialization_reconciler.go
@@ -125,6 +125,10 @@ func (r *ReconcileStorageCluster) ensureCephObjectStores(instance *ocsv1.Storage
 		r.reqLogger.Info("not creating a CephObjectStore because the platform is AWS")
 		return nil
 	}
+	if platform == PlatformGCP {
+		r.reqLogger.Info("not creating a CephObjectStore because the platform is GCP")
+		return nil
+	}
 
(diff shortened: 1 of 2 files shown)


pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -130,6 +130,11 @@ func (r *ReconcileStorageCluster) ensureCephObjectStores(instance *ocsv1.Storage
                 r.reqLogger.Info("not creating a CephObjectStore because the platform is GCP")
                 return nil
         }
+	if platform == PlatformAzure {
```
sabose, line 133: Instead, can we just check if platform is one of the supported platforms and skip CephObjectStore creation?

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -125,6 +125,14 @@ func (r *ReconcileStorageCluster) ensureCephObjectStores(instance *ocsv1.Storage
...
 		return nil
 	}
+	if platform == PlatformGCP {
+		r.reqLogger.Info("not creating a CephObjectStore because the platform is GCP")
+		return nil
+	}
+	if platform == PlatformAzure {
+                r.reqLogger.Info("not creating a CephObjectStore because the platform is Azure")
+                return nil
+        }
```
jarrpa, line 135: Tiny nit, but the indentation here is wrong. :)

PR #495, opened 2020-05-04
Add get subscription command in must-gather

crombus:
Fixing: https://bugzilla.redhat.com/show_bug.cgi?id=1823886

### DIFF
--- a/must-gather/collection-scripts/gather
+++ b/must-gather/collection-scripts/gather
@@ -41,6 +41,7 @@ commands+=("get pvc -n openshift-storage")
 commands+=("get pv")
 commands+=("get sc")
 commands+=("get infrastructures.config -oyaml")
+commands+=("get subscription -n openshift-storage")
 
 # Run the Collection of Resources to list
 for command in "${commands[@]}"; do


umangachapagain:
Instead of adding individual commands like this, we need to collect OCP must-gather which is useful in multiple cases.

https://github.com/openshift/ocs-operator/pull/494
@jarrpa WDYT?

crombus:
I added this separate because it was bug.   

> Instead of adding individual commands like this, we need to collect OCP must-gather which is useful in multiple cases.
> 
I agree with this. I thought of creating a separate file for oc commands which can be collected in both plain and yaml texts. does this sounds good?

jarrpa:
> > Instead of adding individual commands like this, we need to collect OCP must-gather which is useful in multiple cases.

While yes, OCP must-gather should be run as part of dumping debug info, this change is valid and crucial to the OCS must-gather. Best I know, OCP must-gather doesn't look at the openshift-storage namespace, so we'd never see this subscription anyway. I would also ask to add a fetch for any CatalogSources and OperatorSources, allowing either to return nothing as valid.

> I agree with this. I thought of creating a separate file for oc commands which can be collected in both plain and yaml texts. does this sounds good?

I'm not sure I understand what you mean here. By separate file, do you mean a separate bash script?

umangachapagain:
> > > Instead of adding individual commands like this, we need to collect OCP must-gather which is useful in multiple cases.
> 
> While yes, OCP must-gather should be run as part of dumping debug info, this change is valid and crucial to the OCS must-gather. 
>
If OCP must-gather already gives this info, why do we need OCS must-gather to again collect it?

> Best I know, OCP must-gather doesn't look at the openshift-storage namespace, so we'd never see this subscription anyway. 
>
I think it does. The only reason why it might not do so currently is that, we delete "openshift-storage" namespace before we run must-gather which is silly.
>
> I would also ask to add a fetch for any CatalogSources and OperatorSources, allowing either to return nothing as valid.
>
Instead of adding individual commands for each things, let's just use OCP must-gather? These are all OCP resources and they should be gathering it's logs!!
And, OperatorSources is deprecated since OCP 4.2, and we don't use it. So, no need to collect that for sure.

crombus:
> > > Instead of adding individual commands like this, we need to collect OCP must-gather which is useful in multiple cases.
> 
> While yes, OCP must-gather should be run as part of dumping debug info, this change is valid and crucial to the OCS must-gather. Best I know, OCP must-gather doesn't look at the openshift-storage namespace, so we'd never see this subscription anyway. I would also ask to add a fetch for any CatalogSources and OperatorSources, allowing either to return nothing as valid.
> 
> > I agree with this. I thought of creating a separate file for oc commands which can be collected in both plain and yaml texts. does this sounds good?
> 
> I'm not sure I understand what you mean here. By separate file, do you mean a separate bash script?

yes , a separate file just for oc commands.

jarrpa:
> > While yes, OCP must-gather should be run as part of dumping debug info, this change is valid and crucial to the OCS must-gather.
> 
> If OCP must-gather already gives this info, why do we need OCS must-gather to again collect it?
> 
> > Best I know, OCP must-gather doesn't look at the openshift-storage namespace, so we'd never see this subscription anyway.
> 
> I think it does. The only reason why it might not do so currently is that, we delete "openshift-storage" namespace before we run must-gather which is silly.
> 
> > I would also ask to add a fetch for any CatalogSources and OperatorSources, allowing either to return nothing as valid.
> 
> Instead of adding individual commands for each things, let's just use OCP must-gather? These are all OCP resources and they should be gathering it's logs!!
> And, OperatorSources is deprecated since OCP 4.2, and we don't use it. So, no need to collect that for sure.

I swear I remember multiple cases where I've had Support or QE send an OCP must-gather and it did not have anything for openshift-storage... please prove me wrong, and then I'll relent. :)

PR #496, opened 2020-05-05
removes CSV check from CI test target

umangachapagain:
removing the CSV check from ocs-operator-ci test target
to remove dev overhead of maintaining CSV even for
unrelated changes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -137,7 +137,7 @@ verify-generated: update-generated
 	@echo "Verifying generated code"
 	hack/verify-generated.sh
 
-ocs-operator-ci: shellcheck-test gofmt golint govet unit-test build verify-latest-csv verify-generated verify-latest-deploy-yaml
+ocs-operator-ci: shellcheck-test gofmt golint govet unit-test build verify-generated verify-latest-deploy-yaml
 
 red-hat-storage-ocs-ci:
 	@echo "Running red-hat-storage ocs-ci test suite"


umangachapagain:
> I don't agree with removing it entirely. We still NEED to check if the CSV is valid and up to date when changes are made to the code that require it. 
>
Agreed. My plan was to add a separate test spec(in CI) for this. That way it won't block other tests and can be checked when required.
>
> What we should be doing is fixing the `verify-latest-csv` test to not need to regenerate just for a timestamp. Unless I'm missing something that would still catch these things?
>
I don't think timestamp change is part of checksum validation. Even if it is, I'm not sure if we can just "not need" this part. Will check out the possibilities.

jarrpa:
> > I don't agree with removing it entirely. We still NEED to check if the CSV is valid and up to date when changes are made to the code that require it.
> 
> Agreed. My plan was to add a separate test spec(in CI) for this. That way it won't block other tests and can be checked when required.
> (PR: [openshift/release#8935](https://github.com/openshift/release/pull/8935))

Ah, I did not see this when I made the comment. This is a great idea! I'll definitely allow this PR once the openshift/release PR merges.

> > What we should be doing is fixing the `verify-latest-csv` test to not need to regenerate just for a timestamp. Unless I'm missing something that would still catch these things?
> 
> I don't think timestamp change is part of checksum validation. Even if it is, I'm not sure if we can just "not need" this part. Will check out the possibilities.

I think I see what you mean. Yeah, this is not straightforward.

jarrpa:
The verify-latest-csv job merged.
/lgtm
/test ocs-operator-e2e-aws

PR #497, opened 2020-05-06
Changes in the Ceph Cluster CR for external cluster

rajatsingh25aug:
This is the new CephCluster CR that needs to be injected:
```
apiVersion: ceph.rook.io/v1
kind: CephCluster
metadata:
  name: openshift-storage-external->>namespace can be changed according to the requirement
  namespace: openshift-storage-external
spec:
  external:
    enable: true
  crashCollector:
    disable: true
```

### DIFF
--- a/pkg/controller/storagecluster/reconcile.go
+++ b/pkg/controller/storagecluster/reconcile.go
@@ -931,7 +931,9 @@ func newExternalCephCluster(sc *ocsv1.StorageCluster, cephImage string) *cephv1.
 			External: cephv1.ExternalSpec{
 				Enable: true,
 			},
-			DataDirHostPath: "/var/lib/rook",
+			CrashCollector: cephv1.CrashCollectorSpec{
+				Disable: true,
+			},
 		},
 	}
 	return externalCephCluster


rajatsingh25aug:
cc @aruniiird @leseb 
Since I cannot add reviewers so tagging you guys.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -931,6 +931,9 @@ func newExternalCephCluster(sc *ocsv1.StorageCluster, cephImage string) *cephv1.
 			External: cephv1.ExternalSpec{
 				Enable: true,
 			},
+			CrashCollector: cephv1.CrashCollectorSpec{
+				Disable: false,
```
leseb, line 935:
```suggestion
				Disable: true,
```

pkg/controller/storagecluster/reconcile.go
```diff
@@ -931,6 +931,9 @@ func newExternalCephCluster(sc *ocsv1.StorageCluster, cephImage string) *cephv1.
 			External: cephv1.ExternalSpec{
 				Enable: true,
 			},
+			CrashCollector: cephv1.CrashCollectorSpec{
+				Disable: true,
+			},
 			DataDirHostPath: "/var/lib/rook",
```
leseb, line 937: Actually, this is not needed. You can remove it.

rajatsingh25aug: May I know why?
This is the field that rook is expecting right?.

leseb: It's not needed anymore in the CephCluster external configuration. So please remove `DataDirHostPath`.

jarrpa: My understanding is that `DataDirHostPath` is used for mons, and since we're not deploying any mons it's pointless to have it.

rajatsingh25aug: Please check now.

jarrpa:
Functional tests failing due to an unknown networking error, bypassing.

/override ci/prow/ocs-operator-e2e-aws

PR #498, opened 2020-05-07
Create ocs-on-a-disconnected-environment.md

dannyzaken:
copied disconnected env doc from [here](https://gist.github.com/dannyzaken/c7989f4481c2c060d1c6c5c5df4f2ec9) so it can be edited by other contributors

### DIFF
--- a/docs/ocs-on-a-disconnected-environment.md
+++ b/docs/ocs-on-a-disconnected-environment.md
@@ -1,0 +1,190 @@
+# OCS on a Disconnected environment
+
+This doc is based on [this document][1]
+
+## Motivation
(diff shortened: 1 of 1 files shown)


jarrpa:
My understanding is that since we are delivered as part of the redhat-operators catalog that we'd be automatically mirrored into the disconnected environment. Is this not the case? Or is this for getting non-GA content out into a disconnected environment?

dannyzaken:
@jarrpa this is the preliminary doc I prepared for a disconnected environment, which contained some workarounds to make it work in 4.2. It is currently maintained in a private gist in my GitHub account (link above). 
I wanted to put it here so it can serve as a basis so others can update it. I think eventually it should also describe how to install non-GA versions in a disconnected env

PR #500, opened 2020-05-08
Ceph: Create resources automatically for external Cluster

rajatsingh25aug:
The JSON BLOB will be passed from the UI and the appropriate resources will be created one by one.

This PR is also dependent on @aruniiird PR #505 . I will keep rebasing my PR until its merged.


pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -61,6 +61,16 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
...
+	_, err := r.newStorageClasses(instance)
```
rajatsingh25aug, line 67: @umangachapagain can you please let me know where and how to create resources from this test resources meanwhile since the script PR is in WIP?. Thanks

pkg/apis/ocs/v1/storagecluster_types.go
```diff
@@ -66,6 +66,14 @@ type StorageDeviceSet struct {
...
 	Config          StorageDeviceSetConfig       `json:"config,omitempty"`
 	DataPVCTemplate corev1.PersistentVolumeClaim `json:"dataPVCTemplate"`
+	// Following Fields will be present when we will get the BLOB from the UI
+	UserID         string `json:"userID"`
+	UserKey        string `json:"userKey"`
+	FilesystemName string `json:"filesystem-name,omitempty"`
+	PoolName       string `json:"pool-name,omitempty"`
+	ClusterName    string `json:"cluster-name,omitempty"`
+	MaxMonID       string `json:"maxMonId,omitempty"`
+	Fsid           string `json:"fsid,omitempty"`
```
rajatsingh25aug, line 76: These were the fields that were coming from the JSON BLOB @aruniiird , looks good?. Thanks

pkg/controller/storagecluster/reconcile.go
```diff
@@ -232,6 +235,7 @@ func (r *ReconcileStorageCluster) Reconcile(request reconcile.Request) (reconcil
 		// for external cluster, we have a different set of ensure functions
 		ensureFs = []ensureFunc{
 			r.ensureCephCluster,
+			r.ensureExternalClusterConfig,
```
rajatsingh25aug, line 238: @jarrpa this is the function call that we wanted to call as we discussed right?. Thanks

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -61,6 +70,103 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
...
+					OwnerReferences: []metav1.OwnerReference{ownerRef},
+				},
+			}
+			err = controllerutil.SetControllerReference(instance, s, r.scheme)
+			if err != nil {
+				return err
+			}
+			_, err = controllerutil.CreateOrUpdate(context.TODO(), r.client, s, func() error {
+				for k, v := range d.Data {
+					s.Data[k] = []byte(v)
```
leseb, line 156: ditto

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -61,6 +70,103 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
...
+					Name:            d.Name,
+					Namespace:       d.Name,
+					OwnerReferences: []metav1.OwnerReference{ownerRef},
+				},
+			}
+			err = controllerutil.SetControllerReference(instance, cm, r.scheme)
+			if err != nil {
+				return err
+			}
+			_, err = controllerutil.CreateOrUpdate(context.TODO(), r.client, cm, func() error {
```
leseb, line 118: What about just using `r.client.Create`? We don't want to update anything here, also the check `!errors.IsAlreadyExists(err)` later will make more sense.

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -16,8 +17,16 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
+type ExternalResources []struct {
+	Kind           string            `json:"kind"`
+	Data           map[string]string `json:"data,omitempty"`
+	Name           string            `json:"name"`
+	FilesystemName string            `json:"filesystem-name,omitempty"`
```
leseb, line 24: please add

PoolName string            `json:"pool-name,omitempty"`

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -61,6 +70,103 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
...
+				return err
+			}
+			_, err = controllerutil.CreateOrUpdate(context.TODO(), r.client, cm, func() error {
+				cm.Data = d.Data
+				return nil
+			})
+			if err != nil && !errors.IsAlreadyExists(err) {
+				return err
+			}
+		case "StorageClass":
```
leseb, line 125: The StorageClass case needs to be adapted to reflect with the existing logic to create SC. The data has no meaning here. The only things that matter here are:

* for cephfs: `d.FilesystemName`
* for rbd: `d.PoolName`
* for obc: `d.Data['endpoint']`

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -61,6 +70,103 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
...
 
+// ensureExternalClusterConfig ensures that Storagclass for the external cluster exist
+func (r *ReconcileStorageCluster) ensureExternalClusterConfig(instance *ocsv1.StorageCluster, reqLogger logr.Logger) error {
+	// check for the status boolean value accepted or not
+	if instance.Status.ExternalSecretFound {
+		return nil
+	}
+	found := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "rook-ceph-external-cluster-details",
```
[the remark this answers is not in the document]
rajatsingh25aug: Not sure if you meant as in `const found` instead of just the variable name `found`?.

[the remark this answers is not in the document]
leseb: use a constant for `rook-ceph-external-cluster-details`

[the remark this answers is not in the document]
rajatsingh25aug: I think you were talking about this 
```
const SecretName =  rook-ceph-external-cluster-details
const Namespace = openshift-storage
```

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -61,6 +70,103 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
...
+					Namespace:       d.Name,
+					OwnerReferences: []metav1.OwnerReference{ownerRef},
+				},
+			}
+			err = controllerutil.SetControllerReference(instance, cm, r.scheme)
+			if err != nil {
+				return err
+			}
+			_, err = controllerutil.CreateOrUpdate(context.TODO(), r.client, cm, func() error {
+				cm.Data = d.Data
```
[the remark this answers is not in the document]
rajatsingh25aug: Not sure what you meant by that, can you be please more precise?. Thanks

pkg/controller/storagecluster/initialization_reconciler.go
rajatsingh25aug: What about just using `r.client.Create`? We don't want to update anything here,.
> Do you mean like this?. `_,err = r.client.Create(context.TODO(),s)`

 also the check `!errors.IsAlreadyExists(err)` later will make more sense.
>Do you meant that to check for `!errors.IsAlreadyExists(err) at the very end of the switch cases?.

leseb: Yes and keep the check `!errors.IsAlreadyExists(err)` intact.

pkg/controller/storagecluster/initialization_reconciler.go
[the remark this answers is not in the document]
leseb: Why not doing `cm.Data = d.Data` when creating `cm`?

pkg/controller/storagecluster/initialization_reconciler.go
rajatsingh25aug: I believe that this is where the "names" will come for the corresponding Storageclass?.
Can you please give more details that if we don't have to use Switch case for storage class creation then what?
I couldn't come up with another logic. Thanks

leseb: ocs-op already has a logic to create the rbd and cephfs pool, just re-use it and add a new one for the OBC.

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -61,6 +70,103 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
...
+		return nil
+	}
+	found := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "rook-ceph-external-cluster-details",
+			Namespace: "openshift-storage",
+		},
+
+		Data: map[string][]byte{
+			"external_cluster_details": []byte(""),
```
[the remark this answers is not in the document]
rajatsingh25aug: Done

pkg/controller/storagecluster/initialization_reconciler.go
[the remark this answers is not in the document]
rajatsingh25aug: Done,thanks

pkg/controller/storagecluster/initialization_reconciler.go
rajatsingh25aug: Added, thanks

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -16,8 +17,23 @@ import (
...
+	Kind           string            `json:"kind"`
+	Data           map[string]string `json:"data,omitempty"`
+	Name           string            `json:"name"`
+	PoolName       string            `json:"pool-name,omitempty"`
+	FilesystemName string            `json:"filesystem-name,omitempty"`
+}
+
+const (
+	secretName      = "rook-ceph-external-cluster-details"
+	secretNamespace = "openshift-storage"
```
leseb, line 30: This feels off to have it hardcoded, the op should know from the `ReconcileStorageCluster` I assume.

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -61,6 +77,72 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
...
+	// check for the status boolean value accepted or not
+	if instance.Status.ExternalSecretFound {
+		return nil
+	}
+	found := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      secretName,
+			Namespace: secretNamespace,
+		},
+		Data: map[string][]byte{
```
leseb, line 91: I don't think you need this Data section, just get it from what is going to be fetched.

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -16,8 +17,23 @@ import (
...
+	Data           map[string]string `json:"data,omitempty"`
+	Name           string            `json:"name"`
+	PoolName       string            `json:"pool-name,omitempty"`
+	FilesystemName string            `json:"filesystem-name,omitempty"`
+}
+
+const (
+	secretName      = "rook-ceph-external-cluster-details"
+	secretNamespace = "openshift-storage"
+	clusterDetails  = "external_cluster_details"
```
leseb, line 31:
```suggestion
	externalClusterSecretDataKeyName  = "external_cluster_details"
```

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -16,8 +17,23 @@ import (
...
+)
+
 // ensureStorageClasses ensures that StorageClass resources exist in the desired
 // state.
+
```
leseb, line 36: unnecessary empty line

rajatsingh25aug: My vscode apparently don't look for empty lines at times, Fixed it, thanks

pkg/controller/storagecluster/initialization_reconciler.go
rajatsingh25aug: If we won't hardcode it, how we are we going to set the values for `Name` and `Namespace`  for secrets and configmap at line `116` down below?.

pkg/controller/storagecluster/initialization_reconciler.go
rajatsingh25aug: Sure, will remove that the Data section, Since we are already catching it at line 106 I believe.

pkg/controller/storagecluster/initialization_reconciler.go
leseb: I don't know, doesn't `ReconcileStorageCluster` provide it to you?

pkg/controller/storagecluster/initialization_reconciler.go
jarrpa: I can't see what the code was, but in general I thought the idea was to just blindly create whatever data we're given. In which case there should no need to integrate any of this information in other parts of the code.

jarrpa: @leseb Wait, so the `d.Data` field contains nothing of value for the StorageClass?? Why aren't the parameters there? They're just a `map[string]string` like `d.Data` is. Are we expected to distinguish what kind of parameters to set based on whether certain fields are non-null?

jarrpa: @leseb Wait, so the `d.Data` field doesn't contain the `spec.Parameters` value for the StorageClass? Why not? They're just a `map[string]string` like `d.Data` is. Are we expected to distinguish what kind of parameters to set based on whether certain fields are non-null?

leseb: Well since you do have the parameters already I don't see why they should be coming from the JSON blob. The blob reflects what the external env is, there is no concept of SC there.

Because the SC creation logic is already present we just need to slightly re-adapt it whether this is an external or converged cluster.

jarrpa: ...okay, but why are some of the parameters explicit fields and only one of them is a `map[string]string`?

leseb: So that we don't alter the map. We thought it'd make a clearer difference for Secret and ConfigMap `Data` because the JSON actually builds the `cm.Data`. So you can blindly copy it.

We only make a difference for the SC, which still has a Data field, but I see your point, for an SC we can rename it to `Parameters` if you prefer but still you would have to complete the rest of it. Let us know soon.

rajatsingh25aug:
I have pulled changes from @aruniiird PR for OBC storageClass. Will keep rebasing it with mine.
@jarrpa can you please take a look?.
I have to just write the logic for `CephFs` and `RBD` storageclass creation and that's it. Can you please leave your thoughts?.
thanks

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -78,11 +90,106 @@ func (r *ReconcileStorageCluster) newOBCStorageClass(initData *ocsv1.StorageClus
...
+				if errors.IsNotFound(err) {
+					reqLogger.Info("Creating Ceph Secret: 	" + s.Name)
+					err = r.client.Create(context.TODO(), s)
+					if err != nil {
+						return err
+					}
+				}
+				return err
+			}
+		case "StorageClass":
```
jarrpa, line 166: My understanding is that there is not *one* resource of kind `StorageClass`, but three, one for each actual StorageClass that will be created. Also, you do not want to do this here in this function. In this switch case, you only want to add the resource `d` to a new `[]ExternalResource` list defined outside the `for _, d := range data` loop. The rest you'll want to do in `newStorageClasses()`.

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -78,11 +90,106 @@ func (r *ReconcileStorageCluster) newOBCStorageClass(initData *ocsv1.StorageClus
...
+	instance.Status.ExternalSecretFound = true
+	return nil
+}
+
 // newStorageClasses returns the StorageClass instances that should be created
 // on first run.
 func (r *ReconcileStorageCluster) newStorageClasses(initData *ocsv1.StorageCluster) ([]*storagev1.StorageClass, error) {
 	persistentVolumeReclaimDelete := corev1.PersistentVolumeReclaimDelete
 	allowVolumeExpansion := false
+	// if !instance.Spec.ExternalStorage.Enable { //Converged Mode
```
jarrpa, line 192: Don't commit commented out code. ;)

pkg/apis/ocs/v1/storagecluster_types.go
```diff
@@ -179,3 +181,12 @@ type StorageClusterList struct {
 func init() {
 	SchemeBuilder.Register(&StorageCluster{}, &StorageClusterList{})
 }
+
+// ExternalResources containes a list of External Cluster Resources
+type ExternalResources struct {
```
jarrpa, line 186: Rename to `ExternalResource`

pkg/apis/ocs/v1/storagecluster_types.go
```diff
@@ -104,11 +104,13 @@ type StorageClusterStatus struct {
...
-	CephBlockPoolsCreated       bool `json:"cephBlockPoolsCreated,omitempty"`
-	CephObjectStoreUsersCreated bool `json:"cephObjectStoreUsersCreated,omitempty"`
-	CephFilesystemsCreated      bool `json:"cephFilesystemsCreated,omitempty"`
+	StorageClassesCreated       bool                `json:"storageClassesCreated,omitempty"`
+	CephObjectStoresCreated     bool                `json:"cephObjectStoresCreated,omitempty"`
+	CephBlockPoolsCreated       bool                `json:"cephBlockPoolsCreated,omitempty"`
+	CephObjectStoreUsersCreated bool                `json:"cephObjectStoreUsersCreated,omitempty"`
+	CephFilesystemsCreated      bool                `json:"cephFilesystemsCreated,omitempty"`
+	ExternalSecretFound         bool                `json:"externalSecretFound,omitempty"`
+	ExternalStorageClasses      []ExternalResources `json:"externalStorageClasses,omitempty"`
```
jarrpa, line 113: Move this above the booleans and provide a comment on what it's for. See `FailureDomain`.

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -16,6 +17,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
+const (
+	secretName                       = "rook-ceph-external-cluster-details"
+	externalClusterSecretDataKeyName = "external_cluster_details"
```
jarrpa, line 22:
```suggestion
	externalClusterDetailsSecret = "rook-ceph-external-cluster-details"
	externalClusterDetailsKey    = "external_cluster_details"
```

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -78,11 +90,106 @@ func (r *ReconcileStorageCluster) newOBCStorageClass(initData *ocsv1.StorageClus
...
+			}
+		}
+	}
+	instance.Status.ExternalSecretFound = true
+	return nil
+}
+
 // newStorageClasses returns the StorageClass instances that should be created
 // on first run.
 func (r *ReconcileStorageCluster) newStorageClasses(initData *ocsv1.StorageCluster) ([]*storagev1.StorageClass, error) {
```
jarrpa, line 189: Here you want to take the relevant values that we're setting for the external StorageClasses (`fsName`, `pool`, and `endpoint`) and define them above `ret`. Then you'll want a `switch` against the ExternalResources in `initData.Status.ExternalStorageClasses`, setting the appropriate value for each name found.

pkg/controller/storagecluster/reconcile.go
jarrpa: Yes, but move it above `r.ensureCephCluster`.

pkg/controller/storagecluster/initialization_reconciler.go
rajatsingh25aug: Oh my bad, I didn't look that, Thanks

pkg/apis/ocs/v1/storagecluster_types.go
rajatsingh25aug: Are you talking about just the `ExternalStorageClasses` or both the `ExternalStorageClasses` and `ExternalSecretFound`?.  I also believe to move the latter with the former as both are only being used when there's an External Cluster.
Wdyt?.

pkg/controller/storagecluster/initialization_reconciler.go
rajatsingh25aug: > My understanding is that there is not _one_ resource of kind `StorageClass`, but three, one for each actual StorageClass that will be created. Also, you do not want to do this here in this function. In this switch case, you only want to add the resource `d` to a new `[]ExternalResource` list defined outside the `for _, d := range data` loop. The rest you'll want to do in `newStorageClasses()`.

When you say "The rest", you meant the `case` just for the StorageClass right?. Meaning that we should remove that and create as you said?. IIRC, we need the switch Case for creation of the  `Secrets` and `ConfigMaps`.

pkg/apis/ocs/v1/storagecluster_types.go
rajatsingh25aug: Will Do, Thanks

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -55,9 +70,117 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
...
+				}
+				return err
+			}
+		case "StorageClass":
+			scs, err := r.newStorageClasses(instance)
+			if err != nil {
+				return err
+			}
+			scCephFs := scs[0]
+			scCephFs.Parameters["fsName"] = d.FilesystemName
```
leseb, line 172: This is part of the Secret no the StorageClass. See:


```
  {
    "filesystem-name": "myfs",
    "kind": "Secret",
    "data": {
      "adminKey": "AQBV66pefnqmERAAKYhoO2XK5mUIGKSN4J/URw==",
      "adminID": "csi-cephfs-provisioner"
    },
    "name": "rook-csi-cephfs-provisioner",
    "pool-name": "myfs-data0"
  },
```

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -55,9 +70,117 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
...
+			scRBD := scs[1]
+			scRBD.Parameters["pool"] = d.PoolName
```
leseb, line 174: This is part of the Provisioner Secret not the StorageClass. See the JSON blob:

```
{
    "kind": "Secret",
    "data": {
      "userKey": "AQBU66peVcBCKxAAoPQ4tKlJ9gxvrH3MRnDRHQ==",
      "userID": "csi-rbd-provisioner"
    },
    "name": "rook-csi-rbd-provisioner",
    "pool-name": "rbd-pool"
}
```

rajatsingh25aug: I didn't quite understand what you meant by `Provisioner Secret` here?. I may be wrong here but, how are we going to differentiate if it's "Secret" or a "Provisioner Secret"?.

leseb: We just moved these into its own StorageClass section, let me comment your code again

pkg/controller/storagecluster/initialization_reconciler.go
leseb, line 172: Now, this is d.Data["filesystem-name"]

pkg/controller/storagecluster/initialization_reconciler.go
leseb: Now, this is d.Data["pool-name"]

pkg/controller/storagecluster/initialization_reconciler.go
rajatsingh25aug: I don't see any field such as "filesystem-name" under Data.
I was referring this JSON blob of yours https://gist.github.com/leseb/8791cba01e371c0b65895085979626c4#file-cluster-json-L57
Also, above in the comments, you mentioned this
```
for cephfs: d.FilesystemName
for rbd: d.PoolName
for obc: d.Data['endpoint']
```

leseb: Sorry, just updated the gist, now it's all good!

rajatsingh25aug: Thanks, Changes made in context with the updated changes that you made in the JSON.

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -86,12 +85,123 @@ func (r *ReconcileStorageCluster) newOBCStorageClass(initData *ocsv1.StorageClus
...
+	}
+	// storageClasses stores all the StorageClasses coming from the JSON
+	var storageClasses map[string]map[string]string = make(map[string]map[string]string)
+	for _, d := range data {
+		switch d.Kind {
+		case "ConfigMap":
+			cm := &corev1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:            d.Name,
+					Namespace:       d.Name, // TODO: What will be namespace given for ConfigMap?. d.Name or instance.Namespace?.
```
jarrpa, line 128: Yes, all this must be in the same namespace as the StorageCluster representing the external CephCluster.

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -86,12 +85,123 @@ func (r *ReconcileStorageCluster) newOBCStorageClass(initData *ocsv1.StorageClus
...
+						return err
+					}
+				}
+				return err
+			}
+		case "Secret":
+			s := &corev1.Secret{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:            d.Name,
+					Namespace:       d.Name, // TODO: What will be namespace for Secret?. d.Name or instance.Namespace?.
```
jarrpa, line 149: Same as above.

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -86,12 +85,123 @@ func (r *ReconcileStorageCluster) newOBCStorageClass(initData *ocsv1.StorageClus
...
+		Kind:       instance.Kind,
+		Name:       instance.Name,
+	}
+	var data []ocsv1.ExternalResource
+	err = json.Unmarshal(found.Data[externalClusterDetailsKey], &data)
+	if err != nil {
+		return fmt.Errorf("could not parse json blob, got %+v", err)
+	}
+	// storageClasses stores all the StorageClasses coming from the JSON
+	var storageClasses map[string]map[string]string = make(map[string]map[string]string)
```
jarrpa, line 121: Don't have a map for the data, just do `var storageClasses []ExternalResource` and then `storageClasses = append(storageClasses, d)`.

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -86,12 +85,123 @@ func (r *ReconcileStorageCluster) newOBCStorageClass(initData *ocsv1.StorageClus
...
+		}
+		for k, v := range scData {
+			testStorageClass.Parameters[k] = v
+		}
+	}
+	// creating all the storageClass once we set the values
+	err = r.createStorageClasses(scs, reqLogger)
+	if err != nil {
+		return err
+	}
```
jarrpa, line 200: Don't do *any* of this here, you want all this integrated into `newStorageClasses()`. You need to create a new StorageCluster.Status field and save `storageClasses` there.

pkg/controller/storagecluster/initialization_reconciler.go
jarrpa: See below.

pkg/controller/storagecluster/initialization_reconciler.go
rajatsingh25aug: Here's what I was thinking, something like this.

```
// This if block goes on the very first line of newStorageClasses() 

if StorageCluster.Status.ExterClusterFound {
// Move the Entire code from  line 122 to 200 here in this If block.
}
```

On a side note, what are we going to achieve moving under `newStorageClasses()` here instead of having a separate function?.

pkg/apis/ocs/v1/storagecluster_types.go
```diff
@@ -179,3 +182,12 @@ type StorageClusterList struct {
...
 	SchemeBuilder.Register(&StorageCluster{}, &StorageClusterList{})
 }
+
+// ExternalResources containes a list of External Cluster Resources
+type ExternalResource struct {
+	Kind           string            `json:"kind"`
+	Data           map[string]string `json:"data,omitempty"`
+	Name           string            `json:"name"`
+	PoolName       string            `json:"pool-name,omitempty"`
+	FilesystemName string            `json:"filesystem-name,omitempty"`
```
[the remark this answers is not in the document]
leseb: +1 to be more precise both `PoolName` and `FilesystemName` must be removed.

pkg/controller/defaults/placements.go
```diff
@@ -7,6 +7,7 @@ import (
 )
 
 var (
+	app = "app"
```
leseb, line 10: What about a constant?
Same for `rook-ceph-rgw`?

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -55,9 +70,135 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
...
+		switch d.Kind {
+		case "ConfigMap":
+			cm := &corev1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:            d.Name,
+					Namespace:       d.Name, // TODO: What will be namespace given for ConfigMap?. d.Name or instance.Namespace?.
+					OwnerReferences: []metav1.OwnerReference{ownerRef},
+				},
+			}
+			cm.Data = d.Data
```
leseb, line 132: Why not doing this directly when creation `cm`? This seems unnecessary.

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -55,9 +70,135 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
...
+			found := &corev1.ConfigMap{}
+			err := r.client.Get(context.TODO(), types.NamespacedName{Name: d.Name, Namespace: instance.Namespace}, found)
```
leseb, line 134: Make `types.NamespacedName{Name: d.Name, Namespace: instance.Namespace}` a variable instead repeating it multiple times.

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -55,9 +70,135 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
...
+	if err != nil {
+		return fmt.Errorf("could not parse json blob, got %+v", err)
+	}
+	// storageClasses stores all the StorageClasses coming from the JSON
+	var storageClasses map[string]map[string]string = make(map[string]map[string]string)
+	for _, d := range data {
+		switch d.Kind {
+		case "ConfigMap":
+			cm := &corev1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{
```
leseb, line 126: Because `ObjectMeta` is the same in all objects, please make a variable instead of repeating this all the time.

pkg/apis/ocs/v1/storagecluster_types.go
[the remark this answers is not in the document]
rajatsingh25aug: Thanks, I didn't had the new JSON structure to look into. Will do

pkg/controller/storagecluster/initialization_reconciler.go
rajatsingh25aug: Sure

pkg/apis/ocs/v1/storagecluster_types.go
```diff
@@ -179,3 +182,10 @@ type StorageClusterList struct {
 func init() {
 	SchemeBuilder.Register(&StorageCluster{}, &StorageClusterList{})
 }
+
+// ExternalResources containes a list of External Cluster Resources
+type ExternalResource struct {
+	Kind string            `json:"kind"`
+	Data map[string]string `json:"data,omitempty"`
```
leseb, line 189:
```suggestion
	Data map[string]string `json:"data"`
```

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -55,12 +70,138 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
...
+			OwnerReferences: []metav1.OwnerReference{ownerRef},
+		}
+		objectKey := types.NamespacedName{Name: d.Name, Namespace: instance.Namespace}
+		switch d.Kind {
+		case "ConfigMap":
+			cm := &corev1.ConfigMap{
+				ObjectMeta: objectMeta,
+				Data:       d.Data,
+			}
+			found := &corev1.ConfigMap{}
```
leseb, line 119: Wouldn't that be always true? The `found` cm needs to contain the `objectMeta`.

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -55,12 +70,138 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
...
+			}
+		case "Secret":
+			s := &corev1.Secret{
+				ObjectMeta: objectMeta,
+				Data:       make(map[string][]byte),
+			}
+			for k, v := range d.Data {
+				s.Data[k] = []byte(v)
+			}
+			found := &corev1.Secret{}
```
leseb, line 139: ditto

pkg/apis/ocs/v1/storagecluster_types.go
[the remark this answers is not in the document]
rajatsingh25aug: Addressed

pkg/controller/storagecluster/initialization_reconciler.go
rajatsingh25aug: Addressed

pkg/controller/storagecluster/initialization_reconciler.go
rajatsingh25aug: Addressed

pkg/controller/storagecluster/initialization_reconciler.go
rajatsingh25aug: Done, thanks

pkg/controller/storagecluster/initialization_reconciler.go
rajatsingh25aug: Done

pkg/controller/storagecluster/initialization_reconciler.go
rajatsingh25aug: True, Addressed

pkg/controller/storagecluster/initialization_reconciler.go
rajatsingh25aug: Addressed

pkg/apis/ocs/v1/storagecluster_types.go
rajatsingh25aug: Addressed

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -55,9 +70,118 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
...
+		return nil
+	}
+	found := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      externalClusterDetailsSecret,
+			Namespace: instance.Namespace,
+		},
+	}
+	err := r.client.Get(context.TODO(), types.NamespacedName{Name: found.Name, Namespace: found.Namespace}, found)
+	if err != nil {
```
leseb, line 89: I think we should check for `if errors.IsNotFound(err) {` and potentially log something useful to show up in the op logs.

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -55,9 +70,118 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
...
+	ownerRef := metav1.OwnerReference{
+		UID:        instance.UID,
+		APIVersion: instance.APIVersion,
+		Kind:       instance.Kind,
+		Name:       instance.Name,
+	}
+	var data []ocsv1.ExternalResource
+	err = json.Unmarshal(found.Data[externalClusterDetailsKey], &data)
+	if err != nil {
+		return fmt.Errorf("could not parse json blob, got %+v", err)
```
leseb, line 101: Most of the time the code does `return err` so all of a sudden it feels off to name the error. As a best practice, I'd encourage you to always name the errors.

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -55,9 +70,118 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
...
+		case "ConfigMap":
+			cm := &corev1.ConfigMap{
+				ObjectMeta: objectMeta,
+				Data:       d.Data,
+			}
+			found := &corev1.ConfigMap{ObjectMeta: objectMeta}
+			err := r.client.Get(context.TODO(), objectKey, found)
+			if err != nil {
+				if errors.IsNotFound(err) {
+					reqLogger.Info("Creating Ceph ConfigMap:	" + cm.Name)
```
leseb, line 123: Since there is no `Infof` I'd suggest using `fmt.Sprintf` instead of string concat.

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -55,9 +70,118 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
...
+					if err != nil {
+						return err
+					}
+				}
+				return err
+			}
+		case "Secret":
+			s := &corev1.Secret{
+				ObjectMeta: objectMeta,
+				Data:       make(map[string][]byte),
```
leseb, line 134: Why using this `d.Data` instead of the `make`? Why is it different between the Secret and the ConfigMap?

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -55,9 +70,118 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
...
+			// Setting the PoolName for RBD StorageClass
+			testStorageClass = scs[1]
+		} else if scName == "ceph-rgw" {
+			// Setting the Endpoint for OBC StorageClass
+			testStorageClass = scs[2]
+		}
+		for k, v := range scData {
+			testStorageClass.Parameters[k] = v
+		}
+	}
```
leseb, line 178: Why can't you do this in the `case "StorageClass":` above?

pkg/apis/ocs/v1/storagecluster_types.go
```diff
@@ -179,3 +182,10 @@ type StorageClusterList struct {
 func init() {
 	SchemeBuilder.Register(&StorageCluster{}, &StorageClusterList{})
 }
+
+// ExternalResources containes a list of External Cluster Resources
+type ExternalResource struct {
+	Kind string            `json:"kind"`
+	Data map[string]string `json:"data"`
+	Name string            `json:"name"`
+}
```
umangachapagain, line 191: Since this is used internally and by the initialization reconciler, move this to initialization_reconciler.go.
This should not be here as this file indicates API changes.

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -55,9 +70,118 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
...
+				Data:       d.Data,
+			}
+			found := &corev1.ConfigMap{ObjectMeta: objectMeta}
+			err := r.client.Get(context.TODO(), objectKey, found)
+			if err != nil {
+				if errors.IsNotFound(err) {
+					reqLogger.Info("Creating Ceph ConfigMap:	" + cm.Name)
+					err = r.client.Create(context.TODO(), cm)
+					if err != nil {
+						return err
```
umangachapagain, line 126: add error message.

pkg/controller/storagecluster/initialization_reconciler.go
umangachapagain, line 123:
```suggestion
					reqLogger.Info("Creating Ceph ConfigMap: " + cm.Name)
```

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -55,9 +70,118 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
...
+					}
+				}
+				return err
```
umangachapagain, line 129: add error message. what is this error?

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -55,9 +70,118 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
...
+				Data:       make(map[string][]byte),
+			}
+			for k, v := range d.Data {
+				s.Data[k] = []byte(v)
+			}
+			found := &corev1.Secret{ObjectMeta: objectMeta}
+			err := r.client.Get(context.TODO(), objectKey, found)
+			if err != nil {
+				if errors.IsNotFound(err) {
+					reqLogger.Info("Creating Ceph Secret: 	" + s.Name)
```
umangachapagain, line 143:
```suggestion
					reqLogger.Info("Creating Ceph Secret: " + s.Name)
```

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -55,9 +70,118 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
...
+					reqLogger.Info("Creating Ceph ConfigMap:	" + cm.Name)
+					err = r.client.Create(context.TODO(), cm)
+					if err != nil {
+						return err
+					}
+				}
+				return err
+			}
+		case "Secret":
+			s := &corev1.Secret{
```
umangachapagain, line 132: Can we call it `secret` or `sec` instead of `s`?

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -55,9 +70,118 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
...
+				s.Data[k] = []byte(v)
+			}
+			found := &corev1.Secret{ObjectMeta: objectMeta}
+			err := r.client.Get(context.TODO(), objectKey, found)
+			if err != nil {
+				if errors.IsNotFound(err) {
+					reqLogger.Info("Creating Ceph Secret: 	" + s.Name)
+					err = r.client.Create(context.TODO(), s)
+					if err != nil {
+						return err
```
umangachapagain, line 146: add error message.

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -55,9 +70,118 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
...
+					}
+				}
+				return err
```
umangachapagain, line 149: add error message.

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -55,9 +70,118 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
...
+			storageClasses[d.Name] = d.Data
+		}
+	}
+	scs, err := r.newStorageClasses(instance)
+	if err != nil {
+		return err
+	}
+	if len(scs) < 3 {
+		scs = append(scs, r.newOBCStorageClass(instance))
+	}
```
umangachapagain, line 161: What is this about?

pkg/controller/storagecluster/initialization_reconciler.go
umangachapagain, line 178: I'm confused here and comments aren't helping. Can you explain?

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -55,9 +70,118 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
...
+		for k, v := range scData {
+			testStorageClass.Parameters[k] = v
+		}
+	}
+	// creating all the storageClass once we set the values
+	err = r.createStorageClasses(scs, reqLogger)
+	if err != nil {
+		return err
+	}
+	instance.Status.ExternalSecretFound = true
```
umangachapagain, line 184: Shouldn't you update this status right after you found the secret, instead of waiting for the whole function to succeed?
If Secret is found but some other error occurred, this wouldn't be updated right?

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -55,9 +70,118 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
 			}
 		}
 	}
+	return nil
+}
 
-	instance.Status.StorageClassesCreated = true
-
+// ensureExternalCluster ensures that Storagclass for the external cluster exist
+func (r *ReconcileStorageCluster) ensureExternalCluster(instance *ocsv1.StorageCluster, reqLogger logr.Logger) error {
```
umangachapagain, line 77: If this function is supposed to ensure StorageClass for External Cluster, shouldn't it be named accordingly? `ensureExternalCluster` seems misleading.

umangachapagain: I see that this function also creates ConfigMaps and Secrets. So `ensureExternalClusterObjects` or `ensureExternalClusterResources` etc sounds more appropriate to me.

umangachapagain:
Why does an Independent Mode PR depend on Converged Mode PR?

umangachapagain:
CI error: `./pkg/apis/ocs/v1/storagecluster_types.go:186:1: comment on exported type ExternalResource should be of the form "ExternalResource ..." (with optional leading article)`

leseb:
> Why does an Independent Mode PR depend on Converged Mode PR?

For the SC (for OBC) creation logic.

pkg/controller/storagecluster/initialization_reconciler.go
rajatsingh25aug: alright

pkg/controller/storagecluster/initialization_reconciler.go
rajatsingh25aug: I didn't quite understand the `always name the errors`. Can you please explain a bit more?. Thanks

pkg/controller/storagecluster/initialization_reconciler.go
rajatsingh25aug: The reason we do that because `corev1.Secret{}` struct has `Data` field of type  `map[string][]byte` whereas  `corev1.ConfigMap{}` struct has  `Data` field of type `map[string]string`.

So yes, the Data field is different for both and that is why we did the above operation

pkg/controller/storagecluster/initialization_reconciler.go
rajatsingh25aug: **Please ask clarification questions if needed.**
The reason is that whenever we call `newStorageClass` it will return us 3 Storage classes.
So once we will have all the StorageClass information at one place, then we can process them And start modifying them.

I hope that makes sense :)

pkg/controller/storagecluster/initialization_reconciler.go
rajatsingh25aug: That makes sense. Will do right after we got the data.

pkg/controller/storagecluster/initialization_reconciler.go
rajatsingh25aug: Storing All the storageclasses in `scs` and then setting the value based on their fsname.

https://gist.github.com/leseb/8791cba01e371c0b65895085979626c4https://gist.github.com/leseb/8791cba01e371c0b65895085979626c4

pkg/controller/storagecluster/initialization_reconciler.go
rajatsingh25aug: This shouldn't be here

pkg/apis/ocs/v1/storagecluster_types.go
rajatsingh25aug: Yes, we were thinking of doing the same.

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -28,9 +34,18 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
...
+	}
+	instance.Status.StorageClassesCreated = true
+	return nil
+}
+
+func (r *ReconcileStorageCluster) createStorageClasses(scs []*storagev1.StorageClass, reqLogger logr.Logger) error {
 	for _, sc := range scs {
 		existing := storagev1.StorageClass{}
-		err = r.client.Get(context.TODO(), types.NamespacedName{Name: sc.Name, Namespace: sc.Namespace}, &existing)
+		err := r.client.Get(context.TODO(), types.NamespacedName{Name: sc.Name, Namespace: sc.Namespace}, &existing)
```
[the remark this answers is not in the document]
rajatsingh25aug: Can you explain a bit more?.

deploy/crds/ocs_v1_storagecluster_crd.yaml
```diff
@@ -167,6 +167,10 @@ spec:
                 - lastTransitionTime
                 type: object
               type: array
+            externalSecretFound:
+              description: ExternalSecretFound checks if there's an external cluster
+                found or not
```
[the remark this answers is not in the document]
rajatsingh25aug: It should be Secret instead of Cluster. Will fix it.

pkg/controller/storagecluster/initialization_reconciler.go
rajatsingh25aug: Alright

pkg/controller/storagecluster/initialization_reconciler.go
rajatsingh25aug: newStorageClass might not return OBC storageClass so checking the number of storageClasses returned

pkg/controller/storagecluster/initialization_reconciler.go
leseb: Instead of returning `err` format the error just like `fmt.Errorf("could not parse json blob, got %+v", err)` and use a meaningful message.

pkg/controller/storagecluster/initialization_reconciler.go
leseb: Sounds good.

pkg/controller/storagecluster/initialization_reconciler.go
rajatsingh25aug: Done, at the very top

pkg/controller/storagecluster/initialization_reconciler.go
rajatsingh25aug: done

pkg/controller/storagecluster/initialization_reconciler.go
rajatsingh25aug: I did this.
```
reqLogger.Error(err, "could not parse json blob")
		return err
```

pkg/controller/storagecluster/initialization_reconciler.go
rajatsingh25aug: We require the OBC storage Class and if the `newStorageClass` didn't return the OBC StorageClass, we will create it.

pkg/controller/storagecluster/initialization_reconciler.go
rajatsingh25aug: We're still following the switch cases approach as modifying the `newStorageClasses ` will I believe does not make much sense as that function job is to just create StorageClasses and nothing else.

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -28,9 +44,18 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
 	if err != nil {
 		return err
 	}
+	err = r.createStorageClasses(scs, reqLogger)
+	if err != nil {
+		return err
```
leseb, line 49: name your error

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -55,12 +80,140 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
...
+	}
+	var data []ExternalResource
+	err = json.Unmarshal(found.Data[externalClusterDetailsKey], &data)
+	if err != nil {
+		reqLogger.Error(err, "could not parse json blob")
+		return err
+	}
+	scs, err := r.newStorageClasses(instance)
+	if err != nil {
+		return err
```
leseb, line 116: name your error

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -55,12 +80,140 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
...
+			if err != nil {
+				if errors.IsNotFound(err) {
+					reqLogger.Info(fmt.Sprintf("creating configmap: %s", cm.Name))
+					err = r.client.Create(context.TODO(), cm)
+					if err != nil {
+						reqLogger.Error(err, "creation of configmap failed")
+						return err
+					}
+				} else {
+					reqLogger.Error(err, "Unable the get the configmap")
```
leseb, line 147:
```suggestion
					reqLogger.Error(err, "unable the get the configmap")
```

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -55,12 +80,140 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
...
+			}
+			for k, v := range d.Data {
+				sc.Parameters[k] = v
+			}
+		}
+	}
+	// creating all the storageClass once we set the values
+	err = r.createStorageClasses(scs, reqLogger)
+	if err != nil {
+		return err
```
leseb, line 194: name your error

pkg/controller/storagecluster/initialization_reconciler.go
rajatsingh25aug: I'm not quite sure what you meant here?. 
So instead of using `err` I should use so something like `newStorageClasserr`?.

leseb: I told you not to return `err` but explain the error, somewhere else you logged and returned `err` not sure who recommended that but go ahead and follow that pattern

jarrpa: `newStorageClasses()` should be responsible for elaborating on its own errors, we just need to pass it up. This is outside the scope of this PR.

pkg/controller/storagecluster/initialization_reconciler.go
jarrpa: `createStorageClasses()` should be responsible for elaborating on its own errors, we just need to pass it up. This is outside the scope of this PR.

pkg/controller/storagecluster/initialization_reconciler.go
jarrpa: `createStorageClasses()` should be responsible for elaborating on its own errors, we just need to pass it up. This is outside the scope of this PR.

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -55,9 +80,120 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
...
+	}
+	scs, err := r.newStorageClasses(instance)
+	if err != nil {
+		return err
+	}
+	// newStorageClass might not return OBC storageClass so checking the
+	// number of storageClasses returned
+	if len(scs) < 3 {
+		scs = append(scs, r.newOBCStorageClass(instance))
+	}
```
jarrpa, line 122: Don't do this. If you can't rely on `newStorageClasses()` to create the OBC class, fix it.

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -55,9 +80,117 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
...
 	}
+	return nil
+}
 
-	instance.Status.StorageClassesCreated = true
-
+// ensureExternalStorageClusterResources ensures that requested resources for the external cluster
+// being created
+func (r *ReconcileStorageCluster) ensureExternalStorageClusterResources(instance *ocsv1.StorageCluster, reqLogger logr.Logger) error {
+	// check for the status boolean value accepted or not
```
aruniiird, line 89:
```suggestion
	// if external storage is not enabled, this function should not be called
	if !instance.Spec.ExternalStorage.Enable {
		return nil
	}
	// check for the status boolean value accepted or not
```

Even though this function is being put into the `else` part (where `Spec.ExternalStorage.Enable` is true) of the reconcile method, as an independent function, it should only be called in an external storage environment.
@jarrpa , what do you think?

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -125,7 +258,8 @@ func (r *ReconcileStorageCluster) newStorageClasses(initData *ocsv1.StorageClust
 		},
 	}
 	// OBC StorageClass will be added only if it is not a cloud platform
-	if platform, err := r.platform.GetPlatform(r.client); err == nil && !isValidCloudPlatform(platform) {
+	platform, err := r.platform.GetPlatform(r.client)
+	if initData.Spec.ExternalStorage.Enable || err == nil && (!isValidCloudPlatform(platform)) {
```
aruniiird, line 262:
```suggestion
	// OBC storageclass will be returned only in TWO conditions,
	// a. either 'externalStorage' is enabled
	// OR
	// b. current platform is not a cloud-based platform
	if initData.Spec.ExternalStorage.Enable || err == nil && (!isValidCloudPlatform(platform)) {
```
Can you add some comments on similar lines...

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -16,6 +17,21 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
+const (
+	externalClusterDetailsSecret = "rook-ceph-external-cluster-details"
+	externalClusterDetailsKey    = "external_cluster_details"
+	cephfs                       = "cephfs"
+	cephRBD                      = "ceph-rbd"
+	cephRGW                      = "ceph-rgw"
```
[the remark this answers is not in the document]
rajatsingh25aug: Addressed

pkg/controller/storagecluster/initialization_reconciler.go
rajatsingh25aug: Done, fixed it by adding `initData.Spec.ExternalStorage.Enable` condition.

pkg/controller/storagecluster/initialization_reconciler.go
rajatsingh25aug: SUre, will do

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -55,12 +80,137 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
...
+			} else if d.Name == cephRGWStorageClassName {
+				// Setting the Endpoint for OBC StorageClass
+				sc = scs[2]
+			}
+			for k, v := range d.Data {
+				sc.Parameters[k] = v
+			}
+		}
+	}
+	// creating all the storageClass once we set the values
```
umangachapagain, line 188:
```suggestion
	// creating all the storageClasses once we set the values
```

pkg/controller/storagecluster/initialization_reconciler.go
rajatsingh25aug: Done

pkg/controller/storagecluster/initialization_reconciler.go
rajatsingh25aug: Done

pkg/apis/ocs/v1/zz_generated.openapi.go
```diff
@@ -306,6 +306,13 @@ func schema_pkg_apis_ocs_v1_StorageClusterStatus(ref common.ReferenceCallback) c
 							Format:      "",
 						},
 					},
+					"externalSecretFound": {
+						SchemaProps: spec.SchemaProps{
+							Description: "ExternalSecretFound checks if there's an external secret found or not",
+							Type:        []string{"boolean"},
+							Format:      "",
+						},
+					},
```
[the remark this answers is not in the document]
rajatsingh25aug: These changes came from running `make gen-latest-csv`

rajatsingh25aug:
cc @umangachapagain  @jarrpa I wrote one test case with best of my knowledge by taking reference from earlier test cases in the same file. It was passing for me, but please feel free to comment and suggestion as I think that we can write more precise test case for it.

Since I didn't have many ideas about unit testing in OCS op, I wrote what I felt was correct. Please drop your comments. Thanks

pkg/apis/ocs/v1/zz_generated.openapi.go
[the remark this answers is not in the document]
umangachapagain: Then your project is setup incorrectly.
it should be in GOPATH/openshift/ocs-operator not GOPATH/ocs-operator

pkg/apis/ocs/v1/zz_generated.openapi.go
```diff
@@ -11,14 +11,14 @@ import (
...
+		"github.com/ocs-operator/pkg/apis/ocs/v1.OCSInitialization":                  schema_pkg_apis_ocs_v1_OCSInitialization(ref),
+		"github.com/ocs-operator/pkg/apis/ocs/v1.OCSInitializationSpec":              schema_pkg_apis_ocs_v1_OCSInitializationSpec(ref),
+		"github.com/ocs-operator/pkg/apis/ocs/v1.OCSInitializationStatus":            schema_pkg_apis_ocs_v1_OCSInitializationStatus(ref),
+		"github.com/ocs-operator/pkg/apis/ocs/v1.StorageCluster":                     schema_pkg_apis_ocs_v1_StorageCluster(ref),
+		"github.com/ocs-operator/pkg/apis/ocs/v1.StorageClusterInitialization":       schema_pkg_apis_ocs_v1_StorageClusterInitialization(ref),
+		"github.com/ocs-operator/pkg/apis/ocs/v1.StorageClusterInitializationSpec":   schema_pkg_apis_ocs_v1_StorageClusterInitializationSpec(ref),
+		"github.com/ocs-operator/pkg/apis/ocs/v1.StorageClusterInitializationStatus": schema_pkg_apis_ocs_v1_StorageClusterInitializationStatus(ref),
+		"github.com/ocs-operator/pkg/apis/ocs/v1.StorageClusterStatus":               schema_pkg_apis_ocs_v1_StorageClusterStatus(ref),
```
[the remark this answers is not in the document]
rajatsingh25aug: These changes came from running make gen-latest-csv

deploy/crds/ocs_v1_storagecluster_crd.yaml
```diff
@@ -167,6 +167,10 @@ spec:
                 - lastTransitionTime
                 type: object
               type: array
+            externalSecretFound:
+              description: ExternalSecretFound checks if there's an external secret
```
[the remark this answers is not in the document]
rajatsingh25aug: These changes came from running make gen-latest-CSV and I'm not making these changes manually

pkg/controller/storagecluster/initialization_reconciler.go
rajatsingh25aug: Addressed

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -55,12 +80,137 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
...
+			if err != nil {
+				if errors.IsNotFound(err) {
+					reqLogger.Info(fmt.Sprintf("creating secret: %s", sec.Name))
+					err = r.client.Create(context.TODO(), sec)
+					if err != nil {
+						reqLogger.Error(err, "creation of secret failed")
+						return err
+					}
+				} else {
+					reqLogger.Error(err, "unable the get the secret")
```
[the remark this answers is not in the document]
rajatsingh25aug: Addressed

pkg/apis/ocs/v1/zz_generated.openapi.go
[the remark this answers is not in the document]
rajatsingh25aug: Oh yes, my `openhsift` folder is missing. Thanks for the tip

pkg/controller/storagecluster/initialization_reconciler_test.go
```diff
@@ -178,6 +180,30 @@ func TestInitStorageClusterResourcesUpdate(t *testing.T) {
...
+			Namespace: "openshift-storage",
+		},
+	}
+
+	reconciler := createFakeInitializationStorageClusterReconciler(t, cr)
+	err := reconciler.client.Create(nil, cr)
+	result, err := reconciler.Reconcile(request)
+	assert.NoError(t, err)
+	assert.Equal(t, reconcile.Result{}, result)
+}
```
jarrpa, line 205: Follow the pattern in `TestInitStorageClusterResourcesUpdate()`:

1. Do a `reconciler.client.Create()` call for every resource we expect to be created, ensure no errors
2. Run `reconciler.Reconcile()` as you do now
3. Create a new function called `assertExpectedExternalResources()`
4. Within that function, use the `var externalResources` I suggested you make, iterate through the list, and for each item:
    1. Create an empty struct of the appropriate type (ConfigMap, Secret, StorageClass)
    2. Run `reconciler.client.Get()` into the empty struct
    3. Assert that the appropriate data from the `ExternalResource` item made it into the resource you just fetched

deploy/crds/ocs_v1_storagecluster_crd.yaml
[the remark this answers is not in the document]
jarrpa: This still needs to be changed in the place it's being generated from, see below.

pkg/controller/storagecluster/initialization_reconciler_test.go
```diff
@@ -216,6 +258,7 @@ func TestInitStorageClusterResourcesUpdateOnUnknownPlatform(t *testing.T) {
 }
 
 func assertExpectedResources(t assert.TestingT, reconciler ReconcileStorageCluster, cr *api.StorageCluster, request reconcile.Request) {
+
```
jarrpa, line 261: Extraneous newline, remove

pkg/controller/storagecluster/initialization_reconciler_test.go
```diff
@@ -352,6 +395,59 @@ func assertExpectedResources(t assert.TestingT, reconciler ReconcileStorageClust
...
+	result, err := reconciler.Reconcile(request)
+	assert.NoError(t, err)
+	assert.Equal(t, reconcile.Result{}, result)
+	assertExpectedExternalResources(t, reconciler, request, jsonBlob)
+}
+
+func assertExpectedExternalResources(t assert.TestingT, reconciler ReconcileStorageCluster, request reconcile.Request, resources []byte) {
+	var data []ExternalResource
+	json.Unmarshal(resources, &data)
+	for _, d := range data {
```
jarrpa, line 426: Instead of `d`, rename this to `expected`, just for a bit more clarity.

pkg/controller/storagecluster/initialization_reconciler_test.go
```diff
@@ -352,6 +395,59 @@ func assertExpectedResources(t assert.TestingT, reconciler ReconcileStorageClust
...
+		request.Name = d.Name
+		switch d.Kind {
+		case "ConfigMap":
+			actual := &corev1.ConfigMap{}
+			err := reconciler.client.Get(nil, request.NamespacedName, actual)
```
jarrpa, line 431: Assert no error first.

pkg/controller/storagecluster/initialization_reconciler_test.go
```diff
@@ -352,6 +395,59 @@ func assertExpectedResources(t assert.TestingT, reconciler ReconcileStorageClust
...
+			assert.Equal(t, d.Data, actual.Data)
+			assert.NoError(t, err)
+		case "Secret":
+			actual := &corev1.Secret{}
+			err := reconciler.client.Get(nil, request.NamespacedName, actual)
```
jarrpa, line 436: Assert no error first.

pkg/controller/storagecluster/initialization_reconciler_test.go
```diff
@@ -352,6 +395,59 @@ func assertExpectedResources(t assert.TestingT, reconciler ReconcileStorageClust
...
+			assert.Equal(t, d.Data, actual.Data)
+			assert.NoError(t, err)
+		case "StorageClass":
+			actual := &storagev1.StorageClass{}
+			err := reconciler.client.Get(nil, request.NamespacedName, actual)
```
jarrpa, line 441: Assert no error first, before the for loop.

pkg/controller/storagecluster/initialization_reconciler_test.go
```diff
@@ -352,6 +395,59 @@ func assertExpectedResources(t assert.TestingT, reconciler ReconcileStorageClust
...
+func TestEnsureExternalStorageClusterResources(t *testing.T) {
+	jsonBlob, _ := json.Marshal(ExternalResources)
+
+	request := reconcile.Request{
+		NamespacedName: types.NamespacedName{
+			Name:      "ocsinit",
+			Namespace: "",
+		},
+	}
+	cr := &corev1.Secret{
```
jarrpa, line 407: This should not be `cr`, name it something like `externalSecret`. Don't forget to call `reconciler.client.Create()` as well.

pkg/controller/storagecluster/initialization_reconciler_test.go
```diff
@@ -352,6 +395,59 @@ func assertExpectedResources(t assert.TestingT, reconciler ReconcileStorageClust
 	assert.Equal(t, len(expectedCos[0].OwnerReferences), 1)
 }
 
+func TestEnsureExternalStorageClusterResources(t *testing.T) {
```
jarrpa, line 398: The main issue with this test is that you forgot to create an initial StorageCluster CR that is marked as being `external: true`. So create a StorageCluster as is done in `TestInitStorageClusterResourcesUpdate`, and set that as `cr`.

pkg/controller/storagecluster/initialization_reconciler_test.go
rajatsingh25aug: addressed

jarrpa:
The ocs-operator-ci failure looks to be a problem with the CI. Running the test locally succeeded, so overriding it.

/override ci/prow/ocs-operator-ci

rajatsingh25aug:
Thanks all for this PR, Learned a lot 👯 
cc @jarrpa @leseb @aruniiird

PR #501, opened 2020-05-11
BUG 1834327 Add ROOK_CSI_ALLOW_UNSUPPORTED_VERSION to rook operator env

Madhu-1:
The ceph-CSI version of both upstream and downstream are different
ex: the upstream versions are 2.0.0,2.1.0 etc but downstream
built version are release-4.3,release4.4 etc.
we cannot verify the supported downstream version with upstream
version.so setting ROOK_CSI_ALLOW_UNSUPPORTED_VERSION to "true"
in rook operator ENV to skip the version check for the
CSI pods.


Madhu-1:
/assign @jarrpa 
CC @travisn

Madhu-1:
@travisn https://github.com/rook/rook/pull/5412 need to be backported to downstream to have a complete fix for this one

Madhu-1:
> @travisn [rook/rook#5412](https://github.com/rook/rook/pull/5412) need to be backported to downstream to have a complete fix for this one

this is already part of downstream https://github.com/openshift/rook/blob/release-4.5/pkg/operator/ceph/csi/spec.go#L594-L600

Madhu-1:
> @Madhu-1 - I am not entirely sure I fully understand the problem yet: Are you saying that the the ceph-csi branch release-4.5, which is based on the upstream 2.1.0 version, builds to have the software report `release-4.5` as version, and _not_ the actual software version 2.1.0?...  That would be really surprising to me!

version is the cephcsi version tag we will be passing when building the cephcsi binary as in [1].

Yes, the version in the downstream builds reports is the version it is built for i.e release-4.2 release-4.3, etc but in upstream its upstream release version i.e v2.1.0 v2.1.1 etc. as both are different the rook validation will fail due to the different built version [2]
we have two options here disable the version check in downstream(IMO version check  will not be that much useful in downstream)  or provide the upstream version as version tag when building the cephcsi downstream in [1]

[1] https://github.com/ceph/ceph-csi/blob/master/Makefile#L39 
[2] https://github.com/rook/rook/blob/master/pkg/operator/ceph/csi/version.go#L29-L32

@travisn what's your suggestion on this one?

travisn:
> Yes, the version in the downstream builds reports is the version it is built for i.e release-4.2 release-4.3, etc but in upstream its upstream release version i.e v2.1.0 v2.1.1 etc. as both are different the rook validation will fail due to the different built version [2]
we have two options here disable the version check in downstream(IMO version check will not be that much useful in downstream) or provide the upstream version as version tag when building the cephcsi downstream in [1]

@Madhu-1 Agreed that we don't have a reason to keep the version check downstream. We test and verify the version we ship with. The check is really just needed upstream where we don't want to support all combinations.

obnoxxx:
@Madhu-1  writes:
> version is the cephcsi version tag we will be passing when building the cephcsi binary as in [1].
> 
> Yes, the version in the downstream builds reports is the version it is built for i.e release-4.2 release-4.3, etc but in upstream its upstream release version i.e v2.1.0 v2.1.1 etc. as both are different the rook validation will fail due to the different built version [2]
we have two options here disable the version check in downstream(IMO version check will not be that much useful in downstream) or provide the upstream version as version tag when building the cephcsi downstream in [1]

Ugh, this is really surprising to me.
Being an old-school RPM guy, I am so used to our downstream builds having the upstream version and maybe a build revision or a special tag that indicates that this is a d/s spin-off of the upstream version. Claiming that ceph-csi is "4.5" when there is only 2.1 in the upstream is really confusing to me. IMHO this shouldn't be done. But then again, maybe this is me being old-school... ;-)

> [1] https://github.com/ceph/ceph-csi/blob/master/Makefile#L39
> [2] https://github.com/rook/rook/blob/master/pkg/operator/ceph/csi/version.go#L29-L32

obnoxxx:
This error in the ocs-operator-e2e-aws tests is weird:

```
Compiling functests...
Failed to compile functests:
go: downloading k8s.io/apimachinery v0.0.0-20190404173353-6a84e37a896d 
...
 go: finding github.com/matttproud/golang_protobuf_extensions v1.0.1
# runtime/cgo
exec: "gcc": executable file not found in $PATH
make: *** [Makefile:102: build-functest] Error 1 
```

wtf? @jarrpa - any clue?

obnoxxx:
https://github.com/openshift/ocs-operator/pull/506 - this may be an attempt to fix this... but it does not seem to be sufficient

jarrpa:
> Being an old-school RPM guy, I am so used to our downstream builds having the upstream version and maybe a build revision or a special tag that indicates that this is a d/s spin-off of the upstream version. Claiming that ceph-csi is "4.5" when there is only 2.1 in the upstream is really confusing to me. IMHO this shouldn't be done. But then again, maybe this is me being old-school... ;-)

I'm with you too, this looks bad. But that's a discussion for another thread.

Madhu-1:
@jarrpa we need to backport this to release-v4.5, will the bot is going to do it or we have to do it manually?

PR #505, opened 2020-05-12
Adding OBC StorageClass in converged mode

aruniiird:
Made the required changes to provide OBC StorageClass in converged mode.


aruniiird:
@jarrpa , @leseb , can you please take a look at this PR...

FYI: @rajatsingh25aug , you may have to rebase your PR#500 on top of this, so that you will get the OBC StorageClass at your disposal...

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -61,6 +61,23 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
...
+	retSC := &storagev1.StorageClass{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: fmt.Sprintf("%s-ceph-retain-bucket", initData.Name),
+		},
+		Provisioner:   fmt.Sprintf("%s.%s/bucket", initData.Namespace, initData.ClusterName),
+		ReclaimPolicy: &reclaimPolicy,
+		Parameters: map[string]string{
+			"objectStoreName":      fmt.Sprintf("%s-object-store", initData.Namespace),
+			"objectStoreNamespace": initData.Namespace,
+			"region":               "us-east-1", // TODO: how to get the current region
```
aruniiird, line 75: Can someone help me out here to get the current `region`, rather than a hardcoded region?
@leseb , in your sample OBC SC, you have provided `region` as `us-east-1`, but I guess we are trying to add the current region.

leseb: Let's keep it hardcoded for now, we don't need to really expose this.

leseb: It's more an S3 artifact to used to create the bucket and is present for historical reason AFAIR.

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -61,6 +61,23 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
 	return nil
 }
 
+func (r *ReconcileStorageCluster) newOBCStorageClass(initData *ocsv1.StorageCluster) *storagev1.StorageClass {
+	reclaimPolicy := corev1.PersistentVolumeReclaimRetain
+	retSC := &storagev1.StorageClass{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: fmt.Sprintf("%s-ceph-retain-bucket", initData.Name),
```
leseb, line 68: the retain or delete name should be based on `reclaimPolicy`.

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -61,6 +61,23 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
...
+		},
+		Provisioner:   fmt.Sprintf("%s.%s/bucket", initData.Namespace, initData.ClusterName),
```
leseb, line 70:
```suggestion
		Provisioner:   fmt.Sprintf("%s.ceph.rook.io/bucket", initData.Namespace),
```

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -61,6 +61,23 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
...
+		ReclaimPolicy: &reclaimPolicy,
+		Parameters: map[string]string{
+			"objectStoreName":      fmt.Sprintf("%s-object-store", initData.Namespace),
```
leseb, line 73:
```suggestion
			"objectStoreName":     generateNameForCephObjectStore(initData),
```

pkg/controller/storagecluster/initialization_reconciler.go
obnoxxx: Does it need to be an existing region then? Or can we just make up a name, so that it is not as confusing? :thinking:

pkg/controller/storagecluster/initialization_reconciler.go
aruniiird: @leseb , from where will `reclaimPolicy` get that value? Currently `reclaimPolicy` is just a local variable and is not set from any outside condition.
From your sample SC from the converged mode doc,
```json
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: rook-ceph-retain-bucket
provisioner: openshift-storage.ceph.rook.io/bucket
reclaimPolicy: Retain
parameters:
 objectStoreName: openshift-storage-object-store
 objectStoreNamespace: openshift-storage
 region: us-east-1
```
and it just sets the `reclaimPolicy` to `Retain`, which is the value of `corev1.PersistentVolumeReclaimRetain`.

pkg/controller/storagecluster/initialization_reconciler.go
aruniiird: Addressed...

pkg/controller/storagecluster/initialization_reconciler.go
aruniiird: Addressed...

pkg/controller/storagecluster/initialization_reconciler.go
aruniiird: Ok, I got your point, please ignore the above comment...

aruniiird: Done...

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -61,6 +61,23 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
 	return nil
 }
 
+func (r *ReconcileStorageCluster) newOBCStorageClass(initData *ocsv1.StorageCluster) *storagev1.StorageClass {
+	reclaimPolicy := corev1.PersistentVolumeReclaimRetain
```
leseb, line 65: Looks like csi uses `PersistentVolumeReclaimDelete`, let's follow the same pattern

pkg/controller/defaults/placements.go
```diff
@@ -213,5 +213,41 @@ var (
 				},
 			},
 		},
+
+		"rgw-multi-instances": rook.Placement{
```
jarrpa, line 217: Why are you creating a new entry? Can't you just modify the `rgw` one?

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -61,6 +61,23 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
...
+	retSC := &storagev1.StorageClass{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: fmt.Sprintf("%s-ceph-%s-bucket", initData.Name, reclaimPolicy),
+		},
+		Provisioner:   fmt.Sprintf("%s.ceph.rook.io/bucket", initData.Namespace),
+		ReclaimPolicy: &reclaimPolicy,
+		Parameters: map[string]string{
+			"objectStoreName":      generateNameForCephObjectStore(initData),
+			"objectStoreNamespace": initData.Namespace,
+			"region":               "us-east-1",
```
jarrpa, line 75: I can't imagine we want to hardcode a value here, do we?

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -106,6 +123,7 @@ func (r *ReconcileStorageCluster) newStorageClasses(initData *ocsv1.StorageClust
 				"csi.storage.k8s.io/node-stage-secret-namespace":  initData.Namespace,
 			},
 		},
+		r.newOBCStorageClass(initData),
```
jarrpa, line 126: My understanding is that we do not want to create the OBC StorageClass in all scenarios. There are only two scenarios we want to deploy it in:

1. External StorageClusters that have an RGW endpoint defined in their JSON blob
2. Converged StorageClusters that are not in a cloud environment, similar to the behavior in `ensureCephObjectStores()`

pkg/controller/defaults/placements.go
aruniiird: The `rgw` entry can still be reused on a single `instance` mode. As you have pointed out we don't require `OBC` StorageClass in all scenarios, we may be switching between the (rgw / rgw-multi-instances) `Placement`s...

pkg/controller/storagecluster/initialization_reconciler.go
leseb: We do, just like the CSI SCs.

pkg/controller/storagecluster/initialization_reconciler.go
leseb: Let me double-check for 2.

leseb: Ok looks like there is a mechanism in place already whether we should deploy CephObjectStore, so we should create the SC only if that's the case.

pkg/controller/storagecluster/initialization_reconciler.go
aruniiird: Done...

pkg/controller/storagecluster/initialization_reconciler.go
aruniiird: @leseb , @jarrpa , made the change accordingly...

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -121,19 +157,13 @@ func (r *ReconcileStorageCluster) ensureCephObjectStores(instance *ocsv1.Storage
...
-	}
-	if platform == PlatformGCP {
-		r.reqLogger.Info("not creating a CephObjectStore because the platform is GCP")
-		return nil
-	}
-	if platform == PlatformAzure {
-		r.reqLogger.Info("not creating a CephObjectStore because the platform is Azure")
+	if r.isCloudPlatform(platform) {
+		r.reqLogger.Info("Cloud Platform: '" + string(platform) +
+			"', detected. Not creating CephObjectStore.")
```
leseb, line 162:
```suggestion
		r.reqLogger.Infof("Cloud Platform: %q, detected. Not creating CephObjectStore.", string(platform))
```

Doesn't `reqLogger` have an `Infof`?

aruniiird: Nope... It only has `Info()` method for logging... Or I could use `fmt.Sprintf()` inside the `Info()` method...

leseb: 😨 ok...

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -121,19 +157,13 @@ func (r *ReconcileStorageCluster) ensureCephObjectStores(instance *ocsv1.Storage
...
+		// we should exit from this function early itself.
+		instance.Status.CephObjectStoresCreated = true
```
aruniiird, line 164: @jarrpa , in places where ever we make changes to `instance.Status.<variables>`, are we not supposed to call,
`r.client.Status().Update(context.TODO(), instance)`
method and check the return error status.

pkg/controller/defaults/placements.go
```diff
@@ -213,5 +213,41 @@ var (
...
+						LabelSelector: &metav1.LabelSelector{
+							MatchExpressions: []metav1.LabelSelectorRequirement{
+								metav1.LabelSelectorRequirement{
+									Key:      "app",
+									Operator: metav1.LabelSelectorOpIn,
+									Values:   []string{"rook-ceph-rgw"},
+								},
+							},
+						},
+						TopologyKey: "kubernetes.io/hostname",
```
Madhu-1, line 230: we can make use of label which are already defined in kube package https://github.com/kubernetes/api/blob/master/core/v1/well_known_labels.go#L20

pkg/controller/defaults/placements.go
```diff
@@ -213,5 +213,41 @@ var (
...
 		},
+
+		"rgw-multi-instances": rook.Placement{
+			PodAntiAffinity: &corev1.PodAntiAffinity{
+				RequiredDuringSchedulingIgnoredDuringExecution: []corev1.PodAffinityTerm{
+					corev1.PodAffinityTerm{
+						LabelSelector: &metav1.LabelSelector{
+							MatchExpressions: []metav1.LabelSelectorRequirement{
+								metav1.LabelSelectorRequirement{
+									Key:      "app",
```
Madhu-1, line 224: we can define key and value as a variable and use it, as its used in multiple places.

aruniiird: @Madhu-1 , `metav1.LabelSelectorRequirement` is a `struct` not a `map[string]string`, so I am not sure how will I put an already defined key/value variable pair in their place.

Yes, I see that "app" string is used in lot of places, we could put that into a variable and use it in all the places...

Madhu-1: I meant to define a variable for app and rook-ceph-rgw. We can do it later that should not block this pr

pkg/controller/storagecluster/initialization_reconciler.go
jarrpa: This is already taken care of in this block, which runs after every `ensure` function: https://github.com/openshift/ocs-operator/blob/master/pkg/controller/storagecluster/reconcile.go#L239-L266

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -121,19 +157,13 @@ func (r *ReconcileStorageCluster) ensureCephObjectStores(instance *ocsv1.Storage
...
-	if platform == PlatformAzure {
-		r.reqLogger.Info("not creating a CephObjectStore because the platform is Azure")
+	if r.isCloudPlatform(platform) {
+		reqLogger.Info("Cloud Platform: '" + string(platform) +
+			"', detected. Not creating CephObjectStore.")
+		// we should exit from this function early itself.
+		instance.Status.CephObjectStoresCreated = true
 		return nil
 	}
-
```
jarrpa, line 136: Undo this removal.

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -107,10 +124,29 @@ func (r *ReconcileStorageCluster) newStorageClasses(initData *ocsv1.StorageClust
...
+	}
 	return ret, nil
 }
 
+// isCloudPlatform returns true if the platform is cloud type
+// requires an explicit argument of type 'CloudPlatformType',
+// so that getting the platform and dealing with errors are
+// done outside of this function
+func (r *ReconcileStorageCluster) isCloudPlatform(platform CloudPlatformType) bool {
+	cloudPlatfroms := map[CloudPlatformType]struct{}{
```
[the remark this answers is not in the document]
aruniiird: In this map way, it will be easier to add or remove a platforms from the cloud-platform (user or developer don't have to write another if condition or delete a block of code to remove it).

> This way you don't have to maintain a separate list here.
but we are any way maintaining a list either through map or directly in the code (through a sequence of if conditions)

But if you strongly oppose or prefer the condition way more, will change it.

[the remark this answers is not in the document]
aruniiird: Please ignore the above comment, just now only I saw the `ValidCloudPlatforms`.. will change the code...

pkg/controller/defaults/placements.go
```diff
@@ -213,5 +214,41 @@ var (
...
+							LabelSelector: &metav1.LabelSelector{
+								MatchExpressions: []metav1.LabelSelectorRequirement{
+									metav1.LabelSelectorRequirement{
+										Key:      app,
+										Operator: metav1.LabelSelectorOpIn,
+										Values:   []string{"rook-ceph-rgw"},
+									},
+								},
+							},
+							TopologyKey: "topology.kubernetes.io/zone",
```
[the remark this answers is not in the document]
aruniiird: I couldn't use it, because in my go `k8s-api` package don't have that particular `Topology` (above L24th variable) ... Tried to update `go mod vendor` didn't see any change in the `well_known_labels.go` file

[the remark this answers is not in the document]
jarrpa: Ohh... I wonder if it's not in the version we're vendoring...

What about this one? https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/core/v1/well_known_labels.go#L22 It will still be supported for a while, we can change it once the topology labels are in our vendors.

pkg/controller/defaults/placements.go
aruniiird: Thanks for the clarification @jarrpa . Changed the existing `rgw` topology map itself.

pkg/controller/defaults/placements.go
aruniiird: Thanks @Madhu-1 ... Changed accordingly...

aruniiird:
> More comments left. In addition, there are a number of things going on here that should be in their own commits:
> 
> 1. Using well-known label variables instead of strings and using an `app` variable
> 2. Creating the `isCloudPlatform()` function
> 3. Changing the CephObjectStore configuration (and Placement?)
> 4. Adding the OBC StorageClass
> 
> Please split up your commit along those lines.

Separated each changes into different logical commits.
@jarrpa , @leseb , @Madhu-1 , please take a look...

pkg/controller/defaults/placements.go
```diff
@@ -7,6 +7,7 @@ import (
 )
 
 var (
+	app = "app"
```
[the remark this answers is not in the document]
aruniiird: Done...

pkg/controller/defaults/placements.go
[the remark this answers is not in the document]
aruniiird: Ok... Made the change accordingly...

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -107,10 +124,29 @@ func (r *ReconcileStorageCluster) newStorageClasses(initData *ocsv1.StorageClust
 			},
 		},
 	}
-
+	// OBC StorageClass will be created only if it is not a cloud platform
+	if platform, err := r.platform.GetPlatform(r.client); err == nil &&
+		!r.isCloudPlatform(platform) {
```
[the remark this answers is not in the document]
aruniiird: Done...

pkg/controller/storagecluster/initialization_reconciler.go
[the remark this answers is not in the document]
aruniiird: Changed the function accordingly...

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -121,19 +157,13 @@ func (r *ReconcileStorageCluster) ensureCephObjectStores(instance *ocsv1.Storage
...
-		return nil
-	}
-	if platform == PlatformGCP {
-		r.reqLogger.Info("not creating a CephObjectStore because the platform is GCP")
-		return nil
-	}
-	if platform == PlatformAzure {
-		r.reqLogger.Info("not creating a CephObjectStore because the platform is Azure")
+	if r.isCloudPlatform(platform) {
+		reqLogger.Info("Cloud Platform: '" + string(platform) +
```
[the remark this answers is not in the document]
aruniiird: Done.

pkg/controller/storagecluster/initialization_reconciler.go
aruniiird: Done...

aruniiird:
> It looks like your third commit could still be two commits: one to make the changes to the RGW Placement and the CephObjectStore config, and another for adding the OBC StorageClass.
> 
> Even so, did you mean to remove the NodeAffinity and Toleration from the RGW Pod? I would imagine they should remain in the default Placement. If not, explain why.

That was my bad (sleepy head) (=
Can you please check now...

aruniiird:
> Even so, did you mean to remove the NodeAffinity and Toleration from the RGW Pod? I would imagine they should remain in the default Placement. If not, explain why.

@leseb ^^^

leseb:
> > Even so, did you mean to remove the NodeAffinity and Toleration from the RGW Pod? I would imagine they should remain in the default Placement. If not, explain why.
> 
> @leseb ^^^

@jarrpa we are adding anti-affinity because we have 2 instances instead of one so we want them to run on a different machine, just like we do for CephFilesystem.

Makese sense?

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -111,6 +111,19 @@ func (r *ReconcileStorageCluster) newStorageClasses(initData *ocsv1.StorageClust
 	return ret, nil
 }
 
+// isCloudPlatform returns true, if the platform is of cloud type
+// requires an explicit argument of 'CloudPlatformType',
+// so that getting the platform and dealing with errors are
+// all done outside of this function
+func isCloudPlatform(platform CloudPlatformType) bool {
+	for _, eachPlatform := range ValidCloudPlatforms {
```
umangachapagain, line 139:
```suggestion
	for _, validCloudPlatform := range ValidCloudPlatforms {
```

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -121,19 +134,12 @@ func (r *ReconcileStorageCluster) ensureCephObjectStores(instance *ocsv1.Storage
...
-	if platform == PlatformAWS {
-		r.reqLogger.Info("not creating a CephObjectStore because the platform is AWS")
-		return nil
-	}
-	if platform == PlatformGCP {
-		r.reqLogger.Info("not creating a CephObjectStore because the platform is GCP")
+	if isCloudPlatform(platform) {
+		reqLogger.Info(fmt.Sprintf("not creating a CephObjectStore because the platform is '%s'", platform))
+		// as the platform is not supported, we should not re-enter the function
+		instance.Status.CephObjectStoresCreated = true
```
umangachapagain, line 160: Logger says we are not creating CephObjectStore. Why does the status say `True` (i.e Created)?

pkg/controller/defaults/placements.go
```diff
@@ -104,29 +104,21 @@ var (
 		},
 
 		"rgw": rook.Placement{
-			NodeAffinity: &corev1.NodeAffinity{
```
umangachapagain, line 105: Why are we removing this?

pkg/controller/defaults/placements.go
```diff
@@ -104,29 +104,21 @@ var (
...
+									Operator: metav1.LabelSelectorOpIn,
+									Values:   []string{"rook-ceph-rgw"},
 								},
 							},
 						},
+						TopologyKey: corev1.LabelHostname,
 					},
 				},
-			},
-			Tolerations: []corev1.Toleration{
```
umangachapagain, line 119: Why are we removing this?

aruniiird:
@jarrpa , met all the review / concerns raised... Can you please take a look...

pkg/controller/storagecluster/initialization_reconciler.go
aruniiird: If you don't put the status true, the code will re-enter the function (till this point) every time operator reconciles and we don't want to do that...

pkg/controller/defaults/placements.go
aruniiird: Can you check the answer provided by @leseb for a similar question
https://github.com/openshift/ocs-operator/pull/505/#issuecomment-628454073

pkg/controller/defaults/placements.go
aruniiird: See the above comment..

pkg/controller/storagecluster/initialization_reconciler.go
aruniiird: Changed...

pkg/controller/storagecluster/initialization_reconciler.go
umangachapagain: So we need to change when/how reconcile is triggered. We should not be reporting false status. This is something the user can see.

pkg/controller/defaults/placements.go
umangachapagain: It explains why we have PodAntiAffinity. It does not explain why we removed NodeAffinity and Tolerations which are defaults.

pkg/controller/defaults/placements.go
umangachapagain: https://github.com/openshift/ocs-operator/pull/505/#discussion_r425237838

umangachapagain:
> @jarrpa , met all the review / concerns raised... Can you please take a look...

3rd commit still needs to be broken.
Please mark all resolved conversations as resolved, so it's clear what is left out.

aruniiird:
> Commit 3 does more than one thing (CSV update, Placement change, Gateway spec change). Good idea to break it since they're not related.
> Commit 3 and 4 both have CSV changes. But I don't see changes that should be affecting CSV.

3rd update, `Changes for 'GatewaySpec'`, is basically one entity all related to `GatewaySpec` change (to be specific, `Instances: 2` change) and we changed the `Placement` because of the multiple instances.

I postponed the CSV creation to the last commit... I will amend the said commit with CSV updations...

aruniiird:
@umangachapagain ,

> > Commit 3 does more than one thing (CSV update, Placement change, Gateway spec change). Good idea to break it since they're not related.
> > Commit 3 and 4 both have CSV changes. But I don't see changes that should be affecting CSV.
> 
> 3rd update, `Changes for 'GatewaySpec'`, is basically one entity all related to `GatewaySpec` change (to be specific, `Instances: 2` change) and we changed the `Placement` because of the multiple instances.
> 
> I postponed the CSV creation to the last commit... I will amend the said commit with CSV updations...

I re-ran the following `make` commands for verification, (on top of the above said commit)
`make ocs-operator`
`make update-generated`
`make gen-latest-csv`

but I couldn't see any CSV changes (except for the time timestamp updations). The reason, I could think of is that, these are internal specs and not something which we populate to a user.

pkg/controller/storagecluster/initialization_reconciler.go
aruniiird: Ok... I have removed that line, but again just want to bring the attention back to the multiple call of `r.platform.GetPlatform()` function in the module. The function will be called for every reconcile, even when we realized (in the previous attempt itself) that platform is non-usable.

Addressing that issue can be go in as a performance bug fix, and should not block this PR.

pkg/controller/defaults/placements.go
aruniiird: I have created the `Placement` according to the sample @leseb has provided in the doc. I had a chat with @leseb , he mentioned that the provided placement is the minimal / required sample, so I will add the `Toleration` and `NodeAffinity` (in addition to the changes)

pkg/controller/storagecluster/initialization_reconciler.go
aruniiird: Added back the original `rgw` placement with changes required for the anti-affinity (into the `Changes for 'GatewaySpec', required for a multi-instances` commit)...
@umangachapagain , please take a look...

pkg/controller/defaults/placements.go
aruniiird: Added back the original `rgw` placement with changes required for the anti-affinity (into the `Changes for 'GatewaySpec', required for a multi-instances` commit)...
@umangachapagain , please take a look...

pkg/controller/defaults/placements.go
aruniiird: Brought it back, with the latest changes...

aruniiird:
@jarrpa , @umangachapagain , can you take look at the changes and let me know any further changes are required or not...

pkg/controller/defaults/placements.go
```diff
@@ -125,20 +127,34 @@ var (
 				},
 			},
 			PodAntiAffinity: &corev1.PodAntiAffinity{
+				RequiredDuringSchedulingIgnoredDuringExecution: []corev1.PodAffinityTerm{
```
umangachapagain, line 130: In this case, if there's lesser nodes available, pod will never come up. Is that the desired behaviour?

leseb: Isn't that the same issue for CephFilesystem already?

aruniiird:
> Code looks good overall. Now please add a unit test for `newOBCStorageClass()`.

@jarrpa , added the new StorageClass into the tests, can you please take a look...

pkg/controller/storagecluster/initialization_reconciler_test.go
```diff
@@ -191,6 +198,14 @@ func assertExpectedResources(t assert.TestingT, reconciler ReconcileStorageClust
 
 	expected, err := reconciler.newStorageClasses(cr)
 	assert.NoError(t, err)
+	if isValidCloudPlatform(reconciler.platform.platform) {
```
jarrpa, line 201: This is not quite sufficient, since it will never fully test the code path in a single run. You need to simulate both cloud and non-cloud platforms. Leave this test as the non-cloud environment, then copy it into a new test function called something like `TestCloudInitStorageClusterResourcesUpdate` where you repeat the test for all valid cloud platforms. You may want to take the bulk of `TestInitStorageClusterResourcesUpdate` into its own utility function that you can call from your new test function.

pkg/controller/storagecluster/initialization_reconciler_test.go
```diff
@@ -191,6 +198,14 @@ func assertExpectedResources(t assert.TestingT, reconciler ReconcileStorageClust
...
+		assert.Equal(t, len(expected), 2)
+	} else {
+		assert.Equal(t, len(expected), 3)
+		request.Name = "ocsinit-ceph-Delete-bucket"
+		err = reconciler.client.Get(nil, request.NamespacedName, actualSc3)
+		assert.NoError(t, err)
```
jarrpa, line 207: Test more than just thing. Do the ownerref and field checks the other StorageClasses get.

pkg/controller/storagecluster/initialization_reconciler_test.go
jarrpa: Alternatively, refactor `TestInitStorageClusterResourcesUpdate` itself to cycle through all valid cloud platforms (and one non-cloud test)

pkg/controller/storagecluster/initialization_reconciler_test.go
aruniiird: Done...

aruniiird:
@jarrpa , I have made some refactoring to the `initialization_reconciler_test.go` file.
Added the tests for OBC storageclass as well.
Can you please take a look...

pkg/controller/storagecluster/initialization_reconciler_test.go
aruniiird: Added cloud / non-cloud platform tests for `Update` and `Create`

pkg/controller/defaults/placements.go
jarrpa: I think this makes sense, yes.

pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -61,6 +61,23 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
 	return nil
 }
 
+func (r *ReconcileStorageCluster) newOBCStorageClass(initData *ocsv1.StorageCluster) *storagev1.StorageClass {
+	reclaimPolicy := corev1.PersistentVolumeReclaimDelete
+	retSC := &storagev1.StorageClass{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: fmt.Sprintf("%s-ceph-%s-bucket", initData.Name, reclaimPolicy),
```
jarrpa, line 68: I'm not sure if I missed this earlier, but I don't like this naming scheme. First, you should create a `generateNameForCephRgwSC()` function to hold the name generation. Second, the name should follow the pattern of the other StorageClasses, just `<name>-ceph-rgw`. Since we're only ever creating one of these right now, there's no need to specify the reclaimPolicy in the name.

aruniiird: Added `generateNameForCephRgwSC()` function
Added tests to cover all the platforms.

PR #506, opened 2020-05-12
adds gcc to build tools

umangachapagain:


umangachapagain:
> I think I see what you're trying to do here. But it doesn't seem to be sufficient, unfortunately.
>
It seems that CI will use Dockerfile from master branch to build. So, untill we merge this, we might not know if it's resolved. 
>
Also, even if CI passes on this one and we merge, we might just hit another dependency issue.
I tried it on local and it works for me, so we just need to keep fixing this until we have no more issues.

> Also, providing a bit more context to the commit message / PR text would be helpful. smiley
>
Added context.

PR #482, opened 2020-04-22
[WIP] Updating to operator-sdk 0.17.0

jackyalbo:
operator-sdk 0.17.0
k8s api 1.17.4
controller-runtime 0.5.2
golang 1.13

supported noobaa-version
noobaa-core 5.5.0-20200422
noobaa-operator: 2.3.0-20200505

Signed-off-by: jackyalbo <jalbo@redhat.com>

PR #509, opened 2020-05-13
Adds Structural schema to OCS CRDs

umangachapagain:
+  adds build machinery provided by OpenShift
    
    build-machinery-go is a library of Make targets and scripts
    provided by OpenShift and used by various projects. This
    library containts wide range of targets for Golang, OpenShift,
    Operators, and Kubernetes code-generation tools.
    
    Golang target can be used to set minimum Go version requirement,
    set `go env`, build binary, run tests etc.
    
    Operators and code-generation targets can be used to generate
    structural schema for CRDs, verify schema etc.

+  adds minimum Golang version requirement
    
    This commit sets minimum required Golang version to `1.13.3`.
    This is also the version used in CI for building and running
    operator tests.

+  adds structural schema to CRDs
    
    structural schema is mandatory on new versions of Kubernetes
    and OpenShift without which CRD validation fails. This is
    different from OpenAPI schema validation and is also used
    by OpenShift Console to display information about the Spec
    to the users.
 
   `make update-generated` adds structural schema to CRD. Internally
    it calls update-codegen-crds which wraps controller-gen for  
    updating CRDs.

Since this PR sets minimum Go version to `1.13.3` and we currently use `1.12` in Go modules, let's wait for #482 which will upgrade the code-base to Go 1.13 before enforcing it.


umangachapagain:
/hold
CRDs are not recognized. Looking into it.

PR #510, opened 2020-05-14
Avoid running osd being marked out

xenolinux:
This PR prevents running osd from being marked out if running osd id is
provided to the osd removal job template.

### DIFF
--- a/pkg/controller/storagecluster/reconcile.go
+++ b/pkg/controller/storagecluster/reconcile.go
@@ -1191,6 +1191,20 @@ func newCleanupJob(sc *ocsv1.StorageCluster) *batchv1.Job {
 		"app": "ceph-toolbox-job-${FAILED_OSD_ID}",
 	}
 
+	// The purgeOSDScript finds osd status for given FAILED_OSD_ID whether it's up or down. The action will be taken according to osd status. If osd is up and running, it won't be marked out. If osd is down it can be taken out of the cluster and purged.
+	const purgeOSDScript = `
+set -x
+
+osd_status=$(ceph osd tree | grep "osd.${FAILED_OSD_ID} " | awk '{print $5}') 
+if [[ "$osd_status" == "up" ]]; then 
+  echo "OSD ${FAILED_OSD_ID} is up and running."
+  echo "Please check if you entered correct ID of failed osd!"
+else 
+  echo "OSD ${FAILED_OSD_ID} is down. Proceeding to mark out and purge"
+  ceph osd out osd.${FAILED_OSD_ID} 
+  ceph osd purge osd.${FAILED_OSD_ID} --force --yes-i-really-mean-it
+fi`
+
 	job := &batchv1.Job{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "Job",
@@ -1238,7 +1252,11 @@ func newCleanupJob(sc *ocsv1.StorageCluster) *batchv1.Job {
 									ReadOnly:  true,
 								},
 							},
-							Command: []string{"bash", "-c", "ceph osd out osd.${FAILED_OSD_ID};ceph osd purge osd.${FAILED_OSD_ID} --force --yes-i-really-mean-it"},
+							Command: []string{
+								"/bin/bash",
+								"-c",
+								purgeOSDScript,
+							},
 						},
 					},
 					InitContainers: []corev1.Container{


pkg/controller/storagecluster/reconcile.go
```diff
@@ -1190,6 +1190,11 @@ func newCleanupJob(sc *ocsv1.StorageCluster) *batchv1.Job {
 		"app": "ceph-toolbox-job-${FAILED_OSD_ID}",
 	}
 
+	const purgeOSDScript = `set -x
+if [[ \"$osd_status\" == \"up\" ]]; then echo \"OSD ${FAILED_OSD_ID} is up and running. Please check if you entered correct ID of failed osd!\"; else echo \"OSD ${FAILED_OSD_ID} is down. Proceeding to mark out and purge\"
```
travisn, line 1194: Do you really need to escape the quotes? Since it's inside the multi-line string, that's not needed for the golang string. Was it failing in the job without the quotes?

pkg/controller/storagecluster/reconcile.go
```diff
@@ -1190,6 +1190,11 @@ func newCleanupJob(sc *ocsv1.StorageCluster) *batchv1.Job {
...
+ceph osd out osd.${FAILED_OSD_ID} 
+ceph osd purge osd.${FAILED_OSD_ID} --force --yes-i-really-mean-it; fi`
```
travisn, line 1196: nit: i'd suggest `fi` on the next line for readability of the script

pkg/controller/storagecluster/reconcile.go
travisn: Would be helpful for readability if the `else` is also on the next line. In other words, the pattern:
```
if <condition>; then
  <action>
else
  <action>
fi
```

pkg/controller/storagecluster/reconcile.go
```diff
@@ -1190,6 +1190,11 @@ func newCleanupJob(sc *ocsv1.StorageCluster) *batchv1.Job {
...
+
```
leseb, line 1197:
```suggestion
	const purgeOSDScript = `
set -x
if [[ "$osd_status" == "up" ]]; then 
  echo "OSD ${FAILED_OSD_ID} is up and running."
  echo "Please check if you entered correct ID of failed osd!"
else
  echo "OSD ${FAILED_OSD_ID} is down. Proceeding to mark out and purge"
  ceph osd out osd.${FAILED_OSD_ID} 
  ceph osd purge osd.${FAILED_OSD_ID} --force --yes-i-really-mean-it
fi`

```

pkg/controller/storagecluster/reconcile.go
sabose, line 1194: also add the command to get the osd_status for the passed osd_id

pkg/controller/storagecluster/reconcile.go
```diff
@@ -1190,6 +1190,11 @@ func newCleanupJob(sc *ocsv1.StorageCluster) *batchv1.Job {
 		"app": "ceph-toolbox-job-${FAILED_OSD_ID}",
 	}
 
+	const purgeOSDScript = `set -x
```
travisn, line 1193: You're missing the line from the [script](https://gist.github.com/travisn/782cf2ef9f4956d7799d658fca12942c) that sets `$osd_status`.

pkg/controller/storagecluster/reconcile.go
xenolinux: ack

pkg/controller/storagecluster/reconcile.go
xenolinux: ack

xenolinux: ack

pkg/controller/storagecluster/reconcile.go
xenolinux: yes

pkg/controller/storagecluster/reconcile.go
leseb: @xenolinux please use this, it has proper indent.

xenolinux: okay

xenolinux:
> Please squash your commits.

yep squashing it

xenolinux:
> It looks like you reverted instead of squashed.

yeah. Fixing it

xenolinux:
> One more squash

okay

travisn:
@xenolinux After you test the template by installing the private OCS build, could you post the template spec here for review? For example, the output of `oc get template ocs-osd-removal -o yaml`. Thanks!

xenolinux:
> Your commit message has 2 sign-off, one is enough

Changed back to one sign-off

xenolinux:
> @xenolinux After you test the template by installing the private OCS build, could you post the template spec here for review? For example, the output of `oc get template ocs-osd-removal -o yaml`. Thanks!

Okay. I will post it

xenolinux:
> Can you add a better commit message that explains what you did?

Sure

pkg/controller/storagecluster/reconcile.go
```diff
@@ -1237,7 +1251,11 @@ func newCleanupJob(sc *ocsv1.StorageCluster) *batchv1.Job {
 									ReadOnly:  true,
 								},
 							},
-							Command: []string{"bash", "-c", "ceph osd out osd.${FAILED_OSD_ID};ceph osd purge osd.${FAILED_OSD_ID} --force --yes-i-really-mean-it"},
+							Command: []string{
+								"/bin/bash",
+								"-c",
+								fmt.Sprintf(purgeOSDScript),
```
travisn, line 1257: no need to format it
```suggestion
								purgeOSDScript,
```

xenolinux: okay. I will change it

PR #512, opened 2020-05-15
Bug 1834327: Add ROOK_CSI_ALLOW_UNSUPPORTED_VERSION to rook operator env

Madhu-1:
The ceph-CSI version of both upstream and downstream are different
ex: the upstream versions are 2.0.0,2.1.0 etc but downstream
built version are release-4.3,release4.4 etc.
we cannot verify the supported downstream version with upstream
version.so setting ROOK_CSI_ALLOW_UNSUPPORTED_VERSION to "true"
in rook operator ENV to skip the version check for the
CSI pods.

(cherry picked from commit c79848481039e97a217036b9e6f61c7a4af3dd10)


jarrpa:
We don't need backport PRs for 4.5 yet, we haven't hit feature freeze.

PR #521, opened 2020-05-18
Update resource requests/limits: osd,mds,noobaa

mmgaggle:
Update OSD resources now that THP issue is addressed.

https://bugzilla.redhat.com/show_bug.cgi?id=1836359

Update MDS and Noobaa CPU request so we can fit OCS in a smaller footprint

https://bugzilla.redhat.com/show_bug.cgi?id=1836367

### DIFF
--- a/pkg/controller/defaults/resources.go
+++ b/pkg/controller/defaults/resources.go
@@ -12,11 +12,11 @@ var (
 		"osd": corev1.ResourceRequirements{
 			Requests: corev1.ResourceList{
 				corev1.ResourceCPU:    resource.MustParse("1"),
-				corev1.ResourceMemory: resource.MustParse("4Gi"),
+				corev1.ResourceMemory: resource.MustParse("5Gi"),
 			},
 			Limits: corev1.ResourceList{
-				corev1.ResourceCPU:    resource.MustParse("2"),
-				corev1.ResourceMemory: resource.MustParse("8Gi"),
+				corev1.ResourceCPU:    resource.MustParse("3"),
+				corev1.ResourceMemory: resource.MustParse("5Gi"),
 			},
 		},
 		"mon": corev1.ResourceRequirements{
@@ -31,7 +31,7 @@ var (
 		},
 		"mds": corev1.ResourceRequirements{
 			Requests: corev1.ResourceList{
-				corev1.ResourceCPU:    resource.MustParse("3"),
+				corev1.ResourceCPU:    resource.MustParse("1"),
 				corev1.ResourceMemory: resource.MustParse("8Gi"),
 			},
 			Limits: corev1.ResourceList{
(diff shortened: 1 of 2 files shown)


mmgaggle:
CPU isn’t a eviction signal, so it shouldn’t be problematic. It’s a
compressible resource.

https://kubernetes.io/docs/tasks/administer-cluster/out-of-resource/#eviction-signals

https://kubernetes.io/docs/tasks/administer-cluster/out-of-resource/#evicting-end-user-pods

Doing it with memory would be risky.

guymguym:
Why does that changes the problem? OCS pods will still become non Guaranteed for scheduling due to `pod.status.qosClass: Burstable` so even if the eviction signal will be fired by another resource starvation (say due to inodes or whatever), it will be the OCS pods that get evicted over Guaranteed pods.

guymguym:
I think it's largely a matter of decision if OCS supports running as a "second class citizen" on Kubernetes nodes, because even CPU requests which doesn't match CPU limits will cause the entire pod qosClass to be reduced:
>For a Pod to be given a QoS class of Guaranteed:
>- Every Container in the Pod must have a memory limit and a memory request, and they must be the same.
>- Every Container in the Pod must have a CPU limit and a CPU request, and they must be the same.

pkg/controller/defaults/resources.go
```diff
@@ -12,11 +12,11 @@ var (
 		"osd": corev1.ResourceRequirements{
 			Requests: corev1.ResourceList{
 				corev1.ResourceCPU:    resource.MustParse("1"),
-				corev1.ResourceMemory: resource.MustParse("4Gi"),
+				corev1.ResourceMemory: resource.MustParse("5Gi"),
 			},
 			Limits: corev1.ResourceList{
-				corev1.ResourceCPU:    resource.MustParse("2"),
```
oritwas, line 18: The OSD will be Burstable Qos class in order to make it Guaranteed we should set the same value for the cpu as well.

obnoxxx:
/hold

This needs more discussion. Moving most of the pods to burstable QoS by lowering CPU requests is mostly what this PR does. It contradicts our earlier requirements to have guaranteed QoS. 

For the need to fit into smaller footprints, we might not want to change the general requests, but possibly establish a more dynamic mechanism. Like provide different sets of defaults for various scenarios. Alternatively, smaller footprints can be supported by custom `StorageCluster` CRs.

mmgaggle:
This is the path of least resistance to realizing a lower minimum footprint. 

We haven't made downward adjustments of our minimum supported sizing because a manual CR would be required. This is at odds with the suggestion that a custom CR could effect a smaller footprint.

I'm all for a more dynamic mechanism, but we'll need it, or something like this PR, to hit the minimum sizing criteria for 4.5.

RGW, MDS, and Noobaa (on non-EC2 infra) *depend* on OSDs being operational, which are not "guaranteed" pods.  I didn't think it would be so controversial to have those other components be consistent with how we're currently setting resources requests and limits on OSDs. Instead are we suggesting that OSDs should have requests = limits and become guaranteed?

PR #522, opened 2020-05-20
Support metadata pvc for a given OSD

kesavanvt:
**Description :**

- Rook has added support to configure a metadata PVC for an OSD in [https://github.com/rook/rook/pull/4818](https://github.com/rook/rook/pull/4818)

- This PR is about enabling the same support via OCS so that users have flexibility in configuring OSD using mixed devices - flash devices for metadata when only HDD is available for OSD.
- It can be used by creating a new metadataPVCTemplate.

A template will look like this:
```
dataPVCTemplate:
    spec:
      # IMPORTANT: Change the storage class depending on your environment
      storageClassName: local-block
      accessModes:
        - ReadWriteOnce
      resources:
        requests:
          # Replace the size with the size of localPV prepared
          storage: 500Gi
      volumeMode: Block
metadataPVCTemplate:
    spec:
      # IMPORTANT: Change the storage class depending on your environment
      storageClassName: local-block-meta
      accessModes:
        - ReadWriteOnce
      resources:
        requests:
          # Replace the size with the size of localPV prepared
          storage: 100Gi
      volumeMode: Block
```
**Design Doc**: [Support metadata pvc for a given OSD](https://docs.google.com/document/d/1rnqkoAytVpI3d-1vyqCeL9XTlAJhgXkeZW95WzRgUFY/edit?usp=sharing)


kesavanvt:
@jarrpa please review

sp98:
@kesavanvt  I believe that `Metadata` support in rook is only in version 1.3.x and not in 1.2.x.  Should that be of any concerns ?

kesavanvt:
> @kesavanvt I believe that `Metadata` support in rook is only in version 1.3.x and not in 1.2.x. Should that be of any concerns ?

The operator-sdk update has been merged [https://github.com/openshift/ocs-operator/pull/520](https://github.com/openshift/ocs-operator/pull/520), now OCS4.5 uses rook 1.3.x version.

sp98:
> As part of the Rook implementation, it's required that the metadata name of the template must be `name: metadata` to work. How is this handled by this code? Thanks

I don't think that's a hard requirement in rook.  [Rook checks if the we have more than one devices. And then uses the second device as `metadata` device](https://github.com/rook/rook/blob/25766746482e10aa97c9e776ae9bde3acc95ac9b/pkg/daemon/ceph/osd/daemon.go#L181-L198). 

[ Later it just uses the type to configure the `metadata` device.](https://github.com/rook/rook/blob/25766746482e10aa97c9e776ae9bde3acc95ac9b/pkg/daemon/ceph/osd/daemon.go#L416-L419) 

I might be missing something here.

leseb:
> > As part of the Rook implementation, it's required that the metadata name of the template must be `name: metadata` to work. How is this handled by this code? Thanks
> 
> I don't think that's a hard requirement in rook. [Rook checks if the we have more than one devices. And then uses the second device as `metadata` device](https://github.com/rook/rook/blob/25766746482e10aa97c9e776ae9bde3acc95ac9b/pkg/daemon/ceph/osd/daemon.go#L181-L198).
> 
> [ Later it just uses the type to configure the `metadata` device.](https://github.com/rook/rook/blob/25766746482e10aa97c9e776ae9bde3acc95ac9b/pkg/daemon/ceph/osd/daemon.go#L416-L419)
> 
> I might be missing something here.

@sp98 That is right, thanks for reminding me how my code works haha. I thought I had that has a requirement but it looks like I relaxed it somehow :-).

pkg/controller/storagecluster/reconcile.go
```diff
@@ -43,6 +43,11 @@ type StorageClassProvisionerType string
 // ensureFunc which encapsulate all the 'ensure*' type functions
 type ensureFunc func(*ocsv1.StorageCluster, logr.Logger) error
 
+const (
+	dataPVCName     = "data"
+	metadataPVCName = "metadata"
```
anubhav-here, line 48: I think it is a good practice to keep all the constants in the same block for the same file :)

kesavanvt: Done and noted :)

pkg/controller/storagecluster/reconcile.go
```diff
@@ -1054,6 +1068,12 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster) []rook.StorageClassDevi
 				Portable:             ds.Portable,
 				TuneSlowDeviceClass:  ds.Config.TuneSlowDeviceClass,
 			}
+
+			err := validateMetadataPvc(sc)
+			if err == nil {
+				ds.MetadataPVCTemplate.ObjectMeta.Name = metadataPVCName
+				set.VolumeClaimTemplates = []corev1.PersistentVolumeClaim{ds.DataPVCTemplate, ds.MetadataPVCTemplate}
+			}
```
jarrpa, line 1076: What happens when `err != nil`? Shouldn't we be handling that?

kesavanvt: `err != nil `means the MetadataPVCTemplate is not defined, 

Since MetadataPVCTemplate is optional , no modification is required in newCephCluster when MetadataPVCTemplate not defined, so handling wouldn't be required.

MetadataPVCTemplate is added to the volumeclaim list only when it is defined ( when `err == nil` )

jarrpa: I don't think that's a good idea. When an admin creates a StorageDeviceSet like this, he expects to have a metadata device. To simply ignore it and create the OSDs without it is bad since they can't just go back and fix it by changing the CR. We should return the error and refuse to continue reconciling.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -850,6 +852,18 @@ func determineFailureDomain(sc *ocsv1.StorageCluster) string {
 	return failureDomain
 }
 
+// validateMetadataPvc checks the MetadataPvc of the given
+// StorageCluster for completeness and correctness
+func validateMetadataPvc(sc *ocsv1.StorageCluster) error {
```
jarrpa, line 857: Looking at this again, this should not be its own function. This should be part of the `validateStorageDeviceSet()` function, with an `if` to check if the metadataPVCTemplate is set. Similarly the unit tests should be updated to reflect that.

pkg/controller/storagecluster/reconcile.go
kesavanvt: Correct me if i am wrong,

- if we return error and refuse to reconcile when metadatapvc is not defined, then providing the MetadataPVCTemplate in the storage cluster yaml becomes a mandatory thing. (i.e) user has to provide both DataPVCTemplate and MetadataPVCTemplate.

- Based on my understanding, the MetadataPVCTemplate is optional and the user can create the storage cluster even without providing the MetadataPVCTemplate in storage cluster yaml

pkg/controller/storagecluster/reconcile.go
kesavanvt: The reason for not integrating this check into `validateStorageDeviceSet()`.
- When on using only DataPVCTemplate , the below statements in reconcile.go would return an error 

```
		err = r.validateStorageDeviceSets(instance)
		if err != nil {
			reqLogger.Error(err, "Failed to validate StorageDeviceSets")
		}
	}
```

pkg/controller/storagecluster/reconcile.go
jarrpa: I would imagine you're missing a check for `ds.MetadataPVCTemplate != nil` or something.

anubhav-here:
@kesavanvt PR needs rebase

pkg/controller/storagecluster/reconcile.go
kesavanvt: Thanks, Fixed .
Implemented the template as a pointer for null check

pkg/controller/storagecluster/reconcile.go
kesavanvt: Validation for metadatapvc template has been integrated with `validateStorageDeviceSet()` method

jarrpa:
` uncommitted CSV changes. run 'make gen-latest-csv' and commit results. `

PR #523, opened 2020-05-22
adds support for appregistry

umangachapagain:
Brought back the old `make ocs-registry` which build ocs-registry in appregistry format.
We can still use `make operator-bundle` and `make operator-index` to use the new format.

All CI and Build scripts will use appregistry format by default.


hack/common.sh
```diff
@@ -7,6 +7,9 @@ GO111MODULE="on"
 GOPROXY="https://proxy.golang.org"
 GOROOT="${GOROOT:-go env GOROOT}"
 
+# Current DEV version of the CSV
+export CSV_VERSION=4.5.0
```
jarrpa, line 11: Follow the pattern for the other variables, create a DEFAULT for 4.5 and then accept a env variable override. Also, why does this need to be exported unlike everything else? I imagine it's so that it can be read by the csv_merger. Leave a comment to that effect.

umangachapagain: We don't have a default CSV version. It's holding the current CSV version.

PR #525, opened 2020-05-26
Added support for multus in OCS

rohan47:
[//]: # (Please describe your changes in detail)
[//]: # (If your changes affect the CR fields, a syntax for the CR change should be provided)
[//]: # (If a bug is fixed then the behavior change of before and after should be mentioned)
[//]: # (Be sure to run ocs-operator ci before pushing `make ocs-operator-ci`)
**Description of your changes:**
Multus enables pods to attach multiple networks.
The StorageClusterSpec now includes a network section where user can
specify public network and cluster network which references the
NetworkAttachmentDefinition object in multus.

Note: A working OCP cluster with multus is a prerequisite.
Also NetworkAttachmentDefinition objects should be created before creating the StorageCluster CR.

The StorageCluster CR has the following fields
```shell
  network:                             
    public: ocs-public  
    cluster: ocs-cluster
```
Here `ocs-public` and `ocs-cluster` are the NetworkAttachmentDefinition objects in multus which should exist before the StorageCluster CR is created.

**Which issue is resolved by this Pull Request:**
Resolves #

**Checklist:**

- [ ] Documentation has been updated, if necessary.
- [ ] Unit tests have been added, if necessary.
- [ ] Generate CSV (`make gen-latest-csv`)
- [ ] Raise Integration tests issue if necessary.
- [x] Code generation (`make update-generated`) has been run to update object specifications, if necessary.


rohan47:
Will add Unit tests

pkg/controller/storagecluster/reconcile.go
```diff
@@ -913,6 +913,15 @@ func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, nodeCount int, r
 	} else {
 		reqLogger.Info(fmt.Sprintf("No monDataDirHostPath, monPVCTemplate or storageDeviceSets configured for storageCluster %s", sc.GetName()))
 	}
+	multusNetwork := sc.Spec.Network
+	multusNetworkNil := ocsv1.NetworkSpec{}
+	multusSelector := make(map[string]string)
```
leseb, line 918: Create the map inside the condition instead?

pkg/apis/ocs/v1/storagecluster_types.go
```diff
@@ -28,6 +28,13 @@ type StorageClusterSpec struct {
 	MonDataDirHostPath string                                 `json:"monDataDirHostPath,omitempty"`
 	// Version specifies the version of StorageCluster
 	Version string `json:"version,omitempty"`
+	Network NetworkSpec `json:"network,omitempty"`
+}
+
+// NetworkSpec defines the spec for utilizing multiple networks
+type NetworkSpec struct {
+	Public string `json:"public,omitempty"`
```
umangachapagain, line 36: Add a comment about what are the expected values and what is the default.

pkg/apis/ocs/v1/storagecluster_types.go
```diff
@@ -28,6 +28,13 @@ type StorageClusterSpec struct {
...
+	Cluster string `json:"cluster,omitempty"`
```
umangachapagain, line 37: Add a comment about what are the expected values and what is the default.

pkg/apis/ocs/v1/storagecluster_types.go
```diff
@@ -28,6 +28,13 @@ type StorageClusterSpec struct {
 	MonDataDirHostPath string                                 `json:"monDataDirHostPath,omitempty"`
 	// Version specifies the version of StorageCluster
 	Version string `json:"version,omitempty"`
+	Network NetworkSpec `json:"network,omitempty"`
```
umangachapagain, line 31: Add a comment with a short description of this spec. The comment is converted into a description to assist users.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -913,6 +913,15 @@ func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, nodeCount int, r
...
+	if multusNetwork != multusNetworkNil {
```
umangachapagain, line 919: When we provide some random value in `sc.Spec.Network`, this condition is met and we use multus. Shouldn't we validate the user input?

pkg/controller/storagecluster/reconcile.go
```diff
@@ -913,6 +913,15 @@ func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, nodeCount int, r
...
+		cephCluster.Spec.Network.Provider = "multus"
```
umangachapagain, line 920: Export "multus" as a constant `multusNetworkProvider`.

pkg/controller/storagecluster/reconcile.go
rohan47: Do you mean fetching the NetworkAttachmentDefinition object by the value specified by the user and confirming it?
Well, it makes sense to validate this way.

leseb: The NAD is presented by the UI so this is safe, we will get something that exists. Later on Rook-Ceph will validate the NAD too. Not sure if we need to duplicate that validation "just" for the manual bootstrap case. However, I can see that making sense.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -924,6 +926,15 @@ func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, nodeCount int, r
 	} else {
 		reqLogger.Info(fmt.Sprintf("No monDataDirHostPath, monPVCTemplate or storageDeviceSets configured for storageCluster %s", sc.GetName()))
 	}
+	multusNetwork := sc.Spec.Network
```
leseb, line 929: What about adding a method to the spec to validate whether networking is enabled?
So you do something like `if sc.Spec.Network.IsMultus()` and then you enter your condition.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -924,6 +926,15 @@ func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, nodeCount int, r
...
+	multusNetworkNil := ocsv1.NetworkSpec{}
+	if multusNetwork != multusNetworkNil {
+		multusSelector := make(map[string]string)
+		cephCluster.Spec.Network.Provider = networkProvider
```
leseb, line 933: Instead of assigning individually, what about doing something like: `cephCluster.Spec.Network = ocsv1.NetworkSpec`? If the structure is identical we should be able to just pass it to Rook, right?

pkg/controller/storagecluster/reconcile.go
rohan47: Well, if we validate we make it more reliable and an error can be thrown even before the operator goes and creates CephCluster. 
Validation can be part of a followup PR.

pkg/controller/storagecluster/reconcile.go
rohan47: `cephCluster.Spec.Network` and `ocsv1.NetworkSpec` doesn't have identical structure so can't pass it directly to rook.

leseb: Why can't it be identical?

rohan47: I wanted to enforce `Public` and `Cluster` instead of the selectors as they are in Rook
I wanted to hardcode `multus` for the time being so not making `cephCluster.Spec.Network` and `ocsv1.NetworkSpec` identical

This is the Rook Network Spec
```go
type NetworkSpec struct {
	Provider  string `json:"provider"`
	Selectors map[string]string `json:"selectors"`
}
```

This is the one for OCS
```go
type NetworkSpec struct {
    Public  string `json:"public"`
    Cluster string `json:"cluster"`
}
```
So for the OCS spec the user is bound to specify Public and Cluster.

If this does not make sense then I will make both the spec same.

rohan47: Made the changes

pkg/controller/storagecluster/reconcile.go
rohan47: done

pkg/controller/storagecluster/storagecluster_controller_test.go
```diff
@@ -711,3 +712,48 @@ func TestNodeTopologyMapLabelSelector(t *testing.T) {
 	}
 	assert.Equal(t, nodeTopologyMap, sc.Status.NodeTopologies)
 }
+
+//TestStorageClusterOnMultus tests if multus configurations in StorageCluster are successfully applied to CephClusterCR
```
leseb, line 716:
```suggestion
// TestStorageClusterOnMultus tests if multus configurations in StorageCluster are successfully applied to CephClusterCR
```

pkg/controller/storagecluster/storagecluster_controller_test.go
```diff
@@ -711,3 +712,48 @@ func TestNodeTopologyMapLabelSelector(t *testing.T) {
...
+func TestStorageClusterOnMultus(t *testing.T) {
+	// When only public network is specified.
+	createStorageClusterWithMultus(t,"public-network","")
```
leseb, line 719:
```suggestion
	createStorageClusterWithMultus(t, "public-network", "")
```

leseb: Is it me or the linter should catch this.

pkg/controller/storagecluster/storagecluster_controller_test.go
```diff
@@ -711,3 +712,48 @@ func TestNodeTopologyMapLabelSelector(t *testing.T) {
...
+	// When only cluster network is specified .
+	createStorageClusterWithMultus(t,"","cluster-network")
+	// When both public and cluster network are specified.
+	createStorageClusterWithMultus(t,"public-network","cluster-network")
```
leseb, line 723: Case where nothing is specified?

pkg/controller/storagecluster/storagecluster_controller_test.go
```diff
@@ -711,3 +712,48 @@ func TestNodeTopologyMapLabelSelector(t *testing.T) {
 	}
 	assert.Equal(t, nodeTopologyMap, sc.Status.NodeTopologies)
 }
+
+//TestStorageClusterOnMultus tests if multus configurations in StorageCluster are successfully applied to CephClusterCR
+func TestStorageClusterOnMultus(t *testing.T) {
+	// When only public network is specified.
+	createStorageClusterWithMultus(t,"public-network","")
+	// When only cluster network is specified .
+	createStorageClusterWithMultus(t,"","cluster-network")
```
leseb, line 721: This should be an error case, where is the error?

pkg/controller/storagecluster/storagecluster_controller_test.go
umangachapagain: Looks like linter isn't catching a lot of these issues.

rohan47:
looking into CI issues

pkg/controller/storagecluster/storagecluster_controller_test.go
```diff
@@ -573,3 +574,54 @@ func TestNodeTopologyMapLabelSelector(t *testing.T) {
...
+// TestStorageClusterOnMultus tests if multus configurations in StorageCluster are successfully applied to CephClusterCR
+func TestStorageClusterOnMultus(t *testing.T) {
+	// When only public network is specified.
+	createStorageClusterWithMultus(t, "public-network", "")
+	// When both public and cluster network are specified.
+	createStorageClusterWithMultus(t, "public-network", "cluster-network")
+	// When only cluster network is specified, this will be an error case .
+	createStorageClusterWithMultus(t, "", "cluster-network")
+	// When public network and cluster network is empty, this will be an error case.
+	createStorageClusterWithMultus(t, "", "")
```
jarrpa, line 587: Instead of doing multiple function calls, specify this as a case table: https://github.com/openshift/ocs-operator/blob/master/pkg/controller/storagecluster/cephcluster_test.go#L18-L44 You can take the contents of `createStorageClusterWithMultus()` as the body of the `for` loop.

docs/multus.md
```diff
@@ -0,0 +1,109 @@
...
+
+The selector keys are required to be public and cluster where each represents:
+`public`: client communications with the cluster (reads/writes)
+`cluster`: internal Ceph replication network
+Based on the configuration, the operator will do the following:
+if only the public network is specified both communication and replication will happen on that network
+if both public and cluster networks are specified the first one will run the communication network and the second the replication network
+In order to work, each network key value must match a `NetworkAttachmentDefinition` object name in Multus. For example, you can do:
+`public: “my-public-storage-network”`
+`cluster: “my-replication-storage-network”`
```
jarrpa, line 50:
```suggestion
The selector keys are required to be public and cluster where each represents:
* `public`: client communications with the cluster (reads/writes)
* `cluster`: internal Ceph replication network

Based on the configuration, the operator will do the following:
* If only the public network is specified both communication and replication will happen on that network
* If both public and cluster networks are specified the first one will run the communication network and the second the replication network

In order to work, each network key value must match a `NetworkAttachmentDefinition` object name in Multus.
```

docs/multus.md
```diff
@@ -0,0 +1,109 @@
...
+
+For multus network provider, an already working cluster with Multus networking is required. Network attachment definition that later will be attached to the cluster needs to be created before the Cluster CRD. If Rook cannot find the provided Network attachment definition it will fail running the Ceph OSD pods. You can add the Multus network attachment selection annotation selecting the created network attachment definition on selectors.
+
+Following is an example of multus NetworkAttachmentDefinition
+Create a `public` and `cluster` network by creating a `NetworkAttachmentDefinition` object.
+Example NetworkAttachmentDefinition
```
jarrpa, line 56:
```suggestion
The following are an examples of valid multus NetworkAttachmentDefinitions:
```

pkg/controller/storagecluster/reconcile.go
```diff
@@ -55,6 +55,10 @@ osd_memory_target_cgroup_limit_ratio = 0.5
 	CleanupPolicyDelete CleanupPolicyType = "yes-really-destroy-data"
 	//Name of MetadataPVCTemplate
 	metadataPVCName = "metadata"
+	// Hardcoding networkProvider to multus and this can be changed later to accomodate other providers
+	networkProvider           = "multus"
+	publicNetworkSelectorKey  = "public"
+	clusterNetworkSelectorKey = "cluster"
```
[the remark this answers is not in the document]
rohan47: `cephcluster.go` is in the same package so can't define it there as well
Shall I move it to `cephcluster.go` ?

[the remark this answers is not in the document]
jarrpa: Er, yes, that. Bad English on my part. ;)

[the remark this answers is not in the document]
rohan47: done

PR #529, opened 2020-05-26
Update Rook version from 1.3.2 to 1.3.4

bipuladh:
Changes required to support the external cluster are only present in Rook 1.3.4.


deploy/olm-catalog/ocs-operator/manifests/ocs-operator.clusterserviceversion.yaml
```diff
@@ -1245,7 +1250,7 @@ spec:
                 - name: OPERATOR_NAME
                   value: ocs-operator
                 - name: ROOK_CEPH_IMAGE
-                  value: rook/ceph:v1.3.2
+                  value: rook/ceph:v1.3.4
```
travisn, line 1253: If you use this tag `rook/ceph:v1.3.4-4.gb97be10` for the image you will pick up the latest change for the default `data` field on the device sets.

bipuladh: ack.

PR #531, opened 2020-05-26
Bug 1833317: Added script to avoid healthy osd from marking out

obnoxxx:
This change prevents running osd from being marked out if running osd id is
provided by a user to the osd removal job template.
A constant purgeOSDScript is created with the bash scipt in it which finds the osd status of given a FAILED_OSD_ID.
If osd is up and running, osd will not be marked out. If osd status is down, it will be
taken out of the cluster and purged.

(cherry picked from commit 969317a8c350174766a200b7c595dc43c1b9bd79)


jarrpa:
Just gonna do this proactively since we already verified it doesn't break anything else in master.

/override ci/prow/ocs-operator-e2e-aws

PR #533, opened 2020-05-27
[WIP] Decode the encoded json blob from external secret

aruniiird:
Corrected the logic where an encoded json blob
was accidentally 'unmarshalled' directly into go struct.

Fixes: https://github.com/openshift/ocs-operator/issues/532


pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -100,14 +107,14 @@ func (r *ReconcileStorageCluster) ensureExternalStorageClusterResources(instance
...
 	}
-	ownerRef := metav1.OwnerReference{
-		UID:        instance.UID,
-		APIVersion: instance.APIVersion,
-		Kind:       instance.Kind,
-		Name:       instance.Name,
+	decodedJSONBlob, err := decodeJSONBlob(found.Data[externalClusterDetailsKey])
+	if err != nil {
+		reqLogger.Error(err, "failed to decode the json blob from external secret")
+		_ = r.client.Delete(context.TODO(), found)
```
umangachapagain, line 113: Why are we deleting this? And, why ignoring the return value?
What happens when Delete fails?

aruniiird: @umangachapagain ,

> Why are we deleting this?

We are trying to delete the secret, which has an invalid `data` part (or wrongly encoded `data`).
If we don't delete the secret, on each reconcile we will try to read the secret (again and again) and try to decode the badly encoded `data` (which we already failed to decode last time itself). There is no point in continuously doing this.

> Why ignore the return value?

We are already encountered an error (which is stored in `err` variable, 110 line no:), and has to be propagated up. There is no point in checking the deletion error. Even if we fail (err not nil case) reconcile will call `ensureExternalStorageClusterResources` and (as the secret is not yet deleted) we will reach this point again.

aruniiird:
> Remove all the unrelated changes. Rebase on latest master.

Already rebased to the latest upstream master...
All the file deletions in this PR is mainly done by `make gen-latest-csv` call...

pkg/controller/storagecluster/initialization_reconciler.go
aruniiird: Another point raised by @cloudbehl was that, it will be difficult for a user to check / diagnose what went wrong if we delete the secret. When told about the continuous error message (which will be raised on each reconcile call), he mentioned that in the UI the error will be shown only once (not each time OCS-Operator logs the error). So that the user won't be bothered much.

Will remove the deletion of the secret then...

aruniiird:
This is no longer needed, as the fix is provided in the rook-python script itself.

PR #536, opened 2020-05-28
go.mod: replace non-existent tag

raghavendra-talur:
The openshift/api project deleted their tags at some point, because of
which we get the following error for go commands:

```
go:
github.com/openshift/api@v3.9.1-0.20190924102528-32369d4db2ad+incompatible:
invalid pseudo-version: preceding tag (v3.9.0) not found
```

There is at least one more project which is referencing the tag(Noobaa)
which makes it harder to change the require line right now. Therefore, I
have modified the requirement through a replace line.

For more information refer to
https://github.com/openshift/api/issues/456

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -33,6 +33,7 @@ require (
 // Pinned to kubernetes-1.17.4
 replace (
 	github.com/Azure/go-autorest => github.com/Azure/go-autorest v13.3.2+incompatible // Required by OLM
+	github.com/openshift/api => github.com/openshift/api v0.0.0-20190924102528-32369d4db2ad
 	k8s.io/api => k8s.io/api v0.17.4
 	k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.17.4
 	k8s.io/apimachinery => k8s.io/apimachinery v0.17.4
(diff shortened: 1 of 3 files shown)


raghavendra-talur:
To clarify, we still refer to the same commit of openshift/api that we were referring to. It is only the reference to it that has changed from v3.9.1-1- to v0.0.0-

raghavendra-talur:
> Do we need to run `make deps-update` ?

Thanks! Done.

rohan47:
I have tested the changes.

raghavendra-talur:
> The patch looks clean and simple enough. @raghavendra-talur also explained why it's needed. But why doesn't my build fail? What is needed to hit the problem?


@obnoxxx and I debugged this a little more. There are two dirs on the system where the packages are available:

```
$GOPATH/pkg/mod/<URL of the package>
$GOPATH/pkg/mod/cache/download/<URL of the package>
```

After removing openshift/api from both these locations, `go mod tidy` and `go mod vendor` commands fail if the GOPROXY is set to "direct" only.

Also note that the Makefile sets the GOPROXY to `https://proxy.golang.org` and therefore `make deps-update` succeeds even if the shell has GOPROXY set to "direct".

raghavendra-talur:
> @raghavendra-talur ok. thanks for the explanation. So the change is correct. But we would not ever hit it because we are using a go proxy. ;-)

Right. That got me wondering why I found this error in the first place. So here is the reason; 

- vim-go and coc.nvim both use gopls as the language server for autocomplete functionality

- vscode uses gopls as the language server for autocomplete functionality 

and because go mod vendor wasn't working *outside* the Makefile, these tools were failing to parse/index the codebase.

PR #539, opened 2020-06-01
fix: OBC StorageClass for external cluster

leseb:
When deploying an external Rook-Ceph cluster, the OCS-operator does not
deploy an CephObjectStore CR, this means the StorageClass parameter
`objectStoreName` becomes invalid. We must remove it otherwise, the user
creation on OBC will fail.

### DIFF
--- a/pkg/controller/storagecluster/initialization_reconciler_test.go
+++ b/pkg/controller/storagecluster/initialization_reconciler_test.go
@@ -276,6 +276,11 @@ func assertExpectedResources(t assert.TestingT, reconciler ReconcileStorageClust
 		assert.Equal(t, expected[2].Provisioner, actualSc3.Provisioner)
 		assert.Equal(t, expected[2].ReclaimPolicy, actualSc3.ReclaimPolicy)
 		assert.Equal(t, expected[2].Parameters, actualSc3.Parameters)
+		// Doing a bit more validation for the RGW SC since some fields differ whether
+		// we do independent or converged mode, typically "objectStoreName" param must exist
+		assert.NotEmpty(t, actualSc3.Parameters["objectStoreName"], actualSc3.Parameters)
+		assert.NotEmpty(t, actualSc3.Parameters["region"], actualSc3.Parameters)
+		assert.Equal(t, 3, len(actualSc3.Parameters))
 	}
 
 	// The created StorageClasses should not have any ownerReferences set. Any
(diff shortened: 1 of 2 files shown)


pkg/controller/storagecluster/initialization_reconciler_test.go
```diff
@@ -276,6 +276,7 @@ func assertExpectedResources(t assert.TestingT, reconciler ReconcileStorageClust
 		assert.Equal(t, expected[2].Provisioner, actualSc3.Provisioner)
 		assert.Equal(t, expected[2].ReclaimPolicy, actualSc3.ReclaimPolicy)
 		assert.Equal(t, expected[2].Parameters, actualSc3.Parameters)
+		assert.Equal(t, 3, len(actualSc3.Parameters))
```
jarrpa, line 279: Wouldn't it be better to test if `actualSc2.Parameters["objectStoreName"]` exists? At the very least, add a comment explaining why this is being tested separately from the other parameters if it's indeed needed.

pkg/controller/storagecluster/initialization_reconciler_test.go
```diff
@@ -450,6 +451,12 @@ func assertExpectedExternalResources(t assert.TestingT, reconciler ReconcileStor
 			for param, value := range expected.Data {
 				assert.Equal(t, value, actual.Parameters[param])
 			}
+			if actual.Name == "ocsinit-ceph-rgw" {
+				assert.NotEmpty(t, actual.Parameters["endpoint"], actual.Parameters["endpoint"])
+				assert.NotEmpty(t, actual.Parameters["region"], actual.Parameters)
+				assert.Equal(t, actual.Parameters["region"], "us-east-1")
+				assert.Equal(t, 3, len(actual.Parameters), actual.Parameters)
+			}
```
jarrpa, line 459: Why are you making a special case for the RGW StorageClass? Isn't this already covered in the preceding for loop? If you're specifically checking for the absence of `objectStoreName`, you could just check to see if it exists in the array or not.

pkg/controller/storagecluster/initialization_reconciler_test.go
```diff
@@ -450,6 +451,12 @@ func assertExpectedExternalResources(t assert.TestingT, reconciler ReconcileStor
 			for param, value := range expected.Data {
 				assert.Equal(t, value, actual.Parameters[param])
 			}
+			if actual.Name == "ocsinit-ceph-rgw" {
```
jarrpa, line 454: I would think you should check for the generated name rather than a hard-coded name, no? Also, please add a comment explaining why this is a special case.

pkg/controller/storagecluster/initialization_reconciler_test.go
leseb: > Why are you making a special case for the RGW StorageClass? Isn't this already covered in the preceding for loop?

We need a special case for RGW SC because it is different whether we do independent or converged mode.
> If you're specifically checking for the absence of `objectStoreName`, you could just check to see if it exists in the array or not.

Fixed!

pkg/controller/storagecluster/initialization_reconciler_test.go
leseb: I can't really check for the generated name since the blob contains all the SC's names so I need to hardcode which one I want to address.

PR #540, opened 2020-06-02
Disable CephObjectStore when the cloud provider is Azure and gcp.

crombus:
when these cloud provider are used then default backing store should be used instead of rgw.

### DIFF
--- a/pkg/controller/storagecluster/initialization_reconciler.go
+++ b/pkg/controller/storagecluster/initialization_reconciler.go
@@ -448,7 +448,7 @@ func (r *ReconcileStorageCluster) ensureCephObjectStoreUsers(instance *ocsv1.Sto
 	if err != nil {
 		return err
 	}
-	if platform == PlatformAWS {
+	if isValidCloudPlatform(platform) {
 		return nil
 	}
 
--- a/pkg/controller/storagecluster/initialization_reconciler_test.go
+++ b/pkg/controller/storagecluster/initialization_reconciler_test.go
@@ -212,7 +212,7 @@ func createUpdateRuntimeObjects(cp *CloudPlatform) []runtime.Object {
 	}
 
 	// Create 'cephobjectstoreuser' only for non-aws platforms
-	if cp.platform != PlatformAWS {
+	if !isValidCloudPlatform(cp.platform) {
 		cosu := &cephv1.CephObjectStoreUser{
 			ObjectMeta: metav1.ObjectMeta{
 				Name: "ocsinit-cephobjectstoreuser",
@@ -329,7 +329,7 @@ func assertExpectedResources(t assert.TestingT, reconciler ReconcileStorageClust
 	}
 	request.Name = "ocsinit-cephobjectstoreuser"
 	err = reconciler.client.Get(nil, request.NamespacedName, actualCosu)
-	if reconciler.platform.platform == PlatformAWS {
+	if isValidCloudPlatform(reconciler.platform.platform) {
 		assert.Error(t, err)
 	} else {
 		assert.NoError(t, err)
--- a/pkg/controller/storagecluster/platform_detection.go
+++ b/pkg/controller/storagecluster/platform_detection.go
@@ -66,9 +66,6 @@ func (p *CloudPlatform) getPlatform(c client.Client) (CloudPlatformType, error) 
 }
 
 func isValidCloudPlatform(p CloudPlatformType) bool {
-	if p == PlatformUnknown {
-		return true
-	}
 	for _, cp := range ValidCloudPlatforms {
 		if p == cp {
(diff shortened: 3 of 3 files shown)


pkg/controller/storagecluster/initialization_reconciler.go
```diff
@@ -291,7 +291,7 @@ func (r *ReconcileStorageCluster) ensureCephObjectStoreUsers(instance *ocsv1.Sto
 	if err != nil {
 		return err
 	}
-	if platform == PlatformAWS {
+	if isValidCloudPlatform(platform) {
```
rohantmp, line 294: This will disable it for UPI environments too where the type will by `PlatformUnknown`. @obnoxxx is that is what is currently recommended? If not, I suggest with going with an approach similar to that in `ensureCephObjectStores`

obnoxxx: First of all, it is a bit counter-intuitive to have a function `isValidCloudPlatform()` return true for the valid cloud platforms **and for "unknown"** ...

Then, I agree that `ensureCephObjectStores()` and `ensureCephObjectStoreUsers()` should use the same mechansim. Ideally that's a function called in both places. So I'd expect the PR to factor the checking out of `ensureCephObjectStores()` in a first patch and then use that new function in `ensureCephObjectStoreUsers()`. That function could be a "method" under CloudPlatform itself.

obnoxxx: @crombus , @rohantmp

crombus: @obnoxxx  how about these changes https://github.com/openshift/ocs-operator/compare/65a83a40a6a924eca1d869187a4a26d0e9a3b499..190d683067652680d61db83ae6a441a87528f476

obnoxxx: @crombus That is the right direction, but I still sense some code duplication. The whole for loop can go into a function. the slightly confusing bit is that then the function is extremely similar to the existing `isValidCloudPlatform()`, just without the block

```
if p == PlatformUnknown {
    return true
}
```

So, instead of your `GetValidCloudPlatforms`, lets do something like:

```
func isValidCloudPlatformAndNotUnknown(p CloudPlatformType) bool {
    for _, cp := range ValidCloudPlatforms {
        return true
    }
    return false
}
```
And use this in `ensureCephObjectStores` like this:
```
    ...
    platform, err := r.platform.GetPlatform(r.client)
    if err != nil {
        return err
    }
    if isValidCouldPlatformAndNotUnknown(platform) {
        r.reqLogger.Info(fmt.Sprintf("not creating a CephObjectStore because the platform is %T", platform))
        return nil
    }
    ...
```

We just need a better name for the function `isValidCouldPlatformAndNotUnknown`, e.g. rename `isValidCloudPlatform` and use that name ;-)

obnoxxx: Okay... After all this back and forth, I just realized that when following @rohantmp 's recommendation to do what we do in `ensureCephObjectStores()`, I now realized that I had been looking at old code, and that this is using `isValidCoudPlatform()` since PR #505.... 

So doing what @crombus proposed in this PR makes this completely consistent with what `ensureCephObjectStores()` already does.

We can have a separate discussion of whether we want to include or exclude `unknown`.

obnoxxx:
I'm actually struggling to understand why the test is failing at verifying CephObjectStoreUser, and didn't previously fail at verifying CephObjectStore .... :thinking:

pkg/controller/storagecluster/initialization_reconciler.go
rohantmp: isValidCloudPlatform was originally just to ensure that the cached value was in the list of types.

We could have is isKnownCloudPlatform

crombus:
the test-case works fine if i add didderent if loops for each platform if platform == PlatformAWS {
                return nil
        }
        if platform == PlatformAzure {
                return nil
        }

obnoxxx:
@crombus - here is what is needed:

https://github.com/obnoxxx/ocs-operator/commit/f59a37f5093756596969044015aeaaaa51d810b3

my branch https://github.com/obnoxxx/ocs-operator/commits/pr/540 is based on top of this PR (and freshly rebased to master).

rohantmp:
https://github.com/rohantmp/ocs-operator/commit/8e2f8afd9d5f0d933e5c0b6f610d45e2893e3b09

My take:
Concerns addressed
- unit tests work now
- functests need to be updated
- don't disable for type: unknown (baremetal)

obnoxxx:
added the test fix patch and rebased to current master

PR #544, opened 2020-06-02
Makefile: run deps-update when go.mod changes

raghavendra-talur:
Whenever go.mod is changed, it is required that deps-update is run.
Change the Makefile to make it automatic.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,6 +48,7 @@ all: ocs-operator ocs-registry ocs-must-gather
 	deps-update
 
 deps-update:
+	@echo "Running deps-update"
 	go mod tidy && go mod vendor
 
 operator-sdk:
--- a/Makefile
+++ b/Makefile
@@ -44,7 +44,8 @@ all: ocs-operator ocs-registry ocs-must-gather
 	update-generated \
 	ocs-operator-ci \
 	red-hat-storage-ocs-ci \
-	unit-test
+	unit-test \
+	deps-update
 
 deps-update:
 	go mod tidy && go mod vendor
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,7 @@ operator-sdk:
 
 ocs-operator-openshift-ci-build: build
 
-build:
+build: deps-update
 	@echo "Building the ocs-operator binary"
 	mkdir -p build/_output/bin
 	env GOOS=$(TARGET_GOOS) GOARCH=$(TARGET_GOARCH) go build -i -ldflags="-s -w" -mod=vendor -o build/_output/bin/ocs-operator ./cmd/manager


umangachapagain:
Can you explain a bit about how this works for those of us who don't understand the make magic?

raghavendra-talur:
> As discussed offline, this change is wrong.
> 
> * With `deps-update` phony, it will always be executed, and the added dependency to `go.mod` does not have any effect. (Wouldn't have an effect without phony either, unless you create a file named `deps-update`...)
> * since nothing depends on the `deps-update` target, the pr also has no immediate effect. the right thing would be to have `build` depend on `deps-update`.

I have made these changes now:

- Build depends on deps-update
- deps-update is PHONY
- Fixed the PHONY list in a new commit by fixing a bug in the Makefile syntax

raghavendra-talur:
@umangachapagain @obnoxxx @jarrpa The Makefile has some lines in it before the `all` target (which is the first target in the file) because of which just running `make` does not build `all` target.


This is the output when `make` is run. I see that only two targets are built, `operator-sdk` and `update-generated`.
```
$ make --dry-run 
echo "Ensuring operator-sdk"
hack/ensure-operator-sdk.sh
Using existing controller-gen from "_output/tools/bin/controller-gen"
Using existing yq from "_output/tools/bin/yq"
'_output/tools/bin/controller-gen' schemapatch:manifests="./deploy/crds" paths="./pkg/apis/ocs/v1" output:dir="./deploy/crds"
echo Updating generated files
hack/generate-k8s-openapi.sh
echo Running update-codegen-crds
make update-codegen-crds
```

If I comment those lines, then the target `all` becomes the default target. Here is the output:
```
$ make --dry-run                                                                                                                                                                                                     
echo "Running deps-update"
go mod tidy && go mod vendor
echo "Building the ocs-operator binary"
mkdir -p build/_output/bin
env GOOS=linux GOARCH=amd64 go build -i -ldflags="-s -w" -mod=vendor -o build/_output/bin/ocs-operator ./cmd/manager
echo "Building the ocs-operator image"
hack/build-operator.sh
echo "Building ocs-registry image in appregistry format"
hack/build-appregistry.sh
echo "Building the ocs-must-gather image"
hack/build-must-gather.sh
```

I have a commit at the top of this branch to show the difference but I understand that it is not the right fix. If you guide me on the expected behavior of PR #509 , I will modify the commit.

obnoxxx:
@raghavendra-talur : See #550 for a fix for the `all` target. And an unrelated fix to PHONY...

obnoxxx:
Oh, I see that you already spotted the missing newline escape in PHONY definition here. I can remove it from #550. ;-)

obnoxxx:
And for the record, the fix for the `all` target is not required for this PR. The main thing is that `build` triggers `deps-update`, not that `all` does it.

FWIW, the actual change to have build depend on deps-update looks good.
But I would actually split it into 3 patches:
1. `build` depends on `deps-update`
1. make `deps-update` phony
1. add the echo output
I know that not everyone would agree to that. ;-)

raghavendra-talur:
> And for the record, the fix for the `all` target is not required for this PR. The main thing is that `build` triggers `deps-update`, not that `all` does it.
> 

I have removed the fixes for missing escape and the test patch for `all` target. Let us fix that in PR #550 

> FWIW, the actual change to have build depend on deps-update looks good.
> But I would actually split it into 3 patches:
> 
> 1. `build` depends on `deps-update`
> 2. make `deps-update` phony
> 3. add the echo output
> 
> I know that not everyone would agree to that. ;-)

Have split the original commit into 3 commits as per the suggestion.

PR #550, opened 2020-06-08
Makefile: let `make` do the same as `make all`

obnoxxx:
The `all` target is not executed when only `make` is invoked because of the include and call directives before it. This moves the all target above the includes so that it is the first target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,13 +10,13 @@ export GOPROXY=https://proxy.golang.org
 # Export GOROOT. Required for OPERATOR_SDK to work correctly for generate commands.
 export GOROOT=$(shell go env GOROOT)
 
+all: ocs-operator ocs-registry ocs-must-gather
+
 include $(addprefix ./vendor/github.com/openshift/build-machinery-go/make/, \
   targets/openshift/crd-schema-gen.mk \
 )
 
 $(call add-crd-gen,ocsv1,./pkg/apis/ocs/v1,./deploy/crds,./deploy/crds)
-
-all: ocs-operator ocs-registry ocs-must-gather
 
 .PHONY: \
 	build \


obnoxxx:
@raghavendra-talur @jarrpa  - FYI

raghavendra-talur:
@obnoxxx IIUC, we need the include directive before the ocs-operator target as it generates some files that is required for the build of ocs-operator. 

I think @umangachapagain and @jarrpa can verify if the above is correct.

obnoxxx:
@raghavendra-talur my patch does not move the `ocs-operator` target before the include. Only the `all` target. And even so, afaik, the order of the targets wouldn't matter much in the Makefile, except for the first target which would be executed by `make`...

raghavendra-talur:
> @raghavendra-talur my patch does not move the `ocs-operator` target before the include. Only the `all` target. And even so, afaik, the order of the targets wouldn't matter much in the Makefile, except for the first target which would be executed by `make`...

I still think that is going to be a problem. By moving the `all` target above the include and the call directives, only the following targets are being built.

- ocs-operator
- ocs-registry
- ocs-must-gather

I don't know under what conditions the CRDs are to be generated and at what points. I don't believe a fix is possible by moving the include directive, rather it would require creating a new target that calls the makefile in the build-machinery dir.

obnoxxx:
Moved the PHONY fix to a separate PR as this one here might be more controversial. :-)

obnoxxx:
@raghavendra-talur :
>  I still think that is going to be a problem. By moving the all target above the include and the call directives, only the following targets are being built.
>
> * ocs-operator
> * ocs-registry
> * ocs-must-gather

This is precisely what the `all` target is defined to trigger. `make -n all` has exactly the same output with and without this patch.

The purpose of this patch is to have `make` do the same as `make all`. (Usually the sole purpose of the `all` target is to be called from `make`.)

We can raise a separate topic if you think that the `all` target is not complete. 

> I don't know under what conditions the CRDs are to be generated and at what points. I don't believe a fix is possible by moving the include directive, rather it would require creating a new target that calls the makefile in the build-machinery dir.

I don't think that we usually need to call the generation targets.
If we do, we need to invoke them / depend on them in some of our targets (like build / all / ...).
But I think these are mostly needed to be envoked (explicitly) if we use a new sdk version or if we change something in our custom resource definitions etc.

Anyway, we shouldn't rely on these being invoked by accident, just because they are the first target.

@umangachapagain can confirm

PR #554, opened 2020-06-09
Don't set CSI provisioner node affinity in default config

obnoxxx:
The CSI provisioners are not specific to the ceph cluster.
Rook can manage multiple ceph clusters.
And so the csi provisioners should not have the node affinities
set (by default).


obnoxxx:
@rohantmp @cloudbehl @anmolsachan @anubhav-here @umangachapagain FYI

obnoxxx:
/retest

the errors look strange. works for me locally

obnoxxx:
Updated to remove the `CSI_PROVISIONER_NODE_AFFINITY` from the CSV too.

PR #556, opened 2020-06-09
Delete the storageclasses during uninstall

raghavendra-talur:
At the time of installation, storageclasses are created for ceph-rbd and cephfs. We remove the same in the uninstall stage.

Failure to delete them is not considered fatal and does not block the uninstallation.

### DIFF
--- a/pkg/controller/storagecluster/uninstall_reconciler.go
+++ b/pkg/controller/storagecluster/uninstall_reconciler.go
@@ -1,0 +1,46 @@
+package storagecluster
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/go-logr/logr"
+	ocsv1 "github.com/openshift/ocs-operator/pkg/apis/ocs/v1"
+	storagev1 "k8s.io/api/storage/v1"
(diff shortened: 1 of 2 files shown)


obnoxxx:
@raghavendra-talur - what's the reason for holding the PR? The code looks good from a superficial look.

pkg/controller/storagecluster/uninstall_reconciler.go
```diff
@@ -0,0 +1,44 @@
...
+		return err
+	}
+	for _, sc := range scs {
+		existing := storagev1.StorageClass{}
+		err := r.client.Get(context.TODO(), types.NamespacedName{Name: sc.Name, Namespace: sc.Namespace}, &existing)
+
+		switch {
+		case err == nil:
+			if existing.DeletionTimestamp != nil {
+				reqLogger.Info(fmt.Sprintf("StorageClass %s is already marked for deletion", existing.Name))
```
obnoxxx, line 28: should we exit the flow here? Or do you want to schedule deletion again?

PR #557, opened 2020-06-09
Fixes to use empty labelSelector

anubhav-here:


pkg/controller/defaults/placements.go
```diff
@@ -9,24 +9,25 @@ import (
...
-					NodeSelectorTerms: []corev1.NodeSelectorTerm{
-						corev1.NodeSelectorTerm{
-							MatchExpressions: []corev1.NodeSelectorRequirement{
-								corev1.NodeSelectorRequirement{
-									Key:      NodeAffinityKey,
-									Operator: corev1.NodeSelectorOpExists,
-								},
-							},
+	// DefaultNodeSelector is the NodeAffinity to be used when labelSelector is nil
+	DefaultNodeSelector = &corev1.NodeAffinity{
```
umangachapagain, line 13: If it is `DefaultNodeAffinity`, why name it `DefaultNodeSelector`?

anubhav-here: Ah! Will fix that

PR #558, opened 2020-06-11
StorageCluster: delete the default node selector labels during uninstall

raghavendra-talur:
This PR removes the default node affinity labels from the worker nodes when the StorageCluster is deleted.

 If the StorageCluster spec provided a list of node selector labels or expressions, we don't remove them because a subset of them might be used by the user for other services    too.

### DIFF
--- a/pkg/controller/storagecluster/reconcile.go
+++ b/pkg/controller/storagecluster/reconcile.go
@@ -429,6 +429,27 @@ func (r *ReconcileStorageCluster) validateStorageDeviceSets(sc *ocsv1.StorageClu
 	return nil
 }
 
+func (r *ReconcileStorageCluster) getStorageClusterEligibleNodes(sc *ocsv1.StorageCluster, reqLogger logr.Logger) (*corev1.NodeList, error) {
+	nodes := &corev1.NodeList{}
+	selector, err := metav1.LabelSelectorAsSelector(sc.Spec.LabelSelector)
+	if err != nil {
+		return nodes, err
+	}
+	nodeMatchLabel := map[string]string{defaults.NodeAffinityKey: ""}
+	if sc.Spec.LabelSelector != nil {
+		err = r.client.List(context.TODO(), nodes, MatchingLabelsSelector{Selector: selector})
+		if err != nil {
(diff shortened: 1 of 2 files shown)


raghavendra-talur:
This PR needs a rebase after #556 is merged.

obnoxxx:
Rebased on top of master after #556 has been merged. @raghavendra-talur FYI

pkg/controller/storagecluster/reconcile.go
```diff
@@ -429,25 +430,25 @@ func (r *ReconcileStorageCluster) validateStorageDeviceSets(sc *ocsv1.StorageClu
...
 		}
 	}
-	return nodes, err
+
+	selector, err = metav1.LabelSelectorAsSelector(labelSelector)
+	err = r.client.List(context.TODO(), nodes, MatchingLabelsSelector{Selector: selector})
+	if err != nil {
+		return nodes, err
+	}
+	return nodes, nil
```
obnoxxx, line 451: This could be simplified to 

```suggestion
	return nodes, err
```

pkg/controller/storagecluster/reconcile.go
```diff
@@ -429,25 +430,25 @@ func (r *ReconcileStorageCluster) validateStorageDeviceSets(sc *ocsv1.StorageClu
...
-		}
+		labelSelector = sc.Spec.LabelSelector
 	} else {
-		err = r.client.List(context.TODO(), nodes, client.MatchingLabels(nodeMatchLabel))
-		if err != nil {
-			return nodes, err
+		labelSelector = &metav1.LabelSelector{
+			MatchLabels: map[string]string{defaults.NodeAffinityKey: ""},
 		}
 	}
```
obnoxxx, line 444: Just a nit but this could be streamlined even more (IMHO) to:

```
	labelSelector := &metav1.LabelSelector{
		MatchLabels: map[string]string{defaults.NodeAffinityKey: ""},
	}
	if sc.Spec.LabelSelector != nil {
		labelSelector = sc.Spec.LabelSelector
	}
```

PR #569, opened 2020-06-18
independent mode: expose rgw endpoint as a label

leseb:
When doing indepenent mode, we must instruct Noobaa where to find the
external Ceph rgw endpointhen we fetch the externalClusterDetailsSecret
Secret, we extract the Rgw StorageClass as well as its parameters.
Currently, we only have a single parameter called "endpoint", which is
used to configure OBC.
Now we also expose that endpoint as a label on the Nooba CR so Noobaa
can determine whether the endpoint is external or not.

### DIFF
--- a/pkg/controller/storagecluster/noobaa_system_reconciler.go
+++ b/pkg/controller/storagecluster/noobaa_system_reconciler.go
@@ -19,6 +19,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
+const (
+	// externalRgwEndpointLabelName is the name of the label that will be used by Noobaa to configure the S3 endpoint
+	externalRgwEndpointLabelName = "rgw-endpoint"
+)
+
 func (r *ReconcileStorageCluster) ensureNoobaaSystem(sc *ocsv1.StorageCluster, reqLogger logr.Logger) error {
 	// find cephCluster
 	foundCeph := &cephv1.CephCluster{}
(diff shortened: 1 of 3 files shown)


pkg/controller/storagecluster/noobaa_system_reconciler.go
```diff
@@ -87,6 +92,11 @@ func (r *ReconcileStorageCluster) setNooBaaDesiredState(nb *nbv1.NooBaa, sc *ocs
 	nb.Labels = map[string]string{
 		"app": "noobaa",
 	}
+	// If independent mode, we pass the endpoint value as a label to Noobaa
+	// so it can connect to the RGW S3 endpoint
+	if sc.Spec.ExternalStorage.Enable {
+		nb.Labels[externalRgwEndpointLabelName] = externalRgwEndpoint
```
jarrpa, line 98: Instead of saving the variable between functions, could you instead Get() the RGW StorageClass, verify it exists, and read the value from there? Then we could keep the consts in one file and avoid the global var.

jarrpa: This may also make it easier to refactor in case we ever want to use this for converged use cases, like in baremetal.

leseb: IMHO, it feels overkill to fetch the SC, this var is initialized at the same time we build the parameters for the SC.

The converged mode will never have such use case, this is a small patch for 4.5 only. 4.6 will change and have a similar experience for both converged and independent.

jarrpa: I also dislike one-off fixes that are supposed to be reverted later, but  as long as the whole process will be overhauled I'll live with it. :)

PR #580, opened 2020-06-23
encore rgw endpoint for external mode

leseb:
Since labels value cannot contain any `:` when Noobaa reads it when an
error like:

```
Error while reconciling: NooBaa.noobaa.io "noobaa" is invalid:
metadata.labels: Invalid value: "10.70.56.202:8080":
a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.',
and must start and end with an alphanumeric character
(e.g. 'MyValue', or 'my_value', or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')
```

So let's encode the endpoint and decode it later with Nooba.

### DIFF
--- a/pkg/controller/storagecluster/noobaa_system_reconciler_test.go
+++ b/pkg/controller/storagecluster/noobaa_system_reconciler_test.go
@@ -266,7 +266,8 @@ func assertNoobaaResource(t *testing.T, reconciler ReconcileStorageCluster) {
 	err = reconciler.client.Get(nil, request.NamespacedName, fNoobaa)
 	assert.NoError(t, err)
 	assert.NotEmpty(t, fNoobaa.Labels[externalRgwEndpointLabelName])
-	assert.Equal(t, fNoobaa.Labels[externalRgwEndpointLabelName], "10.20.30.40:50")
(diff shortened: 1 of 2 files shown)


pkg/controller/storagecluster/noobaa_system_reconciler_test.go
```diff
@@ -266,7 +266,7 @@ func assertNoobaaResource(t *testing.T, reconciler ReconcileStorageCluster) {
 	err = reconciler.client.Get(nil, request.NamespacedName, fNoobaa)
 	assert.NoError(t, err)
 	assert.NotEmpty(t, fNoobaa.Labels[externalRgwEndpointLabelName])
-	assert.Equal(t, fNoobaa.Labels[externalRgwEndpointLabelName], "10.20.30.40:50")
+	assert.Equal(t, fNoobaa.Labels[externalRgwEndpointLabelName], "MTAuMjAuMzAuNDA6NTA=")
```
umangachapagain, line 269: Can you add a comment that this is encoded value for `"10.20.30.40:50"` which we randomly chose? Just to save confusion later on.

PR #589, opened 2020-06-26
 Remove the host from crush map

xenolinux:
This change will remove the host associated with the failed osd from the
crush map. When node removal process is done and all osds have been
removed from that host, the empty host will be removed from crush map.

### DIFF
--- a/pkg/controller/storagecluster/reconcile.go
+++ b/pkg/controller/storagecluster/reconcile.go
@@ -850,9 +850,11 @@ func newCleanupJob(sc *ocsv1.StorageCluster) *batchv1.Job {
 	}
 
 	// The purgeOSDScript finds osd status for given FAILED_OSD_ID whether it's up or down. The action will be taken according to osd status. If osd is up and running, it won't be marked out. If osd is down it can be taken out of the cluster and purged.
+	// HOST_TO_REMOVE variable contains the host name associated with the failed osd. When all osds are removed from the host, the host will be removed from crush map
 	const purgeOSDScript = `
 set -x
 
+HOST_TO_REMOVE=$(ceph osd find osd.${FAILED_OSD_ID} | grep "host" | tail -n 1 | awk '{print $2}' | cut -d'"' -f 2)
 osd_status=$(ceph osd tree | grep "osd.${FAILED_OSD_ID} " | awk '{print $5}')
 if [[ "$osd_status" == "up" ]]; then
 echo "OSD ${FAILED_OSD_ID} is up and running."
@@ -861,6 +863,8 @@ else
 echo "OSD ${FAILED_OSD_ID} is down. Proceeding to mark out and purge"
 ceph osd out osd.${FAILED_OSD_ID}
 ceph osd purge osd.${FAILED_OSD_ID} --force --yes-i-really-mean-it
+echo "Attempting to remove the parent host. Errors can be ignored if there are other OSDs on the same host"
+ceph osd crush rm $HOST_TO_REMOVE
 fi`
 
 	job := &batchv1.Job{


xenolinux:
Labeled WIP because local testing of code is remaining.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -911,6 +913,7 @@ else
   echo "OSD ${FAILED_OSD_ID} is down. Proceeding to mark out and purge"
   ceph osd out osd.${FAILED_OSD_ID} 
   ceph osd purge osd.${FAILED_OSD_ID} --force --yes-i-really-mean-it
+  ceph osd crush rm $HOST_TO_REMOVE
```
travisn, line 916: Will errors be ignored for this command, right? We just need to confirm the job succeeds even if this fails.

xenolinux: Yeah. The job should succeed. The command will remove host successfully if none osd is present under the host if osd is present, the host won't be removed, since it's not empty. The command will throw such a message.

xenolinux: For eg: ```[root@rhhi-cap2m /]# ceph osd crush rm  rhhi-cap3m-lab-eng-blr-redhat-com
Error ENOTEMPTY: (39) Directory not empty```

xenolinux:
The job succeeds with the following output. 
~~~
mac ~ % oc logs -f ocs-osd-removal-2-bbh8c 
OSD 2 is down. Proceeding to mark out and purge.
marked out osd.2. 
purged osd.2
removed item id -3 name 'ip-10-0-140-29' from crush map  <--- removes host
~~~

travisn:
> The job succeeds with the following output.
> 
> ```
> mac ~ % oc logs -f ocs-osd-removal-2-bbh8c 
> OSD 2 is down. Proceeding to mark out and purge.
> marked out osd.2. 
> purged osd.2
> removed item id -3 name 'ip-10-0-140-29' from crush map  <--- removes host
> ```

LGTM. And for completeness what does the log look like if the host fails to be removed when there is another osd on the same host?

xenolinux:
> > The job succeeds with the following output.
> > ```
> > mac ~ % oc logs -f ocs-osd-removal-2-bbh8c 
> > OSD 2 is down. Proceeding to mark out and purge.
> > marked out osd.2. 
> > purged osd.2
> > removed item id -3 name 'ip-10-0-140-29' from crush map  <--- removes host
> > ```
> 
> LGTM. And for completeness what does the log look like if the host fails to be removed when there is another osd on the same host?

@travisn Yes. I will update that too.

xenolinux:
The case where the host is not empty and it won't get removed. The logs will look like this:
mac ~ % oc logs -f ocs-osd-removal-0-5pcqz
OSD 0 is down. Proceeding to mark out and purge.
marked out osd.0. 
purged osd.0
Error ENOTEMPTY: (39) Directory not empty

travisn:
> The case where the host is not empty and it won't get removed. The logs will look like this:
> 
> ```
> OSD 0 is down. Proceeding to mark out and purge.
> marked out osd.0. 
> purged osd.0
> Error ENOTEMPTY: (39) Directory not empty
> ```

Looks good. So that the user knows they can ignore the last error, perhaps always first print a message such as:
```
Attempting to remove the parent host. Errors can be ignored if there are other OSDs on the same host.
```

xenolinux:
> Will the `osd crush rm` just fail if there are still OSDs on the host? As such, wouldn't the Job fail and retry indefinitely?

@jarrpa If there are osds present `ceph osd crush rm` will return `Error ENOTEMPTY: (39) Directory not empty`. It will only remove the host form crush when none osds are present under that host. The job won't fail and retry indefinitely. The job will purge osd. After that `osd crush rm` will be run. Depending upon the host associated with the failed osd is empty or not, the host will/won't be taken out of the crush map.

jarrpa:
My understanding is that a Kubernetes Job will retry until it Completes successfully. In this case the Job is running `/bin/bash`, meaning that the script that's being executed has to return 0 (no error) for the command to succeed. As such, if the script ever returns an error, the Job will try again.

Does the script, as currently written, not try to run more than once? Will it always return 0?

travisn:
> My understanding is that a Kubernetes Job will retry until it Completes successfully. In this case the Job is running `/bin/bash`, meaning that the script that's being executed has to return 0 (no error) for the command to succeed. As such, if the script ever returns an error, the Job will try again.
> 
> Does the script, as currently written, not try to run more than once? Will it always return 0? What should the admin do if either `ceph osd out` or `ceph osd purge` fail and the OSD is never actually removed?

@xenolinux Jose has a good point that we need to confirm that when the host removal command fails, that it only prints an error and doesn't cause the job to restart again and again. The error message should be printed, but should not cause the job to restart. But since the restart policy on the job is `RestartPolicyNever` I think it should be working already even if it fails.

xenolinux:
@jarrpa  The job pod goes in `completed` state and it prints only printing error at the end.

PR #511, opened 2020-05-14
Creating a new noobaa SCC

jackyalbo:
We want to create a noobaa own secure context in order to fix these two issues:
https://bugzilla.redhat.com/show_bug.cgi?id=1835125
https://bugzilla.redhat.com/show_bug.cgi?id=1804168

Currently trying to validate if this fix indeed resolves our upgrade issue

Signed-off-by: jackyalbo jalbo@redhat.com

PR #591, opened 2020-06-28
Reverting Noobaa's SCC priority

jackyalbo:
Changing the priority leads to unexpected behaviour - so partially reverting
BZ https://bugzilla.redhat.com/show_bug.cgi?id=1851697

CI Failure issue: https://bugzilla.redhat.com/show_bug.cgi?id=1850148

Signed-off-by: jackyalbo jalbo@redhat.com

### DIFF
--- a/pkg/controller/ocsinitialization/sccs.go
+++ b/pkg/controller/ocsinitialization/sccs.go
@@ -184,8 +184,6 @@ func newNooBaaSCC(namespace string) *secv1.SecurityContextConstraints {
 	scc.Users = []string{
 		fmt.Sprintf("system:serviceaccount:%s:noobaa", namespace),
 	}
-	priority := int32(11) // 1 higher than anyuid default one
-	scc.Priority = &priority
 
 	return scc
 }


jackyalbo:
> Hi @jackyalbo, a few points:
> 
> * I don't fully understand the description of the PR / the commit message yet: The bug referenced (https://bugzilla.redhat.com/show_bug.cgi?id=1850148) is an openshift Bug and has nothing in our code to revert.
> * Instead, the SCC was introduced in PR #511, related to BZ https://bugzilla.redhat.com/show_bug.cgi?id=1804168 and as far as I can tell, this is not a complete revert of that patch, but just removes two lines.
> * Can you explain a little more what the "unexpected behavior" is that leads to this?

Hey @obnoxxx I updated my PR explanation, hope it makes things clearer:
When I wrote the PR there wasn't a BZ on us yet, only the CI issue it causes. 
Now we have one and I attached the link. Thanks for your feedback

jackyalbo:
@obnoxxx commit was updated as well, with your proposed format

PR #593, opened 2020-06-30
Move ObjectBucket and ObjectBucketClaim CRDs to owned instead of required.

obnoxxx:
This repeats the change of #424 that was undone in #517 because of problems in OLM that are fixed meanwhile.


obnoxxx:
@jarrpa @umangachapagain @nimrod-becker @jackyalbo FYI

PR #534, opened 2020-05-27
added rook-csi-resizer to relatedImages

dannyzaken:
* rook-csi-resizer image was added to `csv-merger.go` without adding it to the `relatedImages`. this prevents it from being automatically mirrored in a disconnected environment.
*  without the fix disconnected env setup requires another manual step

PR #595, opened 2020-07-01
Bug 1847982: Release 4.5: backport pr #534

obnoxxx:
Backport pr #534


obnoxxx:
@umangachapagain FYI.

But why are there no tests and no protections?
Is the bot not active?

PR #601, opened 2020-07-02
Design document for admission controllers in OCS

vbnrh:
This is a new branch for work done in https://github.com/openshift/ocs-operator/pull/390

Adds a design document for supporting admission controllers in OCS

### DIFF
--- a/docs/design/admission-webhooks.md
+++ b/docs/design/admission-webhooks.md
@@ -1,0 +1,67 @@
+## Background
+
+This proposal is to add support for admission webhooks for validation and mutation for API requests on storagecluster CRs
+
+An admission webhook are HTTP callbacks that intercept and pre-process requests to the Kubernetes API server for a specific resource, prior to persistence of the resource, but after the request is authenticated and authorized
+
+There are two type of admission webhooks:   
+ 
+**MutatingAdmissionWebhook** - During the admission process, the mutating admission webhooks can perform tasks on the resource, such as injecting affinity labels or filling default values
+
+**ValidatingAdmissionWebhook** - At the end of the admission process, the validating admission plug-in can be used to make sure an object is configured properly, for example ensuring affinity labels are as expected. If the validation passes, the api server will persist the changes to the resource, otherwise the api server will fail the request and return an error to the client
+
+Admission webhooks are implemented  as part of a specialized web server that is deployed on the cluster. 
+In order to expose the web server and the webhooks to the API server a MutatingWebhookConfiguration / ValidatingWebhookConfiguration with proper configuration are also needed to be installed on the cluster
+
+For more information on admission webhooks (admission plugins) please follow the links below:
+
+[General information about admission webhooks](https://docs.openshift.com/container-platform/4.5/architecture/admission-plug-ins.html)
+
+[API reference for the MutatingWebhookConfiguration resour  ce](
+https://docs.openshift.com/container-platform/4.4/rest_api/extension_apis/mutatingwebhookconfiguration-admissionregistration-k8s-io-v1.html)
+
+
+[API reference for the ValidatingWebhookConfiguration resource](https://docs.openshift.com/container-platform/4.4/rest_api/extension_apis/validatingwebhookconfiguration-admissionregistration-k8s-io-v1.html)
+
+## User Scenario
+
+The following is a list of use cases suggested to be solved using admission webhooks:
+
+1. Ensuring a finalizer exists on the storagecluster CR.  
+1. Prevent a request from changing spec.multiCloudGateway.reconcileStrategy from “Managed” to  any other value.  
+1. Ensuring that replica count cannot be configured after deviceSet is created.
+
+## Requirements
+
+1. Webhooks web server which will implement the validation and mutation operations
+2. ValidatingWebhookConfig / MutatingWebhookConfig installed in the cluster 
+3. TLS certificates to be used by the web server to allow authenticated and secure communication between the API server and the webhooks web server. 
+
+
+## Implementation 
+
+**Code**
+
+The web server will implement 2 specific paths:
+/mutate - Exposes the implementation of the mutation hooks for storagecluster CRs
+/validate - Exposes the implementation of the validation hooks for  (Validating Webhook)storagecluster CRs
+All other paths should return 404.
+
+**Image**  
+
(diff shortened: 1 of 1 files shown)


vbnrh:
@jarrpa @umangachapagain @aruniiird @travisn Please review

docs/design/crd-validation/README.md
```diff
@@ -0,0 +1,79 @@
+## Background
```
travisn, line 1: @jarrpa What is the right place for design docs in OCS? I would think this doc should have a name/path something like `/design/admission-controller.md`.

docs/design/crd-validation/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+This proposal is to add support for admission controllers in OCS. An admission controller is a piece of code that intercepts requests to the Kubernetes API server prior to persistence of the object, but after the request is authenticated and authorized
+
+ There are two special controllers: MutatingAdmissionWebhook and ValidatingAdmissionWebhook.
+ 
+ Mutating controllers may modify the objects they admit but validation controllers are only allowed to validate requests.
```
travisn, line 7: Do we need mutating controllers for OCS? I was assuming we just needed the validating controller to start with.

docs/design/crd-validation/README.md
```diff
@@ -0,0 +1,79 @@
...
+4. RBAC Components 
+
+## Implementation 
+
+As shown in the above diagram, the admission control process proceeds in two phases. In the first phase, mutating admission controllers are run. In the second phase, validating admission controllers are run. Note again that some of the controllers are both.
+
+The admission controllers intercept requests based on the values given in the configuration. In this config, we have to provide the details on 
+* What resources should it be looking for ? (Pods, Service)
+* What api version and group does it belong to ? *Example : ApiVersion = (v1, v1beta) Group version = (admissionregistration.k8s.io)*
+* What kind of operation should it intercept ? (Create, Update, Delete)
```
travisn, line 38: Why intercept a delete? Operators commonly use finalizers to handle those scenarios.

docs/design/crd-validation/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+As shown in the above diagram, In the OCP cluster, The management of certificates would be completely done by the Openshift Service CA Operator. This operator would be responsible for handling all the certificate management duties like generating and renewing them when OCS is deployed in Openshift Environment.
+
+The first step would be to generate the certificates and assign them to the proper components. On annotating any service with 
+
+`service.beta.openshift.io/serving-cert-secret-name:<secret-name>` 
+
+We have to deploy the the annotated service at the start to generate the signed certificates. This can done manually  through the command 
+
+`oc create -f ${SERVICE_NAME}.yaml`
```
travisn, line 60: To use admission controllers in OCS, we need it to be completely automated. We can't ask the admin to run any `oc` commands. For example, can we create the service via the CSV?

docs/design/crd-validation/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+**Output**
+```
+oc describe service <service-name>
+...
+Annotations:              service.beta.openshift.io/serving-cert-secret-name: <service-name>
+                          service.beta.openshift.io/serving-cert-signed-by: openshift-service-serving-signer@1556850837
+```
+The Openshift Sevice CA operator automatically generates a signed serving certificate and key pair into a secret in the same namespace as the service. 
+This secret will be mounted onto the server pod. Our server would be looking for tls certificates and automatically set them for TLS communication.
```
travisn, line 74: We'll need to understand the sequence of events. For example, would the OCS operator wait for the secret to be generated before launching the admission controller server? Or could we rely on the pod startup to wait for the secret to exist before starting? The latter seems better, but the pod would also appear in a failed state for some (hopefully short) time until the secret is created.

docs/design/crd-validation/README.md
vbnrh: Yes, thats correct we are implementing validatingwebhooks for the start. This is just a brief heads up on what type of admission controllers are available and what they do.

docs/design/crd-validation/README.md
vbnrh: It is an option that is available to use if we want :) Another briefing on the fields in validatingwebhookconfiguration. Let me know if you want that removed.

docs/design/crd-validation/README.md
vbnrh: added a section for `setup in ocs` which addresses this

docs/design/crd-validation/README.md
vbnrh: added a section for `setup in ocs` which addresses this

docs/design/crd-validation/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+If any of the controllers in either phase reject the request, the entire request is rejected immediately and an error is returned to the end-user.
+
+## Certificate Management in OCP
+
+![Prod_AC_Image](./ac-ocp.png)
+
+As shown in the above diagram, In the OCP cluster, The management of certificates would be completely done by the Openshift Service CA Operator. This operator would be responsible for handling all the certificate management duties like generating and renewing them when OCS is deployed in Openshift Environment.
+
+The first step would be to generate the certificates and assign them to the proper components. On annotating any service with
```
jarrpa, line 54:
```suggestion
The first step would be to generate the certificates and assign them to the proper components. The certificates are stored in Secrets, and the Secrets are referenced by the following annotation:
```

docs/design/crd-validation/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+![Prod_AC_Image](./ac-ocp.png)
+
+As shown in the above diagram, In the OCP cluster, The management of certificates would be completely done by the Openshift Service CA Operator. This operator would be responsible for handling all the certificate management duties like generating and renewing them when OCS is deployed in Openshift Environment.
+
+The first step would be to generate the certificates and assign them to the proper components. On annotating any service with 
+
+`service.beta.openshift.io/serving-cert-secret-name:<secret-name>` 
+
+We have to deploy the the annotated service at the start to generate the signed certificates. This can done manually  through the command.
```
jarrpa, line 58:
```suggestion
We have to deploy the the annotated service at the start to generate the signed certificates. This can done manually  through the command:
```

vbnrh:
remove-lifecycle stale

vbnrh:
@jarrpa @umangachapagain PTAL

jarrpa (approved):
/lgtm

PR #563, opened 2020-06-16
Refactor: Moving StorageClusterInit in a new file.

rajatsingh25aug:
Moved the logic for `StorageClusterInit` in a new file for better clarity and understanding.

Signed-off-by: RAJAT SINGH <rajasing@redhat.com>

PR #609, opened 2020-07-03
Undo and cleanly redo whitespace changes included in #563

obnoxxx:
PR #563 sneaked in some whitespace changes and both messed up the commits a bit and removed too many white spaces (removed intentation and not only superfluous trailing spaces). 

This makes backporting of patches in this area difficult.

This PR first undoes the changes (on top of PR #589) and then adds only the intended trailing whitespace patch separately on top. So if we need to backport more patches in this area in the future, we can just pick the top patch first and have clean cherry-picks.


obnoxxx:
@rajatsingh25aug - FYI

PR #613, opened 2020-07-07
Fix problematic permissions

umangachapagain:
- sdk 0.17.x recursively finds roles/rolebindings in bundlemanifests/
directory which causes issues in Permissions spec of CSV. So,
removing bundlemanifests/ while generating CSV and restoring it
afetrwards to bypass the recursive check.

- run `make gen-latest-csv` with new source-manifests to
get rid of the problematic permissions block.


hack/source-manifests.sh
```diff
@@ -25,9 +25,12 @@ if [ -z "$NOOBAA_IMAGE" ] || [ -z "$ROOK_IMAGE" ]; then
 fi
 
 # always start fresh and remove any previous artifacts that may exist.
+rm -rf $OCS_FINAL_DIR
```
obnoxxx, line 28: adding `rm -rf $OCS_FINAL_DIR` here, we can also remove this from `hack/generate-unified-csv.sh`

obnoxxx, line 28: I don't really know if this OCS_FINAL_DIR cleanup is entierly related to the change with the bundledmanifests. It seems it should be a separate commit (in this PR).

hack/source-manifests.sh
```diff
@@ -25,9 +25,12 @@ if [ -z "$NOOBAA_IMAGE" ] || [ -z "$ROOK_IMAGE" ]; then
...
 rm -rf $OUTDIR_TEMPLATES
 mkdir -p $OUTDIR_TEMPLATES
 mkdir -p $OUTDIR_CRDS $OUTDIR_BUNDLEMANIFESTS
+cp deploy/bundlemanifests/*.yaml $OUTDIR_BUNDLEMANIFESTS/
+rm deploy/bundlemanifests/*.yaml
```
obnoxxx, line 33: I think this is the main trick explained in the commit message. Since we are trying to trick the operator-sdk (if I understood it correctly), it might be good to add an explaning comment here (and at the bottom when we copy it back), since this is not obvious from reading the code.

hack/source-manifests.sh
umangachapagain: It is related. otherwise the same issue occurs as contents of `bundlemanifests/` are copied inside `OCS_FINAL_DIR` for packaging.

jarrpa:
We should see what we can do to avoid this workaround in the future.

hack/source-manifests.sh
obnoxxx: Where is this happening? It is not very obvious: before this patch, OCS_FINAL_DIR was not referenced in this script at all...

PR #618, opened 2020-07-08
Properly append LabelSelector

jarrpa:
The value of StorageCluster.Spec.LabelSelector was not being properly
appended to the generated Placements of the OCS components.
Specifically, adding a new NodeSelectorTerm results in a selector term that
is ORed with any other terms, whereas adding a new
NodeSelectorRequirement to an existing NodeSelectorTerm would result in
an AND of all requirements in that term. This commit makes it so the
LabelSelector field is ANDed with any other existing node selectors in
the Placement, properly omitting a selector if an empty LabelSelector is
specified.

### DIFF
--- a/pkg/controller/storagecluster/cephcluster.go
+++ b/pkg/controller/storagecluster/cephcluster.go
@@ -335,8 +335,7 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster) []rook.StorageClassDevi
 						Operator: corev1.NodeSelectorOpIn,
 						Values:   []string{topologyKeyValues[topologyIndex]},
 					}
-					nodeSelectorTerms := placement.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms
-					nodeSelectorTerms[0].MatchExpressions = append(nodeSelectorTerms[0].MatchExpressions, nodeZoneSelector)
+					appendNodeRequirements(&placement, nodeZoneSelector)
 				}
 			} else {
 				placement = ds.Placement
--- a/pkg/controller/storagecluster/placement.go
+++ b/pkg/controller/storagecluster/placement.go
@@ -23,13 +23,11 @@ func getPlacement(sc *ocsv1.StorageCluster, component string) rookv1.Placement {
 	if sc.Spec.LabelSelector == nil {
 		placement.NodeAffinity = defaults.DefaultNodeAffinity
 	} else {
-		term := convertLabelToNodeSelector(*sc.Spec.LabelSelector)
-		if len(term.MatchExpressions) != 0 {
-			placement.NodeAffinity = &corev1.NodeAffinity{
-				RequiredDuringSchedulingIgnoredDuringExecution: &corev1.NodeSelector{
-					NodeSelectorTerms: []corev1.NodeSelectorTerm{term},
-				},
-			}
+		// If the StorageCluster specifies a label selector, append it to the
+		// node affinity, creating it if it doesn't exist.
+		reqs := convertLabelToNodeSelectorRequirements(*sc.Spec.LabelSelector)
+		if len(reqs) != 0 {
+			appendNodeRequirements(&placement, reqs...)
 		}
 	}
 
@@ -51,15 +49,15 @@ func getPlacement(sc *ocsv1.StorageCluster, component string) rookv1.Placement {
 	return placement
 }
 
-//convertLabelToNodeSelector returns NodeSelectorTerm type from a given LabelSelector
-func convertLabelToNodeSelector(labelSelector metav1.LabelSelector) corev1.NodeSelectorTerm {
-	term := corev1.NodeSelectorTerm{}
+//convertLabelToNodeSelectorRequirements returns a NodeSelectorRequirement list from a given LabelSelector
+func convertLabelToNodeSelectorRequirements(labelSelector metav1.LabelSelector) []corev1.NodeSelectorRequirement {
+	reqs := []corev1.NodeSelectorRequirement{}
 	for key, value := range labelSelector.MatchLabels {
 		req := corev1.NodeSelectorRequirement{}
 		req.Key = key
 		req.Operator = corev1.NodeSelectorOpIn
 		req.Values = append(req.Values, value)
-		term.MatchExpressions = append(term.MatchExpressions, req)
+		reqs = append(reqs, req)
 	}
 	numIter := len(labelSelector.MatchExpressions)
 	for i := 0; i < numIter; i++ {
@@ -67,9 +65,23 @@ func convertLabelToNodeSelector(labelSelector metav1.LabelSelector) corev1.NodeS
 		req.Key = labelSelector.MatchExpressions[i].Key
 		req.Operator = corev1.NodeSelectorOperator(labelSelector.MatchExpressions[i].Operator)
 		req.Values = labelSelector.MatchExpressions[i].Values
-		term.MatchExpressions = append(term.MatchExpressions, req)
+		reqs = append(reqs, req)
 	}
-	return term
+	return reqs
+}
+
+func appendNodeRequirements(placement *rookv1.Placement, reqs ...corev1.NodeSelectorRequirement) {
+	if placement.NodeAffinity == nil {
+		placement.NodeAffinity = &corev1.NodeAffinity{}
+	}
+	if placement.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution == nil {
+		placement.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution = &corev1.NodeSelector{}
+	}
+	nodeSelector := placement.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution
+	if len(nodeSelector.NodeSelectorTerms) == 0 {
+		nodeSelector.NodeSelectorTerms = append(nodeSelector.NodeSelectorTerms, corev1.NodeSelectorTerm{})
+	}
+	nodeSelector.NodeSelectorTerms[0].MatchExpressions = append(nodeSelector.NodeSelectorTerms[0].MatchExpressions, reqs...)
 }
 
 // MatchingLabelsSelector filters the list/delete operation on the given label
--- a/pkg/controller/storagecluster/placement.go
+++ b/pkg/controller/storagecluster/placement.go
@@ -20,11 +20,17 @@ func getPlacement(sc *ocsv1.StorageCluster, component string) rookv1.Placement {
 		in := defaults.DaemonPlacements[component]
 		(&in).DeepCopyInto(&placement)
 	}
-	if sc.Spec.LabelSelector == nil {
+
+	// If no placement is specified for the given component and the
+	// StorageCluster has no label selector, set the default node
+	// affinity.
+	if placement.NodeAffinity == nil && sc.Spec.LabelSelector == nil {
 		placement.NodeAffinity = defaults.DefaultNodeAffinity
-	} else {
-		// If the StorageCluster specifies a label selector, append it to the
-		// node affinity, creating it if it doesn't exist.
+	}
+
+	// If the StorageCluster specifies a label selector, append it to the
+	// node affinity, creating it if it doesn't exist.
+	if sc.Spec.LabelSelector != nil {
 		reqs := convertLabelToNodeSelectorRequirements(*sc.Spec.LabelSelector)
 		if len(reqs) != 0 {
 			appendNodeRequirements(&placement, reqs...)
--- a/pkg/controller/storagecluster/placement_test.go
+++ b/pkg/controller/storagecluster/placement_test.go
@@ -79,7 +79,7 @@ func TestGetPlacement(t *testing.T) {
 	sc = &ocsv1.StorageCluster{}
 	mockStorageCluster.DeepCopyInto(sc)
 	sc.Spec.Placement = mockPlacements
-	assert.Equal(t, defaultLabelPlacement["all"], getPlacement(sc, "all"))
+	assert.Equal(t, mockPlacements["all"], getPlacement(sc, "all"))
 
 	// Case 3: LabelSelector to modify the default placements correctly
 	sc = &ocsv1.StorageCluster{}
--- a/pkg/controller/storagecluster/placement_test.go
+++ b/pkg/controller/storagecluster/placement_test.go
@@ -16,160 +16,135 @@ const (
 	MasterAffinityKey = "node-role.kubernetes.io/master"
 )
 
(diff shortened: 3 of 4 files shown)


pkg/controller/storagecluster/placement.go
```diff
@@ -51,25 +55,39 @@ func getPlacement(sc *ocsv1.StorageCluster, component string) rookv1.Placement {
...
 		req.Key = labelSelector.MatchExpressions[i].Key
 		req.Operator = corev1.NodeSelectorOperator(labelSelector.MatchExpressions[i].Operator)
 		req.Values = labelSelector.MatchExpressions[i].Values
-		term.MatchExpressions = append(term.MatchExpressions, req)
+		reqs = append(reqs, req)
+	}
+	return reqs
+}
+
+func appendNodeRequirements(placement *rookv1.Placement, reqs ...corev1.NodeSelectorRequirement) {
```
umangachapagain, line 79: This seems like an enhancement which isn't related to LabelSelectors. Should be in it's own commit.

pkg/controller/storagecluster/cephcluster.go
```diff
@@ -335,8 +335,7 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster) []rook.StorageClassDevi
 						Operator: corev1.NodeSelectorOpIn,
 						Values:   []string{topologyKeyValues[topologyIndex]},
 					}
-					nodeSelectorTerms := placement.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms
-					nodeSelectorTerms[0].MatchExpressions = append(nodeSelectorTerms[0].MatchExpressions, nodeZoneSelector)
+					appendNodeRequirements(&placement, nodeZoneSelector)
```
umangachapagain, line 338: Not related to LabelSelector. Should be in different commit (probably with enhancement commit from below)

obnoxxx:
@umangachapagain writes:
>  I think it's time we add a functional test for this i.e. Create StorageCluster with empty/custom LabelSelectors and see if it succeeds. Because, we had unit tests before that passed and yet the feature wasn't working as expected.

We can add a func test, but not sure it's needed for this PR: If there were unit tests and they didn't catch it, then they were not correctly catching the case. We are effectively aiming at fixing a crash bug here.

I am still reviewing the patch but I somehow feel that we could get a more minimal patch for fixing the crash for backporting, and then can do a bigger rework on top of it for master. Regarding the test, ideally, we'd have a separate patch to the test case applied after a separate code change commit that will fail (trigger the segfault) if we revert the code change.

pkg/controller/storagecluster/placement.go
```diff
@@ -51,25 +55,39 @@ func getPlacement(sc *ocsv1.StorageCluster, component string) rookv1.Placement {
...
+	if placement.NodeAffinity == nil {
+		placement.NodeAffinity = &corev1.NodeAffinity{}
+	}
+	if placement.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution == nil {
+		placement.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution = &corev1.NodeSelector{}
+	}
```
umangachapagain, line 85: if `placement.NodeAffinity == nil` then why do we need to check for `placement.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution == nil`? It'll obviously be `nil`.

pkg/controller/storagecluster/placement_test.go
```diff
@@ -13,108 +13,105 @@ import (
...
-				Key:      WorkerAffinityKey,
-				Operator: metav1.LabelSelectorOpExists,
+	cases := []struct {
+		placements        rookv1.PlacementSpec
+		labelSelector     *metav1.LabelSelector
+		expectedPlacement rookv1.Placement
+	}{
+		{
+			// Case 1: Defaults are preserved i.e no placement and no label selector
+			placements:    rookv1.PlacementSpec{},
```
umangachapagain, line 63: doesn't this mean "empty" placement and not "nil" (or "no" placement) ??

pkg/controller/storagecluster/placement_test.go
```diff
@@ -13,108 +13,105 @@ import (
...
-	sc = &ocsv1.StorageCluster{}
-	mockStorageCluster.DeepCopyInto(sc)
-	sc.Spec.Placement = mockPlacements
-	sc.Spec.LabelSelector = &metav1.LabelSelector{
-		MatchExpressions: []metav1.LabelSelectorRequirement{
-			metav1.LabelSelectorRequirement{
-				Key:      MasterAffinityKey,
-				Operator: metav1.LabelSelectorOpExists,
+		{
+			// Case 2: The configured Placements override the defaults
```
umangachapagain, line 71: I still see default under expectedPlacement?? and, the configured placement is actually empty.

pkg/controller/storagecluster/placement_test.go
```diff
@@ -13,108 +13,105 @@ import (
...
+			expectedPlacement: rookv1.Placement{
+				NodeAffinity: &corev1.NodeAffinity{
+					RequiredDuringSchedulingIgnoredDuringExecution: &workerNodeSelector,
+				},
+				Tolerations: defaults.DaemonPlacements["all"].Tolerations,
+			},
+		},
+		{
+			// Case 4: The LabelSelector modifies the configured Placements correctly
+			placements:    emptyPlacements,
```
umangachapagain, line 91: Shouldn't we add some real placement instead of empty to test this?
If placement and labelSelector is provided, what is the behaviour?

pkg/controller/storagecluster/placement.go
obnoxxx: I think this pattern is perfectly fine to avoid special cases and deep nesting. As long as this check covers the special case of (originally) `placement.NodeAffinity == nil` in which we have set `placement.NodeAffinity` specially.

pkg/controller/storagecluster/placement_test.go
obnoxxx: I think it means placement has not been provided?

pkg/controller/storagecluster/placement_test.go
obnoxxx: I think placement has been provided but it just maps `"all"` to an empty placement, which is overridden by the default placement?

pkg/controller/storagecluster/placement_test.go
obnoxxx: Yes, this should also be covered.

pkg/controller/storagecluster/placement_test.go
```diff
@@ -13,108 +13,105 @@ import (
...
-	mockStorageCluster.DeepCopyInto(sc)
-	sc.Spec.LabelSelector = &metav1.LabelSelector{
-		MatchExpressions: []metav1.LabelSelectorRequirement{},
+	for _, c := range cases {
+		sc := &ocsv1.StorageCluster{}
+		mockStorageCluster.DeepCopyInto(sc)
+		sc.Spec.Placement = c.placements
+		sc.Spec.LabelSelector = c.labelSelector
+		expectedPlacement := c.expectedPlacement
+		assert.Equal(t, expectedPlacement, getPlacement(sc, "all"))
```
obnoxxx, line 115: Better split the invocation of `getPlacement` into a separate line and assign the result to a variable. It will then be easier to see what happens if getPlacement crashes, etc.

pkg/controller/storagecluster/cephcluster.go
obnoxxx: Well, it is related: This is about bad behavior (crashing) on nil `placement.NodeAffinity` when we have set an empty LabelSelector.  This very line was segfaulting.

This new function `appendNodeRequirements` is also invoked in the changed function `getPlacement`.

I agree though, that we might want to first fix the crash with less refactoring / enhancements, if possible, and then rework the code.

pkg/controller/storagecluster/placement.go
```diff
@@ -20,16 +20,20 @@ func getPlacement(sc *ocsv1.StorageCluster, component string) rookv1.Placement {
 		in := defaults.DaemonPlacements[component]
 		(&in).DeepCopyInto(&placement)
 	}
-	if sc.Spec.LabelSelector == nil {
+
+	// If no placement is specified for the given component and the
+	// StorageCluster has no label selector, set the default node
+	// affinity.
+	if placement.NodeAffinity == nil && sc.Spec.LabelSelector == nil {
```
obnoxxx, line 27: This hunk for instance, is pretty much independent of the other changes, afaict. It should be in its own commit, imho.

pkg/controller/storagecluster/placement.go
```diff
@@ -20,16 +20,20 @@ func getPlacement(sc *ocsv1.StorageCluster, component string) rookv1.Placement {
...
-				},
-			}
+	}
+
+	// If the StorageCluster specifies a label selector, append it to the
+	// node affinity, creating it if it doesn't exist.
+	if sc.Spec.LabelSelector != nil {
+		reqs := convertLabelToNodeSelectorRequirements(*sc.Spec.LabelSelector)
+		if len(reqs) != 0 {
+			appendNodeRequirements(&placement, reqs...)
```
obnoxxx, line 36: I am still somewhat confused that new `appendNodeRequirements()` is invoked here in `getPlacement()` but also in `newStorageClassDeviceSets()`, after the invocation of `getPlacement()`. Is this redundant? :thinking: ...
At least the code is still too convoluted to easily understand all the cases.

obnoxxx:
To illustrate my points, here is a RFC branch (https://github.com/obnoxxx/ocs-operator/tree/fix-empty-label-selector-crash) with a test case that catches the segfault (https://github.com/obnoxxx/ocs-operator/commit/f9845636163a8fbe3abbdafed7b6e1e1e1c2edc6) and a minimal fix that just removes the segfault (https://github.com/obnoxxx/ocs-operator/commit/2fe6bbe237c35f3f43241865957901fbe61851a9).

Not saying that this conceptually the right fix, and we need a lot of what @jarrpa has done, at least in master.

For the fix, the conceptual question I have is whether the right thing would be to catch the `nil` `NodeAffinity` in `newStorageClassDeviceSets` and just not dereference it there *or* to make sure in `getPlacements` to never have a `nil` `NodeAffinity`. @jarrpa 's patch here does both, which seems somewhat redundant, but maybe I'm missing a point here.

obnoxxx:
Update: as an alternative, reverted the change to `newStorageClassDeviceSets` and adapted `getPlacement()` instead: https://github.com/obnoxxx/ocs-operator/commit/f93ff81c831d2456860c57bc88617b4e2057a65b

This requires an adaptation of the existing test for `getPlacement` for the change of behavior for empty node selector: https://github.com/obnoxxx/ocs-operator/commit/a69a02d55b68be71da8cd1e03f6daf201cd82921 (these two commits should be squashed if we decided to go down that path, but kept them separate here just to illustrate)

obnoxxx:
raised #619 to see how the func tests go.
I polished the branch a bit.
The broken up state is still visible in https://github.com/obnoxxx/ocs-operator/commits/fix-empty-label-selector-crash-wip

pkg/controller/storagecluster/placement_test.go
jarrpa: Correct. Since the Placement field isn't a pointer, you can't set it to nil, you have to set it to an empty struct.

pkg/controller/storagecluster/placement.go
jarrpa: A StorageClassDeviceSet can take a TopologyKey value that must alter the generated Placement for all OSDs to be specific to the device set. As such, it has an additional thing to append after the initial Placement generation. To condense this further I'd have to modify `getPlacement()` to take an optional TopologyKey parameter, which was more refactoring than I thought was necessary.

pkg/controller/storagecluster/placement.go
jarrpa: This very function fixes the actual segfault being reported. Because the same code was in two places, the other segfult would have been hit once the original one had been cleared. Making it a common function fixes both spots.

pkg/controller/storagecluster/placement.go
jarrpa: First you check for nil in the NodeAffinity, since you cannot guarantee it will be set. Similarly, Req... is a pointer, and that field is ALSO not guaranteed to be set (it would be a NodeAffinity with Preferred... set instead).

pkg/controller/storagecluster/placement_test.go
jarrpa: Since the value being returned is from a list of non-pointers, it will never be nil. An empty struct is the only thing that can be returned if no placement was explicitly set.

pkg/controller/storagecluster/placement_test.go
jarrpa: Correct, the LabelSelector always modifies the NodeAffinity of all Placements. Only when the LabelSelector is defined as empty (Case 5) will the NodeAffinity not be set. This gives preference for the behavior of ALWAYS tying our Pods to storage nodes, with the requirement to take a manual step to disable it.

pkg/controller/storagecluster/placement_test.go
jarrpa: I can see the need for that, will add another test case.

pkg/controller/storagecluster/placement_test.go
jarrpa: Sure.

pkg/controller/storagecluster/placement_test.go
```diff
@@ -94,7 +94,49 @@ func TestGetPlacement(t *testing.T) {
...
+	sc.Spec.Placement = map[rookv1.KeyType]rookv1.Placement{
+		"all": rookv1.Placement{
+			Tolerations: []corev1.Toleration{
+				corev1.Toleration{
+					Key:      defaults.NodeTolerationKey,
+					Operator: corev1.TolerationOpEqual,
+					Value:    "true",
+					Effect:   corev1.TaintEffectNoSchedule,
+				},
+			},
```
obnoxxx:
```suggestion
			Tolerations: defaults.DaemonPlacements["all"].Tolerations,
```

Isn't this the same?

(this is in the commit `Update TestGetPlacement() for proper LabelSelector behavior `)

obnoxxx:
Looks pretty much good now. Thanks for splitting it up! Just a few minor things:

* the Update test to ... patches should be squashed with the preceding code-changes so that these pass unit tests
*  The mentioned simplification to the `Update TestGetPlacement() for proper LabelSelector behavior ` patch
* I still need to figure out why the test case you added from my branch does not segfault any more (when executed without all the other patches... :see_no_evil: )

obnoxxx:
@JohnStrunk FYI

PR #619, opened 2020-07-08
[WIP] Fix empty label selector crash

obnoxxx:
This is a more minimal attempt at #618 to fix the segfault in `newStorageClassDeviceSets` when an empty labelselector is provided.

This is more minimal for easier backporting.
It also provides a test case as a separate patch that demonstrates the segfault without this patch.

The refactoring and enhancement of #618 could follow on top.


obnoxxx:
@jarrpa @umangachapagain  raise this as a PR to see the results of functional tests.

obnoxxx:
not needed any more

PR #624, opened 2020-07-09
Add must-gather image to csv-merger relatedImages

crombus:
https://bugzilla.redhat.com/show_bug.cgi?id=1847980


tools/csv-merger/csv-merger.go
```diff
@@ -855,6 +862,8 @@ func main() {
 		log.Fatal("--noobaa-db-image is required")
 	} else if *ocsContainerImage == "" {
 		log.Fatal("--ocs-image is required")
+	} else if *mustGatherImage == "" {
+		log.Fatal("--must-gather-image is required")
```
umangachapagain, line 866: Why? We don't need must-gather image to generate CSV whatsoever.
This should be kept optional and only be used by whoever needs it.

tools/csv-merger/csv-merger.go
```diff
@@ -749,6 +750,12 @@ func injectCSVRelatedImages(r *unstructured.Unstructured) error {
 			"image": *noobaaDBContainerImage,
 		})
 	}
+	if *mustGatherImage != "" {
+		relatedImages = append(relatedImages, map[string]interface{}{
+			"name":  "must-gather",
```
umangachapagain, line 755: let's call this `ocs-must-gather` just to be clear and not confuse with `ocp-must-gather`

tools/csv-merger/csv-merger.go
```diff
@@ -749,6 +750,12 @@ func injectCSVRelatedImages(r *unstructured.Unstructured) error {
 			"image": *noobaaDBContainerImage,
 		})
 	}
+	if *ocsMustGatherImage != "" {
+		relatedImages = append(relatedImages, map[string]interface{}{
+			"name":  "ocs-must-gather-image",
```
obnoxxx, line 755: This is a bit inconsistent with the others images. When following the pattern, we would do:
```suggestion
			"name":  "ocs-must-gather",
```

hack/generate-latest-csv.sh
```diff
@@ -49,4 +49,5 @@ $CSV_CHECKSUM \
 	--noobaa-core-image="$NOOBAA_CORE_IMAGE" \
 	--noobaa-db-image="$NOOBAA_DB_IMAGE" \
 	--ocs-image="$OCS_IMAGE" \
+	--ocs-must-gather-image="$MUST_GATHER_FULL_IMAGE_NAME" \
```
obnoxxx, line 52: I would follow the pattern of adding a variable `MUST_GATHER_IMAGE` to this script and set it by default to "$MUST_GATHER_FULL_IMAGE_NAME" (in the init section above). This would allow for overriding. Not strictly needed, but nice and consistent.

hack/generate-unified-csv.sh
```diff
@@ -60,6 +60,7 @@ $CSV_MERGER \
 	--noobaa-core-image="$NOOBAA_CORE_IMAGE" \
 	--noobaa-db-image="$NOOBAA_DB_IMAGE" \
 	--ocs-image="$OCS_IMAGE" \
+	--ocs-must-gather-image="$MUST_GATHER_FULL_IMAGE_NAME" \
```
obnoxxx, line 63: Same, suggest to add  a variable.

obnoxxx:
error of verify-latest-csv is strange. `uncommitted CSV changes. run 'make gen-latest-csv' and commit results.` For me locally, it's working fine.... :thinking:

obnoxxx:
Failure may be due to this:
```
Openshift CI detected, deploying using image registry.svc.ci.openshift.org/ci-op-pmkky199/stable:ocs-registry
```

hack/generate-latest-csv.sh
```diff
@@ -19,6 +19,7 @@ export NOOBAA_CORE_IMAGE=${NOOBAA_CORE_IMAGE:-"noobaa/noobaa-core:5.5.0-20200422
 export NOOBAA_DB_IMAGE=${NOOBAA_DB_IMAGE:-"centos/mongodb-36-centos7"}
 export CEPH_IMAGE=${CEPH_IMAGE:-"ceph/ceph:v14.2"}
 export OCS_IMAGE=${OCS_IMAGE:-"${IMAGE_REGISTRY}/${REGISTRY_NAMESPACE}/${OPERATOR_IMAGE_NAME}:${IMAGE_TAG}"}
+export OCS_MUST_GATHER_IMAGE=${OCS_MUST_GATHER_IMAGE:-"${MUST_GATHER_FULL_IMAGE_NAME}"}
 
 echo "=== Generating DEV CSV with the following vars ==="
 echo -e "\tCSV_VERSION=$CSV_VERSION"
```
obnoxxx, line 25: Do we want to add the output of the used `OCS_MUST_GATHER_IMAGE` to this output?

hack/generate-unified-csv.sh
```diff
@@ -60,6 +60,7 @@ $CSV_MERGER \
 	--noobaa-core-image="$NOOBAA_CORE_IMAGE" \
 	--noobaa-db-image="$NOOBAA_DB_IMAGE" \
 	--ocs-image="$OCS_IMAGE" \
+	--ocs-must-gather-image="$OCS_MUST_GATHER_IMAGE" \
```
obnoxxx, line 63: Not sure if `OCS_MUST_GATHER` should be added to the check for required env vars (and a few more, fwiw...)

hack/generate-latest-csv.sh
crombus: this is not what you meant.?

crombus: yes, this is the suggested way because all the above export commands uses the same format.

hack/generate-unified-csv.sh
crombus: must-gather is a optional flag. so i guess no need.

obnoxxx: I don't fully understand this. How is must-gather optional? The script unconditionally specifies `--ocs-must-gather=....` and if the VAR is not set, I gues this command would fail? That is required, imho, and it applies to all parameters, but I must be missing something.

obnoxxx: But I agree that this is a seprate topic. not part of this PR. :-)

PR #625, opened 2020-07-09
Add csv resource

crombus:
https://bugzilla.redhat.com/show_bug.cgi?id=1853025

### DIFF
--- a/must-gather/collection-scripts/gather
+++ b/must-gather/collection-scripts/gather
@@ -42,6 +42,7 @@ commands+=("get pv")
 commands+=("get sc")
 commands+=("get infrastructures.config -oyaml")
 commands+=("get subscription -n openshift-storage")
+commands+=("get csv -n openshift-storage")
 
 # Run the Collection of Resources to list
 for command in "${commands[@]}"; do


must-gather/collection-scripts/gather
```diff
@@ -42,6 +42,7 @@ commands+=("get pv")
 commands+=("get sc")
 commands+=("get infrastructures.config -oyaml")
 commands+=("get subscription -n openshift-storage")
+commands+=("get csv")
```
petr-balogh, line 45: I think that more debug info will be to get yaml output instead of only list of CSVs in text format. (Or we can do both but don't think it's necessary).
```suggestion
commands+=("get csv -o yaml")
```

obnoxxx: good idea - @crombus what do you think?

crombus: the change you are asking already exists in must-gather. https://github.com/openshift/ocs-operator/blob/f6dcb364571ed3d7497a25a3698df679160e18ad/must-gather/collection-scripts/gather_ceph_resources#L115

petr-balogh: Cool, I didn't know it's located there. Just checked the one of the must-gather logs and it's really there: http://magna002.ceph.redhat.com/ocsci-jenkins/openshift-clusters/jnk-pr2470-b2266/jnk-pr2470-b2266_20200709T112044/logs/failed_testcase_ocs_logs_1594296567/test_write_file_to_bucket_ocs_logs/ocs_must_gather/quay-io-rhceph-dev-ocs-must-gather-sha256-a9b65f901470bcd9467474d640e230c0b1a96c9a512284be53deb6962107873e/ceph/namespaces/openshift-storage/operators.coreos.com/clusterserviceversions/ 

So then it's fine to have only `get csv`, but haven't you missed the namespace `-n openshift-storage`?

Thanks

crombus: yes i'll add that. Thanks!

must-gather/collection-scripts/gather
```diff
@@ -42,6 +42,7 @@ commands+=("get pv")
 commands+=("get sc")
 commands+=("get infrastructures.config -oyaml")
 commands+=("get subscription -n openshift-storage")
+commands+=("get csv -n openshift-storage")
```
obnoxxx, line 45: What about Petr's suggestion to use `-o yaml` ?
```suggestion
commands+=("get csv -n openshift-storage -o yaml")
```

crombus: https://github.com/openshift/ocs-operator/blob/f6dcb364571ed3d7497a25a3698df679160e18ad/must-gather/collection-scripts/gather_ceph_resources#L115

crombus: it already exists.

obnoxxx: hmm, but then why are we doing this at all? (I don't understand the difference/relationship between these various scripts...

crombus: this is for only oc commands.  the changes which are in yaml uses inspect cmd.

obnoxxx: I still don't understand: the inspect command is also an oc command... Why do we need to get info about the csv twice once with `oc adm inspect` and once with `oc get`? (I don't know much about how must-gather works, so this may be dumb and feel free to ignore me...) And in general, why not use `-o yaml` for the "normal" oc commands?

crombus: ocp must-gather uses inspect underlying.  So I try to follow that and use inspect only. This oc command was request from support so they can you some data which is in readeable formatt like describe.  This continued and now we have so many commands. But the target is still to capture data using inspect. I'll add a refactor PR which align these changes. there is bug which is targeted for 4.6.

petr-balogh: @obnoxxx when I am looking at the cluster and want to know it's installed an ok I am checking`oc get csv -n openshift-storage ` and it's output a lot. It will definitely be worth to have this human readable format in must gather data instead of go in big csv yaml file.

obnoxxx: @petr-balogh oh ok.  I thought you requested `-o yaml` - so you withdraw that request?

petr-balogh: Cause I thought we don't have that as well ;) But @crombus told me where it's collected.

PR #629, opened 2020-07-10
Shellcheck cleanup - make shellcheck actually run and fail

obnoxxx:
I was recently wondering why I am spotting obvious shellcheck errors in scripts but didn't see CI / test failures. Now looking into the shellcheck-test script, I realized that the return code from shellcheck was not properly interpreted, and all errors went without notice.

This PR fixes this and allows the shellcheck-test to fail. In order to be able to push this, the PR first fixes all shellcheck errors in scripts that have accumulated (12 scripts are fixed). Additionally, the hack/common.sh script is not skipped any more, but just disables the unused variable test.

Finally, a few additional aspects of the shellcheck-test script are improved: the dependency on GOPATH is removed (which is not required to be set for our module based code), so that one can run this without GOPATH set. And an informative output about the number of scripts with errors is added.

### DIFF
--- a/hack/shellcheck-test.sh
+++ b/hack/shellcheck-test.sh
@@ -2,6 +2,8 @@
   
 # Check for shell syntax & style.
 
+source hack/common.sh
+
 test_syntax() {
         bash -n "${1}"
 }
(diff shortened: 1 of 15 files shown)


hack/shellcheck-test.sh
```diff
@@ -27,35 +29,38 @@ BASE_DIR="$(cd "${SCRIPT_DIR}/.." && pwd)"
...
-                ((failed=err))
+                ((failed+=err))
         else
                 echo "${script}: ok" >&2
         fi
 done
+
+echo "${failed} scripts with errors were found"
+
 exit ${failed}
```
crombus, line 65: do we want to exit the script with number of files failed with errors instead of 1

obnoxxx: yes. !=0 is error

crombus: cool

hack/source-manifests.sh
```diff
@@ -71,7 +71,7 @@ gen_args="generate csv --csv-version=$TMP_CSV_VERSION --output-dir=$OUTDIR_TEMPL
 if [ -n "$CSV_REPLACES_VERSION" ]; then
 	gen_args="$gen_args --from-version=$CSV_REPLACES_VERSION"
 fi
-$OPERATOR_SDK $gen_args
+$OPERATOR_SDK "$gen_args"
```
nb-ohad, line 74: This change breaks `make gen-latest-csv` when running on darwin (macos) 

The error is:
`Error: unknown command "generate csv --csv-version=9999.9999.9999 --output-dir=build/_output/csv-templates" for "operator-sdk"`

Removing the quotes (reverting the change) fix the issue

PR #630, opened 2020-07-11
[WIP] Shellcheck cleanup - try making mandatory

obnoxxx:
This is like PR #629 but on top, I want to see if shellcheck is actually (being made) available in the openshift-ci / prow environment. Not meant for merging...

### DIFF
--- a/hack/build-must-gather.sh
+++ b/hack/build-must-gather.sh
@@ -4,4 +4,4 @@ set -e
 
 source hack/common.sh
 
-${IMAGE_BUILD_CMD} build -f must-gather/Dockerfile -t ${MUST_GATHER_FULL_IMAGE_NAME} must-gather/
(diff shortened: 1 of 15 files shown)


obnoxxx:
@umangachapagain @jarrpa - ok, so indeed there's no shellcheck in prow and we can't seem to install it... Is there a possibility to add it to the images?

jarrpa:
@obnoxxx I believe not. The scripts exists purely in the github repo and not in the container images, which only gets pulled in the build image AFAIK. We'd have to explicitly create a new job with a new image that has the full repo and shellcheck installed.

umangachapagain:
> @umangachapagain @jarrpa - ok, so indeed there's no shellcheck in prow and we can't seem to install it... Is there a possibility to add it to the images?

We should be able to. Just like we do lint check. I'll give it a shot.

PR #633, opened 2020-07-13
Inhibit PDB alerts for OSDs

umangachapagain:
PDB alerts for OSDs are inhibited via alertmanager config
to prevent misguiding information about PDB being provided
to the adming via alerts. The provided PDB values are expected
by design and so the admin should not be alerted about it.

test - added unit test to getInhibitionRules()
E2E - added functional tests to verify PDB alert inhibition 

### DIFF
--- a/pkg/controller/storagecluster/reconcile.go
+++ b/pkg/controller/storagecluster/reconcile.go
@@ -1,6 +1,7 @@
 package storagecluster
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -103,6 +104,17 @@ func (r *ReconcileStorageCluster) Reconcile(request reconcile.Request) (reconcil
 
 	if err := versionCheck(instance, reqLogger); err != nil {
 		return reconcile.Result{}, err
+	}
+
+	// always inhibit PDB alerts for OSDs on Internal Clusters
+	// if failed to inhibit PDB alert, log error and continue
+	// this is to unblock Storage Cluster operations in case
+	// of alertmanager failures
+	if !instance.Spec.ExternalStorage.Enable {
+		err = inhibitPDBAlert(r.client, r.reqLogger)
+		if err != nil {
+			reqLogger.Error(err, "unable to inhibit PDB alerts")
+		}
 	}
 
 	// Check for active StorageCluster only if Create request is made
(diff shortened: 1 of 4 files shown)


pkg/controller/storagecluster/reconcile.go
```diff
@@ -958,3 +966,61 @@ fi`
...
+	inhibitRules := `
+	- "equal":
+	  - "namespace"
+	  - "alertname"
+	  "source_match_re":
+	   "poddisruptionbudget" : "^rook-ceph-osd-[0-9]+$"
+	  "target_match_re":
+	    "poddisruptionbudget" : "^rook-ceph-osd-[0-9]+$"`
+	oldConfig := alertmanagerConfigSecret.Data["alertmanager.yaml"]
+	oldconfigBytes := bytes.SplitAfter(oldConfig, []byte("\"inhibit_rules\":"))
```
jarrpa, line 1008: Are we always guaranteed that the "inhibit_rules" field will be the last in the YAML manifest?

pkg/controller/storagecluster/storagecluster_controller_test.go
```diff
@@ -759,3 +759,149 @@ func TestNodeTopologyMapLabelSelector(t *testing.T) {
 	}
 	assert.Equal(t, nodeTopologyMap, sc.Status.NodeTopologies)
 }
+
+func Test_getInhibitionRules(t *testing.T) {
```
jarrpa, line 763: No underscore.

pkg/controller/storagecluster/storagecluster_controller_test.go
```diff
@@ -759,3 +759,149 @@ func TestNodeTopologyMapLabelSelector(t *testing.T) {
...
+			name: "Inhibition rule, 1 Storagecluster",
+			args: args{
+				actualSecret:         secretWithInhibition,
+				expectedSecret:       secretWithInhibition,
+				storageClusterLength: 1,
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
```
jarrpa, line 901: Why are you doing a `t.Run()` here? I don't see that we do it for any of the other tests. I might just be ignorant, though.

umangachapagain: This is auto-generated. Can be used to run tests in parallel.
Since, we're not using it anywhere else, dropping it for now.

pkg/controller/storagecluster/reconcile.go
umangachapagain: No, in fact it will never be the last.
We split the file wherever we find `inhibit_rules:` and append rules to it.

umangachapagain:
> Move the PDB-related functions into their own file, and similarly the PDB-related tests into their own related file.

The function is used only by reconcile.go and will not be referenced outside of it's scope. I don't think we need to complicate it.

> Any reason we're not testing `inhinitPDBAlert()`?

We are testing it in E2E as it needs to make real calls to server.

pkg/controller/storagecluster/reconcile.go
jarrpa: But when you do the split, you seem to be assuming that everything after `inhibit_rules:` are subsections of that block. So I ask again, is that lock always guaranteed to be last in the file?

umangachapagain: Yes. That's ensured by openshift-monitoring operator. There'll always be a block of default `inhibit_rules:`.

jarrpa: As long as it's also guaranteed to be at the end of the manifest, that's fine by me.

simonpasquier: The user is free to edit/change the secret to dispatch notifications wherever needed. The default configuration secret is just a minimal configuration to get started. I'm not sure it's reasonable to have an operator modifying it under the hood.
At the very least, you need to parse the YAML data instead of manipulating text strings.

umangachapagain: @simonpasquier When I looked into the alertmanager job in cluster monitoring operator, it looked like it creates/updates the secret. But, from what you are saying, it creates the secret and doesn't update it (user has to do it). That definitely adds a few more cases to be handled.

>  I'm not sure it's reasonable to have an operator modifying it under the hood.

We can't ask the user to do it because it needs to be done on all clusters and that's an overhead we don't want to add.
We'd be happy getting rid of the PDB alert altogether, but as discussed elsewhere it has importance in some other projects.

> At the very least, you need to parse the YAML data instead of manipulating text strings.

I tried doing that but didn't find any schema or type definition for the config file. It's defined as a block of string and so I chose string manipulation (which is close to how we'd do it manually via CLI or Edit YAML options).
Is there an example that parses this string?

pkg/controller/storagecluster/reconcile.go
```diff
@@ -958,3 +970,64 @@ fi`
...
+	err := client.List(context.TODO(), storageClusterList)
+	if err != nil {
+		return fmt.Errorf("couldn't list Storage Clusters. %v", err)
+	}
+	storageClusterLength := len(storageClusterList.Items)
+
+	alertmanagerConfigSecret := &corev1.Secret{}
+	alertmanagerConfigSecretName := "alertmanager-main"
+	alertmanagerConfigSecretNamespace := "openshift-monitoring"
+	err = client.Get(context.TODO(), types.NamespacedName{Name: alertmanagerConfigSecretName, Namespace: alertmanagerConfigSecretNamespace}, alertmanagerConfigSecret)
```
umangachapagain, line 989: @simonpasquier This line returns a "not Found" error, but the same thing works on E2E test (https://github.com/openshift/ocs-operator/pull/633/files#diff-5643ef1ea1266f32e13238304f0cd425R136). Any clue why that might happen?

If it was an RBAC issue, i'd expect it to complain saying I can't "get" secret instead of "not found" while it clearly exists.

simonpasquier: Hmm, not sure why you'd get a "not found" error...

umangachapagain:
As confirmed with OpenShift Monitoring team in our discussions, this is not the right thing to do.
Operator should not modify the user's Config Map (Users can add/remove inhibition rules as they want).

Closing this PR and searching for a better fix..

PR #634, opened 2020-07-13
Add MCG spec section to the storage cluster CRD

nb-ohad:
And reconcile it as part for the noobaa system reconciler 
Currently focusing on exposing noobaa endpoint specs via the `storageclsuter` CRD

Signed-off-by: Ohad Mitrani omitrani@redhat.com


Makefile
```diff
@@ -40,7 +40,7 @@ $(call add-crd-gen,ocsv1,./pkg/apis/ocs/v1,./deploy/crds,./deploy/crds)
 	shellcheck-test \
 	gofmt \
 	golint \
-	govet \
+	gitgovet \
```
jarrpa, line 43: Isn't the target still named `govet`? At least it is in this version of the file.

nb-ohad: Yes, you are right, I guess I just wrote `git` while the focus was on the editor.
I will revert and push

nb-ohad: done

PR #635, opened 2020-07-13
Add TargetSizeRatio of .49 to RGW data pool

obnoxxx:
The purpose of the `target_size_ratio` is to configure the pool for some pre-optimization of pg placement when the autoscaler is active. If the `target-size-ratio` is set, then up to that threshold of capacity, the pre-optimization will be active. Above the threshold, it is changing the normal non-optimized mode.

This change sets the `target_size_ratio` for the RGW pool to `.49` like for the rbd pools and the cephfs data pool as recommended by ceph experts (starting with RHCS 4.1).

### DIFF
--- a/pkg/controller/storagecluster/cephobjectstores.go
+++ b/pkg/controller/storagecluster/cephobjectstores.go
@@ -92,6 +92,7 @@ func (r *ReconcileStorageCluster) newCephObjectStoreInstances(initData *ocsv1.St
 					FailureDomain: initData.Status.FailureDomain,
 					Replicated: cephv1.ReplicatedSpec{
 						Size: 3,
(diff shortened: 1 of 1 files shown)


obnoxxx:
this was originally done for https://bugzilla.redhat.com/show_bug.cgi?id=1850538

but it turned out we did not need it to fix that bz. see the history and explanations:

https://bugzilla.redhat.com/show_bug.cgi?id=1850538#c24
https://bugzilla.redhat.com/show_bug.cgi?id=1850538#c25
https://bugzilla.redhat.com/show_bug.cgi?id=1850538#c30

So this is a pre-optimization setting for the rgw data pool.

jarrpa (approved):

PR #636, opened 2020-07-13
Fix OCS tools and enable "make" tasks on MacOs environments

nb-ohad:
- Consolidate `OS_TYPE` definition.
- Select `shellcheck` installation source based on os
- Remove dependency on global `shellcheck` and always install to prevent any problems that might occur if the dev env has a different version of `shellcheck`
- use `GOPATH` to locate `golint`
- Remove platform from operator-sdk version

Signed-off-by: Ohad Mitrani omitrani@redhat.com


hack/shellcheck-test.sh
```diff
@@ -15,28 +15,30 @@ test_shellcheck() {
...
-        fi
+SC_VERSION="stable"
+echo "Installing shellcheck... (${SC_VERSION?})" >&2
+FILE="${OUTDIR_TOOLS}/shellcheck"
+if [ "$OS_TYPE" == "Darwin" ]; then
+        wget -qO- "https://github.com/koalaman/shellcheck/releases/download/${SC_VERSION?}/shellcheck-${SC_VERSION?}.darwin.x86_64.tar.xz" | tar -xJv
+else
+        wget -qO- "https://storage.googleapis.com/shellcheck/shellcheck-${SC_VERSION?}.linux.x86_64.tar.xz" | tar -xJv
 fi
+cp -f "shellcheck-${SC_VERSION}/shellcheck" "${FILE}"
```
jarrpa, line 36: Don't install it every time. We want to skip this if the binary is already downloaded.

hack/common.sh
```diff
@@ -77,6 +77,8 @@ OCS_SUBSCRIPTION_CHANNEL=${OCS_SUBSCRIPTION_CHANNEL:-alpha}
 UPGRADE_FROM_OCS_REGISTRY_IMAGE="${UPGRADE_FROM_OCS_REGISTRY_IMAGE:-quay.io/ocs-dev/ocs-registry:4.2.0}"
 UPGRADE_FROM_OCS_SUBSCRIPTION_CHANNEL="${UPGRADE_FROM_OCS_SUBSCRIPTION_CHANNEL:-$OCS_SUBSCRIPTION_CHANNEL}"
 
+OS_TYPE=$(uname)
```
jarrpa, line 80: The move of `OS_TYPE` should be its own commit.

hack/ensure-opm.sh
```diff
@@ -20,4 +19,4 @@ if [ ! -x "${OPM}" ]; then
 	chmod +x ${OPM}
 else
 	echo "Using opm cached at ${OPM}"
-fi
\ No newline at end of file
+fi
```
jarrpa, line 22: Remove this change

hack/shellcheck-test.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-  
+
```
jarrpa, line 2: Separate commit, can be the same as the shellcheck error fix

hack/shellcheck-test.sh
```diff
@@ -63,5 +65,5 @@ done
 
 echo "${failed} scripts with errors were found"
 
-exit ${failed}
+exit "${failed}"
```
jarrpa, line 68: Separate commit, can be the same as the whitespace fix.

hack/shellcheck-test.sh
```diff
@@ -15,28 +15,30 @@ test_shellcheck() {
...
-        wget -qO- "https://storage.googleapis.com/shellcheck/shellcheck-${scversion?}.linux.x86_64.tar.xz" | tar -xJv
-        cp -f "shellcheck-${scversion}/shellcheck" "${FILE}"
-        if [ -f "$FILE" ]; then
-                SHELLCHECK="${FILE}"
-        fi
+SC_VERSION="stable"
+echo "Installing shellcheck... (${SC_VERSION?})" >&2
+FILE="${OUTDIR_TOOLS}/shellcheck"
+if [ "$OS_TYPE" == "Darwin" ]; then
+        wget -qO- "https://github.com/koalaman/shellcheck/releases/download/${SC_VERSION?}/shellcheck-${SC_VERSION?}.darwin.x86_64.tar.xz" | tar -xJv
```
jarrpa, line 32: Are you unable to use the `storage.googleapis.com` version? I would hope we could simply set a variable to be either `darwin` or `linux`. How does this work by only providing `SC_VERSION`?

hack/ensure-opm.sh
nb-ohad: Why? all other bash scripts have a new line in the end
It is a common trait of many editors to add a newline at the end of a source file

hack/shellcheck-test.sh
nb-ohad: The darwin version of `storage.googleapis.com` is deprecated and when you run it, it prompts you to use this one

hack/shellcheck-test.sh
nb-ohad: Do you want me the search for it under the tools folder?

hack/common.sh
nb-ohad: done

hack/ensure-opm.sh
nb-ohad: I removed it

hack/shellcheck-test.sh
nb-ohad: done

hack/shellcheck-test.sh
nb-ohad: done

hack/shellcheck-test.sh
jarrpa: I think it should be fine to just replace `if [[ -z "${SHELLCHECK}" ]]; then` with `if [ -f "${SHELLCHECK}" ]; then`. If we assume we will always call shellcheck from the same place, I think all we'd have to do is check if the file exists.

hack/shellcheck-test.sh
jarrpa: It seems like the github releases should be more authoritative in general. Could we use the same link for both linux and darwin?

nb-ohad: done

hack/shellcheck-test.sh
nb-ohad: Fixed it differently, will not download if already exists in the tools folder

nb-ohad:
Also - Fixed the typo

hack/shellcheck-test.sh
obnoxxx: I see you did what @jarrpa suggested: use the same base url (the github one) for both cases but did that in a later commit. Would be better to do this right away.

Also please construct the url using a conditional and then just have one occurence of the `wget | tar` command, similar to https://github.com/openshift/ocs-operator/blob/master/hack/ensure-operator-sdk.sh and https://github.com/noobaa/noobaa-operator/blob/master/build/install-operator-sdk.sh.

hack/shellcheck-test.sh
```diff
@@ -15,28 +15,30 @@ test_shellcheck() {
                         return 0
                 fi
                 #shell check -x -e SC2181,SC2029,SC1091,SC1090,SC2012 "${1}"
-                shellcheck -x -e SC2181 "${1}"
+                "${SHELLCHECK}" -x -e SC2181 "${1}"
```
obnoxxx, line 18: dang, this was not even using the variable... good catch! :grin:

hack/shellcheck-test.sh
nb-ohad: I don't think it is a good way to just build the URL based on parts because we are not guaranteed to have the same URL structure for each platform. For example, up to the last commit, the Linux case was taken from google cloud and not GitHub.
But in order to not code `wget` and `tar` twice, as requested, I now save the URL (source) in a separate var then run the commands using this var. 

I think this should be an acceptable compromise.

nb-ohad:
@obnoxxx 
Regarding the splitting to commits. All `shellcheck` changes are now in one commit (merged the review changes into the original commits) and I don't see any value in farther breakdown.

This is a 24 lines of code change that is already broken to 5 diff commits.

hack/shellcheck-test.sh
obnoxxx: Well, I just wanted to highlight that @jarrpa asked you to use the same url for linux and darwin (changing the one used for linux), and you said "done" but you actually didn't do it... :wink:

nb-ohad: It cannot be the exact same URL in any case. 
Doing the diff part via `var` or `if` does not matter.
The main thing here is that both URLs point to Github which is more  authoritative

obnoxxx:
@nb-ohad writes:
> Regarding the splitting to commits. All shellcheck changes are now in one commit (merged the review changes into the original commits) and I don't see any value in farther breakdown.
> 
> This is a 24 lines of code change that is already broken to 5 diff commits.

This is irrelevant. We can have 24 commits with 1 line change each if that's the most logical spearation or one patch with thousands of lines of change. The number of lines per patch should not be the metric but the separation of the changes introduced.  My general rule is:

A commit should
- do one thing,
- do this one thing completely,
- do nothing else.
- pass tests (to allow for using git bisect)

So if the commit message says "do this and that" it should usually be two commits... :smiley:  I know this is different from some other projects, but I think there are very good reasons for these rules.

The commit for shellcheck changes is doing 3 things:

* I makes use of the `SHELLCHECK` variable in the `test_shellcheck` function. That fixes the case of not-found system shellcheck
* It adds support for Darwin for the shellcheck test.
* it removes the use of any system shellcheck, always using downloaded.

These three should be 3 separate commits. The first will be just a 1-line change. That's ok!

Finally, the fixing of the shellcheck errors should come *before* the changes to the shellcheck script since using the newer downloaded shellcheck reveals this additional error.

hack/lint.sh
```diff
@@ -2,8 +2,9 @@
 
 RETVAL=0
 GENERATED_FILES="zz_generated.*.go"
+GOLINT="${GOPATH}/bin/golint"
```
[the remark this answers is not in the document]
obnoxxx: Just to follow up to the commit-splitting, this *could* also have been 3 PRs:
* shellcheck fixes
* golint fix
* make verify-generated fix (OPERATOR_SDK)

I wouldn't insist on this, but it would allow unrelated topics to move independently, without one having to wait for the resolution of discussion about the other... :wink:

nb-ohad:
@obnoxxx 
I am sorry but your breakdown does not make sense.
The bullets have dependencies between them. 

You cannot fix the darwin issue without removing the dependency on the system`shellcheck`, and you cannot remove the dependency on system `shellcheck` without installing the darwin `shellcheck`. 

> Finally, the fixing of the `shellcheck` errors should come before the changes to the `shellcheck` script since using the newer downloaded `shellcheck` reveals this additional error.

This could not come first (from my point of view) because it does depend on me running `shellcheck` on my env to find out the error.

I agree with most of the other points you gave, but it is really hard to have to code it in one order (the natural order that discovers the errors) then reorder everything to an order that will not work on the dev env.

hack/lint.sh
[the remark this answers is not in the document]
nb-ohad: > AFAICT, we don't expect or require GOPATH to be set since we're using go modules.
So how would you fix the issue, where would you take `golint` from

hack/shellcheck-test.sh
nb-ohad: I also think that er interrupt "the same URL"  as different things

obnoxxx:
> I am sorry but your breakdown does not make sense.

:grin: let's see if I can explain better.

> The bullets have dependencies between them.
> 
> You cannot fix the darwin issue without removing the dependency on the systemshellcheck, and you cannot remove the dependency on system shellcheck without installing the darwin shellcheck.

Sure, they are logically entirely independent. The "darwin fix" essentially is: **if** you download shellcheck, then make sure you download for the right platform. It is nothing to do with **whether** you install shellcheck...



>>    Finally, the fixing of the shellcheck errors should come before the changes to the shellcheck script since using the newer downloaded shellcheck reveals this additional error.
> 
> This could not come first (from my point of view) because it does depend on me running shellcheck on my env to find out the error.

Well, on non-darwin systems, or systems with installed shellcheck, the test would pass before. Your patches make it fail. And then you fix it again. This is the order in which this is done. But it brakes bisectability of the git tree. So the order should be reversed.

> I agree with most of the other points you gave, but it is really hard to have to code it in one order (the natural order that discovers the errors) then 

I understand that the reversing of the order might be a bit counter-intuitive at first, but it is the right thing to do if it keeps the tests passing for all commits.

And I don't understand what you mean by this:

> reorder everything to an order that will not work on the dev env.

Why would the order not work on the dev env? Even your order would only start working after you commit the shellcheck fixes. So if you commit them before the test-script commits, the first commit that enables darwin for you will already pass the tests. If you have the "natural" order, this first commit will fail tests.

nb-ohad:
@obnoxxx 
You are looking at it from the perspective of a Linux user. 

Let me try to explain it differently
I don't push commits I cannot verify locally. And I cannot verify anything because the baseline (master) is broken and `shellcheck` is not installed correctly for Darwin. 

So the first commit must be to fix `shellcheck` for Darwin in order to be able to verify anything at all.
Now this brings with it all the other fixes for `shellcheck` including using the non-global one 

If on the other hand the master branch was not broken, you could have argued for any other order and I might have accepted it.

>Well, on non-darwin systems, or systems with installed shellcheck, the test would pass before. 

This is not correct, `sheelcheck` did not pass even for @jarrpa (on master) because of a file containing a `shellcheck` error.

Any order that does not enable me to verify the commits on my machine is a "wrong" order in my book, it is like working blindly. Imagine for a second that we switched to a Darwin based ci and try to follow the order of the commits you requested and I think you will get my point.

hack/lint.sh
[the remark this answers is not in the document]
jarrpa: What about the following:
> Maybe we can use the noobaa-operator way of doing go run golang.org/x/lint/golint ?

hack/shellcheck-test.sh
```diff
@@ -15,29 +15,38 @@ test_shellcheck() {
...
-        if [ -f "$FILE" ]; then
-                SHELLCHECK="${FILE}"
+        if [ -f "$SHELLCHECK" ]; then
+                rm -rf "./shellcheck-${SC_VERSION?}"
+        else
+                unset SHELLCHECK
         fi
 fi
 
+
```
jarrpa, line 49: Remove this errant newline.

jarrpa:
I am happy with the ordering and split of commits. They make logical sense and solve the singular problems they intend to solve. Still waiting on the golint conversation, just to be sure.

/ok-to-test

hack/shellcheck-test.sh
nb-ohad: Sorry, it popped up again after the commit edits. I will fix

jarrpa:
@nb-ohad Also looks like you have uncommitted CSV changed. Run `make generate-latest-csv` and commit those in a separate commit.

obnoxxx:
@jarrpa writes:
> @nb-ohad Also looks like you have uncommitted CSV changed. Run make generate-latest-csv and commit those in a separate commit.

If you're referring to the failure of the verify-latest-csv test, that seems to always fail the CI. But pass locally (for me).

jarrpa:
@obnoxxx Running it locally, I get the following status, even in master:
```
uncommitted CSV changes. run 'make gen-latest-csv' and commit results.
make: *** [Makefile:95: verify-latest-csv] Error 1
 jrivera  …  github.com  openshift  ocs-operator  git status
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   hack/latest-csv-checksum.md5

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	deploy/crds/ocs.openshift.io_ocsinitializations.yaml
	deploy/crds/ocs.openshift.io_storageclusterinitializations.yaml
	deploy/crds/ocs.openshift.io_storageclusters.yaml
```
I thought we intentionally got rid of those, so I don't know why those files are there now... @umangachapagain, thoughts?

nb-ohad:
@jarrpa It should be very new because I am running `make gen-latest-csv` locally (synced with master yesterday) and it passes for me

obnoxxx:
@nb-ohad 
> You are looking at it from the perspective of a Linux user.

Well, yes. But that is not the point. I am mainly looking at it from the perspective of developer on who's system the tests were working.

> Let me try to explain it differently
> I don't push commits I cannot verify locally. And I cannot verify anything because the baseline (master) is broken and shellcheck is not installed correctly for Darwin.

Since the test was broken on Darwin before (couldn't even run), this is irrelevant. The only relevant part is what is the situation on a system that could run the tests before.

Also, on your system, you can only verify the shellcheck-script patch in combination with the shellcheck fixes. The order in which they are added should not matter to you. When both are applied, the test will run and pass for you. But for me, adding the shellcheck-script-fix patch first breaks tests, and then adding the shellcheck fixes after it fixes them again. That's why it is wrong and I am insisting on the different order.

One important aspect here is really that the order of the comments as they are landing in the branch eventually is not necessarily the order in which they were created by the developer. 

> So the first commit must be to fix shellcheck for Darwin in order to be able to verify anything at all.
Now this brings with it all the other fixes for shellcheck including using the non-global one

As mentioned, they should be 3 separate commits. They are logically separate. And can be separately verified on my system. (And you can fix it if you don't insist on using your laptop, but check on a different machine.)

> If on the other hand the master branch was not broken, you could have argued for any other order and I might have accepted it.

The master branch is not broken for linux with system installed shellcheck (on my fedora32 box).

> > Well, on non-darwin systems, or systems with installed shellcheck, the test would pass before.

> This is not correct, sheelcheck did not pass even for @jarrpa (on master) because of a file containing a shellcheck error.

I don't know that. It works for me. And it would for @jarrpa if he installed shellcheck from fedora.

> Any order that does not enable me to verify the commits on my machine is a "wrong" order in my book, it is like working blindly.

Understood, but again, the order in which you develop and test the patches is not necessarily the order in the resulting branch, because of the requirement of having tests pass as much as possible at least on each commit.

>  Imagine for a second that we switched to a Darwin based ci

Oh, no, I don't want to imagin that!! :grin: 

>  and try to follow the order of the commits you requested and I think you will get my point.

It would not really change the order I requested. It should not matter to you as the patch author, since this way or another, the first patch you commit would in that order not be verifiable for you on the darwin system. Only when both patches are in the branch, they can be verified. But it makes a difference on other platforms.

I guess at some point @jarrpa will just accept and merge the PR, but until then, I am really insisting on my points. :stuck_out_tongue_winking_eye:

obnoxxx:
@jarrpa:
> @obnoxxx Running it locally, I get the following status, even in master:

no issue for me locally. Try doing `git clean -dxf` prior to running `make gen-latest-csv`. And are you using GOPATH? Try not to.

nb-ohad:
> Since the test was broken on Darwin before (couldn't even run), this is irrelevant. The only relevant part is what is the situation on a system that could run the tests before.

> The master branch is not broken for linux with system installed shellcheck (on my fedora32 box).

The test would have been also broken for you if you have used the stable branch even on Linux.

We need to fix the `shellcheck` to a fixed version so all machines will run the same exact test.
This will solve all the problems and this discussion will become meaningless

obnoxxx:
> The test would have been also broken for you if you have used the stable branch even on Linux.

Not really, because due to the one bug that you fix (not using the SHELLCHECK variable), a system-installed shellcheck is anyway the only way to currently run shellcheck in current merged master... :see_no_evil: 

> We need to fix the shellcheck to a fixed version so all machines will run the same exact test.

I agree that this would be best!
One problem is that we currently don't seem to have a way to install shellcheck from the internet in the CI:

https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/openshift_ocs-operator/630/pull-ci-openshift-ocs-operator-master-ocs-operator-ci/1281914511510474752

(Triggered from test-PR https://github.com/openshift/ocs-operator/pull/630)

> This will solve all the problems and this discussion will become meaningless

Not quite, since the history of commits will still matter. But if we had had afixed version before this PR, and/or the stuff working in a reference machine (CI), then this whole discussion about re-ordering might not have happened at all. :wink:

nb-ohad:
@obnoxxx @jarrpa 
I think we have to move the `Remove platform from operator-sdk version` commit (last one) to the start to solve the `verify-latest-csv` failure

obnoxxx:
@nb-ohad 
> I think we have to move the Remove platform from operator-sdk version commit (last one) to be the first one in order to solve the `verify-latest-csv` failure

Why would the order matter?
Also, the verify-latest-csv test is not mandatory. ;-)

obnoxxx:
Oh, surprising how fast this merged after adding the `lgtm` label even if the e2e tests had failed and would have needed to be rerun. :thinking:

PR #641, opened 2020-07-21
Design document for Custom Resource Metrics Exporter

umangachapagain:
This is the design for Metrics Exporter which will be
used to export metrics related to Custom Resources of
interest to OCS.

### DIFF
--- a/docs/design/metrics-exporter.md
+++ b/docs/design/metrics-exporter.md
@@ -1,0 +1,93 @@
+# Custom Resource Metrics Exporter
+
+# Summary
+
+Custom Resource Metrics Exporter provides an insight into the OpenShift Container Storage (OCS) StorageCluster by aggregating data from various Custom Resources
+and exposing metrics about it. The exposed metrics will be in Prometheus compliant format and can be used to build alerts and charts.
+
+# Motivation
+
+The main motivation behind this exporter is to improve user experience for the consumers of OCS. Currently, cluster state and other custom resource related information can only be found by digging deeper into multiple layers via CLI or Console.
+
+To improve that, this exporter will surface out those details in the form of metrics which can then be consumed by Prometheus for analysis as required. Alerts will be created on top of that to notify the user.
+
+# Proposal
+
+The exporter will watch for events on desired Custom Resources and Namespaces and create meaningful metrics with labels. The metric type will be one of Counters, Gauge, Histogram or Summary. Help text and description will be provided along with the metrics for easy understanding.
+
+For example, we will provide RGW health check information based on the CephObjectStore CR. Required parts of the Custom Resource spec is presented below.
+
+```YAML
+apiVersion: ceph.rook.io/v1
+kind: CephObjectStore
+metadata:
+  name: default
+  namespace: openshift-storage
+spec:
+  ...
+  gateway:
+    ...
+    externalRgwEndpoints:
+      - ip: 10.70.56.202
+  ...
+status:
+  bucketStatus:
+    details: >-
+      failed to create object user
+      "rook-ceph-internal-s3-user-checker-9e599ac7-229c-43d4-8e23-40974c9e9177".
+      error code 1 for object store "default": failed to create s3 user: exit
+      status 5
+    health: Failure
+    lastChecked: '2020-07-16T19:20:58Z'
+  info:
+    endpoint: 'http://rook-ceph-rgw-default.openshift-storage:8080'
+  phase: Ready
+```
+
+Name field will be added to label so that the users can uniquely identify the metrics and corelate to historical data. Being an ID, it is unbounded and can have high cardinality.
+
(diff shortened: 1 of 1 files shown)


docs/design/metrics-exporter.md
```diff
@@ -0,0 +1,93 @@
...
+  phase: Ready
+```
+
+Name field will be added to label so that the users can uniquely identify the metrics and corelate to historical data. Being an ID, it is unbounded and can have high cardinality.
+
+Namespace field can have many values depending on the number of namespaces in the cluster. To limit this value, only `openshift-storage` will be used by default and users can enable watching more namespaces using flags.
+
+Health field will have a pre-determined set of values. All these values are considered for label.
+
+Based on these data, we create a Promethues Gauge metric `ocs_rgw_health_status` whose value will be 0(healthy), 1(warning) or 2(failure).
```
leseb, line 53:
```suggestion
Based on these data, we create a Prometheus Gauge metric `ocs_rgw_health_status` whose value will be 0(healthy), 1(warning) or 2(failure).
```

leseb: We don't really have Warning state for the `ocs_rgw_health_status`, it's either `0(healthy)` or `2(failure)`.

docs/design/metrics-exporter.md
```diff
@@ -0,0 +1,93 @@
...
+Labels will be added to provide context as described above.
+
+The exposed metrics will look like:
+
+```
+ocs_rgw_health_status {name="default", namespace="openshift-storage",  rgw_endpoints="10.70.56.202", health="Failure"} 2
```
leseb, line 59: There should always be a single endpoint at this stage, the one provided by the Service. So essentially `Status.Info["endpoint"]`.

docs/design/metrics-exporter.md
```diff
@@ -0,0 +1,93 @@
...
+```
+
+# Goals
+
++ Expose Prometheus metrics from Custom Resources.
++ Expose Prometheus metrics about the exporter itself.
++ Allow configuration via agruments (like setting log level and server port).
```
leseb, line 66:
```suggestion
+ Allow configuration via arguments (like setting log level and server port).
```

docs/design/metrics-exporter.md
```diff
@@ -0,0 +1,93 @@
...
++ Add basic logging.
++ Allow exporter to run in-cluster or standalone.
+  + in-cluster runs in a pod inside the cluster whose CR needs to be monitored.
```
leseb, line 69: So one pod per CR to watch?

anmolsachan: I think there will be one pod only.

umangachapagain: One pod (or maybe Replica 2) per cluster.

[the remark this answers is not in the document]
anmolsachan: nit ^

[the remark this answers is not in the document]
umangachapagain: got rid of ` whose CR needs to be monitored.`.

docs/design/metrics-exporter.md
```diff
@@ -0,0 +1,93 @@
+# Custom Resource Metrics Exporter
+
+# Summary
+
+Custom Resource Metrics Exporter provides an insight into the OpenShift Container Storage Cluster (OCS) by aggregating data from various Custom Resources
```
jarrpa, line 5: `OpenShift Container Storage (OCS) StorageCluster`

docs/design/metrics-exporter.md
```diff
@@ -0,0 +1,93 @@
...
+and exposing metrics about it. The exposed metrics will be in Prometheus compliant format and can be used to build alerts and charts.
+
+# Motivation
+
+The main motivation behind this exporter is to improve user experience for the consumers of OCS. Currently, cluster state and other custom resource related information can only be found by digging deeper into multiple layers via CLI or Console.
+
+To improve that, this exporter will surface out those details in the form of metrics which can then be consumed by Promethues for analysis as required. Alerts will be created on top of that to notify the user.
```
jarrpa, line 12: `Prometheus`

docs/design/metrics-exporter.md
```diff
@@ -0,0 +1,93 @@
...
+  phase: Ready
+```
+
+Name field will be added to label so that the users can uniquely identify the metrics and corelate to historical data. Being an ID, it is unbounded and can have high cardinality.
+
+Namespace field can have many values depending on the number of namespaces in the cluster. To limit this value, only `openshift-storage` will be used by default and users can enable watching more namespaces using flags.
+
+Health field will have a pre-determined set of values. All these values are considered for label.
+
+Based on these data, we create a Promethues Gauge metric `ocs_rgw_health_status` whose value will be 0(Healthy) or 1(Failure).
```
jarrpa, line 53: `Prometheus`

PR #646, opened 2020-07-23
Refactor: Get rid of pkg/deploy-manager

rajatsingh25aug:
Removed `pkg/deploy-manager` and moved all the files into `functests` in a relevant manner.

[JIRA LINK ](https://issues.redhat.com/browse/RHSTOR-1127).


functests/storageclusterinitialization_test.go
```diff
@@ -11,9 +11,9 @@ import (
 	ocsv1 "github.com/openshift/ocs-operator/pkg/apis/ocs/v1"
 	scController "github.com/openshift/ocs-operator/pkg/controller/storagecluster"
 
-	deploymanager "github.com/openshift/ocs-operator/pkg/deploy-manager"
 	storagev1 "k8s.io/api/storage/v1"
 	"k8s.io/apimachinery/pkg/types"
+
```
jarrpa, line 16: Remove this change.

jarrpa:
Well, looks like you did not check the CI locally before pushing. :) Please *always* run `make ocs-operator-ci` before pushing. Do so now and fix the problems.

rajatsingh25aug:
I Didn't check the ci but I do checked the verify latest csv, it passed
locally, not sure why it's falling here.

obnoxxx:
@rajatsingh25aug :
> I Didn't check the ci but I do checked the verify latest csv, it passed locally, not sure why it's falling here.

I don't know why, but the `verify-latest-csv` never passes in the CI, and it is also not required for merging. But as a habit, `make ocs-operator-ci` should always be run locally before submitting a PR for review. (In contrast to the e2e tests, this can easily be run locally.)

functests/storageclusterinitialization_test.go
rajatsingh25aug: Done!

tools/cluster-deploy/cluster-deploy.go
```diff
@@ -6,36 +6,34 @@ import (
 	"log"
 	"os"
 
-	deploymanager "github.com/openshift/ocs-operator/pkg/deploy-manager"
+	deploymanager "github.com/openshift/ocs-operator/functests"
```
obnoxxx, line 9: What I am not too sure about is whether this is actually a clean approach: to import the functests in the tool and use the variables and the cmdline params defined in the functest... Shouldn't the params be defined in the tool?

rajatsingh25aug: @jarrpa ^^

jarrpa: ...yeah, no, that ain't right. Sorry I didn't catch this earlier. The ONLY line that should change in this file is line 9. Everything else should be reverted.

rajatsing:
There were only 2 functions that were taking API client as an input.

- WaitForPVCBound
- WaitForJobSucceeded

This makes changes very small rather what we were doing earlier!.

functests/pvc_creation_test.go
```diff
@@ -43,12 +43,13 @@ var _ = Describe("PVC Creation", func() {
 
 		Context("create pvc", func() {
 			It("and verify bound status", func() {
+				deployManager, err := deploymanager.NewDeployManager()
```
jarrpa, line 46: Don't do this, this is already taken care of earlier: https://github.com/openshift/ocs-operator/pull/646/files#diff-6b1cc57a75b70f1c156ec6ea5d8e4b43R22

rajatsing: Done!

functests/pvc_creation_test.go
```diff
@@ -15,7 +15,7 @@ import (
 
 var _ = Describe("PVC Creation", func() {
 	var k8sClient *kubernetes.Clientset
-
+	var deployManager *deploymanager.DeployManager
```
jarrpa, line 18: Add a newline after this. Also make the same change in `data_validation_test.go`.

rajatsing: They are being **declared**  
https://github.com/openshift/ocs-operator/blob/48295f5bbaa5fa89fceb0c44f4e32ef5bbe1f4ec/functests/data_validation_test.go#L22
and on Line 51
and they are being used in their own way in both of those functions
So I believe, we don't need a declaration there

jarrpa: It can be set on line 22, that's what we want, but the variable needs to be *declared* at the top-level `Describe()`, just after the `var k8sClient` declaration.

PR #647, opened 2020-07-23
Bug 1860034: Toolbox updated to point to the renamed secret keys

travisn:
A recent Rook refactor renamed the keys that are used for ceph operations, which was an internal implementation detail except for the toolbox which consumes these keys.

Signed-off-by: Travis Nielsen <tnielsen@redhat.com>


travisn:
Note that the prior key will still be supported by rook in upgraded clusters. This change was only necessary for new clusters starting in 4.6. No need to worry about the toolbox being in the initialization reconcile since the backward compatibility will remain for that key.

PR #648, opened 2020-07-27
 Avoid bash script to purge osds in job template

xenolinux:
Removed the bash script as we would be running an operator in order to
perform osd removal procedure.

### DIFF
--- a/pkg/controller/storagecluster/reconcile.go
+++ b/pkg/controller/storagecluster/reconcile.go
@@ -530,8 +530,15 @@ func (r *ReconcileStorageCluster) ensureJobTemplates(sc *ocsv1.StorageCluster, r
 		}
 		osdCleanUpTemplate.Parameters = []openshiftv1.Parameter{
 			{
-				Name:     "FAILED_OSD_ID",
-				Required: true,
+				Name:        "FAILED_OSD_IDS",
+				DisplayName: "OSD IDs",
+				Required:    true,
+				Description: `
+The parameter OSD IDs needs a comma-separated list of numerical FAILED_OSD_IDs 
+when a single job removes multiple OSDs. 
+If the expected comma-separated format is not used, 
+or an ID cannot be converted to an int, 
+or if an OSD ID is not found, errors will be generated in the log and no OSDs would be removed.`,
 			},
 		}
 		return controllerutil.SetControllerReference(sc, osdCleanUpTemplate, r.scheme)
@@ -544,7 +551,7 @@ func (r *ReconcileStorageCluster) ensureJobTemplates(sc *ocsv1.StorageCluster, r
 
 func newCleanupJob(sc *ocsv1.StorageCluster) *batchv1.Job {
 	labels := map[string]string{
-		"app": "ceph-toolbox-job-${FAILED_OSD_ID}",
+		"app": "ceph-toolbox-job-${FAILED_OSD_IDS}",
 	}
 
 	// Annotation template.alpha.openshift.io/wait-for-ready ensures template readiness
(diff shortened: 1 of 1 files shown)


pkg/controller/storagecluster/reconcile.go
```diff
@@ -651,7 +651,7 @@ func (r *ReconcileStorageCluster) ensureJobTemplates(sc *ocsv1.StorageCluster, r
 		}
 		osdCleanUpTemplate.Parameters = []openshiftv1.Parameter{
 			{
-				Name:     "FAILED_OSD_ID",
+				Name:     "FAILED_OSD_IDS",
```
jarrpa, line 654: Why is this plural? What's the expected format? What happens if there's spaces?

xenolinux: The FAILED_OSD_IDS would represent the removal of more than one osd at the same time.

jarrpa: That only answers the first of three question.

xenolinux: I think the format should include underscores. Since we are using `FAILED_OSD_IDS` as a parameter to the openshift template when we trigger the template to remove those osds - `oc process -n openshift-storage ocs-osd-removal -p FAILED_OSD_ID=${failed_osd_id} | oc create -f -`. `FAILED_OSD_IDS` will be replaced by the osd ids- 0,1 as such. I am not sure what will happen if there are spaces instead of underscores though, haven't tested that way.

jarrpa: This seems exceedingly error prone. You could easily create a list like `FAILED_OSD_IDS="1 2 3"` which would be a valid parameter as far as `oc process` is concerned. There should be a well-documented format (my preferences would be for commas), and validation that if anything other than numbers or the divisor are detected that the job fail (ideally the processing should fail as well, but I don't think we can do parameter validation in templates).

travisn: To clarify, the Rook implementation expects a comma-separated list of numerical IDs. Rook will iterate over each of the IDs so that a single job can trigger the removal of multiple OSDs if desired. If not in the expected comma-separated format, or an ID cannot be converted to an int, or if an OSD ID is not found, the invalid ID(s) will be skipped and an error will be logged in the job. For example, if you use spaces, you'll just get errors in the log and no OSDs would be removed.

travisn:
@xenolinux This is ready for merge, right? It still has the do-not-merge label. A BZ was just opened related to this as well. https://bugzilla.redhat.com/show_bug.cgi?id=1879008

xenolinux:
> @xenolinux This is ready for merge, right? It still has the do-not-merge label. A BZ was just opened related to this as well. https://bugzilla.redhat.com/show_bug.cgi?id=1879008

Yes. But before merging I wanted to confirm - Volumes and VolumeMounts that mount an emptyDir to /var/lib/rook and ROOK_LOG_LEVEL. That will be needed right?

agarwal-mudit:
https://bugzilla.redhat.com/show_bug.cgi?id=1879008 is a blocker for OCS4.6

pkg/controller/storagecluster/reconcile.go
jarrpa: This information should be captured in the description of the parameter. I believe there's a Parameter.Description field: https://docs.openshift.com/container-platform/4.5/rest_api/template_apis/template-template-openshift-io-v1.html#template-template-openshift-io-v1

While we're at it, a more readable `DisplayName` like "OSD IDs" would probably be useful.

xenolinux: Okay. Noted. Two things to add: description and displayName of the parameter.

xenolinux: @jarrpa Could you please review? I added a description and a DisplayName for the parameter.

travisn:
> > @xenolinux This is ready for merge, right? It still has the do-not-merge label. A BZ was just opened related to this as well. https://bugzilla.redhat.com/show_bug.cgi?id=1879008
> 
> Yes. But before merging I wanted to confirm - Volumes and VolumeMounts that mount an emptyDir to /var/lib/rook and ROOK_LOG_LEVEL. That will be needed right?

@xenolinux ROOK_LOG_LEVEL is optional, but I don't recall if the emptyDir for /var/lib/rook is necessary. Were you able to test these changes from the OCS template? That should should if it's necessary.

jarrpa (approved):

jarrpa:
Sorry for the long delays. Thank you for adding the description!

travisn:
@xenolinux Were you able to verify this is working in OpenShift?

PR #654, opened 2020-07-29
Ints for requests and limits are valid

mmgaggle:
Mirror what is done by Kube

https://github.com/kubernetes-sigs/controller-tools/pull/381/files#diff-83f69cf64da4a50ca73fc80e1de20c25R1774-R1778

### DIFF
--- a/deploy/crds/ocs.openshift.io_storageclusters_crd.yaml
+++ b/deploy/crds/ocs.openshift.io_storageclusters_crd.yaml
@@ -181,7 +181,11 @@ spec:
                             resources allowed. More info: https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/'
                           type: object
                           additionalProperties:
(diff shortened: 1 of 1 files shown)


sttts:
Lgtm.

JohnStrunk:
Isn't the CRD generated automatically from *_types.go by operator-sdk?
I think we need to add the proper annotations there such that the CRD is generated correctly (if possible).

umangachapagain:
> Isn't the CRD generated automatically from *_types.go by operator-sdk?
> I think we need to add the proper annotations there such that the CRD is generated correctly (if possible).

Yes.
While we want to reach this state, directly changing the CRD spec will not work.
There are multiple steps to get this done properly.

umangachapagain:
Added complete fix here: https://github.com/openshift/ocs-operator/pull/659

PR #661, opened 2020-07-31
Remove RGW storage class as required from the UI validation

bipuladh:
Since all deployments do not have RGW. Hence removing RGW as a required parameter in the UI validation as well. The changes in the exporter script are done at https://github.com/rook/rook/pull/5899. 

cc @leseb 
/assign @jarrpa


bipuladh:
I ran `make gen-latest-csv`

PR #662, opened 2020-07-31
Add labels for architectures that are supported

manojnkumar:
For the operators to be available from the UI,  architecture abels are needed.


manojnkumar:
> @umangachapagain is right, we don't edit the CSV directly, it's impossible to maintain by hand. General CSV changes that don't rely on parsing go structs should go in `tools/csv-merger/csv-merger.go`. Also, are we actually testing OCS across all these architectures, or are these just the architectires OCP is supported on?

The last time I looked at this @umangachapagain @jarrpa is that there was a boot-strap issue. I could not add the ppc64le architecture on `ppc64le` because `make gen-latest-csv` depends on pulling the container images, and the images pulled were for `x86_64`.

manojnkumar:
Messed up on rebase. Closing. Will open a new PR.

PR #663, opened 2020-08-03
enable pruning and add CRD verification for Pruning

umangachapagain:
+ adding verify-crds.sh that will check if pruning is enabled
for CRDs and if it has metadata description (which is not
allowed in Structural Schema).


umangachapagain:
Pruning will be enabled by default when using `apiextensions/v1`.

PR #666, opened 2020-08-03
Remove excess logs from OCS operator

rajatsing:
Remove unwanted logs that gets generated in OCS operator, which makes looking into the logs a bit difficult.

### DIFF
--- a/pkg/controller/storagecluster/reconcile.go
+++ b/pkg/controller/storagecluster/reconcile.go
@@ -114,7 +114,7 @@ func (r *ReconcileStorageCluster) Reconcile(request reconcile.Request) (reconcil
 			instance.Status.Phase = statusutil.PhaseIgnored
 			phaseErr := r.client.Status().Update(context.TODO(), instance)
 			if phaseErr != nil {
-				reqLogger.Error(phaseErr, "Failed to set PhaseIgnored")
+				reqLogger.Info("Status Update Error", "StatusUpdateErr", "Failed to set PhaseProgressing")
 				return reconcile.Result{}, phaseErr
 			}
 			return reconcile.Result{}, nil
@@ -138,7 +138,7 @@ func (r *ReconcileStorageCluster) Reconcile(request reconcile.Request) (reconcil
 		instance.Status.Phase = statusutil.PhaseProgressing
 		phaseErr := r.client.Status().Update(context.TODO(), instance)
 		if phaseErr != nil {
-			reqLogger.Error(phaseErr, "Failed to set PhaseProgressing")
+			reqLogger.Info("Status Update Error", "StatusUpdateErr", "Failed to set PhaseProgressing")
 		}
 	}
 
@@ -149,7 +149,7 @@ func (r *ReconcileStorageCluster) Reconcile(request reconcile.Request) (reconcil
 		statusutil.SetProgressingCondition(&instance.Status.Conditions, reason, message)
 		err = r.client.Status().Update(context.TODO(), instance)
 		if err != nil {
-			reqLogger.Error(err, "Failed to add conditions to status")
+			reqLogger.Info("Status Update Error", "StatusUpdateErr", "Failed to add conditions to status")
 			return reconcile.Result{}, err
 		}
 	}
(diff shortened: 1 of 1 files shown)


pkg/controller/storagecluster/statusUpdate.go
```diff
@@ -0,0 +1,15 @@
...
+import (
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+)
+
+// statusUpdate updates verbose log to a shorter and concise log
+func (r *ReconcileStorageCluster) statusUpdate(err error) error {
+	if errors.IsConflict(err) {
+		return fmt.Errorf("the object has been updated")
```
jarrpa, line 12: `s/updated/modified/`

pkg/controller/storagecluster/uninstall_reconciler.go
```diff
@@ -152,9 +152,8 @@ func (r *ReconcileStorageCluster) deleteCephCluster(sc *ocsv1.StorageCluster, re
 		if errors.IsNotFound(err) {
 			reqLogger.Info("Uninstall: CephCluster not found")
 			return nil
-		} else {
-			return fmt.Errorf("Uninstall: Unable to retrive cephCluster: %v", err)
 		}
+		return fmt.Errorf("Uninstall: Unable to retrive cephCluster: %v", err)
```
rajatsing, line 156: @jarrpa I made these change since this was pointed by the local CI.

deploy/crds/ocs.openshift.io_storageclusters_crd.yaml
```diff
@@ -181,11 +181,7 @@ spec:
                             resources allowed. More info: https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/'
                           type: object
                           additionalProperties:
-                            pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
-                            anyOf:
-                            - type: integer
-                            - type: string
-                            x-kubernetes-int-or-string: true
+                            type: string
```
rajatsing, line 184: Its strange that `verify-latest-csv` is failing until and unless I won't commit these changes that gets generated upon running `make-update-generated` on my local machine!.

pkg/controller/storagecluster/modify.go
```diff
@@ -0,0 +1,18 @@
...
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// StatusUpdate updates verbose log to a shorter and concise log
+func (r *ReconcileStorageCluster) StatusUpdate(obj runtime.Object) error {
+	err := r.client.Status().Update(context.TODO(), obj)
+	if errors.IsConflict(err) {
+		return fmt.Errorf("the object has been updated")
```
jarrpa, line 15:
```suggestion
		return fmt.Errorf("the object has been modified")
```

rajatsing: Yes, was planning to add test cases. I can think of only 2. What's the third one?.

-An Object is updated
- An object is not updated
- what about third?

jarrpa: 1. object has not been modified
2. object has been modified
3. object doesn't exist

rajatsing: done!

umangachapagain:
This PR in it's current form will be obsolete after #717 .

I suggest that we revamp this PR to set correct log level and keys for all Status Update logs. This will allow logs filtering.

Currently we set status update failures as Error. Instead we could set it as Warning which will disable the stacktrace (in Production Clusters).

Then instead of `reqLogger.Error(phaseErr, "Failed to set PhaseProgressing")` if we do `reqLogger.Error(phaseErr, "Status Update Error", "StatusUpdateErr",  "Failed to set PhaseProgressing")` it'll drastically reduce log level and allow filtering it out when needed.

umangachapagain:
> @umangachapagain Can you please explain what this accomplishes? You say to change it to a Warning, but we still use the Error() function. Nothing about the parameters you're proposing explains how this is now suddenly a warning. 
>
Ah, my bad. we need to use `reqlogger.Warning()`. Copy-paste mistake. 

>Also, best I knew the first string is intended to be a format string. With no formatting, what do the subsequent fields accomplish?
>
first string is message and subsequent fields are key-value pairs of interface.

So if for all Warning logs, if you have the same message, it'll be considered as duplicate and suppressed.

Remaining key-value pairs are to provide more context and can be used to filter logs (we don't need it but probably Logging stack can benefit from it)

PR #668, opened 2020-08-03
Add csi log level to rook-ceph-operator-config configmap

Madhu-1:
as the CSI log level is set to `0` by default no debug and info logs will be printed it will be
hard to debug csi. This commit sets the csi log level as "5".


Madhu-1:
@umangachapagain rebased

umangachapagain:
@Madhu-1 Do we want this to be user configurable? If not, it's better to add it to the Env Vars in CSV. WDYT?

Madhu-1:
> @Madhu-1 Do we want this to be user configurable? If not, it's better to add it to the Env Vars in CSV. WDYT?

Yes if the admin doesn't want to see all debug/info logs he can configure this value

PR #671, opened 2020-08-04
Added the pattern to the end of the command

xenolinux:
Added || true to the end of the command. If the command ceph osd crush rm
gives output with error message, the script will abort
and the job would fail. This change ensures that the script doesn't
abort and the job runs without a failure.


xenolinux:
> LGTM, were you able to verify in a cluster with multiple osds per node that this ignores the error?

@travisn I am testing it currently.

PR #672, opened 2020-08-05
Bug 1840539: [release 4.5] Added the pattern to the end of the command

agarwal-mudit:
Added || true to the end of the command. If the command ceph osd crush rm
gives output with error message, the script will abort
and the job would fail. This change ensures that the script doesn't
abort and the job runs without a failure.


agarwal-mudit:
Backport PR for https://github.com/openshift/ocs-operator/pull/671

agarwal-mudit:
@jarrpa @umangachapagain PTAL, this is a backport PR to 4.5

PR #687, opened 2020-08-06
Containerize operator binary build

jarrpa:

### DIFF
--- a/build/Dockerfile.build
+++ b/build/Dockerfile.build
@@ -1,0 +1,10 @@
+FROM registry.access.redhat.com/ubi8/ubi-minimal as builder
+
+RUN microdnf install -y make golang; microdnf -y clean all
+
+FROM builder
+
+RUN mkdir -p /output
+COPY . /ocs-operator
+WORKDIR /ocs-operator
+RUN make go-build OUTDIR_BIN=/output
--- a/Makefile
+++ b/Makefile
@@ -168,7 +168,7 @@ verify-generated: update-generated
 	@echo "Verifying generated code"
 	hack/verify-generated.sh
 
-ocs-operator-ci: shellcheck-test gofmt golint govet unit-test build verify-generated verify-latest-deploy-yaml
+ocs-operator-ci: shellcheck-test gofmt golint govet unit-test verify-generated verify-latest-deploy-yaml
 
 red-hat-storage-ocs-ci:
 	@echo "Running red-hat-storage ocs-ci test suite"
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,3 @@
-TARGET_GOOS=linux
-TARGET_GOARCH=amd64
-
 # Set minimum required Golang version to v1.13.3
 GO_REQUIRED_MIN_VERSION=1.13.3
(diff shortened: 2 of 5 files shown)


jarrpa:
@umangachapagain PTAL. I tried to optimize it by either detecting code changes in the host directory or reusing the build cache on the host, but I couldn't figure anything out. If you have any ideas on how to make it faster, please let me know. Otherwise I think we can live with the extended build times.

umangachapagain:
@nb-ohad had some idea about containerized builds.

I believe the basic idea was to use `operator-sdk build` instead.

build/Dockerfile.build
```diff
@@ -0,0 +1,10 @@
+FROM registry.access.redhat.com/ubi8/ubi-minimal as builder
+
+RUN microdnf install -y make golang; microdnf -y clean all
+
+FROM builder
+
+RUN mkdir -p /output
+COPY . /ocs-operator
+WORKDIR /ocs-operator
+RUN make deps-update && make go-build OUTDIR_BIN=/output
```
umangachapagain, line 10:
```suggestion
RUN make build-go OUTDIR_BIN=/output
```
since build-go already calls deps-update

jarrpa: Legit.

jarrpa:
> @nb-ohad had some idea about containerized builds.
> 
> I believe the basic idea was to use `operator-sdk build` instead.

It was, but my investigation leads me to believe that `operator-sdk build` does NOT perform containerized builds of the operator binaries, merely builds the operator container image. Furthermore, there is no way to NOT build the operator container image, so we can't call the command from within the CI anyway.

nb-ohad:
@jarrpa You are correct. I checked on the noobaa side and got to the conclusions that it is not a "containerized" build but a dockerized build (which means build using docker)

Still, there are benefits of using `operator-sdk build` as the image itself will not targeted for the local OP system the build is running on. It will be targeted to run as a container.

jarrpa:
> Still, there are benefits of using `operator-sdk build` as the image itself will not targeted for the local OP system the build is running on. It will be targeted to run as a container.

@nb-ohad I don't understand what this means. I looked at the debug logs, and it runs these two commands:
```
DEBU[0000] Running []string{"go", "build", "-o", "/home/jrivera/projects/golang/src/github.com/openshift/ocs-operator/build/_output/bin/ocs-operator", "-gcflags", "all=-trimpath=/home/jrivera/projects/golang/src/github.com/openshift", "-asmflags", "all=-trimpath=/home/jrivera/projects/golang/src/github.com/openshift", "-mod=vendor", "github.com/openshift/ocs-operator/cmd/manager"} 
DEBU[0029] Running []string{"docker", "build", "-f", "build/Dockerfile", "-t", "jarrpa/ocs-operator", "."}
```
My understanding is that the `go build` above uses the system environment (e.g. go version installed on you laptop) as opposed to any standard environment (e.g. starting the build within a container based on `registry.access.redhat.com/ubi8/ubi-minimal`). As such, each developer will be potentially building binaries using different go tools in different runtime environments.

And that still doesn't address the problem that we **can't run this in the CI** because there's no way to NOT call the docker command.

PR #688, opened 2020-08-07
Update verify targets

jarrpa:
This PR updates the `make verify-*` targets to be more user-friendly. It removes the need for the CSV checksum file and provides better output for when the verifications fails. While there are no workflow changes in terms of what a developer has to do, the following differences will be noticed:

When running CSV targets locally:
* `make verify-latest-csv` now always calls `gen-latest-csv`
* `make gen-latest-csv` will still source the required CSV manifests from other operator containers, but it will also save the extracted files under `deploy/csv-templates/`.
  * This directory must be added to the commit that also commits the final CSV changes.

When the ci-operator runs the CSV targets:
* `make gen-latest-csv` now detects if we're running in the OCP CI
  * If so, don't source CSV manifests from other operators (since those require docker commands)
  * Take the manifests from `deploy/csv-templates`


jarrpa:
> But, storing all CRDs locally to prevent sourcing of other operator images won't work. It adds us the overhead to update these CRDs everytime there is an Upstream update.

I don't see much in the way of overhead. It all gets updated when you run `make gen-latest-csv` locally, so there are no additional commands needed to run compared to before this change. Just add all the changes to the same CSV commits as before.

> I think we should still pull the images, just ensure that we pull the right one and not take just anyting from cache.

The problem is we can't pull the images in the CI. The ci-operator launches all its tests in containers, and those containers can't run docker commands.

umangachapagain:
> > But, storing all CRDs locally to prevent sourcing of other operator images won't work. It adds us the overhead to update these CRDs everytime there is an Upstream update.
> 
> I don't see much in the way of overhead. It all gets updated when you run `make gen-latest-csv` locally, so there are no additional commands needed to run compared to before this change.
>
how will it update?
Can you list the steps to update and generate latest CSV with this PR (locally and in CI)?

jarrpa:
PR rebased and I updated the description with the requested information. It doesn't entirely make sense to add it to the individual commits, since the changes to the final workflow is spread across multiple commits.

jarrpa:
Rebased and updated to incorporate latest Rook-ceph image.

umangachapagain:
2 tests on `ocs-operator-e2e-aws` fails with a panic. 

/retest

umangachapagain:
These tests weren't failing before. I don't see why it would fail on this PR.
Can you take a look at it before we merge?

/hold
/lgtm

umangachapagain:
Seems the CI issue is unrelated.

/unhold

umangachapagain (approved):

umangachapagain (approved):

PR #691, opened 2020-08-10
enables RBD per image IO stats collection by default

umangachapagain:
Part of https://github.com/rook/rook/pull/5680

### DIFF
--- a/pkg/controller/storagecluster/cephblockpools.go
+++ b/pkg/controller/storagecluster/cephblockpools.go
@@ -28,6 +28,7 @@ func (r *ReconcileStorageCluster) newCephBlockPoolInstances(initData *ocsv1.Stor
 					Size:            3,
 					TargetSizeRatio: .49,
 				},
+				EnableRBDStats: true,
 			},
 		},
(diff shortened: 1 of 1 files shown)


pkg/controller/storagecluster/cephblockpools.go
```diff
@@ -28,6 +28,7 @@ func (r *ReconcileStorageCluster) newCephBlockPoolInstances(initData *ocsv1.Stor
 					Size:            3,
 					TargetSizeRatio: .49,
 				},
+				EnableRBDStats: true,
```
leseb, line 31: Why hard-coding it?

jarrpa: Why not? If it is OCS recommendation to enable it, and it will consistently provide value, we are well within our rights to simply enable it in all cases. Our goal is to minimize configuration points, after all.

umangachapagain: Yes, it was requested for this to be always enabled.

PR #697, opened 2020-08-12
Disable validation of CephFS secrets

bipuladh:
Since CephFS might not be present in all clusters the validation on the UI should be loosened to allow creation without providing `rook-csi-cephfs-node`, `rook-csi-cephfs-provisioner` Secret objects.


pkg/controller/storagecluster/external_resources.go
```diff
@@ -37,6 +38,36 @@ type ExternalResource struct {
 	Name string            `json:"name"`
 }
 
+// filterUnavailableStorageClasses removes SC's that are not present in the ExternalResource array
+func (r *ReconcileStorageCluster) filterUnavailableStorageClasses(scs []*storagev1.StorageClass, data []ExternalResource) (filteredScs []*storagev1.StorageClass, disableCephFS bool) {
+	disableCephFS = true
+	// RBD should always be present
+	filteredScs = append(filteredScs, scs[1])
+	for _, d := range data {
```
jarrpa, line 46: Don't cycle through all the data. Cycle through `scs`, check which StorageClassName const they match, and append to filteredScs that way.

PR #701, opened 2020-08-12
Bug 1862449: [release-4.5][Backport] Create only those StorageClasses which are provided in the external input file

aruniiird:
In External Cluster mode, create only those StorageClasses which are
provided in JSON input file.


pkg/controller/storagecluster/external_resources.go
```diff
@@ -54,39 +54,38 @@ func (r *ReconcileStorageCluster) ensureExternalStorageClusterResources(instance
...
+		Name:       instance.Name,
+	}
+	scs, err := r.newStorageClasses(instance)
+	if err != nil {
+		reqLogger.Error(err, "failed to create StorageClasses")
+		return err
+	}
+	// make a ZERO length array with capacity same as 'scs' length
+	// will store only the provided StorageClasses
+	availableSCs := make([]*storagev1.StorageClass, 0, len(scs))
```
leseb, line 88: Why not making the length dynamic instead of calling `newStorageClasses()`? It seems overkill to call it just to get a fixed length...

aruniiird: The above `make([]<type>, 0, size)` call will create a ZERO length array, with capacity of `size`. It is dynamic (as long as you call `append()` function to add items into the array).
The only advantage of the above `make` is while doing an `append`, the array don't have to relocate (no extra garbage collection required) itself in order to accommodate newer elements.
More details here: https://stackoverflow.com/questions/33834742/remove-and-adding-elements-to-array-in-go-lang

pkg/controller/storagecluster/external_resources.go
```diff
@@ -54,39 +55,58 @@ func (r *ReconcileStorageCluster) ensureExternalStorageClusterResources(instance
...
-                reqLogger.Error(err, "could not create ExternalStorageClusterResource")
-                return err
-        }
-	// creating all the storageClasses once we set the values
-	err = r.createStorageClasses(scs, reqLogger)
+	instance.Status.ExternalSecretFound = true
+	return nil
+}
+
+func (r *ReconcileStorageCluster) setRookCSICephFS(
```
aruniiird, line 73: Added the function from @bipuladh 's PR: https://github.com/openshift/ocs-operator/pull/697

pkg/controller/storagecluster/external_resources.go
```diff
@@ -123,25 +143,38 @@ func (r *ReconcileStorageCluster) createExternalStorageClusterResources(ownerRef
...
+			availableSCs = append(availableSCs, sc)
 		}
 	}
+	// creating only the available storageClasses
+	err = r.createStorageClasses(availableSCs, reqLogger)
+	if err != nil {
+		reqLogger.Error(err, "failed to create needed StorageClasses")
+		return err
+	}
+	if err = r.setRookCSICephFS(enableRookCSICephFS, instance, reqLogger); err != nil {
```
aruniiird, line 173: Here, we could have added a `if` condition first, to check something like,
```suggestion
	if !enableRookCSICephFS {
		if err = r.setRookCSICephFS(enableRookCSICephFS, instance, reqLogger); err != nil {
```
That is, when `CephFS` storageclass is not enabled, then only call the `setRookCSICephFS()` to disable the CSI drivers.
My reasoning to not to call it that way is, on a later stage when we decide to `reconcile` external resource creation, the above part of the code don't have to be changed.
Let me know what others think?

pkg/controller/storagecluster/external_resources.go
leseb: I understand the underlying effect, I'm just not sure if it's worth the optimization given that we call `r.newStorageClasses(instance)`. I'd say that's almost confusing too unless you add a comment explaining the rationale.

aruniiird: About calling `newStorageClasses()` function, we require the actual StorageClasses to be created first and then we use each SC to fix their parameters according to the JSON input. We can see those changes in [Line#147](https://github.com/openshift/ocs-operator/pull/701/files#diff-e5a5352dff74ec9aa443933a8a24fd0eR147)

pkg/controller/storagecluster/external_resources.go
jarrpa: It took me a bit to parse, but I think I understand. The point is you always want to set the CephFS CSI config to *something*, whether true or false. This is good!

pkg/controller/storagecluster/external_resources.go
```diff
@@ -54,39 +59,58 @@ func (r *ReconcileStorageCluster) ensureExternalStorageClusterResources(instance
...
+		APIVersion: instance.APIVersion,
+		Kind:       instance.Kind,
+		Name:       instance.Name,
+	}
+	scs, err := r.newStorageClasses(instance)
+	if err != nil {
+		reqLogger.Error(err, "failed to create StorageClasses")
+		return err
+	}
+	// this flag set the 'ROOK_CSI_ENABLE_CEPHFS' flag
```
jarrpa, line 110:
```suggestion
	// this flag sets the 'ROOK_CSI_ENABLE_CEPHFS' flag
```

pkg/controller/storagecluster/external_resources.go
jarrpa: I agree, this looks like a needless optimization.

pkg/controller/storagecluster/external_resources.go
```diff
@@ -54,39 +59,58 @@ func (r *ReconcileStorageCluster) ensureExternalStorageClusterResources(instance
...
+	enableRookCSICephFS := false
+	// to store only the provided StorageClasses
```
jarrpa, line 112:
```suggestion
	// this stores only the StorageClasses specified in the Secret
```

pkg/controller/storagecluster/external_resources.go
```diff
@@ -123,25 +147,38 @@ func (r *ReconcileStorageCluster) createExternalStorageClusterResources(ownerRef
...
 				sc = scs[1]
 			} else if d.Name == cephRgwStorageClassName {
 				// Set the external rgw endpoint variable for later use on the Noobaa CR (as a label)
 				// Replace the colon with an underscore, otherwise the label will be invalid
 				externalRgwEndpointReplaceColon := strings.Replace(d.Data[externalCephRgwEndpointKey], ":", "_", -1)
 				externalRgwEndpoint = externalRgwEndpointReplaceColon
 
-				// Setting the Endpoint for OBC StorageClass
+				// Setting the 'sc' point to OBC StorageClass
 				sc = scs[2]
```
jarrpa, line 163: I just realized something... `newStorageClasses()` doesn't always return three StorageClasses. The OBC StorageClass may not always be present. Should we have a safeguard in case there is a misconfiguration over the RGW StorageClass?

aruniiird: @jarrpa , we don't require the check, why because the function `ensureExternalStorageClusterResources()` will only be called if we are in external mode (i.e; `Spec.ExternalStorage.Enable` is true) and `newStorageClasses()` function will allways return the third OBC storage class (`scs[2]`) without fail in external mode.

pkg/controller/storagecluster/external_resources.go
aruniiird: Done...

pkg/controller/storagecluster/external_resources.go
aruniiird: Done...

pkg/controller/storagecluster/external_resources.go
aruniiird: Done...

pkg/controller/storagecluster/external_resources.go
aruniiird: Changed...

pkg/controller/storagecluster/external_resources.go
obnoxxx: I agree this should be caught in the function, not outside.

leseb:
> In general, the changes look clean.
> But I am wondering why we are only making the cephfs csi driver configurable this way?
> Rook also knows `ROOK_CSI_ENABLE_RBD` ...

CephFS gets a "special" treatment because the external cluster might not have any MDS daemons running whereas RBD will always be there. RBD is provided by OSDs and you cannot have a functioning Ceph cluster with no OSDs :).

obnoxxx:
@leseb  writes:
> CephFS gets a "special" treatment because the external cluster might not have any MDS daemons running whereas RBD will always be there. RBD is provided by OSDs and you cannot have a functioning Ceph cluster with no OSDs :).

True, but I thought this was not about the presence of daemons (only) but about the presence of the required pools, but maybe that's not even the point? So is the control only about making sure that the CSI driver has the required daemons that it needs to talk to?

Anyway, rook has `ROOK_CSI_ENABLE_RBD`, this is why this struck my as asymmetric..

leseb:
> @leseb writes:
> 
> > CephFS gets a "special" treatment because the external cluster might not have any MDS daemons running whereas RBD will always be there. RBD is provided by OSDs and you cannot have a functioning Ceph cluster with no OSDs :).
> 
> True, but I thought this was not about the presence of daemons (only) but about the presence of the required pools, but maybe that's not even the point? So is the control only about making sure that the CSI driver has the required daemons that it needs to talk to?

With the script, it's about whether we detect a filesystem or fs pool was specified (and exists along with a filesystem). I just generalized when I said "mds daemon".

> 
> Anyway, rook has `ROOK_CSI_ENABLE_RBD`, this is why this struck my as asymmetric..

Yes, if someone only wants to use the FS over the block... Not so common but still possible.

pkg/controller/storagecluster/external_resources.go
jarrpa: Hm.... okay, I'll let it slide for now.

aruniiird:
Added tests which I think should be there for optional resources. They require tweaking.
@jarrpa , @leseb let me know, if you see other test scenarios which should be added.

PR #705, opened 2020-08-13
Metrics Exporter

umangachapagain:
This Pull Request adds a Metrics Exporter to OCS Operator project based on design discussion at #641 

Approved design document: https://github.com/openshift/ocs-operator/blob/master/docs/design/metrics-exporter.md

Currently, this metrics exporter runs alongside the component operators and watched CephObjectStore CR. It translates CephObjectStore Status info into metrics in format directly consumable by Prometheus.

For more information on Prometheus and related concepts: https://prometheus.io/docs/introduction/overview/
For more information on exporters: https://prometheus.io/docs/instrumenting/writing_exporters/


metrics/deploy/deployment.yaml
```diff
@@ -0,0 +1,34 @@
...
+  selector:
+    matchLabels:
+      app.kubernetes.io/component: ocs-metrics-exporter
+      app.kubernetes.io/name: ocs-metrics-exporter
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/component: ocs-metrics-exporter
+        app.kubernetes.io/name: ocs-metrics-exporter
+        app.kubernetes.io/version: 0.0.1
```
jarrpa, line 21: Are we going to be shipping this as a new build in the product? If so, shouldn't this be the same as the current ocs-operator version?

jarrpa:
Before I forget: we should probably be adding unit tests for this service where possible. I have no familiarity with mocking environments for HTTP services, however, so I don't know how much work that is.

umangachapagain:
/hold
Depends on #710

metrics/deploy/deployment.yaml
umangachapagain: No new build is required. Binary is added to ocs-operator image and run as a separate deployment internally.

metrics/internal/collectors/ceph-object-store.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+// Collect implements prometheus.Collector interface
+func (c *CephObjectStoreCollector) Collect(ch chan<- prometheus.Metric) {
+	cephObjectStoreLister := cephv1listers.NewCephObjectStoreLister(c.Informer.GetIndexer())
+	cephObjectStores := getAllObjectStores(cephObjectStoreLister, c.AllowedNamespaces)
+	if len(cephObjectStores) > 0 {
+		c.collectObjectStoreHealth(cephObjectStores, ch)
+	} else {
+		ch <- prometheus.NewInvalidMetric(c.RGWHealthStatus, fmt.Errorf("CephObjectStore not found"))
```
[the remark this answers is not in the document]
umangachapagain: Removing `NewInvalidMetric` from all places as not all errors should return error 500.

metrics/internal/handler/handler.go
```diff
@@ -0,0 +1,67 @@
...
+	}, []string{"code"})
+
+	registry.MustRegister(
+		requestsTotal,
+		requestsInFlight,
+		requestDuration,
+	)
+
+	return promhttp.InstrumentHandlerDuration(
+		requestDuration.MustCurryWith(prometheus.Labels{"code": "200"}),
```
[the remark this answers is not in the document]
umangachapagain: `promhttp.InstrumentHandlerDuration` expects a `ObserverVec` but `GetMetricsWith` returns `Observer`.
So using `MustCurryWith` gives us `ObserverVec` and panics in case of error.

go.mod
```diff
@@ -19,7 +20,9 @@ require (
 	github.com/openshift/custom-resource-status v0.0.0-20190812200727-7961da9a2eb7
 	github.com/operator-framework/operator-lifecycle-manager v0.0.0-20200321030439-57b580e57e88
 	github.com/operator-framework/operator-sdk v0.17.0
+	github.com/prometheus/client_golang v1.5.1
```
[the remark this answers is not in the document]
umangachapagain: Upgraded to v1.7.1. Version is independent of OCP version.

metrics/deploy/deployment.yaml
jarrpa: :+1:

jarrpa:
> > In general, don't split the function call and the function definition across multiple commits.
> 
> These commits are in exact sequence in which I built things. Making any kind of changes means I'll need to re-work on the entire code (move code in and out, merge commits, fix conflicts etc etc). In the interest of time, can we remove this restriction?
> IMO, having Function definition in one commit and Function Call in the commit after won't cause anything to break.

It's not that it'll cause things to break, it's that it makes it hard to review and understand the history in the future. I will not remove this request in the interest of time.

> > A few additional things:
> > ```
> > * For each resulting commit, add a proper description where one is missing. I'm still not confident I understand what this is doing, internally
> > ```
> 
> We could discuss that here, or if you want I can add a comments in code but I don't agree with explaining the code internals in a commit message (anything more than what the commit does and how to use it).

Sure, more code comments would be appreciated.

metrics/internal/handler/handler.go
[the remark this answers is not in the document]
simonpasquier: Sorry I wasn't clear. What I meant is that you should pass `requestDuration` to `promhttp.InstrumentHandlerDuration` without calling `MustCurryWith()` at all.

```suggestion
	return promhttp.InstrumentHandlerDuration(
		requestDuration,
```

obnoxxx:
@umangachapagain writes:
> These commits are in exact sequence in which I built things.

Indeed, this shows clearly, and it is normal that for bigger changes, this is how an early stage of the branch looks like, but this is not usually how it will be arranged and committed in the end.

> Making any kind of changes means I'll need to re-work on the entire code (move code in and out, merge commits, fix conflicts etc etc).

Correct. This is part of the software development process. The resulting commits need to be good, not just the resulting code.

Some guidelines:

* each commit should build
* each commit should pass tests

If you don't want to spend a lot of time interactively rebasing, moving, re-arranging, etc, you can also squash more aggressively: In this case, introducing a new component/feature as a new file or even an entirely new directory subtree, you can just add the whole file or subtree in the final form in a single commit. But make sure it builds (i.e. add vendor and dep updates in the same commit), and fix failing tests in the same commit (if any). You can possibly add additional tests with separate commits.

This PR does not explain what it is actually aiming to do. The description should be updated.
(Same for commit messages, but there's a bit more context there than in the PR.)

metrics/internal/handler/handler.go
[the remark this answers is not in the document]
umangachapagain: ok. Fixed.

obnoxxx:
@umangachapagain - thanks for the summary description!

tools/csv-merger/csv-merger.go
```diff
@@ -835,6 +855,52 @@ func copyManifests() {
 	}
 }
 
+func getMetricsExporterDeployment() appsv1.DeploymentSpec {
```
jarrpa, line 858: Since we're already creating the Service and such in the StorageCluster reconciliation, why are we doing this here instead of the operator code? As you say, it's not an operator, it doesn't make sense to treat it both as a separate component and a part of reconciliation.

metrics/internal/collectors/ceph-object-store_test.go
```diff
@@ -0,0 +1,294 @@
...
+	}
+}
+
+func TestNewCephObjectStoreCollector(t *testing.T) {
+	type args struct {
+		opts *options.Options
+	}
+	tests := []struct {
+		name string
+		args args
```
jarrpa, line 95: Yes! I love seeing case tables from the get-go. :) But, instead of creating a separate `args` struct, just define the arguments as part of the main struct.

anmolsachan:
Just ran this on a Internal cluster mode: 
Works fine in terms of functionality
Screenshot from 2020-08-25 20-51-42
Screenshot from 2020-08-25 20-48-56

umangachapagain:
Before I make any more changes I'd like to get more opinions on this PR.
I'd like to take this as a learning opportunity and understand how a feature should be developed (development process vs final implementation).

/cc @leseb @travisn @humblec @agarwal-mudit @cloudbehl @anmolsachan

metrics/internal/version/version.go
```diff
@@ -0,0 +1,8 @@
+package version
+
+const version = "0.0.1"
```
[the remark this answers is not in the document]
umangachapagain: Can do. But, weren't we also talking about getting rid of the version in operator?

And, since exporter is a separate component (like rook noobaa csi etc) used by OCS, can't it have a different version?

I'm fine with using any version that aligns to our practices.

tools/csv-merger/csv-merger.go
umangachapagain: It's just like you install Rook via CSV and CephCluster(and it's services) via reconcile.

I kept deployment via CSV as it'll now be centrally managed along with other components (no need to separately reconcile deployment) via OLM. 

Also, it doesn't really fit into OCS reconcile because it's not a part of StorageCluster Initialization or Creation etc.
However, Service and Service Monitor fit into reconcile as they are needed wherever StorageCluster is created. Makes sense?

P.S. It's always treated as separate component. If you closely look at reconcile, Service/Service Monitor state doesn't affect OCS Operator Readiness unlike other components (which marks OCS as Not Ready during reconcile).

metrics/internal/collectors/ceph-object-store_test.go
umangachapagain: This was auto-generated as part of standard go unit test generation. And, I retained it as it seemed cleaner to keep `args` isolated from rest of the fields.

umangachapagain:
>     * While waiting to bring in option flags until later makes sense in the development process, it does not make sense for the final implementation. Please merge most of the work of "adds flags/options for configuration flexibility" into "initalize metrics servers" and continue adding flags as necessary in the following patches (e.g. AllowedNamespaces).
>
While I understand that you want to see things squashed into one as you see it as Final Implementation, I see it as each commit adding something valuable to reach the Final desired state (probably because it took a month of development and iterations before raising this PR to reach this state).

There is no need for Configuration Options to be exposed before where it currently is. Till this point in commit it's just setting up the framework and tools and Options are hardly of any use. What's the use of Configuration in a system that does nothing?

Also, each commit should handle one thing and one thing only. So when I am initializing server, I don't see the commit also adding options to initialize server differently.

>
>       * Related, there's still a split between when you define `getAllObjectStores()` and when it gets used. It's fine to bring it in with the main definition of the `CephObjectStoreCollector`, but if so put the full implementation of the collector in one patch.

> 
Seems like a mess-up during rearrange and squash. Will rearrange and squash to fix it.

>
>     * Merge the changes of "fixes unit-tests" into the patches that broke the unit tests.
>
We've always kept tests in a separate commit. Let's keep it that way.

jarrpa:
> > * While waiting to bring in option flags until later makes sense in the development process, it does not make sense for the final implementation. Please merge most of the work of "adds flags/options for configuration flexibility" into "initalize metrics servers" and continue adding flags as necessary in the following patches (e.g. AllowedNamespaces).
> 
> While I understand that you want to see things squashed into one as you see it as Final Implementation, I see it as each commit adding something valuable to reach the Final desired state (probably because it took a month of development and iterations before raising this PR to reach this state).
>
> There is no need for Configuration Options to be exposed before where it currently is. Till this point in commit it's just setting up the framework and tools and Options are hardly of any use. What's the use of Configuration in a system that does nothing?
>
> Also, each commit should handle one thing and one thing only. So when I am initializing server, I don't see the commit also adding options to initialize server differently.

In general, options are something that should be a fundamental part of any program. In addition, the understanding of this new binary is not meaningfully served by having this core feature added separately from the initial commit, while I was reviewing it at first I thought it was broken because you were hardcoding so many values. It is an artifact of development, not crucial to the understanding of the software.

When you introduce the option flags you already define seven flags before adding the new one. As such, there was *always* reason to have the flags there in the first place. Speaking of, `AllowedNamespaces` is brought in but nothing is done with it. 

> >   * Related, there's still a split between when you define `getAllObjectStores()` and when it gets used. It's fine to bring it in with the main definition of the `CephObjectStoreCollector`, but if so put the full implementation of the collector in one patch.
> 
> Seems like a mess-up during rearrange and squash. Will rearrange and squash to fix it.
> 
> > * Merge the changes of "fixes unit-tests" into the patches that broke the unit tests.
> 
> We've always kept tests in a separate commit. Let's keep it that way.

I was talking specifically about "updates unit-tests to account for monitoring resources", but I see I was mistaken. You are indeed enhancing the unit test, not fixing it because your changes broke it. :) 

> P.S. You can see each commit in this PR as a PR in itself. In that case, wouldn't the current sequence and way of doing things make sense?

The current sequence generally makes sense, yes! I am happy with the vast majority of them. :) The ones I outlined are the only ones that I think need to be modified or squashed.

metrics/internal/version/version.go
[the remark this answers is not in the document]
jarrpa: I wanted to get rid of the `Spec.Status` field, yes, but the operator binary itself should have some notion of identity as to which version it is. It may still make sense to have a field like `Status.LastReconcilerVersion`. I think since this binary is just another part of the same project, it makes sense to have it match the version of the project/operator.

obnoxxx:
@umangachapagain writes:
> Before I make any more changes I'd like to get more opinions on this PR.
> I'd like to take this as a learning opportunity and understand how a feature should be developed (development process vs final implementation).

Don't get my previous comment wrong: I think you are doing this just great! :+1:  The way you develop the feature initially is first of all personal, and your way was certainly logical and consistent. It is also fine to first put a PR up with the branch in that form because it might give extra context for the reviewer. But this is usually a WIP and early-review state, and the last mile of the software development is to bring the branch / PR in the final mergeable state. And that is usually a different one that the first state when creating the patchset, because for the ultimate state, some other standards can apply. E.g. it is important that all patches compile and ideally pass tests (for the sake of bisectability). Also each patch should be logically complete, do one thing (and only one thing), and do it completely. A little bit on the reasons: The final state of the merged branch should be so that it is as easy as possible for people reading the code to understand the patches. Also it should facilitate finding bugs / regressions in the branch (git bisect). Additionally, we want to make it easy to do backports (this is e.g. why we've been asking to separate the csv-gerneration from other commits).

So usually, I will be asking to split patches up more. But asking to squash fixes for failing tests with the changes in the code. Additional tests can be added in separate commits. In this case though, I think squashing things more is best as opposed to splitting more, since you are introducing new files and mostly new subdirectories. So eventually, it will be okay to introduce whole file and couple of files even in one patch. 

Instead of going into more details, I just wanted to share my general standards and recommendations here, with a certain amount of reasoning. Hope it helps.

In general, it is important to go that last mile of the development process, and in a bigger patchset like this, it is more work. But it's time well spent, and so we should not try to take too big shortcuts. :smiley:

pkg/controller/storagecluster/initialization_reconciler_test.go
```diff
@@ -1,11 +1,16 @@
...
+
 	nbv1 "github.com/noobaa/noobaa-operator/v2/pkg/apis/noobaa/v1alpha1"
 	appsv1 "k8s.io/api/apps/v1"
-	"testing"
 
+<<<<<<< HEAD
 	snapapi "github.com/kubernetes-csi/external-snapshotter/v2/pkg/apis/volumesnapshot/v1beta1"
+=======
+	monitoringv1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
+>>>>>>> a72e7430... updates unit-tests to account for monitoring resources
```
obnoxxx, line 13: This is some accidentially commited merge conflict marker, I guess from a reworking of the branch.
I think we need both parts (and the order reversed to satisfy gofmt).

Also this patch should be squashed with the previous patch, otherwise it fails unit tests.
```suggestion
	monitoringv1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
	snapapi "github.com/kubernetes-csi/external-snapshotter/v2/pkg/apis/volumesnapshot/v1beta1"
```

obnoxxx (changes requested):
Hi @umangachapagain thanks for updating and reworking the PR.

I reviewed it a lilttle more thoroughly from a conceptual point of view and the patches are mostly fine now and pass build and make test.

* The exeption is commit `updates unit-tests to account for monitoring resources` which needs to be fixed (see inline) and needs to be squashed with the previous commit `reconciles metrics exporter along with storagecluster` so that it passes tests.

* One more point is that you only start building the exporter after adding all the files and vendor updates. That's probably OK but even if you put the commit to trigger its build right at the beginning, all commits build and pass unit tests (i checked that). It is basically up to you if you want to squash all the patches up to `updates Build script and Dockerfile` into one big commit to drop the new components. But you can also leave it like this.

@jarrpa had a few more comments on some implementation details, but from a conceptual, patch organization status this is pretty fine now. Thanks!

tools/csv-merger/csv-merger.go
jarrpa: Gotcha, thanks! :+1:

metrics/internal/collectors/ceph-object-store_test.go
jarrpa: All right, I'll take it.

umangachapagain:
> Looks excellent! Thank you for all your hard work. But you still need to fix the failing images test.

https://github.com/openshift/ocs-operator/pull/705#issuecomment-678131325

PR #709, opened 2020-08-18
Design: OSD Topology Spread Constraints

jarrpa:

### DIFF
--- a/docs/design/osd-topology-spread-constraints.md
+++ b/docs/design/osd-topology-spread-constraints.md
@@ -1,0 +1,116 @@
+# FEATURE: OSD Topology Spread Constraints
+
+## Summary
+
+Implement the use of [Topology Spread Constraints] to distribute the Ceph OSD
+Pods across the configured failure domains.
+
+[Topology Spread Constraints](https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints/)
+
+## Motivation
+
+Currently, the primary mechanism for spreading the OSD Pods is to use Pod
+AntiAffinities. In addition, the OSD Pods are grouped into three
+StorageClassDeviceSets (SCDS) in the CephCluster CR, with each SCDS locked to a
+specific failure domain. This presents two problems:
+
+1. AntiAffinity breaks down when there are more Pods than available Nodes. As
+   soon as this happens, the Kubernetes scheduler acts as if the AntiAffinity
+   doesn't exist, which means that an even distribution of Pods can not be
+   guaranteed.
+2. OSD Pods are locked to only three failure domains, even when more than three
+   exist.
+
+### Goals
+
+Provide an even-as-possible distribution of OSD Pods across all available
+failure domains based on Pod count as opposed to the default of resource
+availability.
+
+### Non-Goals
+
+Always maintain an even distribution of OSD Pods. OSDs can scale at a rate that
+does not match the number of Nodes in a given failure domain, as such there
+will sometimes be uneven distribution of OSD Pods across Nodes in a given
+failure domain.
+
+## Proposal
+
+Implement the use of Topology Spread Constraints for Ceph OSD Pods. Include the
+ability to specify them in the `StorageCluster.StorageDeviceSet` spec so they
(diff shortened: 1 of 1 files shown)


jarrpa:
@travisn @obnoxxx @umangachapagain PTAL

cc: @kesavanvt

docs/design/osd-topology-spread-constraints.md
```diff
@@ -0,0 +1,104 @@
...
+Since we will no longer need to explicitly distinguish between exact failure
+domains, the constraints themselves would be identical across all SCDSes. This
+means that all Pods will be schedulable across all failure domains, regardless
+of which SCDS it belongs to.
+
+For upgrade scenarios, existing OSD Pods would still remain with the old
+Placement configurations. However, because all OSD Pods have appropriate app
+labels to identify them, they would still influence any new OSD Pods created
+with Topology Spread Constraints, allowing the new OSD Pods to be scheduled
+in a more even pattern across failure domains.
```
travisn, line 81: How about adding some specifics about the placement options that will be applied consistently across the three device sets:
- Topology spread constraints for OSD Prepare jobs across the desired failure domain (e.g. `zone` if 3+ zones are available, otherwise `host`)
- Topology spread constraints for OSD daemons across hosts (OSDs will already be tied to a failure domain based on their PV affinity based on where the OSD prepare job ran)
- Node affinity to the nodes selected by the OCS admin

Anything else?

docs/design/osd-topology-spread-constraints.md
```diff
@@ -0,0 +1,104 @@
...
+
+**Phase 2**
+
+The currently proposed solution for the upgrade scenario is to enhance
```
travisn, line 85: The only downside of this phase never being implemented is that OCS would indefinitely have 3 device sets instead of 1, basically just an implementation detail. Not sure the complexity of the upgrade will be worth the change in rook since everything is already working as required for phase 1.

jarrpa: And I don't want OCS to indefinitely generate three device sets. :) It will become a vestigial implementation detail that we will have to explain until the end of time (or a v2 API, whichever comes first). Also, I vaguely remember looking at the Rook-Ceph code and thinking it wouldn't be all that complex, and even that it would be a cleaner way to handle OSD detection overall.

docs/design/osd-topology-spread-constraints.md
jarrpa: Done!

docs/design/osd-topology-spread-constraints.md
obnoxxx: The implicit downside of OCS always having 3 device sets is that OCS can not use more than 3 failure domains. E.g. if you have more than 3 zones (AZs), then still only three would get used. This is s serious limitation that we want to get rid of. We *might* be able to implement a more complicated algorithm in ocs-operator, but the approach to do this in rook with one SCDS seems more robust and correct to me.

It is also important to remember that we originally introduced the splitting of the SDSs into 3 SCDSs only to work around for the absence of the TSCs, hoping/planning to get rid of this splitting again once TSCs are available.

docs/design/osd-topology-spread-constraints.md
```diff
@@ -0,0 +1,115 @@
...
+Rook-Ceph to detect existing OSD Pods and determine if they are valid for a
+given SCDS. Once this is done, we will be able to submit a final patch to
+change the CephCluster generation from three SCDSes to one.
+
+### Test Plan
+
+These changes are under the hood changes, and should require no additional
+e2e testing in our CI.
```
obnoxxx, line 104: At least for phase2, this depends on whether the CI is looking at the creation of multiple SCDSs in rook-ceph.  Also, the CI could and probably should create many OSDs and check the spreading across nodes.

jarrpa: QE should already be testing OSD creation at scale, that is not anything new that is enabled by this feature.

obnoxxx (approved):
/lgtm

PR #710, opened 2020-08-19
enhances containerized build

umangachapagain:
Uses multi-stage build to minimize image size.
ubi8/go-toolset is used as builder image to build ocs-operator
and copy it into final image with ubi8/ubi-minimal as base.
Final image will be around 170MB in comparison to previous 2GB+.

### DIFF
--- a/build/Dockerfile.build
+++ b/build/Dockerfile.build
@@ -1,10 +1,9 @@
-FROM registry.access.redhat.com/ubi8/ubi-minimal as builder
+FROM registry.access.redhat.com/ubi8/go-toolset AS builder
 
-RUN microdnf install -y make golang; microdnf -y clean all
-
-FROM builder
-
-RUN mkdir -p /output
 COPY . /ocs-operator
 WORKDIR /ocs-operator
-RUN make build-go OUTDIR_BIN=/output
+RUN make build-go
+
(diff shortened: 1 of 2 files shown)


build/Dockerfile.build
```diff
@@ -1,10 +1,9 @@
...
+FROM registry.access.redhat.com/ubi8/go-toolset AS builder
 
-RUN microdnf install -y make golang; microdnf -y clean all
-
-FROM builder
-
-RUN mkdir -p /output
 COPY . /ocs-operator
 WORKDIR /ocs-operator
-RUN make build-go OUTDIR_BIN=/output
```
jarrpa, line 10: How do we get the binary out from the container?

umangachapagain: Why do we need to get it out? It won't be of any use.

If you want it, you'll have to run the container and copy it.
I tried adding volume mount while building, it doesn't work.

jarrpa: The entire point of the containerized build is to get the binary to put it into the final operator container image. It doesn't do anything to build a binary in a container and then do nothing with it. So the binary built in the `build/Dockerfile.build` container and use it as the main binary in `build/Dockerfile`.

PR #714, opened 2020-08-20
Whitelist instead of Blacklist to hide internal objects

umangachapagain:
adding API whitelist to populate internal-objects.
This allows us to hide any API that we do not want to expose
to user without having to append individual APIs to the
internal objects list. Instead we add APIs that we want to
expose and the rest is automatically hidden.

### DIFF
--- a/tools/csv-merger/csv-merger.go
+++ b/tools/csv-merger/csv-merger.go
@@ -55,6 +55,13 @@ var (
 	inputManifestsDir = flag.String("manifests-directory", "", "The directory containing the extra manifests to be included in the registry bundle")
 
 	outputDir = flag.String("olm-bundle-directory", "", "The directory to output the unified CSV and CRDs to")
+
+	// List of APIs which should be exposed in Console
+	exposedAPIs = []string{
+		"storageclusters.ocs.openshift.io",
+		"backingstores.noobaa.io",
+		"bucketclasses.noobaa.io",
+	}
 )
 
 type templateData struct {
@@ -385,6 +392,16 @@ func marshallObject(obj interface{}, writer io.Writer, modifyUnstructuredFunc fu
 	return nil
 }
 
+// Checks whether a string is contained within a slice
+func contains(slice []string, s string) bool {
+	for _, item := range slice {
+		if item == s {
+			return true
+		}
+	}
+	return false
+}
+
 func generateUnifiedCSV() *csvv1.ClusterServiceVersion {
 
 	csvs := []string{
@@ -404,6 +421,8 @@ func generateUnifiedCSV() *csvv1.ClusterServiceVersion {
 			MediaType: "image/svg+xml",
 		},
 	}
+
+	ocsCSV.Annotations["operators.operatorframework.io/internal-objects"] = ""
 
 	templateStrategySpec := &csvv1.StrategyDetailsDeployment{
 		ClusterPermissions: []csvv1.StrategyDeploymentPermissions{},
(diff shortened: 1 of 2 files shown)


tools/csv-merger/csv-merger.go
```diff
@@ -405,6 +415,9 @@ func generateUnifiedCSV() *csvv1.ClusterServiceVersion {
 		},
 	}
 
+	ocsCSV.Annotations["operators.operatorframework.io/internal-objects"] = ""
+	nonInternalAPIs := []string{"storageclusters.ocs.openshift.io", "backingstores.noobaa.io", "bucketclasses.noobaa.io"}
```
jarrpa, line 419: Rename this to `exposedAPIs` and make it a global var or const with each string in its own line. This will make it easier to see change sin the future.

tools/csv-merger/csv-merger.go
```diff
@@ -610,7 +630,7 @@ The NooBaa operator deploys and manages the [NooBaa][2] Multi-Cloud Gateway on O
 	ocsCSV.Annotations["support"] = "Red Hat"
 	ocsCSV.Annotations["capabilities"] = "Full Lifecycle"
 	ocsCSV.Annotations["categories"] = "Storage"
-	ocsCSV.Annotations["operators.operatorframework.io/internal-objects"] = `["cephclusters.ceph.rook.io", "cephblockpools.ceph.rook.io", "cephobjectstores.ceph.rook.io", "cephobjectstoreusers.ceph.rook.io", "cephnfses.ceph.rook.io","cephclients.ceph.rook.io", "noobaas.noobaa.io", "objectbuckets.objectbucket.io","objectbucketclaims.objectbucket.io","ocsinitializations.ocs.openshift.io", "storageclusterinitializations.ocs.openshift.io", "cephfilesystems.ceph.rook.io"]`
+	// ocsCSV.Annotations["operators.operatorframework.io/internal-objects"] = `["cephclusters.ceph.rook.io", "cephblockpools.ceph.rook.io", "cephobjectstores.ceph.rook.io", "cephobjectstoreusers.ceph.rook.io", "cephnfses.ceph.rook.io","cephclients.ceph.rook.io", "noobaas.noobaa.io", "objectbuckets.objectbucket.io","objectbucketclaims.objectbucket.io","ocsinitializations.ocs.openshift.io", "storageclusterinitializations.ocs.openshift.io", "cephfilesystems.ceph.rook.io"]`
```
jarrpa, line 633: Remove this comment.

tools/csv-merger/csv-merger.go
```diff
@@ -437,6 +450,13 @@ func generateUnifiedCSV() *csvv1.ClusterServiceVersion {
 			}
 		}
 	}
+	// whitelisting APIs
+	for _, definition := range ocsCSV.Spec.CustomResourceDefinitions.Owned {
+		if !contains(nonInternalAPIs, definition.Name) {
+			ocsCSV.Annotations["operators.operatorframework.io/internal-objects"] = ocsCSV.Annotations["operators.operatorframework.io/internal-objects"] + "," + "\"" + definition.Name + "\""
+		}
+	}
+	ocsCSV.Annotations["operators.operatorframework.io/internal-objects"] = "[" + strings.Replace(ocsCSV.Annotations["operators.operatorframework.io/internal-objects"], ",", "", 1) + "]"
```
jarrpa, line 459: Instead of doing a replace, use the index from the range in the for loop above to see if you're at index zero, and if so add `[` instead of `,`. Then here you can just add the final `]`

PR #638, opened 2020-07-19
Ceph object store for external cluster

aruniiird:
CephObjectStore creation in external cluster mode according to the latest rook-ceph changes

PR #716, opened 2020-08-20
Enable Monitoring for external mode

anmolsachan:
This PR is to enable monitoring on the external mode in OCS. 
The changes have been done in keeping mind that monitoring will be mandatory for OCS as the UI depends on it.

Depends on #638  for some functions and shall be merged post that.

Please review commit:  Enabled monitoring for external mode

### DIFF
--- a/deploy/olm-catalog/ocs-operator/manifests/ocs-operator.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/ocs-operator/manifests/ocs-operator.clusterserviceversion.yaml
@@ -697,7 +697,7 @@ metadata:
       cjoge30iLmZvcm1hdChlcnIpKQo=
     external.features.ocs.openshift.io/validation: '{"secrets":["rook-ceph-operator-creds",
       "rook-csi-rbd-node", "rook-csi-rbd-provisioner"], "configMaps": ["rook-ceph-mon-endpoints",
-      "rook-ceph-mon"], "storageClasses": ["ceph-rbd"]}'
+      "rook-ceph-mon"], "storageClasses": ["ceph-rbd"], "cephClusters": ["monitoring-endpoint"]}'
     features.ocs.openshift.io/enabled: '["external"]'
     olm.skipRange: '>=0.0.1 <4.6.0'
     operatorframework.io/cluster-monitoring: "true"
--- a/pkg/controller/storagecluster/cephcluster.go
+++ b/pkg/controller/storagecluster/cephcluster.go
@@ -61,7 +61,7 @@ func (r *ReconcileStorageCluster) ensureCephCluster(sc *ocsv1.StorageCluster, re
 	var cephCluster *cephv1.CephCluster
 	// Define a new CephCluster object
 	if sc.Spec.ExternalStorage.Enable {
-		cephCluster = newExternalCephCluster(sc, r.cephImage)
+		cephCluster = newExternalCephCluster(sc, r.cephImage, r.monitoringIP)
 	} else {
 		cephCluster = newCephCluster(sc, r.cephImage, r.nodeCount, reqLogger)
 	}
@@ -261,7 +261,7 @@ func validateMultusSelectors(selectors map[string]string) error {
 	return nil
 }
 
-func newExternalCephCluster(sc *ocsv1.StorageCluster, cephImage string) *cephv1.CephCluster {
+func newExternalCephCluster(sc *ocsv1.StorageCluster, cephImage string, monitoringIP string) *cephv1.CephCluster {
 	labels := map[string]string{
 		"app": sc.Name,
 	}
@@ -282,8 +282,14 @@ func newExternalCephCluster(sc *ocsv1.StorageCluster, cephImage string) *cephv1.
 				ManagePodBudgets:               false,
 				ManageMachineDisruptionBudgets: false,
 			},
+			Monitoring: cephv1.MonitoringSpec{
+				Enabled:              true,
+				RulesNamespace:       sc.Namespace,
+				ExternalMgrEndpoints: []corev1.EndpointAddress{{IP: monitoringIP}},
+			},
 		},
 	}
+
 	return externalCephCluster
 }
 
--- a/pkg/controller/storagecluster/external_resources.go
+++ b/pkg/controller/storagecluster/external_resources.go
@@ -7,6 +7,7 @@ import (
 	"net"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/go-logr/logr"
 	ocsv1 "github.com/openshift/ocs-operator/pkg/apis/ocs/v1"
@@ -195,6 +196,30 @@ func (r *ReconcileStorageCluster) createExternalStorageClusterResources(instance
 		}
 		objectKey := types.NamespacedName{Name: d.Name, Namespace: instance.Namespace}
 		switch d.Kind {
+		case "CephCluster":
+			monitoringIP, ok := d.Data["MonitoringEndpoint"]
+			if !ok {
+				err := fmt.Errorf(
+					"Monitoring Endpoint not present in the external cluster secret %s",
+					externalClusterDetailsSecret)
+				reqLogger.Error(err, "Failed to get Monitoring IP.")
+				return err
+			}
+			monitoringPort, ok := d.Data["MonitoringPort"]
+			if !ok {
+				err := fmt.Errorf(
+					"Monitoring Port not present in the external cluster secret %s",
+					externalClusterDetailsSecret)
+				reqLogger.Error(err, "Failed to get Monitoring Port.")
+				return err
+			}
+			err := validateMonitoringEndpoint(monitoringIP, monitoringPort, reqLogger)
+			if err != nil {
+				reqLogger.Error(err, "Monitoring validation failed")
+				return err
+			}
+			reqLogger.Info("Monitoring Information found. Monitoring will be enabled on the external cluster")
+			r.monitoringIP = monitoringIP
 		case "ConfigMap":
 			cm := &corev1.ConfigMap{
 				ObjectMeta: objectMeta,
@@ -309,6 +334,23 @@ func (r *ReconcileStorageCluster) createExternalStorageClusterSecret(sec *corev1
 			reqLogger.Error(err, "unable the get the secret")
 			return err
 		}
+	}
+	return nil
+}
+
+// To check if endpoint is a VALID ip and is REACHABLE or not
+func validateMonitoringEndpoint(monitoringIP string, monitoringPort string, reqLogger logr.Logger) error {
+	_, err := net.LookupIP(monitoringIP)
+	if err != nil {
+		reqLogger.Error(err, "Monitoring endpoint is not a valid IPv4 IP")
+		return err
+	}
+	endpoint := net.JoinHostPort(monitoringIP, monitoringPort)
+	con, err := net.DialTimeout("tcp", endpoint, 5*time.Second)
+	defer con.Close()
+	if err != nil {
+		reqLogger.Error(err, fmt.Sprintf("Monitoring Endpoint (%s) is not reachable", endpoint))
+		return err
 	}
 	return nil
 }
--- a/tools/csv-merger/csv-merger.go
+++ b/tools/csv-merger/csv-merger.go
@@ -651,7 +651,7 @@ The NooBaa operator deploys and manages the [NooBaa][2] Multi-Cloud Gateway on O
 	ocsCSV.Annotations["features.ocs.openshift.io/enabled"] = `["external"]`
 	// Used by UI to validate user uploaded metdata
 	// Metadata is used to connect to an external cluster
-	ocsCSV.Annotations["external.features.ocs.openshift.io/validation"] = `{"secrets":["rook-ceph-operator-creds", "rook-csi-rbd-node", "rook-csi-rbd-provisioner"], "configMaps": ["rook-ceph-mon-endpoints", "rook-ceph-mon"], "storageClasses": ["ceph-rbd"]}`
+	ocsCSV.Annotations["external.features.ocs.openshift.io/validation"] = `{"secrets":["rook-ceph-operator-creds", "rook-csi-rbd-node", "rook-csi-rbd-provisioner"], "configMaps": ["rook-ceph-mon-endpoints", "rook-ceph-mon"], "storageClasses": ["ceph-rbd"], "cephClusters": ["monitoring-endpoint"]}`
 	// Injecting the RHCS exporter script present in Rook CSV
 	ocsCSV.Annotations["external.features.ocs.openshift.io/export-script"] = rookCSV.GetAnnotations()["externalClusterScript"]
 	if *timestamp == "true" {
--- a/pkg/controller/storagecluster/storagecluster_controller.go
+++ b/pkg/controller/storagecluster/storagecluster_controller.go
@@ -143,6 +143,7 @@ type ReconcileStorageCluster struct {
 	conditions      []conditionsv1.Condition
 	phase           string
 	cephImage       string
+	monitoringIP    string
 	noobaaDBImage   string
 	noobaaCoreImage string
 	nodeCount       int


anmolsachan:
@aruniiird @leseb Please review

deploy/olm-catalog/ocs-operator/manifests/ocs-operator.clusterserviceversion.yaml
```diff
@@ -697,7 +697,8 @@ metadata:
       cjoge30iLmZvcm1hdChlcnIpKQo=
     external.features.ocs.openshift.io/validation: '{"secrets":["rook-ceph-operator-creds",
       "rook-csi-rbd-node", "rook-csi-rbd-provisioner"], "configMaps": ["rook-ceph-mon-endpoints",
-      "rook-ceph-mon"], "storageClasses": ["ceph-rbd"]}'
+      "rook-ceph-mon"], "storageClasses": ["ceph-rbd"], "CephCluster": ["MonitoringEndpoint", 
+      "MonitoringPort"]}'
```
umangachapagain, line 701: Where is this `"MonitoringPort"` coming from? I can't see it in CSV merger.

pkg/controller/storagecluster/cephcluster.go
```diff
@@ -229,7 +233,8 @@ func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, nodeCount int, r
 	return cephCluster
 }
 
-func newExternalCephCluster(sc *ocsv1.StorageCluster, cephImage string) *cephv1.CephCluster {
+func newExternalCephCluster(sc *ocsv1.StorageCluster, cephImage string,
```
umangachapagain, line 236: Why did you add `reqLogger logr.Logger` if you are not using it?

pkg/controller/storagecluster/cephcluster.go
```diff
@@ -380,3 +391,30 @@ func newCephDaemonResources(custom map[string]corev1.ResourceRequirements) map[s
...
+	}
+	for _, d := range data {
+		switch d.Kind {
+		case "CephCluster":
+			monitoringIP, ok := d.Data["MonitoringEndpoint"]
+			if !ok {
+				err := fmt.Errorf(
+					"Monitoring Endpoint not present in the external cluster secret %s",
+					externalClusterDetailsSecret)
+				reqLogger.Error(err, err.Error())
```
umangachapagain, line 410:
```suggestion
				reqLogger.Error(err, "failed to get Monitoring Endpoint information")
```

pkg/controller/storagecluster/external_resources.go
```diff
@@ -171,6 +172,28 @@ func (r *ReconcileStorageCluster) createExternalStorageClusterResources(scs []*s
...
+					"Monitoring Port not present in the external cluster secret %s",
+					externalClusterDetailsSecret)
+				reqLogger.Error(err, err.Error())
+				return err
+			}
+			err := validateMonitoringEndpoint(monitoringIP, monitoringPort, reqLogger)
+			if err != nil {
+				reqLogger.Error(err, "Monitoring validation failed.")
+				return err
+			}
```
umangachapagain, line 196: This whole thing looks like a duplicate of `getExternalMonitoringInfo()`. Can we move everything there (including validation) there? I don't see anything being created with IP or Ports here.

pkg/controller/storagecluster/external_resources.go
```diff
@@ -271,3 +294,28 @@ func (r *ReconcileStorageCluster) createExternalStorageClusterSecret(sec *corev1
...
+	endpoint := monitoringIP + ":" + monitoringPort
+	err = checkEndpointReachable(endpoint, 5*time.Second, reqLogger)
+	if err != nil {
+		reqLogger.Error(err, fmt.Sprintf("Monitoring Endpoint (%s) is not reachable.", endpoint))
+		return err
+	}
+	return nil
+}
+
+func checkEndpointReachable(endpoint string, timeout time.Duration, reqLogger logr.Logger) error {
```
umangachapagain, line 314: Is this necessary?

I think we can skip this test. If IP is valid, pass it and create service and servicemonitor. If it is not reachable Prometheus will mark the target accordingly. We can probably build an alert about it then?

The problem with this approach is it's not very reliable (IMO). A timeout of 5 sec and a oneof operation doesn't reflect the real state of connection. What if it becomes unreachable after 5s? or it's unreachable for 10s and is reachable after that?

tools/csv-merger/csv-merger.go
```diff
@@ -594,7 +594,7 @@ The NooBaa operator deploys and manages the [NooBaa][2] Multi-Cloud Gateway on O
 	ocsCSV.Annotations["features.ocs.openshift.io/enabled"] = `["external"]`
 	// Used by UI to validate user uploaded metdata
 	// Metadata is used to connect to an external cluster
-	ocsCSV.Annotations["external.features.ocs.openshift.io/validation"] = `{"secrets":["rook-ceph-operator-creds", "rook-csi-rbd-node", "rook-csi-rbd-provisioner"], "configMaps": ["rook-ceph-mon-endpoints", "rook-ceph-mon"], "storageClasses": ["ceph-rbd"]}`
+	ocsCSV.Annotations["external.features.ocs.openshift.io/validation"] = `{"secrets":["rook-ceph-operator-creds", "rook-csi-rbd-node", "rook-csi-rbd-provisioner"], "configMaps": ["rook-ceph-mon-endpoints", "rook-ceph-mon"], "storageClasses": ["ceph-rbd"], "CephCluster": ["monitoring-endpoint"]}`
```
umangachapagain, line 597: I don't see `Monitoring Port` here but it's on CSV?

pkg/controller/storagecluster/external_resources.go
```diff
@@ -182,3 +294,28 @@ func (r *ReconcileStorageCluster) createExternalStorageClusterSecret(sec *corev1
...
 }
+
+// To check if endpoint is a VALID ip and is REACHABLE or not
+func validateMonitoringEndpoint(monitoringIP string, monitoringPort string, reqLogger logr.Logger) error {
+	_, err := net.LookupIP(monitoringIP)
+	if err != nil {
+		reqLogger.Error(err, "Monitoring endpoint is not a valid IPv4 Ip.")
+		return err
+	}
+	endpoint := monitoringIP + ":" + monitoringPort
```
aruniiird, line 305: Since we are in golang and already importing `net` package, use `net.JoinHostPort()` function (which will take care of IPV6 as well in the future)
```suggestion
	endpoint := net.JoinHostPort(monitoringIP, monitoringPort)
```
PS: again recommending that we use `MonitoringEndpoint` as single arg with `<IP>:<port>` value...

pkg/controller/storagecluster/cephcluster.go
aruniiird: Yes. Usually if you don't have any error to report (or info to print), we don't require `reqLogger`.

pkg/controller/storagecluster/external_resources.go
aruniiird, line 196: If you could combine the functionalities into a single one as @umangachapagain has suggested that will be much appreciated...

```suggestion
		case "CephCluster":
			monitoringIP, okMonIP := d.Data["MonitoringEndpoint"]
			monitoringPort, okMonPort := d.Data["MonitoringPort"]
			// if any of the above is missing return an error
			if !okMonIP || !okMonPort {
				err := fmt.Errorf(
					"Both 'MonitoringEndpoint' and 'MonitoringPort' values should be present in secret %s",
					externalClusterDetailsSecret)
				reqLogger.Error(err, err.Error())
				return err
			}
			err := validateMonitoringEndpoint(monitoringIP, monitoringPort, reqLogger)
			if err != nil {
				reqLogger.Error(err, "Monitoring validation failed.")
				return err
			}
```

deploy/olm-catalog/ocs-operator/manifests/ocs-operator.clusterserviceversion.yaml
anmolsachan: Missed it. I will update

pkg/controller/storagecluster/external_resources.go
anmolsachan: I will update it, The port is required to test the endpoint reachability, though we don't pass it in the Monitoring Spec. Rook already has the default port set. It is a safety net for the case if user enters ports different from(very rare) rook and tell them they need to run it on the default RHCS port.

pkg/controller/storagecluster/external_resources.go
anmolsachan: This for the validation as to when OCP is not able to reach ceph-mgr cluster, even if the IP is valid. We stop the control flow here. Thus Cephcluster and other objects are not created on the first hand. User can then just delete the storage cluster quickly, check the connectivity and then again try
If later on the connection goes bad its an infrastructure problem and we go the Prometheus way. What do you say ?@umangachapagain

tools/csv-merger/csv-merger.go
anmolsachan: Yes, its not required. It was an error in the CSV. I will update it there. Only monitoring-endpoint is required.

pkg/controller/storagecluster/external_resources.go
anmolsachan: @aruniiird Its how the rook script provides it, in separate arguments. If we need to change we will have to make a change in rook first.

pkg/controller/storagecluster/external_resources.go
anmolsachan: I will refactor the code. Thanks.

pkg/controller/storagecluster/external_resources.go
umangachapagain: Isn't this done when CephCluster is getting Connected via Rook? 
If CephCluster is Connected we know OCP can reach it. Do we need same validations on other layers?

@leseb any thoughts on this?

leseb: This check seems a bit redundant to what Rook does internally and somehow "weak" as well since it's just a dial on the address **but** I like the fact we fail early so I'd keep it around. Also, 5s is more than enough, 5s network-wise feels like an eternity, so if after 5s we don't get any reply it's fair to assume the endpoint is not reachable.

pkg/controller/storagecluster/cephcluster.go
```diff
@@ -380,3 +392,68 @@ func newCephDaemonResources(custom map[string]corev1.ResourceRequirements) map[s
 
 	return resources
 }
+
+func (r *ReconcileStorageCluster) getExternalMonitoringInfo(
+	sc *ocsv1.StorageCluster, reqLogger logr.Logger) (string, error) {
+	data, err := r.retrieveExternalSecretData(sc, reqLogger)
```
leseb, line 398: Shouldn't this be in the same code that already fetches the external secret?

pkg/controller/storagecluster/cephobjectstores.go
```diff
@@ -26,42 +26,47 @@ func (r *ReconcileStorageCluster) ensureCephObjectStores(instance *ocsv1.Storage
...
+func (r *ReconcileStorageCluster) createCephObjectStores(cephObjectStores []*cephv1.CephObjectStore, reqLogger logr.Logger) error {
 	for _, cephObjectStore := range cephObjectStores {
 		existing := cephv1.CephObjectStore{}
 		err := r.client.Get(context.TODO(), types.NamespacedName{Name: cephObjectStore.Name, Namespace: cephObjectStore.Namespace}, &existing)
 		switch {
 		case err == nil:
 			if existing.DeletionTimestamp != nil {
-				reqLogger.Info(fmt.Sprintf("Unable to restore init object because %s is marked for deletion", existing.Name))
-				return fmt.Errorf("failed to restore initialization object %s because it is marked for deletion", existing.Name)
+				err := fmt.Errorf("failed to restore initialization object %s because it is marked for deletion",
```
leseb, line 48: Is this really a failure if the CR is being deleted?

pkg/controller/storagecluster/cephobjectstores.go
```diff
@@ -26,42 +26,47 @@ func (r *ReconcileStorageCluster) ensureCephObjectStores(instance *ocsv1.Storage
...
+					existing.Name)
+				reqLogger.Error(err,
+					fmt.Sprintf("Unable to restore init object because %s is marked for deletion", existing.Name))
```
leseb, line 51: Which restore?

leseb (changes requested):
I'd be nice to see some unit tests added for this scenario. Thanks.

pkg/controller/storagecluster/cephcluster.go
```diff
@@ -380,3 +392,68 @@ func newCephDaemonResources(custom map[string]corev1.ResourceRequirements) map[s
...
+	endpoint := net.JoinHostPort(monitoringIP, monitoringPort)
+	err = checkEndpointReachable(endpoint, 5*time.Second, reqLogger)
+	if err != nil {
+		reqLogger.Error(err, fmt.Sprintf("Monitoring Endpoint (%s) is not reachable.", endpoint))
+		return err
+	}
+	return nil
+}
+
+func checkEndpointReachable(endpoint string, timeout time.Duration, reqLogger logr.Logger) error {
```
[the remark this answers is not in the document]
anmolsachan: I wanted to reuse it for some other work as well. Will delete at its won't be required.

pkg/controller/storagecluster/cephcluster.go
anmolsachan: if I put it there then I will have to save monitoring IP in some object for the later use when external cephcluster object is created. That is why I am calling it again here.

pkg/controller/storagecluster/cephobjectstores.go
anmolsachan: this is part of https://github.com/openshift/ocs-operator/pull/638

pkg/controller/storagecluster/cephobjectstores.go
anmolsachan: this part of https://github.com/openshift/ocs-operator/pull/638

pkg/controller/storagecluster/cephcluster.go
```diff
@@ -47,7 +49,11 @@ func (r *ReconcileStorageCluster) ensureCephCluster(sc *ocsv1.StorageCluster, re
 	var cephCluster *cephv1.CephCluster
 	// Define a new CephCluster object
 	if sc.Spec.ExternalStorage.Enable {
-		cephCluster = newExternalCephCluster(sc, r.cephImage)
+		monitoringIP, err := r.getExternalMonitoringInfo(sc, reqLogger)
+		if err != nil {
+			return err
+		}
+		cephCluster = newExternalCephCluster(sc, r.cephImage, monitoringIP)
```
jarrpa, line 56: I see no need to alter the function signature for this. Call `getExternalMonitoringInfo()` from `newExternalCephCluster()`.

pkg/controller/storagecluster/cephcluster.go
jarrpa: Makes sense... I also don't entirely like it, but it's sensible enough.

tools/csv-merger/csv-merger.go
jarrpa, line 597:
```suggestion
	ocsCSV.Annotations["external.features.ocs.openshift.io/validation"] = `{"secrets":["rook-ceph-operator-creds", "rook-csi-rbd-node", "rook-csi-rbd-provisioner"], "configMaps": ["rook-ceph-mon-endpoints", "rook-ceph-mon"], "storageClasses": ["ceph-rbd"], "cephCluster": ["monitoring-endpoint"]}`
```

pkg/controller/storagecluster/cephcluster.go
anmolsachan: @jarrpa The reason I did so was that the fuction _newExternalCephCluster_ does not have the _**receiver type** of r *ReconcileStorageCluster_ . Would you rather have that in place of monitoringIP to call getExternalMonitoringInfo() from inside the newExternalCephCluster() ?

tools/csv-merger/csv-merger.go
anmolsachan: @jarrpa The reason for keeping it in caps was : https://github.com/rook/rook/blob/85d72f70b06f7fb480229994cf8c139616aa088a/cluster/examples/kubernetes/ceph/create-external-cluster-resources.py#L470 .

pkg/controller/storagecluster/cephcluster.go
leseb: Why not just saving it in a package variable?

pkg/controller/storagecluster/cephcluster.go
jarrpa: Ah... shows what I know. :stuck_out_tongue_winking_eye: Okay, I think I can live with this. Let's not break the pattern of the `new*()` functions for this.

pkg/controller/storagecluster/cephcluster.go
jarrpa: Hmm... yes, I think this would make sense. Since this function performs a `Get()`, there's a chance that the information/generation could change between the two instances of this function. Rather than a package var though, maybe a new field in the `ReconcileStorageCluster` struct.

tools/csv-merger/csv-merger.go
jarrpa: Shouldn't matter. All the other entries also have an initial uppercase letter as their official Kind names, yet we still have them lowercase. In fact, is should be plural, so `cephClusters`.

anmolsachan: Okay sure, I will change it :)

anmolsachan: Ack, I will update. Thanks

jarrpa:
Looks like we need to address some unit test failures.

/lgtm cancel

anmolsachan:
@jarrpa I changed the PR a little as per your and Sebastien comment. I have put it in a separate commit https://github.com/openshift/ocs-operator/pull/716/commits/a6b977171dea755e621e4522c65e949e19a241e3 . Would this be better? Otherwise, I will remove the commit and fix the tests. Thanks

tools/csv-merger/csv-merger.go
obnoxxx: done

obnoxxx (approved):
It seems all requests have been addressed.

/lgtm

PR #718, opened 2020-08-21
StorageCluster: add uninstall mode

raghavendra-talur:
Rook and Noobaa are now both defaulting to wait for removal of existing
PVCs and OBCs before deleting their operands. In cases where the admin
wants to enforce a forced cleanup, this label on the StorageCluster
should be set.

uninstall-mode.ocs.openshift.io="yes-really-uninstall-even-if-in-use"

Co-authored-by: Nitin Goyal <nigoyal@redhat.com>

### DIFF
--- a/pkg/controller/storagecluster/reconcile.go
+++ b/pkg/controller/storagecluster/reconcile.go
@@ -49,8 +49,8 @@ osd_memory_target_cgroup_limit_ratio = 0.5
 	monCountOverrideEnvVar = "MON_COUNT_OVERRIDE"
 	// EBS represents AWS EBS provisioner for StorageClass
 	EBS StorageClassProvisionerType = "kubernetes.io/aws-ebs"
-	// CleanupPolicyAnnotation defines the cleanup policy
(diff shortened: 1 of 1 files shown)


raghavendra-talur:
> I only see one commit, is there supposed to be more? Also, why not just extend the cleanup policy?

We were confused about the Noobaa field and its use. As you can see with the updated commits, the cleanup policy in Noobaa spec actually refers to the uninstall behavior, specifically, whether it is graceful or forced. 

This PR introduces a new label in the StorageCluster which determines the uninstall mode for both Rook and Noobaa. As of now, we are using it only to set the uninstall mode for Noobaa. The commit for Rook will come in a later PR after making relevant changes in Rook.

Added comments to make it clean in the code too, let us know if it not clear.

cc: @iamniting

raghavendra-talur:
@jarrpa Now, with the new PR #730 and with the updates in this PR, we have addressed all the review suggestions.

Changes:
1. Moved from labels to annotations
2. Clubbed the annotations under uninstall.ocs.openshift.io

cc: @iamniting

PR #721, opened 2020-08-24
Change the default max number of noobaa endpoint

nb-ohad:


nb-ohad:
@umangachapagain `gen-latest-csv` does not produce any diff on my machine

PR #675, opened 2020-08-05
Reconcile CephObjectStores, CephObjectStoreUsers, and CephFilesystems

priyanka19-98:
Modified the handling of functions `ensureCephObjectStore()` , `ensureCephObjectStoreUsers()`, `ensureCephFileSystems()` by removing check for each function related Status and removing boolean from StorageClusterStatus struct.

Signed off by: Priyanka Jiandani <pjiandan@redhat.com>

PR #722, opened 2020-08-24
Revert PR #675

jarrpa:
This reverts #675 due to some concerns over potential race conditions with deletion of some of the resources affected.


obnoxxx (approved):
We don't really need the whitespace change here, but then.... it doesn't harm.

/lgtm

PR #727, opened 2020-08-25
Supporting OSD Topology Spread Constraints

kesavanvt:
**Description** 
This PR is about supporting Topology Spread Constraints (**Phase 1 changes**) that will allow OSDs to evenly distribute across the required topology domain.

Design : https://github.com/openshift/ocs-operator/pull/709

Signed-off-by: Kesavan Vellalore Thiruvenkadasamy kvellalo@redhat.com

### DIFF
--- a/pkg/controller/defaults/placements.go
+++ b/pkg/controller/defaults/placements.go
@@ -90,6 +90,36 @@ var (
 			},
 		},
 
+		"osd-prepare": rook.Placement{
+			Tolerations: []corev1.Toleration{
+				corev1.Toleration{
+					Key:      NodeTolerationKey,
+					Operator: corev1.TolerationOpEqual,
+					Value:    "true",
+					Effect:   corev1.TaintEffectNoSchedule,
+				},
+			},
+			PodAntiAffinity: &corev1.PodAntiAffinity{
+				PreferredDuringSchedulingIgnoredDuringExecution: []corev1.WeightedPodAffinityTerm{
+					corev1.WeightedPodAffinityTerm{
+						Weight: 100,
+						PodAffinityTerm: corev1.PodAffinityTerm{
+							LabelSelector: &metav1.LabelSelector{
+								MatchExpressions: []metav1.LabelSelectorRequirement{
+									metav1.LabelSelectorRequirement{
+										Key:      appLabelSelectorKey,
+										Operator: metav1.LabelSelectorOpIn,
+										Values:   []string{"rook-ceph-osd-prepare"},
+									},
+								},
+							},
+							TopologyKey: corev1.LabelHostname,
+						},
+					},
+				},
+			},
+		},
+
 		"rgw": rook.Placement{
 			Tolerations: []corev1.Toleration{
 				corev1.Toleration{
--- a/pkg/controller/defaults/placements.go
+++ b/pkg/controller/defaults/placements.go
@@ -120,6 +120,30 @@ var (
 			},
 		},
 
+		"osd-tsc": rook.Placement{
+			Tolerations: []corev1.Toleration{
+				corev1.Toleration{
+					Key:      NodeTolerationKey,
+					Operator: corev1.TolerationOpEqual,
+					Value:    "true",
+					Effect:   corev1.TaintEffectNoSchedule,
+				},
+			},
+			TopologySpreadConstraints: getTopologySpreadConstraintsSpec("rook-ceph-osd"),
+		},
+
+		"osd-prepare-tsc": rook.Placement{
+			Tolerations: []corev1.Toleration{
+				corev1.Toleration{
+					Key:      NodeTolerationKey,
+					Operator: corev1.TolerationOpEqual,
+					Value:    "true",
+					Effect:   corev1.TaintEffectNoSchedule,
+				},
+			},
+			TopologySpreadConstraints: getTopologySpreadConstraintsSpec("rook-ceph-osd-prepare"),
+		},
+
 		"rgw": rook.Placement{
 			Tolerations: []corev1.Toleration{
 				corev1.Toleration{
@@ -203,3 +227,24 @@ var (
 		},
 	}
 )
+
+func getTopologySpreadConstraintsSpec(selectorValue string) []corev1.TopologySpreadConstraint {
+	topologySpreadConstraints := []corev1.TopologySpreadConstraint{
+		corev1.TopologySpreadConstraint{
+			MaxSkew:           1,
+			TopologyKey:       corev1.LabelHostname,
+			WhenUnsatisfiable: "ScheduleAnyway",
+			LabelSelector: &metav1.LabelSelector{
+				MatchExpressions: []metav1.LabelSelectorRequirement{
+					metav1.LabelSelectorRequirement{
+						Key:      appLabelSelectorKey,
+						Operator: metav1.LabelSelectorOpIn,
+						Values:   []string{selectorValue},
+					},
+				},
+			},
+		},
+	}
+
+	return topologySpreadConstraints
+}
--- a/pkg/controller/storagecluster/cephcluster.go
+++ b/pkg/controller/storagecluster/cephcluster.go
@@ -19,6 +19,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/version"
 	"k8s.io/client-go/tools/reference"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
@@ -63,7 +64,7 @@ func (r *ReconcileStorageCluster) ensureCephCluster(sc *ocsv1.StorageCluster, re
 	if sc.Spec.ExternalStorage.Enable {
 		cephCluster = newExternalCephCluster(sc, r.cephImage, r.monitoringIP)
 	} else {
-		cephCluster = newCephCluster(sc, r.cephImage, r.nodeCount, reqLogger)
+		cephCluster = newCephCluster(sc, r.cephImage, r.nodeCount, r.serverVersion, reqLogger)
 	}
 
 	// Set StorageCluster instance as the owner and controller
@@ -161,7 +162,7 @@ func (r *ReconcileStorageCluster) ensureCephCluster(sc *ocsv1.StorageCluster, re
 }
 
 // newCephCluster returns a CephCluster object.
-func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, nodeCount int, reqLogger logr.Logger) *cephv1.CephCluster {
+func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, nodeCount int, serverVersion *version.Info, reqLogger logr.Logger) *cephv1.CephCluster {
 	labels := map[string]string{
 		"app": sc.Name,
 	}
@@ -201,7 +202,7 @@ func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, nodeCount int, r
 				RulesNamespace: "openshift-storage",
 			},
 			Storage: rook.StorageScopeSpec{
-				StorageClassDeviceSets: newStorageClassDeviceSets(sc),
+				StorageClassDeviceSets: newStorageClassDeviceSets(sc, serverVersion),
 			},
 			Placement: rook.PlacementSpec{
 				"all": getPlacement(sc, "all"),
@@ -322,11 +323,15 @@ func getMonCount(nodeCount int) int {
 }
 
 // newStorageClassDeviceSets converts a list of StorageDeviceSets into a list of Rook StorageClassDeviceSets
-func newStorageClassDeviceSets(sc *ocsv1.StorageCluster) []rook.StorageClassDeviceSet {
+func newStorageClassDeviceSets(sc *ocsv1.StorageCluster, serverVersion *version.Info) []rook.StorageClassDeviceSet {
 	storageDeviceSets := sc.Spec.StorageDeviceSets
 	topologyMap := sc.Status.NodeTopologies
 
 	var storageClassDeviceSets []rook.StorageClassDeviceSet
+
+	// For kube server version 1.19 and above, topology spread constraints are used for OSD placements.
+	// For kube server version below 1.19, NodeAffinity and PodAntiAffinity are used for OSD placements.
+	supportTSC := serverVersion.Major >= defaults.KubeMajorTopologySpreadConstraints && serverVersion.Minor >= defaults.KubeMinorTopologySpreadConstraints
 
 	for _, ds := range storageDeviceSets {
 		resources := ds.Resources
@@ -338,6 +343,11 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster) []rook.StorageClassDevi
 		topologyKeyValues := []string{}
 		noPlacement := ds.Placement.NodeAffinity == nil && ds.Placement.PodAffinity == nil && ds.Placement.PodAntiAffinity == nil
 		noPreparePlacement := ds.PreparePlacement.NodeAffinity == nil && ds.PreparePlacement.PodAffinity == nil && ds.PreparePlacement.PodAntiAffinity == nil
+
+		if supportTSC {
+			noPlacement = noPlacement && ds.Placement.TopologySpreadConstraints == nil
+			noPreparePlacement = noPreparePlacement && ds.PreparePlacement.TopologySpreadConstraints == nil
+		}
 
 		if noPlacement {
 			if topologyKey == "" {
@@ -370,19 +380,33 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster) []rook.StorageClassDevi
 			preparePlacement := rook.Placement{}
 
 			if noPlacement {
-				in := getPlacement(sc, "osd")
-				(&in).DeepCopyInto(&placement)
+				if supportTSC {
+					in := getPlacement(sc, "osd-tsc")
+					(&in).DeepCopyInto(&placement)
 
-				if noPreparePlacement {
-					in := getPlacement(sc, "osd-prepare")
-					(&in).DeepCopyInto(&preparePlacement)
-				}
-
-				if len(topologyKeyValues) >= replica {
-					topologyIndex := i % len(topologyKeyValues)
-					setTopologyForAffinity(&placement, topologyKeyValues[topologyIndex], topologyKey)
 					if noPreparePlacement {
-						setTopologyForAffinity(&preparePlacement, topologyKeyValues[topologyIndex], topologyKey)
+						in := getPlacement(sc, "osd-prepare-tsc")
+						(&in).DeepCopyInto(&preparePlacement)
+
+						if len(topologyKeyValues) >= replica {
+							preparePlacement.TopologySpreadConstraints[0].TopologyKey = topologyKey
+						}
+					}
+				} else {
+					in := getPlacement(sc, "osd")
+					(&in).DeepCopyInto(&placement)
+
+					if noPreparePlacement {
+						in := getPlacement(sc, "osd-prepare")
+						(&in).DeepCopyInto(&preparePlacement)
+					}
+
+					if len(topologyKeyValues) >= replica {
+						topologyIndex := i % len(topologyKeyValues)
+						setTopologyForAffinity(&placement, topologyKeyValues[topologyIndex], topologyKey)
+						if noPreparePlacement {
+							setTopologyForAffinity(&preparePlacement, topologyKeyValues[topologyIndex], topologyKey)
+						}
 					}
 				}
 
--- a/pkg/controller/storagecluster/cephcluster.go
+++ b/pkg/controller/storagecluster/cephcluster.go
@@ -337,6 +337,7 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster) []rook.StorageClassDevi
 		topologyKey := ds.TopologyKey
 		topologyKeyValues := []string{}
 		noPlacement := ds.Placement.NodeAffinity == nil && ds.Placement.PodAffinity == nil && ds.Placement.PodAntiAffinity == nil
+		noPreparePlacement := ds.PreparePlacement.NodeAffinity == nil && ds.PreparePlacement.PodAffinity == nil && ds.PreparePlacement.PodAntiAffinity == nil
 
 		if noPlacement {
 			if topologyKey == "" {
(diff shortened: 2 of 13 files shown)


pkg/controller/storagecluster/cephcluster.go
```diff
@@ -291,9 +291,10 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster) []rook.StorageClassDevi
 
 		topologyKey := ds.TopologyKey
 		topologyKeyValues := []string{}
-		noPlacement := ds.Placement.NodeAffinity == nil && ds.Placement.PodAffinity == nil && ds.Placement.PodAntiAffinity == nil
+		noPlacement := ds.Placement.NodeAffinity == nil && ds.Placement.PodAffinity == nil && ds.Placement.PodAntiAffinity == nil && ds.Placement.TopologySpreadConstraints == nil
+		noPreparePlacement := ds.PreparePlacement.NodeAffinity == nil && ds.PreparePlacement.PodAffinity == nil && ds.PreparePlacement.PodAntiAffinity == nil && ds.PreparePlacement.TopologySpreadConstraints == nil
```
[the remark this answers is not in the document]
kesavanvt: Sounds good to me, have placed the required changes. Now the OCS expects either both or neither of these to be set!.

pkg/controller/storagecluster/cephcluster.go
```diff
@@ -291,9 +291,10 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster) []rook.StorageClassDevi
...
 
-		if noPlacement {
+		if noPreparePlacement {
```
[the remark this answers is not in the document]
kesavanvt: Thanks, have removed the check for noPlacement as we are already doing above!.

pkg/controller/storagecluster/cephcluster_test.go
```diff
@@ -167,9 +167,10 @@ func TestStorageClassDeviceSetCreation(t *testing.T) {
 	sc1.Spec.StorageDeviceSets = mockDeviceSets
 	sc1.Status.NodeTopologies = &api.NodeTopologyMap{
 		Labels: map[string]api.TopologyLabelValues{
-			zoneTopologyLabel: []string{
-				"zone1",
-				"zone2",
+			defaults.RackTopologyKey: []string{
```
jarrpa, line 170: Don't modify the existing test cases, only add more test cases. If any of these changes are required for the previous patch to pass unit tests, those changes should be made in the previous commit.

jarrpa:
Something else to consider: naturally this should stay as WIP until we start working on OCS 4.7. However, OCS 4.7 still needs to run on OCP 4.6, which will not have the [fix that we need](https://github.com/kubernetes/kubernetes/pull/91258) to make TSCs viable. Can we somehow perform a check to see if the required setting is correctly configured?

jarrpa:
Something else to consider: OCS 4.6 still needs to run on OCP 4.5, which will not have the fix that we need to make TSCs viable. Can we somehow perform a check to see if the required setting is correctly configured?

pkg/controller/storagecluster/cephcluster.go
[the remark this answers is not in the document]
travisn: Line 295 still needs the `&& noPlacement` because of the fallback to use osd daemon placement if no osd prepare placement is specified.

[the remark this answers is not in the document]
kesavanvt: now `&& noPlacement` is handled in Line 341.

pkg/controller/storagecluster/cephcluster.go
```diff
@@ -321,24 +322,18 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster) []rook.StorageClassDevi
 
 		for i := 0; i < replica; i++ {
 			placement := rook.Placement{}
+			prepareplacement := &rook.Placement{}
 
-			if noPlacement {
-				in := getPlacement(sc, "osd")
+			if noPreparePlacement {
```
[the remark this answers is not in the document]
kesavanvt: done

pkg/controller/storagecluster/cephcluster.go
```diff
@@ -321,24 +322,18 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster) []rook.StorageClassDevi
...
-					nodeZoneSelector := corev1.NodeSelectorRequirement{
-						Key:      topologyKey,
-						Operator: corev1.NodeSelectorOpIn,
-						Values:   []string{topologyKeyValues[topologyIndex]},
-					}
-					appendNodeRequirements(&placement, nodeZoneSelector)
+					prepareplacement.TopologySpreadConstraints[0].TopologyKey = topologyKey
 				}
 			} else {
+				prepareplacement = &ds.PreparePlacement
```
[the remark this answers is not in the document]
kesavanvt: From this, i have inferred that 

- if either of Placement or PreparePlacement is set in yaml its config will be mapped to `PreparePlacement`.
 `Placement` takes the default host failuredomain and have modified accordingly.

pkg/controller/storagecluster/cephcluster_test.go
kesavanvt: Small tweak was required for the existing tests - have added in seperate commit

pkg/controller/storagecluster/cephcluster_test.go
```diff
@@ -262,5 +262,84 @@ func TestStorageClassDeviceSetCreation(t *testing.T) {
...
+	sc4.Spec.StorageDeviceSets = mockDeviceSets
+	sc4.Status.NodeTopologies = &api.NodeTopologyMap{
+		Labels: map[string]api.TopologyLabelValues{
+			defaults.RackTopologyKey: []string{
+				"rack1",
+				"rack2",
+				"rack3",
+			},
+		},
+	}
```
jarrpa, line 275: Move this up with the other sc definitions.

pkg/controller/storagecluster/cephcluster_test.go
```diff
@@ -228,10 +229,14 @@ func TestStorageClassDeviceSetCreation(t *testing.T) {
 			lenOfMatchExpression: 1,
 		},
 	}
-
+	// Testing StorageClassDeviceSetCreation for kube version below 1.19
+	serverVersion := &version.Info{
+		Major: "1",
+		Minor: "18",
+	}
```
jarrpa, line 236: Move this chunk above the case tables definition.

pkg/controller/storagecluster/cephcluster_test.go
```diff
@@ -262,5 +262,84 @@ func TestStorageClassDeviceSetCreation(t *testing.T) {
...
+			if c.lenOfMatchExpression == 0 {
+				assert.Nil(t, scds.Placement.NodeAffinity)
+			} else {
+			matchExpressions := scds.Placement.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms[0].MatchExpressions
+			assert.Equal(t, c.lenOfMatchExpression, len(matchExpressions))
+			}
+			topologyKey := scds.PreparePlacement.TopologySpreadConstraints[0].TopologyKey
+
+			if c.topologyKey == "rack" {
+				assert.Equal(t, defaults.RackTopologyKey, topologyKey)
```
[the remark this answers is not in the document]
kesavanvt: `c.topologyKey` is used for filtering among the cases defined, when it is of rack the topologyKey is ensured to hold the value `topology.rook.io/rack` instead of `failure-domain.beta.kubernetes.io/zone`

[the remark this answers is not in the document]
jarrpa: ...ah! Gotcha. Thanks!

kesavanvt:
> Don't just make a hard split between the code paths like that, they need mix more. In particular, you're ignoring the PreparePlacement entirely in the "old" code path.

Now the PreparePlacement has been integrated into the "old" code path and the fallback to main OSD placement is handled in OCS code base as well.

jarrpa (changes requested):
Thank you for splitting the commits, definitely made it much easier to review. :)

pkg/controller/storagecluster/cephcluster.go
```diff
@@ -321,14 +322,19 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster) []rook.StorageClassDevi
...
 				in := getPlacement(sc, "osd")
 				(&in).DeepCopyInto(&placement)
+				in = getPlacement(sc, "osd-prepare")
+				(&in).DeepCopyInto(prepareplacement)
 
 				if len(topologyKeyValues) >= replica {
 					podAffinityTerms := placement.PodAntiAffinity.PreferredDuringSchedulingIgnoredDuringExecution
 					podAffinityTerms[0].PodAffinityTerm.TopologyKey = topologyKey
+					podAffinityTerms = prepareplacement.PodAntiAffinity.PreferredDuringSchedulingIgnoredDuringExecution
+					podAffinityTerms[0].PodAffinityTerm.TopologyKey = topologyKey
```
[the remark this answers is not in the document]
kesavanvt: Done!

pkg/controller/storagecluster/cephcluster.go
```diff
@@ -325,42 +361,58 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster) []rook.StorageClassDevi
...
-					podAffinityTerms[0].PodAffinityTerm.TopologyKey = topologyKey
-					podAffinityTerms = prepareplacement.PodAntiAffinity.PreferredDuringSchedulingIgnoredDuringExecution
-					podAffinityTerms[0].PodAffinityTerm.TopologyKey = topologyKey
-
-					topologyIndex := i % len(topologyKeyValues)
-					nodeZoneSelector := corev1.NodeSelectorRequirement{
-						Key:      topologyKey,
-						Operator: corev1.NodeSelectorOpIn,
-						Values:   []string{topologyKeyValues[topologyIndex]},
+				if supportTSC {
```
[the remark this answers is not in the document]
kesavanvt: If i understand correctly, another `if supportTSC` needs to be added under `if len(topologyKeyValues) >= replica` for placement logic and have modified accordingly .

pkg/controller/storagecluster/cephcluster.go
```diff
@@ -383,7 +383,32 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster) []rook.StorageClassDevi
 					}
 					appendNodeRequirements(&placement, nodeZoneSelector)
 				}
+
+				if !noPreparePlacement {
+					prepareplacement = ds.PreparePlacement
+				}
+
+				if noPreparePlacement {
```
jarrpa, line 432:
```suggestion
				} else {
```

pkg/controller/storagecluster/cephcluster.go
```diff
@@ -366,14 +367,13 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster) []rook.StorageClassDevi
 
 		for i := 0; i < replica; i++ {
 			placement := rook.Placement{}
+			prepareplacement := rook.Placement{}
 
 			if noPlacement {
 				in := getPlacement(sc, "osd")
 				(&in).DeepCopyInto(&placement)
-
```
jarrpa, line 373: Remove this change

pkg/controller/storagecluster/cephcluster.go
```diff
@@ -321,13 +330,34 @@ func getMonCount(nodeCount int) int {
...
+}
+
 // newStorageClassDeviceSets converts a list of StorageDeviceSets into a list of Rook StorageClassDeviceSets
-func newStorageClassDeviceSets(sc *ocsv1.StorageCluster) []rook.StorageClassDeviceSet {
+func newStorageClassDeviceSets(sc *ocsv1.StorageCluster, serverVersion *version.Info) []rook.StorageClassDeviceSet {
 	storageDeviceSets := sc.Spec.StorageDeviceSets
 	topologyMap := sc.Status.NodeTopologies
 
+	var noPlacement bool
+	var noPreparePlacement bool
```
jarrpa, line 354: I do see the need for defining them before instnatiating them, but why are you defining them up here? I can't see anything that would mean we need them outside the scope of the `for` loop they were originally in, but I might just be blind.

pkg/controller/storagecluster/cephcluster.go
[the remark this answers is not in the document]
jarrpa: :+1:

jarrpa (changes requested):
Looking better.

pkg/controller/storagecluster/cephcluster.go
```diff
@@ -370,37 +406,51 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster) []rook.StorageClassDevi
...
+					in := getPlacement(sc, "osd-tsc")
+					(&in).DeepCopyInto(&placement)
+				} else {
+					in := getPlacement(sc, "osd")
+					(&in).DeepCopyInto(&placement)
 
-				if len(topologyKeyValues) >= replica {
-					placement.PodAntiAffinity.PreferredDuringSchedulingIgnoredDuringExecution[0].PodAffinityTerm.TopologyKey = topologyKey
+					if len(topologyKeyValues) >= replica {
+						placement.PodAntiAffinity.PreferredDuringSchedulingIgnoredDuringExecution[0].PodAffinityTerm.TopologyKey = topologyKey
```
kesavanvt, line 417: This commit shows the white space difference in UI, unable to catch this in IDE as well as in git diff. not sure why its showing in here.

pkg/controller/storagecluster/cephcluster.go
kesavanvt: done

pkg/controller/storagecluster/cephcluster.go
kesavanvt: fixed!

pkg/controller/storagecluster/cephcluster.go
```diff
@@ -321,24 +330,52 @@ func getMonCount(nodeCount int) int {
...
 		topologyKey := ds.TopologyKey
 		topologyKeyValues := []string{}
-		noPlacement := ds.Placement.NodeAffinity == nil && ds.Placement.PodAffinity == nil && ds.Placement.PodAntiAffinity == nil
 
-		if noPlacement {
+		if supportTSC {
+			noPlacement = ds.Placement.NodeAffinity == nil && ds.Placement.PodAffinity == nil && ds.Placement.PodAntiAffinity == nil && ds.Placement.TopologySpreadConstraints == nil
+			noPreparePlacement = ds.PreparePlacement.NodeAffinity == nil && ds.PreparePlacement.PodAffinity == nil && ds.PreparePlacement.PodAntiAffinity == nil && ds.PreparePlacement.TopologySpreadConstraints == nil
+		} else {
+			noPlacement = ds.Placement.NodeAffinity == nil && ds.Placement.PodAffinity == nil && ds.Placement.PodAntiAffinity == nil
```
travisn, line 374: You could simplify lines 370-376 with a pattern such as:
```
noPlacement := ds.Placement.NodeAffinity == nil && ds.Placement.PodAffinity == nil && ds.Placement.PodAntiAffinity == nil
noPreparePlacement := ds.PreparePlacement.NodeAffinity == nil && ds.PreparePlacement.PodAffinity == nil && ds.PreparePlacement.PodAntiAffinity == nil
if supportTSC {
  noPlacement = noPlacement && ds.Placement.TopologySpreadConstraints == nil
  noPreparePlacement = noPreparePlacement && ds.PreparePlacement.TopologySpreadConstraints == nil
}
```

pkg/controller/storagecluster/cephcluster.go
```diff
@@ -366,24 +403,61 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster) []rook.StorageClassDevi
...
+							Key:      topologyKey,
+							Operator: corev1.NodeSelectorOpIn,
+							Values:   []string{topologyKeyValues[topologyIndex]},
+						}
+						appendNodeRequirements(&placement, nodeZoneSelector)
 					}
-					appendNodeRequirements(&placement, nodeZoneSelector)
 				}
+
+				if !noPreparePlacement {
```
travisn, line 429: Lines 409-427 and lines 429-455 are very similar. Can you refactor to a helper method to avoid duplication of the pattern?

pkg/controller/defaults/placements.go
```diff
@@ -90,6 +90,90 @@ var (
 			},
 		},
 
+		"osd-prepare": rook.Placement{
+			Tolerations: []corev1.Toleration{
```
travisn, line 94: Since the "all" placement has the toleration, you shouldn't need to duplicate it for the OSDs. Although perhaps the storageClassDeviceSets pick up the "all" placement, so you might need it. If you do need it, you could factor out to a helper method so you don't have the same toleration code duplicated.

pkg/controller/defaults/placements.go
```diff
@@ -90,6 +90,90 @@ var (
...
+				corev1.Toleration{
+					Key:      NodeTolerationKey,
+					Operator: corev1.TolerationOpEqual,
+					Value:    "true",
+					Effect:   corev1.TaintEffectNoSchedule,
+				},
+			},
+			PodAntiAffinity: &corev1.PodAntiAffinity{
+				PreferredDuringSchedulingIgnoredDuringExecution: []corev1.WeightedPodAffinityTerm{
```
travisn, line 103: Can you factor this to a helper method to avoid duplication of the whole antiaffinity spec?

pkg/controller/defaults/placements.go
```diff
@@ -90,6 +90,90 @@ var (
...
+		"osd-tsc": rook.Placement{
+			Tolerations: []corev1.Toleration{
+				corev1.Toleration{
+					Key:      NodeTolerationKey,
+					Operator: corev1.TolerationOpEqual,
+					Value:    "true",
+					Effect:   corev1.TaintEffectNoSchedule,
+				},
+			},
+			TopologySpreadConstraints: []corev1.TopologySpreadConstraint{
```
travisn, line 132: Refactor to a helper method to avoid duplication of the whole TSC spec?

travisn (changes requested):

pkg/controller/defaults/placements.go
kesavanvt: Done, refactored TSC spec to a helper method

pkg/controller/defaults/placements.go
kesavanvt: Noted!. as discussed, will plan to refactor it in a separate PR.

pkg/controller/defaults/placements.go
kesavanvt: Noted!. as discussed, will plan to refactor it in a separate PR.

pkg/controller/storagecluster/cephcluster.go
kesavanvt: Refactored the setting of topology key/values for affinity placements to a helper method and modified the `if structure` to avoid code duplication as mush as possible.

pkg/controller/storagecluster/cephcluster.go
```diff
@@ -366,24 +400,46 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster) []rook.StorageClassDevi
...
+				} else {
+					in := getPlacement(sc, "osd")
+					(&in).DeepCopyInto(&placement)
+
+					if noPreparePlacement {
+						in := getPlacement(sc, "osd-prepare")
+						(&in).DeepCopyInto(&preparePlacement)
+					}
+
+					if len(topologyKeyValues) >= replica {
```
travisn, line 427: Can you explain why the topology keys are related to the replica? I would have thought we would always use a consistent topology key instead of using a different one depending on `i` as we iterate.

kesavanvt: As part of the old design, the topology key that we get from default.placements is `kubernetes.io/hostname`, this get replaced either by rack or zone in the `for loop`. the topology key doesn't depend on each `i `, it will be the same for every iteration. Only the values of node selector changes for each `i` like (us-east-2a, us-east-2b, us-east-2c)

pkg/controller/storagecluster/cephcluster.go
```diff
@@ -166,6 +169,12 @@ func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, nodeCount int, r
 		"app": sc.Name,
 	}
 
+	// determining the kube server version
+	serverVersion, err := determineKubeVersion()
+	if err != nil {
+		reqLogger.Error(err, "failed to determine the kube server version.")
```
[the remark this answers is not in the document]
kesavanvt: I agree on returning an error and stop the flow, as the user will be unaware with the branching that it takes.
Since we are returning an error in case of failure detection, faced difficulties in the existing unit tests to succeed (unable to fake the serverVersion retrieval). As a better approach, have integrated the serverVersion field in `ReconcileStorageCluster struct` and modified accordingly.

travisn (approved):

pkg/controller/storagecluster/cephcluster.go
```diff
@@ -366,24 +367,32 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster) []rook.StorageClassDevi
...
 				if len(topologyKeyValues) >= replica {
-					podAffinityTerms := placement.PodAntiAffinity.PreferredDuringSchedulingIgnoredDuringExecution
-					podAffinityTerms[0].PodAffinityTerm.TopologyKey = topologyKey
-
-					topologyIndex := i % len(topologyKeyValues)
-					nodeZoneSelector := corev1.NodeSelectorRequirement{
-						Key:      topologyKey,
-						Operator: corev1.NodeSelectorOpIn,
-						Values:   []string{topologyKeyValues[topologyIndex]},
+					setTopologyForAffinity(&placement, i, topologyKeyValues, topologyKey)
```
jarrpa: Since the value of `topologyIndex` would be the same for both function calls, I would rather see `topologyIndex := i % len(topologyKeyValues)` out here and just pass `topologyKeyValues[topologyIndex]` directly rather than the list and index number.

pkg/controller/defaults/placements.go
```diff
@@ -203,3 +227,24 @@ var (
 		},
 	}
 )
+
+func getTopologySpreadConstraintsSpec(value string) []corev1.TopologySpreadConstraint {
```
jarrpa, line 231:
```suggestion
func getTopologySpreadConstraintsSpec(selectorValue string) []corev1.TopologySpreadConstraint {
```

jarrpa (changes requested):

jarrpa (approved):
My sincerest apologies for all the delays. Thank you so much for your patience. Excellent work!

jarrpa:
Looks like there's a valid CI failure. :(

kesavanvt:
> Looks like there's a valid CI failure. :(

Recent commit to master had changes in storagecluster_controller_test.go which lead to import conflicts with my changes and that has been fixed now. Thanks for the extended support :)

jarrpa (approved):

PR #731, opened 2020-08-28
StorageCluster: set uninstall defaults

raghavendra-talur:
The default policy for uninstall mode is set to graceful and the default for the cleanup policy is set to delete.

These annotations are set only if the annotations are not already set on the StorageCluster. If set, even as an empty string, we don't make any changes.


raghavendra-talur:
@jarrpa Please take a look. This is to be merged after #730 and #718 

cc: @iamniting

PR #733, opened 2020-08-31
Deploy OCS prometheus alerts

umangachapagain:
Follows up  on: #705  

Latest 3 Commits are relevant to this PR.


umangachapagain:
/hold
needs #705 to be merged

hack/build-container.sh
```diff
@@ -4,4 +4,5 @@ set -e
 
 source hack/common.sh
 
+cp $PROMETHEUS_RULE_PATH/*rules*.yaml $OUTDIR
```
anmolsachan, line 7:
```suggestion
cp $PROMETHEUS_RULES_PATH/*rules*.yaml $OUTDIR
```
NIT, Since everywhere we call them rules.  Also, change in other places

pkg/controller/storagecluster/prometheus.go
```diff
@@ -0,0 +1,112 @@
...
+	monitoringv1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	k8sYAML "k8s.io/apimachinery/pkg/util/yaml"
+)
+
+const (
+	internalPrometheusRuleFilepath = "/ocs-prometheus-rules/prometheus-ocs-rules.yaml"
+	externalPrometheusRuleFilepath = "/ocs-prometheus-rules/prometheus-ocs-rules-external.yaml"
```
jarrpa, line 20: Why are you storing them as YAML files instead of building them as go structs? Especially given the minimal differences between the two?

hack/build-container.sh
jarrpa: +1

obnoxxx (changes requested):
@umangachapagain - thanks for the PR!

To summarize what's needed here:

- rebase on current master
- fix the tests
- s/PROMETHEUS_RULE_PATH/PROMETHEUS_RULES_PATH/g
- @jarrpa 's comment about storing as yaml

obnoxxx:
@umangachapagain @anmolsachan - FYI created PR #737 as a substitute for this to be try and get it to completion while you guys are out on well deserved holiday :smiley:

obnoxxx:
This can be closed since #737 has been merged.

pkg/controller/storagecluster/prometheus.go
umangachapagain: > Why are you storing them as YAML files instead of building them as go structs?
>
That's what every monitoring project does.
It's much easier to store a YAML and apply it than to create Go structs for each alert.
Also, YAML files are more flexible in terms of updating as required and adding more alerts. With Go struct, you'll have to build binary each time there is even a small change in alert. Decoupling is very helpful and scalable.

> Especially given the minimal differences between the two?
>
That minimal difference between the two is what makes them work on Internal and External cluster.

You might not see it now as this is just the tip of the iceberg but soon you'll see why YAML (separate for Internal and External mode) is a better option.  Specially, once we add more and more alerts and the automation to generate these YAML files.

Also, if we store YAML files, we can just `oc create -f ` at any time to test it out on cluster with custom changes without deploying full stack.

PR #736, opened 2020-09-01
StorageCluster: propagate uninstall mode

raghavendra-talur:
Rook waits for the deletion of the PVCs before deleting the cephCluster.
To override the wait, we can set the "allowUninstallWithVolumes" option
in the Cleanup Policy of the cephCluster.

With this commit, if the UninstallMode is set to forced in the
StorageCluster, we set the allowUninstallWithVolumes option to true in
the cephCluster.

Co-authored-by: Nitin Goyal <nigoyal@redhat.com>


raghavendra-talur:
@jarrpa @obnoxxx This PR requires an update in rook that is in review. Once https://github.com/rook/rook/pull/6188 is merged, we can retest and merge this.

cc: @iamniting

obnoxxx:
https://github.com/rook/rook/pull/6188 is merged now.
But we need to update the vendored / dependent rook copy in ocs-operator in this PR, afaict.

raghavendra-talur:
@jarrpa @obnoxxx I believe the changes are all good now.

However, one thing to note is that there are no Rook images for the commit we want and is based on 1.4 branch. We have added `rook/ceph:v1.4.0.283.g4df2549` image of rook which is actually based on the master branch.

Let me know if there is a better way to do it.

cc: @iamniting

PR #737, opened 2020-09-01
Deploy OCS prometheus alerts - substitute for PR #733

obnoxxx:
Follow-up to PR #705 
Substitute for PR #733 in order to make progress while @umangachapagain is out.


umangachapagain:
Interesting changes.

1. Why is Rook vendor change a part of this PR? These aren't related in any way.
2. What is the justification for using Go structs instead of YAML for alerting rules?
     I have added my reasons for using YAML instead of Go Structs here: https://github.com/openshift/ocs-operator/pull/733#discussion_r481863240

PR #707, opened 2020-08-17
Fix and refactor failing e2e tests

jarrpa:
There are several problems with the e2e-aws tests that have been preventing them from successfully running on PRs. This PR fixes all currently known problems. It includes some refactoring of the OCS uninstallation to deal with intermittent errors in the AfterSuite.

Signed-off-by: Jose A. Rivera <jarrpa@redhat.com>

PR #738, opened 2020-09-01
functests: Additional fixes and optimizations

jarrpa:


Makefile
```diff
@@ -135,7 +135,7 @@ build-functest:
 
 functest: build-functest
 	@echo "Running ocs developer functional test suite"
-	hack/functest.sh
+	hack/functest.sh $(ARGS)
```
umangachapagain, line 138: Didn't know it can take arguments. Is there an example of how I can make use of available args?

umangachapagain (commented):
+ hack: Preserve must-gather output in ci-operator Jobs 
 CI jobs already preserves must-gather logs. What is this commit doing different?

+ cluster-clean: keep CephCluster finalizer 
 We don't use it in E2E. Shouldn't Uninstall be a separate PR with parts to enable Uninstall testing during E2E?

+ hack: Preserve must-gather output in ci-operator Jobs
  CI jobs already preserves must-gather logs. What is this commit doing different?

+  deploymanager: Fix nul pointer error with StorageCluster 
 We can drop this. There's a fix posted with better handling.

+  deploymanager: Fix broken methods
    > Since gomerga does not exist in the deploy-manager
package, we have to remove it and replace it with proper error handling.

   I don't understand this. Isn't Gomega just a library that you import and use? What was breaking?

+  e2e: Fix panicking tests 
 How does removing `BeforeEach()` fix panics? Can you add a little more detail into commit desc?

+   deploymanager: Fix broken methods  and e2e: Cleanup PVC tests 
 Can we merge these 2? It'll be more clear as to why we removed certain things and how the new functions work.

+  e2e: Create test Namespace after OCS installation 
 Why does it matter? I don't see any issues with creating Namespaces before installing OCS.

+  Move e2e test suites into subdirectories 
 This commit does more than just move things. Can you add a little more details into commit messages?

+  e2e: Whitespace and style fixes
 Let's merge this with the commit that has white-spaces and typo fixes.

Makefile
umangachapagain, line 138: What are the args that we can pass? Is it documented somewhere?

functests/ocs/data_validation_test.go
```diff
@@ -1,30 +1,27 @@
...
 
 func DataValidationTest() {
-	var k8sClient *kubernetes.Clientset
-	var deployManager *deploymanager.DeployManager
-
-	BeforeEach(func() {
-		RegisterFailHandler(Fail)
-		deployManager, err := deploymanager.NewDeployManager()
-		Expect(err).To(BeNil())
+	dm := tests.DeployManager
```
umangachapagain, line 18: If we are using a global DeployManager what is the need to assign it to a different local variable?
We should be able to use the global var directly right? Since we don't even mutate it.

hack/functest.sh
```diff
@@ -2,7 +2,8 @@
 
 source hack/common.sh
 
-$OUTDIR_BIN/functests --ocs-registry-image="${CATALOG_FULL_IMAGE_NAME}" \
+"$OUTDIR_BIN/${GINKGO_TEST_SUITE}_tests" -ginkgo.v \
+        --ocs-registry-image="${CATALOG_FULL_IMAGE_NAME}" \
```
umangachapagain, line 6: remove extra indentation

pkg/deploy-manager/common.go
```diff
@@ -30,69 +40,114 @@ func (t *DeployManager) CreateNamespace(namespace string) error {
 	return nil
 }
 
+// RemoveAllFinalizers removes all finalizers from every object in the target namespace
+func (t *DeployManager) RemoveAllFinalizers(namespace string) error {
```
umangachapagain, line 44: If things work as expected, why would we need to force Removal of Finalizer from every object in given Namespace?
I see you removed it from script but added it to test code. I'm confused.

pkg/deploy-manager/common.go
```diff
@@ -105,32 +160,79 @@ func (t *DeployManager) GetDeploymentImage(name string) (string, error) {
...
-			return err
+func (t *DeployManager) WaitForPVCBound(pvc *k8sv1.PersistentVolumeClaim, namespace string) error {
+	pvc, err := t.k8sClient.CoreV1().PersistentVolumeClaims(namespace).Create(pvc)
+	if err != nil && !errors.IsAlreadyExists(err) {
+		return err
+	}
+
+	lastReason := ""
+	timeout := 100 * time.Second
+	interval := 1 * time.Second
```
umangachapagain, line 171: Instead of retrying every second for 100 seconds, can we just do retry every 5 seconds for 10 times?
This is mentioned in K8s guidelines and practices too. Creating a PVC and binding it shouldn't take 100 seconds in a properly functioning system. If it is, something needs to be fixed. Adding more timeout just gives more window for bugs to sneak in.

pkg/deploy-manager/common.go
```diff
@@ -105,32 +160,79 @@ func (t *DeployManager) GetDeploymentImage(name string) (string, error) {
...
-			return err
+func (t *DeployManager) WaitForJobSucceeded(job *batchv1.Job, namespace string) error {
+	job, err := t.k8sClient.BatchV1().Jobs(namespace).Create(job)
+	if err != nil && !errors.IsAlreadyExists(err) {
+		return err
+	}
+
+	lastReason := ""
+	timeout := 200 * time.Second
+	interval := 1 * time.Second
```
umangachapagain, line 205: Do we need to retry every second? Can't we get away with every minute?
I don't suppose this job will complete within 1 or 2 or 5 seconds.

And, having a hard-coded 200 seconds timeout doesn't look right. We can't be waiting for 200s for each job.
Can we make it a parameter and pass some realistic values? it'll drastically reduce our test run time.

umangachapagain (changes requested):
Lots of good additions. Will make testing a lot more easier.

umangachapagain:
So half the commits here are merged via https://github.com/openshift/ocs-operator/pull/707. And there was no indication in either PR that these are related? Ok.

jarrpa:
Oh geez, sorry about this. I was just raising this one to trigger the OCP CI for additional testing, I wasn't intended to have it be for review (hence the [WIP] tag). Some of the weirdness in how the commits were split was due to the order in which I discovered and/or fixed them. In particular, a lot of the latter ones were done after #707 was stable so I didn't want to introduce any new changes that weren't specifically in response to it's review.

To answer some of your remaining comments:

> * e2e: Use DeployManager namespace
>    Why is this only for tools tests?

This was the only test I found that actually needed it. All the other instances I believe used methods rather than direct client calls.

> * e2e: Remove unnecessary fail handlers
>   This can be merged with commit that moves tests around and needs explanation why we don't need it any more.

Sorry I didn't catch this sooner. I added an explanation in the commit message.

umangachapagain (approved):

PR #739, opened 2020-09-02
restores YAML manifests for Prometheus rules

umangachapagain:
This restores prometheus rules YAML deleted by  commit
5a4b96929585cd7ed9dd217dbe864884859c79f7 and uses it for deployment
instead of Go Structs. This helps in decoupling monitoring resources
from operator resources. Also, at later stage these files will be
generated dynamically and will be scalable.

**CI fix :** https://github.com/openshift/release/pull/11528

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -4,3 +4,4 @@ USER nobody
 
 ADD _output/bin/ocs-operator /usr/local/bin/ocs-operator
 ADD _output/bin/metrics-exporter /usr/local/bin/metrics-exporter
+ADD _output/*rules*.yaml /ocs-prometheus-rules/
--- a/build/Dockerfile.build
+++ b/build/Dockerfile.build
@@ -8,3 +8,4 @@ FROM registry.access.redhat.com/ubi8/ubi-minimal
 
 COPY --from=builder /ocs-operator/build/_output/bin/ocs-operator /usr/local/bin/ocs-operator
 COPY --from=builder /ocs-operator/build/_output/bin/metrics-exporter /usr/local/bin/metrics-exporter
+COPY --from=builder /ocs-operator/build/_output/*rules*.yaml /ocs-prometheus-rules/
--- a/hack/build-operator.sh
+++ b/hack/build-operator.sh
@@ -4,4 +4,5 @@ set -e
 
 source hack/common.sh
 
+cp $PROMETHEUS_RULES_PATH/*rules*.yaml $OUTDIR
 ${IMAGE_BUILD_CMD} build -f build/Dockerfile -t "${OPERATOR_FULL_IMAGE_NAME}" build/
--- a/metrics/deploy/prometheus-ocs-rules-external.yaml
+++ b/metrics/deploy/prometheus-ocs-rules-external.yaml
@@ -1,1 +1,25 @@
-# this is just a temporary empty stub to satisfy the CI
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    prometheus: k8s
+    role: alert-rules
+  name: prometheus-ocs-rules
+  namespace: openshift-storage
+spec:
+  groups:
+  - name: cluster-services-alert.rules
+    rules:
+    - alert: ClusterObjectStoreState
+      annotations:
+        description: Cluster Object Store is in unhealthy state for more than 15s.
+          Please check Ceph cluster health or RGW connection.
+        message: Cluster Object Store is in unhealthy state. Please check Ceph cluster
+          health or RGW connection.
+        severity_level: error
+        storage_type: RGW
+      expr: |
+        ocs_rgw_health_status{job="ocs-metrics-exporter"} > 1
+      for: 15s
+      labels:
+        severity: critical
--- a/metrics/deploy/prometheus-ocs-rules.yaml
+++ b/metrics/deploy/prometheus-ocs-rules.yaml
@@ -1,1 +1,27 @@
-# this is just a temporary empty stub to satisfy the CI
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    prometheus: k8s
+    role: alert-rules
+  name: prometheus-ocs-rules
+  namespace: openshift-storage
+spec:
+  groups:
+  - name: ocs.rules
+    rules: []
+  - name: cluster-services-alert.rules
+    rules:
+    - alert: ClusterObjectStoreState
+      annotations:
+        description: Cluster Object Store is in unhealthy state for more than 15s.
+          Please check Ceph cluster health.
+        message: Cluster Object Store is in unhealthy state. Please check Ceph cluster
+          health.
+        severity_level: error
+        storage_type: RGW
+      expr: |
+        ocs_rgw_health_status{job="ocs-metrics-exporter"} > 1
+      for: 15s
+      labels:
+        severity: critical
--- a/openshift-ci/Dockerfile.deploy
+++ b/openshift-ci/Dockerfile.deploy
@@ -4,6 +4,7 @@ ENV LANG=en_US.utf8
 
 COPY ocs-operator /usr/local/bin/ocs-operator
 COPY metrics-exporter /usr/local/bin/metrics-exporter
+COPY *rules*.yaml /ocs-prometheus-rules/
 USER 1001
 
 ENTRYPOINT [ "/usr/local/bin/ocs-operator" ]
--- a/pkg/controller/storagecluster/prometheus.go
+++ b/pkg/controller/storagecluster/prometheus.go
@@ -1,33 +1,41 @@
 package storagecluster
 
 import (
+	"bytes"
 	"context"
 	"fmt"
-
-	ocsv1 "github.com/openshift/ocs-operator/pkg/apis/ocs/v1"
+	"io/ioutil"
+	"os"
+	"path/filepath"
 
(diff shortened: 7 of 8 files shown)


obnoxxx:
**Comment:**
* The yamls were not simply deleted in the mentioned patch, but since there was no apparent reason for the unnecissarily complicated yaml mechanism, they were replaced by something that was deemed simpler and more robust.  

**Questions:**
* What is the reason for wanting to decouple the monitoring rules from operator resources?
* It seems very fragile to want to store yaml files in the container and load them at runtime (unless you want to overwrite them at runtime). Usually having them in go code seems more robust.
* what are scalable rules?

**One more comment:**
* If the rules are generated from some other souce that is probably simpler than the yaml, then these yamls should not just be dropped into this repository and commited as god-given but they should either be pulled by the build process or the source and the generation mechansim should be in this repo and invoked at build time.

umangachapagain:
> **Comments:**
> 
>     * The yamls were not simply deleted in the mentioned patch, but since there was no apparent reason for the unnecissarily complicated yaml mechanism, they were replaced by something that was deemed simpler and more robust.
>
Unnecessarily complicated YAML? Can you elaborate on that?
Who deemed it simpler and robust? Did we have consensus before making the change?

>
>     * It seems rather fragile to want to store yaml files in the container and load them at runtime (unless you want to overwrite them at runtime). Usually having them in go code seems more robust.
> 
Can you explain why it "seems" fragile?
Even today OCS runs alerts derived from Rook which stores alerts as YAMLs (which magically reaches their repo) and we haven't had a single complaint.
Many OpenShift projects use this same pattern and is well established across several other monitoring projects. Some take a step forward and convert YAML to binary dump in Go and use it.

> 
> **Questions:**
> 
>     * What is the reason for wanting to decouple the monitoring rules from operator resources?
>
+ We don't want monitoring to get mixed up with Operator.
+ We want to be able to work on it independently. Add to it, update it or delete it without any dependencies on OCS operator.
+ We don't want to build OCS every time we need to test out a change and be blocked on Operator bugs or development cycles.
+ Like we've discussed on other forums, not everyone understands what goes behind building these things,  we don't want anyone making changes to it unknowingly.
+ We don't want the maintainers/reviewers of OCS who might not have knowledge about monitoring to be burdened with reviewing or merging PRs they don't understand. This has been working well with Rook and many Kubernetes projects.

>
>     * what are scalable rules?
> 
Currently you see 1 alert and you are already so confused. Consider 10 or 20. Consider having it across namespaces. Consider different configuration parameters that might not be known until later. Consider cases where you need to make a small threshold change in all 20 of these alerts. Consider requiring a subset of these alerts or a mix of alerts from multiple projects.

Can you simplify all these with Go without having to develop yet another framework?
In the middle of release, can you make easy amends to these alerts if you use Go structs?
If anyone decides they want more alerts than what is provided by operator, how do they get that without rebuilding/redeploying OCS or waiting for the next release?

> 
> **One more comment:**
> 
>     * If the rules are generated from some other souce that is probably simpler than the yaml, then these yamls should not just be dropped into this repository and commited as god-given but they should either be pulled by the build process or the source and the generation mechansim should be in this repo and invoked at build time.
>
This work is in progress : https://github.com/openshift/ocs-operator/pull/740
All the libraries/framework are god-given. I don't consider this any different :)

obnoxxx:
> > **Comments:**
> > ```
> > * The yamls were not simply deleted in the mentioned patch, but since there was no apparent reason for the unnecissarily complicated yaml mechanism, they were replaced by something that was deemed simpler and more robust.
> > ```
> 
> Unnecessarily complicated YAML? Can you elaborate on that?

It is not about yaml being complicated. Having to load information from a file on disk is much more complicated and error-prone than just using internal code structures.

> Who deemed it simpler and robust? Did we have consensus before making the change?

There was not even consensus on the yaml approach. The yaml approach was in the proposed patch. There were questions about why this approach was taken and it was suggested to use go structs instead, but they were unanswered and we had to make some progress. Because we could always revisit and re-discuss we went ahead, having reached consensus between the people who were available at the time. Now we are re-discussing it to increase mutual understanding and end up with the best solution, so that should be fine. :smiley: 

> > ```
> > * It seems rather fragile to want to store yaml files in the container and load them at runtime (unless you want to overwrite them at runtime). Usually having them in go code seems more robust.
> > ```
> 
> Can you explain why it "seems" fragile?

As said above, relying on reading files from disk is obviously more fragile, i.e. gives more chances for something going wrong than simply reading the information from compiled-in code structures. So unless there is a good reason, we should not do the file hassle (irrespective of yaml or other format). 

> Even today OCS runs alerts derived from Rook which stores alerts as YAMLs (which magically reaches their repo) 

Can you point me to the mechanisms? Do you mean "ocs-operator" or the whole OCS system? Where are those alerts?

> and we haven't had a single complaint.

Maybe the reviewers have not reviewed that closely when this was done, or didn't have the same knee-jerk reaction as me when it comes to cleanliness of approach, layering, transparency etc.

> Many OpenShift projects use this same pattern and is well established across several other monitoring projects. Some take a step forward and convert YAML to binary dump in Go and use it.

You should educate us (me, @jarrpa , ...) on the monitoring. The main problem is that we don't have background information and context. So we were just reacting to things that struck us as strange or sub-obtimal. @jarrpa has already set up a meeting for us to discuss. :-)

> > **Questions:**
> > ```
> > * What is the reason for wanting to decouple the monitoring rules from operator resources?
> > ```
> 
>     * We don't want monitoring to get mixed up with Operator.

Why is it in the ocs-operator repo then? :wink: 

>     * We want to be able to work on it independently. Add to it, update it or delete it without any dependencies on OCS operator.

Independence is not achieved when you commit it into the ocs-operator repo. Well at least not independence in both directions, just in one...

>     * We don't want to build OCS every time we need to test out a change and be blocked on Operator bugs or development cycles.

Do you mean "build ocs-operator" or what do you mean with "build OCS"?
How can you test the alerts/rules of ocs-operator without  ocs-operator?

>     * Like we've discussed on other forums,

Not sure which forums - have not heard this.

>  not everyone understands what goes behind building these things,  we don't want anyone making changes to it unknowingly.

If not everyone who should understand it understands it, then we need to educate people. It should be possible. Otherwise these things should not enter this repository, imho.

This sounds a bit as if you do not want the ocs-operator repo maintainers and developers to interfere with the monitoring work, but want them to just accept code or data that comes out of it without asking. This does not seem like a very good collaboration pattern, but rather one-sided or partial.

>     * We don't want the maintainers/reviewers of OCS  who might not have knowledge about monitoring to be burdened with reviewing or merging PRs they don't understand.

But this is precisely what happens when artifacts (yamls) that were created elsewhere is dumped into the repo without any context and background. These are always PRs that no-one only looking at the ocs-operator repo has a chance to understand. If the sources and generation mechanisms are added here, then someone has a chance to understand. Then it is traceable and comprehensible.  :wink: 

IMHO, there are two good solutions for it:

- Don't commit these monitoring pieces here at all, have a separate project for it. (Not sure if that works.)
- Commit them with the sources and generating mechanisms. (or importing the generation mechs as libs)

Two partial solutions would be:

- Commit the yamls, but with very clear documentation in code / doc files where their origin is etc.
- Don't commit the yamls just like that but pull them from their origin, just like e.g. the dependent projects' csvs are also pulled.

> > ```
> > * what are scalable rules?
> > ```
> 
> Currently you see 1 alert and you are already so confused. Consider 10 or 20. Consider having it across namespaces. Consider different configuration parameters that might not be known until later. Consider cases where you need to make a small threshold change in all 20 of these alerts. Consider requiring a subset of these alerts or a mix of alerts from multiple projects.

So it is not a "scalable rule", but expecting to have more than 1 rule in the future. Got it. I was just confused by the term.

> Can you simplify all these with Go without having to develop yet another framework?
> In the middle of release, can you make easy amends to these alerts if you use Go structs?
> If anyone decides they want more alerts than what is provided by operator, how do they get that without rebuilding/redeploying OCS or waiting for the next release?

Sure, we can change the rules in go structs. Maybe it's a bit more cumbersome than yamls, or the sources for the yamls.
The question would be **how often** do we expect changes?

Furthermore, if we generate the yamls from some source, couldn't we generate go structs from source too and thereby avoid the hassle of reading files? If we are generating from another source anyway, the resulting format shouldn't really matter.

The only good reason for files that I see if the users want to change the rules at run time. If we only ever change them at container build time, it could be in (generated) structs as well. So the unanswered question from above remains:

Do we expect that users would want to change the rules at runtime, i.e. without rebuilding contianers?

(If yes, that's a good reason for having a separate file, actually the only good reason I heard so far.)

> > **One more comment:**
> > ```
> > * If the rules are generated from some other souce that is probably simpler than the yaml, then these yamls should not just be dropped into this repository and commited as god-given but they should either be pulled by the build process or the source and the generation mechansim should be in this repo and invoked at build time.
> > ```
> 
> This work is in progress : #740
> All the libraries/framework are god-given. I don't consider this any different :)

With "god-given" I meant that the yamls are just dropped into the repo without any indication about their origin and about the systems behind it. The libraries instead are pulled from their official locations. It's entirely transparent. Very different!
 
The corresponding thing to using libraries would be to import the generation mechanisms as a library from another repo and commit the sources for the rules into this repo and use the imported lib to generate the yamls.


These comments are getting longer and longer. We should really discuss it directly to some extent. But also good to have it written out. :smiley: :smiley: :smiley:

jarrpa (commented):
Okay, code looks good, best I can tell. Now that we have the dummy files in place and the OCP CI config is updated, please rebase this PR

jarrpa (approved):
The new e2e tests pass. Great! LGTM

PR #746, opened 2020-09-03
Fixed cluster failure when Netwok is not specified.

rohan47:
When Network was not specified in the storagecluster CR, the cluster
creation failed as it was checking for some value that was not defined.

### DIFF
--- a/pkg/controller/storagecluster/cephcluster.go
+++ b/pkg/controller/storagecluster/cephcluster.go
@@ -51,7 +51,7 @@ func (r *ReconcileStorageCluster) ensureCephCluster(sc *ocsv1.StorageCluster, re
 		}
 	}
 
-	if sc.Spec.Network.IsMultus() {
+	if isMultus(sc.Spec.Network) {
 		err := validateMultusSelectors(sc.Spec.Network.Selectors)
 		if err != nil {
 			return err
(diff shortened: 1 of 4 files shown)


rohan47:
@leseb @umangachapagain

pkg/controller/storagecluster/cephcluster.go
```diff
@@ -51,10 +51,12 @@ func (r *ReconcileStorageCluster) ensureCephCluster(sc *ocsv1.StorageCluster, re
...
-	if sc.Spec.Network.IsMultus() {
-		err := validateMultusSelectors(sc.Spec.Network.Selectors)
-		if err != nil {
-			return err
+	if sc.Spec.Network != nil {
+		if sc.Spec.Network.IsMultus() {
+			err := validateMultusSelectors(sc.Spec.Network.Selectors)
+			if err != nil {
+				return err
+			}
```
umangachapagain, line 59: This is same as
https://github.com/openshift/ocs-operator/pull/746/files#diff-d59be10a9ed83fc2cbc7d93e0e430388R245-R248

Is it possible to add a common function? If later we forget that there are 2 places to make changes, it'll cause problems.

rohan47: done

pkg/controller/storagecluster/cephcluster.go
```diff
@@ -51,7 +51,7 @@ func (r *ReconcileStorageCluster) ensureCephCluster(sc *ocsv1.StorageCluster, re
 		}
 	}
 
-	if sc.Spec.Network.IsMultus() {
+	if isMultus(sc.Spec.Network) {
```
umangachapagain, line 54: Don't we need to add NetworkSpec to CephCluster if it isMultus?

rohan47: `ensureCephCluster()` returns cephCluster spec only so I didn't put it inside `isMultus()` in order not to change the function definition.

jarrpa (approved):

rohan47:
> This makes sense, thank you! Please move the unrelated whitespace changes to a different commit.

Moved all the whitespace changes except for this one.
I was not able to revert it.
https://github.com/openshift/ocs-operator/pull/746/files#diff-3ab4add16e230d7d70aa431f83ce9551L616

PR #747, opened 2020-09-03
Add more Go linters golangci-lint

crombus:
golangci-lint is the latest tool to consolidate and manage various Go linters. It handles fetching and running for us, so we only have to fetch one tool without relying on system dependencies and we can control them all from a simple config file. This PR replaces the gofmt, govet, and golint linters as well as enabling a few other common ones.


jarrpa:
I added a WIP tag since this is not passing any tests. Should this PR be closed?

hack/golangci_lint.sh
```diff
@@ -0,0 +1,22 @@
...
+test_golangci() {
+        golangci-lint run ./...
+        if [ $? -eq 0 ]; then
+                check=1
+        fi
+}	
+
+golangci-lint --version 
+
+if [ $? -eq 0 ]; then
```
jarrpa, line 16: What is this test for?

jarrpa (changes requested):
This is long overdue, so thank you!

While I appreciate the discipline, I think there are more commits than are strictly necessary, and some of them have errors in how they're split. I suggest the following:

* Merge the first two commits and place it at the end of the patch series
* Merge the last three commits
* Merge the remaining commits to have one commit per file. For each commit:
  * Mention in the commit message that this these are fixes from golint-ci
  * Leave a comment in the code for every instance of `//nolint` explaining why you're skipping the linting

hack/golangci_lint.sh
crombus: just to confirm local system has golangci. then no need to install.

jarrpa: We shouldn't be using a binary on the system, much less installing a binary somewhere else in the system. If possible we should always download the binary to a location within the repo and call that.

Makefile
```diff
@@ -44,6 +44,7 @@ $(call add-crd-gen,ocsv1,./pkg/apis/ocs/v1,./deploy/crds,./deploy/crds)
 	gofmt \
 	golint \
 	govet \
```
jarrpa, line 46: If these linters are included in the metalinter, they should be removed from the Makefile.

jarrpa (changes requested):
A few comments left on the implementation. Also, missing CSV changes.

hack/golangci_lint.sh
crombus: fixed

crombus:
While looking, I found that we can create conf file https://golangci-lint.run/usage/configuration/ .

This is better way then just running golang-ci.

if we create this I can enable gofmt and golint then I can remove those.

hack/golangci_lint.sh
```diff
@@ -0,0 +1,22 @@
...
+test_golangci() {
+        golangci-lint run ./...
+        if [ $? -eq 0 ]; then
+                check=1
+        fi
+}	
+
+echo "Installing golangci-lint in bin"
+
+wget -O- -nv https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s v1.31.0
```
jarrpa, line 16: If I'm understanding this correctly, this installs it into the system's `/bin`, which we don't want to do.

jarrpa:
> While looking, I found that we can create conf file https://golangci-lint.run/usage/configuration/ .
> 
> This is better way then just running golang-ci.
> 
> if we create this I can enable gofmt and golint then I can remove those.

Sure, go for it!

obnoxxx:
@crombus, @jarrpa - recently, some projects in our ecosystem started using the revive tool https://github.com/mgechev/revive for linting, e.g.https://github.com/ceph/go-ceph and the experimentalhttps://github.com/obnoxxx/samba-operator . Just a suggestion since we're talking about changing the listing in ops-operator...

crombus:
> @crombus, @jarrpa - recently, some projects in our ecosystem started using the revive tool https://github.com/mgechev/revive for linting, e.g.https://github.com/ceph/go-ceph and the experimental https://github.com/obnoxxx/samba-operator . Just a suggestion since we're talking about changing the listing in ops-operator...

This looks good if we decide to go with revive in that case we disable golint in the conf and add the revive toml file.

crombus:
@jarrpa most of the requested changes  are done for make golangci-lint it is failing . for local it passed.

hack/golangci_lint.sh
```diff
@@ -0,0 +1,24 @@
...
+test_golangci() {
+        "${OUTDIR_TOOLS}"/golangci-lint run ./...
+        if [ $? -ne 0 ]; then
+                check=1
+        fi
+}	
+
+echo "Installing golangci-lint"
+
+wget -O- -nv https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b "$(go env GOPATH)"/openshift/ocs-operator/build/_output/tools v1.31.0
```
jarrpa, line 18: This is incorrect. You don't want to use GOPATH, you want to use `${OUTDIR_TOOLS}`.

hack/golangci_lint.sh
```diff
@@ -0,0 +1,26 @@
...
+
+# check for golangci-lint in *.go files
+
+check=0
+
+test_golangci() {
+        "${OUTDIR_TOOLS}"/golangci-lint run ./...
+        if [ $? -ne 0 ]; then
+                check=1
+        fi
```
jarrpa, line 13: I'm sorry I didn't catch this earlier, but you shouldn't need to do this. The function `test_golangci` will automatically return the return code of the last statement it executes. Being that there's only one command in this function, it really doesn't need to be it's own function at all. you can just run the command in the main block. I would also recommend setting `set -e` at the top of the file, since if at any point we hit an error it would make sense to exit the script.

jarrpa:
This seems related to the ci failures: https://github.com/golangci/golangci-lint/issues/734

obnoxxx (changes requested):
Can the PR description be updated to explain: What is the problem we are trying to solve. And how we are solving it?

jarrpa (approved):
I updated the PR description. Code and implementation looks good!

/lgtm

PR #748, opened 2020-09-03
Reconcile external resources

aruniiird:
Reconciling external cluster resources by adding a checksum on external secret data.

Whenever external secret data is changed a new checksum is generated, which is then compared with the previously generated checksum. Since they don't match, `ensure` function proceeds forward, thus updating the resources.

### DIFF
--- a/pkg/controller/storagecluster/external_resources.go
+++ b/pkg/controller/storagecluster/external_resources.go
@@ -2,6 +2,7 @@ package storagecluster
 
 import (
 	"context"
+	"crypto/sha512"
 	"encoding/json"
 	"fmt"
 	"net"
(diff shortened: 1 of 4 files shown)


aruniiird:
Tests need to be dusted up a little bit, otherwise ready for initial reviews...
@jarrpa , @umangachapagain , @leseb , please take a look...

pkg/controller/storagecluster/external_resources.go
```diff
@@ -161,16 +196,18 @@ func (r *ReconcileStorageCluster) newExternalCephObjectStoreInstances(
...
 		return nil
 	}
 	err := r.createExternalStorageClusterResources(instance, reqLogger)
 	if err != nil {
 		reqLogger.Error(err, "could not create ExternalStorageClusterResource")
 		return err
 	}
-	instance.Status.ExternalSecretFound = true
+	// as there are no errors, update the external checksum status
+	if externalSecretDataHash, err := r.externalSecretDataChecksum(instance); err == nil {
```
jarrpa, line 208: Should something happen if `err != nil`? I'd think we can't rightly say we succeeded if we can't calculate the new checksum...

jarrpa, line 208: I don't think we should be calling `externalSecretDataChecksum()` here, since it is a separate call to Get() the Secret and the data may have changed since the last call. I think you should just remove the `sameExternalSecretData()` function and do the comparison directly in `ensureExternalStorageClusterResources()`. Make sense?

pkg/controller/storagecluster/external_resources.go
```diff
@@ -193,7 +230,7 @@ func (r *ReconcileStorageCluster) createExternalStorageClusterResources(instance
 	var availableSCs []*storagev1.StorageClass
 	data, err := r.retrieveExternalSecretData(instance, reqLogger)
 	if err != nil {
-		reqLogger.Error(err, "Failed to retrieve external resources")
+		reqLogger.Error(err, "failed to retrieve external resources")
```
jarrpa, line 233: :+1: :+1:

pkg/controller/storagecluster/external_resources_test.go
```diff
@@ -385,3 +395,57 @@ func assertCephObjectStore(t *testing.T, reconciler ReconcileStorageCluster, rem
 		assert.Equal(t, hostFound, cObjS.Spec.Gateway.ExternalRgwEndpoints[0].IP)
 	}
 }
+
+func TestExternalResourceReconcile(t *testing.T) {
+	reconciler := createExternalClusterReconciler(t)
+	assertReconciliationOfExternalResource(t, reconciler)
+}
+
+func assertReconciliationOfExternalResource(t *testing.T, reconciler ReconcileStorageCluster) {
```
jarrpa, line 404: Any reason to define this as a separate function? It doesn't look like it reduces cyclomatic complexity and is only ever called once.

jarrpa (changes requested):
Oh thank you, this is a MUCH better approach!

pkg/controller/storagecluster/external_resources.go
aruniiird: The original idea was, as we have updated all the resources returning an error here doesn't make any sense. In worse case scenario we will be reconciling again and again, which we will do any way (even after sending an error, as we are unable to set the newer checksum into the status).

aruniiird: Yes, it make sense. I have made a minor change, which should address both the concerns (one in the above comment as well).

pkg/controller/storagecluster/external_resources_test.go
aruniiird: This is a common pattern we follow (at least in this file), that is; a `TestXXX` function which has multiple `assertXXX` function calls. As I mentioned, tests are not complete yet and are just written for the proof of concept (that reconciliation is working). It needs to be polished further, so will be changed.

jarrpa: Okay, that's good enough!

jarrpa (approved):
All righty, thanks!

Issue #750, opened 2020-09-04
Cannot generate latest CSV


bipuladh:
Running `make gen-latest-csv` gives the following error in the latest master( commit 45969159f996d7e7369)
```
make gen-latest-csv
Ensuring operator-sdk
hack/ensure-operator-sdk.sh
Using operator-sdk cached at build/_output/tools/operator-sdk-v0.17.0
Generating latest development CSV version using predefined ENV VARs.
hack/generate-latest-csv.sh
=== Generating DEV CSV with the following vars ===
        CSV_VERSION=4.6.0
        ROOK_IMAGE=rook/ceph:v1.4.0.283.g4df2549
        NOOBAA_IMAGE=noobaa/noobaa-operator:2.3.0
        NOOBAA_CORE_IMAGE=noobaa/noobaa-core:5.5.0
        NOOBAA_DB_IMAGE=centos/mongodb-36-centos7
        CEPH_IMAGE=ceph/ceph:v14.2
        OCS_IMAGE=quay.io/ocs-dev/ocs-operator:latest
cp: cannot stat 'deploy/bundlemanifests/*.yaml': No such file or directory
make: *** [Makefile:86: gen-latest-csv] Error 1
```

jarrpa:
Are you working off of a clean master? What does `git status` say?

PR #751, opened 2020-09-04
Fix 'make gen-latest-csv' failure

aruniiird:
'make gen-latest-csv' command fails when 'deploy/bundlemanifests/'
directory dont have any 'yaml' files in it. This is a fix for it.

Fix Issue: https://github.com/openshift/ocs-operator/issues/750

### DIFF
--- a/hack/source-manifests.sh
+++ b/hack/source-manifests.sh
@@ -33,8 +33,8 @@ mkdir -p $OUTDIR_CRDS $OUTDIR_BUNDLEMANIFESTS
 # directory which causes issues in Permissions spec of CSV. So,
 # removing depoy/bundlemanifests/ while generating CSV and keeping a
 # backup in $OUTDIR_BUNDLEMANIFESTS
-cp $BUNDLEMANIFESTS_DIR/*.yaml $OUTDIR_BUNDLEMANIFESTS/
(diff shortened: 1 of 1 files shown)


jarrpa:
...this doesn't seem right. Why would we build the CSV with no role manifests?

jarrpa:
I am noticing that you did not commit the `deploy/csv-template` changes in your previous PR. Can you add those?

bipuladh:
I did in the commit mentioned above. However, after pulling 1c30d14be it seems to be working but two files named `dummy.yaml`  are getting created on their own. (Can be seen in https://github.com/openshift/ocs-operator/pull/754)

jarrpa (changes requested):
There has been no follow up on #750, so I can't accept this as a valid PR. It also still strikes me as wrong that we would even need to do this.

aruniiird:
> There has been no follow up on #750, so I can't accept this as a valid PR. It also still strikes me as wrong that we would even need to do this.

@jarrpa , the line we have changed here is a `copy` statement (which is for taking backup) and the script was failing, without any error message, as there were no files in it. Even though this may be a rare occurrence, IMHO, we should not be failing at a point which could be avoided.

nbalacha:
There is a different PR up for this same problem at https://github.com/openshift/ocs-operator/pull/891

PR #754, opened 2020-09-07
Update CSV to enable Minimal Deployment and Encryption in UI

bipuladh:


tools/csv-merger/csv-merger.go
```diff
@@ -648,7 +648,7 @@ The NooBaa operator deploys and manages the [NooBaa][2] Multi-Cloud Gateway on O
 	// Feature gating for Console. The array values are unique identifiers provided by the console.
 	// This can be used to enable/disable console support for any supported feature
 	// Example: "features.ocs.openshift.io/enabled": `["external", "foo1", "foo2", ...]`
-	ocsCSV.Annotations["features.ocs.openshift.io/enabled"] = `["external"]`
+	ocsCSV.Annotations["features.ocs.openshift.io/enabled"] = `["external", "minimal_deployment", "encryption"]`
```
umangachapagain, line 651:
```suggestion
	ocsCSV.Annotations["features.ocs.openshift.io/enabled"] = `["external", "minimal-deployment", "encryption"]`
```
Can we do this for consistency?

umangachapagain (changes requested):
Remove `deploy/bundlemanifests/dummy.yaml` & `deploy/olm-catalog/ocs-operator/manifests/dummy.yaml`

tools/csv-merger/csv-merger.go
jarrpa: +1

bipuladh: This is what we have used in the UI, changing this would require changes in the UI since it's a string value, can we push this as it is.

umangachapagain: If changing UI is tough, I don't have issues with this either.

umangachapagain:
CI fails because of `Back-off pulling image "default-route-openshift-image-registry.apps.build02.ci.devcluster.openshift.com/ci-op-23q0dsi2/stable:ocs-operator"`. It's a known issue. @jarrpa We need to fix this.

tools/csv-merger/csv-merger.go
jarrpa: Unfortunately, it does seem like it would not be a trivial change, but only because of a matter of process. We'd need a new BZ and it would have to go through QE verification. As much as I hate this, I'm compelled to let it slide. Please make sure to avoid this in the future.

jarrpa (approved):

PR #755, opened 2020-09-07
Fixed typo in yamls

nbalacha:

### DIFF
--- a/deploy/csv-templates/noobaa-csv.yaml
+++ b/deploy/csv-templates/noobaa-csv.yaml
@@ -92,7 +92,7 @@ spec:
         - urn:alm:descriptor:io.kubernetes:Secret
       version: v1alpha1
     - description: Storage target spec such as aws-s3, s3-compatible, ibm-cos, PV's
-        and more. Used in BacketClass to construct data placement policies.
+        and more. Used in BucketClass to construct data placement policies.
       displayName: Backing Store
       kind: BackingStore
       name: backingstores.noobaa.io
--- a/deploy/olm-catalog/ocs-operator/manifests/ocs-operator.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/ocs-operator/manifests/ocs-operator.clusterserviceversion.yaml
@@ -972,7 +972,7 @@ spec:
         - urn:alm:descriptor:io.kubernetes:Secret
       version: v1alpha1
     - description: Storage target spec such as aws-s3, s3-compatible, ibm-cos, PV's
-        and more. Used in BacketClass to construct data placement policies.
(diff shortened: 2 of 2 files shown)


obnoxxx (commented):
Thanks for your PR, @nbalacha .

These files are generated, the content is pulled in from the noobaa-operator. 

This is where it needs to be fixed, afaict: https://github.com/noobaa/noobaa-operator/blob/master/pkg/olm/olm.go#L206

nbalacha:
> Thanks for your PR, @nbalacha .
> 
> These files are generated, the content is pulled in from the noobaa-operator.
> 
> This is where it needs to be fixed, afaict: https://github.com/noobaa/noobaa-operator/blob/master/pkg/olm/olm.go#L206

Thank you @obnoxxx . I have submitted a PR on the nooba-operator.
However, if these are generated files, should they be checked into git for this operator?

obnoxxx:
> > Thanks for your PR, @nbalacha .
> > These files are generated, the content is pulled in from the noobaa-operator.
> > This is where it needs to be fixed, afaict: https://github.com/noobaa/noobaa-operator/blob/master/pkg/olm/olm.go#L206
> 
> Thank you @obnoxxx . I have submitted a PR on the nooba-operator.

Thanks @nbalacha! - https://github.com/noobaa/noobaa-operator/pull/415 - it's merged.

> However, if these are generated files, should they be checked into git for this operator?

Well, we pull these artifacts from dependent components, generate our csv and commit all that. I agree that it would conceptually be better not to check in these pulled and generated files, but there are some reasons, I think for instance being able to autonomously build without pulling from the internet if nothing has changed in the dependent projects or we don't pull a new version. Also being able to just test this in the CI.

@umangachapagain and @jarrpa will be able to explain better.

In order to pull the changes in, you would update the `LATEST_NOOBAA_IMAGE` variable in `hack/common.sh` https://github.com/openshift/ocs-operator/blob/master/hack/common.sh#L56 once the patch is backported to some appropriate version of noobaa. Then you would run `make gen-latest-csv` and commit the resulting changes.

See e.g PR #655 for the latest noobaa-image update. But that was before the introduction of the above variable. Furthermore, that PR has both the change of the version in the source file and the resulting generated changes in the same commit, which is not best practice: We prefer to have the generated changes committed separately to simplify backports. For an update of the rook image with separte commits, you can see that in PR #736.

nbalacha:
Thanks @obnoxxx . Should I close this PR and open a new one then?

obnoxxx:
@nbalacha - up to you, you can also force-push over this and re-use it for the noobaa update. But it might take a bit until we can get it. Not sure.

jarrpa:
Given that it's been over two weeks, I'm going to close this PR.

PR #760, opened 2020-09-10
fixed getPlacement() for custom placement configurations

umangachapagain:
Checking spec.placement length==0 is not required in case
a placement key is not found as there could be other keys
present. So it is enough to fallback to default irrespective
of length in cases where key is not found. Otherwise, it could
cause Nil Pointer Dereference error.

PodAntiAffinity can have many different combinations. This
commit takes it into account while determining placement for
certain components.

Fixes: Bug 1877582


jarrpa (approved):
An excellent PR.

PR #765, opened 2020-09-15
Add test TestSetRookUninstallandCleanupPolicy

iamniting:


raghavendra-talur (approved):

pkg/controller/storagecluster/uninstall_reconciler_test.go
```diff
@@ -49,3 +52,57 @@ func assertStorageClusterUninstallAnnotation(
...
 	}
 }
+
+func TestSetRookUninstallandCleanupPolicy(t *testing.T) {
+
+	for _, eachPlatform := range allPlatforms {
+		cp := &CloudPlatform{platform: eachPlatform}
+		t, reconciler, sc, _ := initStorageClusterResourceCreateUpdateTestWithPlatform(t, cp, nil)
+
+		// there are two annotations which will be 4 combinations, test all 4 combinations
```
obnoxxx, line 62: Can't we parametrize the 4 cases? creating tables (arrays) with the values of the annotations and the inputs for the corresponding assert function call? and then loop over the tables instead of writing all 4 cases out explicitly?

iamniting: Done thanks for the review.

obnoxxx (approved):
thanks for rewriting this, looks good now!

/lgtm

PR #769, opened 2020-09-16
Upgrade build tools to Fedora 33 and Golang 1.15.2

umangachapagain:
Upgrading build tools to latest Fedora release.
Golang 1.15.2 has important security fixes.

### DIFF
--- a/openshift-ci/Dockerfile.tools
+++ b/openshift-ci/Dockerfile.tools
@@ -1,7 +1,6 @@
-FROM fedora:30 as build-tools
+FROM fedora:33 as build-tools
 ENV GOPATH /go
 ENV GOBIN /go/bin
-ENV GOROOT /usr/local/go
 ENV GOCACHE /go/.cache
 ARG GO_PACKAGE_PATH=github.com/openshift/ocs-operator
 
@@ -14,12 +13,10 @@ RUN dnf -y install \
     tar \
     findutils \
     python3 \
+    go \
     && dnf clean all
 
-# dnf install golang pulls go1.12, so manually pulling go1.13.3
-RUN wget https://dl.google.com/go/go1.13.3.linux-amd64.tar.gz
-RUN tar -xzf go1.13.3.linux-amd64.tar.gz && mv go /usr/local
-ENV PATH=$GOPATH/bin:$GOROOT/bin:$PATH
+ENV PATH=$GOPATH:$GOBIN:$PATH
 ENV GO111MODULE=on
 
 # get required golang tools and OC client


umangachapagain:
CI fails because the Dockerfile required for it to pass is in this PR. We can merge this PR as the commits are arranged such that it would pass outside of CI.

jarrpa (commented):
So two major things:

1. I don't see an explanation as to why this is being done, nor any consequences for doing this change.
2. Why shouldn't the CI be expected to pass?

umangachapagain:
> So two major things:
> 
>     1. I don't see an explanation as to why this is being done, nor any consequences for doing this change.
>
There are no consequences. We're already doing it downstream.
I didn't think we needed an explanation to upgrade Go to higher version (it gives us some important bug fixes which is obvious for any upgrade).
>
>     2. Why shouldn't the CI be expected to pass?
>
Because this PR contains the fix for CI and the running CI tests are using Dockerfile from master.
Let me know if you have any idea on how to resolve it differently.

jarrpa:
> > So two major things:
> > 1. I don't see an explanation as to why this is being done, nor any consequences for doing this change.
> 
> There are no consequences. We're already doing it downstream.
> I didn't think we needed an explanation to upgrade Go to higher version (it gives us some important bug fixes which is obvious for any upgrade).

Right, but usually we only do it when there's a particular bug fix or incompatibility with another project. If there's no such thing, I don't see the need.

> > 2. Why shouldn't the CI be expected to pass?
> 
> Because this PR contains the fix for CI and the running CI tests are using Dockerfile from master.
> Let me know if you have any idea on how to resolve it differently.

Looking at what we already have, the Makefile only specifies a minimum version, so developers could be using a later version than the one used in the CI. Could we create a separate PR to update the tools Dockerfile, maybe with a bump of the fedora version, and see if we can get something that gives us >1.15.2 without having to be explicit about it? Then we could merge the Makefile change with this one.

raghavendra-talur:
@jarrpa @umangachapagain I am not sure if RHEL/CentOS has 1.15 of golang yet. Should we wait to bump up the min version?

iamniting (approved):

umangachapagain:
> @jarrpa @umangachapagain I am not sure if RHEL/CentOS has 1.15 of golang yet. Should we wait to bump up the min version?

Downstream builds are already using 1.15. I don't think we need to wait.

umangachapagain:
@jarrpa I've made the switch to Fedora 33 and Golang 1.15.2.
We can drop the Makefile changes for now.

jarrpa:
Okay, I like what I see. One last thing, please add the sentence `Golang 1.15.2 has important security fixes.` to the commit message as well, not just the PR description.

obnoxxx (approved):
Last requests have been addressed.

/lgtm

PR #773, opened 2020-09-21
Introduce flexible scaling

nbalacha:
Allow the cluster to be expanded by one host at a time

### DIFF
--- a/pkg/apis/ocs/v1/storagecluster_types.go
+++ b/pkg/apis/ocs/v1/storagecluster_types.go
@@ -40,6 +40,10 @@ type StorageClusterSpec struct {
 	// ManagedResources specifies how to deal with auxiliary resources reconciled
 	// with the StorageCluster
 	ManagedResources ManagedResourcesSpec `json:"managedResources,omitempty"`
+	// If enabled, sets the failureDomain to host, allowing devices to be
+	// distributed evenly across all nodes, regardless of distribution in zones
+	// or racks.
+	FlexibleScaling  bool     `json:"flexibleScaling,omitempty"`
 }
 
 // ManagedResourcesSpec defines how to reconcile auxiliary resources
--- a/pkg/controller/storagecluster/cephcluster.go
+++ b/pkg/controller/storagecluster/cephcluster.go
@@ -346,8 +346,11 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster, serverVersion *version.
 			resources = defaults.DaemonResources["osd"]
 		}
 
+		portable := ds.Portable
+
 		topologyKey := ds.TopologyKey
 		topologyKeyValues := []string{}
+
 		noPlacement := ds.Placement.NodeAffinity == nil && ds.Placement.PodAffinity == nil && ds.Placement.PodAntiAffinity == nil
 		noPreparePlacement := ds.PreparePlacement.NodeAffinity == nil && ds.PreparePlacement.PodAffinity == nil && ds.PreparePlacement.PodAntiAffinity == nil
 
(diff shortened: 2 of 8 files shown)


nbalacha:
Starting this off in order to get reviews early. This does not have the TCS changes yet.

jarrpa (changes requested):
Thank you for this! Unfortunately the branch is in a very unclean state.

* The commits still need to be cleaned up.
* You should not be including any changes to the default/managed resources, so please remove those.
* The TSC changes should already be merged, so you probably need a rebase I'm guessing?

jarrpa (changes requested):
Ah, I missed something... don't change the number of StorageClassDeviceSets we generate. That needs to come later after we make a change in Rook-ceph.

pkg/controller/storagecluster/cephcluster.go
```diff
@@ -379,7 +387,7 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster, serverVersion *version.
 			// enable the new behavior while the console is waiting
 			// to be updated.
 			// TODO: Remove this behavior when OCP console is updated
-			count = count / 3
+			count = count / defaults.DeviceSetReplica
```
jarrpa, line 390: This doesn't seem related to this PR, please remove it.

pkg/controller/storagecluster/topology.go
```diff
@@ -264,6 +269,7 @@ func (r *ReconcileStorageCluster) reconcileNodeTopologyMap(sc *ocsv1.StorageClus
 		}
 	}
 
+
```
jarrpa, line 272: Remove this newline.

pkg/apis/ocs/v1/storagecluster_types.go
```diff
@@ -40,6 +40,8 @@ type StorageClusterSpec struct {
 	// ManagedResources specifies how to deal with auxiliary resources reconciled
 	// with the StorageCluster
 	ManagedResources ManagedResourcesSpec `json:"managedResources,omitempty"`
+
+	FlexibleScaling bool            `json:"flexibleScaling,omitempty"`
```
jarrpa, line 44: This needs a comment explaining what the field enables.

jarrpa (changes requested):
Okay, this is much better, and makes sense. Aside from a few small changes, this also needs a test to verify the new functionality works correctly. Add a new test case to `TestFailureDomain()` for flexible scaling.

pkg/controller/storagecluster/cephcluster.go
nbalacha: Ok. I'll send it in a different PR.

pkg/controller/storagecluster/topology_test.go
```diff
@@ -246,7 +271,16 @@ func TestFailureDomain(t *testing.T) {
...
+			label: "Case 3", // storagecluster has predefined failure domain of `host`
+			storageCluster: &api.StorageCluster{
+				Status: api.StorageClusterStatus{
+					FailureDomain: "host",
+				},
+			},
+			expectedFailureDomain: "host",
+		},
+		{
+			label: "Case 4", // storagecluster with three or more zone topology labels
```
jarrpa, line 283: Add this as case 5. I see what you're going for, but here I would value the cleaner patch and chronological order over other logical grouping of test cases.

pkg/controller/storagecluster/topology_test.go
```diff
@@ -278,6 +312,32 @@ func TestFailureDomain(t *testing.T) {
 			},
 			expectedFailureDomain: "rack",
 		},
+		{
+			label: "Case 6", // storagecluster with three or more zone topology labels
```
jarrpa, line 316: Isn't this the same as case 4?

jarrpa (changes requested):

jarrpa (approved):

PR #761, opened 2020-09-11
Move the uninstall code from reconcile.go to uninstall_reconcile.go to organize it in better way.

iamniting:
It should be merged after https://github.com/openshift/ocs-operator/pull/753

PR #781, opened 2020-09-23
Bug 1760426: [release-4.6] Move the uninstall code from reconcile.go to uninstall_reconcile.go to organize it in better way.

raghavendra-talur:
This is a cherrypick of #761


obnoxxx (approved):
/lgtm

this should merge after #781

PR #789, opened 2020-09-28
Remove StorageClusterInitialization

jarrpa:
We want finer control over which resources we reconcile and which we don't. This PR removes the need for a StorageClusterInitialization CR by implementing a set of string to control how a given type of resource is reconciled. The default is set to only reconcile if it doesn't exist, to maintain compatibility with current behavior on upgrade.


pkg/controller/storagecluster/reconcile.go
```diff
@@ -25,6 +25,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
+// ReconcileStratery is a string representing how we want to reconcile
+// (or not) a particular resource
+type ReconcileStratery string
```
obnoxxx, line 30: `s/ReconcileStatery/ReconcileStrategy/` everywhere?

obnoxxx (changes requested):
General direction looks great.
First thing that caught my eye: `ReconcileStratery` ...

pkg/controller/storagecluster/reconcile.go
jarrpa: HAH! At least the typo was consistent ;D Fixed.

obnoxxx: yeah, it built...

obnoxxx (approved):
/lgtm

thanks for the update

PR #794, opened 2020-09-30
UnitTests: Add test TestDeleteStorageClasses

iamniting:
Add test for deleteStorageClasses in uninstall_reconciler_test.go

cc: @raghavendra-talur

### DIFF
--- a/pkg/controller/storagecluster/uninstall_reconciler_test.go
+++ b/pkg/controller/storagecluster/uninstall_reconciler_test.go
@@ -7,6 +7,8 @@ import (
 	api "github.com/openshift/ocs-operator/pkg/apis/ocs/v1"
 	cephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
 	"github.com/stretchr/testify/assert"
+	storagev1 "k8s.io/api/storage/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 )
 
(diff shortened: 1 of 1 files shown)


jarrpa (changes requested):
Looks good, the tests make sense. Please implement the test function using case tables.

iamniting:
> Looks good, the tests make sense. Please implement the test function using case tables.

Hi, Jose thanks for looking into it. I did not get what you are saying. Can you pls elaborate?

iamniting:
> > Looks good, the tests make sense. Please implement the test function using case tables.
> 
> Hi, Jose thanks for looking into it. I did not get what you are saying. Can you pls elaborate?

I guess I understood it and changed it.

raghavendra-talur (approved):

pkg/controller/storagecluster/uninstall_reconciler_test.go
```diff
@@ -102,3 +104,30 @@ func assertCephClusterCleanupPolicy(
 	assert.Equal(t, CleanupPolicyConfirmation, cephCluster.Spec.CleanupPolicy.Confirmation)
 	assert.Equal(t, AllowUninstallWithVolumes, cephCluster.Spec.CleanupPolicy.AllowUninstallWithVolumes)
 }
+
+func TestDeleteStorageClasses(t *testing.T) {
```
umangachapagain, line 108: Use Table Driven approach and simplify the code a little bit.

umangachapagain (changes requested):

jarrpa:
> > > Looks good, the tests make sense. Please implement the test function using case tables.
> > 
> > 
> > Hi, Jose thanks for looking into it. I did not get what you are saying. Can you pls elaborate?
> 
> I guess I understood it and changed it.

Sadly, no. :) As Umanga also says, follow the pattern of using a struct that is a list of test cases, as is done here: https://github.com/openshift/ocs-operator/pull/797/files

iamniting:
> > > > Looks good, the tests make sense. Please implement the test function using case tables.
> > > 
> > > 
> > > Hi, Jose thanks for looking into it. I did not get what you are saying. Can you pls elaborate?
> > 
> > 
> > I guess I understood it and changed it.
> 
> Sadly, no. :) As Umanga also says, follow the pattern of using a struct that is a list of test cases, as is done here: https://github.com/openshift/ocs-operator/pull/797/files

It should be resolved now

pkg/controller/storagecluster/uninstall_reconciler_test.go
iamniting: done

pkg/controller/storagecluster/uninstall_reconciler_test.go
```diff
@@ -102,3 +104,53 @@ func assertCephClusterCleanupPolicy(
...
+
+	testList := []struct {
+		isStorageClassExistBeforedelete bool
+		isStorageClassExistAfterdelete  bool
+	}{
+		// case 1: verify storage classes are present and delete them
+		{true, false},
+		// case 2: verify storage classes does not exist and delete should not get error out
+		{false, false},
+	}
```
jarrpa, line 118: This is the right general idea, but it is too condensed to be immediately readable. I suggest something like this:

```suggestion
	cases := []struct {
		label string
		storageClassExists bool
	}{
		{
			label: "case 1", // verify storage classes are present and delete them
			storageClassExists: true,
		},
		{
			label: "case 2", // verify storage classes does not exist and delete should not get error out
			storageClassExists: false,
		},
	}
```

jarrpa (changes requested):
An improvement, but I think the test itself is a little too convoluted. Rename `isStorageClassExistBeforedelete` to `storageClassExists` and remove `isStorageClassExistAfterdelete` (just test for `assert.True(t, errors.IsNotFound(err))`.

jarrpa (approved):

nbalacha (commented):
/lgtm

PR #796, opened 2020-09-30
[RHSTOR-1428] Introduce new controller for updating PersistentVolumes

jarrpa:
This PR introduces a new controller for managing the volume attributes of Ceph-CSI PersistentVolumes. As part of the implementation, is also introduces the use of Predicate from the Kubernetes controller-runtime. This allows the controller to filter resource events, only queuing those which match all predicate conditions.

Alongside this, this PR implements the use of Predicates with the StorageCluster controller as well, both as an enhancement and a means to verify the use of Predicates in the CI.

### DIFF
--- a/deploy/olm-catalog/ocs-operator/manifests/ocs-operator.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/ocs-operator/manifests/ocs-operator.clusterserviceversion.yaml
@@ -1341,6 +1341,16 @@ spec:
           verbs:
           - get
         - apiGroups:
+          - ""
+          resources:
+          - persistentvolumes
+          verbs:
+          - get
+          - list
+          - watch
+          - update
+          - patch
+        - apiGroups:
           - apps
           resources:
           - deployments
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -35,6 +35,16 @@ rules:
   verbs:
   - get
 - apiGroups:
+  - ""
+  resources:
+  - persistentvolumes
+  verbs:
+  - get
+  - list
+  - watch
+  - update
+  - patch
+- apiGroups:
   - apps
   resources:
   - deployments
--- a/pkg/controller/persistentvolume/persistentvolume_controller.go
+++ b/pkg/controller/persistentvolume/persistentvolume_controller.go
@@ -1,0 +1,123 @@
+package persistentvolume
+
+import (
+	"strings"
+
+	"github.com/openshift/ocs-operator/pkg/controller/util"
+
+	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
+	csiExpansionSecretName      = "csi.storage.k8s.io/controller-expand-secret-name"
+	csiExpansionSecretNamespace = "csi.storage.k8s.io/controller-expand-secret-namespace"
+	csiRBDDriverSuffix          = ".rbd.csi.ceph.com"
+	csiCephFSDriverSuffix       = ".cephfs.csi.ceph.com"
+)
+
+var log = logf.Log.WithName("controller_persistentvolume")
+
+// Add creates a new PersistentVolume Controller and adds it to the Manager.
+// The Manager will set fields on the Controller and Start it when the Manager
+// is Started.
+func Add(mgr manager.Manager) error {
+	return add(mgr, newReconciler(mgr))
+}
+
+// newReconciler returns a new reconcile.Reconciler
+func newReconciler(mgr manager.Manager) reconcile.Reconciler {
+	r := &ReconcilePersistentVolume{
+		client:    mgr.GetClient(),
+		scheme:    mgr.GetScheme(),
+		reqLogger: log,
+	}
+
+	return r
+}
+
+// reconcilePV determines whether we want to reconcile a given PV
+func reconcilePV(obj runtime.Object) bool {
+	pv, ok := obj.(*corev1.PersistentVolume)
+	if !ok {
+		return false
+	}
+
+	drivers := []string{
+		csiRBDDriverSuffix,
+		csiCephFSDriverSuffix,
+	}
+
+	for _, driver := range drivers {
+		if pv.Spec.CSI != nil && strings.HasSuffix(pv.Spec.CSI.Driver, driver) {
+			if pv.Spec.StorageClassName != "" {
+				secretRef := pv.Spec.CSI.ControllerExpandSecretRef
+				if secretRef == nil || secretRef.Name == "" {
+					return true
+				}
+			}
+			return false
+		}
+	}
+
+	return false
+}
+
+var pvPredicate = predicate.Funcs{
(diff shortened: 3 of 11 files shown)


pkg/controller/persistentvolume/reconcile.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+	sc := &storagev1.StorageClass{}
+
+	err := r.client.Get(context.TODO(), types.NamespacedName{Name: scName}, sc)
+	if err != nil {
+		reqLogger.Error(err, "Error getting StorageClass")
+		return err
+	}
+
+	patch := client.MergeFrom(pv)
```
humblec, line 67: Before reconciling, we have to check the retrieved SC has the `AllowVolumeExpansion` set to `true.` @jarrpa

pkg/controller/persistentvolume/reconcile.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+	if secretNamespace, ok := sc.Parameters[csiExpansionSecretNamespace]; ok {
+		secretRef.Namespace = secretNamespace
+	}
+
+	newPV := &corev1.PersistentVolume{}
+	pv.DeepCopyInto(newPV)
+
+	newPV.Spec.CSI.ControllerExpandSecretRef = secretRef
+	err = r.client.Patch(context.TODO(), newPV, patch)
```
humblec, line 82: Cant we patch existing PV object ?

deploy/olm-catalog/ocs-operator/manifests/ocs-operator.clusterserviceversion.yaml
```diff
@@ -1240,6 +1240,7 @@ spec:
           - pods
           - services
           - endpoints
+          - persistentvolumes
```
humblec, line 1243: Does ocs operator already have the RBAC for patching PV?

deploy/role.yaml
```diff
@@ -21,6 +21,7 @@ rules:
   - pods
   - services
   - endpoints
+  - persistentvolumes
```
humblec, line 24: Looks like we are very relaxed on RBAC ( All verbs on above objects . I dont think thats necessary) which is not a good practise in general.

pkg/controller/persistentvolume/persistentvolume_controller.go
```diff
@@ -0,0 +1,113 @@
...
+func reconcilePV(obj runtime.Object) bool {
+	pv, ok := obj.(*corev1.PersistentVolume)
+	if !ok {
+		return false
+	}
+
+	if pv.Spec.CSI != nil && strings.Contains(pv.Spec.CSI.Driver, "csi.ceph.com") {
+		secretRef := pv.Spec.CSI.ControllerExpandSecretRef
+		if secretRef == nil || secretRef.Name == "" {
+			if pv.Spec.StorageClassName != "" {
```
humblec, line 56: Please move SC check to above, we dont want to come down or validate secretRef or its entry if SC is nil

jarrpa: Sure, makes sense.

pkg/controller/persistentvolume/reconcile.go
jarrpa: No we don't. By virtue of being OCS we know if our version of the Ceph-CSI drivers support expansion or not. Based on that, we have to account for the scenario where a PV references a SC (which does not allow expansion) that gets deleted and recreated to allow expansion, by which time (not after) the secret ref needs to already be present.

pkg/controller/persistentvolume/reconcile.go
jarrpa: Yes, but only for adding fields that aren't already set.

deploy/role.yaml
jarrpa: I won't change the RBAC for other objects, but I can implement a more restricted set of verbs for PVs specifically.

pkg/controller/persistentvolume/persistentvolume_controller.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+// reconcilePV
+func reconcilePV(obj runtime.Object) bool {
+	pv, ok := obj.(*corev1.PersistentVolume)
+	if !ok {
+		return false
+	}
+
+	if pv.Spec.CSI != nil && strings.Contains(pv.Spec.CSI.Driver, "csi.ceph.com") {
```
Madhu-1, line 53: can this cause issue if some driver comes with csi.ceph.com in their driver name? as we have the exact driver names in ocs-operator can we do full name comparison?

jarrpa: We don't have the exact driver names in this controller. The full driver names are generated from the Rook-Ceph operator's namespace. We don't have access to the Rook-Ceph operator, since we don't want to assume the name of the operator's Deployment. Still, I can make it a little more explicit by specifying `rbd` and `cephfs`. If we could have the suffixes exposed via the go code, that would be even better, but right now they're just strings in the code: https://github.com/rook/rook/blob/662bde6d4a7cdd51090bea8bb73b314376f47ad2/pkg/operator/ceph/csi/spec.go#L220-L221

Madhu-1: ocs-operator creates default cephfs and rbd storageclass and snapshotter for the csi drivers . Is it possible to do same thing here?

Madhu-1: Okay got it.

jarrpa: Addressed.

pkg/controller/persistentvolume/persistentvolume_controller.go
jarrpa: Addressed.

deploy/role.yaml
jarrpa: Addressed.

pkg/controller/persistentvolume/persistentvolume_controller.go
```diff
@@ -0,0 +1,123 @@
...
+		return false
+	}
+
+	drivers := []string{
+		csiRbdDriverSuffix,
+		csiCephfsDriverSuffix,
+	}
+
+	for _, driver := range drivers {
+		if pv.Spec.CSI != nil && strings.Contains(pv.Spec.CSI.Driver, driver) {
```
Madhu-1, line 62: use `strings.HasSuffix` ?

jarrpa: Good idea

jarrpa: Done

pkg/controller/persistentvolume/persistentvolume_controller.go
```diff
@@ -0,0 +1,123 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
+	csiExpansionSecretName      = "csi.storage.k8s.io/controller-expand-secret-name"
+	csiExpansionSecretNamespace = "csi.storage.k8s.io/controller-expand-secret-namespace"
+	csiRbdDriverSuffix          = ".rbd.csi.ceph.com"
```
umangachapagain, line 25:
```suggestion
	csiRBDDriverSuffix          = ".rbd.csi.ceph.com"
```

pkg/controller/persistentvolume/persistentvolume_controller.go
```diff
@@ -0,0 +1,123 @@
...
+	csiCephfsDriverSuffix       = ".cephfs.csi.ceph.com"
```
umangachapagain, line 26:
```suggestion
	csiCephFSDriverSuffix       = ".cephfs.csi.ceph.com"
```

umangachapagain (changes requested):

pkg/controller/persistentvolume/persistentvolume_controller.go
jarrpa: Done

obnoxxx (approved):
Latest requests by @umangachapagain have been addressed. We seem to be all set!

/lgtm

PR #804, opened 2020-10-05
refactor: helper functions to generate placement structs

rajatsing:
- Added helper functions to generate placement structs and added more function params to change values for a specific struct. 

- This will make each struct more flexible and will also reduce the complexity of structures that need to be written by hand.

### DIFF
--- a/pkg/controller/defaults/placements.go
+++ b/pkg/controller/defaults/placements.go
@@ -11,229 +11,121 @@ var (
 	appLabelSelectorKey = "app"
 	// DefaultNodeAffinity is the NodeAffinity to be used when labelSelector is nil
 	DefaultNodeAffinity = &corev1.NodeAffinity{
-		RequiredDuringSchedulingIgnoredDuringExecution: &corev1.NodeSelector{
-			NodeSelectorTerms: []corev1.NodeSelectorTerm{
-				corev1.NodeSelectorTerm{
-					MatchExpressions: []corev1.NodeSelectorRequirement{
-						corev1.NodeSelectorRequirement{
-							Key:      NodeAffinityKey,
-							Operator: corev1.NodeSelectorOpExists,
-						},
-					},
-				},
-			},
-		},
+		RequiredDuringSchedulingIgnoredDuringExecution: getOcsNodeSelector(),
 	}
 	// DaemonPlacements map contains the default placement configs for the
 	// various OCS daemons
 	DaemonPlacements = map[string]rook.Placement{
 		"all": rook.Placement{
 			Tolerations: []corev1.Toleration{
-				corev1.Toleration{
-					Key:      NodeTolerationKey,
-					Operator: corev1.TolerationOpEqual,
-					Value:    "true",
-					Effect:   corev1.TaintEffectNoSchedule,
-				},
+				getOcsToleration(),
 			},
 		},
 
 		"mon": rook.Placement{
 			PodAntiAffinity: &corev1.PodAntiAffinity{
 				PreferredDuringSchedulingIgnoredDuringExecution: []corev1.WeightedPodAffinityTerm{
-					corev1.WeightedPodAffinityTerm{
-						Weight: 100,
-						PodAffinityTerm: corev1.PodAffinityTerm{
-							LabelSelector: &metav1.LabelSelector{
-								MatchExpressions: []metav1.LabelSelectorRequirement{
-									metav1.LabelSelectorRequirement{
-										Key:      appLabelSelectorKey,
-										Operator: metav1.LabelSelectorOpIn,
-										Values:   []string{"rook-ceph-mon"},
-									},
-								},
-							},
-							TopologyKey: corev1.LabelHostname,
-						},
-					},
+					getWeightedPodAffinityTerm(100, "rook-ceph-mon"),
 				},
 			},
 		},
 
 		"osd": rook.Placement{
 			Tolerations: []corev1.Toleration{
-				corev1.Toleration{
-					Key:      NodeTolerationKey,
-					Operator: corev1.TolerationOpEqual,
-					Value:    "true",
-					Effect:   corev1.TaintEffectNoSchedule,
-				},
+				getOcsToleration(),
 			},
 			PodAntiAffinity: &corev1.PodAntiAffinity{
 				PreferredDuringSchedulingIgnoredDuringExecution: []corev1.WeightedPodAffinityTerm{
-					corev1.WeightedPodAffinityTerm{
-						Weight: 100,
-						PodAffinityTerm: corev1.PodAffinityTerm{
-							LabelSelector: &metav1.LabelSelector{
-								MatchExpressions: []metav1.LabelSelectorRequirement{
-									metav1.LabelSelectorRequirement{
-										Key:      appLabelSelectorKey,
-										Operator: metav1.LabelSelectorOpIn,
-										Values:   []string{"rook-ceph-osd"},
-									},
-								},
-							},
-							TopologyKey: corev1.LabelHostname,
-						},
-					},
+					getWeightedPodAffinityTerm(100, "rook-ceph-osd"),
 				},
 			},
 		},
 
 		"osd-prepare": rook.Placement{
 			Tolerations: []corev1.Toleration{
-				corev1.Toleration{
-					Key:      NodeTolerationKey,
-					Operator: corev1.TolerationOpEqual,
-					Value:    "true",
-					Effect:   corev1.TaintEffectNoSchedule,
-				},
+				getOcsToleration(),
 			},
 			PodAntiAffinity: &corev1.PodAntiAffinity{
 				PreferredDuringSchedulingIgnoredDuringExecution: []corev1.WeightedPodAffinityTerm{
-					corev1.WeightedPodAffinityTerm{
-						Weight: 100,
-						PodAffinityTerm: corev1.PodAffinityTerm{
-							LabelSelector: &metav1.LabelSelector{
-								MatchExpressions: []metav1.LabelSelectorRequirement{
-									metav1.LabelSelectorRequirement{
-										Key:      appLabelSelectorKey,
-										Operator: metav1.LabelSelectorOpIn,
-										Values:   []string{"rook-ceph-osd-prepare"},
-									},
-								},
-							},
-							TopologyKey: corev1.LabelHostname,
-						},
-					},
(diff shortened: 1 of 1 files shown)


pkg/controller/defaults/placements.go
```diff
@@ -248,3 +91,70 @@ func getTopologySpreadConstraintsSpec(selectorValue string) []corev1.TopologySpr
 
 	return topologySpreadConstraints
 }
+
+func setWeightedPodAffinityTerm(selector string) []corev1.WeightedPodAffinityTerm {
```
jarrpa, line 95:
```suggestion
func getWeightedPodAffinityTerm(selector string) []corev1.WeightedPodAffinityTerm {
```

pkg/controller/defaults/placements.go
```diff
@@ -248,3 +91,70 @@ func getTopologySpreadConstraintsSpec(selectorValue string) []corev1.TopologySpr
...
+					},
+				},
+				TopologyKey: corev1.LabelHostname,
+			},
+		},
+	}
+	return WeightedPodAffinityTerm
+}
+
+func setPodAffinityTerm(selector string) []corev1.PodAffinityTerm {
```
jarrpa, line 116:
```suggestion
func getPodAffinityTerm(selector string) []corev1.PodAffinityTerm {
```

pkg/controller/defaults/placements.go
```diff
@@ -248,3 +91,70 @@ func getTopologySpreadConstraintsSpec(selectorValue string) []corev1.TopologySpr
...
+					},
+				},
+			},
+			TopologyKey: corev1.LabelHostname,
+		},
+	}
+	return podAffinityTerm
+}
+
+func setToleration() []corev1.Toleration {
```
jarrpa, line 134:
```suggestion
func getOcsToleration() []corev1.Toleration {
```

pkg/controller/defaults/placements.go
```diff
@@ -248,3 +91,70 @@ func getTopologySpreadConstraintsSpec(selectorValue string) []corev1.TopologySpr
...
+			Key:      NodeTolerationKey,
+			Operator: corev1.TolerationOpEqual,
+			Value:    "true",
+			Effect:   corev1.TaintEffectNoSchedule,
+		},
+	}
+	return toleration
+}
+
+func setNodeSelector() *corev1.NodeSelector {
```
jarrpa, line 146:
```suggestion
func getOcsNodeSelector() *corev1.NodeSelector {
```

jarrpa (changes requested):
Looks handy! The function names could use some work. Also, try and avoid generic commit descriptions. Try something like "refactor: helper functions to generate placement structs"

pkg/controller/defaults/placements.go
rajatsing: Done!

pkg/controller/defaults/placements.go
rajatsing: Done

pkg/controller/defaults/placements.go
rajatsing: Done

pkg/controller/defaults/placements.go
rajatsing: Done

umangachapagain (changes requested):
I don't see any value add to this.

I prefer duplicate code (for something as static as this) rather than a bunch of function calls.
We could just define global vars/constants and re-use it (if duplication is a concern)? Saves a lot of calls.

Unless I'm completely missing the point, I don't agree with this refactor as it is.

rajatsing:
@umangachapagain  above is [disscussed here](https://github.com/openshift/ocs-operator/pull/727/files#r485077430) and [here](https://github.com/openshift/ocs-operator/pull/727/files#r485076961)

jarrpa:
Why is reducing duplication not inherently valuable? I believe we've already run into scenarios where we forgot to update some components when we decided to implement a change that was supposed to affect all components.

The functions also allow us to generate structs that are largely identical save for one field. If  we did it with vars/consts we'd have large structs that were largely the same. Again, why is this not valuable?

rajatsing:
@umangachapagain do you have any reviews in it , otherwise @jarrpa  can go ahead and merge this!

umangachapagain:
> Why is reducing duplication not inherently valuable?

It's never inherently valuable. Even if it is, it's negligible in this case.

> I believe we've already run into scenarios where we forgot to update some components when we decided to implement a change that was supposed to affect all components.

I don't see this re-factor helping us fix those issues. Even with this re-factor one can mess it up (vs the variables which was pretty obvious).

What about when we want to make changes exclusive to a single component? We then add a bunch of if..else which will then blow out of proportion? 

> The functions also allow us to generate structs that are largely identical save for one field. If we did it with vars/consts we'd have large structs that were largely the same.

But, is it always going to be identical? Is there no possibility that we'd maybe like to configure placement for each component differently (other than the name)?

> Again, why is this not valuable?

It's not "not valuable". It's just not significantly valuable. Here's why I think so

+ it doesn't significantly simplify the code
+ it doesn't make it easier to custom configure each component (we get what the function throws at us)
+ retrieving values would require less cycles then generating values
+ this file is still largely static. The configuration would probably change once in a release or two (functions just feels like an overkill)

I'm not against this re-factor, but I believe it's half way there.
It reduced duplication at the cost of configurability and extra processing cycle.
If it can give me configurability, I'd be more than happy with it. But, that's just me.

Since it's not a very big change, I'm not blocking this with my comment.
I'd like the value add of this refactor to be explained (PR and Commit descriptions) and justified rather than just calling it a "refactor".

umangachapagain:
> @umangachapagain above is [disscussed here](https://github.com/openshift/ocs-operator/pull/727/files#r485077430) and [here](https://github.com/openshift/ocs-operator/pull/727/files#r485076961)

Can't reach those comments.

pkg/controller/defaults/placements.go
```diff
@@ -11,219 +11,62 @@ var (
...
-										Operator: metav1.LabelSelectorOpIn,
-										Values:   []string{"rook-ceph-mon"},
-									},
-								},
-							},
-							TopologyKey: corev1.LabelHostname,
-						},
-					},
-				},
+				PreferredDuringSchedulingIgnoredDuringExecution: getWeightedPodAffinityTerm("rook-ceph-mon"),
```
jarrpa, line 25:
```suggestion
				PreferredDuringSchedulingIgnoredDuringExecution: []corev1.WeightedPodAffinityTerm{
					getAppWeightedPodAffinityTerm(100, "rook-ceph-mon"),
				}
```

pkg/controller/defaults/placements.go
```diff
@@ -248,3 +91,70 @@ func getTopologySpreadConstraintsSpec(selectorValue string) []corev1.TopologySpr
 
 	return topologySpreadConstraints
 }
+
+func getWeightedPodAffinityTerm(selector string) []corev1.WeightedPodAffinityTerm {
```
jarrpa, line 95:
```suggestion
func getAppWeightedPodAffinityTerm(weight int, selector string) corev1.WeightedPodAffinityTerm {
```

pkg/controller/defaults/placements.go
```diff
@@ -11,219 +11,62 @@ var (
...
-										Operator: metav1.LabelSelectorOpIn,
-										Values:   []string{"rook-ceph-rgw"},
-									},
-								},
-							},
-							TopologyKey: corev1.LabelZoneFailureDomain,
-						},
-					},
-				},
+				RequiredDuringSchedulingIgnoredDuringExecution:  getPodAffinityTerm("rook-ceph-rgw"),
```
jarrpa, line 56:
```suggestion
				RequiredDuringSchedulingIgnoredDuringExecution:  []corev1.PodAffinityTerm{
					getAppPodAffinityTerm("rook-ceph-rgw"),
				}
```

jarrpa (changes requested):
I agree with a number of Umanga's comments. To start, explain in your commit message (and PR description) what this refactor gains us. Specifically, I see it as reducing the complexity of structures that need to be written by hand.

I also agree that this should probably be more configurable to be more meaningful. As such, I'd like you to take any functions that return a list and only return a single element that would go in that list. I've provided an example in a comment below.

rajatsing:
I considered your changes @jarrpa and I think if I start making changes that you suggested, it will make code more complicated and in turn, the whole point of "refactoring" will be of no use.
I think It's better to close it?.
@umangachapagain wdyt?.

jarrpa:
I completely disagree. Refactoring is more than just simple code. I hate seeing the same exact framework or even exact configuration duplicated in multiple places. While Umanga has a point that these things don't change very frequently (if some would at all), that is no reason to just let the error potential sit around unaddressed. It's all just BZs waiting to come up in QE.

pkg/controller/defaults/placements.go
rajatsing: added, please check.

rajatsing: added, please check!.

pkg/controller/defaults/placements.go
rajatsing: added, please check!.

pkg/controller/defaults/placements.go
rajatsing: added, please check!.

rajatsing:
@jarrpa @umangachapagain 
I have added more specific commit and description. Also, based on the suggestion above, I have modified the code in a way that each struct can be modified based on their own individual's value.

jarrpa (changes requested):
Code looks good, now merge the two commits. :) We don't need to know how you reacted to the review comments in the commit history, just have a clean commit that can be understood (and reverted later if needed).

rajatsing:
> Code looks good, now merge the two commits. :) We don't need to know how you reacted to the review comments in the commit history, just have a clean commit that can be understood (and reverted later if needed).

Absolutely!. When I was reviewing the code myself, I forgot about to review the older commit sadly. But nvm, I'll squash them and form a new commit.

rajatsing:
Commit fixed!.

rajatsing:
I hate it when CI runs on local but fails on remote 👎 .

jarrpa (changes requested):
One last thing I failed to catch earlier: Instead of taking a `string` and setting it as an element of a `[]string{}`, take `...string` and set it directly as the full list.

jarrpa (approved):
LGTM, thank you!

PR #819, opened 2020-10-12
Fix Unbalanced distribution of OSD pods by TSC for more than 3 AZs

### LINKED ISSUE #817: StorageCluster: Unbalaced distribution of OSDs by Topology spread on cluster with more than 3 AZs (zones, rack)
On a cluster with more than 3 AZs, the OSD pod distribution is not even when on expanding the storage. 

During expansion, some of the OSDs fails to schedule on the new AZs. 

Steps to reproduce

1. Create a 6 node OCP cluster based on 6 AZs (east-us 1a,1b,1c,1d,1e,1f)
2. Create initial OCS cluster selecting all the nodes to get 3 OSDs
3. Expand the storage once to get total of 6 OSDs

Expected result : The OSD should be equally distributed across the 6 AZs each with one OSD

kesavanvt:
Fixes #817 

### DIFF
--- a/pkg/controller/defaults/placements.go
+++ b/pkg/controller/defaults/placements.go
@@ -141,7 +141,7 @@ var (
 					Effect:   corev1.TaintEffectNoSchedule,
 				},
 			},
-			TopologySpreadConstraints: getTopologySpreadConstraintsSpec("rook-ceph-osd-prepare"),
+			TopologySpreadConstraints: getTopologySpreadConstraintsSpec("rook-ceph-osd-prepare", "rook-ceph-osd"),
 		},
 
 		"rgw": rook.Placement{
@@ -228,7 +228,7 @@ var (
 	}
 )
 
-func getTopologySpreadConstraintsSpec(selectorValue string) []corev1.TopologySpreadConstraint {
+func getTopologySpreadConstraintsSpec(selectorValue ...string) []corev1.TopologySpreadConstraint {
 	topologySpreadConstraints := []corev1.TopologySpreadConstraint{
 		corev1.TopologySpreadConstraint{
 			MaxSkew:           1,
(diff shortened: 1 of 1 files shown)


iamniting (approved):

travisn (commented):
This change looks good for now, though the two phase scheduling of OSDs will still have corner issues even with this approach. For example, even during initial OSD creation, if some OSD prepare jobs are completed before all of them are scheduled, then may not end up balanced. This would be more likely to happen in a cluster with many OSDs. The only solution I can think of is for the prepare jobs to say in the running state until the operator instructs them to complete, which will be a much bigger change for a future release... @kesavanvt Want to open an issue on Rook for this? thanks

kesavanvt:
> This change looks good for now, though the two phase scheduling of OSDs will still have corner issues even with this approach. For example, even during initial OSD creation, if some OSD prepare jobs are completed before all of them are scheduled, then may not end up balanced. This would be more likely to happen in a cluster with many OSDs. The only solution I can think of is for the prepare jobs to say in the running state until the operator instructs them to complete, which will be a much bigger change for a future release... @kesavanvt Want to open an issue on Rook for this? thanks

I agree on this and expectt there will be an imbalance if some OSD prepare jobs are completed before all of them are scheduled. I haven't faced this scenario till now just because i was deploying/expanding with default no of OSDs (i.e) 3 OSDs.  I will proceed on testing by deploying more OSDs at a time and see how its behaving with current changes.

pkg/controller/defaults/placements.go
```diff
@@ -228,7 +228,7 @@ var (
 	}
 )
 
-func getTopologySpreadConstraintsSpec(selectorValue string) []corev1.TopologySpreadConstraint {
+func getTopologySpreadConstraintsSpec(selectorValue []string) []corev1.TopologySpreadConstraint {
```
jarrpa, line 231: Rather than taking an explicit `[]string`, can you take `string...`?

iamniting: +1

kesavanvt: thanks, done

jarrpa (approved):

PR #820, opened 2020-10-12
ceph: Add missing fields in osd purge job

xenolinux:
Add ROOK_MON_ENDPOINTS, VolumeMounts, Volumes and other required fields in the OCS osd-purge job spec
to match the job spec with rook job https://github.com/rook/rook/blob/master/cluster/examples/kubernetes/ceph/osd-purge.yaml

### DIFF
--- a/pkg/controller/storagecluster/reconcile.go
+++ b/pkg/controller/storagecluster/reconcile.go
@@ -593,16 +593,90 @@ func newCleanupJob(sc *ocsv1.StorageCluster) *batchv1.Job {
 				Spec: corev1.PodSpec{
 					RestartPolicy:      corev1.RestartPolicyNever,
 					ServiceAccountName: "rook-ceph-system",
+					Volumes: []corev1.Volume{
+						{
+							Name:         "ceph-conf-emptydir",
+							VolumeSource: corev1.VolumeSource{EmptyDir: &corev1.EmptyDirVolumeSource{}},
+						},
+						{
+							Name:         "rook-config",
+							VolumeSource: corev1.VolumeSource{EmptyDir: &corev1.EmptyDirVolumeSource{}},
+						},
+					},
+
 					Containers: []corev1.Container{
 						{
 							Name:  "operator",
 							Image: os.Getenv("ROOK_CEPH_IMAGE"),
-							Command: []string{
+							Args: []string{
 								"ceph",
 								"osd",
 								"remove",
 								"--osd-ids=${FAILED_OSD_IDS}",
-								"--namespace=" + sc.Namespace,
+							},
+							VolumeMounts: []corev1.VolumeMount{
+								{
+									Name:      "ceph-conf-emptydir",
+									MountPath: "/etc/ceph",
+								},
+								{
+									Name:      "rook-config",
+									MountPath: "/var/lib/rook",
+								},
+							},
+							Env: []corev1.EnvVar{
+								{
+									Name: "ROOK_MON_ENDPOINTS",
+									ValueFrom: &corev1.EnvVarSource{
+										ConfigMapKeyRef: &corev1.ConfigMapKeySelector{
+											Key:                  "data",
+											LocalObjectReference: corev1.LocalObjectReference{Name: "rook-ceph-mon-endpoints"},
+										},
+									},
+								},
+								{
+									Name:  "POD_NAMESPACE",
+									Value: sc.Namespace,
+								},
+								{
+									Name: "ROOK_CEPH_USERNAME",
+									ValueFrom: &corev1.EnvVarSource{
+										SecretKeyRef: &corev1.SecretKeySelector{
+											Key:                  "ceph-username",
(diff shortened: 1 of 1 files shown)


pkg/controller/storagecluster/reconcile.go
```diff
@@ -597,6 +614,17 @@ func newCleanupJob(sc *ocsv1.StorageCluster) *batchv1.Job {
 							},
 						},
 					},
+					Env: []corev1.EnvVar{
```
travisn, line 617: What about all the env variables in [this example](https://github.com/rook/rook/blob/release-1.4/cluster/examples/kubernetes/ceph/osd-purge.yaml#L18-L48)? Are they not needed?

pkg/controller/storagecluster/reconcile.go
```diff
@@ -584,10 +584,27 @@ func newCleanupJob(sc *ocsv1.StorageCluster) *batchv1.Job {
...
+					},
+
 					Containers: []corev1.Container{
 						{
 							Name:  "operator",
 							Image: os.Getenv("ROOK_CEPH_IMAGE"),
+							VolumeMounts: []corev1.VolumeMount{
+								{
+									MountPath: "/var/lib/rook",
+									Name:      "rook-config",
```
travisn, line 601: Should this be an env var instead of a volume like in the [rook example](https://github.com/rook/rook/blob/release-1.4/cluster/examples/kubernetes/ceph/osd-purge.yaml#L18-L48)?

travisn (changes requested):

pkg/controller/storagecluster/reconcile.go
xenolinux: Added other variables to the code

pkg/controller/storagecluster/reconcile.go
```diff
@@ -595,6 +602,62 @@ func newCleanupJob(sc *ocsv1.StorageCluster) *batchv1.Job {
 								"--osd-ids=${FAILED_OSD_IDS}",
 								"--namespace=" + sc.Namespace,
 							},
+							VolumeMounts: []corev1.VolumeMount{
+								{
+									Name:      "ceph-conf-emptydir",
+									MountPath: "/etc/ceph",
+								},
+							},
+							Env: []corev1.EnvVar{
```
travisn, line 611: Is `POD_NAMESPACE` not needed? It's in the rook upstream example.

pkg/controller/storagecluster/reconcile.go
xenolinux: Yes. Updated

pkg/controller/storagecluster/reconcile.go
xenolinux: They will be needed

pkg/controller/storagecluster/reconcile.go
xenolinux: Updated...

travisn (commented):
LGTM, just pending to hear how your testing goes.

xenolinux:
> LGTM, just pending to hear how your testing goes.

Yes. I will update the result

travisn (commented):
The new changes look good in the template, but why all the other changes? They aren't related to this fix.

xenolinux:
Update: The code is tested and is working properly.

obnoxxx:
ocs-operator-ci works for me locally, retriggered the test

obnoxxx:
same error. strange

xenolinux:
> The new changes look good in the template, but why all the other changes? They aren't related to this fix.

Ack. Removed those auto-generated files.

xenolinux:
> ocs-operator-ci works for me locally, retriggered the test

It's working locally for me as well!

travisn (approved):

pkg/controller/storagecluster/reconcile.go
```diff
@@ -593,16 +593,90 @@ func newCleanupJob(sc *ocsv1.StorageCluster) *batchv1.Job {
...
+							VolumeSource: corev1.VolumeSource{EmptyDir: &corev1.EmptyDirVolumeSource{}},
+						},
+					},
+
 					Containers: []corev1.Container{
 						{
 							Name:  "operator",
 							Image: os.Getenv("ROOK_CEPH_IMAGE"),
-							Command: []string{
+							Args: []string{
```
jarrpa, line 611: Why are you switching to `Args` instead of `Command`? Shouldn't you ideally be using both?

travisn: By not specifying the `Command`, it will use the default container entrypoint, which is the same pattern used by the rook operator.

xenolinux: When the job was triggered using `command`, it was failing to run the ceph commands with this error message- `Error initializing cluster client: ObjectNotFound('RADOS object not found (error calling conf_read_file)',)`.

jarrpa (approved):

PR #824, opened 2020-10-13
Use the defaults.DeviceSetReplica value when dividing the count

nbalacha:
If replica==0, the code sets the replica to defaults.DeviceSetReplica and then divides the count by 3 (the value of defaults.DeviceSetReplica). It is better to use the same value in both cases instead of a hardcoded value.


jarrpa:
Something's wrong, there shouldn't be ten commits in this PR... nonetheless, I don't think we should do this. If we still have StorageClusters created with `replica` 0 that rely on this behavior and then change the internal default to something other than 3, it will break their clusters.

nbalacha:
> Something's wrong, there shouldn't be ten commits in this PR... nonetheless, I don't think we should do this. If we still have StorageClusters created with `replica` 0 that rely on this behavior and then change the internal default to something other than 3, it will break their clusters.

Strange - I didn't notice the 10 commits. Will clean those up.
The problem here is that we are setting the replica value to defaults.DeviceSetReplica (whose value is 3) but dividing by the hardcoded value 3. This could lead to very different values if the 2 don't match. This could end up with a number of storageClassDeviceSets which do not equal to count. if we are going to manipulate the data, we should use the same value throughout.

However, the larger problem is that count may not be divisible by 3 in which case we have an incorrect value anyway.

obnoxxx:
I see a merge master commit here. I recommend using `git rebase` instead of `git merge`.
You should fetch current origin, rebase on current upstream master, resulting in just a single patch on top (for all I can tell from the commits).

PR #833, opened 2020-10-14
[release-4.6] Bug 1873580: Remove StorageClusterInitialization

jarrpa:
We want finer control over which resources we reconcile and which we don't. This PR removes the need for a StorageClusterInitialization CR by implementing a set of string to control how a given type of resource is reconciled. The default is set to only reconcile if it doesn't exist, to maintain compatibility with current behavior on upgrade.

This PR includes backports for #789, #811, #822, and #834


obnoxxx (approved):
/lgtm

thanks for the backport!

obnoxxx (approved):
/lgtm

obnoxxx (approved):
/lgtm

Now it should be complete... :-)

PR #835, opened 2020-10-14
Use the defaults.DeviceSetReplica value everywhere

nbalacha:
instead of a hardcoded value

### DIFF
--- a/pkg/controller/storagecluster/cephcluster.go
+++ b/pkg/controller/storagecluster/cephcluster.go
@@ -387,7 +387,7 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster, serverVersion *version.
 			// enable the new behavior while the console is waiting
 			// to be updated.
 			// TODO: Remove this behavior when OCP console is updated
-			count = count / 3
+			count = count / defaults.DeviceSetReplica
 		}
 
 		for i := 0; i < replica; i++ {


jarrpa:
> The problem here is that we are setting the replica value to defaults.DeviceSetReplica (whose value is 3) but dividing by the hardcoded value 3. This could lead to very different values if the 2 don't match. This could end up with a number of storageClassDeviceSets which do not equal to count. if we are going to manipulate the data, we should use the same value throughout.
> 
> However, the larger problem is that count may not be divisible by 3 in which case we have an incorrect value anyway.

Yes, that is a known problem. And we have to live with it. Again, we can not change the expected behavior because there are StorageClusters from OCS 4.2 (and I think OCS 4.3?) that have `replica = 0` (from when they were created from older versions of the OCP console) and depend on that value *always* being 3. By using the variable, we could conceivably change it away from 3, thus introducing problems for those old clusters.

The only way around this would be to somehow forcefully update the StorageCluster.Spec of all those old clusters, which we should not do. Essentially, we're stuck with this until we change our API version.

nbalacha:
> > The problem here is that we are setting the replica value to defaults.DeviceSetReplica (whose value is 3) but dividing by the hardcoded value 3. This could lead to very different values if the 2 don't match. This could end up with a number of storageClassDeviceSets which do not equal to count. if we are going to manipulate the data, we should use the same value throughout.
> > However, the larger problem is that count may not be divisible by 3 in which case we have an incorrect value anyway.
> 
> Yes, that is a known problem. And we have to live with it. Again, we can not change the expected behavior because there are StorageClusters from OCS 4.2 (and I think OCS 4.3?) that have `replica = 0` (from when they were created from older versions of the OCP console) and depend on that value _always_ being 3. By using the variable, we could conceivably change it away from 3, thus introducing problems for those old clusters.
> 
> The only way around this would be to somehow forcefully update the StorageCluster.Spec of all those old clusters, which we should not do. Essentially, we're stuck with this until we change our API version.

In that case shouldn't we not set the replica to 3 instead of defaults.DeviceSetReplica? My concern is that we are using different values for the 2 assignments. We should use the same for both.

raghavendra-talur (approved):

jarrpa:
> In that case shouldn't we not set the replica to 3 instead of defaults.DeviceSetReplica? My concern is that we are using different values for the 2 assignments. We should use the same for both.

For newer clusters we don't want to hardcode the value directly, so the defaults value makes sense. Really the larger problem we have is that we can't allow the replica count to change after the device set has been created, so effectively we currently can't ever change the default anyway! :joy: I don't want to introduce even the potential uncertainty for clusters with an older configuration until we can address that larger problem.

obnoxxx:
> > In that case shouldn't we not set the replica to 3 instead of defaults.DeviceSetReplica? My concern is that we are using different values for the 2 assignments. We should use the same for both.
> 
> For newer clusters we don't want to hardcode the value directly, so the defaults value makes sense. Really the larger problem we have is that we can't allow the replica count to change after the device set has been created, so effectively we currently can't ever change the default anyway! joy I don't want to introduce even the potential uncertainty for clusters with an older configuration until we can address that larger problem.

So has this been resolved in any way? .... :thinking: 

putting a 

/hold 

on this until it's clear

nbalacha:
> > > In that case shouldn't we not set the replica to 3 instead of defaults.DeviceSetReplica? My concern is that we are using different values for the 2 assignments. We should use the same for both.
> > 
> > 
> > For newer clusters we don't want to hardcode the value directly, so the defaults value makes sense. Really the larger problem we have is that we can't allow the replica count to change after the device set has been created, so effectively we currently can't ever change the default anyway! joy I don't want to introduce even the potential uncertainty for clusters with an older configuration until we can address that larger problem.
> 
> So has this been resolved in any way? .... 
> 
> putting a
> 
> /hold
> 
> on this until it's clear

I would still like to use the same value for both lines - either a hardcoded 3 or the default value. The current approach is prone to errors.

jarrpa:
> I would still like to use the same value for both lines - either a hardcoded 3 or the default value. The current approach is prone to errors.

I still argue that this opens us up for more errors, since if the default ever changes (say we make it configurable via environment variables in the operator) we have changed the behavior of how the CephCluster is generated for an existing cluster which then introduces a problem. I think what we want is a check in an admission controller to prevent modification of this field in an Update. After that's in place I'd be fine with taking in this PR.

PR #843, opened 2020-10-15
noobaa must-gather - Add important notifications collection

nb-ohad:
- Add creation of a `notificaiton.txt` file to the noobaa must-gather flow
- Add a notification for the situation where noobaa is deployed in the cluster but it is not managed by the storagecluster CR

### DIFF
--- a/must-gather/collection-scripts/gather_noobaa_resources
+++ b/must-gather/collection-scripts/gather_noobaa_resources
@@ -56,3 +56,13 @@ do
         { timeout 120 oc -n "${ns}" logs --all-containers "${pod}" &> "${LOG_DIR}"/"${pod}".log; } >> "${BASE_COLLECTION_PATH}"/gather-debug.log 2>&1
     done
 done
+
+# Add important notfications to a notifications.txt file at the root of the noobaa collection path
+UNMANAGED_NOOBAA_NOTIF="The noobaa deployed on this cluster is not managed by the storagecluster CR and will not react to configuration changes made to storagecluster CR"
+
+NOOBAA_RECONCILE_STRATEGY=$( oc -n ${ns} get storagecluster -o jsonpath='{.items[0].spec.multiCloudGateway.reconcileStrategy}' 2> /dev/null)
+if [ "${NOOBAA_RECONCILE_STRATEGY}" = "ignore" ] || [ "${NOOBAA_RECONCILE_STRATEGY}" = "standalone" ]; then 
+    if [ "$(oc -n ${ns} get noobaa -o name 2> /dev/null) " ]; then 
+        echo "- ${UNMANAGED_NOOBAA_NOTIF}" >> ${NOOBAA_COLLLECTION_PATH}/notifications.txt
+    fi
+fi


must-gather/collection-scripts/gather_noobaa_resources
```diff
@@ -56,3 +56,15 @@ do
         { timeout 120 oc -n "${ns}" logs --all-containers "${pod}" &> "${LOG_DIR}"/"${pod}".log; } >> "${BASE_COLLECTION_PATH}"/gather-debug.log 2>&1
     done
 done
+
+# Add important notfications to a notifications.txt file at the root of the noobaa collection path
+UNMANAGED_NOOBAA_NOTIF="The noobaa deployed on this cluster is not managed by the storagecluster CR and will not react to configuration changes made to the CR"
+
+NOOBAA_RECONCILE_STRATEGY=$( oc -n ${ns} get storagecluster -o jsonpath='{.items[0].spec.multiCloudGateway.reconcileStrategy}' 2> /dev/null)
+if [ "${NOOBAA_RECONCILE_STRATEGY}" ] && [ "${NOOBAA_RECONCILE_STRATEGY}" != "managed" ]
```
jarrpa, line 64: Do we need both of these tests? Wouldn't the second one already cover the first since the first will only ever fail when it's `""`, which is `!= "managed"`?

must-gather/collection-scripts/gather_noobaa_resources
```diff
@@ -56,3 +56,15 @@ do
...
+then
```
jarrpa, line 65:
```suggestion
if [ "${NOOBAA_RECONCILE_STRATEGY}" ] && [ "${NOOBAA_RECONCILE_STRATEGY}" != "manage" ]; then 
```

must-gather/collection-scripts/gather_noobaa_resources
```diff
@@ -56,3 +56,15 @@ do
...
 done
+
+# Add important notfications to a notifications.txt file at the root of the noobaa collection path
+UNMANAGED_NOOBAA_NOTIF="The noobaa deployed on this cluster is not managed by the storagecluster CR and will not react to configuration changes made to the CR"
+
+NOOBAA_RECONCILE_STRATEGY=$( oc -n ${ns} get storagecluster -o jsonpath='{.items[0].spec.multiCloudGateway.reconcileStrategy}' 2> /dev/null)
+if [ "${NOOBAA_RECONCILE_STRATEGY}" ] && [ "${NOOBAA_RECONCILE_STRATEGY}" != "managed" ]
+then 
+    if [ "$(oc -n ${ns} get noobaa -o name 2> /dev/null) " ]
+    then
```
jarrpa, line 67:
```suggestion
    if [ "$(oc -n ${ns} get noobaa -o name 2> /dev/null) " ]; then 
```

must-gather/collection-scripts/gather_noobaa_resources
```diff
@@ -56,3 +56,15 @@ do
         { timeout 120 oc -n "${ns}" logs --all-containers "${pod}" &> "${LOG_DIR}"/"${pod}".log; } >> "${BASE_COLLECTION_PATH}"/gather-debug.log 2>&1
     done
 done
+
+# Add important notfications to a notifications.txt file at the root of the noobaa collection path
+UNMANAGED_NOOBAA_NOTIF="The noobaa deployed on this cluster is not managed by the storagecluster CR and will not react to configuration changes made to the CR"
```
jarrpa, line 61:
```suggestion
UNMANAGED_NOOBAA_NOTIF="The noobaa deployed on this cluster is not managed by the storagecluster CR and will not react to configuration changes"
```

jarrpa (changes requested):
Makes perfect sense. One question and a few style nits. Note that the configured option is `manage` and not `managed`.

must-gather/collection-scripts/gather_noobaa_resources
jarrpa: Actually, wait, this is even easier. You need to check for `standalone` or `ignore`, not `manage`. Anything that is not one of those two will be treated the same as `manage` (the implicit default).

must-gather/collection-scripts/gather_noobaa_resources
nb-ohad: It will react to config changes on the Noobaa CR, I want this to be very specific.

must-gather/collection-scripts/gather_noobaa_resources
nb-ohad: Ok, if this is the styling guide

must-gather/collection-scripts/gather_noobaa_resources
nb-ohad: The condition here is `if a value exists and the value is not manage` 
If we remove the first section we will get `if the value is not manage` 
which differ under the case where the value does not exist

must-gather/collection-scripts/gather_noobaa_resources
nb-ohad: done

must-gather/collection-scripts/gather_noobaa_resources
nb-ohad: done

must-gather/collection-scripts/gather_noobaa_resources
nb-ohad: I changed the ending to: `... will not react to configuration changes made to storagecluster CR`

must-gather/collection-scripts/gather_noobaa_resources
jarrpa: Yes, you don't want that. All you want is "is the value `ignore` or `standalone`?". Everything that is not those two words is equivalent to "manage".

nb-ohad: done

jarrpa (approved):
/ok-to-test

PR #848, opened 2020-10-19
Bug 1888614: Changed the execution of connection closing, to avoid closing uncreated connection.

anmolsachan:
Signed-off-by: anmol sachan <anmol13694@gmail.com>


nbalacha (commented):
IIUC, the issue here is that the con was not created because the endpoint was unreachable so trying to access con.Close() it will fail. The defer statement is not the problem - it is the incorrect error handling in this case.

Can you update the commit message to add more detail on the cause?

anmolsachan:
> IIUC, the issue here is that the con was not created because the endpoint was unreachable so trying to access con.Close() it will fail. The defer statement is not the problem - it is the incorrect error handling in this case.
> 
Yes, That is correct.
> Can you update the commit message to add more detail on the cause?
Sure.

anmolsachan:
@nbalacha Done. Thanks for the review.

jarrpa (approved):

PR #849, opened 2020-10-19
Update rook module and image

raghavendra-talur:
Update the rook dependency and the image to a version which has the fix to prevent a crash during the uninstall procedure.

PR #850, opened 2020-10-19
fix ceph cluster delete ordering

raghavendra-talur:
Rook ceph requires that the deletion request for the cephCluster should
    come before the delete request for the other ceph resources. Not doing
    so can result in some of the ceph resources being stuck in deletion
    phase forever.

In the reconciler, we considered a pending uninstall operation as an
    error. It resulted in slower reconciliation because of exponential
    backoff. It is sufficient if the reconciler returns with a requeue
    request but not an error.


raghavendra-talur:
@jarrpa @obnoxxx Merge this after #849 

cc @iamniting

iamniting (approved):

pkg/controller/storagecluster/reconcile.go
```diff
@@ -188,7 +189,8 @@ func (r *ReconcileStorageCluster) Reconcile(request reconcile.Request) (reconcil
 		if contains(instance.GetFinalizers(), storageClusterFinalizer) {
 			err = r.deleteResources(instance, reqLogger)
 			if err != nil {
-				return reconcile.Result{}, err
+				reqLogger.Info("Uninstall in progress", "Status", err)
+				return reconcile.Result{RequeueAfter: time.Second * time.Duration(1)}, nil
```
jarrpa, line 193: FINALLY, a different use of a `reconcile.Result{}`! :joy:

jarrpa (approved):

PR #855, opened 2020-10-20
Bug 1886873: fix ceph cluster delete ordering

raghavendra-talur:
This is a backport of #849 and #850 .

Manually created, I reran the `make deps-update` and `make gen-latest-csv` commands because I wasn't sure if we can cherrypick the generated changes. The result looks identical though.


raghavendra-talur:
/retitle Bug 1886873: fix ceph cluster delete ordering

jarrpa (approved):

raghavendra-talur:
@jarrpa The tests are failing due to what looks like a test infra error `  netifaces.c:1:10: fatal error: Python.h: No such file or directory`.  How to debug?

go.mod
```diff
@@ -23,7 +23,7 @@ require (
 	github.com/operator-framework/operator-sdk v0.17.0
 	github.com/prometheus/client_golang v1.7.1
 	github.com/prometheus/client_model v0.2.0
-	github.com/rook/rook v1.4.3-0.20200901124914-0b665e6e3315
+	github.com/rook/rook v1.4.6
```
travisn, line 26: Is this used in the downstream builds? v1.4.6 has diverged a bit from what we're using downstream. Some fixes have been backported to the 1.4 branch upstream, but not to the downstream 4.6 branch. OCS should really be referencing the downstream rook branch under github.com/openshift/rook, but it's really a separate topic than for this PR.

PR #859, opened 2020-10-22
Add new property in storagedeviceset

crombus:
add new property in storagedeviceset

DeviceType is the property which will allow admin to add the device type in Storagecluster CR.

### DIFF
--- a/deploy/crds/ocs.openshift.io_storageclusters_crd.yaml
+++ b/deploy/crds/ocs.openshift.io_storageclusters_crd.yaml
@@ -1463,6 +1463,19 @@ spec:
                             phase:
                               description: Phase represents the current phase of PersistentVolumeClaim.
                               type: string
+                    deviceType:
+                      description: DeviceType is the value of device type in this
+                        StorageDeviceSet. It can have one of the three values (SSD,
+                        HDD, NVMe)
+                      type: string
+                      enum:
+                      - SSD
+                      - ssd
+                      - HDD
+                      - hdd
+                      - NVMe
+                      - NVME
+                      - nvme
                     metadataPVCTemplate:
                       description: PersistentVolumeClaim is a user's request for and
                         claim to a persistent volume
--- a/deploy/csv-templates/crds/ocs.openshift.io_storageclusters_crd.yaml
+++ b/deploy/csv-templates/crds/ocs.openshift.io_storageclusters_crd.yaml
@@ -1463,6 +1463,19 @@ spec:
                             phase:
                               description: Phase represents the current phase of PersistentVolumeClaim.
                               type: string
+                    deviceType:
+                      description: DeviceType is the value of device type in this
+                        StorageDeviceSet. It can have one of the three values (SSD,
+                        HDD, NVMe)
+                      type: string
+                      enum:
+                      - SSD
+                      - ssd
+                      - HDD
+                      - hdd
+                      - NVMe
+                      - NVME
+                      - nvme
                     metadataPVCTemplate:
                       description: PersistentVolumeClaim is a user's request for and
                         claim to a persistent volume
--- a/deploy/olm-catalog/ocs-operator/manifests/storagecluster.crd.yaml
+++ b/deploy/olm-catalog/ocs-operator/manifests/storagecluster.crd.yaml
@@ -1456,6 +1456,19 @@ spec:
                               type: string
                           type: object
                       type: object
+                    deviceType:
+                      description: DeviceType is the value of device type in this
+                        StorageDeviceSet. It can have one of the three values (SSD,
+                        HDD, NVMe)
+                      enum:
+                      - SSD
+                      - ssd
+                      - HDD
+                      - hdd
+                      - NVMe
+                      - NVME
+                      - nvme
+                      type: string
                     metadataPVCTemplate:
                       description: PersistentVolumeClaim is a user's request for and
                         claim to a persistent volume
--- a/pkg/apis/ocs/v1/storagecluster_types.go
+++ b/pkg/apis/ocs/v1/storagecluster_types.go
@@ -99,6 +99,13 @@ type StorageDeviceSet struct {
 	// +optional
 	Replica int `json:"replica,omitempty"`
 
+	// DeviceType is the value of device type in
+	// this StorageDeviceSet. It can have one of the
+	// three values (SSD, HDD, NVMe)
+	// +kubebuilder:validation:Enum=SSD;ssd;HDD;hdd;NVMe;NVME;nvme
+	// +optional
+	DeviceType string `json:"deviceType,omitempty"`
+
 	// TopologyKey is the Kubernetes topology label that the
 	// StorageClassDeviceSets will be distributed across. Ignored if
 	// Placement is set
--- a/pkg/controller/storagecluster/reconcile.go
+++ b/pkg/controller/storagecluster/reconcile.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/blang/semver"
@@ -62,6 +63,15 @@ osd_memory_target_cgroup_limit_ratio = 0.5
 	ReconcileStrategyManage ReconcileStrategy = "manage"
 	// ReconcileStrategyStandalone also means never reconcile (NooBaa)
 	ReconcileStrategyStandalone ReconcileStrategy = "standalone"
+
+	// DeviceTypeSSD represents the DeviceType SSD
+	DeviceTypeSSD = "ssd"
+
+	// DeviceTypeHDD represents the DeviceType HDD
+	DeviceTypeHDD = "hdd"
+
+	// DeviceTypeNVMe represents the DeviceType NVMe
+	DeviceTypeNVMe = "nvme"
 )
 
 var storageClusterFinalizer = "storagecluster.ocs.openshift.io"
@@ -405,6 +415,13 @@ func (r *ReconcileStorageCluster) validateStorageDeviceSets(sc *ocsv1.StorageClu
 		if ds.MetadataPVCTemplate != nil {
 			if ds.MetadataPVCTemplate.Spec.StorageClassName == nil || *ds.MetadataPVCTemplate.Spec.StorageClassName == "" {
 				return fmt.Errorf("failed to validate StorageDeviceSet %d: no StorageClass specified for metadataPVCTemplate", i)
+			}
+		}
+		if ds.DeviceType != "" {
+			if ( DeviceTypeSSD == strings.ToLower(ds.DeviceType )) || ( DeviceTypeHDD == strings.ToLower(ds.DeviceType )) || ( DeviceTypeNVMe == strings.ToLower(ds.DeviceType )) {
+				metav1.SetMetaDataAnnotation(&sc.ObjectMeta, "crushDeviceClass", ds.DeviceType)
+			} else {
+				return fmt.Errorf("failed to validate DeviceType %q: no Device of this type", ds.DeviceType)
 			}
 		}
 	}


nbalacha (changes requested):
There is a typo in the commit message - "stroageclusterdeviceset"
I also think it would be better to say storageDeviceSet instead of storageClusterDeviceSet (which does not exist).

pkg/controller/storagecluster/reconcile.go
```diff
@@ -406,6 +406,11 @@ func (r *ReconcileStorageCluster) validateStorageDeviceSets(sc *ocsv1.StorageClu
 				return fmt.Errorf("failed to validate StorageDeviceSet %d: no StorageClass specified for metadataPVCTemplate", i)
 			}
 		}
+		if ds.DeviceType != "" {
+			if ds.DeviceType == "SSD" || ds.DeviceType == "HDD" || ds.DeviceType == "NVMe" {
```
vbnrh, line 410: ignore the case.

vbnrh (changes requested):

deploy/crds/ocs.openshift.io_storageclusters_crd.yaml
```diff
@@ -1447,6 +1447,9 @@ spec:
                           phase:
                             description: Phase represents the current phase of PersistentVolumeClaim.
                             type: string
+                  devicetype:
+                    description: DeviceType is the value of device type in this StorageDeviceSet
```
vbnrh, line 1451: DeviceType represents the value of underlying storage device used (restricted to 3 values; SSD, NVME, HDD)

deploy/csv-templates/crds/ocs.openshift.io_storageclusters_crd.yaml
```diff
@@ -1447,6 +1447,9 @@ spec:
                           phase:
                             description: Phase represents the current phase of PersistentVolumeClaim.
                             type: string
+                  devicetype:
+                    description: DeviceType is the value of device type in this StorageDeviceSet
```
vbnrh, line 1451: DeviceType represents the value of underlying storage device used (restricted to 3 values; SSD, NVME, HDD)

deploy/olm-catalog/ocs-operator/manifests/storagecluster.crd.yaml
```diff
@@ -1440,6 +1440,9 @@ spec:
                             type: string
                         type: object
                     type: object
+                  devicetype:
+                    description: DeviceType is the value of device type in this StorageDeviceSet
```
vbnrh, line 1444: DeviceType represents the value of underlying storage device used (restricted to 3 values; SSD, NVME, HDD)

vbnrh (changes requested):

pkg/controller/storagecluster/reconcile.go
```diff
@@ -406,6 +407,11 @@ func (r *ReconcileStorageCluster) validateStorageDeviceSets(sc *ocsv1.StorageClu
 				return fmt.Errorf("failed to validate StorageDeviceSet %d: no StorageClass specified for metadataPVCTemplate", i)
 			}
 		}
+		if ds.DeviceType != "" {
+			if strings.EqualFold(ds.DeviceType, "SSD") || strings.EqualFold(ds.DeviceType, "HDD") || strings.EqualFold(ds.DeviceType, "NVMe") {
```
iamniting, line 411: It looks like you wanted to use a not condition here as per the error msg.

iamniting, line 411: It is good to define these values as types and use it here

iamniting (changes requested):

pkg/controller/storagecluster/reconcile.go
crombus: yes. I'll change that.

crombus: can you explain more what types and how can we ensure to ignore case insensitivity.

iamniting: something like this https://github.com/openshift/ocs-operator/blob/master/pkg/controller/storagecluster/uninstall_reconciler.go#L24-L43

crombus: why would I want to use this approach. when I want to use devicetype only once in file. it is not like `ssd` or `ssd` is being used more than once.

iamniting: It was just a suggestion and totally up to you.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -406,6 +407,11 @@ func (r *ReconcileStorageCluster) validateStorageDeviceSets(sc *ocsv1.StorageClu
 				return fmt.Errorf("failed to validate StorageDeviceSet %d: no StorageClass specified for metadataPVCTemplate", i)
 			}
 		}
+		if ds.DeviceType != "" {
+			if !strings.EqualFold(ds.DeviceType, "SSD") || !strings.EqualFold(ds.DeviceType, "HDD") || !strings.EqualFold(ds.DeviceType, "NVMe") {
```
jarrpa, line 411: Don't just compare strings here. Set them as consts and create an array of valid types. Also, are they actually case-sensitive/mixed case, or can they be all lowercase?

jarrpa (changes requested):
You check the value od `ds.DeviceType` but then you don't seem to do anything with it. What's the point of this PR?

Also, please split the CSV changes into a separate commit.

pkg/controller/storagecluster/reconcile.go
jarrpa: I would make it a demand. :) It is good practice to use explicit types rather than just comparing raw strings. You could then also create an array of these types, and just search to see if the array contains the given value.

pkg/apis/ocs/v1/storagecluster_types.go
```diff
@@ -106,6 +106,12 @@ type StorageDeviceSet struct {
 	// +optional
 	Replica int `json:"replica,omitempty"`
 
+	// DeviceType is the value of device type in
+	// this StorageDeviceSet. It can have only
+	// three values (SSD, HDD, NVMe)
+	// +optional
+	DeviceType string `json:"devicetype,omitempty"`
```
umangachapagain, line 113: If these are the only possible values we should be able to error out immediately if any other value is passed. Can you look into adding that?

umangachapagain, line 113:
```suggestion
	DeviceType string `json:"deviceType,omitempty"`
```

pkg/controller/storagecluster/reconcile.go
```diff
@@ -406,6 +407,11 @@ func (r *ReconcileStorageCluster) validateStorageDeviceSets(sc *ocsv1.StorageClu
...
+				return fmt.Errorf("failed to validate DeviceType %s: no Device of this type", ds.DeviceType)
```
umangachapagain, line 412:
```suggestion
				return fmt.Errorf("failed to validate DeviceType %q: no device of this type", ds.DeviceType)
```

umangachapagain (changes requested):

pkg/controller/storagecluster/reconcile.go
crombus: they all can be lower case also. mixed case can be for only NVMe. It is just a logical grouping which will added as a metadata for osd.

pkg/controller/storagecluster/reconcile.go
crombus: the main concern is to check for case sensitivity. https://github.com/openshift/ocs-operator/pull/859#discussion_r513376637

crombus:
After checking ideally it the device type should be in upper case or lower case an array can be made with 6 values(upper case and lower case). Then I can check for `ds.DeviceType` contain in the string. @jarrpa @iamniting

pkg/apis/ocs/v1/storagecluster_types.go
crombus: there can be case sensitive case. https://github.com/openshift/ocs-operator/pull/859#issuecomment-717880007

nbalacha:
> add new property in stroagedeviceset
> DeviceType

Nit: Typo. 

"stroagedeviceset" should be "storagedeviceset"

crombus:
> After checking ideally it the device type should be in upper case or lower case an array can be made with 6 values(upper case and lower case). Then I can check for `ds.DeviceType` contain in the string. @jarrpa @iamniting

NVMe is  abbreviated like this I think adding this in the array too.

pkg/controller/storagecluster/reconcile.go
jarrpa: If that's the case then you don't nee dto codify the multiple case types. Just force the string to always be lowercase and compare those values.

crombus: ok. I'll update the PR

pkg/apis/ocs/v1/storagecluster_types.go
jarrpa: This still needs to be addressed

pkg/controller/storagecluster/reconcile.go
jarrpa: You have not finished the requested changes. Please define the strings as consts, preferably with a new `type DeviceType string`.

pkg/controller/storagecluster/reconcile.go
jarrpa: This still needs to be addressed.

pkg/apis/ocs/v1/storagecluster_types.go
```diff
@@ -43,7 +43,7 @@ type StorageClusterSpec struct {
 	// If enabled, sets the failureDomain to host, allowing devices to be
 	// distributed evenly across all nodes, regardless of distribution in zones
 	// or racks.
-	FlexibleScaling  bool     `json:"flexibleScaling,omitempty"`
+	FlexibleScaling bool `json:"flexibleScaling,omitempty"`
```
jarrpa, line 46: This does not belong in this PR, please remove it.

jarrpa (changes requested):
Multiple review comments remain unaddressed. In addition, the title of this PR is vague and inadequate. Please change the subject to "Add DeviceType to StorageDeviceSet" for *both* the PR and the commit.

crombus:
@jarrpa one the test for crd verification failing. how to solve that?

pkg/controller/storagecluster/reconcile.go
```diff
@@ -33,6 +34,9 @@ type ReconcileStrategy string
 // StorageClassProvisionerType is a string representing StorageClass Provisioner. E.g: aws-ebs
 type StorageClassProvisionerType string
 
+// DeviceTypes is list of valid device type from which storagecluster can be created
+var DeviceTypes = []string{"nvme", "ssd", "hdd"}
```
iamniting, line 38: This is still not done, make them constants as per https://github.com/openshift/ocs-operator/pull/859

iamniting (changes requested):

jarrpa:
> @jarrpa one the test for crd verification failing. how to solve that?

Remove the changes to the CRD validation test, as they are invalid.

pkg/controller/storagecluster/reconcile.go
crombus: I went with this approach https://github.com/openshift/ocs-operator/pull/859#discussion_r516121654.  Not against of adding const it is just that there will be devices which will be added in future as of now PCI will be one. creating const for each one. and array of devices which will better? and there will one benefit of using const that will be when we will tune the osd. So i am like 5050 on both approaches .

jarrpa: I am 100% for consts, so since you're not fully against it I'd say go ahead with the change. :)

nbalacha:
> add new property in stroagedeviceset
> DeviceType

The typo still exists.

Please add more detail as to the use case for this - when is this property used?

pkg/controller/storagecluster/reconcile.go
crombus: consts it is. Added.

jarrpa:
Please run `make ocs-operator-ci` before pushing to GitHub. :)

pkg/controller/storagecluster/reconcile.go
```diff
@@ -33,6 +34,9 @@ type ReconcileStrategy string
 // StorageClassProvisionerType is a string representing StorageClass Provisioner. E.g: aws-ebs
 type StorageClassProvisionerType string
 
+// DeviceTypes is a string representing DeviceType
```
jarrpa, line 37: New structs shouldn't be plural.

```suggestion
// DeviceType is a string representing a storage device type
```

jarrpa (changes requested):

pkg/controller/storagecluster/reconcile.go
crombus: done

test/crd_validation_test.go
```diff
@@ -36,7 +36,6 @@ func TestSampleCustomResources(t *testing.T) {
 	root := "./../deploy/crds"
 	crdCrMap := map[string]string{
 		"ocs.openshift.io_ocsinitializations_crd.yaml": "ocs_v1_ocsinitialization_cr",
-		"ocs.openshift.io_storageclusters_crd.yaml":    "ocs_v1_storagecluster_cr",
```
jarrpa, line 39: This is not correct. You want to maintain the validation of StorageCluster, you just have to change the validation parameters.

jarrpa (changes requested):
Code changes look good, but the change in the test is invalid.

pkg/apis/ocs/v1/storagecluster_types.go
```diff
@@ -106,6 +106,12 @@ type StorageDeviceSet struct {
 	// +optional
 	Replica int `json:"replica,omitempty"`
 
+	// DeviceType is the value of device type in
+	// this StorageDeviceSet. It can have only
+	// three values (SSD, HDD, NVMe)
+	// +optional
+	DeviceType string `json:"deviceType,omitempty"`
```
umangachapagain, line 113:
```suggestion
	// DeviceType is the value of device type in
	// this StorageDeviceSet. It can have only
	// three values (SSD, HDD, NVMe)
	// +kubebuilder:validation:Enum=SSD;ssd;HDD;hdd;NVMe;NVME;nvme
	// +optional
	DeviceType string `json:"deviceType,omitempty"`
```

This will validate user input at client side.

pkg/apis/ocs/v1/storagecluster_types.go
```diff
@@ -106,6 +106,12 @@ type StorageDeviceSet struct {
 	// +optional
 	Replica int `json:"replica,omitempty"`
 
+	// DeviceType is the value of device type in
+	// this StorageDeviceSet. It can have only
+	// three values (SSD, HDD, NVMe)
```
umangachapagain, line 111:
```suggestion
	// this StorageDeviceSet. It can have one of
	// three values (SSD, HDD, NVMe)
```

pkg/controller/storagecluster/reconcile.go
```diff
@@ -62,6 +66,15 @@ osd_memory_target_cgroup_limit_ratio = 0.5
...
 	ReconcileStrategyStandalone ReconcileStrategy = "standalone"
+
+	// DeviceTypeSSD represents the DeviceType SSD
+	DeviceTypeSSD DeviceType = "ssd"
+
+	// DeviceTypeHDD represents the DeviceType HDD
+	DeviceTypeHDD DeviceType = "hdd"
+
+	// DeviceTypeNVMe represents the DeviceType NVMe
+	DeviceTypeNVMe DeviceType = "nvme"
```
umangachapagain, line 77: Since the variables have `DeviceType` prefix, I don't see the benefit for using `DeviceType` alias for `string` type.

We can either do `SSD DeviceType = "ssd"` or `DeviceTypeSSD string = "ssd"`

umangachapagain (changes requested):

pkg/controller/storagecluster/reconcile.go
crombus: I did this because it has been done similar for adding const and naming just SSD or HDD fails for golint test

umangachapagain: Like in Line 44 53 57,  `DeviceTypeSSD = "ssd"` should be enough too.

crombus: done

jarrpa:
CSV test failures seem valid, please investigate. Looks like it may need a rebase?

jarrpa (approved):
/lgtm

PR #868, opened 2020-10-27
Added NodeAffinity for rgw pods.

rajatsing:
Fixes the issue where RGW pods were running on non-OCS nodes.

BZ Link: https://bugzilla.redhat.com/show_bug.cgi?id=1883828

### DIFF
--- a/pkg/controller/defaults/placements.go
+++ b/pkg/controller/defaults/placements.go
@@ -82,6 +82,22 @@ var (
 					getWeightedPodAffinityTerm(100, "rook-ceph-rgw"),
 				},
 			},
+			NodeAffinity: &corev1.NodeAffinity{
+				RequiredDuringSchedulingIgnoredDuringExecution: getOcsNodeSelector(),
+				PreferredDuringSchedulingIgnoredDuringExecution: []corev1.PreferredSchedulingTerm{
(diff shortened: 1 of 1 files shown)


pkg/controller/defaults/placements.go
```diff
@@ -82,6 +82,22 @@ var (
 					getWeightedPodAffinityTerm(100, "rook-ceph-rgw"),
 				},
 			},
+			NodeAffinity: &corev1.NodeAffinity{
+				RequiredDuringSchedulingIgnoredDuringExecution: getOcsNodeSelector(),
+				PreferredDuringSchedulingIgnoredDuringExecution: []corev1.PreferredSchedulingTerm{
```
raghavendra-talur, line 87: Looks like getOcsNodeSelector() only uses the default node selector label. If the StorageCluster.spec has a label selector given by the user then it will not be considered. 

May I suggest that you split getStorageClusterEligibleNodes() to do this?

raghavendra-talur (changes requested):

jarrpa:
This PR is invalid. The RGW Placement is being generated and set correctly with the proper NodeAffinity in the CephCluster CR. This seems to be a problem in Rook-Ceph.

PR #872, opened 2020-10-28
NodeAffinity for rgw pods.

rajatsing:
BZ Link: https://bugzilla.redhat.com/show_bug.cgi?id=1883828

### DIFF
--- a/pkg/controller/storagecluster/cephobjectstores.go
+++ b/pkg/controller/storagecluster/cephobjectstores.go
@@ -107,7 +107,7 @@ func (r *ReconcileStorageCluster) newCephObjectStoreInstances(initData *ocsv1.St
 				Gateway: cephv1.GatewaySpec{
 					Port:      80,
 					Instances: 2,
-					Placement: defaults.DaemonPlacements["rgw"],
+					Placement: getPlacement(initData, "rgw"),
 					Resources: defaults.GetDaemonResources("rgw", initData.Spec.Resources),
 				},
 			},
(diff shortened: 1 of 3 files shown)


rajatsing:
@jarrpa Please review again!.
Thanks

pkg/controller/storagecluster/cephobjectstores.go
```diff
@@ -107,7 +107,7 @@ func (r *ReconcileStorageCluster) newCephObjectStoreInstances(initData *ocsv1.St
 				Gateway: cephv1.GatewaySpec{
 					Port:      80,
 					Instances: 2,
-					Placement: defaults.DaemonPlacements["rgw"],
+					Placement: getPlacement(initData, "rgw"),
```
umangachapagain, line 110: I don't think this solves the problem.

The original issue was that `defaults.DaemonPlacements["rgw"]` doesn't include `NodeAffinity` which was moved out for some reason, and that still persists here.

When placement is not specified in StorageCluster, `getPlacement` falls back to `defaults.DaemonPlacements["rgw"]` and it still wouldn't get `NodeAffinity`. We need to inject NodeAffinity in that case.

Yes, the unit-tests pass because they were designed to pass. But it fails in most of the real world scenarios.

umangachapagain (changes requested):

pkg/controller/storagecluster/cephobjectstores.go
jarrpa: `getPlacement()` sets the NodeAffinity: https://github.com/openshift/ocs-operator/blob/master/pkg/controller/storagecluster/placement.go#L27-L29

jarrpa: I just verified that the RGW Pods are properly receiving their NodeAffinity. LGTM.

jarrpa (approved):
/lgtm

PR #874, opened 2020-10-28
Check MCG.Reconcile strategy before deleting Noobaa resources

nbalacha:
Delete Noobaa resources only they are being managed
by the OCS operator.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -9,13 +9,11 @@ import (
 	"runtime"
 	"time"
 
-	"github.com/go-logr/zapr"
-	"go.uber.org/zap"
-	"go.uber.org/zap/zapcore"
-
 	monitoringv1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
+	"github.com/go-logr/zapr"
 	snapapi "github.com/kubernetes-csi/external-snapshotter/v2/pkg/apis/volumesnapshot/v1beta1"
 	nbapis "github.com/noobaa/noobaa-operator/v2/pkg/apis"
+	openshiftConfigv1 "github.com/openshift/api/config/v1"
 	openshiftv1 "github.com/openshift/api/template/v1"
 	"github.com/openshift/ocs-operator/pkg/apis"
 	ocsv1 "github.com/openshift/ocs-operator/pkg/apis/ocs/v1"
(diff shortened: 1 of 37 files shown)


nbalacha:
Should the same checks be done for the other Managed resources as well?

jarrpa:
I think this makes sense thus far.

> Should the same checks be done for the other Managed resources as well?

Now that you mention it, that doesn't sound like a bad idea. @raghavendra-talur or @iamniting do either of you have any thoughts on that?

vendor/modules.txt
```diff
@@ -153,6 +153,7 @@ github.com/onsi/gomega/matchers/support/goraph/node
 github.com/onsi/gomega/matchers/support/goraph/util
 github.com/onsi/gomega/types
 # github.com/openshift/api v3.9.1-0.20190924102528-32369d4db2ad+incompatible => github.com/openshift/api v0.0.0-20190924102528-32369d4db2ad
+github.com/openshift/api/config/v1
```
iamniting, line 156: It looks like this Change is not part of this PR

iamniting (changes requested):
> I think this makes sense thus far.
> 
> > Should the same checks be done for the other Managed resources as well?
> 
> Now that you mention it, that doesn't sound like a bad idea. @raghavendra-talur or @iamniting do either of you have any thoughts on that?

Make sense to me

iamniting:
> Delete Noobaa resources only they are being managed
> by the OCS operator.

@nbalacha I think 'if' is missing here.

Can you pls prefix "StorageCluster:" as we are following that pattern for all our delete or uninstall commits.

[1]. https://github.com/openshift/ocs-operator/blame/fac013610cf1bf1f0f50ae9c7f17d2163e004055/pkg/controller/storagecluster/uninstall_reconciler.go

deploy/deploy-with-olm.yaml
```diff
@@ -26,7 +26,7 @@ spec:
...
     mediatype: image/svg+xml
-  image: quay.io/ocs-dev/ocs-registry:latest
+  image: quay.io/nibalach/ocs-registry:master
```
iamniting, line 29: Same as before

iamniting (changes requested):

PR #876, opened 2020-10-30
must-gather: some improvements in snapshot info collection

agarwal-mudit:
1. change log file names to rbd_vol_and_snap_info &
   cephfs_subvol_and_snap_info
2. add subvolumegroup name (csi)
3. add snapshot count for cephFS subvolumes

Signed-off-by: Mudit Agarwal <muagarwa@redhat.com>

### DIFF
--- a/must-gather/collection-scripts/gather_ceph_resources
+++ b/must-gather/collection-scripts/gather_ceph_resources
@@ -143,8 +143,7 @@ for ns in $namespaces; do
 
         # Collecting snapshot info for ceph rbd volumes
         printf "collecting snapshot info for ceph rbd volumes \n" | tee -a  "${BASE_COLLECTION_PATH}"/gather-debug.log
-        COMMAND_OUTPUT_FILE=${COMMAND_OUTPUT_DIR}/rbd_snap_info.log
-        JSON_COMMAND_OUTPUT_FILE=${COMMAND_JSON_OUTPUT_DIR}/rbd_snap_info_format_json-pretty.log
+        COMMAND_OUTPUT_FILE=${COMMAND_OUTPUT_DIR}/rbd_vol_and_snap_info
         # Inspecting ceph block pools for ceph rbd
         blockpools=$(oc get cephblockpools.ceph.rook.io -n openshift-storage -o jsonpath="{range .items[*]}{@.metadata.name}{'\n'}{end}")
         for bp in $blockpools; do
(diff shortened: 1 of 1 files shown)


crombus (approved):

must-gather/collection-scripts/gather_ceph_resources
```diff
@@ -156,20 +155,22 @@ for ns in $namespaces; do
...
 
         # Collecting snapshot information for ceph subvolumes
-        printf "collecting snapshot info for ceph subvolumes \n" | tee -a "${BASE_COLLECTION_PATH}"/gather-debug.log
-        COMMAND_OUTPUT_FILE=${COMMAND_OUTPUT_DIR}/cephfs_snap_info.log
+        printf "collecting snapshot info for cephFS subvolumes \n" | tee -a "${BASE_COLLECTION_PATH}"/gather-debug.log
+        COMMAND_OUTPUT_FILE=${COMMAND_OUTPUT_DIR}/cephfs_subvol_and_snap_info
 
         # Inspecting CephFS filesystems
         filesystems=$(oc get cephfilesystems.ceph.rook.io -n openshift-storage -o jsonpath="{range .items[*]}{@.metadata.name}{'\n'}{end}")
         for fs in $filesystems; do
```
Madhu-1, line 163: define a variable for csi group name it will be easy to replace it later if we start supporting multiple subvolume groups

agarwal-mudit: Didn't add it because in that case we would have to run one more loop and whole logic would change.

crombus: agreed. @agarwal-mudit can create var with csi

crombus: don't add a loop . just a var

agarwal-mudit: I can do it but not required because if we support multiple subvolumegroups we need a complete different logic where we would have to find all the subvolumegroups and then extract all the information for each subvolumegroup one by one. And thus whole logic which is there today would change. Let me know if you guys still want me to add the variable

Madhu-1: Yes, correct we need to first get all group in a volume and then check subvolumes in each group.

how about below logic we need to add one more for loop for groups in a volume
```
filesystems=$(oc get cephfilesystems.ceph.rook.io -n openshift-storage -o jsonpath="{range .items[*]}{@.metadata.name}{'\n'}{end}")
        for fs in $filesystems; do
           subvolgroups=$(oc -n openshift-storage exec "${HOSTNAME}"-helper -- bash -c "ceph fs subvolumegroup ls $fs | jq --raw-output '.[].name' ")
           for group in $subvolgroups:
               ---
```
if we don't it we don't need to hardcode CSI right?

agarwal-mudit: Fixed

Madhu-1: I still see the same code, missed to push the code?

agarwal-mudit: Yeah, testing it after the change.

Not adding the loop as of now, just added the variable.

agarwal-mudit: pushed now, PTAL

Madhu-1 (approved):

PR #878, opened 2020-11-02
StorageCluster: Check MCG.Reconcile before deleting Noobaa resources

nbalacha:
Delete Noobaa resources only if they are being managed by
the OCS operator.


iamniting (approved):

raghavendra-talur (approved):
The changes look good to me from uninstall perspective.

I don't know if there are other places in ocs-operator where we are checking if noobaa is enabled/controlled by ocs-operator but if so better to make a function out of this code.

jarrpa (approved):

Issue #879, opened 2020-11-02
StorageCluster: Unbalaced distribution of OSDs by Topology spread on cluster with local PVs


kesavanvt:
the OSD pod distribution is not even for cluster with multiple nodes within a failure domain for local PVs

Steps to reproduce

1. Create a 9 node OCP cluster based on 3 AZs (east-us 1a,1b,1c) . 3 nodes per AZ
2. Attach 2 disk volumes for each node and create local PVs using LSO
3. Create initial OCS cluster selecting the storageClass created on the above step
4. Expand the storage twice

Actual result : within an AZ, single node had two OSDs, leaving one node to be unused.

Expected result : The OSDs should be equally distributed across each node within an AZ. (each node with 1 OSD)

kesavanvt:
Currently, the placement for prepare pod is zone/rack based on the OCS calculation and its not distributed equally within a failure domain . The PVs created by prepare Jobs will contain further restrictions on where the PV can be mounted. For dynamic provisioning, OSD distribution is maintained as the PVs restriction is the failure domain of prepare job zone/rack. However for local PVs , the PVs restriction is to the host, so OSD placement is dependent on the placement of prepare pod. Thus multiple topology spread needs to be set for prepare pods. Top level would be for zone/rack , the bottom level would be for host. In case of flexible scaling it would be set to host without multiple TSC.


 ```
 topologySpreadConstraints:
    - maxSkew: 1
      topologyKey: failure-domain.beta.kubernetes.io/zone
      whenUnsatisfiable: ScheduleAnyway
      labelSelector:
        matchExpressions:
          - key: app
            operator: In
            values:
              - rook-ceph-osd-prepare
              - rook-ceph-osd
    - maxSkew: 1
      topologyKey: kubernetes.io/hostname
      whenUnsatisfiable: ScheduleAnyway
      labelSelector:
        matchExpressions:
          - key: app
            operator: In
            values:
              - rook-ceph-osd-prepare
              - rook-ceph-osd
```

PR #880, opened 2020-11-02
Fix Adding multiple topology constraints for osd prepare pod

kesavanvt:
Fixes : #879 

Signed-off-by: kesavan kvellalo@redhat.com

### DIFF
--- a/pkg/controller/defaults/placements.go
+++ b/pkg/controller/defaults/placements.go
@@ -66,7 +66,7 @@ var (
 				getOcsToleration(),
 			},
 			TopologySpreadConstraints: []corev1.TopologySpreadConstraint{
-				getTopologySpreadConstraintsSpec(1, "rook-ceph-osd-prepare"),
+				getTopologySpreadConstraintsSpec(1, "rook-ceph-osd-prepare", "rook-ceph-osd"),
 			},
 		},
 
--- a/pkg/controller/defaults/placements.go
+++ b/pkg/controller/defaults/placements.go
@@ -66,6 +66,8 @@ var (
 				getOcsToleration(),
 			},
 			TopologySpreadConstraints: []corev1.TopologySpreadConstraint{
+				// getTopologySpreadConstraintsSpec populates values required for topology spread constraints.
+				// TopologyKey gets updated in newStorageClassDeviceSets after determining it from determineFailureDomain.
 				getTopologySpreadConstraintsSpec(1, "rook-ceph-osd-prepare", "rook-ceph-osd"),
 			},
 		},
--- a/pkg/controller/storagecluster/cephcluster.go
+++ b/pkg/controller/storagecluster/cephcluster.go
@@ -402,6 +402,12 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster, serverVersion *version.
 						(&in).DeepCopyInto(&preparePlacement)
 
 						if len(topologyKeyValues) >= replica {
+							// If topologyKey is not host, append additional topology spread constarint to the
+							// default preparePlacement. This serves even distribution at the host level
+							// within a failure domain (zone/rack).
+							if topologyKey != corev1.LabelHostname {
+								preparePlacement.TopologySpreadConstraints = append(preparePlacement.TopologySpreadConstraints, preparePlacement.TopologySpreadConstraints[0])
+							}
 							preparePlacement.TopologySpreadConstraints[0].TopologyKey = topologyKey
 						}
 					}


jarrpa:
I odn't quite understand what's going on... are you adding the same TSC twice? What does that accomplish?

kesavanvt:
> I odn't quite understand what's going on... are you adding the same TSC twice? What does that accomplish?

Basically, i have added twice because one TSC will act for zone/rack based on the runtime calculation (the topology key will be set at runtime) and the other would be for host by defaultwhich makes OSDs to distribute equally among the zone/rack. In case of flexible scaling, only one TSC will be applied (for host failure domain)

jarrpa:
...OH! I see now. Please add a comment to that effect over the new one you're adding. Specifically mention the function that would modify the first TSC and that having duplicate TSCs is not a problem.

pkg/controller/defaults/placements.go
```diff
@@ -66,7 +66,8 @@ var (
 				getOcsToleration(),
 			},
 			TopologySpreadConstraints: []corev1.TopologySpreadConstraint{
-				getTopologySpreadConstraintsSpec(1, "rook-ceph-osd-prepare"),
+				getTopologySpreadConstraintsSpec(1, "rook-ceph-osd-prepare", "rook-ceph-osd"),
+				getTopologySpreadConstraintsSpec(1, "rook-ceph-osd-prepare", "rook-ceph-osd"),
```
nbalacha, line 70: I think the reverse approach would be easier to understand - add the extra TSC entry and set it to zone/rack at runtime only if flexible scaling is disabled.

pkg/controller/storagecluster/cephcluster.go
```diff
@@ -403,6 +403,10 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster, serverVersion *version.
 						in := getPlacement(sc, "osd-prepare-tsc")
 						(&in).DeepCopyInto(&preparePlacement)
 
+						if sc.Spec.FlexibleScaling == true {
```
nbalacha, line 406: Please compare against topologyKey instead of FlexibleScaling because the storageDeviceSets also has a topologyKey.

nbalacha (changes requested):

kesavanvt:
> ...OH! I see now. Please add a comment to that effect over the new one you're adding. Specifically mention the function that would modify the first TSC and that having duplicate TSCs is not a problem.

I have added the necessary comments. Unfortunately placement with duplicate TSC leads to error ( refer https://kubernetes.io/blog/2020/05/introducing-podtopologyspread/ ). so i have added the statements to remove multi level TSC wherever necessary in case of host failure domain.

pkg/controller/defaults/placements.go
```diff
@@ -66,7 +66,14 @@ var (
 				getOcsToleration(),
 			},
 			TopologySpreadConstraints: []corev1.TopologySpreadConstraint{
-				getTopologySpreadConstraintsSpec(1, "rook-ceph-osd-prepare"),
+				// getTopologySpreadConstraintsSpec populate values required for topology spread constraints.
```
umangachapagain, line 69:
```suggestion
				// getTopologySpreadConstraintsSpec populates values required for topology spread constraints.
```

pkg/controller/defaults/placements.go
umangachapagain: Agree with @nbalacha 
This looks like a mess to me.

pkg/controller/storagecluster/cephcluster.go
```diff
@@ -403,8 +403,16 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster, serverVersion *version.
...
+						// Eliminating default multi level topology spread constraint in case of host failure domain.
+						if sc.Spec.FlexibleScaling {
+							preparePlacement.TopologySpreadConstraints = preparePlacement.TopologySpreadConstraints[:1]
+						}
+
 						if len(topologyKeyValues) >= replica {
 							preparePlacement.TopologySpreadConstraints[0].TopologyKey = topologyKey
+						} else {
+							// Eliminating default multi level topology spread consrtaints.
+							preparePlacement.TopologySpreadConstraints = preparePlacement.TopologySpreadConstraints[:1]
```
umangachapagain, line 415: If Line 408 executes, what is the value of this line? I don't think this does any change to TSC.

umangachapagain (changes requested):

pkg/controller/defaults/placements.go
jarrpa: +1

kesavanvt: Thanks,required changes have been done!

pkg/controller/storagecluster/cephcluster.go
kesavanvt: Thanks, comparing against the topology key seems to be appropriate and changes are done!

pkg/controller/storagecluster/cephcluster.go
kesavanvt: Now the changes have been made to add multiple TSC only when zone/rack is detected as failure domain.

jarrpa (approved):
Changes look good! A much cleaner and straightforward patch. :)

/lgtm

jarrpa (approved):
/lgtm

PR #884, opened 2020-11-03
must-gather: add ceph crash collection func

crombus:
add ceph crash collection func and make it
run it a process for every storage node

### DIFF
--- a/must-gather/collection-scripts/gather_ceph_resources
+++ b/must-gather/collection-scripts/gather_ceph_resources
@@ -224,6 +224,7 @@ for ns in $namespaces; do
         crash_collection &
         # collecting PID for bg jobs
         pids[${idx}]=$!
+        # shellcheck disable=SC2046
         idx=$idx+1
     done
 
--- a/must-gather/collection-scripts/gather_ceph_resources
+++ b/must-gather/collection-scripts/gather_ceph_resources
@@ -205,14 +205,31 @@ for ns in $namespaces; do
 
     oc delete -f pod_helper.yaml
 
+    crash_collection(){
+        oc debug nodes/"${node}" --to-namespace="${ns}" -- bash -c "sleep 5m" &
+        while [ "$(oc get pods -n openshift-storage | grep "${node//./}-debug" | awk '{print $2}')" != "1/1" ] ; do
+            sleep 3
+            echo "waiting for the debug pod to be in ready state"
+        done
+        oc rsync -n "${ns}" "$(oc get pods -n "${ns}"| grep "${node//./}-debug"| awk '{print $1}')":/host/var/lib/rook/openshift-storage/crash/ "${CRASH_OUTPUT_DIR}"
+    }
+
+    # creating a counter variable for collecting PID in array
+    idx=0
     # Collecting ceph crash dump
     for node in $(oc get nodes -l cluster.ocs.openshift.io/openshift-storage='' --no-headers | awk '/Ready/{print $1}'); do
         printf "collecting crash logs  from node %s \n"  "${node}" | tee -a  "${BASE_COLLECTION_PATH}"/gather-debug.log
         CRASH_OUTPUT_DIR=${CEPH_COLLECTION_PATH}/crash_${node}
         mkdir -p "${CRASH_OUTPUT_DIR}"
-        oc debug nodes/"${node}" --to-namespace="${ns}" -- bash -c "sleep 5m" &
-        sleep 60
-        oc rsync -n "${ns}" "$(oc get pods -n "${ns}"| awk '/debug/{print $1}')":/host/var/lib/rook/openshift-storage/crash/ "${CRASH_OUTPUT_DIR}"
-        wait
+        crash_collection &
+        # collecting PID for bg jobs
+        pids[${idx}]=$!
+        idx=$idx+1
     done
+
+    # wait for all pids
+    echo "waiting for ${pids[*]} to terminate"
+    wait "${pids[@]}"
+
+    echo "ceph core dump collection completed"
 done


must-gather/collection-scripts/gather_ceph_resources
```diff
@@ -205,14 +205,20 @@ for ns in $namespaces; do
 
     oc delete -f pod_helper.yaml
 
+    crash_collection(){
+        oc debug nodes/"${node}" --to-namespace="${ns}" -- bash -c "sleep 5m" &
+        sleep 60
+        oc rsync -n "${ns}" "$(oc get pods -n "${ns}"| grep "${node//./}"-debug| awk '{print $1}')":/host/var/lib/rook/openshift-storage/crash/ "${CRASH_OUTPUT_DIR}"
+        wait
```
jarrpa, line 212: I believe this wait would wait the full five minutes for the sleep to finish even if the rsync finishes. Is this what we want?

crombus: yes because otherwise it debug pods will be running in bg after MG completes it's run.

travisn: Why do we have to wait? Just remove the `wait` and let the `oc debug` command timeout on its own. It's not going to hurt anything by staying around. We must be efficient about gathering logs from every node, and can't afford to sleep when connecting to many nodes.

must-gather/collection-scripts/gather_ceph_resources
```diff
@@ -205,14 +205,20 @@ for ns in $namespaces; do
 
     oc delete -f pod_helper.yaml
 
+    crash_collection(){
+        oc debug nodes/"${node}" --to-namespace="${ns}" -- bash -c "sleep 5m" &
+        sleep 60
```
travisn, line 210: Instead of sleeping, how about a simple loop that will check every few seconds if the pod is ready? Then as soon as it is ready we can continue. Again, it's important to avoid long sleeps.

travisn (changes requested):

must-gather/collection-scripts/gather_ceph_resources
crombus: yes that can be done.

must-gather/collection-scripts/gather_ceph_resources
leseb: Why not just `wait` for the `rsync` to be done since they run in // and get rid of the `sleep 5m`?
So we exit as soon as we have finished rsyncing logs.

crombus:
if we want let the debug pod created by MG stay in the system. I will be fine to remove the wait. And in that case sleep for debug pod can be longer also.

must-gather/collection-scripts/gather_ceph_resources
crombus: oc debug pod needs to be in // because it rsh into the pod when run it.

obnoxxx: there must be a better way! maybe the polling every few seconds that travis mentioned in a different comment. or some real timeout wrapper from coreutils. but the 5m sleep does not scale. :-)

obnoxxx (changes requested):

must-gather/collection-scripts/gather_ceph_resources
jarrpa: That's for the `sleep 60` timer. Here we just want to get rid of the wait, since we want to exit EITHER when the debug Pod times out and disappears or the rsync finishes, whichever comes first. The `sleep 5m` above is to keep the debug Pod alive, since it will just immediately exit and terminate once the provide command finishes.

must-gather/collection-scripts/gather_ceph_resources
crombus: done

must-gather/collection-scripts/gather_ceph_resources
crombus: removed the wait

must-gather/collection-scripts/gather_ceph_resources
```diff
@@ -205,14 +205,22 @@ for ns in $namespaces; do
...
     # Collecting ceph crash dump
     for node in $(oc get nodes -l cluster.ocs.openshift.io/openshift-storage='' --no-headers | awk '/Ready/{print $1}'); do
         printf "collecting crash logs  from node %s \n"  "${node}" | tee -a  "${BASE_COLLECTION_PATH}"/gather-debug.log
         CRASH_OUTPUT_DIR=${CEPH_COLLECTION_PATH}/crash_${node}
         mkdir -p "${CRASH_OUTPUT_DIR}"
-        oc debug nodes/"${node}" --to-namespace="${ns}" -- bash -c "sleep 5m" &
-        sleep 60
-        oc rsync -n "${ns}" "$(oc get pods -n "${ns}"| awk '/debug/{print $1}')":/host/var/lib/rook/openshift-storage/crash/ "${CRASH_OUTPUT_DIR}"
-        wait
+        crash_collection &
```
leseb, line 222: You need to register each PID and `wait` for them at the end.

leseb (changes requested):

must-gather/collection-scripts/gather_ceph_resources
crombus: the wait cmd after the func will take care for that.

leseb: This would wait for **any** background jobs. We don't want that, we want to wait for the jobs **we** spawned.

crombus: done

must-gather/collection-scripts/gather_ceph_resources
```diff
@@ -205,14 +205,33 @@ for ns in $namespaces; do
...
+        # collecting PID for bg jobs
+        pids[${count}]=$!
+        count=`expr ${count} + 1`
```
leseb, line 227: do not use backticks
```suggestion
        count=$(expr ${count} + 1)
```

must-gather/collection-scripts/gather_ceph_resources
```diff
@@ -205,14 +205,33 @@ for ns in $namespaces; do
...
     done
+
+    # wait for all pids
+    for pid in ${pids[@]}; do
+        echo "waiting for ${pid}"
+        wait ${pid}
```
leseb, line 233:
```suggestion
        wait ${pids[@]}
```

You don't need the loop you can just print the array directly.

leseb (changes requested):

must-gather/collection-scripts/gather_ceph_resources
crombus: done

must-gather/collection-scripts/gather_ceph_resources
crombus: done

must-gather/collection-scripts/gather_ceph_resources
```diff
@@ -205,14 +205,32 @@ for ns in $namespaces; do
 
     oc delete -f pod_helper.yaml
 
+    crash_collection(){
+        oc debug nodes/"${node}" --to-namespace="${ns}" -- bash -c "sleep 5m" &
+        while [ "$(oc get pods -n openshift-storage | grep "${node//./}-debug" | awk '{print $2}')" != "1/1" ] ; do
```
leseb, line 210: Do these pods have labels we could use?

So we could do something like:

```bash
[[ "$(oc get pods -n openshift-storage -l app=DEBUG_LABEL -o 'jsonpath={..status.conditions[?(@.type=="Ready")].status}') != "True" ]]
```

must-gather/collection-scripts/gather_ceph_resources
```diff
@@ -205,14 +205,32 @@ for ns in $namespaces; do
...
+            sleep 3
+            echo "waiting for the debug pod to be in ready state"
+        done
+        oc rsync -n "${ns}" "$(oc get pods -n "${ns}"| grep "${node//./}-debug"| awk '{print $1}')":/host/var/lib/rook/openshift-storage/crash/ "${CRASH_OUTPUT_DIR}"
```
leseb, line 214: I also wonder if this `"$(oc get pods -n "${ns}"| grep "${node//./}-debug"| awk '{print $1}')"` could be simplified.

must-gather/collection-scripts/gather_ceph_resources
```diff
@@ -205,14 +205,32 @@ for ns in $namespaces; do
...
-        wait
+        crash_collection &
+        # collecting PID for bg jobs
+        pids[${count}]=$!
+        # shellcheck disable=SC2003,SC2046
+        $(count)=$(expr $(count) + 1)
     done
+
+    # wait for all pids
+    echo "waiting for ${pids[*]}"
```
leseb, line 232:
```suggestion
    echo "waiting for ${pids[*]} to terminate"
```

leseb (changes requested):
Almost there.

must-gather/collection-scripts/gather_ceph_resources
jarrpa: I agree, this would be cleaner and more reliable.

must-gather/collection-scripts/gather_ceph_resources
jarrpa: :+1:

must-gather/collection-scripts/gather_ceph_resources
```diff
@@ -205,14 +205,31 @@ for ns in $namespaces; do
...
         printf "collecting crash logs  from node %s \n"  "${node}" | tee -a  "${BASE_COLLECTION_PATH}"/gather-debug.log
         CRASH_OUTPUT_DIR=${CEPH_COLLECTION_PATH}/crash_${node}
         mkdir -p "${CRASH_OUTPUT_DIR}"
-        oc debug nodes/"${node}" --to-namespace="${ns}" -- bash -c "sleep 5m" &
-        sleep 60
-        oc rsync -n "${ns}" "$(oc get pods -n "${ns}"| awk '/debug/{print $1}')":/host/var/lib/rook/openshift-storage/crash/ "${CRASH_OUTPUT_DIR}"
-        wait
+        crash_collection &
+        # collecting PID for bg jobs
+        pids[${count}]=$!
```
jarrpa, line 226: I can't find where this variable is defined outside the `for` loop. Is it missing or am I missing it?

jarrpa (changes requested):

must-gather/collection-scripts/gather_ceph_resources
crombus: there can be multiple debug pods how to know which one is for which.

must-gather/collection-scripts/gather_ceph_resources
crombus: same here

must-gather/collection-scripts/gather_ceph_resources
crombus: this can be done

must-gather/collection-scripts/gather_ceph_resources
crombus: line no 218

must-gather/collection-scripts/gather_ceph_resources
crombus: I checked no labels

must-gather/collection-scripts/gather_ceph_resources
jarrpa: That's `count`, not `pids`.

must-gather/collection-scripts/gather_ceph_resources
jarrpa: Ah well! I can probably live with this for now.

jarrpa (changes requested):

must-gather/collection-scripts/gather_ceph_resources
crombus: pids not declared separately. This works fine.

jarrpa: ...how is this possible? If you don't declare an array outside the scope of the for loop it should disappear once the for loop exits.

jarrpa: Okay, so I wrote a test script:
```bash
  #!/bin/bash
  
  idx=0
  for i in 1 2 3; do
      echo $i &
      pids[${idx}]=$!
      idx=${idx}+1
  done
  
  # wait for all pids
  echo "waiting for ${pids[*]} to terminate"
  wait "${pids[@]}"
```
And found that indeed `pids` is preserved outside the scope of the for loop. This comes as a bit of a surprise, but okay.

I will point out one thing, though: I was able to get rid of the `expr` call and just do basic variable arithmetic. Is there any reason you couldn't also do that? Would that resolve the shellcheck errors?

must-gather/collection-scripts/gather_ceph_resources
```diff
@@ -205,14 +205,32 @@ for ns in $namespaces; do
...
+        oc debug nodes/"${node}" --to-namespace="${ns}" -- bash -c "sleep 5m" &
+        while [ "$(oc get pods -n openshift-storage | grep "${node//./}-debug" | awk '{print $2}')" != "1/1" ] ; do
+            sleep 3
+            echo "waiting for the debug pod to be in ready state"
+        done
+        oc rsync -n "${ns}" "$(oc get pods -n "${ns}"| grep "${node//./}-debug"| awk '{print $1}')":/host/var/lib/rook/openshift-storage/crash/ "${CRASH_OUTPUT_DIR}"
+    }
+
+    # creating a counter variable for collecting PID in array
+    count=1
```
jarrpa, line 218: Rename this to something like `index` or `idx`. Any reason it starts at 1? Otherwise since it's an array index it should start at 0.

must-gather/collection-scripts/gather_ceph_resources
crombus: I was not able to do that without expr but if you are able then I might have missed something. let me check that again.

must-gather/collection-scripts/gather_ceph_resources
crombus: it is for indexing. It should start with zero. I'll update this also.

crombus:
updated the PR. With required changes.

must-gather/collection-scripts/gather_ceph_resources
leseb: I don't mean to go too deep into this but you don't even need to declare the index. You can append in the array directly: `array+=($pid)`. And initialize the array with `array=()`.

must-gather/collection-scripts/gather_ceph_resources
crombus: certain things were made for my initial logic after the review code changes are done. So the dependent code changes can also be modify.

leseb (approved):

must-gather/collection-scripts/gather_ceph_resources
jarrpa: Also valid, but if it works don't break it. :joy:

jarrpa (approved):
Excellent, a much improved PR!

/lgtm

PR #885, opened 2020-11-03
Refactor status updates into a post reconcile phase and add component image reconcile status

nb-ohad:
- Consolidate all `storagecluster` status updates into a post reconcile phase (see commit comment for full explanation) 
- Refactor image info on the reconciler from props into a string map 
- Add reconcile status info for component images


pkg/controller/storagecluster/reconcile.go
```diff
@@ -97,6 +96,37 @@ func (r *ReconcileStorageCluster) Reconcile(request reconcile.Request) (reconcil
...
+		if sc.Status.Images[name] == nil {
+			sc.Status.Images[name] = &ocsv1.ComponentImageStatus{}
+		}
+		sc.Status.Images[name].DesiredImage = image
+	}
+
+	// Reconcile changes to the cluster
+	result, reconcileError := r.reconcilePhases(reqLogger, sc, request)
+
+	// Apply status changes to  the stroagecluster
```
umangachapagain, line 110:
```suggestion
	// Apply status changes to the storagecluster
```

pkg/controller/storagecluster/reconcile.go
```diff
@@ -97,6 +96,37 @@ func (r *ReconcileStorageCluster) Reconcile(request reconcile.Request) (reconcil
...
+	statusError := r.client.Status().Update(context.TODO(), sc)
+	if statusError != nil {
+		reqLogger.Info("Status Update Error", "StatusUpdateErr", "Could not update storagecluster status")
+	}
+
+	// Reconcile errors have highr priority then status update errors
```
umangachapagain, line 116:
```suggestion
	// Reconcile errors have higher priority than status update errors
```

umangachapagain (changes requested):

umangachapagain (commented):
Good additions/re-factor.

Some doubts:
 
With central status update, wouldn't the intermediate phase info be lost?
Say we go from Initialization -> Progressing -> Expanding -> Ready.
With this change we'd only see "Ready" and nothing in between? (Remember this is reflected on UI)

The "Expanding" phase was already visible only for a fraction of second, with this change I don't expect even that to happen. How do we tackle that?

pkg/controller/storagecluster/storagecluster_controller.go
```diff
@@ -41,19 +41,21 @@ func Add(mgr manager.Manager) error {
 }
 
 func (r *ReconcileStorageCluster) initializeImageVars() error {
-	r.cephImage = os.Getenv("CEPH_IMAGE")
-	r.noobaaCoreImage = os.Getenv("NOOBAA_CORE_IMAGE")
-	r.noobaaDBImage = os.Getenv("NOOBAA_DB_IMAGE")
+	r.images = map[string]string{
+		"ceph":        os.Getenv("CEPH_IMAGE"),
+		"noobaa-core": os.Getenv("NOOBAA_CORE_IMAGE"),
+		"noobaa-db":   os.Getenv("NOOBAA_DB_IMAGE"),
```
umangachapagain, line 47: Can we define these keys as constants? Since it's referenced at multiple places that'll probably help us limit the changes if we ever refactor this.

umangachapagain (changes requested):

pkg/controller/storagecluster/storagecluster_controller.go
nb-ohad: I believe that after the change, this is the only place where we refer to these env vars.
The only other places I could found are shell scripts were moving to constants is irrelevant

pkg/controller/storagecluster/reconcile.go
nb-ohad: done

pkg/controller/storagecluster/reconcile.go
nb-ohad: done

nb-ohad:
Umanga, thank you for your comments.

> With central status update, wouldn't the intermediate phase info be lost?
> Say we go from Initialization -> Progressing -> Expanding -> Ready.
> With this change we'd only see "Ready" and nothing in between? (Remember this is reflected on UI)

I will argue to the contrary: 
Each reconcile invocation should not take long, as such any intermediate updates to the status can be ignored safely as they are hardly visible. 

Not we have 2 other cases:
1. The reconcile has an early exit because of an error, 
    - The code ensures that the correct status is visible on the CR
2. The reconcile has an early exit because of an early return
    - The code also ensures that the correct status is visible on the CR

> The "Expanding" phase was already visible only for a fraction of a second, with this change I don't expect even that to happen. > How do we tackle that?

I will argue that if a phase is hardly visible we should not care to refect it at all. 
But If we have an early exit in any phase (without an error) it is already guaranteed to be reflected in the CR status.

As an example, I can give the case where we return a `nil` for error and a `reconcile.Result` with a `RequeueAfter` field set to a value when we are waiting for uninstall. In that case, the phase will be reflected in the status.

If the "Expending" phase can potentially take a long time (which is not the case as for now) we should refactor the code to issue all the changes then schedule a new reconcile a few seconds later instead of waiting for it in the code. 
If we do that we will have the "Expending" reflected in the status.

pkg/controller/storagecluster/storagecluster_controller.go
umangachapagain: What about the places where you are updating `r.Images["ceph"]` etc?

nb-ohad: Ok, I understand now. 
You meant `ceph`, `noobaa-core` and `noobaa-db`.
I was thinking you were referring to `CEPH_IMAGE`, `NOOBAA_CORE_IMAGE` and `NOOBAA_DB_IMAGE`

In that case, we can do constants for the keys in order to deflate the value range.

nb-ohad: done

umangachapagain:
> If the "Expending" phase can potentially take a long time (which is not the case as for now) we should refactor the code to issue all the changes then schedule a new reconcile a few seconds later instead of waiting for it in the code.
If we do that we will have the "Expending" reflected in the status.

Expanding happens for a considerable amount of time, but the Phase itself disappears immediately. We've had request to make it more visible. So it'd be great to see enhancements for that.

That said, I have nothing against this approach where we ignore intermediate phase for reduced API calls and chances of errors.

umangachapagain (approved):

nb-ohad:
@umangachapagain 
The best practices for writing operators dictate that you should never wait for long tasks inside the reconcile invocation. 
In case of a long operation, we should just lunch the operation, update the status, and return a `reconcile.Result` to trigger another reconcile later.

 If we will refactor out code to work this way we will get the desired behavior for  "Expanding".
The code change in this PR will support such a change in the future.

I would like to see such a change but not as part of this PR as this change is out of the scope of what I am trying to accomplish here.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -84,9 +84,8 @@ func (r *ReconcileStorageCluster) Reconcile(request reconcile.Request) (reconcil
 	reqLogger := r.reqLogger.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
 
 	// Fetch the StorageCluster instance
-	instance := &ocsv1.StorageCluster{}
-	err := r.client.Get(context.TODO(), request.NamespacedName, instance)
-	if err != nil {
+	sc := &ocsv1.StorageCluster{}
```
jarrpa, line 87: This variable rename probably belongs in another commit.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -98,6 +97,30 @@ func (r *ReconcileStorageCluster) Reconcile(request reconcile.Request) (reconcil
 		return reconcile.Result{}, err
 	}
 
+	// Reconcile changes to the cluster
+	result, reconcileError := r.reconcilePhases(reqLogger, sc, request)
+
+	// Apply status changes to  the stroagecluster
```
jarrpa:
```suggestion
	// Apply status changes to  the StorageCluster
```

pkg/controller/storagecluster/reconcile.go
```diff
@@ -120,20 +143,14 @@ func (r *ReconcileStorageCluster) Reconcile(request reconcile.Request) (reconcil
...
 			return reconcile.Result{}, nil
 		}
 	} else if instance.Status.Phase == statusutil.PhaseIgnored {
 		return reconcile.Result{}, nil
 	}
 
 	if !instance.Spec.ExternalStorage.Enable {
-		err = r.validateStorageDeviceSets(instance)
-		if err != nil {
+		if err := r.validateStorageDeviceSets(instance); err != nil {
```
jarrpa, line 160: Probably belongs in the same other commit as the variable rename

pkg/controller/storagecluster/reconcile.go
```diff
@@ -168,35 +176,29 @@ func (r *ReconcileStorageCluster) Reconcile(request reconcile.Request) (reconcil
 			reqLogger.Info("Finalizer not found for storagecluster. Adding finalizer")
 			instance.ObjectMeta.Finalizers = append(instance.ObjectMeta.Finalizers, storageClusterFinalizer)
 			if err := r.client.Update(context.TODO(), instance); err != nil {
-				reqLogger.Info("Status Update Error", "StatusUpdateErr", "Failed to update storagecluster with finalizer")
+				reqLogger.Info("Update Error", "MetaUpdateErr", "Failed to update storagecluster with finalizer")
```
jarrpa, line 186: Move out of this commit.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -168,35 +176,29 @@ func (r *ReconcileStorageCluster) Reconcile(request reconcile.Request) (reconcil
...
 				return reconcile.Result{}, err
 			}
 		}
 
-		err = r.reconcileUninstallAnnotations(instance, reqLogger)
-		if err != nil {
+		if err := r.reconcileUninstallAnnotations(instance, reqLogger); err != nil {
```
jarrpa, line 191: Move out of this commit.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -168,35 +176,29 @@ func (r *ReconcileStorageCluster) Reconcile(request reconcile.Request) (reconcil
...
 		instance.Status.Phase = statusutil.PhaseDeleting
-		phaseErr := r.client.Status().Update(context.TODO(), instance)
-		if phaseErr != nil {
-			reqLogger.Info("Status Update Error", "StatusUpdateErr", "Failed to set PhaseDeleting")
-		}
 
 		if contains(instance.GetFinalizers(), storageClusterFinalizer) {
-			err = r.deleteResources(instance, reqLogger)
-			if err != nil {
+			if err := r.deleteResources(instance, reqLogger); err != nil {
```
jarrpa, line 200: Move out of this commit.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -168,35 +176,29 @@ func (r *ReconcileStorageCluster) Reconcile(request reconcile.Request) (reconcil
...
 				reqLogger.Info("Uninstall in progress", "Status", err)
 				return reconcile.Result{RequeueAfter: time.Second * time.Duration(1)}, nil
 			}
 			reqLogger.Info("Removing finalizer")
 			// Once all finalizers have been removed, the object will be deleted
 			instance.ObjectMeta.Finalizers = remove(instance.ObjectMeta.Finalizers, storageClusterFinalizer)
 			if err := r.client.Update(context.TODO(), instance); err != nil {
-				reqLogger.Info("Status Update Error", "StatusUpdateErr", "Failed to remove finalizer from storagecluster")
+				reqLogger.Info("Update Error", "MetaUpdateErr", "Failed to remove finalizer from storagecluster")
```
jarrpa, line 208: Move out of this commit.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -291,12 +275,12 @@ func (r *ReconcileStorageCluster) Reconcile(request reconcile.Request) (reconcil
...
 		r := ready.NewFileReady()
-		err = r.Set()
-		if err != nil {
+		if err := r.Set(); err != nil {
 			reqLogger.Error(err, "Failed to mark operator ready")
 			return reconcile.Result{}, err
 		}
-		if instance.Status.Phase != statusutil.PhaseClusterExpanding && !instance.Spec.ExternalStorage.Enable {
+		if instance.Status.Phase != statusutil.PhaseClusterExpanding &&
+			!instance.Spec.ExternalStorage.Enable {
```
jarrpa, line 290: Move out of this commit.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -291,12 +275,12 @@ func (r *ReconcileStorageCluster) Reconcile(request reconcile.Request) (reconcil
 		// If no operator whose conditions we are watching reports an error, then it is safe
 		// to set readiness.
 		r := ready.NewFileReady()
-		err = r.Set()
-		if err != nil {
+		if err := r.Set(); err != nil {
```
jarrpa, line 285: Move out of this commit.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -324,8 +308,7 @@ func (r *ReconcileStorageCluster) Reconcile(request reconcile.Request) (reconcil
 		// If for any reason we marked ourselves !upgradeable...then unset readiness
 		if conditionsv1.IsStatusConditionFalse(instance.Status.Conditions, conditionsv1.ConditionUpgradeable) {
 			r := ready.NewFileReady()
-			err = r.Unset()
-			if err != nil {
+			if err := r.Unset(); err != nil {
```
jarrpa, line 318: Move out of this commit.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -341,23 +324,16 @@ func (r *ReconcileStorageCluster) Reconcile(request reconcile.Request) (reconcil
...
-	err = r.enableMetricsExporter(instance)
-	if err != nil {
+	if err := r.enableMetricsExporter(instance); err != nil {
 		reqLogger.Error(err, "failed to reconcile metrics exporter")
 		return reconcile.Result{}, err
 	}
 
-	err = r.enablePrometheusRules(instance.Spec.ExternalStorage.Enable)
-	if err != nil {
+	if err := r.enablePrometheusRules(instance.Spec.ExternalStorage.Enable); err != nil {
```
jarrpa, line 343: Move out of this commit.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -341,23 +324,16 @@ func (r *ReconcileStorageCluster) Reconcile(request reconcile.Request) (reconcil
...
-		reqLogger.Info("Status Update Error", "StatusUpdateErr", "Failed to set PhaseProgressing")
-		return reconcile.Result{}, phaseErr
-	}
 
 	// enable metrics exporter at the end of reconcile
 	// this allows storagecluster to be instantiated before
 	// scraping metrics
-	err = r.enableMetricsExporter(instance)
-	if err != nil {
+	if err := r.enableMetricsExporter(instance); err != nil {
```
jarrpa, line 338: Move out of this commit.

pkg/apis/ocs/v1/storagecluster_types.go
```diff
@@ -43,7 +43,7 @@ type StorageClusterSpec struct {
 	// If enabled, sets the failureDomain to host, allowing devices to be
 	// distributed evenly across all nodes, regardless of distribution in zones
 	// or racks.
-	FlexibleScaling  bool     `json:"flexibleScaling,omitempty"`
+	FlexibleScaling bool `json:"flexibleScaling,omitempty"`
```
jarrpa, line 46: Move out of this commit.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -96,17 +96,24 @@ func (r *ReconcileStorageCluster) Reconcile(request reconcile.Request) (reconcil
...
+			sc.Status.Images[name] = &ocsv1.ComponentImageStatus{}
+		}
+		sc.Status.Images[name].DesiredImage = image
+	}
 
 	// Reconcile changes to the cluster
 	result, reconcileError := r.reconcilePhases(reqLogger, sc, request)
 
-	// Apply status changes to  the stroagecluster
+	// Apply status changes to the storagecluster
```
jarrpa, line 110: Move out of this commit.

pkg/controller/storagecluster/reconcile.go
```diff
@@ -96,17 +96,24 @@ func (r *ReconcileStorageCluster) Reconcile(request reconcile.Request) (reconcil
...
 	statusError := r.client.Status().Update(context.TODO(), sc)
 	if statusError != nil {
 		reqLogger.Info("Status Update Error", "StatusUpdateErr", "Could not update storagecluster status")
 	}
 
-	// Reconcile errors have highr priority then status update errors
+	// Reconcile errors have higher priority than status update errors
```
jarrpa, line 116: Move out of this commit.

jarrpa (changes requested):
Solid work, though the commits need some cleanup. All the formatting and spelling fixes should go in a separate commit. You can also merge the constants commit into the image info refactor commit.

pkg/controller/storagecluster/reconcile.go
nb-ohad: This is not a var rename, this is a totally new function that took the name of the old one (`reconcile`). 
The code of the old function was renamed `reconcilePhases` and there I did not touch the name of the var (it is still instance). It just that git diff is a little bit awkward in these situations. 
As such I don't see a problem with this line

pkg/controller/storagecluster/reconcile.go
nb-ohad: done

pkg/controller/storagecluster/reconcile.go
nb-ohad: done

pkg/controller/storagecluster/reconcile.go
nb-ohad: done

pkg/controller/storagecluster/reconcile.go
nb-ohad: done

pkg/controller/storagecluster/reconcile.go
nb-ohad: done

pkg/controller/storagecluster/reconcile.go
nb-ohad: done

pkg/controller/storagecluster/reconcile.go
nb-ohad: done

pkg/controller/storagecluster/reconcile.go
nb-ohad: done

pkg/controller/storagecluster/reconcile.go
nb-ohad: done

pkg/controller/storagecluster/reconcile.go
nb-ohad: done

pkg/controller/storagecluster/reconcile.go
nb-ohad: done

pkg/apis/ocs/v1/storagecluster_types.go
nb-ohad: This is an automatic change done by running `gofmt` (which I run before every commit)

pkg/controller/storagecluster/reconcile.go
nb-ohad: done

pkg/controller/storagecluster/reconcile.go
nb-ohad: done

nb-ohad:
@jarrpa 
As a general note, I have moved all the changes you requested into a new commit called `tidy reconcile code` which is now the first commit in this PR. This leaves only meaningful changes in all the other commits

pkg/apis/ocs/v1/storagecluster_types.go
jarrpa: That's fine, please move it into the tidy commit.

nb-ohad: done

jarrpa (commented):
Commits look good now. Waiting on CI fixes.

pkg/apis/ocs/v1/storagecluster_types.go
```diff
@@ -184,8 +184,20 @@ type StorageClusterStatus struct {
 
 	// ExternalSecretHash holds the checksum value of external secret data.
 	ExternalSecretHash string `json:"externalSecretHash,omitempty"`
+
+	// Images holds the image reconclie status for all images reconciled by the operator
```
nbalacha, line 188: Nit: typo : reconclie

jarrpa (changes requested):
On its surface the code looks good, but the e2e failures seem valid. Have you been able to run this successfully in your own clusters?

pkg/apis/ocs/v1/storagecluster_types.go
```diff
@@ -185,8 +185,20 @@ type StorageClusterStatus struct {
...
+	Images map[string]ComponentImageStatusRef `json:"imageReconcileState,omitempty"`
```
jarrpa, line 190: Since the description shows up in the CSV along with the field, the JSON field name should match the Go field name. I like the name `Images` better, so please also update the JSON name to match.

jarrpa, line 190: Something seems to be wrong with the CSV generation here, I think it doesn't like having a list of pointers. Unless there's any reason not to, can you update this to `Images []ComponentImageStatus`?

nb-ohad: Sorry, my bad.
I have started with an ImageReconcileState than changed to Images but forgot to change the json filed name

jarrpa (approved):
/lgtm

pkg/apis/ocs/v1/storagecluster_types.go
nb-ohad: done

nb-ohad: fixed, not using maps anymore

pkg/apis/ocs/v1/storagecluster_types.go
nb-ohad: done

PR #886, opened 2020-11-04
Remove rgw-endpoint label hack

nb-ohad:
The hack was used to pass the address for an RGW endpoint in external mode before proper support introduced to noobaa operator

Signed-off-by: nb-ohad <mitrani.ohad@gmail.com>

### DIFF
--- a/pkg/controller/storagecluster/external_resources.go
+++ b/pkg/controller/storagecluster/external_resources.go
@@ -34,12 +34,6 @@ const (
 	rookEnableCephFSCSIKey     = "ROOK_CSI_ENABLE_CEPHFS"
 )
 
-var (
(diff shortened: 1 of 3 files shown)


umangachapagain:
> The hack was used to pass the address for an RGW endpoint in external mode before proper support introduced to noobaa operator
> 
Did we upgrade NooBaa Operator to the version that has this support in OCS?

umangachapagain (approved):

leseb (commented):
Can't we just git-revert some commits instead of the manual removal?

nb-ohad:
@leseb 
Not sure, we can try
But I think too much water has flowed under the bridge

jarrpa (approved):
Great! Happy to see this.

/lgtm

PR #888, opened 2020-11-04
hack: build netgo osusergo in binary instead of using clibs

iamniting:
Fix Error version `GLIBC_2.32' not found (required by ocs-operator)
via building static binary.

The Error is seen when the dynamic binary is built on a OS where higher
clibs version is installed and ran on the OS where lower clibs version
is installed.

### DIFF
--- a/hack/go-build.sh
+++ b/hack/go-build.sh
@@ -5,5 +5,5 @@ set -e
 source hack/common.sh
 
 mkdir -p ${OUTDIR_BIN}
-go build -i -ldflags="-s -w" -o ${OUTDIR_BIN}/ocs-operator ./cmd/manager
-go build -i -ldflags="-s -w" -o ${OUTDIR_BIN}/metrics-exporter ./metrics/main.go
+go build -i -tags 'netgo osusergo' -ldflags="-s -w" -o ${OUTDIR_BIN}/ocs-operator ./cmd/manager
+go build -i -tags 'netgo osusergo' -ldflags="-s -w" -o ${OUTDIR_BIN}/metrics-exporter ./metrics/main.go
--- a/hack/go-build.sh
+++ b/hack/go-build.sh
@@ -5,5 +5,5 @@ set -e
 source hack/common.sh
 
 mkdir -p ${OUTDIR_BIN}
-go build -i -tags 'netgo osusergo' -ldflags="-s -w" -o ${OUTDIR_BIN}/ocs-operator ./cmd/manager
-go build -i -tags 'netgo osusergo' -ldflags="-s -w" -o ${OUTDIR_BIN}/metrics-exporter ./metrics/main.go
+go build -tags 'netgo osusergo' -ldflags="-s -w" -o ${OUTDIR_BIN}/ocs-operator ./cmd/manager
+go build -tags 'netgo osusergo' -ldflags="-s -w" -o ${OUTDIR_BIN}/metrics-exporter ./metrics/main.go


iamniting:
cc @raghavendra-talur

iamniting:
@obnoxxx @umangachapagain What are your thoughts on this?

hack/go-build.sh
```diff
@@ -5,5 +5,5 @@ set -e
 source hack/common.sh
 
 mkdir -p ${OUTDIR_BIN}
-go build -i -ldflags="-s -w" -o ${OUTDIR_BIN}/ocs-operator ./cmd/manager
```
raghavendra-talur, line 8: Why did you remove the "-i" flag?

iamniting: The -i flag installs the packages that are dependencies of the target.

Which we have already done IMO and it was causing the below error which I did not found any solution.

```
$ go build -tags 'netgo osusergo' -i -ldflags="-s -w" -o _output/tools/bin/ocs-operator ./cmd/manager
go build os/user: copying /home/nigoyal/.cache/go-build/47/47f22a329c0d011f534ff48e6501350f57139e9dd43c056e0856da7c8d4739e4-d: open /usr/lib/golang/pkg/linux_amd64/os/user.a: permission denied
go build net: copying /home/nigoyal/.cache/go-build/dd/dd6a332ebdbafca3686494fa00f07973256745f3bbd9da76887318b33e5ed44e-d: open /usr/lib/golang/pkg/linux_amd64/net.a: permission denied
```
The above was not working out even after giving sudo and moreover, I don't want to give sudo permissions at all just for building one go program.
```
$ go build -tags 'netgo osusergo' -ldflags="-s -w" -o _output/tools/bin/ocs-operator ./cmd/manager
Success
```

raghavendra-talur: With that explanation, the change is acceptable to me. 

Please include this in the commit message though.

iamniting: will do

umangachapagain: Honestly I am not very comfortable accepting `it was causing the below error which I did not found any solution` as a reason at the moment. At least we should know why we hit that issue? Could it be solved in some other way? Are we sure removing this flag won't backfire?

All these questions because I'm not well versed with Go builds.

jarrpa: Agreed with Umanga, we should do a bit more research. Apart from a more detailed explanation, it should be its own commit.

hack/go-build.sh
```diff
@@ -5,5 +5,5 @@ set -e
...
-go build -i -ldflags="-s -w" -o ${OUTDIR_BIN}/metrics-exporter ./metrics/main.go
+go build -tags 'netgo osusergo' -ldflags="-s -w" -o ${OUTDIR_BIN}/ocs-operator ./cmd/manager
```
jarrpa, line 8: Explain what these tags do.

jarrpa (changes requested):

crombus (changes requested):
I have a little different approach on this see if it works fine.

1) create new file go-build-static.sh add step 2 &3 in this file.
2) using sed commands change the two line in comman.sh
3) run make ocs-operator
4) add this as make ocs-operator-fedora in Makefile

with this approach I think we will not change the existing code and it will solve your problem also.

iamniting:
> I have a little different approach on this see if it works fine.
> 
> 1. create new file go-build-static.sh add step 2 &3 in this file.
> 2. using sed commands change the two line in comman.sh
> 3. run make ocs-operator
> 4. add this as make ocs-operator-fedora in Makefile
> 
> with this approach I think we will not change the existing code and it will solve your problem also.

It is not just a fedora problem or my problem, It is a generic problem with golang dynamic build. You can read about it at [1].

[1]. https://oddcode.daveamit.com/2018/08/16/statically-compile-golang-binary/

crombus:
that is fine if not fedora. you can change the name to make ocs-operator-static.

iamniting:
> that is fine if not fedora. you can change the name to make ocs-operator-static.

I am still in favor of static binaries as static binaries have everything plugged together and no other dependencies are required. So whatever we will build that will be the only thing which will get used and there will be no surprises later.

jarrpa (approved):
Okay, looking through a bit more discussion online, I'm comfortable with this change. The main reason I was confused is that I thought one of the promises of Go was that the binaries were all self-contained! :) But seems that's gotten muddied over the years.

/lgtm

iamniting:
Thanks, @jarrpa and @raghavendra-talur

PR #891, opened 2020-11-05
build: Move the rbac yamls outside deploy/

nbalacha:
The deploy/bundlemanifest directory contains extra rbac yamls.
While building, source-manifests.sh moves these yamls to a different directory 
to prevent OCS CSV from messing up the RBAC in the CSV. They are
later copied back to deploy/bundlemanifests once the CSVs are generated.
If the source-manifests.sh fails after deleting the yamls but before
they are copied back, subsequent builds fail because the yamls are
missing.

This fix moves the rbacs to a directory outside the deploy directory
to prevent the OCS CSV issue so the workaround is no longer needed.

Fixes: #750 


umangachapagain (approved):

iamniting (approved):

jarrpa (changes requested):
Makes sense to me. But for aesthetics, could we just move them to `deploy/rbac`? Or does the SDK search for them there as well?

umangachapagain:
> Makes sense to me. But for aesthetics, could we just move them to `deploy/rbac`? Or does the SDK search for them there as well?

It does. We can't have it under `deploy/`.

jarrpa:
Okay, then rename `extra-rbacs` to just `rbac`.

umangachapagain:
> Okay, then rename `extra-rbacs` to just `rbac`.

I asked for "extra-rbacs" so that we don't confuse it with the default rbacs that we already have in `deploy/`.

jarrpa:
That's a clunky name, and no less vague than just rbac. What's the difference between these and the default rbac definitions?

raghavendra-talur:
> That's a clunky name, and no less vague than just rbac. What's the difference between these and the default rbac definitions?

+1

Does that also mean that these "extra-rbacs" aren't used anywhere and just exist in the repo? When would someone use them and is there a readme for such use-case?

nbalacha:
> > That's a clunky name, and no less vague than just rbac. What's the difference between these and the default rbac definitions?
> 
> +1
> 
> Does that also mean that these "extra-rbacs" aren't used anywhere and just exist in the repo? When would someone use them and is there a readme for such use-case?

They are being used . The rbac directory is passed to the CVS_MERGER --manifests-directory=$BUNDLEMANIFESTS_DIR in the hack/generate-unified-csv.sh

nbalacha:
> That's a clunky name, and no less vague than just rbac. What's the difference between these and the default rbac definitions?

Have we decided on the directory name? If yes, I can update the PR.

umangachapagain:
> > That's a clunky name, and no less vague than just rbac. What's the difference between these and the default rbac definitions?
> 
> Have we decided on the directory name? If yes, I can update the PR.

Let's go ahead with `rbac` for now.

umangachapagain:
> That's a clunky name, and no less vague than just rbac. What's the difference between these and the default rbac definitions?

These are additional rbacs that we manually added into the bundle.
Default rbacs are generated by sdk and are minimal (concerned only with ocs operator deployment).

umangachapagain (approved):
Test failure seems to be an infra issue.
Builds are working fine.

/lgtm

PR #895, opened 2020-11-12
Default managed resources to 'manage'

jarrpa:

### DIFF
--- a/pkg/controller/storagecluster/reconcile.go
+++ b/pkg/controller/storagecluster/reconcile.go
@@ -54,8 +54,8 @@ osd_memory_target_cgroup_limit_ratio = 0.5
 
 	// ReconcileStrategyUnknown is the same as default
 	ReconcileStrategyUnknown ReconcileStrategy = ""
-	// ReconcileStrategyDefault means reconcile once and ignore if it exists
-	ReconcileStrategyDefault ReconcileStrategy = "default"
+	// ReconcileStrategyInit means reconcile once and ignore if it exists
+	ReconcileStrategyInit ReconcileStrategy = "init"
 	// ReconcileStrategyIgnore means never reconcile
 	ReconcileStrategyIgnore ReconcileStrategy = "ignore"
 	// ReconcileStrategyManage means always reconcile
--- a/pkg/controller/storagecluster/storageclasses.go
+++ b/pkg/controller/storagecluster/storageclasses.go
@@ -51,7 +51,7 @@ func (r *ReconcileStorageCluster) createStorageClasses(scs []*storagev1.StorageC
 			return err
 		} else {
 			reconcileStrategy := ReconcileStrategy(instance.Spec.ManagedResources.StorageClasses.ReconcileStrategy)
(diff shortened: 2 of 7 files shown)


obnoxxx (approved):
/lgtm

iamniting (approved):

pkg/controller/storagecluster/reconcile.go
```diff
@@ -54,8 +54,8 @@ osd_memory_target_cgroup_limit_ratio = 0.5
 
 	// ReconcileStrategyUnknown is the same as default
```
nbalacha, line 55: Does this mean that ReconcileStrategyUnknown is now the same as ReconcileStrategyManage?

pkg/controller/storagecluster/storageclasses.go
```diff
@@ -51,7 +51,7 @@ func (r *ReconcileStorageCluster) createStorageClasses(scs []*storagev1.StorageC
 			return err
 		} else {
 			reconcileStrategy := ReconcileStrategy(instance.Spec.ManagedResources.StorageClasses.ReconcileStrategy)
-			if reconcileStrategy == ReconcileStrategyDefault || reconcileStrategy == ReconcileStrategyUnknown {
+			if reconcileStrategy == ReconcileStrategyInit {
```
nbalacha, line 54: Shouldn't this also return without reconciling if reconcileStrategy is ignore?
Is ReconcileStrategyStandalone only applicable to Nooba?

Ditto for the other changes.

nbalacha: The delete<resource> function for each of these resources should check if the reconcileStrategy is ReconcileStrategyInit before attempting to delete them.

obnoxxx: > Shouldn't this also return without reconciling if reconcileStrategy is ignore?

If we have `ReconcileStrategyIgnore`, then we don't even call this function. This is checked above in `ensureStorageClasses`.  For the other cases, the `Ignore` case is checked further up in the same function. In general: 

* ignore ==> don't call the reconcile logic at all
* init ==> only create if it does not exist
* manage ==> create and update

> Is ReconcileStrategyStandalone only applicable to Nooba?

Correct. @nb-ohad wanted it to behave differently than the other resources.



> The delete function for each of these resources should check if the reconcileStrategy is ReconcileStrategyInit before attempting to delete them.

Hmmm, interesting point. We currently only delete the storage resources in uninstall, not in reconciliation, as it seems. We might need to talk about it, but it is not a new thing.

obnoxxx (approved):

pkg/controller/storagecluster/storageclasses.go
jarrpa: > Shouldn't this also return without reconciling if reconcileStrategy is ignore?
> Is ReconcileStrategyStandalone only applicable to Nooba?

It already does at the beginning of the function.

> The delete function for each of these resources should check if the reconcileStrategy is ReconcileStrategyInit before attempting to delete them.

I would say no, we should be deleting everything during uninstall. There's no reason for the CRs to remain when the StorageCluster is gone.

PR #896, opened 2020-11-16
must-gather: Use awk to find Ready nodes only

crombus:
Use awk to find only Ready nodes as NotReady
nodes has Ready so there is no distinction.

### DIFF
--- a/must-gather/collection-scripts/gather_ceph_resources
+++ b/must-gather/collection-scripts/gather_ceph_resources
@@ -210,11 +210,19 @@ for ns in $namespaces; do
 
     crash_collection(){
         oc debug nodes/"${node}" --to-namespace="${ns}" -- bash -c "sleep 5m" &
-        while [ "$(oc get pods -n openshift-storage | grep "${node//./}-debug" | awk '{print $2}')" != "1/1" ] ; do
-            sleep 3
-            echo "waiting for the debug pod to be in ready state"
+        debug_pod_ready=false
+        for i in {0..300..3}; do
+            if [ "$(oc get pods -n openshift-storage | grep "${node//./}-debug" | awk '{print $2}')" != "1/1" ] ; then
+                debug_pod_ready=true
+                break
+            else
+                sleep 3
+                echo "waiting for the debug pod to be in ready state"
+            fi
         done
-        oc rsync -n "${ns}" "$(oc get pods -n "${ns}"| grep "${node//./}-debug"| awk '{print $1}')":/host/var/lib/rook/openshift-storage/crash/ "${CRASH_OUTPUT_DIR}"
+        if $debug_pod_ready; then
+            oc rsync -n "${ns}" "$(oc get pods -n "${ns}"| grep "${node//./}-debug"| awk '{print $1}')":/host/var/lib/rook/openshift-storage/crash/ "${CRASH_OUTPUT_DIR}"
+        fi
     }
 
     # creating a counter variable for collecting PID in array
--- a/must-gather/collection-scripts/gather_ceph_resources
+++ b/must-gather/collection-scripts/gather_ceph_resources
@@ -191,8 +191,11 @@ for ns in $namespaces; do
         done
     done
 
+    # Add Ready nodes to the list
+    nodes=$(oc get nodes -l cluster.ocs.openshift.io/openshift-storage='' --no-headers | awk '/\yReady\y/{print $1}')
+
     # Collecting ceph prepare volume logs
-    for node in $(oc get nodes -l cluster.ocs.openshift.io/openshift-storage='' --no-headers | grep -w 'Ready' | awk '{print $1}'); do
+    for node in ${nodes}; do
         printf "collecting prepare volume logs from node %s \n"  "${node}" | tee -a  "${BASE_COLLECTION_PATH}"/gather-debug.log
         NODE_OUTPUT_DIR=${CEPH_COLLECTION_PATH}/namespaces/${ns}/osd_prepare_volume_logs/${node}
         mkdir -p "${NODE_OUTPUT_DIR}"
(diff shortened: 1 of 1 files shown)


crombus:
@leseb NotReady has Ready so NotReady nodes was also getting in the list.

leseb (approved):
lgtm

jarrpa:
Could you still use awk but use the `\y` flag before and after Ready? https://www.gnu.org/software/gawk/manual/html_node/GNU-Regexp-Operators.html

crombus:
> Could you still use awk but use the `\y` flag before and after Ready? https://www.gnu.org/software/gawk/manual/html_node/GNU-Regexp-Operators.html

@jarrpa the link you shared it is to be used only with gawk not awk. if we want to use gawk that needs to be installed in must-gather pod.

iamniting (approved):

jarrpa:
> > Could you still use awk but use the `\y` flag before and after Ready? https://www.gnu.org/software/gawk/manual/html_node/GNU-Regexp-Operators.html
> 
> @jarrpa the link you shared it is to be used only with gawk not awk. if we want to use gawk that needs to be installed in must-gather pod.

`awk` is mapped to `gawk` in RH OSes. You can run `awk --help` inside the must-gather container to verify, which I just did.

crombus:
> > > Could you still use awk but use the `\y` flag before and after Ready? https://www.gnu.org/software/gawk/manual/html_node/GNU-Regexp-Operators.html
> > 
> > 
> > @jarrpa the link you shared it is to be used only with gawk not awk. if we want to use gawk that needs to be installed in must-gather pod.
> 
> `awk` is mapped to `gawk` in RH OSes. You can run `awk --help` inside the must-gather container to verify, which I just did.

Yup working fine with awk used \y .  Thanks, I was not aware of this.

must-gather/collection-scripts/gather_ceph_resources
```diff
@@ -217,7 +217,7 @@ for ns in $namespaces; do
     # creating a counter variable for collecting PID in array
     idx=0
     # Collecting ceph crash dump
-    for node in $(oc get nodes -l cluster.ocs.openshift.io/openshift-storage='' --no-headers | awk '/Ready/{print $1}'); do
+    for node in $(oc get nodes -l cluster.ocs.openshift.io/openshift-storage='' --no-headers |awk '/\yReady\y/{print $1}'); do
```
jarrpa, line 226: Add back the whitespace removed here.

must-gather/collection-scripts/gather_ceph_resources
```diff
@@ -207,9 +207,15 @@ for ns in $namespaces; do
...
-            echo "waiting for the debug pod to be in ready state"
+            wait_period=$wait_period+3
+            if [ $wait_period -gt 300 ];then
+                break
+            else
+                sleep 3
+                echo "waiting for the debug pod to be in ready state"
+            fi
         done
         oc rsync -n "${ns}" "$(oc get pods -n "${ns}"| grep "${node//./}-debug"| awk '{print $1}')":/host/var/lib/rook/openshift-storage/crash/ "${CRASH_OUTPUT_DIR}"
```
jarrpa, line 220: If the debug Pod never comes up, we shouldn't try to run the rsync command. Maybe track another variable `debug_pod_timeout=false` and set it to `true` if we time out. Alternatively, you could restructure the while loop as follows:

```bash
debug_pod_ready=false
for i in {0..300..3); do
  if <check_debug_pod>; then
    debug_pod_ready=true
    break
  else
    sleep 3
  fi
done

if debug_pod_ready; then
  <oc_rsync>
fi
```

must-gather/collection-scripts/gather_ceph_resources
jarrpa: Actually, since this and the previous for conditional are now the same, could you perform the check once and save the results into an array? Then you can just loop over the array and save a call to `oc get`.

jarrpa (changes requested):

must-gather/collection-scripts/gather_ceph_resources
```diff
@@ -191,9 +191,13 @@ for ns in $namespaces; do
...
     done
 
+    # Add Ready nodes to the list
+    nodes=()
+
     # Collecting ceph prepare volume logs
-    for node in $(oc get nodes -l cluster.ocs.openshift.io/openshift-storage='' --no-headers | grep -w 'Ready' | awk '{print $1}'); do
+    for node in $(oc get nodes -l cluster.ocs.openshift.io/openshift-storage='' --no-headers | awk '/\yReady\y/{print $1}'); do
         printf "collecting prepare volume logs from node %s \n"  "${node}" | tee -a  "${BASE_COLLECTION_PATH}"/gather-debug.log
+        nodes+=("$(node)")
```
jarrpa, line 200: You shouldn't need to do this. Can't you just save the output of the command to `nodes` directly and just loop through it?

must-gather/collection-scripts/gather_ceph_resources
obnoxxx: this has been done

must-gather/collection-scripts/gather_ceph_resources
obnoxxx: this has been done

must-gather/collection-scripts/gather_ceph_resources
crombus: yes

must-gather/collection-scripts/gather_ceph_resources
obnoxxx: This is the last remaining request. It should work roughly like this:
```
nodes=$(oc get nodes --no-headers | awk '/\yReady\y/{print $1}')
for node in $nodes ; do
    ...
done
```

crombus:
@obnoxxx no need to disable shellcheck adding just the brackets works fine for the loop also and shellcheck is all statisfied.

obnoxxx (approved):
/lgtm

thanks for the update

PR #905, opened 2020-11-22
Adding KMS support to OCS Operator

aruniiird:
[//]: # (Please describe your changes in detail)
[//]: # (If your changes affect the CR fields, a syntax for the CR change should be provided)
[//]: # (If a bug is fixed then the behavior change of before and after should be mentioned)
[//]: # (Be sure to run ocs-operator ci before pushing `make ocs-operator-ci`)

**Summary:**

Adding KMS support through OCS Operator

**Description of your changes:**

OCS Operator will check whether KMS (Key Management System) feature is enabled or not through
StorageCluster Spec (under 'Encryption' -> 'KeyManagementService' -> 'Enable' = true).
If KMS is enabled, operator will look for a configmap, named "ocs-kms-connection-details",
fetch details from it and update the 'CephCluster' Spec. Once successfully updated,
rook will access the necessary info through the CephCluster Spec and in turn try to store
the LUKS's (Linux Unified Key Setup) Key Encryption Key (KEK) inside a KMS.

Main pre-requisite for this feature is a valid KMS Service provider should be
running and should be reachable from other nodes in the cluster.

More internal details could be found at: https://github.com/rook/rook/pull/6474

**Which issue is resolved by this Pull Request:**
Resolves #

**Checklist:**

- [ ] Documentation has been updated, if necessary.
- [x] Unit tests have been added, if necessary.
- [x] Generate CSV (`make gen-latest-csv`)
- [ ] Raise Integration tests issue if necessary.
- [x] Code generation (`make update-generated`) has been run to update object specifications, if necessary.


go.mod
```diff
@@ -23,7 +23,7 @@ require (
 	github.com/operator-framework/operator-sdk v0.17.0
 	github.com/prometheus/client_golang v1.7.1
 	github.com/prometheus/client_model v0.2.0
-	github.com/rook/rook v1.4.6
+	github.com/rook/rook v1.5.1
```
obnoxxx, line 26: FYI the update of rook to v1.5.1 is now covered in a separate PR #912

nbalacha:
Questions:
1. How is this used?
2. Does this need to be deleted if StorageCluster is deleted?

go.mod
aruniiird: Thanks @obnoxxx , once the initial testing is complete, will re-base my PR on top of that...

aruniiird:
> Questions:
> 
>     1. How is this used?
> 


@nbalacha , for KMS support, OCS-Operator has to pass two things to `CephCluster` spec (which rook will pick once it is updated),
i. a map with KMS connection / configuration details
ii. a generated token from the KMS.

So we are adding the above two items into the `StorageCluster` spec (and the values are being passed from the UI).

A general how KMS is being used could be found here: https://kubernetes.io/docs/tasks/administer-cluster/kms-provider/

>     2. Does this need to be deleted if StorageCluster is deleted?

In my opinion, we are not creating any specific resource with KMS. We are only updating the CephCluster spec. So it should be dealt once `CephCluster` is deleted. 
@leseb , could answer it better. ^^^

nbalacha:
> > Questions:
> > ```
> > 1. How is this used?
> > ```
> 
> @nbalacha , for KMS support, OCS-Operator has to pass two things to `CephCluster` spec (which rook will pick once it is updated),
> i. a map with KMS connection / configuration details
> ii. a generated token from the KMS.

Please add this information to the commit message.  It will be useful in the future.

pkg/controller/storagecluster/kms_resources.go
```diff
@@ -0,0 +1,56 @@
...
+			Name:      kmsConfigMapName,
+			Namespace: instance.ObjectMeta.Namespace,
+		},
+		kmsConfigMap,
+	)
+	if err != nil {
+		reqLogger.Error(err, fmt.Sprintf("Unable to get '%s' config map", kmsConfigMapName))
+		return err
+	}
+	found.Spec.Security.KeyManagementService.ConnectionDetails = kmsConfigMap.Data
```
nbalacha, line 53: Why not do this at the time the CephCluster object is created?Is there a reason this is being done separately?

pkg/controller/storagecluster/kms_resources.go
```diff
@@ -0,0 +1,56 @@
...
+	"k8s.io/apimachinery/pkg/types"
+)
+
+func (r *ReconcileStorageCluster) ensureKMSSystem(instance *ocsv1.StorageCluster, reqLogger logr.Logger) error {
+	// if there is no 'Security' spec to inspect, return silently
+	if instance.Spec.Security == nil {
+		return nil
+	}
+	if instance.Spec.Security.KeyManagementService != nil &&
+		instance.Spec.Security.KeyManagementService.ConnectionDetails == "" {
```
leseb, line 22: I'd extract the last condition and make it its own `if` statement, it's clearer/easier to read.

pkg/controller/storagecluster/kms_resources.go
```diff
@@ -0,0 +1,56 @@
...
+			Name:      generateNameForCephCluster(instance),
+			Namespace: instance.Namespace,
+		},
+		found,
+	)
+	if err != nil {
+		reqLogger.Error(err, fmt.Sprintf("CephCluster yet to start."))
+		return err
+	}
+	kmsConfigMapName := strings.TrimSpace(instance.Spec.Security.KeyManagementService.ConnectionDetails)
```
leseb, line 40: Let check the CM before fetching the `CephCluster` so if it fails, it's one less API call... (no need to call for `CephCluster`)

pkg/controller/storagecluster/kms_resources.go
leseb: Agreed, let's just build the encryption config altogether with the other CR options.

leseb (changes requested):

pkg/controller/storagecluster/kms_resources.go
nbalacha: Yes. This can be added to newCephCluster(). No need for another ensure function if no new CR is being created.

pkg/controller/storagecluster/kms_resources.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+	// return error, if user enables KMS without enabling 'Encryption' option
+	if instance.Spec.Encryption.KeyManagementService.Enable && !instance.Spec.Encryption.Enable {
+		return fmt.Errorf("KeyManagementService cannot be enabled without enabling 'Encryption'")
+	}
+	if cephCluster == nil {
+		return fmt.Errorf("Empty CephCluster provided")
+	}
+	ctx := context.TODO()
+	kmsConfigMap := &corev1.ConfigMap{}
```
umangachapagain, line 35:
```suggestion
	kmsConfigMap := corev1.ConfigMap{}
```

pkg/controller/storagecluster/kms_resources.go
```diff
@@ -0,0 +1,51 @@
...
+	err := r.client.Get(ctx,
+		types.NamespacedName{
+			Name:      KMSConfigMapName,
+			Namespace: instance.ObjectMeta.Namespace,
+		},
+		kmsConfigMap,
```
umangachapagain, line 41:
```suggestion
		&kmsConfigMap,
```

pkg/controller/storagecluster/kms_resources.go
```diff
@@ -0,0 +1,51 @@
...
+	)
+	if err != nil {
+		reqLogger.Error(err, fmt.Sprintf("%s. Unable to get '%s' config map",
```
umangachapagain, line 44:
```suggestion
		reqLogger.Error(err, fmt.Sprintf("%s. Unable to get %q config map.",
```

umangachapagain (changes requested):

go.mod
aruniiird: PR rebased on top of Umanga's rook updation PR

pkg/controller/storagecluster/kms_resources.go
aruniiird: Yup I agree and made the change...

There were questions around here like,
a. What if we change the data in config map (like VAULT_ADDR, PORT etc), we are not watching the objects?
b. What happens when token or config map get deleted?

pkg/controller/storagecluster/kms_resources.go
aruniiird: Simplified this part. Now KMS will be enabled under `Encryption` spec (no need to pass the token or configmap names).
Spec yaml may look something like this. 

```yaml
encryption:
    enable: true
    kms:
        enable: true
```
When `kms` is enabled, operator will check for the predefined configmap and token names, and update the CephCluster accordingly.

@nbalacha ^^^

pkg/controller/storagecluster/kms_resources.go
aruniiird: Changed this part, please take a look.

pkg/controller/storagecluster/kms_resources.go
aruniiird: Done

pkg/controller/storagecluster/kms_resources.go
aruniiird: Done

pkg/controller/storagecluster/kms_resources.go
aruniiird: Done

pkg/controller/storagecluster/kms_resources.go
```diff
@@ -0,0 +1,51 @@
...
+		return nil
+	}
+	// return error, if user enables KMS without enabling 'Encryption' option
+	if instance.Spec.Encryption.KeyManagementService.Enable && !instance.Spec.Encryption.Enable {
+		return fmt.Errorf("KeyManagementService cannot be enabled without enabling 'Encryption'")
+	}
+	if cephCluster == nil {
+		return fmt.Errorf("Empty CephCluster provided")
+	}
+	ctx := context.TODO()
```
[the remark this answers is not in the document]
aruniiird: Done

pkg/controller/storagecluster/kms_resources.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+func (r *ReconcileStorageCluster) addKmsToCephCluster(instance *ocsv1.StorageCluster, cephCluster *cephv1.CephCluster, reqLogger logr.Logger) error {
+	// if 'KMS' is not enabled, return silently
+	if !instance.Spec.Encryption.KeyManagementService.Enable {
+		return nil
+	}
+	// return error, if user enables KMS without enabling 'Encryption' option
+	if instance.Spec.Encryption.KeyManagementService.Enable && !instance.Spec.Encryption.Enable {
+		return fmt.Errorf("KeyManagementService cannot be enabled without enabling 'Encryption'")
```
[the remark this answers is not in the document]
aruniiird: Removed this check, as I later understood that KMS can be enabled without enabling the `Encryption` field.

pkg/controller/storagecluster/kms_resources.go
```diff
@@ -0,0 +1,51 @@
...
+	KMSTokenSecretName = "ocs-kms-token"
+)
+
+func (r *ReconcileStorageCluster) addKmsToCephCluster(instance *ocsv1.StorageCluster, cephCluster *cephv1.CephCluster, reqLogger logr.Logger) error {
+	// if 'KMS' is not enabled, return silently
+	if !instance.Spec.Encryption.KeyManagementService.Enable {
+		return nil
+	}
+	// return error, if user enables KMS without enabling 'Encryption' option
+	if instance.Spec.Encryption.KeyManagementService.Enable && !instance.Spec.Encryption.Enable {
```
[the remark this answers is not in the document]
aruniiird: Removed this condition, as it no longer applies.

pkg/controller/storagecluster/kms_resources.go
```diff
@@ -0,0 +1,51 @@
...
+const (
+	// KMSConfigMapName configmap name which KMS config details
+	KMSConfigMapName = "ocs-kms-connection-details"
+	// KMSTokenSecretName name of the secret which has KMS token details
+	KMSTokenSecretName = "ocs-kms-token"
+)
+
+func (r *ReconcileStorageCluster) addKmsToCephCluster(instance *ocsv1.StorageCluster, cephCluster *cephv1.CephCluster, reqLogger logr.Logger) error {
+	// if 'KMS' is not enabled, return silently
+	if !instance.Spec.Encryption.KeyManagementService.Enable {
```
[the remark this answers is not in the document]
aruniiird: Not really. This check safe guards the function from using rather a convention based approach of adding an extra `if` condition at every place whereever this function is being called. Something like,
```go
if sc.Spec.Encryption.KeyManagementService.Enable {
    r.addKmsToCephCluster(sc, cephCluster, reqLogger)
}
```

pkg/controller/storagecluster/kms_resources.go
```diff
@@ -0,0 +1,51 @@
...
+	ocsv1 "github.com/openshift/ocs-operator/pkg/apis/ocs/v1"
+	cephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/go-logr/logr"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+const (
+	// KMSConfigMapName configmap name which KMS config details
```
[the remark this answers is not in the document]
aruniiird: Done

pkg/controller/storagecluster/cephcluster.go
```diff
@@ -64,7 +64,11 @@ func (r *ReconcileStorageCluster) ensureCephCluster(sc *ocsv1.StorageCluster, re
 	if sc.Spec.ExternalStorage.Enable {
 		cephCluster = newExternalCephCluster(sc, r.cephImage, r.monitoringIP)
 	} else {
-		cephCluster = newCephCluster(sc, r.cephImage, r.nodeCount, r.serverVersion, reqLogger)
+		var err error
+		cephCluster, err = r.newCephClusterWithKMS(sc, r.cephImage, r.nodeCount, r.serverVersion, reqLogger)
```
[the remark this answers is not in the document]
aruniiird: I tried different scenarios and finally arrived at this change. The main difficulty is that we have to change the existing functional type of the current `newCephCluster` function (which is being used in multiple places, including in different tests).

For the above requirement, TWO main changes are needed to the existing function,

a. We require an instance of `ReconcileStorageCluster`, which is needed to retrieve KMS details. So, either we have to pass it as an argument or add this function as a method of the `ReconcileStorageCluster` object. Both of which changes the existing functional type of `newCephCluster`.

b. Once we decide to add/attach KMS details to the CephCluster, we have to return the error value as well along with `CephCluster` object, which again changes functional type.

In the current way, the existing function is not changed. If user / developer wants a KMS attached CephCluster they have to use `ReconcileStorageCluster`'s method `newCephClusterWithKMS`.

pkg/controller/storagecluster/kms_resources.go
```diff
@@ -0,0 +1,51 @@
...
+	KMSConfigMapName = "ocs-kms-connection-details"
+	// KMSTokenSecretName name of the secret which has KMS token details
+	KMSTokenSecretName = "ocs-kms-token"
+)
+
+func (r *ReconcileStorageCluster) addKmsToCephCluster(instance *ocsv1.StorageCluster, cephCluster *cephv1.CephCluster, reqLogger logr.Logger) error {
```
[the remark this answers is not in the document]
aruniiird: Even if we add `getKMSConfig` we have to deal with the error which it returns.

If you see, in the existing code it is decoupled. `newCephCluster()` function is a standalone function which returns a healthy cephcluster. If we want KMS details to be attached, use `ReconcileStorageCluster`'s method. I have tried to explain the difficulty in clubbing these two functionalities (of creating a cephcluster and attaching KMS to it).

pkg/controller/storagecluster/cephcluster.go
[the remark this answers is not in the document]
umangachapagain: > I tried different scenarios and finally arrived at this change. The main difficulty is that we have to change the existing functional type of the current `newCephCluster` function (which is being used in multiple places, including in different tests).
>
Honestly, I'd be fine with that. Specially, if it means improved code readability and clean & clear code.
 
> For the above requirement, TWO main changes are needed to the existing function,
> 
> a. We require an instance of `ReconcileStorageCluster`, which is needed to retrieve KMS details. So, either we have to pass it as an argument or add this function as a method of the `ReconcileStorageCluster` object. Both of which changes the existing functional type of `newCephCluster`.
> 
> b. Once we decide to add/attach KMS details to the CephCluster, we have to return the error value as well along with `CephCluster` object, which again changes functional type.
> 
> In the current way, the existing function is not changed. If user / developer wants a KMS attached CephCluster they have to use `ReconcileStorageCluster`'s method `newCephClusterWithKMS`.
>
I see what you mean. But, still I'm not very comfortable with multiple generator functions for the same object. And, `KMS` defines behaviour of `CephCluster` and not `ReconcileStorageCluster`, so I would prefer not to have `r.newCephClusterWithKMS()` or `r.addKmsToCephCluster()`.

Can we do this instead:
We retain `newCephCluster()` that returns a basic `CephCluster` object. We then add a method to modify it's behaviour i.e `CephCluster.addKMS(client r.client)` and the call to this modifier will be safeguarded which shouldn't be that bad as we'll do it once. 

So code would look something like
```
cephCluster = newCephCluster(sc, r.cephImage, r.nodeCount, r.serverVersion, reqLogger)
if sc.Spec.Encryption.KeyManagementService.Enable{
 err := cephCluster.addKMS(r.client, ...)
  if err != nil{
     //reconcile
  }
}
```
 Wouldn't this be more manageable, read better and share less data?

pkg/controller/storagecluster/kms_resources.go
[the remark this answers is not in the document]
umangachapagain: I get your point.

I'd argue that we make this call once throughout the code base and making such call should be a conscious decision (which means if..else before making the call). But, that's just my opinion.

pkg/controller/storagecluster/kms_resources.go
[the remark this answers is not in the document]
umangachapagain: It's fine for a get function to return error and the caller to handle it. That's what every get call in k8s does.

umangachapagain (changes requested):

pkg/controller/storagecluster/kms_resources.go
```diff
@@ -0,0 +1,78 @@
...
+	if kmsConfigMapData == nil {
+		return fmt.Errorf("Please provide a valid configmap data as argument.")
+	}
+	kmsProviderName, ok := kmsConfigMapData[KMSProviderKey]
+	if !ok {
+		return nil
+	}
+	// the provider's address key should be in the format, ${PROVIDER_NAME_IN_CAPS}_ADDR.
+	// For example, for 'vault' KMS provider, the address key will be, 'VAULT_ADDR'
+	kmsProviderAddressKey := fmt.Sprintf("%s_ADDR", strings.ToUpper(kmsProviderName))
```
leseb, line 72: We have 0 guarantee that other KMS implementation will have a `"%s_ADDR"` format so I would `switch case` on `KMSProviderKey` and if `VaultKMSProvider` we use `VAULT_ADDR`.

aruniiird: I was trying to avoid any dependency between a new KMS provider addition and code change (at least in OCS-Operator code changes). If we use a conditional approach (adding `if` or `switch` according to the KMS provider), as you mentioned, for any new KMS provider we will have to make code change in the operator.

Another approach (inorder to avoid code changes) is to make the address key (in the configmap data) agnostic to KMS provider, that means for any provider we use the same key, let say `KMS_ENDPOINT` (or something like that).

Let me know your opinion, I'm ok to make changes either way (conditonal or any of the above way).

pkg/controller/storagecluster/cephcluster.go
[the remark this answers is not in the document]
aruniiird: > I see what you mean. But, still I'm not very comfortable with multiple generator functions for the same object.

This is a common scenario in `golang` where you use different functions to initialize an object in different ways.
Golang's [context](https://golang.org/pkg/context/) package is an example, where we have functions like[
WithCancel](https://golang.org/pkg/context/#WithCancel) , [WithDeadline](https://golang.org/pkg/context/#WithDeadline), etc functions which returns different `Context`s according to the input/argument `Context`.
More such functions could be found with `New***()` names, where objects are created according to different arguments passed (something like `NewBuffer(buf []byte) *Buffer` ,  `NewBufferString(s string) *Buffer` etc)

> And, KMS defines behaviour of CephCluster and not ReconcileStorageCluster, so I would prefer not to have r.newCephClusterWithKMS() or r.addKmsToCephCluster().

I totally agree with you.

> We retain newCephCluster() that returns a basic CephCluster object. We then add a method to modify it's behaviour i.e CephCluster.addKMS(client r.client) and the call to this modifier will be safeguarded which shouldn't be that bad as we'll do it once.


`CephCluster` is an object from `github.com/rook/rook/pkg/apis/ceph.rook.io/v1` (nicknamed `cephv1`) and we won't be able to add functions into it directly from our operator code base. Or we have to create a type alias and add the function to it, which complicates the matter.
Another question I have is, if we already know that we require a `CephCluster` with the KMS details, why not make both the lines (creation of cephcluster and the safeguard `if` condition following it) together into a single function and just call it. This will be a much safer approach and this function can be reused in places where we require a `CephCluster` with KMS details (let's say, in writting test cases).

If you are convinced about my first point (about different functions being used to initialize same object and it is a common practice in `golang` world), we can remove all the `ReconcileStorageCluster` methods (that is `r.methods`) and put them as standard methods. So there will be TWO methods to create converged mode `CephCluster`.

a. `newCephCluster` : the existing method without any change
b. `newCephClusterWithKMS`: a new standalone method (like the above one) with extra `client` parameter passed. Along with `CephCluster`, this will return an additional error value.


If you are still not convinced, we have to go with the longer approach of changing the existing `newCephCluster` function (which will now take a `client` and returns an additional error). The con/down-side with this approach is all the test files, which uses this method, have to be changed and additionally we won't be able to create a ceph cluster (which does not have KMS enabled) without using a `client`.

My recommendation will be to add two methods instead of one.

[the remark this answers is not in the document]
aruniiird: I have updated PR (on the agreed point that we should remove the methods from the `ReconcileStorageCluster` object and made them standalone).

pkg/controller/storagecluster/kms_resources.go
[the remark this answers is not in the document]
jarrpa: This is actually in keeping with our ensure functions, which we call blindly and then check for booleans at the top of the function. I'm okay with this.

jarrpa (changes requested):
Rather than going the `newCephClusterWithKMS()` route, I would prefer to go with Umanga's suggestion of modifying `newCephCluster()`, but without needing to pass in the kube client. You could call `getKMSConfigMap()` ahead of time, and then pass in the ConfigMap pointer (or nil if it's disabled). It's fine if we have to update a bunch of other places by changing the function signature if we maintain better code readability.

Apart from that, I think we're still missing a fundamental explanation of what this feature is and what it does. Please add a sentence or so to the initial commit message explaining what KMS support is aiming to do. "do the needful" is astoundingly vague. :)

pkg/controller/storagecluster/kms_resources.go
leseb: > I was trying to avoid any dependency between a new KMS provider addition and code change (at least in OCS-Operator code changes). If we use a conditional approach (adding `if` or `switch` according to the KMS provider), as you mentioned, for any new KMS provider we will have to make code change in the operator.

Yes, I understood that.

> 
> Another approach (inorder to avoid code changes) is to make the address key (in the configmap data) agnostic to KMS provider, that means for any provider we use the same key, let say `KMS_ENDPOINT` (or something like that).
> 

If make it agnostic, how do we configure the KMS?
Are you saying we should translate `KMS_ENDPOINT` to the right value but inside the Rook code?

> Let me know your opinion, I'm ok to make changes either way (conditonal or any of the above way).

I still think we should go with the conditional since what we have now will only work for Vault.
Thanks

aruniiird: Yes changed it to a map, thus user/developer has to just add the KMS provider and address key into the map and no need to edit the code further.

aruniiird:
> Rather than going the `newCephClusterWithKMS()` route, I would prefer to go with Umanga's suggestion of modifying `newCephCluster()`, but without needing to pass in the kube client. You could call `getKMSConfigMap()` ahead of time, and then pass in the ConfigMap pointer (or nil if it's disabled). It's fine if we have to update a bunch of other places by changing the function signature if we maintain better code readability.
> 


Yes, now I think finally got what you and Umanga where saying (=
Changes are added, @umangachapagain , @jarrpa , please take a look


> Apart from that, I think we're still missing a fundamental explanation of what this feature is and what it does. Please add a sentence or so to the initial commit message explaining what KMS support is aiming to do. "do the needful" is astoundingly vague. :)

Added some more details to the PR commit message. @jarrpa please take a look...

aruniiird:
Added basic tests as well...

nbalacha:
> Signed-off-by: Arun Kumar Mohan [amohan@redhat.com](mailto:amohan@redhat.com)

Please add more details to the commit message.

pkg/controller/storagecluster/kms_resources.go
```diff
@@ -0,0 +1,86 @@
...
+	kmsConfigMap := corev1.ConfigMap{}
+	err := client.Get(context.TODO(),
+		types.NamespacedName{
+			Name:      KMSConfigMapName,
+			Namespace: instance.ObjectMeta.Namespace,
+		},
+		&kmsConfigMap,
+	)
+	// if validation function is provided, use it
+	if kmsValidateFunc != nil {
```
nbalacha, line 58: Should also check if client.Get() returned an error.

pkg/controller/storagecluster/kms_resources.go
```diff
@@ -0,0 +1,86 @@
...
+		err = kmsValidateFunc(&kmsConfigMap)
+	}
+	return &kmsConfigMap, err
+}
+
+// isKMSProviderReachable function checks whether the provided address is reachable or not.
+// This function won't validate any other cases and only returns an error if the provided
+// KMS provider address is not reachable. All other validations will be done on rook side.
+func isKMSProviderReachable(kmsConfigMap *corev1.ConfigMap) error {
```
nbalacha, line 67: A better func signature would be to return a bool as well as the err

nbalacha (changes requested):

pkg/controller/storagecluster/kms_resources.go
jarrpa: +1

pkg/controller/storagecluster/kms_resources.go
jarrpa: +1

pkg/controller/storagecluster/cephcluster_test.go
```diff
@@ -350,3 +353,75 @@ func TestStorageClassDeviceSetCreation(t *testing.T) {
...
+	cm.Name = KMSConfigMapName
+	cm.Data = make(map[string]string)
+	cm.Data["KMS_PROVIDER"] = kmsProvider
+	cm.Data["VAULT_ADDR"] = kmsAddr
+	cm.Data["VAULT_BACKEND_PATH"] = "ocs"
+	cm.Data["VAULT_NAMESPACE"] = "my-ocs-namespace"
+	return cm
+}
+
+func TestKMSConfigChanges(t *testing.T) {
```
jarrpa, line 368: Please refactor this test to use case tables.

jarrpa (changes requested):
Code is almost done, and the new tests should be refactored. Apart from that, the initial commit message should have some changes and it should be set as the PR description:

```diff
- OCS Operator will check whether KMS (Key Management System) feature is enabled or not through
- StorageCluster Spec (under 'Encryption' -> 'KeyManagementService' -> 'Enable' = true).
- If KMS is enabled, operator will look for a configmap, named "ocs-kms-connection-details",
- fetch details from it and update the 'CephCluster' Spec. Once successfully updated,
- rook will access the necessary info through the CephCluster Spec and in turn try to store
- the LUKS's (Linux Unified Key Setup) Key Encryption Key (KEK) inside a KMS.
+ OCS Operator will check whether KMS (Key Management System) feature is enabled or not through
+ StorageCluster Spec (under 'Encryption' -> 'KeyManagementService' -> 'Enable' = true).
+ If KMS is enabled, the operator will look for a ConfigMap, named "ocs-kms-connection-details",
+ fetch details from it and update the CephCluster Spec. Once successfully updated,
+ rook will access the necessary info through the CephCluster Spec and in turn try to store
+ the LUKS (Linux Unified Key Setup) Key Encryption Key (KEK) inside a KMS.
```

pkg/controller/storagecluster/kms_resources.go
aruniiird: @nbalacha , the bool value returned won't have much use, right?. Return `error` itself will show that whether reachable function is successful (absence of error) or not (presence of error), and this is a common practice in golang.

In the case where no error is being raised during the execution of function (something like `isNameSame(name1, name2 string) bool`), yes, then definitely return a bool value (no need of an extra error).

A scenario where I would have used multiple return values would be, if I was returning a set of URLs which are not reachable along with the error.

A more strict go convention would be to remove `is` prefix from `isKMSProviderReachable` and make the function name further simpler (like `kmsProviderReachable`). Then we use multiple `is` prefixed functions, something like `isValidURL(err error) bool` or `isValidProtocol(err error) bool` etc; to figure out what is the reason behind the returned error.

Golang's [os package](https://golang.org/pkg/os/) has more such examples.

jarrpa: > A more strict go convention would be to remove `is` prefix from `isKMSProviderReachable` and make the function name further simpler (like `kmsProviderReachable`). Then we use multiple `is` prefixed functions, something like `isValidURL(err error) bool` or `isValidProtocol(err error) bool` etc; to figure out what is the reason behind the returned error.

That also sounds good!

pkg/controller/storagecluster/kms_resources.go
aruniiird: Yup, missed that... Changed...

pkg/controller/storagecluster/cephcluster_test.go
aruniiird: Done

aruniiird:
@jarrpa , I have converted the tests into table format and updated the PR comments.
Please take a look...

pkg/controller/storagecluster/cephcluster_test.go
jarrpa: The general idea is there, but the format is not ideal. There should be a struct of parameters, followed by a list of parameters for individual test cases, and then one for loop to run the same exact test over all values.

aruniiird: Updated the code...

jarrpa: You still have two separate loops. They seem like they could be one loop with a small if statement.

jarrpa: Also, the test parameters usually take a case number with a descriptor of what the case is testing

aruniiird: > You still have two separate loops. They seem like they could be one loop with a small if statement.

But isn't a better idea to add the valid cases and invalid cases separate, rather than cloging the code with if conditions (where even the valid cases has to go through the condition, just to fail). This way it is easier, in the future, to add more new cases (as the code changes)

> Also, the test parameters usually take a case number with a descriptor of what the case is testing

Checking on this...

jarrpa: Merging the loops would give me better confidence that the tests are testing the same code. In this specific instance it's not a problem, but I would like to promote the overall pattern.

aruniiird: @jarrpa , I don't see a point in adding a case label just for the sake of it. The point I am trying to make here is the following,

- a. These labels are not used any where else except in the top struct array elments. We could have even used comments instead

- b. We could have easily generated this case label from loop index itself, why add a static text

- c. Case labels would have been much useful if we were adding a description of what test-scenario we are performing with the following parameters.

Nevertheless I have added the labels (and tried to use them) inorder to make the tests compatible with other tests.

aruniiird: > Merging the loops would give me better confidence that the tests are testing the same code. In this specific instance it's not a problem, but I would like to promote the overall pattern.

Sorry @jarrpa , saw this comment after the initial push. Let me know if you still want this change...

jarrpa: >     * c. Case labels would have been much useful if we were adding a description of what test-scenario we are performing with the following parameters.

There's actually something we can do to take advantage of that, but I'm waiting for the feature freeze and SDK overhaul before bringing it in.

> > Merging the loops would give me better confidence that the tests are testing the same code. In this specific instance it's not a problem, but I would like to promote the overall pattern.
> 
> Sorry @jarrpa , saw this comment after the initial push. Let me know if you still want this change...

Yes please. :)

aruniiird: Done...

pkg/controller/storagecluster/kms_resources.go
nbalacha: I would still prefer adding a bool return for cleaner function definition - an Is<do-something> implies a bool value. 
I checked the os package and the only functions I could see which return only an error are those that perform actions - chmod, chdir etc etc. That would hold in this case if the function was called checkKMSProvider or similar.

aruniiird: @nbalacha , your point is fair, as the function name starts with an `is***` we expect it to have a boolean return value.
Changing the function name to `func reachKMSProvider(configmap) error`. Hope this is fine.

aruniiird:
@jarrpa , all the review  comments are met... but `ci-e2e-aws` tests are failing... can you take a look.

obnoxxx:
> @jarrpa , all the review comments are met... but `ci-e2e-aws` tests are failing... can you take a look.

the failing tests are `red-hat-storage-ocs-ci-e2e-aws` - this is currently always failing, not because of this PR.

pkg/controller/storagecluster/kms_resources.go
```diff
@@ -0,0 +1,89 @@
...
+// reachKMSProvider function checks whether the provided address is reachable or not.
+// This function won't validate any other cases and only returns an error if the provided
+// KMS provider address is not reachable. All other validations will be done on rook side.
+func reachKMSProvider(kmsConfigMap *corev1.ConfigMap) error {
+	if kmsConfigMap == nil {
+		return fmt.Errorf("please provide a valid config map")
+	}
+	kmsProviderName, ok := kmsConfigMap.Data[KMSProviderKey]
+	if !ok {
+		return nil
```
nbalacha, line 76: Is this returns nil, the newCephCluster will proceed without checking if the Endpoint is reachable.

aruniiird:
Added KMS Configs to Nooba spec as well.
Noobaa tests need to be a bit polished, will do that in a while.

pkg/controller/storagecluster/kms_resources.go
aruniiird: Yes @nbalacha , as mentioned in the comments this function will only check the reachability if an endpoint is available (and returns error accordingly).
All the other negative cases (like endpoint not given or a different KMS provider is given etc.) are not handled in OCS-Operator. It will be taken care in rook operator.

aruniiird:
@nimrod-becker , added Noobaa related changes, can you please take a look...
@jarrpa , @umangachapagain , @nbalacha , can we have another round of reviews (to cover latest changes).
Thanks

umangachapagain (changes requested):
Changing NooBaa operator version requires more than just bumping it in go.mod. We need to ensure we use the correct images, update the generated files. It can be a separate PR that goes in before this one.

That way we can keep track of version bumps in a simpler way rather than finding it squashed between commits of a bigger feature. And, also isolate failures to a single point.

obnoxxx:
> Changing NooBaa operator version requires more than just bumping it in go.mod. We need to ensure we use the correct images, update the generated files. It can be a separate PR that goes in before this one.
> 
> That way we can keep track of version bumps in a simpler way rather than finding it squashed between commits of a bigger feature. And, also isolate failures to a single point.

Fully agree.
Please refer to previous bump-noobaa-version-PRs for a pattern.

aruniiird:
> Changing NooBaa operator version requires more than just bumping it in go.mod. We need to ensure we use the correct images, update the generated files. It can be a separate PR that goes in before this one.
> 
> That way we can keep track of version bumps in a simpler way rather than finding it squashed between commits of a bigger feature. And, also isolate failures to a single point.

Created a separate [PR#959](https://github.com/openshift/ocs-operator/pull/959) for noobaa operator updation and rebased this on top of that...

pkg/controller/storagecluster/cephcluster_test.go
```diff
@@ -355,3 +359,82 @@ func TestStorageClassDeviceSetCreation(t *testing.T) {
...
+		}
+		// don't start dummy servers for invalid tests
+		if !strings.HasPrefix(kmsArgs.label, "invalid") {
+			startServerAt(kmsArgs.kmsAddress)
+		}
+		// have to initialize the image status,
+		// without which the code will throw a 'nil pointer' exception
+		reconciler.initializeImagesStatus(cr)
+		err := reconciler.ensureCephCluster(cr, reconciler.reqLogger)
+		if strings.HasPrefix(kmsArgs.label, "invalid") {
```
jarrpa, line 421: Don't fork test logic based on the label name. Introduce a boolean like `expectedFailure` and set it to true for the negative tests.

pkg/controller/storagecluster/cephcluster_test.go
```diff
@@ -355,3 +359,82 @@ func TestStorageClassDeviceSetCreation(t *testing.T) {
...
+			assert.Error(t, err, fmt.Sprintf("Failed: %q. Expected an error", kmsArgs.label))
+		} else {
+			assert.NoError(t, err, fmt.Sprintf("Failed: %q. Not expecting any error. Recieved: %v", kmsArgs.label, err))
+		}
+		// following part of the tests are only for valid tsets
+		if !strings.HasPrefix(kmsArgs.label, "invalid") {
```
jarrpa, line 427: You can just remove these three lines and have the valid test path in the same stanza.

pkg/controller/storagecluster/noobaa_system_reconciler_test.go
```diff
@@ -395,3 +395,104 @@ func getReconciler(t *testing.T, objs ...runtime.Object) ReconcileStorageCluster
 		platform: &Platform{},
 	}
 }
+
+type testKMSArgs struct {
+	kmsProvider string
+	kmsAddress  string
+	testLabel   string
+	valid       bool
+}
```
jarrpa, line 404: Don't define a separate struct, just do an inline struct in the test function.

pkg/controller/storagecluster/noobaa_system_reconciler_test.go
```diff
@@ -395,3 +395,104 @@ func getReconciler(t *testing.T, objs ...runtime.Object) ReconcileStorageCluster
 		platform: &Platform{},
 	}
 }
+
+type testKMSArgs struct {
+	kmsProvider string
+	kmsAddress  string
+	testLabel   string
+	valid       bool
```
jarrpa, line 403: Since this is almost always true, I'd suggest changing it to something like `expectedFailure` and only setting it to true in the invalid cases.

pkg/controller/storagecluster/noobaa_system_reconciler_test.go
```diff
@@ -395,3 +395,104 @@ func getReconciler(t *testing.T, objs ...runtime.Object) ReconcileStorageCluster
...
+func assertNoobaaKMSConfig(t *testing.T, kmsArgs *testKMSArgs, cr *v1.StorageCluster) {
+	ctxTodo := context.TODO()
+	noobaaObj := &v1alpha1.NooBaa{}
+	noobaaObj.Name = "noobaa"
+	kmsCM := createDummyKMSConfigMap(kmsArgs.kmsProvider, kmsArgs.kmsAddress)
+
+	reconciler := createFakeInitializationStorageClusterReconciler(t, noobaaObj)
+	if err := reconciler.client.Create(ctxTodo, kmsCM); err != nil {
+		t.Errorf("Unable to create KMS configmap: %w", err)
+		t.FailNow()
```
jarrpa, line 439: Why is this test using this pattern of Errorf and FailNow? Isn't that the entire reason we have assert?

jarrpa (changes requested):
Code looks good, tests need a little cleanup.

Also, do we really want to combine the Noobaa support into this PR? Can that be its own separate PR?

pkg/controller/storagecluster/noobaa_system_reconciler_test.go
aruniiird: `assert.<functions>` will fall through even when the assert fails.

```go
func TestAsserts(t *testing.T) {
	err := fmt.Errorf("This is an error")
	assert.NoErrorf(t, err, "Not expecting any error, but received: %v", err)
	// if assert worked as expected, we should not reach the below statements
	err = nil
	assert.Errorf(t, err, "Expecting an error, but received: %v", err)
}
```

Output
```
$ go test -v -run "TestAsserts*"     
                                                                                                
=== RUN   TestAsserts
tmp_testgo:7:                                                                                               
                Error Trace:    tmp_test.go:7                                                                    
                Error:          Received unexpected error:                                                                                                              This is an error                                                                                        
                Test:           TestAsserts                         
                Messages:       Not expecting any error, but received: This is an error
tmp_test.go:9:                           
                Error Trace:    tmp_test.go:9 
                Error:          An error is expected but got nil.                                                                                       Test:           TestAsserts          
                Messages:       Expecting an error, but received: <nil>                                                                 
--- FAIL: TestAsserts (0.00s)                                                                                                           
FAIL                                                                                                                                    
exit status 1                                                                                                                           
my_assert_tests 0.037s
```

It is ok to provide assert functions  to assess different failures of the `main component` (continue after each failure and check the next functionality). But what happens if creation of `main component` itself fails? No point in assert through it's functionality failures now (as the component itself is not there).
So I had to use `FailNow()` function in places where the main component itself is not created (or failed to create); ie creation of `CephCluster` resource fails, no point in going further down to check the failures in `ensureNoobaaSystem()` functionality or to check KMS config values.

aruniiird:
> Code looks good, tests need a little cleanup.
> 
> Also, do we really want to combine the Noobaa support into this PR? Can that be its own separate PR?

Moved the noobaa kms changes to [PR#961](https://github.com/openshift/ocs-operator/pull/961)

nbalacha (commented):
Looks good to me.

pkg/controller/storagecluster/noobaa_system_reconciler_test.go
aruniiird: Done

pkg/controller/storagecluster/noobaa_system_reconciler_test.go
aruniiird: Done..

pkg/controller/storagecluster/cephcluster_test.go
aruniiird: Done...

pkg/controller/storagecluster/cephcluster_test.go
aruniiird: Done...

obnoxxx:
this is dependent on #959.
Needs to be rebased and merged after that PR.

obnoxxx:
/hold

holding until #959 is merged

jarrpa (approved):
/hold cancel
/lgtm

PR #906, opened 2020-11-23
Storageclass and snapshot management

egafford:
Storageclass and snapshot management

Prior to this change, StorageClass and Snapshot management
strategies were in two monolithic structs under the
ManagedResourcesSpec. This would allow users to fall out of sync
between their StorageClass and Snapshot management policies
and the management policies of their CephFilesystem, CephBlockPool,
and CephObjectStore. This patch enables users to disable StorageClass
and Snapshot management as part of the ReconcileStrategy of each
Ceph storage type.
    
Signed-off-by: Elise Gafford <egafford@redhat.com>


jarrpa:
Overall, the code as-is makes sense, but I had a though (sorry!). What if we went with something like `Spec.ManagedResources.CephBlockPools.StorageClasses.ReconcileStrategy` instead of booleans? This would serve two purposes: 1. It would preserve the `init` behavior (I know no know uses it, but still), and 2. it introduces a section we could use to later specify explicit StorageClass configurations. Thoughts?

egafford:
> Overall, the code as-is makes sense, but I had a though (sorry!). What if we went with something like `Spec.ManagedResources.CephBlockPools.StorageClasses.ReconcileStrategy` instead of booleans? This would serve two purposes: 1. It would preserve the `init` behavior (I know no know uses it, but still), and 2. it introduces a section we could use to later specify explicit StorageClass configurations. Thoughts?

First, I believe that the 'init' behavior is preserved in these commits. The `if reconcileStrategy == ReconcileStrategyInit` test is still in the build, so StorageClasses and SnapshotClasses will not be modified after creation even if disableStorageClass == false. Unless you're talking about a different `init` behavior that's not coming to mind for me.

My concern with your notion is that it introduces some of the same out-of-sync possibilities that the more monolithic ManagedResources.StorageClass struct introduced. It's true that at present, DisableStorageClass == false and ReconcileStrategy == [non-ignore] is pretty meaningless. Still, offering ReconcileStrategy _and_ a StorageClasses.ReconcileStrategy under CephBlockPools (and the others) creates much more of a sense that these configurations are and should be independent. As I understand the semantics of the change, having a more managed StorageClass and a less managed underlying Ceph storage resource is meaningless. I could be wrong in this, though.

Thoughts?

obnoxxx:
> Overall, the code as-is makes sense, but I had a though (sorry!). What if we went with something like `Spec.ManagedResources.CephBlockPools.StorageClasses.ReconcileStrategy` instead of booleans? This would serve two purposes: 1. It would preserve the `init` behavior (I know no know uses it, but still), and 2. it introduces a section we could use to later specify explicit StorageClass configurations. Thoughts?

Having a section for storage class is not a bad thing, but I don't think that adding a ReconcileStrategy for the storage class subordinate to the reconcile strategy for the actual storage is helpful. I agree with @egafford  that this would rather add to the confusion. The logic implemented currently is this:

* If ReconcileStrategy for storage type foo (CephFilesystem, CephBlockPool, CephObjectStore) is not set to ignore, then (and only then) the corresponding DisableStorageClass setting will be looked at. If it s not disabled, the storage class with be treated the same way as the storage itself (manage / init). If it is disabled, then the storage class will be ignored.

I can not imagine cases where, we'll really want to differentiate. As @egafford said, why would one want to have a storage class less managed or more managed than the corresponding underlying storage? I'd say being able to turn the storage class part on and off is sufficient here.

Regarding the possible config section: If we want to introduce it now, we could create a subsection and instead of a ReconcileStrategy have the bool in there:

`Spec.ManagedResources.CephBlockPools.StorageClasses.Disable` (rather than a full reconcile strategy).

@jarrpa , @egafford  - what do you think?

obnoxxx:
@jarrpa, thinking about it and discussing with @egafford a bit more, let me summarize like this:

* A boolean is sufficient for controlling management of storage classes (if enabled, it will follow the strategy of the underlying storage)
* Introducing a subsection for possible future config options for the SCs is definitely interesting.
* However, while it can be considered, the boolean control for SC management does not *need* to be subordinate to this section. It can even be argued that it is cleaner to have it in parallel to the section, so that the section only controls what gets into the SC, not *whether* the SC gets created in the first place...

In the light of the last point, I think we can safely stick with the current Bools without preventing future additions or making them very awkward or unnatural.

egafford:
> @jarrpa, thinking about it and discussing with @egafford a bit more, let me summarize like this:
> 
>     * A boolean is sufficient for controlling management of storage classes (if enabled, it will follow the strategy of the underlying storage)
> 
>     * Introducing a subsection for possible future config options for the SCs is definitely interesting.
> 
>     * However, while it can be considered, the boolean control for SC management does not _need_ to be subordinate to this section. It can even be argued that it is cleaner to have it in parallel to the section, so that the section only controls what gets into the SC, not _whether_ the SC gets created in the first place...

Right, even if we have a `Spec.ManagedResources.PickACephType.StorageClasses` subsection, it seems to me to be equally semantically valid for a field that disables the StorageClass as a whole be a child or a sibling of that subsection. We don't seem to have any defined precedent on this sort of thing elsewhere in the project. Because it's so close to my indifference point, I'd argue for the one that requires less framework. (Also, SnapshotClasses are not terribly configurable as I understand it, but keeping consistency with StorageClasses still makes sense to me, so that's another rather small argument in favor of `DisableStorageClass` instead of `PickACephType.StorageClass.Disable`, though not a compelling argument on its own.)
> 
> In the light of the last point, I think we can safely stick with the current Bools without preventing future additions or making them very awkward or unnatural.

obnoxxx (approved):
This looks good to me. Approving. I will allow some time for @jarrpa to comment, but intend to merge tomorrow if I don't hear a veto. :-)

jarrpa (approved):
/lgtm

jarrpa:
Sounds good. I do dislike inverse booleans (where true means don't do a thing), but I know that for behavior preservation it's necessary.

egafford:
> Sounds good. I do dislike inverse booleans (where true means don't do a thing), but I know that for behavior preservation it's necessary.

Inverse booleans are in fact terrible. I am sort of astonished at Go unmarshalling that this is not an easily solvable problem, from all I've read.

Issue #756, opened 2020-09-08
RFE: Display "storage cluster" resource as a required custom resource for OCS on UI 


afreen23:
**Proposal:**

Openshift console uses `operatorframework.io/initialization-resource` annotation to display a custom resource as required on UI. OCS can leverage this annotation to allow displaying "storage cluster" CR as a required resource.

This will make users to know what to start with after installing the OCS operator from UI, thereby providing an improved UX

**UI mocks after adding the annotation in operator CSV**

http://openshift.github.io/openshift-origin-design/designs/administrator/olm/operand-creation-at-operator-install/

**Annotation Usage Example (Portworx):**
```
{
  "metadata": {
    "annotations": {
      "operatorframework.io/initialization-resource": "{\n    \"apiVersion\": \"core.libopenstorage.org/v1alpha1\",\n    \"kind\": \"StorageCluster\",\n    \"metadata\": {\n      \"name\": \"portworx-initvalue\",\n      \"namespace\": \"test-operator\"\n    },\n    \"spec\": {\n      \"userInterface\": {\n        \"enabled\": true\n      },\n      \"autopilot\": {\n        \"enabled\": false\n      }\n    }\n  }\n"
    }
  }
}
```
_Annotation key:_ `operatorframework.io/initialization-resource` 
_Annotation value_:  a single value of the default yaml of the CR which should have - apiVersion, kind and metadata fields. Other fields can be added as well as per the CRD

cc @obnoxxx

jarrpa:
Is there a Jira story about this? Since this is focused on the product experience, it should be tracked in a place where PM and Engineering management have visibility to it so we can plan for it accordingly.

afreen23:
Yes, it is included in 4.7 planning under "UX research based enhancments [Feature]". 
Surely after planning exit we will have a JIRA created.

bipuladh:
https://github.com/openshift/ocs-operator/pull/907 
Fixes this issue

PR #907, opened 2020-11-23
Add annotation in CSV to make StorageCluster the initialization resource

bipuladh:
- Add annotation in CSV to make StorageCluster the initialization resource
- Used by the OLM UI to redirect user to Storage Cluster Creation Page
- An example of the CR is required for it as it would use the example as the default value


umangachapagain (approved):
Maybe add some info (on PR and Commit Description) on initialization resources and how they're used?

obnoxxx:
> Maybe add some info (on PR and Commit Description) on initialization resources and how they're used?

Yes, please! 😄

tools/csv-merger/csv-merger.go
```diff
@@ -672,6 +672,56 @@ The NooBaa operator deploys and manages the [NooBaa][2] Multi-Cloud Gateway on O
...
+                "spec": {
+                    "accessModes": [
+                        "ReadWriteOnce"
+                    ],
+                    "resources": {
+                        "requests": {
+                            "storage": "10Gi"
+                        }
+                    },
+                    "storageClassName": "gp2"
```
nbalacha, line 696: Do the values of the various fields matter here? gp2 is aws specific. Ditto for the storage size values and storageDeviceSet definition.

bipuladh: No, it does not matter. It's similar to https://github.com/openshift/ocs-operator/pull/907/files#diff-fa6116989ff8c68dd54e4dacc28aa0a80bf370ff31c1a1100cf693a068fde005R746

nbalacha: This link points to this PR itself.
How is this annotation used in the UI and why does it need to be an example instance of the StorageCluster itself?

umangachapagain: UI parses all annotations in CSV and uses it for various purposes. In this case, this particular annotation would instruct the UI to list StorageCluster as a required resource. Meaning when user subscribes to operator they'll be visually suggested to instantiate StorageCluster.

It needs to be an example instance of the StorageCluster itself. Because UI uses different parts of it for different purposes. For one this would be used in old school "Edit YAML" button to provide an example spec. It uses metadata to fetch respective data model and do some validations. This is similar in pattern to using `alm-examples` annotation.

@bipuladh Did I get it wrong?

jarrpa: This would be useful to have in the commit description. :)

obnoxxx (changes requested):
PR description and Commit messages should explain and provide context:
What is the problem we are trying to solve?
How are we addressing it?
What does the PR / Patch do?

(This is a general request, not only for this PR.)

bipuladh:
Thanks @obnoxxx I have updated the commit message and the description to the PR. :)

jarrpa (approved):
/lgtm

jarrpa (changes requested):
There are still CSV changes missing. Please run `make gen-latest-csv` and make sure you commit all files changed.

bipuladh:
> There are still CSV changes missing. Please run `make gen-latest-csv` and make sure you commit all files changed.

Hi, I ran the command on top of the commit and nothing changed. I can see the annotation is added to the generated YAML file.

afreen23:
Closes #756

PR #909, opened 2020-11-23
Track platforms that support external mode from CSV

bipuladh:
- Used by UI to block the radio button for External Mode
- UI fetches the CSV and compares it with the infrastructure it's running on

### DIFF
--- a/tools/csv-merger/csv-merger.go
+++ b/tools/csv-merger/csv-merger.go
@@ -728,6 +728,9 @@ The NooBaa operator deploys and manages the [NooBaa][2] Multi-Cloud Gateway on O
         }
     }
 	`
+	// Used by UI to track platforms that support External Mode
+	// None is reported as the Infrastructure type for some UPI/Baremetal (non-automated) environment
+	ocsCSV.Annotations["external.features.ocs.openshift.io/supported-platforms"] = `["BareMetal", "None", "VSphere", "OpenStack", "oVirt"]`
 	ocsCSV.Annotations["alm-examples"] = `
 [
     {
--- a/deploy/olm-catalog/ocs-operator/manifests/ocs-operator.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/ocs-operator/manifests/ocs-operator.clusterserviceversion.yaml
@@ -971,6 +971,7 @@ metadata:
       bGYuZmFpbCgiQW4gZXhjZXB0aW9uIHdhcyBleHBlY3RlZCIpCiAgICAgICAgICAgIGV4Y2VwdCBF
       eGVjdXRpb25GYWlsdXJlRXhjZXB0aW9uIGFzIGVycjoKICAgICAgICAgICAgICAgIHByaW50KCJF
       eGNlcHRpb24gdGhyb3duIHN1Y2Nlc3NmdWxseToge30iLmZvcm1hdChlcnIpKQo=
+    external.features.ocs.openshift.io/supported-platforms: '["BareMetal", "None", "VSphere", "OpenStack", "oVirt"]'
     external.features.ocs.openshift.io/validation: '{"secrets":["rook-ceph-operator-creds", "rook-csi-rbd-node", "rook-csi-rbd-provisioner"], "configMaps": ["rook-ceph-mon-endpoints", "rook-ceph-mon"], "storageClasses": ["ceph-rbd"], "cephClusters": ["monitoring-endpoint"]}'
     features.ocs.openshift.io/enabled: '["external", "minimal_deployment", "encryption"]'
     olm.skipRange: '>=0.0.1 <4.7.0'


tools/csv-merger/csv-merger.go
```diff
@@ -657,6 +657,8 @@ The NooBaa operator deploys and manages the [NooBaa][2] Multi-Cloud Gateway on O
 	ocsCSV.Annotations["external.features.ocs.openshift.io/validation"] = `{"secrets":["rook-ceph-operator-creds", "rook-csi-rbd-node", "rook-csi-rbd-provisioner"], "configMaps": ["rook-ceph-mon-endpoints", "rook-ceph-mon"], "storageClasses": ["ceph-rbd"], "cephClusters": ["monitoring-endpoint"]}`
 	// Injecting the RHCS exporter script present in Rook CSV
 	ocsCSV.Annotations["external.features.ocs.openshift.io/export-script"] = rookCSV.GetAnnotations()["externalClusterScript"]
+	// Used by UI to track platforms that support External Mode
+	ocsCSV.Annotations["external.features.ocs.openshift.io/supported-platforms"] = `["Baremetal", "None", "VSphere", "OpenStack", "oVirt"]`
```
umangachapagain, line 661: Can we not have "None" as a supported platform?

umangachapagain (changes requested):

tools/csv-merger/csv-merger.go
bipuladh: Why?

obnoxxx (commented):
What is the reason for introducing this?

Also why would we not support external mode on any platform?

tools/csv-merger/csv-merger.go
cloudbehl: UPI/baremetal env generally reports the infra as None(Non-automated) env. So let's not remove it.

umangachapagain: If it's not in this list, can we consider it's "None"?

If not, we need at least a comment on the code explaining what "None" is so that it doesn't look like we support 'none' platform.

bipuladh: If it's not on the list then it could be Cloud Platforms. I guess we can add a comment explaining what `None` is.

bipuladh:
> What is the reason for introducing this?
> 
> Also why would we not support external mode on any platform?

QE raised a BZ a while back stating that we should not enable External Mode on Cloud Platforms as we don't support it there and would confuse the user. 
We first kept this filter in the UI. But that requires changes in the UI when new platforms are supported hence we have moved it to the CSV so that the UI doesn't have to be updated when we choose to support the new platform for external mode.

obnoxxx:
> > What is the reason for introducing this?
> > Also why would we not support external mode on any platform?
> 
> QE raised a BZ a while back stating that we should not enable External Mode on Cloud Platforms as we don't support it there and would confuse the user.

Hmm, could you reference this BZ? I don't recall us making such a restriction in support statement. If the concern is whether ceph is supported in the cloud environments, it should not be our (OCS's) problem: In principle, if the latency is low enough, the ceph cluster could be installed elsewhere where it is supported, and the consuming OCS cluster could still be on a cloud platform...

> We first kept this filter in the UI. But that requires changes in the UI when new platforms are supported hence we have moved it to the CSV so that the UI doesn't have to be updated when we choose to support the new platform for external mode.

I am trying to understand what you are saying: Are you saying that this list is not used by OCS operator, only kept here in the CSV, and it is used by the UI? In which way is it used - does it prevent the deployment when the platform is not in the list?

tools/csv-merger/csv-merger.go
jarrpa: Comment for "None" is still needed. Also, `Baremetal` should be `BareMetal`.

jarrpa:
I agree with Michael that we should not go forward with this, and indeed should remove the code in the console as well. There is nothing in our release notes or other documentation limiting which platforms may have an external cluster: https://access.redhat.com/documentation/en-us/red_hat_openshift_container_storage/4.5/html-single/planning_your_deployment/index#external-mode-requirements_rhocs

/hold

bipuladh:
> Hmm, could you reference this BZ? I don't recall us making such a restriction in support statement. If the concern is whether ceph is supported in the cloud environments, it should not be our (OCS's) problem: In principle, if the latency is low enough, the ceph cluster could be installed elsewhere where it is supported, and the consuming OCS cluster could still be on a cloud platform...
> 
The BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1851013
> I am trying to understand what you are saying: Are you saying that this list is not used by OCS operator, only kept here in the CSV, and it is used by the UI? In which way is it used - does it prevent the deployment when the platform is not in the list?
Yes, the UI will parse this list from CSV compare it with the Infra type, and disable it based on that. Similar to how we use `features.ocs.openshift.io/enabled` to enable features in the UI.

bipuladh:
> I agree with Michael that we should not go forward with this, and indeed should remove the code in the console as well. There is nothing in our release notes or other documentation limiting which platforms may have an external cluster: https://access.redhat.com/documentation/en-us/red_hat_openshift_container_storage/4.5/html-single/planning_your_deployment/index#external-mode-requirements_rhocs
> 
> /hold

Talked with the PM and he agrees that we should disable it for cloud environments.

umangachapagain (approved):

nbalacha:
> > What is the reason for introducing this?
> > Also why would we not support external mode on any platform?
> 
> QE raised a BZ a while back stating that we should not enable External Mode on Cloud Platforms as we don't support it there and would confuse the user.
> We first kept this filter in the UI. But that requires changes in the UI when new platforms are supported hence we have moved it to the CSV so that the UI doesn't have to be updated when we choose to support the new platform for external mode.

Please include this information in the commit msg.

jarrpa:
> > > What is the reason for introducing this?
> > > Also why would we not support external mode on any platform?
> > 
> > 
> > QE raised a BZ a while back stating that we should not enable External Mode on Cloud Platforms as we don't support it there and would confuse the user.
> > We first kept this filter in the UI. But that requires changes in the UI when new platforms are supported hence we have moved it to the CSV so that the UI doesn't have to be updated when we choose to support the new platform for external mode.
> 
> Please include this information in the commit msg.

Yes, please update the first commit description to the following:
```
 Track platforms that support external mode from CSV (csv-merger)

- Used by UI to block the radio button for External Mode
- UI fetches the CSV and compares it with the infrastructure it's running on
- This explicitly excludes cloud platforms, as we do not support external clusters on them yet
```

bipuladh:
@jarrpa updated the commit message as requested. PTAL.

jarrpa (approved):
/lgtm

PR #910, opened 2020-11-23
updates CSV version to v4.7.0

umangachapagain:


obnoxxx (changes requested):
afaict, this is missing the change in `version/versoin.go`.

As usual, I suggest to split the commit to the one introducing the change in non-generated code (`hack/common.sh` and `version/version.go`) and a second patch to commit the result of `make gen-latest-csv`.

obnoxxx:
compare 1785b2bb719a345d615e6b5640e1bd168d2dcb9c and a914676d03b06a8eb5624f7b5b26b5f83de765b0.

umangachapagain:
> afaict, this is missing the change in `version/versoin.go`.
> 
> As usual, I suggest to split the commit to the one introducing the change in non-generated code (`hack/common.sh` and `version/version.go`) and a second patch to commit the result of `make gen-latest-csv`.

Fixed it.

obnoxxx (approved):
Thanks for the update!

/lgtm

PR #911, opened 2020-11-23
fixes typo in verify-generated.sh

umangachapagain:
This typo causes issues when verify-generated.sh is
run and there are outdated changes.


obnoxxx (approved):
/lgtm

PR #918, opened 2020-11-25
skips CRD Validation tests

umangachapagain:
Skipping CRD validation test as we are moving to
newer CRD schema whereas the test uses older schema
and needs some investigation and updates before it
can be used again.

### DIFF
--- a/test/crd_validation_test.go
+++ b/test/crd_validation_test.go
@@ -65,6 +65,7 @@ func validateCustomResources(t *testing.T, root string, crd string, prefix strin
 }
 
 func TestCompleteCRD(t *testing.T) {
+	t.Skip("Skipping CRD Validation")
 	root := "./../deploy/crds"
 	crdStructMap := map[string]interface{}{
 		"ocs.openshift.io_ocsinitializations_crd.yaml": &v1.OCSInitialization{},


obnoxxx (commented):
@umangachapagain, thanks for the patch!

Prepending the test function name with `P` seems somewhat arbitrary to me.
According to https://golang.org/pkg/testing/, the `t.Skip()` method can be used as a more standard and more explicit way of skipping tests.

umangachapagain:
> @umangachapagain, thanks for the patch!
> 
> Prepending the test function name with `P` seems somewhat arbitrary to me.
> According to https://golang.org/pkg/testing/, the `t.Skip()` method can be used as a more standard and more explicit way of skipping tests.

Updated.

iamniting (approved):

test/crd_validation_test.go
```diff
@@ -65,6 +65,7 @@ func validateCustomResources(t *testing.T, root string, crd string, prefix strin
 }
 
 func TestCompleteCRD(t *testing.T) {
+	t.Skip("Skipping CRD Validation")
```
iamniting, line 68: Is not better to write a small msg here why we are skipping, like commit-msg. and TODO also

obnoxxx (approved):
/lgtm

thanks for the update!

PR #919, opened 2020-11-26
fix invalid YAML in generated CRD files

umangachapagain:
" --- " in CRD spec descriptions causes the YAML to be invalid.
We replace it but not "---" which is used as separator.

### DIFF
--- a/tools/csv-merger/csv-merger.go
+++ b/tools/csv-merger/csv-merger.go
@@ -843,7 +843,14 @@ func copyCrds(ocsCSV *csvv1.ClusterServiceVersion) {
 		}
 
 		fmt.Printf("reading CRD file %s\n", inputFile)
-
+		// CRDs that refer to 'ObjectReference' fetches
+		// spec description from K8s in incorrect YAML format.
+		// This can not be prevented and documentation needs to
+		// be updated in Kubernetes. Until then the work around is
+		// to remove all occurences of ' --- ' in description.
+		// "---" is used as YAML separator and should not be replaced.
+		// https://github.com/kubernetes/api/blob/master/core/v1/types.go#L5287-L5301
+		crdBytes = []byte(strings.ReplaceAll(string(crdBytes), " --- ", " "))
 		entries := strings.Split(string(crdBytes), "---")
 		for _, entry := range entries {
 			crd := extv1beta1.CustomResourceDefinition{}


tools/csv-merger/csv-merger.go
```diff
@@ -843,7 +843,14 @@ func copyCrds(ocsCSV *csvv1.ClusterServiceVersion) {
...
 		fmt.Printf("reading CRD file %s\n", inputFile)
-
+		// CRDs that refer to 'ObjectReference' fetches
+		// spec description from K8s in incorrect YAML format.
+		// This can not be prevented and documentation needs to
+		// be updated in Kubernetes. Until then the work around is
+		// to remove all occurences of ' --- ' in description.
+		// "---" is used as YAML separator and should not be replaced.
+		// https://github.com/kubernetes/api/blob/master/core/v1/types.go#L5287-L5301
+		crdBytes = []byte(strings.ReplaceAll(string(crdBytes), " --- ", ""))
```
obnoxxx, line 853: * Can you give an example of a yaml that contains `" --- "`?
* Is replacing `" --- "` with the empty string `""` actually optimal? Wouldn't it be better to replace it with something like `" === "`? Or at least a space `" "`?
* is the yaml really invalid or is our splitting problematic? If the latter, then couldn't we make the split string more specific to match a pattern where the line consists exactly of `"---"`?

umangachapagain: >     * Can you give an example of a yaml that contains `" --- "`?
> 
https://github.com/noobaa/noobaa-operator/blob/5.7/deploy/crds/noobaa.io_noobaas_crd.yaml#L961

>     * Is replacing `" --- "` with the empty string `""` actually optimal? Wouldn't it be better to replace it with something like `" === "`? Or at least a space `" "`?
> 
Didn't use any other characters as it could have some other meaning. "" or " " anything is fine. I don't think it makes a big difference.

>     * is the yaml really invalid or is our splitting problematic? If the latter, then couldn't we make the split string more specific to match a pattern where the line consists exactly of `"---"`?
>
YAML is invalid and so our splitting doesn't work well.

We don't want to match "---" exactly as it's valid and is needed for split to work.
I tried different pattern matching and nothing works. Looking for " --- " was the simplest way that fixed our problems.

What I found interesting is when we perform read on YAML files, it all becomes a single block of string and "\n" does not exist. So can't match for something like `^-{3}\n$` and skip those during replacement.

obnoxxx: > > ```
> > * Can you give an example of a yaml that contains `" --- "`?
> > ```
> 
> https://github.com/noobaa/noobaa-operator/blob/5.7/deploy/crds/noobaa.io_noobaas_crd.yaml#L961

Thanks!

> > ```
> > * Is replacing `" --- "` with the empty string `""` actually optimal? Wouldn't it be better to replace it with something like `" === "`? Or at least a space `" "`?
> > ```
> 
> 
> Didn't use any other characters as it could have some other meaning. "" or " " anything is fine. I don't think it makes a big difference.

Well, with `""` you might concatenate two words, with `" "` you don't. Can we use `" "`? If someone might read it, this would be better.

> > ```
> > * is the yaml really invalid or is our splitting problematic? If the latter, then couldn't we make the split string more specific to match a pattern where the line consists exactly of `"---"`?
> > ```
> 
> YAML is invalid and so our splitting doesn't work well.
> [...]
> What I found interesting is when we perform read on YAML files, it all becomes a single block of string and "\n" does not exist. So can't match for something like `^-{3}\n$` and skip those during replacement.

Ah, that explains! Otherwise this would exactly have been my attempt.
This is actually quite poor, not allowing arbitrary strings in description fields. 🤦

umangachapagain: Fixed it.

obnoxxx (approved):
/lgtm

obnoxxx:
@umangachapagain note that this is in principle still incomplete: splitting on `"---"` will also split on `"-----`" or `"a---b"`...

umangachapagain:
> @umangachapagain note that this is in principle still incomplete: splitting on `"---"` will also split on `"-----`" or `"a---b"`...

Yes. Need to find a better way to split and generate. Till then this unblocks us as I'm pretty sure we won't be adding anything like that.

PR #924, opened 2020-11-30
testing must-gather collection

umangachapagain:


jarrpa (changes requested):
I don't want to do this. We should run the must-gather as a separate test, so this already-burdened test doesn't take even longer than it already does for a successful run.

umangachapagain:
> I don't want to do this. We should run the must-gather as a separate test, so this already-burdened test doesn't take even longer than it already does for a successful run.

I'm not suggesting we do this. This is me debugging CI to figure out why I don't get must-gather for OCS. There's a bug hidden.

PR #925, opened 2020-11-30
Storageclass and snapshotclass refactor

egafford:
In order to create a minimal test halo for the release of version 4.6, minimal change to the handling of StorageClass and SnapshotClass suggested that we identify the Ceph storage type of each StorageClass or SnapshotClass by slice index for the purpose of looking up management options. This approach will often become somewhat brittle over time. To better support future changes, this patch set creates wrapping structs around StorageClasses and SnapshotClasses that bundle them with their deployment options.

### DIFF
--- a/pkg/controller/storagecluster/storageclasses.go
+++ b/pkg/controller/storagecluster/storageclasses.go
@@ -17,20 +17,26 @@ import (
 const (
 	// The following constants are the indices at which StorageClasses are returned from newStorageClasses and in
 	// which they should be passed to createStorageClasses.
-	cephFileSystemIndex  = 0
-	cephBlockPoolIndex   = 1
-	cephObjectStoreIndex = 2
+	cephFileSystemIndex = 0
+	cephBlockPoolIndex  = 1
 )
+
+// StorageClassConfiguration provides configuration options for a StorageClass.
+type StorageClassConfiguration struct {
+	storageClass      *storagev1.StorageClass
+	reconcileStrategy ReconcileStrategy
+	disable           bool
+}
 
 // ensureStorageClasses ensures that StorageClass resources exist in the desired
 // state.
 func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCluster, reqLogger logr.Logger) error {
-	scs, err := r.newStorageClasses(instance)
+	scs, err := r.newStorageClassConfigurations(instance)
 	if err != nil {
 		return err
 	}
 
-	err = r.createStorageClasses(scs, instance, reqLogger)
+	err = r.createStorageClasses(scs, reqLogger)
 	if err != nil {
 		return err
 	}
@@ -38,28 +44,12 @@ func (r *ReconcileStorageCluster) ensureStorageClasses(instance *ocsv1.StorageCl
 	return nil
 }
 
-func (r *ReconcileStorageCluster) createStorageClasses(scs []*storagev1.StorageClass, instance *ocsv1.StorageCluster, reqLogger logr.Logger) error {
-	for index, sc := range scs {
-		// In the case of an external cluster, some scs may be unavailable. In this case we should move on.
-		if sc == nil {
+func (r *ReconcileStorageCluster) createStorageClasses(sccs []StorageClassConfiguration, reqLogger logr.Logger) error {
+	for _, scc := range sccs {
+		if scc.reconcileStrategy == ReconcileStrategyIgnore || scc.disable {
 			continue
 		}
-		reconcileStrategy := ReconcileStrategyIgnore
-		disableStorageClass := false
-		switch index {
-		case cephFileSystemIndex:
-			reconcileStrategy = ReconcileStrategy(instance.Spec.ManagedResources.CephFilesystems.ReconcileStrategy)
-			disableStorageClass = instance.Spec.ManagedResources.CephFilesystems.DisableStorageClass
-		case cephBlockPoolIndex:
-			reconcileStrategy = ReconcileStrategy(instance.Spec.ManagedResources.CephBlockPools.ReconcileStrategy)
-			disableStorageClass = instance.Spec.ManagedResources.CephBlockPools.DisableStorageClass
-		case cephObjectStoreIndex:
-			reconcileStrategy = ReconcileStrategy(instance.Spec.ManagedResources.CephObjectStores.ReconcileStrategy)
-			disableStorageClass = instance.Spec.ManagedResources.CephObjectStores.DisableStorageClass
-		}
-		if reconcileStrategy == ReconcileStrategyIgnore || disableStorageClass {
-			continue
-		}
+		sc := scc.storageClass
 		existing := &storagev1.StorageClass{}
 		err := r.client.Get(context.TODO(), types.NamespacedName{Name: sc.Name, Namespace: sc.Namespace}, existing)
 
@@ -73,7 +63,7 @@ func (r *ReconcileStorageCluster) createStorageClasses(scs []*storagev1.StorageC
 		} else if err != nil {
 			return err
 		} else {
-			if reconcileStrategy == ReconcileStrategyInit {
+			if scc.reconcileStrategy == ReconcileStrategyInit {
 				continue
 			}
 			if existing.DeletionTimestamp != nil {
@@ -98,89 +88,106 @@ func (r *ReconcileStorageCluster) createStorageClasses(scs []*storagev1.StorageC
 	return nil
 }
 
-func newCephFilesystemStorageClass(initData *ocsv1.StorageCluster) *storagev1.StorageClass {
+// newCephFilesystemStorageClassConfiguration generates configuration options for a Ceph Filesystem StorageClass.
+func newCephFilesystemStorageClassConfiguration(initData *ocsv1.StorageCluster) StorageClassConfiguration {
 	persistentVolumeReclaimDelete := corev1.PersistentVolumeReclaimDelete
 	allowVolumeExpansion := true
-	return &storagev1.StorageClass{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: generateNameForCephFilesystemSC(initData),
-			Annotations: map[string]string{
-				"description": "Provides RWO and RWX Filesystem volumes",
+	managementSpec := initData.Spec.ManagedResources.CephFilesystems
+	return StorageClassConfiguration{
+		storageClass: &storagev1.StorageClass{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: generateNameForCephFilesystemSC(initData),
+				Annotations: map[string]string{
+					"description": "Provides RWO and RWX Filesystem volumes",
+				},
+			},
+			Provisioner:   fmt.Sprintf("%s.cephfs.csi.ceph.com", initData.Namespace),
+			ReclaimPolicy: &persistentVolumeReclaimDelete,
+			// AllowVolumeExpansion is set to true to enable expansion of OCS backed Volumes
+			AllowVolumeExpansion: &allowVolumeExpansion,
+			Parameters: map[string]string{
+				"clusterID": initData.Namespace,
+				"fsName":    fmt.Sprintf("%s-cephfilesystem", initData.Name),
+				"csi.storage.k8s.io/provisioner-secret-name":            "rook-csi-cephfs-provisioner",
+				"csi.storage.k8s.io/provisioner-secret-namespace":       initData.Namespace,
+				"csi.storage.k8s.io/node-stage-secret-name":             "rook-csi-cephfs-node",
+				"csi.storage.k8s.io/node-stage-secret-namespace":        initData.Namespace,
+				"csi.storage.k8s.io/controller-expand-secret-name":      "rook-csi-cephfs-provisioner",
+				"csi.storage.k8s.io/controller-expand-secret-namespace": initData.Namespace,
 			},
 		},
-		Provisioner:   fmt.Sprintf("%s.cephfs.csi.ceph.com", initData.Namespace),
-		ReclaimPolicy: &persistentVolumeReclaimDelete,
-		// AllowVolumeExpansion is set to true to enable expansion of OCS backed Volumes
-		AllowVolumeExpansion: &allowVolumeExpansion,
-		Parameters: map[string]string{
-			"clusterID": initData.Namespace,
-			"fsName":    fmt.Sprintf("%s-cephfilesystem", initData.Name),
-			"csi.storage.k8s.io/provisioner-secret-name":            "rook-csi-cephfs-provisioner",
-			"csi.storage.k8s.io/provisioner-secret-namespace":       initData.Namespace,
-			"csi.storage.k8s.io/node-stage-secret-name":             "rook-csi-cephfs-node",
-			"csi.storage.k8s.io/node-stage-secret-namespace":        initData.Namespace,
-			"csi.storage.k8s.io/controller-expand-secret-name":      "rook-csi-cephfs-provisioner",
-			"csi.storage.k8s.io/controller-expand-secret-namespace": initData.Namespace,
-		},
+		reconcileStrategy: ReconcileStrategy(managementSpec.ReconcileStrategy),
+		disable:           managementSpec.DisableStorageClass,
 	}
 }
 
-func newCephBlockPoolStorageClass(initData *ocsv1.StorageCluster) *storagev1.StorageClass {
+// newCephBlockPoolStorageClassConfiguration generates configuration options for a Ceph Block Pool StorageClass.
+func newCephBlockPoolStorageClassConfiguration(initData *ocsv1.StorageCluster) StorageClassConfiguration {
 	persistentVolumeReclaimDelete := corev1.PersistentVolumeReclaimDelete
 	allowVolumeExpansion := true
-	return &storagev1.StorageClass{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: generateNameForCephBlockPoolSC(initData),
-			Annotations: map[string]string{
-				"description": "Provides RWO Filesystem volumes, and RWO and RWX Block volumes",
+	managementSpec := initData.Spec.ManagedResources.CephBlockPools
+	return StorageClassConfiguration{
+		storageClass: &storagev1.StorageClass{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: generateNameForCephBlockPoolSC(initData),
+				Annotations: map[string]string{
+					"description": "Provides RWO Filesystem volumes, and RWO and RWX Block volumes",
+				},
+			},
+			Provisioner:   fmt.Sprintf("%s.rbd.csi.ceph.com", initData.Namespace),
+			ReclaimPolicy: &persistentVolumeReclaimDelete,
+			// AllowVolumeExpansion is set to true to enable expansion of OCS backed Volumes
+			AllowVolumeExpansion: &allowVolumeExpansion,
+			Parameters: map[string]string{
+				"clusterID":                 initData.Namespace,
+				"pool":                      generateNameForCephBlockPool(initData),
+				"imageFeatures":             "layering",
+				"csi.storage.k8s.io/fstype": "ext4",
+				"imageFormat":               "2",
+				"csi.storage.k8s.io/provisioner-secret-name":            "rook-csi-rbd-provisioner",
+				"csi.storage.k8s.io/provisioner-secret-namespace":       initData.Namespace,
+				"csi.storage.k8s.io/node-stage-secret-name":             "rook-csi-rbd-node",
+				"csi.storage.k8s.io/node-stage-secret-namespace":        initData.Namespace,
+				"csi.storage.k8s.io/controller-expand-secret-name":      "rook-csi-rbd-provisioner",
+				"csi.storage.k8s.io/controller-expand-secret-namespace": initData.Namespace,
 			},
 		},
-		Provisioner:   fmt.Sprintf("%s.rbd.csi.ceph.com", initData.Namespace),
-		ReclaimPolicy: &persistentVolumeReclaimDelete,
-		// AllowVolumeExpansion is set to true to enable expansion of OCS backed Volumes
-		AllowVolumeExpansion: &allowVolumeExpansion,
-		Parameters: map[string]string{
-			"clusterID":                 initData.Namespace,
-			"pool":                      generateNameForCephBlockPool(initData),
-			"imageFeatures":             "layering",
-			"csi.storage.k8s.io/fstype": "ext4",
-			"imageFormat":               "2",
-			"csi.storage.k8s.io/provisioner-secret-name":            "rook-csi-rbd-provisioner",
-			"csi.storage.k8s.io/provisioner-secret-namespace":       initData.Namespace,
-			"csi.storage.k8s.io/node-stage-secret-name":             "rook-csi-rbd-node",
-			"csi.storage.k8s.io/node-stage-secret-namespace":        initData.Namespace,
-			"csi.storage.k8s.io/controller-expand-secret-name":      "rook-csi-rbd-provisioner",
-			"csi.storage.k8s.io/controller-expand-secret-namespace": initData.Namespace,
-		},
+		reconcileStrategy: ReconcileStrategy(managementSpec.ReconcileStrategy),
+		disable:           managementSpec.DisableStorageClass,
 	}
 }
 
-func newCephOBCStorageClass(initData *ocsv1.StorageCluster) *storagev1.StorageClass {
+// newCephOBCStorageClassConfiguration generates configuration options for a Ceph Object Store StorageClass.
+func newCephOBCStorageClassConfiguration(initData *ocsv1.StorageCluster) StorageClassConfiguration {
 	reclaimPolicy := corev1.PersistentVolumeReclaimDelete
-	retSC := &storagev1.StorageClass{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: generateNameForCephRgwSC(initData),
-			Annotations: map[string]string{
-				"description": "Provides Object Bucket Claims (OBCs)",
+	managementSpec := initData.Spec.ManagedResources.CephObjectStores
+	return StorageClassConfiguration{
+		storageClass: &storagev1.StorageClass{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: generateNameForCephRgwSC(initData),
+				Annotations: map[string]string{
+					"description": "Provides Object Bucket Claims (OBCs)",
+				},
+			},
+			Provisioner:   fmt.Sprintf("%s.ceph.rook.io/bucket", initData.Namespace),
+			ReclaimPolicy: &reclaimPolicy,
+			Parameters: map[string]string{
+				"objectStoreNamespace": initData.Namespace,
+				"region":               "us-east-1",
+				"objectStoreName":      generateNameForCephObjectStore(initData),
 			},
 		},
-		Provisioner:   fmt.Sprintf("%s.ceph.rook.io/bucket", initData.Namespace),
-		ReclaimPolicy: &reclaimPolicy,
-		Parameters: map[string]string{
-			"objectStoreNamespace": initData.Namespace,
-			"region":               "us-east-1",
-			"objectStoreName":      generateNameForCephObjectStore(initData),
-		},
+		reconcileStrategy: ReconcileStrategy(managementSpec.ReconcileStrategy),
+		disable:           managementSpec.DisableStorageClass,
 	}
-	return retSC
 }
 
-// newStorageClasses returns the StorageClass instances that should be created
+// newStorageClassConfigurations returns the StorageClassConfiguration instances that should be created
 // on first run.
-func (r *ReconcileStorageCluster) newStorageClasses(initData *ocsv1.StorageCluster) ([]*storagev1.StorageClass, error) {
-	ret := []*storagev1.StorageClass{
-		newCephFilesystemStorageClass(initData),
-		newCephBlockPoolStorageClass(initData),
+func (r *ReconcileStorageCluster) newStorageClassConfigurations(initData *ocsv1.StorageCluster) ([]StorageClassConfiguration, error) {
+	ret := []StorageClassConfiguration{
+		newCephFilesystemStorageClassConfiguration(initData),
+		newCephBlockPoolStorageClassConfiguration(initData),
 	}
 	// OBC storageclass will be returned only in TWO conditions,
 	// a. either 'externalStorage' is enabled
@@ -188,7 +195,7 @@ func (r *ReconcileStorageCluster) newStorageClasses(initData *ocsv1.StorageClust
 	// b. current platform is not a cloud-based platform
 	platform, err := r.platform.GetPlatform(r.client)
 	if initData.Spec.ExternalStorage.Enable || err == nil && !avoidObjectStore(platform) {
-		ret = append(ret, newCephOBCStorageClass(initData))
+		ret = append(ret, newCephOBCStorageClassConfiguration(initData))
 	}
 	return ret, nil
 }
--- a/pkg/controller/storagecluster/storageclasses.go
+++ b/pkg/controller/storagecluster/storageclasses.go
@@ -98,7 +98,64 @@ func (r *ReconcileStorageCluster) createStorageClasses(scs []*storagev1.StorageC
 	return nil
 }
 
-func (r *ReconcileStorageCluster) newOBCStorageClass(initData *ocsv1.StorageCluster) *storagev1.StorageClass {
+func newCephFilesystemStorageClass(initData *ocsv1.StorageCluster) *storagev1.StorageClass {
+	persistentVolumeReclaimDelete := corev1.PersistentVolumeReclaimDelete
+	allowVolumeExpansion := true
+	return &storagev1.StorageClass{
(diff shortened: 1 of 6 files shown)


pkg/controller/storagecluster/storageclasses.go
```diff
@@ -98,6 +98,57 @@ func (r *ReconcileStorageCluster) createStorageClasses(scs []*storagev1.StorageC
 	return nil
 }
 
+func (r *ReconcileStorageCluster) newFilesystemStorageClass(initData *ocsv1.StorageCluster) *storagev1.StorageClass {
```
jarrpa:
```suggestion
func (r *ReconcileStorageCluster) newCephFilesystemStorageClass(initData *ocsv1.StorageCluster) *storagev1.StorageClass {
```

pkg/controller/storagecluster/storageclasses.go
```diff
@@ -98,6 +98,57 @@ func (r *ReconcileStorageCluster) createStorageClasses(scs []*storagev1.StorageC
...
+			"csi.storage.k8s.io/provisioner-secret-namespace":       initData.Namespace,
+			"csi.storage.k8s.io/node-stage-secret-name":             "rook-csi-cephfs-node",
+			"csi.storage.k8s.io/node-stage-secret-namespace":        initData.Namespace,
+			"csi.storage.k8s.io/controller-expand-secret-name":      "rook-csi-cephfs-provisioner",
+			"csi.storage.k8s.io/controller-expand-secret-namespace": initData.Namespace,
+		},
+	}
+}
+
+func (r *ReconcileStorageCluster) newBlockPoolStorageClass(initData *ocsv1.StorageCluster) *storagev1.StorageClass{
```
jarrpa:
```suggestion
func (r *ReconcileStorageCluster) newCephBlockPoolStorageClass(initData *ocsv1.StorageCluster) *storagev1.StorageClass{
```

jarrpa (changes requested):
Oh I absolutely love this, thank you! I've been meaning to get rid of that indexing...

Overall looks sound. Two naming nits.

jarrpa:
Make sure to run `make ocs-operator-ci` before pushing.

pkg/controller/storagecluster/storageclasses.go
```diff
@@ -14,13 +14,11 @@ import (
...
 )
 
-const(
-	// The following constants are the indices at which StorageClasses are returned from newStorageClasses and in
-	// which they should be passed to createStorageClasses.
-	cephFileSystemIndex  = 0
-	cephBlockPoolIndex   = 1
-	cephObjectStoreIndex = 2
-)
+type StorageClassConfiguration struct{
```
umangachapagain, line 17: Exported type requires a comment or the linter fails

pkg/controller/storagecluster/volumesnapshotterclasses.go
```diff
@@ -28,6 +28,12 @@ const (
 	snapshotterSecretNamespace = "csi.storage.k8s.io/snapshotter-secret-namespace"
 )
 
+type SnapshotClassConfiguration struct{
```
umangachapagain, line 31: Exported type requires a comment or the linter fails

pkg/controller/storagecluster/storageclasses.go
umangachapagain, line 17:
```suggestion
// StorageClassConfiguration provides configuration options for a Storage Class
type StorageClassConfiguration struct{
```

pkg/controller/storagecluster/volumesnapshotterclasses.go
umangachapagain:
```suggestion
// SnapshotClassConfiguration provides configuration options for a Snapshot Class
type SnapshotClassConfiguration struct{
```

pkg/controller/storagecluster/storageclasses.go
```diff
@@ -98,30 +80,12 @@ func (r *ReconcileStorageCluster) createStorageClasses(scs []*storagev1.StorageC
...
-			"objectStoreName":      generateNameForCephObjectStore(initData),
-		},
-	}
-	return retSC
-}
-
-// newStorageClasses returns the StorageClass instances that should be created
-// on first run.
-func (r *ReconcileStorageCluster) newStorageClasses(initData *ocsv1.StorageCluster) ([]*storagev1.StorageClass, error) {
+func (r *ReconcileStorageCluster) newFilesystemStorageClass(initData *ocsv1.StorageCluster) *StorageClassConfiguration {
```
umangachapagain, line 83:
```suggestion
func (r *ReconcileStorageCluster) newFilesystemStorageClass(initData *ocsv1.StorageCluster) StorageClassConfiguration {
```

pkg/controller/storagecluster/storageclasses.go
```diff
@@ -140,7 +104,17 @@ func (r *ReconcileStorageCluster) newStorageClasses(initData *ocsv1.StorageClust
 				"csi.storage.k8s.io/controller-expand-secret-namespace": initData.Namespace,
 			},
 		},
-		&storagev1.StorageClass{
+		reconcileStrategy: ReconcileStrategy(managementSpec.ReconcileStrategy),
+		disable: managementSpec.DisableStorageClass,
+	}
+}
+
+func (r *ReconcileStorageCluster) newBlockPoolStorageClass(initData *ocsv1.StorageCluster) *StorageClassConfiguration{
```
umangachapagain, line 112:
```suggestion
func (r *ReconcileStorageCluster) newBlockPoolStorageClass(initData *ocsv1.StorageCluster) StorageClassConfiguration{
```

pkg/controller/storagecluster/storageclasses.go
```diff
@@ -162,6 +136,38 @@ func (r *ReconcileStorageCluster) newStorageClasses(initData *ocsv1.StorageClust
 				"csi.storage.k8s.io/controller-expand-secret-namespace": initData.Namespace,
 			},
 		},
+		reconcileStrategy: ReconcileStrategy(managementSpec.ReconcileStrategy),
+		disable: managementSpec.DisableStorageClass,
+	}
+}
+
+func (r *ReconcileStorageCluster) newOBCStorageClass(initData *ocsv1.StorageCluster) *StorageClassConfiguration {
```
umangachapagain, line 144:
```suggestion
func (r *ReconcileStorageCluster) newOBCStorageClass(initData *ocsv1.StorageCluster) StorageClassConfiguration {
```

pkg/controller/storagecluster/storageclasses.go
```diff
@@ -162,6 +136,38 @@ func (r *ReconcileStorageCluster) newStorageClasses(initData *ocsv1.StorageClust
...
+			},
+		},
+		reconcileStrategy: ReconcileStrategy(managementSpec.ReconcileStrategy),
+		disable: managementSpec.DisableStorageClass,
+	}
+}
+
+// newStorageClasses returns the StorageClass instances that should be created
+// on first run.
+func (r *ReconcileStorageCluster) newStorageClasses(initData *ocsv1.StorageCluster) ([]*StorageClassConfiguration, error) {
```
umangachapagain, line 167:
```suggestion
func (r *ReconcileStorageCluster) newStorageClasses(initData *ocsv1.StorageCluster) ([]StorageClassConfiguration, error) {
```

pkg/controller/storagecluster/volumesnapshotterclasses.go
```diff
@@ -46,30 +52,38 @@ func newVolumeSnapshotClass(instance *ocsv1.StorageCluster, snapShotterType Snap
 	return retSC
 }
 
-func newSnapshotClasses(instance *ocsv1.StorageCluster) []*snapapi.VolumeSnapshotClass {
-	scs := []*snapapi.VolumeSnapshotClass{
-		newVolumeSnapshotClass(instance, cephfsSnapshotter),
-		newVolumeSnapshotClass(instance, rbdSnapshotter),
+func newCephFilesystemSnapshotClass(instance *ocsv1.StorageCluster) *SnapshotClassConfiguration {
```
umangachapagain, line 55:
```suggestion
func newCephFilesystemSnapshotClass(instance *ocsv1.StorageCluster) SnapshotClassConfiguration {
```

pkg/controller/storagecluster/volumesnapshotterclasses.go
```diff
@@ -46,30 +52,38 @@ func newVolumeSnapshotClass(instance *ocsv1.StorageCluster, snapShotterType Snap
...
-		switch index {
-		case cephFileSystemIndex:
-			reconcileStrategy = ReconcileStrategy(instance.Spec.ManagedResources.CephFilesystems.ReconcileStrategy)
-			disableSnapshotClass = instance.Spec.ManagedResources.CephFilesystems.DisableSnapshotClass
-		case cephBlockPoolIndex:
-			reconcileStrategy = ReconcileStrategy(instance.Spec.ManagedResources.CephBlockPools.ReconcileStrategy)
-			disableSnapshotClass = instance.Spec.ManagedResources.CephBlockPools.DisableSnapshotClass
-		}
-		if reconcileStrategy == ReconcileStrategyIgnore || disableSnapshotClass {
+func newCephBlockPoolSnapshotClass(instance *ocsv1.StorageCluster) *SnapshotClassConfiguration {
```
umangachapagain, line 63:
```suggestion
func newCephBlockPoolSnapshotClass(instance *ocsv1.StorageCluster) SnapshotClassConfiguration {
```

pkg/controller/storagecluster/volumesnapshotterclasses.go
```diff
@@ -46,30 +52,38 @@ func newVolumeSnapshotClass(instance *ocsv1.StorageCluster, snapShotterType Snap
...
+	return &SnapshotClassConfiguration{
+		snapshotClass:     newVolumeSnapshotClass(instance, rbdSnapshotter),
+		reconcileStrategy: ReconcileStrategy(instance.Spec.ManagedResources.CephFilesystems.ReconcileStrategy),
+		disable:           instance.Spec.ManagedResources.CephFilesystems.DisableSnapshotClass,
+	}
+}
+
+func newSnapshotClasses(instance *ocsv1.StorageCluster) []*SnapshotClassConfiguration {
```
umangachapagain, line 71:
```suggestion
func newSnapshotClasses(instance *ocsv1.StorageCluster) []SnapshotClassConfiguration {
```

umangachapagain (commented):
While I'm happy to see indices go, I'd be happier if we could get rid of Builder functions returning Pointers to Objects (where/if it makes sense).

I believe this would allow callers to operate on their own copies of objects without any risk of it being tampered by some other caller. I'm not sure if sharing objects built by builder does any good. WDYT?

egafford:
> While I'm happy to see indices go, I'd be happier if we could get rid of Builder functions returning Pointers to Objects (where/if it makes sense).
> 
> I believe this would allow callers to operate on their own copies of objects without any risk of it being tampered by some other caller. I'm not sure if sharing objects built by builder does any good. WDYT?

Thanks @umangachapagain. I'll certainly see to the linting changes; much appreciated. As to the question of passing pointers vs. raw structs, I'm not certain I see the risk case that you're describing. Our prior build passed pointers to StorageClasses (and indeed, the StorageClassConfiguration struct now contains such a pointer). I believe that any given caller to (for instance) newCephFilesystemStorageClass will receive a pointer to a new instantiation of the StorageClassConfiguration struct. From there, yes, any one caller can change its own instantiated object through the pointer, but this may be desirable and would also be true if it modified a raw struct and passed it on to another function in its own logical flow.

Am I missing something fundamental in Golang? Do the current functions as written effectively create a flyweight single object and return pointers to that same object to all callers? I am certain that if we created a StorageClassConfiguration as a package variable and returned a pointer to that, side effects between callers could cause problems, but I think the present logic gives a pointer to a new object to each caller (which can then cause its own side effects in its own logical flow, but that might actually be useful).

Or am I missing an error case that I'm failing to see? Looking around the codebase we tend to pass pointers rather than raw structs, but if there's a risk that I'm missing I'm happy to make the changes you suggest.

jarrpa:
> > While I'm happy to see indices go, I'd be happier if we could get rid of Builder functions returning Pointers to Objects (where/if it makes sense).
> > I believe this would allow callers to operate on their own copies of objects without any risk of it being tampered by some other caller. I'm not sure if sharing objects built by builder does any good. WDYT?
> 
> Thanks @umangachapagain. I'll certainly see to the linting changes; much appreciated. As to the question of passing pointers vs. raw structs, I'm not certain I see the risk case that you're describing. Our prior build passed pointers to StorageClasses (and indeed, the StorageClassConfiguration struct now contains such a pointer). I believe that any given caller to (for instance) newCephFilesystemStorageClass will receive a pointer to a new instantiation of the StorageClassConfiguration struct. From there, yes, any one caller can change its own instantiated object through the pointer, but this may be desirable and would also be true if it modified a raw struct and passed it on to another function in its own logical flow.
> 
> Am I missing something fundamental in Golang? Do the current functions as written effectively create a flyweight single object and return pointers to that same object to all callers? I am certain that if we created a StorageClassConfiguration as a package variable and returned a pointer to that, side effects between callers could cause problems, but I think the present logic gives a pointer to a new object to each caller (which can then cause its own side effects in its own logical flow, but that might actually be useful).
> 
> Or am I missing an error case that I'm failing to see? Looking around the codebase we tend to pass pointers rather than raw structs, but if there's a risk that I'm missing I'm happy to make the changes you suggest.

I think you're right that it's a non-issue, at best it would be one of aesthetics. Unless I'm also missing something.

umangachapagain:
> > > While I'm happy to see indices go, I'd be happier if we could get rid of Builder functions returning Pointers to Objects (where/if it makes sense).
> > > I believe this would allow callers to operate on their own copies of objects without any risk of it being tampered by some other caller. I'm not sure if sharing objects built by builder does any good. WDYT?
> > 
> > 
> > Thanks @umangachapagain. I'll certainly see to the linting changes; much appreciated. As to the question of passing pointers vs. raw structs, I'm not certain I see the risk case that you're describing. Our prior build passed pointers to StorageClasses (and indeed, the StorageClassConfiguration struct now contains such a pointer). I believe that any given caller to (for instance) newCephFilesystemStorageClass will receive a pointer to a new instantiation of the StorageClassConfiguration struct. From there, yes, any one caller can change its own instantiated object through the pointer, but this may be desirable and would also be true if it modified a raw struct and passed it on to another function in its own logical flow.
> > Am I missing something fundamental in Golang? Do the current functions as written effectively create a flyweight single object and return pointers to that same object to all callers? I am certain that if we created a StorageClassConfiguration as a package variable and returned a pointer to that, side effects between callers could cause problems, but I think the present logic gives a pointer to a new object to each caller (which can then cause its own side effects in its own logical flow, but that might actually be useful).
> > Or am I missing an error case that I'm failing to see? Looking around the codebase we tend to pass pointers rather than raw structs, but if there's a risk that I'm missing I'm happy to make the changes you suggest.
> 
> I think you're right that it's a non-issue, at best it would be one of aesthetics. Unless I'm also missing something.

I'm not a Go expert in any way, but my concern was (i) since we pass pointers up to the caller, there's a risk of this memory address being rewritten or cleared so when caller looks at this address it might not get what it is expecting. (Do we check that the object at the address we're trying to read from is actually the object we expected) Maybe this is not a real thing in which case there's no issue here. (ii)  Since we're not trying to share this object and mutate it among the callers, can't we get away with passing a copy to all the callers instead of passing reference to the copies? This seems like a lot of memory allocation at a place where it's not required? (iii) Like you said, each caller gets its own instantiated object and is free to make changes if it wishes to. But, what value does the caller get from having a pointer to this object? If caller or any other method wish to mutate this object, it can just take a reference of this object and not the object at the bottom of the stack.

Maybe I'm completely wrong here hence I made a "Comment' and not "Request Changes". I'm not sure if I'd qualify it as aesthetic change yet. And, yes we've been doing this across the code base but that shouldn't be the reason to continue doing it.

egafford:
> > > > While I'm happy to see indices go, I'd be happier if we could get rid of Builder functions returning Pointers to Objects (where/if it makes sense).
> > > > I believe this would allow callers to operate on their own copies of objects without any risk of it being tampered by some other caller. I'm not sure if sharing objects built by builder does any good. WDYT?
> > > 
> > > 
> > > Thanks @umangachapagain. I'll certainly see to the linting changes; much appreciated. As to the question of passing pointers vs. raw structs, I'm not certain I see the risk case that you're describing. Our prior build passed pointers to StorageClasses (and indeed, the StorageClassConfiguration struct now contains such a pointer). I believe that any given caller to (for instance) newCephFilesystemStorageClass will receive a pointer to a new instantiation of the StorageClassConfiguration struct. From there, yes, any one caller can change its own instantiated object through the pointer, but this may be desirable and would also be true if it modified a raw struct and passed it on to another function in its own logical flow.
> > > Am I missing something fundamental in Golang? Do the current functions as written effectively create a flyweight single object and return pointers to that same object to all callers? I am certain that if we created a StorageClassConfiguration as a package variable and returned a pointer to that, side effects between callers could cause problems, but I think the present logic gives a pointer to a new object to each caller (which can then cause its own side effects in its own logical flow, but that might actually be useful).
> > > Or am I missing an error case that I'm failing to see? Looking around the codebase we tend to pass pointers rather than raw structs, but if there's a risk that I'm missing I'm happy to make the changes you suggest.
> > 
> > 
> > I think you're right that it's a non-issue, at best it would be one of aesthetics. Unless I'm also missing something.
> 
> I'm not a Go expert in any way, but my concern was (i) since we pass pointers up to the caller, there's a risk of this memory address being rewritten or cleared so when caller looks at this address it might not get what it is expecting. (Do we check that the object at the address we're trying to read from is actually the object we expected?) Maybe this is not a real thing in which case there's no issue here. (ii) Since we're not trying to share this object and mutate it among the callers, can't we get away with passing a copy to all the callers instead of passing reference to the copies? This seems like a lot of memory allocation at a place where it's not required? (iii) Like you said, each caller gets its own instantiated object and is free to make changes if it wishes to. But, what value does the caller get from having a pointer to this object? If caller or any other method wish to mutate this object, it can just take a reference of this object and not the object at the bottom of the stack?

So my understanding is that of one of the fundamental design goals of Golang was to use pointers (and thus make pass-by-reference and pass-by-value explicit), but to disallow pointer math entirely. As pointer math is hard to debug, easy to get wrong, and more or less never strictly needed, this idea seems pretty great. So as I understand it, Golang protects us from the danger of rewriting the address.

However, you have two points here. The second is more about efficiency and the Golang idiom. In the Golang runtime, is it more efficient to pass around a (smaller) memory address on the stack but need to dereference to get to the object, or is it more efficient to pass the (larger) raw struct itself? And perhaps more importantly, what is idiomatic in Golang and will make it easier for future developers to understand naturally? This question I'm truly not sure about, as I've worked in Golang less than others on the team, and it's a good one to ask.

@jarrpa, any thoughts on idiomatic Golang and preference for passing pointers or raw structs?

> Maybe I'm completely wrong here hence I made a "Comment' and not "Request Changes". I'm not sure if I'd qualify it as aesthetic change yet. And, yes we've been doing this across the code base but that shouldn't be the reason to continue doing it.

pkg/controller/storagecluster/storageclasses.go
umangachapagain, line 83:
```suggestion
func newFilesystemStorageClass(initData *ocsv1.StorageCluster) *StorageClassConfiguration {
```

pkg/controller/storagecluster/storageclasses.go
umangachapagain, line 112:
```suggestion
func newBlockPoolStorageClass(initData *ocsv1.StorageCluster) *StorageClassConfiguration{
```

pkg/controller/storagecluster/storageclasses.go
umangachapagain, line 144:
```suggestion
func newOBCStorageClass(initData *ocsv1.StorageCluster) *StorageClassConfiguration {
```

pkg/controller/storagecluster/storageclasses.go
```diff
@@ -162,6 +136,38 @@ func (r *ReconcileStorageCluster) newStorageClasses(initData *ocsv1.StorageClust
...
+	ret := []*StorageClassConfiguration{
+		r.newFilesystemStorageClass(initData),
+		r.newBlockPoolStorageClass(initData),
```
umangachapagain, line 170: These calls don't need a `ReconcileStorageCluster` receiver as they neither access it nor mutate it.

pkg/controller/storagecluster/volumesnapshotterclasses.go
umangachapagain, line 71: This and all other builder functions(i.e new<--->() ) should be renamed to reflect the fact that it now returns `SnapshotClassConfiguration` which consists of `SnapshotClass` and other properties.

umangachapagain (changes requested):

pkg/controller/storagecluster/volumesnapshotterclasses.go
obnoxxx: Yeah, this linter requirement is just *so silly* because it encourages people to add meaningless and redundant comments ... :roll_eyes:

jarrpa: Some comments may end up being redundant because the type/function name is sufficiently descriptive, but I find that having the linter error forces a change that may guide my brain to examine whether I really understand the struct or if the name could be improved.

jarrpa:
> However, you have two points here. The second is more about efficiency and the Golang idiom. In the Golang runtime, is it more efficient to pass around a (smaller) memory address on the stack but need to dereference to get to the object, or is it more efficient to pass the (larger) raw struct itself? And perhaps more importantly, what is idiomatic in Golang and will make it easier for future developers to understand naturally? This question I'm truly not sure about, as I've worked in Golang less than others on the team, and it's a good one to ask.
> 
> @jarrpa, any thoughts on idiomatic Golang and preference for passing pointers or raw structs?

I did a bit of looking around and I found something that may mean Umanga's suggestion is more idiomatic:

>Passing by value often is cheaper
>
> Even though Go looks a bit like C, its compiler works differently. And C analogy does not always work with Go. Passing by value in Go may be significantly cheaper than passing by pointer. This happens because Go uses escape analysis to determine if variable can be safely allocated on function’s stack frame, which could be much cheaper then allocating variable on the heap. Passing by value simplifies escape analysis in Go and gives variable a better chance to be allocated on the stack.

https://goinbigdata.com/golang-pass-by-pointer-vs-pass-by-value/

So, I'll say we should go ahead with his suggestion.

egafford:
> > However, you have two points here. The second is more about efficiency and the Golang idiom. In the Golang runtime, is it more efficient to pass around a (smaller) memory address on the stack but need to dereference to get to the object, or is it more efficient to pass the (larger) raw struct itself? And perhaps more importantly, what is idiomatic in Golang and will make it easier for future developers to understand naturally? This question I'm truly not sure about, as I've worked in Golang less than others on the team, and it's a good one to ask.
> > @jarrpa, any thoughts on idiomatic Golang and preference for passing pointers or raw structs?
> 
> I did a bit of looking around and I found something that may mean Umanga's suggestion is more idiomatic:
> 
> > Passing by value often is cheaper
> > Even though Go looks a bit like C, its compiler works differently. And C analogy does not always work with Go. Passing by value in Go may be significantly cheaper than passing by pointer. This happens because Go uses escape analysis to determine if variable can be safely allocated on function’s stack frame, which could be much cheaper then allocating variable on the heap. Passing by value simplifies escape analysis in Go and gives variable a better chance to be allocated on the stack.
> 
> https://goinbigdata.com/golang-pass-by-pointer-vs-pass-by-value/
> 
> So, I'll say we should go ahead with his suggestion.

Cool; will do. Thanks y'all.

umangachapagain:
> > 
> > I'm not a Go expert in any way, but my concern was (i) since we pass pointers up to the caller, there's a risk of this memory address being rewritten or cleared so when caller looks at this address it might not get what it is expecting. (Do we check that the object at the address we're trying to read from is actually the object we expected?) Maybe this is not a real thing in which case there's no issue here. (ii) Since we're not trying to share this object and mutate it among the callers, can't we get away with passing a copy to all the callers instead of passing reference to the copies? This seems like a lot of memory allocation at a place where it's not required? (iii) Like you said, each caller gets its own instantiated object and is free to make changes if it wishes to. But, what value does the caller get from having a pointer to this object? If caller or any other method wish to mutate this object, it can just take a reference of this object and not the object at the bottom of the stack?
> 
> So my understanding is that of one of the fundamental design goals of Golang was to use pointers (and thus make pass-by-reference and pass-by-value explicit), but to disallow pointer math entirely. As pointer math is hard to debug, easy to get wrong, and more or less never strictly needed, this idea seems pretty great. So as I understand it, Golang protects us from the danger of rewriting the address.
> 
Yes, you're correct. Golang protects us from this issue. There is no data corruption or integrity issue here like I thought. But, this came at the cost of heap allocations (Go's escape analysis did this) which also means more work for Garbage Collector? 

This became clear when I ran a basic benchmarking test. While returning pointers `newStorageClasses()` was doing `24 allocation/operation` and cycles were spent to put object into heap, retrieve it back and GC etc. Without pointers we were at `22 allocation/operation`.

https://github.com/openshift/ocs-operator/pull/925#issuecomment-737356925 validates the finding.

jarrpa:
The ocs-operator-ci failure is a flake. But will wait for the rebase to trigger it again.

egafford:
> The ocs-operator-ci failure is a flake. But will wait for the rebase to trigger it again.

Groovy; seemed that way from my local results but I wanted to investigate a bit more before assuming anything to _not_ be my fault. :)

I'll go ahead and rebase then. Thanks.

obnoxxx (commented):
Great start, thanks.

Looking at these pieces of code in the last couple of days, I started to have a slightly different idea for a refactoring.

Currently it is like this:
```
ensureStorageClasses() {
  scs = newStorageClasses()

  createStorageClasses(scs)
}
```

One of the problem is that we do this as a list in the first place.
Secondly, a problem is that some of the filtering happens in `newStorageClasses` and some in `createStorageClasses`. 

So I was thinking of doing it roughly like this:

```
func ensureStorageClasses() {
  ensureRbdStorageClass()
  ensureCephFsStorageClass()
  ensureOBCStorageClas()
}

func ensureRbdStorageClass() {
  if avoidRbdStorageClass() {
    return
  }
  sc := newRbdStorageClass()
  createStorageClass(sc)
}

func ensureCephfsStorageClass() {
  if avoidCephFsStorageClass() {
     return
  }
  sc := newCepFsStorageClass()
  createStorageClass(sc)
}

func ensureObcStorageClass() {
  if avoidObcStorageClass() {
    return
  }
  sc := newObcStorageClass()
  createStorageClass(sc)
}   
```

The newFooStorageClass() would be exactly from your first commit.
And all the filtering goes into the avoidFooSC functions.

Just a rough sketch.

What do you think?

FWIW, yesterday, I have started a patch to first consolidate the filtering in createStorageClasses first.

obnoxxx:
/hold

holding b/c of discussion of approach

obnoxxx:
FYI: #954  is the WIP PR where I consolidated the filtering for storage classes.

jarrpa:
Responding to @obnoxxx 's proposed refactor:

There's a consequence that this would involve some refactoring to the external resources code as well, since we would now need to do individual create wrapper function calls for each StorageClass as we encounter them rather than creating them all after all information is gathered.

This would also necessitate breaking up the StorageClusterConfiguration struct. The reason being that the struct contains the information as to whether the SC would be created or not. Admittedly, this is just a convenience to not duplicate the code in `createStorageClasses()`. From that standpoint I prefer leaving things as they are, to avoid code duplication. The only readon for the split in filtering is because the OBC SC needs to also check `if initData.Spec.ExternalStorage.Enable || err == nil && !avoidObjectStore(platform) `, but otherwise the filtering checks are the same.

obnoxxx:
> Responding to @obnoxxx 's proposed refactor:
> 
> There's a consequence that this would involve some refactoring to the external resources code as well, since we would now need to do individual create wrapper function calls for each StorageClass as we encounter them rather than creating them all after all information is gathered.

Right, there was something about that caller... But it might be fine to refactor that a bit too, since this is a little horrible anyway... :wink: 

> This would also necessitate breaking up the StorageClusterConfiguration struct.

Well, in my approach, the `StorageClassConfiguration` does not necessarily exist. We might not  need it.
But I do not understand why we would need to break it up yet, since it is very generic.

> The reason being that the struct contains the information as to whether the SC would be created or not.

Right, but that information is rather generic. Let me thing a bit more about it to understand better.

> Admittedly, this is just a convenience to not duplicate the code in `createStorageClasses()`. From that standpoint I prefer leaving things as they are, to avoid code duplication.

I didn't intend to introduce a lot of code duplication. But get rid of lists and case/if-else statements. 

> The only readon for the split in filtering is because the OBC SC needs to also check `if initData.Spec.ExternalStorage.Enable || err == nil && !avoidObjectStore(platform) `, but otherwise the filtering checks are the same.

This is not needed, afaict, and I started to get rid of it in #945. 

Let me think a bit more about it...

egafford:
So what I'm seeing here on review is that we do have a different set of filter criteria in the managed vs external ObjectStore. We could try to implement a formal strategy or behavior pattern around this filter. (The caller of createStorageClasses could pass a filter set defined by an interface.) That would likely be more complex than the current implementation, though, and I'm not certain there's a real improvement in maintainability by doing so.

I'm not in favor of pushing logic specific to the external ObjectStore case into storageclasses.go; to my mind that would be a pretty clear violation of encapsulation that would make the code less easy to maintain.

> > Responding to @obnoxxx 's proposed refactor:
> > There's a consequence that this would involve some refactoring to the external resources code as well, since we would now need to do individual create wrapper function calls for each StorageClass as we encounter them rather than creating them all after all information is gathered.
> 
> Right, there was something about that caller... But it might be fine to refactor that a bit too, since this is a little horrible anyway... wink
> 
> > This would also necessitate breaking up the StorageClusterConfiguration struct.
> 
> Well, in my approach, the `StorageClassConfiguration` does not necessarily exist. We might not need it.
> But I do not understand why we would need to break it up yet, since it is very generic.
> 
> > The reason being that the struct contains the information as to whether the SC would be created or not.
> 
> Right, but that information is rather generic. Let me thing a bit more about it to understand better.
> 
> > Admittedly, this is just a convenience to not duplicate the code in `createStorageClasses()`. From that standpoint I prefer leaving things as they are, to avoid code duplication.
> 
> I didn't intend to introduce a lot of code duplication. But get rid of lists and case/if-else statements.
> 
> > The only readon for the split in filtering is because the OBC SC needs to also check `if initData.Spec.ExternalStorage.Enable || err == nil && !avoidObjectStore(platform) `, but otherwise the filtering checks are the same.
> 
> This is not needed, afaict, and I started to get rid of it in #945.
> 
> Let me think a bit more about it...

obnoxxx:
> So what I'm seeing here on review is that we do have a different set of filter criteria in the managed vs external ObjectStore. We could try to implement a formal strategy or behavior pattern around this filter. (The caller of createStorageClasses could pass a filter set defined by an interface.) That would likely be more complex than the current implementation, though, and I'm not certain there's a real improvement in maintainability by doing so.

Right... I think it's not worth it.
 
> I'm not in favor of pushing logic specific to the external ObjectStore case into storageclasses.go; to my mind that would be a pretty clear violation of encapsulation that would make the code harder to maintain.

Currently, there is some external specific logic in storageclasses.go, but not needed (any more).

I think the current patch here is a clear improvement as is and is also preparing for possible further changes.

egafford:
For what it's worth, @obnoxxx, on _another_ review, I think I see what you're saying, which is to just call directly into the creation functions for each storage type directly from the clients of storageclasses.go (whether local or external).

I'm also seeing that we create other k8s resources (ConfigMaps and Secrets, notably) inline in the case statement of external_resources, so it wouldn't be out of keeping with our current codebase. (This was the piece I was missing earlier.)

Sorry for not fully getting your proposal earlier. I'll poke at this tomorrow and see whether it's straightforward to implement. If it's not, I think you're correct that this PR is a step in the right direction as-is.

egafford:
@obnoxxx, after some examination playing around with external_resources.go, in order to make the individual functions for {avoid, create} * {fs, blockpool, obc}, we would have to duplicate a fair bit of code, especially around error handling and negative cases. It may be that a deeper refactoring of external_resources.go itself would allow us to do this sensibly. Still, while I do very much like your proposed approach from a code cleanliness point of view, I think this patch may well be the right approach, for now at least.

umangachapagain (approved):

pkg/controller/storagecluster/storageclasses.go
```diff
@@ -14,13 +14,6 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 )
 
-const (
-	// The following constants are the indices at which StorageClasses are returned from newStorageClasses and in
-	// which they should be passed to createStorageClasses.
-	cephFileSystemIndex = 0
-	cephBlockPoolIndex  = 1
-)
-
```
jarrpa: I believe this block should be in the previous commit.

jarrpa (changes requested):
The code looks good, but I think there's a small cleanup required between commits.

pkg/controller/storagecluster/storageclasses.go
egafford: @jarrpa, while the OBC index was removed in the previous commit, createSnapshotClasses in volumsnapshotterclasses.go still relies on these two index values in its switching block.

jarrpa: Ah! I see. Thanks!

jarrpa (approved):
/lgtm

PR #926, opened 2020-12-01
updates Dockerfile.registry for apiextensions/v1 support

umangachapagain:
updating Dockerfile.registry to be able to support operator
registry builds with CRDs using apiextensions/v1.

also adds nsswitch.conf to fix the issue with non-root
users not being able to write this config.


jarrpa (approved):
/lgtm

PR #930, opened 2020-12-02
CSV generator for OCS on OpenShift Dedicated

umangachapagain:
For OCS on OpenShift Dedicated efforts, we have a new operator that will watch the Storage Cluster resource.
This new operator needs to be deployed along with the other operators.

However, this new deployment is only concerned with OCS on OpenShift Dedicated (OSD in short; not the Cpeh OSD). So, adding a flag to inject the deployment into CSV specially for OSD.

The generated CSV will not be committed into the OCS Operator repository in it's current state.
It might be worth looking into decoupling these generation tools and processes into it's own repository.

Usage:
`GEN_OSD=true make gen-latest-csv`

`GEN_OSD` is set to `false` by default.

### DIFF
--- a/tools/csv-merger/csv-merger.go
+++ b/tools/csv-merger/csv-merger.go
@@ -40,6 +40,7 @@ var (
 	noobaaCSVStr       = flag.String("noobaa-csv-filepath", "", "path to noobaa csv yaml file")
 	ocsCSVStr          = flag.String("ocs-csv-filepath", "", "path to ocs csv yaml file")
 	timestamp          = flag.String("timestamp", "false", "bool value to enable/disable timestamp changes in CSV")
+	osd                = flag.String("gen-osd", "false", "bool value to generate CSV for OSD")
 
 	rookContainerImage       = flag.String("rook-image", "", "rook operator container image")
 	cephContainerImage       = flag.String("ceph-image", "", "ceph daemon container image")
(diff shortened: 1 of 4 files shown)


obnoxxx:
@umangachapagain - can you provide some context in the PR message?

jarrpa:
> @umangachapagain - can you provide some context in the PR message?

Indeed. Please explain what OSD is and what the deployer does.

umangachapagain:
@nb-ohad Please review and verify the PR description. I've added what I understood (and it may not be correct)

@obnoxxx @jarrpa PTAL.

nb-ohad:
@umangachapagain Can we change all OSD references to OSD deployer?
for example, instead of `LATEST_OSD_IMAGE` let`s do `LATEST_OSD_DEPLOYER_IMAGE` 
and instead of `gen-osd` let's do `gen-osd-deployer-csv`.

I think it will help in minimizing the confusion.

nb-ohad:
Also regarding the usage I would prefer a new make target, so instead of `GEN_OSD=true make gen-latest-csv` we will do `make `get-latest-csv-for-osd` or `get-latest-csv-for-openshift-dedicated` if we want to be more explicit, but i think the name si to long.

jarrpa:
I like the PR description, now please also add it in the commit message.

tools/csv-merger/csv-merger.go
```diff
@@ -975,6 +996,45 @@ func getMetricsExporterDeployment() appsv1.DeploymentSpec {
...
+		Template: corev1.PodTemplateSpec{
+			ObjectMeta: metav1.ObjectMeta{
+				Labels: map[string]string{
+					"app": "ocs-osd-deployer",
+				},
+			},
+			Spec: corev1.PodSpec{
+				Containers: []corev1.Container{
+					{
+						Name:            "deployer",
```
iamniting, line 1020: I would suggest the pod name should be the same as the deployment name.

iamniting (changes requested):

obnoxxx:
/hold

I am just putting a protective hold on this as there is a discussion going whether to put the csv generation somewhere else.
This doesn't prevent us from continuing on this PR and getting it into shape.

jarrpa:
I don't think this is valid anymore? Is it?

nb-ohad:
It should be with some changes

umangachapagain:
This is not required as a better alternative was found.

PR #931, opened 2020-12-02
install ginkgo but don't edit the go.mod file

raghavendra-talur:
If ginkgo binary is not found on the system when running the functest,
we were installing ginkgo using go get. The downside to that is that it
updates the go.mod and go.sum file. It is not the intention.

Go has install command which helps us to only install the binary without
updating go.mod and go.sum. This commit updates the hack script to only
install ginkgo.


raghavendra-talur:
cc @iamniting

jarrpa (approved):
/lgtm

PR #932, opened 2020-12-03
functests: improvements to allow StorageCluster with different configurations

raghavendra-talur:
This will be required for testing scenarios where we want to install StorageClusters that are not of default config.

This is in preparation to the arbiter feature.

### DIFF
--- a/functests/ocs/storagecluster_creation_test.go
+++ b/functests/ocs/storagecluster_creation_test.go
@@ -11,7 +11,6 @@ import (
 	tests "github.com/openshift/ocs-operator/functests"
 	ocsv1 "github.com/openshift/ocs-operator/pkg/apis/ocs/v1"
 	"github.com/openshift/ocs-operator/pkg/controller/util"
-	deploymanager "github.com/openshift/ocs-operator/pkg/deploy-manager"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -32,7 +31,7 @@ func initSCCreation() (*SCCreation, error) {
 }
 
 func (sccObj *SCCreation) populateDuplicateSC() error {
-	defaultStorageCluster, err := deploymanager.DefaultStorageCluster()
+	defaultStorageCluster, err := tests.DeployManager.DefaultStorageCluster()
 	if err != nil {
 		return err
 	}
--- a/pkg/deploy-manager/config.go
+++ b/pkg/deploy-manager/config.go
@@ -27,8 +27,11 @@ const DefaultStorageClusterName = "test-storagecluster"
 // DefaultStorageClassRBD is the name of the ceph rbd storage class the test suite installs
 const DefaultStorageClassRBD = DefaultStorageClusterName + "-ceph-rbd"
 
-// MinOSDsCount represents the minimum number of OSDs required for this testsuite to run.
-const MinOSDsCount = 3
+const minOSDsCount = 3
+
+func (t *DeployManager) getMinOSDsCount() int {
+	return minOSDsCount
+}
 
 //nolint:errcheck // ignoring err check as causing failures
 func init() {
--- a/pkg/deploy-manager/config.go
+++ b/pkg/deploy-manager/config.go
@@ -40,14 +40,33 @@ func init() {
 	nbv1.SchemeBuilder.AddToScheme(scheme.Scheme)
 }
 
+// storageClusterConfig stores the configuration of the Storage Cluster that is/will be
+// created by the deploy manager.
+type storageClusterConfig struct {
+	// As deploy manager is mainly used as a replacement for openshift/console,
+	// this should capture the options that are presented in the console for OCS
+	// installation.
+	// For example Encryption, Arbiter etc
+	// Eligibility Check:
+	// Is this option presented in the console and/or does this option cause a
+	// change in the inital Storage Cluster CR provided to the ocs-operator? Yes, it belongs here.
+
+	// Functional tests also use the deploy manager for deployment and if any
+	// state needs to be stored for the storageCluster, it belongs here.
+
+	// TODO: Move all the in-built defaults to this struct. These defaults are
+	// either hardcoded or are defined as globals in the deploy manager.
+}
+
 // DeployManager is a util tool used by the functional tests
 type DeployManager struct {
-	olmClient      *olmclient.Clientset
-	k8sClient      *kubernetes.Clientset
-	rookClient     *rookclient.Clientset
-	ocsClient      *rest.RESTClient
-	crClient       crclient.Client
-	parameterCodec runtime.ParameterCodec
+	olmClient          *olmclient.Clientset
+	k8sClient          *kubernetes.Clientset
+	rookClient         *rookclient.Clientset
+	ocsClient          *rest.RESTClient
+	crClient           crclient.Client
+	parameterCodec     runtime.ParameterCodec
+	storageClusterConf *storageClusterConfig
 }
 
 // GetCrClient is the function used to retrieve the controller-runtime client
@@ -80,7 +99,7 @@ func (t *DeployManager) GetNamespace() string {
 	return InstallNamespace
 }
 
-// NewDeployManager is the way to create a DeployManager struct
+// NewDeployManager creates a DeployManager struct with default configuration
 func NewDeployManager() (*DeployManager, error) {
 	codecs := serializer.NewCodecFactory(scheme.Scheme)
 	parameterCodec := runtime.NewParameterCodec(scheme.Scheme)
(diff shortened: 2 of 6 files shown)


raghavendra-talur:
cc @iamniting 

@jarrpa @obnoxxx @umangachapagain Please review.

functests/ocs/storagecluster_creation_test.go
```diff
@@ -32,7 +31,7 @@ func initSCCreation() (*SCCreation, error) {
 }
 
 func (sccObj *SCCreation) populateDuplicateSC() error {
-	defaultStorageCluster, err := deploymanager.DefaultStorageCluster()
+	defaultStorageCluster, err := tests.DeployManager.DefaultStorageCluster()
```
obnoxxx, line 34: I think this hunk is in the wrong commit - it should be in the "use global deployManager" commit for all I can tell.

obnoxxx (changes requested):

obnoxxx:
There was a linter error in the last commit. Pushed a fix. We can squash it later if it works... @raghavendra-talur

pkg/deploy-manager/storagecluster.go
```diff
@@ -119,7 +119,7 @@ func DefaultStorageCluster() (*ocsv1.StorageCluster, error) {
 			StorageDeviceSets: []ocsv1.StorageDeviceSet{
 				{
 					Name:     "example-deviceset",
-					Count:    MinOSDsCount,
+					Count:    t.storageClusterConfig.minOSDsCount,
```
umangachapagain, line 122: If it returns a default StorageCluster, shouldn't this be set to default as well?

and for custom Storage Cluster we can have StorageClusterWithOptions(StorageClusterOptions).

pkg/deploy-manager/config.go
```diff
@@ -36,6 +36,7 @@ func init() {
 	nbv1.SchemeBuilder.AddToScheme(scheme.Scheme)
 }
 
+// StorageClusterOptions - silly comment enforced by linter
```
umangachapagain, line 39: I'd like to see comments about what it is and what it can do. I'm already confused with this blank struct.

umangachapagain (changes requested):

pkg/deploy-manager/config.go
```diff
@@ -27,23 +27,31 @@ const DefaultStorageClusterName = "test-storagecluster"
...
 	rookcephv1.SchemeBuilder.AddToScheme(scheme.Scheme)
 	nbv1.SchemeBuilder.AddToScheme(scheme.Scheme)
 }
 
+type StorageClusterOptions struct {
+}
+
+type storageClusterConfig struct {
+	minOSDsCount int
+}
```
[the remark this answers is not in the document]
obnoxxx: indeed, i wanted to ask the same and was distracted reading through the code :wink:

pkg/deploy-manager/config.go
obnoxxx: sure - this was just to pass the tests and see if we hit any other failures after that ;-)

However, I am really irritated by that linter thing failing on missing comments. it often leads to meaningless and redundant comments. :frowning_man:

jarrpa: See my previous response to that complaint, I still like having the comments for review purposes.

raghavendra-talur: This one is totally my fault. Will update along with other fixes soon.

IMO,
1. Linter must fail, that is the only way to get documentation.(Easy to enforce)
2. The description should tell more about it than the name. (Difficult to enforce)

pkg/deploy-manager/config.go
```diff
@@ -77,8 +104,17 @@ func (t *DeployManager) GetNamespace() string {
...
 
-// NewDeployManager is the way to create a DeployManager struct
+// NewDeployManager creates a DeployManager struct with default configuration
 func NewDeployManager() (*DeployManager, error) {
+	return newDeployManager(GetDefaultStorageClusterOptions())
+}
+
+// NewDeployManagerWithOptions creates a DeployManager struct with user provided configuration
+func NewDeployManagerWithOptions(opts *StorageClusterOptions) (*DeployManager, error) {
+	return newDeployManager(opts)
```
jarrpa, line 114: This seems to just be a passthrough to a function with identical parameters... why? Should this just be the function currently being called?

pkg/deploy-manager/config.go
```diff
@@ -27,23 +27,31 @@ const DefaultStorageClusterName = "test-storagecluster"
...
-	ocsClient      *rest.RESTClient
-	crClient       crclient.Client
-	parameterCodec runtime.ParameterCodec
+	olmClient            *olmclient.Clientset
+	k8sClient            *kubernetes.Clientset
+	rookClient           *rookclient.Clientset
+	ocsClient            *rest.RESTClient
+	crClient             crclient.Client
+	parameterCodec       runtime.ParameterCodec
+	storageClusterConfig *storageClusterConfig
```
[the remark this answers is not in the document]
jarrpa: I think this is fine as an organizational measure. As we expand the number of parameters, changing this struct doesn't pollute the general DeployManager structure.

pkg/deploy-manager/config.go
```diff
@@ -37,14 +37,41 @@ func init() {
...
+	// provided to the ocs-operator?
+
+	// TODO: Move all the in-built defaults to this struct. These defaults are
+	// either hardcoded or are defined as globals in the deploy manager.
+}
+
+// storageClusterConfig stores the current configuration for the Storage Cluster.
+// This gets created when the deploy manager is instantiated and does not guarantee that the StorageCluster is created yet.
+// This is private to the deploy manager. It can have members which map one to
+// one to StorageClusterOptions or
```
jarrpa, line 61: Incomplete comment?

pkg/deploy-manager/config.go
obnoxxx: I am all for documentation/comments - where reasonable.
I am against silly and redundant comments, which is what happens frequently...

Anyway: Just remember to run `make ocs-operator-ci` before creating a PR 😉

pkg/deploy-manager/config.go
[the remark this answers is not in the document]
raghavendra-talur: Thanks for making me think harder. I have dropped StorageClusterOptions. To configure, we will have methods on the DeployManager, instead of having a big struct of options.

pkg/deploy-manager/config.go
```diff
@@ -76,8 +84,17 @@ func (t *DeployManager) GetNamespace() string {
...
 func NewDeployManager() (*DeployManager, error) {
+	return newDeployManager(GetDefaultStorageClusterOptions())
+}
+
+// NewDeployManagerWithOptions creates a DeployManager struct with user provided configuration
+func NewDeployManagerWithOptions(opts *StorageClusterOptions) (*DeployManager, error) {
+	return newDeployManager(opts)
+}
+
+func newDeployManager(opts *StorageClusterOptions) (*DeployManager, error) {
```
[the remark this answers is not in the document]
raghavendra-talur: obsolete with the update. Done.

pkg/deploy-manager/config.go
```diff
@@ -142,12 +159,30 @@ func NewDeployManager() (*DeployManager, error) {
...
 	}, nil
 }
+
+// GetDefaultStorageClusterOptions returns a StorageClusterOptions type prefilled with default values
+// Clients can modify it and pass the struct when creating a DeployManager
+func GetDefaultStorageClusterOptions() *StorageClusterOptions {
+	return &StorageClusterOptions{}
+}
+
+func getStorageClusterConfigFromOptions(opts *StorageClusterOptions) (*storageClusterConfig, error) {
```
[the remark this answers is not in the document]
raghavendra-talur: Done.

pkg/deploy-manager/config.go
raghavendra-talur: `make`,`make build-functest` and `make ocs-operator-ci` all three passed :)

pkg/deploy-manager/config.go
raghavendra-talur: Obsolete with the update.

pkg/deploy-manager/config.go
raghavendra-talur: Removed.

obnoxxx (approved):

pkg/deploy-manager/config.go
```diff
@@ -27,8 +27,11 @@ const DefaultStorageClusterName = "test-storagecluster"
 // DefaultStorageClassRBD is the name of the ceph rbd storage class the test suite installs
 const DefaultStorageClassRBD = DefaultStorageClusterName + "-ceph-rbd"
 
-// MinOSDsCount represents the minimum number of OSDs required for this testsuite to run.
-const MinOSDsCount = 3
+const minOSDsCount = 3
+
+func (t *DeployManager) getMinOSDsCount() int {
```
umangachapagain, line 32: if it's always going to return a constant, what benefit does this method add?

raghavendra-talur: This is in preparation for arbiter. MinOSDCount for arbiter is 4. This function will return the right value as it also has access to the storagcluster configuration and can check if arbiter is enabled.

umangachapagain (commented):
Since there is no design document or a reference PR that uses these improvements it is not very clear what value it adds or how it will be used. Can we have a PR that can assist in understanding this enhancement?

Here, I just see an empty struct being plugged in without anything concrete.

pkg/deploy-manager/config.go
umangachapagain: ok. That makes the intent clearer.
Can we add a generic comment for that?

raghavendra-talur:
> Since there is no design document or a reference PR that uses these improvements it is not very clear what value it adds or how it will be used. Can we have a PR that can assist in understanding this enhancement?
> 
> Here, I just see an empty struct being plugged in without anything concrete.

I understand. Raised a WIP PR #940 . If it helps, we can get all commits into a single PR too.

umangachapagain (approved):
/lgtm

PR #936, opened 2020-12-03
updates openshift/api to enable Quickstart APIs for Console

umangachapagain:
Updating openshift/api dependency to release-4.7.
It still supports Kubernetes 1.19.z and adds Quickstart API required
by the OCS Console.


umangachapagain:
@dannyzaken I see NooBaa is dependent on `openshift/api/routes` v3, will this cause any trouble?

dannyzaken:
I am not familiar with the changes and how compatible it is. 
@nb-ohad do you have any idea if it will affect noobaa-operator?

nb-ohad:
@dannyzaken not really, but why does it matter? We are not exposing routes apis in our CRDs so I don't see an issue here

jarrpa (approved):
/lgtm

PR #937, opened 2020-12-03
Adding minKubeVersion to CSV

umangachapagain:
minKubeVersion of 1.16.0 ensures that we don't attempt installing on lower versions which does not support apiextensions/v1.

### DIFF
--- a/tools/csv-merger/csv-merger.go
+++ b/tools/csv-merger/csv-merger.go
@@ -648,6 +648,11 @@ The NooBaa operator deploys and manages the [NooBaa][2] Multi-Cloud Gateway on O
 	if *skipRange != "" {
 		ocsCSV.Annotations["olm.skipRange"] = *skipRange
 	}
+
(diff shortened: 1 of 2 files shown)


jarrpa (approved):
/lgtm

tools/csv-merger/csv-merger.go
```diff
@@ -648,6 +648,11 @@ The NooBaa operator deploys and manages the [NooBaa][2] Multi-Cloud Gateway on O
 	if *skipRange != "" {
 		ocsCSV.Annotations["olm.skipRange"] = *skipRange
 	}
+
+	// apiextensions/v1 is available only on Kubernetes 1.16+
+	// This ensures that we don't try to install on lower versions ok K8s
```
obnoxxx, line 653: ok ==> of

obnoxxx (changes requested):

PR #938, opened 2020-12-03
cephcluster: add annotation for each deviceType in cephCR

crombus:
Problem:  crushDeviceClass annotations not being reflected in cephCR.

Important: So to verify that deviceClass for an osd is updated. Though it logical grouping. Won't effect the performance.

Solution: set annotation in cephCR.

### DIFF
--- a/pkg/controller/storagecluster/cephcluster.go
+++ b/pkg/controller/storagecluster/cephcluster.go
@@ -471,6 +471,12 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster, serverVersion *version.
 				placement = ds.Placement
 			}
 
+			// Annotation crushDeviceClass ensures osd with different CRUSH device class than the one detected by Ceph
+			annotations := map[string]string{
+				"crushDeviceClass": ds.DeviceType,
+			}
+			ds.DataPVCTemplate.Annotations = annotations
+
 			set := rook.StorageClassDeviceSet{
 				Name:                 fmt.Sprintf("%s-%d", ds.Name, i),
 				Count:                count,
--- a/pkg/controller/storagecluster/storagecluster_controller_test.go
+++ b/pkg/controller/storagecluster/storagecluster_controller_test.go
@@ -70,8 +70,14 @@ var mockCephClusterNamespacedName = types.NamespacedName{
 
 var storageClassName = "gp2"
 var volMode = corev1.PersistentVolumeBlock
+var annotations = map[string]string{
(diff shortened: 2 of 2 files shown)


jarrpa:
*Always* run `make ocs-operator-ci` before pushing, so we don't waste AWS resources. :)

Also, why are we doing this? I thought the CRUSH device type was only for OSDs.

obnoxxx (changes requested):
Please provide context in the PR description and in commit messages. E.g.:

* What is the problem we are trying to solve / feature request.
* Why is this important.
* How are we addressing it.
* etc

...

crombus:
> _Always_ run `make ocs-operator-ci` before pushing, so we don't waste AWS resources. :)
> 
ok.
> Also, why are we doing this? I thought the CRUSH device type was only for OSDs.
when I am setting https://github.com/openshift/ocs-operator/blob/136e35698cb4b2bb2172c95bb4fd777847c369e9/pkg/controller/storagecluster/reconcile.go#L428

This is not being reflected in cephclusterCR.

crombus:
> @crombus - just out of curiosity: why is your branch for this PR called `validation_test`? 🤔

the branch was created to write a test for the deviceType and validate that metadata is bing set.

Doing that I came to know that metadata should be set in cephcluster. So did that first here and second commit will be for test.

pkg/controller/storagecluster/cephcluster.go
```diff
@@ -485,6 +490,8 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster, serverVersion *version.
 				Encrypted:            sc.Spec.Encryption.Enable,
 			}
 
+			ds.DataPVCTemplate.Annotations = annotations
+			set.VolumeClaimTemplates = append(set.VolumeClaimTemplates, ds.DataPVCTemplate)
```
obnoxxx, line 494: Doesn't this mean we'll have two copies of `ds.DataPVCTemplate` in set?
Wouldn't it be better to first set the annotations into `ds.DataPVCTemplate.Annotation` and then define `set`?

crombus: added the annotations  before set.

obnoxxx (approved):
much better, thanks!

/lgtm

pkg/controller/storagecluster/cephcluster.go
obnoxxx: thanks, much better!

PR #940, opened 2020-12-04
Add Stretch configuration to the Storage Cluster

raghavendra-talur:
This PR adds the arbiter mode of deployment to the Storage Cluster.

Two new fields are added to the Storage Cluster

* ArbiterSpec
* NodeTopologies

If ArbiterSpec.Enable is set to true *and* if
NodeTopologies.ArbiterLocation is set then the Ceph Cluster is
configured with arbiter enabled.

If ArbiterSpec.Enable is true and no NodeTopologies.ArbiterLocation is
provided then the Storage Cluster fails to reconcile.

If ArbiterSpec.Enable is false then NodeTopologies.ArbiterLocation is
ignored even if set.

The arbiter mode differs from the non-arbiter mode in the following
ways:

* The minimum and the maximum number of MONs are 5 instead of 3.
* The OSDs run only in 2 of the 3 failure domains. The third failure
  domain is only used to host a MON pod.
* Mininum number of OSDs in each failure domain is 2 instead of 1.
* The CRUSH Map is configured in replica 4 configuration instead of 3.
* The OSDs can be scaled in multiples of 2 and not 3.
* The placement algorithm doesn't restrict the arbiter MON pod to nodes
  which have the default/user-provided OCS node label. Rest of the ceph
  pods have the nodeAffinity same as in the non-arbiter mode.
* The arbiter MON pod is scheduled to run on any of the nodes in the
  ArbiterLocation in the failure domain.
* The arbiter MON pod has toleration for the master node taint by default. It
  can be disabled by setting ArbiterSpec.DisableMasterNodeToleration to true.

ocs-operator's placement strategy for non-arbiter mode is to select a failure
domain and then spread the ceph resources across the failure domain.

Keeping with the same principle, ocs-operator determines the failure domain
automatically in the the arbiter mode too and tries to match one of the failure
domain locations to the NodeTopologies.ArbiterLocation.

In both the modes, it might sometimes be desirable that the user gets to
specify the failure domain through the Storage Cluster and not rely on the
automatic detection. They may also wish to specify the node topology
themselves.  For this reason, we did not add the ArbiterLocation field to the
ArbiterSpec and added it to the TopologyMap struct instead.

Work needs to be done to reconcile StorageClusterSpec.FlexibleScaling,
StorageClusterSpec.NodeTopologies.FailureDomain, and
StorageClusterStatus.FailureDomain to let users choose the node topology and
the failure domain. This can be done in the future, we have mad


pkg/apis/ocs/v1/storagecluster_types.go
```diff
@@ -155,6 +156,16 @@ type EncryptionSpec struct {
...
 }
 
+// ArbiterSpec defines the arbiter options for the StorageCluster. If defined,
+// the Storage Cluster is deployed in stretch configuration. In stretch
+// configuration, a minimum of 2 OSD pods and 2 mon pods are created in each of
+// the 2 data zones. In the third zone(arbiter), only a mon pod is created.
+// Arbiter mon pods don't require the nodes to have the ocs node label.
+type ArbiterSpec struct {
+	// +optional
+	FailureDomainValue string `json:"failureDomainValue,omitempty"`
```
raghavendra-talur, line 166: @jarrpa @umangachapagain @obnoxxx @iamniting Although I still need to work on the placement code for arbiter, I would like to get your opinion on this change in the API.

Rook needs both the failure domain name and the 3 values for that failure domain to configure the stretch cluster. ocs-operator determines the failure domain name internally and therefore requires only the value of the failure domain that will be arbiter.

Is `FailureDomainValue` an acceptable name ? If not, any suggestions?

raghavendra-talur: Here is the change in Rook https://github.com/rook/rook/pull/6489/commits/b91f4211c9833f89b107599cd6fd091f4c7f06f2#diff-23d4ca6ed288cf0c307912d860846319f62eb5f60846961916c0395a82d76cfaR123

travisn: 1. How about calling it `failureDomainLabel`, the same as in Rook?
2. Seems like this property should be passed by the UI, depending on the failure domains supported by the UI. No need to rely on the internal computation of failure domain in OCS in arbiter clusters IMO.

pkg/deploy-manager/arbiter.go
```diff
@@ -0,0 +1,67 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// electArbiterZone is a helper function just for the tests. In real deployments, the arbiter zone will be picked by the user.
```
travisn, line 11: If this is just for tests, can it be included in the test code? Or at a minimum, let's include `Test` in the function name.

pkg/deploy-manager/arbiter.go
```diff
@@ -0,0 +1,67 @@
...
+// In our tests, we will just pick the zone with the least number of nodes as the arbiter zone.
+func (t *DeployManager) electArbiterZone() (arbiterZoneElect string, err error) {
+	nodes, err := t.k8sClient.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{LabelSelector: "node-role.kubernetes.io/worker"})
```
travisn, line 14: Might there be a zone with only a master and no worker? The common arbiter scenario is to have just the master node in the arbiter zone and no worker, but maybe that won't apply in the clusters where this test will run.

pkg/deploy-manager/storagecluster.go
```diff
@@ -350,13 +350,22 @@ func (t *DeployManager) labelWorkerNodes() error {
 	}
 
 	labeledCount := 0
+	arbiterZone, err := t.GetArbiterZone()
```
travisn, line 353: This is a node label, right? This may help clarify.
```suggestion
	arbiterZoneLabel, err := t.GetArbiterZoneLabel()
```

pkg/deploy-manager/config.go
```diff
@@ -27,8 +27,15 @@ const DefaultStorageClusterName = "test-storagecluster"
...
 const DefaultStorageClassRBD = DefaultStorageClusterName + "-ceph-rbd"
 
-// MinOSDsCount represents the minimum number of OSDs required for this testsuite to run.
-const MinOSDsCount = 3
+const minOSDsCount = 3
+const minOSDsCountArbiter = 4
+
+func (t *DeployManager) getMinOSDsCount() int {
+	if t.IsArbiterEnabled() {
+		return minOSDsCountArbiter
```
travisn, line 35: This is the only change I see related to OSDs. Will this cause the `storageClassDeviceSet.Count` to be `4`? I believe the node affinity, topology spread constraints, etc would be the same in a stretch cluster as other non-arbiter clusters, as long as the topology spread constraints are reliable to spread the ODSs evenly first across the two zones, then across the hosts within zones. For example, say we have 4 worker nodes in each of the two non-arbiter zones, giving us 8 worker nodes where OSDs could be started. We need to ensure two OSDs end up in each zone, rather than all four OSDs potentially in the same zone. Do the OCS topology spread constraints handle that?

travisn (changes requested):

pkg/apis/ocs/v1/storagecluster_types.go
raghavendra-talur: If I am not wrong, ocs-operator has been determining the failure domain itself. The UI too, for the arbiter config is going to pass only the value of the failure domain(for example us-east-2b).

IIUC,  in the UI, the failure domain is always shown as a zone.

travisn: Rook expects the full label to be set in the `failureDomainLabel`, such as `topology.kubernetes.io/zone`, not just the failure domain. Is this not the same setting that OCS will pass on to rook? Or is this setting just intended to be `zone` (in this example)?

raghavendra-talur: Based on @jarrpa's and @travisn's comments, the name of the field is now ArbiterLocation, this is the input given to the ocs-operator in the StorageCluster CR and is just the value of the failure domain that is for arbiter. The failure domain label that is passed to rook is a complete label.

pkg/deploy-manager/arbiter.go
raghavendra-talur: This code belongs to a different package in the ocs-operator, deploy-manager which is sort of a replacement/alternative to the openshift console. Therefore is it not prefixed with `test`. It is not compiled into the ocs-operator binary though.

pkg/deploy-manager/arbiter.go
raghavendra-talur: Yes, I am making an assumption for the deploy-manager that the arbiter pods will be deployed on a worker node. This is to keep the deploy manager simple. The placement algorithm in the ocs-operator now has the master node toleration set for the arbiter pods.

pkg/deploy-manager/config.go
raghavendra-talur: Yes. The above scenario is covered; added in the `placement` related commit.

raghavendra-talur:
@jarrpa @travisn @obnoxxx @umangachapagain @iamniting I believe this is PR is now ready for review.

I have provided design details in the description of the PR.

obnoxxx:
fyi, functest was failing like this before the update:
```
Failure [21.196 seconds]
[BeforeSuite] BeforeSuite 
/go/src/github.com/openshift/ocs-operator/functests/ocs/ocs_test.go:17
  Expected
      <*errors.errorString | 0xc00032a230>: {
          s: "No arbiter nodes found, we need atleast 1",
      }
  to be nil
  /go/src/github.com/openshift/ocs-operator/functests/setup_teardown.go:17
------------------------------ 
```

@raghavendra-talur - what did you change regarding this in the last update?

raghavendra-talur:
> fyi, functest was failing like this before the update:
> 
> ```
> Failure [21.196 seconds]
> [BeforeSuite] BeforeSuite 
> /go/src/github.com/openshift/ocs-operator/functests/ocs/ocs_test.go:17
>   Expected
>       <*errors.errorString | 0xc00032a230>: {
>           s: "No arbiter nodes found, we need atleast 1",
>       }
>   to be nil
>   /go/src/github.com/openshift/ocs-operator/functests/setup_teardown.go:17
> ------------------------------ 
> ```
> 
> @raghavendra-talur - what did you change regarding this in the last update?

The last update only added csv files that are generated when `make gen-latest-csv` is run.

Regarding the functest failure, the arbiter mode of functest expects 5 worker nodes. I will edit it such that it can run with the existing infra and/or I will update here after editing the functest ci job to provide more worker nodes.

iamniting (changes requested):

pkg/apis/ocs/v1/storagecluster_types.go
umangachapagain: > Based on @jarrpa's and @travisn's comments, the name of the field is now ArbiterLocation
>
Sounds good, but where is this comment?

umangachapagain (changes requested):
Commit - `deploy: add files updates by 'make gen-latest-csv'` should have changes related to Rook CRDs only. It seems there are some unwanted changes from a later commit.

pkg/controller/storagecluster/cephcluster.go
```diff
@@ -343,8 +345,17 @@ func newExternalCephCluster(sc *ocsv1.StorageCluster, cephImage string, monitori
 	return externalCephCluster
 }
 
-func getMonCount(nodeCount int) int {
-	count := defaults.MonCountMin
+func getMinNodes(sc *ocsv1.StorageCluster) int {
+
+	if arbiterEnabled(sc) {
+		// We need a minimum of 2 OSDs in each of the two zones when arbiter is enabled
+		return defaults.ArbiterModeDeviceSetReplica * 2
```
umangachapagain, line 352: `defaults.ArbiterModeDeviceSetReplica` is already 2. Why do we need to multiply by 2?

pkg/controller/storagecluster/cephcluster.go
```diff
@@ -564,3 +575,43 @@ func (r *ReconcileStorageCluster) checkTuneStorageDevices(ds ocsv1.StorageDevice
 	// not a known disk type, don't tune
 	return diskSpeedUnknown, nil
 }
+
+func generateStretchClusterSpec(sc *ocsv1.StorageCluster) *cephv1.StretchClusterSpec {
```
umangachapagain, line 579:
```suggestion
func generateStretchClusterSpec(sc *ocsv1.StorageCluster) cephv1.StretchClusterSpec {
```

pkg/controller/storagecluster/cephcluster.go
```diff
@@ -564,3 +575,43 @@ func (r *ReconcileStorageCluster) checkTuneStorageDevices(ds ocsv1.StorageDevice
...
+			Name:    zone,
+			Arbiter: false,
+		})
+	}
+	stretchClusterSpec.Zones = append(stretchClusterSpec.Zones, cephv1.StretchClusterZoneSpec{
+		Name:    sc.Spec.NodeTopologies.ArbiterLocation,
+		Arbiter: true,
+	})
+
+	return &stretchClusterSpec
```
umangachapagain, line 599:
```suggestion
	return stretchClusterSpec
```

pkg/controller/storagecluster/reconcile.go
```diff
@@ -190,6 +190,13 @@ func (r *ReconcileStorageCluster) reconcilePhases(
 		}
 	}
 
+	if err := validateArbiterSpec(instance, reqLogger); err != nil {
+		instance.Status.Phase = statusutil.PhaseError
+		r.client.Update(context.TODO(), instance)
```
umangachapagain, line 195: Error needs to be handled.

umangachapagain (changes requested):

pkg/controller/storagecluster/reconcile.go
iamniting: I am doing this intentionally, I just want to update it, if it does not update it I don't care much about it.
Moreover, If we are going to handle update err then we will miss arbiter err and I do not want to club them together.

umangachapagain: Well this approach isn't going to make the linter happy.

I'd suggest this if your intent is to ignore the error
```suggestion
		_ = r.client.Update(context.TODO(), instance)
```

obnoxxx: Agreed. Be explicit (by using `_ :=`when ignoring return values, in particular errors.

pkg/controller/storagecluster/cephcluster.go
obnoxxx: quoting from the definition:
```
ArbiterModeDeviceSetReplica is the default number of Rook-Ceph StorageClassDeviceSets per StorageCluster StorageDeviceSet when arbiter is enabled
```

In normal mode we have `3`here.
Now, we want 2 replicas per zone. (So that we have two copies in each zone.) 
But I think we want to make the constant factor of `2`here configurable too.
I think for this release it might be ok, but later, we want to be able to do one copy per zone...

raghavendra-talur: Good question:
There are two equations that we need
`minDevices  == minNodes == replicaPerFailureDomain * numFailureDomains`
`minAdditionalDevicesForExpansion == minAdditionalNodesPerExpansion == numFailureDomains`

1. In the non-arbiter cases, the minimum requirement is replica 1 within the failure domain and 3 failure domains in the cluster. To expand the cluster, you can start with the addition of 1 OSD per failure domain. Therefore,
`minDevices  == minNodes == 1 * 3`
`minAdditionalDevicesForExpansion == minAdditionalNodesPerExpansion == 3`

2. In the arbiter case, however, the minimum requirement is replica 2 within the failure domain and 2 failure domains in the cluster. To expand the cluster, you can start with the addition of 1 OSD per failure domain. Therefore,
`minDevices  == minNodes == 2 * 2`
`minAdditionalDevicesForExpansion == minAdditionalNodesPerExpansion == 2`

With that explanation, and to incorporate what @obnoxxx suggested, I will update the code to do the following:
1. add a default for ReplicaPerFailureDomain for arbiter and non arbiter configurations
2. add functions that capture the above equations

raghavendra-talur: Updated

pkg/controller/storagecluster/cephcluster.go
raghavendra-talur: cephv1.MonSpec requires *cephv1.StretchClusterSpec. It just makes it easy to match with what Cephv1 defines it as.

pkg/controller/storagecluster/cephcluster.go
raghavendra-talur: cephv1.MonSpec requires *cephv1.StretchClusterSpec. It just makes it easy to match with what Cephv1 defines it as.

pkg/controller/storagecluster/reconcile.go
iamniting: Done

pkg/controller/storagecluster/cephcluster.go
```diff
@@ -422,7 +444,7 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster, serverVersion *version.
 			// enable the new behavior while the console is waiting
 			// to be updated.
 			// TODO: Remove this behavior when OCP console is updated
-			count = count / 3
+			count = count / replica
```
obnoxxx, line 447: Just highlighting this hunk to @jarrpa 's attention - this was discussed very controversally before, e.g. #835  ... :smile:

pkg/controller/storagecluster/reconcile.go
```diff
@@ -697,3 +701,13 @@ func newCleanupJob(sc *ocsv1.StorageCluster) *batchv1.Job {
 
 	return job
 }
+
+func validateArbiterSpec(sc *ocsv1.StorageCluster, reqLogger logr.Logger) error {
+
+	if sc.Spec.Arbiter.Enable {
+		if sc.Spec.NodeTopologies.ArbiterLocation == "" {
```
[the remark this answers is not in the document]
obnoxxx: done

pkg/controller/storagecluster/reconcile.go
```diff
@@ -697,3 +701,13 @@ func newCleanupJob(sc *ocsv1.StorageCluster) *batchv1.Job {
...
+			return fmt.Errorf("arbiter is set to enable but no arbiterLocation has been provided")
```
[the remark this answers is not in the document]
obnoxxx: done

pkg/controller/storagecluster/reconcile.go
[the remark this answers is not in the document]
iamniting: Done

pkg/controller/storagecluster/reconcile.go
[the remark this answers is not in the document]
iamniting: Done

pkg/controller/storagecluster/cephcluster.go
jarrpa: Yes, don't do this. We want to freeze this behavior as hard as possible.

pkg/controller/storagecluster/cephcluster.go
```diff
@@ -343,8 +340,12 @@ func newExternalCephCluster(sc *ocsv1.StorageCluster, cephImage string, monitori
 	return externalCephCluster
 }
 
+func getMinNodes(sc *ocsv1.StorageCluster) int {
+
+	return defaults.DeviceSetReplica
+}
+
 func getMonCount(nodeCount int) int {
-	count := defaults.MonCountMin
```
jarrpa, line 347: This removes the only reference to this variable in the code. Either keep this logic, or remove the MonCountMin variable.

pkg/controller/storagecluster/cephcluster.go
jarrpa: I think this is fine in this case.

pkg/controller/storagecluster/cephcluster.go
jarrpa: I think this is fine in this case.

pkg/controller/storagecluster/topology.go
```diff
@@ -219,7 +219,7 @@ func (r *ReconcileStorageCluster) ensureNodeRacks(
 // reconcileNodeTopologyMap builds the map of all topology labels on all nodes
 // in the storage cluster
 func (r *ReconcileStorageCluster) reconcileNodeTopologyMap(sc *ocsv1.StorageCluster, reqLogger logr.Logger) error {
-	minNodes := getMinNodes(sc)
+	minNodes := getMinimumNodes(sc)
```
jarrpa, line 222: Unnecessary function rename?

hack/functest.sh
```diff
@@ -13,3 +13,16 @@ if [ $? -ne 0 ]; then
...
+	--ocs-subscription-channel="${OCS_SUBSCRIPTION_CHANNEL}" \
+	--upgrade-from-ocs-registry-image="${UPGRADE_FROM_OCS_REGISTRY_IMAGE}" \
+	--upgrade-from-ocs-subscription-channel="${UPGRADE_FROM_OCS_SUBSCRIPTION_CHANNEL}" \
+	--ocs-cluster-uninstall="${OCS_CLUSTER_UNINSTALL}" \
+	--arbiter "$@"
+
+if [ $? -ne 0 ]; then
+	echo "ERROR: Arbiter mode functest failed."
+	exit 1
+fi
```
jarrpa, line 28: I don't want to do this, we are in effect doubling the length of the functional tests. We need to think about how we want to run this... for now I would just say to not add this.

jarrpa (changes requested):
In addition to the comments below, I see a number of places where there are unnecessary newlines in functions. It's a matter of taste, to be sure, but I don't like blank lines at the beginning and end of functions. If possible I'd like to remove the blank lines in the commits they are introduced, but if it's too much of a hassle I would accept a single formatting commit at the end of the PR.

umangachapagain (changes requested):
I'd like to see all the Rook/Ceph image and dependency changes at one place and compact. Probably a separate PR for it just like we did with NooBaa changes for KMS.

With these changes we're looking at some unsupported, custom image references (as the spec `AllowUnsupported: true` rightly says). It'd be safer to test this separately.

There's a CEPH_IMAGE env var in the CSV which didn't change (it should).
There's some unwanted changes in a commit that upgrades Rook.

Does it sound reasonable? @obnoxxx @jarrpa

pkg/controller/storagecluster/cephcluster.go
raghavendra-talur: Done

pkg/controller/storagecluster/cephcluster.go
raghavendra-talur: Done

pkg/controller/storagecluster/topology.go
raghavendra-talur: This was a function that was added in a different commit of the same PR. I have updated the commits to not have this function rename.

hack/functest.sh
raghavendra-talur: Done

raghavendra-talur:
> I'd like to see all the Rook/Ceph image and dependency changes at one place and compact. Probably a separate PR for it just like we did with NooBaa changes for KMS.

Makes sense. Created #964 
> 
> With these changes we're looking at some unsupported, custom image references (as the spec `AllowUnsupported: true` rightly says). It'd be safer to test this separately.

This was introduced for testing. I have removed it now.

> 
> There's a CEPH_IMAGE env var in the CSV which didn't change (it should).
> There's some unwanted changes in a commit that upgrades Rook.
This is also updated in #964 now.

> 
> Does it sound reasonable? @obnoxxx @jarrpa

I will rebase this PR once #964 is merged.

obnoxxx:
> > I'd like to see all the Rook/Ceph image and dependency changes at one place and compact. Probably a separate PR for it just like we did with NooBaa changes for KMS.
> 
> Makes sense. Created #964
> 
> > With these changes we're looking at some unsupported, custom image references (as the spec `AllowUnsupported: true` rightly says). It'd be safer to test this separately.
> 
> This was introduced for testing. I have removed it now.
> 
> > There's a CEPH_IMAGE env var in the CSV which didn't change (it should).
> > There's some unwanted changes in a commit that upgrades Rook.
> > This is also updated in #964 now.
> 
> > Does it sound reasonable? @obnoxxx @jarrpa

Sounds good, @umangachapagain !

> I will rebase this PR once #964 is merged.

You can already rebase now, if #964 merges as is, this won't need to be rebased again (merge will take care of it). @raghavendra-talur

jarrpa:
| /test ocs-operator-ci
-- | --

pkg/controller/storagecluster/cephblockpools.go
```diff
@@ -14,15 +14,24 @@ import (
...
+			Size:                     4,
+			TargetSizeRatio:          .49,
+			ReplicasPerFailureDomain: 2,
+		}
+	}
 	return cephv1.ReplicatedSpec{
 		Size:            3,
 		TargetSizeRatio: .49,
 	}
+
```
jarrpa, line 28: Remove

pkg/controller/storagecluster/cephblockpools.go
```diff
@@ -14,15 +14,24 @@ import (
...
 }
 
 // newCephBlockPoolInstances returns the cephBlockPool instances that should be created
 // on first run.
 func (r *ReconcileStorageCluster) newCephBlockPoolInstances(initData *ocsv1.StorageCluster) ([]*cephv1.CephBlockPool, error) {
+
```
jarrpa, line 34: Remove

pkg/controller/storagecluster/storagecluster_controller_test.go
```diff
@@ -3,9 +3,10 @@ package storagecluster
 import (
 	"context"
 	"fmt"
-	consolev1 "github.com/openshift/api/console/v1"
 	"testing"
 
+	consolev1 "github.com/openshift/api/console/v1"
+
```
jarrpa, line 9: Is this auto-formatting? Either way it should go with the other OCP imports below.

jarrpa (changes requested):
I think overall the code looks good, and the commits are clean. A few cosmetic changes are requested.

jarrpa (approved):
/lgtm

PR #943, opened 2020-12-07
Updating the annotation for newly enabled features

aruniiird:
The newly enabled features, like 'kms', 'arbiter' and 'flexible-scaling'
are added.


tools/csv-merger/csv-merger.go
```diff
@@ -657,7 +657,7 @@ The NooBaa operator deploys and manages the [NooBaa][2] Multi-Cloud Gateway on O
 	// Feature gating for Console. The array values are unique identifiers provided by the console.
 	// This can be used to enable/disable console support for any supported feature
 	// Example: "features.ocs.openshift.io/enabled": `["external", "foo1", "foo2", ...]`
-	ocsCSV.Annotations["features.ocs.openshift.io/enabled"] = `["external", "minimal_deployment", "encryption"]`
```
jarrpa, line 660: Why are these features being removed, are we not supporting them anymore? I would imagine we should be adding to this list, not replacing.

bipuladh: Because these features have already been introduced in OCS 4.6 we don't need to feature guard them for -1 versions, so we are replacing the list.

umangachapagain:
Needs backport to `release-4.7`

PR #864, opened 2020-10-23
add new property TuneFastDeviceClass

crombus:
Signed-off-by: crombus <pkundra@redhat.com>

PR #946, opened 2020-12-07
Refactor device tuning code

obnoxxx:
Reviewing PR #864 made me consider the naming of functions and variables and the structure of functions. I found it confusing that it talked about throttling internally and converted this to tuning toggles for rook, while ocs-operator is not doing any throttling. Furthermore, having two lists of fast and slow device storage classes, but having the the detection function hardcode which provisioner to apply each list against did not seem very intuitive and robust to me.

This PR refactors the code to be (IMHO) more obvious and more prone to future expansions of known devices.


jarrpa (approved):
/lgtm

PR #953, opened 2020-12-09
Refactor Platform_detection.go

crombus:

### DIFF
--- a/controllers/storagecluster/cephblockpools_test.go
+++ b/controllers/storagecluster/cephblockpools_test.go
@@ -28,11 +28,11 @@ func TestCephBlockPools(t *testing.T) {
 		cp := &Platform{platform: eachPlatform}
 		for _, c := range cases {
 			var objects []runtime.Object
-			if c.createRuntimeObjects {
-				objects = createUpdateRuntimeObjects(cp)
-			}
 			t, reconciler, cr, request := initStorageClusterResourceCreateUpdateTestWithPlatform(
 				t, cp, objects)
+			if c.createRuntimeObjects {
+				objects = createUpdateRuntimeObjects(t, cp, reconciler) //nolint:staticcheck //no need to use objects as they update in runtime
+			}
 			assertCephBlockPools(t, reconciler, cr, request)
 		}
 	}
--- a/controllers/storagecluster/cephobjectstores_test.go
+++ b/controllers/storagecluster/cephobjectstores_test.go
@@ -26,11 +26,11 @@ func TestCephObjectStores(t *testing.T) {
 		cp := &Platform{platform: eachPlatform}
 		for _, c := range cases {
 			var objects []runtime.Object
-			if c.createRuntimeObjects {
-				objects = createUpdateRuntimeObjects(cp)
-			}
 			t, reconciler, cr, request := initStorageClusterResourceCreateUpdateTestWithPlatform(
 				t, cp, objects)
+			if c.createRuntimeObjects {
+				objects = createUpdateRuntimeObjects(t, cp, reconciler) //nolint:staticcheck //no need to use objects as they update in runtime
+			}
 			assertCephObjectStores(t, reconciler, cr, request)
 		}
 	}
@@ -48,7 +48,9 @@ func assertCephObjectStores(t *testing.T, reconciler StorageClusterReconciler, c
 	err = reconciler.Client.Get(context.TODO(), request.NamespacedName, actualCos)
 	// for any cloud platform, 'cephobjectstore' should not be created
 	// 'Get' should have thrown an error
-	if avoidObjectStore(reconciler.platform.platform) {
+	avoid, avoidErr := reconciler.PlatformsShouldAvoidObjectStore()
+	assert.NoError(t, avoidErr)
+	if avoid {
 		assert.Error(t, err)
 	} else {
 		assert.NoError(t, err)
--- a/controllers/storagecluster/initialization_reconciler_test.go
+++ b/controllers/storagecluster/initialization_reconciler_test.go
@@ -50,7 +50,7 @@ func createStorageCluster(scName, failureDomainName string,
 	return cr
 }
 
-func createUpdateRuntimeObjects(cp *Platform) []runtime.Object {
+func createUpdateRuntimeObjects(t *testing.T, cp *Platform, r StorageClusterReconciler) []runtime.Object {
 	csfs := &storagev1.StorageClass{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: "ocsinit-cephfs",
@@ -73,42 +73,39 @@ func createUpdateRuntimeObjects(cp *Platform) []runtime.Object {
 	}
 	updateRTObjects := []runtime.Object{csfs, csrbd, cfs, cbp}
 
-	if !avoidObjectStore(cp.platform) {
+	avoid, err := r.PlatformsShouldAvoidObjectStore()
+	assert.NoError(t, err)
+	if !avoid {
 		csobc := &storagev1.StorageClass{
 			ObjectMeta: metav1.ObjectMeta{
 				Name: "ocsinit-ceph-rgw",
 			},
 		}
 		updateRTObjects = append(updateRTObjects, csobc)
-	}
 
-	// Create 'cephobjectstoreuser' only for non-aws platforms
-	if !avoidObjectStore(cp.platform) {
+		// Create 'cephobjectstoreuser' only for non-aws platforms
 		cosu := &cephv1.CephObjectStoreUser{
 			ObjectMeta: metav1.ObjectMeta{
 				Name: "ocsinit-cephobjectstoreuser",
 			},
 		}
 		updateRTObjects = append(updateRTObjects, cosu)
-	}
 
-	// Create 'cephobjectstore' only for non-cloud platforms
-	if !avoidObjectStore(cp.platform) {
+		// Create 'cephobjectstore' only for non-cloud platforms
 		cos := &cephv1.CephObjectStore{
 			ObjectMeta: metav1.ObjectMeta{
 				Name: "ocsinit-cephobjectstore",
 			},
 		}
 		updateRTObjects = append(updateRTObjects, cos)
-	}
-	// Create 'route' only for non-cloud platforms
-	if !avoidObjectStore(cp.platform) {
-		cos := &routev1.Route{
+
+		// Create 'route' only for non-cloud platforms
+		cosr := &routev1.Route{
 			ObjectMeta: metav1.ObjectMeta{
 				Name: "ocsinit-cephobjectstore",
 			},
 		}
-		updateRTObjects = append(updateRTObjects, cos)
+		updateRTObjects = append(updateRTObjects, cosr)
 	}
 	return updateRTObjects
 }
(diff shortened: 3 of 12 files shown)


pkg/controller/storagecluster/cephblockpools_test.go
```diff
@@ -28,11 +28,11 @@ func TestCephBlockPools(t *testing.T) {
...
 		for _, c := range cases {
 			var objects []runtime.Object
-			if c.createRuntimeObjects {
-				objects = createUpdateRuntimeObjects(cp)
-			}
 			t, reconciler, cr, request := initStorageClusterResourceCreateUpdateTestWithPlatform(
 				t, cp, objects)
+			if c.createRuntimeObjects {
+				objects = createUpdateRuntimeObjects(reconciler)
+			}
```
obnoxxx, line 35: why is this hunk moved?

pkg/controller/storagecluster/cephfilesystem_test.go
```diff
@@ -27,11 +27,11 @@ func TestCephFileSystem(t *testing.T) {
...
 		for _, c := range cases {
 			var objects []runtime.Object
-			if c.createRuntimeObjects {
-				objects = createUpdateRuntimeObjects(cp)
-			}
 			t, reconciler, cr, request := initStorageClusterResourceCreateUpdateTestWithPlatform(
 				t, cp, objects)
+			if c.createRuntimeObjects {
+				objects = createUpdateRuntimeObjects(reconciler)
+			}
```
obnoxxx, line 34: why is this hunk moved?

obnoxxx (commented):
Hi @crombus - thanks for working on the refactoring!
When we discussed it earlier today, I had imagined to approach it a bit different.

* Writing a wrapper around current getplatform and avoidobjectstore instead of changing current avoidobjectstore, we can keep the diff smaller and avoid having to change the test code immediately.
* If we let the wrapper hang off of `ReconcileStorageCluster`, we can avoid the extra argument.

Here are a few raw patches to illustrate what I meant: https://github.com/obnoxxx/ocs-operator/commits/refactor-avoid-object-store

When thinking about it initially, I had thought of just having a bool return, but we don't want to ignore the error, so the callers don't get simplified significantly... Therefore it is possibly not really worth the hassle. But I just hate having to repeat the same pattern call GetPlatform, then call avoidObjectStore in multiple locations, that's what triggered this, so it may not get much shorter, but at least less error-prone due to less repetition.

On top of such changes one could start using the wrapper in the tests as well, but at least it's not necessary to change all test code right away.

jarrpa (approved):
Overall this looks good. If you could do one more rebase, I'll LGTM it.

umangachapagain (commented):
I am trying to understand how this re-factor is helpful or better?

The way I see it, we still need the `platform` value in the caller for this newly added wrapper.
So why not just `GetPlatform()` first and then call `AvoidObjectStore(platform)` ?

This wrapper seems to hide the fact that we aren't actually avoiding ObjectStore but instead trying to decide if we should, based on the platform we are running on.
The naming makes it worse. When I call `AvoidObjectStore()` I expect it to avoid object store and tell me if it was or wasn't able to do so rather than telling me if I can do it and on what platform am I running.

/hold

jarrpa:
> I am trying to understand how this re-factor is helpful or better?
> 
> The way I see it, we still need the `platform` value in the caller for this newly added wrapper.
> So why not just `GetPlatform()` first and then call `AvoidObjectStore(platform)` ?

For the purposes of determining if we will deploy RGW, we have no need to know what the platform actually is. This method is way more streamlined. One function call to get both the boolean and platform (if applicable) makes more sense to me.

> This wrapper seems to hide the fact that we aren't actually avoiding ObjectStore but instead trying to decide if we should, based on the platform we are running on.
> The naming makes it worse. When I call `AvoidObjectStore()` I expect it to avoid object store and tell me if it was or wasn't able to do so rather than telling me if I can do it and on what platform am I running.

I don't know if this is a difference in dialect, but to me this renaming is much clearer and much more intuitive. Being more explicit would be awkward, all I can think of is something like `ShouldAvoidObjectStore()`, which I don't really like.

jarrpa:
@crombus Thank you for the updates! You're going to hate me for this, but I've had a change of heart. :pray: 

In reviewing a separate PR, I decided that it would make more sense to not return the platform with this new function. My reasoning is that since the value is *only* used for two log messages, we can just make the GetPlatform() call right before the log message. In addition, I saw that there are points in the unit tests where avoidObjectStore() is called that we should also probably update.

controllers/storagecluster/cephblockpools_test.go
```diff
@@ -28,11 +28,13 @@ func TestCephBlockPools(t *testing.T) {
...
-			if c.createRuntimeObjects {
-				objects = createUpdateRuntimeObjects(cp)
-			}
 			t, reconciler, cr, request := initStorageClusterResourceCreateUpdateTestWithPlatform(
 				t, cp, objects)
+			if c.createRuntimeObjects {
+				objects = createUpdateRuntimeObjects(cp, reconciler)
+				t, reconciler, cr, request = initStorageClusterResourceCreateUpdateTestWithPlatform(
+					t, cp, objects)
+			}
```
jarrpa, line 37: I'm not quite sure I understand the need for this change... how does it relate to the platform detection changes? Also, why are you running `initStorageClusterResourceCreateUpdateTestWithPlatform()` twice? That seems like a waste. Either you run it with or without the new objects. It seems you should just change the one line to add the new parameter.

controllers/storagecluster/cephobjectstores_test.go
```diff
@@ -48,7 +50,8 @@ func assertCephObjectStores(t *testing.T, reconciler StorageClusterReconciler, c
 	err = reconciler.Client.Get(context.TODO(), request.NamespacedName, actualCos)
 	// for any cloud platform, 'cephobjectstore' should not be created
 	// 'Get' should have thrown an error
-	if avoidObjectStore(reconciler.platform.platform) {
+	avoid, _ := reconciler.PlatformsShouldAvoidObjectStore()
```
jarrpa, line 53: You are still ignoring these errors. Don't. Check them and fail as appropriate.

controllers/storagecluster/initialization_reconciler_test.go
```diff
@@ -101,7 +102,7 @@ func createUpdateRuntimeObjects(cp *Platform) []runtime.Object {
 		updateRTObjects = append(updateRTObjects, cos)
 	}
 	// Create 'route' only for non-cloud platforms
-	if !avoidObjectStore(cp.platform) {
+	if !avoid {
```
jarrpa, line 105: Combine all these into one `if` stanza. They all check the same thing with no logic in between.

jarrpa (changes requested):
The more I look at this, the more I start to frown at the whole Platform implementation... but that refactor is a separate PR. For now I think you're on the right track within the limitations of what we have. Still a few comments to resolve.

umangachapagain:
> For the purposes of determining if we will deploy RGW, we have no need to know what the platform actually is. This method is way more streamlined. One function call to get both the boolean and platform (if applicable) makes more sense to me.
>

Knowing what the platform actually is is exactly what we need to determine if we deploy RGW. And if we need to know it, why not make it explicit?

I'd be ok with this refactor if we  weren't doing
```
if avoid {
		platform, err := r.platform.GetPlatform(r.Client)
		...
		r.Log.Info(fmt.Sprintf("not creating a CephObjectStore because the platform is '%s'", platform))
		...
	}
```
right after the new wrapper returns.

My suggestion: `PlatformsShouldAvoidObjectStore` returns Platform, Bool, Error. Use the returned Platform for logging instead of fetching it again. This also ensures that you log the `Platform` against which you did the `avoidObjectStore` check and not your own copy of it which might change or you might not be able to retrieve it?

jarrpa:
> I'd be ok with this refactor if we weren't doing
> 
> ```
> if avoid {
> 		platform, err := r.platform.GetPlatform(r.Client)
> 		...
> 		r.Log.Info(fmt.Sprintf("not creating a CephObjectStore because the platform is '%s'", platform))
> 		...
> 	}
> ```
> 
> right after the new wrapper returns.
> 
> My suggestion: `PlatformsShouldAvoidObjectStore` returns Platform, Bool, Error. Use the returned Platform for logging instead of fetching it again. This also ensures that you log the `Platform` against which you did the `avoidObjectStore` check and not your own copy of it which might change or you might not be able to retrieve it?

This is part of why I want to redo the entire platform detection implementation. We shouldn't be trying to both discover *and* retrieve the platform type in the same function. Even now we only ever check it once per instance of ocs-operator, we store it in the reconciler once we know it. IMO we should have `r.DetectPlatform()`, which runs once at the beginning, and `r.GetPlatform()` everywhere else. What's the likelihood that the platform OCP is running on changes *while the operator is still running*?

controllers/storagecluster/initialization_reconciler_test.go
```diff
@@ -72,42 +72,38 @@ func createUpdateRuntimeObjects(cp *Platform) []runtime.Object {
 	}
 	updateRTObjects := []runtime.Object{csfs, csrbd, cfs, cbp}
 
-	if !avoidObjectStore(cp.platform) {
+	avoid, _ := r.PlatformsShouldAvoidObjectStore()
```
jarrpa, line 75:
```suggestion
	avoid, avoidErr := reconciler.PlatformsShouldAvoidObjectStore()
	assert.NoError(t, avoidErr)
```

controllers/storagecluster/storageclasses_test.go
```diff
@@ -43,7 +43,8 @@ func assertStorageClasses(t *testing.T, reconciler StorageClusterReconciler, cr
 	err = reconciler.Client.Get(context.TODO(), request.NamespacedName, actualSc3)
 	// on a cloud platform, 'Get' should throw an error,
 	// as OBC StorageClass won't be created
-	if avoidObjectStore(reconciler.platform.platform) {
+	avoid, _ := reconciler.PlatformsShouldAvoidObjectStore()
```
jarrpa, line 46:
```suggestion
	avoid, avoidErr := reconciler.PlatformsShouldAvoidObjectStore()
	assert.NoError(t, avoidErr)
```

jarrpa (changes requested):
Two final things to change in the tests. We really should not be ignoring returned errors.

controllers/storagecluster/storageclasses_test.go
crombus: I might have missed these. updated now?

jarrpa (approved):
Thank you for your extended patience!

I'd like to at least mention the desire for two subsequent changes after this merges:
1. Refactor the while `GetPlatform()` thing so that we have an explicit `DetectPlatform()`/`SetPlatform()` function we call at the beginning of reconciliation. That way we can stop passing in a client we won't be using.
2. I thought of a slight optimization that would follow after this (maybe just as another commit in the same PR) that may make this look neater:
   ```
  	if avoid {
  		r.Log.Info(fmt.Sprintf("not creating a Ceph RGW route because the platform is '%s'", r.GetPlatform()))
  		return nil
  	}
   ```
   We should figure out a way to record this somewhere more official, but I'll leave that as an offline exercise.

/lgtm

controllers/storagecluster/storageclasses_test.go
jarrpa: yup!

umangachapagain (approved):

PR #954, opened 2020-12-10
[WIP] storageclasses: consolidate checks for skipping storage class creation

obnoxxx:
In principle, newStorageClasse() provides the in-memory storage
class structures, and createStorageClasses() creates them on the
cluster using the client.

Now we have some mechanisms to select which storage classes to
create and which ones to skip.  Up to now, this logic was scattered
across the two functions: The newStorageClasses() pre-selected whether
to add the OBC storage class under certain circumstances, and recently,
the configuration for reconcile strategies was added to the storage
cluster, and the filtering based on this is present in
createStorageClasses().

This patch consolidates the checks in the createStorageClasses function,
making the newStorageClasses function very simple again.

One small side effect in behavior is that previously, any possible error
from GetPlatform() was ignored, and we were *not* creating the OBC
storage class in this case. Now we correctly error out if GetPlatform
fails.

### DIFF
--- a/pkg/controller/storagecluster/storageclasses.go
+++ b/pkg/controller/storagecluster/storageclasses.go
@@ -44,6 +44,18 @@ func (r *ReconcileStorageCluster) createStorageClasses(scs []*storagev1.StorageC
 		if sc == nil {
 			continue
 		}
+
+		if index == cephObjectStoreIndex {
+			platform, err := r.platform.GetPlatform(r.client)
+			if err != nil {
+				return err
+			}
+			if avoidObjectStore(platform) {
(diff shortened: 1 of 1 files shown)


obnoxxx:
This will currently fail CI, since the tests need to be adapted:

https://github.com/openshift/ocs-operator/blob/master/pkg/controller/storagecluster/storageclasses_test.go#L40

The `assertStorageClasses` test is actually conceptually wrong, since it uses `newStorageClasses` and assumes there is no other filtering done in `createStorageClasses`. Etc. Will update.

jarrpa:
@obnoxxx Do you want to revisit this one?

PR #944, opened 2020-12-07
cephcluster: tune devices according to the deviceType

crombus:
Signed-off-by: crombus <pkundra@redhat.com>

PR #955, opened 2020-12-10
cephcluster: tune disks according to plaftorm

crombus:
set tunefastdeviceclass according to plaftorm
which are not correctly detected by ceph.
modify tc according to the changes.

### DIFF
--- a/controllers/storagecluster/cephcluster.go
+++ b/controllers/storagecluster/cephcluster.go
@@ -641,6 +641,14 @@ func (r *StorageClusterReconciler) checkTuneStorageDevices(ds ocsv1.StorageDevic
 		return dt.speed, nil
 	}
 
+	tuneFastDevices, err := r.DevicesDefaultToFastForThisPlatform()
+	if err != nil {
+		return diskSpeedUnknown, err
+	}
+	if tuneFastDevices {
+		return diskSpeedFast, nil
+	}
+
 	// not a known disk type, don't tune
 	return diskSpeedUnknown, nil
 }
--- a/controllers/storagecluster/platform_detection.go
+++ b/controllers/storagecluster/platform_detection.go
@@ -18,6 +18,13 @@ var AvoidObjectStorePlatforms = []configv1.PlatformType{
 	configv1.AzurePlatformType,
 }
 
+// TuneFastPlatforms is a list of all PlatformTypes where TuneFastDeviceClass has to be set True.
+var TuneFastPlatforms = []configv1.PlatformType{
+	configv1.OvirtPlatformType,
+	configv1.IBMCloudPlatformType,
+	configv1.AzurePlatformType,
+}
+
 // Platform is used to get the CloudPlatformType of the running cluster in a thread-safe manner
 type Platform struct {
 	platform configv1.PlatformType
@@ -54,4 +61,20 @@ func avoidObjectStore(p configv1.PlatformType) bool {
 		}
 	}
 	return false
+}
+
+func (r *StorageClusterReconciler) DevicesDefaultToFastForThisPlatform() (bool, error) {
+	c := r.Client
+	platform, err := r.platform.GetPlatform(c)
+	if err != nil {
+		return false, err
+	}
+
+	for _, tfplatform := range TuneFastPlatforms {
+		if platform == tfplatform {
+			return true, nil
+		}
+	}
+
+	return false, nil
 }
--- a/controllers/storagecluster/storagecluster_controller_test.go
+++ b/controllers/storagecluster/storagecluster_controller_test.go
@@ -264,6 +264,7 @@ func TestThrottleStorageDevices(t *testing.T) {
 		deviceSets     []api.StorageDeviceSet
 		storageCluster *api.StorageCluster
 		expectedSpeed  diskSpeed
+		platform       *Platform
 	}{
 		{
 			label: "Case 1", // storageclass is gp2 or io1
@@ -424,10 +425,40 @@ func TestThrottleStorageDevices(t *testing.T) {
 			storageCluster: &api.StorageCluster{},
 			expectedSpeed:  diskSpeedFast,
 		},
+		{
+			label: "Case 7", // storageclass is neither gp2 nor io1 but platform is Azure
+			storageClass: &storagev1.StorageClass{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "st1",
+				},
+				Provisioner: string(EBS),
+				Parameters: map[string]string{
+					"type": "st1",
+				},
+			},
+			deviceSets: []api.StorageDeviceSet{
+				{
+					Name:  "mock-sds",
+					Count: 3,
+					DataPVCTemplate: corev1.PersistentVolumeClaim{
+						Spec: corev1.PersistentVolumeClaimSpec{
+							StorageClassName: &fakestorageClassName,
+						},
+					},
+					Portable: true,
+				},
+			},
+			platform:       &Platform{platform: configv1.AzurePlatformType},
+			storageCluster: &api.StorageCluster{},
+			expectedSpeed:  diskSpeedFast,
+		},
 	}
 
 	for _, tc := range testcases {
 		reconciler := createFakeStorageClusterReconciler(t, tc.storageCluster, tc.storageClass)
+		if tc.platform != nil {
+			reconciler.platform = tc.platform
+		}
 		for _, ds := range tc.deviceSets {
 			actualSpeed, err := reconciler.checkTuneStorageDevices(ds)
 			assert.NoError(t, err)
--- a/controllers/storagecluster/storagecluster_controller_test.go
+++ b/controllers/storagecluster/storagecluster_controller_test.go
@@ -852,7 +852,7 @@ func createFakeStorageClusterReconciler(t *testing.T, obj ...runtime.Object) Sto
 		Scheme:        scheme,
 		serverVersion: &k8sVersion.Info{},
 		Log:           logf.Log.WithName("controller_storagecluster_test"),
-		platform:      &Platform{},
+		platform:      &Platform{platform: configv1.NonePlatformType},
 	}
 }
 


pkg/controller/storagecluster/cephcluster.go
```diff
@@ -50,6 +50,8 @@ var knownDiskTypes = []knownDiskType{
 	{diskSpeedFast, AzureDisk, "managed-premium"},
 }
 
+// AvoidObjectStorePlatforms is a list of all PlatformTypes where CephObjectStores will not be deployed.
```
obnoxxx, line 53: Is this line added by accident?

pkg/controller/storagecluster/platform_detection.go
```diff
@@ -55,3 +61,18 @@ func avoidObjectStore(p configv1.PlatformType) bool {
 	}
 	return false
 }
+
+func (p *Platform) TunePlatformForFastDevices(c client.Client) (bool, error) {
```
obnoxxx, line 65: can you do 
```
func (r *ReconcileStorageCluster) TunePlatformForFastDevices() (bool, error) {
  platform := r.platform
  client := r.client
  ...
}
```

Makes it easier on the caller, Imho.

pkg/controller/storagecluster/platform_detection.go
```diff
@@ -18,6 +18,12 @@ var AvoidObjectStorePlatforms = []configv1.PlatformType{
 	configv1.AzurePlatformType,
 }
 
+var TuneFastPlatforms = []configv1.PlatformType{
```
obnoxxx, line 21: I think you may have wanted to add a comment here instead of in the other file?

obnoxxx (changes requested):
As discussed before, looks good in general, but some requests inside.

pkg/controller/storagecluster/cephcluster.go
crombus: yes has to be added in the file

pkg/controller/storagecluster/platform_detection.go
crombus: done.

pkg/controller/storagecluster/platform_detection.go
crombus: yes

pkg/controller/storagecluster/storagecluster_controller_test.go
```diff
@@ -304,11 +304,12 @@ func TestThrottleStorageDevices(t *testing.T) {
 							StorageClassName: &fakestorageClassName,
 						},
 					},
-					Portable: true,
+					Portable:   true,
+					DeviceType: "hdd",
```
obnoxxx, line 308: This seems to be a change to the test case that is not related to this code change. 🤔 

It explicitly configures the type "hdd" in the device set and correctly changes the expected result to Slow. But isn't that the test for the previous PR #944 ?

obnoxxx (changes requested):
Code changes look good now, thanks!

There is one change in the test code that seems to not really belong into this commit. (see inline) - it could be a separate patch to test the effects of PR #944.

And shouldn't we have a test case for this new behaviour?

pkg/controller/storagecluster/storagecluster_controller_test.go
crombus: to solve this error `	Error Trace:	storagecluster_controller_test.go:426
        	Error:      	Received unexpected error:
        	            	could not get infrastructure details to determine cloud platform: infrastructures.config.openshift.io "cluster" not found
`

crombus: and yes for this we should have modify the creation of cluster with platform adding new test cases for this addidtion.

obnoxxx: @crombus - hmmm - wouldn't it be better / more natural to fix this problem by changing the cluster structure in the test so that the platform detection can not fail? It does not seem to be conceptually correct to fix it by avoiding the call in the first place by adding a configuration we know will be checked first... 

Can the platform detection fail in the real world? (not mocked-up test code)

pkg/controller/storagecluster/cephcluster.go
```diff
@@ -550,6 +550,14 @@ func (r *ReconcileStorageCluster) checkTuneStorageDevices(ds ocsv1.StorageDevice
 		return dt.speed, nil
 	}
 
+	ret, err := r.TunePlatformForFastDevices()
```
umangachapagain, line 553: Probably a more simpler implementation would be to check if `TuneFastPlatforms` contains `r.platform.GetPlatform()`.
That way I wouldn't have to jump to definition of `TunePlatformForFastDevices` to understand what is happening and why we are tuning.

Currently it seems like `TunePlatformForFastDevices` is doing tuning intensive stuff while in reality it's just checking if `TuneFastPlatforms` contains `r.platform.GetPlatform()`.

As a follow-up I'd look into re-factoring platform detection and maybe storing it in ReconcileStorageCluster if it makes more sense.

pkg/controller/storagecluster/storagecluster_controller_test.go
umangachapagain, line 308: I'd like to see cases of different possible values of DeviceType covered. Also, a case for invalid DeviceType.

umangachapagain (changes requested):

pkg/controller/storagecluster/storagecluster_controller_test.go
crombus: yes, it can fail. And cration of cluster with platform is the correct fix

pkg/controller/storagecluster/cephcluster.go
crombus: @obnoxxx ^^

crombus: For me I feel both ways are good and maybe if adding comment solve this problem that is also a way to go.

pkg/controller/storagecluster/storagecluster_controller_test.go
obnoxxx: Right, but this is not the topic of this PR. That was introduced in a different PR. So we can (and should) enhance that in a separate PR.

pkg/controller/storagecluster/platform_detection.go
umangachapagain: I'd argue that we keep the old way.

It seems like we made it easier on the caller but we actually made it difficult.
While with the old implementation caller only required a `*Platform` and a `client.Client`, with the suggested change caller would now need to create a concrete `*ReconcileStorageCluster` even though it isn't necessary.

It is easy to ignore this as we seem to have a `*ReconcileStorageCluster` always available but we shouldn't rely on that.
Also, since this tuning is a behavior that depends on platform rather than the reconciler, it should really be a method on Platform.

pkg/controller/storagecluster/cephcluster.go
umangachapagain: "If your code needs a comment to explain what it does, it isn't really a readable/understandable code". 
Read this somewhere and I relate.

We could inline the call, rename the method, use better variables etc to make it better. Adding a comment would be the last thing I'd suggest.

pkg/controller/storagecluster/storagecluster_controller_test.go
umangachapagain: If this is not the topic of this PR, why does it add `DeviceType: "hdd"` to test case?
Honestly, I've lost track of "Tune disks" PRs and enhancements :sweat_smile:

pkg/controller/storagecluster/cephcluster.go
obnoxxx: > Probably a more simpler implementation would be to check if `TuneFastPlatforms` contains `r.platform.GetPlatform()`.

Well, that is exactly what the function does!?

> That way I wouldn't have to jump to definition of `TunePlatformForFastDevices` to understand what is happening and why we are tuning.

That's more a naming thing than an implementation thing. I explicitly asked @crombus to write such a function to *not* have the implementation bleed into the caller. I didn't want to repeat the (imho) bad approach of `aviodObjectStore()` where each caller first has to go and call `GetPlatform` and then insert it into `avoidObjectStore`. This is bad encapsulation/abstraction, and error prone. If there is more than one caller, it's also code duplication.

So it boils down to: we only need to name the function / vars better to make it more obvious what is going on.

> Currently it seems like `TunePlatformForFastDevices` is doing tuning intensive stuff while in reality it's just checking if `TuneFastPlatforms` contains `r.platform.GetPlatform()`.

Right, there needs to be a better name. I like it when in some languages, like ruby, functions can end with a '?' symbol. That makes it more obvious. But here we don't have that option. We could do something like this:

```
tuneFastDevices, err := r.DevicesDefaultToFastForThisPlatform()
...
```

or 

```
tuneDevicesFast, err := DoTuneDevicesFastForPlatform()
...
```
or similar...

> As a follow-up I'd look into re-factoring platform detection and maybe storing it in ReconcileStorageCluster if it makes more sense.

Possible. I have started a discussion with @crombus already about possible refactorings of the platform-detection code triggered by this change. Storing data sounds like an optimization. But let's get the APIs good in the first place. And let's not add new stuff that needs immediate refactoring afterwards. And to reiterate my initial point: I think the current approach is the right encapsulation and just needs better naming.

pkg/controller/storagecluster/platform_detection.go
obnoxxx: We can have an internal method on platform.
But I request the convenience wrapper on the reconciler, since it has all we need, and it avoids code duplication (like with `avoidObjectStore`) when there might be multiple callers.
This pattern of GetPlatform and then invoke the other method is cumbersome and error-prone imho.

obnoxxx: in the end, I won't insist too much if folks are more inline with the original approach. I think both approaches have pros and cons. ;-)

pkg/controller/storagecluster/storagecluster_controller_test.go
obnoxxx: > If this is not the topic of this PR, why does it add `DeviceType: "hdd"` to test case?

This is exactly what I told @crombus: Using this to fix the test failure is wrong. It is just avoiding the code path.
Instead we need to make the fake cluster so that it has platform info, so the new code does not error out.

> Honestly, I've lost track of "Tune disks" PRs and enhancements sweat_smile

:wink:

pkg/controller/storagecluster/cephcluster.go
obnoxxx: @crombus - can you rename the function and variable along the above lines to make the code more obvious?

obnoxxx (changes requested):

jarrpa:
@crombus Hey, can we revisit this?

crombus:
All requests have been addressed except from changing the test-case accordingly .

I get this error
```gp2". multiple group-version-kinds associated with type *v1.StorageClass, refusing to guess at one```

for changes in tc
 ```for _, tc := range testcases {
                reconciler := createFakeInitializationStorageClusterReconcilerWithPlatform(t, tc.platform, tc.storageCluster, tc.storageClass)
                for _, ds := range tc.deviceSets {
                        actualSpeed, err := reconciler.checkTuneStorageDevices(ds)
                        assert.NoError(t, err)
                        assert.Equalf(t, tc.expectedSpeed, actualSpeed, "[%q]: failed to get expected output", tc.label)
                }
        }```

obnoxxx:
Updated the tc fix (by adding a default platform entry to the reconciler) and added a specific tc for the new code path after discussing with @crombus .

let's see what the test runs bring

obnoxxx (approved):
This looks good to me now.

/lgtm

PR #956, opened 2020-12-10
Add configurable number of CephObjectStore gateway instances

egafford:
In many use cases, 1 is the most sensible number of RGW instances
to create in a StorageCluster, but it is necessary to allow the
user to create more. This change makes this value configurable
with a default of 1.

### DIFF
--- a/pkg/apis/ocs/v1/storagecluster_types.go
+++ b/pkg/apis/ocs/v1/storagecluster_types.go
@@ -78,6 +78,7 @@ type ManageCephFilesystems struct {
 type ManageCephObjectStores struct {
 	ReconcileStrategy   string `json:"reconcileStrategy,omitempty"`
 	DisableStorageClass bool   `json:"disableStorageClass,omitempty"`
+	GatewayInstances    int32  `json:"gatewayInstances,omitempty"`
 }
 
 // ManageCephObjectStoreUsers defines how to reconcile CephObjectStoreUsers
--- a/pkg/controller/storagecluster/cephobjectstores.go
+++ b/pkg/controller/storagecluster/cephobjectstores.go
@@ -83,6 +83,10 @@ func (r *ReconcileStorageCluster) createCephObjectStores(cephObjectStores []*cep
 // newCephObjectStoreInstances returns the cephObjectStore instances that should be created
 // on first run.
 func (r *ReconcileStorageCluster) newCephObjectStoreInstances(initData *ocsv1.StorageCluster, reqLogger logr.Logger) ([]*cephv1.CephObjectStore, error) {
+	gatewayInstances := initData.Spec.ManagedResources.CephObjectStores.GatewayInstances
+	if gatewayInstances == 0 {
+		gatewayInstances = int32(defaults.CephObjectStoreGatewayInstances)
+	}
 	ret := []*cephv1.CephObjectStore{
 		{
 			ObjectMeta: metav1.ObjectMeta{
@@ -106,7 +110,7 @@ func (r *ReconcileStorageCluster) newCephObjectStoreInstances(initData *ocsv1.St
 				},
 				Gateway: cephv1.GatewaySpec{
 					Port:      80,
-					Instances: 2,
+					Instances: gatewayInstances,
 					Placement: getPlacement(initData, "rgw"),
 					Resources: defaults.GetDaemonResources("rgw", initData.Spec.Resources),
 				},
--- a/deploy/crds/ocs.openshift.io_storageclusters_crd.yaml
+++ b/deploy/crds/ocs.openshift.io_storageclusters_crd.yaml
@@ -169,6 +169,9 @@ spec:
                     properties:
                       disableStorageClass:
                         type: boolean
+                      gatewayInstances:
+                        type: integer
+                        format: int32
                       reconcileStrategy:
                         type: string
               monDataDirHostPath:
--- a/deploy/csv-templates/crds/ocs/ocs.openshift.io_storageclusters_crd.yaml
+++ b/deploy/csv-templates/crds/ocs/ocs.openshift.io_storageclusters_crd.yaml
@@ -169,6 +169,9 @@ spec:
                     properties:
                       disableStorageClass:
                         type: boolean
+                      gatewayInstances:
+                        type: integer
+                        format: int32
                       reconcileStrategy:
                         type: string
               monDataDirHostPath:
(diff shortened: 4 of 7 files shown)


egafford:
Reasonably straightforward. I have mixed feelings about the variable name: ManageCephObjectStoreGateway seems like rather a lot, but ManageGateway seems a bit nonspecifc. Very open to thoughts on this point.

deploy/csv-templates/crds/ocs.openshift.io_storageclusters_crd.yaml
```diff
@@ -3562,7 +3570,7 @@ spec:
                   to distribute its data replicas for the default CephBlockPool
                 type: string
               images:
-                description: Images holds the image reconclie status for all images
+                description: Images holds the image reconcile status for all images
```
obnoxxx, line 3573: I am wondering why this wasn't covered in earlier commits? Did we forget to run csv generation and commit? 🤔

obnoxxx: can we have this part as a separate commit?

pkg/controller/storagecluster/cephobjectstores.go
```diff
@@ -83,6 +83,10 @@ func (r *ReconcileStorageCluster) createCephObjectStores(cephObjectStores []*cep
 // newCephObjectStoreInstances returns the cephObjectStore instances that should be created
 // on first run.
 func (r *ReconcileStorageCluster) newCephObjectStoreInstances(initData *ocsv1.StorageCluster, reqLogger logr.Logger) ([]*cephv1.CephObjectStore, error) {
+	gatewayInstances := int32(defaults.CephObjectStoreGatewayInstances)
+	if initData.Spec.ManagedResources.CephObjectStores.ManageGateway.GatewayInstances != 0 {
+		gatewayInstances = 1
```
obnoxxx, line 88: I think it should be:

```
gatewayInstances = initData.Spec.ManagedResources.CephObjectStores.ManageGateway.GatewayInstances
```

I think non-default value (>1) should be covered by an additional test case. Currently  it only covers `1`, so it wasn't caught.

pkg/apis/ocs/v1/storagecluster_types.go
```diff
@@ -70,8 +70,14 @@ type ManageCephFilesystems struct {
 
 // ManageCephObjectStores defines how to reconcile CephObjectStores
 type ManageCephObjectStores struct {
-	ReconcileStrategy   string `json:"reconcileStrategy,omitempty"`
-	DisableStorageClass bool   `json:"disableStorageClass,omitempty"`
+	ReconcileStrategy   string        `json:"reconcileStrategy,omitempty"`
+	DisableStorageClass bool          `json:"disableStorageClass,omitempty"`
+	ManageGateway       ManageGateway `json:"manageGateway,omitempty"`
```
obnoxxx, line 75: Interesting thought to add the substructure for `ManageGateway` - do you anticipate other settings for Gateway? Otherwise directly having `GatewayInstances`inside the `ManageObjectStore` struct would also have been sufficient. @jarrpa - what do you think?

obnoxxx (changes requested):
Great, thanks for raising this, @egafford !
A few comments inline.

@oritwas - FYI!

pkg/apis/ocs/v1/storagecluster_types.go
egafford: Hi @obnoxxx, I agree that we could simply put all ObjectStore config settings directly into the ManageObjectStore struct. It was my thought to follow the form of the rook cephv1 struct, to build familiarity and ease of use for users of both projects. Still, this is not necessary at all. I'm happy to change this if we think it's better to just flatten these management structs. It would certainly simplify the interface, and whatever we do here, we should likely follow form later, so we should certainly get it right.

deploy/csv-templates/crds/ocs.openshift.io_storageclusters_crd.yaml
egafford: Ah, odd. And yes, I'll separate that from the main commit.

pkg/apis/ocs/v1/storagecluster_types.go
jarrpa: I'm in favor of having it directly in the ManageCephObjectStores struct. We want to avoid mimicking the underlying APIs where possible, in favor of a cleaner API informed by our support statements. My understanding is that there is only ever one CephObjectStore per CephCluster, as well.

deploy/csv-templates/crds/ocs.openshift.io_storageclusters_crd.yaml
obnoxxx: This has been merged as #958 meanwhile.

pkg/apis/ocs/v1/storagecluster_types.go
obnoxxx: I am also inclining to omitting the extra layer.

I don't fully understand how the 1:1 correspondence would have an impact on this though:
Instead of having a sub-structure for the gateway options, we would need to have a list of gateway options (or a list of substructures). Or do you mean that a list of substructures would be more natural and easy for multiple object stores?

pkg/controller/storagecluster/cephobjectstores.go
obnoxxx: @egafford  -  I am misunderstanding this?

pkg/apis/ocs/v1/storagecluster_types.go
egafford: I'm good with just keeping this in a flat struct. I was close to my indifference point on this anyway.

pkg/controller/storagecluster/cephobjectstores.go
obnoxxx: this has been taken care of by the latest update

pkg/apis/ocs/v1/storagecluster_types.go
obnoxxx: this has been taken care of

obnoxxx (changes requested):
This PR should be rebased on current master (to get rid of the `reconclie` typo fix, which is already in master). (It would not be strictly necessary, but I prefer a clean rebase instead of a merge taking care of this.) Other than that - looks good to me!

obnoxxx (approved):
looks good to me, thanks!

jarrpa (approved):
/lgtm

PR #964, opened 2020-12-17
update rook to include arbiter/stretch cluster changes

raghavendra-talur:
This PR updates the rook image and go module version to include arbiter/stretch cluster changes.


umangachapagain (approved):

obnoxxx (approved):
/lgtm

raghavendra-talur:
Looking into the `ocs-operator-bundle-e2e-aws` failure. The deployment is failing. Checking if I failed to push any change.

raghavendra-talur:
@jarrpa @obnoxxx @umangachapagain The deployment is failing because ceph v16 isn't supported in rook yet and the CephClusterSpec we build in the ocs-operator always has `AllowUnsupported` set to `false`.

I can think of three ways of fixing this problem:
1. Introduce a field in the StorageClusterSpec to let the users enable `AllowUnsupported` and it becomes part of our API.
2. Use an annotation but don't advertise it. However, it will still be part of the build and users may potentially use it.
3. Use a go build tag and only set it to true when go build is run with that tag.

Thoughts?

raghavendra-talur:
> @jarrpa @obnoxxx @umangachapagain The deployment is failing because ceph v16 isn't supported in rook yet and the CephClusterSpec we build in the ocs-operator always has `AllowUnsupported` set to `false`.
> 
> I can think of three ways of fixing this problem:
> 
> 1. Introduce a field in the StorageClusterSpec to let the users enable `AllowUnsupported` and it becomes part of our API.
> 2. Use an annotation but don't advertise it. However, it will still be part of the build and users may potentially use it.
> 3. Use a go build tag and only set it to true when go build is run with that tag.
> 
> Thoughts?

There is another option of ENV which can override this setting. This is very similar to the annotation option ; we don't advertise but users can set if they know the option name.

umangachapagain:
I don't like the idea of adding configuration options to allow something unsupported.

Since this is for CI only, can we explicitly patch the CephCluster CR in the functest?
This is just a temporary fix and we need to revert the moment we have a supported release, instead of adding a feature to work around unsupported configurations.

raghavendra-talur:
> I don't like the idea of adding configuration options to allow something unsupported.

Agreed.

> 
> Since this is for CI only, can we explicitly patch the CephCluster CR in the functest?
> This is just a temporary fix and we need to revert the moment we have a supported release, instead of adding a feature to work around unsupported configurations.

I considered this but editing cephCluster is equal to messing with device under test. We should not do it.

I am currently testing a LDFLAG approach where I change the value for it at build time. We can revert when we have the right image. What is your opinion about LDFLAG approach?

obnoxxx:
Having something made available only with a specific build flag/tag that we set for a testing/ci build sounds like the best approach to me.

obnoxxx:
> Having something made available only with a specific build flag/tag that we set for a testing/ci build sounds like the best approach to me.

However, we should consider that we might later want to run our e2e test suite as part of downstream ocs-ci tests...

raghavendra-talur:
@obnoxxx @umangachapagain Added commits that use the ldflag approach. Please take a look.

jarrpa (approved):
/lgtm

PR #969, opened 2020-12-18
Removes the OSD IDs from the osd removal job name

xenolinux:
When triggering the job for multiple OSDs removal, the template includes comma along with OSD IDs in the job name. This will cause the template to fail.
This commit will give the osd removal job constant and valid name.

### DIFF
--- a/controllers/storagecluster/reconcile.go
+++ b/controllers/storagecluster/reconcile.go
@@ -631,7 +631,7 @@ func newCleanupJob(sc *ocsv1.StorageCluster) *batchv1.Job {
 			APIVersion: "batch/v1",
 		},
 		ObjectMeta: metav1.ObjectMeta{
-			Name:        "ocs-osd-removal-${FAILED_OSD_IDS}",
+			Name:        "ocs-osd-removal-job",
 			Namespace:   sc.Namespace,
 			Labels:      labels,
 			Annotations: annotations,


pkg/controller/storagecluster/reconcile.go
```diff
@@ -602,7 +602,7 @@ func newCleanupJob(sc *ocsv1.StorageCluster) *batchv1.Job {
 			APIVersion: "batch/v1",
 		},
 		ObjectMeta: metav1.ObjectMeta{
-			Name:        "ocs-osd-removal-${FAILED_OSD_IDS}",
```
egafford, line 605: As a question rather than a request, the information regarding failed OSD IDs does seem useful, and it sounds as though it is only commas that are causing issue in the job name. Would it be possible and useful to replace the commas with a character that would not cause the template to fail (such as `-`)?

xenolinux: Yes. Commas are the issue since the template is including them in the job name. It is possible to replace a comma with a hyphen here https://github.com/rook/rook/blob/master/cmd/rook/ceph/osd.go#L269. That way OSD IDs can stay in the job name. Just to add if we have a constant job name, we also can check what OSD ID the job has been run for in the job logs.

travisn: To be clear, we are saying there are two ways we could approach this: 
1. Use a constant name for the job and keep the comma-separated IDs
2. Require multiple OSDs to be specified with a `-` separator instead of a comma, then change Rook for the new parsing. The `-` would have to be specified by the user since the `${FAILED_OSD_IDS}` variable must stay in the template, to be evaluated at runtime. Or perhaps we could have to add some bash replacement to the documentation, but that's messy IMO.

I vote for the first option. It keeps things simple, and more natural for the user by keeping the comma separator. Option two seems unnatural for the user to specify `-` as a list separator.

iamniting:
I think we should divide this commit info into multiple lines.
Screenshot from 2021-01-13 12-32-10

agarwal-mudit:
@travisn PTAL

travisn (approved):
LGTM

obnoxxx (approved):
/lgtm

obnoxxx:
What's going on with this PR? Why is there no tide on it?...

PR #973, opened 2020-12-25
Source Manifest uses template instead of OCS_IMAGE for backward compatibility

umangachapagain:
Before migrating to new project structure and tools, CSV Merger was using templates added by source-manifests to inject OCS_IMAGE value. When we moved to using Kustomize, we set the images directly instead of first templating and replacing.
This broke the build pipeline as it didn't safeguard itself against uninitialized OCS_IMAGE var.

This PR reverts back to using templates instead of Kustomize to unblock the build pipeline.
Alternatively, build pipeline can set OCS_IMAGE env var a little early and all should be well (without this PR)

### DIFF
--- a/hack/source-manifests.sh
+++ b/hack/source-manifests.sh
@@ -84,11 +84,14 @@ function gen_ocs_csv() {
 	gen_args="generate kustomize manifests -q"
 	# shellcheck disable=SC2086
 	$OPERATOR_SDK $gen_args
-	pushd config/manager
-	$KUSTOMIZE edit set image ocs-dev/ocs-operator="$OCS_IMAGE"
-	popd
 	$KUSTOMIZE build config/manifests | $OPERATOR_SDK generate bundle -q --overwrite=false --version "$CSV_VERSION"
 	mv "$GOPATH"/src/github.com/openshift/ocs-operator/bundle/manifests/*clusterserviceversion.yaml $OCS_CSV
+	# Make variables templated for csv-merger tool
+	if [ "$OS_TYPE" == "Darwin" ]; then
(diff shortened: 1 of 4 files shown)


hack/source-manifests.sh
```diff
@@ -84,12 +84,15 @@ function gen_ocs_csv() {
...
 	$KUSTOMIZE build config/manifests | $OPERATOR_SDK generate bundle -q --overwrite=false --version "$CSV_VERSION"
 	mv "$GOPATH"/src/github.com/openshift/ocs-operator/bundle/manifests/*clusterserviceversion.yaml $OCS_CSV
-	cp config/crd/bases/* $ocs_crds_outdir
+	# Make variables templated for csv-merger tool
+	if [ "$OS_TYPE" == "Darwin" ]; then
+		sed -i '' "s/REPLACE_IMAGE/{{.OcsOperatorImage}}/g" $OCS_CSV
+	else
+		sed -i "s/REPLACE_IMAGE/{{.OcsOperatorImage}}/g" $OCS_CSV
+	fi
+		cp config/crd/bases/* $ocs_crds_outdir
```
obnoxxx, line 95: unintended indentation?

```suggestion
	cp config/crd/bases/* $ocs_crds_outdir
```

obnoxxx (changes requested):

hack/source-manifests.sh
umangachapagain: yes, fixed.

umangachapagain:
Build pipeline is fixed. Closing this PR as we decided to use Kustomize instead of Go Templates for simplicity.

PR #972, opened 2020-12-24
hack: fix OCS Operator image in source-manifests.sh

umangachapagain:
Signed-off-by: Umanga Chapagain <chapagainumanga@gmail.com>

PR #974, opened 2020-12-25
[release-4.7] Source Manifest uses template instead of OCS_IMAGE for backward compatibility

umangachapagain:
Backport of #973 

#972  is not required as this makes it obsolete.


umangachapagain:
/hold
wait for #973  to merge

PR #981, opened 2021-01-07
update readme

umangachapagain:


obnoxxx (approved):
Thanks for the overhaul!
At some point, if it keeps growing, we might want to consider splitting this into multiple docs, but that can be discussed later..
We should also probably make our minds up w.r.t american versus british spelling (e.g labeling vs labelling ...) :smiley: 

/lgtm

PR #983, opened 2021-01-08
Remove usage of outdated coreos/prometheus-operator

umangachapagain:
replacing coreos/prometheus-operator with prometheus/prometheus-operator
which has the latest APIs.


obnoxxx (approved):
/lgtm

PR #985, opened 2021-01-11
Non default sc unit tests

egafford:
Enable non-default StorageClusters for unit testing

Prior to this change, creation of a default StorageCluster was
tightly coupled with test initialization functions. As such, many
of our unit tests only cover the default happy path. This change
introduces the option of providing a non-default StorageCluster
for test initialization while leaving the signatures of existing
initialization functions unchanged. It also uses this framework
for the testing of ObjectStore gateway instances, as this was
the case that inspired the change as a whole.
    
### DIFF
--- a/controllers/storagecluster/cephobjectstores_test.go
+++ b/controllers/storagecluster/cephobjectstores_test.go
@@ -2,6 +2,7 @@ package storagecluster
 
 import (
 	"context"
+	"fmt"
 	"testing"
 
 	api "github.com/openshift/ocs-operator/api/v1"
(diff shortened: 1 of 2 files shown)


controllers/storagecluster/cephobjectstores_test.go
```diff
@@ -12,14 +13,23 @@ import (
...
+	var cases = []CephObjectStoreTestCase{
+		{
+			label:                "case 1", // Ensure that cephObjectStores are created on non-cloud Platform with default values
+			createRuntimeObjects: false,
+			gatewayInstances:     1,
+		},
 		{
-			label:                "case 1", // Ensure that cephObjectStores are created on non-cloud Platform
+			label:                "case 2", // Ensure that cephObjectStores are created on non-cloud Platform with configured values
 			createRuntimeObjects: false,
```
iamniting, line 31: I see that `createRuntimeObjects` is false in both cases. If that's the case can not we get rid of this value?

egafford: I thought that was something of a curiosity myself. It seems to be standard across the Ceph storage interface test case structs, though, and there is some value in consistency for ease of modification of all three tests when it comes up. I certainly don't have strong feelings on this either way, though it's unrelated to the current change set.

controllers/storagecluster/cephobjectstores_test.go
```diff
@@ -12,14 +13,23 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
+type CephObjectStoreTestCase struct {
+	label                string
+	createRuntimeObjects bool
+	gatewayInstances     int32
+}
```
jarrpa, line 20: I generally like keeping the case struct as defined inline. I don't see much of a benefit to giving them their own unique name. If anything, I'd advocate for keeping it here just for consistency with the other tests.

jarrpa (changes requested):
Sorry for the delay, this somehow slipped my attention.

Overall the change makes sense, I just have one formatting change request. After that and a rebase, it'd be good to go.

PR #988, opened 2021-01-12
Must-gather: Update OWNERS file

crombus:
Add rajatsing as reviwers and
remove ashishranjan738.

### DIFF
--- a/must-gather/OWNERS
+++ b/must-gather/OWNERS
@@ -1,6 +1,6 @@
 reviewers:
-  - ashishranjan738
   - crombus
+  - rajatsing
 approvers:
-  - ashishranjan738
   - crombus
+  - rajatsing


obnoxxx (commented):
What's the purpose of only adding @rajatsing to the reviewers list but not approvers?

Acdording to the comments in the man OWNERS file  https://github.com/openshift/ocs-operator/blob/master/OWNERS#L1, "approvers" can add the "approved" label and "reviewers" can add the "lgtm" label. We usually add folks to both and don't make the distinction.

obnoxxx:
@crombus  - What I was saying is that you may be misinterpreting the effect of putting someone into the "reviewers" list: If you put him in the reviewers list, then his "/lgtm" will be authoritative and will lead to get patches merged. At least that is my understanding.

And we usually add folks to both lists.

@jarrpa  to keep me honest

jarrpa:
Michael is right, you want him under approvers not reviewers to prevent /LGTM access. But I think this is entirely unnecessary, you could just add him to both with the understanding he shouldn't LGTM anything just yet.

obnoxxx (changes requested):
As @jarrpa confirmed, you should not only put him in reviewers, as it gives access to lgtm.
Either put him in approvers only, or put him in both.
I think both will be fine.

crombus:
Understood and updated.

obnoxxx (approved):
/lgtm

PR #993, opened 2021-01-15
fix misspelling across codebase

umangachapagain:

### DIFF
--- a/controllers/util/predicates.go
+++ b/controllers/util/predicates.go
@@ -7,7 +7,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 )
 
-// ComposePredicates will compose a variable number of predicte and return a predicate that
+// ComposePredicates will compose a variable number of predicates and return a predicate that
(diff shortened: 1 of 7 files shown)


controllers/util/predicates.go
```diff
@@ -7,7 +7,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 )
 
-// ComposePredicates will compose a variable number of predicte and return a predicate that
+// ComposePredicates will compose a variable number of predicate and return a predicate that
```
obnoxxx, line 10:
```suggestion
// ComposePredicates will compose a variable number of predicates and return a predicate that
```

obnoxxx (changes requested):
Good catches, but I think there's still one error. See inline.

umangachapagain:
> Good catches, but I think there's still one error. See inline.

Gotta catch 'em all.

obnoxxx (approved):
/lgtm

PR #1004, opened 2021-01-20
Makefile: re-add target update-generated

obnoxxx:
The target update-generated was removed in the recent project structure
overhaul, but it was still listed PHONY and referenced in the
verify-generated script. It's useful to have this separate target, so
re-adding it instead of cleaning up.


jarrpa (approved):
/lgtm

PR #1016, opened 2021-01-29
Use Golang 1.15 for security fixes to Golang std library

umangachapagain:
Fixes BZ: 1906016

### DIFF
--- a/functests/must-gather/must_gather_test.go
+++ b/functests/must-gather/must_gather_test.go
@@ -3,31 +3,31 @@ package must_gather_test
 import (
 	"testing"
 
-	. "github.com/onsi/ginkgo"
-	. "github.com/onsi/gomega"
+	"github.com/onsi/ginkgo"
+	"github.com/onsi/gomega"
 
 	tests "github.com/openshift/ocs-operator/functests"
 )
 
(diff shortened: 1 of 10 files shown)


jarrpa (approved):

functests/must-gather/must_gather_test.go
```diff
@@ -3,6 +3,7 @@ package must_gather_test
 import (
 	"testing"
 
+	"github.com/onsi/ginkgo"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
```
jarrpa, line 8: Do these need to be removed as well?

jarrpa (changes requested):
Apart from failing tests (not sure if they're valid), I think there's a piece missing in the e2e tests.

functests/must-gather/must_gather_test.go
umangachapagain: Good catch. That's where the tests are failing.

jarrpa:
Looks like there are still build failures, try running this locally to hunt it down.

jarrpa (approved):
/lgtm

PR #1019, opened 2021-01-29
Bug 1909793: Backport: [Release 4.6] adds tunefastdeviceclass property

crombus:


crombus:
Dependent on https://github.com/openshift/ocs-operator/pull/1021

pkg/controller/storagecluster/storagecluster_controller_test.go
```diff
@@ -163,6 +163,55 @@ func TestNonWatchedResourceNameNotFound(t *testing.T) {
 	assert.Equal(t, reconcile.Result{}, result)
 }
 
+func TestThrottleStorageDevices(t *testing.T) {
```
obnoxxx, line 166: This function is actually added in previous commits. Shouldn't we cherry-pick those commits separately? The original commit had only small modifications to the test function.

obnoxxx (changes requested):
There are a few hunks that don't belong to the original commit. Not sure we should merge like this - mainly in the test code so maybe?... 

Prelimiary request changes...

@jarrpa

obnoxxx:
JFTR, @jarrpa agreed to require making this backport cleaner. So I'll leave the change request. @crombus

pkg/controller/storagecluster/storagecluster_controller_test.go
crombus: yes and in release-4.6 the test function is not there https://github.com/openshift/ocs-operator/blob/release-4.6/pkg/controller/storagecluster/storagecluster_controller_test.go

crombus: the commit which adds the test is lost after the sdk update.

crombus: https://github.com/openshift/ocs-operator/pull/759/commits/2994cee1f5facd2d633be03035c89cfa38d30bbf
 found the commit

obnoxxx: The commit is not lost, but it is just on a different file than the test is in the current master: 2994cee1f5facd2d633be03035c89cfa38d30bbf

But this is actually too much different context.
I guess for this backport, it is ok to just add the function like this.
It *could* have been added as a separate commit to keep the actual commit to add TuneFastDeviceClass closer to the original, but I won't insist.

obnoxxx (approved):
Looks clean now, thanks for the update, @crombus .

As discussed, we can add the test function like this, because the history is too convoluted.

/lgtm

PR #1020, opened 2021-01-29
Increase memory footprint for rook-ceph-osd pods

gitsridhar:
In ppc64le environment, we are seeing OOM exceptions in osd pods
in ceph cluster. We would like to increase the memory footprint
for osd pods only to 10Gi from 5Gi.

Signed-off-by: gitsridhar <svenkat@us.ibm.com>

### DIFF
--- a/controllers/defaults/resources.go
+++ b/controllers/defaults/resources.go
@@ -12,11 +12,11 @@ var (
 		"osd": {
 			Requests: corev1.ResourceList{
 				corev1.ResourceCPU:    resource.MustParse("2"),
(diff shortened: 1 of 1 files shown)


jarrpa (changes requested):
I don't think this is acceptable. We are already being requested to have as small of a memory footprint as possible across our other platforms. A more appropriate solution would be to implement a way of changing the default resource allocation based on the platform (if we can detect it). Otherwise the only solution I can think of would have to rely on a StorageCluster CR that specifies the desired resource requirements explicitly, managed through external automation.

PR #1021, opened 2021-02-01
Bug 1909793: [release-4.6] update rook to 1.4.9

crombus:


obnoxxx:
/hold 

Holding this for now: We need to be careful what we do with release branches. We can not update dependencies to newer versions just like that.

Especially for the downstream builds, we need to make sure to use the correct version and if necessary backport necessary changes to dependent components.

Can't we use the rook release-4.6 branch state for ocs-operator release-4.6 branch dependency?

travisn:
Agreed, rook upstream diverges from what we have downstream since we don't take all upstream fixes to the downstream branch. Code dependencies really must be made against the downstream branch instead of upstream.

agarwal-mudit:
@obnoxxx Can we merge this now, AFAIU this is required for upstream and will not affect downstream in any manner. Please correct me if I am wrong.

obnoxxx:
I don't really understand why this PR has not been changed to update to the latest state of the rook downstream `release-4.6` branch as suggested. :-)

I admit we have not done that in the past.
And probably, it would not work just like that, because we might need container builds to do the csv generation?

But let's try to establish this model in the future.

obnoxxx (approved):
All the previous things being said, I think we can (and have to) move forward with this for now.
But in the future I would like to work with release branches for release branches :-)

obnoxxx (approved):
/lgtm

PR #1027, opened 2021-02-04
StorageCluster: add test cases for insufficient zones and regions

raghavendra-talur:
Add test cases which ensure that when insufficient zones/regions are
available, we assign racks automatically.

### DIFF
--- a/controllers/storagecluster/topology_test.go
+++ b/controllers/storagecluster/topology_test.go
@@ -71,7 +71,7 @@ func TestReconcileNodeTopologyMap(t *testing.T) {
 			expectedNodeCount: 3,
 		},
 		{
-			label:          "Case 2", // failure domain is not set
+			label:          "Case 2", // failure domain is not set and sufficient zones available
 			nodeList:       &corev1.NodeList{},
 			storageCluster: &api.StorageCluster{},
 			failureDomain:  "",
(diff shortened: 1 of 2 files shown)


controllers/storagecluster/topology_test.go
```diff
@@ -91,11 +94,135 @@ func TestReconcileNodeTopologyMap(t *testing.T) {
...
+				},
+			},
+			expectedNodeCount: 3,
+		},
 	}
 
 	for _, tc := range testcases {
 		mockStorageCluster.DeepCopyInto(tc.storageCluster)
-		mockNodeList.DeepCopyInto(tc.nodeList)
+		if tc.useDefaultNodeList == true {
```
jarrpa, line 223: Rather than introducing a new boolean, can you check to see if the `nodeList` is empty? Probably just checking `len(tc.nodeList.Items) == 0` would suffice.

raghavendra-talur: Thank you! This reads so much better. Updated.

jarrpa (approved):
/lgtm

jarrpa:
The e2e tests aren't passing, but it looks like it could be broken... since this is only for go tests and the ocs-operator-ci job passes, overriding the e2e tests.

PR #1028, opened 2021-02-04
StorageCluster: fix logic to determine minimum nodes

raghavendra-talur:
The existing code was only applicable for clusters that had failure
domains per replica set to one. In case of arbiter, we have replicas per
failure domain set to 2.


jarrpa (approved):
/lgtm

PR #1031, opened 2021-02-04
Enable new topology labels from Kubernetes

raghavendra-talur:
In ocs-operator we have been using two deprecated  topology label prefixes `failure-domain.beta.kubernetes.io` and `failure-domain.kubernetes.io`. 

The recommended label to use is `topology.kubernetes.io`. In this PR, we add support for the new label while also adhering to the following rules:
1. In a new install use only the new label
2. In existing setups, switch to the new label only if all the nodes have a equivalence of the labels i.e if node1 has zone1 set on the old labels then the new label should also have zone1 as its value.
3. In existing setups, if the values for all three labels don't match then we continue to use the old labels.
4. We also start persisting the failure domain key and values in the StorageCluster Status. This will probably make it easy to make a forced switch of label in a later release.

### DIFF
--- a/controllers/storagecluster/reconcile.go
+++ b/controllers/storagecluster/reconcile.go
@@ -87,11 +87,36 @@ osd_memory_target_cgroup_limit_ratio = 0.5
 
 var storageClusterFinalizer = "storagecluster.ocs.openshift.io"
 
+const labelZoneRegionWithoutBeta = "failure-domain.kubernetes.io/region"
+const labelZoneFailureDomainWithoutBeta = "failure-domain.kubernetes.io/zone"
+const labelRookPrefix = "topology.rook.io"
+
 var validTopologyLabelKeys = []string{
-	"failure-domain.beta.kubernetes.io",
-	"failure-domain.kubernetes.io",
-	"kubernetes.io/hostname",
-	"topology.rook.io",
+	// This is the most preferred key as kubernetes recommends zone and region
+	// labels under this key.
+	corev1.LabelZoneRegionStable,
+
+	// These two are retained only to have backward compatibility; they are
+	// deprecated by kubernetes. If topology.kubernetes.io key has same label we
+	// will skip the next two from the topologyMap.
+	corev1.LabelZoneRegion,
+	labelZoneRegionWithoutBeta,
+
+	// This is the most preferred key as kubernetes recommends zone and region
+	// labels under this key.
+	corev1.LabelZoneFailureDomainStable,
+
+	// These two are retained only to have backward compatibility; they are
+	// deprecated by kubernetes. If topology.kubernetes.io key has same label we
+	// will skip the next two from the topologyMap.
+	corev1.LabelZoneFailureDomain,
+	labelZoneFailureDomainWithoutBeta,
+
+	// This is the kubernetes recommended label to select nodes.
+	corev1.LabelHostname,
+
+	// This label is used to assign rack based topology.
+	labelRookPrefix,
 }
 
 // +kubebuilder:rbac:groups=ocs.openshift.io,resources=*,verbs=get;list;watch;create;update;patch;delete
--- a/controllers/storagecluster/reconcile.go
+++ b/controllers/storagecluster/reconcile.go
@@ -285,10 +285,6 @@ func (r *StorageClusterReconciler) reconcilePhases(
 			r.Log.Error(err, "Failed to set node topology map")
 			return reconcile.Result{}, err
 		}
-
-		if instance.Status.FailureDomain == "" {
-			instance.Status.FailureDomain = determineFailureDomain(instance)
-		}
 	}
 
 	// in-memory conditions should start off empty. It will only ever hold
--- a/controllers/storagecluster/topology.go
+++ b/controllers/storagecluster/topology.go
@@ -38,19 +38,31 @@ func (r *StorageClusterReconciler) getStorageClusterEligibleNodes(sc *ocsv1.Stor
 	return nodes, err
 }
 
-// determineFailureDomain determines the appropriate Ceph failure domain based
+// getFailureDomain returns the failure domain that was determined at the time of node topology reconcilation
+func getFailureDomain(sc *ocsv1.StorageCluster) string {
+	return sc.Status.FailureDomain
+}
+
+// setFailureDomain determines the appropriate Ceph failure domain based
 // on the storage cluster's topology map
-func determineFailureDomain(sc *ocsv1.StorageCluster) string {
+func setFailureDomain(sc *ocsv1.StorageCluster) {
+
+	// We don't change the failure domain after it is determined
 	if sc.Status.FailureDomain != "" {
-		return sc.Status.FailureDomain
+		return
 	}
 
-	if sc.Spec.FlexibleScaling {
-		return "host"
-	}
-
-	topologyMap := sc.Status.NodeTopologies
+	// default is rack
 	failureDomain := "rack"
+
+	// But if FlexiableScaling is enabled then we select host as failure domain
+	// as we need +1 scaling
+	if sc.Spec.FlexibleScaling {
+		failureDomain = "host"
+	}
+
+	// If sufficient zones are available then we select zone as the failure domain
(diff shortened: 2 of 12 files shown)


raghavendra-talur:
@jarrpa @umangachapagain @obnoxxx Please take a look.

cc: @iamniting

controllers/storagecluster/topology.go
```diff
@@ -38,27 +38,40 @@ func (r *StorageClusterReconciler) getStorageClusterEligibleNodes(sc *ocsv1.Stor
...
 	for label, labelValues := range topologyMap.Labels {
 		if strings.Contains(label, "zone") {
 			if (len(labelValues) >= 2 && arbiterEnabled(sc)) || (len(labelValues) >= 3) {
 				failureDomain = "zone"
 			}
 		}
 	}
-	return failureDomain
+
+	sc.Status.FailureDomain = failureDomain
```
iamniting, line 85: I think we should update sc here

controllers/storagecluster/reconcile.go
```diff
@@ -87,11 +87,39 @@ osd_memory_target_cgroup_limit_ratio = 0.5
...
-	"topology.rook.io",
+const labelZoneRegionWithoutBeta = "failure-domain.kubernetes.io/region"
+const labelZoneFailureDomainWithoutBeta = "failure-domain.kubernetes.io/zone"
+const labelRookPrefix = "topology.rook.io"
+
+func getValidTopologyLabelKeys() []string {
+
+	var s []string
+
+	s = append(s,
```
iamniting, line 98: instead of append should not we create a slice here itself like `s := []string{"a", "b", "c", "d"}`

controllers/storagecluster/topology.go
```diff
@@ -43,12 +43,23 @@ func getFailureDomain(sc *ocsv1.StorageCluster) string {
...
+//	return sc.Status.FailureDomainValues
+//}
+
 // setFailureDomain determines the appropriate Ceph failure domain based
 // on the storage cluster's topology map
 func setFailureDomain(sc *ocsv1.StorageCluster) {
 
 	// We don't change the failure domain after it is determined
 	if sc.Status.FailureDomain != "" {
+		sc.Status.FailureDomainKey, sc.Status.FailureDomainValues = sc.Status.NodeTopologies.GetKeyValues(sc.Status.FailureDomain)
```
iamniting, line 62: don't we want to update SC here

controllers/storagecluster/topology.go
```diff
@@ -286,3 +299,83 @@ func (r *StorageClusterReconciler) reconcileNodeTopologyMap(sc *ocsv1.StorageClu
...
+			match = false
+		}
+		for _, key := range keys {
+			vnext, found := node.Labels[key]
+			if !found {
+				match = false
+			}
+			if v != vnext {
+				match = false
+			}
```
iamniting, line 325: These 2 if statements can be combined into or statement

iamniting (changes requested):

controllers/storagecluster/topology.go
raghavendra-talur: Because of the recent refactoring, all the updates to the sc.Status are handled outside of the reconcilePhases() function. Not updating the sc status to follow the same pattern.

controllers/storagecluster/reconcile.go
raghavendra-talur: Choosing to create using append to keep future diffs cleaner.

controllers/storagecluster/topology.go
raghavendra-talur: Same reason as above.

controllers/storagecluster/topology.go
raghavendra-talur: I guess it is a matter of taste and I prefer it this way.

obnoxxx: I agree. Separate statements are easier to read and debug.

iamniting (approved):

controllers/storagecluster/reconcile.go
```diff
@@ -87,11 +87,39 @@ osd_memory_target_cgroup_limit_ratio = 0.5
...
-var validTopologyLabelKeys = []string{
-	"failure-domain.beta.kubernetes.io",
-	"failure-domain.kubernetes.io",
-	"kubernetes.io/hostname",
-	"topology.rook.io",
+const labelZoneRegionWithoutBeta = "failure-domain.kubernetes.io/region"
+const labelZoneFailureDomainWithoutBeta = "failure-domain.kubernetes.io/zone"
+const labelRookPrefix = "topology.rook.io"
+
+func getValidTopologyLabelKeys() []string {
```
jarrpa, line 94: Why create a getter to generate a string list every time instead of retaining the global variable and using the getter to return the global variable? Having the global variable with one item per line (as before) also helps keep diffs clean.

controllers/storagecluster/topology.go
```diff
@@ -280,3 +293,83 @@ func (r *StorageClusterReconciler) reconcileNodeTopologyMap(sc *ocsv1.StorageClu
...
+		}
+	}
+	return match
+}
+
+func filterDuplicateLabels(sc *ocsv1.StorageCluster, nodes *corev1.NodeList, topologyMap *ocsv1.NodeTopologyMap) {
```
jarrpa, line 325: Pretty obvious by the function name, but it may help to have a short comment above it regardless.

controllers/storagecluster/topology.go
```diff
@@ -280,3 +293,83 @@ func (r *StorageClusterReconciler) reconcileNodeTopologyMap(sc *ocsv1.StorageClu
...
+		delete(topologyMap.Labels, corev1.LabelZoneRegion)
+		delete(topologyMap.Labels, "failure-domain.kubernetes.io/region")
+		return
+	}
+
+	// We have a failure domain selected already AND we know that it is not one
+	// of the new labels. Make a conservative switch to the new labels only if
+	// all the three zone labels have the same value for each node. To keep the
+	// behavior same as in the previous releases, we will delete the new labels
+	// from the topologyMap If the user wishes to use new labels, they have to
```
jarrpa, line 358:
```suggestion
	// from the topologyMap. If the user wishes to use new labels, they have to
```

controllers/storagecluster/topology.go
```diff
@@ -280,3 +293,83 @@ func (r *StorageClusterReconciler) reconcileNodeTopologyMap(sc *ocsv1.StorageClu
...
+
+func filterDuplicateLabels(sc *ocsv1.StorageCluster, nodes *corev1.NodeList, topologyMap *ocsv1.NodeTopologyMap) {
+
+	if "" == getFailureDomain(sc) {
+		//This is the first time we are determining the failure domain
+
+		if topologyMap.ContainsKey(corev1.LabelZoneFailureDomainStable) {
+			// New label is found, use it and discard old
+			delete(topologyMap.Labels, corev1.LabelZoneFailureDomain)
+			delete(topologyMap.Labels, "failure-domain.kubernetes.io/zone")
```
jarrpa, line 333: Throughout this function, would it make sense to use the consts you defined in reconcile.go?

jarrpa (changes requested):
Sooo many apologies for the delay on this one. Overall the logic looks good, just a few small change requests.

controllers/storagecluster/reconcile.go
raghavendra-talur: Done

controllers/storagecluster/topology.go
raghavendra-talur: Done

controllers/storagecluster/topology.go
raghavendra-talur: Done

controllers/storagecluster/topology.go
raghavendra-talur: Done

jarrpa (approved):
/lgtm

raghavendra-talur:
/test ocs-operator-ci

Reason: Flaky test TestKMSConfigChanges

PR #1032, opened 2021-02-04
Must-gather: update pre-install.sh script

crombus:
add oc wait cmd and wait for the condition
to be true. make the check function run
in bg and wait for the pids to finish.

### DIFF
--- a/must-gather/collection-scripts/pre-install.sh
+++ b/must-gather/collection-scripts/pre-install.sh
@@ -47,36 +47,31 @@ labels(){
 }
 
 check_for_debug_pod(){
-    for node in ${nodes}; do
-            check_debug_pod=flase
-            if [ "$(oc get pods -n openshift-storage | grep "${node//./}-debug" | awk '{print $2}')" != "1/1" ] ; then
-                echo "waiting for the ${node//./}-debug pod to be in ready state"
-                break
-            else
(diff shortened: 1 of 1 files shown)


jarrpa (approved):
/lgtm

must-gather/collection-scripts/pre-install.sh
```diff
@@ -47,36 +47,31 @@ labels(){
...
-                break
-            else
-                pod_name=$(oc get pods -n openshift-storage | grep "${node//./}-debug" | awk '{print $1}')
-                oc label pod "$pod_name" "${node//./}"-debug='ready'
-                check_debug_pod=true
-            fi
-    done
+    debug_pod_name=$(oc get pods -n openshift-storage | grep "${node//./}-debug" | awk '{print $1}')
+    # sleep for 60 seconds giving time for debug pod to get created
+    sleep 60
```
rajatsing, line 52: just a query, do you think that 60 seconds is enough?. You must have tried it, how was the observation/experience of 60 seconds timeout, is it enough to decide?.

rajatsing (commented):
LGTM, got busy with my PR, but jose got this in already!.

must-gather/collection-scripts/pre-install.sh
crombus: 60 seconds is just for oc to register the pod. for the condition we are waiting for 200s

PR #1034, opened 2021-02-04
StorageCluster: generate cephReplicatedSpec to abstract out arbiter values

raghavendra-talur:
Instead of hardcoding it, we get the value generated using a function.

### DIFF
--- a/controllers/storagecluster/cephblockpools.go
+++ b/controllers/storagecluster/cephblockpools.go
@@ -14,20 +14,6 @@ import (
 
 type ocsCephBlockPools struct{}
 
-func generateCephReplicatedSpec(initData *ocsv1.StorageCluster) cephv1.ReplicatedSpec {
-	if arbiterEnabled(initData) {
-		return cephv1.ReplicatedSpec{
-			Size:                     4,
-			TargetSizeRatio:          .49,
-			ReplicasPerFailureDomain: 2,
-		}
-	}
-	return cephv1.ReplicatedSpec{
-		Size:            3,
-		TargetSizeRatio: .49,
-	}
-}
-
 // newCephBlockPoolInstances returns the cephBlockPool instances that should be created
 // on first run.
 func (r *StorageClusterReconciler) newCephBlockPoolInstances(initData *ocsv1.StorageCluster) ([]*cephv1.CephBlockPool, error) {
@@ -39,7 +25,7 @@ func (r *StorageClusterReconciler) newCephBlockPoolInstances(initData *ocsv1.Sto
 			},
 			Spec: cephv1.PoolSpec{
 				FailureDomain:  determineFailureDomain(initData),
-				Replicated:     generateCephReplicatedSpec(initData),
+				Replicated:     generateCephReplicatedSpec(initData, "data"),
 				EnableRBDStats: true,
 			},
 		},
--- a/controllers/storagecluster/cephcluster.go
+++ b/controllers/storagecluster/cephcluster.go
@@ -408,6 +408,15 @@ func getReplicasPerFailureDomain(sc *ocsv1.StorageCluster) int {
 	return defaults.ReplicasPerFailureDomain
 }
 
+// getCephPoolReplicatedSize returns the default replica per cluster count for a
+// StorageCluster type
+func getCephPoolReplicatedSize(sc *ocsv1.StorageCluster) uint {
+	if arbiterEnabled(sc) {
+		return uint(4)
+	}
+	return uint(3)
+}
+
 // getMinimumNodes returns the minimum number of nodes that are required for the Storage Cluster of various configurations
 func getMinimumNodes(sc *ocsv1.StorageCluster) int {
 	// Case 1: When replicasPerFailureDomain is 1.
--- a/controllers/storagecluster/generate.go
+++ b/controllers/storagecluster/generate.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	ocsv1 "github.com/openshift/ocs-operator/api/v1"
+	cephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
 )
 
 func generateNameForCephCluster(initData *ocsv1.StorageCluster) string {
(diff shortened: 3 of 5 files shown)


controllers/storagecluster/cephblockpools.go
```diff
@@ -14,18 +14,16 @@ import (
...
-	return cephv1.ReplicatedSpec{
-		Size:            3,
-		TargetSizeRatio: .49,
+func generateCephReplicatedSpec(initData *ocsv1.StorageCluster, poolType string) cephv1.ReplicatedSpec {
+	crs := cephv1.ReplicatedSpec{}
+
+	crs.Size = getReplicaSize(initData)
+	crs.ReplicasPerFailureDomain = uint(getReplicasPerFailureDomain(initData))
+	if "data" == poolType {
+		crs.TargetSizeRatio = .49
```
travisn, line 23: Is the targetSizeRatio expected to be set on all data pools? I don't believe it was set previously on the object data pool. Now we need to set it on the object pool as well? That seems correct to me as well, but want to confirm it's an intentional change.

controllers/storagecluster/cephblockpools.go
```diff
@@ -14,18 +14,16 @@ import (
...
-		return cephv1.ReplicatedSpec{
-			Size:                     4,
-			TargetSizeRatio:          .49,
-			ReplicasPerFailureDomain: 2,
-		}
-	}
-	return cephv1.ReplicatedSpec{
-		Size:            3,
-		TargetSizeRatio: .49,
+func generateCephReplicatedSpec(initData *ocsv1.StorageCluster, poolType string) cephv1.ReplicatedSpec {
```
jarrpa, line 17: Since this function is going to be used for multiple CRs, I'd move it to a different file. Not sure if it should be an existing one or a new one, I'll leave that up to you.

controllers/storagecluster/cephblockpools.go
raghavendra-talur: Yes, the target size ratio was begin set for datapools in the cephfilesystem and the objectstore both.
See 
https://github.com/openshift/ocs-operator/pull/1034/files#diff-4cd9b9e8e155fb315c3d558aa21839ed627a878432c8c11e13fb811669809be7L142 
and 
https://github.com/openshift/ocs-operator/pull/1034/files#diff-fabbfe52f64e79da27f9fff14bbd71e9a6bc122fd84073a8a1a7debc4aa9ac30L38

controllers/storagecluster/cephblockpools.go
raghavendra-talur: Moved to generate.go. It is the closest in usage, generates artifacts considering ceph related configuration in the storage cluster spec.

Let me know if you want to move the other newly introduced get functions too. I was thinking of moving them to defaults.go but left them as is.

controllers/storagecluster/cephblockpools.go
travisn: That commit to set the targetSizeRatio on the object store was also recent, right? I wasn't clear if that was intentional. Was there a separate BZ for that?

raghavendra-talur: I misunderstood your comment. Looked at the previous commit. It is from https://github.com/openshift/ocs-operator/pull/635 .

The BZ and the relevant comments are 
* https://bugzilla.redhat.com/show_bug.cgi?id=1850538#c24
*  https://bugzilla.redhat.com/show_bug.cgi?id=1850538#c25
*  https://bugzilla.redhat.com/show_bug.cgi?id=1850538#c30

travisn: That's what I was looking for, thanks!

controllers/storagecluster/cephcluster.go
```diff
@@ -405,6 +405,13 @@ func getReplicasPerFailureDomain(sc *ocsv1.StorageCluster) int {
 	return defaults.ReplicasPerFailureDomain
 }
 
+func getReplicaSize(sc *ocsv1.StorageCluster) uint {
```
jarrpa, line 408: Rename this to something like `getCephPoolReplicatedSize()`, and please add a quick comment on what this function is for.

controllers/storagecluster/generate.go
```diff
@@ -55,3 +56,15 @@ func generateNameForSnapshotClassDriver(initData *ocsv1.StorageCluster, snapshot
 func generateNameForSnapshotClassSecret(snapshotType SnapshotterType) string {
 	return fmt.Sprintf("rook-csi-%s-provisioner", snapshotType)
 }
+
+func generateCephReplicatedSpec(initData *ocsv1.StorageCluster, poolType string) cephv1.ReplicatedSpec {
```
jarrpa, line 60: Add a comment for the function.

jarrpa (changes requested):
Gah, sorry for the delay, fell behind again. A few more changes requested.

controllers/storagecluster/cephcluster.go
raghavendra-talur: done

controllers/storagecluster/generate.go
raghavendra-talur: done

jarrpa (approved):
/lgtm

raghavendra-talur:
The unit test is failing with the error
```
cephcluster_test.go:427: Failed: "case 2". Error: dial tcp [::1]:12321: connect: connection refused
--- FAIL: TestKMSConfigChanges (0.00s)
```

I have been seeing these issues on my laptop too. @jarrpa @aruniiird I need help in debugging this test. This test requires creation of a LISTEN port on the host.

raghavendra-talur:
/test ocs-operator-ci
--

jarrpa:
> The unit test is failing with the error
> 
> ```
> cephcluster_test.go:427: Failed: "case 2". Error: dial tcp [::1]:12321: connect: connection refused
> --- FAIL: TestKMSConfigChanges (0.00s)
> ```
> 
> I have been seeing these issues on my laptop too. @jarrpa @aruniiird I need help in debugging this test. This test requires creation of a LISTEN port on the host.

...odd, I didn't think it'd get replicated locally. At least I haven't run into it lately. Usually when I see something like that I think it's a testing flake due to some rate limiting something or other with the openshift-dev AWS account.

agarwal-mudit:
@raghavendra-talur This needs manual BP

PR #1041, opened 2021-02-05
Bug 1914203: [release-4.7] StorageCluster: fix logic to determine minimum nodes

agarwal-mudit:
The existing code was only applicable for clusters that had failure
domains per replica set to one. In case of arbiter, we have replicas per
failure domain set to 2.

(cherry picked from commit 6836e24569a5f54a0b39661504f689321d2f9c35)


agarwal-mudit:
Created bp PR for #1028 because of the conflicts.

@raghavendra-talur

PR #1042, opened 2021-02-05
Define reconcile strategy for rook-config-override

jarrpa:


controllers/storagecluster/reconcile.go
```diff
@@ -513,7 +520,7 @@ func (obj *ocsCephConfig) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.S
 		}
 	}
 	val, ok := found.Data["config"]
-	if !ok || val != rookConfigData || !ownerRefFound {
+	if !ok || (val != defaultRookConfigData && val != sc.Spec.RookConfig) || !ownerRefFound {
```
obnoxxx, line 523: I think we should compare it against `configData`. This current comparison would always lead to updating the config map if `RookConfig`    is not set in the storagecluster spec.

controllers/storagecluster/reconcile.go
```diff
@@ -476,6 +476,13 @@ func (r *StorageClusterReconciler) validateStorageDeviceSets(sc *ocsv1.StorageCl
 // ensureCreated ensures that a ConfigMap resource exists with its Spec in
 // the desired state.
 func (obj *ocsCephConfig) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.StorageCluster) error {
+	var configData string
```
obnoxxx, line 479: We might want to call this `rookConfigData` just like the previous constant which is now `defaultRookConfigData`.

controllers/storagecluster/reconcile.go
obnoxxx: And if we named the variable `rookConfigData` above (instead of `configData`,  then this code would remain unchanged! 😁

obnoxxx (changes requested):
I like this in general, but some detail comments and requests inline.

jarrpa:
Variables renamed! Don't know why I didn't see it first. :stuck_out_tongue_closed_eyes:

umangachapagain (commented):
Why can't we directly update the ConfigMap?
Or, worst case - can't we just pass the ConfigMap object instead of string literal?

jarrpa:
> Why can't we directly update the ConfigMap?

Isn't that exactly what we're doing now? We Get(), compare, and Update() the ConfigMap as needed.

> Or, worst case - can't we just pass the ConfigMap object instead of string literal?

This is just exposing a field in the StorageCluster CR for the Rook-Ceph config string. I don't see a reason to force the entire ConfigMap struct when the only thing that matters is one string literal.

umangachapagain:
> > Why can't we directly update the ConfigMap?
> 
> Isn't that exactly what we're doing now? We Get(), compare, and Update() the ConfigMap as needed.
> 
I mean't "we the user". OCS Operator should just have a flag to make it customizable?

> > Or, worst case - can't we just pass the ConfigMap object instead of string literal?
> 
> This is just exposing a field in the StorageCluster CR for the Rook-Ceph config string. I don't see a reason to force the entire ConfigMap struct when the only thing that matters is one string literal.
>
There is also no gain in forcing a string for config, when there's already a dedicated ConfigMap for it.

I believe that was your initial thought as well: https://bugzilla.redhat.com/show_bug.cgi?id=1914475#c5

jarrpa:
> > > Why can't we directly update the ConfigMap?
> > 
> > Isn't that exactly what we're doing now? We Get(), compare, and Update() the ConfigMap as needed.
> 
> I mean't "we the user". OCS Operator should just have a flag to make it customizable?

Ah, gotcha.

> I believe that was your initial thought as well: https://bugzilla.redhat.com/show_bug.cgi?id=1914475#c5

Yeah, true... maybe that would be easier for people to understand? Folks would probably be more familiar with editing a ConfigMap directly than figuring out how to it indirectly from another CR. It would also guard us against any future (though unlikely) changes to the format of the ConfigMap. @obnoxxx ?

jarrpa:
...I just had another thought: we probably shouldn't be controlling this from the StorageCluster CR at all. After all, this is a configuration for the rook-ceph-*operator*, not an individual CephCluster. We even initialize the ConfigMap in OCSInitialization, we hsould be implementing this there!

obnoxxx:
> ...I just had another thought: we probably shouldn't be controlling this from the StorageCluster CR at all. After all, this is a configuration for the rook-ceph-_operator_, not an individual CephCluster. We even initialize the ConfigMap in OCSInitialization, we hsould be implementing this there!

This argumentation seems to be very logical and compelling to me.

obnoxxx:
> > ...I just had another thought: we probably shouldn't be controlling this from the StorageCluster CR at all. After all, this is a configuration for the rook-ceph-_operator_, not an individual CephCluster. We even initialize the ConfigMap in OCSInitialization, we hsould be implementing this there!
> 
> This argumentation seems to be very logical and compelling to me.

@jarrpa  - so what does this mean for this PR?

jarrpa:
> > > ...I just had another thought: we probably shouldn't be controlling this from the StorageCluster CR at all. After all, this is a configuration for the rook-ceph-_operator_, not an individual CephCluster. We even initialize the ConfigMap in OCSInitialization, we hsould be implementing this there!
> > 
> > 
> > This argumentation seems to be very logical and compelling to me.
> 
> @jarrpa - so what does this mean for this PR?

Oh, it needs to be entirely redone. I was thinking of just overhauling the branch.

jarrpa:
So I'm glad I left this sitting around for a few days. Turns out that I was *entirely* wrong, and the `rook-config-override` ConfigMap only applies to the CephCluster that is in its same Namespace. As such, it makes perfect sense to keep it in the StorageCluster reconciliation. So now the question is: do we force the admin to provide the `rook-config-override` data in the StorageCluster CR or do we have them edit the ConfigMap directly? The latter would mean that they'd have to manually create the ConfigMap prior to creating the StorageCluster for the settings to affect the CephCluster at creation time.

travisn:
> So I'm glad I left this sitting around for a few days. Turns out that I was _entirely_ wrong, and the `rook-config-override` ConfigMap only applies to the CephCluster that is in its same Namespace. As such, it makes perfect sense to keep it in the StorageCluster reconciliation. So now the question is: do we force the admin to provide the `rook-config-override` data in the StorageCluster CR or do we have them edit the ConfigMap directly? The latter would mean that they'd have to manually create the ConfigMap prior to creating the StorageCluster for the settings to affect the CephCluster at creation time.

The behavior I'd expect is to have an option like other resources whether to reconcile always, only create initially, etc. By default we would set it reconcile. If they change it to be unreconciled, then OCS wouldn't ever update it, and they could add their desired settings. If settings are needed at initial cluster creation, they would need to create the CM before creating the cluster.

If the configmap is updated after initial cluster creation, the settings could still be applied by restarting the daemon that is affected. Some settings affect the rook operator, some affect the mons, osds, etc.

jarrpa:
> The behavior I'd expect is to have an option like other resources whether to reconcile always, only create initially, etc. By default we would set it reconcile. If they change it to be unreconciled, then OCS wouldn't ever update it, and they could add their desired settings. If settings are needed at initial cluster creation, they would need to create the CM before creating the cluster.
>
> If the configmap is updated after initial cluster creation, the settings could still be applied by restarting the daemon that is affected. Some settings affect the rook operator, some affect the mons, osds, etc.

I can live with that.

obnoxxx (approved):
/lgtm

PR #1043, opened 2021-02-08
StorageCluster:Fixed placement config for rack failure domain and modified labelselector

kesavanvt:
This PR provides the folowing changes

- To have effective OSD distribution, 
    - Added changes to have a common key as topology spread label selector as it matches both prepare and OSD pod.
    - Added hard constraints for prepare placement failure domain(zone/rack) in case of multiple topology spread constraints.  

- Restricts the movement of portable volumes between failure domain (rack) while node drain by imposing hard constraint in topology spread. 

### DIFF
--- a/controllers/storagecluster/cephcluster.go
+++ b/controllers/storagecluster/cephcluster.go
@@ -547,9 +547,10 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster, serverVersion *version.
 							// default preparePlacement. This serves even distribution at the host level
 							// within a failure domain (zone/rack).
 							if topologyKey != corev1.LabelHostname {
-								preparePlacement.TopologySpreadConstraints = append(preparePlacement.TopologySpreadConstraints, preparePlacement.TopologySpreadConstraints[0])
+								addStrictFailureDomainTSC(&preparePlacement, topologyKey)
+							} else {
+								preparePlacement.TopologySpreadConstraints[0].TopologyKey = topologyKey
 							}
-							preparePlacement.TopologySpreadConstraints[0].TopologyKey = topologyKey
 						}
 					}
 				} else {
@@ -728,4 +729,14 @@ func getCephObjectStoreGatewayInstances(sc *ocsv1.StorageCluster) int32 {
 		return int32(defaults.ArbiterCephObjectStoreGatewayInstances)
 	}
 	return int32(defaults.CephObjectStoreGatewayInstances)
+}
+
+// addStrictFailureDomainTSC adds hard topology constraints at failure domain level
+// and uses soft topology constraints within falure domain (across host).
+func addStrictFailureDomainTSC(placement *rook.Placement, topologyKey string) {
+	newTSC := placement.TopologySpreadConstraints[0]
+	newTSC.TopologyKey = topologyKey
+	newTSC.WhenUnsatisfiable = "DoNotSchedule"
+
+	placement.TopologySpreadConstraints = []corev1.TopologySpreadConstraint{newTSC, placement.TopologySpreadConstraints[0]}
 }
--- a/controllers/storagecluster/cephcluster.go
+++ b/controllers/storagecluster/cephcluster.go
@@ -541,11 +541,19 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster, serverVersion *version.
 					if noPreparePlacement {
 						in := getPlacement(sc, "osd-prepare-tsc")
 						(&in).DeepCopyInto(&preparePlacement)
+					}
 
-						if len(topologyKeyValues) >= replica {
-							// If topologyKey is not host, append additional topology spread constarint to the
-							// default preparePlacement. This serves even distribution at the host level
-							// within a failure domain (zone/rack).
+					if len(topologyKeyValues) >= replica {
+						// Hard constraints are set in OSD placement for portable volumes with rack failure domain
+						// domain as there is no node affinity in PVs. This restricts the movement of OSDs
+						// between failure domain.
+						if portable && !strings.Contains(topologyKey, "zone") {
+							addStrictFailureDomainTSC(&placement, topologyKey)
+						}
+						// If topologyKey is not host, append additional topology spread constarint to the
+						// default preparePlacement. This serves even distribution at the host level
+						// within a failure domain (zone/rack).
+						if noPreparePlacement {
 							if topologyKey != corev1.LabelHostname {
 								addStrictFailureDomainTSC(&preparePlacement, topologyKey)
 							} else {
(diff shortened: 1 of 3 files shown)


jarrpa (changes requested):
Okay, I think I get this... it would help if you updated the commit message (not PR description!) to describe the problem and how this addresses it a bit more.

controllers/storagecluster/cephcluster.go
```diff
@@ -531,11 +531,18 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster, serverVersion *version.
...
 
-						if len(topologyKeyValues) >= replica {
-							// If topologyKey is not host, append additional topology spread constarint to the
-							// default preparePlacement. This serves even distribution at the host level
-							// within a failure domain (zone/rack).
+					if len(topologyKeyValues) >= replica {
+						if portable && !strings.Contains(topologyKey, "zone") {
+							placement.TopologySpreadConstraints = append(placement.TopologySpreadConstraints, placement.TopologySpreadConstraints[0])
+							placement.TopologySpreadConstraints[0].TopologyKey = topologyKey
+							placement.TopologySpreadConstraints[0].WhenUnsatisfiable = "DoNotSchedule"
```
travisn, line 540: Why not always set this? Without this, the only guard we would have against moving the OSDs across failure domains is the PV constraints. In some environments, even if we're running in zones, it's possible that the PVs might be schedulable across zones, although maybe not a likely scenario.

jarrpa: My hope would be that the PVs would always come with constraints on where they can be used. Is this not the case?

kesavanvt:
> Okay, I think I get this... it would help if you updated the commit message (not PR description!) to describe the problem and how this addresses it a bit more.

Done, Thanks

controllers/storagecluster/cephcluster.go
kesavanvt: Verified most of the cloud providers that their PVs hold the constraints in case of zone failure domain. Also, in case of zones the OSD pod placement is dependent on prepare job placement as the PV created from it holds the actual constraint. Keeping this into consideration, we apply soft constraints for zones.

travisn: @jarrpa For example, OCS puts the artificial constraints on racks. LSO isn't aware of the racks, so OCS needs to put constraints on the pods so they don't move across those boundaries we've chosen. Rook doesn't touch the PVs, it only consumes them through the PVCs. 
@kesavanvt My question is really why we shouldn't apply this constraint to zones. The constraint wouldn't really be needed for zones since the PVs have the constraints, but it shouldn't hurt either. Consistent code is easier to maintain than special cases. I just don't see the need for the special case here.

kesavanvt: IIRC we proceeded on `ScheduleAnyway`(soft constraint) keeping both "rook-ceph-osd-prepare", "rook-ceph-osd" as label selector for prepare pod placement, as topology spread only takes into consideration on the running pods and not on the completed jobs. It may be possible that it breaks the even spread at some point which is not likely to happen. Even if it does, the subsequent osd expansion would try to balance the even spread. Keeping this into account, having hard constraint for the OSD pod would be a problem for zones as it depends on pv constraint which is again dependent on prepare pod placement.

jarrpa: This latest reasoning works for me. 👍

controllers/storagecluster/cephcluster.go
```diff
@@ -531,11 +531,18 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster, serverVersion *version.
...
 						in := getPlacement(sc, "osd-prepare-tsc")
 						(&in).DeepCopyInto(&preparePlacement)
+					}
 
-						if len(topologyKeyValues) >= replica {
-							// If topologyKey is not host, append additional topology spread constarint to the
-							// default preparePlacement. This serves even distribution at the host level
-							// within a failure domain (zone/rack).
+					if len(topologyKeyValues) >= replica {
+						if portable && !strings.Contains(topologyKey, "zone") {
```
jarrpa, line 537: I know it's already in the PR, but please add a comment here describing why this new condition is required.

controllers/storagecluster/cephcluster.go
```diff
@@ -531,11 +531,18 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster, serverVersion *version.
...
+							placement.TopologySpreadConstraints = append(placement.TopologySpreadConstraints, placement.TopologySpreadConstraints[0])
```
jarrpa, line 538: This is a little funky. I think it would be a bit easier to read if you create a new variable like `newTSC`, set it to constraint `[0]`, modified it, then end with `[0] = {newTSC, [0]}`. Thoughts?

controllers/storagecluster/cephcluster_test.go
```diff
@@ -338,7 +338,11 @@ func TestStorageClassDeviceSetCreation(t *testing.T) {
 			assert.Equal(t, deviceSet.DataPVCTemplate, scds.VolumeClaimTemplates[0])
 			assert.Equal(t, true, scds.Portable)
 			assert.Equal(t, c.sc.Spec.Encryption.Enable, scds.Encrypted)
-			assert.Equal(t, getPlacement(c.sc, "osd-tsc"), scds.Placement)
+			if scds.Portable && c.topologyKey == "rack" {
+				assert.Equal(t, scds.Placement.TopologySpreadConstraints[0].WhenUnsatisfiable, corev1.UnsatisfiableConstraintAction("DoNotSchedule"))
```
jarrpa, line 342: While I know it'd be a hassle, you should check the full structure `scds.Placement.TopologySpreadConstraints`, like `len() == 2`, etc..

jarrpa (changes requested):
Code is mostly good. Please include the information in the PR description in the commit message. You can omit the sample YAML.

controllers/storagecluster/placement.go
```diff
@@ -130,3 +130,13 @@ func setTopologyForAffinity(placement *rookv1.Placement, selectorValue string, t
 	}
 	appendNodeRequirements(placement, nodeZoneSelector)
 }
+
+// setMultipleTSC assigns hard topology constraints at failure domain level
+// and soft topology constraints within falure domain (across host).
+func setMultipleTSC(placement *rookv1.Placement, topologyKey string) {
```
jarrpa, line 136: This function name is too generic for what it actually does. Rename this to something like `addStrictFailureDomainTSC()`. Also, move this to cephcluster.go

jarrpa (changes requested):

jarrpa (approved):
/lgtm

PR #1045, opened 2021-02-08
Bug 1902192: [release-4.7] Upgrade rook to v1.5.6 and K8s to 0.20.2

aruniiird:
This is a backport PR created to upgrade rook to v1.5.6 and corresponding k8s dependency being upgraded to v0.20.2 on release-4.7 branch.


jarrpa (approved):
/lgtm

agarwal-mudit:
/retitle Bug 1902192: [release-4.7] Upgrade rook to v1.5.6 and K8s to 0.20.2

obnoxxx:
Just to re-iterate what I recently commented on a release-4.6 PR: We need to move to a model where the don't follow upstream release branches of components, but on corresponding downstream release-4.X branches. I know it will require some work to enable generating csvs etc, but it would be the correct way to do dependencies in the downstream release branches.

PR #1044, opened 2021-02-08
must-gather:do not create  helper pod if storagecluster is deleted

rajatsing:
with storage cluster deleted, MG should not attempt to bring up helper pod.

Signed-off-by: RAJAT SINGH <rajasing@redhat.com>

PR #1048, opened 2021-02-09
Must-gather: add bg process for ceph cmds

crombus:
add bg process for ceph cmds. create
temp log file for ceph bg cmds and
cat them to gather-debug.log file.

### DIFF
--- a/must-gather/collection-scripts/gather_ceph_resources
+++ b/must-gather/collection-scripts/gather_ceph_resources
@@ -125,9 +125,9 @@ for ns in $namespaces; do
         printf "collecting snapshot info for ceph rbd volumes \n" | tee -a  "${BASE_COLLECTION_PATH}"/gather-ceph-debug.log
         COMMAND_OUTPUT_FILE=${COMMAND_OUTPUT_DIR}/rbd_vol_and_snap_info
         # Inspecting ceph block pools for ceph rbd
-        blockpools=$(oc get cephblockpools.ceph.rook.io -n openshift-storage -o jsonpath="{range .items[*]}{@.metadata.name}{'\n'}{end}")
+        blockpools=$(timeout 60 oc get cephblockpools.ceph.rook.io -n openshift-storage -o jsonpath="{range .items[*]}{@.metadata.name}{'\n'}{end}")
         for bp in $blockpools; do
-            images=$(oc -n openshift-storage exec "${HOSTNAME}"-helper -- bash -c "rbd ls -p $bp")
+            images=$(timeout 60 oc -n openshift-storage exec "${HOSTNAME}"-helper -- bash -c "rbd ls -p $bp")
             for image in $images; do
                 { timeout 120 oc -n "${ns}" exec "${HOSTNAME}"-helper -- bash -c "rbd info $image --pool $bp" >> "${COMMAND_OUTPUT_FILE}"; } >> "${COMMAND_ERR_OUTPUT_DIR}"/gather-rbd-image-"${image}"-debug.log 2>&1 &
                 pids_ceph+=($!)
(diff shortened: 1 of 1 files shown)


must-gather/collection-scripts/gather_ceph_resources
```diff
@@ -139,21 +145,24 @@ for ns in $namespaces; do
...
                  snaps=$(oc -n openshift-storage exec "${HOSTNAME}"-helper -- bash -c "ceph fs subvolume snapshot ls $fs $subvol $svg | jq --raw-output '.[].name'")
                  count=$(oc -n openshift-storage exec "${HOSTNAME}"-helper -- bash -c "ceph fs subvolume snapshot ls $fs $subvol $svg | jq --raw-output '.[].name' | wc -l")
                  { printf "Snapshot count in subvolume: %s=%s\n" "${subvol}" "${count}" >> "${COMMAND_OUTPUT_FILE}"; }
                  for snap in $snaps; do
                      { printf "Information for snapshot: %s\n" "${snap}" >> "${COMMAND_OUTPUT_FILE}"; }
-                     { timeout 120 oc -n "${ns}" exec "${HOSTNAME}"-helper -- bash -c "ceph fs subvolume snapshot info $fs $subvol $snap $svg --connect-timeout=15" >> "${COMMAND_OUTPUT_FILE}"; } >> "${BASE_COLLECTION_PATH}"/gather-debug.log 2>&1
+                     { timeout 120 oc -n "${ns}" exec "${HOSTNAME}"-helper -- bash -c "ceph fs subvolume snapshot info $fs $subvol $snap $svg --connect-timeout=15" >> "${COMMAND_OUTPUT_FILE}"; } >> "${BASE_COLLECTION_PATH}"/gather-ceph-debug.log 2>&1 &
+                 pids_ceph+=($!)
```
jarrpa, line 155: Probably should fix this indentation.

must-gather/collection-scripts/gather_ceph_resources
```diff
@@ -139,21 +145,24 @@ for ns in $namespaces; do
              subvols=$(oc -n openshift-storage exec "${HOSTNAME}"-helper -- bash -c "ceph fs subvolume ls $fs $svg | jq --raw-output '.[].name' ")
              for subvol in $subvols; do
                  { printf "Information for subvolume: %s\n" "${subvol}" >> "${COMMAND_OUTPUT_FILE}"; }
-                 { timeout 120 oc -n "${ns}" exec "${HOSTNAME}"-helper -- bash -c "ceph fs subvolume info $fs $subvol $svg --connect-timeout=15" >> "${COMMAND_OUTPUT_FILE}"; } >> "${BASE_COLLECTION_PATH}"/gather-debug.log 2>&1
+                 { timeout 120 oc -n "${ns}" exec "${HOSTNAME}"-helper -- bash -c "ceph fs subvolume info $fs $subvol $svg --connect-timeout=15" >> "${COMMAND_OUTPUT_FILE}"; } >> "${BASE_COLLECTION_PATH}"/gather-ceph-debug.log 2>&1
```
jarrpa, line 148: Should this also be sent into the background?

must-gather/collection-scripts/gather_ceph_resources
```diff
@@ -139,21 +145,24 @@ for ns in $namespaces; do
...
+                     { timeout 120 oc -n "${ns}" exec "${HOSTNAME}"-helper -- bash -c "ceph fs subvolume snapshot info $fs $subvol $snap $svg --connect-timeout=15" >> "${COMMAND_OUTPUT_FILE}"; } >> "${BASE_COLLECTION_PATH}"/gather-ceph-debug.log 2>&1 &
+                 pids_ceph+=($!)
                  done
              done
         done
+        printf "waiting for pids to finish" | tee -a  "${BASE_COLLECTION_PATH}"/gather-ceph-debug.log
+        wait "${pids_ceph[@]}"
     }
-    if [ "$(oc get pods -n openshift-storage -l  must-gather-helper-pod='' | awk '{print $2}')" == "1/1" ] ; then
+    if [ "$(oc get pods -n openshift-storage --no-headers -l  must-gather-helper-pod='' | awk '{print $2}')" == "1/1" ] ; then
```
jarrpa, line 162: This seems like an unrelated change? If so move to a separate PR, probably one that includes further changes to remove the need for awk using templates in the `oc get` calls.

must-gather/collection-scripts/gather_ceph_resources
```diff
@@ -102,33 +102,39 @@ for ns in $namespaces; do
...
+             { timeout 120 oc -n "${ns}" exec "${HOSTNAME}"-helper -- bash -c "${ceph_commands[$i]} --connect-timeout=15" >> "${COMMAND_OUTPUT_FILE}"; } >> "${BASE_COLLECTION_PATH}"/gather-ceph-debug.log 2>&1 &
+             pids_ceph+=($!)
+             { timeout 120 oc -n "${ns}" exec "${HOSTNAME}"-helper -- bash -c "${ceph_commands[$i]} --connect-timeout=15 --format json-pretty" >> "${JSON_COMMAND_OUTPUT_FILE}"; } >> "${BASE_COLLECTION_PATH}"/gather-ceph-debug.log 2>&1 &
+             pids_ceph+=($!)
         done
         for i in $(timeout 120 oc -n "${ns}" exec "${HOSTNAME}"-helper -- bash -c "ceph osd lspools --connect-timeout=15"|awk '{print $2}'); do
-             { timeout 120 oc -n "${ns}" exec "${HOSTNAME}"-helper -- bash -c "rbd ls -p $i" >> "${COMMAND_OUTPUT_DIR}/pools_rbd_$i"; } >> "${BASE_COLLECTION_PATH}"/gather-debug.log 2>&1;
+             { timeout 120 oc -n "${ns}" exec "${HOSTNAME}"-helper -- bash -c "rbd ls -p $i" >> "${COMMAND_OUTPUT_DIR}/pools_rbd_$i"; } >> "${BASE_COLLECTION_PATH}"/gather-ceph-debug.log 2>&1;
```
jarrpa, line 118: Should this be sent into the background?

jarrpa (changes requested):
In general this makes sense, but there's still a problem with the logging. By outputting the stdout of multiple background processes to the same log file there is no guarantee that the output lines will be coherent since multiple jobs could output lines between each other. You should output each command's stdout to separate files, then later aggregate them into the main debug log. If they're all in the same directory you don't need to worry about the exact file names, just loop over all files in the directory.

must-gather/collection-scripts/gather_ceph_resources
rajatsing: right, this should go in the background.

must-gather/collection-scripts/gather_ceph_resources
rajatsing: I have made this change in my PR #1044 you can remove it.

must-gather/collection-scripts/gather_ceph_resources
rajatsing: right!. You missed `&` here as well.

rajatsing (changes requested):
You can remove the `--no-header` line since I already caught it in my PR #1044 and you have also missed a couple of processes to be moved in background.

must-gather/collection-scripts/gather_ceph_resources
crombus: created another PR

crombus:
@jarrpa I didn’t cat the output of logs in gather-debug.log.  Reason they are still up and down it ls not in order. I suggest to have it in the log folder unless you have some other suggestions.

rajatsing (approved):

must-gather/collection-scripts/gather_ceph_resources
```diff
@@ -94,66 +94,79 @@ for ns in $namespaces; do
     ceph_collection(){
         COMMAND_OUTPUT_DIR=${CEPH_COLLECTION_PATH}/must_gather_commands
         COMMAND_JSON_OUTPUT_DIR=${CEPH_COLLECTION_PATH}/must_gather_commands_json_output
+        COMMAND_LOG_OUTPUT_DIR=${CEPH_COLLECTION_PATH}/logs
```
jarrpa, line 97: I think it might make sense to name this `COMMAND_ERR_OUTPUT_DIR`? The main purpose of this is to collect the output to fd 2, right?

jarrpa (changes requested):
I think it's totally okay if they're not in the same file. However, given that, it may just be easier to have both stdout and stderr output to the same file regardless. I can imagine someone looking through must-gather would find it convenient if they just had one file to determine the output (successful or failure) of a single command. Thoughts?

Also, merge the first and last commits.

must-gather/collection-scripts/gather_ceph_resources
crombus: yes. I'll update it

crombus:
> I think it's totally okay if they're not in the same file. However, given that, it may just be easier to have both stdout and stderr output to the same file regardless. I can imagine someone looking through must-gather would find it convenient if they just had one file to determine the output (successful or failure) of a single command. Thoughts?
> 
 As of now if the cmd has failed then the file will be empty and will be removed at end by MG.

For end user we are providing just the data which was collected.  If at any point the data is not collected or cmd failed to find the reason we look in gather-debug.log.

I understand your point but in that case to list which cmds fail and which passed will be difficult as they will not be in one location. We have to check each file to find that it is failed or passed.

> Also, merge the first and last commits.

jarrpa (approved):
/lgtm

jarrpa:
> For end user we are providing just the data which was collected. If at any point the data is not collected or cmd failed to find the reason we look in gather-debug.log.
> 
> I understand your point but in that case to list which cmds fail and which passed will be difficult as they will not be in one location. We have to check each file to find that it is failed or passed.

A solid argument, thanks!

PR #1053, opened 2021-02-10
Localizing build process.

renan-campos:
I noticed that the bundle was being pulled from the Go directory and not necassarilly the build repository.


jarrpa (approved):
/lgtm

PR #1054, opened 2021-02-10
OCS-Operator now supports multiple external monitoring IPs

aruniiird:
A ceph cluster will contain an active 'mgr' (manager) and
(may have) multiple stand-by 'mgr'-s.
This change will allow OCS-Operator to accept multiple
(comma / space separated) endpoints/IP addresses, which
will be passed to external ceph-cluster spec.

This will help rook (which reads the ceph-cluster spec internally)
to change over to the next 'mgr' when the current active 'mgr' fails.

### DIFF
--- a/controllers/storagecluster/external_resources.go
+++ b/controllers/storagecluster/external_resources.go
@@ -83,6 +83,10 @@ func sha512sum(tobeHashed []byte) (string, error) {
 	return fmt.Sprintf("%x", h.Sum(nil)), nil
 }
 
+func parseMonitoringIPs(monIP string) []string {
+	return strings.Fields(strings.ReplaceAll(monIP, ",", " "))
+}
+
 func (r *StorageClusterReconciler) externalSecretDataChecksum(instance *ocsv1.StorageCluster) (string, error) {
 	found, err := r.retrieveSecret(externalClusterDetailsSecret, instance)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


aruniiird:
@jarrpa , can you please take a look at this PR... This PR is to add an array of MGR pods (one active and other standby mgrs) and pass it to the ceph cluster.
Why is it needed?: We are unable to get metrics if the active mgr goes down, as we only pass active mgr pod IP.

controllers/storagecluster/external_resources.go
```diff
@@ -240,17 +240,28 @@ func (r *StorageClusterReconciler) createExternalStorageClusterResources(instanc
...
 					"Monitoring Endpoint not present in the external cluster secret %s",
 					externalClusterDetailsSecret)
 				r.Log.Error(err, "Failed to get Monitoring IP.")
 				return err
 			}
+			// replace any comma in the monitoring ip string with space
+			// and then collect individual (non-empty) items' array
+			monIPArr := strings.Fields(strings.ReplaceAll(monitoringIP, ",", " "))
 			monitoringPort := d.Data["MonitoringPort"]
 			if monitoringPort != "" {
```
jarrpa, line 255: This may be outside this PR (or at least a separate commit in this PR), but is there a reason we don't do `checkEndpointReachable()` if `monitoringPort == ""`? Is there no default port to fall back on, or should we consider it invalid to not provide a port?

jarrpa (changes requested):
In general I like what I see. A few questions at the top:

* Is there a limitation that forces the incoming list to be comma separated instead of space separated? I see that there is a possibility to use `string.FieldFunc()`, but at that point you're writing a small inline function... so may as well use `ReplaceAll()`.
* Since you're using the same conversion twice, what about creating a new function like `func parseMgrMonitoringIPs(data string) []string {}` or something? Part of `external_resources.go`.

controllers/storagecluster/external_resources.go
aruniiird: If port is not given, we (from OCS-Operator) are not providing a default. Rook will internally provide a default port when the given port is empty.

aruniiird:
> In general I like what I see. A few questions at the top:
> 
>     * Is there a limitation that forces the incoming list to be comma separated instead of space separated? I see that there is a possibility to use `string.FieldFunc()`, but at that point you're writing a small inline function... so may as well use `ReplaceAll()`.

In the current implementation, we accept both space and comma as delimiters. As mentioned in the comment, we replace all the commas with space, then use `strings.Fields()` function to split this modified (no-comma-only-space) IP-string into an array of strings.

> 
>     * Since you're using the same conversion twice, what about creating a new function like `func parseMgrMonitoringIPs(data string) []string {}` or something? Part of `external_resources.go`.

The initial thought was, as it is only a single line, `strings.Fields(strings.ReplaceAll(monitoringIP, ",", " "))`, why make a separate function. But reading through your comment, it may be a good approach to provide a function to parse the monitoring IPs. Please take a look.

controllers/storagecluster/external_resources.go
jarrpa: :+1:

jarrpa (approved):
/lgtm

PR #1055, opened 2021-02-10
Add a master branch configuration for openshift-ci

egafford:
At present, all OCS operator CI builds use fixed container versions
for each other OCS project. This will continue to be appropriate,
but a non-blocking CI job to notify us of compatibility issues
between the master branches of multiple OCS components would be
useful. This change introduces new scripts for the easy generation
of an OCS registry that will pull the most recent container image
from the master branch of each component project.

Notes on implementation:
- I chose to use the most recent published container image for each
  project, rather than building from the most recent code for each,
  in order to leave each project team fully responsible for their own
  build process and able to change it without breaking any CI jobs.
  This is a tradeoff, as teams may not update container images as
  often as code, but I chose maintainability.
- I also chose to use jq rather than parsing JSON using basic OS
  tools like awk and sed. Because of this, I opted to build from the
  ubi8/ubi-minimal image rather than the ubi8/go-toolset image
  (which requires subscription management to install tools).
  Because of this, I chose to alter the generated CSV rather than
  generating a CSV using env variables to specify master branch
  versions.

Signed-off-by: Elise Gafford <egafford@redhat.com>

### DIFF
--- a/openshift-ci/Dockerfile.registry.build.master
+++ b/openshift-ci/Dockerfile.registry.build.master
@@ -1,0 +1,24 @@
+FROM registry.access.redhat.com/ubi8/ubi-minimal:latest AS stage
+
+COPY . /ocs-operator
+
+WORKDIR /ocs-operator
+
+RUN microdnf install -y jq
+RUN hack/generate-master-appregistry.sh
+
+
+FROM quay.io/openshift/origin-operator-registry:latest
+
+COPY --from=stage /ocs-operator/build/_output/appregistry/olm-catalog /registry/ocs-catalog
+
+# replaces ocs-operator image with the one built by openshift ci
+RUN find /registry/ocs-catalog/ -type f -exec sed -i "s|image\: .*/ocs-operator:.*$|image: default-route-openshift-image-registry.apps.build02.ci.devcluster.openshift.com/${OPENSHIFT_BUILD_NAMESPACE}/stable:ocs-operator|g" {} \; || :
+
+# Initialize the database
+RUN initializer --manifests /registry/ocs-catalog --output bundles.db
+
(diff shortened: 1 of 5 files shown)


egafford:
This patch cannot have caused e2e test failures; other than the Makefile it only adds new files which are not involved in current e2e tests.

openshift-ci/Dockerfile.registry.build.master
```diff
@@ -0,0 +1,24 @@
...
+RUN microdnf install -y jq
+RUN hack/generate-master-appregistry.sh
+
+
+FROM quay.io/openshift/origin-operator-registry:latest
+
+COPY --from=stage /ocs-operator/build/_output/appregistry/olm-catalog /registry/ocs-catalog
+
+# replaces ocs-operator image with the one built by openshift ci
+RUN find /registry/ocs-catalog/ -type f -exec sed -i "s|image\: .*/ocs-operator:.*$|image: default-route-openshift-image-registry.apps.build02.ci.devcluster.openshift.com/${OPENSHIFT_BUILD_NAMESPACE}/stable:ocs-operator|g" {} \; || :
```
jarrpa, line 16: A note on this: The CI team would REALLY prefer we don't do this, but I don't think they've managed to come up with a way to not require us to use the hardcoded URL. Their preference would probably for us to not build an appregistry container anymore and just use bundles, since they have worked out automation for building index images of bundles specified in the repo. I'm fine with merging this for now, but if we can find a way to do without it later that would be ideal.

jarrpa (approved):
There are a few forward-looking things to consider, but they do not need to block this PR. Since we can't filly test this out until we have accompanying CI changes, best to get this through ASAP.

/lgtm

PR #1058, opened 2021-02-11
storagecluster: Update uninstall error as an events

iamniting:

### DIFF
--- a/controllers/util/events.go
+++ b/controllers/util/events.go
@@ -6,4 +6,7 @@ const (
 
 	// EventReasonValidationFailed is used when the StorageCluster spec validation fails
 	EventReasonValidationFailed = "FailedValidation"
+
+	// EventReasonUninstallPending is used when the StorageCluster uninstall is Pending
+	EventReasonUninstallPending = "UninstallPending"
 )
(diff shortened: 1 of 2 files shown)


iamniting:
@raghavendra-talur This is the shortest way with minimum lines of change I can think of to update the last status of uninstall events. Let me know your thoughts on this.

iamniting:
Depends on https://github.com/openshift/ocs-operator/pull/1060. Pls do not merge this one before #1060

controllers/util/events.go
```diff
@@ -6,4 +6,7 @@ const (
 
 	// EventReasonValidationFailed is used when the StorageCluster spec validation fails
 	EventReasonValidationFailed = "FailedValidation"
+
+	// EventReasonUninstallFailed is used when the StorageCluster uninstall fails
+	EventReasonUninstallFailed = "FailedUninstall"
```
raghavendra-talur, line 11: I think the event is generated in two cases:
1. valid error where we know for sure that it won't progress
2. error where we are just waiting for the uninstall to progress in a sub component.
Unfortunately, you don't get to know the category in the caller.

Therefore, "FailedUninstall" might not be the best wording to use here. How about 
* UninstallInProgress
* UninstallPending

iamniting: I like UninstallPending, Updated the same

raghavendra-talur (approved):

jarrpa (approved):
/lgtm

PR #1060, opened 2021-02-11
storagecluster: Change event type from error to warning

iamniting:
Fix Error via changing event type
"E0210 23:17:48.027136       1 event.go:334] Unsupported event type: 'Error'"

Fix https://bugzilla.redhat.com/show_bug.cgi?id=1913357#c8


raghavendra-talur (approved):

jarrpa (changes requested):
Can you explain why this is needed, namely why EventTypeError was used even though it's invalid?

iamniting:
> Can you explain why this is needed, namely why EventTypeError was used even though it's invalid?

When we have sent the PR before we did not realize this. This came recently as part of a failed QA Bug https://bugzilla.redhat.com/show_bug.cgi?id=1913357#c8

jarrpa (approved):
A bit odd that we didn't catch this ahead of time, but I'm willing to skip a deeper RCA.

/lgtm

PR #1062, opened 2021-02-11
[WIP] StorageCluster: let the server start in a synchronous way

raghavendra-talur:

### DIFF
--- a/controllers/storagecluster/cephcluster_test.go
+++ b/controllers/storagecluster/cephcluster_test.go
@@ -409,7 +409,11 @@ func assertCephClusterKMSConfiguration(t *testing.T, kmsArgs struct {
 
 	// don't start dummy servers for invalid tests
 	if !kmsArgs.failureExpected {
-		startServerAt(kmsArgs.kmsAddress)
+		err := startServerAt(kmsArgs.kmsAddress)
+		if err != nil {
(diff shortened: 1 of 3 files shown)


obnoxxx (changes requested):
Can you explain in the commit message (and PR description) what you want to achieve with the PR? What is the problem to solve? And what is the approach/solution.

Further, the unit tests fail, mostly with the rgw server creation like this:

```
=== RUN   TestOptionalExternalStorageClusterResources
    external_resources_test.go:138: failed to create the server: Could not create a server listen tcp4 127.0.0.1:20854: bind: address already in use
```

jarrpa:
> Can you explain in the commit message (and PR description) what you want to achieve with the PR? What is the problem to solve? And what is the approach/solution.
> 
> Further, the unit tests fail, mostly with the rgw server creation like this:
> 
> ```
> === RUN   TestOptionalExternalStorageClusterResources
>     external_resources_test.go:138: failed to create the server: Could not create a server listen tcp4 127.0.0.1:20854: bind: address already in use
> ```

Yup, we discussed this yesterday. My theory is that this error has been happening all along, we've just been ignoring it. If you look at the original code, we made function call to start goroutines but never did anything with the return value. Taylor said he had an idea for it, so let's wait and see.

PR #1063, opened 2021-02-12
Create a route for Ceph RGW

nbalacha:
This change creates a route for the Ceph RGW service.


jarrpa (changes requested):
Code looks great. Please also split the generated CSV artifacts and the vendor changes into their own respective commits.

controllers/storagecluster/initialization_reconciler_test.go
```diff
@@ -99,7 +100,15 @@ func createUpdateRuntimeObjects(cp *Platform) []runtime.Object {
 		}
 		updateRTObjects = append(updateRTObjects, cos)
 	}
-
+	// Create 'route' only for non-cloud platforms
+	if !avoidObjectStore(cp.platform) {
+		cos := &routev1.Route{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "ocsinit-cephobjectstore",
```
travisn, line 107: Can we do without the `init`?
```suggestion
				Name: "ocs-cephobjectstore",
```

controllers/storagecluster/routes.go
```diff
@@ -0,0 +1,154 @@
...
+// ensureDeleted deletes the CephObjectStores owned by the StorageCluster
+func (obj *ocsCephRGWRoutes) ensureDeleted(r *StorageClusterReconciler, sc *ocsv1.StorageCluster) error {
+	foundRoute := &routev1.Route{}
+	routes, err := r.newCephRGWRoutes(sc)
+	if err != nil {
+		return err
+	}
+
+	for _, route := range routes {
+		err := r.Client.Get(context.TODO(), types.NamespacedName{Name: route.Name, Namespace: sc.Namespace}, foundRoute)
```
travisn, line 55: Instead of `Get()`, `Delete()`, `Get()`, you could just `Delete()` and then ignore the error if `IsNotFound(err)`

controllers/storagecluster/routes.go
```diff
@@ -0,0 +1,154 @@
...
+				return err
+			}
+		case errors.IsNotFound(err):
+			r.Log.Info(fmt.Sprintf("creating route %s", route.Name))
+			err = r.Client.Create(context.TODO(), route)
+			if err != nil {
+				r.Log.Error(err, fmt.Sprintf("failed to create CephObjectStore object: %s", route.Name))
+				return err
+			}
+		}
```
travisn, line 117: If the error is something other that `IsNotFound(err)`, where is that handled? Seems like it should fail the reconcile.

controllers/storagecluster/routes.go
jarrpa: This is the general pattern we're following in other files. The main thing this avoid is sending a `Delete()` is the deletion timestamp is not zero. But, conversely, we are wasting time on a `Get()` every time the deletion timestamp is zero, so it may not be any better. Also, we probably don't gain anything from distinguishing between "Found and not marked for deletion" and "Found and marked for deletion", since we are already committed to deleting by the time we get to this function.

Might not be a bad idea to refactor the other `ensureDeleted()` functions in a later PR.

controllers/storagecluster/routes.go
jarrpa: +1

jarrpa (changes requested):

controllers/storagecluster/routes.go
nbalacha: This would be better done in a separate PR to change the behaviour in all the files.

controllers/storagecluster/initialization_reconciler_test.go
nbalacha: The name of the cephobject store in the cephobjectstore_test.go is ocsinit-cephobjectstore. The test sets the route name to the same name as the cephobjectstore for the tests.

On an actual cluster, the route name will be ocs-storagecluster-cephobjectstore.

controllers/storagecluster/routes.go
nbalacha: This is the same behaviour as that of the types being reconciled.  None of them check for any error type.

This would also be a good candidate for a new PR where the behaviour is modified for all types.

jarrpa (approved):
/lgtm

PR #1066, opened 2021-02-12
Bug 1927885: [Release 4.7] reverting readiness logic to match behavior of 4.6

renan-campos:
With the change to operator-sdk v1, the readiness probe changed from a simple file check to an HTTP get readiness probe.
In the process of accommodating this, the readiness logic in 4.7 changed slightly: [the ocs-operator would report ready even when the StorageCluster was in a progressing phase](https://bugzilla.redhat.com/show_bug.cgi?id=1927885). 

Through reading the code and manually testing, I found this to be the 4.6 readiness behavior:
  When ocs-operator first gets created: 1/1
  When storagecluster creation is in progress: 0/1
  When storagecluster creation completes: 1/1
  When storagecluster deleted: 1/1
  If the storagecluster is deleted while in progress: 0/1
  When storagecluster creation is reinitiated: 0/1

After making the changes, I deployed on an AWS Openshift cluster and manually checked that the behavior matches.

### DIFF
--- a/controllers/storagecluster/readiness.go
+++ b/controllers/storagecluster/readiness.go
@@ -1,0 +1,25 @@
+package storagecluster
+
+import (
+	"errors"
+	"net/http"
+
+	"sigs.k8s.io/controller-runtime/pkg/healthz"
+)
+
+var ready bool
+
+var ReadinessChecker healthz.Checker = func(_ *http.Request) error {
+	if ready {
+		return nil
+	}
+	return errors.New("StorageCluster is not ready yet")
+}
+
+func ReadinessSet() {
+	ready = true
+}
+
+func ReadinessUnset() {
+	ready = false
+}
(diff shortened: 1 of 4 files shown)


controllers/storagecluster/readiness.go
```diff
@@ -0,0 +1,25 @@
+package storagecluster
+
+import (
+	"errors"
+	"net/http"
+
+	"sigs.k8s.io/controller-runtime/pkg/healthz"
+)
+
+var ready bool
```
jarrpa, line 10: Rename this to something more specific, like `operatorReady`.

jarrpa (changes requested):
The function itself makes sense, however there are some fundamental structural changes:

- The changes to the CSV generation should be a separate PR entirely. I know you need it to facilitate your workflow, but it is not related to this fix.
- This PR needs to be done against master, not release-4.7. Once it merges we would then cherrypick it to release-4.7. The master PR also doesn't need a BZ associated.

controllers/storagecluster/readiness.go
renan-campos: Thanks for the feedback, I'll close this PR and reopen with the master branch, leaving out the CSV generation changes. I also changed the global variable name as posted here.

PR #1067, opened 2021-02-12
Bug 1927885: reverted readiness logic to match behavior of 4.6

renan-campos:
With the change to operator-sdk v1, the readiness probe changed from a simple file check to an HTTP get readiness probe.
In the process of accommodating this, the readiness logic in 4.7 changed slightly: [the ocs-operator would report ready even when the StorageCluster was in a progressing phase](https://bugzilla.redhat.com/show_bug.cgi?id=1927885).

Through reading the code and manually testing, I found this to be the 4.6 readiness behavior:
When ocs-operator first gets created: 1/1
When storagecluster creation is in progress: 0/1
When storagecluster creation completes: 1/1
When storagecluster deleted: 1/1
If the storagecluster is deleted while in progress: 0/1
When storagecluster creation is reinitiated: 0/1

After making the changes, I deployed on an AWS Openshift cluster and manually checked that the behavior matches.

### DIFF
--- a/main.go
+++ b/main.go
@@ -45,7 +45,6 @@ import (
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 	ctrl "sigs.k8s.io/controller-runtime"
-	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 	// +kubebuilder:scaffold:imports
 )
@@ -159,10 +158,11 @@ func main() {
 	}
 
 	// Add readiness probe
-	if err := mgr.AddReadyzCheck("readyz", healthz.Ping); err != nil {
+	if err := mgr.AddReadyzCheck("readyz", storagecluster.ReadinessChecker); err != nil {
 		setupLog.Error(err, "unable add a readiness check")
 		os.Exit(1)
 	}
+	storagecluster.ReadinessSet()
 
 	setupLog.Info("starting manager")
 	if err := mgr.Start(ctrl.SetupSignalHandler()); err != nil {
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -159,10 +158,12 @@ func main() {
 	}
 
 	// Add readiness probe
-	if err := mgr.AddReadyzCheck("readyz", healthz.Ping); err != nil {
+	if err := mgr.AddReadyzCheck("readyz", storagecluster.ReadinessChecker); err != nil {
 		setupLog.Error(err, "unable add a readiness check")
 		os.Exit(1)
 	}
+	storagecluster.ReadinessSet()
+	defer storagecluster.ReadinessUnset()
```
jarrpa, line 166: Hmm... I'm not sure this is entirely necessary. Looking to the end of `main()`, the only thing that could happen is that `mgr.Start()` exits, at which point the container process would stop and the Pod would restart.

jarrpa (changes requested):
Yes, good. One small question/comment that I don't have an exact answer to. Also, please add a longer explanation to the commit message, like: 
```
With the change to operator-sdk v1, the readiness probe changed from a simple file check to an HTTP get readiness probe.
In the process of accommodating this, the readiness logic in 4.7 changed such that the ocs-operator would report ready even when the StorageCluster was in a progressing phase.
```
No need to link the BZ in the commit message itself.

jarrpa:
Also, a note: no need to reference a BZ for the master branch PR. Technically doesn't hurt, I suppose, but it's not a procedural requirement.

main.go
renan-campos: That's a good point, calling ReadinessUnset() when the program is closing isn't actually going to do anything since at that point the http server used as the readiness probe will already be down.

jarrpa (approved):
/lgtm

PR #1069, opened 2021-02-12
Bug 1915807: [release-4.7] StorageCluster: generate cephReplicatedSpec to abstract out arbiter values

raghavendra-talur:
StorageCluster: generate cephReplicatedSpec
    
Instead of hardcoding it, we get the value generated using a function.


jarrpa (approved):
/lgtm

PR #1070, opened 2021-02-12
StorageCluster: fix the logic for the minimum number of nodes

raghavendra-talur:
We always want to rely on the internal defaults for minimum number of
nodes. They are 4 for arbiter cluster and 3 for non-arbiter cluster.

In cases where the storage device set has a value greater than the
default we make the assumption that the user wants that many nodes.

Co-authored-by: Michael Adam <obnox@redhat.com>


raghavendra-talur:
@jarrpa @obnoxxx Please take a look.

jarrpa (approved):
/lgtm

PR #1071, opened 2021-02-12
Bug 1927885: [release-4.7] reverted readiness logic to match behavior of 4.6

renan-campos:
With the move to operator-sdk v1, the readiness probe changed from a simple file
check to an HTTP GET call. In the process of accommodating this, the readiness
logic in 4.7 changed slightly: the ocs-operator would report ready even when the
StorageCluster was in a progressing phase.

The expected readiness behavior is as follows:
- When ocs-operator first gets created: 1/1
- When storagecluster creation is in progress: 0/1
- When storagecluster creation completes: 1/1
- When storagecluster is deleted: 1/1
- If the storagecluster is deleted while in progress: 0/1
- When storagecluster creation is reinitiated: 0/1

Backport of #1067

### DIFF
--- a/controllers/storagecluster/readiness.go
+++ b/controllers/storagecluster/readiness.go
@@ -1,0 +1,25 @@
+package storagecluster
+
+import (
+	"errors"
+	"net/http"
+
+	"sigs.k8s.io/controller-runtime/pkg/healthz"
+)
+
+var operatorReady bool
+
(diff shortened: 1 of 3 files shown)


jarrpa:
/retitle Bug 1927885: [release-4.7] reverted readiness logic to match behavior of 4.6

jarrpa (approved):
/lgtm

PR #1072, opened 2021-02-12
Allow metrics to be disabled

jarrpa:
This PR allows the admin to specify the reconciliation strategy for the resources created to export metrics.


controllers/storagecluster/reconcile.go
```diff
@@ -396,14 +396,16 @@ func (r *StorageClusterReconciler) reconcilePhases(
 	// enable metrics exporter at the end of reconcile
 	// this allows storagecluster to be instantiated before
 	// scraping metrics
-	if err := r.enableMetricsExporter(instance); err != nil {
-		r.Log.Error(err, "failed to reconcile metrics exporter")
-		return reconcile.Result{}, err
-	}
+	if instance.Spec.MetricsExporter != nil && ReconcileStrategy(instance.Spec.MetricsExporter.ReconcileStrategy) != ReconcileStrategyIgnore {
```
nb-ohad, line 399: I believe this will change the current default behaviour which is to enable metrics exporter.
After this change when `instance.Spec.MetricsExporter == nil` the first half of the if statement will fail and the if block will be skipped which is the opposite of the desired behaviour

nb-ohad (changes requested):

controllers/storagecluster/reconcile.go
jarrpa: ....ahhhh, yup! Fixed. I hope.

nb-ohad (approved):

umangachapagain (approved):

jarrpa:
@nb-ohad I believe you wanted this for OCS 4.7, yes?

nb-ohad:
Yes, we will skip the backport to 4.6 as it seems that we are going to update the deployer to use 4.7 as soon as it is available

PR #1074, opened 2021-02-12
Bug 1915807: [release-4.7] Enable new topology labels from Kubernetes

raghavendra-talur:
In ocs-operator we have been using two deprecated  topology label prefixes `failure-domain.beta.kubernetes.io` and `failure-domain.kubernetes.io`. 

The recommended label to use is `topology.kubernetes.io`. In this PR, we add support for the new label while also adhering to the following rules:
1. In a new install use only the new label
2. In existing setups, switch to the new label only if all the nodes have a equivalence of the labels i.e if node1 has zone1 set on the old labels then the new label should also have zone1 as its value.
3. In existing setups, if the values for all three labels don't match then we continue to use the old labels.
4. We also start persisting the failure domain key and values in the StorageCluster Status. This will probably make it easy to make a forced switch of label in a later release.

### DIFF
--- a/api/v1/storagecluster_types.go
+++ b/api/v1/storagecluster_types.go
@@ -215,6 +215,16 @@ type StorageClusterStatus struct {
 	// +optional
 	FailureDomain string `json:"failureDomain,omitempty"`
 
+	// FailureDomainKey is the specific key used to find the locations available
+	// under a failure domain. For example topology.kubernetes.io/zone
+	// +optional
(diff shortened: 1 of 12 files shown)


raghavendra-talur:
Created a backport manually as the https://github.com/openshift/ocs-operator/pull/990 isn't backport and was leading to conflicts.

jarrpa (approved):
/lgtm

PR #1075, opened 2021-02-13
storagecluster: add MinimumNodes field

jarrpa:
I think there is benefit to finally adding an field to explicitly specify the minimum number of nodes required for the StorageCluster, for admins that know what they want. This would override the current algorithm that tries to determine the minimum number of nodes based on the replica count of the configured StorageDeviceSets and a few other StorageCluster fields.

Due to continued confusion, we could also consider deprecating the `Replica` field for StorageDeviceSets in favor of just using `Count` to explicitly represent the total number of OSDs. In so doing, we could move to using the MinimumNodes field exclusively, with an implicit default of 3 (or 4 for arbiter mode).

The only potential downside is that this would no longer allow the StorageDeviceSet to represent an estimate of the total *available* space of the StorageCluster (once Ceph replication is taken into account). I've personally never been fond of it anyway, since the estimate is rarely ever going to be close to accurate. This change would have no impact on the OpenShift Console, since they use [hardcoded constants](https://github.com/openshift/console/blob/master/frontend/packages/ceph-storage-plugin/src/constants/common.ts#L16-L17) for their replica calculations anyway.

### DIFF
--- a/api/v1/storagecluster_types.go
+++ b/api/v1/storagecluster_types.go
@@ -65,6 +65,9 @@ type StorageClusterSpec struct {
 	// ArbiterSpec specifies the storage cluster options related to arbiter.
 	// If Arbiter is enabled, ArbiterLocation in the NodeTopologies must be specified.
 	Arbiter ArbiterSpec `json:"arbiter,omitempty"`
+	// MinimumNodes explicitly defines the minimum number of nodes for the
+	// StorageCluster
+	MinimumNodes int `json:"minimumNodes,omitempty"`
 }
 
 // KeyManagementServiceSpec provides a way to enable KMS
--- a/config/crd/bases/ocs.openshift.io_storageclusters.yaml
+++ b/config/crd/bases/ocs.openshift.io_storageclusters.yaml
@@ -398,6 +398,10 @@ spec:
                         type: string
                     type: object
                 type: object
+              minimumNodes:
+                description: MinimumNodes explicitly defines the minimum number of
+                  nodes for the StorageCluster
+                type: integer
               monDataDirHostPath:
                 type: string
               monPVCTemplate:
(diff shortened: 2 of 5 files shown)


iamniting (approved):
like the idea

iamniting:
@raghavendra-talur FYI

obnoxxx (approved):
I like the idea in general. And it does not seem to harm since it's anyway only overriding default behaviour and does not have to be set.

But I think we need to have a more general discussion how we would possibly re-design the whole historically grown replica vs count vs expectation of estimated capacity mess.

raghavendra-talur:
> I like the idea in general. And it does not seem to harm since it's anyway only overriding default behaviour and does not have to be set.
> 
> But I think we need to have a more general discussion how we would possibly re-design the whole historically grown replica vs count vs expectation of estimated capacity mess.

I like the idea too. In addition to the above, we also need to consider the implication of user specifying a value that is too low. For example, if the user specifies 1 node as the minimum then we will end up in a state where not enough OSDs can be scheduled. IIUC, in such scenarios, ocs-operator phase would not be ready because it detected the requirements mismatch. With this change, the phase might be shown as ready. This however can be part of the larger discussion.

For this PR, I would prefer to see a log and/or an event generated where we warn the users if their specified value is lesser than what the algorithm computed.

jarrpa:
> For this PR, I would prefer to see a log and/or an event generated where we warn the users if their specified value is lesser than what the algorithm computed.

I agree something like this would be useful, but a log message would just spam the log every reconciliation. Maybe a new Status field?

raghavendra-talur:
> > For this PR, I would prefer to see a log and/or an event generated where we warn the users if their specified value is lesser than what the algorithm computed.
> 
> I agree something like this would be useful, but a log message would just spam the log every reconciliation. Maybe a new Status field?

Sorry I missed this. Yes, a status condition works too.

PR #1081, opened 2021-02-16
added node affinity to schedule osd on its parent prepare pod FD

kesavanvt:
This PR fixes the differences seen in the ceph osd tree and the actual osd placement 

- Added node affinity for prepare/osd placements thus making the OSD schedule on the same failure domain as of prepare pod

### DIFF
--- a/controllers/storagecluster/cephcluster.go
+++ b/controllers/storagecluster/cephcluster.go
@@ -544,16 +544,19 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster, serverVersion *version.
 					}
 
 					if len(topologyKeyValues) >= replica {
-						// Hard constraints are set in OSD placement for portable volumes with rack failure domain
+						// Node affinity is set in OSD and prepared placement for portable volumes with rack failure
 						// domain as there is no node affinity in PVs. This restricts the movement of OSDs
-						// between failure domain.
+						// between failure domain and OSDs will land on the same failure domain as of prepare pod.
 						if portable && !strings.Contains(topologyKey, "zone") {
-							addStrictFailureDomainTSC(&placement, topologyKey)
-						}
-						// If topologyKey is not host, append additional topology spread constarint to the
-						// default preparePlacement. This serves even distribution at the host level
-						// within a failure domain (zone/rack).
-						if noPreparePlacement {
+							topologyIndex := i % len(topologyKeyValues)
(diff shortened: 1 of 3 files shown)


controllers/storagecluster/cephcluster.go
```diff
@@ -544,16 +544,19 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster, serverVersion *version.
...
-						// between failure domain.
+						// between failure domain and OSDs will land on the same failure domain as of prepare pod.
 						if portable && !strings.Contains(topologyKey, "zone") {
-							addStrictFailureDomainTSC(&placement, topologyKey)
-						}
-						// If topologyKey is not host, append additional topology spread constarint to the
-						// default preparePlacement. This serves even distribution at the host level
-						// within a failure domain (zone/rack).
-						if noPreparePlacement {
+							topologyIndex := i % len(topologyKeyValues)
```
jarrpa, line 551: I think this reintroduces an old problem we were trying to fix with TopologySpreadConstraints. With this, we will never be able to use more than `replica` number of non-zone failure domains, so if the admin goes and configures six racks we will only (by default) ever use the first three. A more ideal solution would be to determine the node the prepare Pod will be scheduled to, but of course that requires the Pod to be scheduled in the first place, and setting the NodeAffinity would have to happen dynamically in Rook-Ceph, not ocs-operator... I'm not sure how to resolve this, but at least raising the problem.

controllers/storagecluster/cephcluster.go
```diff
@@ -544,16 +544,19 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster, serverVersion *version.
...
+							setTopologyForAffinity(&placement, topologyKeyValues[topologyIndex], topologyKey)
+							if noPreparePlacement {
```
jarrpa, line 553: So if we have a defined prepare placement, and it doesn't have a NodeAffinity, isn't it possible the prepare pod will end up scheduled in some other failure domain?

jarrpa (changes requested):

travisn (commented):
Per discussion in chat, Rook really needs to enforce the same failure domain for the OSD daemon pods where the OSD prepare job ran, or else the OSD could have a different failure domain than what is reported in the CRUSH map. Assuming Rook can fix this, I don't believe a change is needed in the OCS operator.

PR #1084, opened 2021-02-17
Bug 1915730: [release-4.7] Create a route for Ceph RGW

nbalacha:
The OCS operator will create a route for the
Ceph RGW service.

Backport of PR #1063


obnoxxx (changes requested):
The last commit, fmt errors, does not exist in master. If anything, it should be squashed with the first comme (code changes), since it removes one line the first patch added.

But the question is : why isn't this commit there in master? Do we need to add it to master, too?

nbalacha:
> The last commit, fmt errors, does not exist in master. If anything, it should be squashed with the first comme (code changes), since it removes one line the first patch added.
> 
> But the question is : why isn't this commit there in master? Do we need to add it to master, too?

The code in the master is significantly different because of patches which refactor the code. Those patches are not part of 4.7 because of which the original backport would not apply.

nbalacha:
> The last commit, fmt errors, does not exist in master. If anything, it should be squashed with the first comme (code changes), since it removes one line the first patch added.

Done.

obnoxxx (approved):
Thanks for the update!

/lgtm

PR #1085, opened 2021-02-17
Bug 1909793: Fix fasttune check for Azure storageclass

nbalacha:
The Azure Storage class does not have a "type" parameter.
Fixed to check for storageaccounttype instead.

### DIFF
--- a/pkg/controller/storagecluster/cephcluster.go
+++ b/pkg/controller/storagecluster/cephcluster.go
@@ -26,7 +26,7 @@ import (
 
 var throttleDiskTypes = []string{"gp2", "io1"}
 
-var throttleFastDiskTypes = []string{"managed-premium"}
+var throttleFastDiskTypes = []string{"StandardSSD_LRS", "Premium_LRS", "UltraSSD_LRS"}
 
 const (
 	// Hardcoding networkProvider to multus and this can be changed later to accomodate other providers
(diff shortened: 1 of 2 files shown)


obnoxxx (changes requested):
all commits should all be squashed into one, as the second two fix the first.

I really wonder how this could have been added in the first place. Haven't we tested the azure part? ... 🤔

jarrpa (changes requested):
Best I know, the changes make sense. The commits need some cleanup, though: All three commits can be merged into one. Normally, yes, we request that new unit tests be added in a separate commit, but changes to the tests so that they don't break as a result of code changes should be part of the same commit, that way you can be sure that when you check out the commit you can still run the tests.

nbalacha:
> Best I know, the changes make sense. The commits need some cleanup, though: All three commits can be merged into one. Normally, yes, we request that new unit tests be added in a separate commit, but changes to the tests so that they don't break as a result of code changes should be part of the same commit, that way you can be sure that when you check out the commit you can still run the tests.

Done.

jarrpa (approved):
/lgtm

obnoxxx:
This is confusing at first. This patch is only on release-4.6 and not in master.

At second sight, this is because the throttling/disktuning changes in master were a whole series of PR, which I originally recommended to backport entirely and not selectively. But it was decided to backport only the PR that was most relevant for Azure disk tuning. As it turns out that PR was apparently buggy, and therefore needed to be fixed.

IMHO the appropriate fix would have been to backport the other master PRs for throttling on top. (Unless of course they are still broken too, in which case we need a master PR in addition to this one, and it should actually have been done first.)

@nbalacha , @crombus , @agarwal-mudit , @jarrpa

PR #1087, opened 2021-02-18
Delete kms resources while uninstalling StorageCluster

aruniiird:
KMS resources, such as ConfigMaps and SecretToken which were created at the time of StorageCluster creation, are lurking around even after we uninstall the StorageCluster.

This PR is to delete those resources at the time of StorageCluster uninstallation.

### DIFF
--- a/controllers/storagecluster/cephcluster.go
+++ b/controllers/storagecluster/cephcluster.go
@@ -98,10 +98,15 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
 	if sc.Spec.ExternalStorage.Enable {
 		cephCluster = newExternalCephCluster(sc, r.images.Ceph, r.monitoringIP, r.monitoringPort)
 	} else {
-		kmsConfigMap, err := getKMSConfigMap(sc, r.Client, reachKMSProvider)
+		kmsConfigMap, err := getKMSConfigMap(sc, r.Client)
 		if err != nil {
 			r.Log.Error(err, "failed to procure KMS config")
 			return err
+		}
+		if kmsConfigMap != nil {
+			if err = reachKMSProvider(kmsConfigMap); err != nil {
+				return err
+			}
 		}
 		cephCluster = newCephCluster(sc, r.images.Ceph, r.nodeCount, r.serverVersion, kmsConfigMap, r.Log)
 	}
--- a/controllers/storagecluster/kms_resources.go
+++ b/controllers/storagecluster/kms_resources.go
@@ -30,18 +30,9 @@ var (
 	}
 )
 
-// kmsConfigMapValidateFunc is a functional type,
-// which validates the given configmap and returns an error if any
-type kmsConfigMapValidateFunc func(*corev1.ConfigMap) error
-
-var (
-	// just to make sure 'reachKMSProvider' function follows the validate function type
-	_ kmsConfigMapValidateFunc = reachKMSProvider
-)
-
 // getKMSConfigMap function try to return a KMS ConfigMap.
 // if 'kmsValidateFunc' function is present it try to validate the retrieved config map.
-func getKMSConfigMap(instance *ocsv1.StorageCluster, client client.Client, kmsValidateFunc kmsConfigMapValidateFunc) (*corev1.ConfigMap, error) {
+func getKMSConfigMap(instance *ocsv1.StorageCluster, client client.Client) (*corev1.ConfigMap, error) {
 	// if 'KMS' is not enabled, nothing to fetch
 	if !instance.Spec.Encryption.KeyManagementService.Enable {
 		return nil, nil
@@ -56,10 +47,6 @@ func getKMSConfigMap(instance *ocsv1.StorageCluster, client client.Client, kmsVa
 	)
 	if err != nil {
 		return nil, err
-	}
-	// if validation function is provided, use it
-	if kmsValidateFunc != nil {
-		err = kmsValidateFunc(&kmsConfigMap)
 	}
 	return &kmsConfigMap, err
 }
--- a/controllers/storagecluster/kms_resources.go
+++ b/controllers/storagecluster/kms_resources.go
@@ -35,7 +35,7 @@ var (
 type kmsConfigMapValidateFunc func(*corev1.ConfigMap) error
 
 var (
-	// just to make sure 'isKMSProviderReachable' function follows the validate function type
+	// just to make sure 'reachKMSProvider' function follows the validate function type
 	_ kmsConfigMapValidateFunc = reachKMSProvider
 )
 
--- a/controllers/storagecluster/kms_resources.go
+++ b/controllers/storagecluster/kms_resources.go
@@ -7,9 +7,11 @@ import (
 
 	ocsv1 "github.com/openshift/ocs-operator/api/v1"
 	corev1 "k8s.io/api/core/v1"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const (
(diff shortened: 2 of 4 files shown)


controllers/storagecluster/uninstall_reconciler.go
```diff
@@ -311,6 +311,12 @@ func (r *StorageClusterReconciler) deleteResources(sc *ocsv1.StorageCluster) err
 		return err
 	}
 
+	kmsRMgr := &kmsResourceManager{}
```
raghavendra-talur, line 314: This doesn't look right to me. Why can't we have the kmsResourceManager in the list of objects between lines 290 and 299?

If the deletion should happen only after ceph resources are deleted, you can add kmsResourceManager after ocsCephBlockPools.

raghavendra-talur (changes requested):

raghavendra-talur:
cc: @iamniting

controllers/storagecluster/uninstall_reconciler.go
iamniting: +1

controllers/storagecluster/kms_resources.go
```diff
@@ -39,6 +39,29 @@ var (
 	_ kmsConfigMapValidateFunc = reachKMSProvider
 )
 
+type kmsResourceManager struct{}
+
+func (kmsRMgr *kmsResourceManager) ensureDeleted(r *StorageClusterReconciler, sc *ocsv1.StorageCluster) error {
```
iamniting, line 44: IMO we should add logs as well for errors and info

controllers/storagecluster/kms_resources.go
```diff
@@ -39,6 +39,29 @@ var (
...
+	// if 'KMS' is not enabled, nothing to delete
+	if !sc.Spec.Encryption.KeyManagementService.Enable {
+		return nil
+	}
+	kmsConfigMap, err := getKMSConfigMap(sc, r.Client, nil)
+	if err != nil {
+		return err
+	}
+	err = r.Client.Delete(context.TODO(), kmsConfigMap)
```
iamniting, line 53: Suppose `kmsconfigMap` is still in deleting phase should we go ahead and delete the `token` if no I think we should check if `kmsconfigMap` gets deleted.

iamniting (changes requested):

controllers/storagecluster/uninstall_reconciler.go
aruniiird: The main reason we cannot add the `kmsResourceManager` into the array is, they are a list of `resourceManager` inerface objects. `kmsResourceManager` has only one method `ensureDeleted()` as we are not creating anything it don't have `ensureCreated()` method.

We could add a dummy `ensureCreated()` function and make `kmsResourceManager` object compliant to the `resourceManager` interface

iamniting: If we are not creating it then who is creating it? Is rook creating it? if yes then we should delete it from the rook itself. In general, we should not delete such objects which we are not creating.

Either we can create a dummy `ensureCreated()` or we can change the `ensureDeleted` to `deleteKms*` which will be a method on the `StorageClusterReconciler` itself.

controllers/storagecluster/kms_resources.go
aruniiird: They are independent, so yes we could delete kmsToken whether any error occurs (or not) during kmsConfigMap deletion... Will make the change...

controllers/storagecluster/kms_resources.go
aruniiird: Then we should change the signature of `ensureDeleted()` and `ensureCreated()` functions and add `logger` parameter as well, right?

iamniting: No, we have a logger in the StorageClusterReconciler

controllers/storagecluster/uninstall_reconciler.go
aruniiird: It is the console UI (while creating StorageCluster) which takes the input from user and creates these KMS ConfigMaps and SecretToken resources. But while removing/uninstalling the StorageCluster, we no longer require these resources and we could safely remove them...

jarrpa: In that case, I say go with Nitin's suggestion to just make an explicit `deleteKms*()` function rather than creating a custom struct.

controllers/storagecluster/kms_resources.go
iamniting: I guess you missed this.

iamniting (changes requested):

controllers/storagecluster/kms_resources.go
aruniiird: Thanks for pointing @iamniting ... Updated

aruniiird:
I have made the needed changes...
@iamniting , @raghavendra-talur , @jarrpa , please take a look...

controllers/storagecluster/kms_resources.go
```diff
@@ -35,10 +37,51 @@ var (
...
+			ensureDeleteErrors = append(ensureDeleteErrors, err)
+		}
+	}
+	// throw the first error happened
+	// TODO: wrap all the errors and return a single error
+	for _, err := range ensureDeleteErrors {
+		if err != nil {
+			return err
+		}
+	}
```
aruniiird, line 80: Is there any (simple) way I can consolidate all the errors and send out a single error? I thought of many complicated approaches... (=
@iamniting , @raghavendra-talur , @jarrpa ^^^

controllers/storagecluster/kms_resources.go
```diff
@@ -35,10 +37,51 @@ var (
...
 )
 
+type kmsResourceManager struct{}
+
+// ensureCreated a dummy function to make 'kmsResourceManager' compliant to 'resourceManager' interface
+func (kmsRMgr *kmsResourceManager) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.StorageCluster) error {
+	return nil
+}
+
+func (kmsRMgr *kmsResourceManager) ensureDeleted(r *StorageClusterReconciler, sc *ocsv1.StorageCluster) error {
```
umangachapagain, line 51: I don't like what we're doing in here. Give me sometime to think about it.

controllers/storagecluster/kms_resources.go
iamniting: As of today, we are not going ahead in any of `ensureDeleted` func's if there is any err. We are returning back from the first error itself. @raghavendra-talur What do you think about the above approach? Where we are deleting everything and consolidate errors if any.

iamniting (approved):
looks good to me

controllers/storagecluster/kms_resources.go
```diff
@@ -35,10 +37,54 @@ var (
 type kmsConfigMapValidateFunc func(*corev1.ConfigMap) error
 
 var (
-	// just to make sure 'isKMSProviderReachable' function follows the validate function type
+	// just to make sure 'reachKMSProvider' function follows the validate function type
 	_ kmsConfigMapValidateFunc = reachKMSProvider
 )
 
+func deleteKMSResoucres(r *StorageClusterReconciler, sc *ocsv1.StorageCluster) error {
```
iamniting, line 44: Can you pls add Uninstall keywords in the logs

iamniting (changes requested):

controllers/storagecluster/kms_resources.go
raghavendra-talur: +1

controllers/storagecluster/kms_resources.go
aruniiird: If `A` depends on `B`, yes, we cannot delete `A` until we delete `B`. But in this case, `A` and `B` are two independent items/resources, that's why I was trying to delete all of them, no matter whether the other one fails (or not).

PS: this was one of the old times irritating case of `Ctrl+A` delete of large sets of files in `some OS-es`, where the `Delete` fails at the first occurrence of any corrupted file and we have to again unselect the corrupted file (one-by-one) and do the process all over again (and again). (=

controllers/storagecluster/kms_resources.go
aruniiird: Done

controllers/storagecluster/kms_resources.go
```diff
@@ -39,6 +41,50 @@ var (
...
+	// if 'KMS' is not enabled, nothing to delete
+	if !sc.Spec.Encryption.KeyManagementService.Enable {
+		return nil
+	}
+	var allGetErrors []error = make([]error, 2)
+	var allRuntimeObjs []runtime.Object = make([]runtime.Object, 2)
+	allRuntimeObjs[0], allGetErrors[0] = getKMSConfigMap(sc, r.Client, nil)
+	allRuntimeObjs[1], allGetErrors[1] = getKMSSecretToken(sc, r.Client)
+	allKMSResourceNames := []string{KMSConfigMapName, KMSTokenSecretName}
```
jarrpa, line 53: Instead of doing all this, it would probably be more idiomatic to use goroutines, channels, and WaitGroups, with an inline struct something like this:

```
kmsResources := []{
  Name string
  Get KMSGetFunc
}{
  {KMSConfigMapName, getKMSConfigMap},
  {KMSTokenSecretName, getKMSSecretToken},
}
```

My syntax may be off. Yes this would mean refactoring `getKMSConfigMap()` to not take a validation function, but I'd want that to happen anyway. Since it only gets called twice and only one of those times actually uses a validation function I think it's a reasonable refactor.

jarrpa (changes requested):

controllers/storagecluster/kms_resources.go
```diff
@@ -39,6 +41,50 @@ var (
 	_ kmsConfigMapValidateFunc = reachKMSProvider
 )
 
+func deleteKMSResoucres(r *StorageClusterReconciler, sc *ocsv1.StorageCluster) error {
```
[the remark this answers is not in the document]
iamniting: +1 I missed this

iamniting (changes requested):

controllers/storagecluster/kms_resources.go
aruniiird: I have implemented with the loop logic itself (not changed to the `go` functions and channels yet)... Please take a look and let me know...

controllers/storagecluster/cephcluster.go
```diff
@@ -98,11 +98,14 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
 	if sc.Spec.ExternalStorage.Enable {
 		cephCluster = newExternalCephCluster(sc, r.images.Ceph, r.monitoringIP, r.monitoringPort)
 	} else {
-		kmsConfigMap, err := getKMSConfigMap(sc, r.Client, reachKMSProvider)
+		kmsConfigMap, err := getKMSConfigMap(sc, r.Client)
 		if err != nil {
 			r.Log.Error(err, "failed to procure KMS config")
 			return err
 		}
+		if err = reachKMSProvider(kmsConfigMap); err != nil {
```
jarrpa, line 106: It seems that in your refactor, you are now trying to run `reachKMSProvider()` on every reconcile, whereas before you only ran it if `kmsConfigMap != nil`. I don't fully understand in what scenario it's okay to have a `nil` ConfigMap pointer with also having returned an error, but this seems to be what's causing your unit test failures.

jarrpa (changes requested):
Overall, I like what I'm seeing, and it is in line with my earlier requested changes. Again, I will stress, if you think this is good enough for now and would rather save the channel refactor for later I will accept this. Please leave a TODO comment here to that effect, and we can also maybe work on a Jira story to capture the broader discussion of using goroutines elsewhere to parallelize resource operations.

That said, the unit test failures seem valid, so please address those.

controllers/storagecluster/cephcluster.go
aruniiird: True, I missed that check. Added the check and TODO comment. @jarrpa , please take a look

jarrpa: :+1:

jarrpa (approved):
Excellent! Changes look good, to the best of my knowledge. Now we just wait for the tests.

/lgtm

PR #1088, opened 2021-02-19
Disable CephObjectStore when the cloud provider is IBMCloud

ambiknai:
Signed-off-by: Ambika Nair ambiknai@in.ibm.com

When the Cloud Provider is IBMCloud and Cos Secret is present, disable deployment of CephObjectStore

### DIFF
--- a/controllers/storagecluster/initialization_reconciler_test.go
+++ b/controllers/storagecluster/initialization_reconciler_test.go
@@ -2,6 +2,7 @@ package storagecluster
 
 import (
 	"context"
+	"os"
 	"testing"
 
 	snapapi "github.com/kubernetes-csi/external-snapshotter/v2/pkg/apis/volumesnapshot/v1beta1"
@@ -145,6 +146,8 @@ func initStorageClusterResourceCreateUpdateTestWithPlatform(
 	result, err := reconciler.Reconcile(request)
 	assert.NoError(t, err)
 	assert.Equal(t, reconcile.Result{}, result)
+	err = os.Setenv("WATCH_NAMESPACE", cr.Namespace)
+	assert.NoError(t, err)
 
 	return t, reconciler, cr, request
 }
--- a/controllers/storagecluster/platform_detection.go
+++ b/controllers/storagecluster/platform_detection.go
@@ -6,9 +6,20 @@ import (
 	"sync"
 
 	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/ocs-operator/controllers/util"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	//IBMCloud COS[Cloud Object Storage] secret name
+	ibmCloudCosSecretName = "ibm-cloud-cos-creds"
+
+	//IBMCloudPlatform with COS Secret
+	IBMCloudCosPlatformType configv1.PlatformType = "IBMCloudCosPlatform"
 )
 
 // AvoidObjectStorePlatforms is a list of all PlatformTypes where CephObjectStores will not be deployed.
@@ -16,6 +27,7 @@ var AvoidObjectStorePlatforms = []configv1.PlatformType{
 	configv1.AWSPlatformType,
 	configv1.GCPPlatformType,
 	configv1.AzurePlatformType,
+	IBMCloudCosPlatformType,
 }
 
 // TuneFastPlatforms is a list of all PlatformTypes where TuneFastDeviceClass has to be set True.
@@ -51,6 +63,16 @@ func (p *Platform) getPlatform(c client.Client) (configv1.PlatformType, error) {
 	}
 
 	p.platform = infrastructure.Status.Platform //nolint:staticcheck
+
+	// if IBMCloudPlatformType check for COS secret in cluster
+	if p.platform == configv1.IBMCloudPlatformType {
+		platformIBM, platErr := getActualIBMPlatformType(c)
+		if platErr != nil {
+			return "", fmt.Errorf("Error checking COS secret in IBMCloud: %v", platErr)
+		}
+		p.platform = platformIBM
+	}
+
 	return p.platform, nil
 }
 
@@ -61,6 +83,40 @@ func avoidObjectStore(p configv1.PlatformType) bool {
 		}
 	}
 	return false
+}
+
+func getActualIBMPlatformType(c client.Client) (configv1.PlatformType, error) {
+	isSecretPresent, secErr := IsCosSecretPresent(c)
+	if secErr != nil {
+		return "", fmt.Errorf("Error checking COS secret in IBMCloud: %v", secErr)
+	}
+	if isSecretPresent {
+		// IsIBMCloud and COS Secret present.
+		// return new CloudProviderType
+		return IBMCloudCosPlatformType, nil
+	}
+	//COS secret is not present in IBMCloudPlatform
+	return configv1.IBMCloudPlatformType, nil
+}
+
+// IsCosSecretPresent checks for ibm-cos-cred secret in the concerned namespace
+// if platform is IBMCloud, enable CephObjectStore only if ibm-cloud-cos-creds secret is not present
+// in the target namespace
+func IsCosSecretPresent(c client.Client) (bool, error) {
+	// TODO: better way to get target namespace
+	ns, nsErr := util.GetWatchNamespace()
+	if nsErr != nil {
+		return false, nsErr
+	}
+	foundSecret := &corev1.Secret{}
+	err := c.Get(context.TODO(), types.NamespacedName{Name: ibmCloudCosSecretName, Namespace: ns}, foundSecret)
+	if err != nil && errors.IsNotFound(err) {
+		return false, nil
+	} else if err != nil {
+		return false, err
+	}
+	// Secret is present.
+	return true, nil
 }
 
 func (r *StorageClusterReconciler) DevicesDefaultToFastForThisPlatform() (bool, error) {
--- a/controllers/storagecluster/storageclasses_test.go
+++ b/controllers/storagecluster/storageclasses_test.go
@@ -12,7 +12,7 @@ import (
 )
 
 var allPlatforms = append(AvoidObjectStorePlatforms,
-	configv1.NonePlatformType, configv1.PlatformType("NonCloudPlatform"))
+	configv1.NonePlatformType, configv1.PlatformType("NonCloudPlatform"), configv1.IBMCloudPlatformType)
 
 func TestStorageClasses(t *testing.T) {
 	for _, eachPlatform := range allPlatforms {


controllers/storagecluster/platform_detection.go
```diff
@@ -62,6 +67,57 @@ func avoidObjectStore(p configv1.PlatformType) bool {
 	}
 	return false
 }
+func isIBMPlatformWithCosSecret(p configv1.PlatformType, namespace string, c client.Client) (bool, error) {
```
jarrpa, line 70: Newline before this, as well as a brief comment. Even though it's not exported, it would be good to know what this is checking and why.

controllers/storagecluster/platform_detection.go
```diff
@@ -62,6 +67,57 @@ func avoidObjectStore(p configv1.PlatformType) bool {
...
+// IsIBMPlatform returns true if this cluster is running on IBM Cloud
+// IBM Satellite clusters are not considered as IBMCloudPlatform
+func IsIBMPlatform(p configv1.PlatformType, c client.Client) (bool, error) {
+	isIBM := true
+	nodeList := &corev1.NodeList{}
+	err := c.List(context.TODO(), nodeList)
+	if err != nil {
+		return isIBM, fmt.Errorf("failed to list kubernetes nodes: %s", err)
+	}
+	// Incase of Satellite, cluster is deployed in user provided infrastructure
```
jarrpa, line 99:
```suggestion
	// In case of Satellite, cluster is deployed in user provided infrastructure
```

controllers/storagecluster/platform_detection.go
```diff
@@ -62,6 +67,57 @@ func avoidObjectStore(p configv1.PlatformType) bool {
...
+	if strings.Contains(nodeList.Items[0].Spec.ProviderID, "/sat-") {
+		isIBM = false
+	}
+
+	return isIBM, nil
+}
+
+// Check for ibm-cos-cred secret in the concerned namespace
```
jarrpa, line 107:
```suggestion
// IsCosSecretPresent checks for ibm-cos-cred secret in the concerned namespace
```

controllers/storagecluster/platform_detection.go
```diff
@@ -3,14 +3,19 @@ package storagecluster
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"strings"
+	"sync"
 )
 
+//IBMCloud secret name
```
jarrpa, line 16: Please explain what `Cos` means here.

jarrpa (changes requested):
Overall this change makes sense. However, is there any reason `isIBMPlatformWithCosSecret()` can't just be called from inside `avoidObjectStore()`? It seems like that would save a lot of effort and make it easier to maintain.

rohan47:
LGTM

controllers/storagecluster/platform_detection.go
ambiknai: I have made the suggested changes. Please could you take a look @jarrpa

controllers/storagecluster/initialization_reconciler_test.go
```diff
@@ -49,7 +50,8 @@ func createStorageCluster(scName, failureDomainName string,
 	return cr
 }
 
-func createUpdateRuntimeObjects(cp *Platform) []runtime.Object {
+func createUpdateRuntimeObjects(cp *Platform, namespace string, c client.Client) []runtime.Object {
```
jarrpa, line 53: Rather than passing in the parameters required for `avoidObjectStore()`, see if you can make a separate call to `avoidObjectStore()` outside this function and just pass in the boolean.

While you're at it, go ahead and group all the objects behind an `if avoidObjectStore()` stanza into one longer `if` statement. We shouldn't need to get the same boolean four times if the value is never going to change. However, please make this change its own commit.

controllers/storagecluster/platform_detection.go
```diff
@@ -54,13 +60,64 @@ func (p *Platform) getPlatform(c client.Client) (configv1.PlatformType, error) {
 	return p.platform, nil
 }
 
-func avoidObjectStore(p configv1.PlatformType) bool {
+func avoidObjectStore(p configv1.PlatformType, namespace string, c client.Client) (bool, error) {
```
jarrpa, line 63: I think we can save ourselves a lot of headache if we make this function a method for `type StorageClusterReconciler`. Both the PlatformType and client.Client come from information in the reconciler. So, instead of taking a PlatformType all you would have to do is pass in a StorageCluster to read the Namespace.

I would request that the refactoring of this function into a method be done as its own commit (without taking a StorageCluster parameter), and then add the IBM changes (with the StorageCluster parameter).

jarrpa (changes requested):
Apologies for the delay, I wanted to take some time to sit back and think about this, and the review got away from me.

Overall I support what this change is trying to accomplish, but it still feels a bit intrusive and inelegant. For the second commit, please explain what problem your changes are fixing. If they are a result of the changes you made, please merge the changes into the same commit as the code that breaks the tests.

controllers/storagecluster/platform_detection.go
ambiknai: @jarrpa Thanks for reviewing the code
Agree that both PlatformType and client.Client come from information in the reconciler. Since we need the platform type in logger, Eg [controllers/storagecluster/cephobjectstores.go]
```
if avoidObjectStore(platform) {
		r.Log.Info(fmt.Sprintf("not creating a CephObjectStore because the platform is '%s'", platform))
		return nil
	}
```
I would return `platform` as well from `avoidObjectStore` method in addition to the bool and error.
In short  fetching the platform part will also be moved to `avoidObjectStore` method so that the code is not repeated in `cephobjectstores.go` file as well as `platform_detection.go` file
```
platform, err := r.platform.GetPlatform(r.Client)
```
Please suggest if this approach is good to go.

controllers/storagecluster/platform_detection.go
```diff
@@ -54,13 +61,68 @@ func (p *Platform) getPlatform(c client.Client) (configv1.PlatformType, error) {
 	return p.platform, nil
 }
 
-func avoidObjectStore(p configv1.PlatformType) bool {
+func (r *StorageClusterReconciler) avoidObjectStore(instance *ocsv1.StorageCluster) (bool, configv1.PlatformType, error) {
```
jarrpa, line 64: Don't return the PlatformType here. Since it is not always needed (you ignore it a lot, actually), just have whatever is calling this function also get the PlatformType itself.

controllers/storagecluster/cephobjectstores_test.go
```diff
@@ -48,7 +51,8 @@ func assertCephObjectStores(t *testing.T, reconciler StorageClusterReconciler, c
 	err = reconciler.Client.Get(context.TODO(), request.NamespacedName, actualCos)
 	// for any cloud platform, 'cephobjectstore' should not be created
 	// 'Get' should have thrown an error
-	if avoidObjectStore(reconciler.platform.platform) {
+	isavoidObjectStore, _, _ := reconciler.avoidObjectStore(cr)
```
jarrpa, line 54: I see you're mostly interested in checking the err from the Get() call above. What happens if `avoidObjectStore()` fails with an error? This means that we can't trust the value of `isavoidObjectStore` because a value of `false` may not actually be correct, it's just the default. You should check the error returned from `avoidObjectStore()` separately, and return it in all cases if not nil.

jarrpa, line 54: For all instances like this, fix the capitalization:
```suggestion
	isAvoidObjectStore, _, _ := reconciler.avoidObjectStore(cr)
```

jarrpa (changes requested):
My sincerest apologies for the delay.

Code is looking much better, but still some comments to resolve. Also, we don't keep track of changes requested by review separately, so please merge all the review fixes into the original commits. Speaking of, do the changes in the first commit break the unit tests such that they don't pass unless "Fix unit test cases" also exists? If so, merge that one into the first commit as well.

controllers/storagecluster/cephobjectstores.go
```diff
@@ -22,15 +22,14 @@ func (obj *ocsCephObjectStores) ensureCreated(r *StorageClusterReconciler, insta
...
+	isavoidObjectStore, platform, err := r.avoidObjectStore(instance)
 	if err != nil {
 		return err
 	}
-	if avoidObjectStore(platform) {
+	if isavoidObjectStore {
 		r.Log.Info(fmt.Sprintf("not creating a CephObjectStore because the platform is '%s'", platform))
 		return nil
 	}
-
```
jarrpa, line 33: Don't remove this newline.

controllers/storagecluster/platform_detection.go
jarrpa: I don't think this is valid, especially since you just ignore the platform value in multiple places. I think it's fine to duplicate the GetPlatform() line where needed, in this case. and honestly, since you only use the platform for the log message, you can just move the GetPlatform() call to right above the Log.Info() call under the `if` statement.

jarrpa:
And another apology, but I completely forgot about a pre-existing PR. #953 is aiming to do most of the refactor work this PR is trying to accomplish. I think it is a more complete and correct refactor, so you may want to just rebase your work on top of that and focus on just the IBM-specific changes.

jarrpa:
@ambiknai after reviewing the other PR, I think this PR can be made even simpler. I would ask that you remove any refactoring changes, and just focus on adding the IBM functionality to the code that already exists. We will take care of refactoring any additions you make as needed. So, plese update this PR to *only* include the changes to `platform_detection.go` and maybe the one line in `storageclasses_test.go`.

ambiknai:
> @ambiknai after reviewing the other PR, I think this PR can be made even simpler. I would ask that you remove any refactoring changes, and just focus on adding the IBM functionality to the code that already exists. We will take care of refactoring any additions you make as needed. So, please update this PR to _only_ include the changes to `platform_detection.go` and maybe the one line in `storageclasses_test.go`.

@jarrpa  Trying to understand it better, adding IBM functionality without any refactoring would take it back to how my initial commit was for this PR.
`platform_detection.go` file will have below changes
```
func avoidObjectStore(p configv1.PlatformType, namespace string, c client.Client) (bool, error) {
 	for _, platform := range AvoidObjectStorePlatforms {
 		if p == platform {
 			if p == configv1.IBMCloudPlatformType {
 				isSecretPresent := false
 				isIBM, err := IsIBMPlatform(p, c)
 				if err != nil {
 					return false, err
 				}
 				if isIBM {
 					isSecretPresent, err = IsCosSecretPresent(namespace, c)
 					if err != nil {
 						return false, err
 					}
 					if isSecretPresent {
 						return true, nil
 					}
 					// IsIBM but secret not present
 					return false, nil
 				}
 			}
 			// platform is something other than IBMCloud in AvoidObjectStorePlatforms
 			return true, nil
 		}
 	}
 	return false, nil
 }

 // IsIBMPlatform returns true if this cluster is running on IBM Cloud
 // IBM Satellite clusters are not considered as IBMCloudPlatform
 func IsIBMPlatform(p configv1.PlatformType, c client.Client) (bool, error) {
 	isIBM := true
 	nodeList := &corev1.NodeList{}
 	err := c.List(context.TODO(), nodeList)
 	if err != nil {
 		return isIBM, fmt.Errorf("failed to list kubernetes nodes: %s", err)
 	}
 	// In case of Satellite, cluster is deployed in user provided infrastructure
 	if strings.Contains(nodeList.Items[0].Spec.ProviderID, "/sat-") {
 		isIBM = false
 	}

 	return isIBM, nil
 }

 // IsCosSecretPresent checks for ibm-cos-cred secret in the concerned namespace
 // if platform is IBMCloud, enable CephObjectStore only if ibm-cloud-cos-creds secret is not present
 // in the target namespace
 func IsCosSecretPresent(namespace string, c client.Client) (bool, error) {
 	foundSecret := &corev1.Secret{}
 	err := c.Get(context.TODO(), types.NamespacedName{Name: ibmCloudCosSecretName, Namespace: namespace}, foundSecret)
 	if err != nil && errors.IsNotFound(err) {
 		return false, nil
 	} else if err != nil {
 		return false, err
 	}
 	// Secret is present.
 	return true, nil
 }
 ```
 Hence impacted files would be Few test files + routes.go,cephobjectstoreusers.go ,cephobjectstores.go  to pass namespace and client to `avoidObjectStore` method.

**Comment by jarrpa** [comment]
Yup yup, unfortunately that's the suggestion. PR #953 is an older PR I lost track of, and the refactor there is headed in a direction that's making more sense.

File: controllers/storagecluster/platform_detection.go [hunk 10]
```diff
@@ -35,6 +46,13 @@ type Platform struct {
 func (p *Platform) GetPlatform(c client.Client) (configv1.PlatformType, error) {
 	// if 'platform' is already set just return it
 	if p.platform != "" {
+		if p.platform == configv1.IBMCloudPlatformType {
+			platformIBM, platErr := getActualIBMPlatformType(c)
+			if platErr != nil {
+				return "", fmt.Errorf("Error checking COS secret in IBMCloud: %v", platErr)
+			}
+			p.platform = platformIBM
+		}
```
jarrpa, line 55: I believe you don't need this here, since anything set in `getPlatform()` below will persist here, and I am generally assuming that the underlying platform will not change without restarting the ocs-operator Pod.

controllers/storagecluster/platform_detection.go
```diff
@@ -63,6 +91,39 @@ func avoidObjectStore(p configv1.PlatformType) bool {
...
+	}
+	//COS secret is not present in IBMCloudPlatform
+	return configv1.IBMCloudPlatformType, nil
+}
+
+// IsCosSecretPresent checks for ibm-cos-cred secret in the concerned namespace
+// if platform is IBMCloud, enable CephObjectStore only if ibm-cloud-cos-creds secret is not present
+// in the target namespace
+func IsCosSecretPresent(c client.Client) (bool, error) {
+	ns, nsErr := util.GetWatchNamespace()
```
jarrpa, line 112: Add a `TODO:` comment above this line to indicate we should try and not do this if possible.

jarrpa (changes requested):
Yes, this looks MUCH better! Thank you for all your hard work. Just a little bit more to go:

* Please cut down the commit message to just `Signed-off-by: Ambika Nair <ambiknai@in.ibm.com>`. In general we don't need to keep track of the full development process of the PR, just adequately explain the results. In this case, the current commit subject seems sufficient.

controllers/storagecluster/platform_detection.go
ambiknai: When I tested the change, it was returning from `if p.platform != ""` condition. So had to add the check here

 log results
 ```
 {"level":"info","ts":1615528098.4257863,"logger":"controllers.StorageCluster","msg":"platform already set Inside CephObjectStore:'IBMCloud'","Request.Namespace":"openshift-storage","Request.Name":"ocs-storagecluster"}
{"level":"info","ts":1615528098.425803,"logger":"controllers.StorageCluster","msg":"Inside CephObjectStore:'IBMCloud'","Request.Namespace":"openshift-storage","Request.Name":"ocs-storagecluster"}
{"level":"info","ts":1615528098.5261004,"logger":"controllers.StorageCluster","msg":"creating CephObjectStore ocs-storagecluster-cephobjectstore","Request.Namespace":"openshift-storage","Request.Name":"ocs-storagecluster"}
{"level":"info","ts":1615528098.6885338,"logger":"controllers.StorageCluster","msg":"Creating cephObjectStoreUser ocs-storagecluster-cephobjectstoreuser","Request.Namespace":"openshift-storage","Request.Name":"ocs-
```

File: controllers/storagecluster/initialization_reconciler_test.go [hunk 12]
```diff
@@ -22,9 +20,11 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/version"
+	"os"
```
jarrpa, line 23: Add this line in the block above with the other golang imports.

controllers/storagecluster/initialization_reconciler_test.go
```diff
@@ -22,9 +20,11 @@ import (
...
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"testing"
```
jarrpa, line 27: Don't move this down here, keep it above.

controllers/storagecluster/platform_detection.go
```diff
@@ -3,19 +3,30 @@ package storagecluster
...
-	"sync"
-
 	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/ocs-operator/controllers/util"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sync"
```
jarrpa, line 13: Don't move this down, keep it above.

jarrpa (changes requested):

jarrpa (approved):
/lgtm

agarwal-mudit:
@jarrpa This is not yet required for 4.6, please don't backport now.

agarwal-mudit:
@jarrpa This is approved for 4.6 now, we need to manually backport it because of the conflict.

agarwal-mudit:
Above BP is failing because of the code restructuring we did in 4.7, so please check if that is also required in 4.6 or we can just manually make the changes.
@jarrpa

PR #1089, opened 2021-02-22
Fix camelCase for product title in quickstart

rexagod:
Use camelCase for "OpenShift Container Storage".


controllers/storagecluster/quickstart_constants.go
```diff
@@ -9,7 +9,7 @@ metadata:
   name: "getting-started-ocs"
 spec:
   version: 4.7
-  displayName: "Getting started with Openshift Container Storage"
+  displayName: "Getting started with OpenShift® Container Storage"
```
umangachapagain, line 12: We just use `OpenShift Container Storage` everywhere. Why is this different?

umangachapagain (changes requested):
Commit message can be updated to explain the change that is done and the reason for doing it instead of just linking a Bugzilla.

umangachapagain (commented):
@rexagod PR to master does not really require a bugzilla. We require it only for backports.

controllers/storagecluster/quickstart_constants.go
rexagod: Fixed.

vbnrh: @umangachapagain probably we should keep the (R) for consistency across the quickstarts. The install tour has it in its displayName as UXD wanted it there (for reasons unknown to me) . Either we change it here or we'll have to remove it in console-operator

agarwal-mudit:
/retitle Non-unified reference to OCS in quickstarts

agarwal-mudit:
Have removed BZ number from the title else bot will mark the Bug in MODIFIED state once the fix is merged.
Please use the BZ number while backporting it to 4.7

umangachapagain:
> Commit message can be updated to explain the change that is done and the reason for doing it instead of just linking a Bugzilla.

@rexagod Seems like you missed this. Rest of it looks good.

rexagod:
My bad, I thought you were referring to the issue body (https://github.com/openshift/ocs-operator/pull/1089#issue-577642506) so I updated that instead. Fixing ASAP.

rexagod:
@umangachapagain PTAL.

jarrpa:
The BZ should also be removed form the commit message. Also, the title shouldn't be quite so vague for such a small change. Considering changing the commit message to something like `Fix camelCase for product title in quickstart`

jarrpa (approved):
/lgtm

agarwal-mudit:
@rexagod Please cherry-pick to release-4.7 You can do that automatically by saying /cherry-pick release-4.7

PR #1090, opened 2021-02-23
Add a pkg custom events and start using it for storage cluster

iamniting:
The PR contains a new pkg `customevents` which will allow an operator to limit the events and stop spamming the APIServer.

cc @raghavendra-talur @rohantmp 

PS: I have added this as a new pkg because we need this at the rook layer also.

### DIFF
--- a/controllers/storagecluster/reconcile.go
+++ b/controllers/storagecluster/reconcile.go
@@ -205,7 +205,7 @@ func (r *StorageClusterReconciler) initializeImagesStatus(sc *ocsv1.StorageClust
 func (r *StorageClusterReconciler) validateStorageClusterSpec(instance *ocsv1.StorageCluster, request reconcile.Request) error {
 	if err := versionCheck(instance, r.Log); err != nil {
 		r.Log.Error(err, "Failed to validate version")
-		r.recorder.ReportIfNotPresent(instance, statusutil.EventTypeWarning, statusutil.EventReasonValidationFailed, err.Error())
+		r.recorder.ReportIfNotPresent(instance, corev1.EventTypeWarning, statusutil.EventReasonValidationFailed, err.Error())
 		instance.Status.Phase = statusutil.PhaseError
 		if updateErr := r.Client.Status().Update(context.TODO(), instance); updateErr != nil {
 			return updateErr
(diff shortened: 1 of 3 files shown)


iamniting:
Here is an example output

```
Events:
  Type     Reason            Age                     From                       Message
  ----     ------            ----                    ----                       -------
  Warning  UninstallPending  5m2s                    controller_storagecluster  Uninstall: Waiting on NooBaa system noobaa to be deleted
  Warning  UninstallPending  4m23s (x10 over 4m58s)  controller_storagecluster  Uninstall: Waiting for cephCluster to be deleted

```

```
Events:
  Type     Reason            Age                 From                       Message
  ----     ------            ----                ----                       -------
  Warning  UninstallPending  5m13s               controller_storagecluster  Uninstall: Waiting on NooBaa system noobaa to be deleted
  Warning  UninstallPending  1s (x13 over 5m9s)  controller_storagecluster  Uninstall: Waiting for cephCluster to be deleted
```

It is generating the same events 10 times within 5 minutes and a new cycle goes on after every 5 minutes. 

PS: time and count are configurable

pkg/customevents/events.go
```diff
@@ -0,0 +1,78 @@
...
+
+	// report events x times where x is count
+	count int
+
+	// report events after x minutes
+	eventReportAfterMinutes int
+}
+
+// NewEventReporter returns EventReporter object
+func NewEventReporter(recorder record.EventRecorder, count, reportAfter int) *EventReporter {
```
raghavendra-talur, line 26: If possible think of expanding `count` to `maxCount` or something similar to make it known to the user that it is the max repetition for a particular event.

jarrpa:
When you say "we need this at the rook layer also", do you mean in the rook/rook repo? If that's the case then I hope you're not planning on importing this into rook/rook from this repo...

rohantmp:
> importing this into rook/rook from this repo

Beat me to it. ocs-operator does not really have versions beyond the release branches, and since we import rook, the versioning will be hell even if we don't run into cyclic dep issues. 

I raised the question of where to put this during the ocs-operator huddle yesterday and since the functionality is relatively small, we can just duplicate it in each repo.

pkg/customevents/events.go
```diff
@@ -0,0 +1,78 @@
+package customevents
```
rohantmp, line 1: eventutil or something else would be more accurate

rohantmp: Oh, I see that some event code already exists in the util package. Maybe this should just add to that?

jarrpa: Yeah, probably best to move it into `util/events.go`.

pkg/customevents/events.go
iamniting: Done

pkg/customevents/events.go
iamniting: Done

iamniting:
> When you say "we need this at the rook layer also", do you mean in the rook/rook repo? If that's the case then I hope you're not planning on importing this into rook/rook from this repo...

Yes I was planning to import from this repo, But it looks like duplicating code is better option.

iamniting:
@jose Can we take a look at this one

jarrpa (approved):
Sorry for the delay in review!

Looks good. Offhand I have an instinct to ask for the first and last commits to be squashed, but I think I can live with this.

/lgtm

iamniting:
> Sorry for the delay in review!

No Issues

PR #1092, opened 2021-02-23
Fix the OSD placement logic when arbiter is enabled

raghavendra-talur:
The logic to determine the location affinity for the OSDs should be
    based on the minimum DeviceSetReplica for a Storage Cluster
    configuration and not on the replica count of the StorageDeviceSet.
    
    TODO: The function to generate the StorageClassDevicesSets should be
    refactored further. This is a short term fix to make it easy to
    backport to other branches.

### DIFF
--- a/controllers/storagecluster/cephcluster_test.go
+++ b/controllers/storagecluster/cephcluster_test.go
@@ -458,3 +458,81 @@ func assertCephClusterKMSConfiguration(t *testing.T, kmsArgs struct {
 	}
 	assert.Equal(t, KMSTokenSecretName, cephCluster.Spec.Security.KeyManagementService.TokenSecretName, fmt.Sprintf("Failed: %q. Expected the token-names tobe same", kmsArgs.testLabel))
 }
+
+func TestStorageClassDeviceSetCreationForArbiter(t *testing.T) {
+
+	sc1 := &api.StorageCluster{}
+	sc1.Spec.Arbiter.Enable = true
+	sc1.Spec.NodeTopologies = &api.NodeTopologyMap{
+		ArbiterLocation: "zone3",
+	}
+	sc1.Spec.StorageDeviceSets = getMockDeviceSets("mock", 1, 4, true)
+	sc1.Status.NodeTopologies = &api.NodeTopologyMap{
+		Labels: map[string]api.TopologyLabelValues{
+			zoneTopologyLabel: []string{
+				"zone1",
+				"zone2",
+			},
+		},
+		ArbiterLocation: "zone3",
+	}
+	sc1.Status.FailureDomain = "zone"
+	sc1.Status.FailureDomainKey = zoneTopologyLabel
+	sc1.Status.FailureDomainValues = []string{"zone1", "zone2"}
+
+	sc2 := &api.StorageCluster{}
+	sc2.Spec.Arbiter.Enable = true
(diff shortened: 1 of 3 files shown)


raghavendra-talur:
cc: @iamniting

controllers/storagecluster/cephcluster_test.go
```diff
@@ -332,39 +374,39 @@ func TestStorageClassDeviceSetCreation(t *testing.T) {
 		deviceSet := c.sc.Spec.StorageDeviceSets[0]
 
 		for i, scds := range actual {
-			assert.Equal(t, fmt.Sprintf("%s-%d", deviceSet.Name, i), scds.Name)
-			assert.Equal(t, deviceSet.Count/3, scds.Count)
-			assert.Equal(t, defaults.DaemonResources["osd"], scds.Resources)
-			assert.Equal(t, deviceSet.DataPVCTemplate, scds.VolumeClaimTemplates[0])
-			assert.Equal(t, true, scds.Portable)
-			assert.Equal(t, c.sc.Spec.Encryption.Enable, scds.Encrypted)
+			assert.Equal(t, fmt.Sprintf("%s-%d", deviceSet.Name, i), scds.Name, c.label)
```
jarrpa, line 377: Don't do the addition of labels in this PR... this is something that needs to be discussed further so we can unify on a common strategy for reporting unit tests.

controllers/storagecluster/cephcluster_test.go
```diff
@@ -332,39 +374,39 @@ func TestStorageClassDeviceSetCreation(t *testing.T) {
...
+				assert.Equal(t, c.lenOfMatchExpression, len(matchExpressions), c.label)
 			}
 			topologyKey := scds.PreparePlacement.TopologySpreadConstraints[0].TopologyKey
 
 			if c.topologyKey == "rack" {
-				assert.Equal(t, defaults.RackTopologyKey, topologyKey)
-			} else if len(c.sc.Status.NodeTopologies.Labels[zoneTopologyLabel]) >= defaults.DeviceSetReplica {
-				assert.Equal(t, zoneTopologyLabel, topologyKey)
+				assert.Equal(t, defaults.RackTopologyKey, topologyKey, c.label)
+			} else if len(c.sc.Status.NodeTopologies.Labels[zoneTopologyLabel]) >= (deviceSet.Count / getReplicasPerFailureDomain(c.sc)) {
```
jarrpa, line 406: This change should be included in the same commit as the code changes, since it is technically required to allow the code commit to pass unit tests.

jarrpa (changes requested):
Hmm... yes, I think this makes sense, though you have some unit test failures to address. Also, some of the test changes should be shuffled around. In particular, the addition of the new test cases (and their associated helper function) should be kept in their own commit at the end.

raghavendra-talur:
@jarrpa @obnoxxx @umangachapagain This is a fix that is required for arbiter placement. However, I see that the same bug exists in the non-arbiter mode too. To illustrate the same I have created #1093 .

If we want the behavior where `A replica count higher than 3 in a non-arbiter setup leads to a placement such that the distribution of the OSDs are equal across the failure domain(zones or racks)` then we can take this fix. However, it will be a change in the behavior compared to the previous builds.

jarrpa (approved):
/lgtm

agarwal-mudit:
@raghavendra-talur Please backport to 4.7

PR #1093, opened 2021-02-24
Add test for OSDs placement when the replica count is higher than 3

raghavendra-talur:


raghavendra-talur:
This test should start passing after merging #1092

raghavendra-talur:
@jarrpa @obnoxxx Please take a look.

PR #1095, opened 2021-02-24
Update rook version to get OSD placement fixes

raghavendra-talur:
We update to a version of rook that has placement fixes for OSD. 

Also, we now use the ceph pacific latest images and the stretch cluster feature has some fixes which aren't available in the previous image we were using. This means that we will have to live with a moving target of ceph daemon.


raghavendra-talur:
@jarrpa Please take a look.

jarrpa (approved):
/lgtm

PR #1098, opened 2021-03-01
rename metrics-exporter spec to monitoring

umangachapagain:
Monitoring spec is more generic and suited to our use case
as we configure more than just a metrics-exporter.

We want monitoring to be fully enabled or fully disabled.
i.e. metrics-exporter, PrometheusRule, ServiceMonitor etc
should all be managed via common parent spec.

It will also enable future enhancements under "monitoring"
without making any breaking changes.


jarrpa (approved):

jarrpa (approved):
100% agree.

/lgtm

PR #1099, opened 2021-03-01
unmanage labels for exporter ServiceMonitor

umangachapagain:
Create ServiceMonitor with default labels but do not reconcile
on label updates. This allows users to add custom labels to
ServiceMonitor which can be used by custom Prometheus instances
for monitoring.

### DIFF
--- a/controllers/storagecluster/exporter.go
+++ b/controllers/storagecluster/exporter.go
@@ -171,8 +171,8 @@ func CreateOrUpdateServiceMonitor(r *StorageClusterReconciler, instance *ocsv1.S
 		}
 		return nil, fmt.Errorf("failed to retrieve metrics exporter servicemonitor %v. %v", namespacedName, err)
 	}
-	serviceMonitor.ResourceVersion = oldSm.ResourceVersion
-	err = r.Client.Update(context.TODO(), serviceMonitor)
+	oldSm.Spec = serviceMonitor.Spec
+	err = r.Client.Update(context.TODO(), oldSm)
 	if err != nil {
 		return nil, fmt.Errorf("failed to update metrics exporter servicemonitor %v. %v", namespacedName, err)
 	}


jarrpa:
Rather than do this, why not add the ability to specify the desired labels in the StorageCluster CR? That way there's not a separate step of modifying the labels after creation. In either case, I would still like to see a new section under `Spec.ManagedResources` for the metrics exporter with a `ReconcileStrategy` as a minimum. This way if you want to keep without the ability to specify labels you could just set it to `init` so it gets created with ignoring updates. Given that the objects we generate are fairly static I don't see this as being much of a problem.

nb-ohad:
@joselsegura 
It seems we can use init. But will it protect us against unintentional deletion of the resource?

In any case, this PR is just a quick workaround to the current limitation. 
The proper way, for a future PR, should be to instruct ocs-operator to add a specific label to all resource monitors, this way the agent will not have to know about all the specific resources that need to be labeled

umangachapagain:
> Ignore the previous comment I deleted. Given that the monitoring resources are fairly static, couldn't we just achieve this with a reconcile strategy of `init`?

We could. But, we don't want to init and forget. We want to reconcile on spec changes only.
Also, this is inline with how we update PrometheusRule somewhere else in the code. So this makes the implementation consistent.

jarrpa:
> @jarrpa
> It seems we can use init. But will it protect us against unintentional deletion of the resource?

Yes, the `init` reconcile strategy will recreate the resource. Of course, the label changes would be lost, but that is true for this PR's current solution as well.

> > Ignore the previous comment I deleted. Given that the monitoring resources are fairly static, couldn't we just achieve this with a reconcile strategy of `init`?
> 
> We could. But, we don't want to init and forget. We want to reconcile on spec changes only.
> Also, this is inline with how we update PrometheusRule somewhere else in the code. So this makes the implementation consistent.

Sure, and that should also be changed. We should be adding fields to the StorageCluster API to properly set and maintain any customization to the resources.

Since this is a workaround that should be fixed properly at a future date, please insert a TODO comment and open a Jira issue to track that future change.

nb-ohad:
@jarrpa  There is already a Jira ticket for that as a story under the "ocs to ocs" epic

jarrpa:
Cool, so we just need the comment in the code.

nb-ohad (approved):

umangachapagain:
> Cool, so we just need the comment in the code.

Do we need to add `todo` for RFE?
This PR is not a workaround. It is how we already handle rest of the monitroring resources.

jarrpa (approved):
Okay, after some offline discussion, it seems there is precedent for only reconciling a managed resource's spec. This seems bad, especially if the labels and/or annotations control functionality of how the operator interacts with the resource or even the behavior of the resource itself. But, given the precedent, I won't be too stubborn about this, we can change it later if anyone actually complains.

/lgtm

PR #1102, opened 2021-03-01
Bug 1931590: [release 4.7] Update rook version to get OSD placement fixes

raghavendra-talur:
We update to a version of rook that has placement fixes for OSD.

Also, we now use the ceph pacific latest images and the stretch cluster feature has some fixes which aren't available in the previous image we were using. This means that we will have to live with a moving target of ceph daemon.

This is a manual cherrypick of #1095 as the auto cherry failed.


raghavendra-talur:
@jarrpa @agarwal-mudit Please take a look.

agarwal-mudit:
@obnoxxx This is another instance where we have to use upstream rook version :(

obnoxxx (approved):
/lgtm

PR #1104, opened 2021-03-04
Enabled block pool tab under ocs operator

GowthamShanmugam:


umangachapagain:
@GowthamShanmugam To make changes to csv, you need to update `tools/csv-merger/csv-merger.go`.
In this case, add the API to the list here https://github.com/openshift/ocs-operator/blob/master/tools/csv-merger/csv-merger.go#L58 in 1 commit and in another commit add changes from `make gen-latest-csv`.

jarrpa (changes requested):
+1 on Umanga's change requests. Is there a requirement coming from somewhere to remove the work "Ceph", and if so can you point me to it or someone I can talk to about it?

GowthamShanmugam:
> @GowthamShanmugam To make changes to csv, you need to update `tools/csv-merger/csv-merger.go`.
> In this case, add the API to the list here https://github.com/openshift/ocs-operator/blob/master/tools/csv-merger/csv-merger.go#L58 in 1 commit and in another commit add changes from `make gen-latest-csv`.

ok let me check this

GowthamShanmugam:
> +1 on Umanga's change requests. Is there a requirement coming from somewhere to remove the work "Ceph", and if so can you point me to it or someone I can talk to about it?

I want to add one more tab under OCS operator to display block pools

pool-list

GowthamShanmugam:
Changes done

umangachapagain (commented):
Looks good. But, why do you want to rename `Ceph Block Pool` to `Block Pools`?

GowthamShanmugam:
> Looks good. But, why do you want to rename `Ceph Block Pool` to `Block Pools`?

Since we are displaying pools under the operator the  page we don't need to use the word ceph, design: https://marvelapp.com/prototype/90010b7/screen/77079049

jarrpa:
> Since we are displaying pools under the operator the page we don't need to use the word ceph, design: https://marvelapp.com/prototype/90010b7/screen/77079049

This is disingenuous. Nowhere on the page does it say "Ceph", so admin won't know that they are dealing with a Ceph construct. The term "block pool" is used widely in the industry, but it has a *very specific meaning* within Ceph. We should not be hiding this from the admin at this point. Please keep the name `CephBlockPool`.

GowthamShanmugam:
> > Since we are displaying pools under the operator the page we don't need to use the word ceph, design: https://marvelapp.com/prototype/90010b7/screen/77079049
> 
> This is disingenuous. Nowhere on the page does it say "Ceph", so admin won't know that they are dealing with a Ceph construct. The term "block pool" is used widely in the industry, but it has a _very specific meaning_ within Ceph. We should not be hiding this from the admin at this point. Please keep the name `CephBlockPool`.

It is a shorter name only, Still, we are keeping kind is CephBlockPool only.  Our entire stack is ceph only, so the name ceph is not adding any value.

jarrpa (approved):
> > This is disingenuous. Nowhere on the page does it say "Ceph", so admin won't know that they are dealing with a Ceph construct. The term "block pool" is used widely in the industry, but it has a _very specific meaning_ within Ceph. We should not be hiding this from the admin at this point. Please keep the name `CephBlockPool`.
> 
> It is a shorter name only, Still, we are keeping kind is CephBlockPool only. Our entire stack is ceph only, so the name ceph is not adding any value.

And in just a few months we will no longer be Ceph only, since we will be integrating IBM. :) But whatever. Let the record show there's no way I'll agree this is correct, but I'm being forced to accept it.

/lgtm

PR #1112, opened 2021-03-10
hack/source-manifests: Do not use quotes

b-ranto:
We cannot use quotes around the noobaa dump csv cmd (olm csv yaml)
because otherwise, these will be treated as a single argument instead of
three arguments and the csv generation fails.


jarrpa (approved):
/lgtm

b-ranto:
Updated to unquote one more line, it was still failing when generating the crds as I later noticed.

b-ranto:
FYI: I was able to produce a downstream build with the patch in the current form.

agarwal-mudit (approved):
/lgtm

PR #1113, opened 2021-03-11
Increase mon failover timeout to 15 min for IBMCloudPlatformType

rohan47:
mon C take time to come up on IBMCloudPlatform which timeouts and Mon d gets created. So increase the mon
failover timeout to 15 min for IBMCloudPlatformType to allow enough time
for mon c to come up.

### DIFF
--- a/controllers/storagecluster/cephcluster.go
+++ b/controllers/storagecluster/cephcluster.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"github.com/go-logr/logr"
+	v1 "github.com/openshift/api/config/v1"
 	objectreferencesv1 "github.com/openshift/custom-resource-status/objectreferences/v1"
 	ocsv1 "github.com/openshift/ocs-operator/api/v1"
 	"github.com/openshift/ocs-operator/controllers/defaults"
@@ -116,9 +117,17 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
 		return err
 	}
 
+	platform, err := r.platform.GetPlatform(r.Client)
+	if err != nil {
+		r.Log.Error(err, "Failed to get platform")
+	} else if platform == v1.IBMCloudPlatformType || platform == IBMCloudCosPlatformType {
+		r.Log.Info(fmt.Sprintf("Increasing Mon failover timeout to 15m for %s", platform))
+		cephCluster.Spec.HealthCheck.DaemonHealth.Monitor.Timeout = "15m"
+	}
+
 	// Check if this CephCluster already exists
 	found := &cephv1.CephCluster{}
-	err := r.Client.Get(context.TODO(), types.NamespacedName{Name: cephCluster.Name, Namespace: cephCluster.Namespace}, found)
+	err = r.Client.Get(context.TODO(), types.NamespacedName{Name: cephCluster.Name, Namespace: cephCluster.Namespace}, found)
 	if err != nil {
 		if errors.IsNotFound(err) {
 			if sc.Spec.ExternalStorage.Enable {
--- a/controllers/storagecluster/cephcluster_test.go
+++ b/controllers/storagecluster/cephcluster_test.go
@@ -6,6 +6,7 @@ import (
 	"testing"
 
 	nbv1 "github.com/noobaa/noobaa-operator/v2/pkg/apis/noobaa/v1alpha1"
+	v1 "github.com/openshift/api/config/v1"
 	conditionsv1 "github.com/openshift/custom-resource-status/conditions/v1"
 	api "github.com/openshift/ocs-operator/api/v1"
 	"github.com/openshift/ocs-operator/controllers/defaults"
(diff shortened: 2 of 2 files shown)


rohan47:
not yet tested on an actual cluster
/hold

rohan47:
Tested the code changes.
/hold

controllers/storagecluster/cephcluster.go
```diff
@@ -116,6 +117,11 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
 		return err
 	}
 
+	if r.platform.platform == v1.IBMCloudPlatformType {
```
jarrpa, line 120: There is a new const being introduced in another IBM PR: https://github.com/openshift/ocs-operator/blob/8c0b66ade6c1187c68cdc8334ee7a9331272b45b/controllers/storagecluster/platform_detection.go#L21 (PR: https://github.com/openshift/ocs-operator/pull/1088)

We should probably wait until that one merges and add the new PlatformType in your if statement.

controllers/storagecluster/cephcluster_test.go
```diff
@@ -102,6 +103,50 @@ func TestEnsureCephCluster(t *testing.T) {
 	}
 }
 
+func TestCephClusterMonTimeout(t *testing.T) {
```
jarrpa, line 106: Since this test is not required to pass the unit tests of the code changes, add the test in a separate commit.

jarrpa (changes requested):

controllers/storagecluster/cephcluster_test.go
rohan47: Done

controllers/storagecluster/cephcluster.go
rohan47: Added.

controllers/storagecluster/cephcluster.go
```diff
@@ -116,6 +117,11 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
 		return err
 	}
 
+	if r.platform.platform == v1.IBMCloudPlatformType || r.platform.platform == IBMCloudCosPlatformType {
```
jarrpa, line 120: Don't reference the variable directly, use the GetPlatform() function.

jarrpa (changes requested):

controllers/storagecluster/cephcluster.go
rohan47: Done

jarrpa (approved):
/lgtm

controllers/storagecluster/cephcluster.go
```diff
@@ -116,9 +117,18 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
 		return err
 	}
 
+	platform, err := r.platform.GetPlatform(r.Client)
+	if err != nil {
+		r.Log.Info("Failed to get platform")
```
sp98, line 122: should be logged as error.

controllers/storagecluster/cephcluster_test.go
```diff
@@ -102,6 +103,53 @@ func TestEnsureCephCluster(t *testing.T) {
...
+	// cases for testing
+	cases := []struct {
+		label    string
+		platform v1.PlatformType
+	}{
+		{
+			label:    "case 1",
```
sp98, line 113: its nice to add more details about the `label` rather than just `case 1`.

controllers/storagecluster/cephcluster.go
sp98: shouldn't we return in case of error ?

sp98 (changes requested):

controllers/storagecluster/cephcluster.go
```diff
@@ -116,9 +117,18 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
...
+	}
+	if platform == v1.IBMCloudPlatformType || platform == IBMCloudCosPlatformType {
+		r.Log.Info("Increasing Mon failover timeout for IBM Cloud Platform")
```
sp98, line 125: nit. would be nice to log both the platform name and `900s` timeout.

controllers/storagecluster/cephcluster_test.go
```diff
@@ -102,6 +103,53 @@ func TestEnsureCephCluster(t *testing.T) {
...
+
+		var obj ocsCephCluster
+		err := obj.ensureCreated(&reconciler, sc)
+		assert.NoError(t, err)
+
+		cc := newCephCluster(sc, "", 3, reconciler.serverVersion, nil, log)
+		err = reconciler.Client.Get(context.TODO(), mockCephClusterNamespacedName, cc)
+		assert.NoError(t, err)
+		if c.platform == v1.IBMCloudPlatformType || c.platform == IBMCloudCosPlatformType {
+			fmt.Println(c.platform)
```
sp98, line 145: is the print statement needed here? if yes, then we can add some more context to it. 
say, 
`fmt.Println("Platform:", c.platform)`

sp98 (commented):
nit

controllers/storagecluster/cephcluster.go
rohan47: We are changing the mon failover timeout for IBM platform only, but if it fails I don't think it should be a problem.
So maybe instead of just logging it, Throwing a warning would be better.

sp98: ok. so if we don't want to return error here, then warning should be fine and next if block should be skipped.

controllers/storagecluster/cephcluster_test.go
rohan47: was using this to debug and forgot to remove.
Its not required

controllers/storagecluster/cephcluster.go
rohan47: Logging the error

controllers/storagecluster/cephcluster.go
```diff
@@ -116,9 +117,17 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
 		return err
 	}
 
+	platform, err := r.platform.GetPlatform(r.Client)
+	if err != nil {
+		r.Log.Error(err, "Failed to get platform")
+	} else if platform == v1.IBMCloudPlatformType || platform == IBMCloudCosPlatformType {
+		r.Log.Info("Increasing Mon failover timeout to 900s for %s", platform)
+		cephCluster.Spec.HealthCheck.DaemonHealth.Monitor.Timeout = "900s"
```
travisn, line 125: For readability you could set `15m`, but the result will be the same.
```suggestion
		cephCluster.Spec.HealthCheck.DaemonHealth.Monitor.Timeout = "15m"
```

controllers/storagecluster/cephcluster_test.go
```diff
@@ -102,6 +103,52 @@ func TestEnsureCephCluster(t *testing.T) {
...
 }
 
+func TestCephClusterMonTimeout(t *testing.T) {
+	// cases for testing
+	cases := []struct {
+		label    string
+		platform v1.PlatformType
+	}{
+		{
+			label:    "case 1", // when the platfor is not identified
```
sp98, line 113:
```suggestion
			label:    "case 1", // when the platform is not identified
```

controllers/storagecluster/cephcluster.go
rohan47: done

controllers/storagecluster/cephcluster_test.go
rohan47: Done

jarrpa (approved):
/lgtm

agarwal-mudit:
@rohan47 I think we can remove DNM label now? Its ready to merge, right?

controllers/storagecluster/cephcluster.go
```diff
@@ -116,9 +117,17 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
 		return err
 	}
 
+	platform, err := r.platform.GetPlatform(r.Client)
+	if err != nil {
+		r.Log.Error(err, "Failed to get platform")
+	} else if platform == v1.IBMCloudPlatformType || platform == IBMCloudCosPlatformType {
+		r.Log.Info("Increasing Mon failover timeout to 15m for %s", platform)
```
sp98, line 124: Log usage is not correct.  Its `Info(msg string, keysAndValues ...interface{})`.  You can check other examples in OCS for reference.

sp98 (changes requested):

controllers/storagecluster/cephcluster.go
rohan47: Changed

sp98 (approved):

umangachapagain (commented):
/lgtm

PR #1114, opened 2021-03-11
[Ocs extend cluster limit] Implementing an automated way to recover from read-only mode

pkesavap:
Implementing an automated way to recover from read-only mode

When cluster capacity reaches a critical limit, the cluster goes to read-only, and even delete operations are not permitted on the cluster. This job serves as a quick work way to recover from the read-only mode

Signed-off-by: Prajith Kesava Prasad <pkesavap@redhat.com>

### DIFF
--- a/controllers/storagecluster/reconcile.go
+++ b/controllers/storagecluster/reconcile.go
@@ -670,11 +670,52 @@ or if an OSD ID is not found, errors will be generated in the log and no OSDs wo
 	}
 }
 
+func extendClusterTemplate(sc *ocsv1.StorageCluster) *openshiftv1.Template {
+
+	jobTemplateName := "ocs-extend-cluster"
+	extendClusterCommand := []string{
+		"bash",
+		"-c",
+		`full_ratio_value=$(ceph osd dump | grep -m1 full_ratio | awk '{print $2}')
+		  echo $full_ratio_value
+		if [[ $full_ratio_value == "0.85"  ]]; then
+			   ceph osd set-full-ratio 0.87
+		else
+			   ceph osd set-full-ratio 0.85
+		fi
+		`,
+	}
+
+	return &openshiftv1.Template{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      jobTemplateName,
+			Namespace: sc.Namespace,
+		},
+		Parameters: []openshiftv1.Parameter{
+			{
+				Name:        "RECONFIGURE",
+				DisplayName: "Ratio",
+				Required:    false,
+				Description: `
+				 Currently this design does not require any parameter yet,
+				 included for an additional possiblity to implement something like
+				 --increase-limit and --reset-limit`,
+			},
+		},
+		Objects: []runtime.RawExtension{
+			{
+				Object: newExtendClusterJob(sc, jobTemplateName, extendClusterCommand),
+			},
+		},
+	}
+}
+
 // ensureCreated ensures if the osd removal job template exists
 func (obj *ocsJobTemplates) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.StorageCluster) error {
 
 	tempFuncs := []func(*ocsv1.StorageCluster) *openshiftv1.Template{
 		osdCleanUpTemplate,
+		extendClusterTemplate,
 	}
 
 	for _, tempFunc := range tempFuncs {
@@ -694,6 +735,123 @@ func (obj *ocsJobTemplates) ensureCreated(r *StorageClusterReconciler, sc *ocsv1
 // ensureDeleted is dummy func for the ocsJobTemplates
 func (obj *ocsJobTemplates) ensureDeleted(r *StorageClusterReconciler, sc *ocsv1.StorageCluster) error {
 	return nil
+}
+
+func newExtendClusterJob(sc *ocsv1.StorageCluster, jobTemplateName string, cephCommands []string) *batchv1.Job {
+	labels := map[string]string{
+		"app": "ceph-toolbox-job",
+	}
+
+	// Annotation template.alpha.openshift.io/wait-for-ready ensures template readiness
+	annotations := map[string]string{
+		"template.alpha.openshift.io/wait-for-ready": "true",
+	}
+
+	job := &batchv1.Job{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "Job",
+			APIVersion: "batch/v1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        jobTemplateName,
+			Namespace:   sc.Namespace,
+			Labels:      labels,
+			Annotations: annotations,
+		},
+		Spec: batchv1.JobSpec{
+			Template: corev1.PodTemplateSpec{
+				Spec: corev1.PodSpec{
+
+					InitContainers: []corev1.Container{
+						{
+							Name:            "config-init",
+							Image:           "rook/ceph:master",
+							Command:         []string{"/usr/local/bin/toolbox.sh"},
+							Args:            []string{"--skip-watch"},
+							ImagePullPolicy: "IfNotPresent",
+							Env: []corev1.EnvVar{
+								{
+									Name: "ROOK_CEPH_USERNAME",
+									ValueFrom: &corev1.EnvVarSource{
+										SecretKeyRef: &corev1.SecretKeySelector{
+											Key:                  "ceph-username",
+											LocalObjectReference: corev1.LocalObjectReference{Name: "rook-ceph-mon"},
+										},
+									},
+								},
+								{
+									Name: "ROOK_CEPH_SECRET",
+									ValueFrom: &corev1.EnvVarSource{
+										SecretKeyRef: &corev1.SecretKeySelector{
+											Key:                  "ceph-secret",
+											LocalObjectReference: corev1.LocalObjectReference{Name: "rook-ceph-mon"},
+										},
+									},
+								},
+								{
+									Name:  "POD_NAMESPACE",
+									Value: sc.Namespace,
+								},
+							},
+							VolumeMounts: []corev1.VolumeMount{
+								{
+									Name:      "ceph-config",
+									MountPath: "/etc/ceph",
+								},
+								{
+									Name:      "mon-endpoint-volume",
+									MountPath: "/etc/rook",
+								},
+							},
+						},
+					},
+
+					Containers: []corev1.Container{
+						{
+							Name:  "script",
+							Image: os.Getenv("ROOK_CEPH_IMAGE"),
+							VolumeMounts: []corev1.VolumeMount{
+								{
+									Name:      "ceph-config",
+									MountPath: "/etc/ceph",
+									ReadOnly:  true,
+								},
+							},
+							Command: cephCommands,
+						},
+					},
+					RestartPolicy:      corev1.RestartPolicyNever,
+					ServiceAccountName: "default",
+					Volumes: []corev1.Volume{
+						{
+							Name:         "rook-config",
+							VolumeSource: corev1.VolumeSource{EmptyDir: &corev1.EmptyDirVolumeSource{}},
+						},
+						{
+							Name:         "ceph-config",
+							VolumeSource: corev1.VolumeSource{EmptyDir: &corev1.EmptyDirVolumeSource{}},
+						},
+						{
+							Name: "mon-endpoint-volume",
+							VolumeSource: corev1.VolumeSource{
+								ConfigMap: &corev1.ConfigMapVolumeSource{
+									LocalObjectReference: corev1.LocalObjectReference{Name: "rook-ceph-mon-endpoints"},
+									Items: []corev1.KeyToPath{
+										{
+											Key:  "data",
+											Path: "mon-endpoints",
+										},
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+	return job
+
 }
 
 func newosdCleanUpJob(sc *ocsv1.StorageCluster, jobTemplateName string, cephCommands []string) *batchv1.Job {
--- a/controllers/storagecluster/reconcile.go
+++ b/controllers/storagecluster/reconcile.go
@@ -635,21 +635,21 @@ func remove(slice []string, s string) (result []string) {
 	return
 }
 
-// ensureCreated ensures if the osd removal job template exists
-func (obj *ocsJobTemplates) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.StorageCluster) error {
-	osdCleanUpTemplate := &openshiftv1.Template{
+func osdCleanUpTemplate(sc *ocsv1.StorageCluster) *openshiftv1.Template {
+
+	jobTemplateName := "ocs-osd-removal"
+	osdCleanupArgs := []string{
+		"ceph",
+		"osd",
+		"remove",
+		"--osd-ids=${FAILED_OSD_IDS}",
+	}
+	return &openshiftv1.Template{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      "ocs-osd-removal",
+			Name:      jobTemplateName,
 			Namespace: sc.Namespace,
 		},
-	}
-	_, err := controllerutil.CreateOrUpdate(context.TODO(), r.Client, osdCleanUpTemplate, func() error {
-		osdCleanUpTemplate.Objects = []runtime.RawExtension{
-			{
-				Object: newCleanupJob(sc),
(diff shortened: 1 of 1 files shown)


config/manager/kustomization.yaml
```diff
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: ocs-dev/ocs-operator
-  newName: quay.io/ocs-dev/ocs-operator
+  newName: quay.io/pkesavap/ocs-operator
```
gobindadas: I think this need to change to official repository.

gobindadas (commented):
Can you also please add more info in PR commit message like steps to execute this job etc.

controllers/storagecluster/reconcile.go
```diff
@@ -39,6 +39,7 @@ type resourceManager interface {
 
 type ocsCephConfig struct{}
 type ocsJobTemplates struct{}
+type ocsExtendCluster struct{}
```
jarrpa, line 42: I think it would make more sense to include this as part of the `odsJobTemplates.ensureCreated()` function. Should just be able to add the new Template after the current one in the function.

jarrpa (changes requested):
Your commits should all be squashed, and you should modify/add tests to ensure the Template is properly created. Also, you shouldn't bee removing any of the files you're deleting, so make sure you're running `make gen-latest-csv` and `make ocs-operator-ci`.

jarrpa:
@pkesavap Still waiting for updates (and rebase) on this PR.

controllers/storagecluster/reconcile.go
pkesavap: > I think it would make more sense to include this as part of the `odsJobTemplates.ensureCreated()` function. Should just be able to add the new Template after the current one in the function.

Hi @jarrpa 

Since ocsExtendCluster is a different flow from osdcleanupjob() that makes use of  ocsJobTemplates, i thought it would be better to create a new one for my use case of extend-cluster. 

could you elaborate? , I could only use one struct signature per ensureCreated() method., maybe  im missing something :)

pkesavap:
> @pkesavap Still waiting for updates (and rebase) on this PR.

Apologies @jarrpa , missed out on replying :)  , I have updated the thread with a few doubts I'm currently having. If you could suggest any insights, I can implement them and push them all together.

controllers/storagecluster/reconcile.go
jarrpa: The extend cluster Job is still a Template of a Job. There should be no need to differentiate between different the Jobs, and indeed it would make sense to keep all future Job Templates together unless there is a really good reason to do otherwise.

Since the code between the two `ensureCreated` functions are almost identical, you could restructure the `ocsJobTemplates.ensureCreate()` function to loop through a set of defined Job Templates, of which there would presently only be two. You could then have the `newFooJob()` functions return both the Job spec and Template parameters and set those in the mutate function that way.

You would have to first modify the cleanup Job functions so that you can efficiently add in your job. This would be done in multiple commits:
1. Move cleanup Job parameters into `newCleanupJob()` and return both spec and parameters
2. Refactor `ocsJobTemplates.ensureCreate()` to loop through the defined Jobs (currently just one) in a generic way that reuses all the code you're currently duplicating
3. Add in your new Template
4. Add unit tests to validate your new Template

jarrpa (changes requested):

jarrpa (changes requested):
The code is looking better. I would clean up the current flow to iterate through an array of functions rather than two arrays:
```go
func newExtendTemplate() *openshiftv1.Template {
	jobName := "ocs-extend-cluster"
	const cephCmds = `
set -x
full_ratio_value=$(ceph osd dump | grep -m1 full_ratio | awk '{print $2}')
echo $full_ratio_value
if [[ $full_ratio_value == "0.85"  ]]; then
	ceph osd set-full-ratio $(( 0.85 + 0.02 ))
else
	ceph osd set-full-ratio 0.85
fi
	`;

	return &openshiftv1.Template{
			Name:           jobName,
		},
		Parameters: []openshiftv1.Parameter{...},
		Objects: []runtime.RawExtension{
			{
				Object: newCephJob(jobName, cephCmds),
			},
		},
	}
}
```

And then in `ensureCreated()`:
```go
tempFuncs := []func() *openshiftv1.Template {
	newCleanupTemplate,
	newExtendTemplate,
}

for _, tempFunc := range tempFuncs {
	template := tempFunc()
	_, err := controllerutil.CreateOrUpdate(context.TODO(), r.Client, template, func() error {
		template.Namespace = sc.Namespace
		return controllerutil.SetControllerReference(sc, template, r.Scheme)
	})
	...
}
```

In addition, the commits are still a mess and need to be restructured. Please organize them into the following commits:
1. Move cleanup Job parameters into newCleanupJob() and return both spec and parameters
2. Refactor ocsJobTemplates.ensureCreate() to loop through the defined Jobs (currently just one) in a generic way that reuses all the code you're currently duplicating
3. Add in your new Template
4. Add unit tests to validate your new Template

pkesavap (commented):
> ```go
> template
> ```

The template being 

> The code is looking better. I would clean up the current flow to iterate through an array of functions rather than two arrays:
> 
> ```go
> func newExtendTemplate() *openshiftv1.Template {
> 	jobName := "ocs-extend-cluster"
> 	const cephCmds = `
> set -x
> full_ratio_value=$(ceph osd dump | grep -m1 full_ratio | awk '{print $2}')
> echo $full_ratio_value
> if [[ $full_ratio_value == "0.85"  ]]; then
> 	ceph osd set-full-ratio $(( 0.85 + 0.02 ))
> else
> 	ceph osd set-full-ratio 0.85
> fi
> 	`;
> 
> 	return &openshiftv1.Template{
> 		ObjectMeta: metav1.ObjectMeta{
> 			Name:           jobName,
> 		},
> 		Parameters: []openshiftv1.Parameter{...},
> 		Objects: []runtime.RawExtension{
> 			{
> 				Object: newCephJob(template, cephCmds),
> 			},
> 		},
> 	}
> }
> ```
> 
> And then in `ensureCreated()`:
> 
> ```go
> tempFuncs := []func() *openshiftv1.Template {
> 	newCleanupTemplate,
> 	newExtendTemplate,
> }
> 
> for _, tempFunc := range tempFuncs {
> 	template := tempFunc()
> 	_, err := controllerutil.CreateOrUpdate(context.TODO(), r.Client, template, func() error {
> 		template.Namespace = sc.Namespace
> 		return controllerutil.SetControllerReference(sc, template, r.Scheme)
> 	})
> 	...
> }
> ```
> 
> In addition, the commits are still a mess and need to be restructured. Please organize them into the following commits:
> 
> 1. Move cleanup Job parameters into newCleanupJob() and return both spec and parameters
> 2. Refactor ocsJobTemplates.ensureCreate() to loop through the defined Jobs (currently just one) in a generic way that reuses all the code you're currently duplicating
> 3. Add in your new Template
> 4. Add unit tests to validate your new Template

The template Parameter being parsed in  Object: newCephJob(template, cephCmds),  is inside the template construction code,
that is being returned?, so shouldnt ```  Object: newCephJob(template, cephCmds),  ``` be inside controllerutil.CreateOrUpdate(...) method? @jarrpa

pkesavap:
Kindly disregard the previous comment,

My core logic now requires a separate template. I will push, implementing just that

pkesavap:
@jarrpa  I have updated the patch with the following commits:-

Since my business logic now require another job template, i had to add in another template for extendCluster, rather than reuse the template for ocdRemoval.

I have managed to reuse the same ensureCreated() method, implementing them as you suggested by looping through a set of pre-defined templates.

I could not figure out the adding test part, there was not test written for reconcile strategy, Did you mean to write test for function responsible for generating the job template, if that's the case, there is not external changes coming through parameters except the Bash Command which the job needs to run, everything else is static.

If you could point me to the unit test part, if that still needs to be included it would be a great help!

controllers/storagecluster/reconcile.go
```diff
@@ -635,47 +635,71 @@ func remove(slice []string, s string) (result []string) {
...
+func (obj *ocsJobTemplates) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.StorageCluster) error {
+
+	tempFuncs := []func() *openshiftv1.Template{
+		osdCleanUpTemplate,
+	}
+
+	for _, tempFunc := range tempFuncs {
+		template := tempFunc()
+		_, err := controllerutil.CreateOrUpdate(context.TODO(), r.Client, template, func() error {
+			template.Namespace = sc.Namespace
```
jarrpa, line 727: This is a good use of the mutation function that I did not think about. Unfortunately this does not also address the namespace value for the Job in the Template.

Rather than doing this here you should just pass `sc.Namespace` into `tempFunc()` above. I think it makes sense to force all Template functions to take a Namespace parameter.

controllers/storagecluster/reconcile.go
```diff
@@ -635,47 +635,71 @@ func remove(slice []string, s string) (result []string) {
...
+		})
+
+		if err != nil && !errors.IsAlreadyExists(err) {
+			return fmt.Errorf("failed to create Template : %v", err.Error())
 		}
-		return controllerutil.SetControllerReference(sc, osdCleanUpTemplate, r.Scheme)
-	})
-	if err != nil && !errors.IsAlreadyExists(err) {
-		return fmt.Errorf("failed to create Template: %v", err.Error())
+
```
jarrpa, line 734: Don't add this space

controllers/storagecluster/reconcile.go
```diff
@@ -635,47 +635,71 @@ func remove(slice []string, s string) (result []string) {
...
 	}
 	return nil
+
```
jarrpa, line 737: Don't add this space. If anything, add it *above* `return nil`.

controllers/storagecluster/reconcile.go
```diff
@@ -635,47 +635,71 @@ func remove(slice []string, s string) (result []string) {
 	return
 }
 
-// ensureCreated ensures if the osd removal job template exists
-func (obj *ocsJobTemplates) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.StorageCluster) error {
-	osdCleanUpTemplate := &openshiftv1.Template{
+func osdCleanUpTemplate() *openshiftv1.Template {
+
+	NameSpace := "openshift-storage"
```
jarrpa, line 640: Private/local variables should never be capitalized at the beginning. And in Kubernetes, "Namespace" is one word, not two.

controllers/storagecluster/reconcile.go
```diff
@@ -671,11 +671,54 @@ func osdCleanUpTemplate() *openshiftv1.Template {
 	}
 }
 
+func extendClusterTemplate() *openshiftv1.Template {
+
+	NameSpace := "openshift-storage"
+	jobTemplateName := "ocs-extend-cluster"
+	extendClusterCommand := []string{
+
```
jarrpa, line 679: Do you need this empty line?

controllers/storagecluster/reconcile.go
```diff
@@ -671,11 +671,54 @@ func osdCleanUpTemplate() *openshiftv1.Template {
...
+			   ceph osd set-full-ratio 0.87
+		else
+			   ceph osd set-full-ratio 0.85
+		fi
+		`,
+	}
+
+	return &openshiftv1.Template{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "ocs-extend-cluster",
```
jarrpa, line 694: You should probably use `jobTemplateName` here.

controllers/storagecluster/reconcile.go
```diff
@@ -635,47 +635,71 @@ func remove(slice []string, s string) (result []string) {
...
+	jobTemplateName := "ocs-osd-removal"
+	osdCleanupArgs := []string{
+		"ceph",
+		"osd",
+		"remove",
+		"--osd-ids=${FAILED_OSD_IDS}",
+	}
+	return &openshiftv1.Template{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "ocs-osd-removal",
```
jarrpa, line 650: Use `jobTemplateName` here.

controllers/storagecluster/reconcile.go
```diff
@@ -635,47 +635,114 @@ func remove(slice []string, s string) (result []string) {
...
-The parameter OSD IDs needs a comma-separated list of numerical FAILED_OSD_IDs 
-when a single job removes multiple OSDs. 
-If the expected comma-separated format is not used, 
-or an ID cannot be converted to an int, 
-or if an OSD ID is not found, errors will be generated in the log and no OSDs would be removed.`,
+	The parameter OSD IDs needs a comma-separated list of numerical FAILED_OSD_IDs 
+	when a single job removes multiple OSDs. 
+	If the expected comma-separated format is not used, 
+	or an ID cannot be converted to an int, 
+	or if an OSD ID is not found, errors will be generated in the log and no OSDs would be removed.`,
```
jarrpa, line 663: Any reason you are adding this whitespace? If you're going to do it at least indent it all to line up with the `Description` line like you've done below.

jarrpa (changes requested):
Few more points left. In addition, the first commit's message needs some cleanup. Change it to the following:
```
refactor ocsJobTemplates.ensureCreated() to loop through defined jobs
```

jarrpa:
Also, this is still missing unit tests.

controllers/storagecluster/reconcile.go
pkesavap: done

pkesavap:
> Also, this is still missing unit tests.

Hi, @jarrpa could you kindly refer to  [here](https://github.com/openshift/ocs-operator/pull/1114#issuecomment-817609323), w.r.t unit tests

controllers/storagecluster/reconcile.go
pkesavap: this line is above return nil

jarrpa:
> I could not figure out the adding test part, there was not test written for reconcile strategy, Did you mean to write test for function responsible for generating the job template, if that's the case, there is not external changes coming through parameters except the Bash Command which the job needs to run, everything else is static.

You're checking for the consequences of running the Reconcile() function. So you'll be writing a new unit test function to loop over test cases of running Reconcile() and checking that the appropriate artifacts exist. But... you're right, there is no place currently for adding such tests. As such, I'm going to propose something:

I'd like you to move all the Job Template stuff to a new file, call it `job_templates.go`. Move the `ocsJobTemplates` functions, the Template functions, and the Job functions into this new file. Then create `job_templates_test.go` and start adding unit tests there.

In the interest of time, I'm not going to require that all this be done in this PR. The code looks good enough for now. However, please submit a new PR, based on this one, with new commits to move the code to the new file. Once that PR exists I'll approve this one.

/ok-to-test

pkesavap:
> > I could not figure out the adding test part, there was not test written for reconcile strategy, Did you mean to write test for function responsible for generating the job template, if that's the case, there is not external changes coming through parameters except the Bash Command which the job needs to run, everything else is static.
> 
> You're checking for the consequences of running the Reconcile() function. So you'll be writing a new unit test function to loop over test cases of running Reconcile() and checking that the appropriate artifacts exist. But... you're right, there is no place currently for adding such tests. As such, I'm going to propose something:
> 
> I'd like you to move all the Job Template stuff to a new file, call it `job_templates.go`. Move the `ocsJobTemplates` functions, the Template functions, and the Job functions into this new file. Then create `job_templates_test.go` and start adding unit tests there.
> 
> In the interest of time, I'm not going to require that all this be done in this PR. The code looks good enough for now. However, please submit a new PR, based on this one, with new commits to move the code to the new file. Once that PR exists I'll approve this one.
> 
> /ok-to-test

Hi @jarrpa ,
I have created a separate PR for that, I'm yet to add unit tests to validate them, but I have moved them to a separate file as you suggested. I will continue to work and update the patches there.

jarrpa (approved):
Thank you!

/lgtm

obnoxxx:
@egafford , @jarrpa  -  it seems that there is a problem with the new checks `ci-index-dev-master-dependencies` and `ci-index-fixed-dependencies` which are meant for the nightly runs. They shouldn't appear here as `Required`.

jarrpa:
> @egafford , @jarrpa - it seems that there is a problem with the new checks `ci-index-dev-master-dependencies` and `ci-index-fixed-dependencies` which are meant for the nightly runs. They shouldn't appear here as `Required`.

These are autogenerated jobs. The first one is indeed intended for the nightly runs, but the latter is the replacements for the original `ci-index` job. Still, some tweaks are in order.

PR #1119, opened 2021-03-17
Generate ci-operator bundle for master branches of all projects

egafford:
This change allows ci-operator to generate a bundle for
OCS-operator which pulls from the master branches of the
component projects of OCS.

### DIFF
--- a/hack/generate-master-appregistry.sh
+++ b/hack/generate-master-appregistry.sh
@@ -7,25 +7,4 @@ source hack/generate-appregistry.sh
 
 CSV_FILE=build/_output/appregistry/olm-catalog/ocs-operator/${CSV_VERSION}/ocs-operator.v${CSV_VERSION}.clusterserviceversion.yaml
 
-latest_master_version() {
-  URL=https://registry.hub.docker.com/v2/repositories/$1/tags/
-  FILTER=${2:-.}
-  VERSION=$(curl -s -S "$URL" | jq '."results"[]["name"]' | grep "$FILTER" | sort -nr | head -n1 | tr -d '"')
-  echo "$VERSION"
-}
-
-CEPH_PATTERN="\(travisn\|ceph\)\/ceph:.*$"
-CEPH_VERSION=$(latest_master_version ceph/ceph)
-sed -i "s/${CEPH_PATTERN}/ceph\/ceph:${CEPH_VERSION}/" $CSV_FILE
-
-NOOBAA_CORE_PATTERN="noobaa\/noobaa-core:[0-9.-]*$"
-NOOBAA_CORE_VERSION=$(latest_master_version 'noobaa/noobaa-core' master)
-sed -i "s/${NOOBAA_CORE_PATTERN}/noobaa\/noobaa-core:${NOOBAA_CORE_VERSION}/" $CSV_FILE
-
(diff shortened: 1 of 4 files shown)


hack/generate-master-appregistry.sh
```diff
@@ -7,25 +7,4 @@ source hack/generate-appregistry.sh
...
-sed -i "s/${NOOBAA_CORE_PATTERN}/noobaa\/noobaa-core:${NOOBAA_CORE_VERSION}/" $CSV_FILE
-
-NOOBAA_OPERATOR_PATTERN="noobaa\/noobaa-operator:[0-9.-]*$"
-NOOBAA_OPERATOR_VERSION=$(latest_master_version noobaa/noobaa-operator master)
-sed -i "s/${NOOBAA_OPERATOR_PATTERN}/noobaa\/noobaa-operator:${NOOBAA_OPERATOR_VERSION}/" $CSV_FILE
-
-ROOK_PATTERN="rook\/ceph:v[0-9.-]*$"
-ROOK_VERSION="master"
-sed -i "s/${ROOK_PATTERN}/rook\/ceph:${ROOK_VERSION}/" ${CSV_FILE}
+source hack/generate-master-csv.sh
```
jarrpa, line 10: I imagine this file is still here to preserve a Make target the CI uses? Could it be removed in the future?

jarrpa (changes requested):
Overall the organization looks good. Do we want to have new Make targets to trigger these scripts?

hack/generate-master-appregistry.sh
egafford: The CI actually doesn't use a Make target other than functest. As part of the old patch, I added a Make target to build an ocs-registry container with master-of-everything refs. I actually needed to target the bundle files, but leaving the Make target in seemed useful for dev purposes, so I created this generate-master-csv script that both the generate-master-appregistry and generate-master-bundle scripts call (since really the only difference is the CSV_FILE path).

I'm perfectly happy to remove the ocs-registry-master Make target and generate-master-appregistry script now or in the future; they're not called by the CI and really only exist as a developer convenience. Still, their existence seemed like a happy accident to me for the moment, since they may be useful in reproducing issues brought to light by the master CI job.

egafford: @jarrpa In terms of adding a Make target for building the bundle image at master, I'm uncertain that's useful or in keeping with our previous work.

Because the CI job creates bundles directly from Dockerfiles (and adds them to the ci-index automatically), no current Make target builds the main bundle (with set dependencies) from ci-operator/Dockerfile.bundle. Instead, the operator-bundle Make target builds from deploy/Dockerfile. We also still have the ocs-registry-master make target for the purposes of generating a registry to pull from the same tags as Dockerfile.bundle.master does.

We could add make targets for building both ci-operator bundle images, but given the ocs-registry-master and operator-bundle targets, I think we have the utility we'd gain from adding the ci-operator bundle generation targets already. That being said, it is certainly possible to add them. Thoughts?

jarrpa: Ah, okay. Yeah that makes sense, no need to add it then.

jarrpa (approved):
/lgtm

PR #1123, opened 2021-03-22
must-gather: noobaa obc list command not working

rajatsing:
added a workaround which fixes the issue where noobaa obc list command was not working. This is just a workaround and might be updated in the later releases.

### DIFF
--- a/must-gather/collection-scripts/gather_noobaa_resources
+++ b/must-gather/collection-scripts/gather_noobaa_resources
@@ -21,20 +21,20 @@ noobaa_resources+=(backingstore)
 noobaa_resources+=(bucketclass)
 
 noobaa_cli=()
-noobaa_cli+=("obc list")
 noobaa_cli+=("status")
+noobaa_cli+=("obc list")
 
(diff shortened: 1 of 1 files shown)


rajatsing:
I tried various ways like moving the entire for a loop at a different place but I could not really think of anything else in the end and this worked. While this is not the best solution, but at the moment, this worked for me while others don't.
If you have any suggestions, please let me know.

crombus (changes requested):
As we want this to fix in 4.8 so we don’t want to it this to be working urgently. No deadline so far. we can try to fix the exact issue ok? 

now I will give you a hint to look around this. 

try look into shellcheck look into other cmds. see whay are we skipping.

rajatsing:
It worked 
```
[must-gather-nwh5g] POD -----------collecting dump of status
[must-gather-nwh5g] POD -----------collecting dump of obc list
[must-gather-nwh5g] POD collecting dump of noobaa
```

jarrpa (approved):
/lgtm

PR #1128, opened 2021-03-24
add ManagedResources spec for Ceph dashboard

umangachapagain:
Adding spec to enable/disable ceph dashboard and to enable/disable ssl for it.


jarrpa:
So this is just the spec, and the actual implementation will come later?

What is this intending to modify? Is this targeting the CephCluster? If so, it may be time to create a CephCluster section and shove this under that.

umangachapagain:
> So this is just the spec, and the actual implementation will come later?
> 
Missed the implementation while rebasing with conflicts. Added it now.

> What is this intending to modify? Is this targeting the CephCluster? If so, it may be time to create a CephCluster section and shove this under that.
>
It's passed to CephCluster. This should be fine unless we have lots of CephCluster configurations that we can put under it.
Most of it is at the top level.

jarrpa (approved):
> It's passed to CephCluster. This should be fine unless we have lots of CephCluster configurations that we can put under it. Most of it is at the top level.

Yeah, fair. I suppose if we ever add a "CephCluster" section it'll just be to finally say `reconcileStrategy: ignore` and not much else. :stuck_out_tongue: 

/lgtm

PR #1131, opened 2021-03-24
Bug 1942354: [release-4.6]  hack/source-manifests: Do not use quotes

agarwal-mudit:
We cannot use quotes around the noobaa dump csv cmd (olm csv yaml)
because otherwise, these will be treated as a single argument instead of
three arguments and the csv generation fails.

The same goes for the noobaa crds.

(cherry picked from commit ddb49afd2c480a3bdede1b725a3b15c4df49038a)


agarwal-mudit:
Manual cherry-pick of  #1112

obnoxxx (approved):
/lgtm

PR #1132, opened 2021-03-25
must-gather: skip noobaa collection when storagecluster is not present

rajatsing:
now, the noobaa collection will only happen when the storagecluster will be present, otherwise, it would skip the noobaa collection


jarrpa (approved):
/lgtm

PR #1133, opened 2021-03-25
Enable Noobaa KMS configuration only when the cluster-wide encryption is enabled

aruniiird:
Noobaa's root secret key, used for MCG encryption, should only  be
stored in to KMS server only if cluster wide encryption with StorageClass
encryption enabled.
Corrected the test cases accordingly.


aruniiird:
@jarrpa , please take a look..

jarrpa (approved):
/lgtm

PR #1134, opened 2021-03-25
storagecluster: Change the phase when spec is not valid

iamniting:
As of now if spec is not valid we are logging error and recording events
and status.phase is still in the ready or progressing state. which gives
wrong indication to the user. From now onward we are changing the phase
of the storagecluster to get the user attention.

Fix: https://bugzilla.redhat.com/show_bug.cgi?id=1913357#c16
cc: @raghavendra-talur

### DIFF
--- a/controllers/storagecluster/reconcile.go
+++ b/controllers/storagecluster/reconcile.go
@@ -205,6 +205,10 @@ func (r *StorageClusterReconciler) validateStorageClusterSpec(instance *ocsv1.St
 	if err := versionCheck(instance, r.Log); err != nil {
 		r.Log.Error(err, "Failed to validate version")
 		r.recorder.Event(instance, statusutil.EventTypeWarning, statusutil.EventReasonValidationFailed, err.Error())
+		instance.Status.Phase = statusutil.PhaseError
+		if updateErr := r.Client.Update(context.TODO(), instance); updateErr != nil {
+			return updateErr
+		}
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


iamniting:
@jarrpa Can you pls take a look. We need this for 4.7

controllers/storagecluster/reconcile.go
```diff
@@ -205,20 +205,26 @@ func (r *StorageClusterReconciler) validateStorageClusterSpec(instance *ocsv1.St
 	if err := versionCheck(instance, r.Log); err != nil {
 		r.Log.Error(err, "Failed to validate version")
 		r.recorder.Event(instance, statusutil.EventTypeWarning, statusutil.EventReasonValidationFailed, err.Error())
+		instance.Status.Phase = statusutil.PhaseError
+		_ = r.Client.Update(context.TODO(), instance)
```
jarrpa, line 209: Why are you ignoring this error? I generally don't like that.

iamniting: Because anyway we are going to return another error either it fails or succeeds. I am doing my best effort here. Moreover IMO the Spec error is much more important than the update error.

jarrpa: It is important to know all errors that occurred, otherwise we may inhibit proper troubleshooting of unforeseen failure cases. If something like `Update()` fails we probably have much bigger problems than an invalid version. You already logged the `versionCheck()` err and recorded the Event, so you can either return the `Update()` err if it is not `nil` otherwise return the `versionCheck()` err.

iamniting: Done

jarrpa (approved):
/lgtm

PR #1145, opened 2021-04-01
Use curl instead of wget

obnoxxx:
curl seems to be more standard these days.
E.g. on mac it's there by default.


jarrpa (approved):
/lgtm

PR #1147, opened 2021-04-06
Adds noobaa.namespacestores.io to the exposed APIs

vbnrh:
This commit is required for KNIP-1600 and exposes the namespace store CRD for UI to use.


umangachapagain (commented):
I don't think the KNIP or the fact that UI uses it matters. Maybe explain a little about this api and why it's a good idea to expose in commit message?

umangachapagain (approved):

umangachapagain (approved):
/lgtm

PR #1149, opened 2021-04-08
cephcluster: add storageaccounttype for azuredisk provisioner

crombus:
add storageaccounttype for azuredisk provisioner and remove type for azuredisk as azure StorageClass does not have type parameter. fix the test according the changes.

### DIFF
--- a/controllers/storagecluster/cephcluster.go
+++ b/controllers/storagecluster/cephcluster.go
@@ -37,9 +37,9 @@ const (
 )
 
 type knownDiskType struct {
-	speed            diskSpeed
-	provisioner      StorageClassProvisionerType
-	storageClassType string
+	speed                  diskSpeed
+	provisioner            StorageClassProvisionerType
+	storageClassParameters map[string]string
 }
 
 // These are known disk types where we can't correctly detect the type of the
@@ -47,9 +47,11 @@ type knownDiskType struct {
 // This list allows to specify disks from which storage classes to tune for fast
 // or slow disk optimization.
 var knownDiskTypes = []knownDiskType{
-	{diskSpeedSlow, EBS, "gp2"},
-	{diskSpeedSlow, EBS, "io1"},
-	{diskSpeedFast, AzureDisk, "managed-premium"},
+	{speed: diskSpeedSlow, provisioner: EBS, storageClassParameters: map[string]string{"type": "gp2"}},
+	{speed: diskSpeedSlow, provisioner: EBS, storageClassParameters: map[string]string{"type": "io1"}},
+	{speed: diskSpeedFast, provisioner: AzureDisk, storageClassParameters: map[string]string{"storageaccounttype": "StandardSSD_LRS"}},
+	{speed: diskSpeedFast, provisioner: AzureDisk, storageClassParameters: map[string]string{"storageaccounttype": "Premium_LRS"}},
+	{speed: diskSpeedFast, provisioner: AzureDisk, storageClassParameters: map[string]string{"storageaccounttype": "UltraSSD_LRS"}},
 }
 
 const (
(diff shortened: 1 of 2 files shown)


controllers/storagecluster/cephcluster.go
```diff
@@ -694,6 +697,10 @@ func (r *StorageClusterReconciler) checkTuneStorageDevices(ds ocsv1.StorageDevic
 			continue
 		}
 
+		if dt.storageAccountType != storageClass.Parameters["storageaccounttype"] {
+			continue
+		}
+
```
jarrpa, line 703: Part of the problem here is that you're checking for a parameter that doesn't exist. When you try to fetch a value in a map, Go actually returns two values, the value itself and a boolean. The boolean says whether or not the value was found in the map. Thus, with the refactor I suggested, it would look like this:
```
	for _, dt := range knownDiskTypes {
		if string(dt.provisioner) != storageClass.Provisioner {
			continue
		}

		invalidParameters := false
		for parameter, expectedValue := range dt.StorageClassParameters {
			actualValue, ok := storageClass.Parameters[parameter]
			if ok && actualValue != expectedValue {
				invalidParameters = true
			}
		}

		if invalidParameters {
			continue
		}
		return dt.speed, nil
	}
```

jarrpa (changes requested):
Proposed a cleaner refactor. In addition, please update both the commit message and PR description to indicate why this change is needed, namely that Azure StorageClasses don't specify "type".

config/manager/kustomization.yaml
```diff
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: ocs-dev/ocs-operator
-  newName: quay.io/ocs-dev/ocs-operator
+  newName: quay.io/pkundra/ocs-operator
```
obnoxxx, line 7: I don't think that this should be merged in the official code.
This PR should probably be marked Draft or WIP until this is resolved.

deploy/bundle/manifests/ocs-operator.clusterserviceversion.yaml
```diff
@@ -2407,7 +2407,7 @@ spec:
                   value: noobaa/noobaa-core:5.7.0-20201216
                 - name: NOOBAA_DB_IMAGE
                   value: centos/postgresql-12-centos7
-                image: quay.io/ocs-dev/ocs-operator:latest
+                image: quay.io/pkundra/ocs-operator:latest
```
obnoxxx, line 2410: same comment. don't merge with this. I guess this is for your testing?

deploy/bundle/manifests/ocs-operator.clusterserviceversion.yaml
```diff
@@ -2579,7 +2579,7 @@ spec:
                 - --namespaces=openshift-storage
                 command:
                 - /usr/local/bin/metrics-exporter
-                image: quay.io/ocs-dev/ocs-operator:latest
+                image: quay.io/pkundra/ocs-operator:latest
```
obnoxxx, line 2582: one more...

obnoxxx (changes requested):
I think we should revert the references to the personal quay repo.

deploy/bundle/manifests/ocs-operator.clusterserviceversion.yaml
crombus: yes.

config/manager/kustomization.yaml
crombus: testing of the patch is pending. The test are failing.

jarrpa (commented):
@crombus Has this been verified on Azure?

crombus:
> @crombus Has this been verified on Azure?

not yet still blocked by azure platform.

PR #1153, opened 2021-04-16
Refactor: Move job template code to another file and add unit tests

pkesavap:
This PR has been created to move job templates that reside in reconcile.go file to another file and add unit tests to the job templates to validate them
As a follw up part of this PR: https://github.com/openshift/ocs-operator/pull/1114


jarrpa (changes requested):
Good start. Also move `ensureCreated()` and `ensureDeleted()`.

jarrpa:
Please at least rebase this PR, even if you haven't implemented the tests yet.

controllers/storagecluster/job_templates_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+	configv1 "github.com/openshift/api/config/v1"
+	openshiftv1 "github.com/openshift/api/template/v1"
+	api "github.com/openshift/ocs-operator/api/v1"
+	"github.com/stretchr/testify/assert"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+//assert(check if th)
```
jarrpa, line 15: What's this comment about?

controllers/storagecluster/job_templates_test.go
```diff
@@ -0,0 +1,59 @@
...
+var platforms = append(AvoidObjectStorePlatforms,
```
jarrpa, line 16: Avoid generic variable names like this, at least in this context. Since this is part of `package storagecluster`, this could conflict with other test files setting the same thing in a full test run.

That said, why are you looping over a list of platforms? I'm not aware of any special requirements where the Job Templates shouldn't exist on a particular platform, though I may be missing something of course.

jarrpa (changes requested):
I do so love commits like these. :) A great start!

Along with a few specific comments, I'd like to say more generally: we should be trying to compare expected vs. actual structures as completely as possible. Usually, doing a complete field-by-field or DeepEquals() comparison is a bit overkill, so of course there's room for compromise.

The main thing to keep in mind is that we are testing whether the mock system running a full reconcile generates the results we would expect from running our functions-under-test separately. Since we are trying to compare fairly large structs that are mostly static, it is reasonable to mostly look for values that need to be read or generated from input parameters. 

In addition, this is not only to validate the code that is already there, but also to define what completeness means for anyone creating new Job Templates in the future. So we should determine what (few) things would be common across all such Templates and verify those as well.

Thus, a potential starting list of things to validate may be:
* Assert proper Namespace for Template (already covered) and Job (not explicitly covered)
* Assert proper annotations (and labels?)
* The expected number of Templates are created (this would make it easy to tell if someone has not added unit tests for their new Template. :) )

Rather than separate assert functions per template, see if you can't restructure things so that you can have common code that loops over a list of new structs that define the expected values for each expected template. This will further reduce the load on implementing tests for new Templates in the future.

controllers/storagecluster/job_templates_test.go
pkesavap: Done, Initially i was confused, i have fixed it on the new patch now, This does not require any specific restriction on platforms

controllers/storagecluster/job_templates_test.go
pkesavap: Removed. Mindless commenting left unchecked.

pkesavap:
> I do so love commits like these. :) A great start!
> 
> Along with a few specific comments, I'd like to say more generally: we should be trying to compare expected vs. actual structures as completely as possible. Usually, doing a complete field-by-field or DeepEquals() comparison is a bit overkill, so of course there's room for compromise.
> 
> The main thing to keep in mind is that we are testing whether the mock system running a full reconcile generates the results we would expect from running our functions-under-test separately. Since we are trying to compare fairly large structs that are mostly static, it is reasonable to mostly look for values that need to be read or generated from input parameters.
> 
> In addition, this is not only to validate the code that is already there, but also to define what completeness means for anyone creating new Job Templates in the future. So we should determine what (few) things would be common across all such Templates and verify those as well.
> 
> Thus, a potential starting list of things to validate may be:
> 
> * Assert proper Namespace for Template (already covered) and Job (not explicitly covered)
> * Assert proper annotations (and labels?)
> * The expected number of Templates are created (this would make it easy to tell if someone has not added unit tests for their new Template. :) )
> 
> Rather than separate assert functions per template, see if you can't restructure things so that you can have common code that loops over a list of new structs that define the expected values for each expected template. This will further reduce the load on implementing tests for new Templates in the future.

Hi, @jarrpa :-)
So 
Now I have written tests to make sure that the jobs are created by reconciler without any errors and asserted the  jobs, 
I have also added a new method which keeps track of the count of jobs, and raises an error , if a new job does not have proper unit test 


For the annotations and labels, since it was a constant hard coded in both the job functions, i did not add them, but i can add them if they are needed as well.

controllers/storagecluster/job_templates_test.go
```diff
@@ -0,0 +1,135 @@
...
+	err := reconciler.Client.List(context.TODO(), actualJobList)
+	assert.NoError(t, err)
+	for _, node := range actualJobList.Items {
+		_, found := TestedJob[node.ObjectMeta.Name]
+		if found == true {
+			count++
+		}
+
+	}
+	assert.Equalf(t, len(TestedJob), count, "Unit tests are not added for one or more Job templates")
```
jarrpa, line 134: Not an ideal test, but a clever solution. I'll take it!

controllers/storagecluster/job_templates_test.go
```diff
@@ -0,0 +1,135 @@
...
+		   ceph osd set-full-ratio 0.85
+	fi
+	`,
+}
+var osdCleanupArgsTest = []string{
+	"ceph",
+	"osd",
+	"remove",
+	"--osd-ids=${FAILED_OSD_IDS}",
+}
```
jarrpa, line 35: Defining these variables here will lead to problems if they ever get updated in the runtime code. I would suggest defining them in the code and referencing them in both places they're used.

controllers/storagecluster/job_templates_test.go
```diff
@@ -0,0 +1,135 @@
...
+
+	ocsRemovalTemplate := &openshiftv1.Template{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "ocs-osd-removal",
+			Namespace: cr.Namespace,
+		},
+	}
+	request.Name = "ocs-osd-removal"
+	reconciler.Client.Get(context.TODO(), request.NamespacedName, ocsRemovalTemplate)
+	assert.Equal(t, ocsRemovalTemplate.Parameters[0].Name, expectedOcsRemovalParameterName)
```
jarrpa, line 58: A little clunky, it's usually a good idea to avoid using static index references into arrays for testing. But this works for now.

jarrpa (changes requested):
Keeps getting better. Apart from the specific comments, a few overall ones:

1. You are ignoring the errors returned form your `Get()` calls. Don't do that. You should capture the errors and then `assert.NoError()` to verify the thing you're getting actually exists. Same for the `Create()` calls, there's no reason to continue the test if these fail, and may indicate problems in the infrastructure as well.
2. Squash the last two commits.
3. You did not address several things from my previous review:
   * Assert proper Namespace for Template (already covered) and Job (not explicitly covered)
   * Assert proper annotations (and labels?)
   * Rather than separate assert functions per template, see if you can't restructure things so that you can have common code that loops over a list of new structs that define the expected values for each expected template. This will further reduce the load on implementing tests for new Templates in the future.

controllers/storagecluster/job_templates_test.go
pkesavap: You mean defining inside the method where I'm using it? @jarrpa

pkesavap:
> Keeps getting better. Apart from the specific comments, a few overall ones:
> 
> 1. You are ignoring the errors returned form your `Get()` calls. Don't do that. You should capture the errors and then `assert.NoError()` to verify the thing you're getting actually exists. Same for the `Create()` calls, there's no reason to continue the test if these fail, and may indicate problems in the infrastructure as well.
> 2. Squash the last two commits.
> 3. You did not address several things from my previous review:
>    
>    * Assert proper annotations (and labels?)
>    * Rather than separate assert functions per template, see if you can't restructure things so that you can have common code that loops over a list of new structs that define the expected values for each expected template. This will further reduce the load on implementing tests for new Templates in the future.

@jarrpa Perhaps you missed my previous comment about asserting proper annotation and labels.

since Annotations and labels are constants hardcoded in both the job functions, I did not add them, but I can add them if they are needed as well.

I will work on restructuring the method in a much more general way.

umangachapagain (commented):
While we're at it, I'd like to use `embed` from Go 1.16 and take out all templates into separate YAMLs instead of Go structures and use that. We can discuss the details.

controllers/storagecluster/job_templates_test.go
jarrpa: Correct, I suggest not doing that.

jarrpa:
> While we're at it, I'd like to use `embed` from Go 1.16 and take out all templates into separate YAMLs instead of Go structures and use that. We can discuss the details.

...wow, that is GREAT. We should definitely see if we can take advantage of that, but I'd do that in a later PR.

pkesavap:
Hi @jarrpa  i have addressed all of your previous comments, requesting you to review them.

jarrpa (approved):
I restructured the commits, as well as added one of my own at the end to refactor the unit tests into the more agreeable test cases format. I can only imagine I must've been rather confusing after all this time, so it would be good to take some time to get together and see where the miscommunication lies.

/lgtm

PR #1120, opened 2021-03-18
Bug 1926182: [release-4.7]  Support disabling reconciliation of monitoring related resources using a dedicated reconcile strategy flag

agarwal-mudit:
Backport for https://github.com/openshift/ocs-operator/pull/1072 and https://github.com/openshift/ocs-operator/pull/1098

PR #1154, opened 2021-04-19
updated noobaa core\operator images in common.sh

dannyzaken:
* updated the images to point to latest upstream builds
* fixed dump commands in source-manifests.sh
* Create noobaa-endpoint SCC which is required by noobaa endpoints 

Signed-off-by: Danny Zaken <dannyzaken@gmail.com>

### DIFF
--- a/hack/source-manifests.sh
+++ b/hack/source-manifests.sh
@@ -40,12 +40,12 @@ function dump_noobaa_csv() {
 	rm -rf $noobaa_crds_outdir
 	mkdir -p $noobaa_crds_outdir
 
-	echo "Dumping Noobaa csv using command: $IMAGE_RUN_CMD $NOOBAA_IMAGE $noobaa_dump_csv_cmd"
+	echo "Dumping Noobaa csv using command: $IMAGE_RUN_CMD --entrypoint=/usr/local/bin/noobaa-operator $NOOBAA_IMAGE $noobaa_dump_csv_cmd"
 	# shellcheck disable=SC2086
-	($IMAGE_RUN_CMD "$NOOBAA_IMAGE" $noobaa_dump_csv_cmd) > $NOOBAA_CSV
-	echo "Dumping Noobaa crds using command: $IMAGE_RUN_CMD $NOOBAA_IMAGE $noobaa_dump_crds_cmd"
+	($IMAGE_RUN_CMD --entrypoint=/usr/local/bin/noobaa-operator "$NOOBAA_IMAGE" $noobaa_dump_csv_cmd) > $NOOBAA_CSV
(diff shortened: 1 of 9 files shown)


hack/source-manifests.sh
```diff
@@ -33,8 +33,8 @@ mkdir -p $OUTDIR_TOOLS
 
 # ==== DUMP NOOBAA YAMLS ====
 function dump_noobaa_csv() {
-	noobaa_dump_crds_cmd="crd yaml"
-	noobaa_dump_csv_cmd="olm csv yaml"
+	noobaa_dump_crds_cmd="/usr/local/bin/noobaa-operator crd yaml"
```
obnoxxx, line 36: Interesting. Has anything in the noobaa image changed that requires this change?
Just seeing this line change, I am think that this can not have worked before...

dannyzaken: I took another look, and I will update this PR to use newer images.

we had [this PR](https://github.com/noobaa/noobaa-operator/pull/556) that removed the entrypoint. this broke the dump command. 
since then it was fixed [here](https://github.com/noobaa/noobaa-operator/pull/575), but `noobaa-operator:master-20210308` does not contain the fix. 
we pushed `noobaa/noobaa-operator:master-20210419` yesterday, and it does contain the second PR, so I'll use that one instead. 
I think that I'll add `--entrypoint=/usr/local/bin/noobaa-operator` to the command, just to avoid this in the future

obnoxxx:
shellcheck error. unrelated. ==> #1162

obnoxxx:
* update now that #1120 
* there is an error in verify-latest-csv (uncommitted changes)

dannyzaken:
@obnoxxx any idea why verify-latest-csv is failing? I ran `make gen-latest-csv` and committed the changes (3rd commit in this PR). is there anything I'm missing here?
is it possible there is a difference in the output between mac and Linux?

config/crd/bases/ocs.openshift.io_ocsinitializations.yaml
```diff
@@ -1,169 +1,169 @@
 
 ---
-apiVersion: apiextensions.k8s.io/v1
+apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.4.1
+    controller-gen.kubebuilder.io/version: v0.3.0
```
jarrpa, line 7: This is a problem, you're using an outdated version of controller-gen. Try deleting the binary we download inside the repo and running it again.

dannyzaken:
e2e tests are failing because changes in noobaa-endpoints deployment require a dedicated SCC to allow it to run.
this need to be resolved before merging this PR

jarrpa:
Okay, you can just add that commit to this PR.

jarrpa (approved):
/lgtm

PR #1155, opened 2021-04-19
storagecluster: Increase mons db allocated space

iamniting:
recent cases of mons db require more space then 10G. Increase the space
to 50G.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1944580

### DIFF
--- a/controllers/storagecluster/cephcluster.go
+++ b/controllers/storagecluster/cephcluster.go
@@ -333,7 +333,7 @@ func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, nodeCount int, s
 				StorageClassName: ds.DataPVCTemplate.Spec.StorageClassName,
 				Resources: corev1.ResourceRequirements{
 					Requests: corev1.ResourceList{
-						corev1.ResourceStorage: resource.MustParse("10Gi"),
+						corev1.ResourceStorage: resource.MustParse("50Gi"),
 					},
 				},
 			},


obnoxxx (approved):
this looks good in general.
But are we agreed on 50Gi?
The BZ seems to mention *at least* 50Gi ...

jarrpa:
> this looks good in general.
> But are we agreed on 50Gi?
> The BZ seems to mention _at least_ 50Gi ...

Yes, I believe so.

jarrpa (changes requested):
The code change is simple enough and valid. However, I worry about merging this without any consideration for upgrade scenarios of existing clusters. I believe this should go in Rook, since it handles the creation of the storage volume. @travisn @leseb Thoughts?

leseb:
> The code change is simple enough and valid. However, I worry about merging this without any consideration for upgrade scenarios of existing clusters. I believe this should go in Rook, since it handles the creation of the storage volume. @travisn @leseb Thoughts?

Rook just creates the PVC based on what's provided by the template. If the CSI driver supports volume resize (block + fs) then we are covered. I don't think the mon store needs to be resized, but let me double check. If that's the case, an init container would do the trick!

jarrpa:
> Rook just creates the PVC based on what's provided by the template. If the CSI driver supports volume resize (block + fs) then we are covered. I don't think the mon store needs to be resized, but let me double check. If that's the case, an init container would do the trick!
> 
> Edit: nothing to resize on the mon data store if the fs is resized! Nothing to do on the Rook side.

So changing the CephCluster spec should lead to a modification of the Mon's PVC and thus trigger a volume expansion from the CSI driver?

leseb:
> > Rook just creates the PVC based on what's provided by the template. If the CSI driver supports volume resize (block + fs) then we are covered. I don't think the mon store needs to be resized, but let me double check. If that's the case, an init container would do the trick!
> > Edit: nothing to resize on the mon data store if the fs is resized! Nothing to do on the Rook side.
> 
> So changing the CephCluster spec should lead to a modification of the Mon's PVC and thus trigger a volume expansion from the CSI driver?

Correct.

jarrpa (approved):
Concerns have been addressed!

/lgtm

PR #1156, opened 2021-04-19
refactor: platform_detection.go

crombus:
rename func GetPlatform() to detectPlatform() and call at the
time creation of storagecluster. Store the cloudplatform in var
and create func GetPlatform() to get the cloudplatform.

### DIFF
--- a/controllers/storagecluster/cephcluster.go
+++ b/controllers/storagecluster/cephcluster.go
@@ -117,17 +117,15 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
 		return err
 	}
 
-	platform, err := r.platform.GetPlatform(r.Client)
-	if err != nil {
-		r.Log.Error(err, "Failed to get platform")
-	} else if platform == v1.IBMCloudPlatformType || platform == IBMCloudCosPlatformType {
-		r.Log.Info(fmt.Sprintf("Increasing Mon failover timeout to 15m for %s", platform))
+	platform := r.platform.GetPlatform()
+	if platform == v1.IBMCloudPlatformType || platform == IBMCloudCosPlatformType {
+		r.Log.Info(fmt.Sprintf("Increasing Mon failover timeout to 15m for %s", r.platform.GetPlatform()))
 		cephCluster.Spec.HealthCheck.DaemonHealth.Monitor.Timeout = "15m"
 	}
 
 	// Check if this CephCluster already exists
 	found := &cephv1.CephCluster{}
-	err = r.Client.Get(context.TODO(), types.NamespacedName{Name: cephCluster.Name, Namespace: cephCluster.Namespace}, found)
+	err := r.Client.Get(context.TODO(), types.NamespacedName{Name: cephCluster.Name, Namespace: cephCluster.Namespace}, found)
 	if err != nil {
 		if errors.IsNotFound(err) {
 			if sc.Spec.ExternalStorage.Enable {
--- a/controllers/storagecluster/platform_detection.go
+++ b/controllers/storagecluster/platform_detection.go
@@ -44,22 +44,17 @@ type Platform struct {
 }
 
 // GetPlatform is used to get the CloudPlatformType of the running cluster
-func (p *Platform) GetPlatform(c client.Client) (configv1.PlatformType, error) {
-	// if 'platform' is already set just return it
-	if p.platform != "" {
-		return p.platform, nil
-	}
-	p.mux.Lock()
-	defer p.mux.Unlock()
-
-	return p.getPlatform(c)
+func (p *Platform) GetPlatform() configv1.PlatformType {
+	return p.platform
 }
 
-func (p *Platform) getPlatform(c client.Client) (configv1.PlatformType, error) {
+func (p *Platform) detectPlatform(c client.Client) error {
+	p.mux.Lock()
+	defer p.mux.Unlock()
 	infrastructure := &configv1.Infrastructure{ObjectMeta: metav1.ObjectMeta{Name: "cluster"}}
 	err := c.Get(context.TODO(), types.NamespacedName{Name: infrastructure.ObjectMeta.Name}, infrastructure)
 	if err != nil {
-		return "", fmt.Errorf("could not get infrastructure details to determine cloud platform: %v", err)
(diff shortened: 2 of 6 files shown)


jarrpa (changes requested):
Looks like the test failures are valid: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_ocs-operator/1156/pull-ci-openshift-ocs-operator-master-ocs-operator-bundle-e2e-aws/1384420533352796160/artifacts/ocs-operator-bundle-e2e-aws/ocs-must-gather/artifacts/ocs-must-gather/registry-build02-ci-openshift-org-ci-op-2812dhrd-stable-sha256-8604243ab3ef022b70d6f14a1dd8149430794e5eb8ca5d9f3eac3d3e5af462e0/namespaces/openshift-storage/pods/ocs-operator-7cfb648dd7-bqx8k/ocs-operator/ocs-operator/logs/current.log

controllers/storagecluster/cephcluster.go
```diff
@@ -117,17 +117,14 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
 		return err
 	}
 
-	platform, err := r.platform.GetPlatform(r.Client)
-	if err != nil {
-		r.Log.Error(err, "Failed to get platform")
-	} else if platform == v1.IBMCloudPlatformType || platform == IBMCloudCosPlatformType {
-		r.Log.Info(fmt.Sprintf("Increasing Mon failover timeout to 15m for %s", platform))
+	if r.platform.GetPlatform() == v1.IBMCloudPlatformType || r.platform.GetPlatform() == IBMCloudCosPlatformType {
```
jarrpa, line 120: I would say you should still keep `platform := r.platform.GetPlatform()` line.

controllers/storagecluster/platform_detection.go
```diff
@@ -44,36 +44,30 @@ type Platform struct {
...
+	defer p.mux.Unlock()
 	infrastructure := &configv1.Infrastructure{ObjectMeta: metav1.ObjectMeta{Name: "cluster"}}
 	err := c.Get(context.TODO(), types.NamespacedName{Name: infrastructure.ObjectMeta.Name}, infrastructure)
 	if err != nil {
-		return "", fmt.Errorf("could not get infrastructure details to determine cloud platform: %v", err)
+		return fmt.Errorf("could not get infrastructure details to determine cloud platform: %v", err)
 	}
 
 	p.platform = infrastructure.Status.Platform //nolint:staticcheck
-
```
jarrpa, line 66: Don't remove this line

controllers/storagecluster/platform_detection.go
```diff
@@ -138,12 +128,14 @@ func (r *StorageClusterReconciler) DevicesDefaultToFastForThisPlatform() (bool,
 // PlatformsShouldAvoidObjectStore determines whether an object store should be created
 // for the platform.
 func (r *StorageClusterReconciler) PlatformsShouldAvoidObjectStore() (bool, error) {
-	// Call GetPlatform to get platform
-	platform, err := r.platform.GetPlatform(r.Client)
-	if err != nil {
-		return false, err
+	// call detectPlatform func to detect the platform of running cluster
+	if err := r.platform.detectPlatform(r.Client); err != nil {
+		r.Log.Error(err, "Failed to get platform")
```
jarrpa, line 133: Do we still need to run this here if we're already calling it when setting up the reconciler?

jarrpa (changes requested):

controllers/storagecluster/platform_detection.go
crombus: the issue I was having was I added detectPlatform() func in storagecluster_controller.go but when I deployed the operator in aws I was still seeing rgw pod. Which shouldn’t have come up.

crombus: @jarrpa the above line can be added in GetPlatform() func and can be removed in controller.go file

jarrpa: If I'm understanding that correctly, that would just revert back to the behavior we had before, which I am explicitly asking to remove. It seems to be that we should (theoretically) be able to configure the reconciler before we ever run Reconcile(). That's the core issue I want to resolve.

crombus:
Tested the changes in the aws setup working fine.  For this https://github.com/openshift/ocs-operator/blob/1c0df4cd1cc67bcf608c51115c52d7cd3888fa60/controllers/storagecluster/cephcluster.go#L126-L132
I put debug cmds before as you suggested to see if the platform value is correct.

controllers/storagecluster/reconcile.go
```diff
@@ -167,6 +167,11 @@ func (r *StorageClusterReconciler) Reconcile(ctx context.Context, request reconc
 	// Reconcile changes to the cluster
 	result, reconcileError := r.reconcilePhases(sc, request)
 
+	// call detectPlatform func to detect the platform of running cluster
+	if err := r.platform.detectPlatform(r.Client); err != nil {
```
jarrpa, line 171: Running this every Reconcile() still seems a bit much... I'd like to figure out why we can't do it when we set up the reconciler in the first place, but I'll take it for now.

Something else to consider: Do we really want to continue reconciliation if this returns an error? Not being able to detect the platform could be a temporary network error, and thus if it resolves we would get a different value in a later reconciliation and end up altering all related configuration. This could be a problem. I would say we should `return` here. If `ErrNotFound` is a possibility, we should account for that in `detectPlatform()` and not return an error.

jarrpa (changes requested):

controllers/storagecluster/reconcile.go
crombus: How about I add if condition in detectPlatform() func to check if get platform gives me a platform value that way I can skip the error part in the func. I think it resolves both the problems. Thoughts?

priyanka19-98: @jarrpa  any thoughts on how could we handle error from `detectPlatform()` ? What if we don't return error in `Reconcile()` and then the `platform` value is not updated even when the network issue resolves?

jarrpa:
I believe this has since been taken care of elsewhere. Closing, but if this needs to be revisited please open a new PR.

PR #1141, opened 2021-03-30
Update rook to latest master build.

obnoxxx:
For 4.8 preparation, we need a newer rook image.

PR #1157, opened 2021-04-19
Fix tests after fake client update

raghavendra-talur:
After updating the controller runtime in #1141 , there has been a change in the fake client behavior. This requires update to the tests to get back the original behavior.

### DIFF
--- a/controllers/ocsinitialization/ocsinitialization_controller_test.go
+++ b/controllers/ocsinitialization/ocsinitialization_controller_test.go
@@ -50,14 +50,17 @@ func getTestParams(mockNamespace bool, t *testing.T) (v1.OCSInitialization, reco
 			Namespace: request.Namespace,
 		},
 	}
-	ocsRecon := v1.OCSInitialization{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      request.Name,
-			Namespace: request.Namespace,
-		},
-	}
 
-	return ocs, request, getReconciler(t, &ocsRecon)
+	reconciler := getReconciler(t, &ocs)
+	//The fake client stores the objects after adding a resource version to
+	//them. This is a breaking change introduced in
+	//https://github.com/kubernetes-sigs/controller-runtime/pull/1306.
+	//Therefore we cannot use the fake object that we provided as input to the
+	//the fake client and should use the object obtained from the Get
+	//operation.
+	_ = reconciler.Client.Get(context.TODO(), request.NamespacedName, &ocs)
+
+	return ocs, request, reconciler
 }
 
 func getReconciler(t *testing.T, objs ...client.Object) OCSInitializationReconciler {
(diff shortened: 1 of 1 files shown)


controllers/ocsinitialization/ocsinitialization_controller_test.go
```diff
@@ -229,7 +230,7 @@ func TestCreateSCCs(t *testing.T) {
 		ocs, request, reconciler := getTestParams(false, t)
 
 		if tc.sscCreated {
-			ocs.Status.SCCsCreated = false
+			ocs.Status.SCCsCreated = true
 			//err := reconciler.Client.Update(context.TODO(), &ocs)
 			//assert.NoErrorf(t, err, "[%s]: failed to update ocsInit status", tc.label)
```
jarrpa: Can you move this commit to the end and remove the commented code in one of the other commits? This PR effectively explains why they were removed.

raghavendra-talur: Done.

obnoxxx: interestingly, now after this is done, the previously commented out code is not removed but uncommented again...

raghavendra-talur: Those lines need to be there. They set the precondition for the test and then we do the reconcile to check if the reconciler brings the ocsinitialization to the required state. I have updated the test to check for the precondition and state after the reconciler is run.

obnoxxx:
we're onto something. but there's still something strange.
the first commit fails `make unit-test` while all the others pass.
Apart from the fact that ideally all commits should pass `make unit-test`, it seems that changing the bool in the block with the previously commented out lines of code does not change the result of the assert that follows, and of the test.

obnoxxx:
the failing test is `TestNonWatchedResourceFound` fwiw:

```
=== RUN   TestNonWatchedResourceFound                                                                                                 
    ocsinitialization_controller_test.go:170:                                                                                         
                Error Trace:    ocsinitialization_controller_test.go:170                                                              
                Error:          Received unexpected error:                                                                            
                                resourceVersion can not be set for Create requests                                                    
                Test:           TestNonWatchedResourceFound                                                                           
                Messages:       [Case 1]: failed CREATE of non watched resource                                                       
    ocsinitialization_controller_test.go:175:                                                                                         
                Error Trace:    ocsinitialization_controller_test.go:175                                                              
                Error:          Received unexpected error:         
                                ocsinitializations.ocs.openshift.io "foo" not found                                                   
                Test:           TestNonWatchedResourceFound        
                Messages:       [Case 1]: failed GET of actual resource
    ocsinitialization_controller_test.go:176:                                                                                         
...
```

obnoxxx (approved):
thanks for the updates. this looks good now

obnoxxx:
@jarrpa - final review for you

jarrpa (approved):
/lgtm

PR #1161, opened 2021-04-20
Allow passing explicit DeviceClass

synarete:
For cases where we want to associate OSD and pools with devices on
specific nodes with specific devices, we need a more fine-grained level
of device-type, which is passed on to rook as CRUSH-device-class.

As an example, consider cluster with ssd devices from two different
vendors, where user would want to have different pools, one for each
sub-device type. When combined with node-labeling and this patch, the
task becomes trivial.


jarrpa (approved):
/lgtm

PR #1162, opened 2021-04-20
hack: fix new shellcheck issue SC2269

obnoxxx:
It seems the latest stable shellcheck update has uncovered another issue

SC2269: This variable is assigned to itself, so the assignment does nothing.


umangachapagain (approved):
/lgtm

PR #1163, opened 2021-04-20
Explicit osd weight

synarete:
Allow passing explicit InitialWeight
    
    Allow passing explicit CRUSH initial-weight upon OSD init, using
    annotations mechanism (same as DeviceType/DeviceClass). This may be
    useful for cases where user wants to have more fine-grained control on
    the overall load of specific devices. As an example, consider the
    (real-life) case of a PV which resides on a partition alongside OS
    partition. In such case, user can reduce I/O to the OSD over this PV be
    lowering its weight value.
    
    Issue: https://issues.redhat.com/browse/KNIP-1616


jarrpa (approved):
/lgtm

PR #1164, opened 2021-04-20
do not pass labels as map of single empty string

synarete:
Commit 3fc88d52 ("merge user provided labels with existing ones") passes
'Labels' as map of single entry, with empty string. This, in turn,
causes deployment failure on AWS cluster, with the following message:

failed to create metrics exporter service
openshift-storage/ocs-metrics-exporter. Service \"ocs-metrics-exporter\" is
invalid: [metadata.labels: Invalid value: \"\": name part must be non-empty

Fixed by setting empty Labels map for the default case of Spec.Monitoring


jarrpa (approved):
/lgtm

PR #1170, opened 2021-04-27
hack: Change the order of the IMAGE_BUILD_CMD

iamniting:
check if podman exists first and then check for docker if podman not
found.

error out if both were not found


jarrpa (changes requested):
The errors we're seeing in CI are valid. You should always be running `make ocs-operator-ci` before pushing.

jarrpa (changes requested):
Ah, I see now. It isn't quite sufficient to only run the check for the files where `IMAGE_BUILD_CMD` is used directly, since it is also part of other variables. I would say I see no problem with requiring the existence of at least one container runtime to develop on this project, so just put the check in `common.sh`.

iamniting:
> Ah, I see now. It isn't quite sufficient to only run the check for the files where `IMAGE_BUILD_CMD` is used directly, since it is also part of other variables. I would say I see no problem with requiring the existence of at least one container runtime to develop on this project, so just put the check in `common.sh`.

Made changes as discussed in the chat

hack/build-appregistry.sh
```diff
@@ -3,8 +3,9 @@
...
 
 source hack/common.sh
+source hack/containers-common.sh
 source hack/generate-appregistry.sh
 
 $IMAGE_BUILD_CMD build --no-cache -t "$CATALOG_FULL_IMAGE_NAME" -f build/Dockerfile.registry .
 echo
-echo "Run '${IMAGE_BUILD_CMD} push ${CATALOG_FULL_IMAGE_NAME}' to push ocs-registry to ${IMAGE_REGISTRY}."
\ No newline at end of file
+echo "Run '${IMAGE_BUILD_CMD} push ${CATALOG_FULL_IMAGE_NAME}' to push ocs-registry to ${IMAGE_REGISTRY}."
```
jarrpa, line 11: I still see this change... is this your editor doing this? I would either remove this change or move it into its own commit, since it is an unrelated formatting change.

iamniting: Yes, my editor did it. Moved to separate commit

jarrpa (approved):
Aww yes. Love it!

/lgtm

jarrpa:
Seems you missed one.

iamniting:
> Seems you missed one.

The failure is because of `hack/source-manifests.sh` where we are calling `hack/docker-common.sh` and docker or podman is not present on the machine. As a fix i am going to call `hack/docker-common.sh` in the funcs where it is required not for the whole file. It should resolve the problem.

jarrpa:
Oh right! Okay, then instead call it around line 100, inside the if statement. Also be mindful of indentation.

iamniting:
> Oh right! Okay, then instead call it around line 100, inside the if statement. Also be mindful of indentation.

Changed it, did not realised it before.

jarrpa:
I still see two source calls, we want only one.

iamniting:
> I still see two source calls, we want only one.

Done

jarrpa (approved):
/lgtm

PR #1174, opened 2021-05-02
fix ClusterObjectStoreState Alert empty spec

anmolsachan:
This commit fixes ClusterObjectStoreState alert spec being empty due to
which the rule was not being listed in the Prometheus Rules and alert
was not working. The reason alert spec is empty is because the function
getPrometheusRuleSpecFrom is returning empty object as the rule file
decoding is creating an object that matches the type struct
monitoringv1.PrometheusRule{} which does not match with the type of the
variable ruleSpec which is monitoringv1.PrometheusRuleS

Signed-off-by: Anmol Sachan <anmol13694@gmail.com>

Fix: https://bugzilla.redhat.com/show_bug.cgi?id=1948378

### DIFF
--- a/controllers/storagecluster/prometheus.go
+++ b/controllers/storagecluster/prometheus.go
@@ -79,10 +79,11 @@ func getPrometheusRuleSpecFrom(filePath string) (*monitoringv1.PrometheusRuleSpe
 	if err != nil {
 		return nil, fmt.Errorf("'%s' not readable", filePath)
 	}
-	ruleSpec := monitoringv1.PrometheusRuleSpec{}
-	if err := k8sYAML.NewYAMLOrJSONDecoder(bytes.NewBufferString(string(fileContent)), 1000).Decode(&ruleSpec); err != nil {
+	rule := monitoringv1.PrometheusRule{}
+	if err := k8sYAML.NewYAMLOrJSONDecoder(bytes.NewBufferString(string(fileContent)), 1000).Decode(&rule); err != nil {
 		return nil, err
 	}
+	ruleSpec := rule.Spec
 	return &ruleSpec, nil
 }
 


jarrpa:
Having no understanding of the problem you're trying to solve, I don't see the connection between the commit title and the actual changes. Can you provide a bit more description in the commit message about why/how your change addresses the problem?

anmolsachan:
> Having no understanding of the problem you're trying to solve, I don't see the connection between the commit title and the actual changes. Can you provide a bit more description in the commit message about why/how your change addresses the problem?

@jarrpa Thank you for the review. I have updated the commit and description.

anmolsachan:
@umangachapagain Please review.

jarrpa:
You're still not really explaining how your code solves the problem. As far as I can tell, you are still returning the same thing you were returning before, I don't see why it would be any different.

controllers/storagecluster/prometheus.go
```diff
@@ -79,10 +79,11 @@ func getPrometheusRuleSpecFrom(filePath string) (*monitoringv1.PrometheusRuleSpe
 	if err != nil {
 		return nil, fmt.Errorf("'%s' not readable", filePath)
 	}
-	ruleSpec := monitoringv1.PrometheusRuleSpec{}
-	if err := k8sYAML.NewYAMLOrJSONDecoder(bytes.NewBufferString(string(fileContent)), 1000).Decode(&ruleSpec); err != nil {
```
anmolsachan, line 83: @jarrpa Apologies for the late reply on this, I was out sick. The above code is expected to decode the rule file into an object and also get the spec out of it and put it in the `ruleSpec` variable. It fails to do so as the object that gets created is of type struct `monitoringv1.PrometheusRule{} `and thus the variable ruleSpec remains empty as it of type `monitoringv1.PrometheusRuleSpec{}`

I am trying to do is, to first convert the rule file into its matching type object i.e `monitoringv1.PrometheusRule{`} and then take the spec out to return it.

jarrpa: Sure. Include something about this in the commit message. :slightly_smiling_face:

anmolsachan: @jarrpa Thank You. I added the short version of it to commit and updated the PR description with the same. Please review.

jarrpa (approved):
/lgtm

PR #1178, opened 2021-05-06
Allow passing explicit PrimaryAffinity

synarete:
Allow passing explicit primary-affinity upon OSD's deployment using
annotations mechanism (same as DeviceClass/InitialWeight). This may be
useful for cases where user wants to have more fine-grained control on
the overall (read) load of specific devices. Valid values are within
range [0, 1) (where 1 is the default).

As an example, consider the (real-life) case of a PV which resides on a
partition alongside OS partition. In such case, user can reduce I/O load
of this specific OSD by choosing explicit primary-affinity value.

Issue: https://issues.redhat.com/browse/KNIP-1616
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1924946

### DIFF
--- a/api/v1/storagecluster_types.go
+++ b/api/v1/storagecluster_types.go
@@ -163,6 +163,15 @@ type StorageDeviceSet struct {
 	// +optional
 	InitialWeight string `json:"initialWeight,omitempty"`
 
+	// PrimaryAffinity is an optional OSD primary-affinity value within the
+	// range [0,1). This value influence the way Ceph's CRUSH selection of
+	// primary OSDs. Lower value reduce performance bottlenecks (especially
+	// on read operations). If not set, default value is 1.
+	// https://docs.ceph.com/en/latest/rados/operations/crush-map/#primary-affinity
+	// +kubebuilder:validation:Pattern=`^0.[0-9]+$`
+	// +optional
+	PrimaryAffinity string `json:"primaryAffinity,omitempty"`
+
 	// TopologyKey is the Kubernetes topology label that the
 	// StorageClassDeviceSets will be distributed across. Ignored if
 	// Placement is set
--- a/config/crd/bases/ocs.openshift.io_storageclusters.yaml
+++ b/config/crd/bases/ocs.openshift.io_storageclusters.yaml
@@ -3561,6 +3561,14 @@ spec:
                             type: object
                           type: array
                       type: object
+                    primaryAffinity:
+                      description: PrimaryAffinity is an optional OSD primary-affinity
+                        value within the range [0,1). This value influence the way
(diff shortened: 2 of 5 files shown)


jarrpa (changes requested):
Overall this seems to make sense. Two requests:
* Please provide at least a short description of what "primary affinity" means in the comments in `storagecluster_types.go`
* Split the generated CSV changes into their own commit

synarete:
> Overall this seems to make sense. Two requests:
> 
> * Please provide at least a short description of what "primary affinity" means in the comments in `storagecluster_types.go`
> * Split the generated CSV changes into their own commit

@jarrpa Will fix soon and re-send the PR

api/v1/storagecluster_types.go
```diff
@@ -163,6 +163,14 @@ type StorageDeviceSet struct {
 	// +optional
 	InitialWeight string `json:"initialWeight,omitempty"`
 
+	// PrimaryAffinity is an optional OSD primary-affinity value within the
+	// range [0,1). This value influence the way Ceph's CRUSH selection of
+	// primary OSDs. Lower value reduce performance bottlenecks (especially
+	// on read operations). If not set, default value is 1.
+	// +kubebuilder:validation:Pattern=`^0.[0-9]+$`
+	// +optional
```
jarrpa, line 171:
```suggestion
	// PrimaryAffinity is an optional OSD primary-affinity value within the
	// range [0,1). This value influences Ceph's CRUSH selection of
	// primary OSDs. A lower value reduces the probability that the OSD will
	// be selected. If not set, default value is 1.
	// https://docs.ceph.com/en/latest/rados/operations/crush-map/#primary-affinity
	// +kubebuilder:validation:Pattern=`^0.[0-9]+$`
	// +optional
```

jarrpa (changes requested):
Good split. Suggesting some edits to the additional comment. Your comment gave me a better understanding on what this does, but the language was a little unclear. A direct link to the documentation will probably be useful here, since anyone modifying this value is already in a place where they should know more about this.

api/v1/storagecluster_types.go
synarete: @jarrpa Should I re-submit my PR with those changes?

jarrpa: Yes

jarrpa (approved):
/lgtm

obnoxxx:
@synarete , @jarrpa - I think we need a backport of this to `release-4.8`.

synarete:
> @synarete , @jarrpa - I think we need a backport of this to `release-4.8`.

@jarrpa I issued a PR (#1179)

PR #1185, opened 2021-05-21
fix merge strategy for newCephDaemonResources

umangachapagain:
newCephDaemonResources was taking custom resource requirements
for multiple daemons but only returning resource requirements
of ceph-mgr and mon.

This patch changes it to take all custom resource requirements
and merge it with the defaults. If key exists, it's value will
be overwritten and if it doesn't exist, it will be appended.

### DIFF
--- a/controllers/storagecluster/cephcluster.go
+++ b/controllers/storagecluster/cephcluster.go
@@ -682,11 +682,13 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster, serverVersion *version.
 
 func newCephDaemonResources(custom map[string]corev1.ResourceRequirements) map[string]corev1.ResourceRequirements {
 	resources := map[string]corev1.ResourceRequirements{
-		"mon": defaults.GetDaemonResources("mon", custom),
-		"mgr": defaults.GetDaemonResources("mgr", custom),
+		"mon": defaults.DaemonResources["mon"],
+		"mgr": defaults.DaemonResources["mgr"],
+		"mds": defaults.DaemonResources["mds"],
+		"rgw": defaults.DaemonResources["rgw"],
 	}
 
-	for k := range resources {
+	for k := range custom {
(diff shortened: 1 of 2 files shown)


umangachapagain:
FYI @nb-ohad

controllers/storagecluster/cephcluster.go
```diff
@@ -682,11 +682,13 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster, serverVersion *version.
...
-		"mon": defaults.GetDaemonResources("mon", custom),
-		"mgr": defaults.GetDaemonResources("mgr", custom),
+		"mon": defaults.DaemonResources["mon"],
+		"mgr": defaults.DaemonResources["mgr"],
+		"mds": defaults.DaemonResources["mds"],
+		"rgw": defaults.DaemonResources["rgw"],
 	}
 
-	for k := range resources {
+	for k := range custom {
```
iamniting, line 691: This can be simplified.
```
for k, v := range custom {
    resources[k] = v
}
```

iamniting (changes requested):

controllers/storagecluster/cephcluster.go
nb-ohad: Just a note:

Using `for k, v = range r` will create an in-memory copy of the items in (one per iteration cycle). 
If the items of `r` are of value type and not pointer type this might create observable memory overhead (depending on the size of each item and the number of items in the range). Also mutating the copy will not mutate the original. 

A better simplification could be:
```
for k := range custom {
      resources[k] = custom[k]
}
```

jarrpa: While I agree this can be simplified, it's not strictly part of enabling this feature, so it can be done in another PR.

jarrpa (approved):
/lgtm

PR #1191, opened 2021-05-26
must-gather: collect noobaa logs independent of sc and external sc

rajatsing:

### DIFF
--- a/must-gather/collection-scripts/gather
+++ b/must-gather/collection-scripts/gather
@@ -16,16 +16,18 @@ pre-install.sh
 # Call other gather scripts
 gather_namespaced_resources ${BASE_COLLECTION_PATH} "${SINCE_TIME}"
 gather_clusterscoped_resources ${BASE_COLLECTION_PATH} "${SINCE_TIME}"
+gather_noobaa_resources ${BASE_COLLECTION_PATH} "${SINCE_TIME}"
 
 storageClusterPresent=$(oc get storagecluster -n openshift-storage -o go-template='{{range .items}}{{.metadata.name}}{{"\n"}}{{end}}')
-if [ "$(oc get storagecluster -n openshift-storage -o go-template='{{range .items}}{{.spec.externalStorage.enable}}{{"\n"}}{{end}}')" == true ]; then
-   echo "Skipping the ceph and nooobaa collection as External Storage is enabled" | tee -a  "${BASE_COLLECTION_PATH}"/gather-debug.log
+externalStorageClusterPresent=$(oc get storagecluster -n openshift-storage -o go-template='{{range .items}}{{.spec.externalStorage.enable}}{{"\n"}}{{end}}')
+
+if [[ -n "${externalStorageClusterPresent}" ]]; then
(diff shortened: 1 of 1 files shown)


crombus (changes requested):
We do not want to collect noobaa if stroage cluster is present.

You need different `if` section to do this with same condition storagecluster condition.

rajatsing:
> We do not want to collect noobaa if stroage cluster is present.
> 
> You need different `if` section to do this with same condition storagecluster condition.

Noo, I think we need noobaa collection when storagecluster is present as well as when external storagecluter is present.

jarrpa:
If I understand this correctly, the actual feature is to collect NooBaa information WITHOUT Ceph logs when you have an external StorageCluster, right? If so, a cleaner solution would be to introduce an `if !external` stanza in the existing `else` clause. I would imagine gathering NooBaa resources is independent of whether a StorageCluster exists or not, you should just be gathering any NooBaa resources that exist. This is especially important for our NooBaa-only deployments.

crombus:
my concern regarding the storagecluster was because in last release we had a bug that noobaa shouldn’t be collected if storagecluster not present. Now as we can have noobaa without storagecluster also. So @rajatsing now no checks needed just call the script as we call for ns and clusterdscope scripts and update the log line w.r.this.

rajatsing:
@crombus , I have made changes according to what you indicated. Now, noobaa is independent of storage and noobaa cluster

must-gather/collection-scripts/gather
```diff
@@ -16,16 +16,18 @@ pre-install.sh
...
-   echo "Skipping the ceph and nooobaa collection as External Storage is enabled" | tee -a  "${BASE_COLLECTION_PATH}"/gather-debug.log
+externalStorageClusterPresent=$(oc get storagecluster -n openshift-storage -o go-template='{{range .items}}{{.spec.externalStorage.enable}}{{"\n"}}{{end}}')
+
+if [[ -n "${externalStorageClusterPresent}" ]]; then
+    echo "Skipping the ceph collection as External Storage is enabled" | tee -a  "${BASE_COLLECTION_PATH}"/gather-debug.log
 elif [ -z "${storageClusterPresent}" ]; then
-    echo "Skipping ceph and nooobaa collection as Storage Cluster is not present" | tee -a  "${BASE_COLLECTION_PATH}"/gather-debug.log
+    echo "Skipping ceph collection as Storage Cluster is not present" | tee -a  "${BASE_COLLECTION_PATH}"/gather-debug.log
 else
     echo "Collecting ceph and noobaa logs" | tee -a "${BASE_COLLECTION_PATH}"/gather-debug.log
```
crombus, line 29: remove noobaa

crombus (changes requested):
code changes looks good. Just update the log msg.

must-gather/collection-scripts/gather
rajatsing: Thanks, doing right now.

rajatsing: done!.

crombus (approved):

PR #1193, opened 2021-05-31
Change rookv1.NetworkSpec to cephv1.NetworkSpec and pass full networkspec if it is specified.

rohantmp:
Looking for other places this might be touched in code. Will update then.

Necessary for IPv6 support so that IPFamily and dualstack can be specified: https://issues.redhat.com/browse/KNIP-1467

### DIFF
--- a/api/v1/storagecluster_types.go
+++ b/api/v1/storagecluster_types.go
@@ -19,6 +19,7 @@ package v1
 import (
 	nbv1 "github.com/noobaa/noobaa-operator/v2/pkg/apis/noobaa/v1alpha1"
 	conditionsv1 "github.com/openshift/custom-resource-status/conditions/v1"
+	rookCephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
 	rook "github.com/rook/rook/pkg/apis/rook.io/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -49,7 +50,7 @@ type StorageClusterSpec struct {
 	// Version specifies the version of StorageCluster
 	Version string `json:"version,omitempty"`
 	// Network represents cluster network settings
-	Network *rook.NetworkSpec `json:"network,omitempty"`
+	Network *rookCephv1.NetworkSpec `json:"network,omitempty"`
 	// ManagedResources specifies how to deal with auxiliary resources reconciled
 	// with the StorageCluster
 	ManagedResources ManagedResourcesSpec `json:"managedResources,omitempty"`
--- a/controllers/storagecluster/cephcluster.go
+++ b/controllers/storagecluster/cephcluster.go
@@ -14,7 +14,7 @@ import (
 	ocsv1 "github.com/openshift/ocs-operator/api/v1"
 	"github.com/openshift/ocs-operator/controllers/defaults"
 	statusutil "github.com/openshift/ocs-operator/controllers/util"
-	cephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
+	rookCephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
 	rook "github.com/rook/rook/pkg/apis/rook.io/v1"
 	corev1 "k8s.io/api/core/v1"
 	storagev1 "k8s.io/api/storage/v1"
@@ -102,7 +102,7 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
 		}
 	}
 
-	var cephCluster *cephv1.CephCluster
+	var cephCluster *rookCephv1.CephCluster
 	// Define a new CephCluster object
 	if sc.Spec.ExternalStorage.Enable {
 		cephCluster = newExternalCephCluster(sc, r.images.Ceph, r.monitoringIP, r.monitoringPort)
@@ -136,7 +136,7 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
 	}
 
 	// Check if this CephCluster already exists
-	found := &cephv1.CephCluster{}
+	found := &rookCephv1.CephCluster{}
 	err = r.Client.Get(context.TODO(), types.NamespacedName{Name: cephCluster.Name, Namespace: cephCluster.Namespace}, found)
 	if err != nil {
 		if errors.IsNotFound(err) {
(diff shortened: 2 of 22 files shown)


config/manager/kustomization.yaml
```diff
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: ocs-dev/ocs-operator
-  newName: quay.io/ocs-dev/ocs-operator
+  newName: quay.io/rohantmp/ocs-operator
```
rohan47, line 7: @rohantmp you will change this right and other occurrences?

controllers/storagecluster/cephcluster.go
```diff
@@ -353,8 +353,10 @@ func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, nodeCount int, s
 		reqLogger.Info(fmt.Sprintf("No monDataDirHostPath, monPVCTemplate or storageDeviceSets configured for storageCluster %s", sc.GetName()))
 	}
 	if isMultus(sc.Spec.Network) {
-		cephCluster.Spec.Network.NetworkSpec = *sc.Spec.Network
+		cephCluster.Spec.Network = *sc.Spec.Network
 	}
+
+	reqLogger.Info("AZF --------------------------------------------- networkspec: %+v", *sc.Spec.Network)
```
rohan47, line 359: this was used for debugging right?

rohan47 (changes requested):

controllers/storagecluster/cephcluster.go
rohantmp: Ah, thanks :)

rohantmp:
Oops, pushed the wrong branch and noticed some other errors. @rohan47 hold off on review

rohantmp:
@rohan47 ready for review and I added unit tests

rohantmp:
failed because of:
> the build ci-index-fixed-dependencies-gen failed after 4m1s with reason BuildPodEvicted: The node was low on resource: memory.

rohan47 (approved):

rohantmp:
failed due to:
> failed: the pod ci-op-tpw0rv1d/ocs-operator-bundle-e2e-aws-tests failed after 4m10s (failed containers: ): Evicted The node was low on resource: memory.

controllers/storagecluster/cephcluster.go
```diff
@@ -385,6 +381,23 @@ func validateMultusSelectors(selectors map[string]string) error {
 	return nil
 }
 
+// getNetworkSpec returns cephv1.NetworkSpec after reconciling the
+// storageCluster.Spec.HostNetwork and storageCluster.Spec.Network fields
+func getNetworkSpec(sc ocsv1.StorageCluster) cephv1.NetworkSpec {
+	// only hostnetwork specified
+	networkSpec := cephv1.NetworkSpec{HostNetwork: sc.Spec.HostNetwork}
```
jarrpa, line 388: I'd complain about this override... but I suppose we're stuck with this field forever or until v2, which ever comes first. :stuck_out_tongue:

controllers/storagecluster/cephcluster.go
```diff
@@ -385,6 +381,23 @@ func validateMultusSelectors(selectors map[string]string) error {
...
+
+	// full NetworkSpec specified
+	if sc.Spec.Network != nil {
+		networkSpec = *sc.Spec.Network
+		// both specified
+		if sc.Spec.HostNetwork {
```
jarrpa, line 394: Honestly, is this desired behavior? If a NetworkSpec even exists, I would expect we would just take it as-is, full override of anything else in the StorageCluster Spec.

api/v1/storagecluster_types.go
```diff
@@ -19,6 +19,7 @@ package v1
 import (
 	nbv1 "github.com/noobaa/noobaa-operator/v2/pkg/apis/noobaa/v1alpha1"
 	conditionsv1 "github.com/openshift/custom-resource-status/conditions/v1"
+	cephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
```
jarrpa, line 22: Please please please name this `rookCephv1`.

controllers/storagecluster/cephcluster.go
```diff
@@ -363,7 +359,7 @@ func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, nodeCount int, s
 	return cephCluster
 }
 
-func isMultus(nwSpec *rook.NetworkSpec) bool {
+func isMultus(nwSpec *cephv1.NetworkSpec) bool {
```
jarrpa, line 362: Similarly here, please rename the import to `rookCephv1`.

controllers/storagecluster/cephcluster_test.go
```diff
@@ -671,3 +674,103 @@ func TestNewCephDaemonResources(t *testing.T) {
...
+		},
+	}
+	for _, testCase := range testTable {
+		t.Logf("running testCase %q", testCase.desc)
+		sc := ocsv1.StorageCluster{
+			Spec: testCase.scSpec,
+		}
+		actual := getNetworkSpec(sc)
+		// test Provider, Selectors, HostNetwork, IPFamily, Dualstack
+		diff := cmp.Diff(testCase.expected, actual)
```
jarrpa, line 772: Why are you explicitly doing a diff instead of `assert.Equal()`? Example: https://github.com/openshift/ocs-operator/blob/master/controllers/storagecluster/cephcluster_test.go#L114

jarrpa: ...oh, I just learned about `cmp`! https://pkg.go.dev/github.com/google/go-cmp/cmp

That said, you could probably use `assert.True(t, cmp.Equal(a, b), cmp.Diff(a ,b))` or something to that effect, to at least keep things more consistent with how the rest of the test suite currently works.

We should probably look into switching from `reflect` to `cmp` in the code base, as a future PR.

jarrpa (changes requested):
Holy crap, sorry for letting this fly under my radar for so long. :sweat: 

In general, this looks good. Left a few comments and questions.

This seems like something that calls for a mutating webhook, so we can just automatically convert `Spec.HostNetwork` into this new format and not have to deal with it again. But that can be a separate PR.

And of course, please resolve the merge conflicts.

controllers/storagecluster/cephcluster.go
rohantmp: If either is set, then the user is indicating hostNetwork should be true. 
Scenario:
User sets hostNetwork, but defines NetworkSpec.IPFamily. NetworkSpec is non-nil, and NetworkSpec.hostNetwork defaults to false.

User goes: "huh, I set hostNetwork, why isn't this stupid thing turning on"

api/v1/storagecluster_types.go
rohantmp: Updating for consistency. Kinda wish the whole thing was `cephv1` :laughing:

controllers/storagecluster/cephcluster_test.go
rohantmp: Learned about it from looking at rook code! Changing to that format.

api/v1/storagecluster_types.go
rohantmp: Looks like `cephv1` is more common in the codebase than `rookCephv1`. This and a few files are an outlier! updating `rookCephv1` to `cephv1` in this file :)

rohantmp:
Added one more commit to allow an empty `public` multus selector. Omitting it is valid.

jarrpa (changes requested):
Asking to revert a change and make a few optimizations for human clarity.

rohantmp:
Sorry for the long wait! Made the requested changes :)

rohantmp:
Just saw
> Wait... actually, it seems this problem is already solved for us: https://pkg.go.dev/gotest.tools/assert/cmp?utm_source=godoc#DeepEqual assert/cmp itself uses cmp in its implementation. Can you see if this works?

investigating

rohantmp:
That package does seem good and it used google's cmp, so moved to it for this file.

jarrpa (approved):
/lgtm

PR #1194, opened 2021-05-31
deps: update Golang to 1.16

umangachapagain:
Go 1.16 is not available in upstream builder images.
Dockerfile needs to be updated once it is available.


jarrpa (approved):
/lgtm

PR #1197, opened 2021-06-01
Disable UI support for tainting user selected nodes dedicated

bipuladh:
QE team doesn't have enough bandwidth to test it hence disabling this feature.


jarrpa (approved):
/lgtm

PR #1198, opened 2021-06-01
Re-Change job template name to its original one, 

pkesavap:
Refer: https://bugzilla.redhat.com/show_bug.cgi?id=1966149

The osd removal job name on OCS4.7 and OCS4.8 is different.Thus this PR will set back the original name of the job

### DIFF
--- a/controllers/storagecluster/job_templates.go
+++ b/controllers/storagecluster/job_templates.go
@@ -136,7 +136,7 @@ func newExtendClusterJob(sc *ocsv1.StorageCluster, jobTemplateName string, cephC
 			APIVersion: "batch/v1",
 		},
 		ObjectMeta: metav1.ObjectMeta{
-			Name:        jobTemplateName,
+			Name:        jobTemplateName + "-job",
 			Namespace:   sc.Namespace,
 			Labels:      labels,
 			Annotations: annotations,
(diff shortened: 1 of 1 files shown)


jarrpa (changes requested):
...is there any reason why you're making the names different now? I would just add `-job` to the end of both variables and leave the rest as-is. That said, it's kind of ridiculous to do `nameVar+"-job"` in the tests and not in the actual code. :) Please at least be consistent and do the same in the code.

pkesavap:
Hi @jarrpa 
In OCS 4.7, the template name and the job name were different, the job name had an extension of '-job' [template_name+'-job] appended to the template_name.

As part of [ https://bugzilla.redhat.com/show_bug.cgi?id=1966149], I wanted to be safe and give the template name and job name the same appellation it had in OCS 4.7, thus if the template name has 'ocs-extend-cluster' , then the job name should be 'ocs-extend-cluster-job' . 

Coming to the test cases, the reason why I gave the "-job" was that the method expects job name, and job name = template-name+"-job"

I implemented it this way, to have as minimal change possible along with keeping all the names similar to the one it had in OCS 4.7.

pkesavap:
@jarrpa As discussed offfline, i introduced the "-job" directly inside the job methods, 
with just this change, changes to the test file are also not needed.

jarrpa (approved):
/lgtm

agarwal-mudit:
@pkesavap looks like this needs manual cherry picking as there are some conflicts.
Ping me if you need some help on gchat.

agarwal-mudit:
@pkesavap For release-4.8 you need to make same changes in `./controllers/storagecluster/reconcile.go`
because there is no file named ` controllers/storagecluster/job_templates.go ` in release-4.8

PR #1209, opened 2021-06-04
GateWay spec's 'Instances' field should be minimum 1

aruniiird:
'Instances' field in GateWay spec needs to be atleast ONE.

External cluster creations are failing due to this with
following error in OCS Operator logs,

"error":"CephObjectStore.ceph.rook.io \"ocs-external-storagecluster-cephobjectstore\"
is invalid: spec.gateway.instances: Invalid value: 0: spec.gateway.instances in body
should be greater than or equal to 1"


aruniiird:
@jarrpa , can you please take a look...

jarrpa (approved):
/lgtm

PR #1211, opened 2021-06-06
topology: fix crash in determinePlacementRack

nbalacha:
Add a new rack for a node if there are no empty racks and none
of the nodes in the existing racks belong to the same zone as the new
node.

### DIFF
--- a/controllers/storagecluster/topology.go
+++ b/controllers/storagecluster/topology.go
@@ -96,13 +96,11 @@ func determinePlacementRack(
 	rackList := []string{}
 
 	if len(nodeRacks.Labels) < minRacks {
-		for i := len(nodeRacks.Labels); i < minRacks; i++ {
-			for j := 0; j <= i; j++ {
-				newRack := fmt.Sprintf("rack%d", j)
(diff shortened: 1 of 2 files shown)


nbalacha:
/assign @raghavendra-talur 

Raghavendra, will these changes work with the arbiter feature?

controllers/storagecluster/topology.go
```diff
@@ -96,13 +96,11 @@ func determinePlacementRack(
 	rackList := []string{}
 
 	if len(nodeRacks.Labels) < minRacks {
-		for i := len(nodeRacks.Labels); i < minRacks; i++ {
-			for j := 0; j <= i; j++ {
```
jarrpa, line 100: I really don't remember why this was here... I'm pretty sure I wrote it. Maybe it was an artifact from an abandoned implementation I never removed. Ah well, happy to see it gone. :)

jarrpa (approved):
/lgtm

jarrpa:
> /assign @raghavendra-talur
> 
> Raghavendra, will these changes work with the arbiter feature?

I should mention there was an offline discussion where we decided we believe this should not impact arbiter functionality

PR #1215, opened 2021-06-08
Update the ocs-osd-removal template description

xenolinux:
Reference https://bugzilla.redhat.com/show_bug.cgi?id=1962281
The status of "ocs-osd-removal-job pod" turns completed although removal
job failed to remove osds successfully.
In some cases user might enter wrong input to the FAILED_OSD_IDS
template parameter.
This commit updates the template description to notify the user to always
read the job logs to check whether osds successfully removed or not.


xenolinux:
The status of a k8s job can only be either completed or error. If user gives wrong input for eg[1] then the user must check the logs to know if the expected osd(s) successfully got removed.  To determine whether the user has/has not entered the correct value to the template parameter [FAILED_OSD_IDS] is beyond the coverage of Rook.
[1] ```$ oc process -n openshift-storage ocs-osd-removal \
-p FAILED_OSD_IDS=failed-osd-id1,failed-osd-id2 | oc create -f -```

jarrpa (approved):
/lgtm

PR #1217, opened 2021-06-09
Adding 'thick-provision' storageclass to External cluster

aruniiird:
Thick-provision storage class is required along with RBD storageclass.

### DIFF
--- a/controllers/storagecluster/external_resources.go
+++ b/controllers/storagecluster/external_resources.go
@@ -309,6 +309,10 @@ func (r *StorageClusterReconciler) createExternalStorageClusterResources(instanc
 				enableRookCSICephFS = true
 			} else if d.Name == cephRbdStorageClassName {
 				scc = newCephBlockPoolStorageClassConfiguration(instance, false)
+				// if RBD storageclass is present, we need to add thick provision
+				// storageclass as well into the 'availableSCCs' array
+				availableSCCs = append(availableSCCs,
+					newCephBlockPoolStorageClassConfiguration(instance, true))
 			} else if d.Name == cephRgwStorageClassName {
 				rgwEndpoint := d.Data[externalCephRgwEndpointKey]
 				if err := checkEndpointReachable(rgwEndpoint, 5*time.Second); err != nil {


aruniiird:

controllers/storagecluster/external_resources.go
```diff
@@ -309,6 +310,9 @@ func (r *StorageClusterReconciler) createExternalStorageClusterResources(instanc
 				enableRookCSICephFS = true
 			} else if d.Name == cephRbdStorageClassName {
 				scc = newCephBlockPoolStorageClassConfiguration(instance, false)
+				// if RBD storageclass is present,
+				// we need to enable thick provision storageclass as well
+				enableThickProvisionerSC = true
```
jarrpa, line 315: Is there any reason you can't just do `availableSCCs = append(availableSCCs, newCephBlockPoolStorageClassConfiguration(instance, true))` here? If it's always `true` when the RBD SC is present, why have a boolean for it?

aruniiird: @jarrpa , variable `enableThickProvisionerSC` will only be `true` when we see a RBD storageclass.
At the declaration (at line#241), by default the variable will be `false`.

https://github.com/openshift/ocs-operator/blob/10fe9dad4479e63ede3ef37f76cf70672f59aa76/controllers/storagecluster/external_resources.go#L241

agarwal-mudit: Yeah, so it is always true inside this `if` block so what Jose is saying that we can just call `availableSCCs = append(availableSCCs, newCephBlockPoolStorageClassConfiguration(instance, true))` inside this `if` block itself.
Is there a reason for not doing it?

aruniiird: Oh! ok I got the point... Yeah true, I was thinking a bit wild (like whether to make `scc` an array to hold both the storageclasses and do the append at the end or what if we enter into the condition multiple times).
Yup, suggestion make sense. Making the change

agarwal-mudit (approved):

controllers/storagecluster/external_resources.go
jarrpa: I do agree that it's not up to my preferences for elegance. That said, this is one of those instances where I'm more concerned with the simplicity of the change. No sense trying to go too hard for a one-off like this.

jarrpa (approved):
/lgtm

PR #1220, opened 2021-06-10
must-gather: collect ceph CR's logs when external cluster is deployed

rajatsing:
Currently for External mode, the must-gather skips the whole ceph collection
and therefore some resources which are applicable for external mode are absent
in the logs. Those resources are
* cephobjectstores
* cephobjectstoreusers
* cephclusters

### DIFF
--- a/must-gather/collection-scripts/gather_ceph_resources
+++ b/must-gather/collection-scripts/gather_ceph_resources
@@ -1,4 +1,7 @@
 #!/bin/bash
+
+gather_common_ceph_resources "${BASE_COLLECTION_PATH}" "${SINCE_TIME}"
+
 # Expect base collection path as an argument
 BASE_COLLECTION_PATH=$1
 
@@ -15,10 +18,7 @@ CEPH_COLLECTION_PATH="${BASE_COLLECTION_PATH}/ceph"
 # Ceph resources
 ceph_resources=()
 ceph_resources+=(cephblockpools)
-ceph_resources+=(cephclusters)
 ceph_resources+=(cephfilesystems)
-ceph_resources+=(cephobjectstores)
-ceph_resources+=(cephobjectstoreusers)
 
 # Ceph commands
 ceph_commands=()
--- a/must-gather/collection-scripts/gather_common_ceph_resources
+++ b/must-gather/collection-scripts/gather_common_ceph_resources
@@ -1,0 +1,27 @@
+#!/bin/bash
+
+# Expect base collection path as an argument
+BASE_COLLECTION_PATH=$1
+
+# Expect time option as an argument
+SINCE_TIME=$2
+
+# Use PWD as base path if no argument is passed
+if [ "${BASE_COLLECTION_PATH}" = "" ]; then
+    BASE_COLLECTION_PATH=$(pwd)
+fi
+
+CEPH_COLLECTION_PATH="${BASE_COLLECTION_PATH}/ceph"
+
+# common Ceph resources
+common_ceph_resources=()
+common_ceph_resources+=(cephobjectstores)
+common_ceph_resources+=(cephobjectstoreusers)
+common_ceph_resources+=(cephclusters)
+
+for resource in "${common_ceph_resources[@]}"; do
+    echo "collecting dump ${resource}" | tee -a  "${BASE_COLLECTION_PATH}"/gather-debug.log
+    { oc adm --dest-dir="${CEPH_COLLECTION_PATH}" inspect "${resource}" --all-namespaces --"${SINCE_TIME}"; } >> "${BASE_COLLECTION_PATH}"/gather-debug.log 2>&1
+done
+cat "${BASE_COLLECTION_PATH}"/gather-ceph-debug.log >> "${BASE_COLLECTION_PATH}"/gather-debug.log 2>&1
+rm -rf "${BASE_COLLECTION_PATH}/"gather-ceph-debug.log >> "${BASE_COLLECTION_PATH}"/gather-debug.log 2>&1
--- a/must-gather/collection-scripts/gather
+++ b/must-gather/collection-scripts/gather
@@ -19,13 +19,15 @@ gather_clusterscoped_resources ${BASE_COLLECTION_PATH} "${SINCE_TIME}"
 gather_noobaa_resources ${BASE_COLLECTION_PATH} "${SINCE_TIME}"
 
 storageClusterPresent=$(oc get storagecluster -n openshift-storage -o go-template='{{range .items}}{{.metadata.name}}{{"\n"}}{{end}}')
+externalStorageClusterPresent=$(oc get storagecluster -n openshift-storage -o go-template='{{range .items}}{{.spec.externalStorage.enable}}{{"\n"}}{{end}}')
 
-if [ "$(oc get storagecluster -n openshift-storage -o go-template='{{range .items}}{{.spec.externalStorage.enable}}{{"\n"}}{{end}}')" == true ]; then
-    echo "Skipping the ceph collection as External Storage is enabled" | tee -a  "${BASE_COLLECTION_PATH}"/gather-debug.log
+if [ "${externalStorageClusterPresent}" == true ]; then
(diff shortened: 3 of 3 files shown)


crombus (changes requested):
the directions is right but we duplicating are scripts not a good idea.

Few changes:
1) the new script should be predecessor for ceph script.
means if external mode enabled only the new script should be called 
and if not external setup than both scripts should be called.
2) all the work done in new script should be removed from ceph collection script.
3) the new script is independent of type installation that means whether the installation is external or internal this data should be collected . So name must belike gather_ceph_resources and the name of the old script should be gather_ceph_internal_resources.

rajatsing:
> the directions is right but we duplicating are scripts not a good idea.
> 
> Few changes:
> 
> 1. the new script should be predecessor for ceph script.
>    means if external mode enabled only the new script should be called
>    and if not external setup than both scripts should be called.
> 2. all the work done in new script should be removed from ceph collection script.

No, I can't because what if no cluster exists?. 

> 3. the new script is independent of type installation that means whether the installation is external or internal this data should be collected . So name must belike gather_ceph_resources and the name of the old script should be gather_ceph_internal_resources.
My main reason to do it this way because we are short on time and I want to get this patch in asap, I know I can improve it further, but it would also work with the way it is right now.

agarwal-mudit:
Doesn't look like duplicate code to me or I am missing something.
Agree with file name change but we can do that in another PR.

crombus:
> > the directions is right but we duplicating are scripts not a good idea.
> > Few changes:
> > 
> > 1. the new script should be predecessor for ceph script.
> >    means if external mode enabled only the new script should be called
> >    and if not external setup than both scripts should be called.
> > 2. all the work done in new script should be removed from ceph collection script.
> 
> No, I can't because what if no cluster exists?.
> 
> > 1. the new script is independent of type installation that means whether the installation is external or internal this data should be collected . So name must belike gather_ceph_resources and the name of the old script should be gather_ceph_internal_resources.
> >    My main reason to do it this way because we are short on time and I want to get this patch in asap, I know I can improve it further, but it would also work with the way it is right now.

That is a separate condition why would it conflict with this change.

crombus:
> Doesn't look like duplicate code to me or I am missing something.
https://github.com/openshift/ocs-operator/blob/219c14b46fcee33a4240364f944a10459188a0bb/must-gather/collection-scripts/gather_ceph_resources#L1-L21
> Agree with file name change but we can do that in another PR.

rajatsing:
> > Doesn't look like duplicate code to me or I am missing something.
> > https://github.com/openshift/ocs-operator/blob/219c14b46fcee33a4240364f944a10459188a0bb/must-gather/collection-scripts/gather_ceph_resources#L1-L21
> > 
> > Agree with the file name change but we can do that in another PR.

Done, but we have to do the entire https://github.com/openshift/ocs-operator/blob/master/must-gather/collection-scripts/gather_ceph_resources#L1-L13 for both the files. I also changed the file name to `gather_common_ceph_resource` since when the external cluster is present,  only limited ceph resources are collected.

must-gather/Dockerfile
```diff
@@ -5,5 +5,6 @@ COPY collection-scripts /usr/bin/
 
 RUN mkdir -p  /templates
 COPY templates /templates
+RUN ["chmod", "+x", "/usr/bin/gather_common_ceph_resources"]
```
crombus, line 8: why are you doing this.

crombus (changes requested):

must-gather/collection-scripts/gather_ceph_resources
```diff
@@ -11,14 +14,14 @@ if [ "${BASE_COLLECTION_PATH}" = "" ]; then
 fi
 
 CEPH_COLLECTION_PATH="${BASE_COLLECTION_PATH}/ceph"
+# pre-install-ceph-resources.sh
+
+gather_external_ceph_resources ${BASE_COLLECTION_PATH} "${SINCE_TIME}"
```
jarrpa, line 19: This file doesn't seem to exist...?

must-gather/collection-scripts/gather_common_ceph_resources
```diff
@@ -0,0 +1,28 @@
...
+fi
+
+CEPH_COLLECTION_PATH="${BASE_COLLECTION_PATH}/ceph"
+# pre-install-ceph-resources.sh
+
+# External Ceph resources
+external_ceph_resources=()
+external_ceph_resources+=(cephobjectstores)
+external_ceph_resources+=(cephobjectstoreusers)
+external_ceph_resources+=(cephclusters)
```
jarrpa, line 21: Probably rename these to `common_*`

must-gather/collection-scripts/gather_common_ceph_resources
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+# Expect base collection path as an argument
+BASE_COLLECTION_PATH=$1
```
jarrpa, line 4: Can't you do `BASE_COLLECTION_PATH="${$1:-pwd}`?

must-gather/Dockerfile
jarrpa: Looks like the `+x` is missing in the file script he commited. Just run this locally and `git add` the change.

jarrpa (changes requested):

jarrpa:
Fix the ci errors. Please always run `make ocs-operator-ci` before pushing. :)

rajatsing:
> Fix the ci errors. Please always run `make ocs-operator-ci` before pushing. :)

the ci errors are not coming from must-gather, they are from some other part of ocs-operator.

must-gather/collection-scripts/gather_common_ceph_resources
rajatsing: This will invalidate https://github.com/openshift/ocs-operator/blob/master/must-gather/collection-scripts/gather_ceph_resources#L8-L11

must-gather/collection-scripts/gather_ceph_resources
rajatsing: yes, my bad,some files did not get committed, thanks

must-gather/Dockerfile
rajatsing: yes, thanks, I was not aware of how to deal perm issue while creating new file, I was confused because the files are copied into a docker file, it was an easy fix, thanks

must-gather/collection-scripts/gather_common_ceph_resources
rajatsing: I will raise a separate pr for refactoring

must-gather/collection-scripts/gather_common_ceph_resources
jarrpa: How? This would make it so that if `$1` is not specified just use the output `pwd`. `gather_ceph_resources` will always call it with `BASE_COLLECTION_PATH` as `$1`, so there's no need for the `if` later on in this file.

crombus: @jarrpa the suggestion is correct but we are doing this way in all must-gather scripts. It should be changed to the suggested method in one separate commit. Does it sounds good.

must-gather/collection-scripts/gather_common_ceph_resources
crombus: @rajatsing you are introducing this change in this PR so you can  update these changes in this PR. I don’t think there should need for separate PR. Thoughts?

must-gather/collection-scripts/gather_ceph_resources
```diff
@@ -12,13 +15,12 @@ fi
 
 CEPH_COLLECTION_PATH="${BASE_COLLECTION_PATH}/ceph"
 
+gather_common_ceph_resources "${BASE_COLLECTION_PATH}" "${SINCE_TIME}"
```
crombus, line 18: remove this

crombus (changes requested):

must-gather/collection-scripts/gather_ceph_resources
rajatsing: done!.

jarrpa (approved):
/lgtm

PR #1221, opened 2021-06-10
Bumping NooBaa version

jeniawhite:
Bumped the version/images of NooBaa core and operator


jarrpa (approved):
In the interest of time, I'll take this as-is. In the future, I would ask that you split PRs like this into multiple commits, at least one for any manual changes and then another for all the generated changes.

/ok-to-test
/lgtm

PR #1225, opened 2021-06-14
Don't report Ready when CephCluster has not reached ClusterStateCreated

jarrpa:
There was a strange issue where the StorageCluster would report `Ready` even though it had an unhealthy CephCluster still pending. This was because when the operator detected a divergence that required an `Update()` in the CephCluster it would make the update and then `return nil` before checking the Status of the CephCluster in question.

This PR also refactors `TestEnsureCephCluster()` to make it easier to see the changes that were needed to reproduce the bug and validate the fix.

### DIFF
--- a/controllers/storagecluster/cephcluster.go
+++ b/controllers/storagecluster/cephcluster.go
@@ -155,35 +155,8 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
 		return err
 	}
 
-	// Update the CephCluster if it is not in the desired state
-	if !reflect.DeepEqual(cephCluster.Spec, found.Spec) {
-		r.Log.Info("Updating spec for CephCluster")
-		if !sc.Spec.ExternalStorage.Enable {
-			// Check if Cluster is Expanding
-			if len(found.Spec.Storage.StorageClassDeviceSets) < len(cephCluster.Spec.Storage.StorageClassDeviceSets) {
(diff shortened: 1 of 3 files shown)


umangachapagain (commented):
How does this fix the problem? You're still updating CephCluster and returning nil without looking into Status.

Also, when we discussed this issue long back, I was told that StorageCluster health should not depend on CephCluster health. So, what changed now?

jarrpa:
> How does this fix the problem? You're still updating CephCluster and returning nil without looking into Status.

Moving the DeepEquals() comparison to the bottom lets the function run through the Status checks before trying to update the spec. So if no errors arise we can safely `return nil` since we will have updated `r.conditions` to tell the reconciler the CephCluster is not ready. Just the fact that we detected a divergence and had to issue an `Update()` is not an error.

> Also, when we discussed this issue long back, I was told that StorageCluster health should not depend on CephCluster health. So, what changed now?

Hmm, that's not quite right, so if I gave that impression I apologize. There's probably a number of things going on here. A StorageCluster should reflect the full state of all its components, and if any of them are not 100% healthy its Status should reflect that. What SHOULDN'T be happening is that ocs-operator should not be reporting its readiness based on the state of the StorageCluster. Unfortunately, I think the UI has come to rely on this for some of its behaviors, which we'll probably need to change for ODf 4.9.

umangachapagain (approved):

umangachapagain (commented):
/lgtm

PR #1229, opened 2021-06-16
Add RBAC for creating service monitors from mgr sidecar

travisn:
Stretch clusters have two mgrs, each with a sidecar for updating the services and service monitor if the active mgr changes in Ceph. The mgr pod needs access to update the service monitor.

In order for the monitoring RBAC to be picked up, it needs to be added to the OCS operator in this folder:
https://github.com/openshift/ocs-operator/tree/master/rbac

This corresponds to the RBAC that was added upstream in this PR: https://github.com/rook/rook/pull/8118


jarrpa (approved):
/lgtm

PR #1231, opened 2021-06-17
Add ability to force reconcile, regardless of platform

### LINKED ISSUE #960: Remove platform constraint from CephObjectStore
In the spirit of embracing alternative outcomes for the OCS experience, we discussed providing a means to disable or ignore the platform detection logic that currently prevents CephObjectStore deployments in the public cloud. The reconcileStrategy functionality that was introduced to lend more control over which OCS components are deployed, seems like a natural way to do this, even if we decide we should continue to not install a CephObjectStore by default.

The intent of course is not that this be some hidden flag for folks at Red Hat to use, but rather a choice that customers can make and that the result would be a fully supported configuration.

mmgaggle:
Add the ability to force reconcile cephobjectstores and cephobjecstoreusers
regardless of platform. Useful for testing, workshops, demos. Resolves #960

### DIFF
--- a/controllers/storagecluster/cephobjectstores.go
+++ b/controllers/storagecluster/cephobjectstores.go
@@ -21,22 +21,24 @@ func (obj *ocsCephObjectStores) ensureCreated(r *StorageClusterReconciler, insta
 	reconcileStrategy := ReconcileStrategy(instance.Spec.ManagedResources.CephObjectStores.ReconcileStrategy)
 	if reconcileStrategy == ReconcileStrategyIgnore {
 		return nil
-	} else if reconcileStrategy == ReconcileStrategyForce {
-                avoid = False
-        } else {
-	        avoid, err := r.PlatformsShouldAvoidObjectStore()
-	        if err != nil {
-                        return err
-	        }
-        }
+	}
+
+	avoid, err := r.PlatformsShouldAvoidObjectStore()
+	if err != nil {
+		return err
+	}
 
 	if avoid {
 		platform, err := r.platform.GetPlatform(r.Client)
 		if err != nil {
 			return err
 		}
-		r.Log.Info(fmt.Sprintf("not creating a CephObjectStore because the platform is '%s'", platform))
-		return nil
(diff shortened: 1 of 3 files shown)


jarrpa:
With the way you're doing you're going to have to touch a lot of files to keep it consistent. It might make more sense to introduce a new boolean higher in the spec, something like `Spec.DisablePlatformDetection` that either circumvents `PlatformsShouldAvoidObjectStore()` or makes it always return `false`. The boolean's name needs to be such that the default of `false` preserves current behaviors.

oritwas:
I prefer the force strategy. It can be used for other use cases we will need  to make sure there is no reconciling. Platform is one example but we will have others in the future.

jarrpa:
> I prefer the force strategy. It can be used for other use cases we will need to make sure there is no reconciling. Platform is one example but we will have others in the future.

This is not the use case. In this case this *is* reconciling, we're just ignoring the part that says to not create it. And I don't think there's any other managed resource under the StorageCluster that needs anything like this, we always create all other default CRs.

We could still use the boolean approach, it'd just be renamed `Spec.ManagedResources.CephObjectStore.Force`.

mmgaggle:
While it is technically reconciling, that reconciliation is effectively a no-op with the platform detection in place. So my idea was you would "force" a meaningful reconciliation by ignoring the platform detection.

I did notice that if I continue with my current approach, that I'll also need to add conditional tests for route / service generation in addition to what I already have here for cephobjectstores and cephobjecstoreuser. I did create an operator image and test out what I have, and it resulted in the desired outcome (RGW deployment in EC2).

snecklifter:
If it helps, I've just hit a use-case for this where I had a typo in the local storage class name and was unable to trigger reconciliation. Its not a huge problem as I just tear down the cluster which will always be empty in this scenario however useful to not have to do that.

jarrpa:
I kind of lost track of this one, but closing this PR for now. Feel free to rebase and re-open if it still seems useful.

PR #1239, opened 2021-06-24
Remove quickstarts from OCS

vbnrh:
For 4.9 release, OCS operator will be replaced with the ODF at a higher level. Hence we would want to make sure that new quickstarts accurately represent the workflows for ODF operator.
This commit includes all the relevant changes that are applicable to newer UI that is coming as part of OCP 4.9


controllers/storagecluster/quickstart_constants.go
```diff
@@ -1,200 +1,181 @@
...
-   3. Click the Action menu (⋮)-> Add Storage
-
-   4. Select \"Create new claim\". You would select \"Use existing claim\" here if you wanted to restore an existing persistent volume to a redeployed application.
+        **Connect your application with a PVC**
+ 
+           1. Click Workloads > Deployments in the navigation menu on the left.  
+   
+           2. Select your project from the Project drop-down and find your application in the list of deployments.  
+   
+           3. Click the Action menu &*⋮&* > Add Storage
```
vbnrh, line 48: Just a quick note on this line. 
I am replacing (&*) with ` (backtick)  as  go multiline strings do not support escaping backtick . The string enclosed by backticks are highlighted in markdown.

vbnrh: If anyone got a better method for using backticks with multiline strings. they could suggest it here.

jarrpa (changes requested):
This is not valid for ocs-operator. The quickstarts should be moved to https://github.com/red-hat-data-services/odf-operator for 4.9. Although I'm seeing changes of 4.7 -> 4.9, meaning that these are also out of date for OCS 4.8. That should probably be something to update and backport to release-4.8, but given where we are with dev freeze we may need to raise a BZ and see if QE will approve it.

vbnrh:
> Although I'm seeing changes of 4.7 -> 4.9, meaning that these are also out of date for OCS 4.8. That should probably be something to update and backport to release-4.8, but given where we are with dev freeze we may need to raise a BZ and see if QE will approve it.

The version is just the resource version that i had added in 4.7 when i initially added these, it has no bearing on the functionality of the quickstarts. The quickstart still holds well for OCS 4.8 as well. We may remove the version information if needed.  
Lot of quickstarts in https://github.com/openshift/console-operator/tree/master/quickstarts have similiar pattern.

> This is not valid for ocs-operator. The quickstarts should be moved to https://github.com/red-hat-data-services/odf-operator for 4.9. 

This is an interesting question here and has me confused as well. 
The quickstarts provide guidance for Day 1 and Day 2 operations on the storage cluster which is managed by OCS . ODF on the other hand manages StorageSystem which has no effect on the operations that are listed here.  We are only rebranding the name here to ODF for the UI.

I believe these resources should be managed by OCS itself  and not ODF as the user will be directly interacting with the StorageCluster CR for these operations. 

@jarrpa Let me know your thoughts on this?

jarrpa:
> The version is just the resource version that i had added in 4.7 when i initially added these, it has no bearing on the functionality of the quickstarts. The quickstart still holds well for OCS 4.8 as well. We may remove the version information if needed.
> Lot of quickstarts in https://github.com/openshift/console-operator/tree/master/quickstarts have similiar pattern.

That's fine, if it can be removed all the better. Still, as of now they will he showing "4.7" in OCS 4.8. How urgent that is or isn't is not my call to make, however.

> > This is not valid for ocs-operator. The quickstarts should be moved to https://github.com/red-hat-data-services/odf-operator for 4.9.
> 
> This is an interesting question here and has me confused as well.
> The quickstarts provide guidance for Day 1 and Day 2 operations on the storage cluster which is managed by OCS . ODF on the other hand manages StorageSystem which has no effect on the operations that are listed here. We are only rebranding the name here to ODF for the UI.
> 
> I believe these resources should be managed by OCS itself and not ODF as the user will be directly interacting with the StorageCluster CR for these operations.

Thing is, we are moving away from exposing the StorageCluster. All of the workflows described in these files will be moving to a different portion of the UI branded as "Storage Systems", which are CRs managed by odf-operator. Admins working through the GUI for normal day 1 and day 2 operations will never see the StorageCluster CR.

bipuladh:
This should be inside ODF Operator. And be removed everything from here. Items related to StorageCluster should stay here. The guides need to be totally revamped as per the new dashboard and systems list page.

vbnrh:
> This should be inside ODF Operator. And be removed everything from here. Items related to StorageCluster should stay here. The guides need to be totally revamped as per the new dashboard and systems list page.
> 

Discussed it with @jarrpa earlier.  Working on moving these to ODF. 

@bipuladh What items are you referring to though? 
If we move, we shall be moving both of the QS to ODF. 

The guides already have information corresponding to the current design and will change accordingly if the design changes

umangachapagain (commented):
@vbnrh  If we are moving it to ODF, can we close this PR?

vbnrh:
> @vbnrh If we are moving it to ODF, can we close this PR?

@umangachapagain I'll update this PR to remove Quickstarts.

vbnrh:
@jarrpa I was wondering if we should keep the permissions and uninstaller flow for this release (or move it in some other directory)
If user upgrades from OCS 4.8 to ODF 4.9, we may need to delete the older versions quickstarts as ODF will be creating the newer ones . 

How do we proceed about doing it ?

jarrpa:
Apologies for the extended delay. :pray: 

> @jarrpa I was wondering if we should keep the permissions and uninstaller flow for this release (or move it in some other directory)
> If user upgrades from OCS 4.8 to ODF 4.9, we may need to delete the older versions quickstarts as ODF will be creating the newer ones .
> 
> How do we proceed about doing it ?

We could have an operator make sure the old ones don't exist. That said, if the resource names of the Quickstarts are not visible to the admin we could just leave the names the same and just move the Quickstart functions to odf-operator. Either one would be fine. For now I'll merge this one and we can submit a follow-up PR if we decide to have ocs-operator deal with them.

/lgtm

PR #1240, opened 2021-06-24
logger: add zap flags to commandline

umangachapagain:
adding zap flags to commandline flags allows users to configure
logging at runtime by passing the right flags. These flags are
then used to instantiate a logger which is then used for logging
messages.

`--zap-devel=true|false` for running in development mode
`--zap-encoder=json|console` for output format
`--zap-log-level=debug|info|error` for log level

```
Development Mode defaults(encoder=consoleEncoder,logLevel=Debug,stackTraceLevel=Warn).

Production Mode defaults(encoder=jsonEncoder,logLevel=Info,stackTraceLevel=Error).
```
```
NOTE: this is in preparation for enhanced logging
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -79,22 +79,24 @@ func printVersion() {
 }
 
 func main() {
-	var isDevelopmentEnv bool
 	var probeAddr string
 	var metricsAddr string
 	var enableLeaderElection bool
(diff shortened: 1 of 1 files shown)


jarrpa (approved):
/lgtm

PR #1241, opened 2021-06-24
Updates DEBUG_EXPORTS.sh and adds new make target

umangachapagain:
`make run` to run operator locally and `source ./extras/DEBUG_EXPORTS.sh` to source required environment variables.


jarrpa (approved):
Oh hey, this is a cool idea! LGTM, but is there any reason that we can't automate the global var exports? Looks like there might be a few options available: https://stackoverflow.com/questions/7507810/how-to-source-a-script-in-a-makefile

/lgtm

PR #1249, opened 2021-07-01
MUST-GATHER: LOGS OF BG PROCESS IN PRE-INSTALL SCRIPT

parthdhanjal:
Signed-off-by: parthdhanjal <dparth@redhat.com>

### DIFF
--- a/must-gather/collection-scripts/pre-install.sh
+++ b/must-gather/collection-scripts/pre-install.sh
@@ -56,7 +56,7 @@ labels(){
 check_for_debug_pod(){
     debug_pod_name=$(oc get pods -n openshift-storage | grep "${node//./}-debug" | awk '{print $1}')
     # sleep for 60 seconds giving time for debug pod to get created
-    sleep 60 
+    sleep 60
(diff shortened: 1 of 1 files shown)


must-gather/collection-scripts/pre-install.sh
```diff
@@ -82,5 +82,5 @@ for node in ${nodes}; do
 done
 
 # wait for all pids
```
crombus, line 84: add the if condition if the pid array is empty it should not print.

must-gather/collection-scripts/pre-install.sh
```diff
@@ -56,7 +56,7 @@ labels(){
 check_for_debug_pod(){
     debug_pod_name=$(oc get pods -n openshift-storage | grep "${node//./}-debug" | awk '{print $1}')
     # sleep for 60 seconds giving time for debug pod to get created
-    sleep 60 
+    sleep 60
```
crombus, line 59: maybe a separate commit.

crombus (changes requested):
For the commit msg I think no hard requirement for it be in capital letters.

crombus (approved):

PR #1252, opened 2021-07-02
hack/source-manifests: Update noobaa cmd

b-ranto:
Noobaa operator no longer accepts `olm csv yaml` command for dumping the
noobaa CSV. It is just `olm csv`, now. The new command still works in
older noobaa operators (4.8 at least), only the support for the yaml
suffix was dropped in 4.9.


agarwal-mudit (approved):
/lgtm

PR #1255, opened 2021-07-02
must-gather: remove the standard pod template

agarwal-mudit:
We don't need two pod templates now, removing the standard
pod template and the code around it.

Signed-off-by: Mudit Agarwal <muagarwa@redhat.com>


crombus (approved):
/lgtm

PR #1142, opened 2021-04-01
Annotation for thick-provisioner and pool-management feature

GowthamShanmugam:
Signed-off-by: Gowtham Shanmugasundaram <gshanmug@redhat.com>

PR #1259, opened 2021-07-06
[release-4.8]Bug 1979012: Disable thick provisioning support from UI

bipuladh:
- Change feature guard annotations in CSV to diable thick provisioning from UI

Partial revert of: #1142

/assign @umangachapagain


umangachapagain (approved):

agarwal-mudit:
Can we keep these changes only in release-4.8?
I have done the same in https://github.com/openshift/ocs-operator/pull/1257

I have already checked the same with Eran, we don't want to disable this for 4.9 as of now.

agarwal-mudit (approved):
/lgtm

PR #1271, opened 2021-07-19
Ceph toolbox will run as non-root user

bindrad:
Container will run with privileged permissions but not as a root user.

Signed-off-by: bindrad <dbindra@redhat.com>

### DIFF
--- a/controllers/ocsinitialization/ocsinitialization_controller.go
+++ b/controllers/ocsinitialization/ocsinitialization_controller.go
@@ -59,6 +59,8 @@ func newToolsDeployment(namespace string, rookImage string) *appsv1.Deployment {
 	var replicaOne int32 = 1
 
 	privilegedContainer := true
+	runAsNonRoot := true
+	var runAsUser int64 = 1000
 	return &appsv1.Deployment{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      name,
@@ -106,12 +108,12 @@ func newToolsDeployment(namespace string, rookImage string) *appsv1.Deployment {
 								},
 							},
 							SecurityContext: &corev1.SecurityContext{
-								Privileged: &privilegedContainer,
+								Privileged:   &privilegedContainer,
+								RunAsNonRoot: &runAsNonRoot,
+								RunAsUser:    &runAsUser,
 							},
 							VolumeMounts: []corev1.VolumeMount{
-								{Name: "dev", MountPath: "/dev"},
-								{Name: "sysbus", MountPath: "/sys/bus"},
-								{Name: "libmodules", MountPath: "/lib/modules"},
+								{Name: "ceph-config", MountPath: "/etc/ceph"},
 								{Name: "mon-endpoint-volume", MountPath: "/etc/rook"},
 							},
 						},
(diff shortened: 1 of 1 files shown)


jarrpa (approved):
/lgtm

controllers/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -106,7 +108,9 @@ func newToolsDeployment(namespace string, rookImage string) *appsv1.Deployment {
 								},
 							},
 							SecurityContext: &corev1.SecurityContext{
-								Privileged: &privilegedContainer,
+								Privileged:   &privilegedContainer,
+								RunAsNonRoot: &runAsNonRoot,
+								RunAsUser:    &runAsUser,
 							},
 							VolumeMounts: []corev1.VolumeMount{
 								{Name: "dev", MountPath: "/dev"},
```
travisn, line 116: You'll want to remove all the volume mounts except mon-endpoint-volume. See the [toolbox spec](https://github.com/rook/rook/blob/master/cluster/examples/kubernetes/ceph/toolbox.yaml) we are using upstream that doesn't have the mounts and isn't privileged. To be safe also add the volume for /etc/ceph to match upstream.

jarrpa:
Holding based on @travisn review comment.

/hold

agarwal-mudit:
Also, Elad raised a concern about this change affecting ocs-ci.
So, ideally we should not merge it in 4.8/4.7 before we test this in master/4.9

jarrpa:
> Also, Elad raised a concern about this change affecting ocs-ci.
> So, ideally we should not merge it in 4.8/4.7 before we test this in master/4.9

Fair enough. We can also close the bot PRs and submit final versions if needed.

controllers/ocsinitialization/ocsinitialization_controller.go
bindrad: @travisn I have made the changes according to the comments.

jarrpa: @travisn @bindrad If the upstream spec isn't marked as privileged, is there a reason for our d/s deployment to be privileged?

bindrad: @jarrpa If I run the container as an unprivileged user, the container shows an error as the entry point command can't be executed.

nb-ohad: @jarrpa @travisn It might be that we can do better, but the current changes meet our security concerns and need to go into 4.8. As the deadline for that is near I will suggest to merge this one and introduce any further changes in a future PR

nb-ohad:
> Also, Elad raised a concern about this change affecting ocs-ci.
So, ideally we should not merge it in 4.8/4.7 before we test this in master/4.9

We did test the change manually, running the pod and invoking a set of ceph commands from the pods with failures.

controllers/ocsinitialization/ocsinitialization_controller.go
travisn: @bindrad Did you try removing the security context, or you just found that specifying a non-privileged user failed? Upstream it works fine on openshift to have the default security context. Specifying the user with a privileged context is an improvement over privileged, but would be good to understand why the default security context can't be used.

bindrad: @travisn If I use the default Security context then the container run as a root user

travisn: It would be non-privileged, which I assumed was the intention. But if the goal is to run as a non-root user, then this looks good.

bindrad: Yes, we need to run the container as non-root user

nb-ohad: @travisn I do agree that non-privileged will be better, but lets try to accomplish that in another PR

jarrpa: Sounds good to me :+1:

jarrpa (approved):
/lgtm

PR #1275, opened 2021-07-21
StorageCluster spec and controller changes to enable CephBlockPool mirroring

umangachapagain:
+ New spec for configuring mirroring.
Currently it only accepts "enabled" boolean and defaults to false.
This configuration only affects CephBlockPools being managed by OCS operator.
If `ReconcileStrategy` is set to `Ignore` or user manually creates a `CephBlockPool`,
users will have to manually enable mirroring on those pools.
```YAML
spec:
   mirroring:
      enabled: false
```


api/v1/storagecluster_types.go
```diff
@@ -67,6 +67,8 @@ type StorageClusterSpec struct {
 	// ArbiterSpec specifies the storage cluster options related to arbiter.
 	// If Arbiter is enabled, ArbiterLocation in the NodeTopologies must be specified.
 	Arbiter ArbiterSpec `json:"arbiter,omitempty"`
+	// DataMirroring specifies data mirroring configuration for the storage cluster.
+	DataMirroring MirroringSpec `json:"dataMirroring,omitempty"`
```
leseb, line 71: I wonder if "data" in `DataMirroring` is not implied. What about just `Mirroring`?

api/v1/storagecluster_types.go
```diff
@@ -234,6 +236,14 @@ type EncryptionSpec struct {
 	KeyManagementService KeyManagementServiceSpec `json:"kms,omitempty"`
 }
 
+// MirroringSpec defines if data mirroring should be enabled for the StorageCluster.
+// It is optional and defaults to false.
+type MirroringSpec struct {
+	// +kubebuilder:default:=true
```
Madhu-1, line 242: Are we planning to enable DR by default in 4.9? AFAIK the plan is to enable DR as a day 2 operation.
@ShyamsundarR

api/v1/storagecluster_types.go
umangachapagain: I started with `Mirroring`, then switched to `DataMirroring` because I felt the need to be explicit about it.
But, since `Mirroring` is used everywhere else, I guess it's better to be consistent.

api/v1/storagecluster_types.go
umangachapagain: I'm confused. Didn't we want to enable mirroring by default?

controllers/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -338,6 +338,8 @@ func newRookCephOperatorConfig(namespace string) *corev1.ConfigMap {
 	data["CSI_PROVISIONER_TOLERATIONS"] = defaultCSIToleration
 	data["CSI_PLUGIN_TOLERATIONS"] = defaultCSIToleration
 	data["CSI_LOG_LEVEL"] = "5"
+	data["CSI_ENABLE_OMAP_GENERATOR"] = "true"
+	data["CSI_ENABLE_VOLUME_REPLICATION"] = "true"
```
umangachapagain, line 342: @Madhu-1 @leseb @ShyamsundarR  I am even enabling these by default. Let me know if it isn't a good idea.

Madhu-1: IMo this should be set only if `mirroring.enabled` is `true`.

umangachapagain: But, it is enabled by default!!

Madhu-1: what if the user chooses to change the mirroring.enabled to `false` and also from the upgrade point of view looks like `ensureRookCephOperatorConfig` will not update the configmap if the configmap already exists? what about the upgrade path?

umangachapagain: ahh I forgot about the upgrade path. We need to patch config whenever mirroring is enabled.
Thanks for pointing it out.

api/v1/storagecluster_types.go
umangachapagain: Set it to disabled by default.

controllers/ocsinitialization/ocsinitialization_controller.go
umangachapagain: It's still not clear how we handle operator config. Dropped these changes for now.
It seems this affects more than just pool mirroring.

jarrpa (approved):
/lgtm

PR #1276, opened 2021-07-21
controller: inject default label to CephCluster

umangachapagain:
Injecting default label to CephCluster monitoring labels.
`rook.io/managedBy=<StorageClusterName>` label is used to reference
owner of a CephCluster. This is merged with user provided
labels from StorageCluster.Monitoring.Labels.

If users provide label with `rook.io/managedBy` key, it will be overwritten by default.

```
NOTE:
A mutation webhook injecting labels to StorageCluster would be better,
but we don't have that today. So, this is minimal, less intrusive change
that doesn't require operator to update StorageCluster CR.

We cannot use default markers on api spec as we want this label to persist.
```

### DIFF
--- a/controllers/storagecluster/cephcluster.go
+++ b/controllers/storagecluster/cephcluster.go
@@ -276,6 +276,15 @@ func (obj *ocsCephCluster) ensureDeleted(r *StorageClusterReconciler, sc *ocsv1.
 
 }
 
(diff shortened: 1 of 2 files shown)


jarrpa (changes requested):

jarrpa (approved):
/lgtm

PR #1280, opened 2021-07-23
Clean up before running must-gather

rewantsoni:
Signed-off-by: Rewant Soni <resoni@redhat.com>

### DIFF
--- a/must-gather/collection-scripts/pre-install.sh
+++ b/must-gather/collection-scripts/pre-install.sh
@@ -58,6 +58,22 @@ check_for_helper_pod(){
     oc wait -n openshift-storage --for=condition=Ready pod/"${HOSTNAME}"-helper --timeout=200s
 }
 
+cleanup() {
+  echo "checking for existing must-gather resource" | tee -a "${BASE_COLLECTION_PATH}"/gather-debug.log
+  pods=$(oc get pods --no-headers -n openshift-storage -l must-gather-helper-pod='' | awk '{print $1}')
+  if [ -n "${storageClusterPresent}" ] && [ -n "${pods}" ]; then
+    SAVEIFS=$IFS # Save current IFS
+    IFS=$'\n'    # Change IFS to new line
+    pods=("$pods") # split to array $pods
+    IFS=$SAVEIFS # Restore IFS
+    echo "deleting existing must-gather resource" | tee -a "${BASE_COLLECTION_PATH}"/gather-debug.log
+    for pod in "${pods[@]}"; do
+      oc delete pod "${pod}" -n "${ns}"
+    done
+  fi
+}
+
+cleanup
 deploy
 labels
 pids=()


must-gather/collection-scripts/pre-install.sh
```diff
@@ -58,6 +58,22 @@ check_for_helper_pod(){
     oc wait -n openshift-storage --for=condition=Ready pod/"${HOSTNAME}"-helper --timeout=200s
 }
 
+cleanup() {
+  echo "checking for existing must-gather resource" | tee -a "${BASE_COLLECTION_PATH}"/gather-debug.log
+  pods=$(oc get pods --no-headers -n openshift-storage -l must-gather-helper-pod='' | awk '{print $1}')
+  if [ -n "${storageClusterPresent}" ] && [ -n "${pods}" ]; then
+    SAVEIFS=$IFS # Save current IFS
```
agarwal-mudit, line 65: Just trying to understand, why do we need this logic (IFS/SAVEIFS)

must-gather/collection-scripts/pre-install.sh
```diff
@@ -58,6 +58,22 @@ check_for_helper_pod(){
     oc wait -n openshift-storage --for=condition=Ready pod/"${HOSTNAME}"-helper --timeout=200s
 }
 
+cleanup() {
+  echo "checking for existing must-gather resource" | tee -a "${BASE_COLLECTION_PATH}"/gather-debug.log
+  pods=$(oc get pods --no-headers -n openshift-storage -l must-gather-helper-pod='' | awk '{print $1}')
```
agarwal-mudit, line 63: Are we expecting more than one pods?

must-gather/collection-scripts/pre-install.sh
```diff
@@ -58,6 +58,22 @@ check_for_helper_pod(){
...
+  if [ -n "${storageClusterPresent}" ] && [ -n "${pods}" ]; then
+    SAVEIFS=$IFS # Save current IFS
+    IFS=$'\n'    # Change IFS to new line
+    pods=($pods) # split to array $pods
+    IFS=$SAVEIFS # Restore IFS
+    echo "deleting existing must-gather resource" | tee -a "${BASE_COLLECTION_PATH}"/gather-debug.log
+    for pod in "${pods[@]}"; do
+      oc delete pod "${pod}" -n "${ns}"
```
agarwal-mudit, line 71: Why don't we use the same logic which is there in post-install script

```
if [ -n "$(oc get storagecluster -n openshift-storage -o go-template='{{range .items}}{{.metadata.name}}{{"\n"}}{{end}}')" ]; then
    oc delete -f pod_helper.yaml
fi

```

must-gather/collection-scripts/pre-install.sh
```diff
@@ -58,6 +58,22 @@ check_for_helper_pod(){
     oc wait -n openshift-storage --for=condition=Ready pod/"${HOSTNAME}"-helper --timeout=200s
 }
 
+cleanup() {
+  echo "checking for existing must-gather resource" | tee -a "${BASE_COLLECTION_PATH}"/gather-debug.log
+  pods=$(oc get pods --no-headers -n openshift-storage -l must-gather-helper-pod='' | awk '{print $1}')
+  if [ -n "${storageClusterPresent}" ] && [ -n "${pods}" ]; then
+    SAVEIFS=$IFS # Save current IFS
+    IFS=$'\n'    # Change IFS to new line
+    pods=($pods) # split to array $pods
```
agarwal-mudit, line 67: This should be put in quotes to avoid shellcheck error
"$pods"

must-gather/collection-scripts/pre-install.sh
rewantsoni: The pod template is updated in 
`apply_helper_pod() {
    < ${POD_TEMPLATE} safe_replace "NAMESPACE" "$1" | safe_replace "IMAGE_NAME" "$2" | safe_replace "MUST_GATHER" "$HOSTNAME" > pod_helper.yaml
    oc apply -f pod_helper.yaml
}`
will the variables in pod_helper.yaml be the same when we run a different instance of must-gather? If that is true then we can call the post-install script here and resue the script

agarwal-mudit: Not sure, can you please check once.

rewantsoni: I will try running the clean up with post-install script and see how it works

agarwal-mudit: Ok, thanks for checking. Lets go ahead with this.

agarwal-mudit (approved):
/lgtm

PR #1283, opened 2021-07-26
ocs: replace avoid with skip where possible

nbalacha:
Replace "avoid" with "skip" in functions and variables whenever
it better clarifies what the code is doing.


jarrpa (approved):
/lgtm

PR #1284, opened 2021-07-27
[WIP] Changing the way external endpoint IP/Port being propagated

aruniiird:
Previously external cluster's monitoring endpoints were attached to
Reconcile object's variables and passed to other parts of the code.

Now, we are creating an external endpoint configmap which holds the
monitoring endpoint details. Resources (like CephCluster) which requires
these endpoint details, can access it through the endpoint configmap.


aruniiird:
Redundant. Opened another one with a different logical approach.

PR #1292, opened 2021-08-05
Initial changes for debugging.

rexagod:
"-s -w" ldflags were removed to enable `dlv` support. "all=-N -l"
`gcflags` were added to disable optimizations and inlining, respectively.

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,7 +1,21 @@
+FROM golang:latest AS builder
+
+RUN GOOS=linux GOARCH=amd64 GO111MODULE=on go get github.com/go-delve/delve/cmd/dlv
+
+# ==============================================
+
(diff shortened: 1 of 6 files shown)


jarrpa:
I'd like to explore this again later. The idea was sound, and I would certainly like to have something like this in the project, but better integration with env var toggles would be more ideal.

rexagod:
Hey @jarrpa, thank you for that comment. I agree, and I was personally not happy with this implementation as it left a lot to the devs to figure out. I do aim to make this way more automated, without any manual interventions whatsoever, and I already have some ideas which I'll get to once I'm done with [this](https://www.oreilly.com/library/view/programming-kubernetes/9781492047094/), just so I can make sure they are robust.

Thanks!

PR #1293, opened 2021-08-05
storagecluster: move cephCluster deletion at the end

iamniting:
with the new uninstall behaviour of the rook we should delete
cephCluster at the end after all related resources are deleted.

cc @raghavendra-talur @BlaineEXE

### DIFF
--- a/controllers/storagecluster/uninstall_reconciler.go
+++ b/controllers/storagecluster/uninstall_reconciler.go
@@ -284,7 +284,6 @@ func (r *StorageClusterReconciler) deleteResources(sc *ocsv1.StorageCluster) err
 
 	objs := []resourceManager{
 		&ocsNoobaaSystem{},
-		&ocsCephCluster{},
 		&ocsCephRGWRoutes{},
 		&ocsCephObjectStoreUsers{},
 		&ocsCephObjectStores{},
@@ -292,6 +291,7 @@ func (r *StorageClusterReconciler) deleteResources(sc *ocsv1.StorageCluster) err
 		&ocsCephBlockPools{},
 		&ocsSnapshotClass{},
 		&ocsStorageClass{},
+		&ocsCephCluster{},
 	}
 
 	for _, obj := range objs {


umangachapagain (approved):

BlaineEXE (approved):

umangachapagain (changes requested):
Just realized that this causes other issues.

In https://github.com/openshift/ocs-operator/pull/1224 , I'm trying to ensure that we create CephCluster before anything else that needs it.

Here, we're doing exactly opposite. It makes sense for delete but that's not how we should deal with it. Instead the entire `objs` list needs to be reversed and processed for deletion. Deletion should be the exact opposite of Creation.

iamniting:
> Just realized that this causes other issues.
> 
> In #1224 , I'm trying to ensure that we create CephCluster before anything else that needs it.
> 
> Here, we're doing exactly opposite. It makes sense for delete but that's not how we should deal with it. Instead the entire `objs` list needs to be reversed and processed for deletion. Deletion should be the exact opposite of Creation.

But rook will block deletion of other resources if we will not delete cephcluster first. We tried to preserve the exact opposite of creation in the past but that caused problems and we end up changing the order.

umangachapagain:
> But rook will block deletion of other resources if we will not delete cephcluster first. We tried to preserve the exact opposite of creation in the past but that caused problems and we end up changing the order.
>
That doesn't sound like a normally expected behavior. Consumer needs to be removed first.

It looks like we either break Install or we break Uninstall.

travisn:
> > But rook will block deletion of other resources if we will not delete cephcluster first. We tried to preserve the exact opposite of creation in the past but that caused problems and we end up changing the order.
> 
> That doesn't sound like a normally expected behavior. Consumer needs to be removed first.
> 
> It looks like we either break Install or we break Uninstall.

The behavior we had before was really a hack to force deletion of the resources even though they were still in use. Now Rook will really enforce the proper ordering for uninstall since you have to remove the child resources before you remove the cluster CR. So if the intention is really that OCS force removes the cluster even if resources are in use, OCS could 1) delete the cluster CR, 2) remove the finalizer from the cluster CR to force its removal (thus ignoring the new behavior), and then 3) proceed to delete other resources.

umangachapagain:
```
  conditions:
  - lastHeartbeatTime: "2021-08-19T09:57:39Z"
    lastTransitionTime: "2021-08-19T09:42:16Z"
    message: 'CephCluster "openshift-storage/test-storagecluster-cephcluster" will
      not be deleted until all dependents are removed: CephBlockPools: [test-storagecluster-cephblockpool],
      CephFilesystems: [test-storagecluster-cephfilesystem]'
    reason: ObjectHasDependents
    status: "True"
    type: DeletionIsBlocked
  - lastHeartbeatTime: "2021-08-19T09:57:16Z"
    lastTransitionTime: "2021-08-19T09:40:01Z"
    message: Cluster created successfully
    reason: ClusterCreated
    status: "True"
    type: Ready
  - lastHeartbeatTime: "2021-08-19T09:57:39Z"
    lastTransitionTime: "2021-08-19T09:42:16Z"
    message: Deleting the CephCluster
    reason: ClusterDeleting
    status: "True"
    type: Deleting
```
I took this from CI run on PR to use Rook v1.7 in OCS.
This confirms the behavior @travisn is referring to.

So for a clean uninstall, CephCluster should be deleted last. CephBlockPool, CephFilesystem  etc should be deleted first.
And, for clean install CephCluster should be created first, followed by CephBlockPool, CephFilesystem and StorageClasses etc.
@travisn can you confirm this?
If this is true, this PR needs to move in a different direction as it is creating CephCluster at last.

travisn:
> So for a clean uninstall, CephCluster should be deleted last. CephBlockPool, CephFilesystem etc should be deleted first.

Agreed, the CephCluster should be deleted last for a normal uninstall. However, I believe the intention is that OCS needs to force delete Rook even if there are resources still in the cluster. For example, if there are any buckets or PVs still referencing the cluster, the deletion was being forced by first deleting the cluster CR,. Then the cleanup for buckets and PVs would ignore their checks if the cluster CR was already deleted. Now, the only way to force delete is to force remove finalizers instead of waiting for Rook to remove finalizers. 

> And, for clean install CephCluster should be created first, followed by CephBlockPool, CephFilesystem and StorageClasses etc.
> @travisn can you confirm this?
> If this is true, this PR needs to move in a different direction as it is creating CephCluster at last.

The creation order is not important. If the pool, filesystem, etc are created before the cluster, they will just continuously reconcile until the cluster is initialized. It's not possible to reconcile the pool, filesystem, and other child CRs until the mons, mgr, and OSDs are created.

jarrpa:
> > So for a clean uninstall, CephCluster should be deleted last. CephBlockPool, CephFilesystem etc should be deleted first.
> 
> Agreed, the CephCluster should be deleted last for a normal uninstall. However, I believe the intention is that OCS needs to force delete Rook even if there are resources still in the cluster. For example, if there are any buckets or PVs still referencing the cluster, the deletion was being forced by first deleting the cluster CR,. Then the cleanup for buckets and PVs would ignore their checks if the cluster CR was already deleted. Now, the only way to force delete is to force remove finalizers instead of waiting for Rook to remove finalizers.

That is an intended use case, yes. But i don't think our ability to force a messy uninstall will be affected by this change in order of operations.

> > And, for clean install CephCluster should be created first, followed by CephBlockPool, CephFilesystem and StorageClasses etc.
> > @travisn can you confirm this?
> > If this is true, this PR needs to move in a different direction as it is creating CephCluster at last.
> 
> The creation order is not important. If the pool, filesystem, etc are created before the cluster, they will just continuously reconcile until the cluster is initialized. It's not possible to reconcile the pool, filesystem, and other child CRs until the mons, mgr, and OSDs are created.

This. Now, conceptually, I neither understand nor remember why we didn't create the CephCluster first anyway. Still, since it more or less works fine right now it can be left as separate from the order of deletion.

travisn:
> > > So for a clean uninstall, CephCluster should be deleted last. CephBlockPool, CephFilesystem etc should be deleted first.
> > 
> > 
> > Agreed, the CephCluster should be deleted last for a normal uninstall. However, I believe the intention is that OCS needs to force delete Rook even if there are resources still in the cluster. For example, if there are any buckets or PVs still referencing the cluster, the deletion was being forced by first deleting the cluster CR,. Then the cleanup for buckets and PVs would ignore their checks if the cluster CR was already deleted. Now, the only way to force delete is to force remove finalizers instead of waiting for Rook to remove finalizers.
> 
> That is an intended use case, yes. But i don't think our ability to force a messy uninstall will be affected by this change in order of operations.

Changing the order of operations as proposed in this PR is needed for a clean uninstall. What I'm trying to say is that without additional changes, forced uninstalls are blocked either way.

umangachapagain:
> Changing the order of operations as proposed in this PR is needed for a clean uninstall. What I'm trying to say is that without additional changes, forced uninstalls are blocked either way.
>
I believe we provide options for clean uninstall or force uninstall (done using annotations?) to users. So both should work.

And, wrt to install sequence, technically it doesn't matter. Logically CephCluster should be first. But that is from Rook perspective.
It might still make sense to maintain the right sequence in OCS Operator since we don't wait for things to be ready, we create and forget. At times we create StorageClass before CephCluster is ready and we end up in issues.

PR #1298, opened 2021-08-10
Allow DBStorageClass to be set from Storage Cluster CR for Noobaa

ezio-auditore:
Signed-off-by: Kaustav Majumder kmajumder@redhat.com

### DIFF
--- a/controllers/storagecluster/noobaa_system_reconciler_test.go
+++ b/controllers/storagecluster/noobaa_system_reconciler_test.go
@@ -23,9 +23,10 @@ import (
 )
 
 const (
-	coreEnvVar          = "NOOBAA_CORE_IMAGE"
-	dbEnvVar            = "NOOBAA_DB_IMAGE"
-	defaultStorageClass = "noobaa-ceph-rbd"
+	coreEnvVar                 = "NOOBAA_CORE_IMAGE"
+	dbEnvVar                   = "NOOBAA_DB_IMAGE"
+	defaultStorageClass        = "noobaa-ceph-rbd"
+	noobaDbDefaultStorageClass = "test-storage-class"
 )
 
 var noobaaReconcileTestLogger = logf.Log.WithName("noobaa_system_reconciler_test")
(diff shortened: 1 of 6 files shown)


controllers/storagecluster/noobaa_system_reconciler_test.go
```diff
@@ -304,11 +315,16 @@ func TestSetNooBaaDesiredState(t *testing.T) {
...
 		assert.Equalf(t, noobaa.Labels["app"], "noobaa", "[%s] expected noobaa Label mismatch", c.label)
 		assert.Equalf(t, noobaa.Name, "noobaa", "[%s] noobaa name not set correctly", c.label)
-		assert.Equal(t, *noobaa.Spec.DBStorageClass, fmt.Sprintf("%s-ceph-rbd", c.sc.Name))
+		if c.label == "case 6" {
+			assert.Equal(t, *noobaa.Spec.DBStorageClass, noobaDbDefaultStorageClass)
+			assert.Equal(t, *noobaa.Spec.PVPoolDefaultStorageClass, noobaDbDefaultStorageClass)
+		} else {
+			assert.Equal(t, *noobaa.Spec.DBStorageClass, fmt.Sprintf("%s-ceph-rbd", c.sc.Name))
+			assert.Equal(t, *noobaa.Spec.PVPoolDefaultStorageClass, fmt.Sprintf("%s-ceph-rbd", c.sc.Name))
+		}
```
jarrpa, line 324: Rather than doing this, extend the cases struct to have a new "dbStorageClassName" string field, or something, and set it to `generateNameForCephBlockPoolSC(sc, "")` as the default if not specified. Alternatively you could just do an `if` to check if `Spec.MultiCloudGateway.DbStorageClassName` (watch out for null pointers!) exists and use that if so. In the end I just don't want to see use using two Equal() calls on the same field with different values depending on use case label.

jarrpa (changes requested):
Code looks solid enough, just gotta tweak the test implementation a bit.

agarwal-mudit:
/retitle Allow DBStorageClass to be set from Storage Cluster CR for Noobaa

agarwal-mudit:
For PRs in master, we don't need the BZ number

jarrpa (approved):
/lgtm

PR #1299, opened 2021-08-10
Expose Rook's SecuritySpec in StorageCluster

rexagod:
Exposed Rook's SecuritySpec in StorageCluster to support KMS config
changes for RGW.


api/v1/storagecluster_types.go
```diff
@@ -362,8 +362,9 @@ type StorageCluster struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	Spec   StorageClusterSpec   `json:"spec,omitempty"`
-	Status StorageClusterStatus `json:"status,omitempty"`
+	Spec     StorageClusterSpec      `json:"spec,omitempty"`
+	Status   StorageClusterStatus    `json:"status,omitempty"`
+	Security rookCephv1.SecuritySpec `json:"security,omitempty"`
```
umangachapagain, line 367: This is not the right place for it.
There's already a section for KMS configuration, maybe this should go there.
https://github.com/openshift/ocs-operator/blob/82f2de0694c12323a5d949e35b8d05c9a0125323/api/v1/storagecluster_types.go#L76-L80

and instead of dumping entire spec here, can we just add what's required?
We don't want to mix StorageCluster and CephCluster APIs.

umangachapagain (changes requested):

umangachapagain (commented):
Reword the commit message to explain what SecuritySpec is required and why.

api/v1/storagecluster_types.go
jarrpa: This one was my suggestion. Much as I also don't entirely like it, we already have precedent in that we use rook Placements even in places they're not normally used (say, for NooBaa). While trying to keep the two APIs distinct is usually a good idea, i am more and more of the mind that we are explicitly a Ceph-focused project and should act accordingly. I see no sense in trying to abstract or limit a part of the API like this one, all that's gonna do is force us to figure out a different override when someone wants to use the Security spec directly. Yes at some point we want to say NO and just have the user create a CephCluster CR directly, rather than reimplementing the whole thing, but as long as we are still generating *most* of the CephCluster ourselves I see an argument to be made for reusing parts of their APIs.

umangachapagain: >  we already have precedent in that we use rook Placements even in places they're not normally used (say, for NooBaa)
>

That's a whole different discussion that we need to have. That thing is probably broken (waiting for QE results).

I am not asking to abstract it. I'm just asking to enable it in our own terms, so that we have more control over what is and is not user configurable. If something is possible in Rook it doesn't mean we need to allow it in OCS and our use of rook PlacementSpec is a very good example for that.

If inheriting Rook APIs is the way we want to move forward, we might as well do it for every other resource we create. And at that point, we probably can get rid of StorageCluster.

For this specific PR I asked to not inherit API because we already have a dedicated section for Encryption and under encryption we have KMS. This KMS section was originally created for situations like this where we want to add more options to it. I'm sure you agree to that. Inside of KMS section, directly inheriting the SecuritySpec seems like it'll cause some redundant fields. So instead of inheriting the higher level SecuritySpec, we should probably inherit specific subsection as required. What if SecuritySpec gets updated to contain more than just KMS security config? We'd be automatically inheriting that inside our KMS spec.

If we want all of security spec, we need a security section. and at that point KMS configuration will be scattered all over the place.

jarrpa: I mean, i wouldn't entirely mind getting rid of StorageClusters altogether... but yeah, that's for later. 😂

I'm the meantime, @rexagod go ahead with Umanga's suggestions. I'll get some sleep and see what it looks like in the morning.

umangachapagain (changes requested):
Go changes and generated YAML changes should be in 2 separate commits.

Also, Spec changes looks good but aren't we using it anywhere in operator? We probably need to pass it to CephCluster, right?

rexagod:
Pushed an empty commit to trigger CI.

PR #1301, opened 2021-08-10
Deploy RBD mirror daemon when mirroring is enabled - WIP

GowthamShanmugam:


GowthamShanmugam:
image

controllers/storagecluster/cephrbdmirrors.go
```diff
@@ -0,0 +1,158 @@
...
+		err = r.Client.Get(context.TODO(), types.NamespacedName{Name: cephRbdMirror.Name, Namespace: cephRbdMirror.Namespace}, &existing)
+		switch {
+		case !instance.Spec.Mirroring.Enabled:
+			if err == nil {
+				err = r.deleteCephRbdMirrorInstance(cephRbdMirror, &existing)
+				if err != nil {
+					r.Log.Error(err, "Failed to delete CephRbdMirror.", "CephRbdMirror", klog.KRef(existing.Namespace, existing.Name))
+					return fmt.Errorf("Failed to delete CephRbdMirror %v: %v", existing.Name, err)
+				}
+			}
```
aruniiird, line 90: can you move this part of the code into the below case (where you check `err == nil`) under the condition `!instance.Spec.Mirroring.Enabled`.
That will make all the case-s consistent, that is all the case-s are only for errors.

controllers/storagecluster/cephrbdmirrors.go
```diff
@@ -0,0 +1,158 @@
...
+
+		err = r.Client.Get(context.TODO(), types.NamespacedName{Name: cephRbdMirror.Name, Namespace: sc.Namespace}, foundCephRbdMirror)
+		if err != nil {
+			if errors.IsNotFound(err) {
+				r.Log.Info("Uninstall: CephRbdMirror is deleted.", "CephRbdMirror", klog.KRef(cephRbdMirror.Namespace, cephRbdMirror.Name))
+				continue
+			}
+		}
+		r.Log.Error(err, "Uninstall: Waiting for CephRbdMirror to be deleted.", "CephRbdMirror", klog.KRef(cephRbdMirror.Namespace, cephRbdMirror.Name))
+		return fmt.Errorf("uninstall: Waiting for CephRbdMirror %v to be deleted", cephRbdMirror.Name)
```
aruniiird, line 154: This part is executed when there is **no** error. Question is, if there is no error, why are we making sure once again (is it a norm or practice, that we have to do a double check) the resources doesn't exist.

controllers/storagecluster/cephrbdmirrors.go
GowthamShanmugam: mm why do we mix restore logic and delete logic, i feel handle !instance.Spec.Mirroring.Enabled seperatly so it is easy to handle this case seperatly

api/v1/storagecluster_types.go
```diff
@@ -127,6 +128,13 @@ type ManageCephObjectStoreUsers struct {
 	ReconcileStrategy string `json:"reconcileStrategy,omitempty"`
 }
 
+// ManageCephRbdMirrors defines how to reconcilea CephRbdMirrors
+type ManageCephRbdMirrors struct {
+	ReconcileStrategy    string `json:"reconcileStrategy,omitempty"`
+	DisableStorageClass  bool   `json:"disableStorageClass,omitempty"`
+	DisableSnapshotClass bool   `json:"disableSnapshotClass,omitempty"`
```
umangachapagain, line 135:
```suggestion
```

not required here.

api/v1/storagecluster_types.go
```diff
@@ -127,6 +128,13 @@ type ManageCephObjectStoreUsers struct {
 	ReconcileStrategy string `json:"reconcileStrategy,omitempty"`
 }
 
+// ManageCephRbdMirrors defines how to reconcilea CephRbdMirrors
```
umangachapagain, line 131:
```suggestion
// ManageCephRbdMirrors defines how to reconcile CephRbdMirrors
```

controllers/storagecluster/cephrbdmirrors.go
```diff
@@ -0,0 +1,158 @@
...
+	rbdMirroringSpec.Resources = corev1.ResourceRequirements{
+		Limits: corev1.ResourceList{
+			corev1.ResourceCPU:    resource.MustParse("1"),
+			corev1.ResourceMemory: resource.MustParse("2Gi"),
+		},
+		Requests: corev1.ResourceList{
+			corev1.ResourceCPU:    resource.MustParse("1"),
+			corev1.ResourceMemory: resource.MustParse("2Gi"),
+		},
+	}
```
umangachapagain, line 45: This should be added here: https://github.com/openshift/ocs-operator/blob/0b1d4bd26829b15538b71616a463514f2862dc5a/controllers/defaults/resources.go#L11

controllers/storagecluster/cephrbdmirrors.go
```diff
@@ -0,0 +1,158 @@
...
+		existing := cephv1.CephRBDMirror{}
+		err = r.Client.Get(context.TODO(), types.NamespacedName{Name: cephRbdMirror.Name, Namespace: cephRbdMirror.Namespace}, &existing)
+		switch {
+		case !instance.Spec.Mirroring.Enabled:
+			if err == nil {
+				err = r.deleteCephRbdMirrorInstance(cephRbdMirror, &existing)
+				if err != nil {
+					r.Log.Error(err, "Failed to delete CephRbdMirror.", "CephRbdMirror", klog.KRef(existing.Namespace, existing.Name))
+					return fmt.Errorf("Failed to delete CephRbdMirror %v: %v", existing.Name, err)
+				}
```
umangachapagain, line 89: If CephRBDMirror is not present,this deletion will fail and error out. We need to handle such errors.

controllers/storagecluster/cephrbdmirrors.go
```diff
@@ -0,0 +1,158 @@
...
+			r.Log.Error(err, "Unable to set controller reference for CephRBDMirror.", "CephRBDMirror", klog.KRef(obj.Namespace, obj.Name))
+			return nil, err
+		}
+	}
+	return ret, nil
+}
+
+// ensureCreated ensures that cephRbdMirror resources exist in the desired
+// state.
+func (obj *ocsCephRbdMirrors) ensureCreated(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) error {
```
umangachapagain, line 68: Please refactor this function and try to keep it straightforward. It's difficult to understand what's going on.

controllers/storagecluster/cephrbdmirrors.go
umangachapagain: It is not needed.
Try to delete it, if it fails retry deletion. That's it.

controllers/storagecluster/cephrbdmirrors_test.go
```diff
@@ -0,0 +1,106 @@
...
+)
+
+func TestCephRbdMirror(t *testing.T) {
+	//cases for testing
+	var cases = []struct {
+		label                string
+		createRuntimeObjects bool
+	}{
+		{
+			label:                "case 1",
```
umangachapagain, line 23: add proper names for each cases.

controllers/storagecluster/initialization_reconciler_test.go
```diff
@@ -5,6 +5,8 @@ import (
 	"os"
 	"testing"
 
+	"github.com/imdario/mergo"
+
```
umangachapagain, line 9: remove extra newline.
```suggestion

```

umangachapagain (changes requested):

api/v1/storagecluster_types.go
GowthamShanmugam: ack

api/v1/storagecluster_types.go
GowthamShanmugam: ack

controllers/storagecluster/cephrbdmirrors.go
```diff
@@ -0,0 +1,136 @@
...
+		return nil
+	}
+
+	cephRbdMirrors, err := r.newCephRbdMirrorInstances(instance)
+	if err != nil {
+		return err
+	}
+
+	if !instance.Spec.Mirroring.Enabled {
+		return r.deleteCephRbdMirrorInstance(cephRbdMirrors, "Create")
```
GowthamShanmugam, line 88: Using this Create string to find out from which flow I am triggering the deletion

umangachapagain: It shouldn't matter. A delete is a delete no matter who triggers or how.
Just add a log message saying mirroring is disabled so deleting rbd mirror.

controllers/storagecluster/cephrbdmirrors.go
```diff
@@ -0,0 +1,137 @@
...
+
+type ocsCephRbdMirrors struct{}
+
+func (r *StorageClusterReconciler) fetchCephRbdMirrorInstance(cephRbdMirror *cephv1.CephRBDMirror) (*cephv1.CephRBDMirror, error) {
+	existing := &cephv1.CephRBDMirror{}
+	err := r.Client.Get(context.TODO(), types.NamespacedName{Name: cephRbdMirror.Name, Namespace: cephRbdMirror.Namespace}, existing)
+	return existing, err
+}
+
+func (r *StorageClusterReconciler) deleteCephRbdMirrorInstance(cephRbdMirrors []*cephv1.CephRBDMirror, flowName string) error {
```
umangachapagain, line 25: let's remove the notation of flow name. Why does it matter?

controllers/storagecluster/cephrbdmirrors.go
```diff
@@ -0,0 +1,137 @@
...
+func (r *StorageClusterReconciler) newCephRbdMirrorInstances(initData *ocsv1.StorageCluster) ([]*cephv1.CephRBDMirror, error) {
+	ret := []*cephv1.CephRBDMirror{
+		{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      generateNameForCephRbdMirror(initData),
+				Namespace: initData.Namespace,
+			},
+			Spec: cephv1.RBDMirroringSpec{
+				Count:     1,
+				Resources: defaults.DaemonResources["rbd-mirror"],
```
umangachapagain, line 61: https://github.com/openshift/ocs-operator/blob/b67106a5e4e56c0b8eb98d76f13e93a65f597019/controllers/defaults/utils.go#L10-L15
Use this instead. Resources can be updated from StorageCluster spec.

umangachapagain (changes requested):

controllers/storagecluster/cephrbdmirrors.go
GowthamShanmugam: ack

controllers/storagecluster/cephrbdmirrors.go
GowthamShanmugam: ack

controllers/storagecluster/cephrbdmirrors.go
GowthamShanmugam: I saw all other resource reconcilers are using the prefix "uninstall" in the log inside delete flow. I thought the user wants to see delete related logs separately for easy debugging. Since I am calling the delete function from creating and delete functions, I passing the flow name as parameter to log.

As per above comment(https://github.com/openshift/ocs-operator/pull/1301#discussion_r693757549), I will remove the flow name param.

controllers/storagecluster/cephrbdmirrors.go
GowthamShanmugam: refactoring is done

controllers/storagecluster/cephrbdmirrors.go
```diff
@@ -0,0 +1,136 @@
...
+			if errors.IsNotFound(err) {
+				r.Log.Info("Creating CephRbdMirror.", "CephRbdMirror", klog.KRef(cephRbdMirror.Namespace, cephRbdMirror.Name))
+				err = r.Client.Create(context.TODO(), cephRbdMirror)
+				if err != nil {
+					r.Log.Error(err, "Failed to create CephRbdMirror.", "CephRbdMirror", klog.KRef(cephRbdMirror.Namespace, cephRbdMirror.Name))
+					return err
+				}
+				continue
+			}
+			r.Log.Error(err, "Unable to create CephRbdMirror.", "CephRbdMirror", klog.KRef(cephRbdMirror.Namespace, cephRbdMirror.Name))
```
iamniting, line 104: I think this log msg needs to be corrected as your get call failed not the create call.

controllers/storagecluster/cephrbdmirrors.go
```diff
@@ -0,0 +1,136 @@
...
+		if err != nil {
+			r.Log.Error(err, "Failed to update CephRbdMirror.", "CephRbdMirror", klog.KRef(cephRbdMirror.Namespace, cephRbdMirror.Name))
+			return err
+		}
+	}
+	return nil
+}
+
+// ensureDeleted deletes the CephRbdMirrors owned by the StorageCluster
+func (obj *ocsCephRbdMirrors) ensureDeleted(r *StorageClusterReconciler, sc *ocsv1.StorageCluster) error {
```
iamniting, line 130: should not we check strategy in the deletion part as well

iamniting (changes requested):

controllers/storagecluster/cephrbdmirrors.go
umangachapagain: Yes, we should.

controllers/storagecluster/cephrbdmirrors.go
```diff
@@ -0,0 +1,136 @@
...
+	err := r.Client.Get(context.TODO(), types.NamespacedName{Name: cephRbdMirror.Name, Namespace: cephRbdMirror.Namespace}, existing)
+	return existing, err
+}
+
+func (r *StorageClusterReconciler) deleteCephRbdMirrorInstance(cephRbdMirrors []*cephv1.CephRBDMirror) error {
+	for _, cephRbdMirror := range cephRbdMirrors {
+		existing, err := r.fetchCephRbdMirrorInstance(cephRbdMirror)
+		if err != nil {
+			if !errors.IsNotFound(err) {
+				return fmt.Errorf("Unable to retrieve CephRbdMirror %v: %v", cephRbdMirror.Name, err)
```
umangachapagain, line 30:
```suggestion
				return fmt.Errorf("unable to retrieve CephRbdMirror %v: %v", cephRbdMirror.Name, err)
```

controllers/storagecluster/cephrbdmirrors.go
```diff
@@ -0,0 +1,136 @@
...
+			continue
+		}
+		if cephRbdMirror.GetDeletionTimestamp().IsZero() {
+			r.Log.Info("Deleting CephRbdMirror.", "CephRbdMirror", klog.KRef(cephRbdMirror.Namespace, cephRbdMirror.Name))
+			err := r.Client.Delete(context.TODO(), existing)
+			if err != nil {
+				r.Log.Info("Retrying deletion of CephRbdMirror.", "CephRbdMirror", klog.KRef(existing.Namespace, existing.Name))
+				err := r.Client.Delete(context.TODO(), existing)
+				if err != nil {
+					r.Log.Error(err, "Failed to delete CephRbdMirror.", "CephRbdMirror", klog.KRef(existing.Namespace, existing.Name))
```
umangachapagain, line 41: No need to retry. Just return the error and it will requeue.

controllers/storagecluster/cephrbdmirrors.go
```diff
@@ -0,0 +1,136 @@
...
+					return fmt.Errorf("Failed to delete CephRbdMirror %v: %v", existing.Name, err)
+				}
+			}
```
umangachapagain, line 44: add a log when delete succeeds.

controllers/storagecluster/cephrbdmirrors.go
```diff
@@ -0,0 +1,136 @@
...
+		}
+		if cephRbdMirror.GetDeletionTimestamp().IsZero() {
+			r.Log.Info("Deleting CephRbdMirror.", "CephRbdMirror", klog.KRef(cephRbdMirror.Namespace, cephRbdMirror.Name))
+			err := r.Client.Delete(context.TODO(), existing)
+			if err != nil {
+				r.Log.Info("Retrying deletion of CephRbdMirror.", "CephRbdMirror", klog.KRef(existing.Namespace, existing.Name))
+				err := r.Client.Delete(context.TODO(), existing)
+				if err != nil {
+					r.Log.Error(err, "Failed to delete CephRbdMirror.", "CephRbdMirror", klog.KRef(existing.Namespace, existing.Name))
+					return fmt.Errorf("Failed to delete CephRbdMirror %v: %v", existing.Name, err)
```
umangachapagain, line 42:
```suggestion
					return fmt.Errorf("failed to delete CephRbdMirror %v: %v", existing.Name, err)
```

controllers/storagecluster/cephrbdmirrors.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	cephRbdMirrors, err := r.newCephRbdMirrorInstances(instance)
+	if err != nil {
+		return err
+	}
+
+	if !instance.Spec.Mirroring.Enabled {
+		return r.deleteCephRbdMirrorInstance(cephRbdMirrors)
+	}
```
umangachapagain, line 89: We can't delete if strategy is init only

controllers/storagecluster/cephrbdmirrors.go
umangachapagain:
```suggestion
			r.Log.Error(err, "Failed to get CephRbdMirror.", "CephRbdMirror", klog.KRef(cephRbdMirror.Namespace, cephRbdMirror.Name))
```

controllers/storagecluster/initialization_reconciler_test.go
```diff
@@ -111,8 +117,11 @@ func createUpdateRuntimeObjects(t *testing.T, cp *Platform, r StorageClusterReco
 }
 
 func initStorageClusterResourceCreateUpdateTestWithPlatform(
-	t *testing.T, platform *Platform, runtimeObjs []client.Object) (*testing.T, StorageClusterReconciler, *api.StorageCluster, reconcile.Request) {
+	t *testing.T, platform *Platform, runtimeObjs []client.Object, customSpecs ...api.StorageClusterSpec) (*testing.T, StorageClusterReconciler, *api.StorageCluster, reconcile.Request) {
 	cr := createDefaultStorageCluster()
+	if len(customSpecs) > 0 {
+		_ = mergo.Merge(&cr.Spec, customSpecs[0])
+	}
```
umangachapagain, line 124: handle all values not just the first one.

umangachapagain (changes requested):

controllers/storagecluster/cephrbdmirrors.go
GowthamShanmugam: The goal of reconciling here is to create or update the resource (delete is a separate function call), So returning an error for "get" is not inappropriate.

Better I change the error message like:

r.Log.Error(err, "Failed to ensure CephRbdMirror is present.", "CephRbdMirror", klog.KRef(cephRbdMirror.Namespace, cephRbdMirror.Name))

GowthamShanmugam: means i can't decide to update or delete the CephRbdMirror resource

controllers/storagecluster/cephrbdmirrors.go
GowthamShanmugam: ack

controllers/storagecluster/cephrbdmirrors.go
GowthamShanmugam: ack

controllers/storagecluster/cephrbdmirrors.go
GowthamShanmugam: ack

controllers/storagecluster/cephrbdmirrors.go
GowthamShanmugam: ack

controllers/storagecluster/cephrbdmirrors.go
GowthamShanmugam: as we discussed removing strategy for rbd-mirror

controllers/storagecluster/initialization_reconciler_test.go
GowthamShanmugam: it handles all spec, This is for an optional parameter
func initStorageClusterResourceCreateUpdateTestWithPlatform(
	t *testing.T, platform *Platform, runtimeObjs []client.Object, customSpecs ...api.StorageClusterSpec)

when I pass custom spec it will reived as an array here "customSpecs ...api.StorageClusterSpec"
here we always pass one spec but due to this optional parameter structure it will be received as array-like [spc]
so i am using [0]

PR #1306, opened 2021-08-17
Fixed usage of global mock variables in unit-tests

priyanka19-98:
The global mock variables were being reused in may tests, due to which some tests were
failing to run individually, this commit reuses mock variables by creating a DeepCopy of global mock variables to prevent
modification of global mock variables.

There were multiple duplicate functions to create Fake Scheme for the tests,
this commit keeps only one createFakeScheme function for testing StorageCluster controller.

Above two changes are dependent on each other to make tests run individually and as a whole by `make unit-test` . 

Signed off by: Priyanka Jiandani <pjiandan@redhat.com>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -119,7 +119,7 @@ make unit-test
 
 *It is of special note that many of the mock objects in the [StorageCluster
 tests](./pkg/controller/storagecluster/storagecluster_controller_test.go) are
-reused in many other test cases. They may be useful in developing your own.*
+reused in many other test cases. Please avoid using the global mock variables directly, instead use the mock variables by creating DeepCopy() to prevent modification of global mock variables. They may be useful in developing your own.*
 
 ### Certificate of Origin
 
--- a/controllers/storagecluster/initialization_reconciler_test.go
+++ b/controllers/storagecluster/initialization_reconciler_test.go
@@ -5,20 +5,14 @@ import (
 	"os"
 	"testing"
 
-	snapapi "github.com/kubernetes-csi/external-snapshotter/client/v4/apis/volumesnapshot/v1"
 	nbv1 "github.com/noobaa/noobaa-operator/v2/pkg/apis/noobaa/v1alpha1"
 	configv1 "github.com/openshift/api/config/v1"
 	routev1 "github.com/openshift/api/route/v1"
-	openshiftv1 "github.com/openshift/api/template/v1"
 	api "github.com/openshift/ocs-operator/api/v1"
-	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	cephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
 	"github.com/stretchr/testify/assert"
 	appsv1 "k8s.io/api/apps/v1"
-	batchv1 "k8s.io/api/batch/v1"
-	corev1 "k8s.io/api/core/v1"
 	storagev1 "k8s.io/api/storage/v1"
-	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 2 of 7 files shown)


controllers/storagecluster/storagecluster_controller_test.go
```diff
@@ -81,13 +82,13 @@ var mockStorageClusterWithArbiter = &api.StorageCluster{
 
 var mockCephCluster = &rookCephv1.CephCluster{
 	ObjectMeta: metav1.ObjectMeta{
-		Name:      generateNameForCephCluster(mockStorageCluster),
+		Name:      generateNameForCephCluster(mockStorageCluster.DeepCopy()),
```
umangachapagain, line 85: Even if you don't use DeepCopy here, it's fine. Because, generate function won't modify the object.

controllers/storagecluster/storagecluster_controller_test.go
```diff
@@ -81,13 +82,13 @@ var mockStorageClusterWithArbiter = &api.StorageCluster{
...
 		Namespace: mockStorageCluster.Namespace,
 	},
 }
 
 var mockCephClusterNamespacedName = types.NamespacedName{
-	Name:      generateNameForCephCluster(mockStorageCluster),
+	Name:      generateNameForCephCluster(mockStorageCluster.DeepCopy()),
```
umangachapagain, line 91: Even if you don't use DeepCopy here, it's fine. Because, generate function won't modify the object.

controllers/storagecluster/storagecluster_controller_test.go
```diff
@@ -158,9 +159,9 @@ func getMockDeviceSets(name string, count int, replica int, portable bool) []api
 			Name:                name,
 			Count:               count,
 			Replica:             replica,
-			DataPVCTemplate:     mockDataPVCTemplate,
-			MetadataPVCTemplate: mockMetaDataPVCTemplate,
-			WalPVCTemplate:      mockWalPVCTemplate,
+			DataPVCTemplate:     *mockDataPVCTemplate.DeepCopy(),
```
umangachapagain, line 162: Why did you make it a pointer? Is it required to fix unit test?

controllers/storagecluster/initialization_reconciler_test.go
```diff
@@ -163,8 +157,8 @@ func createFakeInitializationStorageClusterReconciler(t *testing.T, obj ...runti
 func createFakeInitializationStorageClusterReconcilerWithPlatform(t *testing.T,
 	platform *Platform,
 	obj ...runtime.Object) StorageClusterReconciler {
-	scheme := createFakeInitializationScheme(t, obj...)
-	obj = append(obj, mockNodeList)
+	scheme := createFakeScheme(t)
```
umangachapagain, line 160: Earlier this scheme had a lot of api schemes added. Is it still the case or it's just the default scheme?

controllers/storagecluster/storagecluster_controller_test.go
priyanka19-98: yes, true. I wanted to make sure that we don't use mock variables directly regardless of their modification.

controllers/storagecluster/storagecluster_controller_test.go
priyanka19-98: The DeepCopy() returns a pointer, but we don't need a pointer, hence we use the pointer value by `*`

controllers/storagecluster/initialization_reconciler_test.go
priyanka19-98: yes, this scheme has a lot of api schemes.  `createFakeInitializationScheme()` also had a lot of api schemes added, which was duplicated.

umangachapagain (approved):

CONTRIBUTING.md
```diff
@@ -119,7 +119,7 @@ make unit-test
 
 *It is of special note that many of the mock objects in the [StorageCluster
 tests](./pkg/controller/storagecluster/storagecluster_controller_test.go) are
-reused in many other test cases. They may be useful in developing your own.*
+reused in many other test cases. Please avoid using the global mock variables directly, instead use the mock variables by creating DeepCopy() to prevent modification of global mock variables. They may be useful in developing your own.*
```
jarrpa, line 122: This line should be properly wrapped as all the other text in the markdown. But that can be a separate PR, I'm not about to hold the code changes for this at this stage.

PR #1307, opened 2021-08-17
csv: add annotation to hide from UI

jarrpa:


umangachapagain (commented):
`add annotation to hide from UI` sounds a little unclear to me.
Should we just say something like `marking operator as non-standalone` ?

PR #1308, opened 2021-08-19
WIP : Fix for Dual Stack

pkesavap:
WIP: Fix issue of osd pods being up, but  osd shows cluster is unhealthy because of wrong ip assignment in Dual stack


rohantmp:
Reviewing

controllers/storagecluster/reconcile.go
```diff
@@ -526,6 +535,21 @@ func (obj *ocsCephConfig) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.S
 		return nil
 	}
 
+	if sc.Spec.Network.Provider == "" && sc.Spec.Network.DualStack {
+
+		networkConfig := &configv1.Network{ObjectMeta: metav1.ObjectMeta{Name: "cluster"}}
+		if err != nil {
+			return fmt.Errorf("could not get network config details to determine cloud platform: %v", err)
+		}
+		cidr1 := networkConfig.Spec.ClusterNetwork[0].CIDR
```
rohantmp, line 544: let's loop through instead of assuming there are two

controllers/storagecluster/reconcile.go
```diff
@@ -526,6 +535,21 @@ func (obj *ocsCephConfig) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.S
 		return nil
 	}
 
+	if sc.Spec.Network.Provider == "" && sc.Spec.Network.DualStack {
```
rohantmp, line 538: Can we add a comment here saying something like:
```
configure public network if the cluster is dualstack, but not multus
```

controllers/storagecluster/reconcile.go
```diff
@@ -647,3 +672,24 @@ func validateArbiterSpec(sc *ocsv1.StorageCluster, reqLogger logr.Logger) error
 	}
 	return nil
 }
+
+func setNetworkConfiguration(defaultRookConfigData string, val string) (string, error) {
```
rohantmp, line 676: perhaps we can have a more general method like `updateRookConfig(config string, section string, value string )(string, error)`?

rohantmp (changes requested):
requesting some changes

rohantmp:
The context: OSDs can't figure out the right IPs to listen on in a native dualstack environment. Populating the public network ceph config fixes this.

controllers/storagecluster/reconcile.go
pkesavap: Done

controllers/storagecluster/reconcile.go
pkesavap: @rohantmp what does the config and section string indicate?

controllers/storagecluster/reconcile.go
```diff
@@ -526,6 +535,26 @@ func (obj *ocsCephConfig) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.S
...
+	if sc.Spec.Network.Provider == "" && sc.Spec.Network.DualStack {
+
+		networkConfig := &configv1.Network{ObjectMeta: metav1.ObjectMeta{Name: "cluster"}}
+		if err != nil {
+			return fmt.Errorf("could not get network config details to determine cloud platform: %v", err)
+		}
+
+		var cidrName string
+		for _, cidr := range networkConfig.Spec.ClusterNetwork {
+			cidrName += cidr.CIDR
```
rohantmp, line 548: this should be comma separated right?

controllers/storagecluster/reconcile.go
```diff
@@ -46,9 +49,15 @@ mon_osd_backfillfull_ratio = .8
 mon_osd_nearfull_ratio = .75
 mon_max_pg_per_osd = 600
 mon_pg_warn_max_object_skew = 0
+public_network = 10.128.0.0/14,fd01::/48
```
rohantmp, line 52: this key should not be there by default

rohantmp (changes requested):

controllers/storagecluster/reconcile.go
```diff
@@ -647,3 +677,24 @@ func validateArbiterSpec(sc *ocsv1.StorageCluster, reqLogger logr.Logger) error
 	}
 	return nil
 }
+
+func updateRookConfig(defaultRookConfigData string, val string) (string, error) {
```
rohantmp, line 681: I meant we could change the method to accept a section and key to change too so that it can change keys other than `public_network` and in sections other than `global`

rohantmp (changes requested):

controllers/storagecluster/reconcile.go
pkesavap: Done

controllers/storagecluster/reconcile.go
pkesavap: done

controllers/storagecluster/reconcile.go
pkesavap: done

controllers/storagecluster/reconcile.go
```diff
@@ -647,3 +679,24 @@ func validateArbiterSpec(sc *ocsv1.StorageCluster, reqLogger logr.Logger) error
...
+	}
+
+	if val == "" {
+		return "", nil
+	}
+	cfg, err := ini.Load([]byte(defaultRookConfigData))
+	if err != nil {
+		return "", fmt.Errorf("failed ini.Load(): %+v", err)
+	}
+	cfg.Section("global").Key(section).SetValue(val)
```
rohantmp, line 695: `"global"` is the section, and the arg for `Key(arg)` should be another var.

rohantmp (changes requested):
Will ask whether network config will change

controllers/storagecluster/reconcile.go
```diff
@@ -526,6 +533,30 @@ func (obj *ocsCephConfig) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.S
...
+		if err != nil {
+			return fmt.Errorf("could not get network config details to determine cloud platform Prajith anne: %v", err)
+		}
+
+		clusterNetworkLength := len(networkConfig.Spec.ClusterNetwork)
+		var cidrName string
+		for index, cidr := range networkConfig.Spec.ClusterNetwork {
+			cidrName += cidr.CIDR
+			if index != clusterNetworkLength-1 {
+				cidrName += ","
```
rohantmp, line 549: Can we use `strings.Join()`?

rohantmp:
Also @pkesavap, please confirm when you've tested the change on the cluster :)

controllers/storagecluster/reconcile.go
pkesavap: Something like this?
```
	cidrNameArray := []string{}
		for _, cidr := range networkConfig.Spec.ClusterNetwork {
			cidrNameArray = append(cidrNameArray, cidr.CIDR)
		}
		cidrName := strings.Join(cidrNameArray, ",")
		```

Comment (pkesavap) [reply controllers/storagecluster/reconcile.go:549]: @rohantmp ^^

Comment (rohantmp) [reply controllers/storagecluster/reconcile.go:549]: looks good!

File: controllers/storagecluster/reconcile.go [hunk 9]
```diff
@@ -526,6 +531,22 @@ func (obj *ocsCephConfig) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.S
 		return nil
 	}
 
+	// configure public network if the cluster is dualstack, but not multus
+	if sc.Spec.Network != nil && sc.Spec.Network.Provider == "" && sc.Spec.Network.DualStack {
+		networkConfig := &configv1.Network{ObjectMeta: metav1.ObjectMeta{Name: "cluster"}}
```
pkesavap, line 536: @rohantmp shouldnt we do a get call to get the network object here?

rohantmp: Yup! Don't know how I missed that!

controllers/storagecluster/reconcile.go
```diff
@@ -526,6 +531,26 @@ func (obj *ocsCephConfig) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.S
...
+	if sc.Spec.Network != nil && sc.Spec.Network.Provider == "" && sc.Spec.Network.DualStack {
+		log.Info("Prajith DualStack is enabled, but no network provider is specified. Using the default network provider.")
+		//networkConfig := &configv1.Network{}
+		networkConfig := &configv1.Network{}
+		err := r.Client.Get(context.TODO(), types.NamespacedName{Name: "cluster", Namespace: ""}, networkConfig)
+		if err != nil {
+			return fmt.Errorf("could not get network config details to determine cloud platform: %v", err)
+		}
+		cidrNameArray := []string{}
+		for _, cidr := range networkConfig.Spec.ClusterNetwork {
```
nbalacha, line 544: The documentation states that the status (not the spec) should not be read. 

https://docs.openshift.com/container-platform/4.8/rest_api/config_apis/network-config-openshift-io-v1.html

controllers/storagecluster/reconcile.go
```diff
@@ -526,6 +531,26 @@ func (obj *ocsCephConfig) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.S
...
 	}
 
+	// configure public network if the cluster is dualstack, but not multus
+	if sc.Spec.Network != nil && sc.Spec.Network.Provider == "" && sc.Spec.Network.DualStack {
+		log.Info("Prajith DualStack is enabled, but no network provider is specified. Using the default network provider.")
+		//networkConfig := &configv1.Network{}
+		networkConfig := &configv1.Network{}
+		err := r.Client.Get(context.TODO(), types.NamespacedName{Name: "cluster", Namespace: ""}, networkConfig)
+		if err != nil {
+			return fmt.Errorf("could not get network config details to determine cloud platform: %v", err)
```
nbalacha, line 541: How does this help with determining the cloud platform?

nbalacha (changes requested):

controllers/storagecluster/reconcile.go
```diff
@@ -526,6 +531,26 @@ func (obj *ocsCephConfig) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.S
 		return nil
 	}
 
+	// configure public network if the cluster is dualstack, but not multus
+	if sc.Spec.Network != nil && sc.Spec.Network.Provider == "" && sc.Spec.Network.DualStack {
+		log.Info("Prajith DualStack is enabled, but no network provider is specified. Using the default network provider.")
```
nbalacha, line 536: Please remove "Prajith" from the log message.

PR #1320, opened 2021-08-30
Support standalone NooBaa

jarrpa:
This PR introduces two related features: CephCluster reconcile strategy and MCG standalone strategy. Together these allow a StorageCluster to specify finer or harsher control over the operator's resource reconciliation.


jarrpa:
e2e tests passed, but must-gather failed for unrelated reasons.

controllers/storagecluster/reconcile.go
```diff
@@ -53,7 +53,7 @@ const (
 	ReconcileStrategyIgnore ReconcileStrategy = "ignore"
 	// ReconcileStrategyManage means always reconcile
 	ReconcileStrategyManage ReconcileStrategy = "manage"
-	// ReconcileStrategyStandalone also means never reconcile (NooBaa)
+	// ReconcileStrategyStandalone means to renconcile the component exclusively (NooBaa)
```
nbalacha, line 56: Nit: typo reconcile

obnoxxx: The typo needs to be fixed. but could be fixed in a follow-up PR too.

I don't fully understand where the statement of standalone-noobaa vs external mode comes from. If I read the code correctly, in the noobaa-standalone mode, the noobaa reconciler just does not need to worry about the cephcluster at all any more (external or internal).

controllers/storagecluster/reconcile.go
```diff
@@ -340,22 +342,29 @@ func (r *StorageClusterReconciler) reconcilePhases(
...
-			&ocsCephObjectStores{},
-			&ocsCephObjectStoreUsers{},
-			&ocsCephRGWRoutes{},
-			&ocsStorageClass{},
-			&ocsNoobaaSystem{},
-			&ocsStorageQuota{},
-			&ocsSnapshotClass{},
-			&ocsJobTemplates{},
-			&ocsCephRbdMirrors{},
+		if !r.IsNoobaaStandalone {
```
nbalacha, line 345: It looks like this is being checked in the reconcile operation only for non-external clusters.  Is this the expected behaviour?

obnoxxx: Right... Currently it is like this:

* internal-mode + noobaa-standalone ==> noobaa-only
* external-mode + noobaa-standalone ==> external mode (and noobaa-standalone is ignored)

One could argue that noobaa-standalone should alway win.

Or it could be made independent of the cephcluster reconciliation, i.e. noobaa-standalone would mean that noobaa is just not using ceph for backing store. And the ceph cluster could be managed independently, internally or externally.

controllers/storagecluster/reconcile.go
jarrpa: Good catch

controllers/storagecluster/reconcile.go
jarrpa: Specifying both external and noobaa-only is an invalid spec to begin with. The proper thing to do would be to invalidate the spec and refuse to reconcile until it is resolved. After that the issue of which one takes precedent is largely a moot point.

As far as non-ceph noobaa goes, we can already do that by explicitly specifying the noobaa db storageclassname. That's the only thing that ocs-operator generates for noobaa from ceph. Here, the standalone reconcile strategy is shorthand for don't create any Ceph resources. Without it, the user would have to specify `reconcileStrategy: ignore` for every single other managed resource. And if we do that, then we don't need `standalone` since the absence of the RBD storageclass would cause noobaa to use some other default storageclass. Also we would continue to have the (irritating) situation where `standalone == ignore`.

obnoxxx: I think we can merge this is as is for the feature completion and put the rest it as fixes in subsequent PRs

obnoxxx (approved):
as discussed with @jarrpa and @nbalacha , approving as is and deferring fixes to subsequent PRs

/lgtm

controllers/storagecluster/reconcile.go
jarrpa: Just to record it: I agreed to let noobaa-standalone take precedence. But in exploring that, I found additional complications. So a subsequent PR should include at minimum:

1. Let NooBaa-standalone take precedence over external CephCluster
2. Check for StorageClass in NooBaa creation

Additionally for external mode, we should maybe figure out a way to pre-emptively detect if RBD StorageClass exists in external Secret. If it doesn't, StorageClass reconciliation will still succeed, so NooBaa will try to use a StorageClass that doesn't exist. Whether we fail or fall back to defaults is still up for debate.

PR #1324, opened 2021-08-31
Fix  server listen flake forever

jarrpa:
A test flake where sometimes a test would fail due to a mock server not being present has plagued us for far too long. This PR refactors the `startServerAt()` function to properly start and stop a dummy listening server. It also updates the relevant test functions to use `t.Run()` to make sure we are not reusing mock servers between test cases.

### DIFF
--- a/controllers/storagecluster/storagecluster_controller_test.go
+++ b/controllers/storagecluster/storagecluster_controller_test.go
@@ -3,6 +3,8 @@ package storagecluster
 import (
 	"context"
 	"fmt"
+	"net"
+	"regexp"
 	"testing"
 
 	"github.com/blang/semver"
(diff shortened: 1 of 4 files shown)


controllers/storagecluster/storagecluster_controller_test.go
```diff
@@ -1139,3 +1141,44 @@ func getSemVer(version string, majorDiff uint64, islower bool) string {
...
+
+	return ls, nil
+
+}
+
+func startServerAt(t *testing.T, endpoint string) {
+	ls, err := NewListenServer(endpoint)
+	if err != nil {
+		t.Logf("Failed to start ListenServer: %v", err)
+		return
```
BlaineEXE, line 1180: Would it be better to panic the test (or `t.Fatal`) upon this failure since it will cause some tests to fail anyway?

BlaineEXE (approved):
Looks good from what I can see. Relying on `testing.T.Cleanup()` makes it easier to use it seems. Approving with just one nit comment/question.

controllers/storagecluster/storagecluster_controller_test.go
jarrpa: I thought about this one a fair amount. The main issue is that there is *one* test case that purposefully supplies and invalid port. It's not the best, but to change that I would've had to refactor even more auxiliary functions that I did not feel like diving into. :joy: Also, in general, if for some reason the ListenServer fails to start, something else in the test is probably going to break anyway. This was my compromise to at least always highlight it even if it's expected.

nbalacha (approved):
/lgtm

PR #1335, opened 2021-09-13
storagecluster: move topology map reconciliation into its own resourceManager

jarrpa:

### DIFF
--- a/controllers/storagecluster/reconcile.go
+++ b/controllers/storagecluster/reconcile.go
@@ -327,14 +327,6 @@ func (r *StorageClusterReconciler) reconcilePhases(
 		return reconcile.Result{}, returnErr
 	}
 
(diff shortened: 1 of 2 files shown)


nbalacha (commented):
While the code looks fine, this does seem to differ from the other entries in the resource manager which are managing k8s CR objects unlike the topologyMap.  Is that fine?
Can you add the reason for the change in the commit message?

jarrpa:
> While the code looks fine, this does seem to differ from the other entries in the resource manager which are managing k8s CR objects unlike the topologyMap. Is that fine?
> Can you add the reason for the change in the commit message?

Fair enough, done.

obnoxxx (approved):
this looks good to moe

/lgtm

@nbalacha  - I think your comment has been addressed.

PR #1336, opened 2021-09-14
Automate prometheususer creation for ob metrics collector

thotz:
As a prerequisite for object bucket metrics collector, it is expected to have a user named `prometheus-user` in the object store with certain permissions . Otherwise, it won't able to fetch the stat info from backend. This PR automates that creation

### DIFF
--- a/controllers/storagecluster/cephobjectstoreusers.go
+++ b/controllers/storagecluster/cephobjectstoreusers.go
@@ -15,6 +15,8 @@ import (
 
 type ocsCephObjectStoreUsers struct{}
 
+const prometheusUserName = "prometheus-user"
+
 // newCephObjectStoreUserInstances returns the cephObjectStoreUser instances that should be created
 // on first run.
 func (r *StorageClusterReconciler) newCephObjectStoreUserInstances(initData *ocsv1.StorageCluster) ([]*cephv1.CephObjectStoreUser, error) {
@@ -27,6 +29,20 @@ func (r *StorageClusterReconciler) newCephObjectStoreUserInstances(initData *ocs
 			Spec: cephv1.ObjectStoreUserSpec{
 				DisplayName: initData.Name,
 				Store:       generateNameForCephObjectStore(initData),
+			},
+		},
+		{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      prometheusUserName,
+				Namespace: initData.Namespace,
+			},
+			Spec: cephv1.ObjectStoreUserSpec{
+				DisplayName: prometheusUserName,
+				Store:       generateNameForCephObjectStore(initData),
+				// This user needs to read quota info from other users where actual quota is on set at the backend
+				Capabilities: &cephv1.ObjectUserCapSpec{
+					User: "read",
+				},
 			},
 		},
 	}
--- a/controllers/storagecluster/uninstall_reconciler_test.go
+++ b/controllers/storagecluster/uninstall_reconciler_test.go
@@ -570,18 +570,30 @@ func assertTestDeleteCephBlockPools(
 	}
 }
 
-func getFakeCephObjectStoreUser() *cephv1.CephObjectStoreUser {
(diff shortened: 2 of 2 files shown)


thotz:
@umangachapagain @jarrpa @anmolsachan  PTAL

nbalacha:
This PR includes a commit to update the rook version to 1.7.3. Is that required in order to create the prometheus user?
If yes, please do that in a separate PR as that is a big change.

thotz:
> This PR includes a commit to update the rook version to 1.7.3. Is that required in order to create the prometheus user?
> If yes, please do that in a separate PR as that is a big change.

Yes, I will send separate PR for rook update

controllers/storagecluster/cephobjectstoreusers.go
```diff
@@ -29,6 +31,20 @@ func (r *StorageClusterReconciler) newCephObjectStoreUserInstances(initData *ocs
 				Store:       generateNameForCephObjectStore(initData),
 			},
 		},
+		{
```
umangachapagain, line 34: is it safe to create such a user? any user who can access this resource could use it for previlage escalation.

umangachapagain, line 34: is it safe to create such a user? any user who can access this resource could use it for previlage escalation.

thotz: The same security concern applies for the `noobaa-ceph-objectstore-user` and `ocs-storagecluster-cephobjectstoreuser`, even if we create this user manually, this concern issue will be present

umangachapagain (commented):
lgtm

/cc @leseb  Do you think there could be any issues with creating such a user?

controllers/storagecluster/cephobjectstoreusers.go
```diff
@@ -29,6 +31,19 @@ func (r *StorageClusterReconciler) newCephObjectStoreUserInstances(initData *ocs
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      prometheusUserName,
+				Namespace: initData.Namespace,
+			},
+			Spec: cephv1.ObjectStoreUserSpec{
+				DisplayName: prometheusUserName,
+				Store:       generateNameForCephObjectStore(initData),
+				Capabilities: &cephv1.ObjectUserCapSpec{
+					User: "*",
```
leseb, line 43: Why do we need a wildcard? Why `User`? It will be good to understand a bit more which permissions are required.

thotz: Read permission is fine, the the quota is set on the internal user of OBC, so we need to fetch the user stats

controllers/storagecluster/cephobjectstoreusers.go
```diff
@@ -29,6 +31,20 @@ func (r *StorageClusterReconciler) newCephObjectStoreUserInstances(initData *ocs
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      prometheusUserName,
+				Namespace: initData.Namespace,
+			},
+			Spec: cephv1.ObjectStoreUserSpec{
+				DisplayName: prometheusUserName,
+				Store:       generateNameForCephObjectStore(initData),
+				// This user needs to read quota info from other users where actual quota is on set at the backend
+				Capabilities: &cephv1.ObjectUserCapSpec{
+					User: "read",
```
leseb, line 44: Great thanks I think this should address the security concern better now.

controllers/storagecluster/cephobjectstoreusers.go
```diff
@@ -15,6 +15,8 @@ import (
 
 type ocsCephObjectStoreUsers struct{}
 
+var prometheusUserName = "prometheus-user"
```
leseb, line 18:
```suggestion
const prometheusUserName = "prometheus-user"
```

leseb (changes requested):
One nit

umangachapagain (approved):

PR #1257, opened 2021-07-05
Bug 1979263: [release-4.8] revert: creation of thick provisioned storage class in 4.8

agarwal-mudit:
Thick provisioning feature is now dev preview, hence removing
the sc creation.

Signed-off-by: Mudit Agarwal <muagarwa@redhat.com>

PR #1339, opened 2021-09-16
Bug 2000865: [release-4.9] revert: creation of thick provisioned storage class in 4.9 

agarwal-mudit:
Backport of #1257 after resolving the conflicts

/assign @jarrpa 
/assign @obnoxxx

### DIFF
--- a/controllers/storagecluster/external_resources.go
+++ b/controllers/storagecluster/external_resources.go
@@ -307,7 +307,7 @@ func (r *StorageClusterReconciler) createExternalStorageClusterResources(instanc
 				scc = newCephFilesystemStorageClassConfiguration(instance)
 				enableRookCSICephFS = true
 			} else if d.Name == cephRbdStorageClassName {
-				scc = newCephBlockPoolStorageClassConfiguration(instance, false)
+				scc = newCephBlockPoolStorageClassConfiguration(instance)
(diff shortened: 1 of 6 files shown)


nbalacha:
The PR title says revert creation of this provisioned storage class in *4.8*. Please update the title.

Why is this being reverted for 4.9?

agarwal-mudit:
> The PR title says revert creation of this provisioned storage class in _4.8_. Please update the title.
> 
> Why is this being reverted for 4.9?

Because this was first reverted in 4.8 and now in 4.9, reason is same for both the releases:

We wanted to have this feature since OCS4.8 but there are some ceph related issues with this feature  which are not yet fixed.
We hope to get those fixes by the time we release ODF 4.10 but for now we need to revert the changes which were done to enable the feature.

There is one related PR to hide it from UI: #1259 This was also done during 4.8 and need to be done again for 4.9

nbalacha:
@jarrpa , any idea why the red-hat-storage-ocs-ci-e2e-aws  test is failing?

jarrpa:
It always fails. :joy: We haven't maintained it in a very long time. We've marked it as optional so it doesn't block merges, but ideally it shouldn't be triggered at all. Haven't explored how to do that yet.

PR #1345, opened 2021-09-23
ocs: replace appregistry with index image

nbalacha:
Appregistry has been deprecated for a while now and
the recent MCG changes mean that the operator cannot be
deployed with the current ocs-registry. This is a good time
to move to using the index image format.


jarrpa (approved):
/lgtm

malayparida2000:
@nbalacha I think if this PR is stale now, We can close it

PR #1346, opened 2021-09-23
Fixing a minor bug regarding RGW TLS support enabling

aruniiird:
When the TLS secret is retrieved successfully, that is retrieving 'err' is nil,
then only enable the 'SecurePort' and 'SSLCertRef' in CephObjectStore Gateway.


controllers/storagecluster/external_resources.go
```diff
@@ -195,7 +195,8 @@ func (r *StorageClusterReconciler) newExternalCephObjectStoreInstances(
 	}
 	var tlsEnabled bool = false
 	_, err := r.retrieveSecret(cephRgwTLSSecretKey, initData)
-	if err != nil {
+	// if we could retrieve a TLS secret, then enable TLS
+	if err == nil {
```
leseb, line 199: Probably better to use `IsNotFound(err)` instead and handle the error properly.

PR #1355, opened 2021-10-03
Bug 2010041: [release-4.9] Inject default label to noobaa_system_reconciler

liranmauda:
As part of this backport we needed to backport also the commit that replace openshift with red-hat-storage  to avoid conflicts and the commits that update NooBaa operator from v2 to v5


jarrpa (approved):
/lgtm

PR #1360, opened 2021-10-07
BUG 1938647: use SecurityContextConstraints API from Ceph-CSI

nixpanic:
Remove the duplication of the SCC for the CephCSI component, use the
new Ceph-CSI API to get the SCC instead.

Signed-off-by: Niels de Vos <ndevos@redhat.com>

### DIFF
--- a/controllers/ocsinitialization/sccs.go
+++ b/controllers/ocsinitialization/sccs.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 
+	cephcsi "github.com/ceph/ceph-csi/api/deploy/ocp"
 	secv1 "github.com/openshift/api/security/v1"
 	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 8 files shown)


controllers/ocsinitialization/sccs.go
```diff
@@ -103,44 +104,13 @@ func newRookCephSCC(namespace string) *secv1.SecurityContextConstraints {
...
-	}
-	scc.Users = []string{
-		fmt.Sprintf("system:serviceaccount:%s:rook-csi-rbd-plugin-sa", namespace),
-		fmt.Sprintf("system:serviceaccount:%s:rook-csi-rbd-provisioner-sa", namespace),
-		fmt.Sprintf("system:serviceaccount:%s:rook-csi-rbd-attacher-sa", namespace),
-		fmt.Sprintf("system:serviceaccount:%s:rook-csi-cephfs-plugin-sa", namespace),
-		fmt.Sprintf("system:serviceaccount:%s:rook-csi-cephfs-provisioner-sa", namespace),
+	rookValues := cephcsi.SecurityContextConstraintsValues{
+		Namespace: "rook-ceph",
+		Deployer:  "rook",
```
Madhu-1, line 109: this should be `rook-` because the name is `rook-ceph-csi` and all the SCC Users are starts with `rook-`

controllers/ocsinitialization/sccs.go
```diff
@@ -103,44 +104,13 @@ func newRookCephSCC(namespace string) *secv1.SecurityContextConstraints {
...
-		secv1.FSTypeAll,
-	}
-	scc.Users = []string{
-		fmt.Sprintf("system:serviceaccount:%s:rook-csi-rbd-plugin-sa", namespace),
-		fmt.Sprintf("system:serviceaccount:%s:rook-csi-rbd-provisioner-sa", namespace),
-		fmt.Sprintf("system:serviceaccount:%s:rook-csi-rbd-attacher-sa", namespace),
-		fmt.Sprintf("system:serviceaccount:%s:rook-csi-cephfs-plugin-sa", namespace),
-		fmt.Sprintf("system:serviceaccount:%s:rook-csi-cephfs-provisioner-sa", namespace),
+	rookValues := cephcsi.SecurityContextConstraintsValues{
+		Namespace: "rook-ceph",
```
Madhu-1, line 108: need to use `namespace` input parameter as the namespace will be openshift-storage

nixpanic: Oh, right, of course!

controllers/ocsinitialization/sccs.go
nixpanic: This needs to be without the `-`, it will be added automatically when `Deployer` is set:

https://github.com/ceph/ceph-csi/blob/574852e27c84b8ac78ad7412b0f8ca300eb2bb4b/api/deploy/ocp/scc.go#L93-L95

umangachapagain (approved):
We generally split code changes and vendor changes into separate commits.
Rest looks good.

nbalacha (approved):
/lgtm

PR #1365, opened 2021-10-11
build: support AWS/Graviton (aarch64)

synarete:
Build OCS on AWS/Graviton processor (ARM, aarch64). Need to update usage
of external tools:

1) Download proper operator-sdk for local arch, make sure that the
   downloaded image is indeed exacutable on local host.
2) For aarch64 need to build opm on local machine as no image is
   provided.

Align definition of OS & ARCH with installation instruction at:
  https://sdk.operatorframework.io/docs/installation/

### DIFF
--- a/hack/ensure-opm.sh
+++ b/hack/ensure-opm.sh
@@ -5,18 +5,42 @@ set -e
 source hack/common.sh
 
 OPM_VERSION="v1.12.3"
-OPM_PLATFORM="linux-amd64-opm"
-if [ "$OS_TYPE" == "Darwin" ]; then
-	OPM_PLATFORM=darwin-amd64-opm
-fi
+case "$(uname -m)" in
+	x86_64)
+		OPM_ARCH="amd64"
+		;;
+	aarch64)
+		OPM_ARCH="arm64"
+		;;
+	*)
+		OPM_ARCH="$(uname -m)"
+		;;
+esac
+OPM_OS=$(echo -n "${OS_TYPE}" | awk '{print tolower($0)}')
+OPM_PLATFORM="${OPM_OS}-${OPM_ARCH}-opm"
 OPM_URL="https://github.com/operator-framework/operator-registry/releases/download/${OPM_VERSION}/${OPM_PLATFORM}"
 OPM_BIN="opm-${OPM_VERSION}"
 OPM="${OUTDIR_TOOLS}/${OPM_BIN}"
-if [ ! -x "${OPM}" ]; then
-	echo "Downloading opm ${OPM_VERSION} for ${OS_TYPE}"
-	mkdir -p ${OUTDIR_TOOLS}
-	curl -JL ${OPM_URL} -o ${OPM}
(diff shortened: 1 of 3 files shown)


hack/operator-sdk-common.sh
```diff
@@ -3,11 +3,19 @@
 # shellcheck disable=SC2034
 # don't fail on unused variables - this is for sourcing
 
+
+# Align with installation instructions from:
+# https://sdk.operatorframework.io/docs/installation/
+OPERATOR_SDK_ARCH=$(uname -m)
+OPERATOR_SDK_OS=$(uname | awk '{print tolower($0)}')
```
jarrpa, line 10: If we already have `OS_TYPE` do we need to do this? Looks like you could just do this in `common.sh` (including the hack below), you would only need to additionally update `shellcheck-test.sh`. Also, at this point we might want to define "constants" of `OS_TYPE_LINUX` and `OS_TYPE_DARWIN` (in all lowercase) for our expected values and compare against those everywhere. This sounds like it'd be a separate commit before introducing the operator-sdk hack on top of it.

hack/ensure-opm.sh
```diff
@@ -5,18 +5,32 @@ set -e
...
-else
+OPERATOR_REGISTRY_TMPDIR="${OUTDIR_TOOLS}/operator-registry"
+
+if [ -x "${OPM}" ]; then
+	# Case 1: already has opm binray
 	echo "Using opm cached at ${OPM}"
-fi
\ No newline at end of file
+elif [ "${OPM_ARCH}" != "amd64" ]; then
+	# Case 2: image does not exist, need to build from source
```
jarrpa, line 20: ....wow, I can't believe this. :joy: But I see it is indeed necessary for now...

hack/ensure-opm.sh
```diff
@@ -5,18 +5,32 @@ set -e
 source hack/common.sh
 
 OPM_VERSION="v1.12.3"
-OPM_PLATFORM="linux-amd64-opm"
-if [ "$OS_TYPE" == "Darwin" ]; then
-	OPM_PLATFORM=darwin-amd64-opm
-fi
+OPM_ARCH=$(case $(uname -m) in x86_64) echo -n amd64 ;; aarch64) echo -n arm64 ;; *) echo -n "$(uname -m)" ;; esac)
```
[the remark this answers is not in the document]
synarete: @jarrpa Its a copy-paste from https://sdk.operatorframework.io/docs/installation/
However, I agree that multi-line block is more readable. Will change.

hack/operator-sdk-common.sh
```diff
@@ -3,11 +3,19 @@
...
+# This hach was changed in newer operator-sdk versions
```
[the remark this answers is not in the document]
synarete: A hack indeed :)

hack/operator-sdk-common.sh
jarrpa: Fair enough!

PR #1368, opened 2021-10-12
add permissions for ocs-metrics-exporter

thotz:
The ocs-metrics-exporter service misses permissions to access the
secrets and objectbuckets, hence it was not able to metrics.

Signed-off-by: Jiffin Tony Thottan <thottanjiffin@gmail.com>

### DIFF
--- a/deploy/bundle/manifests/exporter-role.yaml
+++ b/deploy/bundle/manifests/exporter-role.yaml
@@ -21,3 +21,17 @@ rules:
     - get
     - list
     - watch
+- apiGroups:
+  - objectbucket.io
+  resources:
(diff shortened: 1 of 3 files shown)


thotz:
> What secrets does this need access to?

It needs to access the RGW endpoint for getting metrics using a `cephobjectstoreuser`. The connection details will be present in the secret for that user

nbalacha:
> > What secrets does this need access to?
> 
> It needs to access the RGW endpoint for getting metrics using a `cephobjectstoreuser`. The connection details will be present in the secret for that user

Is the secret name fixed? Can we provide access only to that secret?

PR #1375, opened 2021-10-18
storagecluster: require valid overprovision capacity value

synarete:
Do not allow zero or negative value for overprovision capacity entries.
In case of non-valid entry, log out the name of the bad entry.

Refs BZ2014805


controllers/storagecluster/reconcile.go
```diff
@@ -627,10 +627,19 @@ func validateArbiterSpec(sc *ocsv1.StorageCluster, reqLogger logr.Logger) error
...
+			return fmt.Errorf("can not define both capacity and percentage for overprovision %s", opc.QuotaName)
+		}
+		if opc.Capacity != nil {
+			val, ok := opc.Capacity.AsInt64()
+			if !ok {
+				return fmt.Errorf("non-valid capacity %q for overprovision %s", opc.Capacity, opc.QuotaName)
+			}
+			if val <= 0 {
+				return fmt.Errorf("illegal capacity value for overprovision %s", opc.QuotaName)
+			}
```
nbalacha, line 642: Should this check for a negative percentage?

controllers/storagecluster/reconcile.go
```diff
@@ -627,10 +627,19 @@ func validateArbiterSpec(sc *ocsv1.StorageCluster, reqLogger logr.Logger) error
...
-			return fmt.Errorf("can not define both capacity and percentage for overprovision")
+			return fmt.Errorf("can not define both capacity and percentage for overprovision %s", opc.QuotaName)
+		}
+		if opc.Capacity != nil {
+			val, ok := opc.Capacity.AsInt64()
+			if !ok {
+				return fmt.Errorf("non-valid capacity %q for overprovision %s", opc.Capacity, opc.QuotaName)
+			}
+			if val <= 0 {
+				return fmt.Errorf("illegal capacity value for overprovision %s", opc.QuotaName)
```
nbalacha, line 641: Perhaps "capacity cannot be a negative value" would be clearer.

PR #1377, opened 2021-10-18
csv: don't require NooBaa CRD

jarrpa:
Unfortunately, since the NooBaa CSV requires the OB/OBC CRDs from us,
this creates a cyclic dependency... and in particular, OLM is terrible
about resolving this and thus leads to race conditions when installing
ODF 4.9. For now, we're removing the NooBaa CRD from the OCS CSV's
required CRDs. Since we don't support ever running ocs-operator outside
the context of ODF, this won't be an issue in production.


iamniting:
@jarrpa We are missing to remove the CRD yaml files from the bundle.
https://github.com/red-hat-storage/ocs-operator/blob/main/deploy/bundle/manifests/objectbucket.crd.yaml
https://github.com/red-hat-storage/ocs-operator/blob/main/deploy/bundle/manifests/objectbucketclaim.crd.yaml

jarrpa:
At this point, this PR is potentially superseded by #1771.

malayparida2000:
@jarrpa  I think if this PR is stale now, We can close it

iamniting:
yes it can be closed

PR #1384, opened 2021-10-19
storagecluster: disable percentage option in overprovision control

synarete:
The 'percentage' option turned out to be more confusing then useful.
Prevent users from using this option for now, but leave it in API for
future enhancement.

Refs BZ2014806


umangachapagain (commented):
If percentage option was confusing, this is even more confusing.
This is bad user experience.

Can we enhance our docs? and make updates to make it less confusing?

PR #1386, opened 2021-10-20
add managedBy label only for mcg-standalone

umangachapagain:
Adding `managedBy` label only when mcg is standalone. This is to prevent collision when mcg is deployed along with Ceph.


jarrpa (approved):
/lgtm

PR #1391, opened 2021-10-26
Enable taint-nodes from the operator CSV

SanjalKatiyar:
Added UI for dedicated nodes with taints in 4.10 (https://github.com/openshift/console/pull/10323)
This PR is to enable that UI.

Signed-off-by: sanjalkatiyar <sanjaldhir@gmail.com>


umangachapagain (commented):
/lgtm

PR #1404, opened 2021-11-22
Fixing for OdfPoolMirroringImageHealth false alerts

GowthamShanmugam:
Modified expression to raise alert only when the corresponding pool is enabled for mirroring.

### DIFF
--- a/metrics/mixin/alerts/mirroring.libsonnet
+++ b/metrics/mixin/alerts/mirroring.libsonnet
@@ -23,7 +23,7 @@
           {
             alert: 'OdfPoolMirroringImageHealth',
             expr: |||
-              (ocs_pool_mirroring_image_health{%(ocsExporterSelector)s}  * on (namespace) group_left() (max by(namespace) (ocs_pool_mirroring_status{%(ocsExporterSelector)s}))) == 1
+              (ocs_pool_mirroring_image_health{%(ocsExporterSelector)s}  * on (namespace, name) group_left() (ocs_pool_mirroring_status{%(ocsExporterSelector)s})) == 1
             ||| % $._config,
             'for': $._config.odfPoolMirroringImageHealthWarningAlertTime,
             labels: {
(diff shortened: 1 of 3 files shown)


GowthamShanmugam:
Old alert expression behavior even after disabled mirroring for pool-1
image

New alert expression behavior after disabled mirroring for pool-1
image

metrics/mixin/alerts/mirroring.libsonnet
```diff
@@ -23,7 +23,7 @@
           {
             alert: 'OdfPoolMirroringImageHealth',
             expr: |||
-              (ocs_pool_mirroring_image_health{%(ocsExporterSelector)s}  * on (namespace) group_left() (max by(namespace) (ocs_pool_mirroring_status{%(ocsExporterSelector)s}))) == 1
```
GowthamShanmugam, line 26: The problem is, even after disabling the mirroring, the old mirroring status is displayed on Blackpool CR. So ```ocs_pool_mirroring_image_health``` will keep reporting the old health status. Here, a query expression is written like if one pool is enabled for mirroring then all other pools which have old mirroring status also get triggering an alert.

CR has old status or new status is does not matter, but a mirroring alert has to be triggered only when mirroring is enabled for the corresponding pool. so I am checking namespace as well as name while grouping.

GowthamShanmugam: @vasukulkarni

metrics/mixin/alerts/mirroring.libsonnet
```diff
@@ -23,7 +23,7 @@
...
+              (ocs_pool_mirroring_image_health{%(ocsExporterSelector)s}  * on (namespace, name) group_left() (ocs_pool_mirroring_status{%(ocsExporterSelector)s})) == 1
```
GowthamShanmugam, line 26: OdfPoolMirroringImageHealth alert will be triggered only when ocs_pool_mirroring_status is "1" (enabled) for the corresponding pool.

PR #1405, opened 2021-11-23
api: make api changes to support ocs-to-ocs externalStorage mode

iamniting:
As part of the PR below tasks are done
* update the API
* add the validation check for the wrong spec
* update the generated files in separate commits

Design Doc PR https://github.com/red-hat-storage/ocs-operator/pull/1409


iamniting:
I will update the manifest changes before merging, once we are good with the API changes.

iamniting:
@jarrpa @umangachapagain @nbalacha pls take a look.
cc @nb-ohad @kesavanvt @sp98

api/v1/storagecluster_types.go
```diff
@@ -137,11 +148,42 @@ type ManageCephObjectStoreUsers struct {
...
 
+// ExternalStorageMode specifies a mode of the external storage
+type ExternalStorageMode string
+
+const (
+	// ModeOcsToOcs specifies a "ocs-to-ocs" mode of the external storage
+	ModeOcsToOcs ExternalStorageMode = "ocs-to-ocs"
+
+	// ModeStanaloneCeph specifies a "standalone-ceph" mode of the external storage
+	ModeStanaloneCeph ExternalStorageMode = "standalone-ceph"
```
humblec, line 159: isnt it `ModeStandaloneCeph` better?

iamniting: Do you mean the value?

iamniting: Got it, that was a typo

iamniting: changed

api/v1/storagecluster_types.go
```diff
@@ -26,13 +26,21 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// ApiServerKind defines the type of servers
+type ApiServerKind string
+
+const (
+	// ProviderApiServer defines the ocs-to-ocs provider API server
+	ProviderApiServer ApiServerKind = "provider"
```
nb-ohad, line 34: Do we have more than one API server type/kind to support? 
I am not aware of any other API server implementation.

api/v1/storagecluster_types.go
```diff
@@ -75,6 +83,9 @@ type StorageClusterSpec struct {
 	// OverprovisionControl specifies the allowed hard-limit PVs overprovisioning relative to
 	// the effective usable storage capacity.
 	OverprovisionControl []OverprovisionControlSpec `json:"overprovisionControl,omitempty"`
+
+	// ApiServers defines the servers deployments which needs to be enabled
+	ApiServers map[ApiServerKind]bool `json:"apiServers,omitempty"`
```
nb-ohad, line 88: This doesn't seem to be very well defined. Even the term API server is very loaded and does not convey the entities that need to be deployed. And I am not even sure what should go into this map. 

I think a better API will be to have a boolean flag indicating that we are going to allow external storage consumers and the fact that we deploy an API server will be hidden as an implementation detail. 

So something like:
```Spec.AllowRemoteStorageConsumers = True/False```

File: api/v1/storagecluster_types.go [hunk 4]
```diff
@@ -137,11 +148,42 @@ type ManageCephObjectStoreUsers struct {
...
+// to be connected to the external cluster
 type ExternalStorageClusterSpec struct {
-	// +optional
+	//+kubebuilder:default:=true
+	// Enable enables the external mode
 	Enable bool `json:"enable,omitempty"`
+
+	//+kubebuilder:default:=standalone-ceph
+	// Mode describes the mode of the external storage
+	Mode ExternalStorageMode `json:"mode,omitempty"`
```
nb-ohad, line 171: Why `Mode`, this is a very generic word? can we try to be more specific?  `ExternalStorage` is by definition defining external mode and we do not want to add another new mode but extend the current one.

Can we rename it to something like `StorageProviderKind`, with 2 possible values: `ocs` and `rhcs` ?

api/v1/storagecluster_types.go
```diff
@@ -137,11 +148,42 @@ type ManageCephObjectStoreUsers struct {
...
+
+	// ConnectionString describes the endpoint and UID of the provider
+	// server to connect to the API server
+	ConnectionString string `json:"connectionString,omitempty"`
+
+	//+kubebuilder:default:="1TB"
+	// RequestedCapacity defines the requested storage on the provider
+	RequestedCapacity string `json:"requestedCapacity,omitempty"`
```
nb-ohad, line 179: I think it will benefit us to have all capacity-related values to be defined using the k8s type for sizes and not just plain strings.

nb-ohad (changes requested):

api/v1/storagecluster_types.go
iamniting: As of now, we have only one, But we don't know about the future it will help us to have more deployments in the future. Maybe we can rename the `ApiServers` in the spec to `Deployments`

nb-ohad: The point is that fact that the API server needs a new deployment is just an implementation detail and we should not expose it in the stroagecluster API (see one of my other comments below for a full explanation of my position)

api/v1/storagecluster_types.go
```diff
@@ -137,11 +148,42 @@ type ManageCephObjectStoreUsers struct {
...
 
+// ExternalStorageMode specifies a mode of the external storage
+type ExternalStorageMode string
+
+const (
+	// ModeOcsToOcs specifies a "ocs-to-ocs" mode of the external storage
+	ModeOcsToOcs ExternalStorageMode = "ocs-to-ocs"
+
+	// ModeStandaloneCeph specifies a "standalone-ceph" mode of the external storage
+	ModeStandaloneCeph ExternalStorageMode = "standalone-ceph"
```
sp98, line 159: would this mean existing users have to update their spec to have `standalone-ceph` mode?

api/v1/storagecluster_types.go
sp98: would this imply that existing customers have to update `StorageProviderKind` to `rhcs` or the upgrades might break?

sp98: How about we don't touch the existing External Storage API at all. Something like: 

```
 type StorageClusterSpec struct {
	ExternalStorage ExternalStorageClusterSpec `json:"externalStorage,omitempty"`
	ExternalOCS ExternalOCSClusterSpec
	...
}

type ExternalOCSClusterSpec struct {
	//+kubebuilder:default:=true
	// Enable enables the external mode

	// ConnectionString describes the endpoint and UID of the provider
	// server to connect to the API server
	ConnectionString string `json:"connectionString,omitempty"`

	//+kubebuilder:default:="1TB"
	// RequestedCapacity defines the requested storage on the provider
	RequestedCapacity string `json:"requestedCapacity,omitempty"`
}```

`ExternalStorage` would mean we are referring to `rhcs`
`ExternalOCS` would mean we are referring to an external ocs cluster. Its spec `ExternalOCSClusterSpec` would have API connection details and other fields we need to update, like `RequestedCapacity`

Comment (nb-ohad) [reply api/v1/storagecluster_types.go:171]: First, we can decide that the `StorageProviderKind` is defaulting to RHCS which means that everything is backward compatible. 

Second, I think that adding a new section will do more harm than good, as you then need to protect against cases when you have both and the API itself is becoming very confusing from an API user standpoint (like the obvious question, why do I have 2 sections? which one should I use?)

File: api/v1/storagecluster_types.go (hunk 6 above)
Comment (nb-ohad) [reply api/v1/storagecluster_types.go:159]: No, the spec will default to using the normal external mode if not specified.

File: api/v1/storagecluster_types.go (hunk 2 above)
Comment (iamniting) [reply api/v1/storagecluster_types.go]: done

File: api/v1/storagecluster_types.go (hunk 3 above)
Comment (iamniting) [reply api/v1/storagecluster_types.go]: done

File: api/v1/storagecluster_types.go (hunk 4 above)
Comment (iamniting) [reply api/v1/storagecluster_types.go]: done

File: api/v1/storagecluster_types.go (hunk 5 above)
Comment (iamniting) [reply api/v1/storagecluster_types.go]: done

**Review by nb-ohad — CHANGES_REQUESTED** [review]

File: api/v1/storagecluster_types.go [hunk 7]
```diff
@@ -137,11 +140,42 @@ type ManageCephObjectStoreUsers struct {
...
-	// +optional
+	//+kubebuilder:default:=true
+	// Enable enables the external mode
 	Enable bool `json:"enable,omitempty"`
+
+	//+kubebuilder:default:=rhcs
+	// StorageProviderKind describes the kind of the external storage
+	StorageProviderKind ExternalStorageKind `json:"storageProviderKind,omitempty"`
+
+	// ConnectionString describes the endpoint and UID of the provider
```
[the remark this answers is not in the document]
sp98: should it be a secret name instead?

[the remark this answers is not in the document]
iamniting: +1 for the secret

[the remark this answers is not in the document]
nb-ohad: @sp98 @iamniting  Why complicate stuff with a secret? It will force the customer to create a secret for non-confidential information? will complicate the setup and generally speaking does not make sense in my eyes.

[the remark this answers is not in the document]
nb-ohad: Ok, it was decided in an offline conversation to go with an encoded string (base64?) provided at the `Spec.ExternalStorage.ExternalStorageKind` field

sp98:
PR description and commit message descriptions are missing.

api/v1/storagecluster_types.go
```diff
@@ -137,11 +140,41 @@ type ManageCephObjectStoreUsers struct {
 	ReconcileStrategy string `json:"reconcileStrategy,omitempty"`
 }
 
+// ExternalStorageKind specifies a kind of the external storage
+type ExternalStorageKind string
+
+const (
+	// KindOcs specifies a "ocs" kind of the external storage
+	KindOcs ExternalStorageKind = "ocs"
```
sp98, line 148:
```suggestion
	KindOCS ExternalStorageKind = "ocs"
```

api/v1/storagecluster_types.go
```diff
@@ -137,11 +140,41 @@ type ManageCephObjectStoreUsers struct {
...
+
+	// KindRhcs specifies a "rhcs" kind of the external storage
+	KindRhcs ExternalStorageKind = "rhcs"
```
sp98, line 151:
```suggestion
	KindRHCS ExternalStorageKind = "rhcs"
```

api/v1/storagecluster_types.go
```diff
@@ -137,11 +140,41 @@ type ManageCephObjectStoreUsers struct {
...
+)
+
 // ExternalStorageClusterSpec defines the spec of the external Storage Cluster
-// to be connected to the local cluster
+// to be connected to the external cluster
```
sp98, line 155: I think the use of `local` was correct here.

api/v1/storagecluster_types.go
```diff
@@ -137,11 +140,41 @@ type ManageCephObjectStoreUsers struct {
...
 type ExternalStorageClusterSpec struct {
-	// +optional
+	//+kubebuilder:default:=true
```
sp98, line 157: Why `default: true`?

sp98 (changes requested):

api/v1/storagecluster_types.go
nb-ohad: Why do we need the enabled field at all, Why does the fact that we have this section on the CR enough for implying external storage?

sp98: existing implementation using `enabled` field.

nb-ohad: Ok, so in that case, we need to keep it to be backward compatible. 
But what is the expected behaviour if someone changes the value from true to false after deployment?

iamniting: > Why default: true?

So that if someone adds any other field under the `externalStorage` in the CR this will get filled automatically to true.

> Why do we need the enabled field at all, Why does the fact that we have this section on the CR enough for implying external storage?

This is still an optional feild while create a CR taking below `omitempty` code into consideration. 

```
ExternalStorage ExternalStorageClusterSpec `json:"externalStorage,omitempty"`
```

api/v1/storagecluster_types.go
nbalacha: The earlier sentence was correct. It refers to the local StorageCluster which will connect to the external ceph cluster

api/v1/storagecluster_types.go
```diff
@@ -137,11 +140,41 @@ type ManageCephObjectStoreUsers struct {
...
 	Enable bool `json:"enable,omitempty"`
+
+	//+kubebuilder:default:=rhcs
+	// StorageProviderKind describes the kind of the external storage
+	StorageProviderKind ExternalStorageKind `json:"storageProviderKind,omitempty"`
+
+	// ConnectionString describes the blob of provider
+	ConnectionString string `json:"connectionString,omitempty"`
+
+	//+kubebuilder:default:="1Ti"
```
iamniting, line 168: @nb-ohad I am thinking to remove this `kubebuilder:default` tag as it does not make sense in case of the rhcs and can mislead.

api/v1/storagecluster_types.go
nb-ohad: @iamniting  What about the question about the behavior upon changes to this field post deployment?

api/v1/storagecluster_types.go
```diff
@@ -137,11 +140,41 @@ type ManageCephObjectStoreUsers struct {
...
+	// RequestedCapacity defines the requested storage on the provider
+	RequestedCapacity resource.Quantity `json:"requestedCapacity,omitempty"`
+}
+
+// ExternalStorageClusterStatus defines the status of the external Storage Cluster
+// to be connected to the external cluster
+type ExternalStorageClusterStatus struct {
+	// ActualCapacity defines the actual storage provided by the provider
+	ActualCapacity resource.Quantity `json:"actualCapacity,omitempty"`
```
nbalacha, line 177: Does this require some sort of Succeded/Failed? status as well ?

api/v1/storagecluster_types.go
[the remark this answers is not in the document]
iamniting: correct and made the changes also to the API comment

api/v1/storagecluster_types.go
iamniting: not really, It should be handled the way it is currently handled in the external mode IMO

api/v1/storagecluster_types.go
iamniting: Can you pls elaborate?

nbalacha: Will setting the default to true mean that this will never be empty? I would prefer that it stay false to keep the behaviour the same.

api/v1/storagecluster_types.go
nbalacha, line 168: I'm not in favour of setting this default here. Let the StorageCluster have the requested value set explicitly.

api/v1/storagecluster_types.go
nbalacha: To clarify - what if someone creates an empty externalstorage struct in the StorageCluster. What is the behaviour with the default:true ?

iamniting: creating empty {} will be a problem as all these default values will be placed in the CR.

sp98: does that mean `kubebuilder:default` and golang's `omitempty` are contradicting and should not be used together?

nbalacha: In that case, please remove the default setting or set it to false.

nb-ohad: Setting it to false is problematic as it indicates that if it is not present we should not establish a connection to the remote server which is undefined behavior. If the sub-struct is present we should act upon it.

api/v1/storagecluster_types.go
nb-ohad: So how are we going to handle an empty value? are we going to hard code the 1TB value in the operator itself?

nbalacha: No, create a CR with the value specified.

nb-ohad: @nbalacha This is not an initial value but needs to be set from the start, with a specific setting we will not be able to onboard. 
The problem is that in order to support backword compatibility with the external mode we need to be able to not specify this field which bring us to the default value solution

api/v1/storagecluster_types.go
nbalacha: Disagree - we use the enable value field to check how to setup the cluster and reconcile the values. Currently an empty ExternalStorage struct is treated as not External storage.
If the value is not present, it would be better to return an error until the struct is edited.

api/v1/storagecluster_types.go
nb-ohad: Also, to the best of my understanding,  the default value should not affect the value in the CT itself. Only the in-memory struct created when reading the CR and in that case we can just ignore it in RHCS external mode

nb-ohad: @iamniting See my response below to @nbalacha comment around default values

nbalacha: But isn't the deployer creating the storagecluster CR and getting the capacity via the UI? Why then should this be a default value? It should be set in the StorageCluster CR created by the deployer.

nb-ohad: This is a product feature and as such, we need to treat this as an external API, deployer, or no deployer. 
So we can either choose to make the field non-optional, which is problematic because we need to support backword compatibility, or provide a default value and ignore it in the case of RHCS.

nbalacha: I'm not sure I understand. If you make the field optional, and require it to be set for OCS, won't it work?

api/v1/storagecluster_types.go
nb-ohad: But in that case what will happen if the value is reverted to false?

nb-ohad: I also don't believe there is a single-use case where the sub-struct is defined and the value is set to false. In an ideal world I will just stop watching that value in deprecate it (and in the future remove it completely)

nbalacha: When and by who would it be reverted? We can call out that the value cannot be changed once set , the way we do for other fields.

api/v1/storagecluster_types.go
nb-ohad: How would you suggest doing that? a reconcile failure? that is an anti-pattern as this is a CR configuration failure. 
I am aware of only 2 ways to deal with this:
1. validation webhook - out of scope right now
2. default value - no apparent downsides as we are already going to touch the external mode code inside ocs operator.

api/v1/storagecluster_types.go
nbalacha: I have seen empty structs in yamls when I read them from the cluster - I want to ensure we do not run into such a problem say if a YAML is edited in the console. I am against using the default value in this case and I do not see a big requirement for setting the default.

nb-ohad: > When and by who would it be reverted? We can call out that the value cannot be changed once set
In that case, let's mark it as immutable. but then there is no point in having it in the first place as if the storage cluster is created with false in this field it is unusable

api/v1/storagecluster_types.go
nbalacha: We also do not provide default values for OSDs in the CRD. I do not see why the same logic does not hold for the external mode in the managed service.

nb-ohad: Think about that differently - the default value allows us to broadcast the actual default value as part of the API itself. 
Assuming we will not take that route we will have to either treat that as a failure (my prev response) or provide a default value hardcoded in golang - which I think is a worse version of the current solution as every piece of code that needs the actual value will have to implement the same logic which can lead to bugs.

We also do not provide default values for OSDs in the CRD - when that was implemented we used the tools that were available for us then. Now we have new tools

nbalacha: I don't like the idea of having a value as part of the API. I would prefer that the value be set in the CR explicitly.
However, I would like to hear from Jose and Umanga on this point.

nb-ohad: Ok, but just another point, if we do not publish the default value in the API it means that we might (by mistake or intention) change it in future code changes which is problematic for such kind of value

nbalacha: I am not in favour of having the default value in the code at all. In my opinion, the best way would be to explicitly specify the value in the CR and check in the reconcile loop. This PR already has to check to validate that the connection string has been provided. This would be an additional check.

nb-ohad: @nbalacha We cannot do that as we cannot make this field mandatory for backward compatibility sake! 
If you have a way to deal with that please explain.

Just to clarify, I do agree that if this was a new section we would mark this field as mandatory and force the client who creates the CR to provide a value.

nb-ohad: Run time checks and reconcile failures are the worst solution as it has no guarantees around what was acted upon and what not and is an anti-pattern for static checks

nb-ohad: > It should be optional and the check-in the code would be to ensure that the field value is provided if the kind is OCS.
This is the exact case I am advocating against.

api/v1/storagecluster_types.go
iamniting: done

api/v1/storagecluster_types.go
iamniting: done

api/v1/storagecluster_types.go
iamniting: done

api/v1/storagecluster_types.go
iamniting: > So how are we going to handle an empty value? are we going to hard code the 1TB value in the operator itself?

I was thinking of the other way in the reconcile, CR validation as it is done in ocs-operator in many cases.

nb-ohad: @iamniting Please refer to the discussion below. 
I think as it stands now we need a 3rd opinion on the matter. Already pinged Umanaga and/or Jose

umangachapagain (commented):
I look at the proposed changes and I'm thinking "why"?

- Why are we mixing things?
  If ocs to ocs is drastically different from ocs to rhcs, then why not keep the API separate from StorageCluster?

- Why are we leaking product/offering details into user facing CRs?
  Remote consumer is a managed service offering. It has different assumptions than normal OCS. These shouldn't just be introduced into user facing CRs.
  
- Why does OCS care about who uses it and how?
  While creating a cluster, do I really need to care about who is going to use it? or how are they going to use it? If someone wants to create a StorageCluster and use it to provide remote storage, that's probably day 3 configuration. I mean Ceph doesn't ask if it's going to be used via Rook or OCS or in what mode.

- Why does OCS need to care about this extra API server?
  API server is an implementation detail. User shouldn't be thinking whether they want it or not. Also. OCS shouldn't be managing this server. This is a layer above OCS.


Maybe these discussions happened and I missed those. In that case point me to the right design docs or even better send a PR with the enhancement proposal, so that this can be discussed in the open. I refuse to review a PR with no context (since it doesn't look like a simple change).

nb-ohad:
> Why are we mixing things?
> If ocs to ocs is drastically different from ocs to rhcs, then why not keep the API separate from StorageCluster?

It's not, it is just a managed external mode and it uses the exact same mechanisms as external mode.

> Why are we leaking product/offering details into user facing CRs?
> Remote consumer is a managed service offering. It has different assumptions than normal OCS. These shouldn't just be > introduced into user facing CRs.

No, it is not, it is a product feature (unrelated to managed service) that we might expose to all customers in 4.11/4.12. 
It is true that the first use case it the Managed Service but we do have other use cases and the entire enhancement is designed to be a customer-facing standalone feature.  

> Why does OCS care about who uses it and how?
> While creating a cluster, do I really need to care about who is going to use it? or how are they going to use it? If someone > wants to create a StorageCluster and use it to provide remote storage, that's probably day 3 configuration. I mean Ceph > doesn't ask if it's going to be used via Rook or OCS or in what mode.

The API Change is to be used by the external cluster not the one providing the storage and as such it is a deployment feature (not a day 2 or day 3 feature), this is just the nature of the external mode.


> Why does OCS need to care about this extra API server?
> API server is an implementation detail. User shouldn't be thinking whether they want it or not. Also. OCS shouldn't be  managing this server. This is a layer above OCS.

We do not see it as a layer on top of OCS, the ability to allow external connection to an OCS cluster is going to be part of the product itself.  OCS operator is already the "above" layer where the actual storage orchestration is done by rook




>

api/v1/storagecluster_types.go
umangachapagain: This PR can either implement the new changes or fix an old implementation detail. We can't do both.
If you think what we're doing is wrong or doesn't make sense, please raise an issue with details and we can continue the discussion. For now, please revert this change @iamniting .

api/v1/storagecluster_types.go
```diff
@@ -137,11 +140,41 @@ type ManageCephObjectStoreUsers struct {
...
 // to be connected to the local cluster
 type ExternalStorageClusterSpec struct {
-	// +optional
+	//+kubebuilder:default:=true
+	// Enable enables the external mode
 	Enable bool `json:"enable,omitempty"`
+
+	//+kubebuilder:default:=rhcs
+	// StorageProviderKind describes the kind of the external storage
+	StorageProviderKind ExternalStorageKind `json:"storageProviderKind,omitempty"`
```
umangachapagain, line 163: Since only 2 values are acceptable, enum(ocs,rhcs) might be a good validation to add here.

api/v1/storagecluster_types.go
```diff
@@ -137,11 +140,41 @@ type ManageCephObjectStoreUsers struct {
...
+
+	// ConnectionString describes the blob of provider
```
umangachapagain, line 165: Comment needs to be more user friendly.

api/v1/storagecluster_types.go
```diff
@@ -137,11 +140,41 @@ type ManageCephObjectStoreUsers struct {
...
 // ExternalStorageClusterSpec defines the spec of the external Storage Cluster
 // to be connected to the local cluster
 type ExternalStorageClusterSpec struct {
-	// +optional
+	//+kubebuilder:default:=true
+	// Enable enables the external mode
 	Enable bool `json:"enable,omitempty"`
+
+	//+kubebuilder:default:=rhcs
+	// StorageProviderKind describes the kind of the external storage
```
umangachapagain, line 162: Comment needs to be user friendly.

api/v1/storagecluster_types.go
```diff
@@ -137,11 +140,41 @@ type ManageCephObjectStoreUsers struct {
...
+	StorageProviderKind ExternalStorageKind `json:"storageProviderKind,omitempty"`
+
+	// ConnectionString describes the blob of provider
+	ConnectionString string `json:"connectionString,omitempty"`
+
+	//+kubebuilder:default:="1Ti"
+	// RequestedCapacity defines the requested storage on the provider
```
umangachapagain, line 169: Should be a user friendly message. This is shown when we do `oc explain storagecluster.spec`.

api/v1/storagecluster_types.go
```diff
@@ -137,11 +140,41 @@ type ManageCephObjectStoreUsers struct {
...
+	RequestedCapacity resource.Quantity `json:"requestedCapacity,omitempty"`
+}
+
+// ExternalStorageClusterStatus defines the status of the external Storage Cluster
+// to be connected to the local cluster
+type ExternalStorageClusterStatus struct {
+	// ActualCapacity defines the actual storage provided by the provider
```
umangachapagain, line 176: Is this the capacity of the Provider OCS or Consumer OCS?

umangachapagain (changes requested):

api/v1/storagecluster_types.go
umangachapagain: Since it's a new field, it's ok to set sensible default. We just need to be careful about how we consume it and have necessary validations.

umangachapagain (changes requested):
3rd and 4th commit can be merged together as they're generated changes and tightly coupled.

api/v1/storagecluster_types.go
iamniting: It will be the capacity that is provided to the consumer by the provider. which means consumer capacity.

api/v1/storagecluster_types.go
iamniting: @umangachapagain reverted the change.

api/v1/storagecluster_types.go
iamniting: done

api/v1/storagecluster_types.go
iamniting: done

api/v1/storagecluster_types.go
iamniting: done

iamniting:
> 3rd and 4th commit can be merged together as they're generated changes and tightly coupled.

done

api/v1/storagecluster_types.go
iamniting: done

umangachapagain (commented):
Looks good to me. @nbalacha @jarrpa do you want to take another look?

umangachapagain:
@iamniting You need to update generated code. CI failure is valid.

iamniting:
> @iamniting You need to update generated code. CI failure is valid.

Done

api/v1/storagecluster_types.go
nb-ohad: @iamniting It think the name is misleading, Can we please rename form `ActualCapacity` to `GrantedCapacity`?

api/v1/storagecluster_types.go
```diff
@@ -305,6 +341,9 @@ type StorageClusterStatus struct {
 	// ExternalSecretHash holds the checksum value of external secret data.
 	ExternalSecretHash string `json:"externalSecretHash,omitempty"`
 
+	// ExternalStorage shows the status of the external cluster
+	ExternalStorage ExternalStorageClusterStatus `json:"externalStorage,omitempty"`
```
nb-ohad, line 345: I don't think this is something we would like to support? What value does it bring? what is the use case?

nb-ohad (changes requested):

api/v1/storagecluster_types.go
iamniting: After discussing it offline we came to a conclusion that we should remove the default tag, so removing it.

jarrpa (approved):
We had some short offline discussions to bring this PR over the line. Overall, @nbalacha and I still have some reservations, but I say not enough to hold the PR and we can address them later. The main request of removing the default value *kubebuilder annotation* so that it does not get encoded in the CRD schema was addressed.

/lgtm

iamniting:
/hold 

To decide where we want to merge this in main or ocs_to_ocs

PR #1406, opened 2021-11-24
misc fixes for rook-ceph toolbox

leseb:
Stop using tini on the toolbox job
Rook-Ceph has stopped using tini in the current devel code which will be
release with Rook 1.8 which will be the basis of the 4.10 branch.
So removing the tini entrypoint command to replace it with a simple bash
shell.

Run the toolbox with the rook user
Rook-Ceph now uses its own user to operate. The Operator runs as "rook"
and not "root" anymore so this should be reflected in the toolbox as
well.

Stop running the toolbox as privileged
The toolbox does not need to run privileged so let's remove this
capability.


nbalacha (approved):
/lgtm

PR #1410, opened 2021-11-25
add a tty to the toolbox

leseb:
Bash's job control needs a TTY to control a terminal, without it the
container fails to run. So adding a TTY to it.


nbalacha (approved):
/lgtm

PR #1411, opened 2021-11-26
toolbox: add `tty:true` to toolbox deployment

subhamkrai:
adding `tty:true` to toolbox deployment
as it need to acces terminal.

### DIFF
--- a/controllers/ocsinitialization/ocsinitialization_controller.go
+++ b/controllers/ocsinitialization/ocsinitialization_controller.go
@@ -58,7 +58,7 @@ func newToolsDeployment(namespace string, rookImage string) *appsv1.Deployment {
 	name := rookCephToolDeploymentName
 	var replicaOne int32 = 1
 
-	privilegedContainer := false
+	privilegedContainer := true
 	runAsNonRoot := true
 	var runAsUser, runAsGroup int64 = 2016, 2016
 	return &appsv1.Deployment{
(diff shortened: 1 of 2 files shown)


subhamkrai:
^^ @leseb

leseb (changes requested):
My bad, I got confused by both jobs, thanks for the follow-up. We need one commit to revert and one to add TTY to the toolbox.

nbalacha:
Removed Privileged = false because the toolbox was unable to access the ceph-keyring.
It was decided to revert the change to get this to work and then continue to debug this.

leseb (changes requested):
I don't think 54d344cec1382244dabd28a6bda2315ca99ce4cc is a revert commit, please use git revert and push again. We don't need the changes from `controllers/storagecluster/job_templates.go` either

nbalacha:
Looks good to me but will wait for @leseb's review.

leseb (changes requested):
Still, not there, we need to revert this https://github.com/red-hat-storage/ocs-operator/pull/1406/commits/bee58ad86273b5efde81e282d2a239022f13f137 properly. Then we can have a commit that add `tty` to the right place.

subhamkrai:
> Still, not there, we need to revert this [bee58ad](https://github.com/red-hat-storage/ocs-operator/commit/bee58ad86273b5efde81e282d2a239022f13f137) properly. Then we can have a commit that adds `tty` to the right place.

Added 3 commits

1. to revert privileged commit
2. to revert the wrong file to add tty
3. added commit to add tty in wright file

leseb (changes requested):
Almost there, when you revert a commit always explain why.

leseb (approved):

agarwal-mudit (approved):

PR #1412, opened 2021-11-26
Add encrypted RBD storage class during deployment

SanjalKatiyar:
Add a new encrypted RBD storage class during cluster deployment, if the customer asked for PV encryption.

URL: https://issues.redhat.com/browse/RHSTOR-2269

### DIFF
--- a/api/v1/storagecluster_types.go
+++ b/api/v1/storagecluster_types.go
@@ -284,8 +284,14 @@ type MonitoringSpec struct {
 // EncryptionSpec defines if encryption should be enabled for the Storage Cluster
 // It is optional and defaults to false.
 type EncryptionSpec struct {
+	// deprecated from OCS 4.10 onwards, acting as a dummy,
+	// UI will keep sending this flag for backward compatibility (OCP 4.10 + OCS 4.9)
 	// +optional
-	Enable               bool                     `json:"enable,omitempty"`
+	Enable bool `json:"enable,omitempty"`
+	// +optional
+	ClusterWide bool `json:"clusterWide,omitempty"`
+	// +optional
+	StorageClass         bool                     `json:"storageClass,omitempty"`
 	KeyManagementService KeyManagementServiceSpec `json:"kms,omitempty"`
 }
 
--- a/controllers/storagecluster/noobaa_system_reconciler.go
+++ b/controllers/storagecluster/noobaa_system_reconciler.go
@@ -180,9 +180,10 @@ func (r *StorageClusterReconciler) setNooBaaDesiredState(nb *nbv1.NooBaa, sc *oc
 	}
 
 	// Add KMS details to Noobaa spec, only if
-	// cluster-wide encryption is enabled (ie; sc.Spec.Encryption.Enable is True)
+	// cluster-wide encryption is enabled
+	// ie, sc.Spec.Encryption.ClusterWide/sc.Spec.Encryption.Enable is True
 	// and KMS ConfigMap is available
-	if sc.Spec.Encryption.Enable {
+	if sc.Spec.Encryption.Enable || sc.Spec.Encryption.ClusterWide {
 		if kmsConfig, err := getKMSConfigMap(KMSConfigMapName, sc, r.Client); err != nil {
 			return err
 		} else if kmsConfig != nil {
--- a/controllers/storagecluster/storageclasses.go
+++ b/controllers/storagecluster/storageclasses.go
@@ -230,6 +230,16 @@ func newCephBlockPoolStorageClassConfiguration(initData *ocsv1.StorageCluster) S
 	}
 }
 
+// newEncryptedCephBlockPoolStorageClassConfiguration generates configuration options for an encrypted Ceph Block Pool StorageClass.
+// when user has asked for PV encryption during deployment.
+func newEncryptedCephBlockPoolStorageClassConfiguration(initData *ocsv1.StorageCluster, serviceName string) StorageClassConfiguration {
+	encryptedStorageClassConfig := newCephBlockPoolStorageClassConfiguration(initData)
+	encryptedStorageClassConfig.storageClass.ObjectMeta.Name = generateNameForEncryptedCephBlockPoolSC(initData)
+	encryptedStorageClassConfig.storageClass.Parameters["encrypted"] = "true"
+	encryptedStorageClassConfig.storageClass.Parameters["encryptionKMSID"] = serviceName
+	return encryptedStorageClassConfig
+}
+
 // newCephOBCStorageClassConfiguration generates configuration options for a Ceph Object Store StorageClass.
 func newCephOBCStorageClassConfiguration(initData *ocsv1.StorageCluster) StorageClassConfiguration {
 	reclaimPolicy := corev1.PersistentVolumeReclaimDelete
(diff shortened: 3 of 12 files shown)


controllers/storagecluster/storageclasses.go
```diff
@@ -271,5 +281,15 @@ func (r *StorageClusterReconciler) newStorageClassConfigurations(initData *ocsv1
 	if initData.Spec.ExternalStorage.Enable || err == nil && !skip {
 		ret = append(ret, newCephOBCStorageClassConfiguration(initData))
 	}
+	// encrypted Ceph Block Pool storageclass will be returned only if
+	// storage-class-wide encryption + kms is enabled and KMS ConfigMap is available
+	if initData.Spec.Encryption.StorageClassWide && initData.Spec.Encryption.KeyManagementService.Enable {
+		kmsConfig, err := getKMSConfigMap(KMSConfigMapName, initData, r.Client)
+		if err == nil && kmsConfig != nil {
+			serviceName := kmsConfig.Data["KMS_SERVICE_NAME"]
```
SanjalKatiyar, line 289: need to check once with Rook and Ceph-CSI whether they are willing to change this key from`KMS_SERVICE_NAME` to `KMS_CONNECTION_NAME`

api/v1/storagecluster_types.go
```diff
@@ -247,8 +247,14 @@ type MonitoringSpec struct {
...
 type EncryptionSpec struct {
+	// deprecated from OCS 4.10 onwards, acting as a dummy,
+	// UI will keep sending this flag for backward compatibility (OCP 4.10 + OCS 4.9)
 	// +optional
-	Enable               bool                     `json:"enable,omitempty"`
+	Enable bool `json:"enable,omitempty"`
+	// +optional
+	ClusterWide bool `json:"clusterWide,omitempty"`
+	// +optional
+	StorageClassWide     bool                     `json:"storageClassWide,omitempty"`
```
GowthamShanmugam, line 257:
```suggestion
	StorageClass     bool                     `json:"storageClassWide,omitempty"`
```

SanjalKatiyar: This is what we are using from the UI: https://github.com/openshift/console/pull/10563/commits/2385083c7c71001d5b89e2657aa72a264a0f2644#r758951150

Same format I used in backend as well.

GowthamShanmugam: Ui is not using the name StorageClassWide while displaying the encryption page, so we better stick to that.

SanjalKatiyar: okay, got it... Will change it to...
	// +optional
	ClusterWide bool `json:"clusterWide,omitempty"`
	// +optional
	StorageClass    bool                     `json:"storageClass,omitempty"`

controllers/storagecluster/storageclasses.go
GowthamShanmugam: As we discussed ```KMS_CONNECTION_NAME``` is not required

GowthamShanmugam:
This change is LGTM


cc: @umangachapagain, @jarrpa requesting for further reviews.

controllers/storagecluster/storageclasses.go
GowthamShanmugam: We better log the err with clear message

SanjalKatiyar: ack...

SanjalKatiyar: added on L292

controllers/storagecluster/noobaa_system_reconciler.go
```diff
@@ -180,9 +180,9 @@ func (r *StorageClusterReconciler) setNooBaaDesiredState(nb *nbv1.NooBaa, sc *oc
 	}
 
 	// Add KMS details to Noobaa spec, only if
-	// cluster-wide encryption is enabled (ie; sc.Spec.Encryption.Enable is True)
+	// cluster-wide encryption is enabled (ie; sc.Spec.Encryption.ClusterWide is True)
 	// and KMS ConfigMap is available
-	if sc.Spec.Encryption.Enable {
+	if sc.Spec.Encryption.ClusterWide {
```
GowthamShanmugam, line 185: no need to check kms is enabled?

SanjalKatiyar: `getKMSConfigMap` will return `nil, nil` if it is not enabled.

GowthamShanmugam: we can avoid one network call :P

GowthamShanmugam: and there is a change if someone enabling encryption for nooba without actually enabling from storagecluster

SanjalKatiyar: Yup, but, it will return `nil` before polling for the resource. It will only poll if KMS is enabled.
https://github.com/red-hat-storage/ocs-operator/blob/1f34b667278a17aa9e82b6128144d50f796696ec/controllers/storagecluster/kms_resources.go#L115

SanjalKatiyar: `and there is a change if someone enables KMS encryption for nooba without actually enabling it from storagecluster` -->
sorry I did not get it, how will this cause an issue ?
I mean, if they are not using storagecluster CR then they should add it to the noobaa CR directly ?? There will be no extra network call.

GowthamShanmugam: ack

GowthamShanmugam:
```
 - ClusterWide and StorageClass are the two different flags to intimate the OCS operator to which type of encryption the user wants to enable.
  - ClusterWide is an OSD level encryption and StorageClass is a PV level encryption.
  - When KMS is enabled, 
        For ClusterWide encryption, KMS connection details are passed into CephCluster CR as well as Noobaa CR for the OSD encryption and RGW encryption. (KMS is optional for ClusterWide encryption).
        For StorageClass encryption, the OCS operator will create one extra StorageClass with encryption enabled. All PVCs which is created using this encrypted StorageClass on openshift-storage namespace are encrypted automatically. to create encrypted PVCs on other namespaces user has to create a tenant token on the corresponding namespace [https://github.com/openshift/ceph-csi/blob/release-4.7/docs/design/proposals/encryption-with-vault-tokens.md#configuration-stored-in-the-tenants-kubernetes-namespace](URL).
```

umangachapagain (changes requested):
This will not work for users upgrading from 4.9 to 4.10.

In OCS 4.9
```
Enable: true
ClusterWide: false (or nil)
StorageClass: false (or nil)
```
which means Encryption is enabled.

When upgraded to OCS 4.10, we ignore `enabled` and other fields are `false` which means encryption is disabled.

SanjalKatiyar:
> This will not work for users upgrading from 4.9 to 4.10.
> 
> In OCS 4.9
> 
> ```
> Enable: true
> ClusterWide: false (or nil)
> StorageClass: false (or nil)
> ```
> 
> which means Encryption is enabled.
> 
> When upgraded to OCS 4.10, we ignore `enabled` and other fields are `false` which means encryption is disabled.

We are taking that in account as well, in the follow up PR (https://github.com/red-hat-storage/ocs-operator/pull/1417)
PTAL: https://github.com/red-hat-storage/ocs-operator/pull/1417/commits/fbbf5f5d47eb02aea4d9604b73fd522546dfc813#r773616046

umangachapagain (approved):

PR #1414, opened 2021-11-30
pass kms vault options for ceph object store spec

thotz:
Currently kms config options are passed to ceph cluster spec, but for
RGW this need to pass seperately and need to set secret engine as
`transit` than `kv`.

Signed-off-by: Jiffin Tony Thottan <thottanjiffin@gmail.com>

### DIFF
--- a/controllers/storagecluster/cephobjectstores.go
+++ b/controllers/storagecluster/cephobjectstores.go
@@ -7,6 +7,7 @@ import (
 	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v1"
 	"github.com/red-hat-storage/ocs-operator/controllers/defaults"
 	cephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -37,8 +38,18 @@ func (obj *ocsCephObjectStores) ensureCreated(r *StorageClusterReconciler, insta
 		r.Log.Info("Platform is set to skip object store. Not creating a CephObjectStore.", "Platform", platform)
 		return nil
 	}
-
-	cephObjectStores, err := r.newCephObjectStoreInstances(instance)
+	kmsConfigMap, err := getKMSConfigMap(KMSConfigMapName, instance, r.Client)
+	if err != nil {
+		r.Log.Error(err, "Failed to procure KMS ConfigMap.", "KMSConfigMap", klog.KRef(instance.Namespace, KMSConfigMapName))
+		return err
+	}
+	if kmsConfigMap != nil {
+		if err = reachKMSProvider(kmsConfigMap); err != nil {
+			r.Log.Error(err, "Address provided in KMS ConfigMap is not reachable.", "KMSConfigMap", klog.KRef(kmsConfigMap.Namespace, kmsConfigMap.Name))
+			return err
+		}
+	}
+	cephObjectStores, err := r.newCephObjectStoreInstances(instance, kmsConfigMap)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


thotz:
PTAL @umangachapagain

thotz:
PTAL @leseb

umangachapagain (commented):
lgtm

controllers/storagecluster/cephobjectstores.go
```diff
@@ -175,6 +186,16 @@ func (r *StorageClusterReconciler) newCephObjectStoreInstances(initData *ocsv1.S
 			r.Log.Error(err, "Failed to set ControllerReference for CephObjectStore.", "CephObjectStore", klog.KRef(obj.Namespace, obj.Name))
 			return nil, err
 		}
+		// if kmsConfig is not 'nil', add the KMS details to CephCluster spec
+		if kmsConfigMap != nil {
+			// Set default KMS_PROVIDER and VAULT_SECRET_ENGINE values, refer https://issues.redhat.com/browse/RHSTOR-1963
+			if _, ok := kmsConfigMap.Data["KMS_PROVIDER"]; !ok {
+				kmsConfigMap.Data["KMS_PROVIDER"] = "vault"
+			}
+			kmsConfigMap.Data["VAULT_SECRET_ENGINE"] = "transit"
```
leseb, line 195: If you change `kmsConfigMap` it will change for the cluster-wide scenario too. So you need to populate a new map instead. If you don't on the next reconcile we will be passing `VAULT_SECRET_ENGINE = transit` to the OSD encryption scenario.
I'm not sure if the CephCluster and CephObjectStore CRs are reconciled later or not by OCS-Op but I'd like to avoid any issues with that. Also, I don't know if later in the code `kmsConfigMap` is reused or not.

So please create another instead of `kmsConfigMap` for the object store so we avoid any potential issues.

leseb (changes requested):

controllers/storagecluster/cephobjectstores.go
```diff
@@ -175,6 +186,28 @@ func (r *StorageClusterReconciler) newCephObjectStoreInstances(initData *ocsv1.S
 			r.Log.Error(err, "Failed to set ControllerReference for CephObjectStore.", "CephObjectStore", klog.KRef(obj.Namespace, obj.Name))
 			return nil, err
 		}
+		// if kmsConfig is not 'nil', add the KMS details to CephCluster spec
+		if kmsConfigMap != nil {
+			// Set default KMS_PROVIDER and VAULT_SECRET_ENGINE values, refer https://issues.redhat.com/browse/RHSTOR-1963
+			if _, ok := kmsConfigMap.Data["KMS_PROVIDER"]; !ok {
```
leseb, line 192: Is there a scenario where `KMS_PROVIDER` does not exist?

controllers/storagecluster/cephobjectstores.go
leseb: Why is this line still needed? `kmsConfigMap` uses `kv` not `transit`.

controllers/storagecluster/cephobjectstores.go
```diff
@@ -175,6 +186,28 @@ func (r *StorageClusterReconciler) newCephObjectStoreInstances(initData *ocsv1.S
...
+				kmsConfigMap.Data["KMS_PROVIDER"] = "vault"
+			}
+			kmsConfigMap.Data["VAULT_SECRET_ENGINE"] = "transit"
+			rgwConnDetails := make(map[string]string)
+			for key, value := range kmsConfigMap.Data {
+				// modify or ignore kv engine related options
+				if key == "VAULT_SECRET_ENGINE" {
```
leseb, line 199: can't we just set `rgwConnDetails.Data["VAULT_SECRET_ENGINE"] = "transit"` once the loop is done?

leseb (changes requested):

controllers/storagecluster/cephobjectstores.go
thotz: I have copied from [CephClusterSpec] (https://github.com/red-hat-storage/ocs-operator/blob/main/controllers/storagecluster/cephcluster.go#L402)

controllers/storagecluster/cephobjectstores.go
thotz: that also works

controllers/storagecluster/cephobjectstores.go
leseb: Ok.

leseb (approved):
/lgtm

umangachapagain (approved):

PR #1416, opened 2021-11-30
ocs-ci: Don't add user to /etc/passwd

nbalacha:
Nobody seems to know why this is required and the
ocs-ci run fails because of insufficient permissions
to do this. Removing this to see where exactly it fails.


jarrpa (approved):
/lgtm

malayparida2000:
@nbalacha Is this required any longer?

PR #1421, opened 2021-12-02
Implement a controller for ocs to ocs provider

iamniting:
create required resources to enable the remote connection on the provider.

### DIFF
--- a/controllers/storagecluster/provider_server.go
+++ b/controllers/storagecluster/provider_server.go
@@ -1,0 +1,264 @@
+package storagecluster
+
+import (
+	"context"
+	"fmt"
+	"math/rand"
+	"os"
+
+	"go.uber.org/multierr"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v1"
+)
+
+const (
+	ocsProviderServerName  = "ocs-provider-server"
+	providerAPIServerImage = "PROVIDER_API_SERVER_IMAGE"
+)
+
+type ocsProviderServer struct{}
+
+func (o *ocsProviderServer) ensureCreated(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) error {
+
+	if !instance.Spec.AllowRemoteStorageConsumers {
+		r.Log.Info("nigoyal Spec.AllowRemoteStorageConsumers is disabled")
+		return o.ensureDeleted(r, instance)
+	}
+
+	r.Log.Info("nigoyal Spec.AllowRemoteStorageConsumers is enabled creating resources")
+
+	var finalErr error
+
+	for _, f := range []func(*StorageClusterReconciler, *ocsv1.StorageCluster) error{
+		o.createSecret,
+		o.createService,
+		o.createDeployment,
+	} {
+		err := f(r, instance)
+		if err != nil {
+			multierr.AppendInto(&finalErr, err)
+			continue
+		}
+	}
+
+	return finalErr
+}
+
+func (o *ocsProviderServer) ensureDeleted(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) error {
+
+	// We do not check instance.Spec.AllowRemoteStorageConsumers because provider can disable this functionality
+	// and we need to delete the resources even the flag is not enabled (uninstall case).
+
+	// This func is directly called by the ensureCreated if the flag is disabled and deletes the resource
+	// Which means we do not need to call ensureDeleted while reconciling unless we are uninstalling
+
+	// NOTE: Do not add the check
+
+	var finalErr error
+
+	for _, resource := range []client.Object{
+		GetProviderAPIServerSecret(instance),
+		GetProviderAPIServerService(instance),
+		GetProviderAPIServerDeployment(instance),
+	} {
+		err := r.Client.Delete(context.TODO(), resource)
+
+		if err != nil && !errors.IsNotFound(err) {
+			r.Log.Error(err, "nigoyal Failed to delete resource", "Kind", resource.GetObjectKind(), "Name", resource.GetName())
+			multierr.AppendInto(&finalErr, err)
+		}
+	}
+
+	if finalErr == nil {
+		r.Log.Info("nigoyal Resource deletion for provider succeeded")
+	}
+
+	return finalErr
+}
+
+func (o *ocsProviderServer) createDeployment(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) error {
+
+	if _, ok := os.LookupEnv(providerAPIServerImage); !ok {
+		return fmt.Errorf("ENV var %s not found", providerAPIServerImage)
+	}
+
+	desiredDeployment := GetProviderAPIServerDeployment(instance)
+	actualDeployment := &appsv1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      desiredDeployment.Name,
+			Namespace: desiredDeployment.Namespace,
+		},
+	}
+
+	_, err := controllerutil.CreateOrUpdate(
+		context.TODO(), r.Client, actualDeployment,
+		func() error {
+			actualDeployment.Spec = desiredDeployment.Spec
+			return controllerutil.SetOwnerReference(instance, actualDeployment, r.Client.Scheme())
+		},
+	)
+	if err != nil && !errors.IsAlreadyExists(err) {
+		r.Log.Error(err, "nigoyal Failed to create/update deployment", "Name", desiredDeployment.Name)
+		return err
+	}
+
+	err = o.ensureDeploymentReplica(actualDeployment, desiredDeployment)
+	if err != nil {
+		r.Log.Error(err, "nigoyal Deployment is not ready", "Name", desiredDeployment.Name)
+		return err
+	}
+
+	r.Log.Info("nigoyal Deployment is running as desired")
+	return nil
+}
+
+func (o *ocsProviderServer) createService(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) error {
+
+	desiredService := GetProviderAPIServerService(instance)
+	actualService := &corev1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      desiredService.Name,
+			Namespace: desiredService.Namespace,
+		},
+	}
+
+	_, err := controllerutil.CreateOrUpdate(
+		context.TODO(), r.Client, actualService,
+		func() error {
+			actualService.Spec = desiredService.Spec
+			return controllerutil.SetOwnerReference(instance, actualService, r.Client.Scheme())
+		},
+	)
+	if err != nil && !errors.IsAlreadyExists(err) {
+		r.Log.Error(err, "nigoyal Failed to create/update service", "Name", desiredService.Name)
+		return err
+	}
+
+	r.Log.Info("nigoyal Service create/update succeeded")
+	return nil
+}
+
+func (o *ocsProviderServer) createSecret(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) error {
+
+	desiredSecret := GetProviderAPIServerSecret(instance)
+	actualSecret := &corev1.Secret{}
+
+	err := r.Client.Get(context.TODO(), client.ObjectKeyFromObject(desiredSecret), actualSecret)
+
+	if err != nil && errors.IsNotFound(err) {
+		err = r.Client.Create(context.TODO(), desiredSecret)
+		if err != nil {
+			r.Log.Error(err, "nigoyal Failed to create Secret", "Name", desiredSecret.Name)
+			return err
+		}
+		r.Log.Info("nigoyal Secret create succeeded", "Name", desiredSecret.Name)
+	} else if err != nil {
+		r.Log.Error(err, "nigoyal Failed to get Secret", "Name", desiredSecret.Name)
+		return err
+	}
+
+	return nil
+}
+
+func (o *ocsProviderServer) ensureDeploymentReplica(actual, desired *appsv1.Deployment) error {
+
+	if actual.Status.AvailableReplicas != *desired.Spec.Replicas {
+		return fmt.Errorf("Deployment %s is not ready", desired.Name)
+	}
+
+	return nil
+}
+
+func GetProviderAPIServerDeployment(instance *ocsv1.StorageCluster) *appsv1.Deployment {
+
+	return &appsv1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      ocsProviderServerName,
+			Namespace: instance.Namespace,
+		},
+		Spec: appsv1.DeploymentSpec{
+			Replicas: func(i int32) *int32 { return &i }(1),
+			Selector: &metav1.LabelSelector{
+				MatchLabels: map[string]string{
+					"app": "ocsProviderApiServer",
+				},
+			},
+			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{
+						"app": "ocsProviderApiServer",
+					},
+				},
+				Spec: corev1.PodSpec{
+					Containers: []corev1.Container{
+						{
+							Name:    "ocs-provider-api-server",
+							Image:   os.Getenv(providerAPIServerImage),
+							Command: []string{"/usr/local/bin/provider-api"},
+							Ports: []corev1.ContainerPort{
+								{
+									Name:          "ocs-provider",
+									ContainerPort: int32(50051),
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func GetProviderAPIServerService(instance *ocsv1.StorageCluster) *corev1.Service {
+
+	return &corev1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      ocsProviderServerName,
+			Namespace: instance.Namespace,
+		},
+		Spec: corev1.ServiceSpec{
+			Selector: map[string]string{
+				"app": "ocsProviderApiServer",
+			},
+			Ports: []corev1.ServicePort{
+				{
+					Port:       int32(50051),
+					TargetPort: intstr.FromString("ocs-provider"),
+				},
+			},
+			Type: corev1.ServiceTypeNodePort,
+		},
+	}
+}
+
+func GetProviderAPIServerSecret(instance *ocsv1.StorageCluster) *corev1.Secret {
+
+	return &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      ocsProviderServerName,
+			Namespace: instance.Namespace,
+		},
+		Immutable: func(flag bool) *bool { return &flag }(true),
+		StringData: map[string]string{
+			"Key": RandomString(1024),
+		},
+	}
+}
+
+//RandomString - Generate a random string of A-Z chars with len = l
+func RandomString(l int) string {
+
+	bytes := make([]byte, l)
+	for i := 0; i < l; i++ {
+		bytes[i] = byte(65 + rand.Intn(25)) //A=65 and Z = 65+25
+	}
+
+	return string(bytes)
+}
--- a/controllers/storagecluster/reconcile.go
+++ b/controllers/storagecluster/reconcile.go
@@ -350,6 +350,7 @@ func (r *StorageClusterReconciler) reconcilePhases(
 			// list of default ensure functions
 			// preserve list order
 			objs = []resourceManager{
+				&ocsProviderServer{},
 				&ocsTopologyMap{},
 				&ocsStorageQuota{},
 				&ocsCephConfig{},
@@ -372,6 +373,9 @@ func (r *StorageClusterReconciler) reconcilePhases(
 			}
 		}
 
+	} else if instance.Spec.ExternalStorage.StorageProviderKind == ocsv1.KindOCS {
+		// ocs to ocs external mode
+		objs = []resourceManager{}
 	} else {
 		// for external cluster, we have a different set of ensure functions
 		// preserve list order
--- a/controllers/storagecluster/storagecluster_controller.go
+++ b/controllers/storagecluster/storagecluster_controller.go
@@ -11,6 +11,7 @@ import (
 	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v1"
 	"github.com/red-hat-storage/ocs-operator/controllers/util"
 	cephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/version"
@@ -117,5 +118,7 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&cephv1.CephCluster{}).
 		Owns(&nbv1.NooBaa{}).
 		Owns(&corev1.PersistentVolumeClaim{}, builder.WithPredicates(pvcPredicate)).
+		Owns(&appsv1.Deployment{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		Owns(&corev1.Service{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
 		Complete(r)
 }
--- a/controllers/storagecluster/uninstall_reconciler.go
+++ b/controllers/storagecluster/uninstall_reconciler.go
@@ -283,6 +283,7 @@ func (r *StorageClusterReconciler) deleteResources(sc *ocsv1.StorageCluster) err
 	}
 
 	objs := []resourceManager{
+		&ocsProviderServer{},
 		&ocsNoobaaSystem{},
 		&ocsCephRGWRoutes{},
 		&ocsCephObjectStoreUsers{},
--- a/deploy/bundle/manifests/ocs-operator.clusterserviceversion.yaml
+++ b/deploy/bundle/manifests/ocs-operator.clusterserviceversion.yaml
@@ -2342,6 +2342,8 @@ spec:
                   value: noobaa/noobaa-core:master-20210912
                 - name: NOOBAA_DB_IMAGE
                   value: centos/postgresql-12-centos7
+                - name: PROVIDER_API_SERVER_IMAGE
+                  value: quay.io/ocs-dev/ocs-operator:latest
                 image: quay.io/ocs-dev/ocs-operator:latest
                 imagePullPolicy: Always
                 name: ocs-operator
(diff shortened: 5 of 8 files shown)


iamniting:
cc @nb-ohad @kesavanvt @sp98 

@jarrpa pls take a look

iamniting:
Will remove `nigoyal` from logs once we are good with the approach as this helps me to test my PR easily.

controllers/storagecluster/generate.go
```diff
@@ -79,3 +79,7 @@ func generateCephReplicatedSpec(initData *ocsv1.StorageCluster, poolType string)
 func generateStorageQuotaName(storageClassName, quotaName string) string {
 	return fmt.Sprintf("ocs-%s-%s", storageClassName, quotaName)
 }
+
+func generateProviderDeploymentName(initData *ocsv1.StorageCluster) string {
+	return fmt.Sprintf("%s-provider-server", initData.Name)
```
nb-ohad, line 84: I think a better name for the deployment would be 'external-endpointorconsumer-endpoint`

iamniting: Let's have more opinions on this. I am listing all 3 names here.

`ocs-storagecluster-external-endpoint`
`ocs-storagecluster-consumer-endpoint`
 `ocs-storagecluster-provider-server`

nb-ohad: But why d we need `ocs-storagecluster` as a prefix? Even if we have multiple storage clusters we need a single server. So the options should be:
`ocs-external-endpoint`
`ocs-consumer-endpoint`
`ocs-provider-server`

iamniting: correct I missed this.

controllers/storagecluster/server.go
```diff
@@ -0,0 +1,179 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v1"
+)
+
+var (
+	ServerImage = "quay.io/sp1098/ocs-operator:consumer-api"
```
iamniting, line 21: Question to the maintainers: How should I feed this image name in the ocs-operator?

controllers/storagecluster/generate.go
nb-ohad: We might want to add `ocs-consumer-server` to the list

controllers/storagecluster/server.go
nb-ohad: I was under the impressions that we have decided to not have a separate image and that the image will be the ocs operator image (running from a different entry point)

iamniting: That is correct, But we still need to either feed the image or fetch it on runtime from ocs-operator deployment.

controllers/storagecluster/server.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+
+	r.Log.Info("nigoyal Deployment is working as expected", "Name", desiredServerDeployment.Name)
+	return nil
+}
+
+func GetProviderServerDeployment(instance *ocsv1.StorageCluster) *appsv1.Deployment {
+
+	return &appsv1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
```
iamniting, line 129: Do we want to add owner references and monitor the resources for any changes? same goes for the service also.

controllers/storagecluster/server.go
```diff
@@ -0,0 +1,187 @@
...
+		},
+		Spec: corev1.ServiceSpec{
+			Selector: map[string]string{
+				"app": "apiServer",
+			},
+			Ports: []corev1.ServicePort{
+				{
+					Port:       int32(50051),
+					TargetPort: intstr.FromString("api-server-port"),
+					NodePort:   int32(30080),
```
iamniting, line 181: We can use node ports in range b/w `30000-32767`, Do we need any kind of approval before using any of the port?

controllers/storagecluster/server.go
nb-ohad: Why cant we just read the image name from the ocs-operator CSV?

nb-ohad: > We can do that but for that we either need a CSV name or we need to filter it out with ocs-operator name where ocs-operator deployment name will be always the same.

We already do exact that in the deployer, you can take a peak here:
https://github.com/red-hat-storage/ocs-osd-deployer/blob/fbc8169d7b7a81147309788813b07c2a88031aa1/controllers/managedocs_controller.go#L1389-L1399

> As far as I know, metrics exporter deployment is a part of the CSV which gets installed on the operator install not on storageCluster installation. So that's a different usecase.

Ok

iamniting: So you would like to filter with the `ocs-operator` prefix. ok I can do that

controllers/storagecluster/server.go
iamniting: I am not sure about that I can give it a try. @sp98 any idea?

controllers/storagecluster/server.go
nb-ohad: This is just one way to do it. Another option is to specify it in an env var as part of the CSV. That way it will be similar to all the other stuff that is already mentioned in the CSV.

controllers/storagecluster/server.go
```diff
@@ -0,0 +1,190 @@
...
+var (
+	ServerImage = "quay.io/sp1098/ocs-operator:consumer-api"
+)
+
+type ocsProviderServer struct{}
+
+func (obj *ocsProviderServer) ensureCreated(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) error {
+
+	if !instance.Spec.AllowRemoteStorageConsumers {
+		r.Log.Info("nigoyal Spec.AllowRemoteStorageConsumers is disabled")
```
jarrpa, line 29: I imagine all these log line will be removed before long. :wink:

controllers/storagecluster/server.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	r.Log.Info("nigoyal Spec.AllowRemoteStorageConsumers is enabled creating resources")
+
+	desiredServerService := GetProviderServerService(instance)
+	desiredServerDeployment := GetProviderServerDeployment(instance)
+
+	var finalErr error
+
+	for _, obj := range []client.Object{
```
jarrpa, line 40: ...don't use `obj` here, this is extremely confusing when the *struct you are writing a method for* is also `obj`. :stuck_out_tongue_closed_eyes: Honestly, though, you should change all the instances of `(obj *ocsProviderServer)` to something like `(srv *ocsProviderServer)`.

controllers/storagecluster/server.go
jarrpa: I'd say yes.

controllers/storagecluster/server.go
jarrpa: This is not required: https://v1-19.docs.kubernetes.io/docs/concepts/services-networking/service/#nodeport
```
      # Optional field
      # By default and for convenience, the Kubernetes control plane will allocate a port from a range (default: 30000-32767)
      nodePort: 30007
```

controllers/storagecluster/server.go
jarrpa: Could you just use downward API field references? https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/#the-downward-api e.g.:

```
      env:
        - name: IMAGE
          valueFrom:
            fieldRef:
              fieldPath: spec.containers[0].image
```

controllers/storagecluster/server.go
```diff
@@ -0,0 +1,190 @@
...
+		desiredServerDeployment,
+		desiredServerService,
+	} {
+		_, err := controllerutil.CreateOrUpdate(
+			context.TODO(), r.Client, obj,
+			func() error { return nil },
```
jarrpa, line 46: I might be forgetting, but I believe the `obj` being passed here would be overridden by `CreateOrUpdate()` with whatever it finds (or doesn't find). See the example here, where the desired PodTemplateSpec is defined separately and outside the `deploy` object: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#CreateOrUpdate Thus, you must use this in-line function to enforce the state you want.

controllers/storagecluster/server.go
```diff
@@ -0,0 +1,190 @@
...
+		}
+
+		r.Log.Info("nigoyal Resource created successfully", "Kind", obj.GetObjectKind(), "Name", obj.GetName())
+	}
+
+	if finalErr != nil {
+		return finalErr
+	}
+
+	if err := obj.isDeploymentReady(r, instance); err != nil {
```
jarrpa, line 62: Why do you need to pass `r` and `instance` into this? I would imagine you should be able to just pass the Deployment you already have and do the calculation.

jarrpa (changes requested):
All the use of "server" as a generic term for a very specific thing is confusing and technically misleading. Please rename `server.go` to something like `provider_server.go`. Similarly the instances of `apiServer` should probably be `ocsConsumerApiServer` or something similarly descriptive.

controllers/storagecluster/server.go
nb-ohad: @jarrpa @ issues here:
1. Referencing any container by an index is pruned to future issues due to container list location changes in future updates. 
2.  I don't think that the downward API support arbitrary spec fields, if I remember correctly there is a list of what can and can't be refernced to. But I might be wrong here.

controllers/storagecluster/server.go
nb-ohad: We use this pattern a lot in the deployer. 
The behaviour is a little bit different"
The object is taken then the result of the get is MERGED into it then the callback is run against the merged object to allow programmatic changes to the content before the create/update call is issued on the result.

The key point is that the merge operation will only affect fields that are defined in the k8s resource. This means that if the object presented to the function contains values that are undefined in the resource the outcome will be a struct that contains a merge of both.

controllers/storagecluster/server.go
iamniting: yes

controllers/storagecluster/server.go
iamniting: changed for loop obj to `resource` and method `obj` to `o`

controllers/storagecluster/server.go
iamniting: done

controllers/storagecluster/server.go
iamniting: removed

controllers/storagecluster/server.go
iamniting: I don't have the actual copy of the deployment, I have only the desired deployment. r is having a client which will help fetch the actual deployment with the help of an instance and verify its status.

controllers/storagecluster/generate.go
iamniting: going with `ocs-provider-server`

iamniting:
> All the use of "server" as a generic term for a very specific thing is confusing and technically misleading. Please rename `server.go` to something like `provider_server.go`. Similarly the instances of `apiServer` should probably be `ocsConsumerApiServer` or something similarly descriptive.

done

controllers/storagecluster/server.go
jarrpa: Looks like it. Okay, set the image as an annotation: https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/#capabilities-of-the-downward-api

controllers/storagecluster/server.go
jarrpa: Sure. And that sounds like something we wouldn't want, right? If the Deployment we Get() has a different PodSpec, wouldn't that override the desired PodSpec you want to actually have?

controllers/storagecluster/server.go
jarrpa: I don't mean the desired Deployment, I mean the actual Deployment. That's what `resources` here gets set to, right? https://github.com/red-hat-storage/ocs-operator/pull/1421/files#diff-85e662b500fee0a341fbae01f02cd05d31663d6f6e355ab6536eb8b2ca0ea38fR53

controllers/storagecluster/server.go
nb-ohad: @jarrpa But why? what is the diff between setting it directly in the ENV var or setting an annotation and then referencing it in the env var? You still need to have it manually set up somewhere!

controllers/storagecluster/server.go
nb-ohad: Usually, you start with an empty obj! you should not think of object passed as the template
For the deployer we do the following:
1. create an "empty" obj
2. run `createOrUpdate`
3. have a callback that takes the content of the template and replace/merge it into the obj

An example: https://github.com/red-hat-storage/ocs-osd-deployer/blob/main/controllers/managedocs_controller.go#L599-L626

nb-ohad: @jarrpa we are in agreement here! the mutate function is the one to enforce desired state

iamniting: Yes we are, I had an discussion with Jose on the same

iamniting: done

controllers/storagecluster/server.go
iamniting: done

controllers/storagecluster/provider_server.go
```diff
@@ -0,0 +1,243 @@
...
+	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v1"
+)
+
+var (
+	ServerImage = "quay.io/sp1098/ocs-operator:provider-api"
+)
+
+type ocsProviderServer struct{}
+
+func (o *ocsProviderServer) ensureCreated(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) error {
```
sp98, line 27: Where is `ensureCreated` being invoked?

sp98 (changes requested):
Unit tests?

controllers/storagecluster/provider_server.go
```diff
@@ -0,0 +1,243 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v1"
+)
+
+var (
+	ServerImage = "quay.io/sp1098/ocs-operator:provider-api"
```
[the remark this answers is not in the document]
iamniting: yeah we are in discussion on this.

controllers/storagecluster/provider_server.go
```diff
@@ -0,0 +1,243 @@
...
+	r.Log.Info("nigoyal Spec.AllowRemoteStorageConsumers is enabled creating resources")
+
+	var finalErr error
+
+	for _, f := range []func(*StorageClusterReconciler, *ocsv1.StorageCluster) error{
+		o.createSecret,
+		o.createService,
+		o.createDeployment,
+	} {
+		err := f(r, instance)
```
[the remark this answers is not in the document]
iamniting: I do want to reduce the reconcile calls wherever we can, thas why trying to create everything even any of them will fail. It should not affect us.

controllers/storagecluster/generate.go
```diff
@@ -79,3 +79,7 @@ func generateCephReplicatedSpec(initData *ocsv1.StorageCluster, poolType string)
 func generateStorageQuotaName(storageClassName, quotaName string) string {
 	return fmt.Sprintf("ocs-%s-%s", storageClassName, quotaName)
 }
+
+func generateProviderDeploymentName() string {
```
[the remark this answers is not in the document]
iamniting: yes will do that

controllers/storagecluster/provider_server.go
```diff
@@ -0,0 +1,243 @@
...
+				},
+			},
+		},
+	}
+}
+
+func GetProviderApiServerService(instance *ocsv1.StorageCluster) *corev1.Service {
+
+	return &corev1.Service{
+		ObjectMeta: metav1.ObjectMeta{
```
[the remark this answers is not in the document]
iamniting: @jarrpa your opinion here?

controllers/storagecluster/generate.go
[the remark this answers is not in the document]
iamniting: done

iamniting:
> Unit tests?

Will create once we finalize the actual code. This will save efforts to change tests again and again.

controllers/storagecluster/provider_server.go
```diff
@@ -0,0 +1,243 @@
...
+	if actual.Status.AvailableReplicas != *desired.Spec.Replicas {
+		return fmt.Errorf("Deployment %s is not ready", desired.Name)
+	}
+
+	return nil
+}
+
+func GetProviderApiServerDeployment(instance *ocsv1.StorageCluster) *appsv1.Deployment {
+
+	return &appsv1.Deployment{
```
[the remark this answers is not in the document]
nb-ohad: I think this is a good idea.

We have the same policy in the ocs-osd-deployer for most of the things that we reconcile. We do not use `embed` but `createOrUpdate` with a mutate function that starts with the template and applies dynamic changes on top of it. But the idea is almost identical.

controllers/storagecluster/provider_server.go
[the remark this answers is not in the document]
nb-ohad: copying my reply from the last comment


> I think this is a good idea.

> We have the same policy in the ocs-osd-deployer for most of the things that we reconcile. We do not use embed but createOrUpdate with a mutate function that starts with the template and applies dynamic changes on top of it. But the idea is almost identical.

controllers/storagecluster/provider_server.go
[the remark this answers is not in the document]
jarrpa: I'm not convinced that using `embed` would be an actual benefit to us... but I could be convinced otherwise.

Specifically, if we move all our static definitions into YAML files, we would then be in a place where we're parsing this text blob into a go struct. My main concern is how easy is it to debug errors? For example, what happens if we've changed a field or entire sub-struct in the go definition but the YAML manifest is not longer valid? Will that be any easier than continuing to hit nil-pointer errors because we keep using pointers to structs?

controllers/storagecluster/provider_server.go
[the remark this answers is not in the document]
jarrpa: I think Nitin's approach is fine. As long as the underlying system won't be broken by it, I generally prefer to just blast out all Create/Update calls at once and let the system resolve itself over time. Here specifically, if two of these succeed but one doe snot due to something like a network hiccup, there's no sense in blocking on just the one failure.

controllers/storagecluster/provider_server.go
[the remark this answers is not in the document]
nb-ohad: I would argue that YAML files are useless, the way we solved it in the deployer is by using const structs for the templates, one example can be found here: https://github.com/red-hat-storage/ocs-osd-deployer/blob/main/templates/alertmanagerconfig.go

This bring the benefits of static typing, and the ability to use code to generate some of the static parts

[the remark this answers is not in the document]
nb-ohad: To complete the picture, the pattern we use for the reconcile is:
```
_, err := ctrl.CreateOrUpdate(r.ctx, r.Client, r.alertmanagerConfig, func() error {
	if err := r.own(r.alertmanagerConfig); err != nil {
		return err
	}
	desired := templates.AlertmanagerConfigTemplate.DeepCopy()
	r.alertmanagerConfig.Spec = desired.Spec
	
	// Some code to apply the none static part of the reconcile
}

```

controllers/storagecluster/server.go
jarrpa: In general, hard-coding container images in code feels wrong. Unless there's a good reason to do it, the default should be to avoid it at all costs.

Here specifically, however, if we take the definition out of the go code and into the manifests that makes it easier for us to integrate it into our make/kustomize workflow so we don't have to worry about keeping track of it separately from what we're already doing to generate the bundle. It also gives us more flexibility as to where it can live, though off-hand if makes more sense (to me) to put the annotation on the Pod itself. Since the spec is generated by automation form the same value, I'm not to worried about writing it to multiple places.

controllers/storagecluster/provider_server.go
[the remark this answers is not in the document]
jarrpa: Yeah, if we can let's do this. :point_up:

controllers/storagecluster/server.go
jarrpa: Ah-ha... just talked to Ohad and I realized I was thinking about the wrong thing. Still, literally because we have the whole kustomize automation process, we can (and probably should) just inject the container image name directly into the Pod's environment variable. Let's do that!

controllers/storagecluster/provider_server.go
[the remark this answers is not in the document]
iamniting: We are doing something like this.

controllers/storagecluster/server.go
iamniting: done

controllers/storagecluster/provider_server.go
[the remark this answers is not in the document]
nb-ohad: That is not the same:
1. The template is kept in memory but live for all reconciles not a single one
2. A template (as an env variable or a file) is more declarative by nature, as it is a definition a declaration of a definition,  while code  (as in a func) is less unless the entire func is a declaration of a variable that is immediately returned. But at that point, there is no reason they have the function in the first place.
3. If you have a function, you do not need a template as a function can take the actual object and mutate it, but then you lose the declarative nature of templates. 


To summarize, my opinion, on the matter:
- Copy from a template is a good solution
- Mutating using code is also a good solution (but less readable and prune to ordering errors) 
- Combining them is does NOT create something that is better then the original options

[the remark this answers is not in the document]
jarrpa: I'm happy with this: https://github.com/red-hat-storage/ocs-operator/pull/1421/commits/bb44cbf3b72380f6aa1013d1c1d2834155e05d1e#diff-85e662b500fee0a341fbae01f02cd05d31663d6f6e355ab6536eb8b2ca0ea38fR91-R101

controllers/storagecluster/provider_server.go
```diff
@@ -0,0 +1,264 @@
...
+	actualDeployment := &appsv1.Deployment{}
+
+	actualDeployment.ObjectMeta = desiredDeployment.ObjectMeta
+	_, err := controllerutil.CreateOrUpdate(
+		context.TODO(), r.Client, actualDeployment,
+		func() error {
+			actualDeployment.Spec = desiredDeployment.Spec
+			return controllerutil.SetOwnerReference(instance, actualDeployment, r.Client.Scheme())
+		},
+	)
```
jarrpa, line 101: I'm happy with this! For me this resolves this discussion below: https://github.com/red-hat-storage/ocs-operator/pull/1421/commits/bb44cbf3b72380f6aa1013d1c1d2834155e05d1e#r765682700

nb-ohad: Copying metadata is very problematic it removes additional labels and annotations that are set from the outside and this is something that we encounter again and again!

The metadata is both input and output we should never treat it as output-only or input only section

nb-ohad: To clarify, I am referring to line 94

iamniting: I agree, resolved the problem at all places.

jarrpa:
Marking my general LGTM on this PR. It needs at least one more push to pass CI, but the fixes don't look like they'd change anything I'd disapprove. :wink:

controllers/storagecluster/provider_server.go
```diff
@@ -0,0 +1,272 @@
...
+	}
+
+	return nil
+}
+
+func (o *ocsProviderServer) isDeploymentReady(actual, desired *appsv1.Deployment) error {
+
+	if actual.Status.AvailableReplicas != *desired.Spec.Replicas {
+		return fmt.Errorf("Deployment %s is not ready", desired.Name)
+	}
```
jarrpa, line 173: I also like this! Look how simple it is. :grin: For a future PR, I would like to see the following:

* Rather than just returning `error`, I would suggest `bool, error`. Yes checking for `err != nil` achieves more or less the same thing, but for most any function starting with `is*` it makes semantic sense for it to be treated as or return a boolean. 
* In the error message, include what criteria weren't met for the Deployment to be considered "ready".

jarrpa (approved):

controllers/storagecluster/provider_server.go
nb-ohad: Why `bool, error` for a sync function that works on memory only. I think `is` function should only return `bool` and the log or error should be created from the outside context. That way the util is more usable for other use cases where you don't actually need an error. For example, think about the `!is` use case

jarrpa: > For example, think about the `!is` use case

Exactly this! Specifically, though, if the `is` function grows to more than just a single check (which it is now), we should understand why the function returned false when it does. And yes, we could effectively log the error in-place and return false, but then what does the human debugging / RCA experience look like? And error message gets propagated back up and could be exposed in the CR Status, but a log message is just a log. I think?

nb-ohad: > if the is function grows to more than just a single check 
So it is not an `is` function anymore, if you insist you can also return a enum string for a reason. Again, no need for error

iamniting: I changed the name of the func itself to the `ensureDeploymentReplica` than the func structure.

iamniting:
@jarrpa Now CI is passing other than the `ocs-operator-ci`. It is working well on my machine. But somehow export is not working in the CI. Any idea what can be wrong here?

controllers/storagecluster/provider_server.go
```diff
@@ -0,0 +1,272 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v1"
+)
+
+var (
+	ServerImage = GetEnvVarOrPanic("PROVIDER_API_SERVER_IMAGE")
```
sp98, line 22:
```suggestion
	serverImage = GetEnvVarOrPanic("PROVIDER_API_SERVER_IMAGE")
```
Name is not used outside the package.

controllers/storagecluster/provider_server.go
```diff
@@ -0,0 +1,272 @@
...
+)
+
+const (
+	OcsProviderServerName = "ocs-provider-server"
```
sp98, line 26:
```suggestion
	ocsProviderServerName = "ocs-provider-server"
```
Name is not used outside the package.

controllers/storagecluster/provider_server.go
```diff
@@ -0,0 +1,272 @@
...
+)
+
+type ocsProviderServer struct{}
```
sp98, line 29: empty struct?

sp98 (changes requested):

controllers/storagecluster/provider_server.go
iamniting: Not now, We may use them in the e2e tests which are not part of the same pkg.

controllers/storagecluster/provider_server.go
iamniting: same as above.

controllers/storagecluster/provider_server.go
iamniting: The struct is required to Implement the `resourceManager` Interface. https://github.com/red-hat-storage/ocs-operator/blob/main/controllers/storagecluster/reconcile.go#L33

jarrpa (approved):
/lgtm

PR #1425, opened 2021-12-10
Introduce annotation to disable features in console

bipuladh:
OCP console will parse `features.ocs.openshift.io/disabled` annotation to disable features.
Disabling features of UI is required to aid migration of components from OCP to ODF repo. 


jarrpa (approved):
/lgtm

PR #1430, opened 2021-12-14
ocs-to-ocs: Implement provider api server

sp98:
This PR implements the gRPC provider api-server. 

Pending items that will be taken up an follow up PR:
- [ ] New onbording ticket generation and validating ticket using public key
- [ ] Add TLS cert to api server 
- [ ] StorageCluster CR resource deletion should not be allowed in case there is one ore more StorageConsumer resources available.

### DIFF
--- a/services/provider/main.go
+++ b/services/provider/main.go
@@ -1,8 +1,12 @@
 package main
 
 import (
+	"context"
 	"flag"
+	"os"
+	"os/signal"
 
+	"github.com/red-hat-storage/ocs-operator/controllers/util"
 	"github.com/red-hat-storage/ocs-operator/services/provider/server"
 	"google.golang.org/grpc"
 	"k8s.io/klog"
@@ -17,6 +21,20 @@ func main() {
 
 	klog.Info("Starting Provider API server")
 
-	var opts []grpc.ServerOption
-	server.Start(*port, opts)
+	namespace, err := util.GetWatchNamespace()
+	if err != nil {
+		klog.Errorf("failed to get provider cluster namespace. %v", err)
+		return
+	}
+
+	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt)
+	defer stop()
+
+	providerServer, err := server.NewOCSProviderServer(ctx, namespace)
+	if err != nil {
+		klog.Errorf("failed to start provider server. %v", err)
+		return
+	}
+
+	providerServer.Start(*port, []grpc.ServerOption{})
 }
--- a/services/provider/server/consumer.go
+++ b/services/provider/server/consumer.go
@@ -1,0 +1,182 @@
+package server
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"sync"
+
+	ocsv1alpha1 "github.com/red-hat-storage/ocs-operator/api/v1alpha1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/klog"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	errTicketAlreadyExists = errors.New("onboarding ticket already used by another storageConsumer")
+)
+
+type ocsConsumerManager struct {
+	client       client.Client
+	namespace    string
+	nameByTicket map[string]string
+	nameByUID    map[types.UID]string
+	mutex        sync.RWMutex
+}
+
+func newConsumerManager(ctx context.Context, client client.Client, namespace string) (*ocsConsumerManager, error) {
+	consumers := &ocsv1alpha1.StorageConsumerList{}
+	err := client.List(ctx, consumers)
+	if err != nil {
+		return nil, fmt.Errorf("failed to list storage consumers. %v", err)
+	}
+
+	nameByTicket := map[string]string{}
+	nameByUID := map[types.UID]string{}
+
+	for _, consumer := range consumers.Items {
+		nameByUID[consumer.UID] = consumer.Name
+
+		if ticket, ok := consumer.GetAnnotations()[TicketAnnotation]; ok {
+			nameByTicket[ticket] = consumer.Name
+		}
+	}
+
+	return &ocsConsumerManager{
+		client:       client,
+		namespace:    namespace,
+		nameByTicket: nameByTicket,
+		nameByUID:    nameByUID,
+	}, nil
+}
+
+// Create creates a new storageConsumer resource, updates the consumer cache and returns the storageConsumer UID
+func (c *ocsConsumerManager) Create(ctx context.Context, name, ticket string, capacity resource.Quantity) (string, error) {
+	c.mutex.RLock()
+	if _, ok := c.nameByTicket[ticket]; ok {
+		c.mutex.RUnlock()
+		klog.Warning("onboarding ticket already in use")
+		return "", errTicketAlreadyExists
+	}
+	c.mutex.RUnlock()
+
+	consumerObj := &ocsv1alpha1.StorageConsumer{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: c.namespace,
+			Annotations: map[string]string{
+				TicketAnnotation: ticket,
+			},
+		},
+		Spec: ocsv1alpha1.StorageConsumerSpec{
+			Capacity: capacity,
+		},
+	}
+
+	err := c.client.Create(ctx, consumerObj)
+	if err != nil {
+		if kerrors.IsAlreadyExists(err) {
+			klog.Warningf("storageConsumer %q already exists", name)
+			return "", err
+		}
+		return "", fmt.Errorf("failed to create storageConsumer resource %q. %v", consumerObj.Name, err)
+	}
+
+	c.mutex.Lock()
+	c.nameByUID[consumerObj.UID] = consumerObj.Name
+	c.nameByTicket[ticket] = consumerObj.Name
+	c.mutex.Unlock()
+
+	klog.Infof("successfully created storageConsumer resource %q", name)
+
+	return string(consumerObj.UID), nil
+}
+
+// Delete deletes the storageConsumer resource using UID and updates the consumer cache
+func (c *ocsConsumerManager) Delete(ctx context.Context, id string) error {
+	uid := types.UID(id)
+	c.mutex.RLock()
+	consumerName, ok := c.nameByUID[uid]
+	if !ok {
+		klog.Warningf("no storageConsumer found with UID %q", id)
+		c.mutex.RUnlock()
+		return nil
+	}
+	c.mutex.RUnlock()
+
+	consumerObj := &ocsv1alpha1.StorageConsumer{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      consumerName,
+			Namespace: c.namespace,
+		},
+	}
+	if err := c.client.Delete(ctx, consumerObj); err != nil {
+		if kerrors.IsNotFound(err) {
+			// update uidStore
+			c.mutex.Lock()
+			delete(c.nameByUID, uid)
+			c.mutex.Unlock()
+			klog.Warningf("storageConsumer %q not found.", consumerObj.Name)
+			return nil
+		}
+		return fmt.Errorf("failed to delete storageConsumer %q. %v", consumerName, err)
+	}
+
+	c.mutex.Lock()
+	delete(c.nameByUID, uid)
+	c.mutex.Unlock()
+
+	klog.Infof("successfully deleted storageConsumer resource %q", consumerName)
+
+	return nil
+}
+
+// UpdateCapacity updates the capacity field in storageConsumer resource
+func (c *ocsConsumerManager) UpdateCapacity(ctx context.Context, id string, capacity resource.Quantity) error {
+	// Get storage consumer resource using UID
+	consumerObj, err := c.Get(ctx, id)
+	if err != nil {
+		return err
+	}
+
+	consumerObj.Spec.Capacity = capacity
+	err = c.client.Update(ctx, consumerObj)
+	if err != nil {
+		return fmt.Errorf("failed to update storageConfig resource %q. %v", consumerObj.Name, err)
+	}
+
+	klog.Infof("successfully updated requested capacity field in the StorageConsumer resource %q", consumerObj.Name)
+
+	return nil
+}
+
+// Get returns a storageConsumer resource using the UID
+func (c *ocsConsumerManager) Get(ctx context.Context, id string) (*ocsv1alpha1.StorageConsumer, error) {
+	uid := types.UID(id)
+
+	c.mutex.RLock()
+	consumerName, ok := c.nameByUID[uid]
+	if !ok {
+		c.mutex.RUnlock()
+		klog.Errorf("no storageConsumer found with the UID %q", id)
+		return nil, fmt.Errorf("no storageConsumer found with the UID %q", id)
+	}
+	c.mutex.RUnlock()
+
+	consumerObj := &ocsv1alpha1.StorageConsumer{}
+	if err := c.client.Get(ctx, types.NamespacedName{Name: consumerName, Namespace: c.namespace}, consumerObj); err != nil {
+		if kerrors.IsNotFound(err) {
+			// update uidStore
+			c.mutex.Lock()
+			delete(c.nameByUID, uid)
+			c.mutex.Unlock()
+			return nil, fmt.Errorf("storageConsumer resource %q not found. %v", consumerName, err)
+		}
+		return nil, fmt.Errorf("failed to get storageConsumer resource with name %q. %v", consumerName, err)
+	}
+
+	return consumerObj, nil
+}
--- a/services/provider/server/server.go
+++ b/services/provider/server/server.go
@@ -6,65 +6,186 @@ import (
 	"net"
 	"os"
 
+	ocsv1alpha1 "github.com/red-hat-storage/ocs-operator/api/v1alpha1"
 	"github.com/red-hat-storage/ocs-operator/services/provider/common"
 	pb "github.com/red-hat-storage/ocs-operator/services/provider/pb"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/reflection"
+	"google.golang.org/grpc/status"
+	corev1 "k8s.io/api/core/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/klog"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
 )
 
-type ocsProviderServer struct {
+const (
+	TicketAnnotation = "ocs.openshift.io/provider-onboarding-ticket"
+)
+
+type OCSProviderServer struct {
 	pb.UnimplementedOCSProviderServer
+	client          client.Client
+	consumerManager *ocsConsumerManager
+	namespace       string
+}
+
+func NewOCSProviderServer(ctx context.Context, namespace string) (*OCSProviderServer, error) {
+	client, err := newClient()
+	if err != nil {
+		return nil, fmt.Errorf("failed to create new client. %v", err)
+	}
+
+	consumerManager, err := newConsumerManager(ctx, client, namespace)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create new OCSConumer instance. %v", err)
+	}
+
+	return &OCSProviderServer{
+		client:          client,
+		consumerManager: consumerManager,
+		namespace:       namespace,
+	}, nil
 }
 
 // OnboardConsumer RPC call to onboard a new OCS consumer cluster.
-func (c *ocsProviderServer) OnboardConsumer(ctx context.Context, req *pb.OnboardConsumerRequest) (*pb.OnboardConsumerResponse, error) {
+func (s *OCSProviderServer) OnboardConsumer(ctx context.Context, req *pb.OnboardConsumerRequest) (*pb.OnboardConsumerResponse, error) {
 	mock := os.Getenv(common.MockProviderAPI)
 	if mock != "" {
 		return mockOnboardConsumer(common.MockError(mock))
 	}
-	return &pb.OnboardConsumerResponse{}, nil
+
+	// Validate capacity
+	capacity, err := resource.ParseQuantity(req.Capacity)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "%q is not a valid storageConsumer capacity: %v", req.Capacity, err)
+	}
+
+	// Validate onboardingTicket
+	// TODO: check if ticket is already used.
+	// TODO: check expiry of the ticket
+	// TODO: validate ticket with public key
+
+	storageConsumerUUID, err := s.consumerManager.Create(ctx, req.ConsumerName, req.OnboardingTicket, capacity)
+	if err != nil {
+		if kerrors.IsAlreadyExists(err) || err == errTicketAlreadyExists {
+			return nil, status.Errorf(codes.AlreadyExists, "failed to create storageConsumer %q. %v", req.ConsumerName, err)
+		}
+		return nil, status.Errorf(codes.Internal, "failed to create storageConsumer %q. %v", req.ConsumerName, err)
+	}
+
+	// TODO: send correct granted capacity
+	return &pb.OnboardConsumerResponse{StorageConsumerUUID: storageConsumerUUID, GrantedCapacity: req.Capacity}, nil
 }
 
 // GetStorageConfig RPC call to onboard a new OCS consumer cluster.
-func (c *ocsProviderServer) GetStorageConfig(ctx context.Context, req *pb.StorageConfigRequest) (*pb.StorageConfigResponse, error) {
+func (s *OCSProviderServer) GetStorageConfig(ctx context.Context, req *pb.StorageConfigRequest) (*pb.StorageConfigResponse, error) {
 	mock := os.Getenv(common.MockProviderAPI)
 	if mock != "" {
 		return mockGetStorageConfig(common.MockError(mock))
 	}
-	return &pb.StorageConfigResponse{}, nil
-}
 
-// OffboardConsumer RPC call to delete the StorageConsumer CR
-func (c *ocsProviderServer) OffboardConsumer(ctx context.Context, req *pb.OffboardConsumerRequest) (*pb.OffboardConsumerResponse, error) {
-	mock := os.Getenv(common.MockProviderAPI)
-	if mock != "" {
-		return mockOffboardConsumer(common.MockError(mock))
+	// Get storage consumer resource using UUID
+	consumerObj, err := s.consumerManager.Get(ctx, req.StorageConsumerUUID)
+	if err != nil {
+		return nil, err
 	}
-	return &pb.OffboardConsumerResponse{}, nil
+
+	// Verify Status
+	switch consumerObj.Status.State {
+	case ocsv1alpha1.StorageConsumerStateFailed:
+		// TODO: get correct error message from the storageConsumer status
+		return nil, status.Errorf(codes.Internal, "storageConsumer status failed")
+	case ocsv1alpha1.StorageConsumerStateConfiguring:
+		return nil, status.Errorf(codes.Unavailable, "waiting for the rook resources to be provisioned")
+	case ocsv1alpha1.StorageConsumerStateDeleting:
+		return nil, status.Errorf(codes.NotFound, "storageConsumer is already in deleting phase")
+	case ocsv1alpha1.StorageConsumerStateReady:
+		// TODO: Return json connection details
+		return &pb.StorageConfigResponse{}, nil
+	}
+
+	return nil, status.Errorf(codes.Unavailable, "storage consumer status is not set")
 }
 
 // UpdateCapacity PRC call to increase or decrease the storage pool size
-func (c *ocsProviderServer) UpdateCapacity(ctx context.Context, req *pb.UpdateCapacityRequest) (*pb.UpdateCapacityResponse, error) {
+func (s *OCSProviderServer) UpdateCapacity(ctx context.Context, req *pb.UpdateCapacityRequest) (*pb.UpdateCapacityResponse, error) {
 	mock := os.Getenv(common.MockProviderAPI)
 	if mock != "" {
 		return mockUpdateCapacity(common.MockError(mock))
 	}
+
+	capacity, err := resource.ParseQuantity(req.Capacity)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "%q is not a valid resource capacity: %v", req.Capacity, err)
+	}
+
+	if err := s.consumerManager.UpdateCapacity(ctx, req.StorageConsumerUUID, capacity); err != nil {
+		if kerrors.IsNotFound(err) {
+			return nil, status.Errorf(codes.NotFound, "failed to update capacity in the storageConsumer resource: %v", err)
+		}
+		return nil, status.Errorf(codes.Internal, "failed to update capacity in the storageConsumer resource: %v", err)
+	}
+
+	// TODO: Return granted capacity correctly.
 	return &pb.UpdateCapacityResponse{}, nil
 }
 
-func Start(port int, opts []grpc.ServerOption) {
+// OffboardConsumer RPC call to delete the StorageConsumer CR
+func (s *OCSProviderServer) OffboardConsumer(ctx context.Context, req *pb.OffboardConsumerRequest) (*pb.OffboardConsumerResponse, error) {
+	mock := os.Getenv(common.MockProviderAPI)
+	if mock != "" {
+		return mockOffboardConsumer(common.MockError(mock))
+	}
+
+	err := s.consumerManager.Delete(ctx, req.StorageConsumerUUID)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to delete storageConsumer resource with the provided UUID. %v", err)
+	}
+
+	return &pb.OffboardConsumerResponse{}, nil
+}
+
+func (s *OCSProviderServer) Start(port int, opts []grpc.ServerOption) {
 	lis, err := net.Listen("tcp", fmt.Sprintf(":%d", port))
 	if err != nil {
 		klog.Fatalf("failed to listen: %v", err)
 	}
 
 	grpcServer := grpc.NewServer(opts...)
-	pb.RegisterOCSProviderServer(grpcServer, &ocsProviderServer{})
+	pb.RegisterOCSProviderServer(grpcServer, s)
 	// Register reflection service on gRPC server.
 	reflection.Register(grpcServer)
 	err = grpcServer.Serve(lis)
 	if err != nil {
 		klog.Fatalf("failed to start gRPC server: %v", err)
 	}
+}
+
+func newClient() (client.Client, error) {
+	scheme := runtime.NewScheme()
+	err := ocsv1alpha1.AddToScheme(scheme)
+	if err != nil {
+		return nil, fmt.Errorf("failed to add ocsv1alpha1 to scheme. %v", err)
+	}
+	err = corev1.AddToScheme(scheme)
+	if err != nil {
+		return nil, fmt.Errorf("failed to add ocsv1alpha1 to scheme. %v", err)
+	}
+
+	config, err := config.GetConfig()
+	if err != nil {
+		klog.Error(err, "failed to get rest.config")
+		return nil, err
+	}
+	client, err := client.New(config, client.Options{Scheme: scheme})
+	if err != nil {
+		klog.Error(err, "failed to create controller-runtime client")
+		return nil, err
+	}
+
+	return client, nil
 }
--- a/services/provider/server/consumer_test.go
+++ b/services/provider/server/consumer_test.go
@@ -1,0 +1,180 @@
+package server
+
+import (
+	"context"
+	"testing"
+
+	api "github.com/red-hat-storage/ocs-operator/api/v1"
+	ocsv1alpha1 "github.com/red-hat-storage/ocs-operator/api/v1alpha1"
+	"github.com/stretchr/testify/assert"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+const (
+	testNamespace = "test"
+)
+
+var (
+	consumer1 = &ocsv1alpha1.StorageConsumer{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "consumer1",
+			Namespace: testNamespace,
+			Annotations: map[string]string{
+				TicketAnnotation: "ticket1",
+			},
+			UID: "uid1",
+		},
+		Spec: ocsv1alpha1.StorageConsumerSpec{
+			Capacity: resource.MustParse("1G"),
+		},
+	}
+
+	consumer2 = &ocsv1alpha1.StorageConsumer{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "consumer2",
+			Namespace: testNamespace,
+			Annotations: map[string]string{
+				TicketAnnotation: "ticket2",
+			},
+			UID: "uid2",
+		},
+	}
+)
+
+func newFakeClient(t *testing.T, obj ...runtime.Object) client.Client {
+	scheme, err := api.SchemeBuilder.Build()
+	assert.NoError(t, err, "unable to build scheme")
+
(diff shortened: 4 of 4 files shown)


services/provider/client/auth_interceptor.go
```diff
@@ -0,0 +1,41 @@
...
+	"context"
+
+	"github.com/red-hat-storage/ocs-operator/services/provider/common"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+	"k8s.io/klog"
+)
+
+type AuthInterceptor struct {
+	accessToken string
```
oritwas, line 13: Let's use a byte array with max length

jarrpa: +1

services/provider/client/auth_interceptor.go
```diff
@@ -0,0 +1,41 @@
...
+func (interceptor *AuthInterceptor) Unary() grpc.UnaryClientInterceptor {
+	return func(
+		ctx context.Context,
+		method string,
+		req, reply interface{},
+		cc *grpc.ClientConn,
+		invoker grpc.UnaryInvoker,
+		opts ...grpc.CallOption,
+	) error {
+		klog.Infof("--> unary interceptor: %q", method)
```
jarrpa, line 29: Do we need the `-->` in the error message?

services/provider/common/common.go
```diff
@@ -0,0 +1,36 @@
...
+
+const ProviderServicePath = "/provider.OCSProvider/"
+
+func TokenAuthMethods() map[string]bool {
+	return map[string]bool{
+		ProviderServicePath + "OnBoardConsumer": true,
+	}
+}
+
+func GetMetaData(ctx context.Context, key string) (string, error) {
```
jarrpa, line 27: Just a nit, but `GetMetadata` would feel more intuitive.

services/provider/client/client.go
```diff
@@ -24,17 +24,33 @@ func NewConsumerClient(cc *grpc.ClientConn, timeout time.Duration) *ConsumerClie
 func NewGRPCConnection(serverAddr string, opts []grpc.DialOption) (*grpc.ClientConn, error) {
 	conn, err := grpc.Dial(serverAddr, opts...)
 	if err != nil {
-		klog.Fatalf("failed to dial: %v", err)
+		return nil, fmt.Errorf("failed to dial: %v", err)
```
jarrpa, line 27: Probably include this in the previous commit? Unless I'm missing something at this code was not introduced in this PR...

services/provider/client/client.go
```diff
@@ -24,17 +24,33 @@ func NewConsumerClient(cc *grpc.ClientConn, timeout time.Duration) *ConsumerClie
...
+
 // OnBoardConsumer to validate the consumer and create StorageConsumer
 // resource on the StorageProvider cluster
-func (cc *ConsumerClient) OnBoardConsumer(token, capacity string) (*pb.OnBoardConsumerResponse, error) {
+func (cc *ConsumerClient) OnBoardConsumer(name, capacity string) (*pb.OnBoardConsumerResponse, error) {
 	req := &pb.OnBoardConsumerRequest{
-		Token:    token,
-		Capacity: capacity,
+		ConsumerName: name,
+		Capacity:     capacity,
```
jarrpa, line 53: Probably move this to the first commit where there's related changes.

services/provider/server/server.go
```diff
@@ -5,43 +5,202 @@ import (
...
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/reflection"
+	"google.golang.org/grpc/status"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/klog"
 )
 
-type ocsProviderServer struct {
+type OCSProviderServer struct {
```
jarrpa, line 21: Why is this being done? I don't see it being used outside this package, so starting with lowercase would be more conventional.

jarrpa (changes requested):

services/provider/client/client.go
sp98: Right. Moved it to the first commit where proto file was updated.

services/provider/client/auth_interceptor.go
sp98: removed.

services/provider/server/server.go
sp98: linter not happy if I do this. 
```
Running golangci-lint
services/provider/server/server.go:28:93: exported func NewOCSProviderServer returns unexported type *github.com/red-hat-storage/ocs-operator/services/provider/server.ocsProviderServer, which can be annoying to use (golint)
func NewOCSProviderServer(cClient consumer.APIUpdater, auth *AuthManager, namespace string) *ocsProviderServer {
```

services/provider/client/client.go
sp98: This could be in the same PR.  I added the `Initialize` method below and realized that `NewGRPCConnection` is not returning error correctly.  `NewGRPCConnection` is only used by `Initialize` method.

services/provider/server/server.go
nb-ohad: You cannot return a "private" (unexported type) from a "public" (exported) function
You either export both or none.

services/provider/client/client.go
```diff
@@ -31,10 +31,10 @@ func NewGRPCConnection(serverAddr string, opts []grpc.DialOption) (*grpc.ClientC
 
 // OnBoardConsumer to validate the consumer and create StorageConsumer
 // resource on the StorageProvider cluster
-func (cc *ConsumerClient) OnBoardConsumer(token, capacity string) (*pb.OnBoardConsumerResponse, error) {
+func (cc *ConsumerClient) OnBoardConsumer(name, capacity string) (*pb.OnBoardConsumerResponse, error) {
```
nb-ohad, line 50: `OnBoardConsumer` -> `OnboardConsumer`

Onboard is a single word (according to google search)

services/provider/server/consumer/api_updater.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	apiClient := &cAPIUpdater{
+		client: crClient,
+	}
+	return apiClient, nil
+}
+
+func (c cAPIUpdater) CreateStorageConsumer(sc *ocsv1alpha1.StorageConsumer) error {
+	return c.client.Create(context.TODO(), sc)
```
nb-ohad, line 59: Why are we using context.TODO() and not not a single context.Background(), there is not need to switch to a new context for every client invocation.

I would suggest using a single context instance per incoming API request processed by the server

services/provider/server/consumer/store.go
```diff
@@ -0,0 +1,68 @@
...
+	"errors"
+	"sync"
+)
+
+// Consumer is the storage consumer
+type Consumer struct {
+	Name  string
+	ID    string
+	Token string
+}
```
nb-ohad, line 13: Is there a reason why we use a local struct here instead of the struct representing the StorageConsumer CR?

services/provider/server/server.go
```diff
@@ -5,43 +5,202 @@ import (
...
+	}
+}
+
+// GenerateToken RPC call to generate a new jwt token for the consumer cluster
+func (s *OCSProviderServer) GenerateToken(ctx context.Context, req *pb.GenerateTokenRequest) (*pb.GenerateTokenResponse, error) {
+	token, err := s.authManager.GetAccessToken()
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to generate access token: %v", err)
+	}
+	return &pb.GenerateTokenResponse{AccessToken: token}, nil
```
nb-ohad, line 42: Why do we have this? We should not provide a public endpoint for token generation

services/provider/server/server.go
```diff
@@ -5,43 +5,202 @@ import (
...
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to validate storage consumer: %v", err)
+	}
+
+	// Delete storageConsumer resource no matter the current state of the resource
+	err = s.consumerAPI.DeleteStorageConsumer(consumerResource)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to delete storage consumer resource: %v", err)
+	}
+
```
nb-ohad, line 114: Where are we going to protect again deletion of StorageConsumers that are still consuming storage (have claims on storage)?

kesavanvt: Should we look for the available PVCs (consuming ocs storage) in the consumer and send the offboarding request only if not exists?

nb-ohad: No, the source of truth should always be the provider. 
@leseb is there anything in rook-ceph we can query to get that information (for both RBD and CephFS)

sp98: Consumer Storage Cluster is deleted ->  Consumer storage cluster deletes the resources its using (RBD/Ceph FS) -> Only after successful deletion of those resources the consumer OCS cluster will call the Offboarding API to delete the StoageConsumer CR . 

This ^^ is one possible approach that we discussed today.

nb-ohad: @sp98 The API server is not an internal component, it is a public endpoint, and as such cannot assume the correct "order of operations" is enforced from the outside. 

It is true that for the "happy path" the strategy you describe can work, but what about bugs? what about attackers? 
The risk here is loss of customer data and as such we must protect against such situations.

services/provider/main.go
```diff
@@ -13,6 +17,58 @@ var (
...
 func main() {
 	flag.Parse()
-	var opts []grpc.ServerOption
-	server.Start(*port, opts)
+
+	klog.Info("Starting Provider API server")
+
+	namespace, err := util.GetWatchNamespace()
+	if err != nil {
+		klog.Errorf("failed to get provider cluster namespace. Error %v", err)
```
leseb, line 25:
```suggestion
		klog.Errorf("failed to get provider cluster namespace. %v", err)
```

The logger is already configured to log Error level so we don't need to explicitly add this.

services/provider/main.go
```diff
@@ -13,6 +17,58 @@ var (
...
+		return
+	}
+
+	consumerAPI, err := consumer.NewConsumerAPIUpdater()
+	if err != nil {
+		klog.Errorf("failed to initialize consumer client. Error %v", err)
```
leseb, line 31: ditto

services/provider/main.go
```diff
@@ -13,6 +17,58 @@ var (
...
+		return
+	}
+
+	// Get master key to encryt/decryt token
+	// TODO: Remove hard coded value for secret name
+	secret, err := consumerAPI.GetSecret("ocs-provider-server", namespace)
+	if err != nil {
+		klog.Errorf("failed to get secret to read the master key. Error %v", err)
```
leseb, line 39: ditto

services/provider/main.go
```diff
@@ -13,6 +17,58 @@ var (
...
+	if masterKey == "" {
+		klog.Errorf("master key is empty")
+		return
+	}
+
+	// Update in-memory user store
+	storageConsumers, err := consumerAPI.ListStorageConsumers(namespace)
+	consumerStore := consumer.NewConsumerStore()
+	if err != nil {
+		klog.Errorf("failed to list storageConsumer resources in %q namespace. Error %v", namespace, err)
```
leseb, line 53: It's an error but the we don't return?

services/provider/main.go
```diff
@@ -13,6 +17,58 @@ var (
...
+	}
+
+	for _, sc := range storageConsumers.Items {
+		newConsumer := consumer.NewConsumer(sc.Name, string(sc.UID), sc.Annotations[common.TokenAnnotation])
+		err := consumerStore.Save(newConsumer)
+		if err != nil {
+			klog.Errorf("failed to intialize in-memory consumer store. Error %v", err)
```
leseb, line 60:
```suggestion
			klog.Errorf("failed to initialize in-memory consumer store. Error %v", err)
```

Is the `in-memory` message relevant?

services/provider/server/auth_interceptor.go
```diff
@@ -0,0 +1,68 @@
...
+
+	return nil
+}
+
+// authorizeToken validates the access token
+func (a *AuthInterceptor) authorizeToken(ctx context.Context) error {
+
+	accessToken, err := common.GetMetadata(ctx, "authorization")
+	if err != nil {
+		return status.Errorf(codes.Unauthenticated, "authorization token is not provided")
```
leseb, line 55: Why not add the `err` in the return?

services/provider/server/consumer/api_updater.go
leseb: Agreed, we need a parent context we pass down to all the callers. Also using a context with cancellation is preferred.

services/provider/server/server.go
leseb: From the Rook perspective, once the CephBlockPool and SubVolumeGroup are in deleting phase, the finalizers will only be released once no RBD image and no SubVolume are present anymore. Until this is the case, the deletion of these CRDs will be blocked. Today, the failure is not reflected in the CR status, but it's quite easy to do. I'll add this.

nb-ohad: @leseb 
I think the CR status is too late as the consumer will get an ack on the delete (as we agreed that it is an async process). We need to fail the delete request before we even change the storageconsumer CR

services/provider/main.go
nb-ohad: maybe failed to initialize local consumer cache? 

As a side note, I think the struct is miss-named and misleading.  
A store usually holds the source of truth for its items while this struct acts more like a cache, so maybe ConsumerCache would be preferable (for both the type and the message)

services/provider/client/client.go
sp98: using `onboard` and `offboard`

services/provider/server/consumer/api_updater.go
sp98: passing the parent context provided by the client. It uses cancel with timeout.

services/provider/server/consumer/store.go
sp98: Just a in-memory store (cache) to track all the consumers and tokens used by them.

services/provider/main.go
sp98: done.

services/provider/main.go
sp98: done

services/provider/main.go
sp98: returning err in the message now.

services/provider/main.go
sp98: using `cache` instead of `store`.

services/provider/main.go
sp98: done

services/provider/main.go
```diff
@@ -13,6 +17,62 @@ var (
 
 func main() {
 	flag.Parse()
-	var opts []grpc.ServerOption
-	server.Start(*port, opts)
+
+	klog.Info("Starting Provider API server")
+
+	ctx := context.Background()
+	ctx, cancel := context.WithCancel(ctx)
```
leseb, line 24: Use `signal.NotifyContext()` instead so that any requests can be canceled if the container main process receives a termination signal. Right now, only the end of execution will terminate this, so it's more or less non-graceful.

services/provider/client/client.go
```diff
@@ -24,23 +24,39 @@ func NewConsumerClient(cc *grpc.ClientConn, timeout time.Duration) *ConsumerClie
...
-// OnBoardConsumer to validate the consumer and create StorageConsumer
+// Initialize initialzes a new provider client
+func Initialize(serverAddr, token string, reqTimeOut time.Duration) (*ConsumerClient, error) {
+	interceptor := NewAuthInterceptor(token)
+
+	// TODO: ADD TLS support
+	opts := []grpc.DialOption{
+		grpc.WithUnaryInterceptor(interceptor.Unary()),
+	}
+	cc, err := NewGRPCConnection(serverAddr, opts)
```
iamniting, line 40: can we also implement the logic to close the connection

services/provider/server/server.go
```diff
@@ -5,43 +5,202 @@ import (
...
+	}
+	return &pb.GenerateTokenResponse{AccessToken: token}, nil
+}
+
+// OnboardConsumer RPC call to onboard a new OCS consumer cluster.
+func (s *OCSProviderServer) OnboardConsumer(ctx context.Context, req *pb.OnboardConsumerRequest) (*pb.OnboardConsumerResponse, error) {
+	// validate capacity
+	capacity, err := resource.ParseQuantity(req.Capacity)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "%q is not a valid resource capacity: %v", req.Capacity, err)
```
iamniting, line 49: I think this should not be an codes.Internal error code

iamniting (changes requested):

services/provider/client/client.go
sp98: I realised that it won't be easy to add close() using  current approach.  So moved back to original changes. You use and close the connection like this : 
```
conn, err := client.NewGRPCConnection("address", "token")

cc := NewConsumerClient(conn, <timeout>)
// use cc

// close cc
cc.Close()
```

services/provider/server/server.go
sp98: Right.  Passing `InvalidArgument` error code.

services/provider/main.go
sp98: Thanks for the suggestion. Using `signal.NotifyContext()`.

services/provider/common/common.go
sp98: done.

services/provider/server/server.go
sp98: marking as resolved.

services/provider/client/client.go
```diff
@@ -2,49 +2,60 @@ package client
...
 	"time"
 
+	"github.com/red-hat-storage/ocs-operator/services/provider/common"
 	pb "github.com/red-hat-storage/ocs-operator/services/provider/pb"
 	"google.golang.org/grpc"
-	"k8s.io/klog"
 )
 
-type ConsumerClient struct {
+type OCSProviderClient struct {
```
oritwas, line 13: Why the rename?  The provider Client is a consumer.
I would have services/provider/server.go and services/consumer/client.go
another option is services/provider/provider.go and services/provider/consumer.go

services/provider/client/client.go
```diff
@@ -2,49 +2,60 @@ package client
...
 	Client  pb.OCSProviderClient
 	timeout time.Duration
 }
```
oritwas, line 16: Add the grpc connection here

nb-ohad (changes requested):

services/provider/server/server.go
```diff
@@ -6,54 +6,206 @@ import (
...
+
+	// Decrypt storage consumer resource UUID
+
+	// Get storage consumer resource using UUID
+	consumerResource, err := s.getStorageConsumerWithUUID(ctx, "uuid")
+	if err != nil {
+		return nil, err
+	}
+	err = s.client.Delete(ctx, consumerResource)
+	if err != nil {
```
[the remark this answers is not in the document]
sp98: done

services/provider/server/server.go
```diff
@@ -6,54 +6,206 @@ import (
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+			Annotations: map[string]string{
+				TokenAnnotation: ticket,
+			},
+		},
+		Spec: ocsv1alpha1.StorageConsumerSpec{
+			Capacity: capacity,
+		},
```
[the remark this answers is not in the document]
sp98: done.

services/provider/server/server.go
```diff
@@ -6,54 +6,206 @@ import (
...
 	mock := os.Getenv(common.MockProviderAPI)
 	if mock != "" {
 		return mockGetStorageConfig(common.MockError(mock))
 	}
-	return &pb.StorageConfigResponse{}, nil
+
+	// Decrypt storage consumer resource UUID
+
+	// Get storage consumer resource using UUID
+	consumerResource, err := s.getStorageConsumerWithUUID(ctx, "uuid")
```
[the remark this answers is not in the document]
sp98: fixed.

services/provider/server/server.go
```diff
@@ -6,54 +6,206 @@ import (
...
+	if mock != "" {
+		return mockUpdateCapacity(common.MockError(mock))
+	}
+	capacity, err := resource.ParseQuantity(req.Capacity)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "%q is not a valid resource capacity: %v", req.Capacity, err)
+	}
+
+	// Get storage consumer resource using UUID
+	consumerResource, err := s.getStorageConsumerWithUUID(ctx, "uuid")
```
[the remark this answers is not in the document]
sp98: fixed.

services/provider/server/server.go
```diff
@@ -6,54 +6,206 @@ import (
...
+func (s *ocsProviderServer) OffboardConsumer(ctx context.Context, req *pb.OffboardConsumerRequest) (*pb.OffboardConsumerResponse, error) {
 	mock := os.Getenv(common.MockProviderAPI)
 	if mock != "" {
 		return mockOffboardConsumer(common.MockError(mock))
 	}
+
+	// Decrypt storage consumer resource UUID
+
+	// Get storage consumer resource using UUID
+	consumerResource, err := s.getStorageConsumerWithUUID(ctx, "uuid")
```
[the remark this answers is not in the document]
sp98: fixed.

services/provider/server/server.go
```diff
@@ -6,54 +6,206 @@ import (
...
+	}
+
+	// Get storage consumer resource using UUID
+	consumerResource, err := s.getStorageConsumerWithUUID(ctx, "uuid")
+	if err != nil {
+		return nil, err
+	}
+
+	consumerResource.Spec.Capacity = capacity
+	err = s.client.Update(ctx, consumerResource)
```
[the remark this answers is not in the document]
sp98: doing that now.

services/provider/server/server.go
```diff
@@ -6,54 +6,206 @@ import (
...
+		return nil, err
+	}
+	client, err := client.New(config, client.Options{Scheme: scheme})
+	if err != nil {
+		klog.Error(err, "failed to create controller-runtime client")
+		return nil, err
+	}
+
+	return client, nil
+}
```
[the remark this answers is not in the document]
sp98: not exporting this function anymore.

services/provider/server/consumer.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	err := c.client.Create(ctx, consumerObj)
+	if err != nil {
+		if errors.IsAlreadyExists(err) {
+			return "", err
+		}
+		return "", fmt.Errorf("failed to create storageConsumer resource %q. %v", consumerObj.Name, err)
+	}
+
```
sp98, line 73: I need to first call `create` to create the object and then `get` to return the `UUID` of the object. Wondering if this is something that can be done using a single API call.

services/provider/server/server.go
```diff
@@ -6,65 +6,214 @@ import (
...
+		return nil, fmt.Errorf("failed to create new client. %v", err)
+	}
+
+	consumer, err := NewOCSConsumer(ctx, client)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create new OCSConumer instance. %v", err)
+	}
+
+	return &ocsProviderServer{
+		client:    client,
```
sp98, line 49: This `client` is not used now. But will be needed when we call rook resources to generate the connection info.

services/provider/server/server.go
```diff
@@ -6,65 +6,214 @@ import (
...
+	capacity, err := resource.ParseQuantity(req.Capacity)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "%q is not a valid storageConsumer capacity: %v", req.Capacity, err)
+	}
+
+	// Validate onboardingTicket
+	isTicketUsed := s.consumer.IsTicketUsed(req.OnboardingTicket)
+	if isTicketUsed {
+		return nil, status.Errorf(codes.InvalidArgument, "onboarding ticket is already used by an existing storageConsumer resource")
+	}
```
nb-ohad, line 72: There should be a couple of steps here:
1. Split the ticket into the encrypted and non-encrypted section
1. Validate the ticket using the provided key (and the two sections) 
2. Unmarsh the non-encrypted ticket into a struct (to be provided by @jarrpa)
3. Check the expiry field to make sure the ticket is not expired 
4. Validate that the ticket is not in use - would suggest saving only the encrypted side on the annotation or the sig field

You can encapsulate the logic in a function, something like `validateTicket`

services/provider/server/server.go
```diff
@@ -6,65 +6,214 @@ import (
...
 }
 
 // GetStorageConfig RPC call to onboard a new OCS consumer cluster.
-func (c *ocsProviderServer) GetStorageConfig(ctx context.Context, req *pb.StorageConfigRequest) (*pb.StorageConfigResponse, error) {
+func (s *ocsProviderServer) GetStorageConfig(ctx context.Context, req *pb.StorageConfigRequest) (*pb.StorageConfigResponse, error) {
 	mock := os.Getenv(common.MockProviderAPI)
 	if mock != "" {
 		return mockGetStorageConfig(common.MockError(mock))
 	}
-	return &pb.StorageConfigResponse{}, nil
```
nb-ohad, line 35: Is there a reason we are missing the implementation for this?

services/provider/server/server.go
```diff
@@ -6,65 +6,214 @@ import (
...
+	consumerObj, err := s.consumer.GetWithID(ctx, s.namespace, req.StorageConsumerUUID)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to get storageConsumer resource with provided UID: %v", err)
+	}
+
+	consumerObj.Spec.Capacity = capacity
+	err = s.consumer.Update(ctx, consumerObj)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to update capacity in the storageConsumer resource: %v", err)
+	}
```
nb-ohad, line 157: See comment on update function definition

nb-ohad (changes requested):

services/provider/server/consumer.go
```diff
@@ -0,0 +1,157 @@
...
+	ocsv1alpha1 "github.com/red-hat-storage/ocs-operator/api/v1alpha1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/klog"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type consumerCache struct {
```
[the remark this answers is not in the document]
nb-ohad: Naming is important as it can help avoid misunderstanding around usage and improve readability. , 
Why use the store if you do not hold any actual entities in the store?

Don't you think something like `consumerById` and `consumerByTicket` fields will be better?

services/provider/server/consumer.go
```diff
@@ -0,0 +1,157 @@
...
+		if _, ok := cache[uid]; !ok {
+			cache[uid] = consumerCache{consumer.Name, uid, ticket}
+		}
+	}
+
+	return cache, nil
+}
+
+// Create creates a new storageConsumer resource, updates the consumer cache and returns the storageConsumer UUID
+func (c *ocsConsumer) Create(ctx context.Context, consumerObj *ocsv1alpha1.StorageConsumer) (string, error) {
```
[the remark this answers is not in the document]
nb-ohad: >  Right now, its looks good but in future if there is more information needed to create the
We should not try to resolve future code organization problems, when we get to that point we will open the discussion again. 

Please note that for public interfaces (like CRs or API endpoints - not code interfaces) we cannot take that approach and think about an arrangement that is open to future changes is important - but it is not the case here.

services/provider/server/consumer.go
```diff
@@ -0,0 +1,157 @@
...
+		return "", fmt.Errorf("failed to create storageConsumer resource %q. %v", consumerObj.Name, err)
+	}
+
+	existingConsumer := &ocsv1alpha1.StorageConsumer{}
+	err = c.client.Get(ctx, types.NamespacedName{Name: consumerObj.Name, Namespace: consumerObj.Namespace}, existingConsumer)
+	if err != nil {
+		return "", fmt.Errorf("failed to get UUID from the storageConsumer resource %q. %v", consumerObj.Name, err)
+	}
+
+	consumerID := string(existingConsumer.UID)
```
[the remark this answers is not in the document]
nb-ohad: So in that case, why are we using a string on our code and not `types.UID`?

services/provider/server/consumer.go
```diff
@@ -0,0 +1,157 @@
...
+	consumerObj := &ocsv1alpha1.StorageConsumer{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      consumer.name,
+			Namespace: namespace,
+		},
+	}
+	err := c.client.Delete(ctx, consumerObj)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			klog.Warningf("storageConsumer %q not found.", consumerObj.Name)
```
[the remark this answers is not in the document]
nb-ohad: OK, but I think you meant UID (unique identifier)  and not UUID (universally unique identifier), there is no guarantee that these IDs are universally unique

services/provider/server/consumer.go
```diff
@@ -0,0 +1,157 @@
...
+	delete(c.cache, id)
+
+	return nil
+}
+
+// Update updates the storageConsumer
+func (c *ocsConsumer) Update(ctx context.Context, consumerObj *ocsv1alpha1.StorageConsumer) error {
+	err := c.client.Update(ctx, consumerObj)
+	if err != nil {
+		return fmt.Errorf("failed to update storageConfig resource %q. %v", consumerObj.Name, err)
```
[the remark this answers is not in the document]
nb-ohad: OK, but I think you meant UID (unique identifier) and not UUID (universally unique identifier), there is no guarantee that these IDs are universally unique

services/provider/server/consumer.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+// IsTicketUsed checks if the ticket is already being used by another storageConsumer
+func (c *ocsConsumer) IsTicketUsed(ticket string) bool {
+	for _, consumer := range c.cache {
+		if consumer.ticket == ticket {
+			return true
+		}
+	}
+
```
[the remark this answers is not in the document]
nb-ohad: Why do you need to delete it from the ticket store at all? We shouldn't let a ticket be reused even if the consumer was offboarded. I understand that if the pod gets restarted we lose that information, but to me it is acceptable.

services/provider/server/server.go
nb-ohad: Ok

services/provider/server/consumer.go
```diff
@@ -0,0 +1,157 @@
...
+		return "", fmt.Errorf("can't create storageConsumer resource without onboarding ticket annotation")
+	}
+
+	err := c.client.Create(ctx, consumerObj)
+	if err != nil {
+		if errors.IsAlreadyExists(err) {
+			return "", err
+		}
+		return "", fmt.Errorf("failed to create storageConsumer resource %q. %v", consumerObj.Name, err)
+	}
```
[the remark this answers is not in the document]
nb-ohad: Please note if you are accepting or rejecting the suggestion, I will resolve after.

services/provider/server/consumer.go
```diff
@@ -0,0 +1,157 @@
...
+
+	existingConsumer := &ocsv1alpha1.StorageConsumer{}
+	err = c.client.Get(ctx, types.NamespacedName{Name: consumerObj.Name, Namespace: consumerObj.Namespace}, existingConsumer)
+	if err != nil {
+		return "", fmt.Errorf("failed to get UUID from the storageConsumer resource %q. %v", consumerObj.Name, err)
+	}
```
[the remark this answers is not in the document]
nb-ohad: Ok, I see, you need this to get the UID. Are we sure that after calling create/update the resource is not already getting updated? I am asking because I remember that from a k8s API standpoint (not operators sdk standpoint) both create and update return the updated CR

services/provider/server/consumer.go
nb-ohad: See my comment below

services/provider/server/consumer.go
```diff
@@ -0,0 +1,158 @@
...
+// Delete deletes the storageConsumer resource using UID and updates the consumer cache
+func (c *ocsConsumerManager) Delete(ctx context.Context, namespace, id string) error {
+	c.mutex.RLock()
+	consumerName, ok := c.uidStore[id]
+	if !ok {
+		klog.Warningf("no storageConsumer found with UUID %q", id)
+		c.mutex.RUnlock()
+		return nil
+	}
+	c.mutex.RUnlock()
```
nb-ohad, line 105: Why do we need to lock on get operations?

services/provider/server/consumer.go
```diff
@@ -0,0 +1,158 @@
...
+	if err := c.client.Delete(ctx, consumerObj); err != nil {
+		if errors.IsNotFound(err) {
+			klog.Warningf("storageConsumer %q not found.", consumerObj.Name)
+			return nil
+		}
+		return fmt.Errorf("failed to delete storageConsumer %q. %v", consumerName, err)
+	}
+
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
```
nb-ohad, line 122: You would like to remove the lock as soon as possible after the index update, do not defer the operation.
You do not need to handle erroneous states because `delete(map, value)` will not fail if the key does not exist.

services/provider/server/consumer.go
```diff
@@ -0,0 +1,158 @@
...
+	consumerID := string(existingConsumer.UID)
+	if _, ok := c.uidStore[consumerID]; !ok {
+		c.uidStore[consumerID] = existingConsumer.Name
+	}
+	if _, ok := c.ticketStore[ticket]; !ok {
+		c.ticketStore[ticket] = existingConsumer.Name
+	}
```
nb-ohad, line 91: No need to check for existence if you are not going to handle the case where it already exists (which I am not saying you should)
```suggestion
	c.uidStore[consumerID] = existingConsumer.Name
	c.ticketStore[ticket] = existingConsumer.Name
```

services/provider/server/consumer.go
```diff
@@ -0,0 +1,158 @@
...
+
+	existingConsumer := &ocsv1alpha1.StorageConsumer{}
+	err = c.client.Get(ctx, types.NamespacedName{Name: consumerObj.Name, Namespace: consumerObj.Namespace}, existingConsumer)
+	if err != nil {
+		return "", fmt.Errorf("failed to get UUID from the storageConsumer resource %q. %v", consumerObj.Name, err)
+	}
+
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
```
nb-ohad, line 84: Do not defer unlocks, we need the lock to be removed as soon as possible. 
In any case map insert operations cannot fail

services/provider/server/consumer.go
```diff
@@ -0,0 +1,158 @@
...
+// Get returns a storageConsumer resource using the UID
+func (c *ocsConsumerManager) Get(ctx context.Context, namespace, id string) (*ocsv1alpha1.StorageConsumer, error) {
+	c.mutex.RLock()
+	consumerName, ok := c.uidStore[id]
+	if !ok {
+		c.mutex.RUnlock()
+		klog.Errorf("no storageConsumer found with the provided UUID %q", id)
+		return nil, fmt.Errorf("no storageConsumer found with the provided UUID %q", id)
+	}
+	c.mutex.RUnlock()
```
nb-ohad, line 149: No need to lock for read operation

services/provider/server/consumer.go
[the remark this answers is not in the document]
sp98: I'm getting the UID as a string from the user.  If I use the map key as `types.UID`, then I have to do `types.UID(id)` every time when I'm trying to get the customer name using the UID.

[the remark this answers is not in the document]
nb-ohad: Ok, from your answer I assume that we cannot use the UID type in the grpc req definitions. 
Please confirm

services/provider/server/consumer.go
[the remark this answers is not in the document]
sp98: using `nameByUID` and `nameByTicket`.

services/provider/server/consumer.go
sp98: discussed on the gchat that we need lock for read operations.

services/provider/server/consumer.go
sp98: done.

services/provider/server/consumer.go
sp98: done.

services/provider/server/consumer.go
sp98: done.

services/provider/server/consumer.go
sp98: that works. Didn't know that. Call the `create` updates the object as well. So need to call `get` again.

nb-ohad (changes requested):

services/provider/server/consumer.go
```diff
@@ -0,0 +1,179 @@
...
+
+	for _, consumer := range consumers.Items {
+		if _, ok := nameByUID[consumer.UID]; !ok {
+			nameByUID[consumer.UID] = consumer.Name
+		}
+
+		if ticket, ok := consumer.GetAnnotations()[TicketAnnotation]; ok {
+			if _, ok := nameByTicket[ticket]; !ok {
+				nameByTicket[ticket] = consumer.Name
+			}
```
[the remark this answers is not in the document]
sp98: done.

services/provider/server/consumer.go
```diff
@@ -0,0 +1,179 @@
...
+		return nil, fmt.Errorf("failed to list storage consumers. %v", err)
+	}
+
+	nameByTicket := map[string]string{}
+	nameByUID := map[types.UID]string{}
+
+	for _, consumer := range consumers.Items {
+		if _, ok := nameByUID[consumer.UID]; !ok {
+			nameByUID[consumer.UID] = consumer.Name
+		}
```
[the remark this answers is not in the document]
sp98: done.

services/provider/server/consumer.go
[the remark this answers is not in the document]
sp98: accepted this suggestion.

services/provider/server/server.go
```diff
@@ -6,65 +6,185 @@ import (
...
+	return nil, status.Errorf(codes.Unavailable, "storage consumer status is not set")
 }
 
 // UpdateCapacity PRC call to increase or decrease the storage pool size
-func (c *ocsProviderServer) UpdateCapacity(ctx context.Context, req *pb.UpdateCapacityRequest) (*pb.UpdateCapacityResponse, error) {
+func (s *OCSProviderServer) UpdateCapacity(ctx context.Context, req *pb.UpdateCapacityRequest) (*pb.UpdateCapacityResponse, error) {
 	mock := os.Getenv(common.MockProviderAPI)
 	if mock != "" {
 		return mockUpdateCapacity(common.MockError(mock))
 	}
```
iamniting, line 120: should not we delete this?

services/provider/server/server.go
```diff
@@ -6,65 +6,185 @@ import (
...
+	capacity, err := resource.ParseQuantity(req.Capacity)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "%q is not a valid resource capacity: %v", req.Capacity, err)
+	}
+
+	err = s.consumerManager.UpdateCapacity(ctx, req.StorageConsumerUUID, capacity)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to update capacity in the storageConsumer resource: %v", err)
+	}
+
```
iamniting, line 131: I think it should also take care of if CR not found and set appropriate code accordingly

iamniting (changes requested):

services/provider/server/consumer.go
```diff
@@ -0,0 +1,175 @@
...
+		nameByUID:    nameByUID,
+	}, nil
+}
+
+// Create creates a new storageConsumer resource, updates the consumer cache and returns the storageConsumer UID
+func (c *ocsConsumerManager) Create(ctx context.Context, name, ticket string, capacity resource.Quantity) (string, error) {
+	c.mutex.RLock()
+	if _, ok := c.nameByTicket[ticket]; ok {
+		c.mutex.RUnlock()
+		return "", fmt.Errorf("onboarding ticket is already being used by another storageConsumer")
```
iamniting, line 56: Can we set appropriate error code here which is `codes.AlreadyExists`?

iamniting (changes requested):

services/provider/server/server.go
sp98: not deleting the mocks as of now.

services/provider/server/consumer.go
nb-ohad: @iamniting This is the consumer manager code, not the server code. 
I don't think we should introduce JRPC codes here

iamniting: In that case we should read this error and set the grpc code where we are calling this func.

services/provider/server/server.go
sp98: sending `NotFound` status code if storageConsumer is not found while updating capacity

services/provider/server/consumer.go
sp98: Sending `AlreadyExists` status code when onboarding ticket is already being used.

iamniting (approved):

nb-ohad (approved):

services/provider/client/auth_interceptor.go
sp98: No onboarding ticket changes in this PR. This comment is not valid anymore. Marking as resolved.

services/provider/client/client.go
sp98: This comment is obsolete.  We have only two commits now. One for implementation and other for unit tests.  Marking this as resolved.

PR #1433, opened 2021-12-14
Decreasing the timeout of 'ODF_standardized_metrics.rules' group to 30s

aruniiird:
In order to make the 'odf_system_iops_total_bytes' recording rule to
collect more granular metric data points, we are decreasing the
recording rule group's time interval to 30s (where default is 1m).


umangachapagain (changes requested):
I believe you also need to add the generated changes.

jarrpa (approved):
/lgtm

PR #1435, opened 2021-12-15
ocs-to-ocs: Add required env var in the API server deployment

iamniting:
STORAGE_CLUSTER_NAME and STORAGE_CLUSTER_UID is required for the garbage
collection of the storage consumer.

### DIFF
--- a/controllers/storagecluster/provider_server.go
+++ b/controllers/storagecluster/provider_server.go
@@ -16,6 +16,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v1"
+	"github.com/red-hat-storage/ocs-operator/controllers/util"
 )
 
 const (
@@ -85,8 +86,16 @@ func (o *ocsProviderServer) ensureDeleted(r *StorageClusterReconciler, instance 
 
 func (o *ocsProviderServer) createDeployment(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) error {
 
-	if _, ok := os.LookupEnv(providerAPIServerImage); !ok {
-		return fmt.Errorf("ENV var %s not found", providerAPIServerImage)
+	var finalErr error
+
+	for _, env := range []string{providerAPIServerImage, util.WatchNamespaceEnvVar} {
+		if _, ok := os.LookupEnv(env); !ok {
+			multierr.AppendInto(&finalErr, fmt.Errorf("ENV var %s not found", env))
+		}
+	}
+
+	if finalErr != nil {
+		return finalErr
 	}
 
 	desiredDeployment := GetProviderAPIServerDeployment(instance)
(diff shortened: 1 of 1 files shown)


controllers/storagecluster/provider_server.go
```diff
@@ -202,6 +202,12 @@ func GetProviderAPIServerDeployment(instance *ocsv1.StorageCluster) *appsv1.Depl
 							Name:    "ocs-provider-api-server",
 							Image:   os.Getenv(providerAPIServerImage),
 							Command: []string{"/usr/local/bin/provider-api"},
+							Env: []corev1.EnvVar{
+								{
+									Name:  "WATCH_NAMESPACE",
+									Value: os.Getenv("WATCH_NAMESPACE"),
```
sp98, line 208: There is a `GetWatchNamespace` method to get this namespace. It has error validations as well. Suggesting to use it here to get the `WATCH_NAMESPACE` value.

sp98 (changes requested):

controllers/storagecluster/provider_server.go
iamniting: I am doing the same as a func for the other env, added this also to the list.

sp98: so this approach ends up fetching the env variable data twice.  Once in `LookupEnv` and again in `Getenv`. Suggesting to have a single call to fetch the env variables.

sp98 (changes requested):

controllers/storagecluster/provider_server.go
iamniting: Actually, I am doing that for a reason `createDeployment` func returns an error where I am checking it for the first time and returning an error if ENV does not exist. `GetProviderAPIServerDeployment` func does not return the error it just returns deployment only that's why not checking in that and I do not want to send ENVS as the argument.

sp98: ok.

controllers/storagecluster/provider_server.go
```diff
@@ -202,6 +210,20 @@ func GetProviderAPIServerDeployment(instance *ocsv1.StorageCluster) *appsv1.Depl
 							Name:    "ocs-provider-api-server",
 							Image:   os.Getenv(providerAPIServerImage),
 							Command: []string{"/usr/local/bin/provider-api"},
+							Env: []corev1.EnvVar{
+								{
+									Name:  "WATCH_NAMESPACE",
```
sp98, line 215:
```suggestion
									Name:  util.WatchNamespaceEnvVar,
```

controllers/storagecluster/provider_server.go
```diff
@@ -202,6 +210,20 @@ func GetProviderAPIServerDeployment(instance *ocsv1.StorageCluster) *appsv1.Depl
...
+									Value: os.Getenv("WATCH_NAMESPACE"),
```
sp98, line 216:
```suggestion
									Value: os.Getenv(util.WatchNamespaceEnvVar),
```

controllers/storagecluster/provider_server.go
```diff
@@ -85,8 +85,16 @@ func (o *ocsProviderServer) ensureDeleted(r *StorageClusterReconciler, instance
 
 func (o *ocsProviderServer) createDeployment(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) error {
 
-	if _, ok := os.LookupEnv(providerAPIServerImage); !ok {
-		return fmt.Errorf("ENV var %s not found", providerAPIServerImage)
+	var finalErr error
+
+	for _, env := range []string{providerAPIServerImage, "WATCH_NAMESPACE"} {
```
sp98, line 90:
```suggestion
	for _, env := range []string{providerAPIServerImage, util.WatchNamespaceEnvVar} {
```

sp98 (changes requested):
nit

controllers/storagecluster/provider_server.go

controllers/storagecluster/provider_server.go

sp98 (approved):
/lgtm

agarwal-mudit (approved):
/lgtm

Issue #1436, opened 2021-12-15
storageProviderKind field takes default value even with internal mode StorageCluster installation 


kesavanvt:
When StorageCluster is installed in internal mode, storageProviderKind takes default value as rhcs

```
spec:
  externalStorage:
      storageProviderKind: rhcs
```

Expected behavior:
it should not be assigned with "rhcs" when StorageCluster is installed in internal mode

PR #1438, opened 2021-12-15
api: remove default value of Spec.ExternalStorage.StorageProviderKind

iamniting:
having default value fills in a value even in the non external mode CR
which leads to a confusion.

instead of using a default value for the traditional external mode we
will treat Spec.ExternalStorage.StorageProviderKind="" as the "rhcs" if
the ExternalStorage is enabled.

Fix: https://github.com/red-hat-storage/ocs-operator/issues/1436


nbalacha (approved):
/lgtm

PR #1440, opened 2021-12-16
External: configure sc properties for external cluster EC RBD pool

subhamkrai:
Adding extra field in storageClass "dataPool" when pool is
erasure coded pool in case of RBD.

Links:
1) storageClass https://github.com/rook/rook/blob/master/deploy/examples/csi/rbd/storageclass-ec.yaml#L45
2) Json Blob in case of EC RBD pool https://github.com/rook/rook/pull/9276#issuecomment-992107767

### DIFF
--- a/controllers/storagecluster/external_resources_test.go
+++ b/controllers/storagecluster/external_resources_test.go
@@ -557,3 +557,52 @@ func TestExternalMonitoringResources(t *testing.T) {
 		})
 	}
 }
+
+func TestErasureCodedExternalResources(t *testing.T) {
+	cr := &api.StorageCluster{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "ocsinit",
+			Namespace: "openshift-storage",
+		},
+		Spec: api.StorageClusterSpec{
+			ExternalStorage: api.ExternalStorageClusterSpec{
+				Enable: true,
+			},
+			Monitoring: &api.MonitoringSpec{
+				ReconcileStrategy: string(ReconcileStrategyIgnore),
+			},
+		},
+	}
+	externalResource := []ExternalResource{
+		{
+			Name: "ceph-rbd",
+			Kind: "StorageClass",
+			Data: map[string]string{
+				"pool": generateNameForCephBlockPool(cr),
+			},
+		},
+		{
+			Name: "ceph-rbd-ec",
+			Kind: "StorageClass",
+			Data: map[string]string{
+				"dataPool": "ec-data-pool",
+				"pool":     "replicated-metadata-pool",
+			},
+		},
+	}
+
+	for _, extR := range externalResource {
+		t.Run(extR.Name, func(t *testing.T) {
+			actualSC := newCephBlockPoolStorageClassConfiguration(cr)
+			// To override the values for external cluster
+			for k, v := range extR.Data {
+				actualSC.storageClass.Parameters[k] = v
+			}
+			assert.NotEmpty(t, actualSC.storageClass.Parameters["clusterID"])
+			assert.Equal(t, extR.Data["dataPool"], actualSC.storageClass.Parameters["dataPool"])
+			assert.Equal(t, extR.Data["pool"], actualSC.storageClass.Parameters["pool"])
+			assert.NotEmpty(t, actualSC.storageClass.Parameters["csi.storage.k8s.io/provisioner-secret-name"])
+			assert.NotEmpty(t, actualSC.storageClass.Parameters["csi.storage.k8s.io/provisioner-secret-namespace"])
+		})
+	}
+}


subhamkrai:
@leseb ^^

subhamkrai:
marked as draft as I need to check how to test external cluster changes.

controllers/storagecluster/storageclasses.go
```diff
@@ -228,6 +228,12 @@ func newCephBlockPoolStorageClassConfiguration(initData *ocsv1.StorageCluster) S
 		disable:           managementSpec.DisableStorageClass,
 		isClusterExternal: initData.Spec.ExternalStorage.Enable,
 	}
+	if dataPool != "" {
```
aruniiird, line 231: We are checking whether the `dataPool` argument is empty or not, but where are we using it?
In the line#233, we are directly creating a value (using the function `generateNameForEcCephBlockPool()`) for the `dataPool` key.

controllers/storagecluster/external_resources_test.go
```diff
@@ -44,7 +44,8 @@ var globalTestExternalResources = []ExternalResource{
 	{
 		Kind: "StorageClass",
 		Data: map[string]string{
-			"pool": "device_health_metrics",
+			"pool":     "device_health_metrics",
```
leseb, line 47: We need a specific test for EC and non-EC. This is just adding EC and not validating it. Also, it removes the non-EC case.

controllers/storagecluster/external_resources.go
```diff
@@ -308,7 +308,7 @@ func (r *StorageClusterReconciler) createExternalStorageClusterResources(instanc
 				scc = newCephFilesystemStorageClassConfiguration(instance)
 				enableRookCSICephFS = true
 			} else if d.Name == cephRbdStorageClassName {
-				scc = newCephBlockPoolStorageClassConfiguration(instance)
+				scc = newCephBlockPoolStorageClassConfiguration(instance, d.Data["dataPool"])
```
leseb, line 311: Don't pass `d.Data["dataPool"]`, better to just pass `d.Data`, in case we need something else in the future. Then in `newCephBlockPoolStorageClassConfiguration()` determine whether `d.Data["dataPool"]` exists or not.

controllers/storagecluster/storageclasses.go
leseb: That's a good point. Where is `dataPool` used? With EC, the SC config looks like this:

https://github.com/rook/rook/blob/fffb86295600196389de81905c3402f9aec6e751/deploy/examples/csi/rbd/storageclass-ec.yaml#L45-L46

leseb (changes requested):

controllers/storagecluster/external_resources_test.go
subhamkrai: added unit test.

controllers/storagecluster/external_resources.go
subhamkrai: now, passing `d.Data` only.

controllers/storagecluster/generate.go
```diff
@@ -27,6 +27,10 @@ func generateNameForCephBlockPool(initData *ocsv1.StorageCluster) string {
 	return fmt.Sprintf("%s-cephblockpool", initData.Name)
 }
 
+func generateNameForEcCephBlockPool(initData *ocsv1.StorageCluster) string {
```
leseb, line 30:
```suggestion
func generateNameForErasureCodedCephBlockPool(initData *ocsv1.StorageCluster) string {
```

controllers/storagecluster/storageclasses.go
```diff
@@ -228,6 +228,13 @@ func newCephBlockPoolStorageClassConfiguration(initData *ocsv1.StorageCluster) S
 		disable:           managementSpec.DisableStorageClass,
 		isClusterExternal: initData.Spec.ExternalStorage.Enable,
 	}
+	if data["dataPool"] != "" {
+		sc.storageClass.Parameters = map[string]string{
```
leseb, line 232: This creates a new map and removes what was allocated before. We just need to edit `sc.storageClass.Parameters["pool"]` and `sc.storageClass.Parameters["dataPool"]`

controllers/storagecluster/storageclasses.go
```diff
@@ -228,6 +228,13 @@ func newCephBlockPoolStorageClassConfiguration(initData *ocsv1.StorageCluster) S
 		disable:           managementSpec.DisableStorageClass,
 		isClusterExternal: initData.Spec.ExternalStorage.Enable,
 	}
+	if data["dataPool"] != "" {
```
leseb, line 231: Add more comments in the code to explain what is happening.

controllers/storagecluster/external_resources_test.go
```diff
@@ -557,3 +557,43 @@ func TestExternalMonitoringResources(t *testing.T) {
...
+			Data: map[string]string{
+				"pool": generateNameForCephBlockPool(cr),
+			},
+		},
+	}
+	for _, extR := range externalResource {
+		t.Run(extR.Name, func(t *testing.T) {
+			sc := newCephBlockPoolStorageClassConfiguration(cr, extR.Data)
+			assert.Equal(t, sc.storageClass.Parameters["dataPool"], extR.Data["dataPool"])
+			assert.Equal(t, sc.storageClass.Parameters["pool"], extR.Data["pool"])
```
leseb, line 596: Also test that previous fields don't go away, e,g: `clusterID`

leseb (changes requested):

controllers/storagecluster/external_resources_test.go
subhamkrai: added.

controllers/storagecluster/storageclasses.go
subhamkrai: right, updated accordingly.

controllers/storagecluster/external_resources_test.go
```diff
@@ -557,3 +557,61 @@ func TestExternalMonitoringResources(t *testing.T) {
...
+			Kind: "StorageClass",
+			Data: map[string]string{
+				"pool": generateNameForCephBlockPool(cr),
+			},
+		},
+		{
+			Name: "ceph-rbd-ec",
+			Kind: "StorageClass",
+			Data: map[string]string{
+				"dataPool": generateNameForCephBlockPool(cr),
```
leseb, line 588:
```suggestion
				"dataPool": "my-pool",
```

`generateNameForCephBlockPool()` is overriden when the cluster is external with the actual external pool name

controllers/storagecluster/storageclasses.go
```diff
@@ -228,6 +228,13 @@ func newCephBlockPoolStorageClassConfiguration(initData *ocsv1.StorageCluster) S
 		disable:           managementSpec.DisableStorageClass,
 		isClusterExternal: initData.Spec.ExternalStorage.Enable,
 	}
+
+	// In case of erasure coded, parameter `dataPool` is not empty. The parameter`dataPool` is the name of pool and `pool` is the name erasure coded pool
+	if data["dataPool"] != "" {
+		sc.storageClass.Parameters["pool"] = generateNameForErasureCodedCephBlockPool(initData)
+		sc.storageClass.Parameters["dataPool"] = generateNameForCephBlockPool(initData)
+	}
+	return sc
```
leseb, line 237: I wonder if this block is actually needed since the cluster is external and we don't support EC in converged AFAIK.
@aruniiird to confirm?

leseb (changes requested):

controllers/storagecluster/external_resources_test.go
subhamkrai: can you point in where it is overriden?

leseb: https://github.com/red-hat-storage/ocs-operator/pull/1440#discussion_r772957395

subhamkrai: updated.

controllers/storagecluster/storageclasses.go
aruniiird: You are right, we don't require `dataPool` value to be set here.
IMO, we don't even have to touch the function `newCephBlockPoolStorageClassConfiguration()` at all because, as you mentioned in the above comment, that `for` loop will provide appropriate values for both `pool` and `dataPool` parameters.

subhamkrai: So, basically we don't need any change just adding unit test to verify working.

controllers/storagecluster/external_resources_test.go
```diff
@@ -557,3 +557,52 @@ func TestExternalMonitoringResources(t *testing.T) {
...
+
+	for _, extR := range externalResource {
+		t.Run(extR.Name, func(t *testing.T) {
+			actualSC := newCephBlockPoolStorageClassConfiguration(cr)
+			// To override the values for external cluster
+			for k, v := range extR.Data {
+				actualSC.storageClass.Parameters[k] = v
+			}
+			assert.NotEmpty(t, actualSC.storageClass.Parameters["clusterID"])
+			assert.Equal(t, extR.Data["dataPool"], actualSC.storageClass.Parameters["dataPool"])
```
leseb, line 602: Where do you test that the `dataPool` name is `ec-data-pool`?

subhamkrai: `"dataPool": "ec-data-pool",` this what I'm passing as JSON blob

leseb: Yes and it should be in the SC parameters, so the unit test should validate this.

subhamkrai: maybe I'm missing something but I'm passing `"dataPool": "ec-data-pool"` in the JSON blob and checking it with storage class parameters `actualSC.storageClass.Parameters["dataPool"])` if provided is equal to expected.

leseb (approved):

agarwal-mudit (approved):
/lgtm

PR #1441, opened 2021-12-17
ocs-to-ocs: fetch Json connection string from rook resources

subhamkrai:
Adding method to fetch Json connection string from rook resources
like secrets, configmap, and services.


subhamkrai:
JFYI, this PR is at top of #1430

services/provider/server/server.go
```diff
@@ -206,3 +212,131 @@ func Start(port int, providerServer *OCSProviderServer, opts []grpc.ServerOption
...
+		"mon-secret":   "mon-secret",
+		"admin-secret": "admin-secret",
+	}
+	name = "rook-ceph-mon"
+	kind = "Secret"
+
+	j = append(j, externalSC.ExternalResource{Name: name, Kind: kind, Data: data})
+
+	// Secret storing CSI RBD node userID and key
+	scCsiRbdNode, err := s.consumerAPI.GetSecret("rook-csi-rbd-node", s.namespace)
```
leseb, line 252: This secret does not exist in the provider cluster. What exists is a CephClient CRD that points to a Secret with the credentials.

services/provider/server/server.go
```diff
@@ -206,3 +212,131 @@ func Start(port int, providerServer *OCSProviderServer, opts []grpc.ServerOption
...
+		"userID":  string(scCsiRbdNode.Data["userID"]),
+		"userKey": string(scCsiRbdNode.Data["userKey"]),
+	}
+	name = "rook-csi-rbd-node"
+	kind = "Secret"
+
+	j = append(j, externalSC.ExternalResource{Name: name, Kind: kind, Data: data})
+
+	// Secret storing CSI RBD provisioner userID and key
+	scCsiRbdProv, err := s.consumerAPI.GetSecret("rook-csi-rbd-provisioner", s.namespace)
```
leseb, line 267: ditto

services/provider/server/server.go
leseb: Don't you get this from `storageConsumer.Status.CephObjects` somewhere?

subhamkrai: The secrets will be created once Ceph client cr is created.

kesavanvt: @leseb yes, we can get the details from storageConsumer.Status.CephObjects. Also, does this secret value needs to be placed in both `csi rbd node` and `csi rbd provisioner` ?

leseb: There are 4 secrets for ceph-csi:

* rook-csi-cephfs-node
* rook-csi-cephfs-provisioner
* rook-csi-rbd-node
* rook-csi-rbd-provisioner

services/provider/server/server.go
```diff
@@ -206,3 +212,121 @@ func Start(port int, providerServer *OCSProviderServer, opts []grpc.ServerOption
...
+		"userKey": string(scCephUser.Data[cephUser]),
+	}
+	name = "rook-ceph-operator-creds"
+	kind = "Secret"
+
+	j = append(j, externalSC.ExternalResource{Name: name, Kind: kind, Data: data})
+
+	// Secret storing CSI RBD node userID and key
+	data = map[string]string{
+		"userID":  cephUser,
```
subhamkrai, line 273: @kesavanvt please confirm if this is correct Thanks

services/provider/server/consumer/api_updater.go
```diff
@@ -20,6 +20,8 @@ type APIUpdater interface {
 	ListStorageConsumers(namespace string) (*ocsv1alpha1.StorageConsumerList, error)
 	UpdateStorageConsumer(sc *ocsv1alpha1.StorageConsumer) error
 	GetSecret(name, namespace string) (*corev1.Secret, error)
+	GetConfigMap(name, namespace string) (*corev1.ConfigMap, error)
```
leseb, line 23: We need to rebase on top of https://github.com/red-hat-storage/ocs-operator/pull/1430 now since the Interface accepts a context.

services/provider/server/server.go
```diff
@@ -92,7 +94,11 @@ func (s *OCSProviderServer) GetStorageConfig(ctx context.Context, req *pb.Storag
 	case ocsv1alpha1.StorageConsumerStateDeleting:
 		return nil, status.Errorf(codes.Internal, "storageConsumer is already in deleting phase")
 	case ocsv1alpha1.StorageConsumerStateReady:
-		// TODO: Return json connection details
+		// use JSON connection string accordingly
+		_, err = s.getJSONConnectionDetails(consumerResource)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get JSON connection string. Error %v", err)
```
leseb, line 100: `fmt.Errorf("failed to get JSON connection string. %v", err)`

services/provider/server/server.go
```diff
@@ -206,3 +212,121 @@ func Start(port int, providerServer *OCSProviderServer, opts []grpc.ServerOption
...
 	}
 }
+
+func (s *OCSProviderServer) getJSONConnectionDetails(consumerResource *ocsv1alpha1.StorageConsumer) ([]externalSC.ExternalResource, error) {
+	var j []externalSC.ExternalResource
+
+	// Configmap with mon endpoints
+	configmap, err := s.consumerAPI.GetConfigMap("rook-ceph-mon-endpoints", s.namespace)
+	if err != nil {
+		return []externalSC.ExternalResource{}, fmt.Errorf("failed to get `rook-ceph-mon-endpoints` configMap. Error %v", err)
```
leseb, line 222: remove: `Error`.

services/provider/server/server.go
```diff
@@ -206,3 +212,121 @@ func Start(port int, providerServer *OCSProviderServer, opts []grpc.ServerOption
...
+	}
+
+	cmData := strings.Split(configmap.Data["data"], ",")
```
leseb, line 225: So that's how you get the first mon? Perhaps adding a comment will help to understand what's going on here.
Also add a new var for the mon address so it's clearer.

services/provider/server/server.go
```diff
@@ -206,3 +212,121 @@ func Start(port int, providerServer *OCSProviderServer, opts []grpc.ServerOption
...
+	data := map[string]string{
+		"data":     cmData[0],
+		"maxMonId": "0",
+		"mapping":  "{}",
+	}
+	name := "rook-ceph-mon-endpoints"
```
leseb, line 231: This is used multiple times and should be a const

services/provider/server/server.go
```diff
@@ -206,3 +212,121 @@ func Start(port int, providerServer *OCSProviderServer, opts []grpc.ServerOption
...
+	kind := "ConfigMap"
+
+	j = append(j, externalSC.ExternalResource{Name: name, Kind: kind, Data: data})
+
+	// Secret storing cluster mon.admin key, fsid and name
+	scMon, err := s.consumerAPI.GetSecret("rook-ceph-mon", s.namespace)
```
leseb, line 237: ditto const

services/provider/server/server.go
```diff
@@ -206,3 +212,121 @@ func Start(port int, providerServer *OCSProviderServer, opts []grpc.ServerOption
...
+	if err != nil {
+		return []externalSC.ExternalResource{}, fmt.Errorf("failed to get `rook-ceph-mon` secret. Error %v", err)
```
leseb, line 239: ditto no Error

services/provider/server/server.go
```diff
@@ -206,3 +212,121 @@ func Start(port int, providerServer *OCSProviderServer, opts []grpc.ServerOption
...
+	}
+	data = map[string]string{
+		"fsid":         string(scMon.Data["fsid"]),
+		"mon-secret":   "mon-secret",
+		"admin-secret": "admin-secret",
+	}
+	name = "rook-ceph-mon"
```
leseb, line 246: ditto const

services/provider/server/server.go
```diff
@@ -206,3 +212,121 @@ func Start(port int, providerServer *OCSProviderServer, opts []grpc.ServerOption
...
+
+	// Ceph user secret
+	storageConsumer, err := s.consumerAPI.GetStorageConsumer(consumerResource.Name, s.namespace)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get storageConsumer resource for the given consumer: %v", err)
+	}
+	cephUser := storageConsumer.Status.CephObjects.CephUser
+	scCephUser, err := s.consumerAPI.GetSecret(fmt.Sprintf("rook-ceph-client-%s", cephUser), s.namespace)
+	if err != nil {
+		return []externalSC.ExternalResource{}, fmt.Errorf("failed to get `rook-ceph-client %s` secret. Error %v", "", err)
```
leseb, line 259: why is `%s` pointing to an emtpty string?

services/provider/server/server.go
```diff
@@ -206,3 +212,121 @@ func Start(port int, providerServer *OCSProviderServer, opts []grpc.ServerOption
...
+		"userID":  "client." + cephUser,
+		"userKey": string(scCephUser.Data[cephUser]),
+	}
+	name = "rook-ceph-operator-creds"
+	kind = "Secret"
+
+	j = append(j, externalSC.ExternalResource{Name: name, Kind: kind, Data: data})
+
+	// Secret storing CSI RBD node userID and key
+	data = map[string]string{
```
leseb, line 272: A loop will probably be better here for the ceph csi secrets.

services/provider/server/server.go
```diff
@@ -206,3 +212,121 @@ func Start(port int, providerServer *OCSProviderServer, opts []grpc.ServerOption
...
+	}
+	name = "rook-csi-rbd-provisioner"
+	kind = "Secret"
+
+	j = append(j, externalSC.ExternalResource{Name: name, Kind: kind, Data: data})
+
+	// Service for monitoring endpoints
+	scMonitoring, err := s.consumerAPI.GetService("rook-ceph-mgr", s.namespace)
+	if err != nil {
+		return []externalSC.ExternalResource{}, fmt.Errorf("failed to get `rook-ceph-mgr` service. Error %v", err)
```
leseb, line 294: ditto no Error string

services/provider/server/server.go
```diff
@@ -206,3 +212,121 @@ func Start(port int, providerServer *OCSProviderServer, opts []grpc.ServerOption
...
+		"MonitoringEndpoint": scMonitoring.Spec.ClusterIP,
+		"MonitoringPort":     fmt.Sprintf("%v", scMonitoring.Spec.Ports[0].Port),
+	}
+	name = "monitoring-endpoint"
+	kind = "CephCluster"
+
+	j = append(j, externalSC.ExternalResource{Name: name, Kind: kind, Data: data})
+
+	// Service having mgr ip and port for ceph dashboard
+	scDashboard, err := s.consumerAPI.GetService("rook-ceph-mgr-dashboard", s.namespace)
```
leseb, line 307: Why do we need this? What is using the ceph dashboard?

leseb (changes requested):

services/provider/server/server.go
subhamkrai: I'm not seeing it. could you point to a reference?

services/provider/server/server.go
subhamkrai: so, if we are not using than I'll remove.

services/provider/server/server.go
leseb: It's a repetitive block, hence using a loop will be better than duplicating code.

services/provider/server/server.go
subhamkrai: added.

services/provider/server/server.go
subhamkrai: added string var.

services/provider/server/server.go
subhamkrai: got it. added loop

services/provider/server/consumer/api_updater.go
subhamkrai: I'll rebase in the end once #1430  is merged otherwise I'll have to rebase all the time that PR is modified.

services/provider/server/server.go
```diff
@@ -206,3 +217,101 @@ func Start(port int, providerServer *OCSProviderServer, opts []grpc.ServerOption
...
+		return nil, fmt.Errorf("failed to get storageConsumer resource for the given consumer: %v", err)
+	}
+	cephUser := storageConsumer.Status.CephObjects.CephUser
+	scCephUser, err := s.consumerAPI.GetSecret(fmt.Sprintf("rook-ceph-client-%s", cephUser), s.namespace)
+	if err != nil {
+		return []externalSC.ExternalResource{}, fmt.Errorf("failed to get `rook-ceph-client-%s` secret. %v", cephUser, err)
+	}
+
+	data = map[string]string{
+		"userID":  "client." + cephUser,
```
kesavanvt, line 271: instead here , cephUser details need to be added in the provisioner and node secrets

services/provider/server/server.go
```diff
@@ -206,3 +217,101 @@ func Start(port int, providerServer *OCSProviderServer, opts []grpc.ServerOption
...
+	kind = "Secret"
+
+	j = append(j, externalSC.ExternalResource{Name: name, Kind: kind, Data: data})
+
+	// Storing CSI RBD node and provisioner userID and key
+	csiSecretList := []string{"node", "provisioner"}
+	for i := range csiSecretList {
+		data = map[string]string{
+			"userID":  "csi-rbd-" + csiSecretList[i],
+			"userKey": string(scCephUser.Data[cephUser]),
```
kesavanvt, line 284:
```
			"userID":  "rook-ceph-client-" + cephUser,
			"userKey": string(scCephUser.Data[cephUser]),
```

services/provider/server/server.go
```diff
@@ -206,3 +217,101 @@ func Start(port int, providerServer *OCSProviderServer, opts []grpc.ServerOption
 		klog.Fatalf("failed to start gRPC server: %v", err)
 	}
 }
+
+func (s *OCSProviderServer) getJSONConnectionDetails(consumerResource *ocsv1alpha1.StorageConsumer) ([]externalSC.ExternalResource, error) {
+	var j []externalSC.ExternalResource
```
sp98, line 222: Need a better var name than `j`

services/provider/server/server.go
```diff
@@ -206,3 +217,101 @@ func Start(port int, providerServer *OCSProviderServer, opts []grpc.ServerOption
...
+	// Configmap with mon endpoints
+	configmap, err := s.consumerAPI.GetConfigMap(monConfigMap, s.namespace)
+	if err != nil {
+		return []externalSC.ExternalResource{}, fmt.Errorf("failed to get %s configMap. %v", monConfigMap, err)
+	}
+
+	// This is done to get address of first mon from the monConfigMap configmap
+	cmData := strings.Split(configmap.Data["data"], ",")
+	// Address of first mon
+	monAddr := cmData[0]
```
leseb, line 233: What if the CM data is empty?

services/provider/server/server.go
```diff
@@ -4,46 +4,314 @@ import (
...
+func (s *OCSProviderServer) getJSONConnectionDetails(consumerResource *ocsv1alpha1.StorageConsumer) ([]externalSC.ExternalResource, error) {
+	var j []externalSC.ExternalResource
+
+	// Configmap with mon endpoints
+	configmap, err := s.consumerAPI.GetConfigMap(monConfigMap, s.namespace)
+	if err != nil {
+		return []externalSC.ExternalResource{}, fmt.Errorf("failed to get %s configMap. %v", monConfigMap, err)
+	}
+
+	// This is done to get address of first mon from the monConfigMap configmap
```
sp98, line 230:
```suggestion
	// Get address of first mon from the monConfigMap configmap
```

services/provider/server/server.go
```diff
@@ -206,3 +217,101 @@ func Start(port int, providerServer *OCSProviderServer, opts []grpc.ServerOption
...
+
+	j = append(j, externalSC.ExternalResource{Name: name, Kind: kind, Data: data})
+
+	// Secret storing cluster mon.admin key, fsid and name
+	scMon, err := s.consumerAPI.GetSecret(monSecret, s.namespace)
+	if err != nil {
+		return []externalSC.ExternalResource{}, fmt.Errorf("failed to get %s secret. %v", monSecret, err)
+	}
+	data = map[string]string{
+		"fsid":         string(scMon.Data["fsid"]),
```
leseb, line 250: What if the CM data is empty? Or if `scMon.Data["fsid"]` does not exist?

services/provider/server/server.go
```diff
@@ -206,3 +217,101 @@ func Start(port int, providerServer *OCSProviderServer, opts []grpc.ServerOption
...
+	}
+	cephUser := storageConsumer.Status.CephObjects.CephUser
+	scCephUser, err := s.consumerAPI.GetSecret(fmt.Sprintf("rook-ceph-client-%s", cephUser), s.namespace)
+	if err != nil {
+		return []externalSC.ExternalResource{}, fmt.Errorf("failed to get `rook-ceph-client-%s` secret. %v", cephUser, err)
+	}
+
+	data = map[string]string{
+		"userID":  "client." + cephUser,
+		"userKey": string(scCephUser.Data[cephUser]),
```
leseb, line 272: ditto

services/provider/server/server.go
```diff
@@ -206,3 +217,101 @@ func Start(port int, providerServer *OCSProviderServer, opts []grpc.ServerOption
...
+		name = "rook-csi-rbd-" + csiSecretList[i]
+		kind = "Secret"
+
+		j = append(j, externalSC.ExternalResource{Name: name, Kind: kind, Data: data})
+	}
+
+	// Service for monitoring endpoints
+	scMonitoring, err := s.consumerAPI.GetService("rook-ceph-mgr", s.namespace)
+	if err != nil {
+		return []externalSC.ExternalResource{}, fmt.Errorf("failed to get `rook-ceph-mgr` service. %v", err)
```
leseb, line 295: why backticks in the error message?

services/provider/server/server.go
```diff
@@ -206,3 +217,101 @@ func Start(port int, providerServer *OCSProviderServer, opts []grpc.ServerOption
...
+	}
+
+	data = map[string]string{
+		"MonitoringEndpoint": scMonitoring.Spec.ClusterIP,
+		"MonitoringPort":     fmt.Sprintf("%v", scMonitoring.Spec.Ports[0].Port),
```
leseb, line 300: use `strconv.Itoa` to convert the int to string

leseb (changes requested):

services/provider/server/server.go
```diff
@@ -4,46 +4,314 @@ import (
...
+		"mon-secret":   "mon-secret",
+		"admin-secret": "admin-secret",
+	}
+	name = monSecret
+	kind = "Secret"
+
+	j = append(j, externalSC.ExternalResource{Name: name, Kind: kind, Data: data})
+
+	// Ceph user secret
+	storageConsumer, err := s.consumerAPI.GetStorageConsumer(consumerResource.Name, s.namespace)
```
sp98, line 260: No need to call GetStorageConsumer again.   `getJSONConnectionDetails` function already has StorageConsumer resource as an argument.

services/provider/server/server.go
```diff
@@ -4,46 +4,314 @@ import (
...
+	case ocsv1alpha1.StorageConsumerStateFailed:
+		// TODO: get correct error message from the storageConsumer status
+		return nil, status.Errorf(codes.Internal, "storageConsumer status failed")
+	case ocsv1alpha1.StorageConsumerStateConfiguring:
+		return nil, status.Errorf(codes.Unavailable, "waiting for the rook resources to be provisioned")
+	case ocsv1alpha1.StorageConsumerStateDeleting:
+		return nil, status.Errorf(codes.Internal, "storageConsumer is already in deleting phase")
+	case ocsv1alpha1.StorageConsumerStateReady:
+		// use JSON connection string accordingly
+		_, err = s.getJSONConnectionDetails(consumerResource)
```
sp98, line 103: result of this function call should be returned in StorageConfigResponse.

sp98 (changes requested):

services/provider/server/server.go
subhamkrai: updated

services/provider/server/server.go
subhamkrai: removed

services/provider/server/server.go
subhamkrai: updated

services/provider/server/server.go
subhamkrai: updated.

services/provider/server/server.go
subhamkrai: I thought we assume clusters with these kinds of resources will be available but anyway added the conditions to check.

sp98 (changes requested):

services/provider/server/server.go
```diff
@@ -4,46 +4,333 @@ import (
...
+		return nil, status.Errorf(codes.Internal, "storageConsumer status failed")
+	case ocsv1alpha1.StorageConsumerStateConfiguring:
+		return nil, status.Errorf(codes.Unavailable, "waiting for the rook resources to be provisioned")
+	case ocsv1alpha1.StorageConsumerStateDeleting:
+		return nil, status.Errorf(codes.Internal, "storageConsumer is already in deleting phase")
+	case ocsv1alpha1.StorageConsumerStateReady:
+		// use JSON connection string accordingly
+		conString, err := s.getJSONConnectionDetails(consumerResource)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get JSON connection string. %v", err)
```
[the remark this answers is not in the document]
subhamkrai: right, I forgot this. Thanks

services/provider/server/consumer/api_updater.go
subhamkrai: decided to rebase since lot was changing which impact my changes.

services/provider/server/api_updater.go
```diff
@@ -84,3 +86,15 @@ func (c cAPIUpdater) GetSecret(ctx context.Context, name, namespace string) (*co
 	err := c.client.Get(ctx, types.NamespacedName{Name: name, Namespace: namespace}, secret)
 	return secret, err
 }
+
+func (c cAPIUpdater) GetConfigMap(ctx context.Context, name, namespace string) (*corev1.ConfigMap, error) {
+	configMap := &corev1.ConfigMap{}
+	err := c.client.Get(context.TODO(), types.NamespacedName{Name: name, Namespace: namespace}, configMap)
```
leseb, line 92: you have `ctx` so you don't need the `context.TODO()`

services/provider/server/api_updater.go
```diff
@@ -84,3 +86,15 @@ func (c cAPIUpdater) GetSecret(ctx context.Context, name, namespace string) (*co
...
+	return configMap, err
+}
+
+func (c cAPIUpdater) GetService(ctx context.Context, name, namespace string) (*corev1.Service, error) {
+	service := &corev1.Service{}
+	err := c.client.Get(context.TODO(), types.NamespacedName{Name: name, Namespace: namespace}, service)
```
leseb, line 98: ditto

sp98:
LGTM.  Lets wait for the dependent PRs to be reviewed and merged.

services/provider/server/server.go
```diff
@@ -204,3 +224,120 @@ func Start(port int, providerServer *OCSProviderServer, opts []grpc.ServerOption
...
+	if err != nil {
+		return []externalSC.ExternalResource{}, fmt.Errorf("failed to get %s secret. %v", monSecret, err)
+	}
+	if string(scMon.Data["fsid"]) == "" {
+		return []externalSC.ExternalResource{}, fmt.Errorf("secret %s data fsid is empty", monSecret)
+	}
+
+	data = map[string]string{
+		"fsid":         string(scMon.Data["fsid"]),
+		"mon-secret":   "mon-secret",
```
BlaineEXE, line 267: Should the `mon-secret` be set to the text "mon-secret" like this?

subhamkrai:
depends on #1446

services/provider/server/server.go
subhamkrai: yes, check [this](https://github.com/rook/rook/blob/master/deploy/examples/create-external-cluster-resources.py#L912-L915).

services/provider/server/server.go
```diff
@@ -188,3 +204,143 @@ func newClient() (client.Client, error) {
 
 	return client, nil
 }
+func (s *OCSProviderServer) getJSONConnectionDetails(consumerResource *ocsv1alpha1.StorageConsumer) ([]*pb.ExternalResource, error) {
+	var extR []*pb.ExternalResource
```
sp98, line 208: you should be able to use the same context that is passed into the `GetStorageConfig` function

subhamkrai: updated

services/provider/server/server.go
```diff
@@ -189,3 +204,177 @@ func newClient() (client.Client, error) {
...
+	data, err := json.Marshal(map[string]string{
+		"data":     monAddr,
+		"maxMonId": "0",
+		"mapping":  "{}",
+	})
+	if err != nil {
+		return []*pb.ExternalResource{}, fmt.Errorf("failed to marshal %s configmap data", monConfigMap)
+	}
+
+	extR = append(extR, &pb.ExternalResource{Name: name, Kind: kind, Data: data})
```
nb-ohad, line 237:
```suggestion
extR = append(extR, &pb.ExternalResource{
    Name: monConfigMap, 
    Kind: "ConfigMap", 
    Data: mustMarshal(map[string]string{
        "data":     monAddr,
        "maxMonId": "0",
        "mapping":  "{}",
    })
})
```

services/provider/server/server.go
nb-ohad: The above will be a panic if the `cmData` is an empty slice.  please check length/size instead

services/provider/server/server.go
```diff
@@ -189,3 +204,177 @@ func newClient() (client.Client, error) {
...
+	data, err = json.Marshal(map[string]string{
+		"fsid":         string(scMon.Data["fsid"]),
+		"mon-secret":   "mon-secret",
+		"admin-secret": "admin-secret",
+	})
+	if err != nil {
+		return []*pb.ExternalResource{}, fmt.Errorf("failed to marshal %s secret data", monSecret)
+	}
+
+	extR = append(extR, &pb.ExternalResource{Name: name, Kind: kind, Data: data})
```
nb-ohad, line 260:
```suggestion 
fsid := string(scMon.Data["fsid"])
if fsid == "" {
    return nil, fmt.Errorf("secret %s data fsid is empty", monSecret)
}
extR = append(extR, &pb.ExternalResource{
      Name: monSecret, 
      Kind: "Secret", 
      Data: mustMarshal(map[string]string{
		"fsid":         string(scMon.Data["fsid"]),
                 "mon-secret":   "mon-secret",
   	         "admin-secret": "admin-secret",
	})
})
```

services/provider/server/server.go
```diff
@@ -189,3 +204,177 @@ func newClient() (client.Client, error) {
...
+
+	name = "monitoring-endpoint"
+	kind = "CephCluster"
+	data, err = json.Marshal(map[string]string{
+		"MonitoringEndpoint": scMonitoring.Spec.ClusterIP,
+		"MonitoringPort":     strconv.Itoa(int(scMonitoring.Spec.Ports[0].Port)),
+	})
+	if err != nil {
+		return []*pb.ExternalResource{}, fmt.Errorf("failed to marshal %s ceph cluster data", name)
+	}
```
nb-ohad, line 281: Suggesting same refactor as above asks, to enhance readability

services/provider/server/server.go
```diff
@@ -189,3 +205,150 @@ func newClient() (client.Client, error) {
...
+		return nil, fmt.Errorf("failed to get %s configMap. %v", monConfigMap, err)
+	}
+
+	// Get address of first mon from the monConfigMap configmap
+	cmData := strings.Split(configmap.Data["data"], ",")
+	if len(cmData) == 0 {
+		return nil, fmt.Errorf("configmap %s data is empty", monConfigMap)
+	}
+
+	extR = append(extR, &pb.ExternalResource{Name: monConfigMap, Kind: "ConfigMap", Data: mustMarshal(map[string]string{
```
nb-ohad, line 224: Can we split the params across lines to make this line more readable?

```suggestion
extR = append(extR, &pb.ExternalResource{
	Name: monConfigMap, 
	Kind: "ConfigMap", 
	Data: mustMarshal(map[string]string{
		"data":     cmData[0], // Address of first mon
		"maxMonId": "0",
		"mapping":  "{}",
	})
})
```

services/provider/server/server.go
```diff
@@ -189,3 +205,150 @@ func newClient() (client.Client, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get %s secret. %v", monSecret, err)
+	}
+
+	fsid := string(scMon.Data["fsid"])
+	if fsid == "" {
+		return nil, fmt.Errorf("secret %s data fsid is empty", monSecret)
+	}
+
+	extR = append(extR, &pb.ExternalResource{Name: monSecret, Kind: "Secret", Data: mustMarshal(map[string]string{
```
nb-ohad, line 242: Same as above comment about line readability

services/provider/server/server.go
```diff
@@ -189,3 +205,150 @@ func newClient() (client.Client, error) {
...
+	err = s.client.Get(ctx, types.NamespacedName{Name: "rook-ceph-mgr", Namespace: s.namespace}, scMonitoring)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get rook-ceph-mgr service. %v", err)
+	}
+
+	if scMonitoring.Spec.ClusterIP == "" || strconv.Itoa(int(scMonitoring.Spec.Ports[0].Port)) == "" {
+		return nil, fmt.Errorf("service rook-ceph-mgr clusterIP or port is empty")
+	}
+
+	extR = append(extR, &pb.ExternalResource{Name: "monitoring-endpoint", Kind: "CephCluster", Data: mustMarshal(map[string]string{
```
nb-ohad, line 259: Same as my last comment about line readability

services/provider/server/server.go
```diff
@@ -189,3 +205,150 @@ func newClient() (client.Client, error) {
...
+				return nil, fmt.Errorf("failed to get %s secret. %v", i.CephClients["node"], err)
+			}
+			data["userID"] = i.Name
+			data["userKey"] = string(cephUserSecret.Data[i.Name])
+
+			extR = append(extR, &pb.ExternalResource{
+				Name: i.Name,
+				Kind: "Secret",
+				Data: mustMarshal(data),
+			})
```
nb-ohad, line 279: What happened to all of the other secrets?

nb-ohad, line 279:
```suggestion
			extR = append(extR, &pb.ExternalResource{
				Name: i.Name,
				Kind: "Secret",
				Data: mustMarshal(map[string]string{
					"userID" = i.name,
					"userKey" = string(cephUserSecret.Data[i.Name]),
				}),
			})
```

This also removes the need for line L265

services/provider/server/server.go
```diff
@@ -189,3 +205,150 @@ func newClient() (client.Client, error) {
...
+
+				data["userID"] = i.Name
+				data["userKey"] = string(cephUserSecret.Data[i.Name])
+
+				extR = append(extR, &pb.ExternalResource{
+					Name: i.CephClients["node"],
+					Kind: "Secret",
+					Data: mustMarshal(data),
+				})
+			}
```
nb-ohad, line 318: Why are we mixing the creation of clients and storage classes?
These are different concerns, you don't need to create them here. Let's move the creation of the client secrets to the CephClinet section above

services/provider/server/server.go
```diff
@@ -189,3 +206,138 @@ func newClient() (client.Client, error) {
...
+			} else {
+				extR = append(extR, &pb.ExternalResource{
+					Name: name,
+					Kind: "Secret",
+					Data: mustMarshal(map[string]string{
+						"userID":  i.Name,
+						"userKey": string(cephUserSecret.Data[i.Name]),
+					}),
+				})
+			}
```
nb-ohad, line 306: xyz

services/provider/server/server.go
```diff
@@ -189,3 +206,138 @@ func newClient() (client.Client, error) {
...
+		} else if i.Kind == "CephBlockPool" {
+			extR = append(extR, &pb.ExternalResource{
+				Name: "ceph-rbd",
+				Kind: "StorageClass",
+				Data: mustMarshal(map[string]string{
+					"pool": i.Name,
+				})})
```
nb-ohad, line 313: zxy

services/provider/server/server.go
```diff
@@ -189,3 +206,145 @@ func newClient() (client.Client, error) {
...
+				Name: "ceph-rbd",
+				Kind: "StorageClass",
+				Data: mustMarshal(map[string]string{
+					"clusterID":                 s.namespace,
+					"pool":                      i.Name,
+					"imageFeatures":             "layering",
+					"csi.storage.k8s.io/fstype": "ext4",
+					"imageFormat":               "2",
+					"csi.storage.k8s.io/provisioner-secret-name":            provisionerCephClient.Status.Info["secretName"],
+					"csi.storage.k8s.io/provisioner-secret-namespace":       s.namespace,
```
nb-ohad, line 323: This is incorrect, we cannot assume the namespace on the consumer is the same as on the provider. remove this overwrite.

services/provider/server/server.go
```diff
@@ -189,3 +206,145 @@ func newClient() (client.Client, error) {
...
+					"csi.storage.k8s.io/node-stage-secret-name":             nodeCephClient.Status.Info["secretName"],
+					"csi.storage.k8s.io/node-stage-secret-namespace":        s.namespace,
```
nb-ohad, line 325: same as above

services/provider/server/server.go
```diff
@@ -189,3 +206,145 @@ func newClient() (client.Client, error) {
...
+					"csi.storage.k8s.io/controller-expand-secret-name":      provisionerCephClient.Status.Info["secretName"],
+					"csi.storage.k8s.io/controller-expand-secret-namespace": s.namespace,
```
nb-ohad, line 327: Same as above

services/provider/server/server.go
```diff
@@ -189,3 +206,145 @@ func newClient() (client.Client, error) {
...
+		Kind: "CephCluster",
+		Data: mustMarshal(map[string]string{
+			"MonitoringEndpoint": scMonitoring.Spec.ClusterIP,
+			"MonitoringPort":     strconv.Itoa(int(scMonitoring.Spec.Ports[0].Port)),
+		})})
+
+	for _, i := range consumerResource.Status.CephResources {
+		switch i.Kind {
+		case "CephClient":
+			cephClient, err := s.getCephClient(ctx, i.Name)
```
nb-ohad, line 277: You don't need the client object itself, so why not make the util function retrieve the secret name instead. 
As it stands now the util function is just a very small wrapper over normal get operation which does not bring a lot of value.

```suggestion
			cephClientSecretName, err := s.getCephClientSecretName(ctx, i.Name)
```

services/provider/server/server.go
```diff
@@ -189,3 +206,142 @@ func newClient() (client.Client, error) {
...
+	return extR, nil
+}
+
+func (s *OCSProviderServer) getCephClientSecretName(ctx context.Context, name string) (string, error) {
+	cephClient := &rookCephv1.CephClient{}
+	err := s.client.Get(ctx, types.NamespacedName{Name: name, Namespace: s.namespace}, cephClient)
+	if err != nil {
+		return "", fmt.Errorf("failed to get cephClient %s secret. %v", name, err)
+	}
+	return cephClient.Status.Info["secretName"], nil
```
nb-ohad, line 338: `cephClient` might not have a status when you try to read it (we might read it before it is reconcile by rook)
You need to protect against this access or else it will result in a panic.

Same goes for the Info map if the Info map is optional on the Status structure (not sure about that, you will need to check the type definition)

services/provider/server/server_test.go
```diff
@@ -0,0 +1,242 @@
...
+	exR, err := server.getExternalResources(ctx, consumerResource)
+	assert.NoError(t, err)
+	for i := range exR {
+		data, err := json.Marshal(mockExtR[i].Data)
+		assert.NoError(t, err)
+		assert.Equal(t, string(exR[i].Data), string(data))
+		assert.Equal(t, exR[i].Name, mockExtR[i].Name)
+		assert.Equal(t, exR[i].Kind, mockExtR[i].Kind)
+
+	}
```
nb-ohad, line 241: This only checks the happy path scenario, what about testing failures as well?

subhamkrai: I'll update the unit test soon.

services/provider/server/server_test.go
```diff
@@ -0,0 +1,298 @@
...
+	exR, err = server.getExternalResources(ctx, consumerResource)
+	assert.Equal(t, err, fmt.Errorf("cephClient cephuser-rbd-node Status.Info is empty"))
+	assert.NotEqual(t, len(mockExtR), len(exR))
+
+	// when storageConsume status is empty
+	exR, err = server.getExternalResources(ctx, &ocsv1alpha1.StorageConsumer{})
+	assert.Equal(t, err, fmt.Errorf("ceph resources are empty"))
+	assert.NotEqual(t, len(mockExtR), len(exR))
+
+}
```
nb-ohad, line 255: I think the test suite should include tests for all possible storage consumer states and verify the correct error code (the error message is really unimportant).
```
       ocsv1alpha1.StorageConsumerStateFailed -> codes.Internal
       ocsv1alpha1.StorageConsumerStateConfiguring -> codes.Unavailable
       ocsv1alpha1.StorageConsumerStateDeleting -> codes.NotFound
```

services/provider/server/server.go
```diff
@@ -189,3 +206,153 @@ func newClient() (client.Client, error) {
...
+		Name: "monitoring-endpoint",
+		Kind: "CephCluster",
+		Data: mustMarshal(map[string]string{
+			"MonitoringEndpoint": scMonitoring.Spec.ClusterIP,
+			"MonitoringPort":     strconv.Itoa(int(scMonitoring.Spec.Ports[0].Port)),
+		})})
+
+	if len(consumerResource.Status.CephResources) == 0 {
+		return nil, fmt.Errorf("ceph resources are empty")
+	}
```
nb-ohad, line 276: This should not be an error, it is ok (from the API server perspective to have an empty `CephResources` list, please remove the if statmenent

services/provider/server/server_test.go
```diff
@@ -0,0 +1,298 @@
...
+	assert.NoError(t, client.Create(ctx, monSc))
+	assert.NoError(t, client.Create(ctx, mgrSvc))
+
+	exR, err = server.getExternalResources(ctx, consumerResource)
+	assert.Equal(t, err, fmt.Errorf("cephClient cephuser-rbd-node Status.Info is empty"))
+	assert.NotEqual(t, len(mockExtR), len(exR))
+
+	// when storageConsume status is empty
+	exR, err = server.getExternalResources(ctx, &ocsv1alpha1.StorageConsumer{})
+	assert.Equal(t, err, fmt.Errorf("ceph resources are empty"))
```
nb-ohad, line 252: The check here is incorrect, you are testing the API server, not the storage consumer controller. 
The API server does not have bais toward the number (or even existence) of resources in the StorageConsumer status as such the return value here will not be an error, it will be an empty array.

services/provider/server/server_test.go
```diff
@@ -0,0 +1,298 @@
...
+		}
+	}
+
+	monCm, monSc, mgrSvc = createMonConfigMapandSecret(server)
+	assert.NoError(t, client.Create(ctx, monCm))
+	assert.NoError(t, client.Create(ctx, monSc))
+	assert.NoError(t, client.Create(ctx, mgrSvc))
+
+	exR, err = server.getExternalResources(ctx, consumerResource)
+	assert.Equal(t, err, fmt.Errorf("cephClient cephuser-rbd-node Status.Info is empty"))
```
nb-ohad, line 247: Why are we validating against an error message, messages might change. The important thing to validate is against the error code on the `getExternalResources` response,

services/provider/server/server_test.go
```diff
@@ -0,0 +1,298 @@
...
+
+		}
+	}
+
+	monCm, monSc, mgrSvc = createMonConfigMapandSecret(server)
+	assert.NoError(t, client.Create(ctx, monCm))
+	assert.NoError(t, client.Create(ctx, monSc))
+	assert.NoError(t, client.Create(ctx, mgrSvc))
+
+	exR, err = server.getExternalResources(ctx, consumerResource)
```
nb-ohad, line 246: You should the public entry point (`GetStorageConfig`) not a private util function `getExternalResources`

services/provider/server/server_test.go
```diff
@@ -0,0 +1,298 @@
...
+	for _, i := range consumerResource.Status.CephResources {
+		if i.Kind == "CephClient" {
+			cephClient, secret := createCephClientAndSecret(i.Name, server)
+			assert.NoError(t, client.Create(ctx, cephClient))
+			assert.NoError(t, client.Create(ctx, secret))
+
+		}
+	}
+
+	exR, err = server.getExternalResources(ctx, consumerResource)
```
nb-ohad, line 211: You should the public entry point (`GetStorageConfig`) not a private util function `getExternalResources`.
You should also very the correct status code on the response

services/provider/server/server_test.go
```diff
@@ -0,0 +1,298 @@
...
+	assert.NoError(t, client.Create(ctx, monCm))
+	assert.NoError(t, client.Create(ctx, monSc))
+	assert.NoError(t, client.Create(ctx, mgrSvc))
+
+	exR, err = server.getExternalResources(ctx, consumerResource)
+	assert.Equal(t, err, fmt.Errorf("cephClient cephuser-rbd-node Status.Info is empty"))
+	assert.NotEqual(t, len(mockExtR), len(exR))
+
+	// when storageConsume status is empty
+	exR, err = server.getExternalResources(ctx, &ocsv1alpha1.StorageConsumer{})
```
nb-ohad, line 251: You should the public entry point (`GetStorageConfig`) not a private util function `getExternalResources`

services/provider/server/server_test.go
```diff
@@ -0,0 +1,298 @@
...
+			assert.NoError(t, client.Create(ctx, secret))
+		}
+	}
+
+	monCm, monSc, mgrSvc := createMonConfigMapandSecret(server)
+	assert.NoError(t, client.Create(ctx, monCm))
+	assert.NoError(t, client.Create(ctx, monSc))
+	assert.NoError(t, client.Create(ctx, mgrSvc))
+
+	exR, err := server.getExternalResources(ctx, consumerResource)
```
nb-ohad, line 179: You should the public entry point (`GetStorageConfig`) not a private util function `getExternalResources`

nb-ohad (changes requested):

services/provider/server/server_test.go
```diff
@@ -0,0 +1,343 @@
...
+			State: ocsv1alpha1.StorageConsumerStateDeleting,
+		},
+	}
+	consumerResource4 = &ocsv1alpha1.StorageConsumer{
+		ObjectMeta: metav1.ObjectMeta{Name: "consumer4", UID: "uid4"},
+		Status:     ocsv1alpha1.StorageConsumerStatus{},
+	}
+)
+
+func TestGetJSONConnectionDetails(t *testing.T) {
```
nb-ohad, line 175: I think we need to change the name of this test, the current name refer to something that does not exists anymore

services/provider/server/server_test.go
```diff
@@ -0,0 +1,343 @@
...
+	server.consumerManager.nameByUID = map[types.UID]string{
+		consumerResource4.UID: consumerResource4.Name,
+	}
+	storageConRes, err = server.GetStorageConfig(ctx, &req)
+	assert.Error(t, err)
+	assert.Equal(t, err, status.Errorf(codes.Unavailable, "storage consumer status is not set"))
+	assert.Nil(t, storageConRes)
+}
+
+func createMonConfigMapandSecret(server *OCSProviderServer) (*v1.ConfigMap, *v1.Secret, *v1.Service) {
```
nb-ohad, line 302:
```suggestion
func createMonConfigMapAndSecret(server *OCSProviderServer) (*v1.ConfigMap, *v1.Secret, *v1.Service) {
```

services/provider/server/server_test.go
```diff
@@ -0,0 +1,343 @@
...
+	assert.NoError(t, err)
+	assert.NotNil(t, storageConRes)
+
+	for i := range mockExtR {
+		data, err := json.Marshal(mockExtR[i].Data)
+		assert.NoError(t, err)
+		assert.Equal(t, storageConRes.ExternalResource[i].Data, data)
+		assert.Equal(t, storageConRes.ExternalResource[i].Kind, mockExtR[i].Kind)
+		assert.Equal(t, storageConRes.ExternalResource[i].Name, mockExtR[i].Name)
+	}
```
nb-ohad, line 240: This tests only the case where there are CephObjects on the storage consumer, 
I think it will also benefit us testing the case where there aren't any

services/provider/server/server_test.go
```diff
@@ -0,0 +1,405 @@
...
+		StorageConsumerUUID: string(consumerResource4.UID),
+	}
+	server.consumerManager.nameByUID = map[types.UID]string{
+		consumerResource4.UID: consumerResource4.Name,
+	}
+	storageConRes, err = server.GetStorageConfig(ctx, &req)
+	assert.Error(t, err)
+	errCode, _ = status.FromError(err)
+	assert.Equal(t, errCode.Code(), codes.Unavailable)
+	assert.Nil(t, storageConRes)
```
nb-ohad, line 303: I think the assertion is wrong, if the status has an empty list of CephObjects then there should be no error. 
It all depends on the status of the storage consumer. 

So if StorageConsumer CR is in Ready state and the list is empty it is an OK response (And there should be a test for it)

subhamkrai: > I think the assertion is wrong, if the status has an empty list of CephObjects then there should be no error. It all depends on the status of the storage consumer.
here I checking for `empty storage consumer status state` which will through `codes.Unavailable`
> 
> So if StorageConsumer CR is in Ready state and the list is empty it is an OK response (And there should be a test for it)
added case when consumer is in ready state but`cephClient` is empty

nb-ohad (approved):

jarrpa (approved):
/lgtm

PR #1442, opened 2021-12-17
Fix no RBD mirroring daemon alert when daemon is down issue

GowthamShanmugam:
What is changed?
  The older approach is to raise the ceph RBD mirror daemon alert using a metric that is derived from CephRBDMirror CR (alert is never raised).
  The new approach is to raise the ceph RBD mirror daemon alert using a metric that is derived from Cephcluster CR.

Example CephCluster status spec:
```
status:
  ceph:
    versions:
      rbd-mirror:
        ceph version 16.2.5 (0883bdea7337b95e4b611c768c0279868462204a) pacific (stable): 1
```

Alerting Rule:
```
((count by(namespace) (ocs_mirror_daemon_count{job="ocs-metrics-exporter"} == 0)) * on(namespace) group_left() (count by(namespace) (ocs_pool_mirroring_status{job="ocs-metrics-exporter"} == 1))) > 0
```

Note: 
  Daemon down alert is raised when at least one pool has mirroring enabled.


GowthamShanmugam:
image

umangachapagain (commented):
Why did it not work before? Was the metric incorrect?
Why did we change to a different metric? How does it work now?

GowthamShanmugam:
> Why did it not work before? Was the metric incorrect? Why did we change to a different metric? How does it work now?

the current approach is to fetch the daemon status from ``CephRbdMirroring```  using ```status.phase``` value. This phase is not exactly pointing to daemon status and it is just a reconcile status. We wrongly exported metrics out of it. Daemon down case is not reflecting anything on this phase value,

I discussed with @leseb about this and he suggested this approach, Now we are finding daemon status from ```CephCluster``` CR status.  Instead of status(no one is providing cluster daemon status) now we are fetching number of daemon are up and running for this cluster.

```
status:
  ceph:
    versions:
      rbd-mirror:
        ceph version 16.2.5 (0883bdea7337b95e4b611c768c0279868462204a) pacific (stable): 1
```

umangachapagain (approved):

PR #1444, opened 2021-12-20
ocs-to-ocs: Implement API calls for the consumer

iamniting:
As part of the PR Implement onboard, upscaling and offboard.

### DIFF
--- a/controllers/storagecluster/cephcluster.go
+++ b/controllers/storagecluster/cephcluster.go
@@ -110,11 +110,9 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
 	var cephCluster *rookCephv1.CephCluster
 	// Define a new CephCluster object
 	if sc.Spec.ExternalStorage.Enable {
-		extRArr, err := r.retrieveExternalSecretData(sc)
-		if err != nil {
-			r.Log.Error(err, "Could not retrieve the External Cluster details.",
-				"CephCluster", klog.KRef(sc.Namespace, sc.Name))
-			return reconcile.Result{}, err
+		extRArr, ok := externalOCSResources[sc.UID]
+		if !ok {
+			return reconcile.Result{}, fmt.Errorf("Unable to retrieve external resource from externalOCSResources")
 		}
 		endpointR, err := findNamedResourceFromArray(extRArr, "monitoring-endpoint")
 		if err != nil {
--- a/controllers/storagecluster/external_ocs.go
+++ b/controllers/storagecluster/external_ocs.go
@@ -1,0 +1,257 @@
+package storagecluster
+
+import (
+	"context"
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"time"
+
+	configv1 "github.com/openshift/api/config/v1"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v1"
+	externalClient "github.com/red-hat-storage/ocs-operator/services/provider/client"
+)
+
+var (
+	// externalOCSResources will hold the ExternalResources for storageclusters
+	// ExternalResources can be accessible using the UID of an storagecluster
+	externalOCSResources = map[types.UID][]ExternalResource{}
+)
+
+const (
+	// grpcCallNames
+	OnboardConsumer  = "OnboardConsumer"
+	OffboardConsumer = "OffboardConsumer"
+	UpdateCapacity   = "UpdateCapacity"
+	GetStorageConfig = "GetStorageConfig"
+)
+
+type connectionDetails struct {
+	OnboardingTicket string `json:"onboardingTicket"`
+	ServerAddress    string `json:"serverAddress"`
+}
+
+// isExternalOCSProvider returns true if it is ocs to ocs ExternalStorage consumer cluster
+func isExternalOCSProvider(instance *ocsv1.StorageCluster) bool {
+	return instance.Spec.ExternalStorage.Enable && instance.Spec.ExternalStorage.StorageProviderKind == ocsv1.KindOCS
+}
+
+func (r *StorageClusterReconciler) getConnectionDetails(instance *ocsv1.StorageCluster) (*connectionDetails, error) {
+
+	connectionInfo, err := base64.StdEncoding.DecodeString(instance.Spec.ExternalStorage.ConnectionString)
+	if err != nil {
+		r.Log.Error(err, "ConnectionString is Invalid, Not able to decode ConnectionString with base64",
+			"ConnectionString", instance.Spec.ExternalStorage.ConnectionString)
+		r.recorder.ReportIfNotPresent(instance, corev1.EventTypeWarning, "Invalid", "ConnectionString is Invalid")
+		return nil, err
+	}
+
+	var c connectionDetails
+	err = json.Unmarshal(connectionInfo, &c)
+	if err != nil {
+		r.Log.Error(err, "Failed to parse connection details, some keys are missing")
+		r.recorder.ReportIfNotPresent(instance, corev1.EventTypeWarning, "Invalid", "ConnectionString is Invalid")
+		return nil, err
+	}
+
+	return &c, nil
+}
+
+// newExternalClusterClient returns the *externalClient.OCSProviderClient
+func (r *StorageClusterReconciler) newExternalClusterClient(instance *ocsv1.StorageCluster) (*externalClient.OCSProviderClient, error) {
+
+	connectionInfo, err := r.getConnectionDetails(instance)
+	if err != nil {
+		return nil, err
+	}
+
+	consumerClient, err := externalClient.NewProviderClient(
+		context.Background(), connectionInfo.ServerAddress, time.Second*10)
+	if err != nil {
+		return nil, err
+	}
+
+	return consumerClient, nil
+}
+
+// onboardConsumer makes an API call to the external storage provider cluster for onboarding
+func (r *StorageClusterReconciler) onboardConsumer(instance *ocsv1.StorageCluster, externalClusterClient *externalClient.OCSProviderClient) (reconcile.Result, error) {
+
+	clusterVersion := &configv1.ClusterVersion{}
+	err := r.Client.Get(context.Background(), types.NamespacedName{Name: "version"}, clusterVersion)
+	if err != nil {
+		r.Log.Error(err, "External-OCS:Failed to get the clusterVersion version of the OCP cluster")
+		return reconcile.Result{}, err
+	}
+
+	connectionInfo, err := r.getConnectionDetails(instance)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	name := fmt.Sprintf("ocs-consumer-%s", clusterVersion.Spec.ClusterID)
+	response, err := externalClusterClient.OnboardConsumer(
+		context.Background(), connectionInfo.OnboardingTicket, name,
+		instance.Spec.ExternalStorage.RequestedCapacity.String())
+	if err != nil {
+		if s, ok := status.FromError(err); ok {
+			r.logGrpcErrorAndReportEvent(instance, OnboardConsumer, err, s.Code())
+		}
+		return reconcile.Result{}, err
+	}
+
+	if response.StorageConsumerUUID == "" || response.GrantedCapacity == "" {
+		err = fmt.Errorf("External-OCS:OnboardConsumer:response is empty")
+		r.Log.Error(err, "empty response")
+		return reconcile.Result{}, err
+	}
+
+	instance.Status.ExternalStorage.ConsumerID = response.StorageConsumerUUID
+	instance.Status.ExternalStorage.GrantedCapacity = resource.MustParse(response.GrantedCapacity)
+
+	return reconcile.Result{}, nil
+}
+
+// offboardConsumer makes an API call to the external storage provider cluster for offboarding
+func (r *StorageClusterReconciler) offboardConsumer(instance *ocsv1.StorageCluster, externalClusterClient *externalClient.OCSProviderClient) (reconcile.Result, error) {
+
+	_, err := externalClusterClient.OffboardConsumer(context.Background(), instance.Status.ExternalStorage.ConsumerID)
+	if err != nil {
+		if s, ok := status.FromError(err); ok {
+			r.logGrpcErrorAndReportEvent(instance, OffboardConsumer, err, s.Code())
+		}
+		return reconcile.Result{}, err
+	}
+
+	return reconcile.Result{}, nil
+}
+
+// updateConsumerCapacity makes an API call to the external storage provider cluster to update the capacity
+func (r *StorageClusterReconciler) updateConsumerCapacity(instance *ocsv1.StorageCluster, externalClusterClient *externalClient.OCSProviderClient) (reconcile.Result, error) {
+
+	response, err := externalClusterClient.UpdateCapacity(
+		context.Background(),
+		instance.Status.ExternalStorage.ConsumerID,
+		instance.Spec.ExternalStorage.RequestedCapacity.String())
+	if err != nil {
+		if s, ok := status.FromError(err); ok {
+			r.logGrpcErrorAndReportEvent(instance, UpdateCapacity, err, s.Code())
+		}
+		return reconcile.Result{}, err
+	}
+
+	responseQuantity, err := resource.ParseQuantity(response.GrantedCapacity)
+	if err != nil {
+		r.Log.Error(err, "Failed to parse GrantedCapacity from UpdateCapacity response.", "GrantedCapacity", response.GrantedCapacity)
+		return reconcile.Result{}, err
+	}
+
+	if !instance.Spec.ExternalStorage.RequestedCapacity.Equal(responseQuantity) {
+		r.Log.Error(err, "GrantedCapacity is not equal to the RequestedCapacity in the UpdateCapacity response.",
+			"GrantedCapacity", response.GrantedCapacity, "RequestedCapacity", instance.Spec.ExternalStorage.RequestedCapacity)
+		r.Log.Error(err, "empty response")
+		return reconcile.Result{}, err
+	}
+
+	instance.Status.ExternalStorage.GrantedCapacity = responseQuantity
+
+	return reconcile.Result{}, nil
+}
+
+// getExternalConfigFromProvider makes an API call to the external storage provider cluster for json blob
+func (r *StorageClusterReconciler) getExternalConfigFromProvider(
+	instance *ocsv1.StorageCluster, externalClusterClient *externalClient.OCSProviderClient) ([]ExternalResource, reconcile.Result, error) {
+
+	response, err := externalClusterClient.GetStorageConfig(context.Background(), instance.Status.ExternalStorage.ConsumerID)
+	if err != nil {
+		if s, ok := status.FromError(err); ok {
+			r.logGrpcErrorAndReportEvent(instance, GetStorageConfig, err, s.Code())
+
+			// storage consumer is not ready yet, requeue after some time
+			if s.Code() == codes.Unavailable {
+				return nil, reconcile.Result{RequeueAfter: time.Second * 5}, nil
+			}
+		}
+
+		return nil, reconcile.Result{}, err
+	}
+
+	var externalResources []ExternalResource
+
+	for _, eResource := range response.ExternalResource {
+
+		data := map[string]string{}
+		err = json.Unmarshal(eResource.Data, &data)
+		if err != nil {
+			r.Log.Error(err, "Failed to Unmarshal response of GetStorageConfig", "Kind", eResource.Kind, "Name", eResource.Name, "Data", eResource.Data)
+			return nil, reconcile.Result{}, err
+		}
+
+		externalResources = append(externalResources, ExternalResource{
+			Kind: eResource.Kind,
+			Data: data,
+			Name: eResource.Name,
+		})
+	}
+
+	return externalResources, reconcile.Result{}, nil
+}
+
+func (r *StorageClusterReconciler) logGrpcErrorAndReportEvent(instance *ocsv1.StorageCluster, grpcCallName string, err error, errCode codes.Code) {
+
+	var msg, eventReason, eventType string
+
+	if grpcCallName == OnboardConsumer {
+		if errCode == codes.InvalidArgument {
+			msg = "Token is invalid. Verify the token again or contact the provider admin"
+			eventReason = "TokenInvalid"
+			eventType = corev1.EventTypeWarning
+		} else if errCode == codes.AlreadyExists {
+			msg = "Token is already used. Contact provider admin for a new token"
+			eventReason = "TokenAlreadyUsed"
+			eventType = corev1.EventTypeWarning
+		}
+	} else if grpcCallName == OffboardConsumer {
+		if errCode == codes.InvalidArgument {
+			msg = "StorageConsumer UID is not valid. Contact the provider admin"
+			eventReason = "UIDInvalid"
+			eventType = corev1.EventTypeWarning
+		}
+	} else if grpcCallName == UpdateCapacity {
+		if errCode == codes.InvalidArgument {
+			msg = "StorageConsumer UID or requested capacity is not valid. Contact the provider admin"
+			eventReason = "UIDorCapacityInvalid"
+			eventType = corev1.EventTypeWarning
+		} else if errCode == codes.NotFound {
+			msg = "StorageConsumer UID not found. Contact the provider admin"
+			eventReason = "UIDNotFound"
+			eventType = corev1.EventTypeWarning
+		}
+	} else if grpcCallName == GetStorageConfig {
+		if errCode == codes.InvalidArgument {
+			msg = "StorageConsumer UID is not valid. Contact the provider admin"
+			eventReason = "UIDInvalid"
+			eventType = corev1.EventTypeWarning
+		} else if errCode == codes.NotFound {
+			msg = "StorageConsumer UID not found. Contact the provider admin"
+			eventReason = "UIDNotFound"
+			eventType = corev1.EventTypeWarning
+		} else if errCode == codes.Unavailable {
+			msg = "StorageConsumer is not ready yet. Will requeue after 5 second"
+			eventReason = "NotReady"
+			eventType = corev1.EventTypeNormal
+		}
+	}
+
+	if msg != "" {
+		r.Log.Error(err, "External-OCS:"+grpcCallName+":"+msg)
+		r.recorder.ReportIfNotPresent(instance, eventType, eventReason, msg)
+	}
+}
--- a/controllers/storagecluster/external_resources.go
+++ b/controllers/storagecluster/external_resources.go
@@ -264,6 +264,13 @@ func (obj *ocsExternalResources) ensureCreated(r *StorageClusterReconciler, inst
 		if r.sameExternalSecretData(instance) {
 			return reconcile.Result{}, nil
 		}
+
+		data, err := r.retrieveExternalSecretData(instance)
+		if err != nil {
+			r.Log.Error(err, "Failed to retrieve external secret resources.")
+			return reconcile.Result{}, err
+		}
+		externalOCSResources[instance.UID] = data
 	}
 
 	err := r.createExternalStorageClusterResources(instance)
@@ -292,6 +299,9 @@ func (obj *ocsExternalResources) ensureDeleted(r *StorageClusterReconciler, inst
 
 // createExternalStorageClusterResources creates external cluster resources
 func (r *StorageClusterReconciler) createExternalStorageClusterResources(instance *ocsv1.StorageCluster) error {
+
+	var err error
+
 	ownerRef := metav1.OwnerReference{
 		UID:        instance.UID,
 		APIVersion: instance.APIVersion,
@@ -302,11 +312,11 @@ func (r *StorageClusterReconciler) createExternalStorageClusterResources(instanc
 	enableRookCSICephFS := false
 	// this stores only the StorageClasses specified in the Secret
 	availableSCCs := []StorageClassConfiguration{}
-	data, err := r.retrieveExternalSecretData(instance)
-	if err != nil {
-		r.Log.Error(err, "Failed to retrieve external secret resources.")
-		return err
+	data, ok := externalOCSResources[instance.UID]
+	if !ok {
+		return fmt.Errorf("Unable to retrieve external resource from externalOCSResources")
 	}
+
 	var extCephObjectStores []*cephv1.CephObjectStore
 	for _, d := range data {
 		objectMeta := metav1.ObjectMeta{
--- a/controllers/storagecluster/external_resources.go
+++ b/controllers/storagecluster/external_resources.go
@@ -230,9 +230,42 @@ func (r *StorageClusterReconciler) newExternalCephObjectStoreInstances(
 // ensureCreated ensures that requested resources for the external cluster
 // being created
 func (obj *ocsExternalResources) ensureCreated(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) (reconcile.Result, error) {
-	if r.sameExternalSecretData(instance) {
-		return reconcile.Result{}, nil
+
+	if isExternalOCSProvider(instance) {
+
+		externalClusterClient, err := r.newExternalClusterClient(instance)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		defer externalClusterClient.Close()
+
+		if instance.Status.ExternalStorage.ConsumerID == "" {
+			res, err := r.onboardConsumer(instance, externalClusterClient)
+			if err != nil || !res.IsZero() {
+				return res, err
+			}
+		} else if instance.Spec.ExternalStorage.RequestedCapacity.Equal(instance.Status.ExternalStorage.GrantedCapacity) {
+			res, err := r.updateConsumerCapacity(instance, externalClusterClient)
+			if err != nil || !res.IsZero() {
+				return res, err
+			}
+		}
+
+		if externalOCSResources[instance.UID] == nil {
+			externalConfig, res, err := r.getExternalConfigFromProvider(instance, externalClusterClient)
+			if err != nil || !res.IsZero() {
+				return res, err
+			}
+			externalOCSResources[instance.UID] = externalConfig
+		}
+		externalClusterClient.Close()
+	} else {
+		// rhcs external mode
+		if r.sameExternalSecretData(instance) {
+			return reconcile.Result{}, nil
+		}
 	}
+
 	err := r.createExternalStorageClusterResources(instance)
 	if err != nil {
 		r.Log.Error(err, "Could not create ExternalStorageClusterResource.", "StorageCluster", klog.KRef(instance.Namespace, instance.Name))
@@ -243,6 +276,17 @@ func (obj *ocsExternalResources) ensureCreated(r *StorageClusterReconciler, inst
 
 // ensureDeleted is dummy func for the ocsExternalResources
 func (obj *ocsExternalResources) ensureDeleted(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) (reconcile.Result, error) {
+
+	if isExternalOCSProvider(instance) {
+		externalClusterClient, err := r.newExternalClusterClient(instance)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		defer externalClusterClient.Close()
+
+		return r.offboardConsumer(instance, externalClusterClient)
+	}
+
 	return reconcile.Result{}, nil
 }
 
--- a/controllers/storagecluster/reconcile.go
+++ b/controllers/storagecluster/reconcile.go
@@ -376,9 +376,6 @@ func (r *StorageClusterReconciler) reconcilePhases(
 			}
 		}
 
-	} else if instance.Spec.ExternalStorage.StorageProviderKind == ocsv1.KindOCS {
-		// ocs to ocs external mode
-		objs = []resourceManager{}
 	} else {
 		// for external cluster, we have a different set of ensure functions
 		// preserve list order
--- a/controllers/storagecluster/reconcile.go
+++ b/controllers/storagecluster/reconcile.go
@@ -241,8 +241,9 @@ func (r *StorageClusterReconciler) validateStorageClusterSpec(instance *ocsv1.St
 		return err
 	}
 
-	if instance.Spec.ExternalStorage.Enable && instance.Spec.ExternalStorage.StorageProviderKind == ocsv1.KindOCS && instance.Spec.ExternalStorage.ConnectionString == "" {
-		err := fmt.Errorf("spec.externalStorage.connectionString can not be empty when spec.externalStorage.storageProviderKind is set to ocs")
+	if isExternalOCSProvider(instance) && (instance.Spec.ExternalStorage.ConnectionString == "" || instance.Spec.ExternalStorage.RequestedCapacity == nil) {
+		err := fmt.Errorf("spec.externalStorage.connectionString or spec.externalStorage.requestedCapacity" +
+			" can not be empty when spec.externalStorage.storageProviderKind is set to ocs")
 		r.Log.Error(err, "Failed to validate Spec for ExternalStorage", "StorageCluster", klog.KRef(instance.Namespace, instance.Name))
 		r.recorder.ReportIfNotPresent(instance, corev1.EventTypeWarning, statusutil.EventReasonValidationFailed, err.Error())
 		instance.Status.Phase = statusutil.PhaseError
--- a/controllers/storagecluster/reconcile.go
+++ b/controllers/storagecluster/reconcile.go
@@ -117,6 +117,7 @@ var validTopologyLabelKeys = []string{
 // +kubebuilder:rbac:groups=snapshot.storage.k8s.io,resources=volumesnapshots;volumesnapshotclasses,verbs=*
 // +kubebuilder:rbac:groups=template.openshift.io,resources=templates,verbs=*
 // +kubebuilder:rbac:groups=config.openshift.io,resources=infrastructures;networks,verbs=get;list;watch
+// +kubebuilder:rbac:groups=config.openshift.io,resources=clusterversions;networks,verbs=get;list;watch
 // +kubebuilder:rbac:groups=console.openshift.io,resources=consolequickstarts,verbs=*
 // +kubebuilder:rbac:groups=apiextensions.k8s.io,resources=customresourcedefinitions,verbs=get;list;watch;create;update
 // +kubebuilder:rbac:groups=route.openshift.io,resources=routes,verbs=*
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -39,6 +39,15 @@ rules:
 - apiGroups:
   - config.openshift.io
   resources:
+  - clusterversions
+  - networks
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - config.openshift.io
+  resources:
   - infrastructures
   - networks
   verbs:
--- a/controllers/storagecluster/uninstall_reconciler.go
+++ b/controllers/storagecluster/uninstall_reconciler.go
@@ -284,6 +284,7 @@ func (r *StorageClusterReconciler) deleteResources(sc *ocsv1.StorageCluster) (re
 	}
 
 	objs := []resourceManager{
+		&ocsExternalResources{},
 		&ocsProviderServer{},
 		&ocsNoobaaSystem{},
 		&ocsCephRGWRoutes{},
--- a/deploy/bundle/manifests/ocs-operator.clusterserviceversion.yaml
+++ b/deploy/bundle/manifests/ocs-operator.clusterserviceversion.yaml
@@ -1499,6 +1499,15 @@ spec:
         - apiGroups:
           - config.openshift.io
           resources:
+          - clusterversions
+          - networks
+          verbs:
+          - get
+          - list
+          - watch
+        - apiGroups:
+          - config.openshift.io
+          resources:
           - infrastructures
           - networks
           verbs:
--- a/deploy/csv-templates/ocs-operator.csv.yaml.in
+++ b/deploy/csv-templates/ocs-operator.csv.yaml.in
@@ -134,6 +134,15 @@ spec:
         - apiGroups:
           - config.openshift.io
           resources:
+          - clusterversions
+          - networks
+          verbs:
+          - get
+          - list
+          - watch
+        - apiGroups:
+          - config.openshift.io
+          resources:
           - infrastructures
           - networks
           verbs:


controllers/storagecluster/consumer.go
```diff
@@ -0,0 +1,177 @@
...
+)
+
+var (
+	externalOCSResources map[types.UID][]ExternalResource
+)
+
+type ocsConsumer struct{}
+
+func (o *ocsConsumer) ensureCreated(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) error {
+	if !(instance.Spec.ExternalStorage.Enable && instance.Spec.ExternalStorage.StorageProviderKind == ocsv1.KindOCS) {
```
sp98, line 22: isn't this condition checked already in `reconcile.go` when populating `resourceManager` ?

controllers/storagecluster/consumer.go
```diff
@@ -0,0 +1,177 @@
...
+		return nil
+	}
+
+	if instance.Status.ExternalStorage.ConsumerID == "" {
+		return o.onBoardConsumer(r, instance)
+	}
+
+	if instance.Status.ExternalStorage.GrantedCapacity.Cmp(*instance.Spec.ExternalStorage.RequestedCapacity) == -1 {
```
sp98, line 30: What is the requested capacity is less? (that is the user wan't to downscale)

controllers/storagecluster/consumer.go
iamniting: it is, but it is better to add here also.

controllers/storagecluster/consumer.go
iamniting: changed the code

controllers/storagecluster/consumer.go
```diff
@@ -0,0 +1,146 @@
...
+
+	if !instance.Status.ExternalStorage.GrantedCapacity.Equal(*instance.Spec.ExternalStorage.RequestedCapacity) {
+		return upscaleOrDownscaleConsumer(r, instance)
+	}
+
+	return nil
+}
+
+func (o *ocsConsumer) ensureDeleted(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) error {
+	if !(instance.Spec.ExternalStorage.Enable && instance.Spec.ExternalStorage.StorageProviderKind == ocsv1.KindOCS) {
```
oritwas, line 38: Why this case is only needed when deleting a consumer?

iamniting: This is not needed only for deletion/uninstall, We do have the same check for creation also. even if we move this to the offBoardConsumer it will still act the same.

controllers/storagecluster/consumer.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+func onBoardConsumer(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) error {
+
+	consumerClient, err := getConsumerClient(instance)
+	if err != nil {
+		return err
+	}
+
+	capacity := func(instance *ocsv1.StorageCluster) string {
```
oritwas, line 52: why a func?

controllers/storagecluster/consumer.go
```diff
@@ -0,0 +1,146 @@
...
+	if externalOCSResources[instance.UID] != nil {
+		return externalOCSResources[instance.UID], nil
+	}
+
+	consumerClient, err := getConsumerClient(instance)
+	if err != nil {
+		return nil, err
+	}
+
+	consumerID := instance.Status.ExternalStorage.ConsumerID
```
oritwas, line 101: Please use instance.Status.ExternalStorage.ConsumerID and remove the consumer ID variable that is not used

controllers/storagecluster/consumer.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+func getExternalConfig(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) ([]ExternalResource, error) {
+	if instance.Spec.ExternalStorage.StorageProviderKind == ocsv1.KindOCS {
+		return getExternalConfigFromProvider(r, instance)
+	}
+	return r.retrieveExternalSecretData(instance)
+}
+
+func parseExternalOcsJSONData(jsonBlob []byte) ([]ExternalResource, error) {
```
oritwas, line 125: As it is a short fun and only used once, lets move the code there

controllers/storagecluster/consumer_grpc_error_handling.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	corev1 "k8s.io/api/core/v1"
+
+	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v1"
+)
+
+// handleOnBoardConsumerErrors returns the bool which tells to requeue or not
+func (r *StorageClusterReconciler) handleOnBoardConsumerErrors(instance *ocsv1.StorageCluster, err error) bool {
```
oritwas, line 12: Why is this a function and not part of onBoardConsumer?

controllers/storagecluster/consumer.go
oritwas: This can be a function isExternalOCS, and we can check on every function.

controllers/storagecluster/reconcile.go
```diff
@@ -375,7 +375,14 @@ func (r *StorageClusterReconciler) reconcilePhases(
 
 	} else if instance.Spec.ExternalStorage.StorageProviderKind == ocsv1.KindOCS {
 		// ocs to ocs external mode
-		objs = []resourceManager{}
+		objs = []resourceManager{
```
oritwas, line 378: why is this change needed?

services/provider/server/api_updater.go
```diff
@@ -0,0 +1,86 @@
...
+	ocsv1alpha1 "github.com/red-hat-storage/ocs-operator/api/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/klog"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+)
+
+type APIUpdater interface {
```
oritwas, line 16: Let call it ConsomerAPI

oritwas: Actually why not call it provider like the protobuf?

services/provider/server/api_updater.go
```diff
@@ -0,0 +1,86 @@
...
+	crClient, err := client.New(config, client.Options{Scheme: scheme})
+	if err != nil {
+		klog.Error(err, "failed to create controller-runtime client")
+		return &cAPIUpdater{}, err
+	}
+
+	apiClient := &cAPIUpdater{
+		client: crClient,
+	}
+	return apiClient, nil
```
oritwas, line 55: why not return &cAPIUpdater { client: crClient, }

services/provider/server/api_updater.go
```diff
@@ -0,0 +1,86 @@
...
+
+	config, err := config.GetConfig()
+	if err != nil {
+		klog.Error(err, "failed to get rest.config")
+		return &cAPIUpdater{}, err
+	}
+	crClient, err := client.New(config, client.Options{Scheme: scheme})
+	if err != nil {
+		klog.Error(err, "failed to create controller-runtime client")
+		return &cAPIUpdater{}, err
```
oritwas, line 49: should it not be return nil, err ?

services/provider/server/auth_interceptor.go
```diff
@@ -0,0 +1,68 @@
...
+
+	"github.com/red-hat-storage/ocs-operator/services/provider/common"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"k8s.io/klog"
+)
+
+// AuthInterceptor is a server interceptor for authenticating access token and StorageConsumer UID
+type AuthInterceptor struct {
```
oritwas, line 14: How about Authenticator?

oritwas, line 14: why not sue authManager directly?

oritwas, line 14: Let's merge it with AuthManager

services/provider/server/consumer_cache.go
```diff
@@ -0,0 +1,68 @@
...
+
+// Consumer is the storage consumer
+type Consumer struct {
+	Name  string
+	ID    string
+	Token string
+}
+
+// NewConsumer returns a new storage consumer
+func NewConsumer(name, id, token string) *Consumer {
```
oritwas, line 16: Do we need this method?

controllers/storagecluster/consumer.go
iamniting: created a func

controllers/storagecluster/consumer.go
iamniting: I created a func in this file itself, as this is a specific case where we use a default value (1Ti) if none is provided.

controllers/storagecluster/consumer.go
iamniting: done

controllers/storagecluster/consumer.go
iamniting: done

services/provider/server/api_updater.go
iamniting: All these service related changes came from the [@sp98's PR](https://github.com/red-hat-storage/ocs-operator/pull/1430)

controllers/storagecluster/consumer_grpc_error_handling.go
iamniting: I think It makes it easy to read and maintain the code. Now we have 4 func for 4 different calls for error handling.

controllers/storagecluster/consumer.go
```diff
@@ -0,0 +1,189 @@
...
+func onBoardConsumer(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) error {
+
+	grpcConnection, consumerClient, err := getConsumerClient(instance)
+	if err != nil {
+		return err
+	}
+	defer grpcConnection.Close()
+
+	// TODO: think about naming convention
+	response, err := consumerClient.OnboardConsumer("my-ocs", getExternalStorageRequestedCapacity(instance))
```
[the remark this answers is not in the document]
iamniting: It is handled in a next line.

controllers/storagecluster/consumer.go
```diff
@@ -0,0 +1,189 @@
...
+		return externalOCSResources[instance.UID], nil
+	}
+
+	grpcConnection, consumerClient, err := getConsumerClient(instance)
+	if err != nil {
+		return nil, err
+	}
+	defer grpcConnection.Close()
+
+	config, err := consumerClient.GetStorageConfig(instance.Status.ExternalStorage.ConsumerID)
```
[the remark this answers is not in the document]
iamniting: It is handled in a next line

controllers/storagecluster/consumer.go
sp98: but why check twice?

controllers/storagecluster/consumer.go
```diff
@@ -0,0 +1,189 @@
...
+		return nil, err
+	}
+
+	externalOCSResources[instance.UID] = externalResources
+
+	return externalOCSResources[instance.UID], nil
+}
+
+// getConsumerClient returns the *providerClient.ConsumerClient
+func getConsumerClient(instance *ocsv1.StorageCluster) (*grpc.ClientConn, *providerClient.ConsumerClient, error) {
```
sp98, line 152: Lets rename it to `getProviderClient`.   It was wrong in my PR as well. I've updated it.  Please rename in other places as well.

controllers/storagecluster/consumer_grpc_error_handling.go
```diff
@@ -0,0 +1,184 @@
...
+
+	// internal error, do requeue
+	if s.Code() == codes.Internal {
+		return err
+	}
+
+	// token not valid, do not requeue
+	if s.Code() == codes.InvalidArgument {
+		reason := "Token is invalid verify the token again or contact provider admin"
+		r.Log.Error(err, "ocs-to-ocs:onboard: "+reason)
```
sp98, line 48: I don't think we want to say `ocs-to-ocs` in logs.

controllers/storagecluster/consumer_grpc_error_handling.go
```diff
@@ -0,0 +1,184 @@
...
+		r.recorder.ReportIfNotPresent(instance, corev1.EventTypeWarning, "TokenInvalid", reason)
+		isErrorResolvable[instance.UID] = false
+		return err
+	}
+
+	// token already used, do not requeue
+	if s.Code() == codes.AlreadyExists {
+		reason := "Token is already used contact provider admin for a new token"
```
sp98, line 56:
```suggestion
		reason := "Token is already used. Contact provider admin for a new token"
```

controllers/storagecluster/consumer_grpc_error_handling.go
```diff
@@ -0,0 +1,184 @@
...
+	// But if it returns error Reconcile will get requeued, We can not determine
+	// if the error is resolvable or not either we need to requeue of not
+	// isErrorResolvable will help us for the same if isErrorResolvable[instance.UID] == false we won't requeue
+	isErrorResolvable = map[types.UID]bool{}
+
+	// requeueAfter holds info about requeue time
+	requeueAfter = map[types.UID]time.Duration{}
+)
+
+// handleOnBoardConsumerErrors returns the bool which tells to requeue or not
```
sp98, line 26: no bool is returned in this method.

controllers/storagecluster/consumer.go
```diff
@@ -0,0 +1,189 @@
...
+		return nil
+	}
+
+	if instance.Status.ExternalStorage.ConsumerID == "" {
+		return onBoardConsumer(r, instance)
+	}
+
+	if instance.Spec.ExternalStorage.RequestedCapacity != nil &&
+		!instance.Status.ExternalStorage.GrantedCapacity.Equal(*instance.Spec.ExternalStorage.RequestedCapacity) {
+		return upscaleOrDownscaleConsumer(r, instance)
```
sp98, line 38: How about `updateConsumerCapacity` ?

controllers/storagecluster/consumer.go
```diff
@@ -0,0 +1,189 @@
...
+	instance.Status.ExternalStorage.ConsumerID = response.StorageConsumerUUID
+	instance.Status.ExternalStorage.GrantedCapacity = resource.MustParse(response.GrantedCapacity)
+
+	return nil
+}
+
+// offBoardConsumer makes an API call to the provider cluster for offboarding
+func offBoardConsumer(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) error {
+
+	grpcConnection, consumerClient, err := getConsumerClient(instance)
```
sp98, line 83: how about creating client  and closing the connection only once inside the `ensureCreated` func.   The client can be passed in as an argument or can be added to StorageClusterReconciler Struct.

controllers/storagecluster/consumer_grpc_error_handling.go
```diff
@@ -0,0 +1,184 @@
...
+
+	// internal error, do requeue
+	if s.Code() == codes.Internal {
+		requeueAfter[instance.UID] = 5 * time.Second
+		return err
+	}
+
+	// UID or capacity not valid, do not requeue
+	if s.Code() == codes.InvalidArgument {
+		reason := "StorageConsumer UID or requested capacity is not valid contact provider admin"
```
sp98, line 125:
```suggestion
		reason := "StorageConsumer UID or requested capacity is not valid. Contact the provider admin"
```

controllers/storagecluster/consumer_grpc_error_handling.go
```diff
@@ -0,0 +1,184 @@
...
+		r.Log.Error(err, "ocs-to-ocs:updateCapacity: "+reason)
+		r.recorder.ReportIfNotPresent(instance, corev1.EventTypeWarning, "UIDorCapacityInvalid", reason)
+		isErrorResolvable[instance.UID] = false
+		return err
+	}
+
+	// UID not found, do not requeue
+	if s.Code() == codes.NotFound {
+		reason := "StorageConsumer UID not found contact provider admin"
```
sp98, line 134:
```suggestion
		reason := "StorageConsumer UID not found. Contact the provider admin"
```

sp98 (changes requested):

controllers/storagecluster/consumer.go
iamniting: done

controllers/storagecluster/consumer_grpc_error_handling.go
iamniting: Actually, I want to differentiate b/w logs of ocs-to-ocs that's why I did this. maybe we can think of some other keyword.

controllers/storagecluster/consumer_grpc_error_handling.go
iamniting: done

controllers/storagecluster/consumer_grpc_error_handling.go
iamniting: changed the comment

controllers/storagecluster/consumer.go
iamniting: done

controllers/storagecluster/consumer_grpc_error_handling.go
iamniting: done

controllers/storagecluster/consumer_grpc_error_handling.go
iamniting: done

controllers/storagecluster/consumer.go
iamniting: It is not just this `ensureCreated` where these API calls are used, these are used at other places also, in that case, I need to write the logic at other `ensureCreated` too.

I don't want to make it part of the reconciler, because we do not need these calls very frequently and do not want to get the client every time.

controllers/storagecluster/consumer.go
```diff
@@ -0,0 +1,188 @@
...
+	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v1"
+	providerClient "github.com/red-hat-storage/ocs-operator/services/provider/client"
+	"github.com/red-hat-storage/ocs-operator/services/provider/common"
+)
+
+var (
+	// externalOCSResources will hold the ExternalResources for storageclusters
+	// ExternalResources can be accessible using the UID of an storagecluster
+	// getExternalConfigFromProvider func sets value for this once and use same again
+	externalOCSResources map[types.UID][]ExternalResource
```
kesavanvt, line 20: Empty map is defined, required to initialize 

`externalOCSResources = map[types.UID][]ExternalResource{}`

iamniting: made changes

controllers/storagecluster/consumer.go
sp98: IMO, the current code makes it impossible/difficult to unit test the onBoard, offboard and updatedCapacity functionality. The client should be passed in as a dependency via `ocsConsumer` struct.

controllers/storagecluster/consumer.go
```diff
@@ -0,0 +1,168 @@
...
+	// externalOCSResources will hold the ExternalResources for storageclusters
+	// ExternalResources can be accessible using the UID of an storagecluster
+	// getExternalConfigFromProvider func sets value for this once and use same again
+	externalOCSResources = map[types.UID][]ExternalResource{}
+)
+
+type ocsConsumer struct{}
+
+// ensureCreated performs onboarding and update capacity
+func (o *ocsConsumer) ensureCreated(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) error {
```
oritwas, line 26: Let's separate it two function one that ensures the consumer is created and one that updates the capacity.

controllers/storagecluster/consumer.go
```diff
@@ -0,0 +1,168 @@
...
+
+	if !isExternalOCSProvider(instance) {
+		return nil
+	}
+
+	if instance.Status.ExternalStorage.ConsumerID == "" {
+		return onBoardConsumer(r, instance)
+	}
+
+	if instance.Spec.ExternalStorage.RequestedCapacity != nil &&
```
oritwas, line 36: How can a consumer be onboarded and not have RequestedCapacity?

controllers/storagecluster/consumer.go
```diff
@@ -0,0 +1,168 @@
...
+
+	response, err := r.providerClient.UpdateCapacity(
+		instance.Status.ExternalStorage.ConsumerID,
+		getExternalStorageRequestedCapacity(instance))
+	err = r.handleUpdateConsumerCapacityErrors(instance, err)
+	if err != nil {
+		return err
+	}
+
+	if response.GrantedCapacity == "" {
```
oritwas, line 94: How can this happen if updateCapacity was successful?

controllers/storagecluster/consumer.go
```diff
@@ -0,0 +1,168 @@
...
+		return externalOCSResources[instance.UID], nil
+	}
+
+	config, err := r.providerClient.GetStorageConfig(instance.Status.ExternalStorage.ConsumerID)
+	err = r.handleGetStorageConfigErrors(instance, err)
+	if err != nil {
+		return nil, err
+	}
+
+	var externalResources []ExternalResource
```
oritwas, line 120: How about GetStorageConfig returned ExternalResource and not a json?

controllers/storagecluster/consumer.go
```diff
@@ -0,0 +1,168 @@
...
+		return nil, err
+	}
+
+	externalOCSResources[instance.UID] = externalResources
+
+	return externalOCSResources[instance.UID], nil
+}
+
+// getConsumerClient returns the *providerClient.ConsumerClient
+func getProviderClient(instance *ocsv1.StorageCluster) (*grpc.ClientConn, *providerClient.OCSProviderClient, error) {
```
oritwas, line 132: shouldn't it called newProviderClient or CreateProviderClient?

oritwas: Why is not the grpc connection part of the providerClient?

controllers/storagecluster/reconcile.go
```diff
@@ -324,6 +327,11 @@ func (r *StorageClusterReconciler) reconcilePhases(
 			if err := r.deleteResources(instance); err != nil {
 				r.Log.Info("Uninstall in progress.", "Status", err)
 				r.recorder.ReportIfNotPresent(instance, corev1.EventTypeWarning, statusutil.EventReasonUninstallPending, err.Error())
+
+				if !isErrorResolvable[instance.UID] {
```
oritwas, line 331: This change is not related to OCS to OCS. Should it be in a separate PR?

controllers/storagecluster/reconcile.go
```diff
@@ -410,6 +433,15 @@ func (r *StorageClusterReconciler) reconcilePhases(
 			message := fmt.Sprintf("Error while reconciling: %v", returnErr)
 			statusutil.SetErrorCondition(&instance.Status.Conditions, reason, message)
 			instance.Status.Phase = statusutil.PhaseError
+
+			if !isErrorResolvable[instance.UID] {
```
oritwas, line 437: see my previous comment

services/provider/client/client.go
```diff
@@ -2,49 +2,55 @@ package client
...
-// NewConsumerClient creates a ConsumerClient to talk to OCS consumer server
-func NewConsumerClient(cc *grpc.ClientConn, timeout time.Duration) *ConsumerClient {
-	return &ConsumerClient{Client: pb.NewOCSProviderClient(cc), timeout: timeout}
+// NewProviderClient creates a client to talk to OCS provider server
+func NewProviderClient(cc *grpc.ClientConn, timeout time.Duration) *OCSProviderClient {
+	return &OCSProviderClient{Client: pb.NewOCSProviderClient(cc), timeout: timeout}
+}
+
+// NewMockProviderClient creates a client to mock the OCS provider server api calls
+func NewMockProviderClient(errorFlow common.MockErrorFlow) *OCSProviderClient {
```
oritwas, line 24: Why do we need mock provider client?

services/provider/client/client_mock.go
```diff
@@ -0,0 +1,87 @@
...
+	"encoding/json"
+
+	"github.com/red-hat-storage/ocs-operator/services/provider/common"
+	pb "github.com/red-hat-storage/ocs-operator/services/provider/pb"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+type MockOCSProviderClient struct {
```
oritwas, line 14: Why do we need a MOCK provider client?

controllers/storagecluster/consumer.go
iamniting: Actually, everything will be called through the `ensureCreated` only as we are implementing the ` resourceManager` interface and we already have 2 different funcs for onboard and update capacity.

https://github.com/red-hat-storage/ocs-operator/blob/main/controllers/storagecluster/reconcile.go#L33

controllers/storagecluster/consumer.go
iamniting: As per the design we are taking 1TB as a default value if a consumer is not specifying any value.

controllers/storagecluster/consumer.go
iamniting: It should not, I am just making sure that the server has set the value.

controllers/storagecluster/reconcile.go
iamniting: But OCS to OCS is the one that requires this functionality and I think creating a separate PR will delay things.

services/provider/client/client.go
iamniting: Actually, we are creating this for testing purposes as our API server and controller work is blocked for now.

services/provider/client/client_mock.go
iamniting: Actually, we are creating this for testing purposes as our API server and controller work is blocked for now.

controllers/storagecluster/consumer.go
iamniting: > shouldn't it called newProviderClient or CreateProviderClient?

This can be done

> Why is grpc connection not part of the providerClient?

@sp98 Can you pls answer this one?

controllers/storagecluster/consumer.go
iamniting: No it does not returns `ExternalResource` it returns `Data []byte `protobuf:"bytes,1,opt,name=data,proto3" json:"data,omitempty"`` in the response

controllers/storagecluster/external_ocs.go
```diff
@@ -0,0 +1,284 @@
...
+	}
+
+	// internal error, do requeue
+	if s.Code() == codes.Internal {
+		return reconcile.Result{RequeueAfter: time.Second * 5}, err
+	}
+
+	// token not valid, do not requeue
+	if s.Code() == codes.InvalidArgument {
+		reason := "Token is invalid verify the token again or contact the provider admin"
```
sp98, line 154:
```suggestion
		reason := "Token is invalid. Verify the token again or contact the provider admin"
```

controllers/storagecluster/external_ocs.go
```diff
@@ -0,0 +1,284 @@
...
+		q := resource.MustParse("1Ti")
+		return &q
+	}
+	return instance.Spec.ExternalStorage.RequestedCapacity
+}
+
+// newProviderClient returns the *grpc.ClientConn and *providerClient.ConsumerClient
+func newProviderClient(instance *ocsv1.StorageCluster) (*grpc.ClientConn, *externalClient.OCSProviderClient, error) {
+
+	// TODO: implement logic for seprating adress and token
```
sp98, line 46:
```suggestion
	// TODO: implement logic for separating address and token
```

controllers/storagecluster/external_ocs.go
```diff
@@ -0,0 +1,284 @@
...
+		instance.Spec.ExternalStorage.ConnectionString, []grpc.DialOption{})
+	if err != nil {
+		return nil, nil, err
+	}
+
+	consumerClient := externalClient.NewProviderClient(grpcConnection, time.Second*10)
+	return grpcConnection, consumerClient, nil
+}
+
+// onBoardConsumer makes an API call to the provider cluster for onboarding
```
sp98, line 58:
```suggestion
// onBoardConsumer makes an API call to the external storage provider cluster for onboarding
```

controllers/storagecluster/external_ocs.go
```diff
@@ -0,0 +1,284 @@
...
+func onBoardConsumer(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) (reconcile.Result, error) {
+
+	clusterVersion := &configv1.ClusterVersion{}
+	err := r.Client.Get(context.TODO(), types.NamespacedName{Name: "version"}, clusterVersion)
+	if err != nil {
+		return reconcile.Result{}, err
```
sp98, line 64: would be nice to send more details with the error (using `fmt.Errorf`)

controllers/storagecluster/external_ocs.go
```diff
@@ -0,0 +1,284 @@
...
+	}
+
+	instance.Status.ExternalStorage.ConsumerID = response.StorageConsumerUUID
+	instance.Status.ExternalStorage.GrantedCapacity = resource.MustParse(response.GrantedCapacity)
+
+	return reconcile.Result{}, nil
+}
+
+// offBoardConsumer makes an API call to the provider cluster for offboarding
+func offBoardConsumer(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) (reconcile.Result, error) {
```
sp98, line 87: Lets change to `offboardConsumer` and `onboardConsumer` (`b` in lowercase) everywhere.

controllers/storagecluster/external_ocs.go
```diff
@@ -0,0 +1,284 @@
...
+
+	_, err := r.externalClient.OffboardConsumer(instance.Status.ExternalStorage.ConsumerID)
+	return r.handleOffboardConsumerErrors(instance, err)
+}
+
+// upscaleOrDownscaleConsumer makes an API call to the provider cluster to update the capacity
```
sp98, line 93:
```suggestion
// updateConsumerCapacity makes an API call to the external storage provider cluster to update the capacity
```

sp98 (changes requested):
few more nits

controllers/storagecluster/external_ocs.go
iamniting: done

controllers/storagecluster/external_ocs.go
iamniting: done

controllers/storagecluster/external_ocs.go
iamniting: done

controllers/storagecluster/external_ocs.go
iamniting: adding log now

controllers/storagecluster/external_ocs.go
iamniting: done

controllers/storagecluster/external_ocs.go
iamniting: done

controllers/storagecluster/external_ocs.go
```diff
@@ -0,0 +1,285 @@
...
+// isExternalOCSProvider returns true if it is ocs to ocs ExternalStorage consumer cluster
+func isExternalOCSProvider(instance *ocsv1.StorageCluster) bool {
+	return instance.Spec.ExternalStorage.Enable && instance.Spec.ExternalStorage.StorageProviderKind == ocsv1.KindOCS
+}
+
+// getExternalStorageRequestedCapacity returns requestedCapacity,
+// if not set it returns a default value 1Ti
+func getExternalStorageRequestedCapacity(instance *ocsv1.StorageCluster) *resource.Quantity {
+	if instance.Spec.ExternalStorage.RequestedCapacity == nil {
+		q := resource.MustParse("1Ti")
```
iamniting, line 37: remove the default value and enforce it from the controller.

controllers/storagecluster/external_ocs.go
```diff
@@ -0,0 +1,285 @@
...
+		return &q
+	}
+	return instance.Spec.ExternalStorage.RequestedCapacity
+}
+
+// newProviderClient returns the *grpc.ClientConn and *providerClient.ConsumerClient
+func newProviderClient(instance *ocsv1.StorageCluster) (*grpc.ClientConn, *externalClient.OCSProviderClient, error) {
```
sp98, line 44: call it `ExternalClusterClient` or `ExternalModeClient`

oritwas, line 44: Why do we need to return the grpc.ClientConn from this function?
We store it in the externalClusterClient, it should be enough to return it.

iamniting: This is required for closing the connection.

services/provider/client/client.go
```diff
@@ -2,49 +2,50 @@ package client
...
-		Token:    token,
-		Capacity: capacity,
+func (cc *OCSProviderClient) OnboardConsumer(token, name, capacity string) (*pb.OnboardConsumerResponse, error) {
+	req := &pb.OnboardConsumerRequest{
+		Token:        token,
+		ConsumerName: name,
+		Capacity:     capacity,
 	}
 
 	ctx, cancel := context.WithTimeout(context.Background(), cc.timeout)
```
nb-ohad, line 41: Why are we creating a new context.Background() for each call? It will be preferable if all calls during a single reconcile run use the same context. This is true for the entire OCS operator reconcile run. Not only the new part.

controllers/storagecluster/reconcile.go
```diff
@@ -389,6 +389,19 @@ func (r *StorageClusterReconciler) reconcilePhases(
...
 		}
+
+		if isExternalOCSProvider(instance) {
+			grpcConnection, externalClusterClient, err := newExternalClusterClient(instance)
+			if err != nil {
+				return reconcile.Result{}, err
+			}
+			r.externalClusterClient = externalClusterClient
+			defer func() {
+				grpcConnection.Close()
```
oritwas, line 400: This should be moved into newExternalClusterClient function

controllers/storagecluster/external_ocs.go
```diff
@@ -0,0 +1,221 @@
...
+
+	consumerClient := externalClient.NewProviderClient(grpcConnection, time.Second*10)
+	return grpcConnection, consumerClient, nil
+}
+
+// onboardConsumer makes an API call to the external storage provider cluster for onboarding
+func onboardConsumer(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) (reconcile.Result, error) {
+
+	clusterVersion := &configv1.ClusterVersion{}
+	err := r.Client.Get(context.TODO(), types.NamespacedName{Name: "version"}, clusterVersion)
```
nb-ohad, line 52: We should not be creating new context objects for each client call. there should be a single context object for an entire reconcile run that is created on the main reconcile function and provided here from the outside.

 Also please use context.Background() to create that context. and not context.TODO().

controllers/storagecluster/external_ocs.go
```diff
@@ -0,0 +1,221 @@
...
+	if err != nil {
+		s, ok := status.FromError(err)
+		if !ok {
+			// This is not gRPC error. The operation must have failed before gRPC
+			// method was called, otherwise we would get gRPC error.
+			return reconcile.Result{}, err
+		}
+		r.logGrpcErrorAndReportEvent(instance, "OnboardConsumer", err, s.Code())
+		return reconcile.Result{}, err
+	}
```
nb-ohad, line 72: Why do we care about the origin of the error? 

But if you really od care pelase use the reversed pattern as it is shorter and much easier to read:
```suggestion
	if err != nil {
		if s, ok := status.FromError(err), ok {
			 r.logGrpcErrorAndReportEvent(instance, "OnboardConsumer", err, s.Code())
		}
		return reconcile.Result{}, err
	}
```

This comment should apply to all of the instances of the original pattern below.

controllers/storagecluster/external_ocs.go
```diff
@@ -0,0 +1,221 @@
...
+	if err != nil {
+		s, ok := status.FromError(err)
+		if !ok {
+			// This is not gRPC error. The operation must have failed before gRPC
+			// method was called, otherwise we would get gRPC error.
+			return reconcile.Result{}, err
+		}
+		r.logGrpcErrorAndReportEvent(instance, "UpdateCapacity", err, s.Code())
+		return reconcile.Result{}, err
+	}
```
nb-ohad, line 119: same as the previous comment

controllers/storagecluster/external_ocs.go
```diff
@@ -0,0 +1,221 @@
...
+
+	response, err := r.externalClusterClient.GetStorageConfig(instance.Status.ExternalStorage.ConsumerID)
+	if err != nil {
+		s, ok := status.FromError(err)
+		if !ok {
+			// This is not gRPC error. The operation must have failed before gRPC
+			// method was called, otherwise we would get gRPC error.
+			return nil, reconcile.Result{}, err
+		}
+		r.logGrpcErrorAndReportEvent(instance, "GetStorageConfig", err, s.Code())
```
nb-ohad, line 143: same as the previous comment

controllers/storagecluster/external_ocs.go
```diff
@@ -0,0 +1,221 @@
...
+		// storage consumer is not ready yet, requeue after some time
+		if s.Code() == codes.Unavailable {
+			return nil, reconcile.Result{RequeueAfter: time.Second * 5}, nil
+		}
+
+		return nil, reconcile.Result{}, err
+	}
+
+	var externalResources []ExternalResource
+	err = json.Unmarshal(response.Data, &externalResources)
```
nb-ohad, line 154: Why do we need to unmarshal the data? I think we agreed that we are going to return the response a list of external resource objects, not as a blob of data.

controllers/storagecluster/external_ocs.go
```diff
@@ -0,0 +1,221 @@
...
+			},
+		},
+	}
+
+	eventDetails, ok := cases[grpcCallName][errCode]
+	if ok {
+		r.Log.Error(err, "External-OCS:"+grpcCallName+":"+eventDetails["msg"])
+		r.recorder.ReportIfNotPresent(instance,
+			eventDetails["eventType"], eventDetails["eventReason"], eventDetails["msg"])
+	}
```
nb-ohad, line 220: If this is not a global const map that the call sites utilize to figure out what to log then it would be preferable to just have a nested switch statement instead (or if statements). 
The current approach is just putting strain on the garbage collector because this very big map is created for each call and immediately after is thrown aside. It is also less readable (just look at the type definition).

I am valuing the simplest solution as it is always more readable and more maintainable, and in this case it is even less amount of code.


```suggestion
	msg := ""
	eventReason := ""
        eventType := ""

	if grpcCallName == OnboardConsumer {
		if errCode == codes.InvalidArgument {
			msg = "Token is invalid. Verify the token again or contact the provider admin";
			eventReason =  "TokenInvalid";
			eventType =  corev1.EventTypeWarning;
		} else if errCode == AlreadyExists {
			msg = "Token is already used. Contact provider admin for a new token";
			reason = "TokenAlreadyUsed";
			eventType = corev1.EventTypeWarning;
		}
		...
	} else if grpcCallName == OffboardConsumer {
	        ....
	} else if ....

        if msg != "" {
            r.Log.Error(err, "External-OCS:"+grpcCallName+":"+msg])
            r.recorder.ReportIfNotPresent(instance, eventType, eventReason,msg)
        }
```

controllers/storagecluster/reconcile.go
```diff
@@ -241,8 +241,11 @@ func (r *StorageClusterReconciler) validateStorageClusterSpec(instance *ocsv1.St
 		return err
 	}
 
-	if instance.Spec.ExternalStorage.Enable && instance.Spec.ExternalStorage.StorageProviderKind == ocsv1.KindOCS && instance.Spec.ExternalStorage.ConnectionString == "" {
-		err := fmt.Errorf("spec.externalStorage.connectionString can not be empty when spec.externalStorage.storageProviderKind is set to ocs")
+	if instance.Spec.ExternalStorage.Enable && instance.Spec.ExternalStorage.StorageProviderKind == ocsv1.KindOCS &&
+		(instance.Spec.ExternalStorage.ConnectionString == "" || instance.Spec.ExternalStorage.RequestedCapacity == nil) {
```
nb-ohad, line 245: You can use the `isExternalOCSProvider` helper to make the condition more clear

```suggestion
	if isExternalOCSProvider(instance) && (instance.Spec.ExternalStorage.ConnectionString == "" || instance.Spec.ExternalStorage.RequestedCapacity == nil) {
```

nb-ohad (changes requested):

controllers/storagecluster/external_ocs.go
```diff
@@ -0,0 +1,221 @@
...
+
+	consumerClient := externalClient.NewProviderClient(grpcConnection, time.Second*10)
+	return grpcConnection, consumerClient, nil
+}
+
+// onboardConsumer makes an API call to the external storage provider cluster for onboarding
+func onboardConsumer(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) (reconcile.Result, error) {
+
+	clusterVersion := &configv1.ClusterVersion{}
```
kesavanvt, line 51: add necessary rbac for clusterversion

controllers/storagecluster/external_ocs.go
```diff
@@ -0,0 +1,221 @@
...
+	return instance.Spec.ExternalStorage.Enable && instance.Spec.ExternalStorage.StorageProviderKind == ocsv1.KindOCS
+}
+
+// newExternalClusterClient returns the *grpc.ClientConn and *providerClient.ConsumerClient
+func newExternalClusterClient(instance *ocsv1.StorageCluster) (*grpc.ClientConn, *externalClient.OCSProviderClient, error) {
+
+	// TODO: implement logic for separating address and token
+
+	grpcConnection, err := externalClient.NewGRPCConnection(
+		instance.Spec.ExternalStorage.ConnectionString, []grpc.DialOption{})
```
kesavanvt, line 39: pass required dial options , for insecure connection and to block until the connection is made.
```
		instance.Spec.ExternalStorage.ConnectionString, []grpc.DialOption{
			grpc.WithInsecure(),
			grpc.WithBlock(),
			grpc.WithTimeout(time.Second * 10),
		})
```

controllers/storagecluster/reconcile.go
```diff
@@ -389,6 +389,19 @@ func (r *StorageClusterReconciler) reconcilePhases(
 			&ocsSnapshotClass{},
 			&ocsNoobaaSystem{},
 		}
+
+		if isExternalOCSProvider(instance) {
+			grpcConnection, externalClusterClient, err := newExternalClusterClient(instance)
```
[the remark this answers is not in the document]
iamniting: that's not possible, because in that case connection will be closed in the newExternalClusterClient itself and we wont be able to make calls.

controllers/storagecluster/external_ocs.go
```diff
@@ -0,0 +1,221 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	consumerClient := externalClient.NewProviderClient(grpcConnection, time.Second*10)
+	return grpcConnection, consumerClient, nil
```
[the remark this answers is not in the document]
iamniting: to close the connection

services/provider/client/client.go
iamniting: This is from the mock PR

controllers/storagecluster/external_ocs.go
iamniting: I can use context.Background() but won't be able to pass the one coming from reconcile.

controllers/storagecluster/reconcile.go
iamniting: It can not be moved as defer will be called exactly after newExternalClusterClient is done and we will lost the connection

controllers/storagecluster/external_ocs.go
iamniting: added context.Background()

controllers/storagecluster/external_ocs.go
iamniting: done

controllers/storagecluster/external_ocs.go
iamniting: done

controllers/storagecluster/external_ocs.go
iamniting: done

controllers/storagecluster/external_ocs.go
iamniting: So we are using ExternalResource on the server-side for all operations and converting it into bytes just before sending the response to the client, that's why we need to unmarshal

controllers/storagecluster/external_ocs.go
iamniting: done

controllers/storagecluster/reconcile.go
iamniting: done

controllers/storagecluster/external_ocs.go
iamniting: done

controllers/storagecluster/external_ocs.go
iamniting: done

controllers/storagecluster/cephcluster.go
```diff
@@ -110,12 +110,7 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
 	var cephCluster *rookCephv1.CephCluster
 	// Define a new CephCluster object
 	if sc.Spec.ExternalStorage.Enable {
-		extRArr, err := r.retrieveExternalSecretData(sc)
-		if err != nil {
-			r.Log.Error(err, "Could not retrieve the External Cluster details.",
-				"CephCluster", klog.KRef(sc.Namespace, sc.Name))
-			return reconcile.Result{}, err
-		}
+		extRArr := externalOCSResources[sc.UID]
```
oritwas, line 113: Don't we need to check the err like before?

controllers/storagecluster/external_ocs.go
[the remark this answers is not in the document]
oritwas: why not close it inside of newProviderClient func?

controllers/storagecluster/cephcluster.go
iamniting: no, we don't need to as we are setting the value in the map and will get errors while setting the value itself and returning it.

controllers/storagecluster/external_ocs.go
nb-ohad: But that is my question, why are we doing it? what was the reason to convert into bytes in the first place where we have a structured response

controllers/storagecluster/external_ocs.go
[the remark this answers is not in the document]
iamniting: > why not close it inside of newProviderClient func in case of an error?

no need to close in that case, as we did get the grpc connection

> I would add a deleteProviderClient that will do the clean up and close the grpc connection (i.e for client offboarding). Another option is to close it inside of client.offboardConsumer

client connection is required for every reconcile and we are not doing a offboard every reconcile and we will have a lot of open connections.

controllers/storagecluster/cephcluster.go
oritwas: OK, but don't we need to check externalOCSResources[sc.UID] is not nil?

iamniting: we can do that, but it should not be nil

iamniting: done

controllers/storagecluster/external_ocs.go
nb-ohad: I think the main ask was to have a single context for all of the calls, it is the important part of the ask.
Just accept it as a parameter

services/provider/client/client.go
nb-ohad: Ok

controllers/storagecluster/external_ocs.go
iamniting: In that case, I have to change the interface again to pass the context and use the context as an argument, I think we can take this one later.

controllers/storagecluster/external_ocs.go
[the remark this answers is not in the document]
nb-ohad: @iamniting I think you missing the point, you do not need to return the connection just provide a function the closes the connection on the client and defer it. I agree with @oritwas that the GRPC code and structures need to be encapsulated inside the client.

controllers/storagecluster/consumer.go
nb-ohad: Not relevant anymore

controllers/storagecluster/reconcile.go
oritwas: But don't we need the grpcConnection later?

controllers/storagecluster/reconcile.go
```diff
@@ -389,6 +388,16 @@ func (r *StorageClusterReconciler) reconcilePhases(
 			&ocsSnapshotClass{},
 			&ocsNoobaaSystem{},
 		}
+
```
iamniting, line 391: move it to the ensure created and ensure deleted

controllers/storagecluster/external_ocs.go
```diff
@@ -0,0 +1,248 @@
...
+	}
+
+	var externalResources []ExternalResource
+
+	for _, eResource := range response.ExternalResource {
+
+		data := map[string]string{}
+		err = json.Unmarshal(eResource.Data, &data)
+		if err != nil {
+			return nil, reconcile.Result{}, err
```
nb-ohad, line 185: Would suggest having `log.Error` to capture which resource (kind + name) had the problem with unmarshaling

controllers/storagecluster/external_resources.go
```diff
@@ -230,9 +230,49 @@ func (r *StorageClusterReconciler) newExternalCephObjectStoreInstances(
...
-	if r.sameExternalSecretData(instance) {
-		return reconcile.Result{}, nil
+
+	if isExternalOCSProvider(instance) {
+
+		externalClusterClient, err := r.newExternalClusterClient(instance)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		r.externalClusterClient = externalClusterClient
```
nb-ohad, line 240: We do not need `r.externalClusterClient` please remove

controllers/storagecluster/external_resources.go
```diff
@@ -230,9 +230,49 @@ func (r *StorageClusterReconciler) newExternalCephObjectStoreInstances(
...
+			}
+		}
+
+		if externalOCSResources[instance.UID] == nil {
+			externalConfig, res, err := r.getExternalConfigFromProvider(instance)
+			if err != nil || !res.IsZero() {
+				return res, err
+			}
+			externalOCSResources[instance.UID] = externalConfig
+		}
```
nb-ohad, line 261: Can we please get an `externalClusterClient.close()` here to close early if everything succeeded?

controllers/storagecluster/external_resources.go
```diff
@@ -230,9 +230,49 @@ func (r *StorageClusterReconciler) newExternalCephObjectStoreInstances(
...
+
+		externalClusterClient, err := r.newExternalClusterClient(instance)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		r.externalClusterClient = externalClusterClient
+		defer r.externalClusterClient.Close()
+
+		if instance.Status.ExternalStorage.ConsumerID == "" {
+			res, err := r.onboardConsumer(instance)
```
nb-ohad, line 244:
```suggestion
			res, err := r.onboardConsumer(instance, externalClusterClient)
```

controllers/storagecluster/external_resources.go
```diff
@@ -230,9 +230,49 @@ func (r *StorageClusterReconciler) newExternalCephObjectStoreInstances(
...
+			if err != nil || !res.IsZero() {
+				return res, err
+			}
+		} else if instance.Spec.ExternalStorage.RequestedCapacity.Equal(instance.Status.ExternalStorage.GrantedCapacity) {
+			res, err := r.updateConsumerCapacity(instance)
```
nb-ohad, line 249:
```suggestion
			res, err := r.updateConsumerCapacity(instance, externalClusterClient)
```

controllers/storagecluster/external_resources.go
```diff
@@ -230,9 +230,49 @@ func (r *StorageClusterReconciler) newExternalCephObjectStoreInstances(
...
+			if err != nil || !res.IsZero() {
+				return res, err
+			}
+		}
+
+		if externalOCSResources[instance.UID] == nil {
+			externalConfig, res, err := r.getExternalConfigFromProvider(instance)
```
nb-ohad, line 256:
```suggestion
			externalConfig, res, err := r.getExternalConfigFromProvider(instance, externalClusterClient)
```

controllers/storagecluster/external_resources.go
```diff
@@ -243,11 +283,26 @@ func (obj *ocsExternalResources) ensureCreated(r *StorageClusterReconciler, inst
...
+
+	if isExternalOCSProvider(instance) {
+		externalClusterClient, err := r.newExternalClusterClient(instance)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		r.externalClusterClient = externalClusterClient
+		defer r.externalClusterClient.Close()
+
+		return r.offboardConsumer(instance)
```
nb-ohad, line 295:
```suggestion
		return r.offboardConsumer(instance, externalClusterClient)
```

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -85,6 +86,9 @@ type StorageClusterReconciler struct {
 	recorder           *util.EventReporter
 	OperatorCondition  conditions.Condition
 	IsNoobaaStandalone bool
+
+	// client required for making grpc call to the external OCS storage cluster
+	externalClusterClient *externalClient.OCSProviderClient
```
nb-ohad, line 91: As discussed, this should not be used anymore, can we remove it?

nb-ohad (changes requested):

controllers/storagecluster/external_resources.go
iamniting: yes we do else we need to pass it.

controllers/storagecluster/external_resources.go
iamniting: no that can not be done, as we need to handle a failure in b/w.

controllers/storagecluster/storagecluster_controller.go
iamniting: done

iamniting: done

controllers/storagecluster/external_resources.go
iamniting: done

controllers/storagecluster/external_ocs.go
iamniting: done

controllers/storagecluster/external_resources.go
iamniting: done

controllers/storagecluster/external_resources.go
iamniting: done

controllers/storagecluster/external_resources.go
iamniting: done

controllers/storagecluster/external_resources.go
iamniting: made changes

controllers/storagecluster/external_ocs.go
```diff
@@ -0,0 +1,250 @@
...
+		context.Background(), connectionInfo.ServerAddress, time.Second*10)
+	if err != nil {
+		return nil, err
+	}
+
+	return consumerClient, nil
+}
+
+// onboardConsumer makes an API call to the external storage provider cluster for onboarding
+func (r *StorageClusterReconciler) onboardConsumer(instance *ocsv1.StorageCluster, grpcClient *externalClient.OCSProviderClient) (reconcile.Result, error) {
```
nb-ohad, line 85: Please rename `grpcClient`

controllers/storagecluster/external_ocs.go
```diff
@@ -0,0 +1,250 @@
...
+		instance.Status.ExternalStorage.ConsumerID,
+		instance.Spec.ExternalStorage.RequestedCapacity.String())
+	if err != nil {
+		if s, ok := status.FromError(err); ok {
+			r.logGrpcErrorAndReportEvent(instance, UpdateCapacity, err, s.Code())
+		}
+		return reconcile.Result{}, err
+	}
+
+	if response.GrantedCapacity == "" {
```
nb-ohad, line 150: Can we check need to match the requested capacity

controllers/storagecluster/external_ocs.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+
+	var externalResources []ExternalResource
+
+	for _, eResource := range response.ExternalResource {
+
+		data := map[string]string{}
+		err = json.Unmarshal(eResource.Data, &data)
+		if err != nil {
+			r.Log.Error(err, "Failed to Unmarshal response of GetStorageConfig", "Kind", eResource.Kind, "Name", eResource.Name)
```
nb-ohad, line 186: Please add the `data` as well

nb-ohad (changes requested):

nb-ohad (approved):

agarwal-mudit (approved):

PR #1445, opened 2021-12-20
storageconsumer: modify api status for rbd, cephfs and cephuser

kesavanvt:
Modify StorageConsumer api to support status on node, provisioner user and subvolume group required for multi tenancy in cephfs.

Signed-off-by: kesavan kvellalo@redhat.com

PR #1446, opened 2021-12-21
StorageConsumer: [Ceph rbd] Add controller changes to provide storage service for consumers

kesavanvt:
This PR adds changes to StorageConsumer controller to provide storage service (rbd) for the consumers . 

Signed-off-by: kesavan kvellalo@redhat.com

### DIFF
--- a/controllers/storageconsumer/storageconsumer_controller.go
+++ b/controllers/storageconsumer/storageconsumer_controller.go
@@ -18,8 +18,14 @@ package controllers
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/go-logr/logr"
+	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v1"
+	"github.com/red-hat-storage/ocs-operator/api/v1alpha1"
+	rookCephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/klog/v2"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -29,14 +35,32 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
+const (
+	storageConsumerFinalizer = "storagesconsumer.ocs.openshift.io"
+)
+
+var (
+	cephResourceIndexMap = map[string]int{}
+	cephResourceIndex    int
+)
+
 // StorageConsumerReconciler reconciles a StorageConsumer object
 type StorageConsumerReconciler struct {
 	client.Client
 	Log    logr.Logger
 	Scheme *runtime.Scheme
+
+	ctx                      context.Context
+	storageConsumer          *ocsv1alpha1.StorageConsumer
+	cephBlockPool            *rookCephv1.CephBlockPool
+	cephClientRBDProvisioner *rookCephv1.CephClient
+	cephClientRBDNode        *rookCephv1.CephClient
+	cephClientHealthChecker  *rookCephv1.CephClient
+	namespace                string
 }
 
 //+kubebuilder:rbac:groups=ocs.openshift.io,resources=storageconsumers,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=ceph.rook.io,resources=cephclients,verbs=get;list;watch;create;update;delete
 //+kubebuilder:rbac:groups=ocs.openshift.io,resources=storageconsumers/status,verbs=get;update;patch
 
 // Reconcile reads that state of the cluster for a StorageConsumer object and makes changes based on the state read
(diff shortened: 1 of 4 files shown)


kesavanvt:
FYI, this PR is on top of https://github.com/red-hat-storage/ocs-operator/pull/1445

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -52,11 +73,229 @@ func (r *StorageConsumerReconciler) Reconcile(ctx context.Context, request recon
...
+
+	return &rookCephv1.CephBlockPool{
+		Spec: rookCephv1.PoolSpec{
+			FailureDomain: failureDomain,
+			Replicated: rookCephv1.ReplicatedSpec{
+				Size:                     3,
+				ReplicasPerFailureDomain: 1,
+				TargetSizeRatio:          .49,
+			},
+			EnableRBDStats: true,
```
umangachapagain, line 222:
```suggestion
			EnableRBDStats: false,
```

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -52,11 +73,229 @@ func (r *StorageConsumerReconciler) Reconcile(ctx context.Context, request recon
...
+
+	// Check status for cephClient resource of rbd.
+	if err := r.get(r.cephClientRBD); err == nil {
+		if r.cephClientRBD.Status.Phase != "Ready" {
+			return false
+		}
+	} else {
+		return false
+	}
+	return true
```
umangachapagain, line 260:
```suggestion
	// Check status for cephBlockPool resource.
	if err := r.get(r.cephBlockPool); err != nil {
			return false
        }

	// Check status for cephClient resource of rbd.
	if err := r.get(r.cephClientRBD); err != nil {
			return false
	}

	return (r.cephBlockPool.Status.Phase == "Ready" && r.cephClientRBD.Status.Phase == "Ready") 
```

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -52,11 +73,229 @@ func (r *StorageConsumerReconciler) Reconcile(ctx context.Context, request recon
...
+
+// Checks whether a string is contained within a slice
+func contains(slice []string, s string) bool {
+	for _, item := range slice {
+		if item == s {
+			return true
+		}
+	}
+	return false
+}
```
umangachapagain, line 289: Don't we already have these functions? Why do we need to replicate it?

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -52,11 +73,229 @@ func (r *StorageConsumerReconciler) Reconcile(ctx context.Context, request recon
...
+
+func (r *StorageConsumerReconciler) own(resource metav1.Object) error {
+	// Ensure storageConsumer ownership on a resource
+	if err := ctrl.SetControllerReference(r.storageConsumer, resource, r.Scheme); err != nil {
+		return err
+	}
+	return nil
+}
```
umangachapagain, line 297: What benefit does this wrapper provide?

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -52,11 +73,229 @@ func (r *StorageConsumerReconciler) Reconcile(ctx context.Context, request recon
...
+	} else {
+		return false
+	}
+	return true
+}
+
+func (r *StorageConsumerReconciler) get(obj client.Object) error {
+	key := client.ObjectKeyFromObject(obj)
+	return r.Client.Get(r.ctx, key, obj)
+}
```
umangachapagain, line 266: What benefit does this wrapper provide?

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -52,11 +84,439 @@ func (r *StorageConsumerReconciler) Reconcile(ctx context.Context, request recon
...
+	storageClusterList := ocsv1.StorageClusterList{}
+	err := r.Client.List(r.ctx, &storageClusterList, client.InNamespace(r.namespace))
+	if err != nil {
+		return fmt.Errorf("Error fetching StorageClusterList. %+v", err)
+	}
+
+	var failureDomain string
+	if len(storageClusterList.Items) == 1 {
+		failureDomain = storageClusterList.Items[0].Status.FailureDomain
+	}
```
nb-ohad, line 223:
```suggestion
	if len(storageClusterList.Items) != 1 {
		// return a an error 
	}
	failureDomain = storageClusterList.Items[0].Status.FailureDomain
```

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -52,11 +84,439 @@ func (r *StorageConsumerReconciler) Reconcile(ctx context.Context, request recon
...
+					ReplicasPerFailureDomain: 1,
+					TargetSizeRatio:          .49,
+				},
+				EnableRBDStats: false,
+				Quotas: rookCephv1.QuotaSpec{
+					MaxSize: &capacity,
+				},
+			},
+		},
+	}
```
nb-ohad, line 241: Orit, can you please verify the config?

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -52,11 +84,439 @@ func (r *StorageConsumerReconciler) Reconcile(ctx context.Context, request recon
...
+	err := r.Client.List(r.ctx, &cephFilesystemList, client.InNamespace(r.namespace))
+
+	if err != nil {
+		return fmt.Errorf("Error fetching CephFilesystemList. %+v", err)
+	}
+
+	var cephFileSystemName string
+	if len(cephFilesystemList.Items) == 1 {
+		cephFileSystemName = cephFilesystemList.Items[0].Name
+	}
```
nb-ohad, line 272: Logic change request:
Less than 1 -> error
More than 1 -> warning
Take the first

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -52,11 +84,439 @@ func (r *StorageConsumerReconciler) Reconcile(ctx context.Context, request recon
...
+			PoolSpec: rookCephv1.PoolSpec{
+				FailureDomain: failureDomain,
+				Replicated: rookCephv1.ReplicatedSpec{
+					Size:                     3,
+					ReplicasPerFailureDomain: 1,
+					TargetSizeRatio:          .49,
+				},
+				EnableRBDStats: false,
+				Quotas: rookCephv1.QuotaSpec{
+					MaxSize: &capacity,
```
leseb, line 237: Does `capacity` parses this regex `^[0-9]+[\.]?[0-9]*([KMGTPE]i|[kMGTPE])?$`?

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -52,11 +84,439 @@ func (r *StorageConsumerReconciler) Reconcile(ctx context.Context, request recon
...
+
+}
+
+func (r *StorageConsumerReconciler) reconcileCephClientRBDProvisioner() error {
+
+	// TODO: Update caps relavent to rbd provisioner
+	desired := &rookCephv1.CephClient{
+		Spec: rookCephv1.ClientSpec{
+			Caps: map[string]string{
+				"mon": "profile rbd",
```
leseb, line 305: You need `"mgr", "allow rw"` too

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -52,11 +84,439 @@ func (r *StorageConsumerReconciler) Reconcile(ctx context.Context, request recon
...
+	return nil
+}
+
+func (r *StorageConsumerReconciler) reconcileCephClientRBDNode() error {
+
+	// TODO: Update caps relavent to rbd node
+	desired := &rookCephv1.CephClient{
+		Spec: rookCephv1.ClientSpec{
+			Caps: map[string]string{
+				"mon": "profile rbd",
```
leseb, line 335: You need `"mgr", "allow rw"` too

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -52,11 +84,439 @@ func (r *StorageConsumerReconciler) Reconcile(ctx context.Context, request recon
...
+	return nil
+}
+
+func (r *StorageConsumerReconciler) reconcileCephClientCephFSProvisioner() error {
+
+	// TODO: Update caps relavent to cephfs provisioner
+	desired := &rookCephv1.CephClient{
+		Spec: rookCephv1.ClientSpec{
+			Caps: map[string]string{
+				"mon": "profile rbd",
```
leseb, line 364: You need `"mgr", "allow rw"` too

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -52,11 +84,439 @@ func (r *StorageConsumerReconciler) Reconcile(ctx context.Context, request recon
...
+	return nil
+}
+
+func (r *StorageConsumerReconciler) reconcileCephClientCephFSNode() error {
+
+	// TODO: Update caps relavent to cephfs node
+	desired := &rookCephv1.CephClient{
+		Spec: rookCephv1.ClientSpec{
+			Caps: map[string]string{
+				"mon": "profile rbd",
```
leseb, line 394: You need `"mgr", "allow rw"` too

controllers/storageconsumer/storageconsumer_controller.go
leseb, line 364:
```suggestion
				"mon": "allow r",
```

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -52,11 +84,439 @@ func (r *StorageConsumerReconciler) Reconcile(ctx context.Context, request recon
...
+	r.storageConsumer.Status.CephObjects.CephRBDNodeUser = r.cephClientRBDNode.Name
+	return nil
+}
+
+func (r *StorageConsumerReconciler) reconcileCephClientCephFSProvisioner() error {
+
+	// TODO: Update caps relavent to cephfs provisioner
+	desired := &rookCephv1.CephClient{
+		Spec: rookCephv1.ClientSpec{
+			Caps: map[string]string{
```
leseb, line 363: Rook uses:

```
	return []string{
		"mon", "allow r",
		"mgr", "allow rw",
		"osd", "allow rw tag cephfs metadata=*",
	}
```

But we need to adapt the osd caps to match the subvolumegroup name.

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -52,11 +84,439 @@ func (r *StorageConsumerReconciler) Reconcile(ctx context.Context, request recon
...
+	r.storageConsumer.Status.CephObjects.CephFSProvisionerUser = r.cephClientCephFSProvisioner.Name
+	return nil
+}
+
+func (r *StorageConsumerReconciler) reconcileCephClientCephFSNode() error {
+
+	// TODO: Update caps relavent to cephfs node
+	desired := &rookCephv1.CephClient{
+		Spec: rookCephv1.ClientSpec{
+			Caps: map[string]string{
```
leseb, line 393: Rook uses:

```
	return []string{
		"mon", "allow r",
		"mgr", "allow rw",
		"osd", "allow rw tag cephfs *=*",
		"mds", "allow rw",
	}
```


But we need to adapt the osd caps to match the subvolumegroup name.

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -52,11 +84,439 @@ func (r *StorageConsumerReconciler) Reconcile(ctx context.Context, request recon
...
+	r.storageConsumer.Status.CephObjects.CephFSNodeUser = r.cephClientCephFSNode.Name
+	return nil
+}
+
+func (r *StorageConsumerReconciler) reconcileCephClientHealthChecker() error {
+
+	// TODO: Update caps relavent to health checker
+	desired := &rookCephv1.CephClient{
+		Spec: rookCephv1.ClientSpec{
+			Caps: map[string]string{
```
leseb, line 423: Rook uses:

```
            'mgr': 'allow command config',
            'mon': 'allow r, allow command quorum_status, allow command version',
            'osd': "allow rwx pool={0}.rgw.meta, " +
            "allow r pool=.rgw.root, " +
            "allow rw pool={0}.rgw.control, " +
            "allow rx pool={0}.rgw.log, " +
            "allow x pool={0}.rgw.buckets.index"
```

Forget about the pool rgw caps if there is no object store and stick with the first 2 lines.

leseb (changes requested):

api/v1alpha1/storageconsumer_types.go
```diff
@@ -46,8 +45,18 @@ type StorageConsumerSpec struct {
...
 	// BlockPoolName holds the name of created ceph block pool.
 	BlockPoolName string `json:"blockPoolName,omitempty"`
-	// CephUser holds the name of created ceph user.
-	CephUser string `json:"cephUser,omitempty"`
+	// CephRBDProvisionerUser holds the name of created ceph rbd provisioner user required for rbd for the ceph-csi driver.
+	CephRBDProvisionerUser string `json:"cephRBDProvisionerUser,omitempty"`
+	// CephRBDNodeUser holds the name of created ceph rbd node user required for rbd for the ceph-csi driver.
+	CephRBDNodeUser string `json:"cephRBDNodeUser,omitempty"`
+	// SubVolumeGroupName holds the name of created subVolume group.
+	SubVolumeGroupName string `json:"subVolumeGroupName,omitempty"`
```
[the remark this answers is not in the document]
kesavanvt: As per the new discussion, will be proceeding with list of ceph resources

controllers/storageconsumer/storageconsumer_controller.go
kesavanvt: As per the new discussion, this flow has been modified.

controllers/storageconsumer/storageconsumer_controller.go
kesavanvt: We have this function in a different controller package (storagecluster controller). Keeping in consideration not modify the existing storage controller code, have introduced the function in this package.

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -52,11 +84,439 @@ func (r *StorageConsumerReconciler) Reconcile(ctx context.Context, request recon
...
+	result, reconcileError := r.reconcilePhases()
+
+	// Apply status changes to the StorageConsumer
+	statusError := r.Client.Status().Update(r.ctx, r.storageConsumer)
+	if statusError != nil {
+		r.Log.Info("Could not update StorageConsumer status.", "StorageConsumer", klog.KRef(r.storageConsumer.Namespace, r.storageConsumer.Name))
+	}
+
+	// Reconcile errors have higher priority than status update errors
+	if reconcileError != nil {
```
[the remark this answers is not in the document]
nb-ohad: Because we want to guarantee that the status is written to k8s even when reconcile have errors.
So we cannot handle the error before writing the status

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -52,11 +84,439 @@ func (r *StorageConsumerReconciler) Reconcile(ctx context.Context, request recon
...
+		return result, reconcileError
+	} else if statusError != nil {
+		return result, statusError
+	} else {
+		return result, nil
+	}
+}
+
+func (r *StorageConsumerReconciler) initReconciler(request reconcile.Request) {
+	r.ctx = context.Background()
```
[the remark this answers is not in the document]
nb-ohad: This is the reconciler init function, it is the one that creates a single context to be used by the entire reconcile invocation.

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -52,11 +84,439 @@ func (r *StorageConsumerReconciler) Reconcile(ctx context.Context, request recon
...
+	statusError := r.Client.Status().Update(r.ctx, r.storageConsumer)
+	if statusError != nil {
+		r.Log.Info("Could not update StorageConsumer status.", "StorageConsumer", klog.KRef(r.storageConsumer.Namespace, r.storageConsumer.Name))
+	}
+
+	// Reconcile errors have higher priority than status update errors
+	if reconcileError != nil {
+		return result, reconcileError
+	} else if statusError != nil {
+		return result, statusError
```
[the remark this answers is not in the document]
nb-ohad: see response to your last comment

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -52,14 +72,366 @@ func (r *StorageConsumerReconciler) Reconcile(ctx context.Context, request recon
...
+	r.cephClientHealthChecker.Name = fmt.Sprintf("%s-%s", "cephclient-health-checker", r.storageConsumer.Name)
+	r.cephClientHealthChecker.Namespace = r.namespace
+
+}
+
+func (r *StorageConsumerReconciler) reconcilePhases() (reconcile.Result, error) {
+
+	r.storageConsumer.Status.State = v1alpha1.StorageConsumerStateConfiguring
+
+	r.storageConsumer.Status.CephResources = nil
```
nb-ohad, line 140: I think this is erroneous, resetting the array means that if this reconcile fails at an early stage will update the StorageConsumer status with an empty status for the reconciled objects, but these objects might exist (and have status) from a prev reconcile invocation. 

I think we should keep prev status for existing resources that were not touched during the current reconciliation. 
To do so, you need to remove this line

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -52,14 +72,366 @@ func (r *StorageConsumerReconciler) Reconcile(ctx context.Context, request recon
...
+
+		if err := r.reconcileCephClientRBDNode(); err != nil {
+			return reconcile.Result{}, err
+		}
+
+		if err := r.reconcileCephClientHealthChecker(); err != nil {
+			return reconcile.Result{}, err
+		}
+
+		var notReady bool
```
nb-ohad, line 168: This has backward logic, for the sake of understandability I think it will be prefer to check for readiness instead of non-readiness

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -52,14 +72,366 @@ func (r *StorageConsumerReconciler) Reconcile(ctx context.Context, request recon
...
+		for index := range r.storageConsumer.Status.CephResources {
+			cephObject := &r.storageConsumer.Status.CephResources[index]
+			if cephObject.Phase != "Ready" {
+				notReady = true
+			}
```
nb-ohad, line 173: You can bail out from the loop (using `break`) the first time you find a resource that is not ready

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -52,14 +72,366 @@ func (r *StorageConsumerReconciler) Reconcile(ctx context.Context, request recon
...
+			r.storageConsumer.Status.State = v1alpha1.StorageConsumerStateReady
+			r.storageConsumer.Status.GrantedCapacity = r.storageConsumer.Spec.Capacity
+		}
+
+	} else {
+		r.storageConsumer.Status.State = v1alpha1.StorageConsumerStateDeleting
+
+		var cephResourcesExists bool
+		if err := r.Client.Delete(r.ctx, r.cephBlockPool); err == nil {
+			cephResourcesExists = true
```
nb-ohad, line 186: This does not make sense, if you managed to issue the delete then the resource should be considered as a non-existing resource. Am I missing something?

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -52,14 +72,366 @@ func (r *StorageConsumerReconciler) Reconcile(ctx context.Context, request recon
...
+		return nil
+	})
+
+	cephClientRBDProvisionerDetails := &ocsv1alpha1.CephResourcesSpec{
+		Kind: r.cephClientRBDProvisioner.Kind,
+		Name: r.cephClientRBDProvisioner.Name,
+	}
+	if r.cephClientRBDProvisioner.Status != nil {
+		cephClientRBDProvisionerDetails.Phase = string(r.cephClientRBDProvisioner.Status.Phase)
+	}
```
nb-ohad, line 311: What will be the phase if there is no status on the rook resource?

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -52,14 +72,366 @@ func (r *StorageConsumerReconciler) Reconcile(ctx context.Context, request recon
...
+		return nil
+	})
+
+	cephClientRBDNodeDetails := &ocsv1alpha1.CephResourcesSpec{
+		Kind: r.cephClientRBDNode.Kind,
+		Name: r.cephClientRBDNode.Name,
+	}
+	if r.cephClientRBDNode.Status != nil {
+		cephClientRBDNodeDetails.Phase = string(r.cephClientRBDNode.Status.Phase)
+	}
```
nb-ohad, line 348: What will be the phase if there is no status on the rook resource?

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -52,14 +72,366 @@ func (r *StorageConsumerReconciler) Reconcile(ctx context.Context, request recon
...
+		Kind: r.cephBlockPool.Kind,
+		Name: r.cephBlockPool.Name,
+		CephClients: map[string]string{
+			"provisioner": r.cephClientRBDProvisioner.Name,
+			"node":        r.cephClientRBDNode.Name,
+		},
+	}
+	if r.cephBlockPool.Status != nil {
+		cephBlockPoolDetails.Phase = string(r.cephBlockPool.Status.Phase)
+	}
```
nb-ohad, line 273: What will be the phase if there is no status on the rook resource?

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -52,14 +72,366 @@ func (r *StorageConsumerReconciler) Reconcile(ctx context.Context, request recon
...
+		return nil
+	})
+
+	cephClientHealthCheckerDetails := &ocsv1alpha1.CephResourcesSpec{
+		Kind: r.cephClientHealthChecker.Kind,
+		Name: r.cephClientHealthChecker.Name,
+	}
+	if r.cephClientHealthChecker.Status != nil {
+		cephClientHealthCheckerDetails.Phase = string(r.cephClientHealthChecker.Status.Phase)
+	}
```
nb-ohad, line 385: What will be the phase if there is no status on the rook resource?

controllers/storageconsumer/storageconsumer_controller.go
nb-ohad: `get(object)` -> `get(ctx, key , obj)`
The wrapper auto-populates the additional info from the object itself and the reconciler, this way we do not need to do it on every call site

controllers/storageconsumer/storageconsumer_controller.go
nb-ohad: The wrapper auto-populates the additional info from the reconciler, this way we do not need to do it on every call site. But I would say this can be better  written as:

```suggestion
func (r *StorageConsumerReconciler) own(resource metav1.Object) error {
	// Ensure storageConsumer ownership on a resource
	return ctrl.SetControllerReference(r.storageConsumer, resource, r.Scheme)
}
```

nb-ohad (changes requested):

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -52,11 +84,439 @@ func (r *StorageConsumerReconciler) Reconcile(ctx context.Context, request recon
...
+	if statusError != nil {
+		r.Log.Info("Could not update StorageConsumer status.", "StorageConsumer", klog.KRef(r.storageConsumer.Namespace, r.storageConsumer.Name))
+	}
+
+	// Reconcile errors have higher priority than status update errors
+	if reconcileError != nil {
+		return result, reconcileError
+	} else if statusError != nil {
+		return result, statusError
+	} else {
```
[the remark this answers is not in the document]
kesavanvt: yes, removed as it is not required

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -52,11 +84,439 @@ func (r *StorageConsumerReconciler) Reconcile(ctx context.Context, request recon
...
+	}
+
+	capacity := r.storageConsumer.Spec.Capacity.String()
+	desired := &rookCephv1.CephBlockPool{
+		Spec: rookCephv1.NamedBlockPoolSpec{
+			PoolSpec: rookCephv1.PoolSpec{
+				FailureDomain: failureDomain,
+				Replicated: rookCephv1.ReplicatedSpec{
+					Size:                     3,
+					ReplicasPerFailureDomain: 1,
```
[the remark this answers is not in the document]
kesavanvt: As we discussed earlier , the information that are provided here is separate from the defaults that storage cluster has.

controllers/storageconsumer/storageconsumer_controller.go
kesavanvt: fixed

controllers/storageconsumer/storageconsumer_controller.go
kesavanvt: fixed

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -52,14 +71,367 @@ func (r *StorageConsumerReconciler) Reconcile(ctx context.Context, request recon
...
+
+		if err := r.reconcileCephClientRBDNode(); err != nil {
+			return reconcile.Result{}, err
+		}
+
+		if err := r.reconcileCephClientHealthChecker(); err != nil {
+			return reconcile.Result{}, err
+		}
+
+		var cephResourcesReady bool = true
```
nb-ohad, line 168:
```suggestion
		cephResourcesReady :=  true
```

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -52,14 +71,367 @@ func (r *StorageConsumerReconciler) Reconcile(ctx context.Context, request recon
...
+			cephResource := &r.storageConsumer.Status.CephResources[index]
+			if cephResource.Phase != "Ready" {
+				cephResourcesReady = false
+				break
+			}
+		}
+
+		if cephResourcesReady {
+			r.storageConsumer.Status.State = v1alpha1.StorageConsumerStateReady
+			r.storageConsumer.Status.GrantedCapacity = r.storageConsumer.Spec.Capacity
```
nb-ohad, line 179: Updating capacity here is problematic, you need to update this as soon as you managed to change the quota on the block pool. If not a later failure will unsync this value with the actual value on the block pool resource

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -52,11 +84,439 @@ func (r *StorageConsumerReconciler) Reconcile(ctx context.Context, request recon
...
+
+	capacity := r.storageConsumer.Spec.Capacity.String()
+	desired := &rookCephv1.CephBlockPool{
+		Spec: rookCephv1.NamedBlockPoolSpec{
+			PoolSpec: rookCephv1.PoolSpec{
+				FailureDomain: failureDomain,
+				Replicated: rookCephv1.ReplicatedSpec{
+					Size:                     3,
+					ReplicasPerFailureDomain: 1,
+					TargetSizeRatio:          .49,
```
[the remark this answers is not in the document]
kesavanvt: Done

controllers/storageconsumer/storageconsumer_controller.go
kesavanvt: capacity is derived from `resource.Quantity` , so the parsing is taken care of it

controllers/storageconsumer/storageconsumer_controller.go
kesavanvt: yes there is no object store , fixed

controllers/storageconsumer/storageconsumer_controller.go
kesavanvt: Noted, this will be handled in a new PR containing cephfs changes

controllers/storageconsumer/storageconsumer_controller.go
kesavanvt: Noted, this will be handled in a new PR containing cephfs changes

controllers/storageconsumer/storageconsumer_controller.go
kesavanvt: Noted, this will be handled in a new PR containing cephfs changes

controllers/storageconsumer/storageconsumer_controller.go
kesavanvt: Noted, this will be handled in a new PR containing cephfs changes

controllers/storageconsumer/storageconsumer_controller.go
kesavanvt: fixed

controllers/storageconsumer/storageconsumer_controller.go
kesavanvt: After issuing the Delete operation , it is not guaranteed that the resource deletion is complete. Once the resource is deleted, it triggers the reconciliation and the flow does not reach this condition block.

controllers/storageconsumer/storageconsumer_controller.go
kesavanvt: there wont be phase in the cephresources list (ie it will not be defined)

controllers/storageconsumer/storageconsumer_controller.go
kesavanvt: This has been removed and introduced logic for modifying the existing cephResources list.

controllers/storageconsumer/storageconsumer_controller.go
kesavanvt: The flow has been modified to fetch the existence of cephResources first

nb-ohad (approved):

subhamkrai (commented):
lgtm

PR #1447, opened 2021-12-21
Report versioning errors in StorageCluster's status

rexagod:
Earlier, only the status of the storage cluster was being updated,
because of which, side effects such as version mismatch came in to play.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -161,7 +161,7 @@ golangci-lint:
 # use 'make functest' to run just the functional tests
 unit-test:
 	@echo "Executing unit tests"
-	go test -v -cover `go list ./... | grep -v "functest"`
+	hack/unit-test.sh
 
 ocs-operator-ci: shellcheck-test golangci-lint unit-test verify-deps verify-generated verify-latest-deploy-yaml
 
--- a/controllers/storagecluster/reconcile.go
+++ b/controllers/storagecluster/reconcile.go
@@ -200,6 +200,12 @@ func (r *StorageClusterReconciler) validateStorageClusterSpec(instance *ocsv1.St
 		r.Log.Error(err, "Failed to validate StorageCluster version.", "StorageCluster", klog.KRef(instance.Namespace, instance.Name))
 		r.recorder.ReportIfNotPresent(instance, corev1.EventTypeWarning, statusutil.EventReasonValidationFailed, err.Error())
 		instance.Status.Phase = statusutil.PhaseError
+		instance.Status.Conditions = append(instance.Status.Conditions, conditionsv1.Condition{
+			Type:    conditionsv1.ConditionDegraded,
+			Status:  corev1.ConditionTrue,
+			Reason:  statusutil.EventReasonValidationFailed,
+			Message: err.Error(),
+		})
 		if updateErr := r.Client.Status().Update(context.TODO(), instance); updateErr != nil {
 			r.Log.Error(updateErr, "Failed to update StorageCluster.", "StorageCluster", klog.KRef(instance.Namespace, instance.Name))
 			return updateErr
--- a/hack/build-functest.sh
+++ b/hack/build-functest.sh
@@ -21,8 +21,9 @@ else
 	echo "GINKO binary found at $GINKGO"
 fi
 
+LDFLAGS="-X github.com/red-hat-storage/ocs-operator/version.Version=${VERSION}"
 
-"$GOBIN"/ginkgo build "functests/${suite}/"
+"$GOBIN"/ginkgo build --ldflags "${LDFLAGS}" "functests/${suite}/"
 
 mkdir -p $OUTDIR_BIN
 mv "functests/${suite}/${suite}.test" "${OUTDIR_BIN}/${suite}_tests"
--- a/hack/generate-unified-csv.sh
+++ b/hack/generate-unified-csv.sh
@@ -5,7 +5,10 @@ set -e
 source hack/common.sh
 
 CSV_MERGER="tools/csv-merger/csv-merger"
-(cd tools/csv-merger/ && go build)
+
+LDFLAGS="-X github.com/red-hat-storage/ocs-operator/version.Version=${VERSION}"
+
+(cd tools/csv-merger/ && go build -ldflags="${LDFLAGS}")
 
 function help_txt() {
 	echo "Environment Variables"
--- a/hack/unit-test.sh
+++ b/hack/unit-test.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+set -e
+
+source hack/common.sh
+
+LDFLAGS="-X github.com/red-hat-storage/ocs-operator/version.Version=${VERSION}"
+
+go test -ldflags="${LDFLAGS}" -v -cover `go list ./... | grep -v "functest"`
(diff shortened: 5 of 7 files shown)


rexagod:
`verify-generated` looks good on local but fails for the CI.

```
❯ make verify-generated

Updating generated code
/home/rexagod/go/bin/controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."
Updating generated manifests
/home/rexagod/go/bin/controller-gen "crd:trivialVersions=true" rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/
bases
Verifying generated code and manifests
hack/verify-generated.sh
Success: no out of source tree changes found for generated files
```

To verify that nothing else crept in, and this branch was in sync with `main`, here are the `git diff` logs.

```
❯ gd upstream/main --name-only

api/v1/zz_generated.deepcopy.go
api/v1alpha1/zz_generated.deepcopy.go
controllers/storagecluster/reconcile.go
```

I'd appreciate any help with this, thanks.

rexagod:
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/red-hat-storage_ocs-operator/1447/pull-ci-red-hat-storage-ocs-operator-main-ocs-operator-bundle-e2e-aws/1473402713831116800#1:build-log.txt%3A157

umangachapagain (changes requested):
I understand why we need to do this, but an operator updating it's own CR is probably not a good idea. We need to re-work on the whole version thing. So, putting this on hold while we explore other options.

/hold

umangachapagain:
It seems late initialization is fine.

Can we patch only the version field from inside version check, instead of updating the entire resource?
If it fails, we'll retry on next reconcile instead of error out.

Changes in this PR can be reverted.
@jarrpa WDYT?

nb-ohad:
@rexagod @jarrpa @umangachapagain 

I think `Patch` is the wrong approach here as it does not guarantee the replacement of the old status with the new status. 
We might start to see the strange status values-based as of merge semantics. 

I will vote to stay with update and maybe add retries

rexagod:
/hold

I think I missed Ohad's comment before pushing the latest commit. Nonetheless, I'll put this PR on hold until we reach a common consensus here.

jarrpa (changes requested):
It took some doing but I think I understand the problem with the Patch() strategy now.

In general a Patch() strategy has its place, and it is not here. Specifically this is because the StorageCluster reconcile loop is designed to collect changes to the Status and push them all at once at the end of the reconcile loop (whether or not reconciliation completed successfully is a different story). As such what Ohad is saying is correct, we shouldn't be "merging" new state information, we should be "replacing" the whole thing since we are taking the role as the source of truth on the current state of the system at that point in time.

Now, it is true that we have a bug in how we deal with the Version field in the spec. But also, the way we're using the version field is completely broken, we should *never* be changing the spec to begin with. I would rather see the version field be dropped entirely and have the "version" information moved to the Status subresource. That will require a bit of design discussion, but shouldn't be too bad.

All that being said, there is still some worthwhile work being done in this PR:

* I actually like the idea of having the version string be empty in version.go, so that should be kept.
* The versionCheck error condition is also a great addition, so keep that as well.

If you want to change this PR to just those two changes, I'm all right with that. But I would rather have a better conversation on whether we can safely *stop using the spec.version field* once and for all. We'll have to coordinate with the UI team to know when they'd be able to update the Console code for that, but as it stands it's already broken so what more harm could it do? ¯\_(ツ)_/¯

umangachapagain:
> 
>     * I actually like the idea of having the version string be empty in version.go, so that should be kept.
>
We use version info to generate a version metric . As long as this doesn't break, I'm fine with it.
https://github.com/red-hat-storage/ocs-operator/blob/4f2dfefe4e014f1191ef88d4cdf831ecde3ef430/metrics/internal/version/version.go#L1-L11

rexagod:
> We use version info to generate a version metric . As long as this doesn't break, I'm fine with it.

It seems `(github.com/red-hat-storage/ocs-operator/metrics/internal/version).GetVersion()` successfully picks up the updated version from `version.Version`.

image

rexagod:
I think we'll need to include the version string since `ldflags` are not in the scope of unit tests, so `version.Version` will always be `""` for them.

https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/red-hat-storage_ocs-operator/1447/pull-ci-red-hat-storage-ocs-operator-main-ocs-operator-ci/1503646574947864576#1:build-log.txt%3A287-292

rexagod:
Bugzilla link: https://bugzilla.redhat.com/show_bug.cgi?id=1855339

umangachapagain (approved):

rexagod:
Alternative to the `LDFLAGS` injection in the `version` package for test binaries (which is what's happening in this PR right now), I just wanted to put it out there that there's also an approach that is IDE or command independent, meaning that it'll allow us to inject latest version while using `go test` or any custom IDE plugin on the fly. The compilation in this case also need not be for the entire tests collection, only a subset of them (e.g., filtering).
***
```diff
diff --git a/version/version.go b/version/version.go
index 9c0e886b96f4176bf6508c679a12499febca5138..f3b884e8a7c711316e95cdf110ed210edb876f62 100644
--- a/version/version.go
+++ b/version/version.go
@@ -1,6 +1,38 @@
 package version
 
+import (
+       "bufio"
+       "os"
+       "path/filepath"
+)
+
 var (
        // Version of the operator
-       Version = "4.11.0"
+       Version = ""
 )
+
+func init() {
+       // The condition below will be false if the OCS binary is already created using the `LDFLAGS` parameters defined in
+       // hack/go-build.sh.
+       // The condition below will be true in scenarios where `LDFLAGS` injection does not happen, for e.g., when running
+       // tests.
+       if Version == "" {
+               fp, _ := filepath.Abs("../hack/common.sh")
+               f, err := os.Open(fp)
+               if err != nil {
+                       panic(err)
+               }
+               defer f.Close()
+               v := "DEFAULT_CSV_VERSION="
+               vLen := len(v)
+               scanner := bufio.NewScanner(f)
+               for scanner.Scan() {
+                       t := scanner.Text()
+                       if len(t) >= vLen && t[:vLen] == v {
+                               Version = t[vLen+1:]               // Remove leading apostrophe
+                               Version = Version[:len(Version)-1] // Remove trailing apostrophe
+                               break
+                       }
+               }
+       }
+}
```

rexagod:
Will `/unhold` once the e2e tests pass.

rexagod:
Follow-up PR: https://github.com/red-hat-storage/ocs-operator/pull/1706 (`spec.version` deprecation).

malayparida2000:
@rexagod I think this PR is stale now & no longer required. So can we close this?

Nikhil-Ladha:
Looking at the latest [bz](https://bugzilla.redhat.com/show_bug.cgi?id=1855339) status, I think we still need this PR.
Also, as the depreciation PR is merged we should be good to test this PR on a nightly/test build, once rebased.

rexagod:
Not sure if we still need this, but since I've moved to a different team recently, I'd appreciate if someone with bandwidth can pick this up.

Closing, thanks.

PR #1448, opened 2021-12-23
Implement mock Provider API server

sp98:
This PR implements mock Provider API server.  This can be used by the OCS operator controller for testing.


services/provider/common/mock.go
```diff
@@ -0,0 +1,95 @@
...
+	UpdateInvalidUID       MockError = "UPDATE_INVALID_UID"
+	UpdateConsumerNotFound MockError = "UPDATE_CONSUMER_NOT_FOUND"
+
+	MockConsumerID = "vMHA0ppPbjg5TlgvMFcaH4QlQEJB68u+1jWQJ9O9xvde8fxz5vBuu2F6bVIY6pAYLVrC3FajrK1KxmhFTzNDow=="
+
+	MockGrantedCapacity = "2Ti"
+)
+
+var (
+	MockProviderAPI      = os.Getenv("MOCK_PROVIDER_API")
```
iamniting, line 35: It will be fetched once while starting, which means even if we change this in b/w it won't affect. I would suggest getting this in every call so that it can be changed on the fly before making a call.

iamniting (changes requested):

services/provider/common/mock.go
sp98: good point. I'll update this.

sp98: done.

umangachapagain (commented):
Is it a standard practice to keep mocks and actual implementations side by side in the same package?

I am wondering if a separate mock package would be better here. Say `services/provider/mock` that can then have `client.go`, `server.go` and `common.go`. When consuming this you'd explicitly import the mock package vs importing the package with actual implementation and using mock.

Also, shouldn't mocks be completely independent/de-coupled/external from the real implementation?

sp98:
> Is it a standard practice to keep mocks and actual implementations side by side in the same package?

Not sure of this. Quick google search didn't give any good answers. 
 
> I am wondering if a separate mock package would be better here. Say `services/provider/mock` that can then have `client.go`, `server.go` and `common.go`. When consuming this you'd explicitly import the mock package vs importing the package with actual implementation and using mock.
> 
> Also, shouldn't mocks be completely independent/de-coupled/external from the real implementation?

If I'm reading it right, I think your first and last statement above could be a bit contradicting. If we decide to move mocks to a different package than actual implementation then it would make more sense to have `mocks be completely independent/de-coupled/external from the real implementation`.  

We have a peculiar situation, where the server mock is a temporary solution (we can discuss this offline), hence server mock and server are using the same interface implementation. 

`Client mock` and `client` are already different implementations. So having `services/provider/client/client.go` and `services/provider/client/client_mock.go` should be fine, I think.

agarwal-mudit (approved):

PR #1450, opened 2022-01-05
controllers: change resourceManager interface

iamniting:
With the current resourceManager interface there is no way to
communicate the reconcile.Result back to the reconcilePhases as it only
returns the error.

change the interface signature to return the reconcile.Result as well
with the error.

Why?
The change is required for ocs to ocs where we want to communicate the
reconcile.Result to the reconcilePhases.
    
While making GPRC calls we will get different error codes and based on
those error codes we will decide whether we need to requeue or not if
requeue when to requeue and this can be done with reconcile.Result. With
the existing design, we can return only errors and reconcile will get
requeue. That's why we are changing the interface to return the
reconcile.Result also.
    
Usecase:
let's say a user will create a storageCluster CR with the wrong
connection string and while making an onboarding request we will get an
error that says invalid connection string. This means there is no point
in re queueing at all as this can not be solved without user interaction.

### DIFF
--- a/controllers/storagecluster/uninstall_reconciler.go
+++ b/controllers/storagecluster/uninstall_reconciler.go
@@ -16,6 +16,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/strategicpatch"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
 // CleanupPolicyType is a string representing cleanup policy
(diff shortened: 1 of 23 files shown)


oritwas (commented):
Please describe why do we need to return reconcile.Result from ensureCreated.

oritwas (approved):

controllers/storagecluster/uninstall_reconciler.go
```diff
@@ -297,20 +299,21 @@ func (r *StorageClusterReconciler) deleteResources(sc *ocsv1.StorageCluster) err
...
 	err = r.deleteNodeTaint(sc)
 	if err != nil {
-		return err
+		return reconcile.Result{RequeueAfter: time.Second * time.Duration(1)}, err
 	}
 
 	err = deleteKMSResources(r, sc)
 	if err != nil {
-		return err
+		return reconcile.Result{RequeueAfter: time.Second * time.Duration(1)}, err
```
sp98, line 316: would be nice to use this result with 1 second timeout as a constant since its used in a lot of places.

sp98 (changes requested):
nit

nbalacha (approved):

PR #1451, opened 2022-01-13
must-gather: stop using tini in must gather helper pod

subhamkrai:
stop using tini in must gather helper pod template. This is the only
the place we are using tini.

### DIFF
--- a/must-gather/templates/pod.template
+++ b/must-gather/templates/pod.template
@@ -33,7 +33,9 @@ spec:
     name: must-gather-helper
     securityContext:
       privileged: true
-      runAsUser: 1000
+      runAsNonRoot: true
+      runAsUser: 2016
+      runAsGroup: 2016
     volumeMounts:
       - mountPath: /dev
         name: dev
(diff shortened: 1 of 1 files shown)


crombus (changes requested):
Changes Looks good but commit msg should be updated.

In the second commit update the must-gather helper pod template not must-gather pod template.

agarwal-mudit:
/hold
1. Requires review from @leseb 
2. Review commits need to be rectified

must-gather/templates/pod.template
```diff
@@ -11,11 +11,12 @@ spec:
...
   - args:
-    - -g
-    - --
+    - -m
+    - -c
     - /usr/local/bin/toolbox.sh
     command:
-    - /tini
+    - /bin/bash
+    tty: true
```
leseb, line 19: This change should be part of the first commit (removal of `tini`)

leseb (changes requested):

leseb (approved):

PR #1453, opened 2022-01-13
Update gRPC provider client

sp98:
- This PR abstracts the gRPC connection by adding it inside the `NewProviderClient` function. 
- Also provides an function to close the connection 
- Passes parent context for each gRPC call.

### DIFF
--- a/services/provider/client/client.go
+++ b/services/provider/client/client.go
@@ -10,19 +10,14 @@ import (
 )
 
 type OCSProviderClient struct {
-	Client  pb.OCSProviderClient
-	timeout time.Duration
+	Client     pb.OCSProviderClient
+	clientConn *grpc.ClientConn
+	timeout    time.Duration
 }
 
-// NewProviderClient creates a client to talk to OCS provider server
-func NewProviderClient(cc *grpc.ClientConn, timeout time.Duration) *OCSProviderClient {
-	return &OCSProviderClient{Client: pb.NewOCSProviderClient(cc), timeout: timeout}
-}
-
-// NewGRPCConnection returns a grpc client connection which can be used to create the consumer client
-// Note: Close the connection after use
-func NewGRPCConnection(serverAddr string, timeout time.Duration) (*grpc.ClientConn, error) {
-	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+// NewProviderClient creates a client to talk to the external OCS storage provider server
+func NewProviderClient(ctx context.Context, serverAddr string, timeout time.Duration) (*OCSProviderClient, error) {
+	apiCtx, cancel := context.WithTimeout(ctx, timeout)
 	defer cancel()
 
 	opts := []grpc.DialOption{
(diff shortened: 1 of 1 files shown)


services/provider/client/client.go
```diff
@@ -10,13 +10,34 @@ import (
...
+	timeout    time.Duration
 }
 
-// NewProviderClient creates a client to talk to OCS provider server
-func NewProviderClient(cc *grpc.ClientConn, timeout time.Duration) *OCSProviderClient {
-	return &OCSProviderClient{Client: pb.NewOCSProviderClient(cc), timeout: timeout}
+// NewProviderClient creates a client to talk to the external OCS storage provider server
+func NewProviderClient(serverAddr string, timeout time.Duration) (*OCSProviderClient, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
```
nb-ohad, line 21: I think this will cancel when you go out of the `NewProviderClient` call

services/provider/client/client.go
```diff
@@ -10,13 +10,34 @@ import (
...
+
+	clientConn, err := grpc.DialContext(ctx, serverAddr, opts...)
+	if err != nil {
+		return nil, fmt.Errorf("failed to dial: %v", err)
+	}
+
+	return &OCSProviderClient{Client: pb.NewOCSProviderClient(clientConn),
+		clientConn: clientConn,
+		timeout:    timeout}, nil
+}
```
nb-ohad, line 36:
```suggestion
	return &OCSProviderClient{
		Client: pb.NewOCSProviderClient(clientConn),
		clientConn: clientConn,
		timeout:    timeout}, nil
}
```

services/provider/client/client.go
```diff
@@ -10,13 +10,34 @@ import (
...
+	clientConn *grpc.ClientConn
+	timeout    time.Duration
 }
 
-// NewProviderClient creates a client to talk to OCS provider server
-func NewProviderClient(cc *grpc.ClientConn, timeout time.Duration) *OCSProviderClient {
-	return &OCSProviderClient{Client: pb.NewOCSProviderClient(cc), timeout: timeout}
+// NewProviderClient creates a client to talk to the external OCS storage provider server
+func NewProviderClient(serverAddr string, timeout time.Duration) (*OCSProviderClient, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
```
nb-ohad, line 20: Please pass the context from the outside. 
For the interim period (until we refactor `context` usage on the for reconcile loop)  we can just create a new context on the call site

nb-ohad (changes requested):

services/provider/client/client.go
```diff
@@ -10,82 +10,85 @@ import (
...
+
+// Close closes the gRPC connection of the external OCS storage provider client
+func (cc *OCSProviderClient) Close() error {
+	return cc.clientConn.Close()
 }
 
 // OnboardConsumer to validate the consumer and create StorageConsumer
 // resource on the StorageProvider cluster
-func (cc *OCSProviderClient) OnboardConsumer(token, name, capacity string) (*pb.OnboardConsumerResponse, error) {
+func (cc *OCSProviderClient) OnboardConsumer(ctx context.Context, token, name, capacity string) (*pb.OnboardConsumerResponse, error) {
```
iamniting, line 46: Can we save the context also in the `OCSProviderClient` and use the same again and again. WDYT?

iamniting (changes requested):

services/provider/client/client.go
nb-ohad: context needs to pass from the outside as the actual context (not code object) of the operation is unknown to the client itself and can change between call sites.

services/provider/client/client.go
```diff
@@ -10,82 +10,103 @@ import (
...
 
-	return cc.Client.OnboardConsumer(ctx, req)
+	return cc.Client.OnboardConsumer(apiCtx, req)
 }
 
 // GetStorageConfig generates the json config for connecting to storage provider cluster
-func (cc *OCSProviderClient) GetStorageConfig(consumerUUID string) (*pb.StorageConfigResponse, error) {
+func (cc *OCSProviderClient) GetStorageConfig(ctx context.Context, consumerUUID string) (*pb.StorageConfigResponse, error) {
+	if cc.Client == nil || cc.clientConn == nil {
+		return nil, fmt.Errorf("provider client or connection is nil")
```
nb-ohad, line 68:
```suggestion
		return nil, fmt.Errorf("provider client or connection was closed")
```

services/provider/client/client.go
```diff
@@ -10,82 +10,103 @@ import (
...
 
-	return cc.Client.OffboardConsumer(ctx, req)
+	return cc.Client.OffboardConsumer(apiCtx, req)
 }
 
 // UpdateCapacity increases or decreases the storage block pool size
-func (cc *OCSProviderClient) UpdateCapacity(consumerUUID, capacity string) (*pb.UpdateCapacityResponse, error) {
+func (cc *OCSProviderClient) UpdateCapacity(ctx context.Context, consumerUUID, capacity string) (*pb.UpdateCapacityResponse, error) {
+	if cc.Client == nil || cc.clientConn == nil {
+		return nil, fmt.Errorf("provider client or connection is nil")
```
nb-ohad, line 100:
```suggestion
		return nil, fmt.Errorf("provider client or connection was closed")
```

nb-ohad (approved):

services/provider/client/client.go
```diff
@@ -10,82 +10,103 @@ import (
...
-	return conn, err
+	return &OCSProviderClient{
+		Client:     pb.NewOCSProviderClient(conn),
+		clientConn: conn,
+		timeout:    timeout}, nil
+}
+
+// Close closes the gRPC connection of the external OCS storage provider client
+func (cc *OCSProviderClient) Close() {
+	_ = cc.clientConn.Close()
```
jarrpa, line 41: Ignoring an error? That sure sounds bad...

jarrpa (changes requested):
This commit is doing a lot. The fact that you have a bulleted list of changes itself means it should be chopped up. Offhand I'd want at least:

* Passing the parent context
* Adding the connection to `OCSProviderClient` + checking for closed connection

services/provider/client/client.go
nb-ohad: The only error that will fire is if the underlying grpc connection is already closed. And as the intent here to close anyway no need to handle the error

services/provider/client/client.go
```diff
@@ -10,82 +10,103 @@ import (
...
 }
 
 // OnboardConsumer to validate the consumer and create StorageConsumer
 // resource on the StorageProvider cluster
-func (cc *OCSProviderClient) OnboardConsumer(token, name, capacity string) (*pb.OnboardConsumerResponse, error) {
+func (cc *OCSProviderClient) OnboardConsumer(ctx context.Context, token, name, capacity string) (*pb.OnboardConsumerResponse, error) {
+	if cc.Client == nil || cc.clientConn == nil {
+		return nil, fmt.Errorf("provider client is closed")
+	}
+
```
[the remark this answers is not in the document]
nb-ohad: The entire construct is an abstraction on top of the JRPC client. 

In this specific abstraction, we only support a 1:1 relation between GRPC connections and clients and both of the objects share the same lifespan. The thing is that we want to release the underlying objects as soon as close() is called (to be garbage collected). So the check above just makes sure that none of the other methods are called after close() was called.

services/provider/client/client.go
```diff
@@ -10,82 +10,103 @@ import (
...
 
-	return conn, err
+	return &OCSProviderClient{
+		Client:     pb.NewOCSProviderClient(conn),
+		clientConn: conn,
+		timeout:    timeout}, nil
+}
+
+// Close closes the gRPC connection of the external OCS storage provider client
+func (cc *OCSProviderClient) Close() {
```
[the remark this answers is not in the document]
nb-ohad: The OCSProviderClient (for better or worse) is an abstraction on top of the underlying GRPC code. c
close() is to prevent leakage of the GRPC implementation outside the abstraction even if it was a single line of code and not three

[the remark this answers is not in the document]
jarrpa: Ah, okay.

services/provider/client/client.go
jarrpa: Fair enough

services/provider/client/client.go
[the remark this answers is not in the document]
jarrpa: Ah-ha, the garbage collection thing is something I didn't think of.... good enough. The only thing I'll say is that we should go a bit further and have something like an `isClosed()` method, since this is already repeated and would mirror the `Closed()` method. Still, no need to hold this PR for that, given the impending feature freeze.

jarrpa (approved):
@nb-ohad assuaged most of my concerns. Since I'm otherwise not asking for further changes, we can keep this PR as a single commit.

/lgtm

services/provider/client/client.go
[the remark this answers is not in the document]
nb-ohad: Agreed about the `isCLosed` observation. It will also allow us to expose that state of the client without exposing the implementation. We will take it in a future PR

PR #1460, opened 2022-01-25
unitTests: Add tests for ocs to ocs provider side code

iamniting:

### DIFF
--- a/controllers/storagecluster/provider_server_test.go
+++ b/controllers/storagecluster/provider_server_test.go
@@ -1,0 +1,252 @@
+package storagecluster
+
+import (
+	"context"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+
+	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v1"
+	"github.com/red-hat-storage/ocs-operator/controllers/util"
+	"github.com/red-hat-storage/ocs-operator/services/provider/server"
+)
+
+func TestOcsProviderServerEnsureCreated(t *testing.T) {
+
+	t.Run("Ensure that Deployment,Service,Secret is created when AllowRemoteStorageConsumers is enabled", func(t *testing.T) {
+
+		r, instance := createSetupForOcsProviderTest(t, true)
+
+		obj := &ocsProviderServer{}
+		_, err := obj.ensureCreated(r, instance)
+
+		assert.Error(t, err)
+		assert.Equal(t, "Deployment ocs-provider-server is not ready", err.Error())
+
+		deployment := &appsv1.Deployment{
+			ObjectMeta: metav1.ObjectMeta{Name: ocsProviderServerName},
+		}
+		assert.NoError(t, r.Client.Get(context.TODO(), client.ObjectKeyFromObject(deployment), deployment))
+		expectedDeployment := GetProviderAPIServerDeploymentForTest(instance)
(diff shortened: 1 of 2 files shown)


controllers/storagecluster/provider_server_test.go
```diff
@@ -0,0 +1,169 @@
...
+
+		assert.Error(t, err)
+		assert.Equal(t, "Deployment ocs-provider-server is not ready", err.Error())
+
+		deployment := &appsv1.Deployment{
+			ObjectMeta: metav1.ObjectMeta{Name: ocsProviderServerName},
+		}
+		assertFound(t, r.Client, deployment)
+		expectedDeployment := GetProviderAPIServerDeployment(instance)
+		assert.Equal(t, deployment.Spec, expectedDeployment.Spec)
```
nb-ohad, line 38: Shouldn't this need to be deep equal? or is `assert.equal` has deep equal semantics?

controllers/storagecluster/provider_server_test.go
```diff
@@ -0,0 +1,169 @@
...
+
+		service := &corev1.Service{
+			ObjectMeta: metav1.ObjectMeta{Name: ocsProviderServerName},
+		}
+		assertFound(t, r.Client, service)
+		expectedService := GetProviderAPIServerService(instance)
+		assert.Equal(t, expectedService.Spec, service.Spec)
```
nb-ohad, line 45: same as last comment

controllers/storagecluster/provider_server_test.go
```diff
@@ -0,0 +1,169 @@
...
+		ObjectMeta: metav1.ObjectMeta{Name: ocsProviderServerName},
+	}
+	assertNotFound(t, cli, service)
+
+	secret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{Name: ocsProviderServerName},
+	}
+	assertNotFound(t, cli, secret)
+
+}
```
nb-ohad, line 131: This function is used by a single test and is very specific, can we just move it into the test itself? 
Right now the reader needs to jump between two distinctive locations in the source code for no real reason.

controllers/storagecluster/provider_server_test.go
```diff
@@ -0,0 +1,169 @@
...
+	})
+}
+
+func assertFound(t *testing.T, cli client.Client, object client.Object) {
+
+	key := client.ObjectKeyFromObject(object)
+
+	err := cli.Get(context.TODO(), key, object)
+	assert.NoError(t, err)
+}
```
nb-ohad, line 103: Why do we need this util when we can just do this (in the call site) instead?
```suggestion
assert.NoError(t, cli.Get(context.TODO(), client.ObjectKeyFromObject(object), object))
```

controllers/storagecluster/provider_server_test.go
```diff
@@ -0,0 +1,169 @@
...
+
+func assertNotFound(t *testing.T, cli client.Client, object client.Object) {
+
+	key := client.ObjectKeyFromObject(object)
+
+	err := cli.Get(context.TODO(), key, object)
+	assert.Error(t, err)
+	assert.True(t, errors.IsNotFound(err))
+}
```
nb-ohad, line 112: `assert.True(t, errors.IsNotFound(err))` is already testing for `assert.Error(t, err)` and the entire thing can be compressed into 
```suggestion
assert.True(t, errors.IsNotFound(cli.Get(context.TODO(), client.ObjectKeyFromObject(object), object)))
```

controllers/storagecluster/provider_server_test.go
iamniting: It does verify it.

controllers/storagecluster/provider_server_test.go
iamniting: same as prev

controllers/storagecluster/provider_server_test.go
```diff
@@ -0,0 +1,169 @@
...
+		_, err := obj.ensureCreated(r, instance)
+
+		assert.Error(t, err)
+		assert.Equal(t, "Deployment ocs-provider-server is not ready", err.Error())
+
+		deployment := &appsv1.Deployment{
+			ObjectMeta: metav1.ObjectMeta{Name: ocsProviderServerName},
+		}
+		assertFound(t, r.Client, deployment)
+		expectedDeployment := GetProviderAPIServerDeployment(instance)
```
nb-ohad, line 37: Why are we testing against data coming from the code base itself? what if we have a bug in `GetProviderAPIServerDeployment`?

nb-ohad (approved):

PR #1470, opened 2022-02-01
api: split connection string into 2 new fields

iamniting:
remove connectionString and create 2 new fields:
- OnboardingTicket
- ExternalServerAddress

### DIFF
--- a/api/v1/storagecluster_types.go
+++ b/api/v1/storagecluster_types.go
@@ -162,9 +162,11 @@ type ExternalStorageClusterSpec struct {
 	// StorageProviderKind Identify the type of storage provider cluster this consumer cluster is going to connect to.
 	StorageProviderKind ExternalStorageKind `json:"storageProviderKind,omitempty"`
 
-	// ConnectionString An encoded string holding connection and identity information
-	// that is needed in order to establish connection with the storage providing cluster.
-	ConnectionString string `json:"connectionString,omitempty"`
+	// StorageProviderEndpoint holds info to establish connection with the storage providing cluster.
+	StorageProviderEndpoint string `json:"storageProviderEndpoint,omitempty"`
+
+	// OnboardingTicket holds an identity information required for consumer to onboard.
+	OnboardingTicket string `json:"onboardingTicket,omitempty"`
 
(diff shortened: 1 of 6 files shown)


api/v1/storagecluster_types.go
```diff
@@ -162,9 +162,11 @@ type ExternalStorageClusterSpec struct {
 	// StorageProviderKind Identify the type of storage provider cluster this consumer cluster is going to connect to.
 	StorageProviderKind ExternalStorageKind `json:"storageProviderKind,omitempty"`
 
-	// ConnectionString An encoded string holding connection and identity information
-	// that is needed in order to establish connection with the storage providing cluster.
-	ConnectionString string `json:"connectionString,omitempty"`
```
sp98, line 167: The code that removes usage of  `ConnectionString` should  also have been in the same commit. Right now its removed in the third commit.

nb-ohad (changes requested):

api/v1/storagecluster_types.go
```diff
@@ -162,9 +162,11 @@ type ExternalStorageClusterSpec struct {
...
+	// OnboardingTicket holds an identity information required for consumer to onboard.
+	OnboardingTicket string `json:"onboardingTicket,omitempty"`
+
+	// ExternalServerAddress holds info to establish connection with the storage providing cluster.
+	ExternalServerAddress string `json:"externalServerAddress,omitempty"`
```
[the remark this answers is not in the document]
iamniting: changed the name

api/v1/storagecluster_types.go
iamniting: Actually, we like to keep API and controller changes in the different commits.

nb-ohad (approved):

api/v1/storagecluster_types.go
sp98: This is a different scenario.  An API field is removed that is being used in the controller.  This commit on its own won't build. IMO, `ConnectionString` and all its references should have been removed in a single PR. 
Approving the PR.

sp98 (approved):

PR #1472, opened 2022-02-01
Update source-manifests.sh

b-ranto:
This makes the script use oc tool to extract the necessary files -- CSVs
and CRDs from rook and noobaa images. This should be less prone to
errors as it does not try to run the containers. It should also help us
sync with downstream where we can't use podman/docker to run the
containers.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,7 +77,7 @@ ocs-must-gather:
 	@echo "Building the ocs-must-gather image"
 	hack/build-must-gather.sh
 
-source-manifests: operator-sdk manifests kustomize
+source-manifests: operator-sdk manifests kustomize oc
 	@echo "Sourcing CSV and CRD manifests from component-level operators"
 	hack/source-manifests.sh
 
@@ -85,7 +85,7 @@ gen-protobuf:
 	@echo "Generating protobuf files for gRPC services"
 	hack/gen-protobuf.sh
 
-gen-latest-csv: operator-sdk manifests kustomize
+gen-latest-csv: operator-sdk manifests kustomize oc
 	@echo "Generating latest development CSV version using predefined ENV VARs."
 	hack/generate-latest-csv.sh
 
@@ -101,7 +101,7 @@ verify-latest-deploy-yaml: gen-latest-deploy-yaml
 	@echo "Verifying deployment yaml changes"
 	hack/verify-latest-deploy-yaml.sh
 
-gen-release-csv: operator-sdk manifests kustomize
+gen-release-csv: operator-sdk manifests kustomize oc
 	@echo "Generating unified CSV from sourced component-level operators"
 	hack/generate-unified-csv.sh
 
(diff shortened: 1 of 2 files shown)


b-ranto:
FYI: This worked sa expected DS.

umangachapagain:
CI is failing for 2 things:
`ocs-operator-ci` is failing shellcheck. Run it locally to fix the script.
`verify-latest-csv` failed because `oc` tool is not available inside the container that generates CSV. We'll have to install it.

b-ranto:
@umangachapagain I fixed the `ocs-operator-ci` shellcheck failures. I'm not sure what to do about the `oc` tool missing. Can you point me to a file that I should patch/fix?

Makefile
```diff
@@ -231,3 +231,20 @@ export KUSTOMIZE=$(GOBIN)/kustomize
...
+	cd $$OC_GEN_TMP_DIR ;\
+	export latest_oc_client_version=$$(curl https://mirror.openshift.com/pub/openshift-v4/clients/ocp/latest/ 2>/dev/null | grep -o \"openshift-client-linux-4.*tar.gz\" | tr -d \") ;\
+	curl -JL https://mirror.openshift.com/pub/openshift-v4/clients/ocp/latest/$${latest_oc_client_version} -o oc.tar.gz ;\
+	tar -xzvf oc.tar.gz ;\
+	mv oc ${GOBIN}/oc ;\
+	rm -rf $$OC_GEN_TMP_DIR ;\
+	}
+export OCS_OC_PATH=$(GOBIN)/oc
+else
+export OCS_OC_PATH=$(shell which oc)
```
jarrpa, line 249: Setting it this way now overrides the intended behavior of allowing the user to specify their own custom `oc` path (which I use regularly :wink: ). See here: https://github.com/red-hat-storage/ocs-operator/blob/main/hack/common.sh#L39

Similarly, if you don't explicit `export` the variable then it won't be available in any of the subshells called by `make`. In general I would rather see this in the `common.sh` file.

Makefile
```diff
@@ -231,3 +231,20 @@ export KUSTOMIZE=$(GOBIN)/kustomize
...
 export KUSTOMIZE=$(shell which kustomize)
 endif
+
+oc:
+ifeq (, $(shell which oc))
+	@{ \
+	set -e ;\
+	OC_GEN_TMP_DIR=$$(mktemp -d) ;\
+	cd $$OC_GEN_TMP_DIR ;\
+	export latest_oc_client_version=$$(curl https://mirror.openshift.com/pub/openshift-v4/clients/ocp/latest/ 2>/dev/null | grep -o \"openshift-client-linux-4.*tar.gz\" | tr -d \") ;\
```
jarrpa, line 241: Is this really the best way to determine the latest version? Shouldn't we be using https://mirror.openshift.com/pub/openshift-v4/clients/ocp/latest/release.txt instead?

Even so, we don't always want to be using the latest OCP binaries. While this is fine for `main`, we probably want to restrict ourselves to the target version of our current release.

hack/source-manifests.sh
```diff
@@ -40,37 +40,42 @@ function dump_noobaa_csv() {
...
 	mkdir -p $noobaa_crds_outdir
 
-	echo "Dumping Noobaa csv using command: $IMAGE_RUN_CMD --entrypoint=/usr/local/bin/noobaa-operator $NOOBAA_IMAGE $noobaa_dump_csv_cmd"
-	# shellcheck disable=SC2086
-	($IMAGE_RUN_CMD --entrypoint=/usr/local/bin/noobaa-operator "$NOOBAA_IMAGE" $noobaa_dump_csv_cmd) > $NOOBAA_CSV
-	echo "Dumping Noobaa crds using command: $IMAGE_RUN_CMD --entrypoint=/usr/local/bin/noobaa-operator $NOOBAA_IMAGE $noobaa_dump_crds_cmd"
-	# shellcheck disable=SC2086
-	($IMAGE_RUN_CMD --entrypoint=/usr/local/bin/noobaa-operator "$NOOBAA_IMAGE" $noobaa_dump_crds_cmd) > $noobaa_crds_outdir/noobaa-crd.yaml
+	echo "Dumping Noobaa csv and crds"
+	${OCS_OC_PATH} image extract --confirm "${NOOBAA_IMAGE}" --path /usr/local/bin/noobaa-operator:./
```
jarrpa, line 44: Use `mktemp` for this as you did for Rook-Ceph.

jarrpa (changes requested):
Oh.... I didn't know about the image extract command! This is exciting!

malayparida2000:
@b-ranto Is this PR still required?

PR #1477, opened 2022-02-02
api: Add StorageProviderEndpoint field in the storageCluster status 

iamniting:
StorageProviderEndpoint will be used to save the endpoint info on the
provider cluster.

### DIFF
--- a/controllers/storagecluster/provider_server.go
+++ b/controllers/storagecluster/provider_server.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"math/rand"
 	"os"
+	"sort"
 
 	"go.uber.org/multierr"
 	appsv1 "k8s.io/api/apps/v1"
@@ -23,6 +24,9 @@ import (
 const (
 	ocsProviderServerName  = "ocs-provider-server"
 	providerAPIServerImage = "PROVIDER_API_SERVER_IMAGE"
+
+	ocsProviderServicePort     = int32(50051)
+	ocsProviderServiceNodePort = int32(31659)
 )
 
 type ocsProviderServer struct{}
@@ -80,6 +84,7 @@ func (o *ocsProviderServer) ensureDeleted(r *StorageClusterReconciler, instance 
 
 	if finalErr == nil {
 		r.Log.Info("Resource deletion for provider succeeded")
+		instance.Status.StorageProviderEndpoint = ""
 	}
 
 	return reconcile.Result{}, finalErr
@@ -155,6 +160,20 @@ func (o *ocsProviderServer) createService(r *StorageClusterReconciler, instance 
 	}
 
 	r.Log.Info("Service create/update succeeded")
+
+	nodeAddresses, err := r.getWorkerNodesInternalIPAddresses()
+	if err != nil {
+		return err
+	}
+
+	if len(nodeAddresses) == 0 {
+		return fmt.Errorf("Did not found any worker node addresses")
+	}
+
+	instance.Status.StorageProviderEndpoint = nodeAddresses[0] + ":" + fmt.Sprint(ocsProviderServiceNodePort)
+
+	r.Log.Info("status.storageProviderEndpoint is updated", "Endpoint", instance.Status.StorageProviderEndpoint)
+
 	return nil
 }
 
@@ -187,6 +206,36 @@ func (o *ocsProviderServer) ensureDeploymentReplica(actual, desired *appsv1.Depl
 	}
 
 	return nil
+}
+
+// getWorkerNodesInternalIPAddresses return slice of Internal IPAddress of worker nodes
+func (r *StorageClusterReconciler) getWorkerNodesInternalIPAddresses() ([]string, error) {
+
+	nodes := &corev1.NodeList{}
+
+	err := r.Client.List(context.TODO(), nodes)
+	if err != nil {
+		r.Log.Error(err, "Failed to list nodes")
+		return nil, err
+	}
+
+	nodeAddresses := []string{}
+
+	for i := range nodes.Items {
+		node := &nodes.Items[i]
+		if _, ok := node.ObjectMeta.Labels["node-role.kubernetes.io/worker"]; ok {
+			for _, address := range node.Status.Addresses {
+				if address.Type == corev1.NodeInternalIP {
+					nodeAddresses = append(nodeAddresses, address.Address)
+					break
+				}
+			}
+		}
+	}
+
+	sort.Strings(nodeAddresses)
+
+	return nodeAddresses, nil
 }
 
 func GetProviderAPIServerDeployment(instance *ocsv1.StorageCluster) *appsv1.Deployment {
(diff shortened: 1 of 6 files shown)


controllers/storagecluster/provider_server.go
```diff
@@ -189,6 +192,38 @@ func (o *ocsProviderServer) ensureDeploymentReplica(actual, desired *appsv1.Depl
...
+		return err
+	}
+
+	nodeAddress := []string{}
+
+	for _, node := range nodes.Items {
+		if _, ok := node.ObjectMeta.Labels["node-role.kubernetes.io/worker"]; ok {
+			for _, address := range node.Status.Addresses {
+				if address.Type == corev1.NodeInternalIP {
+					nodeAddress = append(nodeAddress, address.Address)
```
sp98, line 212: Any reasons for using `nodeAddress` as `[]string`.   If only first entry is needed, then maybe we can use it as string and break the loop here.

iamniting: I am doing this on purpose. As I am not sure about the order of nodes as it is a list. I am getting all of them and sorting it so that I will always use the same one.

controllers/storagecluster/provider_server.go
```diff
@@ -189,6 +192,38 @@ func (o *ocsProviderServer) ensureDeploymentReplica(actual, desired *appsv1.Depl
 	return nil
 }
 
+// updateStorageProviderEndpointInStatus updates the status.storageProviderEndpoint which is required for consumer to establish a connection with provider
+func (r *StorageClusterReconciler) updateStorageProviderEndpointInStatus(instance *ocsv1.StorageCluster, nodePort int32) error {
```
nb-ohad, line 196: Can we have this as a util that returns the address instead of setting the address on the status?
This way you can call it above and update the status as part of the service reconcile

controllers/storagecluster/provider_server.go
iamniting: I am taking a single address only from each node as I have an if condition and there will be only one `NodeInternalIP`, I will put a `break`

iamniting: made changes

controllers/storagecluster/provider_server.go

controllers/storagecluster/provider_server.go
```diff
@@ -155,6 +157,16 @@ func (o *ocsProviderServer) createService(r *StorageClusterReconciler, instance
 	}
 
 	r.Log.Info("Service create/update succeeded")
+
+	nodeAddresses, err := r.GetWorkerNodesInternalIPAddresses()
+	if err != nil {
+		return err
+	}
+
+	instance.Status.StorageProviderEndpoint = nodeAddresses[0] + ":" + fmt.Sprint(actualService.Spec.Ports[0].NodePort)
```
nb-ohad, line 166: I think it is a bad idea to access the array by constant index. Do we give the port a name? if so we can search by name

iamniting: made changes on both

controllers/storagecluster/provider_server.go
```diff
@@ -155,6 +157,29 @@ func (o *ocsProviderServer) createService(r *StorageClusterReconciler, instance
...
 
 	r.Log.Info("Service create/update succeeded")
+
+	nodeAddresses, err := r.getWorkerNodesInternalIPAddresses()
+	if err != nil {
+		return err
+	}
+
+	if len(nodeAddresses) == 0 {
+		return fmt.Errorf("Did not found any worker nodes")
```
nb-ohad, line 167:
```suggestion
		return fmt.Errorf("Did not found any worker node addresses")
```

controllers/storagecluster/provider_server.go
```diff
@@ -155,6 +157,29 @@ func (o *ocsProviderServer) createService(r *StorageClusterReconciler, instance
...
+	}
+
+	var nodePortNumber int32
+
+	for _, port := range actualService.Spec.Ports {
+		if port.TargetPort == intstr.FromString("ocs-provider") {
+			nodePortNumber = port.NodePort
+			break
+		}
+	}
```
nb-ohad, line 177: I this code is not setting the port itself it means that immediately after creation there will be no port on the service, and we will need to wait for the service to be "reconciled" by k8s before a port will be allocated, so trying to get the pod address here is meanigless. 

I think the simplest approch will be to pre-select a port number + name and set it ourselves, so it is always know

controllers/storagecluster/provider_server.go
```diff
@@ -189,6 +214,35 @@ func (o *ocsProviderServer) ensureDeploymentReplica(actual, desired *appsv1.Depl
...
+
+	err := r.Client.List(context.TODO(), nodes)
+	if err != nil {
+		r.Log.Error(err, "Failed to list nodes")
+		return nil, err
+	}
+
+	nodeAddresses := []string{}
+
+	for _, node := range nodes.Items {
```
nb-ohad, line 230: I do not think NodeList.items contain node references, as such this for operation is creating deep copies of the nodes' CR structures. These structures are large and might affect memory and garbage collection. I will suggest to do the following instead:
```suggestion
	for i, _ := range nodes.Items {
		node := &range nodes.Items[i]
	
	
```

controllers/storagecluster/provider_server.go
```diff
@@ -189,6 +214,35 @@ func (o *ocsProviderServer) ensureDeploymentReplica(actual, desired *appsv1.Depl
...
+		if _, ok := node.ObjectMeta.Labels["node-role.kubernetes.io/worker"]; ok {
+			for _, address := range node.Status.Addresses {
```
nb-ohad, line 232: Not sure about the size of the address structure, but the last comment reasoning might also apply here.

As a general note using `for _, item := range slice` should be used when the slide slice either holds small data types (e.g numbers, booleans, etc.) or pointers.

iamniting: This will have only 2 fields keeping it as it is

controllers/storagecluster/provider_server.go
iamniting: I tested this out and verified that it had the port immediately after the creation. As the port is a part of the spec maybe the webhook is adding the value even before reconciling. But I can choose my own port, which will make the code cleaner.

sp98 (changes requested):

controllers/storagecluster/provider_server.go
```diff
@@ -189,6 +208,36 @@ func (o *ocsProviderServer) ensureDeploymentReplica(actual, desired *appsv1.Depl
...
+
+	err := r.Client.List(context.TODO(), nodes)
+	if err != nil {
+		r.Log.Error(err, "Failed to list nodes")
+		return nil, err
+	}
+
+	nodeAddresses := []string{}
+
+	for i := range nodes.Items {
```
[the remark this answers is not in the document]
nb-ohad: @sp98 I asked him to change that, your suggestion will create a big strain on the garbage collector as the node structure is a very big one and include mainly non-ref subtypes. so each iteration will clone that entire structure and will just throw it at the end of the loop.

So we replaced 
```
for i, node := range nodes.Items {
```
with
```
for i := range nodes.Items {
      node := &nodes.Items[i]
```

which will prevent cloning from happening.

nb-ohad (approved):

nb-ohad:
@iamniting I think it will be beneficial to have unit test around this logic, but I guess we can add it in a future PR

sp98 (approved):

PR #1481, opened 2022-02-03
Start using TLS in the API server

iamniting:


controllers/storagecluster/provider_server.go
```diff
@@ -249,6 +252,9 @@ func GetProviderAPIServerService(instance *ocsv1.StorageCluster) *corev1.Service
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      ocsProviderServerName,
 			Namespace: instance.Namespace,
+			Annotations: map[string]string{
+				serviceCertAnnotation: ocsProviderServerName + "-cert",
```
sp98, line 273: how about using a new global variable instead of  `ocsProviderServerName + "-cert"`

sp98 (changes requested):
Nit. Rest looks good to me.

controllers/storagecluster/provider_server.go

sp98 (approved):
Thanks for the quick PR. Looks good to me if there are no issues when testing it.

controllers/storagecluster/provider_server.go
```diff
@@ -139,8 +142,9 @@ func (o *ocsProviderServer) createService(r *StorageClusterReconciler, instance
 	desiredService := GetProviderAPIServerService(instance)
 	actualService := &corev1.Service{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      desiredService.Name,
-			Namespace: desiredService.Namespace,
+			Name:        desiredService.Name,
+			Namespace:   desiredService.Namespace,
+			Annotations: desiredService.Annotations,
```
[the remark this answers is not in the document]
iamniting: I did verified it and it is working well

services/provider/server/server.go
```diff
@@ -155,6 +156,15 @@ func (s *OCSProviderServer) Start(port int, opts []grpc.ServerOption) {
 		klog.Fatalf("failed to listen: %v", err)
 	}
 
+	certFile := "/mnt/cert/tls.crt"
+	keyFile := "/mnt/cert/tls.key"
```
nb-ohad, line 160: This can get out of sync with the mount point added to the service spec. 
Can we have the mount point const in a shared file (maybe util file) and here we will just concatenate the file names?

iamniting: done

controllers/storagecluster/provider_server.go
```diff
@@ -150,6 +154,12 @@ func (o *ocsProviderServer) createService(r *StorageClusterReconciler, instance
 			desiredService.Spec.ClusterIP = actualService.Spec.ClusterIP
 			desiredService.Spec.IPFamilies = actualService.Spec.IPFamilies
 
+			if actualService.Annotations == nil {
+				actualService.Annotations = desiredService.Annotations
+			} else {
+				actualService.Annotations[serviceCertAnnotation] = ocsProviderCertSecretName
+			}
+
```
nb-ohad, line 162: We might want to support more annotation in the template in the future:
```suggestion
			if actualService.Annotations == nil {
				actualService.Annotations = map[string]string{}
			} 
			for key, value := range desiredService.Annotations {
				actualService.Annotations[key] = value
			}
```

services/provider/server/server.go
```diff
@@ -24,6 +25,8 @@ import (
 
 const (
 	TicketAnnotation = "ocs.openshift.io/provider-onboarding-ticket"
+
+	SecretMountPoint = "/mnt/cert"
```
nb-ohad, line 29: This is imported in other packages, can we call it `ProviderCertsMountPoint`?

iamniting: yeah

controllers/storagecluster/provider_server.go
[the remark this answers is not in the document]

nb-ohad (approved):

PR #1482, opened 2022-02-03
Skip SC creation for vault SA based kms encryption

GowthamShanmugam:
4.10 epic covers vault SA based kms encryption only for cluster-wide encryption. 
After discussing with QE and PM, we are decided to skip this feature for now.

### DIFF
--- a/controllers/storagecluster/storageclasses.go
+++ b/controllers/storagecluster/storageclasses.go
@@ -287,8 +287,14 @@ func (r *StorageClusterReconciler) newStorageClassConfigurations(initData *ocsv1
 	if initData.Spec.Encryption.StorageClass && initData.Spec.Encryption.KeyManagementService.Enable {
 		kmsConfig, err := getKMSConfigMap(KMSConfigMapName, initData, r.Client)
 		if err == nil && kmsConfig != nil {
-			serviceName := kmsConfig.Data["KMS_SERVICE_NAME"]
-			ret = append(ret, newEncryptedCephBlockPoolStorageClassConfiguration(initData, serviceName))
+			authMethod, found := kmsConfig.Data["VAULT_AUTH_METHOD"]
+			if found && authMethod != VaultTokenAuthMethod {
+				// for 4.10, skipping SC creation for vault SA based kms encryption
+				r.Log.Info("Only vault token based auth method is supported for PV encryption", "VaultAuthMethod", authMethod)
+			} else {
+				serviceName := kmsConfig.Data["KMS_SERVICE_NAME"]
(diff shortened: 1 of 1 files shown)


SanjalKatiyar:
LGTM

controllers/storagecluster/storageclasses.go
```diff
@@ -287,8 +287,12 @@ func (r *StorageClusterReconciler) newStorageClassConfigurations(initData *ocsv1
 	if initData.Spec.Encryption.StorageClass && initData.Spec.Encryption.KeyManagementService.Enable {
 		kmsConfig, err := getKMSConfigMap(KMSConfigMapName, initData, r.Client)
 		if err == nil && kmsConfig != nil {
-			serviceName := kmsConfig.Data["KMS_SERVICE_NAME"]
-			ret = append(ret, newEncryptedCephBlockPoolStorageClassConfiguration(initData, serviceName))
+			authMethod, found := kmsConfig.Data["VAULT_AUTH_METHOD"]
+			// for 4.10, skipping SC creation for vault SA based kms encryption
```
umangachapagain, line 291: need an INFO level log stating why this was skipped.

controllers/storagecluster/storageclasses.go
```diff
@@ -287,8 +287,12 @@ func (r *StorageClusterReconciler) newStorageClassConfigurations(initData *ocsv1
...
+			if !(found && authMethod == VaultSAAuthMethod) {
```
umangachapagain, line 292:
```suggestion
			if found && authMethod != VaultSAAuthMethod {
```

umangachapagain (changes requested):

controllers/storagecluster/storageclasses.go
GowthamShanmugam: We are supporting multiple KMS services:
    1. Vault
    2. IBM Hpcs
1.  The condition should be if kms service is vault and auth method is not Kubernetes then continue.
2. Also, if KMS service is not Vault then also continue.
 
Just to reduce if condition i used this, 

    !(found && authMethod == VaultSAAuthMethod) always true for non Vault  kms services. because other kms service dont have "VAULT_AUTH_METHOD".
    For Vault it will be true only authMethod is not equal to Kubernetes

GowthamShanmugam: i cant use found && authMethod != VaultSAAuthMethod because found always false for non Vault kms service.

GowthamShanmugam:
@umangachapagain Changes done

umangachapagain (approved):

PR #1484, opened 2022-02-03
Remove clusterresourcequota metrics

synarete:
Exporting clusterresourcequota via OCS is duplication of metrics which are already exported by OpenShift. Removed.

Refs BZ-2016513


umangachapagain:
Removing metrics is fine. But, do we also want to remove alerts? Wouldn't it be better to use existing metrics to generate this alert for OCS specific clusterresourcequotas?

synarete:
> 

@umangachapagain  This is a very good question: should we create alert rules for something which is no longer part of OCS metrics. @jarrpa What is your opinion?

jarrpa (approved):
For the moment I'll say no. If someone is actually impacted by this we can always add them back.

/lgtm

PR #1485, opened 2022-02-03
validate onboarding ticket signature.

sp98:
This PR validates the signature of the onboarding ticket
used by the consumer cluster to onboard a storage provider cluster

Note: This PR assumes that the public key secret to be created by the deployer will be called `onboarding-ticket-key`

### DIFF
--- a/services/provider/server/server.go
+++ b/services/provider/server/server.go
@@ -2,7 +2,13 @@ package server
 
 import (
 	"context"
+	"crypto"
+	"crypto/md5"
+	"crypto/rsa"
+	"crypto/x509"
+	"encoding/base64"
 	"encoding/json"
+	"encoding/pem"
 	"fmt"
 	"net"
 	"os"
@@ -32,9 +38,9 @@ import (
 )
 
 const (
-	TicketAnnotation = "ocs.openshift.io/provider-onboarding-ticket"
-
-	ProviderCertsMountPoint = "/mnt/cert"
+	TicketAnnotation          = "ocs.openshift.io/provider-onboarding-ticket"
+	ProviderCertsMountPoint   = "/mnt/cert"
+	onboardingTicketKeySecret = "onboarding-ticket-key"
 )
 
 const (
@@ -81,9 +87,16 @@ func (s *OCSProviderServer) OnboardConsumer(ctx context.Context, req *pb.Onboard
 	}
 
 	// Validate onboardingTicket
-	// TODO: check if ticket is already used.
 	// TODO: check expiry of the ticket
-	// TODO: validate ticket with public key
+	pubKey, err := s.getOnboardingValidationKey(ctx)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to get public key to validate onboarding ticket for consumer %q. %v", req.ConsumerName, err)
+	}
+
+	if err := validateTicket(req.OnboardingTicket, pubKey); err != nil {
+		klog.Errorf("failed to validate onboarding ticket for consumer %q. %v", req.ConsumerName, err)
+		return nil, status.Errorf(codes.InvalidArgument, "onboarding ticket is not valid. %v", err)
+	}
 
 	storageConsumerUUID, err := s.consumerManager.Create(ctx, req.ConsumerName, req.OnboardingTicket, capacity)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


services/provider/server/server.go
```diff
@@ -201,3 +216,56 @@ func newClient() (client.Client, error) {
...
+
+	key, err := x509.ParsePKIXPublicKey(block.Bytes)
+	if err != nil {
+		return fmt.Errorf("failed to parse public key. %v", err)
+	}
+
+	pubKey := key.(*rsa.PublicKey)
+
+	ticketArr := strings.Split(string(ticket), ".")
+	if len(ticketArr) != 0 {
```
iamniting, line 249:
```suggestion
	if len(ticketArr) == 0 {
```

iamniting (changes requested):

services/provider/server/server.go
```diff
@@ -201,3 +216,56 @@ func newClient() (client.Client, error) {
 
 	return client, nil
 }
+
+func (s *OCSProviderServer) getPublicKey(ctx context.Context, secretName string) (string, error) {
```
nb-ohad, line 220: Can we please rename it to `getOnboardingValidationKey`? the current name is very generic and loses meaning without looking at the call site?

services/provider/server/server.go
```diff
@@ -201,3 +216,56 @@ func newClient() (client.Client, error) {
...
+	if block == nil {
+		return fmt.Errorf("invalid PEM block")
+	}
+
+	key, err := x509.ParsePKIXPublicKey(block.Bytes)
+	if err != nil {
+		return fmt.Errorf("failed to parse public key. %v", err)
+	}
+
+	pubKey := key.(*rsa.PublicKey)
```
nb-ohad, line 246: Why is this parsing of the public key here and not in the `getPublicKey` function? 
It will be better to separate the concerns of getting a usable public key (raw is not usable) and validation of the ticket using that public key (this function)

services/provider/server/server.go
nb-ohad, line 220: Can we please rename it to getOnboardingValidationKey? the current name is very generic and loses meaning without looking at the call site?

services/provider/server/server.go
nb-ohad, line 246: This seems like something that will fit better in the `getPublicKey` function.
Returning a raw public key from `getPublicKey` seems unnecessary and I will prefer to separate the concerns in a way  that will allow this function to only focus on validation the onboarding ticker

services/provider/server/server.go
nb-ohad: I think a better test will be 
```suggestion
	if len(ticketArr) != 2 {
```
As we do assume (later in the function), exactly 2 parts.

services/provider/server/server.go
```diff
@@ -201,3 +216,56 @@ func newClient() (client.Client, error) {
...
+	message, err := base64.StdEncoding.DecodeString(ticketArr[0])
+	if err != nil {
+		return fmt.Errorf("failed to decode payload. %v", err)
+	}
+	signature, err := base64.StdEncoding.DecodeString(ticketArr[1])
+	if err != nil {
+		return fmt.Errorf("failed to decode sigature. %v", err)
+	}
+
+	hash := sha256.Sum256(message)
```
nb-ohad, line 262: Why SHA256? we should use a much simpler hash as we only care about the randomness not the strength of the hash. SHA256 is very expensive. Can we please more to MD5?

services/provider/server/server.go
```diff
@@ -201,3 +216,56 @@ func newClient() (client.Client, error) {
...
+	err = rsa.VerifyPKCS1v15(pubKey, crypto.SHA1, hash[:], signature)
+	if err != nil {
+		return fmt.Errorf("failed to verify onboarding ticket signature. %v", err)
+	}
+
+	klog.Info("Successfully verified onboarding ticket")
+
+	return nil
```
nb-ohad, line 270: Should return the unencoded payload (preferably as a go struct) so the calling function could use the validated ticket content to verify other aspects of the payload (like expiration, and caps in the future))

nb-ohad (changes requested):

services/provider/server/server.go
```diff
@@ -68,9 +76,16 @@ func (s *OCSProviderServer) OnboardConsumer(ctx context.Context, req *pb.Onboard
...
 	// TODO: check expiry of the ticket
-	// TODO: validate ticket with public key
+	pubKey, err := s.getPublicKey(ctx, onboardingTicketKeySecret)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to get public key secret to validate onboarding ticket for consumer %q. %v", req.ConsumerName, err)
+	}
+
+	if err := validateTicket(req.OnboardingTicket, pubKey); err != nil {
+		klog.Errorf("failed to validate onboarding ticket for consumer %q. %v", req.ConsumerName, err)
+		return nil, status.Errorf(codes.InvalidArgument, "onboarding token is not valid. %v", err)
```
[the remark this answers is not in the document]
sp98: done

services/provider/server/server.go
sp98: done

services/provider/server/server.go
sp98: using `getOnboardingValidationKey`.  This returns the `rsa.PublicKey`

services/provider/server/server.go
sp98: using MD5 now.

services/provider/server/server.go
sp98: using `NewOnboardingTicket` to return a go struct with decoded payload and signature.  Now `Validate` function is only concerned about the validation of the ticket.

services/provider/server/server.go
sp98: using getOnboardingValidationKey. This returns the rsa.PublicKey, rather than the raw public key.

nb-ohad (approved):

nb-ohad:
@kesavanvt Can you please review?

services/provider/server/server.go
```diff
@@ -201,3 +216,54 @@ func newClient() (client.Client, error) {
...
+		return fmt.Errorf("invalid ticket")
+	}
+
+	message, err := base64.StdEncoding.DecodeString(ticketArr[0])
+	if err != nil {
+		return fmt.Errorf("failed to decode payload. %v", err)
+	}
+	signature, err := base64.StdEncoding.DecodeString(ticketArr[1])
+	if err != nil {
+		return fmt.Errorf("failed to decode sigature. %v", err)
```
kesavanvt, line 257:
```suggestion
		return fmt.Errorf("failed to decode signature. %v", err)
```

jarrpa (approved):
/lgtm

kesavanvt (approved):
lgtm

PR #1492, opened 2022-02-07
StorageConsumer: [CephFS] Add controller changes to provide storage service for consumers 

kesavanvt:

PR #1496, opened 2022-02-08
ocs-to-ocs: fetch cephFilesystemSubVolumeGroup resources

subhamkrai:
this commit fetch cephFilesystemSubVolumeGroup resources
from rook and append it to the external resource list.

### DIFF
--- a/services/provider/server/server.go
+++ b/services/provider/server/server.go
@@ -5,8 +5,10 @@ import (
 	"crypto"
 	"crypto/md5"
 	"crypto/rsa"
+	"crypto/sha256"
 	"crypto/x509"
 	"encoding/base64"
+	"encoding/hex"
 	"encoding/json"
 	"encoding/pem"
 	"fmt"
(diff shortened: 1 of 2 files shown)


subhamkrai:
on top of #1492

services/provider/server/server.go
```diff
@@ -334,6 +336,40 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
+			provisionerCephClientSecret, err := s.getCephClientSecretName(ctx, i.CephClients["provisioner"])
+			if err != nil {
+				return nil, err
+			}
+
+			extR = append(extR, &pb.ExternalResource{
+				Name: "cephfs",
+				Kind: "StorageClass",
+				Data: mustMarshal(map[string]string{
+					"clusterID": hash(clusterID),
```
nb-ohad, line 361: Why are we splitting the creation of the cluster id into two places? (here and Line 345)
Can we just do it here, or there?

services/provider/server/server.go
```diff
@@ -334,6 +336,40 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
 					"csi.storage.k8s.io/node-stage-secret-name":        nodeCephClientSecret,
 					"csi.storage.k8s.io/controller-expand-secret-name": provisionerCephClientSecret,
 				})})
+		case "CephFilesystemSubVolumeGroup":
+			subVolumeGroup := &rookCephv1.CephFilesystemSubVolumeGroup{}
+			err := s.client.Get(ctx, types.NamespacedName{Name: i.Name, Namespace: s.namespace}, subVolumeGroup)
+			if err != nil {
+				return nil, fmt.Errorf("failed to get %s cephFilesystemSubVolumeGroup. %v", i.Name, err)
+			}
+			clusterID := fmt.Sprintf("%s-%s-file-%s", subVolumeGroup.Namespace, subVolumeGroup.Spec.FilesystemName, subVolumeGroup.Name)
```
nb-ohad, line 345: This is defined very far for the place it is used, any reason to do so?

services/provider/server/server.go
```diff
@@ -363,3 +399,8 @@ func mustMarshal(data map[string]string) []byte {
 	}
 	return newData
 }
+
+func hash(s string) string {
+	h := sha256.Sum256([]byte(s))
+	return hex.EncodeToString(h[:16])
+}
```
nb-ohad, line 406: Why do we need a util function to a 2 liner that is used only once? Can't we just embed this in the call site?

services/provider/server/server.go
```diff
@@ -334,6 +336,40 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
+
+			extR = append(extR, &pb.ExternalResource{
+				Name: "cephfs",
+				Kind: "StorageClass",
+				Data: mustMarshal(map[string]string{
+					"clusterID": hash(clusterID),
+					"csi.storage.k8s.io/provisioner-secret-name":       provisionerCephClientSecret,
+					"csi.storage.k8s.io/node-stage-secret-name":        nodeCephClientSecret,
+					"csi.storage.k8s.io/controller-expand-secret-name": provisionerCephClientSecret,
+				})})
```
nb-ohad, line 365: Why are we splitting the creation of the cluster id into two places? (here and Line 345)
Can we just do it here, or there?

nb-ohad, line 365: A suggestion related to my comment on the hash util (below)

services/provider/server/server.go
nb-ohad, line 361: A suggestion related to my comment on the hash util (below)

PR #1501, opened 2022-02-08
expose option to force remove the OSD in osd removal template

subhamkrai:

### DIFF
--- a/controllers/storagecluster/job_templates.go
+++ b/controllers/storagecluster/job_templates.go
@@ -34,6 +34,8 @@ var osdCleanupArgs = []string{
 	"osd",
 	"remove",
 	"--osd-ids=${FAILED_OSD_IDS}",
+	"--force-osd-removal",
+	"${FORCE_OSD_REMOVAL}",
 }
 
 // ensureCreated ensures if the osd removal job template exists
@@ -84,6 +86,17 @@ OSD removal is an advanced use case.
 In the event of errors or invalid user inputs,
 the Job will attempt to remove as many OSDs as can be processed and complete without returning an error condition.
 Users should always check for errors and success in the log of the finished OSD removal Job.`,
+			},
+			{
+				Name:        "FORCE_OSD_REMOVAL",
+				DisplayName: "Force OSD Removal",
+				Required:    false,
+				Value:       "false",
+				Description: `
+A flag indicating whether the OSD should be forcefully removed. Valid values are true or false.
+The default value is false. If an OSD is being removed that could lead to data loss, the OSD
+will not be removed by default. If you see the osd removal fails and you are sure the OSD
+should be removed, set this flag to true and run the job again.`,
 			},
 		},
 		Objects: []runtime.RawExtension{
--- a/controllers/storagecluster/job_templates_test.go
+++ b/controllers/storagecluster/job_templates_test.go
@@ -23,7 +23,7 @@ func TestJobTemplates(t *testing.T) {
 	}{
 		{
 			jobTemplateName:    "ocs-osd-removal",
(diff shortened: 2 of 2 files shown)


controllers/storagecluster/job_templates.go
```diff
@@ -85,6 +87,13 @@ In the event of errors or invalid user inputs,
 the Job will attempt to remove as many OSDs as can be processed and complete without returning an error condition.
 Users should always check for errors and success in the log of the finished OSD removal Job.`,
 			},
+			{
+				Name:        "FORCE_OSD_REMOVAL",
+				DisplayName: "FORCE_OSD_REMOVAL",
```
travisn, line 92:
```suggestion
				DisplayName: "Force OSD Removal",
```

controllers/storagecluster/job_templates.go
```diff
@@ -85,6 +87,13 @@ In the event of errors or invalid user inputs,
...
+				Required:    false,
```
travisn, line 93: I believe we also need to specify the default value as:
```
Value: "false",
```
Without that I assume the default is just empty, which will likely cause an error in Rook. But we need to test these template changes in an openshift cluster to confirm the default behavior.

controllers/storagecluster/job_templates.go
```diff
@@ -85,6 +87,13 @@ In the event of errors or invalid user inputs,
...
+				Description: `
+The parameter FORCE_OSD_REMOVAL require bool value.`,
```
travisn, line 95:
```suggestion
A flag indicating whether the OSD should be forcefully removed. Valid values are true or false. The default value is false. If an OSD is being removed that could lead to data loss, the OSD will not be removed by default. If you see the osd removal fails and you are sure the OSD should be removed, set this flag to true and run the job again.
```

travisn (changes requested):

subhamkrai:
> The changes look good. Can you comment on the testing you've been able to run? From an offline conversation I think it's good to go, but would be good to summarize it here.

So, when running the template without flag `force-osd-removal`, I was able to verify the default value was `false`. And when I ran the template with the flag, in the logs I was able to see `force-osd-removal=true` but osd was not purged because it was healthy(which is expected).

travisn:
/lgtm
@agarwal-mudit @jarrpa Can we get an approve on this?

subhamkrai:
> @subhamkrai: #1501 failed to apply on top of branch "release-4.8":
> 
> ```
> Applying: expose option to force remove the OSD in osd removal template
> Using index info to reconstruct a base tree...
> A	controllers/storagecluster/job_templates.go
> A	controllers/storagecluster/job_templates_test.go
> Falling back to patching base and 3-way merge...
> CONFLICT (modify/delete): controllers/storagecluster/job_templates_test.go deleted in HEAD and modified in expose option to force remove the OSD in osd removal template. Version expose option to force remove the OSD in osd removal template of controllers/storagecluster/job_templates_test.go left in tree.
> CONFLICT (modify/delete): controllers/storagecluster/job_templates.go deleted in HEAD and modified in expose option to force remove the OSD in osd removal template. Version expose option to force remove the OSD in osd removal template of controllers/storagecluster/job_templates.go left in tree.
> error: Failed to merge in the changes.
> hint: Use 'git am --show-current-patch=diff' to see the failed patch
> Patch failed at 0001 expose option to force remove the OSD in osd removal template
> When you have resolved this problem, run "git am --continue".
> If you prefer to skip this patch, run "git am --skip" instead.
> To restore the original branch and stop patching, run "git am --abort".
> ```

I'll create manual BP for 4.8, 4.7, 4.6

PR #1510, opened 2022-02-09
Update CSV merger to inject Volume Replication Operator CRDs

umangachapagain:
Volume Replication Operator CRDs are not owned by default.
Since it is the part of ocs-operator bundle, we manually
create a mock CSV for volume replication operator and allow
ocs-operator to assume ownership of it's resources.

### DIFF
--- a/deploy/bundle/manifests/volumereplication.crd.yaml
+++ b/deploy/bundle/manifests/volumereplication.crd.yaml
@@ -1,0 +1,165 @@
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.4.1
+  name: volumereplications.replication.storage.openshift.io
+spec:
+  group: replication.storage.openshift.io
+  names:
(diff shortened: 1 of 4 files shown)


iamniting (approved):

deploy/bundle/manifests/volumereplication.crd.yaml
```diff
@@ -0,0 +1,165 @@
+---
+apiVersion: apiextensions.k8s.io/v1
```
Madhu-1, line 2: is this a duplicate copy of the CRD? if there are any changes in the CRD in the volume replication operator. Do the same changes need to be done here also?

umangachapagain: No. Changes go to Rook. We pull it from Rook image.

b-ranto: Rook pulls the CRDs from the volume-replication github repo (and so do we DS) so there is no need to update them in the rook repo.

b-ranto (approved):

umangachapagain:
These resources will be owned by Rook. No change required.

Issue #1515, opened 2022-02-11
Provider/Consumer: Remove grpc.WithInsecure() param while making client connection with API server 


kesavanvt:
Remove grpc.WithInsecure() param while making client connection with API server and include necessary TLS details.

PR #1528, opened 2022-02-17
[WIP] skip verifying the storage provider server's cert and hostname

sp98:
InsecureSkipVerify controls whether a client verifies the server
certificate chain and host name. If InsecureSkipVerify is true, crypto/tls
accepts any certificate presented by the server and any host name in
that certificate

### DIFF
--- a/services/provider/client/client.go
+++ b/services/provider/client/client.go
@@ -2,11 +2,13 @@ package client
 
 import (
 	"context"
+	"crypto/tls"
 	"fmt"
 	"time"
 
 	pb "github.com/red-hat-storage/ocs-operator/services/provider/pb"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
 )
 
 type OCSProviderClient struct {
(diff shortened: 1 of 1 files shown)


iamniting (approved):

services/provider/client/client.go
```diff
@@ -20,8 +22,12 @@ func NewProviderClient(ctx context.Context, serverAddr string, timeout time.Dura
 	apiCtx, cancel := context.WithTimeout(ctx, timeout)
 	defer cancel()
 
+	config := &tls.Config{
+		InsecureSkipVerify: true,
+	}
+
```
Madhu-1, line 28: Is this a temp fix? Disabling security checks is dangerous and should be avoided? why not use valid certificates?

sp98: We are not sharing certs between clients and the server. So I think we are ok to skip this verification on the client side. Agree this is not very secure.   Also [see](https://github.com/red-hat-storage/ocs-operator/issues/1515#issuecomment-1037383999).

Madhu-1: From the official doc 
```
// InsecureSkipVerify controls whether a client verifies the server's
// certificate chain and host name. If InsecureSkipVerify is true, crypto/tls
// accepts any certificate presented by the server and any host name in that
// certificate. In this mode, TLS is susceptible to machine-in-the-middle
// attacks unless custom verification is used. This should be used only for
// testing or in combination with VerifyConnection or VerifyPeerCertificate.
InsecureSkipVerify [bool](https://pkg.go.dev/builtin#bool)
```

sp98: Yes.  I pasted the same comment in the description, but skipped the last couple of lines. ;)

PR #1529, opened 2022-02-17
ocs-to-ocs: update key in secret according to health checker

subhamkrai:
updating key and value in one of the data field in the secret
according to the health checker, name.


subhamkrai:
@leseb @nb-ohad

services/provider/server/server.go
```diff
@@ -278,9 +278,9 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
 		Name: monSecret,
 		Kind: "Secret",
 		Data: mustMarshal(map[string]string{
-			"fsid":         fsid,
-			"mon-secret":   "mon-secret",
-			"admin-secret": "admin-secret",
+			"fsid":          fsid,
+			"mon-secret":    "mon-secret",
+			"ceph-username": fmt.Sprintf("%s-%s", "cephclient-health-checker", consumerResource.Name),
```
leseb, line 283: How about a helper function instead of duplicating the `fmt.Sprintf()`?

subhamkrai:
I'll be updating the unit test

services/provider/server/server.go
leseb: Up to you!

subhamkrai:
> I'll be updating the unit test

done

nb-ohad (approved):

services/provider/server/server.go
```diff
@@ -274,13 +274,15 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
 	extR = append(extR, &pb.ExternalResource{
 		Name: monSecret,
 		Kind: "Secret",
 		Data: mustMarshal(map[string]string{
-			"fsid":         fsid,
-			"mon-secret":   "mon-secret",
-			"admin-secret": "admin-secret",
+			"fsid":          fsid,
+			"mon-secret":    "mon-secret",
+			"ceph-username": fmt.Sprintf("%s-%s", "cephclient-health-checker", consumerResource.Name),
```
leseb, line 285: You need to add `"ceph-secret"` as well which is the keyring of the `"ceph-username"`.

leseb (changes requested):

services/provider/server/server_test.go
```diff
@@ -284,12 +285,15 @@ func TestGetExternalResources(t *testing.T) {
 	assert.NoError(t, err)
 	assert.NotNil(t, storageConRes)
 
-	for i := range mockExtR {
-		data, err := json.Marshal(mockExtR[i].Data)
+	for i := range storageConRes.ExternalResource {
+		extResource := storageConRes.ExternalResource[i]
```
nb-ohad, line 289: Take the address here instead of the value to prevent cloning of the struct

services/provider/server/server_test.go
```diff
@@ -284,12 +285,15 @@ func TestGetExternalResources(t *testing.T) {
...
+		mockResoruce := mockExtR[extResource.Name]
```
nb-ohad, line 290: Same here, take the address.
Also, you do not know if the item is in the map or not (this is what the test is checking)
So better to do:
```suggestion
		mockResoruce, ok := mockExtR[extResource.Name]
		Assert.True(ok)
```

services/provider/server/server_test.go
```diff
@@ -304,12 +308,16 @@ func TestGetExternalResources(t *testing.T) {
 	}
 	storageConRes, err = server.GetStorageConfig(ctx, &req)
 	assert.NoError(t, err)
+	assert.NotEqual(t, storageConRes.ExternalResource, mockExtR)
 	for i := range storageConRes.ExternalResource {
-		data, err := json.Marshal(mockExtR[i].Data)
+		extResource := storageConRes.ExternalResource[i]
+		mockResoruce := mockExtR[extResource.Name]
```
nb-ohad, line 314: Same 2 comments as above

nb-ohad (changes requested):

PR #1543, opened 2022-02-21
ocs-to-ocs: get mgr pod hostIP

subhamkrai:
fetching mgr pod hostIP instead of clusterIP.

### DIFF
--- a/services/provider/server/server.go
+++ b/services/provider/server/server.go
@@ -278,23 +278,44 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
 		return nil, fmt.Errorf("secret %s data fsid is empty", monSecret)
 	}
 
-	// Service for monitoring endpoints
-	scMonitoring := &v1.Service{}
-	err = s.client.Get(ctx, types.NamespacedName{Name: "rook-ceph-mgr", Namespace: s.namespace}, scMonitoring)
+	// Get mgr pod hostIP
+	podList := &corev1.PodList{}
+	listOpts := []client.ListOption{
+		client.InNamespace(s.namespace),
+		client.MatchingLabels(map[string]string{"app": "rook-ceph-mgr"}),
+	}
+	err = s.client.List(ctx, podList, listOpts...)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get rook-ceph-mgr service. %v", err)
+		return nil, fmt.Errorf("failed to list pod with rook-ceph-mgr label. %v", err)
+	}
+	if len(podList.Items) == 0 {
+		return nil, fmt.Errorf("no pods available with rook-ceph-mgr label")
 	}
 
-	if scMonitoring.Spec.ClusterIP == "" || strconv.Itoa(int(scMonitoring.Spec.Ports[0].Port)) == "" {
-		return nil, fmt.Errorf("service rook-ceph-mgr clusterIP or port is empty")
+	mgrPod := &podList.Items[0]
+	var port int32 = -1
+
+	for i := range mgrPod.Spec.Containers {
+		container := &mgrPod.Spec.Containers[i]
+		if container.Name == "mgr" {
+			for j := range container.Ports {
+				if container.Ports[j].Name == "http-metrics" {
+					port = container.Ports[j].ContainerPort
+				}
+			}
+		}
+	}
+
+	if port < 0 {
+		return nil, fmt.Errorf("mgr pod port is empty")
 	}
 
 	extR = append(extR, &pb.ExternalResource{
 		Name: "monitoring-endpoint",
 		Kind: "CephCluster",
 		Data: mustMarshal(map[string]string{
-			"MonitoringEndpoint": scMonitoring.Spec.ClusterIP,
-			"MonitoringPort":     strconv.Itoa(int(scMonitoring.Spec.Ports[0].Port)),
+			"MonitoringEndpoint": mgrPod.Status.HostIP,
+			"MonitoringPort":     strconv.Itoa(int(port)),
 		})})
 
 	for _, i := range consumerResource.Status.CephResources {
--- a/deploy/bundle/manifests/provider-role.yaml
+++ b/deploy/bundle/manifests/provider-role.yaml
@@ -34,3 +34,10 @@ rules:
   - cephclients
   verbs:
   - get
+- apiGroups:
+    - ""
+  resources:
+    - pods
+  verbs:
+    - get
+    - list
--- a/rbac/provider-role.yaml
+++ b/rbac/provider-role.yaml
@@ -34,3 +34,10 @@ rules:
   - cephclients
   verbs:
   - get
+- apiGroups:
+    - ""
+  resources:
+    - pods
+  verbs:
+    - get
+    - list
--- a/services/provider/server/server_test.go
+++ b/services/provider/server/server_test.go
@@ -220,7 +220,6 @@ func TestGetExternalResources(t *testing.T) {
 	objects := []runtime.Object{
 		&v1.ConfigMap{},
 		&v1.Secret{},
-		&v1.Service{},
 		&rookCephv1.CephClient{},
 		consumerResource,
 		consumerResource1,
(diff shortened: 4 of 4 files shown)


services/provider/server/server.go
```diff
@@ -285,6 +285,19 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
 		return nil, fmt.Errorf("failed to get rook-ceph-mgr service. %v", err)
 	}
 
+	// Get mgr pod hostIP
+	podList := &corev1.PodList{}
+	err = s.client.List(ctx, podList, client.MatchingLabels(map[string]string{"app": "rook-ceph-mgr"}))
+	if err != nil || len(podList.Items) == 0 {
+		return nil, fmt.Errorf("failed to list pod with matching lable")
```
nb-ohad, line 292:
```suggestion
		return nil, fmt.Errorf("failed to list pod with rook-ceph-mgr label")
```

services/provider/server/server.go
```diff
@@ -285,6 +285,19 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
+	}
+
+	mgrPod := &v1.Pod{}
+	err = s.client.Get(ctx, types.NamespacedName{Name: podList.Items[0].Name, Namespace: s.namespace}, mgrPod)
+	if err != nil || len(podList.Items) == 0 {
+		return nil, fmt.Errorf("failed to get mgr pod")
+	}
```
nb-ohad, line 299: Why do you need to get the pod? the pod resource is already in the list

sp98 (changes requested):

services/provider/server/server.go
```diff
@@ -285,6 +285,19 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
+
 	if scMonitoring.Spec.ClusterIP == "" || strconv.Itoa(int(scMonitoring.Spec.Ports[0].Port)) == "" {
 		return nil, fmt.Errorf("service rook-ceph-mgr clusterIP or port is empty")
 	}
```
nb-ohad, line 303: This test is wrong, it looks for ClusterIP, not HostIP

nb-ohad (changes requested):

iamniting (changes requested):
Can you also add the permissions to read the pods?

services/provider/server/server.go
```diff
@@ -285,6 +285,19 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
+		return nil, fmt.Errorf("failed to list pod with matching lable")
+	}
+
+	mgrPod := &v1.Pod{}
+	err = s.client.Get(ctx, types.NamespacedName{Name: podList.Items[0].Name, Namespace: s.namespace}, mgrPod)
+	if err != nil || len(podList.Items) == 0 {
+		return nil, fmt.Errorf("failed to get mgr pod")
+	}
+
 	if scMonitoring.Spec.ClusterIP == "" || strconv.Itoa(int(scMonitoring.Spec.Ports[0].Port)) == "" {
```
sp98, line 301: you don't need to call `mgr` service anymore.  The port can be obtained from the mgr pod itself.

nb-ohad: @sp98 How would he know which port to look for?

services/provider/server/server.go
```diff
@@ -278,23 +278,29 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
+	podList := &corev1.PodList{}
+	err = s.client.List(ctx, podList, client.MatchingLabels(map[string]string{"app": "rook-ceph-mgr"}))
+	if err != nil || len(podList.Items) == 0 {
+		return nil, fmt.Errorf("failed to list pod with rook-ceph-mgr label. %v", err)
 	}
 
-	if scMonitoring.Spec.ClusterIP == "" || strconv.Itoa(int(scMonitoring.Spec.Ports[0].Port)) == "" {
-		return nil, fmt.Errorf("service rook-ceph-mgr clusterIP or port is empty")
+	var port int32
+	for i := range podList.Items[0].Spec.Containers[0].Ports {
```
nb-ohad, line 289: Why do we assume the pod has containers? what if Ceph will add additional containers? do we assume it will always be the first one?

I think we should search for the container by name.

services/provider/server/server.go
```diff
@@ -278,23 +278,29 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
+		if podList.Items[0].Spec.Containers[0].Ports[i].Name == "http-metrics" {
+			port = podList.Items[0].Spec.Containers[0].Ports[i].ContainerPort
+		}
```
nb-ohad, line 292: Let's make this a little be more readable 
```suggestion
		portSpec := &podList.Items[0].Spec.Containers[0].Ports;
		if portSpec.Name == "http-metrics" {
			port = portSpec.ContainerPort
		}
```

services/provider/server/server.go
```diff
@@ -278,23 +278,29 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
-	scMonitoring := &v1.Service{}
-	err = s.client.Get(ctx, types.NamespacedName{Name: "rook-ceph-mgr", Namespace: s.namespace}, scMonitoring)
-	if err != nil {
-		return nil, fmt.Errorf("failed to get rook-ceph-mgr service. %v", err)
+	// Get mgr pod hostIP
+	podList := &corev1.PodList{}
+	err = s.client.List(ctx, podList, client.MatchingLabels(map[string]string{"app": "rook-ceph-mgr"}))
+	if err != nil || len(podList.Items) == 0 {
+		return nil, fmt.Errorf("failed to list pod with rook-ceph-mgr label. %v", err)
 	}
```
nb-ohad, line 287: You are going to reference the first pod a couple of times, It will be preferable to take a pointer for it and use it instead of `podList.Items[0]`

```
mgrPod := &podList.Items[0]
```

nb-ohad (changes requested):

services/provider/server/server.go
iamniting: even if we have more than one container we should be able to use any of them IMO.

iamniting:
```
apiVersion: v1
kind: Pod
metadata:
  annotations:
    openshift.io/scc: rook-ceph
    prometheus.io/port: "9283"
    prometheus.io/scrape: "true"
  creationTimestamp: "2022-02-17T09:26:19Z"
  generateName: rook-ceph-mgr-a-5544b8b8d5-
  labels:
    app: rook-ceph-mgr
    app.kubernetes.io/component: cephclusters.ceph.rook.io
    app.kubernetes.io/created-by: rook-ceph-operator
    app.kubernetes.io/instance: a
    app.kubernetes.io/managed-by: rook-ceph-operator
    app.kubernetes.io/name: ceph-mgr
    app.kubernetes.io/part-of: ocs-storagecluster-cephcluster
    ceph_daemon_id: a
    ceph_daemon_type: mgr
    instance: a
    mgr: a
    pod-template-hash: 5544b8b8d5
    rook.io/operator-namespace: openshift-storage
    rook_cluster: openshift-storage
  name: rook-ceph-mgr-a-5544b8b8d5-5czgz
  namespace: openshift-storage
```

If we have more than one pod they should be like `rook-ceph-mgr-b`, ` rook-ceph-mgr-c` and we should be able to use any of them.

nb-ohad: @iamniting I think you are talking about multiple pods, not containers. 
If Ceph will add a sidecar container to the pod and have it as the first container our code will break. 
We should select the right container by name

sp98 (approved):

services/provider/server/server.go
iamniting: my bad, make sense.

services/provider/server/server.go
```diff
@@ -278,23 +278,32 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
+
+	mgrPod := &podList.Items[0]
+	var port int32
+	for i := range mgrPod.Spec.Containers {
+		if mgrPod.Spec.Containers[i].Name == "mgr" {
+			if mgrPod.Spec.Containers[0].Ports[i].Name == "http-metrics" {
+				port = mgrPod.Spec.Containers[0].Ports[i].ContainerPort
+			}
+		}
 	}
```
nb-ohad, line 296: This is wrong, you cannot use the same index for both the container and the port
```suggestion
	for i := range mgrPod.Spec.Containers {
		if mgrPod.Spec.Containers[i].Name == "mgr" {
			container := &mgrPod.Spec.Containers[i]
			for j := range container.Ports {
				if container.Ports[j].Name == "http-metrics" {
					port = container.Ports[j].ContainerPort
				}
			}
		}
	}
```

nb-ohad (changes requested):

PR #1548, opened 2022-02-22
Inject toleration to rook-ceph-tools

weirdwiz:
Get OCS toleration and apply it to rook-ceph-tools to handle non ocs
taints on nodes.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2012084


jarrpa (approved):
/lgtm

PR #1549, opened 2022-02-22
Uninstall: remove CephFilesystemSubVolumeGroup on external mode uninstall 

kesavanvt:
Fix offboarding in the consumer cluster by removing the relevant created external resource


controllers/storagecluster/external_resources.go
```diff
@@ -299,6 +299,30 @@ func (obj *ocsExternalResources) ensureDeleted(r *StorageClusterReconciler, inst
...
+							"CephFilesystemSubVolumeGroup",
+							klog.KRef(cephFilesystemSubVolumeGroup.Namespace, cephFilesystemSubVolumeGroup.Name),
+						)
+						return reconcile.Result{}, fmt.Errorf("uninstall: Failed to delete CephFilesystemSubVolumeGroup %v: %v", cephFilesystemSubVolumeGroup.Name, err)
+					}
+				}
+			}
+		}
+
 		return r.offboardConsumer(instance, externalClusterClient)
```
iamniting, line 326: I think we should offBoard before deleting these resources. Can you move this offboard call before the deletion?

iamniting (changes requested):

controllers/storagecluster/external_resources.go
kesavanvt: Fixed

iamniting: Hi @Madhu-1 here is a code where we delete all resources https://github.com/red-hat-storage/ocs-operator/blob/main/controllers/storagecluster/uninstall_reconciler.go#L316

iamniting (approved):
/lgtm

controllers/storagecluster/external_resources.go
```diff
@@ -299,7 +300,27 @@ func (obj *ocsExternalResources) ensureDeleted(r *StorageClusterReconciler, inst
...
+
+		for _, cephFilesystemSubVolumeGroup := range cephFilesystemSubVolumeGroupList.Items {
+			if err = r.Client.Delete(context.TODO(), &cephFilesystemSubVolumeGroup); err != nil && !errors.IsNotFound(err) {
+				r.Log.Error(
+					err,
+					"Uninstall: Failed to delete CephFilesystemSubVolumeGroup.",
+					"CephFilesystemSubVolumeGroup",
+					klog.KRef(cephFilesystemSubVolumeGroup.Namespace, cephFilesystemSubVolumeGroup.Name),
+				)
+				return reconcile.Result{}, fmt.Errorf("uninstall: Failed to delete CephFilesystemSubVolumeGroup %v: %v", cephFilesystemSubVolumeGroup.Name, err)
```
sp98, line 321:
```suggestion
				return reconcile.Result{}, fmt.Errorf("uninstall: Failed to delete CephFilesystemSubVolumeGroup %q %v", cephFilesystemSubVolumeGroup.Name, err)
```

sp98 (commented):
nit

controllers/storagecluster/external_resources.go
```diff
@@ -299,7 +300,27 @@ func (obj *ocsExternalResources) ensureDeleted(r *StorageClusterReconciler, inst
...
+
+		for _, cephFilesystemSubVolumeGroup := range cephFilesystemSubVolumeGroupList.Items {
+			if err = r.Client.Delete(context.TODO(), &cephFilesystemSubVolumeGroup); err != nil && !errors.IsNotFound(err) {
+				r.Log.Error(
+					err,
+					"Uninstall: Failed to delete CephFilesystemSubVolumeGroup.",
+					"CephFilesystemSubVolumeGroup",
+					klog.KRef(cephFilesystemSubVolumeGroup.Namespace, cephFilesystemSubVolumeGroup.Name),
+				)
+				return reconcile.Result{}, fmt.Errorf("uninstall: Failed to delete CephFilesystemSubVolumeGroup %q. %v", cephFilesystemSubVolumeGroup.Name, err)
```
nb-ohad, line 321: @kesavanvt 
Do we really need an error here? can we do a warning and keep deleting any other resources?
What will happen if we will keep the resource?

kesavanvt: AFAIF, this error will be raised in case of any Kubernetes issue. I think its better to return an error rather than just logging the warning. If we keep the resource, the cephcluster deletion will get stuck .

nb-ohad: Ok, so what about keeping an error flag outside the loop. 
Then for each resource that cannot be deleted add a warning and set the flag to true.
At the end of the loop if the flag is on, return a general error about being unable to delete resources.

WDYT?

agarwal-mudit:
/approve
On Ohad's behalf, there is a comment pending to be taken care but that can be taken in a follow-up PR

agarwal-mudit (approved):
/lgtm

PR #1555, opened 2022-02-24
Provider API server : Reduce length of the user secret name for consumer

kesavanvt:
This PR is to reduce the length of the user secret name that needs to be created in the consumer as Kubernetes expects the length should not be more than 63 characters when on provisioning volumes  

### DIFF
--- a/services/provider/server/server.go
+++ b/services/provider/server/server.go
@@ -338,7 +338,7 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
 				keyProp = "adminKey"
 			}
 			extR = append(extR, &pb.ExternalResource{
-				Name: clientSecretName,
+				Name: strings.ReplaceAll(clientSecretName, "-"+consumerResource.Name, ""),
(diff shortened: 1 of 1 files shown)


services/provider/server/server.go
```diff
@@ -338,7 +338,7 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
 				keyProp = "adminKey"
 			}
 			extR = append(extR, &pb.ExternalResource{
-				Name: clientSecretName,
+				Name: strings.ReplaceAll(clientSecretName, "-"+consumerResource.Name, ""),
```
nb-ohad, line 341: Why are we shortening the names after the fact? 
I think it will be preferable to just create a shorter name in the first place. 
We own that code and can change it there.

nb-ohad (changes requested):

nb-ohad:
Closing this one as we got to an agreement that this fix takes the wrong approach. 
Keeping this open until we have the "right" fix, just creates confusion.

PR #1562, opened 2022-02-24
hack: add tools for working with consumer onboarding tickets

jarrpa:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,8 @@ tools/csv-checksum/csv-checksum
 tools/cluster-deploy/cluster-deploy
 tools/shellcheck
 metrics/mixin/build/vendor
+
+# Ignore ticketgen output
+hack/ticketgen/*.pem
+hack/ticketgen/*.pub
+hack/ticketgen/*.txt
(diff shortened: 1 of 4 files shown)


iamniting (changes requested):
@jarrpa Can you also add a small readme file on how to use it?

jarrpa:
Not a bad idea in general, but since I already have usage text could that suffice? Specifically, I'm wondering about:

* What's the expected state of documentation for the customers, will they need a local README, or is everything expected to be communicated via downstream documentation?
* I heard there was a need to get this merged by Monday, and possibly backported as well?

Let me know if you still want it in this PR and I'll get to it in my morning.

iamniting:
> Not a bad idea in general, but since I already have usage text could that suffice? Specifically, I'm wondering about:

Yes, I see usage text, If I am not wrong usage text will tell about the current script only not the flow.
> 
> * What's the expected state of documentation for the customers, will they need a local README, or is everything expected to be communicated via downstream documentation?

I am not worried about that at this point, I am mainly worried about the ones who is going to try it for the first time before release.
> * I heard there was a need to get this merged by Monday, and possibly backported as well?

Yes that is true
 
> Let me know if you still want it in this PR and I'll get to it in my morning.

I would like to see a readme, It is not really required within this PR itself. you can create a follow-up PR.

sp98 (approved):

PR #1563, opened 2022-02-24
uninstall: delete CephCluster at the start of uninstall

jarrpa:
To properly trigger the new uninstallation logic in Rook-Ceph for a
"graceful" uninstall, we need to mark the CephCluster for deletion
towards the start of the process. This is how we inform Rook-Ceph that a
cluster deletion is in progress.

### DIFF
--- a/controllers/storagecluster/uninstall_reconciler.go
+++ b/controllers/storagecluster/uninstall_reconciler.go
@@ -136,19 +136,9 @@ func (r *StorageClusterReconciler) deleteNodeTaint(sc *ocsv1.StorageCluster) (er
 }
 
 // setRookUninstallandCleanupPolicy sets the uninstall mode and cleanup policy for rook based on the annotation on the StorageCluster
-func (r *StorageClusterReconciler) setRookUninstallandCleanupPolicy(instance *ocsv1.StorageCluster) (err error) {
+func (r *StorageClusterReconciler) setRookUninstallandCleanupPolicy(instance *ocsv1.StorageCluster, cephCluster *cephv1.CephCluster) (err error) {
 
-	cephCluster := &cephv1.CephCluster{}
 	var updateRequired bool
-
-	err = r.Client.Get(context.TODO(), types.NamespacedName{Name: generateNameForCephCluster(instance), Namespace: instance.Namespace}, cephCluster)
-	if err != nil {
-		if errors.IsNotFound(err) {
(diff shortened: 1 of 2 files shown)


nb-ohad:
@jarrpa Do we know how this will affect external mode / provider-consumer mode?
In provider consumer mode strict ordering is important, let's make sure we are not breaking it

@iamniting Can you please review the code and give your opnion

iamniting:
> @jarrpa Do we know how this will affect external mode / provider-consumer mode? In provider consumer mode strict ordering is important, let's make sure we are not breaking it
> 
> @iamniting Can you please review the code and give your opinion

@nb-ohad This change should not break our provider/consumer uninstall.

@jarrpa The ensureDeleted func will always return an error until the cephcluster is present and we won't be able to proceed further at all. If we just want to mark it for deletion we should change the func too.

umangachapagain (changes requested):
If we do this, CephCluster will be marked for deletion but since finalizer is set, it will not be removed yet. So `ensureDelete` is going to return an error and the deletion loop will never proceed forward until Rook removed finalizer.
For Rook to remove finalizer, all the consumers of CephCluster needs to be deleted but we can't do it since we're waiting on CephCluster to be removed.
Seems like a deadlock. This can be fixed by returning `nil` instead of error when CephCluster is present and deletion timestamp is set. But, then that would mean we'll never be able to guarantee that CephCluster resource is cleaned up.

jarrpa:
I verified the above comments, it just took me a while to get a testing cluster up and running.

I reverted the move of the ensureDeleted call and updated the annotating function instead. @travisn's suggestion is ideal, but that's a larger refactor than we really want to do right now.

PTAL

controllers/storagecluster/uninstall_reconciler.go
```diff
@@ -301,13 +300,20 @@ func (r *StorageClusterReconciler) deleteResources(sc *ocsv1.StorageCluster) (re
 		return reconcile.Result{}, err
 	}
 
-	err = r.setRookUninstallandCleanupPolicy(sc)
```
iamniting, line 304: Why are we removing this func call? This is required to tell rook about the kind of uninstall (force/graceful). We should not remove this.

iamniting (changes requested):

controllers/storagecluster/uninstall_reconciler.go
jarrpa: I pushed the wrong commit, this one was out of date

jarrpa: Also note this moved to the newCephCluster() function. I couldn't figure out why we weren't always setting them to begin with.

iamniting (approved):
/lgtm

PR #1564, opened 2022-02-25
[wip]:ocs-to-ocs: use hash to reduce length of user secret

subhamkrai:
this commit uses md5 hash to reduce the length of secret.
This contains storageconsumer name, claimID, cephusertype.
And, also adding annotation to the cephClient so that we can
identify the cephClient.


controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -130,25 +139,43 @@ func (r *StorageConsumerReconciler) initReconciler(request reconcile.Request) {
 	r.cephFilesystemSubVolumeGroup.Namespace = r.namespace
 
 	r.cephClientRBDProvisioner = &rookCephv1.CephClient{}
-	r.cephClientRBDProvisioner.Name = fmt.Sprintf("%s-%s", "cephclient-rbd-provisioner", r.storageConsumer.Name)
+	cephClientName := &cephClientFormat{StorageConsumerName: r.storageConsumer.Name, ClaimID: "rbd", CephUserType: "provisioner"}
+	r.cephClientRBDProvisioner.Name = generateHashForCephClient(cephClientName)
```
nb-ohad, line 143: Using a struct to serialize is an implementation detail and as such, I would expect it to be hidden inside ` generateHashForCephClient`.
```suggestion
	r.cephClientRBDProvisioner.Name = generateHashForCephClient(r.storageConsumer.Name, "rbd", "provisioner")
```

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -130,25 +139,43 @@ func (r *StorageConsumerReconciler) initReconciler(request reconcile.Request) {
...
 	r.cephClientRBDProvisioner.Namespace = r.namespace
+	r.cephClientRBDProvisioner.Annotations["cephUserType"] = fmt.Sprintf("%s-%s-%s", cephClientName.StorageConsumerName, cephClientName.ClaimID, cephClientName.CephUserType)
```
nb-ohad, line 145: 1. Let's use a separate annotation for each
2. Use the annotation name prefix we use for all ocs annotations (I believe it is `ocs.openshift.io`, but please verify it)
3. Move the annotation code into the reconciliation section, not the init section, the code here is used to fill only the info that is needed to fetch the resource. The main reason: if the annotation will not be set in the `createOrUpdate` mutate function then they might be overwritten when we load the resource.


```suggestion
	r.cephClientRBDProvisioner.Annotations["ocs.openshift.io.storageconsumer "] = cephClientName.StorageConsumerName
	r.cephClientRBDProvisioner.Annotations["ocs.openshift.io.storageclaim "] = cephClientName.ClaimID
	r.cephClientRBDProvisioner.Annotations["ocs.openshift.io.cephusertype "] = cephClientName.CephUserType
```

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -130,25 +139,43 @@ func (r *StorageConsumerReconciler) initReconciler(request reconcile.Request) {
...
 
 	r.cephClientRBDNode = &rookCephv1.CephClient{}
-	r.cephClientRBDNode.Name = fmt.Sprintf("%s-%s", "cephclient-rbd-node", r.storageConsumer.Name)
+	cephClientName = &cephClientFormat{StorageConsumerName: r.storageConsumer.Name, ClaimID: "rbd", CephUserType: "node"}
+	r.cephClientRBDNode.Name = generateHashForCephClient(cephClientName)
 	r.cephClientRBDNode.Namespace = r.namespace
+	r.cephClientRBDNode.Annotations["cephUserType"] = fmt.Sprintf("%s-%s-%s", cephClientName.StorageConsumerName, cephClientName.ClaimID, cephClientName.CephUserType)
```
nb-ohad, line 151: same as above

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -130,25 +139,43 @@ func (r *StorageConsumerReconciler) initReconciler(request reconcile.Request) {
...
 
 	r.cephClientCephFSProvisioner = &rookCephv1.CephClient{}
-	r.cephClientCephFSProvisioner.Name = fmt.Sprintf("%s-%s", "cephclient-cephfs-provisioner", r.storageConsumer.Name)
+	cephClientName = &cephClientFormat{StorageConsumerName: r.storageConsumer.Name, ClaimID: "cephfs", CephUserType: "provisioner"}
+	r.cephClientCephFSProvisioner.Name = generateHashForCephClient(cephClientName)
 	r.cephClientCephFSProvisioner.Namespace = r.namespace
+	r.cephClientCephFSProvisioner.Annotations["cephUserType"] = fmt.Sprintf("%s-%s-%s", cephClientName.StorageConsumerName, cephClientName.ClaimID, cephClientName.CephUserType)
```
nb-ohad, line 157: same as above

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -130,25 +139,43 @@ func (r *StorageConsumerReconciler) initReconciler(request reconcile.Request) {
...
 
 	r.cephClientCephFSNode = &rookCephv1.CephClient{}
-	r.cephClientCephFSNode.Name = fmt.Sprintf("%s-%s", "cephclient-cephfs-node", r.storageConsumer.Name)
+	cephClientName = &cephClientFormat{StorageConsumerName: r.storageConsumer.Name, ClaimID: "cephfs", CephUserType: "node"}
+	r.cephClientCephFSNode.Name = generateHashForCephClient(cephClientName)
 	r.cephClientCephFSNode.Namespace = r.namespace
+	r.cephClientCephFSNode.Annotations["cephUserType"] = fmt.Sprintf("%s-%s-%s", cephClientName.StorageConsumerName, cephClientName.ClaimID, cephClientName.CephUserType)
```
nb-ohad, line 163: same as above

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -130,25 +139,43 @@ func (r *StorageConsumerReconciler) initReconciler(request reconcile.Request) {
...
 
 	r.cephClientHealthChecker = &rookCephv1.CephClient{}
 	r.cephClientHealthChecker.Name = fmt.Sprintf("%s-%s", "cephclient-health-checker", r.storageConsumer.Name)
```
nb-ohad, line 166: Why aren't we using the same hashing here, I don't want to make this user as an exception to the rule
```suggestion
	r.cephClientHealthChecker.Name = generateHashForCephClient(r.storageConsumer.Name, "global", "healthChecker")
```

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -130,25 +139,43 @@ func (r *StorageConsumerReconciler) initReconciler(request reconcile.Request) {
...
 	r.cephClientHealthChecker.Namespace = r.namespace
+	r.cephClientHealthChecker.Annotations["cephUserType"] = fmt.Sprintf("%s-%s", "cephclient-health-checker", r.storageConsumer.Name)
```
nb-ohad, line 168: same as comments above

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -130,25 +139,43 @@ func (r *StorageConsumerReconciler) initReconciler(request reconcile.Request) {
...
+
+}
 
+func generateHashForCephClient(cephClientName *cephClientFormat) string {
+	cephClient, err := json.Marshal(cephClientName)
+	if err != nil {
+		panic("failed to marshal")
+	}
+	name := md5.Sum([]byte(cephClient))
+	return hex.EncodeToString(name[:16])
```
nb-ohad, line 178: free-floating util functions (which are not acting on the reconciler) should be positioned at after all reconciler functions

services/provider/server/server.go
```diff
@@ -325,6 +325,11 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
 				return nil, err
 			}
 
+			cephClientAnnotation, err := s.getCephClientAnnotation(ctx, i.Name)
+			if err != nil {
+				return nil, err
+			}
+
```
nb-ohad, line 332: Why are we reading the resource twice? (once in `getCephClientSecretName` and immediatly after in `getCephClientAnnotation`?

I would suggest to refactor  `getCephClientSecretName` into a `getCephClientInformation` util which return what you need:

```
func (s *OCSProviderServer) getCephClientInformation context.Context, name string) (string, string, error) 
```

where the first return value is a string for the name and the second is a bool for cephUserType

services/provider/server/server.go
```diff
@@ -422,6 +427,21 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
+	err := s.client.Get(ctx, types.NamespacedName{Name: name, Namespace: s.namespace}, cephClient)
+	if err != nil {
+		return "", fmt.Errorf("failed to get rook ceph client %s secret. %v", name, err)
+	}
+	if len(cephClient.Annotations) == 0 {
+		return "", fmt.Errorf("rook ceph client %s annotation is empty", name)
+	}
+	if cephClient.Annotations["cephUserType"] == "" {
+		return "", fmt.Errorf("rook ceph client %s annotation `cephUserType` is empty", name)
+	}
```
nb-ohad, line 441: Why do we need this validation here? I think if the annotation does not exist just return an empty string. 
I don't think we should break reconciliation just because of a missing annotation. 
Maybe in the future we will have other kind of users who does not have the annotation on them

nb-ohad (changes requested):

nb-ohad:
Code reviewed without testing. will review testing after all comments are resolved

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -658,3 +692,13 @@ func (r *StorageConsumerReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&rookCephv1.CephClient{}).
 		Complete(r)
 }
+
+func generateHashForCephClient(storageConsumerName, claimID, cephUserType string) string {
```
Madhu-1, line 696: Make this a method of `cephClientFormat`?

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -658,3 +692,13 @@ func (r *StorageConsumerReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	cephClientName := &cephClientFormat{StorageConsumerName: storageConsumerName, ClaimID: claimID, CephUserType: cephUserType}
+	cephClient, err := json.Marshal(cephClientName)
+	if err != nil {
+		panic("failed to marshal")
```
Madhu-1, line 700: return error instead of panic and also return the actual error

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -374,6 +382,13 @@ func (r *StorageConsumerReconciler) reconcileCephFilesystemSubVolumeGroup() erro
 func (r *StorageConsumerReconciler) reconcileCephClientRBDProvisioner() error {
 
 	desired := &rookCephv1.CephClient{
+		ObjectMeta: metav1.ObjectMeta{
+			Annotations: map[string]string{
+				"ocs.openshift.io.storageconsumer": r.storageConsumer.Name,
+				"ocs.openshift.io.storageclaim":    "rbd",
+				"ocs.openshift.io.cephusertype":    "provisioner",
+			},
```
Madhu-1, line 390: add a helper function to set the annotations?

services/provider/server/server.go
```diff
@@ -422,20 +422,27 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
 	if cephClient.Status.Info == nil {
-		return "", fmt.Errorf("rook ceph client %s Status.Info is empty", name)
+		return "", "", fmt.Errorf("rook ceph client %s Status.Info is empty", name)
 	}
 
-	return cephClient.Status.Info["secretName"], nil
+	if len(cephClient.Annotations) == 0 {
+		return "", "", fmt.Errorf("rook ceph client %s annotation is empty", name)
+	}
+	if cephClient.Annotations["ocs.openshift.io.storageclaim"] == "" {
```
Madhu-1, line 441: define const of annotation keys.

controllers/storageconsumer/storageconsumer_controller.go
subhamkrai: could you explain the use case here or do you have examples of how we use it?

controllers/storageconsumer/storageconsumer_controller.go
subhamkrai: added log but keeping panic...since in other PR it was suggested to do panic when marshalling fails by @nb-ohad (if I'm not wrong)so keeping it, let me know if we want to remove it.

controllers/storageconsumer/storageconsumer_controller.go
Madhu-1: instead of adding or passing more parameters make this a method it will help in future also

```go
func (c *cephClientFormat)generateHashForCephClient() string {
cephClient, err := json.Marshal(c)
if err != nil {
	panic("failed to marshal")
 }
return hasvalue?
}
```

controllers/storageconsumer/storageconsumer_controller.go
nb-ohad: Panic is ok here as we are not trying to marshal a user-given input and as such, we should not fail.

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -658,3 +683,26 @@ func (r *StorageConsumerReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&rookCephv1.CephClient{}).
 		Complete(r)
 }
+
+func (c *cephClientFormat) generateHashForCephClient(storageConsumerName, claimID, cephUserType string) string {
```
nb-ohad, line 687: Why do we need this as a function on cephClientFormat? I think we should hide this structure as much as possible, even just make it an internal structure inside the function itself.

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -658,3 +683,26 @@ func (r *StorageConsumerReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	return hex.EncodeToString(name[:16])
+}
+
+func setAnnotations(storageConsumerName, storageClaim, cephUserType string) map[string]string {
+	annotations := map[string]string{
+		storageConsumerAnnotation:     storageConsumerName,
+		storageClaimAnnotation:        storageClaim,
+		storageCephUserTypeAnnotation: cephUserType,
+	}
+	return annotations
```
nb-ohad, line 707: Annotations are usually considered both an input/output, you cannot just override them completely - which is what this util function is doing.

services/provider/server/server.go
```diff
@@ -320,7 +321,7 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
 	for _, i := range consumerResource.Status.CephResources {
 		switch i.Kind {
 		case "CephClient":
-			clientSecretName, err := s.getCephClientSecretName(ctx, i.Name)
+			clientSecretName, annotation, err := s.getCephClientSecretName(ctx, i.Name)
```
nb-ohad, line 324: Why is this var called annotation? it is taken from an annotation but it is not an annotation. 
Why not just call it cephUserType as suggested before?

nb-ohad (changes requested):

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -658,3 +683,26 @@ func (r *StorageConsumerReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&rookCephv1.CephClient{}).
 		Complete(r)
 }
+
+func (c *cephClientFormat) generateHashForCephClient(storageConsumerName, claimID, cephUserType string) string {
+	c.StorageConsumerName = storageConsumerName
+	c.ClaimID = claimID
+	c.CephUserType = cephUserType
```
nb-ohad, line 690: To complete the prev comment:

```suggestion
	var c struct {
 		StorageConsumerName string
 		ClaimID string
 		CephUserType string
	}

	c.StorageConsumerName = storageConsumerName
	c.ClaimID = claimID
	c.CephUserType = cephUserType
```

services/provider/server/server.go
subhamkrai: now it look something like this `clientSecretName, cephClaim, cephUserType, err := s.getCephClientInformation(ctx, i.Name)`

controllers/storageconsumer/storageconsumer_controller.go
subhamkrai: updated accordingly

controllers/storageconsumer/storageconsumer_controller.go
subhamkrai: done

controllers/storageconsumer/storageconsumer_controller.go
subhamkrai: going with this approach https://github.com/red-hat-storage/ocs-operator/pull/1564#discussion_r815784288

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -374,6 +379,7 @@ func (r *StorageConsumerReconciler) reconcileCephFilesystemSubVolumeGroup() erro
 func (r *StorageConsumerReconciler) reconcileCephClientRBDProvisioner() error {
 
 	desired := &rookCephv1.CephClient{
+		ObjectMeta: metav1.ObjectMeta{},
```
nb-ohad, line 382: Why do we need this line?

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -658,3 +672,29 @@ func (r *StorageConsumerReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	cephClient, err := json.Marshal(c)
+	if err != nil {
+		klog.Errorf("failed to marshal ceph client name for consumer %s. %v", storageConsumerName, err)
+		panic("failed to marshal")
+	}
+	name := md5.Sum([]byte(cephClient))
+	return hex.EncodeToString(name[:16])
+}
+
+func annotationObjectMetaType(annotationObjectMeta *metav1.ObjectMeta, storageConsumerName, storageClaim, cephUserType string) {
```
nb-ohad, line 696: First, what is an `annotationObjectMeta`? every ObjectMeta instance has annotations so the name is just convoluted. Also no need to say objectMeta in the name as this is just an interface for an object. Please change the argument name to `obj`

Second, the name `annotationObjectMetaType` doesn't mean anything here or on the call site. Please give function names based on the thing they are trying to accomplish. For example, in this case I can suggest the name `addStorageRelatedAnnotations`

services/provider/server/server.go
```diff
@@ -422,20 +422,27 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
 	if cephClient.Status.Info == nil {
-		return "", fmt.Errorf("rook ceph client %s Status.Info is empty", name)
+		return "", "", "", fmt.Errorf("rook ceph client %s Status.Info is empty", name)
 	}
 
-	return cephClient.Status.Info["secretName"], nil
+	if len(cephClient.Annotations) == 0 {
+		return "", "", "", fmt.Errorf("rook ceph client %s annotation is empty", name)
+	}
+	if cephClient.Annotations["ocs.openshift.io.storageclaim"] == "" || cephClient.Annotations["ocs.openshift.io.cephusertype"] == "" {
```
nb-ohad, line 441: Use the existing constants instead of hard coded strings 
```suggestion
	if cephClient.Annotations[storageClaimAnnotation] == "" || cephClient.Annotations[storageCephUserTypeAnnotation] == "" {
```

services/provider/server/server.go
```diff
@@ -422,20 +422,27 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
+		klog.Warningf("rook ceph client %s annotation is empty", name)
```
nb-ohad, line 442:
```suggestion
		klog.Warningf("rook ceph client %s has missing storage annotations", name)
```

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -388,6 +393,7 @@ func (r *StorageConsumerReconciler) reconcileCephClientRBDProvisioner() error {
 			return err
 		}
 
+		addStorageRelatedAnnotations(&r.cephClientRBDProvisioner.ObjectMeta, r.storageConsumer.Name, "rbd", "provisioner")
```
nb-ohad, line 396:
```suggestion
		addStorageRelatedAnnotations(&r.cephClientRBDProvisioner, r.storageConsumer.Name, "rbd", "provisioner")
```

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -422,11 +428,13 @@ func (r *StorageConsumerReconciler) reconcileCephClientRBDNode() error {
 			},
 		},
 	}
+
 	_, err := ctrl.CreateOrUpdate(r.ctx, r.Client, r.cephClientRBDNode, func() error {
 		if err := r.own(r.cephClientRBDNode); err != nil {
 			return err
 		}
 
+		addStorageRelatedAnnotations(&r.cephClientRBDNode.ObjectMeta, r.storageConsumer.Name, "rbd", "node")
```
nb-ohad, line 437:
```suggestion
		addStorageRelatedAnnotations(&r.cephClientRBDNode, r.storageConsumer.Name, "rbd", "node")
```

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -457,6 +465,8 @@ func (r *StorageConsumerReconciler) reconcileCephClientCephFSProvisioner() error
 		if err := r.own(r.cephClientCephFSProvisioner); err != nil {
 			return err
 		}
+
+		addStorageRelatedAnnotations(&r.cephClientCephFSProvisioner.ObjectMeta, r.storageConsumer.Name, "cephfs", "provisioner")
```
nb-ohad, line 469:
```suggestion
		addStorageRelatedAnnotations(&r.cephClientCephFSProvisioner, r.storageConsumer.Name, "cephfs", "provisioner")
```

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -494,6 +504,8 @@ func (r *StorageConsumerReconciler) reconcileCephClientCephFSNode() error {
 		if err := r.own(r.cephClientCephFSNode); err != nil {
 			return err
 		}
+
+		addStorageRelatedAnnotations(&r.cephClientCephFSNode.ObjectMeta, r.storageConsumer.Name, "cephfs", "node")
```
nb-ohad, line 508:
```suggestion
		addStorageRelatedAnnotations(&r.cephClientCephFSNode, r.storageConsumer.Name, "cephfs", "node")
```

controllers/storageconsumer/storageconsumer_controller.go
subhamkrai: going with this approach as discussed offline
```
if obj.Annotations == nil {
		obj.Annotations = map[string]string{}
	}
```
in this, we don't need above suggestion

controllers/storageconsumer/storageconsumer_controller.go
subhamkrai: same as above comment

controllers/storageconsumer/storageconsumer_controller.go
subhamkrai: same as above comment

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -541,6 +553,7 @@ func (r *StorageConsumerReconciler) reconcileCephClientHealthChecker() error {
 			return err
 		}
 
+		addStorageRelatedAnnotations(&r.cephClientHealthChecker.ObjectMeta, r.storageConsumer.Name, "global", "healthchecker")
```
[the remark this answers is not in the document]
subhamkrai: same as above comment

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -658,3 +671,30 @@ func (r *StorageConsumerReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	if err != nil {
+		klog.Errorf("failed to marshal ceph client name for consumer %s. %v", storageConsumerName, err)
+		panic("failed to marshal")
+	}
+	name := md5.Sum([]byte(cephClient))
+	return hex.EncodeToString(name[:16])
+}
+
+func addStorageRelatedAnnotations(obj *metav1.ObjectMeta, storageConsumerName, storageClaim, cephUserType string) {
+	obj.Annotations = map[string]string{}
```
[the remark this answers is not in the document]
subhamkrai: add if condition to check that so we are good now

controllers/storageconsumer/storageconsumer_controller.go
nb-ohad: @subhamkrai I don't see the connection. Maybe you misunderstood the suggestion, `ObjectMeta` is embedded inside `cephClientRBDProvisioner` which means that when you take the address of ObjectMeta you are actually taking the address of `cephClientRBDProvisioner` which makes `&cephClientRBDProvisioner` equal to `cephClientRBDProvisioner.ObjectMeta` in this context.

What I am suggesting here is to remove the unnecessary `.ObjectMeta` to make the call more readable and the line shorter

subhamkrai: okay, this how it looks now 
`addStorageRelatedAnnotations(r.cephClientCephFSProvisioner, r.storageConsumer.Name, "cephfs", "provisioner")`

```
func addStorageRelatedAnnotations(cephClient *rookCephv1.CephClient, storageConsumerName, storageClaim, cephUserType string) {
	if cephClient.Annotations == nil {
		cephClient.Annotations = map[string]string{}
	}
	cephClient.Annotations[StorageConsumerAnnotation] = storageConsumerName
	cephClient.Annotations[StorageClaimAnnotation] = storageClaim
	cephClient.Annotations[StorageCephUserTypeAnnotation] = cephUserType
}
```

PR #1568, opened 2022-03-01
Add default toleration to rbd-mirror pod

weirdwiz:
This commit adds default OCS toleration rbd-mirror pod to allow it to be
run on nodes with OCS taints.

Fixes : https://bugzilla.redhat.com/show_bug.cgi?id=2057489

### DIFF
--- a/controllers/storagecluster/cephrbdmirrors.go
+++ b/controllers/storagecluster/cephrbdmirrors.go
@@ -57,6 +57,7 @@ func (r *StorageClusterReconciler) newCephRbdMirrorInstances(initData *ocsv1.Sto
 			Spec: cephv1.RBDMirroringSpec{
 				Count:     1,
 				Resources: defaults.GetDaemonResources("rbd-mirror", initData.Spec.Resources),
+				Placement: getPlacement(initData, "rbd-mirror"),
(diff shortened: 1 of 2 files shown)


controllers/storagecluster/cephrbdmirrors.go
```diff
@@ -57,6 +58,16 @@ func (r *StorageClusterReconciler) newCephRbdMirrorInstances(initData *ocsv1.Sto
 			Spec: cephv1.RBDMirroringSpec{
 				Count:     1,
 				Resources: defaults.GetDaemonResources("rbd-mirror", initData.Spec.Resources),
+				Placement: cephv1.Placement{
```
umangachapagain, line 61: We allow user defined placements for all daemons. If placement for rbd-mirror is not supported in Rook, it should be added there first. Then it should be added to ocs-operator in this file (https://github.com/red-hat-storage/ocs-operator/blob/main/controllers/defaults/placements.go). Then in this line, we'll need to merge user defined placement with our defaults.

umangachapagain (changes requested):

jarrpa (approved):
/lgtm

PR #1572, opened 2022-03-02
Define which type of nodes to label on install

bentito:
Adds the word "worker" to the labeling for install notes in README.
Helps user avoid error of labeling master nodes instead.

Signed-off-by: Brett Tofel <btofel@redhat.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ watching and configuring their CustomResources (CRs).
 
 OCS Operator will install its components only on nodes labelled for OCS with `cluster.ocs.openshift.io/openshift-storage=''`.
 
-To label the nodes from CLI,
+To label the worker nodes from CLI,
 
 ```console
 $ oc label nodes <NodeName> cluster.ocs.openshift.io/openshift-storage=''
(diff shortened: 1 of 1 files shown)


jarrpa:
This is not strictly accurate, as we support using non-worker nodes for our Pods and OSDs.

bentito:
> This is not strictly accurate, as we support using non-worker nodes for our Pods and OSDs.

@jarrpa Could you suggest better wording, because from trying to install on a clusterbot generated ("launch 4.9") cluster I didn't get any joy from labeling all the master nodes (dumb I'm sure, but it's what I did)? Installed simply did not progress at all and it was hard to tell what, if anything, was happening.

jarrpa:
You need to add tolerations for all taints on a given node you want to use. The notion of "master" nodes is a somewhat arbitrary openshift designation implemented by a specific taint.

bentito:
> You need to add tolerations for all taints on a given node you want to use. The notion of "master" nodes is a somewhat arbitrary openshift designation implemented by a specific taint.

Understood, but this PR is just about top of the README.md#installation section, should we get into that? I think most people following this section just want something that works to try out OCS operator.

New wording coming...

malayparida2000:
@bentito Is this PR still relevant?

bentito:
Well it hasn't attracted anyone else thinking it's a useful add to the docs, nor did I have time to do the re-wording required. I think we can close it?

bentito:
Oh actually sorry, missed my own commits. It did get reworded. I think it's up to @jarrpa to see if the re-wording it good and it could merge?

PR #1574, opened 2022-03-08
validate correct cephUserType annotation

sp98:
OCS provider server checks for `healthChecker` annotation on the client
but the actual annotation is `healthchecker`. This PR changes
`healthChecker` to `healthchecker`.


sp98:
We should move `healthchecker` and other similar annotations to a common place and export them.  This should be done in a separate PR.

PR #1580, opened 2022-03-08
Reset external resource cache when monitoring endpoints are not reachable

aruniiird:
Verify the monitoring endpoints only once while cephcluster is created.

### DIFF
--- a/controllers/storagecluster/cephcluster.go
+++ b/controllers/storagecluster/cephcluster.go
@@ -123,6 +123,10 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
 		if err := verifyMonitoringEndpoints(monitoringIP, monitoringPort, r.Log); err != nil {
 			r.Log.Error(err, "Could not connect to the Monitoring Endpoints.",
 				"CephCluster", klog.KRef(sc.Namespace, sc.Name))
+			// reset the external resources cache line for this storagecluster to nil,
+			// to allow the next reconcile to fetch an updated  external resources information
+			// from the provider cluster
+			externalOCSResources[sc.UID] = nil
 			return reconcile.Result{}, err
 		}
 		r.Log.Info("Monitoring Information found. Monitoring will be enabled on the external cluster.", "CephCluster", klog.KRef(sc.Namespace, sc.Name))


controllers/storagecluster/cephcluster.go
```diff
@@ -120,12 +133,6 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
 		}
 		monitoringIP := endpointR.Data["MonitoringEndpoint"]
 		monitoringPort := endpointR.Data["MonitoringPort"]
-		if err := verifyMonitoringEndpoints(monitoringIP, monitoringPort, r.Log); err != nil {
-			r.Log.Error(err, "Could not connect to the Monitoring Endpoints.",
-				"CephCluster", klog.KRef(sc.Namespace, sc.Name))
-			return reconcile.Result{}, err
-		}
-		r.Log.Info("Monitoring Information found. Monitoring will be enabled on the external cluster.", "CephCluster", klog.KRef(sc.Namespace, sc.Name))
 		cephCluster = newExternalCephCluster(sc, r.images.Ceph, monitoringIP, monitoringPort)
```
nb-ohad, line 136: Sending the `monitoringIP` to `newExternalCephCluster` will override the one that rook manages internally with the one from the initial config and this can cause problems. Up until now, the `verifyMonitoringEndpoints` prevented that from happening.

controllers/storagecluster/cephcluster.go
```diff
@@ -170,6 +177,15 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
...
 			if sc.Spec.ExternalStorage.Enable {
+				// External Mode and there is no existing CephCluster,
+				// verify the monitoring endpoints only for the first time
+				err := verifyMonitoringEndpointsOfCephCluster(cephCluster, r.Log)
+				if err != nil {
+					r.Log.Error(err, "Could not connect to the Monitoring Endpoints.",
+						"CephCluster", klog.KRef(sc.Namespace, sc.Name))
+					return reconcile.Result{}, err
+				}
+				r.Log.Info("Monitoring Information found. Monitoring will be enabled on the external cluster.", "CephCluster", klog.KRef(sc.Namespace, sc.Name))
```
nb-ohad, line 188: Can we just keep this code in the original location and just move up L175-L176 before that code section? 
It can also solve the issue In my last comment and will aggregate all checks in a single locations (right now you have 2 sections that take the same underlaying assumption - external mode)

controllers/storagecluster/cephcluster.go
```diff
@@ -70,6 +70,19 @@ func arbiterEnabled(sc *ocsv1.StorageCluster) bool {
...
+func verifyMonitoringEndpointsOfCephCluster(cephCluster *rookCephv1.CephCluster, log logr.Logger) error {
+	if cephCluster == nil {
+		return fmt.Errorf("empty ceph cluster passed as argument")
+	}
+	monIPs := make([]string, len(cephCluster.Spec.Monitoring.ExternalMgrEndpoints))
+	for idx, endpointAddress := range cephCluster.Spec.Monitoring.ExternalMgrEndpoints {
+		monIPs[idx] = endpointAddress.IP
+	}
+	monIPStr := strings.Join(monIPs, ",")
+	monPort := fmt.Sprint(cephCluster.Spec.Monitoring.ExternalMgrPrometheusPort)
```
nb-ohad, line 82: Is there really a need for an additional util function that is called only once from a single call site?
Can't we just embed the code in the call site or alternatively add this code to `verifyMonitringEndpoints` (which I believe is also called only once from a single call site)

nb-ohad (changes requested):

controllers/storagecluster/cephcluster.go
aruniiird: At line#136, `newExternalCephCluster()` function will just initialize the variable `cephCluster`. This variable, `cephCluster` will only be used (to create the external cephcluster) when there is no existing CephCluster found (at line#193). Before create the CephCluster, we verify the endpoints (only in case of an external cluster mode).
So setting `monitoringIP` (and port) to `newExternalCephCluster()` function won't set or override the existing CephCluster spec.

controllers/storagecluster/cephcluster.go
aruniiird: We can definitely move the checking of an existing CephCluster, ie L175 and 176 up. But I was trying to avoid this in an effort to not to duplicate the same code (for found logic) in two places.
As I explained previously, the above scenario (of overriding the configurations of an existing CephCluster) will never happen.
Let me know WDYT.

controllers/storagecluster/cephcluster.go
aruniiird: I really agree. I was trying to add a very minimal change (and mostly trying not to touch the existing code).
We can definitely add this new logic into the existing `verifyMonitringEndpoints()` function.

controllers/storagecluster/cephcluster.go
nb-ohad: Yes, I think this is an issue. 
IIUC Before the change the cephcluster was reconciled every iteration while now we only reconcile it on creation - this is a major change in behavior. 

I think we should focus on doing the check only once, not to fully stop reconciliation after creation (as it can create other unrelated bugs)

controllers/storagecluster/cephcluster.go
nb-ohad: Refactor is a refactor, we do not need to be afraid of changes, we have a review process in place and strong QE. 
I will argue that avoiding touching existing code will always create much more convoluted code over a long period of time which will make every future change even harder to reason about

controllers/storagecluster/cephcluster.go
aruniiird: > Yes, I think this is an issue. IIUC Before the change the cephcluster was reconciled every iteration while now we only reconcile it on creation - this is a major change in behavior.

In the new change, we haven't touched any part of the previous code which affects `CephCluster`'s reconciliation. We just moved `verifyMonitoringEndpoints()` function down. No other change. Every other piece of code is as is.


> 
> I think we should focus on doing the check only once, not to fully stop reconciliation after creation (as it can create other unrelated bugs)

We do the check only once and we don't stop/alter the existing reconciling logic.

controllers/storagecluster/cephcluster.go
aruniiird: Ack and I agree.
There were times when comments were like this =)
```We can do all the refactoring later. Add just the minimal code to solve the issue or else it will be very hard to say whether we have introduced a new bug (or not)```

File: controllers/storagecluster/cephcluster.go (hunk 1 above)
Comment (aruniiird) [reply controllers/storagecluster/cephcluster.go:157]: Thanks @nb-ohad  for the clarification. I see that we are using the `cephCluster` object later (line#282 onwards) and use check the `Spec`. With this change, yes that will become a problem/bug.

File: controllers/storagecluster/cephcluster.go [hunk 4]
```diff
@@ -123,6 +123,9 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
 		if err := verifyMonitoringEndpoints(monitoringIP, monitoringPort, r.Log); err != nil {
 			r.Log.Error(err, "Could not connect to the Monitoring Endpoints.",
 				"CephCluster", klog.KRef(sc.Namespace, sc.Name))
+			// reset the external resources cache to nil,
+			// when monitoring endpoints are not reachable
```
nb-ohad, line 127:
```suggestion
			// reset the external resources cache line for this storagecluster to nil, 
			// to allow the next reconcile to fetch an updated  external resources information 
			// from the provider cluster
```

controllers/storagecluster/cephcluster.go
```diff
@@ -123,6 +123,9 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
...
+			externalOCSResources[sc.UID] = nil
```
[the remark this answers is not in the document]
nb-ohad: The short answer: We do not. 
The long answer:  
1. We return an error, which in turn will queue another event,
2. A new reconcile for the new event is triggered
3. The code in `ensure_created` on `external _resources.go` sees that we have no cache line for this storageclsuter and will fetch again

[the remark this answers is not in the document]
aruniiird: My question revolves around the `3` rd step.
When we reset a map variable's (particular) key value to nil, we are not resetting any cache (because the map var `externalOCSResources` will always start as an empty map for each reconcile). [this is my understanding]

In order for the `3` rd step (above mentioned) to be true, the assumption is it has to enter `ensure_created` (on `external_resources.go` file) again in the **same** reconcile call (after we reset map of that storagecluster-id to `nil`)

controllers/storagecluster/cephcluster.go
aruniiird: done

controllers/storagecluster/cephcluster.go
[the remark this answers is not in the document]
aruniiird: Again thanks @nb-ohad . Now I got the context that the map var which we edit is a global/package level one, which won't be reset on each reconcile. Then this solves the confusion.

PR #1583, opened 2022-03-11
ocs-to-ocs: correct the name of healthchecker cephclient

subhamkrai:
due to recent change where we introduced shorter names
for cephClient it required to change healthchecker cephclient
also due to this, we were getting failure during onboard.

### DIFF
--- a/services/provider/server/server.go
+++ b/services/provider/server/server.go
@@ -318,10 +318,10 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
 			"MonitoringPort":     strconv.Itoa(int(port)),
 		})})
 
-	for _, i := range consumerResource.Status.CephResources {
-		switch i.Kind {
+	for _, cephRes := range consumerResource.Status.CephResources {
+		switch cephRes.Kind {
 		case "CephClient":
-			clientSecretName, cephClaim, cephUserType, err := s.getCephClientInformation(ctx, i.Name)
+			clientSecretName, cephClaim, cephUserType, err := s.getCephClientInformation(ctx, cephRes.Name)
 			if err != nil {
 				return nil, err
 			}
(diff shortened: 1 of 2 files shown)


services/provider/server/server_test.go
```diff
@@ -43,7 +43,7 @@ var mockExtR = map[string]*externalResource{
 		Name: "rook-ceph-mon",
 		Kind: "Secret",
 		Data: map[string]string{
-			"ceph-username": "client.cephclient-health-checker-consumer",
+			"ceph-username": "client.7c1507bffc3030e301cdd0f9b6ecd56e",
```
iamniting, line 46: are you sure we need this specific hash or string? This might be a case only for one consumer.

iamniting (changes requested):

services/provider/server/server_test.go
nb-ohad: @iamniting this is test code, it is ok to assume a specific result

subhamkrai: yes it is unique for one consumer as it is unit test so I used specific hash

services/provider/server/server.go
```diff
@@ -356,7 +356,7 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
 					Data: mustMarshal(map[string]string{
 						"fsid":          fsid,
 						"mon-secret":    "mon-secret",
-						"ceph-username": fmt.Sprintf("client.cephclient-health-checker-%s", consumerResource.Name),
+						"ceph-username": fmt.Sprintf("client.%s", i.Name),
```
iamniting, line 359: This i is confusing, I would also suggest that you change the name of the i to something else as i is mainly used for the index which is not the case here.

iamniting (changes requested):

services/provider/server/server.go
subhamkrai: changed to `cephRes`

sp98 (approved):

nb-ohad:
From an offline discussion, it seems like the fix is not complete, as there are still issues with PVC creation. holding for now

PR #1595, opened 2022-03-22
ocs: add RPC calls for storageclassClaim

Madhu-1:
added 3 RPC calls to FulfillStorageClassClaim, RevokeStorageClassClaim, and GetStorageClassClaimConfig

* Fulfill storageClassClaim

On the consumer side, we send RPC request to fulfill the storageclassClaim, once the request is received on the provider side, the storageClassClaim CR will be created and all corresponding ceph resources will also be created. the Consumer will keep on polling with GetStorageClassClaimConfig RPC call to get the resources created for the new storageClassClaim

* Revoke storageClassClaim

On the consumer side, we send RPC request to revoke the storageclassClaim, once the request is received on the provider side, the storageClassClaim CR will be deleted and all corresponding ceph resources will also be deleted. A success or failure response will be sent back.


services/provider/proto/provider.proto
```diff
@@ -109,3 +122,50 @@ message AcknowledgeOnboardingRequest{
...
+    string storageConsumerUUID = 1;
+    // storageclassClaim (UUID) of the consumer cluster.
+    string storageClassClaimUUID = 2;
+    // StorageType of the storageclassClaim
+    enum StorageType {
+      CephFS = 0;
+      RBD = 1;
+   }
+   StorageType storageType = 3;
+}
```
umangachapagain, line 139: we need a field for encryption as well.

services/provider/proto/provider.proto
```diff
@@ -109,3 +122,50 @@ message AcknowledgeOnboardingRequest{
...
+// create the StorageclassClaim CR on provider cluster.
+message ProvisionStorageClassClaimRequest{
+    // K8s UID (UUID) of the consumer cluster
+    string storageConsumerUUID = 1;
+    // storageclassClaim (UUID) of the consumer cluster.
+    string storageClassClaimUUID = 2;
+    // StorageType of the storageclassClaim
+    enum StorageType {
+      CephFS = 0;
+      RBD = 1;
```
umangachapagain, line 136: I understand this is internal, but can we keep it same as api spec?

https://github.com/red-hat-storage/ocs-operator/blob/dcc00921abb43f109204c32b8d7d975b0146c610/api/v1alpha1/storageclassclaim_types.go#L27

services/provider/proto/provider.proto
```diff
@@ -109,3 +122,50 @@ message AcknowledgeOnboardingRequest{
...
+    string storageConsumerUUID = 1;
+    // storageclassClaim (UUID) of the consumer cluster.
+    string storageClassClaimUUID = 2;
+}
+
+// DeProvisionStorageClassClaimResponse holds the response for the DeProvisionStorageClassClaim API request.
+message DeProvisionStorageClassClaimResponse{
+}
+
+// StorageClassClaimConfigRequest holds the information required generate the
```
umangachapagain, line 158:
```suggestion
// StorageClassClaimConfigRequest holds the information required to generate the
```

services/provider/proto/provider.proto
Madhu-1: encryption is part of the storageclass we can handle it on the consumer side as well IMO no need to make it part of the RPC call. WDYT?

umangachapagain: So once we get back StorageClass object in response, we inject encryption spec and create it?
Sounds good.

umangachapagain (approved):

services/provider/client/client.go
```diff
@@ -133,3 +133,54 @@ func (cc *OCSProviderClient) AcknowledgeOnboarding(ctx context.Context, consumer
...
 }
+
+func (cc *OCSProviderClient) ProvisionStorageClassClaim(ctx context.Context, consumerUUID, claimUUID string, storageType pb.ProvisionStorageClassClaimRequest_StorageType) (*pb.ProvisionStorageClassClaimResponse, error) {
+	if cc.Client == nil || cc.clientConn == nil {
+		return nil, fmt.Errorf("provider client is closed")
+	}
+
+	req := &pb.ProvisionStorageClassClaimRequest{
+		StorageConsumerUUID:   consumerUUID,
+		StorageClassClaimUUID: claimUUID,
```
nb-ohad, line 144: If we are provisioning, why do we need to send a local uid to the provider?

services/provider/client/client.go
```diff
@@ -133,3 +133,54 @@ func (cc *OCSProviderClient) AcknowledgeOnboarding(ctx context.Context, consumer
 
 	return cc.Client.AcknowledgeOnboarding(apiCtx, req)
 }
+
+func (cc *OCSProviderClient) ProvisionStorageClassClaim(ctx context.Context, consumerUUID, claimUUID string, storageType pb.ProvisionStorageClassClaimRequest_StorageType) (*pb.ProvisionStorageClassClaimResponse, error) {
```
nb-ohad, line 137: The fact we are creating a storage class claim on the provider is an implementation detail from the perspective of the consumer cluster, so provisioning might be the wrong verb to use here. I will suggest:
`fulfilStorageClassClaim` instead. This way we are not hinting at a specific provider implementation

services/provider/client/client.go
```diff
@@ -133,3 +133,54 @@ func (cc *OCSProviderClient) AcknowledgeOnboarding(ctx context.Context, consumer
...
+	req := &pb.ProvisionStorageClassClaimRequest{
+		StorageConsumerUUID:   consumerUUID,
+		StorageClassClaimUUID: claimUUID,
+		StorageType:           storageType,
+	}
+
+	apiCtx, cancel := context.WithTimeout(ctx, cc.timeout)
+	defer cancel()
+
+	return cc.Client.ProvisionStorageClassClaim(apiCtx, req)
```
nb-ohad, line 151: Are we getting the remote claim UID in the response?

services/provider/client/client.go
```diff
@@ -133,3 +133,54 @@ func (cc *OCSProviderClient) AcknowledgeOnboarding(ctx context.Context, consumer
...
+}
+
+func (cc *OCSProviderClient) DeProvisionStorageClassClaim(ctx context.Context, consumerUUID, claimUUID string) (*pb.DeProvisionStorageClassClaimResponse, error) {
```
nb-ohad, line 154: For the same reason I gave on the comment for the `ProvisionStorageClassClaim` name, I will suggest here: `revokeStorageClassClaim`

services/provider/proto/provider.proto
nb-ohad: I am not sure this is a good approach, I think we want the ability to understand the exact claim configuration on the provider as well, even if we are not going to take action on it. The main reason here is trackability and supportability

services/provider/proto/provider.proto
```diff
@@ -109,3 +122,50 @@ message AcknowledgeOnboardingRequest{
...
+
+// ProvisionStorageClassClaimResponse holds the response for the ProvisionStorageClassClaim API request.
+message ProvisionStorageClassClaimResponse{
+}
+
+// DeProvisionStorageClassClaimRequest holds the information required to delete the
+// StorageclassClaim CR on provider cluster.
+message DeProvisionStorageClassClaimRequest{
+    // K8s UID (UUID) of the consumer cluster
+    string storageConsumerUUID = 1;
```
nb-ohad, line 149: Why do we need to identify the consumer?

services/provider/proto/provider.proto
```diff
@@ -109,3 +122,50 @@ message AcknowledgeOnboardingRequest{
...
+
+// DeProvisionStorageClassClaimResponse holds the response for the DeProvisionStorageClassClaim API request.
+message DeProvisionStorageClassClaimResponse{
+}
+
+// StorageClassClaimConfigRequest holds the information required to generate the
+// json config for StorageClassClaim specific resources.
+message StorageClassClaimConfigRequest{
+    // K8s UID (UUID) of the consumer cluster
+    string storageConsumerUUID =1;
```
nb-ohad, line 162: Why do we need to identify the consumer?

nb-ohad (changes requested):

services/provider/proto/provider.proto
Madhu-1: Yes but this is extra data on the RPC call in both request and the response. if that is fine am fine with adding a field for the encryption as well.

services/provider/client/client.go
Madhu-1: Will change it to the name as we need to create the storageclassClaim CR on the provider side.

services/provider/client/client.go
Madhu-1: for me it's not a problem, this was the names taken from the design discussions

nb-ohad: This is a public API so names are important.

services/provider/client/client.go
Madhu-1: Yes will add it to the Response.

services/provider/client/client.go
```diff
@@ -133,3 +133,50 @@ func (cc *OCSProviderClient) AcknowledgeOnboarding(ctx context.Context, consumer
 
 	return cc.Client.AcknowledgeOnboarding(apiCtx, req)
 }
+
+func (cc *OCSProviderClient) FulFilStorageClassClaim(ctx context.Context, consumerUUID, encryptionMethod string, storageType pb.FulFilStorageClassClaimRequest_StorageType) (*pb.FulFilStorageClassClaimResponse, error) {
```
nb-ohad: Fulfill is one word
```suggestion
func (cc *OCSProviderClient) FulfillStorageClassClaim(ctx context.Context, consumerUUID, encryptionMethod string, storageType pb.FulFilStorageClassClaimRequest_StorageType) (*pb.FulFilStorageClassClaimResponse, error) {
```

services/provider/proto/provider.proto
```diff
@@ -50,8 +63,8 @@ message OnboardConsumerResponse{
 // StorageConfigRequest holds the information required generate the json config for connecting to storage provider cluster
 message StorageConfigRequest{
     // K8s UID (UUID) of the consumer cluster
-     string storageConsumerUUID =1;
- }
+    string storageConsumerUUID =1;
```
nb-ohad, line 66:
```suggestion
    string storageConsumerUUID = 1;
```

nb-ohad (changes requested):

umangachapagain (approved):

agarwal-mudit:
ci infra issue, merging it manually

PR #1597, opened 2022-03-22
set upgrade conditions to true in external mode deployment

umangachapagain:
We were setting `Upgradable: false` in some cases but we
never set it to `Upgradable: true` for external mode
deployments. This caused further upgrades to be blocked.

This commit changes that behavior to set `Upgradable: true`
even for external mode deployments when none of the components
report an error.


umangachapagain:
This might fix upgrade issue. But how do we apply this fix to a cluster that can not be upgraded?
/cc @agarwal-mudit @jarrpa

agarwal-mudit (approved):
/lgtm

PR #1602, opened 2022-03-24
Add the ceph toolbox enable/disable & reconcile to storagecluster 

malayparida2000:
The NewToolsDeployment function has been removed from ocsinitialization
controller and has been added to storagecluster API.

the ensureToolbox() function is added to the StorageClusterReconciler
using the new API method, checking the OCSInitialization & the
StorageClsuter for the EnableCephTools boolean. If either of them is
true the toolbox is created & if both are false toolbox is disabled.

As the toolbox deployment is now handled by the storagecluster
controller so the same logic is being removed from ocsinitialization.


controllers/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -174,13 +175,18 @@ func (r *OCSInitializationReconciler) ensureToolsDeployment(initialData *ocsv1.O
...
+			return err
+		}
 		if !isFound {
 			return r.Client.Create(context.TODO(), toolsDeployment)
 		} else if !reflect.DeepEqual(foundToolsDeployment.Spec, toolsDeployment.Spec) {
 
 			updateDeployment := foundToolsDeployment.DeepCopy()
 			updateDeployment.Spec = *toolsDeployment.Spec.DeepCopy()
-
+
```
iamniting, line 189: this line change should be reverted

iamniting (changes requested):

controllers/storagecluster/uninstall_reconciler.go
```diff
@@ -358,3 +364,21 @@ func (r *StorageClusterReconciler) deleteResources(sc *ocsv1.StorageCluster) (re
...
+	instance := &ocsv1.OCSInitialization{}
+	ocsNamespacedName := ocsinitialization.InitNamespacedName()
+	err := r.Client.Get(context.TODO(), types.NamespacedName{Name: ocsNamespacedName.Name, Namespace: ocsNamespacedName.Namespace}, instance)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return nil
+		}
+		return err
+	}
+	instance.Spec.EnableCephTools = false
```
iamniting, line 378: if it is already false skip the update call and return immediately

iamniting (changes requested):

umangachapagain (changes requested):
When the StorageCluster is deleted and toolbox is not, does it break anything? If not, there is no need for this change.

Toolbox deployment is not a part of StorageCluster creation workflow, and as such it should not be part of StorageCluster deletion workflow.

OCSInitialization is for controlling operator level configurations and wasn't designed to be managed by StorageCluster.

If toolbox is explicitly installed by user via the flag provided on initialization CR, they are responsible for deleting by switching that flag off.

/hold

agarwal-mudit:
> When the StorageCluster is deleted and toolbox is not, does it break anything? If not, there is no need for this change.
> 
> Toolbox deployment is not a part of StorageCluster creation workflow, and as such it should not be part of StorageCluster deletion workflow.
> 
> OCSInitialization is for controlling operator level configurations and wasn't designed to be managed by StorageCluster.
> 
> If toolbox is explicitly installed by user via the flag provided on initialization CR, they are responsible for deleting by switching that flag off.
> 
> /hold

We had the same discussion in the BZ (https://bugzilla.redhat.com/show_bug.cgi?id=1937117) and later decided to fix it because:
1. It is a minor change.
2. It mainly helps QE during there automated ci.

iamniting (approved):

umangachapagain:
> We had the same discussion in the BZ (https://bugzilla.redhat.com/show_bug.cgi?id=1937117) and later decided to fix it because:
>
Discussion in the BZ is limited to "it should be fixed, so let's fix it". Here I am more concerned about how we choose to fix it.
 
>     1. It is a minor change.
> 
It is a minor change that changes how the system was designed to work. So we need to be sure that this is what we want to do.

As an alternate approach, (if toolbox is deployed per StorageCluster) I suggest we move the whole thing from OCSInitialization to StorageCluster. Then toolbox deployment can be owned by StorageCluster via OwnerRef without the need for extra deletion logic.
This requires deprecating older way of enabling/disabling toolbox.

@jarrpa WDYT?

jarrpa:
I'm of the opinion that the toolbox Pod should persist as long as a CephCluster persists. But also, we intentionally decided to only have one toolbox Pod instance. That said, looking at the implementation it seems that a given toolbox Pod has to access Secrets that are specific to any given CephCluster anyway, thus negating the intent of the design. 

@travisn can you weigh in on what the Rook-Ceph stance is on the toolbox Pod when there are multiple CephClusters? As each CephCluster is expected to be in its own Namespace, the argument for a toolbox per cluster makes sense. Still, thoughts?

travisn:
The toolbox pod runs in the same namespace as the ceph cluster. If you have multiple clusters, you would have multiple toolboxes. So it does seem appropriate to reconcile at the same time as the cephcluster.

jarrpa:
Awesome. Unfortunately for us, this has become a slightly bigger endeavor. :joy: 

I can break this down slightly:
1. Introduce a new boolean to the StorageCluster for deploying the toolbox Pod (`StorageCluster.Spec.EnableCephTools` just to keep the naming consistent)
2. I think we can get away with not using the `ensureCreate()`/`ensureDeleted()` pattern for this, just make sure the toolbox reconciliation happens early in the reconciliation loop so it's either present or gone before anything else
3. Mark the toolbox Pod as owned/controlled by the relevant StorageCluster (already done)
4. Figure out the upgrade/deprecation path from OCSInitialization to StorageCluster control. It might make sense to temporarily make the OCSInitialization boolean control the same boolean for *all* StorageClusters. This way the OCSInitialization CR no longer controls the reconciliation of the Pod(s) directly, it just tells all StorageClusters to do it instead.

For now, please go ahead and implement parts 1 to 3 in this PR, as they are the minimum requirements to fix the BZ itself. If we get consensus on 4 in the meantime we can push that here as well, otherwise that can be left to a follow-up PR.

malayparida2000:
@jarrpa , Thanks for the inputs. Would start working on the way you suggested.

nb-ohad:
@malayparida2000 @jarrpa This will break the managed service deployment as we use this field on the OCSInitialization CR and this breaks an existing API without a CR version change

api/v1/ocsinitialization_types.go
```diff
@@ -25,16 +25,6 @@ import (
...
-	// EnableCephTools toggles on whether or not the ceph tools pod
-	// should be deployed.
-	// Defaults to false
-	// +optional
-	EnableCephTools bool `json:"enableCephTools,omitempty"`
-
-	// Tolerations if specified set toolbox ceph tools pod tolerations
-	// Defaults to empty
-	// +optional
-	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
```
nb-ohad, line 37: These are breaking changes to an existing CR interface and we should not do them this way!

If the desire is for breaking changes to an existing CR in release `n` then you will have to do the following:

In release `n`:
1.  introduce a new CR version (v2) with the breaking changes 
2. support backward compatibility for this release
3. Mark the old CR version as deprecated  

In release `n + 1`
1. remove support for the V1 CR 

All of this needs to be done because a CR interface is a public interface that our internal and external customers' automation might rely on. One example is the ODF Managed service deployer.

nb-ohad (changes requested):

api/v1/ocsinitialization_types.go
nb-ohad: @agarwal-mudit FYI

malayparida2000: For now, I have reverted these changes to ocsinitialization, The EnableCephTools & Tolerations will stay here in ocsinitialization for now. 
So basically for now we have now two duplicate ways of enabling
    cephToolBox,
    1. Via Ocsinitialization
    2. Via Storagecluster
    
    If done by patching Ocsinitialization then the toolbox pod will
    still, stay there even after the storage cluster is gone.
    
    If done via patching Storagecluster then whenever the storagecluster
    goes away the toolbox pod goes away automatically as the storagecluster
    is added as an owner to that.

We will figure out an upgrade/deprecation path from OCSInitialization to StorageCluster control later.

controllers/storagecluster/reconcile.go
```diff
@@ -676,3 +690,146 @@ func validateOverprovisionControlSpec(sc *ocsv1.StorageCluster, reqLogger logr.L
...
+
+			updateDeployment := foundToolsDeployment.DeepCopy()
+			updateDeployment.Spec = *toolsDeployment.Spec.DeepCopy()
+
+			return r.Client.Update(context.TODO(), updateDeployment)
+		}
+	} else if isFound {
+		// delete if ceph tools exists and is disabled
+		return r.Client.Delete(context.TODO(), foundToolsDeployment)
+	}
```
nb-ohad, line 833: @malayparida2000 You still have an issue. 
As was discussed on this PR, we are going to keep support for the field on `OCSInitialization` as well.
Assuming that I am enabling with `OCSInitialization` the pod will get deployed, but then when the first reconcile of the storagecluster will happen the flag on the storagecluster will be false and the pod will get removed!

In your condition for deletion you need to also check that the value not set or set to false on the `OCSInitialization` CR

nb-ohad (changes requested):

controllers/storagecluster/reconcile.go
malayparida2000: @nb-ohad  For this I have to get the OCSInitialization CR inside the storagecluster reconciler & check for its value is set as false, But the vice versa also applies i.e I have to get the storage cluster CR inside ocsinitilization reconciler & before disabling the toolbox I have to check the spec for storagecluster CR. 
Or else it will create the same issue you mentioned above, i.e assuming that I am enabling with Storagecluster the pod will get deployed, but then when the first reconciliation of the ocsinitialization will happen the flag on the ocsinitialization will be false and the pod will get removed!
 So I was thinking Instead of this can I change the name of the toolbox that the storagecluster creates slightly? So that the cephtoolbox deployments both of them create are different from each other & both won't interfere with each other. Like the rookCephToolDeploymentName is set as "rook-ceph-tool" in the ocsinitialization, so can we set the name for the same in storagecluster something different, Like "rook-ceph-toolbox"?

iamniting: I don't agree, We should not have 2 toolbox pods at all.

I do understand that we can not remove changes from the ocsinitilization API, But we can remove the toolbox deployment code from the ocsinitilization controller and have this functionality in the storagecluster controller only.

In upgrades, we can check if it already exists and the flag is disabled in the storagecluster we will still update it. If the flag is enabled anyway we will create or update it.

jarrpa (changes requested):

jarrpa:
Comments left. I'd like to see the new commits structured as follows:

1. Move toolbox Deployment function to StorageCluster API
2. Add `ensureToolbox()` function to the StorageClusterReconciler using the new API method, checking the OCSInitialization CR for the `EnableCephTools` boolean
3. Update OCSInitialization logic to not create its own toolbox Deployment

controllers/storagecluster/reconcile.go
nb-ohad: @malayparida2000  @jarrpa  @iamniting

This is more than just supporting the API, removing the functionality from the `ocsinitilization` controller will break existing autoamtions/documentation (not ours) - this includes the Managed Service! 
We need to have a timespan where both flags take effect to allow migration. 

I will start with saying that I am not agreeing with the main statement of this comment https://github.com/red-hat-storage/ocs-operator/pull/1602#discussion_r861282028 and I will explain:

Assuming that we are ok with watching and reading the `ocsinitilization` CR in the `stoagecluster` controller for the next release then we can do the following:
1. the `storagecluster` controller will be the one to deploy the toolbox!
2. the condition for deployment will be an OR operation between the enablement flags on both CRs.
3. the `storagecluster` will watch the `ocsinitilization` CR (which you will need for upgrades anyway) which will trigger a reconcile if the flag changes on `ocsinitilization` CR

The main thing here is that we do not need to sync both CRs, we enable if either one is on and we disable only if both are off!

Running with this code in place will give us the opportunity to deprecate the feature from the `ocsinitilization` controller during a full release, and on the release that comes after it, we can just ignore the flag on `ocsinitilization` CR completely assuming that we gave enough time and warning to our customers to update their automation/documentation

controllers/storagecluster/reconcile.go
```diff
@@ -676,3 +699,61 @@ func validateOverprovisionControlSpec(sc *ocsv1.StorageCluster, reqLogger logr.L
...
+		if !isFound {
+			return r.Client.Create(context.TODO(), toolsDeployment)
+		} else if !reflect.DeepEqual(foundToolsDeployment.Spec, toolsDeployment.Spec) {
+
+			updateDeployment := foundToolsDeployment.DeepCopy()
+			updateDeployment.Spec = *toolsDeployment.Spec.DeepCopy()
+
+			return r.Client.Update(context.TODO(), updateDeployment)
+		}
+	} else if isFound && !sc.Spec.EnableCephTools && !ocsinit.Spec.EnableCephTools {
```
nb-ohad, line 754: when evaluating this condition you already know that both  `sc.Spec.EnableCephTools` and `ocsinit.Spec.EnableCephTools` are false, because if any of them was true you will not get to evaluate the else condition.

```suggestion
	} else if isFound  {
```

nb-ohad (changes requested):

controllers/storagecluster/reconcile.go
malayparida2000: Done

controllers/storagecluster/reconcile.go
jarrpa: I believe the results of this discussion have been met. @nb-ohad PTAL

jarrpa (changes requested):
Overall, this looks great. I have some additional finishing touches required:

* I don't think you need or want to filter the enqueue events for OCSInitialization in the StorageClusterController. Just reconcile all StorageClusters on all OCSInitialization events.
* Instead of fully deleting the toolbox unit tests, move them to `controllers/storagecluster`. I would recommend organizing all the new code into a `cephtoolbox.go` file and a companion `cephtoolbox_test.go` file.
* Obviously, fix the currently broken unit tests. :wink:

controllers/storagecluster/reconcile.go
nb-ohad: Seems Ok

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -113,12 +121,31 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		},
 	}
 
+	enqueueStorageClusterRequest := handler.EnqueueRequestsFromMapFunc(
```
nb-ohad, line 124: I think it will be better if the name of the enqueue command will describe the fact that this request is enqueued from a label.
```suggestion
	enqueueStorageClusterRequestFromLabel := handler.EnqueueRequestsFromMapFunc(
```

malayparida2000:
@jarrpa Have made all the suggested changes.
The cephtoolbox thing has been moved to a new file along with its tests. 
Toolbox is enabled when any one of the ocsinitialization or storagecluster spec is true.
Toolbox is disabled when both of them are false.
All the broken unit tests have been fixed.

malayparida2000:
@nb-ohad Can you please remove the do-not-merge/hold label?

PR #1605, opened 2022-03-24
implement FulfillStorageClassClaim and RevokeStorageClassClaim RPC call

Madhu-1:
* Add storageClassClaimName and consumerUUID for all claim RPC calls.
 Adding claimName and consumerUUID to all claim RPC calls for the identification of the claim on the provider cluster.
 Sending claimName is important to avoid corner cases like below
   * If ocs-operator is restarted on the consumer side after sending the FulfillStorageClassClaim RPC call but before receiving the response
   * If ocs-operator is restarted on the provider side after creating the storageClassClaim CR but before sending the FulfillStorageClassClaim RPC response.
   * If there is a network disturbance in FulfillStorageClassClaim RPC operation.
    
Keeping all these reasons and also to avoid breaking changes in future versions, sending storageClassClaimName and consumerUUID to all claim RPC calls.

* Implement FulfillStorageClassClaim RPC call
  Added the logic for the FulfillStorageClassClaim RPC call. it follows the below flow.
  * Validate the consumerUUID
  * Generate the unique Name for Claim
  * Creates the StorageClassClaim CR on the provider
    cluster.

* Implement RevokeStorageClassClaim RPC
    Added implementation for the RevokeStorageClassClaim RPC to delete the StorageClassClaim CR.


services/provider/server/storageclaim.go
```diff
@@ -0,0 +1,99 @@
...
+	return fmt.Sprintf("storageclassclaim-%s", generateHashForStorageClassClaim(consumerUUID, storageClassClaimName))
+}
+
+// Create creates a new storageClassClaim resource and returns the storageClassClaim ID.
+func (c *ocsStorageClassClaimManager) Create(ctx context.Context, consumerUUID, storageClassClaimName, claimType, encryptionMethod string) (string, error) {
+	generatedClaimName := getStorageClassClaimName(consumerUUID, storageClassClaimName)
+	storageClassClaimObj := &ocsv1alpha1.StorageClassClaim{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: generatedClaimName,
+		},
```
Madhu-1, line 57: @umangachapagain @nb-ohad we need to make consumerUUID as part of the StorageClassClaim API which is missing.

services/provider/server/storageclaim.go
```diff
@@ -0,0 +1,100 @@
...
+		StorageConsumerUUID   string `json:"storageConsumerUUID"`
+		StorageClassClaimName string `json:"storageClassClaimName"`
+	}
+
+	s.StorageConsumerUUID = storageConsumerUUID
+	s.StorageClassClaimName = storageClassClaimName
+
+	claimName, err := json.Marshal(s)
+	if err != nil {
+		klog.Errorf("failed to generate storageClassClaim hash name for %q for consumer %q. %v", storageClassClaimName, storageConsumerUUID, err)
```
nb-ohad, line 41:
```suggestion
		klog.Errorf("failed to marshal a name for a storage class claim based on %v. %v", s, err)
```

services/provider/server/storageclaim.go
```diff
@@ -0,0 +1,100 @@
...
+		panic("failed to marshal")
+	}
+	name := md5.Sum([]byte(claimName))
+	return hex.EncodeToString(name[:16])
```
nb-ohad, line 45: See next comment for full reasoning

```suggestion
	return fmt.Sprintf("storageclassclaim-%s", hex.EncodeToString(name[:16])
```

services/provider/server/storageclaim.go
nb-ohad: Why do we need it as a change to the API? the consumer object should be specified on the claim as an owner ref. 
And if we really want the UID we can just annotate the resource. 

Let's remember that StorageClassClaim should also be usable in converged mode where we do not have storage consumers.

services/provider/server/storageclaim.go
```diff
@@ -0,0 +1,100 @@
...
+			EncryptionMethod: encryptionMethod,
+		},
+	}
+
+	err := c.client.Create(ctx, storageClassClaimObj)
+	if err != nil {
+		if kerrors.IsAlreadyExists(err) {
+			klog.Warningf("storageClassClaim %q for %q already exists", generatedClaimName, storageClassClaimName)
+			return string(storageClassClaimObj.GetUID()), err
+		}
```
nb-ohad, line 70: First, do we really need a warning for this case?
Second, we do need to consider a request that has a different configuration than the original one. It should fail as we are not allowing updates to storage class claims with this enhancement (maybe in the future)

services/provider/server/storageclaim.go
```diff
@@ -0,0 +1,100 @@
...
+		return "", fmt.Errorf("failed to create storageClassClaim resource %q for %q. %w", storageClassClaimObj.Name, storageClassClaimName, err)
```
nb-ohad, line 71:
```suggestion
		return "", fmt.Errorf("failed to create a StorageClassClaim named %q for consumer %q and claim %q. %w", generatedClaimName, consumerUUID, storageClassClaimName,  err)
```

services/provider/server/storageclaim.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	klog.Infof("successfully created storageClassClaim resource %q for %q", generatedClaimName, storageClassClaimName)
```
nb-ohad, line 74:
```suggestion
	klog.Infof("successfully created a StorageClassClaim resource %q for consumer %q and claim %q", generatedClaimName, consumerUUID , storageClassClaimName)
```

services/provider/server/storageclaim.go
```diff
@@ -0,0 +1,100 @@
...
+	generatedClaimName := getStorageClassClaimName(consumerUUID, storageClassClaimName)
+	storageClassClaimObj := &ocsv1alpha1.StorageClassClaim{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      generatedClaimName,
+			Namespace: c.namespace,
+		},
+	}
+	if err := c.client.Delete(ctx, storageClassClaimObj); err != nil {
+		if kerrors.IsNotFound(err) {
+			klog.Warningf("storageClassClaim %q not found for %q", generatedClaimName, storageClassClaimName)
```
nb-ohad, line 91: Do we really need this warning? deletion of an object that is already deleted means success with no implication so I do not see the need for the log.

If we are keeping it, then
```suggestion
			klog.Warningf("StorageClassClaim %q not found for consumer %q and claim %q", generatedClaimName, consumerUUID, storageClassClaimName)
```

services/provider/server/storageclaim.go
```diff
@@ -0,0 +1,100 @@
...
+			return nil
+		}
+		return fmt.Errorf("failed to delete storageClassClaim %q. %v", storageClassClaimName, err)
```
nb-ohad, line 94:
```suggestion
		return fmt.Errorf("failed to delete StorageClassClaim %q for consumer %q and claim %q. %v", generatedClaimName, consumerUUID, storageClassClaimName, err)
```

services/provider/server/storageclaim.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	klog.Infof("successfully deleted storageClassClaim resource %q for %q", generatedClaimName, storageClassClaimName)
```
nb-ohad, line 97:
```suggestion
	klog.Infof("successfully deleted StorageClassClaim %q for consumer %q and claim %q", generatedClaimName, consumerUUID, storageClassClaimName)
```

nb-ohad (changes requested):

services/provider/server/storageclaim.go
Madhu-1: IMO the user who is creating should not add the ownerRef when creating the claim and Not sure adding important input as the annotation is a good idea.

services/provider/server/storageclaim.go
```diff
@@ -0,0 +1,100 @@
...
+		klog.Errorf("failed to generate storageClassClaim hash name for %q for consumer %q. %v", storageClassClaimName, storageConsumerUUID, err)
+		panic("failed to marshal")
+	}
+	name := md5.Sum([]byte(claimName))
+	return hex.EncodeToString(name[:16])
+}
+
+func getStorageClassClaimName(consumerUUID, storageClassClaimName string) string {
+	return fmt.Sprintf("storageclassclaim-%s", generateHashForStorageClassClaim(consumerUUID, storageClassClaimName))
+}
```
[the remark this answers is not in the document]
Madhu-1: keeping this in a single function for now.

services/provider/server/server.go
```diff
@@ -557,12 +563,34 @@ func validateTicket(ticket string, pubKey *rsa.PublicKey) error {
...
+	claimID, err := s.storageClassClaimManager.Create(ctx, req.StorageConsumerUUID, req.StorageClassClaimName, req.StorageType.String(), req.EncryptionMethod)
+	if err != nil {
+		if !kerrors.IsAlreadyExists(err) {
+			return nil, status.Errorf(codes.Internal, "failed to FulfillStorageClassClaim for %q. %v", req.StorageConsumerUUID, err)
+		}
+	}
+
+	return &pb.FulfillStorageClassClaimResponse{
+		StorageClassClaimID: claimID,
+	}, nil
```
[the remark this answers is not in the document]
Madhu-1: This is useful for tracking purposes, even though for today it's not much useful. we can store the claimUUID on the claim Object on the consumer side for identity purposes. If we change the implementation  (breaking changes) this might help RPC calls for resource identification.

services/provider/server/storageclaim.go
```diff
@@ -0,0 +1,100 @@
...
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type ocsStorageClassClaimManager struct {
+	client    client.Client
+	namespace string
+}
```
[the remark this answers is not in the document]
Madhu-1: This is the pattern followed in the consumer.go. as suggested removing the ocs

services/provider/server/storageclaim.go
Madhu-1: >First, do we really need a warning for this case?

This is the followed pattern in the consumer.go

>Second, we do need to consider a request that has a different configuration than the original one. It should fail as we are not allowing updates to storage class claims with this enhancement (maybe in the future)

:+1

services/provider/server/storageclaim.go
Madhu-1: Again this is the followed pattern in consumer.go. this might be helpful for debugging.

services/provider/server/storageclaim.go
```diff
@@ -0,0 +1,100 @@
...
+		StorageClassClaimName string `json:"storageClassClaimName"`
+	}
+
+	s.StorageConsumerUUID = storageConsumerUUID
+	s.StorageClassClaimName = storageClassClaimName
+
+	claimName, err := json.Marshal(s)
+	if err != nil {
+		klog.Errorf("failed to generate storageClassClaim hash name for %q for consumer %q. %v", storageClassClaimName, storageConsumerUUID, err)
+		panic("failed to marshal")
```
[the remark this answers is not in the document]
Madhu-1: `storageclassClaim` is a single word right do we need to separate it out like `storage class claim` while logging in all places?

services/provider/proto/provider.proto
```diff
@@ -147,8 +149,10 @@ message FulfillStorageClassClaimResponse{
 // RevokeStorageClassClaimRequest holds the information required to delete the
 // StorageclassClaim CR on provider cluster.
 message RevokeStorageClassClaimRequest{
-    // storageclassClaim Identified on the provider cluster.
-    string storageClassClaimID = 1;
+    // name of the storageclassclaim on the consumer cluster.
+    string storageClassClaimName = 1;
```
umangachapagain, line 153: What happens on provider, when user deletes and recreates claim on consumer cluster with same name but different spec? Would UUID be better here?

We need a mechanism to cancel older request when we see new updates on claim or it gets re-created.

services/provider/server/storageclaim.go
umangachapagain: Our deletion process depends on StorageClassClaim being owned by StorageConsumer on provider cluster.
Adding explicit ownership should be fine here, since it's internal to the provider and created via automation.

services/provider/server/storageclaim.go
```diff
@@ -0,0 +1,72 @@
...
+	s.StorageClassClaimName = storageClassClaimName
+
+	claimName, err := json.Marshal(s)
+	if err != nil {
+		klog.Errorf("failed to marshal a name for a storage class claim based on %v. %v", s, err)
+		panic("failed to marshal storage class claim name")
+	}
+	name := md5.Sum([]byte(claimName))
+	return fmt.Sprintf("storageclassclaim-%s", hex.EncodeToString(name[:16]))
+}
```
umangachapagain, line 45: Can we add some comments and a unit test to demonstrate how this works?

services/provider/server/server.go
```diff
@@ -74,10 +74,16 @@ func NewOCSProviderServer(ctx context.Context, namespace string) (*OCSProviderSe
 		return nil, fmt.Errorf("failed to create new OCSConumer instance. %v", err)
 	}
 
+	storageClassClaimManager, err := newStorageClassClaimManager(ctx, client, namespace)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create new storageClassClaim instance. %v", err)
```
umangachapagain, line 79:
```suggestion
		return nil, fmt.Errorf("failed to create new StorageClassClaim instance. %v", err)
```

services/provider/server/storageclaim.go
```diff
@@ -0,0 +1,95 @@
...
+	var s struct {
+		StorageConsumerUUID   string `json:"storageConsumerUUID"`
+		StorageClassClaimName string `json:"storageClassClaimName"`
+	}
+	s.StorageConsumerUUID = consumerUUID
+	s.StorageClassClaimName = storageClassClaimName
+
+	claimName, err := json.Marshal(s)
+	if err != nil {
+		klog.Errorf("failed to marshal a name for a storage class claim based on %v. %v", s, err)
```
umangachapagain, line 40:
```suggestion
		klog.Errorf("failed to marshal a name for a storage class claim based on %v. %v", s, err)
```

services/provider/server/storageclaim.go
```diff
@@ -0,0 +1,95 @@
...
+		panic("failed to marshal storage class claim name")
```
umangachapagain, line 41: return an error instead of panic. Panic will shut down everything and exit.

services/provider/server/storageclaim.go
```diff
@@ -0,0 +1,95 @@
...
+		Spec: ocsv1alpha1.StorageClassClaimSpec{
+			Type:             claimType,
+			EncryptionMethod: encryptionMethod,
+		},
+	}
+
+	err := s.client.Create(ctx, storageClassClaimObj)
+	if err != nil {
+		if kerrors.IsAlreadyExists(err) {
+			klog.Warningf("storageClassClaim %q for %q already exists", generatedClaimName, storageClassClaimName)
```
umangachapagain, line 63:
```suggestion
			klog.Warningf("StorageClassClaim %q for %q already exists", generatedClaimName, storageClassClaimName)
```

umangachapagain (changes requested):

services/provider/server/storageclaim.go
Madhu-1: Why this should be done at the RPC level, not at the controller level?

services/provider/proto/provider.proto
Madhu-1: Cancellation can lead to unwanted side effects. the recreation will not go through as the object already exists and it should wait till the deletion is completed.

services/provider/server/storageclaim.go
Madhu-1: what change is expected here? am not able to make out what change is required here.

services/provider/server/storageclaim.go
nb-ohad: @umangachapagain This should be a panic, as this is a dev bug and not a runtime error

Madhu-1: panic is suggested instead of returning error https://github.com/red-hat-storage/ocs-operator/pull/1564#discussion_r815599759

services/provider/server/storageclaim.go
nb-ohad: Two reasons:
1. we want it already set from the creation
2. No need to start a storageconsumer  full reconcile just to add an owner ref

nb-ohad: >IMO the user who is creating should not add the ownerRef when creating the claim and Not sure adding important input as the annotation is a good idea.

It is not the user who is adding it, it is the API server.

services/provider/server/server.go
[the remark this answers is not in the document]
nb-ohad: Why? we already know how to get from a claim name and consumer UID to a remote claim id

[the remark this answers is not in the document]
Madhu-1: that is on the provider side not on the consumer side. Dont we need any tracking on the consumer side?

services/provider/server/storageclaim.go
Madhu-1: done, added code to add ownerRef at the APIServer.

services/provider/proto/provider.proto
nb-ohad: @umangachapagain we did that by design, it is not a side effect

services/provider/server/server.go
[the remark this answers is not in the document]
Madhu-1: Removed claimID from the response forgot to update here.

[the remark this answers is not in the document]
nb-ohad: I think exposing provider implementation details on consumer is problamatic

umangachapagain (commented):
lgtm

/cc @nb-ohad

services/provider/server/storageclaim.go
```diff
@@ -0,0 +1,128 @@
...
+		if kerrors.IsAlreadyExists(err) {
+			newStorageClassClaimObj := &ocsv1alpha1.StorageClassClaim{}
+			getErr := s.client.Get(ctx, client.ObjectKey{Name: generatedClaimName, Namespace: s.namespace}, newStorageClassClaimObj)
+			if getErr != nil {
+				klog.Errorf("failed to get a StorageClassClaim named %q for consumer %q and claim %q. %v", generatedClaimName, consumerUUID, storageClassClaimName, getErr)
+				return getErr
+			}
+			// check if storageClassClaim spec is same and its not getting deleted.
+			if reflect.DeepEqual(storageClassClaimObj.Spec, newStorageClassClaimObj.Spec) && newStorageClassClaimObj.DeletionTimestamp == nil {
+				klog.Warningf("StorageClassClaim %q for %q already exists with same spec", generatedClaimName, storageClassClaimName)
```
nb-ohad, line 85: This is defending not a warning. Also, the way the code is structured means that we will need to duplicate any future work that comes after Line 101 also here. Can we restructure it in a way that will allow flow to resume after the outer if for this case?

nb-ohad (changes requested):

services/provider/server/storageclaim.go
nb-ohad: Maybe something like this
```
if err := s.client.Create(ctx, storageClassClaimObj); err != nil {
	if !kerrors.IsAlreadyExists(err) {
		return fmt.Errorf("failed to create a StorageClassClaim named %q for consumer %q and claim %q. %w", generatedClaimName, consumerUUID, storageClassClaimName, err)
		
	} 
	
	newStorageClassClaimObj := &ocsv1alpha1.StorageClassClaim{}
	if getErr := s.client.Get(ctx, client.ObjectKey{Name: generatedClaimName, Namespace: s.namespace}, newStorageClassClaimObj) {
		klog.Errorf("failed to get a StorageClassClaim named %q for consumer %q and claim %q. %v", generatedClaimName, consumerUUID, storageClassClaimName, getErr)
		return getErr
	}

	// check if the input is different
	if !reflect.DeepEqual(storageClassClaimObj.Spec, newStorageClassClaimObj.Spec) {
		klog.Errorf("StorageClassClaim named %q for consumer %q and claim %q is already exists with different spec (%v) but requested spec (%v)", generatedClaimName, consumerUUID, storageClassClaimName, storageClassClaimObj.Spec, newStorageClassClaimObj.Spec)
		return err;
	}

	// check if the storageClassClaim is getting deleted.
	if newStorageClassClaimObj.DeletionTimestamp != nil {
		klog.Warningf("StorageClassClaim named %q for consumer %q and claim %q is already created but is getting deleted", generatedClaimName, consumerUUID, storageClassClaimName)
	}
}

klog.Infof("successfully created a StorageClassClaim resource %q for consumer %q and claim %q", generatedClaimName, consumerUUID, storageClassClaimName)
return nil
```

Madhu-1: Done.

services/provider/server/storageclaim.go
```diff
@@ -0,0 +1,123 @@
...
+	err := s.client.Create(ctx, storageClassClaimObj)
+	if err != nil {
+		if !kerrors.IsAlreadyExists(err) {
+			return fmt.Errorf("failed to create a StorageClassClaim named %q for consumer %q and claim %q. %w", generatedClaimName, consumerUUID, storageClassClaimName, err)
+		}
+		newStorageClassClaimObj := &ocsv1alpha1.StorageClassClaim{}
+		getErr := s.client.Get(ctx, client.ObjectKey{Name: generatedClaimName, Namespace: s.namespace}, newStorageClassClaimObj)
+		if getErr != nil {
+			klog.Errorf("failed to get a StorageClassClaim named %q for consumer %q and claim %q. %v", generatedClaimName, consumerUUID, storageClassClaimName, getErr)
+			return getErr
```
umangachapagain, line 83: I believe `IsAlreadyExists` error takes precedence over `getErr`. We should return it, so that server can return AlreadyExists code instead of Internal issue.

umangachapagain (approved):

PR #1607, opened 2022-03-28
ocs to ocs: create default storageclassclaims

iamniting:

### DIFF
--- a/controllers/storagecluster/external_resources.go
+++ b/controllers/storagecluster/external_resources.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/go-logr/logr"
 	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v1"
+	ocsv1alpha1 "github.com/red-hat-storage/ocs-operator/api/v1alpha1"
 	statusutil "github.com/red-hat-storage/ocs-operator/controllers/util"
 	cephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -23,6 +24,7 @@ import (
 	"k8s.io/klog/v2"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
@@ -39,6 +41,11 @@ const (
 const (
 	rookCephOperatorConfigName = "rook-ceph-operator-config"
 	rookEnableCephFSCSIKey     = "ROOK_CSI_ENABLE_CEPHFS"
+)
+
+const (
+	// defaultStorageClassClaimLabel is added to all default storage class claims
+	defaultStorageClassClaimLabel = "storageclassclaim.ocs.openshift.io/default"
 )
 
 // ExternalResource contains a list of External Cluster Resources
@@ -288,6 +295,11 @@ func (obj *ocsExternalResources) ensureDeleted(r *StorageClusterReconciler, inst
 
 	if IsOCSConsumerMode(instance) {
 
+		err := r.deleteDefaultStorageClassClaims(instance)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+
 		// skip offboarding if consumer is not onboarded
 		if instance.Status.ExternalStorage.ConsumerID == "" {
 			r.Log.Info("Consumer is not onboarded. Skipping the offboarding request.")
@@ -324,6 +336,107 @@ func (obj *ocsExternalResources) ensureDeleted(r *StorageClusterReconciler, inst
 	}
 
 	return reconcile.Result{}, nil
+}
+
+// claims should be created only once and should not be created/updated again if user deletes/update it.
+func (r *StorageClusterReconciler) createDefaultStorageClassClaims(instance *ocsv1.StorageCluster) error {
+
+	storageClassClaimFile := &ocsv1alpha1.StorageClassClaim{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      generateNameForCephFilesystemSC(instance),
+			Namespace: instance.Namespace,
+			Labels: map[string]string{
+				defaultStorageClassClaimLabel: "true",
+			},
+		},
+		Spec: ocsv1alpha1.StorageClassClaimSpec{
+			Type: "sharedfilesystem",
+		},
+	}
+
+	storageClassClaimBlock := &ocsv1alpha1.StorageClassClaim{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      generateNameForCephBlockPoolSC(instance),
+			Namespace: instance.Namespace,
+			Labels: map[string]string{
+				defaultStorageClassClaimLabel: "true",
+			},
+		},
+		Spec: ocsv1alpha1.StorageClassClaimSpec{
+			Type: "blockpool",
+		},
+	}
+
+	err := r.createAndOwnStorageClassClaim(instance, storageClassClaimFile)
+	if err != nil {
+		return err
+	}
+
+	err = r.createAndOwnStorageClassClaim(instance, storageClassClaimBlock)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (r *StorageClusterReconciler) createAndOwnStorageClassClaim(
+	instance *ocsv1.StorageCluster, claim *ocsv1alpha1.StorageClassClaim) error {
+
+	err := controllerutil.SetOwnerReference(instance, claim, r.Client.Scheme())
+	if err != nil {
+		return err
+	}
+
+	err = r.Client.Create(context.TODO(), claim)
+	if err != nil && !errors.IsAlreadyExists(err) {
+		return err
+	}
+	return nil
+}
+
+func (r *StorageClusterReconciler) deleteDefaultStorageClassClaims(instance *ocsv1.StorageCluster) error {
+
+	storageClassClaims := &ocsv1alpha1.StorageClassClaimList{}
+	err := r.Client.List(context.TODO(), storageClassClaims, &client.ListOptions{
+		Raw: &metav1.ListOptions{LabelSelector: defaultStorageClassClaimLabel}})
+	if err != nil {
+		return err
+	}
+
+	for i := range storageClassClaims.Items {
+		storageClassClaim := &storageClassClaims.Items[i]
+
+		err = r.Client.Delete(context.TODO(), storageClassClaim)
+		if err != nil && !errors.IsNotFound(err) {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (r *StorageClusterReconciler) verifyNoStorageClassClaimsExist(instance *ocsv1.StorageCluster) error {
+
+	storageClassClaims := &ocsv1alpha1.StorageClassClaimList{}
+	err := r.Client.List(context.TODO(), storageClassClaims)
+	if err != nil {
+		return err
+	}
+
+	for i := range storageClassClaims.Items {
+		storageClassClaim := &storageClassClaims.Items[i]
+
+		if _, ok := storageClassClaim.Labels[defaultStorageClassClaimLabel]; !ok {
+			err = fmt.Errorf("Failed to cleanup resources. storageClassClaims are present." +
+				"Delete all storageClassClaims for the cleanup to proceed")
+			r.recorder.ReportIfNotPresent(instance, corev1.EventTypeWarning, "Cleanup", err.Error())
+			r.Log.Error(err, "Waiting for all storageClassClaims to be deleted.")
+			return err
+		}
+	}
+
+	return nil
 }
 
 // createExternalStorageClusterResources creates external cluster resources
--- a/controllers/storagecluster/external_ocs.go
+++ b/controllers/storagecluster/external_ocs.go
@@ -98,6 +98,12 @@ func (r *StorageClusterReconciler) acknowledgeOnboarding(instance *ocsv1.Storage
 		return reconcile.Result{}, err
 	}
 
+	// claims should be created only once and should not be created/updated again if user deletes/update it.
+	err = r.createDefaultStorageClassClaims(instance)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
 	instance.Status.Phase = statusutil.PhaseProgressing
 
 	r.Log.Info("External-OCS:Onboarding is acknowledged successfully.")
--- a/controllers/storagecluster/uninstall_reconciler.go
+++ b/controllers/storagecluster/uninstall_reconciler.go
@@ -286,6 +286,16 @@ func (r *StorageClusterReconciler) verifyNoStorageConsumerExist(instance *ocsv1.
 // Every function that is called within this function should be idempotent
 func (r *StorageClusterReconciler) deleteResources(sc *ocsv1.StorageCluster) (reconcile.Result, error) {
 
+	if IsOCSConsumerMode(sc) {
+		err := r.verifyNoStorageClassClaimsExist(sc)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+	}
+
+	// we do not check if it is a provider before checking storageConsumers CR because of corner case
+	// where user can mark instance.Spec.AllowRemoteStorageConsumers as false and mark CR for deletion immediately.
+	// Which will trigger a deletion without having instance.Spec.AllowRemoteStorageConsumers as true.
 	err := r.verifyNoStorageConsumerExist(sc)
 	if err != nil {
 		return reconcile.Result{}, err


umangachapagain:
> @iamniting @umangachapagain To me, the approach on this PR seems to be the wrong one, because it will not allow the customer to delete the default storage claims (they will get recreated on the next reconcile), which was one of the design goals.
> 
> Can we please move this to OCS initialization?

OCS Initialization doesn't have any context about StorageCluster or mode of deployment.
This logic needs to be in StorageCluster reconciler, but with logic to initialize only.

iamniting:
> > @iamniting @umangachapagain To me, the approach on this PR seems to be the wrong one, because it will not allow the customer to delete the default storage claims (they will get recreated on the next reconcile), which was one of the design goals.
> > Can we please move this to OCS initialization?
> 
> OCS Initialization doesn't have any context about StorageCluster or mode of deployment. This logic needs to be in StorageCluster reconciler, but with logic to initialize only.

We can call this new func while onboarding only and not after that.

umangachapagain:
> 
> We can call this new func while onboarding only and not after that. WDYT? @umangachapagain @nb-ohad
>
@iamniting That is the expectation. Once onboarding is successful, we initialize default claim and let user edit or delete it.
Also, we expect the user to delete it before offboarding.

Madhu-1:
> > We can call this new func while onboarding only and not after that. WDYT? @umangachapagain @nb-ohad
> 
> @iamniting That is the expectation. Once onboarding is successful, we initialize default claim and let user edit or delete it. Also, we expect the user to delete it before offboarding.

We should not expect the user to delete it before offboarding as he didn't create the  Claim. as we are creating by default we need to delete it right?

nb-ohad:
I think the offboarding process itself can clean up these, it has symmetry with the creation of the claims upon onboarding.

iamniting:
> I think the offboarding process itself can clean up these, it has symmetry with the creation of the claims upon onboarding.

We need to make sure all (default/user-created) of them are deleted before offboarding and block storage cluster deletion if they are present.

nb-ohad:
@iamniting I think you misunderstood what I am suggesting. I am suggesting to clean these claims in the storagecluster code on the consumer side as part of the offboarding process - not the offboarding call)

iamniting:
> @iamniting I think you misunderstood what I am suggesting. I am suggesting to clean these claims in the storagecluster code on the consumer side as part of the offboarding process - not the offboarding call)

I did not mean offboarding call, I mean in the offboarding process but before the offboarding call and we should not make offboard call if we have these claims in the consumer. which means we need to wait for them to be deleted.

controllers/storagecluster/external_resources.go
```diff
@@ -326,6 +344,82 @@ func (obj *ocsExternalResources) ensureDeleted(r *StorageClusterReconciler, inst
 	return reconcile.Result{}, nil
 }
 
+func (r *StorageClusterReconciler) createStorageClassClaim(instance *ocsv1.StorageCluster) error {
+
+	for _, desiredStorageClassClaim := range getDefaultStorageClassClaims(instance) {
```
Madhu-1, line 349:
```suggestion
        for _, scc := range getDefaultStorageClassClaims(instance) {
```
Prefer using shorted names for the for loops?

controllers/storagecluster/external_resources.go
```diff
@@ -326,6 +344,82 @@ func (obj *ocsExternalResources) ensureDeleted(r *StorageClusterReconciler, inst
...
+func (r *StorageClusterReconciler) verifyNoStorageClassClaimsExist(instance *ocsv1.StorageCluster) error {
+
+	storageClassClaims := &ocsv1alpha1.StorageClassClaimList{}
+	err := r.Client.List(context.TODO(), storageClassClaims, &client.ListOptions{})
+	if err != nil {
+		return err
+	}
+
+	if len(storageClassClaims.Items) != 0 {
+		err = fmt.Errorf("Failed to cleanup resources. storageClassClaims are present." +
```
Madhu-1, line 388: use errors.New when no error formatting is required.

controllers/storagecluster/generate.go
```diff
@@ -83,3 +83,11 @@ func generateCephReplicatedSpec(initData *ocsv1.StorageCluster, poolType string)
 func generateStorageQuotaName(storageClassName, quotaName string) string {
 	return fmt.Sprintf("ocs-%s-%s", storageClassName, quotaName)
 }
+
+func generateNameForDefaultStorageClassClaimFile(initData *ocsv1.StorageCluster) string {
+	return fmt.Sprintf("%s-claim-file", initData.Name)
```
Madhu-1, line 88: Let the name start with `storageclassclaim-%s-file/block`?

Madhu-1 (changes requested):

controllers/storagecluster/external_resources.go
```diff
@@ -326,6 +344,82 @@ func (obj *ocsExternalResources) ensureDeleted(r *StorageClusterReconciler, inst
 	return reconcile.Result{}, nil
 }
 
+func (r *StorageClusterReconciler) createStorageClassClaim(instance *ocsv1.StorageCluster) error {
```
Madhu-1, line 347: rename function to `createDefaultStorageClassClaims`

controllers/storagecluster/external_resources.go
```diff
@@ -326,6 +344,82 @@ func (obj *ocsExternalResources) ensureDeleted(r *StorageClusterReconciler, inst
...
+		if err != nil && !errors.IsAlreadyExists(err) {
+			return err
+		}
+
+	}
+
+	return nil
+}
+
+func (r *StorageClusterReconciler) deleteStorageClassClaim(instance *ocsv1.StorageCluster) error {
```
Madhu-1, line 366: rename function to `deleteDefaultStorageClassClaims`

controllers/storagecluster/external_resources.go
```diff
@@ -326,6 +334,54 @@ func (obj *ocsExternalResources) ensureDeleted(r *StorageClusterReconciler, inst
...
+func (r *StorageClusterReconciler) createStorageClassClaim(instance *ocsv1.StorageCluster) error {
+
+	StorageClassClaimFile := &ocsv1alpha1.StorageClassClaim{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "storage-class-claim-file",
+			Namespace: instance.Namespace,
+		},
+		Spec: ocsv1alpha1.StorageClassClaimSpec{
+			Type:             "sharedfilesystem",
+			EncryptionMethod: "",
```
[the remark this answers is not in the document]
iamniting: done

controllers/storagecluster/external_resources.go
iamniting: done

controllers/storagecluster/external_resources.go
iamniting: done

controllers/storagecluster/generate.go
iamniting: done

controllers/storagecluster/external_resources.go
iamniting: done

controllers/storagecluster/external_resources.go
iamniting: done

controllers/storagecluster/external_resources.go
```diff
@@ -326,6 +345,81 @@ func (obj *ocsExternalResources) ensureDeleted(r *StorageClusterReconciler, inst
...
+			Name:      generateNameForDefaultStorageClassClaimBlock(instance),
+			Namespace: instance.Namespace,
+		},
+		Spec: ocsv1alpha1.StorageClassClaimSpec{
+			Type: "blockpool",
+		},
+	}
+
+	return []*ocsv1alpha1.StorageClassClaim{StorageClassClaimFile, StorageClassClaimBlock}
+}
```
nb-ohad, line 421: Why do we need this function?
-  it just returns a new instance of hard-coded objects that better to be described at the call site.
-  it will return both in all invocation even if the call site need only one of them 
-  It creates and returns a full nested structure even in cases when we only care about some partial data.

nb-ohad (changes requested):

controllers/storagecluster/external_resources.go
nb-ohad: I do not agree. For a reader, `scc` means nothing while `storageClassClaim` or even just `claim` has a meaning. 
We do not have any limitation on code size these days so why are we insisting on making the codebase less readable?

controllers/storagecluster/generate.go
```diff
@@ -83,3 +83,11 @@ func generateCephReplicatedSpec(initData *ocsv1.StorageCluster, poolType string)
...
 	return fmt.Sprintf("ocs-%s-%s", storageClassName, quotaName)
 }
+
+func generateNameForDefaultStorageClassClaimFile(initData *ocsv1.StorageCluster) string {
+	return fmt.Sprintf("storageclassclaim-%s-file", initData.Name)
+}
+
+func generateNameForDefaultStorageClassClaimBlock(initData *ocsv1.StorageCluster) string {
+	return fmt.Sprintf("storageclassclaim-%s-block", initData.Name)
+}
```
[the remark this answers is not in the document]
iamniting: Using the existing func which were created for storageclasses.

controllers/storagecluster/external_resources.go
```diff
@@ -326,6 +345,81 @@ func (obj *ocsExternalResources) ensureDeleted(r *StorageClusterReconciler, inst
...
+	for _, scc := range getDefaultStorageClassClaims(instance) {
+
+		err := r.Client.Delete(context.TODO(), scc)
+		if err != nil && !errors.IsNotFound(err) {
+			return err
+		}
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
iamniting: done

controllers/storagecluster/external_resources.go
```diff
@@ -326,6 +345,81 @@ func (obj *ocsExternalResources) ensureDeleted(r *StorageClusterReconciler, inst
...
+		}
+
+		err = r.Client.Create(context.TODO(), scc)
+		if err != nil && !errors.IsAlreadyExists(err) {
+			return err
+		}
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
iamniting: removed the loop and added the func

controllers/storagecluster/external_resources.go
```diff
@@ -245,6 +248,12 @@ func (obj *ocsExternalResources) ensureCreated(r *StorageClusterReconciler, inst
 		if instance.Status.ExternalStorage.ConsumerID == "" {
 			return r.onboardConsumer(instance, externalClusterClient)
 		} else if instance.Status.Phase == statusutil.PhaseOnboarding {
+			// create default StorageClassClaim
+			err = r.createDefaultStorageClassClaims(instance)
+			if err != nil {
+				return reconcile.Result{}, err
+			}
```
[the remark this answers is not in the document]
iamniting: done

controllers/storagecluster/external_resources.go
```diff
@@ -288,12 +297,22 @@ func (obj *ocsExternalResources) ensureDeleted(r *StorageClusterReconciler, inst
 
 	if IsOCSConsumerMode(instance) {
 
+		err := r.deleteDefaultStorageClassClaims(instance)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+
```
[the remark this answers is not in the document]
iamniting: checking even before starting uninstall

controllers/storagecluster/external_resources.go
iamniting: removed the func and created variables at call site only

controllers/storagecluster/external_resources.go
```diff
@@ -326,6 +339,105 @@ func (obj *ocsExternalResources) ensureDeleted(r *StorageClusterReconciler, inst
...
+func (r *StorageClusterReconciler) deleteDefaultStorageClassClaims(instance *ocsv1.StorageCluster) error {
+
+	storageClassClaims := &ocsv1alpha1.StorageClassClaimList{}
+	err := r.Client.List(context.TODO(), storageClassClaims, &client.ListOptions{
+		Raw: &metav1.ListOptions{LabelSelector: defaultStorageClassClaimLabel}})
+	if err != nil {
+		return err
+	}
+
+	for _, storageClassClaim := range storageClassClaims.Items {
```
nb-ohad, line 409: `storageClassClaims.Items` is a list of data types, not pointers. 
Using this form of for will create an unnecessary copy of a StorageClassClaim struct (and dependent non pointer structs) on each iteration. 

```suggestion
	for i  := range storageClassClaims.Items {
	    storageClassClaim := &storageClassClaims.Items[i]
```

controllers/storagecluster/external_resources.go
```diff
@@ -326,6 +339,105 @@ func (obj *ocsExternalResources) ensureDeleted(r *StorageClusterReconciler, inst
...
+
+func (r *StorageClusterReconciler) verifyNoStorageClassClaimsExist(instance *ocsv1.StorageCluster) error {
+
+	storageClassClaims := &ocsv1alpha1.StorageClassClaimList{}
+	err := r.Client.List(context.TODO(), storageClassClaims)
+	if err != nil {
+		return err
+	}
+
+	for _, storageClassClaim := range storageClassClaims.Items {
```
nb-ohad, line 428: `storageClassClaims.Items` is a list of data types, not pointers. 
Using this form of for will create an unnecessary copy of a StorageClassClaim struct (and dependent non pointer structs) on each iteration. 

```suggestion
	for i  := range storageClassClaims.Items {
	    storageClassClaim := &storageClassClaims.Items[i]
```

controllers/storagecluster/external_resources.go
```diff
@@ -326,6 +339,105 @@ func (obj *ocsExternalResources) ensureDeleted(r *StorageClusterReconciler, inst
...
+		if _, ok := storageClassClaim.Labels[defaultStorageClassClaimLabel]; !ok {
+			err = inbuilterrors.New("Failed to cleanup resources. storageClassClaims are present." +
+				"Delete all storageClassClaims for the cleanup to proceed")
```
nb-ohad, line 431: Why not use `fmt.Errorf` which is more wildly known?

nb-ohad (changes requested):

controllers/storagecluster/external_resources.go
iamniting: done

controllers/storagecluster/external_resources.go
iamniting: done

controllers/storagecluster/external_resources.go
iamniting: because of this comment https://github.com/red-hat-storage/ocs-operator/pull/1607#discussion_r840252417

controllers/storagecluster/external_resources.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/sha512"
 	"encoding/json"
+	inbuilterrors "errors"
```
umangachapagain, line 7:
```suggestion
	"errors"
```
Keep this as `errors` and rename the other to `apierrors`.

controllers/storagecluster/external_resources.go
```diff
@@ -41,6 +44,11 @@ const (
 	rookEnableCephFSCSIKey     = "ROOK_CSI_ENABLE_CEPHFS"
 )
 
+const (
+	// defaultStorageClassClaimLabel is added to all default storage class claims
+	defaultStorageClassClaimLabel = "claim.ocs.openshift.io/default"
```
umangachapagain, line 49:
```suggestion
	defaultStorageClassClaimLabel = "storageclassclaim.ocs.openshift.io/default"
```

controllers/storagecluster/external_resources.go
```diff
@@ -326,6 +339,108 @@ func (obj *ocsExternalResources) ensureDeleted(r *StorageClusterReconciler, inst
 	return reconcile.Result{}, nil
 }
 
+// claims should be created only once and should not be created/updated again if user deletes/update it.
+func (r *StorageClusterReconciler) createDefaultStorageClassClaims(instance *ocsv1.StorageCluster) error {
+
+	StorageClassClaimFile := &ocsv1alpha1.StorageClassClaim{
```
umangachapagain, line 345:
```suggestion
	storageClassClaimFile := &ocsv1alpha1.StorageClassClaim{
```

controllers/storagecluster/external_resources.go
```diff
@@ -326,6 +339,108 @@ func (obj *ocsExternalResources) ensureDeleted(r *StorageClusterReconciler, inst
...
+			Labels: map[string]string{
+				defaultStorageClassClaimLabel: "true",
+			},
+		},
+		Spec: ocsv1alpha1.StorageClassClaimSpec{
+			Type: "sharedfilesystem",
+		},
+	}
+
+	StorageClassClaimBlock := &ocsv1alpha1.StorageClassClaim{
```
umangachapagain, line 358:
```suggestion
	storageClassClaimBlock := &ocsv1alpha1.StorageClassClaim{
```

controllers/storagecluster/external_resources.go
```diff
@@ -326,6 +339,108 @@ func (obj *ocsExternalResources) ensureDeleted(r *StorageClusterReconciler, inst
...
+
+	err = r.createAndOwnStorageClassClaim(instance, StorageClassClaimBlock)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (r *StorageClusterReconciler) createAndOwnStorageClassClaim(
```
umangachapagain, line 384: Can we refactor this function to take any number of claim objects? We can reduce redundant functions calls that way.

controllers/storagecluster/external_resources.go
```diff
@@ -326,6 +339,108 @@ func (obj *ocsExternalResources) ensureDeleted(r *StorageClusterReconciler, inst
...
+
+	err := controllerutil.SetOwnerReference(instance, claim, r.Client.Scheme())
+	if err != nil {
+		return err
+	}
+
+	err = r.Client.Create(context.TODO(), claim)
+	if err != nil && !errors.IsAlreadyExists(err) {
+		return err
+	}
```
umangachapagain, line 395: Since we want default claims to be created only once, isn't it a bigger issue if there is an existing one with same name? We should probably error out when we think we need to create default claim but for some reason it already exists?

umangachapagain (changes requested):

controllers/storagecluster/external_resources.go
iamniting: I removed this and using fmt itself for error.

controllers/storagecluster/external_resources.go
iamniting: done

controllers/storagecluster/external_resources.go
iamniting: done

controllers/storagecluster/external_resources.go
iamniting: done

controllers/storagecluster/external_resources.go
iamniting: done

controllers/storagecluster/external_resources.go
iamniting: Actually, this is a protection against failure in the creation of a second claim. 

For example, if we fail to create a second one we will return an error which means we will call the same code again to create the second one and the first one will be already existing. If we do not add this condition and fail on it again and again we won't be able to proceed with reconciliation at all.

controllers/storagecluster/external_resources.go
```diff
@@ -326,6 +338,104 @@ func (obj *ocsExternalResources) ensureDeleted(r *StorageClusterReconciler, inst
...
+
+	return nil
+}
+
+func (r *StorageClusterReconciler) createAndOwnStorageClassClaim(
+	instance *ocsv1.StorageCluster, claims ...*ocsv1alpha1.StorageClassClaim) error {
+
+	for _, claim := range claims {
```
nb-ohad, line 381: @iamniting Why do we want back to the for loop? This basically undo the request I had on my last review

iamniting:
@nb-ohad as we are adding the owner reference we can delete the logic to delete the default claims, WDYT @nb-ohad ?

controllers/storagecluster/external_resources.go
nb-ohad: @iamniting Please undo as of comments from last reivew

controllers/storagecluster/external_resources.go
nb-ohad: @iamniting Please undo as of comments from last reivew

controllers/storagecluster/external_resources.go
iamniting: I removed it and made calls twice

controllers/storagecluster/external_resources.go
umangachapagain: In that case, how do you know if the existing claim is the expected default?

controllers/storagecluster/external_resources.go
umangachapagain: Which review comments? If there is a reason why my request doesn't make sense, I'd like to know about it and discuss.

iamniting: https://github.com/red-hat-storage/ocs-operator/pull/1607#discussion_r841260282

controllers/storagecluster/external_resources.go
nb-ohad: 1. This code is only called during onboarding
2. If the customer somehow managed to create an scc with the same name before onboarding then we would like to skip that one, and continue with the creation of the second one

iamniting:
> @nb-ohad as we are adding the owner reference we can delete the logic to delete the default claims, WDYT @nb-ohad ?

As per our experiment with owner reference, we do require the code for deletion.

controllers/storagecluster/external_resources.go
nb-ohad: @umangachapagain the change you requested is just another version of the loop that was there in the first place (see above comment). I don't really care if we have the ability to send multiple CRs to this function as long as we call it one at a time

nb-ohad:
> As per our experiment with owner reference, we do require the code for deletion

Yes, if the owning resource has a finalizer on it, then the owned resource will not get a delete command. This make sense as a finalizer is a signal to k8s that say, I am responsible for my own cleanup

controllers/storagecluster/external_resources.go
umangachapagain: >     2. If the customer somehow managed to create an scc with the same name before onboarding then we would like to skip that one, and continue with the creation of the second one
>
But how do we skip that and create another one with consistent naming? We can not use a different name whenever there is an AlreadyExists error. If we don't return an AlreadyExists error, we won't even know about it.

nb-ohad: > But how do we skip that and create another one with consistent naming?
This is not the use case we are trying to solve. We will not create another one to replace the one the user created. If a customer creates a stoageclassclaim with the same name we assume he knows what he is doing. 

But if the customer only create one of the 2 we will create the second one.

controllers/storagecluster/external_resources.go
nb-ohad: @umangachapagain 
I agree it is a little different, but my point remains. 
Imagen that for the block claim we need to introduce another new step, I think we can agree that both steps that need to apply to the first claim need to go together. 

Without the loop, the change is very straight straightforward, just add some other code after the first call to the function. But if we do it with the requested change, and the mentality of ODF engineers to not touch/refactor code that works, the result will be some new "magic" argument to the function and an `if` inside the function based on that "magic" value - I think we can agree that is no ideal!

If I was the one to write the code, I would not even introduce the function, I would embed the code in the main function, as that function does not have a meaningful logic that should really concern us.  

Every structure that we introduce into the code base is a limiting factor for future changes. 
As stated above, I would not mind that, if the culture was on this project was different.

nb-ohad: At the end of the day, this is very philosophical, and different experiences will result in different opinions. 
There is no actual bug or issue with both approaches, I would suggest letting the contribution decide as we already laied down here our opinions. 

@umangachapagain WDYT?

iamniting: I would like to keep as it is.

PR #1609, opened 2022-03-29
Implement GetStorageClassClaimConfig RPC call

Madhu-1:
Added 4 commits for below

* Add CephResources to the StorageClassClaim status
* Added code to extract the ceph resources from the StorageClassClaim CR status and return back the required details in the
GetStorageClassClaimConfig RPC call.
* Add consumerUUID and storageClassClaimName as annotation when creating the StorageClassClaim CR during FulfillStorageClassClaim RPC
* Remove rbd/cephfs specific resources from GetStorageConfig as GetStorageConfig is used only to get global ceph resources.


services/provider/server/storageclaim.go
```diff
@@ -64,6 +64,10 @@ func (s *storageClassClaimManager) Create(ctx context.Context, consumer *ocsv1al
 			Name:            generatedClaimName,
 			Namespace:       s.namespace,
 			OwnerReferences: []metav1.OwnerReference{ownerRef},
+			Annotations: map[string]string{
+				consumerUUIDAnnotation:          consumerUUID,
```
umangachapagain, line 69: OwnerRef already has consumer UUID. Do we need it as annotation also?
Annotations can't be used as label selector when doing get or list operations. If it's for tracking, label would be a better choice.

services/provider/server/storageclaim.go
```diff
@@ -64,6 +64,10 @@ func (s *storageClassClaimManager) Create(ctx context.Context, consumer *ocsv1al
...
+				storageClassClaimNameAnnotation: storageClassClaimName,
```
umangachapagain, line 70: Why do we need this?

Madhu-1: This is for tracking purposes, adding the claimName as metadata to the claim CR.

services/provider/server/storageclaim.go
Madhu-1: Yes it's for tracking purposes and also helps with listing all claims of a consumer also, let me move it to labels.

umangachapagain (approved):
/hold

waiting for @nb-ohad to ack

services/provider/server/server.go
```diff
@@ -355,7 +357,7 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
 	for _, cephRes := range consumerResource.Status.CephResources {
```
nb-ohad, line 357: As all of the storage consumer direct resources are known ahead of time and the list is static, why do we need to go over the range if we already know need to be created?

My thinking is to remove the whole section and just add the two remaining resources, the health checker client secret and the mon-secret

services/provider/server/server.go
```diff
@@ -366,21 +368,6 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
-				keyProp = "adminKey"
-			}
-			extR = append(extR, &pb.ExternalResource{
-				Name: clientSecretName,
-				Kind: "Secret",
-				Data: mustMarshal(map[string]string{
-					idProp:  cephRes.Name,
-					keyProp: string(cephUserSecret.Data[cephRes.Name]),
-				}),
-			})
```
nb-ohad, line 382: Without this code you will never create the health checker client secret

services/provider/server/storageclaim.go
```diff
@@ -64,6 +65,10 @@ func (s *storageClassClaimManager) Create(ctx context.Context, consumer *ocsv1al
 			Name:            generatedClaimName,
 			Namespace:       s.namespace,
 			OwnerReferences: []metav1.OwnerReference{ownerRef},
+			Labels: map[string]string{
+				consumerUUIDLabel:          consumerUUID,
+				storageClassClaimNameLabel: storageClassClaimName,
+			},
```
nb-ohad, line 71: why labels and not annotations?

services/provider/server/storageclaim.go
```diff
@@ -121,3 +126,19 @@ func (s *storageClassClaimManager) Delete(ctx context.Context, consumerUUID, sto
 
 	return nil
 }
+
+// Get returns the storageClassClaim resource using storageClassClaimName
+// and consumerUUID.
+func (s *storageClassClaimManager) Get(ctx context.Context, consumerUUID, storageClassClaimName string) (*ocsv1alpha1.StorageClassClaim, error) {
+	generatedClaimName := getStorageClassClaimName(consumerUUID, storageClassClaimName)
+	storageClassClaimObj := &ocsv1alpha1.StorageClassClaim{}
+	err := s.client.Get(ctx, types.NamespacedName{Name: generatedClaimName}, storageClassClaimObj)
```
nb-ohad, line 135: Claim is a namespace scoped object, you should pass the namespace name to the get call

services/provider/server/server.go
```diff
@@ -596,5 +526,119 @@ func (s *OCSProviderServer) RevokeStorageClassClaim(ctx context.Context, req *pb
...
 // GetStorageClassClaim RPC call to get the ceph resources for the StorageclassClaim.
 func (s *OCSProviderServer) GetStorageClassClaimConfig(ctx context.Context, req *pb.StorageClassClaimConfigRequest) (*pb.StorageClassClaimConfigResponse, error) {
-	return &pb.StorageClassClaimConfigResponse{}, nil
+	consumerClaimObj, err := s.storageClassClaimManager.Get(ctx, req.StorageConsumerUUID, req.StorageClassClaimName)
+	if err != nil {
+		if kerrors.IsNotFound(err) {
+			return nil, status.Errorf(codes.NotFound, "failed to get storage class claim config %q for %q. %v", req.StorageClassClaimName, req.StorageConsumerUUID, err)
+		}
+		return nil, status.Errorf(codes.Internal, "failed to get storage class claim config %q for %q. %v", req.StorageClassClaimName, req.StorageConsumerUUID, err)
+	}
```
nb-ohad, line 535: This will not work as the errors return from ` s.storageClassClaimManager.Get` are custom ones.
I actually think that analyzing the get error fits more here and as such I will suggest just returning the underlying get error from the manager, as the manager is just an abstraction above k8s client.

services/provider/server/server.go
```diff
@@ -596,5 +526,119 @@ func (s *OCSProviderServer) RevokeStorageClassClaim(ctx context.Context, req *pb
...
+	// state or not before extracting the ceph resources from the StorageClassClaim CR status
+	conString, err := s.getExternalResourcesFromClaim(ctx, consumerClaimObj)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to get external resources. %v", err)
+	}
+
+	return &pb.StorageClassClaimConfigResponse{ExternalResource: conString}, nil
+}
+
+func (s *OCSProviderServer) getExternalResourcesFromClaim(ctx context.Context, storageClassClaim *ocsv1alpha1.StorageClassClaim) ([]*pb.ExternalResource, error) {
```
nb-ohad, line 547: Why is this a separate func? it is basically the implementation for `GetStorageClassClaimConfig` so I see no need to separate it. 

It has a single-use (single call site) and just complicates the reader's life by having him jump all over the file to find different parts of the implementation of the same action.

Can we fold this into the call site?

services/provider/server/server.go
```diff
@@ -596,5 +526,119 @@ func (s *OCSProviderServer) RevokeStorageClassClaim(ctx context.Context, req *pb
...
+	consumerClaimObj, err := s.storageClassClaimManager.Get(ctx, req.StorageConsumerUUID, req.StorageClassClaimName)
+	if err != nil {
+		if kerrors.IsNotFound(err) {
+			return nil, status.Errorf(codes.NotFound, "failed to get storage class claim config %q for %q. %v", req.StorageClassClaimName, req.StorageConsumerUUID, err)
+		}
+		return nil, status.Errorf(codes.Internal, "failed to get storage class claim config %q for %q. %v", req.StorageClassClaimName, req.StorageConsumerUUID, err)
+	}
+
+	// TODO: add status.Phase validation here to check claim is in expected
+	// state or not before extracting the ceph resources from the StorageClassClaim CR status
```
nb-ohad, line 538: Are we waiting for something? Can't we just check for `consumerClaimObj.Status.Phase == "Ready"`?

services/provider/server/server.go
```diff
@@ -596,5 +526,119 @@ func (s *OCSProviderServer) RevokeStorageClassClaim(ctx context.Context, req *pb
...
+
+	return &pb.StorageClassClaimConfigResponse{ExternalResource: conString}, nil
+}
+
+func (s *OCSProviderServer) getExternalResourcesFromClaim(ctx context.Context, storageClassClaim *ocsv1alpha1.StorageClassClaim) ([]*pb.ExternalResource, error) {
+	var extR []*pb.ExternalResource
+	for _, cephRes := range storageClassClaim.Status.CephResources {
+		switch cephRes.Kind {
+		case "CephClient":
+			clientSecretName, _, _, err := s.getCephClientInformation(ctx, cephRes.Name)
```
nb-ohad, line 552: After these changes, the second return value is always `_` in all call sites. 
If we don't need it, let's not return it in the first place.

In addition, the third return value is also meaningless, in this function it is not used and in the original function, it will always be the same value as we are always dealing with the health checker client.
So let's remove it as well and rename the util to `getCephClientSecretName`

nb-ohad (changes requested):

services/provider/server/server.go
```diff
@@ -40,9 +40,11 @@ import (
...
 
 const (
-	TicketAnnotation          = "ocs.openshift.io/provider-onboarding-ticket"
-	ProviderCertsMountPoint   = "/mnt/cert"
-	onboardingTicketKeySecret = "onboarding-ticket-key"
+	TicketAnnotation           = "ocs.openshift.io/provider-onboarding-ticket"
+	ProviderCertsMountPoint    = "/mnt/cert"
+	onboardingTicketKeySecret  = "onboarding-ticket-key"
+	consumerUUIDLabel          = "ocs.openshift.io/storageconsumer-uuid"
+	storageClassClaimNameLabel = "ocs.openshift.io/storageclassclaim-name"
```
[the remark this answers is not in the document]
Madhu-1: Chosen labels over annotation because the consumerID and claimName are added as the identification of the resources and if we add these informations as annotation we cannot do get or list operations based on the annotation.

services/provider/server/server.go
Madhu-1: mon secret is already fetched above i think you meant only health check client.

services/provider/server/server.go
Madhu-1: good catch, missed this one added it now.

services/provider/server/server.go
Madhu-1: As we are returning an error with `%w` from the `s.storageClassClaimManager.Get` IMO it should work as `kerrors.IsNotFound` calls `errors.As` to check the wrapped error message. returning plain k8s error from the `Get` call so that we can return custom errors from here as you said manager is just an abstration.

services/provider/server/server.go
Madhu-1: it just increases the number of lines of code of a single function. thinking more about it even though it's a single time called function to provide abstraction from `Get` and extracting the information from the CR and increases the readability. if you still need this to be changed I can do it. let me know what do you think

services/provider/server/server.go
Madhu-1: removed yet, I didn't optimize anything from previously written functions.

services/provider/server/server.go
Madhu-1: Yes i can do it but below is the reason for not doing it

* Still types (status.Phases) are not defined in the types.go as it need to be added by the server
* Multiple checks need to be added to return the proper GRPC error code.

Let me know if you still want to just add string comparison and keep the TODO and replace it when consts are defined in the storageclassclaim_types.go file.

services/provider/server/server.go
nb-ohad: @Madhu-1 
Num of lines of codes per function should never ever be a consideration. 
There is a wide misconception that small functions are better! Flow is the important matric.

I am providing here two main points to back up the above claim:
1. By breaking stuff into functions you are creating a contract, in the form of function signature and variable scope,  Which by definition is a limiting factor to any code change in the future. And on top of that, the decision to break the flow this way will discourage future contributors to change this decision in the future even in cases where it is necessary (It is just human nature) 
2. There is no obvious location to search for to find the implementation of the action, it is split. It will be split for the reader and it will be split on the call stack when debugging. This just adds unnecessary complexity. Now it is true that on a case by case this seems irrelevant, but by repeating this pattern (splitting functions because you want them to be small) again and again, all over the codebase, the complexity aggregates and becomes meaningfully.

To clarify, I am not against refactoring stuff into functions. But functions, by definition, are a tool for reusability and here I cannot see a case for reusability, not now and not in the future.

services/provider/server/server.go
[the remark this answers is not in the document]
nb-ohad: Ok

services/provider/server/storageclaim.go
nb-ohad: Got the answer from the first reply

services/provider/server/server.go
nb-ohad: So this is going to be implemented in a future PR, correct?

services/provider/server/server.go
nb-ohad: Ok

services/provider/server/server.go
Madhu-1: Yes correct it will be in the future once the implementation is done for storageclaim controller on provider side.

services/provider/server/server.go
```diff
@@ -352,132 +354,73 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
-	}
+	extR = append(extR, &pb.ExternalResource{
+		Name: monSecret,
+		Kind: "Secret",
+		Data: mustMarshal(map[string]string{
+			"fsid":          fsid,
+			"mon-secret":    "mon-secret",
+			"ceph-username": fmt.Sprintf("client.%s", healthCheckName),
+			"ceph-secret":   string(cephUserSecret.Data[healthCheckName]),
+		})})
```
nb-ohad, line 398:
```suggestion
	extR = append(extR, &pb.ExternalResource{
		Name: monSecret,
		Kind: "Secret",
		Data: mustMarshal(map[string]string{
			"fsid":          fsid,
			"mon-secret":    "mon-secret",
			"ceph-username": fmt.Sprintf("client.%s", healthCheckName),
			"ceph-secret":   string(cephUserSecret.Data[healthCheckName]),
		}),
	})
```

nb-ohad (changes requested):

nb-ohad:
@Madhu-1 In addition to the code review above I saw that a lot of unit tests got removed and you didn't write any other unit tests to replace them. Is there a reason?

Madhu-1:
> @Madhu-1 In addition to the code review above I saw that a lot of unit tests got removed and you didn't write any other unit tests to replace them. Is there a reason?

I didn't add any new tests as none of the test cases were removed, As we are removing the cephfs and rbd specific resources from the `GetStorageConfig` i removed the code where we are creating the rb/dcephfs resources and validating the ceph (rbd/cephfs) Resources in GetStorageConfig. Am working on adding the unit testing for the new RPCs (FulfillStorageClassClaim, RevokeStorageClassClaim and GetStorageClassClaimConfig), will be done in the follow-up PR's.

nb-ohad:
Ok, thanks for clarifying

PR #1610, opened 2022-03-29
Bug 2052923: Wildcard permissions removed from rbac definitions

RishabhKodes:
Removed the whitelisted permissions from RBAC definitions and added appropriate permissions.


RishabhKodes:
@jarrpa PTAL

PR #1603, opened 2022-03-24
ocs-to-ocs: storageclassclaim cntrl provider change

subhamkrai:
adding changes in storageClassClaim and storageConsumer controller
to provision required rook resources.

Signed-off-by: subhamkrai <srai@redhat.com>

PR #1623, opened 2022-04-07
ocs-to-ocs: delete ceph resources on SC claim provider

subhamkrai:
adding changes for deletion of Ceph resources on
storageClassClaim on provider side.

### DIFF
--- a/controllers/storageclassclaim/storageclassclaim_controller.go
+++ b/controllers/storageclassclaim/storageclassclaim_controller.go
@@ -20,9 +20,10 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"k8s.io/klog/v2"
 	"reflect"
 	"time"
+
+	"k8s.io/klog/v2"
 
 	"github.com/go-logr/logr"
 	v1 "github.com/red-hat-storage/ocs-operator/api/v1"
@@ -45,11 +46,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
-const (
-	storageClassClaimFinalizer  = "storagesclassclaim.ocs.openshift.io"
-	StorageClassClaimAnnotation = "ocs.openshift.io.storagesclassclaim"
-)
-
 // StorageClassClaimReconciler reconciles a StorageClassClaim object
 // nolint
 type StorageClassClaimReconciler struct {
@@ -69,7 +65,8 @@ type StorageClassClaimReconciler struct {
 }
 
 const (
-	StorageClassClaimFinalizer = "storageclassclaim.ocs.openshift.io"
+	StorageClassClaimFinalizer  = "storageclassclaim.ocs.openshift.io"
+	StorageClassClaimAnnotation = "ocs.openshift.io.storagesclassclaim"
 )
 
 // +kubebuilder:rbac:groups=ocs.openshift.io,resources=storageclassclaims,verbs=get;list;watch;create;update;patch;delete
@@ -410,14 +407,6 @@ func (r *StorageClassClaimReconciler) reconcileProviderPhases() (reconcile.Resul
 	r.storageClassClaim.Status.Phase = v1alpha1.StorageClassClaimCreating
 
 	if r.storageClassClaim.GetDeletionTimestamp().IsZero() {
-		if !controllers.Contains(r.storageClassClaim.GetFinalizers(), storageClassClaimFinalizer) {
-			r.log.Info("Finalizer not found for StorageClassClaim. Adding finalizer.", "StorageClassClaim", klog.KRef(r.storageClassClaim.Namespace, r.storageClassClaim.Name))
-			r.storageClassClaim.SetFinalizers(append(r.storageClassClaim.GetFinalizers(), storageClassClaimFinalizer))
-			if err := r.update(r.storageClassClaim); err != nil {
-				r.log.Error(err, "Failed to update StorageClassClaim with finalizer.", "StorageClassClaim", klog.KRef(r.storageClassClaim.Namespace, r.storageClassClaim.Name))
-				return reconcile.Result{}, err
-			}
-		}
 		if r.storageClassClaim.Spec.Type == "blockpool" {
 
 			if err := r.reconcileCephClientRBDProvisioner(); err != nil {
@@ -457,6 +446,8 @@ func (r *StorageClassClaimReconciler) reconcileProviderPhases() (reconcile.Resul
 			r.storageClassClaim.Status.Phase = v1alpha1.StorageClassClaimReady
 		}
 
+	} else {
+		r.storageClassClaim.Status.Phase = v1alpha1.StorageClassClaimDeleting
 	}
 	return reconcile.Result{}, nil
 }
--- a/controllers/storageconsumer/storageconsumer_controller.go
+++ b/controllers/storageconsumer/storageconsumer_controller.go
@@ -21,7 +21,6 @@ import (
 	"crypto/md5"
 	"encoding/hex"
 	"encoding/json"
-	"fmt"
 
 	"github.com/go-logr/logr"
 	"github.com/red-hat-storage/ocs-operator/api/v1alpha1"
@@ -43,7 +42,6 @@ import (
 )
 
 const (
-	storageConsumerFinalizer      = "storagesconsumer.ocs.openshift.io"
 	StorageConsumerAnnotation     = "ocs.openshift.io.storageconsumer"
 	StorageClaimAnnotation        = "ocs.openshift.io.storageclaim"
 	StorageCephUserTypeAnnotation = "ocs.openshift.io.cephusertype"
@@ -67,6 +65,7 @@ type StorageConsumerReconciler struct {
 //+kubebuilder:rbac:groups=ocs.openshift.io,resources=storageconsumers,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=ceph.rook.io,resources=cephclients,verbs=get;list;watch;create;update;delete
 //+kubebuilder:rbac:groups=ocs.openshift.io,resources=storageconsumers/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=ocs.openshift.io,resources=storageclassclaims,verbs=get;list;
 
 // Reconcile reads that state of the cluster for a StorageConsumer object and makes changes based on the state read
 // and what is in the StorageConsumer.Spec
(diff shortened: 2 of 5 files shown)


subhamkrai:
on top of #1603

nb-ohad:
This PR is highly dependent on #1603 and as there are multiple requests for changes on that PR I will restrain from reviewing this one until #1603 get merged

subhamkrai (commented):
need to address this comment https://github.com/red-hat-storage/ocs-operator/pull/1603#discussion_r851929652

controllers/storageclassclaim/storageclassclaim_controller.go
```diff
@@ -236,6 +236,44 @@ func (r *StorageClassClaimReconciler) reconcileProviderPhases() (reconcile.Resul
...
+					if err := r.delete(cephBlockPool); err != nil {
+						return ctrl.Result{}, fmt.Errorf("unable to delete CephBlockPool : %v", err)
+					}
+				case "CephFilesystemSubVolumeGroup":
+					cephFilesystemSubVolumeGroup := &rookCephv1.CephFilesystemSubVolumeGroup{}
+					cephFilesystemSubVolumeGroup.Name = cephResource.Name
+					cephFilesystemSubVolumeGroup.Namespace = r.namespace
+					if err := r.delete(cephFilesystemSubVolumeGroup); err != nil {
+						return ctrl.Result{}, fmt.Errorf("unable to delete CephFilesystemSubVolumeGroup : %v", err)
+					}
```
nb-ohad, line 273: Why do we need to delete all of these manually?

First, the block pool and subvolumegroups are shared between multiple claims and should not be deleted until the last claim that is using them gets deleted.

Second, the owner references on the dependent resources should automatically accomplish the garbage collection on these resources when the last owner gets removed. It is a built-in feature of k8s.

Please remove this entire code block!

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -185,7 +186,7 @@ func (r *StorageConsumerReconciler) reconcilePhases() (reconcile.Result, error)
 
 		if r.verifyCephResourcesDoNotExist() {
 			r.Log.Info("Removing finalizer from StorageConsumer.", "StorageConsumer", klog.KRef(r.storageConsumer.Namespace, r.storageConsumer.Name))
-			r.storageConsumer.ObjectMeta.Finalizers = remove(r.storageConsumer.ObjectMeta.Finalizers, storageConsumerFinalizer)
+			r.storageConsumer.ObjectMeta.Finalizers = Remove(r.storageConsumer.ObjectMeta.Finalizers, storageConsumerFinalizer)
 			if err := r.Client.Update(r.ctx, r.storageConsumer); err != nil {
 				r.Log.Error(err, "Failed to remove finalizer from StorageConsumer", "StorageConsumer", klog.KRef(r.storageConsumer.Namespace, r.storageConsumer.Name))
 				return reconcile.Result{}, err
```
nb-ohad, line 192: There should be no finalizers on the storageclassclaim in the first place. Adding a finalizer will only work against our  delete process - see same comment on the claim controller

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -250,6 +251,17 @@ func (r *StorageConsumerReconciler) reconcileCephClientHealthChecker() error {
...
+		client.MatchingLabels(map[string]string{ConsumerUUIDLabel: string(r.storageConsumer.UID)}),
+	}
+
+	if err := r.list(storageClassClaimList, storageClassClaimListOption); err == nil && len(storageClassClaimList.Items) > 0 {
+		return false
+	}
+
 	for _, cephResource := range r.storageConsumer.Status.CephResources {
 		switch cephResource.Kind {
 		case "CephClient":
```
nb-ohad, line 267: No need for this manual verification we agreed to use k8s ownership-based deleteion - please see the similar comment on the storageclasscalim controller for more details.

nb-ohad (changes requested):

controllers/storageclassclaim/storageclassclaim_controller.go
```diff
@@ -236,6 +236,44 @@ func (r *StorageClassClaimReconciler) reconcileProviderPhases() (reconcile.Resul
...
+	} else {
+		r.storageClassClaim.Status.Phase = string(v1alpha1.StorageClassClaimStateDeleting)
+
+		if r.verifyCephResourcesDoNotExist() {
+			r.log.Info("Removing finalizer from StorageClassClaim.", "StorageClassClaim", klog.KRef(r.storageClassClaim.Namespace, r.storageClassClaim.Name))
+			r.storageClassClaim.ObjectMeta.Finalizers = controllers.Remove(r.storageClassClaim.ObjectMeta.Finalizers, storageClassClaimFinalizer)
+			if err := r.update(r.storageClassClaim); err != nil {
+				r.log.Error(err, "Failed to remove finalizer from StorageClassClaim", "StorageClassClaim", klog.KRef(r.storageClassClaim.Namespace, r.storageClassClaim.Name))
+				return reconcile.Result{}, err
+			}
```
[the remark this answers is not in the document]
subhamkrai: remove adding/removal of finalizers on both claim and consumer. And, added `blockOwnerDeletion` claim `ownerRef`, also added deletion option foreground deletion.

controllers/storageclassclaim/storageclassclaim_controller.go
```diff
@@ -529,6 +567,37 @@ func addStorageRelatedAnnotations(obj client.Object, storageClassClaimName, stor
...
+			cephFilesystemSubVolumeGroup := &rookCephv1.CephFilesystemSubVolumeGroup{}
+			cephFilesystemSubVolumeGroup.Name = cephResource.Name
+			cephFilesystemSubVolumeGroup.Namespace = r.namespace
+			if err := r.get(cephFilesystemSubVolumeGroup); err == nil || !errors.IsNotFound(err) {
+				return false
+			}
+		}
+	}
+	return true
+}
```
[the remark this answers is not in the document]
subhamkrai: removed

controllers/storageclassclaim/storageclassclaim_controller.go
subhamkrai: removed

controllers/storageconsumer/storageconsumer_controller.go
subhamkrai: removed

controllers/storageclassclaim/storageclassclaim_controller.go
```diff
@@ -218,16 +214,6 @@ func (r *StorageClassClaimReconciler) reconcileConsumerPhases() (reconcile.Resul
...
 
-		// Check if finalizers are present, if not, add them.
-		if !contains(r.storageClassClaim.GetFinalizers(), StorageClassClaimFinalizer) {
-			storageClassClaimRef := klog.KRef(r.storageClassClaim.Name, r.storageClassClaim.Namespace)
-			r.log.Info("Finalizer not found for StorageClassClaim. Adding finalizer.", "StorageClassClaim", storageClassClaimRef)
-			r.storageClassClaim.SetFinalizers(append(r.storageClassClaim.GetFinalizers(), StorageClassClaimFinalizer))
-			if err := r.update(r.storageClassClaim); err != nil {
-				return reconcile.Result{}, fmt.Errorf("failed to update StorageClassClaim [%v] with finalizer: %s", storageClassClaimRef, err)
-			}
-		}
```
nb-ohad, line 229: Why are we removing finalizers from consumer phases? in the consumer, the claim finalizer is needed in order to ensure a revoke call to the provider (which is not. k8s construct and cannot be ensured by k8s delete mechinizem without a finalizer). Please revert this change.

nb-ohad, line 229: Why are we removing finalizers from consumer phases? in the consumer, the claim finalizer is needed in order to ensure a revoke call to the provider (which is not a k8s construct and cannot be ensured by k8s delete mechanism without a finalizer). 

Please revert this change.

controllers/storageclassclaim/storageclassclaim_controller.go
```diff
@@ -410,6 +388,8 @@ func (r *StorageClassClaimReconciler) reconcileProviderPhases() (reconcile.Resul
 			r.storageClassClaim.Status.Phase = v1alpha1.StorageClassClaimReady
 		}
 
+	} else {
+		r.storageClassClaim.Status.Phase = v1alpha1.StorageClassClaimStateDeleting
```
nb-ohad, line 392: Why are we removing the indication that we are at a deletion phase? 
I think this is important information to convey as deletion is not immediate and the claim will stay alive until all sub-resources are deleted.

controllers/storageclassclaim/storageclassclaim_controller.go
```diff
@@ -789,6 +769,11 @@ func addStorageRelatedAnnotations(obj client.Object, storageClassClaimName, stor
 	annotations[controllers.StorageCephUserTypeAnnotation] = cephUserType
 }
 
+func (r *StorageClassClaimReconciler) own(resource metav1.Object) error {
+	// Ensure storageClassClaim ownership on a resource
+	return ctrl.SetControllerReference(r.storageClassClaim, resource, r.Scheme)
+}
```
nb-ohad, line 775: This is in conflict with the `own` function used in consumer mode which sets an OwerReference and not a controller reference. Either way, you cannot use it as a resource can only have a single controller reference and we share resources between multiple claims (block pools etc.)

Please revert, or provide a different implementation

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -286,39 +241,11 @@ func (r *StorageConsumerReconciler) get(obj client.Object) error {
...
-
-// Checks whether a string is contained within a slice
-func Contains(slice []string, s string) bool {
-	for _, item := range slice {
-		if item == s {
-			return true
-		}
-	}
-	return false
-}
```
nb-ohad, line 304: I believe this is used in multiple places (It is a public function) Are we sure it is ok to remove this?

nb-ohad (changes requested):

controllers/storageclassclaim/storageclassclaim_controller.go
subhamkrai: reverted

controllers/storageclassclaim/storageclassclaim_controller.go
subhamkrai: I'm updating here that we are at the deletion stage, until I'm missing something here

controllers/storageclassclaim/storageclassclaim_controller.go
subhamkrai: reverted

controllers/storageconsumer/storageconsumer_controller.go
subhamkrai: Yes, it was throwing golangci lint error also, I make ocs-operator and no error so we are good

PR #1608, opened 2022-03-28
Implement StorageClassClaim creation for consumer cluster

rexagod:
Consumer cluster users should be able to create a StorageClassClaim and get a
StorageClass.

Signed-off-by: Pranshu Srivastava <rexagod@gmail.com>

PR #1628, opened 2022-04-12
Implement StorageClassClaim deletion part for consumer cluster

rexagod:
Consumer cluster users should be able to delete a StorageClassClaim.

***
Please only consider the changes coming in from the last commit, everything else is being worked upon as per the reviews received on #1608.

### DIFF
--- a/controllers/storageclassclaim/storageclassclaim_controller.go
+++ b/controllers/storageclassclaim/storageclassclaim_controller.go
@@ -70,7 +70,6 @@ type StorageClassClaimReconciler struct {
 
 const (
 	StorageClassClaimFinalizer = "storageclassclaim.ocs.openshift.io"
-	StorageClassFinalizer      = "storageclass.ocs.openshift.io"
 )
 
 // +kubebuilder:rbac:groups=ocs.openshift.io,resources=storageclassclaims,verbs=get;list;watch;create;update;patch;delete
@@ -310,8 +309,56 @@ func (r *StorageClassClaimReconciler) reconcileConsumerPhases() (reconcile.Resul
 		// Update the StorageClassClaim status.
 		r.storageClassClaim.Status.Phase = v1alpha1.StorageClassClaimReady
 
+		// Initiate deletion phase if the StorageClassClaim exists.
+	} else if r.storageClassClaim.UID != "" {
+
+		// Deletion phase.
+		// Update the StorageClassClaim status.
+		r.storageClassClaim.Status.Phase = v1alpha1.StorageClassClaimDeleting
+
+		// Delete StorageClass.
+		// Make sure there are no StorageClass consumers left.
+		// Check if StorageClass is in use, if yes, then fail.
+		// Wait until all PVs using the StorageClass under deletion are removed.
+		// Check for any PVs using the StorageClass.
+		pvList := corev1.PersistentVolumeList{}
+		err := r.list(&pvList, client.MatchingFields{
+			"spec.storageClassName": r.storageClassClaim.Name,
+		})
+		if err != nil {
+			return reconcile.Result{}, fmt.Errorf("failed to list PersistentVolumes: %s", err)
+		}
+		if len(pvList.Items) > 0 {
+			return reconcile.Result{}, fmt.Errorf("StorageClass %s is still in use by one or more PV(s)",
+				r.storageClassClaim.Name)
+		}
+
+		// Call `RevokeStorageClassClaim` service on the provider server with StorageClassClaim as a request message.
+		// Check if StorageClassClaim is still exists (it might have been manually removed during the StorageClass
+		// removal above).
+		_, err = providerClient.RevokeStorageClassClaim(
+			r.ctx,
+			r.storageCluster.Status.ExternalStorage.ConsumerID,
+			r.storageClassClaim.Name,
+		)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+
+		storageClass := &storagev1.StorageClass{}
+		storageClass.Name = r.storageClassClaim.Name
+		if err = r.get(storageClass); err != nil && !errors.IsNotFound(err) {
+			return reconcile.Result{}, fmt.Errorf("failed to get StorageClass %s: %s", storageClass.Name, err)
+		}
+		if storageClass.UID != "" {
+
+			if err = r.delete(storageClass); err != nil {
+				return reconcile.Result{}, fmt.Errorf("failed to delete StorageClass %s: %s", storageClass.Name, err)
+			}
+		} else {
+			r.log.Info("StorageClass already deleted.")
+		}
 	}
-	// TODO(@rexagod): Deletion flow for SCC on Consumer side
 
 	return reconcile.Result{}, nil
 }
@@ -424,9 +471,6 @@ func (r *StorageClassClaimReconciler) getCephFSStorageClass(data map[string]stri
 			Annotations: map[string]string{
 				"description": "Provides RWO and RWX Filesystem volumes",
 			},
-			Finalizers: []string{
-				StorageClassFinalizer,
-			},
 		},
 		ReclaimPolicy:        &pvReclaimPolicy,
 		AllowVolumeExpansion: &allowVolumeExpansion,
@@ -446,9 +490,6 @@ func (r *StorageClassClaimReconciler) getCephRBDStorageClass(data map[string]str
 			Annotations: map[string]string{
 				"description": "Provides RWO Filesystem volumes, and RWO and RWX Block volumes",
 			},
-			Finalizers: []string{
-				StorageClassFinalizer,
-			},
 		},
 		ReclaimPolicy:        &pvReclaimPolicy,
 		AllowVolumeExpansion: &allowVolumeExpansion,
@@ -459,11 +500,8 @@ func (r *StorageClassClaimReconciler) getCephRBDStorageClass(data map[string]str
 }
 
 func (r *StorageClassClaimReconciler) createOrReplaceStorageClass(storageClass *storagev1.StorageClass) error {
-	existing := &storagev1.StorageClass{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: r.storageClassClaim.Name,
-		},
-	}
+	existing := &storagev1.StorageClass{}
+	existing.Name = r.storageClassClaim.Name
 
 	if err := r.get(existing); err != nil && !errors.IsNotFound(err) {
 		return fmt.Errorf("failed to get StorageClass: %v", err)
@@ -477,14 +515,8 @@ func (r *StorageClassClaimReconciler) createOrReplaceStorageClass(storageClass *
 	// StorageClass already exists, but parameters have changed. Delete the existing StorageClass and create a new one.
 	if existing.UID != "" {
 
-		// Since we have to update the existing StorageClass, so we will delete the existing storageclass and create a new one.
+		// Since we have to update the existing StorageClass, so we will delete the existing StorageClass and create a new one.
 		r.log.Info("StorageClass needs to be updated, deleting it.", "StorageClass", klog.KRef(storageClass.Namespace, existing.Name))
-
-		// Remove finalizer from StorageClass.
-		storageClass.SetFinalizers(remove(storageClass.Finalizers, StorageClassFinalizer))
-		if err := r.update(storageClass); err != nil {
-			return fmt.Errorf("failed to remove finalizer from StorageClass: %s", err)
-		}
 
 		// Delete the StorageClass.
 		err := r.delete(existing)
@@ -821,13 +853,4 @@ func contains(slice []string, s string) bool {
 		}
 	}
 	return false
-}
-
-func remove(slice []string, s string) []string {
-	for i, item := range slice {
-		if item == s {
-			return append(slice[:i], slice[i+1:]...)
-		}
-	}
-	return slice
 }
--- a/api/v1alpha1/storageclassclaim_types.go
+++ b/api/v1alpha1/storageclassclaim_types.go
@@ -44,6 +44,8 @@ const (
 	StorageClassClaimConfiguring storageClassClaimState = "Configuring"
 	// StorageClassClaimReady represents Ready state of StorageClassClaim
 	StorageClassClaimReady storageClassClaimState = "Ready"
+	// StorageClassClaimDeleting represents Deleting state of StorageClassClaim
+	StorageClassClaimDeleting storageClassClaimState = "Deleting"
 )
 
 // StorageClassClaimStatus defines the observed state of StorageClassClaim


controllers/storageclassclaim/storageclassclaim_controller.go
```diff
@@ -286,8 +286,67 @@ func (r *StorageClassClaimReconciler) reconcileConsumerPhases() (reconcile.Resul
...
+			pvList := &corev1.PersistentVolumeList{}
+			err := r.list(pvList, client.MatchingFields{
+				"spec.storageClassName": r.storageClassClaim.Name,
+			})
+			if err != nil {
+				return reconcile.Result{}, fmt.Errorf("failed to list PVs: %v", err)
+			}
+			if len(pvList.Items) > 0 {
+				return reconcile.Result{}, fmt.Errorf("StorageClass %s is still in use by one or more PV(s)", r.storageClassClaim.Name)
+			}
```
rexagod, line 309: Will remove this (based on recent discussion with Madhu and Ohad).

controllers/storageclassclaim/storageclassclaim_controller.go
```diff
@@ -310,8 +310,54 @@ func (r *StorageClassClaimReconciler) reconcileConsumerPhases() (reconcile.Resul
...
+			} else if errors.IsNotFound(err) {
+				r.log.Info("StorageClass already deleted.")
+			} else {
+				return reconcile.Result{}, fmt.Errorf("failed to get StorageClass %s: %s", existingStorageClass.Name, err)
+			}
+
+			// Call `RevokeStorageClassClaim` service on the provider server with StorageClassClaim as a request message.
+			// Check if StorageClassClaim is still exists (it might have been manually removed during the StorageClass
+			// removal above).
+			if !errors.IsNotFound(r.get(r.storageClassClaim)) {
```
rexagod, line 347: @nb-ohad Is it safe to assume that the `RevokeStorageClassClaim` should only be called when `storageClassClaim` exists? Should we remove this check?

Even if the `storageClassClaim` was deleted manually before the flow reached this line of code, we should still signal for the deletion of the `storageClassClaim` resource on the provider cluster.

controllers/storageclassclaim/storageclassclaim_controller.go
```diff
@@ -310,8 +310,54 @@ func (r *StorageClassClaimReconciler) reconcileConsumerPhases() (reconcile.Resul
 		// Update the StorageClassClaim status.
 		r.storageClassClaim.Status.Phase = v1alpha1.StorageClassClaimReady
 
+	} else {
+
+		// Reconciliation needed if the StorageClassClaim is present.
+		if err = r.get(r.storageClassClaim); !errors.IsNotFound(err) {
```
nb-ohad, line 316: Why do we need to get the claim again? 
Just check `r. storageClassClaim.UID != ""` and fold the `if` into the else above it

controllers/storageclassclaim/storageclassclaim_controller.go
```diff
@@ -310,8 +310,54 @@ func (r *StorageClassClaimReconciler) reconcileConsumerPhases() (reconcile.Resul
...
+			// Deletion phase.
+			// Update the StorageClassClaim status.
+			r.storageClassClaim.Status.Phase = v1alpha1.StorageClassClaimDeleting
+
+			// Delete StorageClass.
+			existingStorageClass := &storagev1.StorageClass{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: r.storageClassClaim.Name,
+				},
+			}
```
nb-ohad, line 327: A small note (not a blocker for the PR), if you are creating an object with only identification information it is more readable to just create an empty object and set `Name` and `Namespace` (if needed). 
The resulting code will be more concise and with no indentation. 

```suggestion
                        existingStorageClass := &storagev1.StorageClass{}
                        existingStorageClass.Name = r.storageClassClaim.Name
```

controllers/storageclassclaim/storageclassclaim_controller.go
```diff
@@ -310,8 +310,54 @@ func (r *StorageClassClaimReconciler) reconcileConsumerPhases() (reconcile.Resul
...
+			if err = r.get(existingStorageClass); err == nil {
+
+				// Remove finalizer from StorageClass.
+				existingStorageClass.SetFinalizers(remove(existingStorageClass.Finalizers, StorageClassFinalizer))
```
nb-ohad, line 331: Please use `GetFinalizers` to get the existing list of finalizers 
```suggestion 
				existingStorageClass.SetFinalizers(remove(existingStorageClass.GetFinalizers(), StorageClassFinalizer))
```

controllers/storageclassclaim/storageclassclaim_controller.go
nb-ohad: Revoking a deleted claim should return an OK response as a deleted claim is a removed claim. 
Let's remove this check.

controllers/storageclassclaim/storageclassclaim_controller.go
```diff
@@ -310,8 +310,54 @@ func (r *StorageClassClaimReconciler) reconcileConsumerPhases() (reconcile.Resul
...
+
+		// Reconciliation needed if the StorageClassClaim is present.
+		if err = r.get(r.storageClassClaim); !errors.IsNotFound(err) {
+
+			// Deletion phase.
+			// Update the StorageClassClaim status.
+			r.storageClassClaim.Status.Phase = v1alpha1.StorageClassClaimDeleting
+
+			// Delete StorageClass.
+			existingStorageClass := &storagev1.StorageClass{
```
nb-ohad, line 323: Why have `existing` in the name of the variable? it just raises questions. Can you please rename it to just `storageClass`?

controllers/storageclassclaim/storageclassclaim_controller.go
```diff
@@ -310,8 +310,54 @@ func (r *StorageClassClaimReconciler) reconcileConsumerPhases() (reconcile.Resul
...
+				},
+			}
+			if err = r.get(existingStorageClass); err == nil {
+
+				// Remove finalizer from StorageClass.
+				existingStorageClass.SetFinalizers(remove(existingStorageClass.Finalizers, StorageClassFinalizer))
+				if err := r.update(existingStorageClass); err != nil {
+					return reconcile.Result{}, fmt.Errorf("failed to remove finalizer from StorageClass: %s", err)
+				}
+				if err = r.delete(existingStorageClass); err != nil {
```
nb-ohad, line 335: I think there is an issue with the order of operations. 
I think we should not delete the storageclass until we got an ack on the revoke operation.

controllers/storageclassclaim/storageclassclaim_controller.go
```diff
@@ -310,8 +310,54 @@ func (r *StorageClassClaimReconciler) reconcileConsumerPhases() (reconcile.Resul
...
+				r.log.Info("StorageClass already deleted.")
+			} else {
+				return reconcile.Result{}, fmt.Errorf("failed to get StorageClass %s: %s", existingStorageClass.Name, err)
+			}
+
+			// Call `RevokeStorageClassClaim` service on the provider server with StorageClassClaim as a request message.
+			// Check if StorageClassClaim is still exists (it might have been manually removed during the StorageClass
+			// removal above).
+			if !errors.IsNotFound(r.get(r.storageClassClaim)) {
+				_, err = providerClient.RevokeStorageClassClaim(r.ctx, r.storageCluster.Status.ExternalStorage.ConsumerID, r.storageClassClaim.Name)
```
nb-ohad, line 348: What will happen if there are still PVs using the resources provisioned by this storageclassclaim? I think we first need to validate customer has deleted all PVs before sending the remove command.

controllers/storageclassclaim/storageclassclaim_controller.go
```diff
@@ -310,8 +310,54 @@ func (r *StorageClassClaimReconciler) reconcileConsumerPhases() (reconcile.Resul
...
+				if err != nil {
+					return reconcile.Result{}, err
+				}
+			}
+
+			// Remove finalizer from StorageClassClaim on consumer cluster.
+			r.storageClassClaim.SetFinalizers(remove(r.storageClassClaim.GetFinalizers(), StorageClassClaimFinalizer))
+			if err = r.Client.Update(r.ctx, r.storageClassClaim); err != nil {
+				return reconcile.Result{}, fmt.Errorf("failed to remove finalizer from StorageClassClaim %s: %s", r.storageClassClaim.Name, err)
```
nb-ohad, line 357: Why do we have a finalizer on the claim object? 
A finalizer on an object work against an ownership-based delete as it prevents k8s from sending the delete operation to the dependent objects. 

If the reason for this finalizer is to keep this object alive until all dependent objects are getting removed than it is better to setup `blockOwnerDeletion: true` on the owner ref and use foreground deletion mode when deleting.

nb-ohad (changes requested):

controllers/storageclassclaim/storageclassclaim_controller.go
rexagod: Removed the PV check based on this: https://github.com/red-hat-storage/ocs-operator/pull/1608#discussion_r860739646, PTAL.

nb-ohad: There is a difference between recreation of a storageclass (which is a superficial change) and revoking a claim which is going to trigger the deletion of backend resources. In the case of the latter I think we do need to validate usage before revoking

controllers/storageclassclaim/storageclassclaim_controller.go
rexagod: I've added a `blockOwnerDeletion` field on the dependant resources. However, I've not removed the finalizer logic as of now, because based on [this](https://unofficial-kubernetes.readthedocs.io/en/latest/concepts/workloads/controllers/garbage-collection/), it demands that:

- A finalizer with value "foregroundDeletion" be added to the list of finalizers.
- Passing the `deleteOptions.propagationPolicy` as `foreground`. This is a bit confusing since we don't call `delete` method on SCClaim (and hence can't pass our own `deleteOptions`, since the SCClaim is already in a deletion phase when we fetch it)?

controllers/storageclassclaim/storageclassclaim_controller.go
```diff
@@ -310,8 +319,60 @@ func (r *StorageClassClaimReconciler) reconcileConsumerPhases() (reconcile.Resul
...
+		}
+
+		// Delete StorageClass.
+		// Make sure there are no StorageClass consumers left.
+		// Check if StorageClass is in use, if yes, then fail.
+		// Wait until all PVs using the StorageClass under deletion are removed.
+		// Check for any PVs using the StorageClass.
+		pvList := corev1.PersistentVolumeList{}
+		err := r.list(&pvList, client.MatchingFields{
+			"spec.storageClassName": r.storageClassClaim.Name,
```
nb-ohad, line 343: This is an optional field in the PV spec. Are we sure we are setting it when creating PVs?
Also, if I remember correctly, `storageClassName` does not support all fields, are we sure this selector works?

controllers/storageclassclaim/storageclassclaim_controller.go
```diff
@@ -310,8 +319,60 @@ func (r *StorageClassClaimReconciler) reconcileConsumerPhases() (reconcile.Resul
 		// Update the StorageClassClaim status.
 		r.storageClassClaim.Status.Phase = v1alpha1.StorageClassClaimReady
 
+	} else if r.storageClassClaim.UID != "" {
```
nb-ohad, line 322: Please add a comment explaining the need for this check

controllers/storageclassclaim/storageclassclaim_controller.go
```diff
@@ -310,8 +319,60 @@ func (r *StorageClassClaimReconciler) reconcileConsumerPhases() (reconcile.Resul
...
+
+		// Delete StorageClass.
+		// Make sure there are no StorageClass consumers left.
+		// Check if StorageClass is in use, if yes, then fail.
+		// Wait until all PVs using the StorageClass under deletion are removed.
+		// Check for any PVs using the StorageClass.
+		pvList := corev1.PersistentVolumeList{}
+		err := r.list(&pvList, client.MatchingFields{
+			"spec.storageClassName": r.storageClassClaim.Name,
+		})
```
nb-ohad, line 344: This check needs to be done before revoking. We do not want to revoke a claim, and delete rook resources related to that claim, if the customer is still using PVs created by that calim

controllers/storageclassclaim/storageclassclaim_controller.go
```diff
@@ -310,8 +319,60 @@ func (r *StorageClassClaimReconciler) reconcileConsumerPhases() (reconcile.Resul
...
+		storageClass.Name = r.storageClassClaim.Name
+		if err = r.get(storageClass); err == nil {
+
+			// Remove finalizer from StorageClass.
+			storageClass.SetFinalizers(remove(storageClass.GetFinalizers(), StorageClassFinalizer))
+			if err := r.update(storageClass); err != nil {
+				return reconcile.Result{}, fmt.Errorf("failed to remove finalizer from StorageClass: %s", err)
+			}
+			if err = r.delete(storageClass); err != nil {
+				return reconcile.Result{}, fmt.Errorf("failed to delete StorageClass %s: %s", storageClass.Name, err)
```
nb-ohad, line 362: We set a finalizer on the storage class, but we do not react to the deletion of storage classes. 
To me it does not make sense, I think the behavior we want is to recreate the storage class in case it got deleted (by a 3rd agent) and we cannot do so if we have a finalizer as the storage class will not get deleted and we will not reinvoke the creation path. 

I think we need to get rid of this finalizer and at the same time validate our creation path is handling recreation in case of an event that was raised as a result of storage class deletion. My guess it is, but we do need to validate.

controllers/storageclassclaim/storageclassclaim_controller.go
```diff
@@ -310,8 +319,60 @@ func (r *StorageClassClaimReconciler) reconcileConsumerPhases() (reconcile.Resul
...
+			}
+		} else if errors.IsNotFound(err) {
+			r.log.Info("StorageClass already deleted.")
+		} else {
+			return reconcile.Result{}, fmt.Errorf("failed to get StorageClass %s: %s", storageClass.Name, err)
+		}
+
```
nb-ohad, line 369: From a readability standpoint, it will be better to handle the errors near the get operation and to move the main code after the error handling.

```suggestion
		if err = r.get(storageClass); err != nil &&  !errors.IsNotFound(err){ 
			return reconcile.Result{}, fmt.Errorf("failed to get StorageClass %s: %s", storageClass.Name, err)
		}
		if (storageClass.UID != "") {
			// Remove finalizer from StorageClass.
			storageClass.SetFinalizers(remove(storageClass.GetFinalizers(), StorageClassFinalizer))
			if err := r.update(storageClass); err != nil {
				return reconcile.Result{}, fmt.Errorf("failed to remove finalizer from StorageClass: %s", err)
			}
			if err = r.delete(storageClass); err != nil {
				return reconcile.Result{}, fmt.Errorf("failed to delete StorageClass %s: %s", storageClass.Name, err)
			}
		} else {
			r.log.Info("StorageClass already deleted.")
		}
```

controllers/storageclassclaim/storageclassclaim_controller.go
```diff
@@ -279,8 +279,17 @@ func (r *StorageClassClaimReconciler) reconcileConsumerPhases() (reconcile.Resul
...
-					err := r.own(secret)
+					secret.OwnerReferences = []metav1.OwnerReference{
+						{
+							APIVersion:         r.storageClassClaim.APIVersion,
+							Kind:               r.storageClassClaim.Kind,
+							Name:               r.storageClassClaim.Name,
+							UID:                r.storageClassClaim.UID,
+							BlockOwnerDeletion: &_true,
+						},
+					}
```
nb-ohad, line 292: Re-thinking the subject. I agree with the fact we need a finalizer, mainly because of the need to invoke an API call during delete (revoke call), so no need for the `BlockOwnerDeletion` field.

In addition, even if we wanted to support the field, I would suggest having it an argument on the `own()` call and not inline the owner ref section.

Can you please revert this back to own?

controllers/storageclassclaim/storageclassclaim_controller.go
```diff
@@ -494,6 +553,15 @@ func (r *StorageClassClaimReconciler) createOrReplaceStorageClass(storageClass *
...
 	r.log.Info("Creating StorageClass.", "StorageClass", klog.KRef(storageClass.Namespace, existing.Name))
+	storageClass.OwnerReferences = []metav1.OwnerReference{
+		{
+			APIVersion:         r.storageClassClaim.APIVersion,
+			Kind:               r.storageClassClaim.Kind,
+			Name:               r.storageClassClaim.Name,
+			UID:                r.storageClassClaim.UID,
+			BlockOwnerDeletion: &_true,
+		},
+	}
```
nb-ohad, line 564: You cannot own a cluster-wide resource from a namespaced resource!
If you used `own` you would get an error (as it is using `SetOwnerReference` internally which preform validation before adding the ownership metadata).

Please remove this code.

nb-ohad (changes requested):

controllers/storageclassclaim/storageclassclaim_controller.go
nb-ohad: Ok, I agree, but for different reasons.

I will start by stating that there is no need to add a finalizer just to make sure dependent resources are getting deleted, this is already taken care of by k8s. But in this case we have special conditions:

1. `storageclass` is NOT owned by `storageclasscalim` as you cannot establish ownership of a clsuter level resource by a namespace level resource
2. we need to ensure that a `revoke` call is dispatched and received on the provider side before we complete the deletion process.

These conditions are a classic use case for the use of finalizers.

controllers/storageclassclaim/storageclassclaim_controller.go
rexagod: It's being set on creation of PVs.

image

> storageClassName does not support all fields

Can you please rephrase, I didn't get this.

nb-ohad: Ho  sorry I meant `client.MatchingFields` not `storageClassName`. I fixed the original comment

controllers/storageclassclaim/storageclassclaim_controller.go
```diff
@@ -223,7 +222,7 @@ func (r *StorageClassClaimReconciler) reconcileConsumerPhases() (reconcile.Resul
 			storageClassClaimRef := klog.KRef(r.storageClassClaim.Name, r.storageClassClaim.Namespace)
 			r.log.Info("Finalizer not found for StorageClassClaim. Adding finalizer.", "StorageClassClaim", storageClassClaimRef)
 			r.storageClassClaim.SetFinalizers(append(r.storageClassClaim.GetFinalizers(), StorageClassClaimFinalizer))
-			if err := r.update(r.storageClassClaim); err != nil {
+			if err = r.update(r.storageClassClaim); err != nil {
```
nb-ohad, line 225: why? 
We should strive to use locality when possible. if there is nothing preventing you from using a local error variable please do so.

controllers/storageclassclaim/storageclassclaim_controller.go
```diff
@@ -280,7 +279,7 @@ func (r *StorageClassClaimReconciler) reconcileConsumerPhases() (reconcile.Resul
 				secret.Name = resource.Name
 				secret.Namespace = r.storageClassClaim.Namespace
 				_, err = controllerutil.CreateOrUpdate(r.ctx, r.Client, secret, func() error {
-					err := r.own(secret)
+					err = r.own(secret)
```
nb-ohad, line 282: Sam as the last comment about variable locality

nb-ohad:
@rexagod I left 2 comments about error variable locality. Please do not consider them as mandatory fixes, otherwise, the PR looks good to me. 

Please squash the commits and then we will merge

PR #1634, opened 2022-04-14
Bug 2075422:Add noobaa operator as dependency in the bundle

iamniting:
The PR fixes the issue of getting mcg 4.10 installed with ocs 4.9

This PR also has fixes from other PR https://github.com/red-hat-storage/ocs-operator/pull/1377 from @jarrpa 

Bug: https://bugzilla.redhat.com/show_bug.cgi?id=2075422


iamniting:
/hold 

After talking to @jarrpa We think that this should be added to the odf-operator as odf is the one that does take care of the subscriptions and all. It requires work in all 3 components odf, ocs, noobaa. We do have a PR for ocs and odf but do require a PR from the noobaa team to delete the dependency on the ocs-operator in their CSV.

jarrpa (changes requested):
To elaborate the above: We explicit avoided this for a reason. At best we might want to backport my changes, but I'm not even sure about that.

nimrod-becker:
How would noobaa upstream work with OBC if we remove the dependency?

iamniting:
closing it as discussed over the meet.

PR #1635, opened 2022-04-18
storagecluster: Add CephNFS deployment

Rakshith-R:
Refer
- https://issues.redhat.com/browse/RHSTOR-2719
- https://issues.redhat.com/browse/RHSTOR-2720

Commits: 

- ####    add `nfs` parameter to storagecluster CR
    
    The following struct has been added to
    storagecluster CR to enable CephNFS CR
    creation with corresponding default nfs
    storageclass.
    
    ```
    spec:
      nfs:
        enable: true
    ```
    
- ####   storagecluster: add capability to deploy cephnfs CR
    
    This commit adds capability to deply cephnfs CR
    when `spec.nfs.enable` is set to true.
    
- ####    storagecluster: add service required for cephnfs
    
    This commit adds capability to deploy k8s service
    which is required by cephnfs. This will be deployed
    when `spec.nfs.enable` is set to true.
    

- ####   storagecluster: deploy `.nfs` cephblockpool required fro cephnfs
    
    This commit adds capability to deploy `.nfs` blockpool CR,
    which is required for cephnfs when `spec.nfs.enable` is set
    to true.

- ####    storagecluster: deploy nfs storageclass when `nfs.enable` is true
    
    This commit adds capability to deploy
    nfs storageclass when `spec.nfs.enable` is set to
    true in storagecluster CR.
    

    Signed-off-by: Rakshith R <rar@redhat.com>


Rakshith-R:
/cc @BlaineEXE @nixpanic @Madhu-1 @agarwal-mudit 
PTAL

api/v1/storagecluster_types.go
```diff
@@ -46,6 +46,7 @@ type StorageClusterSpec struct {
 	MonPVCTemplate     *corev1.PersistentVolumeClaim          `json:"monPVCTemplate,omitempty"`
 	MonDataDirHostPath string                                 `json:"monDataDirHostPath,omitempty"`
 	MultiCloudGateway  *MultiCloudGatewaySpec                 `json:"multiCloudGateway,omitempty"`
+	NFS                *NFSSpec                               `json:"nfs,omitempty"`
```
Madhu-1, line 49: it should be part of ManagedResourcesSpec?

controllers/storagecluster/cephnfs.go
```diff
@@ -0,0 +1,340 @@
...
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	existingCephNFS := cephv1.CephNFS{}
+	err = r.Client.Get(context.TODO(), types.NamespacedName{Name: cephNFS.Name, Namespace: cephNFS.Namespace}, &existingCephNFS)
+	switch {
+	case err == nil:
+		if existingCephNFS.DeletionTimestamp != nil {
+			r.Log.Info("Unable to restore CephNFS because it is marked for deletion.", "CephNFS", klog.KRef(existingCephNFS.Namespace, existingCephNFS.Name))
+			return reconcile.Result{}, fmt.Errorf("failed to restore initialization object %s because it is marked for deletion", existingCephNFS.Name)
```
Madhu-1, line 63:
```suggestion
return reconcile.Result{}, fmt.Errorf("failed to restore initialization object %q because it is marked for deletion", existingCephNFS.Name)
```

controllers/storagecluster/cephnfs.go
```diff
@@ -0,0 +1,340 @@
...
+
+// ensureDeleted deletes the CephNFS resource owned by the StorageCluster
+func (obj *ocsCephNFS) ensureDeleted(r *StorageClusterReconciler, sc *ocsv1.StorageCluster) (reconcile.Result, error) {
+	foundCephNFS := &cephv1.CephNFS{}
+	cephNFS, err := r.newCephNFSInstance(sc)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	err = r.Client.Get(context.TODO(), types.NamespacedName{Name: cephNFS.Name, Namespace: sc.Namespace}, foundCephNFS)
```
Madhu-1, line 94: context.TODO is used in multiple places, create a var and use it?

controllers/storagecluster/cephnfs.go
```diff
@@ -0,0 +1,340 @@
...
+	if cephBlockPool.GetDeletionTimestamp().IsZero() {
+		r.Log.Info("Uninstall: Deleting CephBlockPool.", "CephBlockPool", klog.KRef(cephBlockPool.Namespace, cephBlockPool.Name))
+		err = r.Client.Delete(context.TODO(), foundCephBlockPool)
+		if err != nil {
+			r.Log.Error(err, "Uninstall: Failed to delete CephBlockPool.", "CephBlockPool", klog.KRef(foundCephBlockPool.Namespace, foundCephBlockPool.Name))
+			return reconcile.Result{}, fmt.Errorf("uninstall: Failed to delete CephBlockPool %v: %v", foundCephBlockPool.Name, err)
+		}
+	}
+
+	err = r.Client.Get(context.TODO(), types.NamespacedName{Name: cephBlockPool.Name, Namespace: sc.Namespace}, foundCephBlockPool)
```
Madhu-1, line 219: why do we need more one Get here?

controllers/storagecluster/cephnfs.go
```diff
@@ -0,0 +1,340 @@
...
+	if cephNFS.GetDeletionTimestamp().IsZero() {
+		r.Log.Info("Uninstall: Deleting CephNFS.", "CephNFS", klog.KRef(foundCephNFS.Namespace, foundCephNFS.Name))
+		err = r.Client.Delete(context.TODO(), foundCephNFS)
+		if err != nil {
+			r.Log.Error(err, "Uninstall: Failed to delete CephNFS.", "CephNFS", klog.KRef(foundCephNFS.Namespace, foundCephNFS.Name))
+			return reconcile.Result{}, fmt.Errorf("uninstall: Failed to delete CephNFS %v: %v", foundCephNFS.Name, err)
+		}
+	}
+
+	err = r.Client.Get(context.TODO(), types.NamespacedName{Name: cephNFS.Name, Namespace: sc.Namespace}, foundCephNFS)
```
Madhu-1, line 113: why one more get again here?

controllers/storagecluster/cephnfs.go
```diff
@@ -0,0 +1,340 @@
...
+	case errors.IsNotFound(err):
+		r.Log.Info("Creating CephBlockPool.", "CephBlockPool", klog.KRef(cephBlockPool.Namespace, cephBlockPool.Name))
+		err = r.Client.Create(context.TODO(), cephBlockPool)
+		if err != nil {
+			r.Log.Error(err, "Failed to create CephBlockPool.", "CephBlockPool", klog.KRef(cephBlockPool.Namespace, cephBlockPool.Name))
+			return reconcile.Result{}, err
+		}
+	}
+
+	return reconcile.Result{}, nil
```
Madhu-1, line 190: in case of other errors we are returning nil?

controllers/storagecluster/cephnfs.go
```diff
@@ -0,0 +1,340 @@
...
+	case errors.IsNotFound(err):
+		r.Log.Info("Creating CephNFS.", "CephNFS", klog.KRef(cephNFS.Namespace, cephNFS.Name))
+		err = r.Client.Create(context.TODO(), cephNFS)
+		if err != nil {
+			r.Log.Error(err, "Unable to create CephNFS.", "CephNFS", klog.KRef(cephNFS.Namespace, cephNFS.Name))
+			return reconcile.Result{}, err
+		}
+	}
+
+	return reconcile.Result{}, nil
```
Madhu-1, line 83: returning nil if both error case as no default is defined?

controllers/storagecluster/cephnfs.go
```diff
@@ -0,0 +1,340 @@
...
+
+	nfsService, err := r.newNFSService(instance)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	existingNFSService := v1.Service{}
+	err = r.Client.Get(context.TODO(), types.NamespacedName{Name: nfsService.Name, Namespace: nfsService.Namespace}, &existingNFSService)
+	switch {
+	case err == nil:
```
Madhu-1, line 277: default case is required for others errors

controllers/storagecluster/storageclasses.go
```diff
@@ -231,6 +231,38 @@ func newCephBlockPoolStorageClassConfiguration(initData *ocsv1.StorageCluster) S
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name: generateNameForCephNetworkFilesystemSC(initData),
+				Annotations: map[string]string{
+					"description": "Provides RWO and RWX Filesystem volumes",
+				},
+			},
+			Provisioner:   fmt.Sprintf("%s.nfs.csi.ceph.com", initData.Namespace),
+			ReclaimPolicy: &persistentVolumeReclaimDelete,
+			// AllowVolumeExpansion is set to true to enable expansion of OCS backed Volumes
+			AllowVolumeExpansion: &allowVolumeExpansion,
```
Madhu-1, line 249: its set to false at line 237

Madhu-1 (changes requested):

controllers/storagecluster/cephnfs.go
Rakshith-R: > why one more get again here?

This get is to make sure, the resource is deleted. 
If the resource is still present, waiting error is returned to trigger backoff reconcile, please see line 120-121 below.

Madhu-1: isnt Delete a blocking call?

controllers/storagecluster/cephnfs.go
Rakshith-R: > in case of other errors we are returning nil?

That's the pattern being following in other types too 
https://github.com/red-hat-storage/ocs-operator/blob/49c7a58a2d1ae83caa456366db6bab5691132ec0/controllers/storagecluster/cephblockpools.go#L126

https://github.com/red-hat-storage/ocs-operator/blob/49c7a58a2d1ae83caa456366db6bab5691132ec0/controllers/storagecluster/cephfilesystem.go#L106

Only error nil and notfound cases are handled.

controllers/storagecluster/cephnfs.go
Rakshith-R: > why do we need more one Get here?

same explanation as commented here https://github.com/red-hat-storage/ocs-operator/pull/1635/files#r852009853

This is the pattern that is currently followed in the existing code.

controllers/storagecluster/storageclasses.go
Rakshith-R: > its set to false at line 237

Yup, ceph-csi does not currently allow it. 
and it receives a pointer to a bool. 

Should I change the variable name? any suggestions?

Madhu-1: Just Remove the comment it should be fine.

Rakshith-R: > Just Remove the comment it should be fine.

okay, :laughing: totally missed that comment part.

api/v1/storagecluster_types.go
Rakshith-R: > it should be part of ManagedResourcesSpec?

I don't think adding NFS to ManagedResourcesSpec and related additional options is really required for tech preview, 
maybe we can add that support later on.

controllers/storagecluster/cephnfs.go
Rakshith-R: > isnt Delete a blocking call?

I am not sure, I'll get back after testing it. 

Meanwhile, Other types' `ensureDeleted` follow the same format of checking the resource's existence after deletion.

api/v1/storagecluster_types.go
Madhu-1: AFAIK it should be part of the ManagedResourcesSpec which will provide fine-grained control for the users. For me, it does not feel okay to add the NFS section here its ceph resources like others in the managerResourcesSpec.  @jarrpa @umangachapagain can help here.

controllers/storagecluster/cephnfs.go
Rakshith-R: No, the delete is not blocking.

```
{"level":"info","ts":1650279363.4743314,"logger":"controllers.StorageCluster","msg":"Uninstall: Deleting CephBlockPool.","Request.Namespace":"openshift-storage","Request.Name":"ocs-storagecluster","CephBlockPool":"openshift-storage/ocs-storagecluster-cephblockpool"}
{"level":"error","ts":1650279363.4865177,"logger":"controllers.StorageCluster","msg":"Uninstall: Waiting for CephBlockPool to be deleted.","Request.Namespace":"openshift-storage","Request.Name":"ocs-storagecluster","CephBlockPool":"openshift-storage/ocs-storagecluster-cephblockpool","stacktrace":"github.com/red-hat-storage/ocs-operator/controllers/storagecluster.(*StorageClusterReconciler).deleteResources\n\t/remote-source/app/controllers/storagecluster/uninstall_reconciler.go:333\ngithub.com/red-hat-storage/ocs-operator/controllers/storagecluster.(*StorageClusterReconciler).reconcilePhases\n\t/remote-source/app/controllers/storagecluster/reconcile.go:334\ngithub.com/red-hat-storage/ocs-operator/controllers/storagecluster.(*StorageClusterReconciler).Reconcile\n\t/remote-source/app/controllers/storagecluster/reconcile.go:161\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/remote-source/app/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:298\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/remote-source/app/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:253\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/remote-source/app/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:214"}
```

The r.Log.Error just before last return statement in every `ensureDelete` func is called with a `nil` error which leads to a panic.

api/v1/storagecluster_types.go
BlaineEXE: It seems to me like ManagedResourcesSpec is the right place as well. I see CephBlockPools, CephFileSystems, and CephObjectStores there, and CephNFS is the same form of abstraction as those.

controllers/storagecluster/cephnfs.go
```diff
@@ -0,0 +1,346 @@
...
+			return reconcile.Result{}, err
+		}
+	case errors.IsNotFound(err):
+		r.Log.Info("Creating CephNFS.", "CephNFS", klog.KRef(cephNFS.Namespace, cephNFS.Name))
+		err = r.Client.Create(ctxTODO, cephNFS)
+		if err != nil {
+			r.Log.Error(err, "Unable to create CephNFS.", "CephNFS", klog.KRef(cephNFS.Namespace, cephNFS.Name))
+			return reconcile.Result{}, err
+		}
+	}
```
BlaineEXE, line 82: Where is the case for handling errors? Currently this handles `err == nil` and `err != nil && err.IsNotFound(err)`, but it doesn't handle `err != nil && !errors.IsNotFound(err)` which is the actual error case.

(@jarrpa is this intentional? I also see this left out of the CephFilesystem controller)

I think the easiest way to handle this is like the flow in `ensureDeleted` below that looks like:
```go
if err != nil {
	if errors.IsNotFound(err) {
		// Create new
		r.Log.Info("Creating CephNFS.", "CephNFS", klog.KRef(cephNFS.Namespace, cephNFS.Name))
	}
	// Error case
}
// Non-error case
r.Log.Info("Restoring original CephNFS.", "CephNFS", klog.KRef(cephNFS.Namespace, cephNFS.Name))
```

controllers/storagecluster/cephnfs.go
BlaineEXE: @jarrpa is this "pattern" intentional, or is it a mistake? 

If it's a mistake, we should not repeat it again here.

pkg/deploy-manager/storagecluster.go
```diff
@@ -91,6 +91,9 @@ func (t *DeployManager) DefaultStorageCluster() (*ocsv1.StorageCluster, error) {
 					},
 				},
 			},
+			NFS: &ocsv1.NFSSpec{
+				Enable: true,
+			},
```
BlaineEXE, line 96: Does this mean NFS is enabled by default? If so, this should be made disabled by default.

controllers/storagecluster/uninstall_reconciler.go
```diff
@@ -332,6 +332,9 @@ func (r *StorageClusterReconciler) deleteResources(sc *ocsv1.StorageCluster) (re
 		&ocsCephObjectStoreUsers{},
 		&ocsCephObjectStores{},
 		&ocsCephRbdMirrors{},
+		&ocsCephNFS{},
+		&ocsCephNFSService{},
```
BlaineEXE, line 336: Two things:
1. Following the conventions from above, this should be `ocsCephNFSes` and `ocsCephNFSServices` (plural).
2. Since OCS creates a service for each CephNFS, why not include the service creation and deletion as part of creating/deleting CephNFSes?

controllers/storagecluster/reconcile.go
```diff
@@ -370,6 +370,9 @@ func (r *StorageClusterReconciler) reconcilePhases(
 				&ocsCephCluster{},
 				&ocsCephBlockPools{},
 				&ocsCephFilesystems{},
+				&ocsCephNFSBlockPool{},
+				&ocsCephNFS{},
+				&ocsCephNFSService{},
```
BlaineEXE, line 375: Two things:
1. Following the conventions from above, this should be `ocsCephNFSes` and `ocsCephNFSServices` (plural).
2. Since OCS creates a service for each CephNFS, why not include the service creation and deletion as part of creating/deleting CephNFSes?

controllers/storagecluster/uninstall_reconciler.go
```diff
@@ -332,6 +332,9 @@ func (r *StorageClusterReconciler) deleteResources(sc *ocsv1.StorageCluster) (re
...
+		&ocsCephNFSBlockPool{},
```
BlaineEXE, line 337: Why does this implementation treat the NFS block pool separately from the other `CephBlockPools` yet treat the NFS StorageClass together with the other StorageClasses? Shouldn't all resources follow similar patterns?

@jarrpa do you have implementation preferences on this matter (ocs-op should create a CephBlockPool to configure the shared config location for all CephNFSes)?

controllers/storagecluster/storageclasses.go
```diff
@@ -231,6 +245,38 @@ func newCephBlockPoolStorageClassConfiguration(initData *ocsv1.StorageCluster) S
...
+					"description": "Provides RWO and RWX Filesystem volumes",
+				},
+			},
+			Provisioner:          fmt.Sprintf("%s.nfs.csi.ceph.com", initData.Namespace),
+			ReclaimPolicy:        &persistentVolumeReclaimDelete,
+			AllowVolumeExpansion: &allowVolumeExpansion,
+			Parameters: map[string]string{
+				"clusterID":  initData.Namespace,
+				"nfsCluster": generateNameForCephNFS(initData),
+				"fsName":     fmt.Sprintf("%s-cephfilesystem", initData.Name),
```
BlaineEXE, line 267: Why is there no `generateNameForCephFilesystem()` being used here?

controllers/storagecluster/storageclasses.go
```diff
@@ -231,6 +245,38 @@ func newCephBlockPoolStorageClassConfiguration(initData *ocsv1.StorageCluster) S
...
+	allowVolumeExpansion := false
+	return StorageClassConfiguration{
+		storageClass: &storagev1.StorageClass{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: generateNameForCephNetworkFilesystemSC(initData),
+				Annotations: map[string]string{
+					"description": "Provides RWO and RWX Filesystem volumes",
+				},
+			},
+			Provisioner:          fmt.Sprintf("%s.nfs.csi.ceph.com", initData.Namespace),
```
BlaineEXE, line 261: Why is there no `generateNameForNFSCSIProvisioner()` (or similar) here?

controllers/storagecluster/storageclasses.go
Rakshith-R: Must have missed it, added it now.

controllers/storagecluster/storageclasses.go
Rakshith-R: done

api/v1/storagecluster_types.go
Rakshith-R: According to 
```
	// ManagedResources specifies how to deal with auxiliary resources reconciled
	// with the StorageCluster
	ManagedResources ManagedResourcesSpec `json:"managedResources,omitempty"`
```
This provides extra set of options to controller resources already deployed by OCS-op,
Considering NFS is a new addition, I think we should have it at the top level.
This option will be just to turn it on. 

We can add advanced options later to ManagedResourcesSpec under CephNFSes.

pkg/deploy-manager/storagecluster.go
Rakshith-R: > Does this mean NFS is enabled by default? If so, this should be made disabled by default.

This is only being used for e2e and to start the cluster with scripts
https://github.com/red-hat-storage/ocs-operator/blob/df8317888c7f65d59fe1840a3622843fd061efd2/pkg/deploy-manager/storagecluster.go#L28

controllers/storagecluster/uninstall_reconciler.go
Rakshith-R: > Why does this implementation treat the NFS block pool separately from the other `CephBlockPools` yet treat the NFS StorageClass together with the other StorageClasses? Shouldn't all resources follow similar patterns?
> 
> @jarrpa do you have implementation preferences on this matter (ocs-op should create a CephBlockPool to configure the shared config location for all CephNFSes)?

moved it with cephblockpool

controllers/storagecluster/cephnfs.go
Rakshith-R: Added default case to handle error now.

controllers/storagecluster/cephnfs.go
Rakshith-R: > default case is required for others errors

added

controllers/storagecluster/reconcile.go
Rakshith-R: > Two things:
> 
> 1. Following the conventions from above, this should be `ocsCephNFSes` and `ocsCephNFSServices` (plural).

It now follows this pattern.

> 2. Since OCS creates a service for each CephNFS, why not include the service creation and deletion as part of creating/deleting CephNFSes?

It will lead to a very complicated `ensureCreated` and `ensureDeleted` function. 
It is much cleaner with two different structs.

controllers/storagecluster/cephnfs.go
Rakshith-R: added default case to handle error.

controllers/storagecluster/cephblockpools.go
```diff
@@ -69,6 +69,27 @@ func (r *StorageClusterReconciler) newCephBlockPoolInstances(initData *ocsv1.Sto
 			},
 		},
 	}
+
+	// create `.nfs` cephblockpool if NFS is enabled.
+	if initData.Spec.NFS != nil && initData.Spec.NFS.Enable {
+		ret = append(ret,
+			&cephv1.CephBlockPool{
```
nixpanic, line 76: This is not a pool for block images, it is a RADOS pool that contains objects which are snippets of NFS-Ganesha configuration files. Depending on the Ceph version, the name of the pool should be "nfs-ganesha", or for more recent versions ".nfs" as you have now.

I would have expected Rook to create the pool with the right name when the CephNFS resource is initialized. But maybe there is a reason it should be done here?

controllers/storagecluster/cephnfs.go
```diff
@@ -0,0 +1,274 @@
...
+				Name:      generateNameForCephNFS(initData),
+				Namespace: initData.Namespace,
+			},
+			Spec: cephv1.NFSGaneshaSpec{
+				Server: cephv1.GaneshaServerSpec{
+					Active:    1,
+					Placement: getPlacement(initData, "nfs"),
+					Resources: defaults.GetDaemonResources("nfs", initData.Spec.Resources),
+					// set PriorityClassName for the NFS pods
+					PriorityClassName: openshiftUserCritical,
```
nixpanic, line 36: a comment explaining _why_ it is `openshiftUserCritical` would be more useful than explaining that it gets set

I assume it is critical because Pods may use the NFS-volumes, and without it, they can not get started/stopped?

controllers/storagecluster/storageclasses.go
```diff
@@ -231,6 +245,38 @@ func newCephBlockPoolStorageClassConfiguration(initData *ocsv1.StorageCluster) S
...
+				},
+			},
+			Provisioner:          generateNameForNFSCSIProvisioner(initData),
+			ReclaimPolicy:        &persistentVolumeReclaimDelete,
+			AllowVolumeExpansion: &allowVolumeExpansion,
+			Parameters: map[string]string{
+				"clusterID":  initData.Namespace,
+				"nfsCluster": generateNameForCephNFS(initData),
+				"fsName":     generateNameForCephFilesystem(initData),
+				"server":     generateNameForNFSService(initData),
```
nixpanic, line 268: It would be good to set a volume prefix, otherwise it will inherit the default from the CephFS provisioner. In Ceph-CSI the example sets it to `volumeNamePrefix: nfs-export-`.

controllers/storagecluster/cephblockpools.go
Rakshith-R: > This is not a pool for block images, it is a RADOS pool that contains objects which are snippets of NFS-Ganesha configuration files. Depending on the Ceph version, the name of the pool should be "nfs-ganesha", or for more recent versions ".nfs" as you have now.

This feature will be present from odf 4.11 onwards, which will have ceph pacific 16.2.7+ 
We don't need go worry about backward support as far as we discussed
@

> I would have expected Rook to create the pool with the right name when the CephNFS resource is initialized. But maybe there is a reason it should be done here?

So that extra parameters can be added, replication spec and failure domain 


BlaineEXE can confirm both the above statement.

controllers/storagecluster/storageclasses.go
Rakshith-R: > It would be good to set a volume prefix, otherwise it will inherit the default from the CephFS provisioner. In Ceph-CSI the example sets it to `volumeNamePrefix: nfs-export-`.

Makes sense to me.
@Madhu-1 @BlaineEXE wdyt?

Madhu-1: sounds good to me, make sure this is well tested. `volumeNamePrefix` is not very well tested.

Rakshith-R: > sounds good to me, make sure this is well tested. `volumeNamePrefix` is not very well tested.

There is e2e for this at cephcsi ci https://github.com/ceph/ceph-csi/blob/2b71aac752614182fb0ba5dd956307401d2920b1/e2e/cephfs.go#L420 
Anything else needs to be tested?

controllers/storagecluster/cephnfs.go
Rakshith-R: done

controllers/storagecluster/storageclasses.go
Rakshith-R: added this now

nixpanic:
On Tue, Apr 19, 2022 at 02:19:23AM -0700, Madhu Rajanna wrote:
> @Madhu-1 commented on this pull request.
> 
> 
> 
> > +	return StorageClassConfiguration{
> +		storageClass: &storagev1.StorageClass{
> +			ObjectMeta: metav1.ObjectMeta{
> +				Name: generateNameForCephNetworkFilesystemSC(initData),
> +				Annotations: map[string]string{
> +					"description": "Provides RWO and RWX Filesystem volumes",
> +				},
> +			},
> +			Provisioner:          generateNameForNFSCSIProvisioner(initData),
> +			ReclaimPolicy:        &persistentVolumeReclaimDelete,
> +			AllowVolumeExpansion: &allowVolumeExpansion,
> +			Parameters: map[string]string{
> +				"clusterID":  initData.Namespace,
> +				"nfsCluster": generateNameForCephNFS(initData),
> +				"fsName":     generateNameForCephFilesystem(initData),
> +				"server":     generateNameForNFSService(initData),
> 
> sounds good to me, make sure this is well tested. `volumeNamePrefix` is not very well tested.

For me it works well with the examples in the Ceph-CSI repository :-)

controllers/storagecluster/storageclasses.go
Madhu-1: Yes, we added this in upstream long back. It's only CI tested just to make sure some corner cases like provisioner restart etc are well tested.

BlaineEXE (approved):
Thanks @Rakshith-R. All of this looks good to me as much as I can tell.

BlaineEXE:
It seems that I don't have the best understanding of how openshift's merge bot works. I was just typing up a message that I would wait to merge until  @nixpanic and @Madhu-1 have also approved to make sure I didn't miss anything.

agarwal-mudit:
@BlaineEXE you are a part of https://github.com/red-hat-storage/ocs-operator/blob/main/OWNERS
That's why after your approval it was merged.

@jarrpa @umangachapagain please take a look and if there are further comments @Rakshith-R can address in a follow up PR

PR #1636, opened 2022-04-18
add StorageType as input param for FulfillStorageClassClaim

Madhu-1:
client API should not have the GRPC messages as the input for that reason introducing a new StorageType const so that setting the GRPC storageType can be done inside the FulfillStorageClassClaim client function.

/assign @nb-ohad

### DIFF
--- a/services/provider/client/client.go
+++ b/services/provider/client/client.go
@@ -134,16 +134,29 @@ func (cc *OCSProviderClient) AcknowledgeOnboarding(ctx context.Context, consumer
 	return cc.Client.AcknowledgeOnboarding(apiCtx, req)
 }
 
-func (cc *OCSProviderClient) FulfillStorageClassClaim(ctx context.Context, consumerUUID, storageClassClaimName, encryptionMethod string, storageType pb.FulfillStorageClassClaimRequest_StorageType) (*pb.FulfillStorageClassClaimResponse, error) {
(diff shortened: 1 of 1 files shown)


services/provider/client/client.go
```diff
@@ -134,16 +134,29 @@ func (cc *OCSProviderClient) AcknowledgeOnboarding(ctx context.Context, consumer
 	return cc.Client.AcknowledgeOnboarding(apiCtx, req)
 }
 
-func (cc *OCSProviderClient) FulfillStorageClassClaim(ctx context.Context, consumerUUID, storageClassClaimName, encryptionMethod string, storageType pb.FulfillStorageClassClaimRequest_StorageType) (*pb.FulfillStorageClassClaimResponse, error) {
+type StorageType uint
+
+const (
+	Blockpool StorageType = iota
+	Sharedfilesystem
```
nb-ohad, line 141: The names `Blockpool` and `Sharedfilesystem` are very common throughout the codebase can we prefix them to prevent naming issues?

Maybe `StorageTypeBlockpool` and `StorageTypeSharedfilesystem`

nb-ohad (changes requested):

services/provider/client/client.go
Madhu-1: Done

PR #1639, opened 2022-04-25
[BZ: 2076468]Added extra rook configs for provider clusters

ezio-auditore:
For a provider storage cluster extra rook ceph configs are required to solve a performance problem.
**[global]
osd_pool_default_pg_autoscale_mode = off
osd_pool_default_pg_num = 128
osd_pool_default_pgp_num = 128**

-  Added Unit tests

### DIFF
--- a/controllers/storagecluster/cephconfig.go
+++ b/controllers/storagecluster/cephconfig.go
@@ -40,6 +40,13 @@ osd_memory_target_cgroup_limit_ratio = 0.8
 `
 )
 
+const providerSCRookConfigs = `
+[global]
+osd_pool_default_pg_autoscale_mode = off
+osd_pool_default_pg_num = 128
+osd_pool_default_pgp_num = 128
+`
+
 // ensureCreated ensures that a ConfigMap resource exists with its Spec in
 // the desired state.
 func (obj *ocsCephConfig) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.StorageCluster) (reconcile.Result, error) {
@@ -128,6 +135,7 @@ func updateRookConfig(defaultRookConfigData string, section string, key string, 
 }
 
 func getRookCephConfig(r *StorageClusterReconciler, sc *ocsv1.StorageCluster) (string, error) {
+	rookConfigData := defaultRookConfigData
 	// configure public network if the cluster is dualstack, but not multus
 	if sc.Spec.Network != nil && sc.Spec.Network.Provider == "" && sc.Spec.Network.DualStack {
 		log.Info("DualStack is enabled, and no alternate network provider is detected")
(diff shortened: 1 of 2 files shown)


controllers/storagecluster/cephconfig.go
```diff
@@ -151,5 +157,17 @@ func getRookCephConfig(r *StorageClusterReconciler, sc *ocsv1.StorageCluster) (s
...
+		var rookConfigData string
+		var err error
+		for key, value := range extraRookConfigs {
+			rookConfigData, err = updateRookConfig(defaultRookConfigData, globalSectionKey, key, value)
+			if err != nil {
+				return "", fmt.Errorf("failed to set extra configuration for rook key: %v value: %v | %v", key, value, err)
+			}
+		}
+		return rookConfigData, nil
+	}
```
nb-ohad, line 171: If we start using dual-stack on AWS we will never get to this code, we need a code change that will allow doing each one separately or both depending on the conditions.

controllers/storagecluster/cephconfig.go
```diff
@@ -151,5 +157,17 @@ func getRookCephConfig(r *StorageClusterReconciler, sc *ocsv1.StorageCluster) (s
 		}
 		return rookConfigData, nil
 	}
+	// additonal rook configs when it is a provider cluster
+	if sc.Spec.AllowRemoteStorageConsumers {
+		var rookConfigData string
+		var err error
+		for key, value := range extraRookConfigs {
+			rookConfigData, err = updateRookConfig(defaultRookConfigData, globalSectionKey, key, value)
```
nb-ohad, line 165: This will only apply one of the configs (the last one) as you are always using the `defaultRookConfigData` as your source

Also, why do we need to use  `updateRookConfig` if the config that we are trying to set is static and global?

nb-ohad (changes requested):

controllers/storagecluster/cephconfig.go
ezio-auditore: Added a function to check for AllowRemoteStorageConsumers and append configs accordingly for both cases

controllers/storagecluster/cephconfig.go
ezio-auditore: Ack. Fixed the issue

controllers/storagecluster/cephconfig.go
```diff
@@ -40,6 +40,12 @@ osd_memory_target_cgroup_limit_ratio = 0.8
 `
 )
 
+var providerSCRookConfigs = map[string]string{
+	"osd_pool_pg_autoscale_mode": "off",
+	"osd_pool_default_pg_num":    "128",
+	"osd_pool_default_pgp_num":   "128",
+}
+
```
umangachapagain, line 48: Don't make it a global var. Keep it within the function where it is used.

controllers/storagecluster/cephconfig.go
```diff
@@ -145,11 +164,17 @@ func getRookCephConfig(r *StorageClusterReconciler, sc *ocsv1.StorageCluster) (s
...
+		rookConfigData, err = updateRookConfig(rookConfigData, globalSectionKey, publicNetworkKey, cidrName)
 		if err != nil {
 			return "", fmt.Errorf("failed to set network configuration for rook: %v", err)
 		}
-		return rookConfigData, nil
 	}
-	return defaultRookConfigData, nil
+	// additonal rook configs when it is a provider cluster
+	if sc.Spec.AllowRemoteStorageConsumers {
+		rookConfigData, err = updateRookConfigFromMap(rookConfigData, globalSectionKey, providerSCRookConfigs)
```
umangachapagain, line 174: You could use `updateRookConfig` directly and avoid this new wrapper.

umangachapagain, line 174: You could use `updateRookConfig` directly and avoid this new wrapper.

controllers/storagecluster/cephconfig.go
```diff
@@ -40,6 +40,12 @@ osd_memory_target_cgroup_limit_ratio = 0.8
 `
 )
 
+var providerSCRookConfigs = map[string]string{
+	"osd_pool_pg_autoscale_mode": "off",
+	"osd_pool_default_pg_num":    "128",
+	"osd_pool_default_pgp_num":   "128",
+}
```
nb-ohad, line 47: This should be a const not a var, same for prev config

ezio-auditore:
The correct key for the autoscale mode was osd_pool_default_pg_autoscale_mode. changed it in the new commit

controllers/storagecluster/cephconfig.go
ezio-auditore: Changed implementation to fetch and merge configs using ini library

PR #1643, opened 2022-04-27
Added Parameters to CephBlockPool and CephFS Data Pools For Provider Clusters

ezio-auditore:


ezio-auditore:
Closing this as this a wrong approach. The default ceph block and cepfs data pools need not be configured with the parameters but only the ones created by consumer.

PR #1645, opened 2022-04-28
Added Parameters to CephFilesystemSubVolumeGroup and CephBlockPool for Provider Clusters

ezio-auditore:


agarwal-mudit:
/hold
Can be merged only when 4.10.2 is open for merging

controllers/storagecluster/generate.go
```diff
@@ -83,3 +90,28 @@ func generateCephReplicatedSpec(initData *ocsv1.StorageCluster, poolType string)
...
+		}
+		deviceCount := deviceSet.Count
+		pgNumValue := DefaultPGnumValue
+		if deviceCount > DefaultDCFor4TB {
+			pgNumValue = ProviderPgNumValue
+		}
+		providerParameters := map[string]string{
+			"pg_autoscale_mode": "off",
+			"pg_num":            fmt.Sprintf("%d", pgNumValue),
+			"pgp_num":           fmt.Sprintf("%d", pgNumValue),
```
nb-ohad, line 112:
```suggestion
			"pg_autoscale_mode": "off",
			"pg_num":            strconv.Itoa(pgNumValue),
			"pgp_num":           strconv.Itoa(pgNumValue),
```

controllers/storagecluster/generate.go
```diff
@@ -83,3 +90,28 @@ func generateCephReplicatedSpec(initData *ocsv1.StorageCluster, poolType string)
...
+// generateCephFSProviderParameters function generates extra parameters required for provider storage clusters
+func generateCephFSProviderParameters(initData *ocsv1.StorageCluster) map[string]string {
+	if initData.Spec.AllowRemoteStorageConsumers {
+		deviceSetList := initData.Spec.StorageDeviceSets
+		var deviceSet ocsv1.StorageDeviceSet
+		for _, ds := range deviceSetList {
+			if ds.Name == StorageDeviceName {
+				deviceSet = ds
+			}
+		}
```
nb-ohad, line 103: You need to handle the case where you don't find the specific storage device set and return an error

controllers/storagecluster/generate.go
```diff
@@ -83,3 +90,28 @@ func generateCephReplicatedSpec(initData *ocsv1.StorageCluster, poolType string)
...
+		deviceCount := deviceSet.Count
+		pgNumValue := DefaultPGnumValue
+		if deviceCount > DefaultDCFor4TB {
+			pgNumValue = ProviderPgNumValue
+		}
```
nb-ohad, line 108: This is not the logic we agreed to! this will only work for 4Tib and 20Tib! what will happen with 8Tib, 16Tib etc?

You should calculate the total num of PG available based on the device count from that get a value for a single unit, the ont that we will use for block pools, and set the CephFS block pool to use 3 such units.

I would suggest moving the calculation of the `unit size` into some other location as you will need to reuse it in the storage consumer controller when creating the block pool per consumer.

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -292,6 +292,9 @@ func (r *StorageConsumerReconciler) reconcileCephBlockPool() error {
 				},
 				Parameters: map[string]string{
 					"target_size_ratio": ".49",
+					"pg_autoscale_mode": "off",
+					"pg_num":            "128",
+					"pgp_num":           "128",
```
nb-ohad, line 297: This is not the behavior we agreed to.  You should calculate the total num of PG available based on the device count from that get a value for a single unit that should be used here.

Please see the above comment about cephfs for a more detailed exaplantion.

nb-ohad (changes requested):

controllers/storagecluster/generate.go
ezio-auditore: Have changed implementation as discussed

controllers/storageconsumer/storageconsumer_controller.go
ezio-auditore: Done

controllers/storagecluster/generate.go
```diff
@@ -2,11 +2,19 @@ package storagecluster
...
+	"math"
+	"strconv"
 
 	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v1"
 	cephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
 )
 
+const (
+	StorageDeviceName = "default"
+	MaxPGPerOSD       = 600
```
nb-ohad, line 14: Is 600 the real number? Also why all it a max when it does not donate a maximum value? 
```suggestion
	PGCountPerOSD       = ????
```

controllers/storagecluster/generate.go
```diff
@@ -83,3 +91,38 @@ func generateCephReplicatedSpec(initData *ocsv1.StorageCluster, poolType string)
...
+// generateCephFSProviderParameters function generates extra parameters required for provider storage clusters
+func generateCephFSProviderParameters(initData *ocsv1.StorageCluster) map[string]string {
+	if initData.Spec.AllowRemoteStorageConsumers {
+		deviceSetList := initData.Spec.StorageDeviceSets
+		var deviceSet *ocsv1.StorageDeviceSet = nil
+		for _, ds := range deviceSetList {
+			if ds.Name == StorageDeviceName {
+				deviceSet = &ds
+			}
+		}
```
nb-ohad, line 104: Unless `deviceSetList` is an array of pointers (which I don't think it is), this form of `for-range` loop will create a in-memory copy of the device set object on each iteration, and changes to the resulting object will not be reflected in the original object (the one in the list). 
Please use indexing into the original array to get the address of the source device set 

```suggestion 
		for i := range deviceSetList {
			ds := &deviceSetList[i] 
			if ds.Name == StorageDeviceName {
				deviceSet = ds
			}
		}
```

controllers/storagecluster/generate.go
```diff
@@ -2,11 +2,19 @@ package storagecluster
...
 	"fmt"
+	"math"
+	"strconv"
 
 	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v1"
 	cephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
 )
 
+const (
+	StorageDeviceName = "default"
```
nb-ohad, line 13: Why do we need this constant? it has no meaning outside searching for the backing pool in `generateCephFSProviderParameters`, and the name does not imply anything about provide mode or Ceph fs (the name lack any kind of context). 

Let's just use the actual string in the search operation instead of a constant. Any future change will be made in a single location in both cases, and for someone reading the search code, all information will be stated locally (no need to jump to the top of the file and back again to the code while reading)

controllers/storagecluster/generate.go
```diff
@@ -83,3 +91,38 @@ func generateCephReplicatedSpec(initData *ocsv1.StorageCluster, poolType string)
...
+				"pg_num":            strconv.Itoa(pgNumValue),
+				"pgp_num":           strconv.Itoa(pgNumValue),
+			}
+			return providerParameters
+		}
+
+	}
+	return nil
+}
+func getUnitSizeOfPgs(osdCount int) int {
```
nb-ohad, line 120:
```suggestion
func getPGBaseUnitSize(osdCount int) int {
```

controllers/storagecluster/generate.go
```diff
@@ -83,3 +91,38 @@ func generateCephReplicatedSpec(initData *ocsv1.StorageCluster, poolType string)
...
+	totalPGCount := MaxPGPerOSD * osdCount
+	totalPGCount -= 16 // taking 16 pgs buffer
+	// 5 (block pool consumers) + 4 (ceph fs underlying block pool)
+	pgUnitSize := float64(totalPGCount / 9)
+	pgUnitSize = math.Min(pgUnitSize, MaxPGUnitSize)
+	nearestPowerOf2 := math.Floor(math.Log(pgUnitSize) / math.Log(2))
+	return int(math.Pow(2, nearestPowerOf2))
```
nb-ohad, line 127:
```suggestion
	// leaving out 16 PGs as a buffer
	totalPGCount := (MaxPGPerOSD * osdCount) - 16 
	// 9 = 5 (block pool consumers) + 4 (ceph fs underlying block pool)
	pgUnitSize := math.Min(float64(totalPGCount / 9),  MaxPGUnitSize)	
	// Round to the nearest power of 2
	pgUnitSize := math.Pow(2, math.Floor(math.Log2(pgUnitSize)))
	return int(pgUnitSize)
```

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -43,6 +45,9 @@ const (
 	StorageConsumerAnnotation     = "ocs.openshift.io.storageconsumer"
 	StorageClaimAnnotation        = "ocs.openshift.io.storageclaim"
 	StorageCephUserTypeAnnotation = "ocs.openshift.io.cephusertype"
+	MaxPGPerOSD                   = 600
+	MaxPGUnitSize                 = 128
```
nb-ohad, line 49: Please do not duplicate these numbers/constants in two distinct places (it will just create a divergence in future updates), have a single place to query the values from.

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -43,6 +45,9 @@ const (
...
+	StorageDeviceName             = "default"
```
nb-ohad, line 50: Same as the comment I added above for the other constant named `StorageDeviceName`

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -708,3 +727,13 @@ func addStorageRelatedAnnotations(obj client.Object, storageConsumerName, storag
...
+
+func getUnitSizeOfPgs(osdCount int) int {
+	totalPGCount := MaxPGPerOSD * osdCount
+	totalPGCount -= 16 // taking 16 pgs buffer
+	// 5 (block pool consumers) + 4 (ceph fs underlying block pool)
+	pgUnitSize := float64(totalPGCount / 9)
+	pgUnitSize = math.Min(pgUnitSize, MaxPGUnitSize)
+	nearestPowerOf2 := math.Floor(math.Log(pgUnitSize) / math.Log(2))
+	return int(math.Pow(2, nearestPowerOf2))
+}
```
nb-ohad, line 739: Don't have a duplicate of this function. Move it to somewhere where all code path can use it (maybe utils)

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -277,6 +282,17 @@ func (r *StorageConsumerReconciler) reconcileCephBlockPool() error {
...
+	var deviceSet *ocsv1.StorageDeviceSet = nil
+	for _, ds := range deviceSetList {
+		if ds.Name == StorageDeviceName {
+			deviceSet = &ds
+		}
+	}
+	pgUnitSize := MaxPGUnitSize
+	if deviceSet != nil {
+		pgUnitSize = getUnitSizeOfPgs(deviceSet.Count)
+	}
```
nb-ohad, line 295: This is wrong. If you cannot find the correct device set in provider mode you should return an error and not continue reconciling as this is an erroneous state

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -277,6 +282,17 @@ func (r *StorageConsumerReconciler) reconcileCephBlockPool() error {
 	failureDomain = storageClusterList.Items[0].Status.FailureDomain
 
 	capacity := r.storageConsumer.Spec.Capacity.String()
+	deviceSetList := storageClusterList.Items[0].Spec.StorageDeviceSets
+	var deviceSet *ocsv1.StorageDeviceSet = nil
+	for _, ds := range deviceSetList {
+		if ds.Name == StorageDeviceName {
+			deviceSet = &ds
+		}
+	}
```
nb-ohad, line 291: Same comment the one above regarding searching for the device set by ref/pointer

nb-ohad (changes requested):

controllers/storagecluster/generate.go
ezio-auditore: Done

controllers/storageconsumer/storageconsumer_controller.go
ezio-auditore: Done added to util/cephutil.go

controllers/storagecluster/generate.go
```diff
@@ -83,3 +85,31 @@ func generateCephReplicatedSpec(initData *ocsv1.StorageCluster, poolType string)
 func generateStorageQuotaName(storageClassName, quotaName string) string {
 	return fmt.Sprintf("ocs-%s-%s", storageClassName, quotaName)
 }
+
+// generateCephFSProviderParameters function generates extra parameters required for provider storage clusters
+func generateCephFSProviderParameters(initData *ocsv1.StorageCluster) (map[string]string, error) {
+	if !initData.Spec.AllowRemoteStorageConsumers {
+		return nil, nil
+	}
```
nb-ohad, line 93: redundant check, you are already checking on the call site. 
Assume that if this function is called then the config is needed

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -277,6 +279,20 @@ func (r *StorageConsumerReconciler) reconcileCephBlockPool() error {
...
+		if ds.Name == "default" {
+			deviceSet = ds
+		}
+	}
+	pgUnitSize := util.MaxPGUnitSize
+	if deviceSet != nil {
+		pgUnitSize = util.GetPGBaseUnitSize(deviceSet.Count)
+	} else {
+		return fmt.Errorf("Could not find  device set named default in Storage cluster")
+	}
```
nb-ohad, line 295:
```suggestion
	if deviceSet == nil {
		return fmt.Errorf("Could not find  device set named default in Storage cluster")
	}
	pgUnitSize := util.GetPGBaseUnitSize(deviceSet.Count)
```

controllers/util/cephutil.go
```diff
@@ -0,0 +1,18 @@
+package util
+
+import "math"
+
+const (
+	PGCountPerOSD = 600
+	MaxPGUnitSize = 128
```
nb-ohad, line 7: These constants are not used outside of this file. Please make them private
```suggestion
	pgCountPerOSD = 600
	maxPGUnitSize = 128
```

controllers/storagecluster/generate.go
nb-ohad: @ezio-auditore Please do not resolve without an answer

ezio-auditore: Sorry, 600pgs is set in the ceph config
https://github.com/red-hat-storage/ocs-operator/blob/c89d00a2f30fe098fcc1534e2c99096ba95e491c/controllers/storagecluster/cephconfig.go#L35

nb-ohad:
Looks good to me

nb-ohad:
/hold cancel
This was discussed and approved for 4.10.2

nb-ohad:
Removing the `bugzilla/invalid-bug` label as it seems like an error. 4.10.1 is already out and this BZ should target 4.10.2

nbalacha (approved):

nb-ohad (approved):

PR #1648, opened 2022-04-29
Moving the defaults for rookceph operator from configmap to csv

malayparida2000:
Going forward the configmap created by ocsinitialization will be
reserved purely for customer overrides (higher precedence), and
the csv will have the env vars for our defaults.

We had discussed this here in odf virtual teams space, 
[https://chat.google.com/room/AAAAREGEba8/JO7SyyGIA0k](https://chat.google.com/room/AAAAREGEba8/JO7SyyGIA0k)

Also the corresponding BZ link for the discussion that happened there.
[https://bugzilla.redhat.com/show_bug.cgi?id=1986016](https://bugzilla.redhat.com/show_bug.cgi?id=1986016)

### DIFF
--- a/controllers/ocsinitialization/ocsinitialization_controller.go
+++ b/controllers/ocsinitialization/ocsinitialization_controller.go
@@ -7,7 +7,6 @@ import (
 	"github.com/go-logr/logr"
 	secv1client "github.com/openshift/client-go/security/clientset/versioned/typed/security/v1"
 	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v1"
-	"github.com/red-hat-storage/ocs-operator/controllers/defaults"
 	"github.com/red-hat-storage/ocs-operator/controllers/util"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -177,24 +176,12 @@ func (r *OCSInitializationReconciler) SetupWithManager(mgr ctrl.Manager) error {
 
 // returns a ConfigMap with default settings for rook-ceph operator
 func newRookCephOperatorConfig(namespace string) *corev1.ConfigMap {
-	var defaultCSIToleration = `
-- key: ` + defaults.NodeTolerationKey + `
-  operator: Equal
-  value: "true"
-  effect: NoSchedule`
-
 	config := &corev1.ConfigMap{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      rookCephOperatorConfigName,
 			Namespace: namespace,
 		},
 	}
-	data := make(map[string]string)
-	data["CSI_PROVISIONER_TOLERATIONS"] = defaultCSIToleration
-	data["CSI_PLUGIN_TOLERATIONS"] = defaultCSIToleration
-	data["CSI_LOG_LEVEL"] = "5"
-	data["CSI_ENABLE_CSIADDONS"] = "true"
-	config.Data = data
 
 	return config
 }
--- a/deploy/bundle/manifests/ocs-operator.clusterserviceversion.yaml
+++ b/deploy/bundle/manifests/ocs-operator.clusterserviceversion.yaml
@@ -2714,6 +2714,10 @@ spec:
                       operator: Equal
                       value: "true"
                       effect: NoSchedule
+                - name: CSI_LOG_LEVEL
+                  value: "5"
+                - name: CSI_ENABLE_CSIADDONS
+                  value: "true"
                 - name: NODE_NAME
                   valueFrom:
                     fieldRef:
--- a/tools/csv-merger/csv-merger.go
+++ b/tools/csv-merger/csv-merger.go
@@ -252,6 +252,14 @@ func unmarshalCSV(filePath string) *csvv1.ClusterServiceVersion {
   effect: NoSchedule`,
 			},
 			{
+				Name:  "CSI_LOG_LEVEL",
+				Value: "5",
+			},
+			{
+				Name:  "CSI_ENABLE_CSIADDONS",
+				Value: "true",
+			},
+			{
 				Name: "NODE_NAME",
 				ValueFrom: &corev1.EnvVarSource{
 					FieldRef: &corev1.ObjectFieldSelector{


controllers/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -328,44 +315,3 @@ func (r *OCSInitializationReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-	return config
-}
-
-func (r *OCSInitializationReconciler) ensureRookCephOperatorConfig(initialData *ocsv1.OCSInitialization) error {
-	rookCephOperatorConfig := &corev1.ConfigMap{}
-	err := r.Client.Get(context.TODO(), types.NamespacedName{Name: rookCephOperatorConfigName, Namespace: initialData.Namespace}, rookCephOperatorConfig)
-	if err != nil {
-		if errors.IsNotFound(err) {
-			// If it does not exist, create a ConfigMap with default settings
-			return r.Client.Create(context.TODO(), newRookCephOperatorConfig(initialData.Namespace))
```
Madhu-1, line 362: How about creating an empty configmap with ownerRef as it helps in cleanup also? or else we need to document steps for both create and delete with a specific name.

If we create the empty configmap manually we just need to add documentation on what to update/how to update this configmap.

agarwal-mudit:
@travisn what about these parameters, are these taken care during upgrade?
https://github.com/red-hat-storage/ocs-operator/blob/73503d9f12015195368c3bb7a6047a587866dfeb/controllers/storagecluster/cephconfig.go#L30

malayparida2000:
> @travisn what about these parameters, are these taken care during upgrade?
> 
> https://github.com/red-hat-storage/ocs-operator/blob/73503d9f12015195368c3bb7a6047a587866dfeb/controllers/storagecluster/cephconfig.go#L30

@agarwal-mudit According to Jose  This is a CephCluster-specific config, not a rook-ceph-operator config, so we should not worry about that one.

controllers/ocsinitialization/ocsinitialization_controller.go
malayparida2000: Okay, Have to discuss on this if we should keep the empty configmap or not. @jarrpa what do you say?

nb-ohad:
@Madhu-1 The ocs-osd-deployer operator is using the configmap here: https://github.com/red-hat-storage/ocs-osd-deployer/blob/main/controllers/managedocs_controller.go#L1381

Does this change is going to affect/break the deployer operator?

Madhu-1:
> @Madhu-1 The ocs-osd-deployer operator is using the configmap here: https://github.com/red-hat-storage/ocs-osd-deployer/blob/main/controllers/managedocs_controller.go#L1381

After discussing with @nb-ohad we have a problem with the upgraded cluster. suppose the configmap contains the `CSI_LOG_LEVEL:5` and we are moving it to the ENV variable. This is fine as a customer also wanted this value to be set to `5` what if we want to change the default value to `4` in the next release? One way is to go over the configmap and change the value from `5` to `4` but that creates a problem as the customer wanted it to be `5` in his cluster.  We need to make a decision here about the expectation.

controllers/ocsinitialization/ocsinitialization_controller.go
jarrpa: No, I say! :laughing: Having documentation to "update" versus "create" is a trivial and unhelpful distinction. Just don't create it at all, and don't remove it if it exists.

Madhu-1:
> > > @Madhu-1 The ocs-osd-deployer operator is using the configmap here: https://github.com/red-hat-storage/ocs-osd-deployer/blob/main/controllers/managedocs_controller.go#L1381
> > 
> > 
> > After discussing with @nb-ohad we have a problem with the upgraded cluster. suppose the configmap contains the `CSI_LOG_LEVEL:5` and we are moving it to the ENV variable. This is fine as a customer also wanted this value to be set to `5` what if we're going to change the default value to `4` in the next release? we can update the ENV but what about the configmap, we don't know if the ocs-operator or the customer added the configuration. We can make a statement old configuration will continue to work as it is and we don't override the configurations and the default will be applied to new clusters in ENV another way is to go over the configmap and change the value from `5` to `4` but that creates a problem as the customer wanted it to be `5` in his cluster. We need to make a decision here about the expectation.
> 
> Yes, this has been the core issue from the beginning.

This might be the right time to make this change as we have minimal changes in configmap.

> 
> > In any case, this PR seems like the right fix. If the desire is to change the csi log level from 5 to 4 for upgraded clusters, how about a separate PR to update this value in the configmap as a one-time operation, then never do it again?
> 
> No. The notion of a "one-time operation" is antithetical to the reconciliation process and requires needless overhead in state tracking to determine if we ever did it, plus further work to either continue supporting it or dropping it for future upgrades.
> 
> In the end, all of this is so freaking minor that I'm fine with just not worrying about it. We never used these for truly significant configuration values, so if a customer detects "oh no, my logging is slightly wrong" then we can just either document or tell them how to fix it. Yes it'll be a slight headache in debugging on our end, but it's not like we weren't already dealing with it before we introduced these ConfgiMap values. Maybe we can include a note on optional maintenance in the release notes (or whatever) that admins are encouraged to delete the ConfigMap in question if they're not actively using it.

Sounds good, Thanks @jarrpa.

controllers/ocsinitialization/ocsinitialization_controller.go
Madhu-1: @jarrpa If we dont create the configmap, ocs-deployer might fail here https://github.com/red-hat-storage/ocs-osd-deployer/blob/main/controllers/managedocs_controller.go#L1386-L1389 we need a fix here or in ocs-deployer. cc @nb-ohad

nb-ohad:
> No, I say! 😆 Having documentation to "update" versus "create" is a trivial and unhelpful distinction. Just don't create it at all, and don't remove it if it exists.

@jarrpa I do not agree with this statement. A reconciliation can be a "create reconciliation" where the reconciler just makes sure of the existence of a resource and set an ownership ref on it (not a controller ref) - regardless of the content. 
This use case exists, mainly when you want to ensure that whatever config is there (regardless of the content) will get wiped out with uninstallation. 

I would prefer for the config map to existing as part of the installation because clients (including the deployer) assume it. No it is true that we can manage an update for our internal client that will make sure to create it if it does not exist, but this will just add additional upgrade code (not in the product, but non the less "upgrade code") for us and inconvenient to whatever clients that introduced automation on top of fresh deployment that adds configuration.  

It also introduces more toil when that part of the automation is uninstalling, as it now needs to make sure to delete the configmap.

Madhu-1:
@nb-ohad good to remove hold on this PR?

Madhu-1:
> @nb-ohad good to remove hold on this PR?

@nb-ohad ping.

malayparida2000:
@nb-ohad Can you please take another look here, This is now a blocker for [RHSTOR-2511](https://issues.redhat.com/browse/RHSTOR-2511). cc @agarwal-mudit @humblec @Madhu-1

jarrpa (approved):
/lgtm

jarrpa:
@nb-ohad @Madhu-1 I talked to @malayparida2000 about this, and there was a suggestion to keep a list of the "previous" default values of the ConfigMap in memory, and if we find those values are still the same then we can remove them from the ConfigMap so that the environment variables in the Deployment become the new defaults. I suggested we could fire an Event or similar to inform the admin that we modified/deleted the ConfigMap and that if they don't like this they can recreate it with the old values (we should even be able to provide the exact YAML to use). I'll leave this PR on hold for a few more days, but if no compelling arguments are raised against it I'll just merge it and we can implement further changes in future PRs.

Madhu-1:
> @nb-ohad @Madhu-1 I talked to @malayparida2000 about this, and there was a suggestion to keep a list of the "previous" default values of the ConfigMap in memory, and if we find those values are still the same then we can remove them from the ConfigMap so that the environment variables in the Deployment become the new defaults. I suggested we could fire an Event or similar to inform the admin that we modified/deleted the ConfigMap and that if they don't like this they can recreate it with the old values (we should even be able to provide the exact YAML to use). I'll leave this PR on hold for a few more days, but if no compelling arguments are raised against it I'll just merge it and we can implement further changes in future PRs.

sounds good to me.

humblec:
@malayparida2000 jfyi, we are adding one more control for sidecar logs instead of just having one CSI_LOG_LEVEL . The details are here https://github.com/ceph/ceph-csi/pull/3199 and in rook PR linked there.. so we need to have this configuation refreshed either in this PR or as a follow up PR.

malayparida2000:
@humblec Let’s keep this PR limited to just for the change in implementation, The new values for sidecar logs should be another separate PR in my opinion.

humblec:
> @humblec Let’s keep this PR limited to just for the change in implementation, The new values for sidecar logs should be another separate PR in my opinion.

Yeah,, that should be fine @malayparida2000 :+1:

PR #1649, opened 2022-05-02
Call the Rook API for SCC generation

travisn:
Call the Rook api for SCCs. A rook api is available for generating the correct scc expected for rook components. Update to the new api so we don't have the ocs operator owning an outdated scc spec. The `rook-ceph-rgw` service account was missing in the scc, which is causing RGW to fail to start in 4.11 as seen in this issue: https://bugzilla.redhat.com/show_bug.cgi?id=2075581


jarrpa (approved):
/lgtm

travisn:
The update to reference rook v1.9.2 was separated to #1652, this PR is just focusing on the scc reference.

jarrpa (approved):
/lgtm

jarrpa (approved):
/lgtm

PR #1654, opened 2022-05-05
add status.Phase check for GetStorageClassClaimConfig

Madhu-1:
check the status.Phase for GetStorageClassClaimConfig based on the Phase return success or failure response.

/assign @umangachapagain 
/assign @nb-ohad

### DIFF
--- a/services/provider/server/server.go
+++ b/services/provider/server/server.go
@@ -530,9 +530,22 @@ func (s *OCSProviderServer) GetStorageClassClaimConfig(ctx context.Context, req 
 		return nil, status.Errorf(codes.Internal, "failed to get storage class claim config %q for %q. %v", req.StorageClassClaimName, req.StorageConsumerUUID, err)
 	}
 
-	// TODO: add status.Phase validation here to check claim is in expected
-	// state or not before extracting the ceph resources from the StorageClassClaim CR status
-
+	// Verify Status.Phase
+	msg := fmt.Sprintf("storage class claim %q for %q is in %q phase", req.StorageClassClaimName, req.StorageConsumerUUID, storageClassClaim.Status.Phase)
+	if storageClassClaim.Status.Phase != ocsv1alpha1.StorageClassClaimReady {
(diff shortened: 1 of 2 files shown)


services/provider/server/server.go
```diff
@@ -530,103 +530,112 @@ func (s *OCSProviderServer) GetStorageClassClaimConfig(ctx context.Context, req
...
+	// Verify Status.Phase
+	msg := fmt.Sprintf("storage class claim %q for %q is in %q phase", req.StorageClassClaimName, req.StorageConsumerUUID, storageClassClaim.Status.Phase)
+	switch storageClassClaim.Status.Phase {
+	case ocsv1alpha1.StorageClassClaimFailed:
+		return nil, status.Error(codes.Internal, msg)
+	case ocsv1alpha1.StorageClassClaimInitializing:
+		return nil, status.Error(codes.Unavailable, msg)
+	case ocsv1alpha1.StorageClassClaimCreating:
+		return nil, status.Error(codes.Unavailable, msg)
+	case ocsv1alpha1.StorageClassClaimReady:
```
nb-ohad, line 542: This nesting indent the entirety of the code in this function and produce a switch case inside a switch case. 

```suggestion
        if storageClassClaim.Status.Phase != ocsv1alpha1.StorageClassClaimReady {
		switch storageClassClaim.Status.Phase {
		case ocsv1alpha1.StorageClassClaimInitializing:
			return nil, status.Error(codes.Unavailable, msg)
		case ocsv1alpha1.StorageClassClaimCreating:
			return nil, status.Error(codes.Unavailable, msg)
		default:
			return nil, status.Error(codes.Internal, msg)
	}
	
	// All of the main code of the function goes here without additional indentation 
```

nb-ohad (changes requested):

services/provider/server/server.go
Madhu-1: sounds better :+1:

PR #1655, opened 2022-05-05
Make use of SHA256 digests explicit

mmgaggle:
Some platforms do not default to using SHA256 for digests, instead
using MD5. On one of those platforms, a token will fail verification
at the provider since it explicitly verifies using SHA256.


jarrpa (approved):
/lgtm

PR #1658, opened 2022-05-10
Fix bugs found in the testing

Madhu-1:
* Improve logging in storageclass claim to help to debug in error cases.
* The storageType should be of type string sharedfilesystem or blockpool when creating the storageclassclaim CR it's not always the string format of the auto generated RPC names.
* Added RBAC role for provider api-server for the storageclassclaim CR.
* The fsName is the key in the storageclass parameters matching the key to sending back in GetStorageClassClaimConfig to create the storageclass
* Get operation on the consumer will not have the TypeMeta objects set causing the problem of setting the ownerRef, use the helper function to set the ownerRef

/assign @umangachapagain 
/assign @nb-ohad 
/assign @iamniting

### DIFF
--- a/services/provider/server/server.go
+++ b/services/provider/server/server.go
@@ -500,6 +500,7 @@ func (s *OCSProviderServer) FulfillStorageClassClaim(ctx context.Context, req *p
 	err = s.storageClassClaimManager.Create(ctx, consumerObj, req.StorageClassClaimName, req.StorageType.String(), req.EncryptionMethod)
 	if err != nil {
 		errMsg := fmt.Sprintf("failed to fulfill storage class claim for %q. %v", req.StorageConsumerUUID, err)
+		klog.Error(errMsg)
 		if kerrors.IsAlreadyExists(err) {
 			return nil, status.Error(codes.AlreadyExists, errMsg)
 		}
@@ -514,7 +515,9 @@ func (s *OCSProviderServer) FulfillStorageClassClaim(ctx context.Context, req *p
 func (s *OCSProviderServer) RevokeStorageClassClaim(ctx context.Context, req *pb.RevokeStorageClassClaimRequest) (*pb.RevokeStorageClassClaimResponse, error) {
 	err := s.storageClassClaimManager.Delete(ctx, req.StorageConsumerUUID, req.StorageClassClaimName)
 	if err != nil {
-		return nil, status.Errorf(codes.Internal, "failed to revoke storage class claim %q for %q. %v", req.StorageClassClaimName, req.StorageConsumerUUID, err)
+		errMsg := fmt.Sprintf("failed to revoke storage class claim %q for %q. %v", req.StorageClassClaimName, req.StorageConsumerUUID, err)
+		klog.Error(errMsg)
+		return nil, status.Errorf(codes.Internal, errMsg)
 	}
 
 	return &pb.RevokeStorageClassClaimResponse{}, nil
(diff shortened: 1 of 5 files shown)


services/provider/server/server.go
```diff
@@ -497,9 +497,15 @@ func (s *OCSProviderServer) FulfillStorageClassClaim(ctx context.Context, req *p
 
 	klog.Infof("Found StorageConsumer %q (%q)", consumerObj.Name, req.StorageConsumerUUID)
 
-	err = s.storageClassClaimManager.Create(ctx, consumerObj, req.StorageClassClaimName, req.StorageType.String(), req.EncryptionMethod)
+	storageType := "sharedfilesystem"
+	if req.StorageType == pb.FulfillStorageClassClaimRequest_BLOCKPOOL {
+		storageType = "blockpool"
+	}
```
nb-ohad, line 503: Let's cover each option individually and cover the case of unknown values

```suggestion
	var storageType string
	switch req.StorageType {
	case pb.FulfillStorageClassClaimRequest_BLOCKPOOL:
		storageType = "blockpool"
	case FulfillStorageClassClaimRequest_SHAREDFILESYSTEM:
			storageType := "sharedfilesystem"
	default:
		return fmt.ErrorF("Encountered an unknown stroage type, %s", storageType) 
	
```

nb-ohad (changes requested):

services/provider/server/storageclaim.go
```diff
@@ -79,7 +71,12 @@ func (s *storageClassClaimManager) Create(ctx context.Context, consumer *ocsv1al
 		},
 	}
 
-	err := s.client.Create(ctx, storageClassClaimObj)
+	err := controllerutil.SetOwnerReference(consumer, storageClassClaimObj, s.client.Scheme())
```
nb-ohad, line 74: This will not set the `BlockOwnerDeletion` flag. only SetControllerReference will but in that case we have an issue of a wrong controller set on the owner, Please revert

nb-ohad (changes requested):

services/provider/server/server.go
Madhu-1: Done.

services/provider/server/storageclaim.go
Madhu-1: The manual setting of ownRef was not working. so we added this code when we found issue during Create operation. i brought back the functionality of BlockOwnerDeletion. as this is already testing code can we stick to this one, if we need to go back to the previous version need to check what is going wrong in setting ownRef and need more testing.

Madhu-1: As discussed, changes are done PTAL

services/provider/server/storageclaim.go
```diff
@@ -78,6 +70,15 @@ func (s *storageClassClaimManager) Create(ctx context.Context, consumer *ocsv1al
 			EncryptionMethod: encryptionMethod,
 		},
 	}
+	storageClassClaimObj.SetOwnerReferences([]metav1.OwnerReference{
```
Madhu-1, line 73: @nb-ohad am not sure it will work SetOwnerReferences is an inline function it just sets the ownerRef on object . this is same as before https://github.com/kubernetes/apimachinery/blob/v0.24.0/pkg/apis/meta/v1/meta.go#L172-L174

nb-ohad: I think it didn't work because of Kind and APIVersion issues, with the current change we are taking the information form the Schema and not the object (using GVK), this is the exact same logic that the controller utils function uses and should work

PR #1663, opened 2022-05-11
ocs-to-ocs: revert disabling cephfs csi driver

umangachapagain:
In case of OCS to OCS, consumer OCS cluster has a default
CephFS based StorageClass. So, to allow PV creation using
this StorageClass CephFS CSI driver should also be enabled
by default.

This PR skips the toggle added in #701.

### DIFF
--- a/controllers/storagecluster/external_resources.go
+++ b/controllers/storagecluster/external_resources.go
@@ -561,9 +561,13 @@ func (r *StorageClusterReconciler) createExternalStorageClusterResources(instanc
 		r.Log.Error(err, "Failed to create needed StorageClasses.")
 		return err
 	}
-	if err = r.setRookCSICephFS(enableRookCSICephFS, instance); err != nil {
-		r.Log.Error(err, "Failed to set RookEnableCephFSCSIKey to EnableRookCSICephFS.", "RookEnableCephFSCSIKey", rookEnableCephFSCSIKey, "EnableRookCSICephFS", enableRookCSICephFS)
-		return err
+	// We do not want to disable CephFS csi driver in consumer mode since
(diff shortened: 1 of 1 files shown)


controllers/storagecluster/external_resources.go
```diff
@@ -561,9 +561,13 @@ func (r *StorageClusterReconciler) createExternalStorageClusterResources(instanc
...
-	if err = r.setRookCSICephFS(enableRookCSICephFS, instance); err != nil {
-		r.Log.Error(err, "Failed to set RookEnableCephFSCSIKey to EnableRookCSICephFS.", "RookEnableCephFSCSIKey", rookEnableCephFSCSIKey, "EnableRookCSICephFS", enableRookCSICephFS)
-		return err
+	// We do not want to disable CephFS csi driver in consumer mode since
+	// CephFS storageclass is available by default.
+	if !IsOCSConsumerMode(instance) {
+		if err = r.setRookCSICephFS(enableRookCSICephFS, instance); err != nil {
+			r.Log.Error(err, "Failed to set RookEnableCephFSCSIKey to EnableRookCSICephFS.", "RookEnableCephFSCSIKey", rookEnableCephFSCSIKey, "EnableRookCSICephFS", enableRookCSICephFS)
+			return err
+		}
```
aruniiird, line 570: What if we just enable the flag when we are in `ConsumerMode`? If you see the function, `setRookCSICephFS()`, it won't do any updation if the flag is already true, thus the code look much cleaner.

```suggestion
	if IsOCSConsumerMode(instance) {
	       enableRookCSICephFS = true
	}
```

PR #1666, opened 2022-05-11
Fix storageclassname problem

Madhu-1:
If the storageclass already exits and is backed by another storage-class claim in a different namespace block the new request for the same name in a different namespace.
    
### DIFF
--- a/controllers/storageclassclaim/storageclassclaim_controller.go
+++ b/controllers/storageclassclaim/storageclassclaim_controller.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"reflect"
 	"strconv"
+	"strings"
 	"time"
 
 	"k8s.io/klog/v2"
@@ -199,6 +200,19 @@ func (r *StorageClassClaimReconciler) reconcileConsumerPhases() (reconcile.Resul
 			},
 		}
 		if err = r.get(existing); err == nil {
+			annotation, found := existing.GetAnnotations()[StorageClassClaimAnnotation]
+			if !found {
+				r.log.Error(fmt.Errorf("storageClass %q does not belong to any storageclass claim", existing.Name), "StorageClassClaim validation failed.")
+				r.storageClassClaim.Status.Phase = v1alpha1.StorageClassClaimFailed
+				return reconcile.Result{}, nil
+			}
+			claimNamespacedName := strings.Split(annotation, "/")
+			if claimNamespacedName[0] != r.storageClassClaim.Namespace {
+				r.log.Error(fmt.Errorf("storageClass belongs to %q storageclass claim in %q namespace", r.storageClassClaim.Name, r.storageClassClaim.Namespace), "StorageClassClaim validation failed.")
+				r.storageClassClaim.Status.Phase = v1alpha1.StorageClassClaimFailed
+				return reconcile.Result{}, nil
+			}
+
 			sccType := r.storageClassClaim.Spec.Type
 			sccEncryptionMethod := r.storageClassClaim.Spec.EncryptionMethod
 			_, scIsFSType := existing.Parameters["fsName"]
--- a/controllers/storageclassclaim/storageclassclaim_controller.go
+++ b/controllers/storageclassclaim/storageclassclaim_controller.go
@@ -503,8 +503,7 @@ func (r *StorageClassClaimReconciler) getCephFSStorageClass(data map[string]stri
 	allowVolumeExpansion := true
 	storageClass := &storagev1.StorageClass{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      r.storageClassClaim.Name,
-			Namespace: r.storageClassClaim.Namespace,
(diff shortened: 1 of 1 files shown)


controllers/storageclassclaim/storageclassclaim_controller.go
```diff
@@ -196,6 +197,12 @@ func (r *StorageClassClaimReconciler) reconcileConsumerPhases() (reconcile.Resul
 			},
 		}
 		if err = r.get(existing); err == nil {
+			storageClassNamespace := existing.Annotations[storageClassNamespaceAnnotation]
```
Madhu-1, line 200: @nb-ohad can this cause problem for the upgrades as the annotation is not set in the SC?

controllers/storageclassclaim/storageclassclaim_controller.go
```diff
@@ -199,6 +200,18 @@ func (r *StorageClassClaimReconciler) reconcileConsumerPhases() (reconcile.Resul
...
+				if claimNamespacedName[0] != r.storageClassClaim.Namespace {
+					r.log.Error(fmt.Errorf("storageClass belongs to %q storageclass claim in %q namespace", r.storageClassClaim.Name, r.storageClassClaim.Namespace), "StorageClassClaim validation failed.")
+					r.storageClassClaim.Status.Phase = v1alpha1.StorageClassClaimFailed
+					err = r.update(r.storageClassClaim)
+					if err != nil {
+						r.log.Error(err, "failed to update StorageClassClaim", "storageClassClaim", klog.KRef(r.storageClassClaim.Namespace, r.storageClassClaim.Name))
+					}
+					return reconcile.Result{}, err
+				}
+			}
```
nb-ohad, line 214: Shouldn't we also address the case where the annotation does not exist?

controllers/storageclassclaim/storageclassclaim_controller.go
```diff
@@ -208,7 +221,11 @@ func (r *StorageClassClaimReconciler) reconcileConsumerPhases() (reconcile.Resul
 				r.log.Error(fmt.Errorf("storageClassClaim is not compatible with existing StorageClass"),
 					"StorageClassClaim validation failed.")
 				r.storageClassClaim.Status.Phase = v1alpha1.StorageClassClaimFailed
-				return reconcile.Result{}, nil
+				err = r.update(r.storageClassClaim)
+				if err != nil {
+					r.log.Error(err, "failed to update StorageClassClaim", "storageClassClaim", klog.KRef(r.storageClassClaim.Namespace, r.storageClassClaim.Name))
+				}
+				return reconcile.Result{}, err
```
nb-ohad, line 228: Why are we updating the resource we take as input? If the intent was to update the status then just exist (with or without an error) the reconcile function. we have code that ensures that the status will be updated.

controllers/storageclassclaim/storageclassclaim_controller.go
```diff
@@ -199,6 +200,18 @@ func (r *StorageClassClaimReconciler) reconcileConsumerPhases() (reconcile.Resul
...
 		if err = r.get(existing); err == nil {
+			if annotation, found := existing.GetAnnotations()[StorageClassClaimAnnotation]; found {
+				claimNamespacedName := strings.Split(annotation, "/")
+				if claimNamespacedName[0] != r.storageClassClaim.Namespace {
+					r.log.Error(fmt.Errorf("storageClass belongs to %q storageclass claim in %q namespace", r.storageClassClaim.Name, r.storageClassClaim.Namespace), "StorageClassClaim validation failed.")
+					r.storageClassClaim.Status.Phase = v1alpha1.StorageClassClaimFailed
+					err = r.update(r.storageClassClaim)
+					if err != nil {
+						r.log.Error(err, "failed to update StorageClassClaim", "storageClassClaim", klog.KRef(r.storageClassClaim.Namespace, r.storageClassClaim.Name))
+					}
```
nb-ohad, line 211: Why are we updating the resource we take as input? If the intent was to update the status then just exist (with or without an error) the reconcile function. we have code that ensures that the status will be updated.

nb-ohad (changes requested):

controllers/storageclassclaim/storageclassclaim_controller.go
Madhu-1: missed that part. removed status update.

controllers/storageclassclaim/storageclassclaim_controller.go
Madhu-1: what we should do in this case?

controllers/storageclassclaim/storageclassclaim_controller.go
Madhu-1: removed update call

controllers/storageclassclaim/storageclassclaim_controller.go
nb-ohad: I would say we should fail the validation

PR #1673, opened 2022-05-11
update rook-ceph-mon endpoints cm

parth-gr:
Currently the server was overriding the mon endpoints cm
with only 1 mon-IP,
Updated it to add all the 3 mon IPs

Signed-off-by: parth-gr <paarora@redhat.com>

### DIFF
--- a/services/provider/server/server.go
+++ b/services/provider/server/server.go
@@ -268,9 +268,7 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
 		return nil, fmt.Errorf("failed to get %s configMap. %v", monConfigMap, err)
 	}
 
-	// Get address of first mon from the monConfigMap configmap
-	cmData := strings.Split(configmap.Data["data"], ",")
-	if len(cmData) == 0 {
+	if configmap.Data["data"] == "" {
 		return nil, fmt.Errorf("configmap %s data is empty", monConfigMap)
 	}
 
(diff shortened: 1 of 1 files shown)


services/provider/server/server.go
```diff
@@ -278,7 +278,7 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
 		Name: monConfigMap,
 		Kind: "ConfigMap",
 		Data: mustMarshal(map[string]string{
-			"data":     cmData[0], // Address of first mon
+			"data":     strings.Join(cmData, ", "), // IP Address of all mon's
```
travisn, line 281: Not sure rook can handle the space. Here's a sample from a test cluster. And is the mon name included with the `cmData`?
```
data: a=10.110.186.73:6789,b=10.105.227.50:6789,c=10.104.4.237:6789
```

```suggestion
			"data":     strings.Join(cmData, ","), // IP Address of all mon's
```

services/provider/server/server.go
```diff
@@ -278,7 +278,7 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
 		Name: monConfigMap,
 		Kind: "ConfigMap",
 		Data: mustMarshal(map[string]string{
-			"data":     cmData[0], // Address of first mon
+			"data":     strings.Join(cmData, ","), // IP Address of all mon's
```
nb-ohad, line 281: No need to join information that was split just a few lines above
```suggestion
			"data":    configmap.Data, // IP Address of all mon's
```

You can also remove the split above and change the condition to check empty string instead of len

nb-ohad (changes requested):

services/provider/server/server.go
parth-gr: Done

nb-ohad:
Added a hold to give @iamniting the opportunity to review as well

PR #1676, opened 2022-05-12
fix resource watch problem

Madhu-1:
As we are not setting the Controller OwnerReference on the ceph resources we are creating as part of the StorageClassClaim, we need to set IsController to false to get Reconcile Request of StorageClassClaim for the owned ceph resources updates.

/assign @nb-ohad


controllers/storageclassclaim/storageclassclaim_controller.go
```diff
@@ -321,6 +344,10 @@ func (r *StorageClassClaimReconciler) reconcileConsumerPhases() (reconcile.Resul
 				} else if resource.Name == "ceph-rbd" {
 					storageClass = r.getCephRBDStorageClass(data)
 				}
+				if storageClass.Annotations == nil {
+					storageClass.Annotations = map[string]string{}
+				}
+				storageClass.Annotations[StorageClassClaimNamespaceAnnotation] = r.storageClassClaim.Namespace
```
nb-ohad, line 350: Please use Get/SetAnnotations to read/update the annotations:
```suggestion
				if annotations := storageClass.GetAnnotations(); annotations == nil {
					storageClass.SetAnnotations(map[string]string{
						StorageClassClaimNamespaceAnnotation: r.storageClassClaim.Namespace
					})
				} else {
					annotations[StorageClassClaimNamespaceAnnotation] = r.storageClassClaim.Namespace
				}
```

controllers/storageclassclaim/storageclassclaim_controller.go
```diff
@@ -66,8 +69,9 @@ type StorageClassClaimReconciler struct {
 }
 
 const (
-	StorageClassClaimFinalizer  = "storageclassclaim.ocs.openshift.io"
-	StorageClassClaimAnnotation = "ocs.openshift.io.storagesclassclaim"
+	StorageClassClaimFinalizer           = "storageclassclaim.ocs.openshift.io"
+	StorageClassClaimAnnotation          = "ocs.openshift.io.storagesclassclaim"
+	StorageClassClaimNamespaceAnnotation = "ocs.openshift.io.storagesclassclaimnamespacename"
```
nb-ohad, line 74: A claim is a namespaced resource so an annotation that only specifies the name `StorageClassClaimAnnotation` is meaningless. My suggestion is not to have a second annotation for the namespace but change the value of the claim annotation to be `<namespace>/<name>` (which I believe is a common way to represent a namespace name in a string format)

This change will be beneficial for all of the resources that are annotated with this annotation

controllers/storageclassclaim/storageclassclaim_controller.go
```diff
@@ -151,14 +155,33 @@ func (r *StorageClassClaimReconciler) Reconcile(ctx context.Context, request rec
...
+				return []reconcile.Request{{
+					NamespacedName: types.NamespacedName{
+						Name:      obj.GetName(),
+						Namespace: namespace,
+					},
+				}}
+
+			}
+			return []reconcile.Request{}
+		})
```
nb-ohad, line 171: Although we agreed to match the name of the claim with the name of the storage class, this conversion might change in the future and it will be better to avoid writing code based on this conversion. As suggested in the last comment, let's try to capture the namespace name of the claim in the annotation. 

```suggestion
	enqueueStorageConsumerRequest := handler.EnqueueRequestsFromMapFunc(
		func(obj client.Object) []reconcile.Request {
			if annotation, found := annotation[StorageClassClaimAnnotation]; found {
				parts := strings.split(annotation, "/")
				return []reconcile.Request{{
					NamespacedName: types.NamespacedName{
						Name:      parts[0]
						Namespace: parts[1],
					},
				}}
			} else {
				return []reconcile.Request{}
			}
		},
	)
```

controllers/storageclassclaim/storageclassclaim_controller.go
```diff
@@ -151,14 +155,33 @@ func (r *StorageClassClaimReconciler) Reconcile(ctx context.Context, request rec
 }
 
 func (r *StorageClassClaimReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	enqueueStorageConsumerRequest := handler.EnqueueRequestsFromMapFunc(
```
nb-ohad, line 158: Name here is very generic, can we change that to `enqueueStorageConsumerRequestFromAnnotation`

nb-ohad (changes requested):

controllers/storageclassclaim/storageclassclaim_controller.go
Madhu-1: I chosed not to do it because of the length limitation we can it. If we follow this one we will hit name length limitation if the claim name and the namespace name are too long?

Madhu-1: okay, annotation doesn't have a length limit but labels do. updating as per suggestion but I haven't seen this format anywhere.

controllers/storageclassclaim/storageclassclaim_controller.go
Madhu-1: Done

controllers/storageclassclaim/storageclassclaim_controller.go
nb-ohad: Yes, I was just going to mention that annotation does not have length limitation

PR #1682, opened 2022-05-19
metrics-exporter: add checks to prevent crash

umangachapagain:
+ Check length after splitting string to ensure we have the expected data and we don't end up with IndexOutOfBound error.
+ Remove primary/non-primary check on image to ensure we don't miss negative states.

### DIFF
--- a/metrics/internal/collectors/rbd-mirror.go
+++ b/metrics/internal/collectors/rbd-mirror.go
@@ -119,53 +119,53 @@ func (c *RBDMirrorCollector) Collect(ch chan<- prometheus.Metric) {
 			}
 		}
 		for _, image := range poolData.MirrorStatus.Images {
-			if image.Description == "local image is primary" {
-				for _, site := range image.PeerSites {
-					// site.State can have values like up+stopped, up+error etc.
-					state := strings.SplitN(site.State, "+", 2)
-					if len(state) != 2 {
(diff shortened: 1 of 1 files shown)


metrics/internal/collectors/rbd-mirror.go
```diff
@@ -163,29 +164,40 @@ func (c *RBDMirrorCollector) Collect(ch chan<- prometheus.Metric) {
...
-					err := json.Unmarshal([]byte(description), &desc)
-					if err != nil {
-						klog.Errorf("Failed to unmarshal description of image %q from site %q: %v", image.Name, site.SiteName, err)
-						return
+				}
+				if image.Description == "local image is primary" {
+					// site.Description can have values like "replaying,{...}" "split-brain" etc.
+					siteDescription := strings.SplitN(site.Description, ", ", 2)
+					if len(siteDescription) != 2 {
+						klog.Errorf("Unexpected mirror peer site description %q of image %q to site %q.", site.Description, image.Name, site.SiteName)
```
Madhu-1, line 172: in this case, we can continue? we can remove else check if we continue here

Madhu-1 (approved):
left a question, change LGTM

metrics/internal/collectors/rbd-mirror.go
umangachapagain: fixed.

Madhu-1 (approved):

PR #1697, opened 2022-05-26
Fix metric exporter

malayparida2000:
Currently the ocs-metrics-exporter-image is set to use the ocs-image.
Which is resulting in the ocs-metrics-exporter deployment failing, which
in turn results in failing of the ocs operator csv. So changing it to use the correct image


iamniting (approved):
/lgtm

jarrpa (approved):
/lgtm

PR #1703, opened 2022-06-06
odf to odf: start using LoadBalancer type for ocs-provider-server svc

iamniting:
We were using service of type NodePort which is prune to failures
whenever a used node goes down or removed from a OCP cluster.

Start using the service of type LoadBalancer which creates a external IP
which can be used via consumers.

Bug: https://bugzilla.redhat.com/show_bug.cgi?id=2060487

### DIFF
--- a/controllers/storagecluster/provider_server.go
+++ b/controllers/storagecluster/provider_server.go
@@ -5,7 +5,7 @@ import (
 	"fmt"
 	"math/rand"
 	"os"
-	"sort"
+	"time"
 
 	"go.uber.org/multierr"
 	appsv1 "k8s.io/api/apps/v1"
@@ -26,8 +26,7 @@ const (
 	ocsProviderServerName  = "ocs-provider-server"
 	providerAPIServerImage = "PROVIDER_API_SERVER_IMAGE"
 
-	ocsProviderServicePort     = int32(50051)
-	ocsProviderServiceNodePort = int32(31659)
+	ocsProviderServicePort = int32(50051)
 
 	ocsProviderCertSecretName = ocsProviderServerName + "-cert"
 )
@@ -46,21 +45,23 @@ func (o *ocsProviderServer) ensureCreated(r *StorageClusterReconciler, instance 
 
 	r.Log.Info("Spec.AllowRemoteStorageConsumers is enabled. Creating Provider API resources")
 
-	var finalErr error
-
-	for _, f := range []func(*StorageClusterReconciler, *ocsv1.StorageCluster) error{
-		o.createSecret,
-		o.createService,
-		o.createDeployment,
-	} {
-		err := f(r, instance)
-		if err != nil {
-			multierr.AppendInto(&finalErr, err)
-			continue
-		}
+	if err := o.createSecret(r, instance); err != nil {
+		return reconcile.Result{}, err
 	}
 
-	return reconcile.Result{}, finalErr
+	if res, err := o.createService(r, instance); err != nil {
+		return reconcile.Result{}, err
+	} else if !res.IsZero() {
+		return res, nil
+	}
+
+	if res, err := o.createDeployment(r, instance); err != nil {
+		return reconcile.Result{}, err
+	} else if !res.IsZero() {
+		return res, nil
+	}
+
+	return reconcile.Result{}, nil
 }
 
 func (o *ocsProviderServer) ensureDeleted(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) (reconcile.Result, error) {
@@ -96,7 +97,7 @@ func (o *ocsProviderServer) ensureDeleted(r *StorageClusterReconciler, instance 
 	return reconcile.Result{}, finalErr
 }
 
-func (o *ocsProviderServer) createDeployment(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) error {
+func (o *ocsProviderServer) createDeployment(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) (reconcile.Result, error) {
 
 	var finalErr error
 
@@ -107,7 +108,7 @@ func (o *ocsProviderServer) createDeployment(r *StorageClusterReconciler, instan
 	}
 
 	if finalErr != nil {
-		return finalErr
+		return reconcile.Result{}, finalErr
 	}
 
 	desiredDeployment := GetProviderAPIServerDeployment(instance)
@@ -127,20 +128,21 @@ func (o *ocsProviderServer) createDeployment(r *StorageClusterReconciler, instan
 	)
 	if err != nil && !errors.IsAlreadyExists(err) {
 		r.Log.Error(err, "Failed to create/update deployment", "Name", desiredDeployment.Name)
-		return err
+		return reconcile.Result{}, err
 	}
 
 	err = o.ensureDeploymentReplica(actualDeployment, desiredDeployment)
 	if err != nil {
-		r.Log.Error(err, "Deployment is not ready", "Name", desiredDeployment.Name)
-		return err
+		r.recorder.ReportIfNotPresent(instance, corev1.EventTypeNormal, "Waiting", "Waiting for Deployment to become ready "+desiredDeployment.Name)
+		r.Log.Info("Waiting for Deployment to become ready", "Name", desiredDeployment.Name)
+		return reconcile.Result{RequeueAfter: time.Second * 5}, nil
 	}
 
 	r.Log.Info("Deployment is running as desired")
-	return nil
+	return reconcile.Result{}, nil
 }
 
-func (o *ocsProviderServer) createService(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) error {
+func (o *ocsProviderServer) createService(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) (reconcile.Result, error) {
 
 	desiredService := GetProviderAPIServerService(instance)
 	actualService := &corev1.Service{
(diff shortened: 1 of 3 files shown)


controllers/storagecluster/provider_server.go
```diff
@@ -26,8 +25,7 @@ const (
 	ocsProviderServerName  = "ocs-provider-server"
 	providerAPIServerImage = "PROVIDER_API_SERVER_IMAGE"
 
-	ocsProviderServicePort     = int32(50051)
-	ocsProviderServiceNodePort = int32(31659)
+	ocsProviderServicePort = int32(50051)
```
oritwas, line 28: Will it be possible to make the port configurable i.e part of the CRD?

Madhu-1: If I understand correctly @oritwas is asking Can this port be configurable like the user can provide a custom port when deploying OCS in provider mode? like in CSV/CRD or POD ENV etc?

Madhu-1: not sure POD ENV is useful

iamniting: We can read it from the storagecluster CR if required. I am not in favor of the ENV variables as we have to populate that value either via CSV on the build time or via configmap.

controllers/storagecluster/provider_server.go
```diff
@@ -175,19 +173,22 @@ func (o *ocsProviderServer) createService(r *StorageClusterReconciler, instance
...
+		err = fmt.Errorf("Service %s Status.LoadBalancer.Ingress is empty", actualService.Name)
+		r.Log.Error(err, "Service status is not set yet", "Status", actualService.Status)
 		return err
 	}
 
-	if len(nodeAddresses) == 0 {
-		err = fmt.Errorf("Did not found any worker node addresses")
-		r.Log.Error(err, "Worker nodes count is zero")
+	if actualService.Status.LoadBalancer.Ingress[0].IP != "" {
+		instance.Status.StorageProviderEndpoint = actualService.Status.LoadBalancer.Ingress[0].IP + ":" + fmt.Sprint(ocsProviderServicePort)
```
nb-ohad, line 183:
```suggestion
		instance.Status.StorageProviderEndpoint = fmt.Sprint(%s:$d", actualService.Status.LoadBalancer.Ingress[0].IP, ocsProviderServicePort)
```

controllers/storagecluster/provider_server.go
```diff
@@ -175,19 +173,22 @@ func (o *ocsProviderServer) createService(r *StorageClusterReconciler, instance
...
+	} else if actualService.Status.LoadBalancer.Ingress[0].Hostname != "" {
+		instance.Status.StorageProviderEndpoint = actualService.Status.LoadBalancer.Ingress[0].Hostname + ":" + fmt.Sprint(ocsProviderServicePort)
```
nb-ohad, line 185: Same as above

controllers/storagecluster/provider_server.go
nb-ohad: Why complicate things? just have it as a constant. This is for provider mode which is always deployed on its own cluster which means that the chance of collision is almost 0. So why are trying to solve a non-issue

controllers/storagecluster/provider_server.go
```diff
@@ -48,12 +47,17 @@ func (o *ocsProviderServer) ensureCreated(r *StorageClusterReconciler, instance
...
 		o.createSecret,
 		o.createService,
 		o.createDeployment,
 	} {
-		err := f(r, instance)
+		res, err := f(r, instance)
+
+		if !res.IsZero() {
+			return res, err
+		}
```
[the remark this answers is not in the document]
iamniting: If there is an error only, we are not returning immediately. We are doing all 3 operations, collecting the errors, and returning them together.

controllers/storagecluster/provider_server.go
```diff
@@ -168,32 +172,33 @@ func (o *ocsProviderServer) createService(r *StorageClusterReconciler, instance
...
 	}
 
 	r.Log.Info("Service create/update succeeded")
 
-	nodeAddresses, err := r.getWorkerNodesInternalIPAddresses()
-	if err != nil {
-		return err
+	if len(actualService.Status.LoadBalancer.Ingress) == 0 {
+		r.Log.Info("Waiting for Service Status.LoadBalancer.Ingress", "Service", actualService.Name, "Status", actualService.Status)
+		return reconcile.Result{RequeueAfter: time.Second * 5}, nil
```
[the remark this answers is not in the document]

controllers/storagecluster/provider_server.go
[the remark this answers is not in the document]
iamniting: made changes to return result and error immediately

controllers/storagecluster/provider_server.go
```diff
@@ -46,21 +45,19 @@ func (o *ocsProviderServer) ensureCreated(r *StorageClusterReconciler, instance
...
-			multierr.AppendInto(&finalErr, err)
-			continue
+		res, err := f(r, instance)
+
+		if !res.IsZero() || err != nil {
+			return res, err
 		}
 	}
 
-	return reconcile.Result{}, finalErr
```
nb-ohad, line 63: Why complicate things with a loop? 
You force every create func to have the exact same sig! no args and 2 return values. Why put these restrictions on the developer?

```suggestion
	if err := o.createSecret; err != nil {
		return reconcile.Result{}, err
	}
	
	if result, err: = o.createService; err != nil {
		return err
	} else if !res.IsZero() || {
		return result, nil
	} 
	
	if err := o.createDeployment; err != nil {
		return reconcile.Result{}, err
	}
	
	return reconcile.Result{}, nil
 ```

File: controllers/storagecluster/provider_server.go [hunk 7]
```diff
@@ -96,7 +93,7 @@ func (o *ocsProviderServer) ensureDeleted(r *StorageClusterReconciler, instance
 	return reconcile.Result{}, finalErr
 }
 
-func (o *ocsProviderServer) createDeployment(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) error {
+func (o *ocsProviderServer) createDeployment(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) (reconcile.Result, error) {
```
nb-ohad, line 96: No need to introduce a result into the sig of this func. Please revert

controllers/storagecluster/provider_server.go
```diff
@@ -168,32 +166,37 @@ func (o *ocsProviderServer) createService(r *StorageClusterReconciler, instance
...
+	instance.Status.StorageProviderEndpoint = endpoint
 
 	r.Log.Info("status.storageProviderEndpoint is updated", "Endpoint", instance.Status.StorageProviderEndpoint)
 
-	return nil
+	return reconcile.Result{}, nil
 }
 
-func (o *ocsProviderServer) createSecret(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) error {
+func (o *ocsProviderServer) createSecret(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) (reconcile.Result, error) {
```
nb-ohad, line 199: No need to introduce a result into the sig of this func. Please revert

controllers/storagecluster/provider_server.go
```diff
@@ -168,32 +166,37 @@ func (o *ocsProviderServer) createService(r *StorageClusterReconciler, instance
...
 	r.Log.Info("Service create/update succeeded")
 
-	nodeAddresses, err := r.getWorkerNodesInternalIPAddresses()
-	if err != nil {
-		return err
+	endpoint := ""
+
+	if len(actualService.Status.LoadBalancer.Ingress) != 0 {
+		if actualService.Status.LoadBalancer.Ingress[0].IP != "" {
+			endpoint = fmt.Sprintf("%s:%d", actualService.Status.LoadBalancer.Ingress[0].IP, ocsProviderServicePort)
```
nb-ohad, line 180:
```suggestion
			endpoint = actualService.Status.LoadBalancer.Ingress[0].IP
```

controllers/storagecluster/provider_server.go
```diff
@@ -168,32 +166,37 @@ func (o *ocsProviderServer) createService(r *StorageClusterReconciler, instance
...
+		} else if actualService.Status.LoadBalancer.Ingress[0].Hostname != "" {
+			endpoint = fmt.Sprintf("%s:%d", actualService.Status.LoadBalancer.Ingress[0].Hostname, ocsProviderServicePort)
```
nb-ohad, line 182:
```suggestion
			endpoint = actualService.Status.LoadBalancer.Ingress[0].Hostname
```

controllers/storagecluster/provider_server.go
```diff
@@ -168,32 +166,37 @@ func (o *ocsProviderServer) createService(r *StorageClusterReconciler, instance
...
-		r.Log.Error(err, "Worker nodes count is zero")
-		return err
+	if endpoint == "" {
+		r.recorder.ReportIfNotPresent(instance, corev1.EventTypeNormal, "Waiting", "Waiting for Ingress on service "+actualService.Name)
+		r.Log.Info("Waiting for Ingress on service", "Service", actualService.Name, "Status", actualService.Status)
+		return reconcile.Result{RequeueAfter: time.Second * 5}, nil
 	}
 
-	instance.Status.StorageProviderEndpoint = nodeAddresses[0] + ":" + fmt.Sprint(ocsProviderServiceNodePort)
+	instance.Status.StorageProviderEndpoint = endpoint
```
nb-ohad, line 192:
```suggestion
	instance.Status.StorageProviderEndpoint = fmt.Sprintf("%s:%d", endpoint, ocsProviderServicePort)
```

nb-ohad (changes requested):

controllers/storagecluster/provider_server.go

controllers/storagecluster/provider_server.go

controllers/storagecluster/provider_server.go
iamniting: I changed this one as I am returning a requeue in this one as well when the server pod is not ready.

nb-ohad: ok

PR #1704, opened 2022-06-09
odf-to-odf: initialize map when adding labels

subhamkrai:
In case of upgrade, we are getting error `panic: assignment to entry in nil map`
because while blockpool creation in 4.10, it was created without labels
and in 4.11 we are trying to assign label which is empty.
This commit fix above issue.

### DIFF
--- a/controllers/storageclassclaim/storageclassclaim_controller.go
+++ b/controllers/storageclassclaim/storageclassclaim_controller.go
@@ -458,10 +458,6 @@ func (r *StorageClassClaimReconciler) reconcileProviderPhases() (reconcile.Resul
 		r.cephBlockPool = &rookCephv1.CephBlockPool{}
 		r.cephBlockPool.Name = fmt.Sprintf("cephblockpool-%s", r.storageConsumer.Name)
 		r.cephBlockPool.Namespace = r.OperatorNamespace
-		r.cephBlockPool.Labels = map[string]string{
-			controllers.StorageConsumerNameLabel: r.storageConsumer.Name,
-		}
-
 	} else if r.storageClassClaim.Spec.Type == "sharedfilesystem" {
 		r.cephFilesystemSubVolumeGroup = &rookCephv1.CephFilesystemSubVolumeGroup{}
 		r.cephFilesystemSubVolumeGroup.Name = fmt.Sprintf("cephfilesystemsubvolumegroup-%s", r.storageConsumer.Name)
@@ -625,7 +621,8 @@ func (r *StorageClassClaimReconciler) reconcileCephBlockPool() error {
 			return fmt.Errorf("Could not find  device set named default in Storage cluster")
 		}
 		pgUnitSize := util.GetPGBaseUnitSize(deviceSet.Count)
-		r.cephBlockPool.Labels[controllers.StorageConsumerNameLabel] = r.storageConsumer.Name
+		addLabel(r.cephBlockPool, controllers.StorageConsumerNameLabel, r.storageConsumer.Name)
+
 		r.cephBlockPool.Spec = rookCephv1.NamedBlockPoolSpec{
 			PoolSpec: rookCephv1.PoolSpec{
 				FailureDomain: failureDomain,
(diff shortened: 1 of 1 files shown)


controllers/storageclassclaim/storageclassclaim_controller.go
```diff
@@ -625,7 +625,15 @@ func (r *StorageClassClaimReconciler) reconcileCephBlockPool() error {
 			return fmt.Errorf("Could not find  device set named default in Storage cluster")
 		}
 		pgUnitSize := util.GetPGBaseUnitSize(deviceSet.Count)
-		r.cephBlockPool.Labels[controllers.StorageConsumerNameLabel] = r.storageConsumer.Name
+		if r.cephBlockPool.Labels == nil {
+			r.cephBlockPool.Labels = map[string]string{
+				controllers.StorageConsumerNameLabel: r.storageConsumer.Name,
+			}
+
+		} else {
```
Madhu-1, line 633: Remove this one duplicate line https://github.com/red-hat-storage/ocs-operator/blob/c1d20824e55fb233e2dc077d4ebbda60a07582e0/controllers/storageclassclaim/storageclassclaim_controller.go#L461-L463

controllers/storageclassclaim/storageclassclaim_controller.go
```diff
@@ -625,7 +625,15 @@ func (r *StorageClassClaimReconciler) reconcileCephBlockPool() error {
 			return fmt.Errorf("Could not find  device set named default in Storage cluster")
 		}
 		pgUnitSize := util.GetPGBaseUnitSize(deviceSet.Count)
-		r.cephBlockPool.Labels[controllers.StorageConsumerNameLabel] = r.storageConsumer.Name
+		if r.cephBlockPool.Labels == nil {
+			r.cephBlockPool.Labels = map[string]string{
```
Madhu-1, line 629: Make use of `r.cephBlockPool.SetLabels` helper

Madhu-1 (changes requested):

controllers/storageclassclaim/storageclassclaim_controller.go
```diff
@@ -625,7 +621,16 @@ func (r *StorageClassClaimReconciler) reconcileCephBlockPool() error {
...
+				controllers.StorageConsumerNameLabel: r.storageConsumer.Name,
+			}
+
+		} else {
+			r.cephBlockPool.Labels[controllers.StorageConsumerNameLabel] = r.storageConsumer.Name
+		}
+		r.cephBlockPool.SetLabels(r.cephBlockPool.Labels)
```
Madhu-1, line 632:
```suggestion
if r.cephBlockPool.Labels == nil {
			r.cephBlockPool.SetLabels(map[string]string{
				controllers.StorageConsumerNameLabel: r.storageConsumer.Name,
			}
			)

		} else {
			r.cephBlockPool.Labels[controllers.StorageConsumerNameLabel] = r.storageConsumer.Name
		}
```

controllers/storageclassclaim/storageclassclaim_controller.go
```diff
@@ -625,7 +621,14 @@ func (r *StorageClassClaimReconciler) reconcileCephBlockPool() error {
...
 		}
 		pgUnitSize := util.GetPGBaseUnitSize(deviceSet.Count)
-		r.cephBlockPool.Labels[controllers.StorageConsumerNameLabel] = r.storageConsumer.Name
+		if r.cephBlockPool.Labels == nil {
+			r.cephBlockPool.SetLabels(map[string]string{
+				controllers.StorageConsumerNameLabel: r.storageConsumer.Name,
+			})
+		} else {
+			r.cephBlockPool.Labels[controllers.StorageConsumerNameLabel] = r.storageConsumer.Name
+		}
```
nb-ohad, line 630:
```suggestion
		labels := cephBlockPool.GetLabels();
		if labels == nil {
			cephBlockPool.SetLabels(map[string]string{})
		}
		labels[controllers.StorageConsumerNameLabel] = r.storageConsumer.Name
```

or even better, you can do what we do in the deployer, have a util function like this one:
https://github.com/red-hat-storage/ocs-osd-deployer/blob/main/utils/utils.go#L46-L54

then replace the entire section with 
```suggestion
		AddLabel(cephBlockPool, controllers.StorageConsumerNameLabel, r.storageConsumer.Name)
```
And start using it in all places we need to set a label (but don't do it in this PR :) )

nb-ohad (changes requested):

PR #1706, opened 2022-06-10
Deprecate StorageCluster's `spec.version` field

rexagod:
Deprecate StorageCluster's `spec.version` field and populate the version in
`status.version` field instead. In later releases, we'll remove the
`spec.version` field entirely.


rexagod:
The aforementioned PR should ideally go in before this one.

nb-ohad:
@rexagod You cannot remove a field on a public API without bumping the CRD version. 
Mainly because this will create issues for anyone who is importing this struct (internal or at our customers) and k8s has no way to do an auto-translate without introducing a new version for the CRD.

What we can do, is mark it as deprecated and just remove the code that uses it!

jarrpa (commented):
Code changes look good. Are the changes in `/config` and `/deploy` generated? if so please move them to a separate commit.

jarrpa (approved):
/lgtm

PR #1718, opened 2022-06-16
hack: make golangci-lint.sh more verbose

jarrpa:
This is to help debug problems when running in the OpenShift CI, but
also should be generally useful.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -18,5 +18,5 @@ linters:
     - structcheck
     - unused
     - varcheck
-    - golint
     - gofmt
+    - revive
--- a/controllers/storageclassclaim/storageclassclaim_controller.go
+++ b/controllers/storageclassclaim/storageclassclaim_controller.go
@@ -711,7 +711,7 @@ func (r *StorageClassClaimReconciler) reconcileCephBlockPool() error {
 			return err
 		}
 		deviceSetList := r.storageCluster.Spec.StorageDeviceSets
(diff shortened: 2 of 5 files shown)


iamniting (changes requested):
Instead of this curl and script method why don't we directly use `go install`

jarrpa:
> Instead of this curl and script method why don't we directly use `go install`

Because these are the recommended instructions, after just using GH Actions, per their own documentation: https://golangci-lint.run/usage/install/ This also allows devs to run it locally without pushing to a repo first.

iamniting:
> > Instead of this curl and script method why don't we directly use `go install`
> 
> Because these are the recommended instructions, after just using GH Actions, per their own documentation: https://golangci-lint.run/usage/install/ This also allows devs to run it locally without pushing to a repo first.

Even installing via `go install` allows the same as we are doing other installs like controller-gen etc in the makefile.

jarrpa:
Unrelated, the CI still seems to be facing some networking issues, this time in functests.... I'll keep an eye on this and bug the DPTP team as needed.

iamniting (approved):
/lgtm

jarrpa:
I added a new commit for build-functests, just to see.

jarrpa:
Yup, problem still persists... let's keep this open and retesting until we confirm the issue has been resolved.

/hold

BlaineEXE (approved):

PR #1724, opened 2022-06-20
ocs: Update Volume Snapshots secret name generation

parth-gr:
For external restriced mode the csi-secrets name will
not be default, so updating the name accordingly in the
volume snapshot class so snapshots can work

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=2091998

Signed-off-by: parth-gr <paarora@redhat.com>

### DIFF
--- a/controllers/storagecluster/generate.go
+++ b/controllers/storagecluster/generate.go
@@ -3,6 +3,7 @@ package storagecluster
 import (
 	"fmt"
 	"strconv"
+	"strings"
 
 	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v1"
 	"github.com/red-hat-storage/ocs-operator/controllers/util"
@@ -79,7 +80,23 @@ func generateNameForSnapshotClassDriver(initData *ocsv1.StorageCluster, snapshot
 	return fmt.Sprintf("%s.%s.csi.ceph.com", initData.Namespace, snapshotType)
 }
 
-func generateNameForSnapshotClassSecret(snapshotType SnapshotterType) string {
+func generateNameForSnapshotClassSecret(instance *ocsv1.StorageCluster, snapshotType SnapshotterType) string {
+	if instance.Spec.ExternalStorage.Enable {
+		data, ok := externalOCSResources[instance.UID]
+		if !ok {
+			log.Error(fmt.Errorf("Unable to retrieve external resource from externalOCSResources"),
+				"unable to generate name for snapshot class secret for external mode")
+		}
+		// print the Secret name which contains the prefix as the rook-csi-rbd/cephfs-provisioner default secret name
+		// for example if the secret name is rook-csi-rbd-node-rookStorage-replicapool it will check the prefix with rook-csi-rbd-node if it matches it will return that name
+		for _, d := range data {
+			if d.Kind == "Secret" {
+				if strings.Contains(d.Name, fmt.Sprintf("rook-csi-%s-provisioner", snapshotType)) {
+					return d.Name
+				}
+			}
+		}
+	}
 	return fmt.Sprintf("rook-csi-%s-provisioner", snapshotType)
 }
 
--- a/controllers/storagecluster/volumesnapshotterclasses.go
+++ b/controllers/storagecluster/volumesnapshotterclasses.go
@@ -48,7 +48,7 @@ func newVolumeSnapshotClass(instance *ocsv1.StorageCluster, snapShotterType Snap
 		Driver: generateNameForSnapshotClassDriver(instance, snapShotterType),
 		Parameters: map[string]string{
 			"clusterID":                instance.Namespace,
-			snapshotterSecretName:      generateNameForSnapshotClassSecret(snapShotterType),
+			snapshotterSecretName:      generateNameForSnapshotClassSecret(instance, snapShotterType),
 			snapshotterSecretNamespace: instance.Namespace,
 		},
 		DeletionPolicy: snapapi.VolumeSnapshotContentDelete,


controllers/storagecluster/generate.go
```diff
@@ -79,7 +80,22 @@ func generateNameForSnapshotClassDriver(initData *ocsv1.StorageCluster, snapshot
 	return fmt.Sprintf("%s.%s.csi.ceph.com", initData.Namespace, snapshotType)
 }
 
-func generateNameForSnapshotClassSecret(snapshotType SnapshotterType) string {
+func generateNameForSnapshotClassSecret(instance *ocsv1.StorageCluster, snapshotType SnapshotterType) string {
+	if instance.Spec.ExternalStorage.Enable == true {
```
travisn, line 84:
```suggestion
	if instance.Spec.ExternalStorage.Enable {
```

controllers/storagecluster/generate.go
```diff
@@ -79,7 +80,22 @@ func generateNameForSnapshotClassDriver(initData *ocsv1.StorageCluster, snapshot
...
+		data, ok := externalOCSResources[instance.UID]
+		if !ok {
+			log.Error(fmt.Errorf("Unable to retrieve external resource from externalOCSResources"),
+				"unable to generate name for snapshot class secret for external mode")
+		}
+		// print the Secret name which contains the prefix as the rook-csi-rbd/cephfs-provisioner default secret name
+		for _, d := range data {
+			if d.Kind == "Secret" {
+				if strings.Contains(d.Name, fmt.Sprintf("rook-csi-%s-provisioner", snapshotType)) {
+					return d.Name
```
travisn, line 94: How does `d.Name` compare to the default name? Maybe add an example to the comment.

controllers/storagecluster/generate.go
```diff
@@ -79,7 +80,23 @@ func generateNameForSnapshotClassDriver(initData *ocsv1.StorageCluster, snapshot
...
+		data, ok := externalOCSResources[instance.UID]
+		if !ok {
+			log.Error(fmt.Errorf("Unable to retrieve external resource from externalOCSResources"),
+				"unable to generate name for snapshot class secret for external mode")
+		}
+		// print the Secret name which contains the prefix as the rook-csi-rbd/cephfs-provisioner default secret name
+		// for example if the secret name is rook-csi-rbd-node-rookStorage-replicapool it will check the prefix with rook-csi-rbd-node if it matches it will return that name
+		for _, d := range data {
+			if d.Kind == "Secret" {
+				if strings.Contains(d.Name, fmt.Sprintf("rook-csi-%s-provisioner", snapshotType)) {
```
Madhu-1, line 94: what about the upgraded cluster? do we need to delete and recreate the snapshotclass?

parth-gr: We are not supporting the Restricted modes for the upgraded cluster, will only be used for fresh deployments,

We can think about it for upgrades later too but for that, we need to create fresh CSI secrets too,
And ig for now any changes in the storage class are picked by the ocs-operator by deleting the class and creating it again. Not sure the same is implemented for volume snapshot class too or not, @iamniting

controllers/storagecluster/generate.go
```diff
@@ -79,7 +80,22 @@ func generateNameForSnapshotClassDriver(initData *ocsv1.StorageCluster, snapshot
 	return fmt.Sprintf("%s.%s.csi.ceph.com", initData.Namespace, snapshotType)
 }
 
-func generateNameForSnapshotClassSecret(snapshotType SnapshotterType) string {
+func generateNameForSnapshotClassSecret(instance *ocsv1.StorageCluster, snapshotType SnapshotterType) string {
+	if instance.Spec.ExternalStorage.Enable == true {
+		data, ok := externalOCSResources[instance.UID]
```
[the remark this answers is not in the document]
parth-gr: It's the json data from the python script, initialised by ocs-operator

controllers/storagecluster/generate.go
parth-gr: I tested and even the upgrade was taken over by ocs-operator to update the volume-snapshot class

parth-gr:
The testing looks good the volume snapshot class is been updated to use the desired secret name,

@iamniting @nb-ohad can you have a look onto the cahnges,
And also see it wouldn't affect the managed services.

nb-ohad:
@parth-gr @agarwal-mudit This change has a very high probability to break consumer mode. 
I would highly suggest to test a custom build of this change in consumer mode before merging.

controllers/storagecluster/generate.go
Madhu-1: No, it won't.  these secrets are generated by the python script what Rook generates. and we don't use newVolumeSnapshotClass or generateNameForSnapshotClassSecret in consumer/provider mode. but as you mentioned well its good to test these changes to make sure nothing breaks.

nb-ohad: So these secrets are not also coming from our API server in consumer mode? 
If not, what are the equivalent secrets in our case?

Madhu-1: No, in our case when we create the ceph users rook internally creates a secretes per user and extracts information from it, and returns secret name and data in Apiserver response https://github.com/red-hat-storage/ocs-operator/blob/main/services/provider/server/server.go#L565-L591

nb-ohad: right, thank you for clarifying

parth-gr:
> @parth-gr @agarwal-mudit This change has a very high probability to break consumer mode. I would highly suggest testing a custom build of this change in consumer mode before merging.
> 
> Putting a hold until we have clarity

It just matches the secret name prefix which would not affect the by default secrets, 
I tested this on external cluster, do you also wanted to test me the manged services mode?

parth-gr:
@nb-ohad @Madhu-1 @iamniting any update??

parth-gr:
> > It just matches the secret name prefix which would not affect the by default secrets,
> > I tested this on external cluster, do you also wanted to test me the manged services mode?
> 
> My main concern is not the change to the secret name, as @Madhu-1 already confirmed that in consumer mode we are using a different set of secrets. The concern is the structure of the data in `externalOCSResources`.
> 
> @iamniting @Madhu-1 I would like your opinion on the effect of the storagecluster UID indexing change on the `externalOCSResources` map. Is that going to cause us problems in consumer mode?



> @nb-ohad @Madhu-1 @iamniting any update??

PS: I feel like there is no other parameter added so the total number of values will be the same in the map so no change in ordering, may that's what your question means

controllers/storagecluster/generate.go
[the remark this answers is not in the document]
Madhu-1: AFAIK in consumer mode, we dont call this function nor newVolumeSnapshotClass function also.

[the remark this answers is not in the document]
iamniting: We do call it https://github.com/red-hat-storage/ocs-operator/blob/main/controllers/storagecluster/reconcile.go#L405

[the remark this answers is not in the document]
Madhu-1: it will move skipped for consumer mode here https://github.com/red-hat-storage/ocs-operator/pull/1729/files#diff-ebb1f458b4aa3b51172933994b017c626698cd74c8de4b9fcf63ae92f5edc290R133

controllers/storagecluster/generate.go
```diff
@@ -79,7 +80,23 @@ func generateNameForSnapshotClassDriver(initData *ocsv1.StorageCluster, snapshot
...
-func generateNameForSnapshotClassSecret(snapshotType SnapshotterType) string {
+func generateNameForSnapshotClassSecret(instance *ocsv1.StorageCluster, snapshotType SnapshotterType) string {
+	if instance.Spec.ExternalStorage.Enable {
+		data, ok := externalOCSResources[instance.UID]
+		if !ok {
+			log.Error(fmt.Errorf("Unable to retrieve external resource from externalOCSResources"),
+				"unable to generate name for snapshot class secret for external mode")
+		}
+		// print the Secret name which contains the prefix as the rook-csi-rbd/cephfs-provisioner default secret name
+		// for example if the secret name is rook-csi-rbd-node-rookStorage-replicapool it will check the prefix with rook-csi-rbd-node if it matches it will return that name
```
BlaineEXE, line 91: This statement is a little unclear: "... it will return **that** name..." 
Since there are two secret names being discussed here, "that" is unclear to me.

PR #1728, opened 2022-06-22
odf to odf: keep the old nodeport provider service

iamniting:
In ODF v4.10.z the providerService is of type nodePort so the consumer
can give any worker nodeIP:31659 to connect with the provider. But in
ODF v4.11.z this service is changed to the service of the type of
loadBalancer where the consumers will need to update the
storageProviderEndpoint to hostname:50051. The consumers won't
be able to reach out to the provider unless the storageProviderEndpoint
is updated.

Create both the services to allow consumers still be connected and
change the storageProviderEndpoint as per their cadence.

### DIFF
--- a/controllers/storagecluster/provider_server.go
+++ b/controllers/storagecluster/provider_server.go
@@ -26,7 +26,8 @@ const (
 	ocsProviderServerName  = "ocs-provider-server"
 	providerAPIServerImage = "PROVIDER_API_SERVER_IMAGE"
 
-	ocsProviderServicePort = int32(50051)
+	ocsProviderServicePort     = int32(50051)
+	ocsProviderServiceNodePort = int32(31659)
 
 	ocsProviderCertSecretName = ocsProviderServerName + "-cert"
 )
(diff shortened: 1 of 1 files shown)


controllers/storagecluster/provider_server.go
```diff
@@ -55,6 +56,10 @@ func (o *ocsProviderServer) ensureCreated(r *StorageClusterReconciler, instance
 		return res, nil
 	}
 
+	if err := o.createNodePortService(r, instance); err != nil {
```
nb-ohad, line 59: @iamniting Can we add a comment above the call site to indicate that we are keeping this to enable the transition of node port based connectivity to load balancer based connectivity during 4.11 lifetime. And that this is going to be deprecated and removed in a future version.

nb-ohad (changes requested):

controllers/storagecluster/provider_server.go
iamniting: done

controllers/storagecluster/provider_server.go
```diff
@@ -55,6 +56,13 @@ func (o *ocsProviderServer) ensureCreated(r *StorageClusterReconciler, instance
 		return res, nil
 	}
 
+	// TODO: we are keeping this to enable the transition of node port based connectivity
+	// to load balancer based connectivity during 4.11 lifetime.
+	// And that this is going to be deprecated and removed in a future version.
```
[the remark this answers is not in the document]

PR #1735, opened 2022-06-27
odf-to-odf: Introduce heartbeat mechanism to check connectivity

bindrad:
This PR will enable us to have a heartbeat mechanism where consumer will ping provider API server every minute which will update the status of the corresponding storageConsumer CR.
After this, we can have an alert that will be fired if the heartbeat is not received from a consumer in the provided timeframe.
High-level design doc: https://docs.google.com/document/d/1T9MoK6GMPt__Hs2FOMjBjA9Q0H5KiaOhS4y3CYDKv9U/edit?usp=sharing


services/provider/client/client.go
```diff
@@ -196,3 +196,16 @@ func (cc *OCSProviderClient) GetStorageClassClaimConfig(ctx context.Context, con
 
 	return cc.Client.GetStorageClassClaimConfig(apiCtx, req)
 }
+
+func (cc *OCSProviderClient) ConsumerHeartBeat(ctx context.Context, consumerUUID string) (*pb.ConsumerHeartBeatResponse, error) {
+	if cc.Client == nil || cc.clientConn == nil {
+		return nil, fmt.Errorf("provider client is closed")
+	}
+	req := &pb.ConsumerHeartBeatRequest{
```
nb-ohad, line 204:
```suggestion
	req := &pb.ReportStatusRequest{
```

services/provider/client/client.go
```diff
@@ -196,3 +196,16 @@ func (cc *OCSProviderClient) GetStorageClassClaimConfig(ctx context.Context, con
...
+		StorageConsumerUUID: consumerUUID,
+	}
+	apiCtx, cancel := context.WithTimeout(ctx, cc.timeout)
+	defer cancel()
+
+	return cc.Client.ConsumerHeartBeat(apiCtx, req)
```
nb-ohad, line 210:
```suggestion
	return cc.Client.ReportStatus(apiCtx, req)
```

services/consumer/main.go
```diff
@@ -0,0 +1,50 @@
...
+
+	c := pb.NewOCSProviderClient(conn)
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*10)
+	defer cancel()
+
+	_, err = c.ConsumerHeartBeat(ctx, &pb.ConsumerHeartBeatRequest{StorageConsumerUUID: *storageConsumerUID})
+	if err != nil {
+		klog.Fatalf("Could not update heartbeat timestamp of %v storageConsumer", *storageConsumerUID)
+	}
```
nb-ohad, line 48:
```suggestion
	if err := c.ConsumerHeartBeat(ctx, storageConsumerUID); err != nil {
		klog.Fatalf("Could not report consumer status for %v storageConsumer", storageConsumerUID)
	}
```

controllers/storagecluster/external_ocs_heartbeat.go
```diff
@@ -0,0 +1,82 @@
...
+									},
+								},
+							},
+							RestartPolicy: corev1.RestartPolicyOnFailure,
+						},
+					},
+				},
+			},
+		},
+	}
```
nb-ohad, line 55: Please move into `CreateOrUpdate`, if not the `get` part of `CreateOrUpdate` can override some of this desired state.

controllers/storagecluster/external_ocs_heartbeat.go
```diff
@@ -0,0 +1,82 @@
...
+	cronJob := &batchv1.CronJob{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "heartbeat-to-provider",
+			Namespace: instance.Namespace,
+			OwnerReferences: []metav1.OwnerReference{{
+				UID:        instance.UID,
+				APIVersion: instance.APIVersion,
+				Kind:       instance.Kind,
+				Name:       instance.Name,
+			}},
```
nb-ohad, line 30: Please add owner ref using the utils and not manually (as it does not work as expected)

controllers/storagecluster/external_ocs_heartbeat.go
```diff
@@ -0,0 +1,82 @@
...
+					Template: corev1.PodTemplateSpec{
+						Spec: corev1.PodSpec{
+							Containers: []corev1.Container{
+								{
+									Name:  "heartbeat",
+									Image: os.Getenv(providerAPIServerImage),
+									Command: []string{
+										"/usr/local/bin/consumer-heartbeat",
+										fmt.Sprintf("--storageConsumerUID=%v", instance.Status.ExternalStorage.ConsumerID),
+										fmt.Sprintf("--endpoint=%v", instance.Spec.ExternalStorage.StorageProviderEndpoint),
```
nb-ohad, line 45: Everything here can be read from within the job itself. I think it will be better as in the near future the job will have to collect status and report it back which mean it will need much more info then just connection and identity

controllers/storagecluster/external_ocs_heartbeat.go
```diff
@@ -0,0 +1,82 @@
...
+			}
+			oldCronJob.Spec = cronJob.Spec
+			err := r.Client.Update(context.TODO(), oldCronJob)
+			if err != nil {
+				return fmt.Errorf("failed while updating CronJob: %v", err)
+			}
+		} else {
+			return fmt.Errorf("failed while creating CronJob: %v", err)
+		}
+	}
```
nb-ohad, line 80: Why are we re-implementing `controllerutil.CreateOrUpdate`?

services/consumer/main.go
```diff
@@ -0,0 +1,50 @@
...
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+	"k8s.io/klog"
+
+	pb "github.com/red-hat-storage/ocs-operator/services/provider/pb"
+)
+
+var (
+	storageProviderEndpoint = flag.String("endpoint", "localhost:50051", "The storageProviderEndpoint")
+	storageConsumerUID      = flag.String("storageConsumerUID", "UID", "The UID of storageConsumer CR")
```
nb-ohad, line 19: The default values here do not make sense. In addition all of this info can be read from within the Job which is something that we will need anyway when we start to report actual status

nb-ohad (changes requested):

iamniting (changes requested):
Can we also divide this one commit into multiple commits?

controllers/storagecluster/external_ocs_heartbeat.go
```diff
@@ -0,0 +1,82 @@
...
+			Namespace: instance.Namespace,
+			OwnerReferences: []metav1.OwnerReference{{
+				UID:        instance.UID,
+				APIVersion: instance.APIVersion,
+				Kind:       instance.Kind,
+				Name:       instance.Name,
+			}},
+		},
+		Spec: batchv1.CronJobSpec{
+			Schedule: "* * * * *",
```
Madhu-1, line 33: we should define the schedule internal run every 10s/15s like that isn't it?

services/consumer/main.go
```diff
@@ -0,0 +1,50 @@
+package main
+
+import (
+	"context"
+	"flag"
+	"time"
+
```
Madhu-1, line 7: correct import orders

services/consumer/main.go
```diff
@@ -0,0 +1,50 @@
...
+	conn, err := grpc.Dial(*storageProviderEndpoint, opts...)
+	if err != nil {
+		klog.Fatalf("failed to dial: %v", err)
+	}
+	defer conn.Close()
+
+	c := pb.NewOCSProviderClient(conn)
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*10)
+	defer cancel()
```
Madhu-1, line 43: just use `providerclient.NewProviderClient` ?

controllers/storagecluster/external_ocs_heartbeat.go
nb-ohad: I would say every 1 min or even 5 min ! we don't need to report a heartbeat in the seconds

api/v1alpha1/storageconsumer_types.go
```diff
@@ -66,6 +66,8 @@ type StorageConsumerStatus struct {
 	GrantedCapacity resource.Quantity `json:"grantedCapacity,omitempty"`
 	// CephResources provide details of created ceph resources required for external storage
 	CephResources []*CephResourcesSpec `json:"cephResources,omitempty"`
+	// Timestamp of last heartbeat received from consumer
+	HeartbeatTimestamp metav1.Time `json:"heartbeatTimestamp,omitempty"`
```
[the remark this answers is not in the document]
Madhu-1: as its a json value better to have it empty or some default values show up? in deployment spec i have seen its pointer to time.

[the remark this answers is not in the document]
Madhu-1: if the field is not showing any unwanted value until the status is updated we are good we don't need to change it to pointer. but please check it once.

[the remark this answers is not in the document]
nb-ohad: `omitempty` should take care of a 0 value and this is a timestamp so a value of 0 means 01.01.1970 which is OK for the last time an heartbeat append (never)

controllers/storagecluster/external_ocs_heartbeat.go
bindrad: What does reading from job mean over here?

config/rbac/role.yaml
```diff
@@ -24,6 +24,18 @@ rules:
...
+  resources:
+  - cronjobs
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
```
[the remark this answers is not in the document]
bindrad: Yes, we will need all the CRUD operation verbs

controllers/storagecluster/external_ocs.go
```diff
@@ -107,6 +107,12 @@ func (r *StorageClusterReconciler) acknowledgeOnboarding(instance *ocsv1.Storage
 	instance.Status.Phase = statusutil.PhaseProgressing
 
 	r.Log.Info("External-OCS:Onboarding is acknowledged successfully.")
+
+	// start the cronJob to ping the provider api server
+	if err := r.enableHeartBeatCronJob(instance, externalClusterClient); err != nil {
```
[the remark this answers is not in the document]
bindrad: As suggested by Ohad and Nitin, I'll add it to the reconcile path

services/provider/client/client.go
```diff
@@ -196,3 +196,16 @@ func (cc *OCSProviderClient) GetStorageClassClaimConfig(ctx context.Context, con
 
 	return cc.Client.GetStorageClassClaimConfig(apiCtx, req)
 }
+
+func (cc *OCSProviderClient) ConsumerHeartBeat(ctx context.Context, consumerUUID string) (*pb.ConsumerHeartBeatResponse, error) {
```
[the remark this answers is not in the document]
bindrad: Agreed with all the point
Regarding the 5th point, we need to return a response as gRPC doesn't allow us to send a nil response but it allows us to send an empty response.

services/provider/server/server.go
```diff
@@ -658,3 +658,18 @@ func (s *OCSProviderServer) GetStorageClassClaimConfig(ctx context.Context, req
...
 
 }
+
+// ConsumerHeartBeat rpc call to check if a consumer can reach to the provider.
+func (s *OCSProviderServer) ConsumerHeartBeat(ctx context.Context, req *pb.ConsumerHeartBeatRequest) (*pb.ConsumerHeartBeatResponse, error) {
+	// Update the status in storageConsumer CR
+	if err := s.consumerManager.UpdateTimestamp(ctx, req.StorageConsumerUUID); err != nil {
+		if kerrors.IsNotFound(err) {
+			return nil, status.Errorf(codes.NotFound, "failed to update heartbeatTimestamp in the storageConsumer resource: %v", err)
+		} else if err == errFailedPrecondition {
```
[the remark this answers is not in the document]
bindrad: I thought we should return an error or warning if storageConsumer is disabled

services/consumer/main.go
bindrad: Why? The name of gRPC service is OCSProvider so client name will be NewOCSProviderClient, right?

api/v1alpha1/storageconsumer_types.go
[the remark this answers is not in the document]
bindrad: Updated in latest commit

controllers/storagecluster/external_ocs_heartbeat.go
```diff
@@ -0,0 +1,82 @@
...
+	"os"
+
+	batchv1 "k8s.io/api/batch/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+
+	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v1"
+	externalClient "github.com/red-hat-storage/ocs-operator/services/provider/client"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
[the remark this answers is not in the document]
bindrad: Updated in latest commit

services/consumer/main.go
bindrad: Updated in latest commit

controllers/storagecluster/external_ocs.go
```diff
@@ -107,6 +107,12 @@ func (r *StorageClusterReconciler) acknowledgeOnboarding(instance *ocsv1.Storage
...
+		r.Log.Error(err, "Failed to update cronJob")
+		return reconcile.Result{Requeue: true}, nil
+	}
```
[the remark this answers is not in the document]
bindrad: Updated in latest commit

services/provider/client/client.go
bindrad: Updated in latest commit

services/provider/client/client.go
bindrad: Updated in latest commit

services/consumer/main.go
bindrad: Updated in latest commit

controllers/storagecluster/external_ocs_heartbeat.go
bindrad: Updated in latest commit

services/consumer/main.go
```diff
@@ -0,0 +1,50 @@
...
+
+	_, err = c.ConsumerHeartBeat(ctx, &pb.ConsumerHeartBeatRequest{StorageConsumerUUID: *storageConsumerUID})
+	if err != nil {
```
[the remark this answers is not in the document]
bindrad: Using klog.Exit, updated in latest commit

services/consumer/main.go
nb-ohad: @bindrad I think the ask here is to use the method on the providerclient struct instead of accessing the protobuf struct directly as the client is supposed to be an obstruction of the protobuf struct

controllers/storagecluster/external_ocs_heartbeat.go
nb-ohad: It means that the job code will get the input directly from the storagecluster resources (by reading it) and not via command-line arguments.

controllers/storagecluster/external_resources.go
```diff
@@ -254,6 +254,8 @@ func (obj *ocsExternalResources) ensureCreated(r *StorageClusterReconciler, inst
 			return r.onboardConsumer(instance, externalClusterClient)
 		} else if instance.Status.Phase == statusutil.PhaseOnboarding {
 			return r.acknowledgeOnboarding(instance, externalClusterClient)
+		} else if instance.Status.Phase == statusutil.PhaseReady {
+			return r.enableReportStatusCronJob(instance, externalClusterClient)
```
nb-ohad, line 258: Why do we need to run this only when we are Ready? why can't we just call this in the else section. 
Also, this change means that the code below (the next `else if`)will not apply to Ready state anymore - which I don't think was your original intention.

services/provider/server/consumer.go
```diff
@@ -224,3 +224,18 @@ func (c *ocsConsumerManager) Get(ctx context.Context, id string) (*ocsv1alpha1.S
...
+func (c *ocsConsumerManager) UpdateTimestamp(ctx context.Context, id string) error {
+	consumerObj, err := c.Get(ctx, id)
+	if err != nil {
+		return err
+	}
+	consumerObj.Status.HeartbeatTimestamp = metav1.Now()
+	err = c.client.Status().Update(ctx, consumerObj)
+	if err != nil {
+		return fmt.Errorf("failed to update heartbeatTimestamp status %q. %v", consumerObj.Name, err)
+	}
```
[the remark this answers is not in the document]
nb-ohad: A different way to solve it is a retry loop, but only  in the case where an update failed for the reason of a resourceVersion issue

controllers/storagecluster/external_resources.go
nb-ohad, line 258: Why do we need to run this only when we are Ready? why can't we just call this in the else section?
Also, this change means that the code below (the next else if)will not apply to Ready state anymore - which I don't think was your original intention.

nb-ohad (changes requested):

controllers/storagecluster/external_resources.go
bindrad: > Why do we need to run this only when we are Ready?
Because once the external storageCluster is created that means there is a corresponding storageConsumer on provider so we can hit the gRPC endpoint to update lastHeartbeat timestamp. If storageCluster is in onboarding or connecting state then there is no guarantee that storageConsumer resource is created. So I think it is better to run the cronJob after storageCluster is in ready state.

> why can't we just call this in the else section?
There are other phases as well such as error, ignored and not ready. If storageCluster is in this phase just after creation then I don't think we should start the job as corresponding storageConsumer is not created.

hack/go-build.sh
```diff
@@ -12,3 +12,4 @@ LDFLAGS="-s -w -X github.com/red-hat-storage/ocs-operator/controllers/defaults.I
 go build -tags 'netgo osusergo' -ldflags="${LDFLAGS}" -o ${OUTDIR_BIN}/ocs-operator ./main.go
 go build -tags 'netgo osusergo' -ldflags="${LDFLAGS}" -o ${OUTDIR_BIN}/metrics-exporter ./metrics/main.go
 go build -tags 'netgo osusergo' -ldflags="${LDFLAGS}" -o ${OUTDIR_BIN}/provider-api ./services/provider/main.go
+go build -tags 'netgo osusergo' -ldflags="${LDFLAGS}" -o ${OUTDIR_BIN}/consumer-heartbeat ./services/consumer/main.go
```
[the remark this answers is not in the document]
bindrad: Updated in latest commit

controllers/storagecluster/external_ocs_heartbeat.go
```diff
@@ -0,0 +1,59 @@
...
+	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v1"
+	externalClient "github.com/red-hat-storage/ocs-operator/services/provider/client"
+
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+func (r *StorageClusterReconciler) enableHeartBeatCronJob(instance *ocsv1.StorageCluster, externalClusterClient *externalClient.OCSProviderClient) error {
```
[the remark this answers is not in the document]
bindrad: Updated in latest commit

services/provider/server/consumer.go
```diff
@@ -224,3 +224,18 @@ func (c *ocsConsumerManager) Get(ctx context.Context, id string) (*ocsv1alpha1.S
 
 	return consumerObj, nil
 }
+
+func (c *ocsConsumerManager) UpdateTimestamp(ctx context.Context, id string) error {
```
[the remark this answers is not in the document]
bindrad: Updated in latest commit

services/consumer/main.go
bindrad: Understood, updated in latest commit

services/consumer/main.go
bindrad: Got it, updated in latest commit

services/consumer/main.go
```diff
@@ -0,0 +1,64 @@
...
+
+	config, err := config.GetConfig()
+	if err != nil {
+		klog.Exitf("failed to get config: %v", err)
+	}
+	client, err := client.New(config, client.Options{Scheme: scheme})
+	if err != nil {
+		klog.Exitf("failed to create controller-runtime client: %v", err)
+	}
+	namespace, err := util.GetWatchNamespace()
```
Madhu-1, line 37: use OPERATOR_NAMESPACE env.

controllers/storagecluster/external_ocs_heartbeat.go
```diff
@@ -0,0 +1,70 @@
...
+		if err := controllerutil.SetOwnerReference(instance, cronJob, r.Client.Scheme()); err != nil {
+			return err
+		}
+
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+	return nil
```
Madhu-1, line 69: just return err no need to have a check here?

controllers/storagecluster/external_ocs_heartbeat.go
```diff
@@ -0,0 +1,70 @@
...
+						},
+					},
+				},
+			},
+		}
+		if err := controllerutil.SetOwnerReference(instance, cronJob, r.Client.Scheme()); err != nil {
+			return err
+		}
+
+		return nil
```
Madhu-1, line 64: just return err? no check is required

controllers/storagecluster/external_resources.go
```diff
@@ -261,6 +261,10 @@ func (obj *ocsExternalResources) ensureCreated(r *StorageClusterReconciler, inst
 			}
 		}
 
+		if instance.Status.Phase == statusutil.PhaseReady {
```
iamniting, line 264: We should not check for the status.phase at all, We should just call it and return only if there is an error.

This will create a problem where we will not go ahead with the below code if the status is already ready.

controllers/storagecluster/external_ocs_heartbeat.go
```diff
@@ -0,0 +1,70 @@
...
+								{
+									Name:  "heartbeat",
+									Image: os.Getenv(providerAPIServerImage),
+									Command: []string{
+										"/usr/local/bin/status-reporter",
+									},
+									Env: []corev1.EnvVar{
+										{
+											Name:  util.WatchNamespaceEnvVar,
+											Value: os.Getenv(util.WatchNamespaceEnvVar),
```
iamniting, line 44: I would suggest using instance.namespace here

controllers/storagecluster/external_ocs_heartbeat.go
```diff
@@ -0,0 +1,70 @@
...
+										},
+										{
+											Name:  "STORAGE_CLUSTER_NAME",
```
iamniting, line 47: We should use the constant name here which we defined

controllers/storagecluster/external_ocs_heartbeat.go
iamniting: +1

controllers/storagecluster/external_ocs_heartbeat.go
iamniting: +1

controllers/storagecluster/external_ocs.go
```diff
@@ -107,9 +107,20 @@ func (r *StorageClusterReconciler) acknowledgeOnboarding(instance *ocsv1.Storage
 	instance.Status.Phase = statusutil.PhaseProgressing
 
 	r.Log.Info("External-OCS:Onboarding is acknowledged successfully.")
+
 	return reconcile.Result{Requeue: true}, nil
 }
 
+func (r *StorageClusterReconciler) reconcileConsumerStatusReporterJob(instance *ocsv1.StorageCluster, externalClusterClient *externalClient.OCSProviderClient) (reconcile.Result, error) {
```
iamniting, line 114: IMO we are not required this extra wrapper func as we are just calling one func here which can be called from the calling site itself.

services/consumer/main.go
iamniting, line 37: We should change this according to my other comment about the env in the external_ocs_heartbeat.go

iamniting (changes requested):

services/consumer/main.go
bindrad: Applied Nitin's suggestion

controllers/storagecluster/external_ocs_heartbeat.go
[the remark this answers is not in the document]
bindrad: Let me know if you feel it needs to be updated to `OperatorNamespaceEnvVar`

services/consumer/main.go
bindrad: Updated in latest commit

controllers/storagecluster/external_ocs.go
bindrad: Updated in latest commit

controllers/storagecluster/external_ocs_heartbeat.go
bindrad: Updated in latest commit

controllers/storagecluster/external_ocs_heartbeat.go
bindrad: Updated in latest commit

controllers/storagecluster/external_resources.go
bindrad: Updated in latest commit

controllers/storagecluster/external_ocs_heartbeat.go
bindrad: Updated in latest commit

rbac/consumer-role.yaml
```diff
@@ -0,0 +1,12 @@
+# permissions for end users to edit storageclusters.
```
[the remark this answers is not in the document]
bindrad: copy paste mistake

services/consumer/main.go
bindrad: Let me know if you feel it needs to be updated to `OperatorNamespaceEnvVar`

config/rbac/role.yaml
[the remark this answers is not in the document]
bindrad: Updated in latest commit

controllers/storagecluster/external_ocs.go
```diff
@@ -107,9 +116,64 @@ func (r *StorageClusterReconciler) acknowledgeOnboarding(instance *ocsv1.Storage
...
+							RestartPolicy:      corev1.RestartPolicyOnFailure,
+							ServiceAccountName: "ocs-status-reporter",
+						},
+					},
+				},
+			},
+		}
+		if err := controllerutil.SetOwnerReference(instance, cronJob, r.Client.Scheme()); err != nil {
+			return fmt.Errorf("failed to set owner reference: %v", err)
+		}
```
nb-ohad, line 167: Can we please move this above the spec update?

services/consumer/main.go
```diff
@@ -0,0 +1,63 @@
...
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/klog"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+)
+
+func main() {
+	flag.Parse()
```
nb-ohad, line 21: What is the reason to parse flags when we have no flags defined?

services/consumer/main.go
```diff
@@ -0,0 +1,63 @@
...
+	storageCluster := &ocsv1.StorageCluster{}
+	if err := client.Get(context.Background(), types.NamespacedName{Name: name, Namespace: namespace}, storageCluster); err != nil {
+		klog.Exit("failed to get storageCluster: %v", err)
+	}
+
+	providerClient, err := providerclient.NewProviderClient(
+		context.Background(),
+		storageCluster.Spec.ExternalStorage.StorageProviderEndpoint,
+		10*time.Second,
+	)
```
nb-ohad, line 54: I think you are missing a defer to make sure the client is closed if we exit on errors.

services/consumer/main.go
```diff
@@ -0,0 +1,63 @@
...
+	if !isSet {
+		klog.Exit("STORAGE_CLUSTER_NAME env var not set")
+	}
+	namespace, isSet := os.LookupEnv(storagecluster.StorageClusterNamespaceEnvVar)
+	if !isSet {
+		klog.Exit("STORAGE_CLUSTER_NAMESPACE env var not set")
+	}
+
+	storageCluster := &ocsv1.StorageCluster{}
+	if err := client.Get(context.Background(), types.NamespacedName{Name: name, Namespace: namespace}, storageCluster); err != nil {
```
nb-ohad, line 46: Why are we creating another context? Can you please create a single context, at the start, for the entire flow, and reuse it?

services/consumer/main.go
```diff
@@ -0,0 +1,63 @@
...
+	providerClient, err := providerclient.NewProviderClient(
+		context.Background(),
+		storageCluster.Spec.ExternalStorage.StorageProviderEndpoint,
+		10*time.Second,
+	)
+	if err != nil {
+		klog.Exitf("failed to create grpc client: %v", err)
+	}
+
+	if _, err = providerClient.ReportStatus(context.Background(), storageCluster.Status.ExternalStorage.ConsumerID); err != nil {
```
nb-ohad, line 59: Why are we creating another context? Can you please create a single context, at the start, for the entire flow, and reuse it?

services/provider/server/consumer.go
```diff
@@ -224,3 +224,18 @@ func (c *ocsConsumerManager) Get(ctx context.Context, id string) (*ocsv1alpha1.S
 
 	return consumerObj, nil
 }
+
+func (c *ocsConsumerManager) UpdateConsumerReportedStatus(ctx context.Context, id string) error {
```
nb-ohad, line 228: Let's simplify the name: 
- The actual action doesn't care where the status came from so it makes no sense to have the word `Reported` in the name
- Consumer is given as this is a consumer wrapping class. 
- This func is just updating the LastHeartbeat it makes sense to have that in the name

```suggestion
func (c *ocsConsumerManager) UpdateStatusLastHeatbeat(ctx context.Context, id string) error {
```

nb-ohad (changes requested):

services/provider/server/consumer.go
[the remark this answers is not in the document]
nb-ohad: @bindrad This is still not resolved in the latest code

rbac/consumer-role.yaml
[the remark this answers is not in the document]
nb-ohad: @bindrad This is still there in the latest code

services/provider/server/consumer.go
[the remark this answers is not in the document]
bindrad: Updated in the latest commit

rbac/consumer-role.yaml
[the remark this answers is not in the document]
bindrad: Updated in the latest commit

controllers/storagecluster/external_ocs.go
```diff
@@ -107,9 +116,64 @@ func (r *StorageClusterReconciler) acknowledgeOnboarding(instance *ocsv1.Storage
 	instance.Status.Phase = statusutil.PhaseProgressing
 
 	r.Log.Info("External-OCS:Onboarding is acknowledged successfully.")
+
```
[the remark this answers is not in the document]
bindrad: Updated in the latest commit

controllers/storagecluster/external_ocs.go
```diff
@@ -107,9 +116,64 @@ func (r *StorageClusterReconciler) acknowledgeOnboarding(instance *ocsv1.Storage
...
 }
 
+func (r *StorageClusterReconciler) reconcileConsumerStatusReporterJob(instance *ocsv1.StorageCluster, externalClusterClient *externalClient.OCSProviderClient) (reconcile.Result, error) {
+	// start the cronJob to ping the provider api server
+	cronJob := &batchv1.CronJob{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "report-status-to-provider",
+			Namespace: instance.Namespace,
+		},
+	}
```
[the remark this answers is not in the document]
bindrad: Updated in the latest commit

controllers/storagecluster/external_ocs.go
```diff
@@ -107,9 +116,64 @@ func (r *StorageClusterReconciler) acknowledgeOnboarding(instance *ocsv1.Storage
...
+									Env: []corev1.EnvVar{
+										{
+											Name:  StorageClusterNamespaceEnvVar,
+											Value: instance.Namespace,
+										},
+										{
+											Name:  StorageClusterNameEnvVar,
+											Value: instance.Name,
+										},
+									},
```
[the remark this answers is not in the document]
bindrad: Updated in the latest commit

controllers/storagecluster/external_ocs.go
bindrad: Updated in the latest commit

controllers/storagecluster/external_resources.go
```diff
@@ -261,6 +261,14 @@ func (obj *ocsExternalResources) ensureCreated(r *StorageClusterReconciler, inst
 			}
 		}
 
+		if res, err := r.reconcileConsumerStatusReporterJob(instance, externalClusterClient); err != nil {
+			return res, err
+		}
+
+		if instance.Status.Phase == statusutil.PhaseReady {
+			return r.reconcileConsumerStatusReporterJob(instance, externalClusterClient)
+		}
```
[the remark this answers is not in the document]
bindrad: I missed to remove the older implementation, updated in the latest commit

services/consumer/main.go
bindrad: Updated in the latest commit

services/consumer/main.go
```diff
@@ -0,0 +1,63 @@
...
+		klog.Exitf("failed to get config: %v", err)
+	}
+	client, err := client.New(config, client.Options{Scheme: scheme})
+	if err != nil {
+		klog.Exitf("failed to create controller-runtime client: %v", err)
+	}
+	name, isSet := os.LookupEnv(storagecluster.StorageClusterNameEnvVar)
+	if !isSet {
+		klog.Exit("STORAGE_CLUSTER_NAME env var not set")
+	}
```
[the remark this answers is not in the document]
bindrad: Updated in the latest commit

services/consumer/main.go
```diff
@@ -0,0 +1,63 @@
...
+
+	storageCluster := &ocsv1.StorageCluster{}
+	if err := client.Get(context.Background(), types.NamespacedName{Name: name, Namespace: namespace}, storageCluster); err != nil {
+		klog.Exit("failed to get storageCluster: %v", err)
+	}
+
+	providerClient, err := providerclient.NewProviderClient(
+		context.Background(),
+		storageCluster.Spec.ExternalStorage.StorageProviderEndpoint,
+		10*time.Second,
```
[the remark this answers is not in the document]
bindrad: Updated in the latest commit

services/consumer/main.go
bindrad: Updated in the latest commit

services/consumer/main.go
bindrad: Updated in the latest commit

services/provider/server/consumer.go
bindrad: Updated in the latest commit

services/provider/server/consumer.go
```diff
@@ -224,3 +224,18 @@ func (c *ocsConsumerManager) Get(ctx context.Context, id string) (*ocsv1alpha1.S
...
+	consumerObj, err := c.Get(ctx, id)
+	if err != nil {
+		return err
+	}
+	consumerObj.Status.LastHeartbeat = metav1.Now()
+	err = c.client.Status().Update(ctx, consumerObj)
```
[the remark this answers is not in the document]
bindrad: Updated in the latest commit
We are using patch because update will rewrite the entire object whereas patch will replace the intended field.
Another advantage of using patch is that it will save an API call to server as only the updated data needs to be provided.

services/provider/server/consumer.go
```diff
@@ -224,3 +226,46 @@ func (c *ocsConsumerManager) Get(ctx context.Context, id string) (*ocsv1alpha1.S
...
+	}{{
+		Op:    "replace",
+		Path:  "/status/lastHeartbeat",
+		Value: metav1.Now(),
+	}}
+	data, err := json.Marshal(payload)
+	if err != nil {
+		return fmt.Errorf("failed to encode the payload: %v", err)
+	}
+
```
nb-ohad, line 259:
```suggestion
```

services/provider/server/consumer.go
```diff
@@ -224,3 +226,46 @@ func (c *ocsConsumerManager) Get(ctx context.Context, id string) (*ocsv1alpha1.S
...
+	patch := client.RawPatch(types.JSONPatchType, data)
+
```
nb-ohad, line 261:
```suggestion
```

services/provider/server/consumer.go
```diff
@@ -224,3 +226,46 @@ func (c *ocsConsumerManager) Get(ctx context.Context, id string) (*ocsv1alpha1.S
...
+	payload := []struct {
+		Op    string      `json:"op"`
+		Path  string      `json:"path"`
+		Value interface{} `json:"value"`
+	}{{
+		Op:    "replace",
+		Path:  "/status/lastHeartbeat",
+		Value: metav1.Now(),
+	}}
+	data, err := json.Marshal(payload)
```
nb-ohad, line 255: Let's simplify the anonymous struct to a single array item type, and rename the generic payload and data names to something more meaningful.  

```suggestion
	patchInfo  := struct {
		Op    string      `json:"op"`
		Path  string      `json:"path"`
		Value interface{} `json:"value"`
	} {
		Op:    "replace",
		Path:  "/status/lastHeartbeat",
		Value: metav1.Now(),
	}
	jsonPatchInfo, err := json.Marshal(interface[]{} {patchInfo})
```

nb-ohad (changes requested):

controllers/storagecluster/external_ocs.go
```diff
@@ -110,6 +117,52 @@ func (r *StorageClusterReconciler) acknowledgeOnboarding(instance *ocsv1.Storage
...
+								{
+									Name:  "heartbeat",
+									Image: os.Getenv(providerAPIServerImage),
+									Command: []string{
+										"/usr/local/bin/status-reporter",
+									},
+									Env: []corev1.EnvVar{
+										{
+											Name:  StorageClusterNamespaceEnvVar,
+											Value: cronJob.Namespace,
```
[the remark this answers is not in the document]
bindrad: What does self referencing env var mean in this case? can you give an example of what is expected?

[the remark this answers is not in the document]
nb-ohad: For example:
```
- name: NAMESPACE
          valueFrom:
                    fieldRef:
                              fieldPath: metadata.namespace										
```

services/provider/server/consumer.go
bindrad: Updated in the latest commit

services/provider/server/consumer.go
bindrad: Updated in the latest commit

services/provider/server/consumer.go
bindrad: Updated in the latest commit

services/provider/server/consumer.go
```diff
@@ -224,3 +226,46 @@ func (c *ocsConsumerManager) Get(ctx context.Context, id string) (*ocsv1alpha1.S
...
+	if err != nil {
+		return fmt.Errorf("failed to encode the payload: %v", err)
+	}
```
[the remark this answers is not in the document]
bindrad: Updated in the latest commit

services/provider/server/consumer.go
```diff
@@ -224,3 +226,46 @@ func (c *ocsConsumerManager) Get(ctx context.Context, id string) (*ocsv1alpha1.S
...
+	if !ok {
+		c.mutex.RUnlock()
+		klog.Warningf("no storageConsumer found with UID %q", id)
+		return nil
+	}
+	c.mutex.RUnlock()
+
+	consumerObj := &ocsv1alpha1.StorageConsumer{}
+	consumerObj.Name = consumerName
+	consumerObj.Namespace = c.namespace
```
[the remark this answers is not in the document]
bindrad: Updated in the latest commit

services/consumer/main.go
```diff
@@ -0,0 +1,69 @@
...
+	config, err := config.GetConfig()
+	if err != nil {
+		klog.Exitf("failed to get config: %v", err)
+	}
+	cl, err := client.New(config, client.Options{Scheme: scheme})
+	if err != nil {
+		klog.Exitf("failed to create controller-runtime client: %v", err)
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*10)
```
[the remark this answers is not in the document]
bindrad: Updated in the latest commit

services/consumer/main.go
```diff
@@ -0,0 +1,69 @@
...
+
+	storageClusterList := &ocsv1.StorageClusterList{}
+	if err := cl.List(ctx, storageClusterList, client.InNamespace(namespace)); err != nil {
+		klog.Exitf("failed to list storageClusters: %v", err)
+	}
+
+	switch l := len(storageClusterList.Items); {
+	case l == 0:
+		klog.Exitf("no StorageCluster found")
+	case l != 1:
```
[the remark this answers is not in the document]
bindrad: Updated in the latest commit

services/consumer/main.go
```diff
@@ -0,0 +1,69 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/klog"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+)
+
+func main() {
+	scheme := runtime.NewScheme()
+	if err := ocsv1.AddToScheme(scheme); err != nil {
+		klog.Exitf("failed to add ocsv1 to scheme: %v", err)
```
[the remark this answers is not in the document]
bindrad: Updated in the latest commit

services/consumer/main.go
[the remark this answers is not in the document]
bindrad: Updated in the latest commit

services/consumer/main.go
```diff
@@ -0,0 +1,69 @@
...
+		klog.Exitf("failed to create controller-runtime client: %v", err)
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*10)
+	defer cancel()
+
+	namespace, isSet := os.LookupEnv(storagecluster.StorageClusterNamespaceEnvVar)
+	if !isSet {
+		klog.Exit("STORAGE_CLUSTER_NAMESPACE env var not set")
+	}
```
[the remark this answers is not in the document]
bindrad: Updated in the latest commit

controllers/storagecluster/external_ocs.go
```diff
@@ -35,6 +38,10 @@ const (
 	AcknowledgeOnboarding = "AcknowledgeOnboarding"
 )
 
+const (
+	StorageClusterNamespaceEnvVar = "STORAGE_CLUSTER_NAMESPACE"
```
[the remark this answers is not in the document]
bindrad: Updated in the latest commit

controllers/storagecluster/external_ocs.go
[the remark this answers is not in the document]
bindrad: Thanks for example, updated in the latest commit.

services/provider/server/consumer.go
```diff
@@ -224,3 +226,40 @@ func (c *ocsConsumerManager) Get(ctx context.Context, id string) (*ocsv1alpha1.S
...
+		Value interface{} `json:"value"`
+	}{
+		Op:    "replace",
+		Path:  "/status/lastHeartbeat",
+		Value: metav1.Now(),
+	}
+	jsonPatchInfo, _ := json.Marshal([]interface{}{patchInfo})
+	patch := client.RawPatch(types.JSONPatchType, jsonPatchInfo)
+
+	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
nb-ohad, line 254: I don't think we need RetryOnCOnflict here as there cannot be conflicts when we use patch as the entire operation is atomic to the k8s API.

nb-ohad (changes requested):

services/provider/server/consumer.go
```diff
@@ -224,3 +226,40 @@ func (c *ocsConsumerManager) Get(ctx context.Context, id string) (*ocsv1alpha1.S
...
+		consumerObj := &ocsv1alpha1.StorageConsumer{}
+		consumerObj.Name = consumerName
+		consumerObj.Namespace = c.namespace
+		return c.client.Status().Patch(ctx, consumerObj, patch)
+	})
+	if err != nil {
+		return fmt.Errorf("Failed to patch storageConsumer status: %v", err)
+	}
+	klog.Infof("successfully updated lastHeartbeat field in the StorageConsumer resource")
```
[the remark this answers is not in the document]
bindrad: Updated in the latest commit

services/provider/server/consumer.go
```diff
@@ -224,3 +226,40 @@ func (c *ocsConsumerManager) Get(ctx context.Context, id string) (*ocsv1alpha1.S
...
+	patch := client.RawPatch(types.JSONPatchType, jsonPatchInfo)
+
+	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		consumerObj := &ocsv1alpha1.StorageConsumer{}
+		consumerObj.Name = consumerName
+		consumerObj.Namespace = c.namespace
+		return c.client.Status().Patch(ctx, consumerObj, patch)
+	})
+	if err != nil {
+		return fmt.Errorf("Failed to patch storageConsumer status: %v", err)
```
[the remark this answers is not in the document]
bindrad: Updated in the latest commit

services/provider/server/consumer.go
bindrad: Updated in the latest commit

nb-ohad:
@bindrad PR looks good. Please squash before we merge

bindrad:
@nb-ohad Squashed and rebased the PR

iamniting (approved):
/lgtm

PR #1740, opened 2022-06-29
Enable rbd-mirror daemon restart config automatically

malayparida2000:
Currently, the user has to set this value ceph config set global rbd_mirror_die_after_seconds 3600 via the toolbox,
and that is not a good way for the customer as they don't have toolbox access. So setting this value in the rook-config-override 
configmap.

Note- it will work on fresh clusters not on the upgraded cluster as someone needs to restart rook pods after this configuration is applied

BZ Link-
[https://bugzilla.redhat.com/show_bug.cgi?id=2093266](https://bugzilla.redhat.com/show_bug.cgi?id=2093266)

### DIFF
--- a/controllers/storagecluster/cephconfig.go
+++ b/controllers/storagecluster/cephconfig.go
@@ -28,6 +28,7 @@ const (
 var (
 	defaultRookConfigData = `
 [global]
+rbd_mirror_die_after_seconds = 3600
 bdev_flock_retry = 20
 mon_osd_full_ratio = .85
 mon_osd_backfillfull_ratio = .8


idryomov (approved):
Ack from the RBD perspective

ShyamsundarR (approved):

controllers/storagecluster/cephconfig.go
```diff
@@ -37,6 +38,18 @@ mon_pg_warn_max_object_skew = 0
...
+debug_rbd = 20
+debug_rbd_mirror = 30
+log_file = /var/log/ceph/$cluster-$name.log
+[client.rbd-mirror-peer]
+debug_ms = 1
+debug_rbd = 20
+debug_rbd_mirror = 30
+log_file = /var/log/ceph/$cluster-$name.log
+[mgr]
+mgr/rbd_support/log_level = debug
```
idryomov, line 52: Did you verify that these settings take effect -- `mgr/rbd_support/log_level` in particular?  In the BZ @travisn said that `defaultRookConfigData` doesn't currently use `ceph config assimilate-conf`, so this is not going to work.

idryomov: The following was suggested instead:

> [...] a setting in the mon store can be applied in Rook inside the postMonStartupActions() method [1] (only in the downstream branch such as 4.11).
> The code would look like: 
>```
> if err := monStore.Set("mgr", "mgr/rbd_support/log_level", "debug"); err != nil {>
>  logger.Errorf("failed to set rbd log level to debug. %v", err)
> }
>```
> [1] https://github.com/red-hat-storage/rook/blob/release-4.11/pkg/operator/ceph/cluster/cluster.go#L489

idryomov (changes requested):

agarwal-mudit (approved):
/lgtm

agarwal-mudit:
/hold cancel

We will have separate PR for the debugging related changes

PR #1744, opened 2022-06-30
odf to odf: do not use the existing service for loadbalancer service

iamniting:
Existing node port svc use nodeport 31659, Changing the existing one to
loadbalancer keep using nodeport 31659 and while creating a new nodeport
svc with the same nodeport creates a conflict.

keep the existing nodeport svc as it is and create a new one with
loadbalancer type.

Error seen:
nodePort: Invalid value: 31659: provided port is already allocated

https://bugzilla.redhat.com/show_bug.cgi?id=2099958


iamniting:
close in favor of https://github.com/red-hat-storage/ocs-operator/pull/1748

PR #1753, opened 2022-07-18
odf-to-odf: wrong parameter in storageclass

subhamkrai:
when checking the existing storageclass the parameter
mentioned was wrong instead of this `existing.Parameters["encryptionMethod"]` it will be `existing.Parameters["encryptionKMSID"]`

https://github.com/rook/rook/blob/master/deploy/examples/csi/rbd/storageclass.yaml#L57

### DIFF
--- a/controllers/storageclassclaim/storageclassclaim_controller.go
+++ b/controllers/storageclassclaim/storageclassclaim_controller.go
@@ -54,6 +54,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
+const (
+	storageClassEncryptionParamKey = "encryptionKMSID"
+)
+
 // StorageClassClaimReconciler reconciles a StorageClassClaim object
 // nolint
 type StorageClassClaimReconciler struct {
@@ -223,7 +227,8 @@ func (r *StorageClassClaimReconciler) reconcileConsumerPhases() (reconcile.Resul
 			sccType := r.storageClassClaim.Spec.Type
 			sccEncryptionMethod := r.storageClassClaim.Spec.EncryptionMethod
 			_, scIsFSType := existing.Parameters["fsName"]
-			scEncryptionMethod, scHasEncryptionMethod := existing.Parameters["encryptionMethod"]
+
+			scEncryptionMethod, scHasEncryptionMethod := existing.Parameters[storageClassEncryptionParamKey]
 
 			validStorageClassConfig := true
 			if sccType == "sharedfilesystem" {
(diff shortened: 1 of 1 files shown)


controllers/storageclassclaim/storageclassclaim_controller.go
```diff
@@ -223,7 +223,7 @@ func (r *StorageClassClaimReconciler) reconcileConsumerPhases() (reconcile.Resul
 			sccType := r.storageClassClaim.Spec.Type
 			sccEncryptionMethod := r.storageClassClaim.Spec.EncryptionMethod
 			_, scIsFSType := existing.Parameters["fsName"]
-			scEncryptionMethod, scHasEncryptionMethod := existing.Parameters["encryptionMethod"]
+			scEncryptionMethod, scHasEncryptionMethod := existing.Parameters["encryptionKMSID"]
```
Rakshith-R, line 227: A suggestion
Let's create constants for such parameter keys to avoid such confusion ?

controllers/storageclassclaim/storageclassclaim_controller.go
```diff
@@ -252,7 +252,7 @@ func (r *StorageClassClaimReconciler) reconcileConsumerPhases() (reconcile.Resul
 				r.log.Error(fmt.Errorf("storageClassClaim is not compatible with existing StorageClass"),
 					"StorageClassClaim validation failed.")
 				r.storageClassClaim.Status.Phase = v1alpha1.StorageClassClaimFailed
-				return reconcile.Result{}, nil
+				return reconcile.Result{}, fmt.Errorf("failed to validate storageclassclaim, %t %s %s", scHasEncryptionMethod, sccEncryptionMethod, scEncryptionMethod)
```
Rakshith-R, line 255: The reason maybe a mismatch in fs/block types too ?

Rakshith-R (changes requested):

controllers/storageclassclaim/storageclassclaim_controller.go
nb-ohad, line 255: This change is wrong it will trigger a reconcile, while no reconcile can help in this case. Please revert.

nb-ohad (changes requested):

controllers/storageclassclaim/storageclassclaim_controller.go
Rakshith-R: let's add error logs, right next to the lines where the invalid condition is met.

subhamkrai: reverted but added the log in above since it will help in debug error from next time

controllers/storageclassclaim/storageclassclaim_controller.go
```diff
@@ -54,6 +54,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
+const (
+	scParameterForEncryption = "encryptionKMSID"
```
nb-ohad, line 58:
```suggestion
	storageClassEncryptionParamKey = "encryptionKMSID"
```

nb-ohad (changes requested):

PR #1755, opened 2022-07-21
add metrics for total objects and bucket count

weirdwiz:
this commit adds metrics for total number of objects in the rgw buckets
created by the user, also adds the number of object bucket.

Jira Issue: https://issues.redhat.com/browse/RHSTOR-3471


umangachapagain (changes requested):
Split into 2 commits: 1 for metrics and 1 for rules.
Rest looks good.

umangachapagain (approved):

PR #1758, opened 2022-07-27
Additional rules required for external clusters

ezio-auditore:
Require job:kube_pv:count metric from external ceph clusters

### DIFF
--- a/controllers/storagecluster/prometheus/externalcephrules.yaml
+++ b/controllers/storagecluster/prometheus/externalcephrules.yaml
@@ -8,6 +8,11 @@ metadata:
   namespace: rook-ceph
 spec:
   groups:
+  - name: ceph.rules
+    rules:
+    - expr: |
+        count(kube_persistentvolume_info * on (storageclass)  group_left(provisioner) kube_storageclass_info {provisioner=~"(.*rbd.csi.ceph.com)|(.*cephfs.csi.ceph.com)"})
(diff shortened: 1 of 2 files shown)


ezio-auditore:
ocs-operator-ci runs fine in local
logs : https://pastebin.com/8DG0A6H2
@jarrpa please advise

controllers/storagecluster/prometheus/externalcephrules.yaml
```diff
@@ -8,6 +8,11 @@ metadata:
   namespace: rook-ceph
 spec:
   groups:
+  - name: ceph.rules
+    rules:
+    - expr: |
+        count(kube_persistentvolume_info * on (storageclass)  group_left(provisioner) kube_storageclass_info {provisioner=~"(.*rbd.csi.ceph.com)|(.*cephfs.csi.ceph.com)"})
+      record: job:kube_pv:count
```
weirdwiz, line 15: https://github.com/red-hat-storage/ocs-operator/blob/1e4c078fed92ed0fb14d09e8e4ad85fe2a869a99/controllers/storagecluster/prometheus/localcephrules.yaml#L25

A query like this already exists, would this not apply to your use case?

ezio-auditore: It exists for localCephrules aka for provider clusters. we need the same for the external mode clusters as well

PR #1761, opened 2022-08-01
Fix utilization values mismatch queries

aruniiird:
Fixed latency query and removed the previous (not-needed) fix.

### DIFF
--- a/metrics/mixin/rules/rules.libsonnet
+++ b/metrics/mixin/rules/rules.libsonnet
@@ -36,7 +36,6 @@
       },
       {
         name: 'ODF_standardized_metrics.rules',
-        interval: '30s',
         rules: [
           {
             record: 'odf_system_health_status',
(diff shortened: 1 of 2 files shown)


agarwal-mudit:
@aruniiird Do we have a BZ for this PR? What is the priority?
/retest

bipuladh:
/retitle Bug 2026144: Fix utilization values mismatch queries

agarwal-mudit:
Fix is required for https://bugzilla.redhat.com/show_bug.cgi?id=2026144
However, the bug number is not required in the title for main branch fixes.

metrics/mixin/rules/rules.libsonnet
```diff
@@ -36,7 +36,6 @@
       },
       {
         name: 'ODF_standardized_metrics.rules',
-        interval: '30s',
```
umangachapagain, line 39: Why do we need to remove this interval?

metrics/mixin/rules/rules.libsonnet
```diff
@@ -91,7 +90,7 @@
           {
             record: 'odf_system_latency_seconds',
             expr: |||
-              avg by (namespace, managedBy, job, service)
+              sum by (namespace, managedBy, job, service)
```
umangachapagain, line 93: Can you explain why changing "avg" to "sum" and removing division by 2 works here?
Seems completely different from original query.

metrics/mixin/rules/rules.libsonnet
aruniiird: The BZ for which this PR is initiated is actually a re-opened BZ.
The line which is removed here was the previous fix which was not effective (and QE had to re-open the BZ again) and it is not needed now

umangachapagain (approved):

PR #1764, opened 2022-08-04
odf-to-odf: Report status of consumer storageCluster to provider

bindrad:
Heartbeat cronJob will report the status of consumer storageCluster to the provider cluster and the status will get stored in the corresponding storageConsumer CR.


iamniting (approved):
Changes look good to me, We just need to verify the backward compatibility when the server and client both are not on the same version.

api/v1alpha1/storageconsumer_types.go
```diff
@@ -58,6 +58,13 @@ type CephResourcesSpec struct {
 	CephClients map[string]string `json:"cephClients,omitempty"`
 }
 
+type ConsumerClusterStatus struct {
+	// Timestamp of last heartbeat received from consumer
+	LastHeartbeat metav1.Time `json:"lastHeartbeat,omitempty"`
+	// Phase of corresponding consumer storageCluseter
+	ExternalStorageClusterPhase string `json:"externalStorageClusterPhase,omitempty"`
```
Madhu-1, line 65: a consumer is an external client to the provider? just call this as StorageClusterStatus.

api/v1alpha1/storageconsumer_types.go
```diff
@@ -58,6 +58,13 @@ type CephResourcesSpec struct {
 	CephClients map[string]string `json:"cephClients,omitempty"`
 }
 
+type ConsumerClusterStatus struct {
+	// Timestamp of last heartbeat received from consumer
+	LastHeartbeat metav1.Time `json:"lastHeartbeat,omitempty"`
+	// Phase of corresponding consumer storageCluseter
```
Madhu-1, line 64: `storageCluseter` to `storageCluster`

api/v1alpha1/storageconsumer_types.go
bindrad: Updated in latest commit

api/v1alpha1/storageconsumer_types.go
bindrad: Updated in latest commit

PR #1768, opened 2022-08-10
Initialize rookCephOperatorCM if data is nil

bindrad:

### DIFF
--- a/controllers/storagecluster/external_resources.go
+++ b/controllers/storagecluster/external_resources.go
@@ -70,12 +70,12 @@ func (r *StorageClusterReconciler) setRookCSICephFS(
 		return err
 	}
 	enableDisableFlagStr := fmt.Sprintf("%v", enableDisableFlag)
+	if rookCephOperatorConfig.Data == nil {
+		rookCephOperatorConfig.Data = map[string]string{}
+	}
 	// if the current state of 'ROOK_CSI_ENABLE_CEPHFS' flag is same, just return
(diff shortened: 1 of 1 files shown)


Madhu-1 (approved):

controllers/storagecluster/external_resources.go
```diff
@@ -74,6 +74,9 @@ func (r *StorageClusterReconciler) setRookCSICephFS(
 	if rookCephOperatorConfig.Data[rookEnableCephFSCSIKey] == enableDisableFlagStr {
 		return nil
 	}
+	if rookCephOperatorConfig.Data == nil {
+		rookCephOperatorConfig.Data = map[string]string{}
+	}
```
nb-ohad, line 79: I think you mislocated the check - line 74 is already trying to access the map.

nb-ohad (changes requested):

controllers/storagecluster/external_resources.go
bindrad: It should work on line 77 as well because on line 74 it is a comparison and not an assignment but I have updated in the latest commit

Madhu-1: checking nil map works the map should not be empty when you are trying to assign it.

nb-ohad: Just checked and you guys are right.
But to me, it seems like a quirk of the language.

PR #1770, opened 2022-08-16
Initialize the rookcephoperatorconfig CM while creation & cleanup externalresources test

malayparida2000:
We recently merged [ocs-operator/pull/1648](https://github.com/red-hat-storage/ocs-operator/pull/1648), in which we moved the configurations from the CM to csv env. But we forgot to initialize the CM while it's being created, which may create unexpected behavior elsewhere.

And we also forgot to clean up the externalresources_test, which had tests for those variables which we moved.


jarrpa (approved):
/lgtm

PR #1773, opened 2022-08-18
Add two new variables for Ceph-CSI to set as metadata on RBD & CephFS

malayparida2000:
Refers-https://issues.redhat.com/browse/RHSTOR-3269

**Add two new variables for Ceph-CSI to set as metadata on RBD & CephFS**
     
    CSI_ENABLE_METADATA variable is set directly & CSI_CLUSTER_NAME
    is set from a configmap named ocs-operator-config.
    Both of them can be overridden by editing the CM
    rook-ceph-operator-config to add these value & setting it to any
    required value.
    
**Add new configmap ocs-ocsoperator-config for pure ocs-operator use**
    
    This configmap is used only for configuration in the ocs-operator.
    Customer changes should not be made on this. This Can be useful in
    cases where we have to get something during the runtime and set
    that somewhere.

### DIFF
--- a/controllers/ocsinitialization/ocsinitialization_controller.go
+++ b/controllers/ocsinitialization/ocsinitialization_controller.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/go-logr/logr"
+	configv1 "github.com/openshift/api/config/v1"
 	secv1client "github.com/openshift/client-go/security/clientset/versioned/typed/security/v1"
 	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v1"
 	"github.com/red-hat-storage/ocs-operator/controllers/util"
@@ -17,7 +18,9 @@ import (
 	"k8s.io/klog/v2"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
 // watchNamespace is the namespace the operator is watching.
@@ -150,6 +153,12 @@ func (r *OCSInitializationReconciler) Reconcile(ctx context.Context, request rec
 		instance.Status.RookCephOperatorConfigCreated = true
 	}
 
+	err = r.ensureOCSOperatorConfig(instance)
+	if err != nil {
+		r.Log.Error(err, "Failed to process ConfigMap.", "ConfigMap", klog.KRef(instance.Namespace, "ocs-operator-config"))
+		return reconcile.Result{}, err
+	}
+
 	reason := ocsv1.ReconcileCompleted
 	message := ocsv1.ReconcileCompletedMessage
 	util.SetCompleteCondition(&instance.Status.Conditions, reason, message)
@@ -171,6 +180,23 @@ func (r *OCSInitializationReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&ocsv1.OCSInitialization{}).
 		Owns(&appsv1.Deployment{}).
+		Watches(&source.Kind{Type: &configv1.ClusterVersion{}}, handler.EnqueueRequestsFromMapFunc(func(obj client.Object) []reconcile.Request {
+			return []reconcile.Request{
+				{NamespacedName: types.NamespacedName{
+					Name:      InitNamespacedName().Name,
+					Namespace: InitNamespacedName().Namespace,
+				}},
+			}
+		})).
+		Watches(&source.Kind{Type: &corev1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      ocsOperatorConfigName,
+				Namespace: watchNamespace,
+			},
+		}}, &handler.EnqueueRequestForOwner{
+			OwnerType:    &ocsv1.OCSInitialization{},
+			IsController: true,
+		}).
 		Complete(r)
 }
 
--- a/controllers/ocsinitialization/ocsoperator_config.go
+++ b/controllers/ocsinitialization/ocsoperator_config.go
@@ -1,0 +1,78 @@
+package ocsinitialization
+
+import (
+	"context"
+	"fmt"
+
+	configv1 "github.com/openshift/api/config/v1"
+	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+const (
+	// This configmap is purely for the OCS operator to use
+	ocsOperatorConfigName = "ocs-operator-config"
+)
+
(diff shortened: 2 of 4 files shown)


controllers/ocsinitialization/ocsoperator_config.go
```diff
@@ -0,0 +1,126 @@
+package ocsinitialization
```
Madhu-1, line 1: @malayparida2000 i dont think this commit is required at all. you can set the required configuration in the Rook ENV in csv-merger.go

Madhu-1: never mind, i didnt notice that we cannot add runtime calculation to csi-merger.go

malayparida2000: Yes, And Jose suggested let's have a configmap specifically for OCS-Operator which can be useful in the future also, And let's keep the existing rook-ceph-operator-config CM just for customer use.

controllers/ocsinitialization/ocsoperator_config.go
```diff
@@ -0,0 +1,147 @@
...
+	err = r.updateClusterName(ocsOperatorConfig)
+	if err != nil {
+		return err
+	}
+	// restart the rook-ceph-operator pod if any change has occured to the configmap
+	err = r.Client.Get(context.TODO(), types.NamespacedName{Name: ocsOperatorConfigName, Namespace: initialData.Namespace}, ocsOperatorConfig)
+	if err != nil {
+		return err
+	}
+	if resourceVersion != ocsOperatorConfig.ResourceVersion {
```
jarrpa, line 47: I don't think you need to do all this. You already know in `updateClusterName()` if you had to make any changes, so just use that to determine if you need to restart the operator Pod. Also, since you're only calling it once, it probably doesn't need to be a function. Just restart the operator Pod any time you have to issue an `Update()`.

controllers/ocsinitialization/ocsoperator_config.go
```diff
@@ -0,0 +1,147 @@
...
+		r.Log.Error(err, "Failed to get the clusterVersion version of the OCP cluster")
+		return ""
+	}
+	return fmt.Sprint(clusterVersion.Spec.ClusterID)
+}
+
+// RestartPod deletes a pod marked by a given label
+// It then waits for a given timeout for the pod to resume running state
+func (r *OCSInitializationReconciler) restartRookCephOperatorPod(namespace string) error {
+	pod, err := r.getPodByLabel("app", "rook-ceph-operator", namespace)
```
jarrpa, line 98: This whole function is unnecessary, there's no reason to try and make this a generic function when it's only used twice to get the same thing. In particular, you're already making some assumptions in that you're expecting only one Pod in the list, which I don't think you need to worry about at all. In general, I only see the following things that need to be done:

```
if configMapUpdated {
  List(podList, opts)
  Delete(podList.Items[0])

  PollImmediate(stuff, func() {
    Get(rookCephOperatorDeployment)
    if rookCephOperatorDeployment.Status.UnavailableReplicas != 0 {
      Log.Error("This could be a problem...")
      // Don't do `return err`, we can just advise it's not ready and continue reconciling other stuff.
    }
  })
}

return nil
```

In particular for 4, there's no need to return an error, we just want to mention in the logs that this is probably a bad thing, but still continue reconciliation .

jarrpa (changes requested):
The overall implementation is sound, but I think it can be made simpler and easier to read.

controllers/ocsinitialization/ocsoperator_config.go
malayparida2000: I made this because in the future we might need to add some other values to the configmap just like this cluster-ID, 
And there might be a case where someone may have manually edited the CM via oc edit.

In both cases, the rook-ceph-operator pod should restart to get the new configurations.

controllers/ocsinitialization/ocsoperator_config.go
malayparida2000: Done

controllers/ocsinitialization/ocsoperator_config.go
jarrpa: 1. Don't write functions "for the future". Write simple, readable code that will be easy to move into a function *if* it's ever needed.
2. We are *explicitly* not allowing anyone else to edit this ConfigMap. Just add a `Watch()` to the controller for ConfigMaps, preferably filtered by just ours, and we'll get it every time.

controllers/ocsinitialization/ocsoperator_config.go
```diff
@@ -0,0 +1,142 @@
...
+	resourceVersion := ocsOperatorConfig.ResourceVersion
+
+	//Adding Ownerreference to the configmap if it's not present
+	ownerRefs := ocsOperatorConfig.GetOwnerReferences()
+	if len(ownerRefs) == 0 {
+		err = controllerutil.SetOwnerReference(initialData, ocsOperatorConfig, r.Client.Scheme())
+		if err != nil {
+			return err
+		}
+		err = r.Client.Update(context.TODO(), ocsOperatorConfig)
```
jarrpa, line 46: I see no need to do this as a separate `Update()` call... just use `CreateOrUpdate()` and pass all the required mutations (ownerRef, clusterId update) there.

controllers/ocsinitialization/ocsoperator_config.go
```diff
@@ -0,0 +1,142 @@
...
+		opts := []client.ListOption{
+			client.InNamespace(initialData.Namespace),
+			client.MatchingLabels(map[string]string{"key": "value"}),
+		}
+		err = r.Client.List(context.TODO(), podList, opts...)
+		if err != nil {
+			r.Log.Error(err, "Failed to list rook-ceph-operator pods")
+			return nil
+		}
+		err = r.Client.Delete(context.TODO(), &podList.Items[0])
```
travisn, line 72: Restarting the rook operator pod should not be necessary. Rook has a watcher on this configmap and it will update the CSI driver every time the configmap is updated. So I believe we're already covered to pick up these settings in the configmap if they change.

malayparida2000: But Travis this one is not the rook-ceph-operator configmap, this is a new CM that we are introducing. Rook-ceph-operator will set an env variable by looking here, so I believe unless restarted the new value for env var won't come.

controllers/ocsinitialization/ocsoperator_config.go
malayparida2000: Addressed now

controllers/ocsinitialization/ocsoperator_config.go
malayparida2000: Addressed this one

jarrpa (approved):
/lgtm

controllers/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -171,6 +180,23 @@ func (r *OCSInitializationReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&ocsv1.OCSInitialization{}).
 		Owns(&appsv1.Deployment{}).
+		Watches(&source.Kind{Type: &configv1.ClusterVersion{}}, handler.EnqueueRequestsFromMapFunc(func(obj client.Object) []reconcile.Request {
```
parth-gr, line 183: @malayparida2000 is there a need for a watcher here??

malayparida2000: If somehow the clusterversion CR changes(someone gives a new name for the cluster ID, We should catch that), Thus the watcher

Madhu-1: I dont think that's the case; the clusterID will always remain the same for the entire cluster lifecycle. as its an OCP cluster-ID.

malayparida2000: The spec.clusterID field for a clusterversion cr is actually editable, So someone can actually edit it if they want.

PR #1774, opened 2022-08-22
Update local ceph rules' annotations

rexagod:
Update local ceph rule file based on the latest Ceph master.

### DIFF
--- a/controllers/storagecluster/prometheus/localcephrules.yaml
+++ b/controllers/storagecluster/prometheus/localcephrules.yaml
@@ -46,6 +46,7 @@ spec:
       for: 5m
       labels:
         severity: critical
+        documentation: "https://access.redhat.com/documentation/en-us/red_hat_openshift_data_foundation/4.12/html-single/troubleshooting_openshift_data_foundation/index#resolving-alerts-and-errors-6.1.1_rhodf"
     - alert: CephMgrIsMissingReplicas
       annotations:
         description: Ceph Manager is missing replicas.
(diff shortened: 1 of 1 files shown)


umangachapagain (commented):
We do not just download upstream rules and apply it downstream.

We have a separate project [ceph-mixins](https://github.com/ceph/ceph-mixins) where rules are added or https://github.com/red-hat-storage/ocs-operator/tree/main/metrics/mixin for ocs specific metrics.

The final generated YAML is a K8s object with rules that match OCS deployment. Upstream Ceph rules may not apply as it is.

rexagod:
> Upstream Ceph rules may not apply as it is.

ACK, so does this mean that I should point https://github.com/red-hat-storage/ocs-operator/pull/1774/files#diff-ffea8b477bf82a7034c97d82ad37e9d0a6ace823b7ec27263475fbd423d48095R10 to https://raw.githubusercontent.com/ceph/ceph-mixins/master/extras/manifests/prometheus-ceph-rules.yaml?

rexagod:
But I'm assuming someone (or something) already makes sure that the ceph-mixins alerts are replicated here. If that's the case, we don't need this PR.

rexagod:
@umangachapagain Does the above comment sound good to you?

rexagod:
We'd still need documentation links present here in any case, so updated the local ceph rules file with the *predicted* links.

umangachapagain (commented):
Is it expected for these links to throw a 404 error?

rexagod:
@umangachapagain Yes. These are placeholder links that will be updated by the documentation team before the GA happens in December.

rexagod:
/cc @umangachapagain
Hello Umanga, please review, thanks.

umangachapagain:
> /cc @umangachapagain Hello Umanga, please review, thanks.

Yeah I'm wondering if we should merge a PR that adds broken (as of now) links. Or, if we can wait until the links actually work.

PR #1779, opened 2022-08-23
Bug 2119513: [release 4.8] expose option to force remove the OSD in osd…

subhamkrai:
expose option to force remove the OSD in osd removal template

(cherry picked from commit 9d1e544c033b1232d67bd483091982871898daab)


subhamkrai:
added new patch since ci was failing because of
 ```
hack/verify-generated.sh
diff --git a/api/v1/zz_generated.deepcopy.go b/api/v1/zz_generated.deepcopy.go
index 26525244..d460b37b 100644
--- a/api/v1/zz_generated.deepcopy.go
+++ b/api/v1/zz_generated.deepcopy.go
@@ -1,4 +1,3 @@
-//go:build !ignore_autogenerated
 // +build !ignore_autogenerated
 
 /*
uncommitted generated files. run 'make update-generated' and commit results.
make: *** [Makefile:180: verify-generated] Error 1
```

File: api/v1/zz_generated.deepcopy.go [hunk 1]
```diff
@@ -1,4 +1,3 @@
-//go:build !ignore_autogenerated
```
travisn, line 1: Did just your local build require this update? The downstream build should still be building with the older golang version so I'd suggest reverting it

subhamkrai: actually, the lastest go is adding this IMO, before my commit one the ci was failing with 
```
hack/verify-generated.sh
diff --git a/api/v1/zz_generated.deepcopy.go b/api/v1/zz_generated.deepcopy.go
index 26525244..d460b37b 100644
--- a/api/v1/zz_generated.deepcopy.go
+++ b/api/v1/zz_generated.deepcopy.go
@@ -1,4 +1,3 @@
-//go:build !ignore_autogenerated
// +build !ignore_autogenerated

/*
uncommitted generated files. run 'make update-generated' and commit results.
make: *** [Makefile:180: verify-generated] Error 1
```

PR #1780, opened 2022-08-23
Bug 2119516:[release-4.7]expose option to force remove the OSD in osd

subhamkrai:


subhamkrai:
cc @iamniting

travisn (changes requested):
@subhamkrai Are the changes other than reconcile.go intended to be pulled with this PR? It looks unrelated

subhamkrai:
> @subhamkrai Are the changes other than reconcile.go intended to be pulled with this PR? It looks unrelated

yes, before that the CI was failing, so I @iamniting came up with that commits.

PR #1781, opened 2022-08-23
Bug 2119517: [release 4.6] expose option to force remove the OSD in osd

subhamkrai:
adding option to expose --force-osd-removal <true/false>.


subhamkrai:
/assign @travisn 

cc @agarwal-mudit

PR #1782, opened 2022-08-24
Ceph Non-resilient Pools - Dev Preview

malayparida2000:
JIRA-https://issues.redhat.com/browse/RHSTOR-3719

**Goals-**
_Applications like Cloudera, MongoDB, and others manage their resiliency on the application level.
Storage efficiency is very important for our customers, so we need to stop having unnecessary resilience.
So we have to provide a storage solution that offers Ceph replica 1 for RBD, avoiding unnecessary replication when not required by the application._

**Add a flag to storagecluster CR to enable Ceph Non-Resilient pools**
    
    This flag can be set by simple oc patch command,
    oc patch storagecluster ocs-storagecluster -n openshift-storage
    --type json --patch '[{ "op": "replace", "path":
    "/spec/managedResources/cephNonResilientPools/enable", "value": true }]'

**Turn off health warning for pools with no redundancy**
    
    Ceph will issue a health warning if a pool is configured with no
    redundancy. As we are going to have pools with no redundancy when
    Non-Resilient pools are enabled, we need to turn off this warning

**Create 1 cephblockpool per failureDomain(zone/node)**
    
    One pool is created per node/zone depending on if it's a cloud
    provider cluster or a local cluster. Each pool will specify the
    deviceClass with that node/zone name.

**Add <deviceClass : replicated> to existing default pools**
    
    The default pools created by the OCS operator (rbd, cephfs,
    and object store data pools) have <deviceClass: replicated> when
    Non-Resilient pools are enabled. This is done so that Ceph can
    differentiate placing data on different sets of OSDs.

**Create a new storageclass for the topology constrained pools**
    
    The replica-1 pools are added as "topologyConstrainedPools" in the
    parameter section of the storageclass. The name for the new
    storageclass would be <storageClusterName>-ceph-non-resilient-rbd.

**Create one storageClassDeviceSet for each failureDomain**
    
    Each replica-1 storageClassDeviceSet has the name of its failureDomain
    (node/zone), and an annotation crushDeviceClass : <node/zone_name>.
    The created osd has NodeAffinity to its zone/node.

**Add <deviceClass : replicated> to existing deviceset**
    
    The existing storageClassDeviceSets have the added annotation
    <crushDeviceClass: replicated> when Non-Resilient pools are enabled.

**Autogenerated changes to API & YAML files**


api/v1/storagecluster_types.go
```diff
@@ -133,6 +136,14 @@ type ManageCephBlockPools struct {
 	DisableSnapshotClass bool   `json:"disableSnapshotClass,omitempty"`
 }
 
+// ManageReplica1CephBlockPools defines how to reconcilea CephBlockPools
+type Replica1CephBlockPoolsSpec struct {
```
travisn, line 140: How about just name it for replica 1 so other settings besides pools can be added in the future
```suggestion
type Replica1Spec struct {
```

api/v1/storagecluster_types.go
```diff
@@ -50,6 +50,9 @@ type StorageClusterSpec struct {
 	MonDataDirHostPath string                                 `json:"monDataDirHostPath,omitempty"`
 	MultiCloudGateway  *MultiCloudGatewaySpec                 `json:"multiCloudGateway,omitempty"`
 	NFS                *NFSSpec                               `json:"nfs,omitempty"`
+	// Applications consuming ODF should have the option to avoid replication
+	// of the storage if they don’t need storage redundancy
+	Replica1CephBlockPools *Replica1CephBlockPoolsSpec `json:"replica1CephBlockPools,omitempty"`
```
travisn, line 55:
```suggestion
	Replica1 *Replica1CephBlockPoolsSpec `json:"replica1CephBlockPools,omitempty"`
```

controllers/storagecluster/cephcluster.go
```diff
@@ -727,12 +727,8 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster, serverVersion *version.
 			}
 
 			// Annotation crushDeviceClass ensures osd with different CRUSH device class than the one detected by Ceph
-			crushDeviceClass := ds.DeviceType
```
travisn, line 730: There are still scenarios where other deviceClasses may be set. For now, let's leave the old behavior when replica 1 is disabled, and the new "replicated" device class when it is enabled. We will have to confirm with PM if this is sufficient for the other deviceClass scenarios.

controllers/storagecluster/storageclasses.go
```diff
@@ -355,3 +356,37 @@ func (r *StorageClusterReconciler) newStorageClassConfigurations(initData *ocsv1
 
 	return ret, nil
 }
+
+func gettopologyConstrainedPools(initData *ocsv1.StorageCluster) string {
```
travisn, line 360:
```suggestion
func getTopologyConstrainedPools(initData *ocsv1.StorageCluster) string {
```

controllers/storagecluster/storageclasses.go
```diff
@@ -355,3 +356,37 @@ func (r *StorageClusterReconciler) newStorageClassConfigurations(initData *ocsv1
...
+				PoolName: generateNameForReplica1CephBlockPool(initData, value),
+				DomainSegments: []topologySegment{
+					{
+						DomainLabel: failureDomain,
+						DomainValue: value,
+					},
+				},
+			})
+		}
+		// retunring as string as parameters are of type map[string]string
```
travisn, line 388:
```suggestion
		// returning as string as parameters are of type map[string]string
```

travisn (changes requested):

controllers/storagecluster/cephfilesystem.go
```diff
@@ -42,6 +42,7 @@ func (r *StorageClusterReconciler) newCephFilesystemInstances(initData *ocsv1.St
 				DataPools: []cephv1.NamedPoolSpec{
 					{
 						PoolSpec: cephv1.PoolSpec{
+							DeviceClass:   "replicated",
```
Madhu-1, line 45: please note this should not cause any issue for standalone upgraded clusters and managed service upgraded cluster and also for new managed service feature example https://github.com/red-hat-storage/ocs-operator/pull/1750
cc @nb-ohad

malayparida2000: ack, Will take a look

controllers/storagecluster/cephcluster.go
malayparida2000: done

controllers/storagecluster/generate.go
```diff
@@ -118,6 +133,17 @@ func generateCephReplicatedSpec(initData *ocsv1.StorageCluster, poolType string)
 	return crs
 }
 
+// generateCephReplica1Spec returns the ReplicatedSpec for the cephCluster
+// based on the StorageCluster configuration
+func generateCephReplica1Spec(initData *ocsv1.StorageCluster) cephv1.ReplicatedSpec {
```
travisn, line 138: nit: How about this method should generate the entire blockpool spec instead of just the replicated spec? Or if simpler, just set these properties in the caller and no need for this method
```suggestion
func generateReplica1BlockPoolSpec() cephv1.ReplicatedSpec {
```

controllers/storagecluster/generate.go
```diff
@@ -42,6 +42,17 @@ func generateNameForCephBlockPool(initData *ocsv1.StorageCluster) string {
 	return fmt.Sprintf("%s-cephblockpool", initData.Name)
 }
 
+func generateDeviceClass(initData *ocsv1.StorageCluster) string {
+	if initData.Spec.Replica1 != nil && initData.Spec.Replica1.Enable {
+		return "replicated"
```
travisn, line 47: `"replicated"` is used in several places, how about defining a const?

controllers/storagecluster/cephcluster.go
```diff
@@ -775,6 +779,66 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster, serverVersion *version.
 		}
 	}
 
+	if (sc.Spec.Replica1 != nil) && (sc.Spec.Replica1.Enable) {
+		failureDomain := sc.Status.FailureDomain
+		failureDomainKey := sc.Status.FailureDomainKey
+		failureDomainValues := sc.Status.FailureDomainValues
+
+		// Creating osds
+		for _, value := range failureDomainValues {
```
travisn, line 788:
```suggestion
		for _, deviceClass := range failureDomainValues {
```

controllers/storagecluster/storageclasses.go
```diff
@@ -355,3 +414,37 @@ func (r *StorageClusterReconciler) newStorageClassConfigurations(initData *ocsv1
...
+	failureDomain := initData.Status.FailureDomain
+	failureDomainValues := initData.Status.FailureDomainValues
+
+	type topologySegment struct {
+		DomainLabel string `json:"domainLabel"`
+		DomainValue string `json:"value"`
+	}
+
+	// TopologyConstrainedPool stores the pool name and a list of its associated topology domain values.
+	type constrainedPool struct {
```
travisn, line 428: Can we reference these types in the csi repo instead of redefining them here?

travisn (changes requested):

controllers/storagecluster/storageclasses.go
```diff
@@ -103,6 +104,27 @@ func (r *StorageClusterReconciler) createStorageClasses(sccs []StorageClassConfi
...
+		case strings.Contains(sc.Name, "-ceph-replica1-rbd") && !scc.isClusterExternal:
+			// wait for CephBlockPools to be ready
+			cephBlockPools := cephv1.CephBlockPoolList{}
+			err := r.Client.List(context.TODO(), &cephBlockPools, client.InNamespace(sc.Parameters["clusterID"]))
+			num := len(initData.Status.FailureDomainValues)
+			// if the number of CephBlockPools is less than the number of failure domains+1, skip reconciling StorageClass
+			if err != nil || len(cephBlockPools.Items) < num+1 {
+				skippedSC = append(skippedSC, sc.Name)
+				continue
+			}
```
jarrpa, line 116: This shouldn't be needed at all, because the total number of non-resilient pools should not depend on the number of resilient pools. What I would do is start with `num := 0`, then increment it when you loop over the block pool list and find a "Ready" non-resilient pool. After that is when you should check if the correct number of pools are ready.

controllers/storagecluster/storageclasses.go
```diff
@@ -103,6 +104,27 @@ func (r *StorageClusterReconciler) createStorageClasses(sccs []StorageClassConfi
 				skippedSC = append(skippedSC, sc.Name)
 				continue
 			}
+		case strings.Contains(sc.Name, "-ceph-replica1-rbd") && !scc.isClusterExternal:
+			// wait for CephBlockPools to be ready
+			cephBlockPools := cephv1.CephBlockPoolList{}
+			err := r.Client.List(context.TODO(), &cephBlockPools, client.InNamespace(sc.Parameters["clusterID"]))
+			num := len(initData.Status.FailureDomainValues)
```
jarrpa, line 111: Rather than passing in the StorageCluster spec at all, just read the information from the StorageClass parameters:

```suggestion
		num := len(scc.StorageClass.Parameters["topologyConstrainedPools"])
```

You probably need to do `strings.Split()` or something.

controllers/storagecluster/cephblockpools.go
```diff
@@ -61,6 +61,7 @@ func (r *StorageClusterReconciler) newCephBlockPoolInstances(initData *ocsv1.Sto
 			},
 			Spec: cephv1.NamedBlockPoolSpec{
 				PoolSpec: cephv1.PoolSpec{
+					DeviceClass:    generateDeviceClass(initData),
```
jarrpa, line 64: I don't think this needs to be a function at all. There's no reason to set the DeviceClass for the replica 3 block pools. Or at least, so I assume. @travisn should probably confirm. Really this ties into the fact that we don't want a global boolean for all CephBlockPools, so it can just be removed entirely.

controllers/storagecluster/storageclasses.go
jarrpa: What types are you referring to? Is `constrainedPool` a struct used in the Ceph-CSI repo?

jarrpa (changes requested):
While the specifics on generating the right configurations seem appropriate, having a top-level boolean to control pool resiliency across *all* generated block pools seems counter-intuitive. We treat these nen-resilient pools as a new kind of managed resource, independent from the existing default CephBlockPools.

controllers/storagecluster/cephblockpools.go
travisn: This is the one place where there is some side effect on resilient pools if the replica 1 feature is enabled... The `deviceClass: replicated` is needed on all of the resilient pools so that Ceph can differentiate placing data on different sets of OSDs. If we didn't add this property to the resilient pools, the resilient pools would place data on the OSDs that are meant to be reserved for replica 1 pools.

controllers/storagecluster/storageclasses.go
travisn: They are defined [here](https://github.com/ceph/ceph-csi/blob/1952a9b4b3ff4b00859d5b663d2f3d4a7d07542b/internal/util/topology.go#L129-L139), but since one of them is private (lower-case), it might not be worth asking for a change in the csi repo, and just keep them redefined here.

controllers/storagecluster/cephblockpools.go
jarrpa: Ah-ha... okay. That seems a little odd, but makes sense.

controllers/storagecluster/generate.go
jarrpa: I only see the string used twice, so I don't think that's worth defining a const. That said, does Rook have any defined values for device classes?

jarrpa (approved):
There's a few tiny things I could nag about, but everything looks good for an initial implementation. It's easily extensible if more requirements come in the future.

/lgtm

controllers/storagecluster/cephcluster.go
```diff
@@ -800,6 +804,62 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster, serverVersion *version.
...
+			ds.Count = 1
+			ds.Resources = defaults.DaemonResources["osd"]
+			// passing on existing defaults from existing devcicesets
+			ds.TuneSlowDeviceClass = sc.Spec.StorageDeviceSets[0].Config.TuneSlowDeviceClass
+			ds.TuneFastDeviceClass = sc.Spec.StorageDeviceSets[0].Config.TuneFastDeviceClass
+			annotations := map[string]string{
+				"crushDeviceClass": failureDomainValue,
+			}
+			// using the spec for volumeclaimtemplate from existing devicesets including the storageclass
+			volumeClaimTemplateSpec := storageClassDeviceSets[0].VolumeClaimTemplates[0].Spec
```
travisn, line 821: We really should allow for setting a different volumeClaimTemplate than what is being used by the replica pools. Local storage could have different properties, such as storage class from which to provision the storage, or different size of the storage. This support can come in a future PR after we confirm the requirement with PM, but I would think it's a requirement for GA.

travisn (approved):

PR #1786, opened 2022-08-29
Fix StorageCluster is in error state for few minutes after installing

malayparida2000:
Resolve: https://bugzilla.redhat.com/show_bug.cgi?id=2004027

**Stop setting storagecluster to error phase just after creation**

    While storageclasses are being created it's waiting till underlying
    resources like cephBlockPool, cephFilesystem & NFS are created.
    In this phase, we are returning an error which results in the phase of the
    storagecluster to be in Error Phase creating confusion for the user.
    
    So henceforth this particular error would be caught, A requeue
    Reconcile result would be given and the storagecluster would be marked
    as in the Progressing Phase.

### DIFF
--- a/controllers/storagecluster/reconcile.go
+++ b/controllers/storagecluster/reconcile.go
@@ -434,6 +434,12 @@ func (r *StorageClusterReconciler) reconcilePhases(
 			statusutil.SetErrorCondition(&instance.Status.Conditions, reason, message)
 
 			if instance.Status.Phase != statusutil.PhaseOnboarding {
+				// if the error was due to skipped storage classes
+				// instead of error phase we will set it to progressing
+				if strings.Contains(returnErr.Error(), storageClassSkippedError) {
+					instance.Status.Phase = statusutil.PhaseProgressing
+					return reconcile.Result{Requeue: true}, nil
+				}
 				instance.Status.Phase = statusutil.PhaseError
 			}
 
--- a/controllers/storagecluster/storageclasses.go
+++ b/controllers/storagecluster/storageclasses.go
@@ -18,6 +18,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
+const (
+	storageClassSkippedError = "some StorageClasses were skipped while waiting for pre-requisites to be met"
+)
+
 // StorageClassConfiguration provides configuration options for a StorageClass.
 type StorageClassConfiguration struct {
 	storageClass      *storagev1.StorageClass
@@ -169,7 +173,7 @@ func (r *StorageClusterReconciler) createStorageClasses(sccs []StorageClassConfi
 		}
 	}
 	if len(skippedSC) > 0 {
-		return fmt.Errorf("some StorageClasses [%s] were skipped while waiting for pre-requisites to be met", strings.Join(skippedSC, ","))
+		return fmt.Errorf("%s: [%s]", storageClassSkippedError, strings.Join(skippedSC, ","))
 	}
 	return nil
 }


controllers/storagecluster/reconcile.go
```diff
@@ -433,6 +433,15 @@ func (r *StorageClusterReconciler) reconcilePhases(
 			message := fmt.Sprintf("Error while reconciling: %v", returnErr)
 			statusutil.SetErrorCondition(&instance.Status.Conditions, reason, message)
 
+			// if the error was due to skipped storage classes, we shouldn't set the phase to error
+			// instead of error phase we will set it to initializing
+			if strings.Contains(returnErr.Error(), "some StorageClasses were skipped while waiting for pre-requisites to be met") {
+				if instance.Status.Phase != statusutil.PhaseOnboarding {
+					instance.Status.Phase = statusutil.PhaseInitializing
```
iamniting, line 440: Instead of PhaseInitializing can not we make use of progressing itself

iamniting (changes requested):

controllers/storagecluster/reconcile.go
malayparida2000: We can actually, I wanted to differentiate between these two phases actually. Basically, the Initializing phase suggests the storageclass creation is happening. That's why thought of the new phase. If you think the progressing phase will be suitable for this, then I can use that instead.

iamniting: Earlier we were seeing the progressing phase only when this bug was not present. So I would suggest having the progressing phase.

malayparida2000: Have Updated the PR, Please Take anotherLook

controllers/storagecluster/reconcile.go
```diff
@@ -433,6 +433,15 @@ func (r *StorageClusterReconciler) reconcilePhases(
...
+			if strings.Contains(returnErr.Error(), "some StorageClasses were skipped while waiting for pre-requisites to be met") {
+				if instance.Status.Phase != statusutil.PhaseOnboarding {
+					instance.Status.Phase = statusutil.PhaseProgressing
+				}
+				return reconcile.Result{Requeue: true}, nil
+			}
+
 			if instance.Status.Phase != statusutil.PhaseOnboarding {
 				instance.Status.Phase = statusutil.PhaseError
 			}
```
iamniting, line 447:
```suggestion
			if instance.Status.Phase != statusutil.PhaseOnboarding {
			        // if the error was due to skipped storage classes, we shouldn't set the phase to error
			        // instead of error phase we will set it to progressing
			        if strings.Contains(returnErr.Error(), "some StorageClasses were skipped while waiting for pre-requisites to be met") {         
			                instance.Status.Phase = statusutil.PhaseProgressing
			                return reconcile.Result{Requeue: true}, nil
			        }
				instance.Status.Phase = statusutil.PhaseError
			}
```
I think this would be a better way to write it.

malayparida2000: Yeah, I agree. Didn't think of this approach, but it looks cleaner. Thanks.
Have made the changes.

controllers/storagecluster/reconcile.go
```diff
@@ -434,6 +434,12 @@ func (r *StorageClusterReconciler) reconcilePhases(
 			statusutil.SetErrorCondition(&instance.Status.Conditions, reason, message)
 
 			if instance.Status.Phase != statusutil.PhaseOnboarding {
+				// if the error was due to skipped storage classes
+				// instead of error phase we will set it to progressing
+				if strings.Contains(returnErr.Error(), "some StorageClasses were skipped while waiting for pre-requisites to be met") {
```
iamniting, line 439: Thought: Can we create a constant error string and use the same at both places? It will reduce the chances of someone changing the error and having the bug again. WDYT?

iamniting (changes requested):

controllers/storagecluster/reconcile.go
malayparida2000: Hmm, Yeah agreed. It can be a good step towards avoiding this bug again. Made the change. PTAL.

controllers/storagecluster/storageclasses.go
```diff
@@ -19,6 +19,12 @@ import (
 )
 
 // StorageClassConfiguration provides configuration options for a StorageClass.
+
+const (
+	// StorageClassName is the name of the StorageClass.
+	storageClassSkippedError = "some StorageClasses were skipped while waiting for pre-requisites to be met"
+)
+
```
iamniting, line 27:
```suggestion
const (
	// StorageClassName is the name of the StorageClass.
	storageClassSkippedError = "some StorageClasses were skipped while waiting for pre-requisites to be met"
)

// StorageClassConfiguration provides configuration options for a StorageClass.
```

controllers/storagecluster/storageclasses.go
```diff
@@ -169,7 +175,7 @@ func (r *StorageClusterReconciler) createStorageClasses(sccs []StorageClassConfi
 		}
 	}
 	if len(skippedSC) > 0 {
-		return fmt.Errorf("some StorageClasses [%s] were skipped while waiting for pre-requisites to be met", strings.Join(skippedSC, ","))
+		return fmt.Errorf("%s : [%s]", storageClassSkippedError, strings.Join(skippedSC, ","))
```
iamniting, line 178:
```suggestion
		return fmt.Errorf("%s: [%s]", storageClassSkippedError, strings.Join(skippedSC, ","))
```

iamniting (changes requested):

controllers/storagecluster/storageclasses.go
malayparida2000: Yeah, messed this up, Fixed now.

controllers/storagecluster/storageclasses.go
malayparida2000: Fixed

iamniting (approved):
/lgtm

malayparida2000:
@agarwal-mudit need approve

umangachapagain (approved):

PR #1787, opened 2022-08-29
Bug 2117146: rook: add logCollector in storageCluster

subhamkrai:
rook: upgrade rook to v1.9.9

rook: add logCollector in storageCluster

adding logCollector in storageCluster which will have
fields like enabled to enable/disable logCollector,
periodicity interval after which logs will be rotated
and maxLogSize time after which logs will be rotated.
If both periodicity and maxLogSize what every comes
first will rotate the logs.


subhamkrai:
/assign @travisn 

cc @agarwal-mudit , @iamniting

PR #1788, opened 2022-08-29
Bug 2120268: rook: add logCollector in storageCluster

subhamkrai:
rook: upgrade rook to v1.9.9

rook: add logCollector in storageCluster

adding logCollector in storageCluster which will have
fields like enabled to enable/disable logCollector,
periodicity interval after which logs will be rotated
and maxLogSize time after which logs will be rotated.
If both periodicity and maxLogSize what every comes
first will rotate the logs.


subhamkrai:
/assign @travisn 
seems like BP to 4.10 will be difficult 
```
/go/bin/controller-gen "crd:trivialVersions=true" rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
vendor/github.com/hashicorp/vault/sdk/helper/pluginutil/run_config.go:13:2: //go:build comment without // +build comment
Error: not all generators ran successfully
```

we are upgrading rook from 1.8.3 to 1.9.9 there could be many errors similar to this.

cc @agarwal-mudit @iamniting

travisn:
> /assign @travisn seems like BP to 4.10 will be difficult
> 
> ```
> /go/bin/controller-gen "crd:trivialVersions=true" rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
> vendor/github.com/hashicorp/vault/sdk/helper/pluginutil/run_config.go:13:2: //go:build comment without // +build comment
> Error: not all generators ran successfully
> ```
> 
> we are upgrading rook from 1.8.3 to 1.9.9 there could be many errors similar to this.
> 
> cc @agarwal-mudit @iamniting

@subhamkrai Instead of backporting the PR from 4.11 that has all the changes to the go modules, let's create a PR with only the log rotate setting and regenerate the go modules based on the 4.10 branch. Then the change will be much smaller and the 4.11 go module updates won't cause backport issues.

go.mod
```diff
@@ -28,25 +25,24 @@ require (
 	github.com/operator-framework/operator-lib v0.6.0
 	github.com/operator-framework/operator-lifecycle-manager v0.17.0
 	github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring v0.46.0
-	github.com/prometheus/client_golang v1.11.0
+	github.com/prometheus/client_golang v1.11.1
 	github.com/prometheus/client_model v0.2.0
-	github.com/rook/rook v1.8.3
+	github.com/rook/rook v1.9.9
```
travisn, line 30: The 4.10 branch must be based on v1.8, not on v1.9

subhamkrai: so how we'll get API from rook based on 1.9.9?

subhamkrai (commented):
this changes are coming `make gen-latest-csv` and `make update-generated` https://github.com/red-hat-storage/ocs-operator/pull/1788/commits/a013e88aa649f194add321e61664a1472b597181

deploy/bundle/manifests/cephblockpool.crd.yaml
```diff
@@ -14,7 +14,11 @@ spec:
     singular: cephblockpool
   scope: Namespaced
   versions:
-  - name: v1
+  - additionalPrinterColumns:
```
travisn, line 17: Changes aren't expected in this crd

deploy/bundle/manifests/cephblockpoolradosnamespace.crd.yaml
```diff
@@ -0,0 +1,52 @@
+---
```
travisn, line 1: not expected to backport this crd, let's see if we can restrict to only the changes needed for the log collector

deploy/bundle/manifests/cephfilesystem.crd.yaml
```diff
@@ -314,6 +314,89 @@ spec:
                     nullable: true
                     type: object
                     x-kubernetes-preserve-unknown-fields: true
+                  livenessProbe:
```
travisn, line 317: not expected

hack/common.sh
```diff
@@ -44,9 +44,9 @@ NOOBAA_CSV="$OUTDIR_TEMPLATES/noobaa-csv.yaml"
 ROOK_CSV="$OUTDIR_TEMPLATES/rook-csv.yaml.in"
 OCS_CSV="$OUTDIR_TEMPLATES/ocs-operator.csv.yaml.in"
 
-LATEST_ROOK_IMAGE="rook/ceph:v1.8.3"
-LATEST_NOOBAA_IMAGE="noobaa/noobaa-operator:master-20210912"
-LATEST_NOOBAA_CORE_IMAGE="noobaa/noobaa-core:master-20210912"
+LATEST_ROOK_IMAGE="rook/ceph:v1.9.9"
```
travisn, line 47: We really don't expect to use a 1.9.9 image in the 4.10 branch, which is based on 1.8

travisn (changes requested):

hack/common.sh
subhamkrai: reverted, we should be 1.8 based only now

api/v1/storagecluster_types.go
```diff
@@ -79,6 +79,11 @@ type StorageClusterSpec struct {
 	// AllowRemoteStorageConsumers Indicates that the OCS cluster should deploy the needed
 	// components to enable connections from remote consumers.
 	AllowRemoteStorageConsumers bool `json:"allowRemoteStorageConsumers,omitempty"`
+
+	// Logging represents loggings settings
```
umangachapagain, line 83:
```suggestion
	// LogCollector represents log collector settings
```

umangachapagain (changes requested):

umangachapagain:
Can't find the logs for tests. Please test it manually once.

subhamkrai:
> Can't find the logs for tests. Please test it manually once.

I have tested the original pr and also verified by qe(IIRC) it is just backport to 4.10

hack/common.sh
```diff
@@ -44,7 +44,7 @@ NOOBAA_CSV="$OUTDIR_TEMPLATES/noobaa-csv.yaml"
 ROOK_CSV="$OUTDIR_TEMPLATES/rook-csv.yaml.in"
 OCS_CSV="$OUTDIR_TEMPLATES/ocs-operator.csv.yaml.in"
 
-LATEST_ROOK_IMAGE="rook/ceph:v1.8.3"
+LATEST_ROOK_IMAGE="subham03/rook:down-4.10"
```
subhamkrai, line 47:
```suggestion
LATEST_ROOK_IMAGE="subham03/rook:down-4.10"
```
will change this later, before pr merges

api/v1/storagecluster_types.go
subhamkrai: done

PR #1791, opened 2022-08-30
Set CSI_ENABLE_METADATA by default false

malayparida2000:
We don't want to enable metadata for CSI driver by default.
If a customer wants to enable metadata they can edit the
rook-ceph-operator-config cm to set CSI_ENABLE_METADATA to true.

oc patch cm rook-ceph-operator-config -p $'data:\n "CSI_ENABLE_METADATA":  "true"'


iamniting (approved):
/lgtm

PR #1792, opened 2022-08-31
must-gather: add ceph log last commands

pdvian:
Add ceph log last commands to collect ceph cluster
logs to ease debugging from RADOS perspective. 

NOTE: Currently, the ceph log last command returns only
50 cluster log entries at max even if we specify 10000
(mon_log_max) log entries to be retrieved in the log
last command. This will be fixed with ceph PR#47873.

Signed-off-by: Prashant D <pdhange@redhat.com>


iamniting (approved):
/ok-to-test
/lgtm

PR #1795, opened 2022-09-02
Register non nil collectors

aruniiird:
Registring only non-nil collectors and all the `New` collector functions should either return a proper collector or returns `nil`.
Register function will check for `nil` objects and only register non-nil collector objects.


umangachapagain (changes requested):
Good changes. Thanks.

Please squash the commits and add commit message as described here: [Contributing Guide](https://github.com/red-hat-storage/ocs-operator/blob/main/CONTRIBUTING.md#commit-and-pull-request-messages)

PR #1800, opened 2022-09-07
Give correct name for poolName in the non-resilient pool storageclass

malayparida2000:
Was using the failureDomain as the poolName in the non-resilient pool storageclass. This is incorrect. The poolName should be the name of the pool which is the failureDomainValue.

This was mistakenly left from this PR https://github.com/red-hat-storage/ocs-operator/pull/1782.


iamniting (approved):
/lgtm

PR #1801, opened 2022-09-07
Advanced feature metric should be send per ODF installation

aruniiird:
A single 'ocs-advanced-feature-usage' metric should be send per Openshift installation. The metric value will be set to '1' if any of the ceph-cluster in any namespace is using an advanced-feature (like PV/KMS encryption or external cluster mode etc)

### DIFF
--- a/metrics/internal/collectors/cluster-advance-feature-use.go
+++ b/metrics/internal/collectors/cluster-advance-feature-use.go
@@ -1,9 +1,6 @@
 package collectors
 
 import (
-	"fmt"
-	"strings"
-
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/red-hat-storage/ocs-operator/metrics/internal/options"
 	cephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
@@ -23,9 +20,6 @@ type ClusterAdvanceFeatureCollector struct {
 	AdvancedFeature   *prometheus.Desc
 	Informer          cache.SharedIndexInformer
 	AllowedNamespaces []string
-	// advancedFeatureMap will have 'namespace/clusterName' as key and an integer as value.
-	// Value '1' indicates the cluster is using an advanced feature or else '0'.
-	advancedFeatureMap map[string]int
 }
 
 const (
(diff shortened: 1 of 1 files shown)


aruniiird:
@umangachapagain , ready for review... please take a look...

aruniiird:
As the collector's `Collect()` method will be used concurrently, made a small change.
Instead of using stateful variable in the struct object, moving it to a pure functional variable and made all the struct-functions as `no-side-effect` functions. Thus we don't require any mutex (or any other concurrency mechanism) to call `Collect()` method.

metrics/internal/collectors/cluster-advance-feature-use.go
```diff
@@ -77,102 +69,93 @@ func (c *ClusterAdvanceFeatureCollector) Describe(ch chan<- *prometheus.Desc) {
...
-			break
-		}
+	cephRBDMirrorLister := cephv1listers.NewCephRBDMirrorLister(c.Informer.GetIndexer())
+	cephRBDMirrors := getAllRBDMirrors(cephRBDMirrorLister, c.AllowedNamespaces)
+	if len(cephRBDMirrors) > 0 {
+		advancedFeature = c.advancedFeatureFromCephRBDMirrors(cephRBDMirrors)
 	}
-	return validKey
+
+	c.collectAdvancedFeatureUse(ch, advancedFeature)
```
umangachapagain, line 113: Don't we need to check if `advancedFeature > 0` here?

umangachapagain:
Commit title could be `Send advanced feature usage metric per ODF installation` as per contributing guide.

metrics/internal/collectors/cluster-advance-feature-use.go
aruniiird: Last `collectAdvancedFeatureUse()` call should be called unconditionally so that we should collect (whether) ZERO or ONE and send the metric. This is needed if customer resets a special feature then metric should reflect that (or if customer is not using any advanced feature).

aruniiird:
> Commit title could be `Send advanced feature usage metric per ODF installation` as per contributing guide.

Done

umangachapagain (approved):
/lgtm

PR #1802, opened 2022-09-08
reduce cardinality for odf_system_bucket_count and odf_system_objects_total

weirdwiz:
Remove the labels to reduce the cardinality for the odf_system_bucket_count and odf_system_objects_total metrics.


iamniting (changes requested):
Can you pls update description?

umangachapagain (approved):

PR #1804, opened 2022-09-12
Remove PV controller

Madhu-1:
Path PV with controller expand secret name and secret namespace only if the storageclass
supports expansion operation. If we try to patch the PV with an empty secret name and secret namespace operation will fail.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2125815#c17


controllers/persistentvolume/reconcile_test.go
```diff
@@ -68,8 +75,11 @@ func TestEnsureExpansionSecret(t *testing.T) {
 
 		testSC := &storagev1.StorageClass{}
 		mockStorageClass.DeepCopyInto(testSC)
-		testSC.Parameters[csiExpansionSecretName] = c.secretName
-		testSC.Parameters[csiExpansionSecretNamespace] = c.secretName
+		if c.allowExpansion {
+			testSC.Parameters[csiExpansionSecretName] = c.secretName
+			testSC.Parameters[csiExpansionSecretNamespace] = c.secretName
```
iamniting, line 80: Should not it set namespace as a value?

controllers/persistentvolume/reconcile_test.go
```diff
@@ -82,13 +92,14 @@ func TestEnsureExpansionSecret(t *testing.T) {
...
-		assert.Equal(t, testSC.Parameters[csiExpansionSecretName], actualSecretRef.Name)
-		assert.Equal(t, testSC.Parameters[csiExpansionSecretNamespace], actualSecretRef.Namespace)
+		if c.allowExpansion {
+			actual := &corev1.PersistentVolume{}
+			err = reconciler.Client.Get(context.TODO(), request.NamespacedName, actual)
+			assert.NoError(t, err)
+			actualSecretRef := actual.Spec.PersistentVolumeSource.CSI.ControllerExpandSecretRef
+			assert.Equal(t, testSC.Parameters[csiExpansionSecretName], actualSecretRef.Name)
+			assert.Equal(t, testSC.Parameters[csiExpansionSecretNamespace], actualSecretRef.Namespace)
+		}
```
iamniting, line 102: I think in case of false also we should perform some checks like value is not set

iamniting (changes requested):

controllers/persistentvolume/reconcile_test.go
Madhu-1: Yes correct, its a bug in current code will fix it

controllers/persistentvolume/reconcile_test.go
Madhu-1: its an extra check i dont think its necessary,

Madhu-1:
/hold

still discussing whether to do this fix or remove pv_controller completely. will update later

Madhu-1:
In older releases of ODF (<4.5), we didn't have support for PVC expansion, To support the expansion of older PVC [1] we added a hack in operator to patch all the existing PV for expansion secret and secretnamespace in [2]

[1] https://issues.redhat.com/browse/RHSTOR-1428
[2]  https://github.com/red-hat-storage/ocs-operator/pull/796

We decided to remove this hack in ODF 4.12 as we don't need to support this anymore because if the customer created an SC without expansion, he/she could not expand the PVC. We need to have a KCS and insight rule to
identify the problem if anyone has it.

cc @agarwal-mudit 

/hold cancel

iamniting:
looks good to me, Giving a chance to @umangachapagain to take a look

umangachapagain (approved):

PR #1805, opened 2022-09-12
Add separate 'SharedIndexInformer' to each feature providers

aruniiird:
Previously each advanced feature provider was using the same caching mechanism. This was causing crash while running the cluster. Now each provider has its own caching indexer.

### DIFF
--- a/metrics/internal/collectors/cluster-advance-feature-use.go
+++ b/metrics/internal/collectors/cluster-advance-feature-use.go
@@ -9,6 +9,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
+	k8s "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog"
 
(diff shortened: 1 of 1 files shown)


metrics/internal/collectors/cluster-advance-feature-use.go
```diff
@@ -33,27 +128,33 @@ var _ prometheus.Collector = &ClusterAdvanceFeatureCollector{}
...
+		klog.Errorf("cluster advanced feature collector failed to create client: %v", err)
 		return nil
 	}
 
-	lw := cache.NewListWatchFromClient(client.CephV1().RESTClient(), "cephclusters", metav1.NamespaceAll, fields.Everything())
-	sharedIndexInformer := cache.NewSharedIndexInformer(lw, &cephv1.CephCluster{}, 0, cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc})
+	advFeatureProviders := []AdvancedFeatureProvider{
+		NewCephClusterAdvancedFeatureProvider(client),
+		NewCephObjectStoreAdvancedFeatureProvider(client),
+		NewStorageClassAdvancedFeatureProvider(client),
```
umangachapagain, line 138: StorageClass would need K8s client instead of rook client, right?

aruniiird: Yes, I remember you mentioned about StorageClass being a kubernetes resource. Made the changes, please take a look.

umangachapagain (approved):
/lgtm

PR #1807, opened 2022-09-12
Replace StorageClassClaim reconciliation in Provider clusters

jarrpa:
This PR is a weird one, because we're removing the use of a public API but maintaining its same functionality using its go structs... yeah. :joy: 

With StorageClassClaims being moved to a new component, to be used exclusively as a Consumer-facing resource, the Provider service and operator need to switch away from using their API. Internally we will still use the same go structs, but the Kubernetes objects being manipulated will change, in this case, to ConfigMaps.

### DIFF
--- a/main.go
+++ b/main.go
@@ -38,6 +38,7 @@ import (
 	ocsv1alpha1 "github.com/red-hat-storage/ocs-operator/api/v1alpha1"
 	"github.com/red-hat-storage/ocs-operator/controllers/ocsinitialization"
 	"github.com/red-hat-storage/ocs-operator/controllers/storageclassclaim"
+	"github.com/red-hat-storage/ocs-operator/controllers/storageclassconfig"
 	"github.com/red-hat-storage/ocs-operator/controllers/storagecluster"
 	controllers "github.com/red-hat-storage/ocs-operator/controllers/storageconsumer"
 	"github.com/red-hat-storage/ocs-operator/controllers/util"
@@ -184,6 +185,15 @@ func main() {
 		OperatorNamespace: operatorNamespace,
 	}).SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "StorageClassClaim")
+		os.Exit(1)
+	}
+	if err = (&storageclassconfig.StorageClassConfigReconciler{
+		Cache:             mgr.GetCache(),
+		Client:            mgr.GetClient(),
+		Scheme:            mgr.GetScheme(),
+		OperatorNamespace: operatorNamespace,
+	}).SetupWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create controller", "controller", "StorageClassConfig")
 		os.Exit(1)
 	}
 	// +kubebuilder:scaffold:builder
--- a/services/provider/server/storageclaim.go
+++ b/services/provider/server/storageclaim.go
@@ -10,10 +10,11 @@ import (
 
 	ocsv1alpha1 "github.com/red-hat-storage/ocs-operator/api/v1alpha1"
 	controllers "github.com/red-hat-storage/ocs-operator/controllers/storageconsumer"
+	"gopkg.in/yaml.v2"
 
+	corev1 "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/klog/v2"
 	"k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -65,13 +66,13 @@ func (s *storageClassClaimManager) Create(
 	generatedClaimName := getStorageClassClaimName(consumerUUID, storageClassClaimName)
 
 	storageClassClaimObj := &ocsv1alpha1.StorageClassClaim{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "StorageClassClaim",
+			APIVersion: "ocs.openshift.io/v1",
+		},
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      generatedClaimName,
 			Namespace: s.namespace,
-			Labels: map[string]string{
-				controllers.ConsumerUUIDLabel: consumerUUID,
-				storageClassClaimNameLabel:    storageClassClaimName,
-			},
 		},
 		Spec: ocsv1alpha1.StorageClassClaimSpec{
 			Type:             claimType,
@@ -88,7 +89,24 @@ func (s *storageClassClaimManager) Create(
 		return fmt.Errorf("failed to get gvk for consumer %q. %w", consumerUUID, err)
 	}
 
-	storageClassClaimObj.SetOwnerReferences([]metav1.OwnerReference{
+	claimBytes, err := yaml.Marshal(storageClassClaimObj)
+	if err != nil {
+		return fmt.Errorf("failed to marshal StorageClassClaim for consumer %q. %w", consumerUUID, err)
+	}
+	cmObj := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      storageClassClaimObj.Name,
+			Namespace: s.namespace,
+			Labels: map[string]string{
+				controllers.ConsumerUUIDLabel:      consumerUUID,
+				ocsv1alpha1.StorageClassClaimLabel: storageClassClaimName,
+			},
+		},
+		Data: map[string]string{
+			"StorageClassClaim": string(claimBytes),
+		},
+	}
+	cmObj.SetOwnerReferences([]metav1.OwnerReference{
 		{
 			APIVersion:         gvk.GroupVersion().String(),
 			Kind:               gvk.Kind,
@@ -98,20 +116,27 @@ func (s *storageClassClaimManager) Create(
 		},
 	})
 
-	err = s.client.Create(ctx, storageClassClaimObj)
+	err = s.client.Create(ctx, cmObj)
 	if err != nil {
 		if !kerrors.IsAlreadyExists(err) {
-			return fmt.Errorf("failed to create a StorageClassClaim named %q for consumer %q and claim %q. %w", generatedClaimName, consumerUUID, storageClassClaimName, err)
+			return fmt.Errorf("failed to create a ConfigMap named %q for consumer %q and claim %q. %w", generatedClaimName, consumerUUID, storageClassClaimName, err)
 		}
-		newStorageClassClaimObj := &ocsv1alpha1.StorageClassClaim{}
-		getErr := s.client.Get(ctx, client.ObjectKey{Name: generatedClaimName, Namespace: s.namespace}, newStorageClassClaimObj)
+		existingCM := &corev1.ConfigMap{}
+		getErr := s.client.Get(ctx, client.ObjectKey{Name: cmObj.Name, Namespace: s.namespace}, existingCM)
 		if getErr != nil {
-			klog.Errorf("failed to get a StorageClassClaim named %q for consumer %q and claim %q. %v", generatedClaimName, consumerUUID, storageClassClaimName, getErr)
+			klog.Errorf("failed to get a ConfigMap named %q for consumer %q and claim %q. %v", generatedClaimName, consumerUUID, storageClassClaimName, getErr)
 			return err
 		}
-		// check if the storageClassClaim is getting deleted.
-		if newStorageClassClaimObj.DeletionTimestamp != nil {
+		// check if the StorageClassClaim ConfigMap is getting deleted.
+		if existingCM.DeletionTimestamp != nil {
 			klog.Warningf("StorageClassClaim named %q for consumer %q and claim %q is already created but is getting deleted", generatedClaimName, consumerUUID, storageClassClaimName)
+			return err
+		}
+		newStorageClassClaimObj := &ocsv1alpha1.StorageClassClaim{}
+		claimString := existingCM.Data["StorageClassClaim"]
+		unmarshallErr := yaml.Unmarshal([]byte(claimString), newStorageClassClaimObj)
+		if unmarshallErr != nil {
+			klog.Errorf("failed to unmarshall the StorageClassClaim named %q for consumer %q and claim %q. %v", generatedClaimName, consumerUUID, storageClassClaimName, unmarshallErr)
 			return err
 		}
 		// check if the input is different
@@ -121,7 +146,7 @@ func (s *storageClassClaimManager) Create(
 		}
 	}
 
-	klog.Infof("successfully created a StorageClassClaim resource %q for consumer %q and claim %q", generatedClaimName, consumerUUID, storageClassClaimName)
+	klog.Infof("successfully created a StorageClassClaim ConfigMap %q for consumer %q and claim %q", cmObj.Name, consumerUUID, storageClassClaimName)
 
 	return nil
 }
@@ -130,7 +155,7 @@ func (s *storageClassClaimManager) Create(
 // and consumerUUID.
 func (s *storageClassClaimManager) Delete(ctx context.Context, consumerUUID, storageClassClaimName string) error {
 	generatedClaimName := getStorageClassClaimName(consumerUUID, storageClassClaimName)
-	storageClassClaimObj := &ocsv1alpha1.StorageClassClaim{
+	storageClassClaimObj := &corev1.ConfigMap{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      generatedClaimName,
 			Namespace: s.namespace,
@@ -159,10 +184,22 @@ func (s *storageClassClaimManager) Delete(ctx context.Context, consumerUUID, sto
 func (s *storageClassClaimManager) Get(ctx context.Context, consumerUUID, storageClassClaimName string) (*ocsv1alpha1.StorageClassClaim, error) {
 	generatedClaimName := getStorageClassClaimName(consumerUUID, storageClassClaimName)
 	storageClassClaimObj := &ocsv1alpha1.StorageClassClaim{}
-	err := s.client.Get(ctx, types.NamespacedName{Name: generatedClaimName, Namespace: s.namespace}, storageClassClaimObj)
-	if err != nil {
-		klog.Errorf("failed to get a StorageClassClaim named %q for consumer %q and claim %q. %v", generatedClaimName, consumerUUID, storageClassClaimName, err)
-		return nil, err
+	existingCM := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      generatedClaimName,
+			Namespace: s.namespace,
+		},
+	}
+	getErr := s.client.Get(ctx, client.ObjectKey{Name: existingCM.Name, Namespace: existingCM.Namespace}, existingCM)
+	if getErr != nil {
+		klog.Errorf("failed to get a ConfigMap named %q for consumer %q and claim %q. %v", existingCM.Name, consumerUUID, storageClassClaimName, getErr)
+		return nil, getErr
+	}
+	claimString := existingCM.Data["StorageClassClaim"]
+	unmarshallErr := yaml.Unmarshal([]byte(claimString), storageClassClaimObj)
+	if unmarshallErr != nil {
+		klog.Errorf("failed to unmarshall the StorageClassClaim named %q for consumer %q and claim %q. %v", generatedClaimName, consumerUUID, storageClassClaimName, unmarshallErr)
+		return nil, unmarshallErr
 	}
 
 	return storageClassClaimObj, nil
--- a/api/v1alpha1/storageclassclaim_types.go
+++ b/api/v1alpha1/storageclassclaim_types.go
@@ -52,6 +52,7 @@ const (
 const (
 	StorageClassClaimFinalizer  = "storageclassclaim.ocs.openshift.io"
 	StorageClassClaimAnnotation = "ocs.openshift.io.storagesclassclaim"
+	StorageClassClaimLabel      = "ocs.openshift.io/storageclassclaim"
 	CephFileSystemDataPoolLabel = "cephfilesystem.datapool.name"
 )
 
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -129,6 +129,18 @@ rules:
   - ""
   resources:
   - configmaps
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
   - endpoints
   - events
   - nodes
--- a/controllers/storageclassconfig/storageclassconfig_controller.go
+++ b/controllers/storageclassconfig/storageclassconfig_controller.go
@@ -29,7 +29,10 @@ import (
 	"github.com/red-hat-storage/ocs-operator/api/v1alpha1"
 	"github.com/red-hat-storage/ocs-operator/controllers/storagecluster"
 	controllers "github.com/red-hat-storage/ocs-operator/controllers/storageconsumer"
+	"github.com/red-hat-storage/ocs-operator/controllers/util"
 	rookCephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
+	"gopkg.in/yaml.v2"
+	corev1 "k8s.io/api/core/v1"
 	storagev1 "k8s.io/api/storage/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -38,6 +41,7 @@ import (
 	"k8s.io/klog/v2"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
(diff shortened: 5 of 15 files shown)


nb-ohad (changes requested):

controllers/storageclassclaimsecret/storageclassclaimsecret_controller.go
```diff
@@ -0,0 +1,665 @@
+/*
+Copyright 2020 Red Hat OpenShift Container Storage.
```
Madhu-1, line 2: 2022

controllers/storageclassclaimsecret/storageclassclaimsecret_controller.go
```diff
@@ -0,0 +1,665 @@
...
+// isClaimSecret determines whether we want to reconcile a given Secret
+func isClaimSecret(obj runtime.Object) bool {
+	secretObj, ok := obj.(*corev1.Secret)
+	if !ok {
+		return false
+	}
+
+	annotations := secretObj.GetAnnotations()
+	if _, found := annotations[v1alpha1.StorageClassClaimAnnotation]; found {
+		return true
```
Madhu-1, line 169: just return found here no need to return true and false

controllers/storageclassclaimsecret/storageclassclaimsecret_controller.go
```diff
@@ -0,0 +1,665 @@
...
+		return reconcile.Result{}, fmt.Errorf("no storage consumer owner ref on the storage class claim")
+	}
+
+	if err := r.get(r.storageConsumer); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	if r.storageClassClaim.Spec.Type == "blockpool" {
+		r.cephBlockPool = &rookCephv1.CephBlockPool{}
+		r.cephBlockPool.Name = fmt.Sprintf("cephblockpool-%s", r.storageConsumer.Name)
```
Madhu-1, line 254: blockpool name is changed in 4.12 not sure we are handing that in this PR

controllers/storageclassclaimsecret/storageclassclaimsecret_controller.go
```diff
@@ -0,0 +1,665 @@
...
+		r.cephBlockPool.Namespace = r.OperatorNamespace
+	} else if r.storageClassClaim.Spec.Type == "sharedfilesystem" {
+		r.cephFilesystemSubVolumeGroup = &rookCephv1.CephFilesystemSubVolumeGroup{}
+		r.cephFilesystemSubVolumeGroup.Name = fmt.Sprintf("cephfilesystemsubvolumegroup-%s", r.storageConsumer.Name)
```
Madhu-1, line 258: subvolumegroup name is changed in 4.12 not sure we are handing that in this PR

controllers/storageclassclaimsecret/storageclassclaimsecret_controller.go
```diff
@@ -0,0 +1,665 @@
...
+
+	_, err := ctrl.CreateOrUpdate(r.ctx, r.Client, r.cephBlockPool, func() error {
+		if err := r.own(r.cephBlockPool); err != nil {
+			return err
+		}
+		deviceSetList := r.storageCluster.Spec.StorageDeviceSets
+		var deviceSet *v1.StorageDeviceSet
+		for i := range deviceSetList {
+			ds := &deviceSetList[i]
+			if ds.Name == "default" {
```
Madhu-1, line 338: Looks like this is old code. Do we have a plan to include the IOPS feature in this controller?

Madhu-1:
Need to rebase on top of https://github.com/red-hat-storage/ocs-operator/pull/1750 to include storageProfile feature

controllers/storageclassclaimsecret/storageclassclaimsecret_controller.go
```diff
@@ -0,0 +1,665 @@
...
+	//	util.MetadataChangedPredicate{},
+	//)
+
+	// As we are not setting the Controller OwnerReference on the ceph
+	// resources we are creating as part of the StorageClassClaim, we need to
+	// set IsController to false to get Reconcile Request of StorageClassClaim
+	// for the owned ceph resources updates.
+	enqueueForNonControllerOwner := &handler.EnqueueRequestForOwner{OwnerType: &v1alpha1.StorageClassClaim{}, IsController: false}
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&corev1.Secret{}, builder.WithPredicates(claimPredicate)).
```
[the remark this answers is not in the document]
jarrpa: I didn't seem to have luck with providing both the `claimPredicate` and `predicate.GenerationChangedPredicate{}`, the Claims didn't get updated from `Creating` to `Ready`. Maybe I didn't wait long enough? All I know is that when I removed the predicate the process completed successfully. The following seems to work:

```diff
@@ -202,17 +202,17 @@ func (r *StorageClassClaimSecretReconciler) SetupWithManager(mgr ctrl.Manager) e
                        return []reconcile.Request{}
                })
        // Compose a predicate that is an OR of the specified predicates
-       //secretsPredicate := util.ComposePredicates(
-       //      predicate.GenerationChangedPredicate{},
-       //      util.MetadataChangedPredicate{},
-       //)
+       secretsPredicate := util.ComposePredicates(
+               predicate.GenerationChangedPredicate{},
+               util.MetadataChangedPredicate{},
+       )
 
        // As we are not setting the Controller OwnerReference on the ceph
        // resources we are creating as part of the StorageClassClaim, we need to
        // set IsController to false to get Reconcile Request of StorageClassClaim
        // for the owned ceph resources updates.
        enqueueForNonControllerOwner := &handler.EnqueueRequestForOwner{OwnerType: &v1alpha1.StorageClassClaim{}, IsController: false}
-       return ctrl.NewControllerManagedBy(mgr).
+       return ctrl.NewControllerManagedBy(mgr).WithEventFilter(secretsPredicate).
                For(&corev1.Secret{}, builder.WithPredicates(claimPredicate)).
                Watches(&source.Kind{Type: &rookCephv1.CephBlockPool{}}, enqueueForNonControllerOwner).
                Watches(&source.Kind{Type: &rookCephv1.CephFilesystemSubVolumeGroup{}}, enqueueForNonControllerOwner).
```

There's a lot of weird stuff happening underneath in the builder functions:

* Using [`util.ComposePredicates`](https://github.com/red-hat-storage/ocs-operator/blob/main/controllers/util/predicates.go#L10-L47) generates a single predicate that is an OR of any predicates you pass into it. We put this together so we could catch any of multiple metadata changes we often (though not always) care about.
* The reason the compose exists is that any slice of predicates provided to `builder.WithPredicates()` generates an AND of all listed predicates, which is not always what we want.

I'm not entirely sure why `claimPredicate AND GenerationChangedPredicate` didn't seem to work, but passing the latter through `WithEventFilter()` got the job done. I'm still working on updating the PR for the other comments, but if you end up responding to this before the next push let me know if you'd accept the above diff.

controllers/storageclassclaimsecret/storageclassclaimsecret_controller.go
jarrpa: By technicality, yes, since it needs to maintain parity with whatever the StorageClassClaim controller is doing on the provider side.

controllers/storageclassclaimsecret/storageclassclaimsecret_controller.go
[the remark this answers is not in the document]
nb-ohad: Thanks for reminding me about the AND nature of predicates. But you shouldn't have created your own OR behavior when you have `predicate.Or` implemented by the framework. 

Also, I think `WithEventFilter` is not necessary when using the built-in OR function (or any of the other built-in functions). It will set the predicate on all `Watches` and not only on the `For` watch. 
This can, and will, result in erroneous behavior for other events.

controllers/storageclassclaimsecret/storageclassclaimsecret_controller.go
nb-ohad: @jarrpa we already merged the 4.12 SCC changes for the provider side on `main`. I would suggest to rebase ASAP

controllers/storageclassclaimsecret/storageclassclaimsecret_controller.go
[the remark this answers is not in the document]
jarrpa: You have just saved me so much time, THANK YOU.

controllers/storageclassclaimsecret/storageclassclaimsecret_controller.go
jarrpa: We'll see if a rebase is even feasible, since these are completely new files that have been created as copies. Even so it's not impossible.

controllers/storageclassclaimsecret/storageclassclaimsecret_controller.go
```diff
@@ -0,0 +1,665 @@
...
+	},
+	UpdateFunc: func(e event.UpdateEvent) bool {
+		return isClaimSecret(e.ObjectNew)
+	},
+	GenericFunc: func(e event.GenericEvent) bool {
+		return isClaimSecret(e.Object)
+	},
+}
+
+func (r *StorageClassClaimSecretReconciler) SetupWithManager(mgr ctrl.Manager) error {
```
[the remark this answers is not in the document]
jarrpa: :+1:

controllers/storageclassclaimsecret/storageclassclaimsecret_controller.go
```diff
@@ -0,0 +1,665 @@
...
+		For(&corev1.Secret{}, builder.WithPredicates(claimPredicate)).
+		Watches(&source.Kind{Type: &rookCephv1.CephBlockPool{}}, enqueueForNonControllerOwner).
+		Watches(&source.Kind{Type: &rookCephv1.CephFilesystemSubVolumeGroup{}}, enqueueForNonControllerOwner).
+		Watches(&source.Kind{Type: &rookCephv1.CephClient{}}, enqueueForNonControllerOwner).
+		Watches(&source.Kind{Type: &storagev1.StorageClass{}}, enqueueStorageConsumerRequest).
+		Watches(&source.Kind{Type: &snapapi.VolumeSnapshotClass{}}, enqueueStorageConsumerRequest).
+		Complete(r)
+}
+
+func (r *StorageClassClaimSecretReconciler) reconcileProviderPhases() (reconcile.Result, error) {
```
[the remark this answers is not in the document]
jarrpa: :+1:

controllers/storageclassclaimsecret/storageclassclaimsecret_controller.go
```diff
@@ -0,0 +1,665 @@
...
+
+	claimBytes, marshalErr := json.Marshal(r.storageClassClaim)
+	if marshalErr != nil {
+		r.log.Error(marshalErr, "Failed to marshal StorageClassClaim Secret.")
+		return reconcile.Result{}, marshalErr
+	}
+	r.storageClassClaimSecret.Data["StorageClassClaim"] = claimBytes
+
+	// Apply status changes to the StorageClassClaim
+	statusError := r.Client.Update(r.ctx, r.storageClassClaimSecret)
```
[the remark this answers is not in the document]
jarrpa: :+1:

controllers/storageclassclaimsecret/storageclassclaimsecret_controller.go
jarrpa: :+1:

controllers/storageclassclaimsecret/storageclassclaimsecret_controller.go
```diff
@@ -0,0 +1,665 @@
...
+	// reading storageConsumer Name from storageClassClaimSecret ownerReferences
+	ownerRefs := r.storageClassClaimSecret.GetOwnerReferences()
+	for i := range ownerRefs {
+		if ownerRefs[i].Kind == gvk.Kind {
+			r.storageConsumer = &v1alpha1.StorageConsumer{}
+			r.storageConsumer.Name = ownerRefs[i].Name
+			r.storageConsumer.Namespace = r.OperatorNamespace
+			break
+		}
+	}
```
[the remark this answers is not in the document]
jarrpa: This is on the provider side, so there's no reason to have OwnerRefs on the claim since it's not a real resource. The Secret is the thing that should have OwnerRefs.

controllers/storageclassclaimsecret/storageclassclaimsecret_controller.go
```diff
@@ -0,0 +1,665 @@
...
+// +kubebuilder:rbac:groups=ocs.openshift.io,resources=storageclassclaims,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=ocs.openshift.io,resources=storageclassclaims/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=ceph.rook.io,resources=cephclients,verbs=get;list;watch;create;update;delete
+// +kubebuilder:rbac:groups=ceph.rook.io,resources=cephfilesystemsubvolumegroups,verbs=get;list;watch;create;update;delete
+// +kubebuilder:rbac:groups=ceph.rook.io,resources=cephblockpools,verbs=get;list;watch;create;update;delete
+// +kubebuilder:rbac:groups=ocs.openshift.io,resources=storageclusters,verbs=get;watch;list
+// +kubebuilder:rbac:groups=storage.k8s.io,resources=storageclasses,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=snapshot.storage.k8s.io,resources=volumesnapshotclasses,verbs=get;list;watch;create;delete
+// +kubebuilder:rbac:groups=core,resources=persistentvolumes,verbs=get;list;watch
+// +kubebuilder:rbac:groups=core,resources=secrets,verbs=*
```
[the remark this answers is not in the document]
jarrpa: :+1:

controllers/storageclassclaimsecret/storageclassclaimsecret_controller.go
jarrpa: :+1:

controllers/storageclassclaimsecret/storageclassclaimsecret_controller.go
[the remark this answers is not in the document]
nb-ohad: Ok

controllers/storageclassclaimsecret/storageclassclaimsecret_controller.go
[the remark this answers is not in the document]
jarrpa: Done

controllers/storageclassclaimsecret/storageclassclaimsecret_controller.go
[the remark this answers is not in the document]
jarrpa: Done

controllers/storageclassclaimsecret/storageclassclaimsecret_controller.go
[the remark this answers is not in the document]
jarrpa: Done

controllers/storageclassclaimsecret/storageclassclaimsecret_controller.go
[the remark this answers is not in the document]
jarrpa: Done

controllers/storageclassclaimsecret/storageclassclaimsecret_controller.go
jarrpa: Done

controllers/storageclassclaimsecret/storageclassclaimsecret_controller.go
jarrpa: Done

controllers/storageclassclaimsecret/storageclassclaimsecret_controller.go
jarrpa: Fixed in rebase

controllers/storageclassclaimsecret/storageclassclaimsecret_controller.go
jarrpa: Fixed in rebase

controllers/storageclassclaimsecret/storageclassclaimsecret_controller.go
jarrpa: The rebase itself wasn't too bad, but things immediately broke afterwards. I managed to find and resolve the two small bugs from #1750, then cherry-picked Madhu's fixes until I could do another rebase earlier today. This should now be resolved.

controllers/storageclassclaimsecret/storageclassclaimsecret_controller.go
```diff
@@ -0,0 +1,665 @@
...
+type StorageClassClaimSecretReconciler struct {
+	client.Client
+	Scheme            *runtime.Scheme
+	OperatorNamespace string
+
+	log                          logr.Logger
+	ctx                          context.Context
+	storageConsumer              *v1alpha1.StorageConsumer
+	storageCluster               *v1.StorageCluster
+	storageClassClaim            *v1alpha1.StorageClassClaim
```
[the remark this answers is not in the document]
jarrpa: For the sake of code stability, I think it's sufficient to just remove the CRD from the ocs-operator bundle but leave the API for now. It can be moved or removed later.

controllers/storageclassclaimsecret/storageclassclaimsecret_controller.go
```diff
@@ -0,0 +1,665 @@
...
+	}
+
+	result, reconcileError = r.reconcileProviderPhases()
+
+	claimBytes, marshalErr := json.Marshal(r.storageClassClaim)
+	if marshalErr != nil {
+		r.log.Error(marshalErr, "Failed to marshal StorageClassClaim Secret.")
+		return reconcile.Result{}, marshalErr
+	}
+	r.storageClassClaimSecret.Data["StorageClassClaim"] = claimBytes
```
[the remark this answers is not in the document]
jarrpa: Done

controllers/storageclassclaimsecret/storageclassclaimsecret_controller.go
```diff
@@ -0,0 +1,665 @@
...
+		r.log.Error(err, "Failed to get StorageClassClaim Secret.")
+		return reconcile.Result{}, err
+	}
+
+	r.storageClassClaim = &v1alpha1.StorageClassClaim{}
+	unmarshallErr := json.Unmarshal(r.storageClassClaimSecret.Data["StorageClassClaim"], r.storageClassClaim)
+	if unmarshallErr != nil {
+		r.log.Error(unmarshallErr, "Failed to unmarshal StorageClassClaim Secret.")
+		return reconcile.Result{}, unmarshallErr
+	}
```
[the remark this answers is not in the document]
jarrpa: Done

services/provider/server/storageclaim.go
```diff
@@ -98,19 +98,49 @@ func (s *storageClassClaimManager) Create(
...
+			},
+			Name:      fmt.Sprintf("%s-claim", storageClassClaimObj.Name),
+			Namespace: s.namespace,
+			Labels:    storageClassClaimObj.ObjectMeta.Labels,
+		},
+		Data: map[string][]byte{
+			"StorageClassClaim": claimBytes,
+		},
+	}
+	secretObj.SetOwnerReferences([]metav1.OwnerReference{
```
nb-ohad, line 115: Just a reminder to remove the SetOwnerReferences call on the claim object (located above this code change)

services/provider/server/storageclaim.go
```diff
@@ -98,19 +98,49 @@ func (s *storageClassClaimManager) Create(
...
+		getErr := s.client.Get(ctx, client.ObjectKey{Name: secretObj.Name, Namespace: s.namespace}, existingSecret)
 		if getErr != nil {
-			klog.Errorf("failed to get a StorageClassClaim named %q for consumer %q and claim %q. %v", generatedClaimName, consumerUUID, storageClassClaimName, getErr)
+			klog.Errorf("failed to get a Secret named %q for consumer %q and claim %q. %v", generatedClaimName, consumerUUID, storageClassClaimName, getErr)
+			return err
+		}
+		newStorageClassClaimObj := &ocsv1alpha1.StorageClassClaim{}
+		unmarshallErr := json.Unmarshal(existingSecret.Data["StorageClassClaim"], newStorageClassClaimObj)
+		if unmarshallErr != nil {
+			klog.Errorf("failed to unmarshall the StorageClassClaim named %q for consumer %q and claim %q. %v", generatedClaimName, consumerUUID, storageClassClaimName, unmarshallErr)
```
nb-ohad, line 139: Why do we need to unmarshal here?

controllers/storageclassclaimsecret/storageclassclaimsecret_controller.go
```diff
@@ -0,0 +1,691 @@
...
+// +kubebuilder:rbac:groups=ceph.rook.io,resources=cephblockpools,verbs=get;list;watch;create;update;delete
+// +kubebuilder:rbac:groups=ocs.openshift.io,resources=storageclusters,verbs=get;watch;list
+// +kubebuilder:rbac:groups=storage.k8s.io,resources=storageclasses,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=snapshot.storage.k8s.io,resources=volumesnapshotclasses,verbs=get;list;watch;create;delete
+// +kubebuilder:rbac:groups=core,resources=persistentvolumes,verbs=get;list;watch
+// +kubebuilder:rbac:groups=core,resources=secrets,verbs=get;list;watch;create;update;patch;delete
+
+func (r *StorageClassClaimSecretReconciler) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+	r.log = ctrllog.FromContext(ctx, "StorageClassClaim Secret", request)
+	r.ctx = ctrllog.IntoContext(ctx, r.log)
```
nb-ohad, line 88: You are already getting the `r.log` from the context. There is no need to call `IntoContext` as well

controllers/storageclassclaimsecret/storageclassclaimsecret_controller.go
```diff
@@ -0,0 +1,691 @@
...
+		return reconcile.Result{}, fmt.Errorf("StorageClassClaim Secret is not supported in OCSConsumer mode")
+	}
+
+	result, reconcileError = r.reconcilePhases()
+
+	if claimBytes, marshalErr := json.Marshal(r.storageClassClaim); marshalErr == nil {
+		r.storageClassClaimSecret.Data["StorageClassClaim"] = claimBytes
+	} else {
+		r.log.Error(marshalErr, "Failed to marshal StorageClassClaim Secret.")
+		return reconcile.Result{}, marshalErr
```
nb-ohad, line 140: I think this should be a Panic (or os.Exist), not a returned Error. 
If we cannot marshal it means we have a fundamental code error.

controllers/storageclassclaimsecret/storageclassclaimsecret_controller.go
```diff
@@ -0,0 +1,691 @@
...
+		})
+
+	// TODO: Annotations or labels...?
+	claimAnnotationPredicate := predicate.NewPredicateFuncs(
+		func(obj client.Object) bool {
+			annotations := obj.GetAnnotations()
+			_, found := annotations[v1alpha1.StorageClassClaimAnnotation]
+			return found
+		},
+	)
```
nb-ohad, line 184: Definitely label as we are deling with the identity of the secret.

controllers/storageclassclaimsecret/storageclassclaimsecret_controller.go
```diff
@@ -0,0 +1,691 @@
...
+	metaPredicate := predicate.Or(
+		predicate.GenerationChangedPredicate{},
+		util.MetadataChangedPredicate{},
```
nb-ohad, line 187: I believe that metadata changes are captured via generation change. No need to Or against it.

nb-ohad (changes requested):

main.go
```diff
@@ -178,6 +179,14 @@ func main() {
 		setupLog.Error(err, "unable to create controller", "controller", "StorageClassClaim")
 		os.Exit(1)
 	}
+	if err = (&storageclassclaimsecret.StorageClassClaimSecretReconciler{
+		Client:            mgr.GetClient(),
```
Madhu-1, line 183: Need to do similar cache change here https://github.com/red-hat-storage/ocs-operator/pull/1811

services/provider/server/storageclaim.go
Madhu-1: in line 148 we are comparing existing and requested spec change

Madhu-1 (commented):
expect @nb-ohad comments LGTM.

services/provider/server/storageclaim.go
nb-ohad: In that case, let's move the unmarshal just above that section of the code.

jarrpa: This was to keep the comparison strictly with the Spec as before, otherwise the comparison would also include the Status changes (and Meta, but that should never relaly change).

controllers/storageclassclaimsecret/storageclassclaimsecret_controller.go
jarrpa: Fair enough. This is an artifact from the initial file copy, so I dunno why it's there, but fixing it nonetheless.

services/provider/server/storageclaim.go
jarrpa: Done

controllers/storageclassclaimsecret/storageclassclaimsecret_controller.go
jarrpa: Not always! It seems to vary from Kind to Kind (because of course it does!! :sob: ) so I figure it's safe enough to leave it for now even if it's slightly inefficient. Thus far I haven't seen enough log spam to indicate a problem. Might be different when we switch to ConfigMaps, so I might be able to look more deeply into it for that.

main.go
jarrpa: Done

controllers/storageclassclaimsecret/storageclassclaimsecret_controller.go
jarrpa: Done!

PR #1811, opened 2022-09-14
WaitForCacheSync in storageclassclaim Reconcile

Madhu-1:
When a storageclassclaim CR is created, a Reconcile is triggered. We are using controller runtime client does GET operation which reads from the runtime cache. When a blockpool or subvolumegroup is created, a new reconcile request will be in the queue until the existing Reconcile completes. Once the existing Reconcile completes, we will get Reconcile request for the resource in the queue and try to get the storageclassclaim object here. We have a chance to get the object from the runtime cache instead of fetching it from the apiserver. Because of it, we get stale resources created for subvolumegroup and blockpool.

As a workaround for the above problem, wait for  Cache to be synced during start of Reconcile.

### DIFF
--- a/controllers/storageclassclaim/storageclassclaim_controller.go
+++ b/controllers/storageclassclaim/storageclassclaim_controller.go
@@ -42,6 +42,7 @@ import (
 	"k8s.io/klog/v2"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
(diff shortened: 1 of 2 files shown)


controllers/storageclassclaim/storageclassclaim_controller.go
```diff
@@ -87,6 +89,11 @@ type StorageClassClaimReconciler struct {
 // +kubebuilder:rbac:groups=core,resources=persistentvolumes,verbs=get;list;watch
 
 func (r *StorageClassClaimReconciler) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+	ok := r.Cache.WaitForCacheSync(ctx)
+	if !ok {
+		return reconcile.Result{}, fmt.Errorf("cache sync failed")
+	}
+
```
nb-ohad, line 96:
```suggestion
	if ok := r.Cache.WaitForCacheSync(ctx); !ok {
		return reconcile.Result{}, fmt.Errorf("cache sync failed")
	}

```

nb-ohad (changes requested):

PR #1816, opened 2022-09-16
Enable RBD mirroring debugging by default

malayparida2000:
Resolves- https://bugzilla.redhat.com/show_bug.cgi?id=2127186

[Enable RBD mirroring debugging by default](https://github.com/red-hat-storage/ocs-operator/commit/d8a663103aed3e3a64ac74e9f63bf6e2521a8a53) 

This will help to debug issues during preview of regional DR.
For GA these settings would be removed.

[Apply rbd_mirror_die_after_seconds only when rbd_mirroring is enabled](https://github.com/red-hat-storage/ocs-operator/commit/0eb18f6ac92e6bf244343f0c34c892c67ff206f8) 

Avoiding setting the config even when rbd_mirroring is disabled, to
avoid any future inconsistencies.

[Fix ensureCreated() to update the found configMap](https://github.com/red-hat-storage/ocs-operator/pull/1816/commits/a27a8fbf57de1681ed33a03a702b60b07d87a4f3) 

Currently we are not updating the found configMap with the new values,
We are just updating the local configMap object.
This makes the getRookCephConfig() useless.
This commit fixes that.

### DIFF
--- a/controllers/storagecluster/cephconfig.go
+++ b/controllers/storagecluster/cephconfig.go
@@ -93,10 +93,23 @@ func (obj *ocsCephConfig) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.S
 			ownerRefFound = true
 		}
 	}
-	val, ok := found.Data["config"]
-	if !ok || val != defaultRookConfigData || !ownerRefFound {
+	if !ownerRefFound {
+		found.OwnerReferences = append(found.OwnerReferences, ownerRef)
+		err = r.Client.Update(context.TODO(), found)
+		if err != nil {
+			r.Log.Error(err, "Failed to update Ceph ConfigMap.", "ConfigMap", klog.KRef(sc.Namespace, rookConfigMapName))
+			return reconcile.Result{}, err
+		}
+	}
+	val := found.Data["config"]
+	if val != rookConfigData {
 		r.Log.Info("Updating Ceph ConfigMap.", "ConfigMap", klog.KRef(sc.Namespace, cm.Name))
-		return reconcile.Result{}, r.Client.Update(context.TODO(), cm)
+		found.Data["config"] = rookConfigData
+		err = r.Client.Update(context.TODO(), found)
+		if err != nil {
+			r.Log.Error(err, "Failed to update ConfigMap.", "ConfigMap", klog.KRef(sc.Namespace, rookConfigMapName))
+			return reconcile.Result{}, err
+		}
 	}
 	return reconcile.Result{}, nil
 }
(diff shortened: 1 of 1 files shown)


controllers/storagecluster/cephconfig.go
```diff
@@ -162,5 +161,28 @@ func getRookCephConfig(r *StorageClusterReconciler, sc *ocsv1.StorageCluster) (s
...
+			`[client.rbd-mirror.a]
+debug_ms = 1
+debug_rbd = 20
+debug_rbd_mirror = 30
+log_file = /var/log/ceph/$cluster-$name.log
+[client.rbd-mirror-peer]
+debug_ms = 1
+debug_rbd = 20
+debug_rbd_mirror = 30
+log_file = /var/log/ceph/$cluster-$name.log
```
idryomov, line 196: Doing it this way doesn't cover `mgr/rbd_support/log_level` and also means that none of these settings can be easily reverted (we have learned that the hard way with `rbd_mirror_die_after_seconds`).  I would prefer if all of these were set with `ceph config set` so that users can easily disable verbose logging if it ends up hurting them.

idryomov (changes requested):

malayparida2000:
```
mparida@mparida-mac ~ % oc patch storagecluster $(oc get storagecluster -n openshift-storage -o=jsonpath='{.items[0].metadata.name}')  -n openshift-storage --type json --patch  '[{ "op": "replace", "path": "/spec/mirroring", "value": {"enabled": true} }]'
storagecluster.ocs.openshift.io/ocs-storagecluster patched

mparida@mparida-mac ~ % oc get cm rook-config-override -o yaml
apiVersion: v1
data:
  config: |
    [global]
    bdev_flock_retry             = 20
    mon_osd_full_ratio           = .85
    mon_osd_backfillfull_ratio   = .8
    mon_osd_nearfull_ratio       = .75
    mon_max_pg_per_osd           = 600
    mon_pg_warn_max_object_skew  = 0
    mon_data_avail_warn          = 15
    rbd_mirror_die_after_seconds = 3600

    [osd]
    osd_memory_target_cgroup_limit_ratio = 0.8
    [client.rbd-mirror.a]
    debug_ms = 1
    debug_rbd = 20
    debug_rbd_mirror = 30
    log_file = /var/log/ceph/$cluster-$name.log
    [client.rbd-mirror-peer]
    debug_ms = 1
    debug_rbd = 20
    debug_rbd_mirror = 30
    log_file = /var/log/ceph/$cluster-$name.log
kind: ConfigMap
metadata:
  creationTimestamp: "2022-09-16T12:19:39Z"
  name: rook-config-override
  namespace: openshift-storage
  ownerReferences:
    kind: StorageCluster
    name: ocs-storagecluster
    uid: 45544010-6c71-49ab-a8b3-c46e2bbbc835
  resourceVersion: "168302"
  uid: 5f53a613-2b14-4919-a90b-2d141e78f483


```

malayparida2000:
As the relevant bug is now closed, so this PR is no longer necessary.

PR #1817, opened 2022-09-16
Limit OCS operator leaderElection & cache to operator namespace

malayparida2000:
Currently ocs-operator is relying on the default "AllNamespaces" cache
in controller-runtime, which works by syncing all the Kubernetes
resources in the cluster. This initial informer cache sync is so huge
that it causes a sudden massive spike in the memory usage of the
operator. This causes OOMKilled failures for the ocs-operator pods.

So ocs-operator will now set the leaderElectionNamespace & the Manager
cache to the namespace where the operator is deployed. This will
ensure that the initial cache sync will only sync the resources in the
namespace where the operator is deployed.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2126626

### DIFF
--- a/main.go
+++ b/main.go
@@ -56,6 +56,7 @@ import (
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 	"k8s.io/client-go/util/retry"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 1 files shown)


iamniting (approved):
/lgtm

PR #1814, opened 2022-09-15
Various QoL changes from jarrpa-dev

jarrpa:
Just a few changes that help my development workflow, which I think others would also benefit from. I would like to stop maintaining these changes just in my working branches. :stuck_out_tongue_winking_eye: 

Signed-off-by: Jose A. Rivera <jarrpa@redhat.com>

PR #1818, opened 2022-09-16
Various fixes to initial GitHub Actions

jarrpa:
...that merged too early in #1814 ;)

The golangci-lint changes were tricky, but basically it comes down to some weirdness in running it using Go 1.19 versus Go 1.18 for golangci-lint v1.49.0. Specifically, it was this change in 1.48.0 that was causing us so much pain: https://github.com/golangci/golangci-lint/pull/3002

In addition, the original configuration listed *both* `go-version` and `go-version-file` configuration options, and `go-version` was being taken as the preferred one. Unfortunately, our code still fails against Go 1.19 for some reason, so for the time being we're forcing the Action to target Go 1.18 across the job's config matrix. I kept the matrix at `["1.18","1.19"]` just so we don't have to change it later when we do start moving to 1.19. 

### DIFF
--- a/.github/workflows/ocs-operator-ci.yaml
+++ b/.github/workflows/ocs-operator-ci.yaml
@@ -19,7 +19,7 @@ jobs:
           fetch-depth: 0
 
       - name: Run shellcheck
-        run: make shellcheck
+        run: make shellcheck-test
 
   golangci-lint:
     name: golangci-lint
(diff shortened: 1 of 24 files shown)


.github/workflows/ocs-operator-ci.yaml
```diff
@@ -69,7 +73,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go: ["1.17", "1.18"]
+        go: ["1.18", "1.19"]
         make-target: ["verify-deps", "verify-generated", "verify-latest-csv", "verify-operator-bundle", "verify-latest-deploy-yaml"]
```
malayparida2000, line 77: Do we need this make-target matrix any more?

jarrpa: Good catch!

jarrpa: Removed, and also fixed some of the jobs that had wrong targets :joy:

malayparida2000 (approved):
/lgtm

PR #1828, opened 2022-09-28
Move to file based catalog & containerized build process

malayparida2000:
1.  The SQLite-based catalog format, including the related CLI commands, is a deprecated feature. So we are moving to the latest file-based catalog.
2. The process of building images(building of binary) is now wholly containerized. And the old index & appregistry methods are removed.


malayparida2000:
/hold

I needs a lot of work on openshift-ci side to make this feasible. So this will take time. Putting on hold till then.

PR #1833, opened 2022-10-06
Define registries for Rook and Noobaa images

rexagod:
Lack of registry definition in this case causes the `gen-latest-csv` command to fail when dumping them, due to the lack of an interactive TTY.

Folks may not face this issue if they already have the aliases defined in `.cache/containers/short-name-aliases.conf`. However, it's better to be explicit, so that newer systems that don't have such aliases don't crash on this command.

Also refer: https://github.com/containers/image/blob/main/docs/containers-registries.conf.5.md#note-risk-of-using-unqualified-image-names


hack/common.sh
```diff
@@ -48,8 +48,8 @@ NOOBAA_CSV="$OUTDIR_TEMPLATES/noobaa-csv.yaml"
 ROOK_CSV="$OUTDIR_TEMPLATES/rook-csv.yaml.in"
 OCS_CSV="$OUTDIR_TEMPLATES/ocs-operator.csv.yaml.in"
 
-LATEST_ROOK_IMAGE="rook/ceph:v1.10.0"
-LATEST_NOOBAA_IMAGE="noobaa/noobaa-operator:master-20220916"
+LATEST_ROOK_IMAGE="docker.io/rook/ceph:v1.10.0"
+LATEST_NOOBAA_IMAGE="quay.io/noobaa/noobaa-operator:master-20220916"
```
weirdwiz, line 52: Can we use these variables to construct URL for noobaa and rook like we do for bundle?

https://github.com/red-hat-storage/ocs-operator/blob/316e2b3235bfaf0878a949ad3155c807f58791ec/hack/common.sh#L92

rexagod: ACK, done.

hack/common.sh
```diff
@@ -48,9 +48,13 @@ NOOBAA_CSV="$OUTDIR_TEMPLATES/noobaa-csv.yaml"
...
 OCS_CSV="$OUTDIR_TEMPLATES/ocs-operator.csv.yaml.in"
 
-LATEST_ROOK_IMAGE="rook/ceph:v1.10.0"
-LATEST_NOOBAA_IMAGE="noobaa/noobaa-operator:master-20220916"
-LATEST_NOOBAA_CORE_IMAGE="noobaa/noobaa-core:master-20220916"
+DEFAULT_NOOBAA_BUNDLE_REGISTRY="quay.io/noobaa"
+DEFAULT_NOOBAA_BUNDLE_NAME="noobaa-operator-bundle"
+DEFAULT_NOOBAA_BUNDLE_TAG="master-20220916"
+
+LATEST_ROOK_IMAGE="docker.io/rook/ceph:v1.10.0"
```
umangachapagain, line 55: @travisn Do we still use dockerhub? Didn't we move to quay?

hack/common.sh
```diff
@@ -48,9 +48,13 @@ NOOBAA_CSV="$OUTDIR_TEMPLATES/noobaa-csv.yaml"
...
+LATEST_NOOBAA_IMAGE="${DEFAULT_NOOBAA_BUNDLE_REGISTRY}/noobaa-operator:${DEFAULT_NOOBAA_BUNDLE_TAG}"
+LATEST_NOOBAA_CORE_IMAGE="noobaa/noobaa-core:${DEFAULT_NOOBAA_BUNDLE_TAG}"
```
umangachapagain, line 57: This would need the default registry prefix too, right?

PR #1836, opened 2022-10-11
Pick tolerations applied to cephtoolbox on ocsinitialization & structure the cephtoolbox tolerations on storagecluster CR

malayparida2000:
Resolves-https://bugzilla.redhat.com/show_bug.cgi?id=2131220

In 4.11 in this [#1602](https://github.com/red-hat-storage/ocs-operator/pull/1602), We moved the toolbox enable/disable & reconcile logic to the storagecluster side. But the implementation to still pick up tolerations added to the ocsinitialization CR was not there. So adding it now.

The tolerations on the same thing on the storagecluster side are also being structured properly to maintain uniformity.
And also adding missing implementation for the reconcile strategy.


api/v1/storagecluster_types.go
```diff
@@ -211,13 +211,10 @@ type ManageCephObjectStoreUsers struct {
...
+// ManageCephToolbox defines how to reconcile Ceph toolbox & the placement configuration
 type ManageCephToolbox struct {
-	ReconcileStrategy string `json:"reconcileStrategy,omitempty"` // default behavior here is effectively `ignore`
-
-	// Tolerations if specified set toolbox ceph tools pod tolerations
-	// Defaults to empty
-	// +optional
-	Tolerations []corev1.Toleration `json:"cephTolerations,omitempty"`
+	ReconcileStrategy string               `json:"reconcileStrategy,omitempty"`
+	Placement         rookCephv1.Placement `json:"placement,omitempty"`
```
umangachapagain, line 217: We need to move this to `StorageClusterSpec.Placement` if we want to be consistent with other components.
None of the Manage* struct have a placement spec inside it.

controllers/storagecluster/cephtoolbox.go
```diff
@@ -31,11 +31,21 @@ func (r *StorageClusterReconciler) ensureToolsDeployment(sc *ocsv1.StorageCluste
...
+	ocsinit := &ocsv1.OCSInitialization{}
+	err := r.Client.Get(context.TODO(), types.NamespacedName{Name: "ocsinit", Namespace: namespace}, ocsinit)
+	if err != nil && !errors.IsNotFound(err) {
+		return err
+	}
+
+	// Get tolerations from both the storagecluster CR & ocsinitialization CR
+	// Keeping this for ability to add tolerations from ocsinitialization CR
+	tolerations = append(tolerations, sc.Spec.ManagedResources.CephToolbox.Placement.Tolerations...)
+	tolerations = append(tolerations, ocsinit.Spec.Tolerations...)
```
umangachapagain, line 50: It'd be better to merge it such that there are no duplicate toleration.

umangachapagain (changes requested):

api/v1/storagecluster_types.go
malayparida2000: So we can just take the Placement info from the Storagecluster Spec.Placement.
We can take the tolerations out of that & apply that onto the toolbox. That would work right?
We would no longer need this field for Placement/Toleration inside the ManageCephToolBox.

malayparida2000: Done

controllers/storagecluster/cephtoolbox.go
malayparida2000: Done

api/v1/storagecluster_types.go
```diff
@@ -211,13 +211,9 @@ type ManageCephObjectStoreUsers struct {
 	ReconcileStrategy string `json:"reconcileStrategy,omitempty"`
 }
 
+// ManageCephToolbox defines how to reconcile Ceph toolbox & the placement configuration
```
umangachapagain, line 214:
```suggestion
// ManageCephToolbox defines how to reconcile Ceph toolbox
```

controllers/storagecluster/cephtoolbox.go
```diff
@@ -76,3 +87,15 @@ func (r *StorageClusterReconciler) ensureToolsDeployment(sc *ocsv1.StorageCluste
 	}
 	return nil
 }
+
+func removeDuplicateTolerations(tolerations []corev1.Toleration) []corev1.Toleration {
```
umangachapagain, line 91: please add unit test for this

umangachapagain (changes requested):

api/v1/storagecluster_types.go
malayparida2000: Done

malayparida2000:
@umangachapagain PTAL, We need to merge this today as this is targeted for 4.11.3

umangachapagain (approved):

PR #1854, opened 2022-10-18
Don't use rook master tag in Job

subhamkrai:
we should not be using rook master tag in init container in product deployments
instead, we should be using rook version from env var.


subhamkrai:
/assign @travisn
cc @agarwal-mudit

subhamkrai:
/retitle Bug 2135736: don't use rook master tag in Job

travisn:
/lgtm
We should consider this for backport to supported versions

PR #1863, opened 2022-10-21
Deploy two mgr pods in a cephcluster

malayparida2000:
Recently in upstream Rook we did make 2 mgrs the default, and it has been very stable so we need the OCS operator to make that the default as well.

Resolves-https://bugzilla.redhat.com/show_bug.cgi?id=2130266

### DIFF
--- a/controllers/defaults/defaults.go
+++ b/controllers/defaults/defaults.go
@@ -20,6 +20,8 @@ const (
 )
 
 var (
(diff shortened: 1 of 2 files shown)


iamniting (approved):
We may want to introduce API modifications to the storage cluster to replace the default count. This can be done in the future.

PR #1865, opened 2022-10-25
Stop using Topology Spread Constraints when Arbiter is enabled

malayparida2000:
Using Topology Spread Constraints with Arbiter creates problems. As we use the maxskew as 1 and arbiter zone never has any pods, so after 1 pod each is scheduled in the data zones, the maxskew is violated and the other pods(osd prepare pods) stay in pending state.

Resolves-https://bugzilla.redhat.com/show_bug.cgi?id=2091922


malayparida2000:
/hold

We still have to test this on a stretch cluster setup to verify the solution

PR #1886, opened 2022-11-21
must-gather: collect journal and kernel logs for all nodes

yati1998:
This commit collects the journal and kernel logs
seperately according to the nodes.

Signed-off-by: yati1998 <ypadia@redhat.com>

### DIFF
--- a/must-gather/collection-scripts/gather_ceph_resources
+++ b/must-gather/collection-scripts/gather_ceph_resources
@@ -398,37 +398,22 @@ for ns in $namespaces; do
 
     journal_collection(){
         printf "collecting journal logs from node %s \n" "${node}" | tee -a  "${BASE_COLLECTION_PATH}"/gather-debug.log
-        oc rsync -n "${ns}" "$(oc get pods -n "${ns}" -l "${node//./}"-debug='ready' --no-headers | awk '{print $1}')":/host/var/log/journal.gz "${JOURNAL_OUTPUT_DIR}"
-        oc rsync -n "${ns}" "$(oc get pods -n "${ns}" -l "${node//./}"-debug='ready' --no-headers | awk '{print $1}')":/host/var/log/kernel.gz "${KERNEL_OUTPUT_DIR}"
-    }
-
-    delete_journal_log(){
-        for node in ${nodes}; do
-        printf "Deleting journal logs from node %s \n"  "${node}"
-        oc debug nodes/"${node}" << CMDS
-            chroot /host
-            cd var/log
-            rm -rf journal.gz
-            rm -rf kernel.gz
-            exit
-CMDS
-    done
-    }
-    # creating a counter variable for collecting PID in array
-    pids_log=()
+        oc rsync -n "${ns}" "$(oc get pods -n "${ns}" -l "${node//./}"-debug='ready' --no-headers | awk '{print $1}')":/host/var/log/journal_"${node}".gz "${JOURNAL_OUTPUT_DIR}"
+        oc rsync -n "${ns}" "$(oc get pods -n "${ns}" -l "${node//./}"-debug='ready' --no-headers | awk '{print $1}')":/host/var/log/kernel_"${node}".gz "${KERNEL_OUTPUT_DIR}"
+    }  
     
     for node in ${nodes}; do
-        printf "Generating journal logs from node %s \n"  "${node}"
+        printf "Generating journal logs from node %s \n"  "${node}" | tee -a  "${BASE_COLLECTION_PATH}"/gather-debug.log
         oc debug nodes/"${node}" << CMDS
             chroot /host
             cd var/log
-            journalctl --since "2 days ago" | gzip > journal.gz
-            journalctl -k -o short-iso-precise --utc --no-hostname | gzip > kernel.gz
+            journalctl --since "2 days ago" | gzip > journal_"${node}".gz
+            journalctl -k -o short-iso-precise --utc --no-hostname | gzip > kernel_"${node}".gz
             exit
 CMDS
     done
-
-    # Collecting ceph crash dump
+    # creating a counter variable for collecting PID in array
+    pids_log=()
     for node in ${nodes}; do
         printf "collecting crash, journal and volume logs from node %s \n"  "${node}" | tee -a  "${BASE_COLLECTION_PATH}"/gather-debug.log
         CRASH_OUTPUT_DIR=${CEPH_COLLECTION_PATH}/crash_${node}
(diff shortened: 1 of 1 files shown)


agarwal-mudit (approved):
/lgtm

agarwal-mudit (approved):
/lgtm

must-gather/collection-scripts/gather_ceph_resources
```diff
@@ -444,13 +445,18 @@ CMDS
         crash_collection &
         pids_log+=($!)
         journal_collection &
-        pids_log+=($!)
+        pids_journal+=($!)
```
idryomov, line 448: I'm not following the purpose of introducing `pids_journal` array in addition to existing `pids_log` array.  Aren't the respective processes waited for in exactly the same place?

idryomov (changes requested):

must-gather/collection-scripts/gather_ceph_resources
yati1998: So we have some pid's limit, due to more volume collection, the pids reach their limit and logs are not getting collected further. Hence, we need to introduce a separate pid for journal logs.

idryomov: I don't see how this changes anything wrt. any potential limit (not sure what are referring to exactly).  The same number of processes is launched, it's just that instead of adding all their PIDs to the same array, you are adding `journal_collection` PID to the new array and then almost immediately wait for it.

must-gather/collection-scripts/gather_ceph_resources
```diff
@@ -444,12 +445,17 @@ CMDS
...
             # wait for all pids
             echo "waiting for ${pids_log[*]} to terminate" | tee -a  "${BASE_COLLECTION_PATH}"/gather-debug.log
             wait "${pids_log[@]}"
         fi
+        if [ -n "${pids_journal[*]}" ]; then
+            # wait for all pids
+            echo "waiting for ${pids_journal[*]} to terminate" | tee -a  "${BASE_COLLECTION_PATH}"/gather-debug.log
+            wait "${pids_journal[@]}"
+        fi
         delete_journal_log
```
idryomov, line 459: I don't think this is the right place to call `delete_journal_log`.  Here it's called for *each* node -- when it already iterates over all nodes internally.  See my previous comment:
>  Perhaps you meant to move it outside the loop (since it has a loop of its own) or alternatively make `delete_journal_log` operate on a single node?

idryomov (changes requested):

must-gather/collection-scripts/gather_ceph_resources
yati1998: sorry, missed it. Thanks.

idryomov: I'd suggest renaming this function to `delete_journal_logs` so that there is at least some hint in the name.  Or, better yet, just open-code it similar to journal log generation which is also open-coded: https://github.com/red-hat-storage/ocs-operator/blob/fb00fe37905a8a66e52b79cf97e825734aa9cedb/must-gather/collection-scripts/gather_ceph_resources#L419-L428

yati1998: This is not deleting the logs, it's about generating the logs, open code to this seems fine to me.

yati1998: The print msg describes it all..Let me know what you think @idryomov

idryomov: > The print msg describes it all..Let me know what you think @idryomov

What print msg?

yati1998: `printf "Generating journal logs from node %s \n"  "${node}" | tee -a  "${BASE_COLLECTION_PATH}"/gather-debug.log `

yati1998: line 420 in above code

idryomov: I understand the different between generation and deletion ;)

What I was pointing out is that journal log generation is currently open-coded, while journal log deletion is wrapped into a function with a rather confusing name which is probably what caused you to place it into the loop body by mistake.  I'm suggesting to open-code journal log deletion as well -- thus making that mistake very hard to repeat.

yati1998:
@idryomov can you please review again.

idryomov (commented):
`delete_journal_log`-related bit looks good but I still don't understand the need for a new `pids_journal` array: https://github.com/red-hat-storage/ocs-operator/pull/1886/files#r1029105197

idryomov:
> with the pids in same same array, the journal logs collection is not reached.

What do you mean by "not reached"?

> That I think can be due to reached limit and hence can't accumulate more pids and resulting in not executing the journal collection part. With this change in new array for journal log, this is resolved.

What limit do you have in mind?  The number of PIDs is definitely not the issue here because the new code spawns the *same* number of processes.  The only difference is that one of the returned PIDs is put into a separate array.

must-gather/collection-scripts/gather_ceph_resources
```diff
@@ -398,37 +398,24 @@ for ns in $namespaces; do
...
+            journalctl -k -o short-iso-precise --utc --no-hostname | gzip > kernel_"${node}".gz
             exit
 CMDS
     done
-
-    # Collecting ceph crash dump
+    # creating a counter variable for collecting PID in array
+    pids_log=()
+    # Counter variable for collecting PID for journal logs
+    pids_journal=()
```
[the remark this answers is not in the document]
yati1998: done with the changes, would be fine now.

idryomov (approved):
A ton of improvements could be made here but, considered purely as a fix for `journal.gz` and `kernel.gz` getting overwritten when there is more than one node, LGTM!

must-gather/collection-scripts/gather_ceph_resources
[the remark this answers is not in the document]
idryomov: FWIW this is not addressed: you removed `pids_journal` (good) but `pids_log` would still contain stale PIDs from previous iterations (bad).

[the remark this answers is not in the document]
yati1998: it is getting terminated at the end of every iteration, not sure which stale image you are talking about.

yati1998:
> A ton of improvements could be made here but, considered purely as a fix for `journal.gz` and `kernel.gz` getting overwritten when there is more than one node, LGTM!

yeah sure, will check that and send subsequent PRs.

must-gather/collection-scripts/gather_ceph_resources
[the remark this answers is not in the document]
idryomov: > it is getting terminated at the end of every iteration, 

It's waiting for processes to terminate on their own, yes ...

> not sure which stale image you are talking about.

...  but process IDs (integers, not images) remain in `pids_log` array.

PR #1895, opened 2022-12-16
disable webhook for rook untill offically supported

subhamkrai:
Signed-off-by: subhamkrai <srai@redhat.com>

PR #1896, opened 2022-12-16
Bug 2154165: disable webhook for rook

subhamkrai:
disabling webhook for rook untill is officially supported

Backport of https://github.com/red-hat-storage/ocs-operator/pull/1895


iamniting:
I would prefer to use cherry-pick robot here

subhamkrai:
> I would prefer to use cherry-pick robot here

yes but I created manually BP to avoid merge conflict if any and since I'll be off for a week and this is a blocker from rook that's why

iamniting (approved):
/lgtm

PR #1900, opened 2022-12-16
Bug 2154164: disable webhook for rook untill we support

subhamkrai:
disable webhook untill we say we support it


subhamkrai:
duplicate pr

PR #1901, opened 2022-12-21
Change failure domain value to hostname from host

malayparida2000:
This is to align the failure domain standards with K8s.
This is also to solve the problem of provisioning with replica 1
topology aware provisioning.

### DIFF
--- a/controllers/storagecluster/storageclasses.go
+++ b/controllers/storagecluster/storageclasses.go
@@ -443,11 +443,15 @@ func getTopologyConstrainedPools(initData *ocsv1.StorageCluster) string {
 
 	var topologyConstrainedPools []topologyConstrainedPool
 	for _, failureDomainValue := range initData.Status.FailureDomainValues {
+		failureDomain := initData.Status.FailureDomain
+		if failureDomain == "host" {
+			failureDomain = "hostname"
+		}
 		topologyConstrainedPools = append(topologyConstrainedPools, topologyConstrainedPool{
 			PoolName: generateNameForNonResilientCephBlockPool(initData, failureDomainValue),
 			DomainSegments: []topologySegment{
 				{
-					DomainLabel: initData.Status.FailureDomain,
+					DomainLabel: failureDomain,
 					DomainValue: failureDomainValue,
 				},
 			},


controllers/storagecluster/topology.go
```diff
@@ -83,7 +83,7 @@ func setFailureDomain(sc *ocsv1.StorageCluster) {
 	// But if FlexiableScaling is enabled then we select host as failure domain
```
humblec, line 83: may be good to adjust the part "host" in above comment or make it detailed.

travisn (commented):
This change doesn't affect the `failureDomain` that is set on the ceph pools for rbd, cephfs, or object store, right? Ceph would still expect `failureDomain: host`

malayparida2000:
> This change doesn't affect the `failureDomain` that is set on the ceph pools for rbd, cephfs, or object store, right? Ceph would still expect `failureDomain: host`

It will definitely affect the failure domain everywhere as all the things you mentioned copy the failure domain from the storage cluster status only. This sounds problematic then

malayparida2000:
/hold

This is just a test PR for getting an image on 4.12 release branch to test on the IBM cloud cluster.

travisn:
> > This change doesn't affect the `failureDomain` that is set on the ceph pools for rbd, cephfs, or object store, right? Ceph would still expect `failureDomain: host`
> 
> It will definitely affect the failure domain everywhere as all the things you mentioned copy the failure domain from the storage cluster status only. This sounds problematic then

So we can be clear on the issue with provisioning that we expect this to fix, where exactly does K8s expect the `failureDomain` to be `hostname`? I'm not aware of this standard.

malayparida2000:
> > > This change doesn't affect the `failureDomain` that is set on the ceph pools for rbd, cephfs, or object store, right? Ceph would still expect `failureDomain: host`
> > 
> > 
> > It will definitely affect the failure domain everywhere as all the things you mentioned copy the failure domain from the storage cluster status only. This sounds problematic then
> 
> So we can be clear on the issue with provisioning that we expect this to fix, where exactly does K8s expect the `failureDomain` to be `hostname`? I'm not aware of this standard.

so the thing is the provisioner is expecting the domainlabel to be hostname, but when we create the topology constrained pools storage class we add the domain labels directly from the storagecluster failure domain, which is set to host(not hostname). So there is this mismatch & provisoning doesnt happen.


This was the exact error

```
Event(v1.ObjectReference{Kind:"PersistentVolumeClaim", Namespace:"openshift-storage", Name:"non-resilient-rbd-pvc", UID:"cabf9c36-0b32-43b9-9f9b-ad680bccbe3a", APIVersion:"v1", ResourceVersion:"3503148", FieldPath:""}): type: 'Warning' reason: 'ProvisioningFailed' failed to provision volume with StorageClass "ocs-storagecluster-ceph-non-resilient-rbd": rpc error: code = Internal desc = none of the topology constrained pools matched requested topology constraints : pools ([{PoolName:ocs-storagecluster-cephblockpool-worker-0 DataPoolName: DomainSegments:[{DomainLabel:host DomainValue:worker-0}]} {PoolName:ocs-storagecluster-cephblockpool-worker-1 DataPoolName: DomainSegments:[{DomainLabel:host DomainValue:worker-1}]} {PoolName:ocs-storagecluster-cephblockpool-worker-2 DataPoolName: DomainSegments:[{DomainLabel:host DomainValue:worker-2}]}]) requested topology ({Requisite:[segments:<key:"[topology.openshift-storage.rbd.csi.ceph.com/hostname](http://topology.openshift-storage.rbd.csi.ceph.com/hostname)" value:"worker-0" >  segments:<key:"[topology.openshift-storage.rbd.csi.ceph.com/hostname](http://topology.openshift-storage.rbd.csi.ceph.com/hostname)" value:"worker-1" >  segments:<key:"[topology.openshift-storage.rbd.csi.ceph.com/hostname](http://topology.openshift-storage.rbd.csi.ceph.com/hostname)" value:"worker-2" > ] Preferred:[segments:<key:"[topology.openshift-storage.rbd.csi.ceph.com/hostname](http://topology.openshift-storage.rbd.csi.ceph.com/hostname)" value:"worker-0" >  segments:<key:"[topology.openshift-storage.rbd.csi.ceph.com/hostname](http://topology.openshift-storage.rbd.csi.ceph.com/hostname)" value:"worker-1" >  segments:<key:"[topology.openshift-storage.rbd.csi.ceph.com/hostname](http://topology.openshift-storage.rbd.csi.ceph.com/hostname)" value:"worker-2" > ] XXX_NoUnkeyedLiteral:{} XXX_unrecognized:[] XXX_sizecache:0})
```

malayparida2000:
Closing this test PR as the fix worked & will raise a PR against main branch.

PR #1904, opened 2022-12-25
hack: fix broken build for ARM64-arch

synarete:
When building over AWS-Graviton, need to use ARM64 (aarch64) protoc binaries. Upstream project does not provides proper ARM64 binaries for protoc-gen-go and protoc-gen-go-grpc so we need to build it from source within temporary directory.

Refs RHSTOR-2044


iamniting (commented):
The changes look good to me, Were these changes tested well? CI does not test it on arm ENV so we need to make sure they work well.

hack/common.sh
```diff
@@ -5,9 +5,9 @@
 
 GO111MODULE="on"
 GOPROXY="https://proxy.golang.org"
-GOROOT="${GOROOT:-go env GOROOT}"
+GOROOT="${GOROOT:-$(go env GOROOT)}"
```
iamniting, line 8: from the failure in the CI it looks like we don't have go in the openshift CI. We also need to figure this out.

synarete:
> The changes look good to me, Were these changes tested well? CI does not test it on arm ENV so we need to make sure they work well.

Currently, I am testing the build process over ARM64 using RH's beaker machines. This is just the first step in a more comprehensive cycle of adjusting ocs-operator to ARM64 environment (specifically, AWS Graviton). See https://github.com/red-hat-storage/ocs-operator/issues/1906 for more details.

hack/common.sh
synarete: It is a chicken-and-egg problem:  you first need to be able to build ocs-operator over ARM64 before you may test it via CI.

PR #1905, opened 2022-12-25
[WIP] Adding alert comparison logic

aruniiird:
A standalone program to compare 'PrometheusRule' files and display the differences.


hack/comparealerts/main.go
```diff
@@ -0,0 +1,48 @@
...
+func init() {
+	alertDiffFlagSet.StringVar(&fileUp, "upstream-alert-file", "", "provide an upstream alert file")
+	alertDiffFlagSet.StringVar(&fileDown, "downstream-alert-file", "", "provide a downstream alert file")
+}
+
+func main() {
+	alertDiffFlagSet.Parse(os.Args[1:])
+	klog.Info("Upstream:", fileUp)
+	klog.Info("Downstream:", fileDown)
+	promRuleOne, err := types.NewPrometheusRuleFromFile(fileUp)
```
jmolmo, line 28: The kind of file is not verified, that could cause some confusion, because if for example, one file is a "generic yaml" ., and the other one a right prometheus rule file , the tool will say that all the alerts in the right file are not in the first one.

I would add another additional check, something like:
```
	if promRuleOne.TypeMeta.Kind != "PrometheusRule" {
		klog.Error("Upstream file does not contain a PrometheusRule")
		return
	}
```
The same for the downstream file.

jmolmo (changes requested):
Apart of the validation of the content of the input files, I miss some default behavior ( for example if not provided parameters, or the command is used with the help flag) to provide some context and explanation about what is the use of the tool.

aruniiird:
Moved the changes/pr to a separate/independent location: https://github.com/red-hat-storage/odf-monitoring-tools/pull/2

PR #1908, opened 2023-01-06
remove namespace deps from the bundle

iamniting:
These changes will allow us to install operator in any namespace.


umangachapagain (changes requested):
Changes look good. Squash commits since these are not independent changes. Keep only the generated changes separate.

Please add E2E tests to ensure nothing breaks when deploying in other namespaces.

iamniting:
I have tested this locally and it works well

```
$ oc get storageclusters.ocs.openshift.io -n nigoyal
ocs-storagecluster   60m   Ready              2023-01-11T07:24:10Z   4.12.0
```

iamniting:
> Changes look good. Squash commits since these are not independent changes. Keep only the generated changes separate.

done

> Please add E2E tests to ensure nothing breaks when deploying in other namespaces.

I'm thinking about adding the e2e after all of these other changes are completed. Though I have only tested this locally for the time being.

umangachapagain:
> 
> > Please add E2E tests to ensure nothing breaks when deploying in other namespaces.
> 
> I'm thinking about adding the e2e after all of these other changes are completed. Though I have only tested this locally for the time being.
>
At least some basic sanity tests would be better for such a change. You can always add to it later when more changes are done. Also, it'll help valid newer changes on top of this change.

iamniting:
> > > Please add E2E tests to ensure nothing breaks when deploying in other namespaces.
> > 
> > 
> > I'm thinking about adding the e2e after all of these other changes are completed. Though I have only tested this locally for the time being.
> 
> At least some basic sanity tests would be better for such a change. You can always add to it later when more changes are done. Also, it'll help valid newer changes on top of this change.

Hi @umangachapagain I created a PR in the release repo, But to get it tested properly we will require these changes to be merged. https://github.com/openshift/release/pull/35362

umangachapagain (approved):
/lgtm

PR #1909, opened 2023-01-06
Deploy 1 mgr deamon by default & Allow deploying 2 ceph mgr daemons(1 active and 1 standby)

malayparida2000:
Jira- https://issues.redhat.com/browse/RHSTOR-3860

Ocs operator will deploy 1 ceph mgr daemon by default. User can deploy 2 ceph mgr daemons(1 active and 1 standby) by
setting `spec.mgr.enableActivePassive` to true. When the active mgr fails, the standby mgr will be promoted to active mgr.
This will provide a highly available mgr.

There can be a delay (15-30s) for ceph to promote the passive mgr when the active mgr goes down.
Rook queries ceph every 20s if there is a new active mgr & it will update the required services to direct traffic to the new active mgr.

This PR  will also undo the changes done here https://github.com/red-hat-storage/ocs-operator/pull/1863


api/v1/storagecluster_types.go
```diff
@@ -48,8 +48,11 @@ type StorageClusterSpec struct {
 	StorageDeviceSets  []StorageDeviceSet                     `json:"storageDeviceSets,omitempty"`
 	MonPVCTemplate     *corev1.PersistentVolumeClaim          `json:"monPVCTemplate,omitempty"`
 	MonDataDirHostPath string                                 `json:"monDataDirHostPath,omitempty"`
-	MultiCloudGateway  *MultiCloudGatewaySpec                 `json:"multiCloudGateway,omitempty"`
-	NFS                *NFSSpec                               `json:"nfs,omitempty"`
+	// MgrActivePassive can be set as true to deploy 2 ceph manager pods, one active and one standby
+	// Ceph will promote the standby mgr when the active mgr goes down due to any reason
+	MgrActivePassive  bool                   `json:"mgrActivePassive,omitempty"`
```
iamniting, line 53: I think we should rename it to the `Enable ActivePassiveMgr`

iamniting (changes requested):

api/v1/storagecluster_types.go
travisn: Since in the future there could also be other mgr settings, another idea is to define a struct for MgrSpec, and put this setting there. For example, the setting would then be: `mgr.enableActivePassive`

iamniting: +1

malayparida2000: done

malayparida2000:
@iamniting PTAL

api/v1/storagecluster_types.go
```diff
@@ -216,6 +217,13 @@ type ManageCephToolbox struct {
 	ReconcileStrategy string `json:"reconcileStrategy,omitempty"`
 }
 
+// MgrSettings defines the settings for the Ceph Manager
+type MgrSettings struct {
```
iamniting, line 221:
```suggestion
type MgrSpec struct {
```

Lets rename this

iamniting (changes requested):

api/v1/storagecluster_types.go
malayparida2000: Avoided naming it as MgrSpec as MgrSpec is an actual struct in rook api, And it has its specific field. Naming it MgrSpec may create confusion. See here
https://pkg.go.dev/github.com/rook/rook@v1.10.0/pkg/apis/ceph.rook.io/v1#MgrSpec

travisn: MgrSpec seems more consistent, other structs defined in these types have the `Spec` suffix even if it is the same as a Rook type. Since it's in a different package, there isn't a conflict.

malayparida2000: Ack, Done

iamniting (approved):
/lgtm

PR #1910, opened 2023-01-07
Warn in logs if both in-transit encryption & compression are enabled

malayparida2000:
This adds a warning in the logs if both in-transit encryption and compression are enabled. This will help the user to understand the behaviour.

Combining encryption with compression reduces the level of security of messages between peers. In case both encryption and compression are enabled, compression setting will be ignored and message will not be compressed. 
Reference-
https://issues.redhat.com/browse/RHSTOR-2516
https://issues.redhat.com/browse/RHSTOR-3328

### DIFF
--- a/controllers/storagecluster/reconcile.go
+++ b/controllers/storagecluster/reconcile.go
@@ -532,6 +532,21 @@ func (r *StorageClusterReconciler) reconcilePhases(
 		}
 	}
 
+	// For security of encrypted messagges, When both encryption and compression are enabled,
+	// compression setting will be ignored and message will not be compressed.
+	// Refer https://docs.ceph.com/en/quincy/rados/configuration/msgr2/#confval-ms_compress_secure
+	networkSpec := instance.Spec.Network
+	if networkSpec != nil && networkSpec.Connections != nil &&
+		networkSpec.Connections.Encryption != nil && networkSpec.Connections.Encryption.Enabled &&
+		networkSpec.Connections.Compression != nil && networkSpec.Connections.Compression.Enabled {
+		r.Log.Info("Both in-transit encryption & compression are enabled. " +
+			"To protect security of encrypted messages ceph will ignore compression")
+		r.recorder.ReportIfNotPresent(instance, corev1.EventTypeWarning, "EncryptionAndCompressionEnabled",
+			"Both in-transit encryption & compression are enabled. "+
+				"To protect security of encrypted messages ceph will ignore compression")
+
+	}
+
 	// Ensure that verbose logging is enabled when RBD mirroring is enabled
 	// TODO: This is a temporary arrangement, this is to be removed when RDR goes to GA
 	result, err := r.ensureRbdMirrorDebugLogging(instance)


controllers/storagecluster/reconcile.go
```diff
@@ -532,6 +532,12 @@ func (r *StorageClusterReconciler) reconcilePhases(
 		}
 	}
 
+	// For security of encrypted messages, When both encryption and compression are enabled, compression setting will be ignored and message will not be compressed.
+	networkSpec := instance.Spec.Network
+	if networkSpec != nil && networkSpec.Connections != nil && networkSpec.Connections.Encryption != nil && networkSpec.Connections.Encryption.Enabled && networkSpec.Connections.Compression != nil && networkSpec.Connections.Compression.Enabled {
```
iamniting, line 537:
```suggestion
	if networkSpec != nil && networkSpec.Connections != nil && 
	    networkSpec.Connections.Encryption != nil && networkSpec.Connections.Encryption.Enabled && 
	    networkSpec.Connections.Compression != nil && networkSpec.Connections.Compression.Enabled {
```
Can we break this into multiple lines something like this?

controllers/storagecluster/reconcile.go
```diff
@@ -532,6 +532,12 @@ func (r *StorageClusterReconciler) reconcilePhases(
...
+		r.Log.Info("Both in-transit encryption & compression are enabled. To protect security of encrypted messages ceph will ignore the compression setting by default here. To override this behavior,refer https://docs.ceph.com/en/quincy/rados/configuration/msgr2/#confval-ms_compress_secure")
```
iamniting, line 538: Let's split this log line into multiple lines. Also, I don't think we should have a ref to the upstream doc here.

iamniting (changes requested):

controllers/storagecluster/reconcile.go
malayparida2000: @agarwal-mudit Can you clarify if it's appropriate to pass this ceph doc in the logs? for overriding the default? I believe What Nitin is saying is true & we shouldn't pass this as overriding what ceph does by default may cause problems.

malayparida2000: As discussed with Mudit, Removing the doc link.

malayparida2000:
@iamniting @travisn PTAL

controllers/storagecluster/reconcile.go
```diff
@@ -532,6 +532,17 @@ func (r *StorageClusterReconciler) reconcilePhases(
...
 
+	// For security of encrypted messages, When both encryption and compression are enabled,
+	// compression setting will be ignored and message will not be compressed.
+	// Refer https://docs.ceph.com/en/quincy/rados/configuration/msgr2/#confval-ms_compress_secure
+	networkSpec := instance.Spec.Network
+	if networkSpec != nil && networkSpec.Connections != nil &&
+		networkSpec.Connections.Encryption != nil && networkSpec.Connections.Encryption.Enabled &&
+		networkSpec.Connections.Compression != nil && networkSpec.Connections.Compression.Enabled {
+		r.Log.Info("Both in-transit encryption & compression are enabled. ")
+		r.Log.Info("To protect security of encrypted messages ceph will ignore the compression setting by default here.")
```
iamniting, line 543: Divide the line into 2 but do not use r.log twice. I would also like to have an k8s event for the same.

iamniting (changes requested):

controllers/storagecluster/reconcile.go
```diff
@@ -532,6 +532,17 @@ func (r *StorageClusterReconciler) reconcilePhases(
...
+	}
```
travisn, line 544: @malayparida2000 Where is the compression setting ignored? I only see the logging added here.

iamniting: Is not that going to be handled in the rook?

malayparida2000: @travisn  Ceph will ignore the compression setting, see here https://docs.ceph.com/en/quincy/rados/configuration/msgr2/#confval-ms_compress_secure

travisn: Ah I missed that, thanks

iamniting:
The PR needs few changes which i have discussed with Malay, He will push the changes once he is back.

iamniting (approved):
/lgtm
/hold Malay will update the results of testing

Madhu-1:
@malayparida2000 have we tested this feature by creating both cephfs PVC and rbd pvc and pods? just checking to make sure everything works as expected.

malayparida2000:
@Madhu-1 If we create the pvc they are bound, then if we try to create a pod to use the pvc the pods stay forver in container creation as of now. I belive it has to do something with the bugs that Travis & You were discussing on the chat room. This PR is just finishing up the code changes work required on OCS side for the epic in advance, As the ocs ooperator already passes its network spec directly to cephcluster we didn't require any code changes here.

Madhu-1:
> @Madhu-1 If we create the pvc they are bound, then if we try to create a pod to use the pvc the pods stay forver in container creation as of now. I belive it has to do something with the bugs that Travis & You were discussing on the chat room. This PR is just finishing up the code changes work required on OCS side for the epic in advance, As the ocs ooperator already passes its network spec directly to cephcluster we didn't require any code changes here.

Thanks for checking, IMO its good to validate the complete workflow to make sure the feature works

PR #1911, opened 2023-01-07
fix io/ioutil deprecation warning from golangci-lint run

malayparida2000:
io/ioutil" has been deprecated since Go 1.16: As of Go 1.16, the same functionality is now provided by package io or package os, and those implementations should be preferred in new code.


iamniting (approved):
/lgtm

PR #1914, opened 2023-01-11
Build ocs-operator over arm64 machine

synarete:
Build ocs-operator over ARM64 (aarch64) machine:
- Fixes cases where using build tools from wrong arch.
- Fixes cases where using base image which is not suitable for ARM64

Tested over AWS-Graviton vm.


umangachapagain:
@synarete Do we know why tests are failing? I hope it is not some CI limitation.

synarete:
> @synarete Do we know why tests are failing? I hope it is not some CI limitation.

@umangachapagain I am trying to figure it out but could not find the root cause so far. Any ideas?

PR #1917, opened 2023-01-16
Add fusion operator bundle

umangachapagain:


hack/source-manifests.sh
```diff
@@ -77,35 +75,59 @@ function dump_rook_csv() {
...
+	rm -rf $fcs_crds_outdir
+	mkdir -p $fcs_crds_outdir
+
+	gen_args="generate kustomize manifests --input-dir config/manifests/fcs-operator --output-dir config/manifests/fcs-operator --package fcs-operator -q"
+	# shellcheck disable=SC2086
+	$OPERATOR_SDK $gen_args
+	pushd config/manager
+	$KUSTOMIZE edit set image ocs-dev/ocs-operator="$OCS_IMAGE"
+	popd
+	$KUSTOMIZE build config/manifests/fcs-operator | $OPERATOR_SDK generate bundle -q --output-dir deploy/fcs-operator --kustomize-dir config/manifests/fcs-operator --package fcs-operator --version "$CSV_VERSION"
```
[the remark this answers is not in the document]
umangachapagain: Added.

hack/verify-operator-bundle.sh
```diff
@@ -6,3 +6,5 @@ source hack/ensure-operator-sdk.sh
 source hack/docker-common.sh
 
 ./"${OPERATOR_SDK}" bundle validate "$(dirname $OCS_FINAL_DIR)" --verbose
+./"${OPERATOR_SDK}" bundle validate "$(dirname $FCS_FINAL_DIR)" --verbose
```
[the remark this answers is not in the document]
umangachapagain: Added a check. So now it is expected to fail if bundle doesn't exist.

iamniting (approved):
/lgtm
/hold for malay to take a look

deploy/Dockerfile
```diff
@@ -7,18 +7,13 @@ LABEL operators.operatorframework.io.bundle.package.v1=ocs-operator
...
 
-COPY bundle/manifests /manifests/
-COPY bundle/metadata/annotations.yaml /metadata/annotations.yaml
+COPY ocs-operator/manifests /manifests/
+COPY ocs-operator/metadata/annotations.yaml /metadata/annotations.yaml
 
 # These are three labels needed to control how the pipeline should handle this container image
 # This first label tells the pipeline that this is a bundle image and should be
 # delivered via an index image
 LABEL com.redhat.delivery.operator.bundle=true
```
malayparida2000, line 16: Do we need these labels anymore? They were here when we didn't have separate build methods for the bundle Downstream, But now the build team keeps their own separate build Dockerfiles, So I guess we can do away with these labels.

hack/common.sh
```diff
@@ -41,7 +41,7 @@ GINKGO_TEST_SUITE="${GINKGO_TEST_SUITE:-ocs}"
 # This env var allows developers to point to a custom oc tool that isn't in $PATH
 # defaults to just using the 'oc' binary provided in $PATH
 OCS_OC_PATH="${OCS_OC_PATH:-oc}"
-OCS_FINAL_DIR="deploy/bundle/manifests"
+OCS_FINAL_DIR="deploy/ocs-operator/manifests"
```
malayparida2000, line 44: If we have decided to move the directories, Lets stop using the deploy directory. Let's have a directory called bundle at the root of the repo. And store bundle-related stuff inside that.

OCS_FINAL_DIR="bundle/ocs-operator/manifests"

malayparida2000: @iamniting WDYT?

hack/generate-unified-csv.sh
```diff
@@ -42,29 +42,58 @@ fi
...
-	--ocs-image="$OCS_IMAGE" \
-	--ocs-metrics-exporter-image="$OCS_METRICS_EXPORTER_IMAGE" \
-	--ocs-must-gather-image="$OCS_MUST_GATHER_IMAGE" \
-	--crds-directory="$OUTDIR_CRDS" \
-	--manifests-directory=$BUNDLEMANIFESTS_DIR \
-	--olm-bundle-directory="$OCS_FINAL_DIR" \
-	--timestamp="$TIMESTAMP" \
-	--rook-csiaddons-image="$ROOK_CSIADDONS_IMAGE"
+if [ "$FUSION" == "true" ]; then
+	# Merge component-level operators into ocs CSV
```
malayparida2000, line 46: This will be fcs CSV I suppose, instead of ocs CSV

hack/generate-unified-csv.sh
```diff
@@ -42,29 +42,58 @@ fi
...
-	--noobaa-db-image="$NOOBAA_DB_IMAGE" \
-	--ocs-image="$OCS_IMAGE" \
-	--ocs-metrics-exporter-image="$OCS_METRICS_EXPORTER_IMAGE" \
-	--ocs-must-gather-image="$OCS_MUST_GATHER_IMAGE" \
-	--crds-directory="$OUTDIR_CRDS" \
-	--manifests-directory=$BUNDLEMANIFESTS_DIR \
-	--olm-bundle-directory="$OCS_FINAL_DIR" \
-	--timestamp="$TIMESTAMP" \
-	--rook-csiaddons-image="$ROOK_CSIADDONS_IMAGE"
+if [ "$FUSION" == "true" ]; then
```
malayparida2000, line 45: In the if else condition I see just --ocs-csv-filepath being different, Can we do something like,
where we just set a variable like $OCS_CSV_FILEPATH accordingly if FUSION is true or false.
Then just call the CSV_MERGER tool without the if else

Makefile
```diff
@@ -118,11 +119,11 @@ verify-operator-bundle: operator-sdk
 	hack/verify-operator-bundle.sh
 
 operator-bundle:
-	@echo "Building ocs-operator-bundle image"
```
malayparida2000, line 121: Can we keep this info like 
@echo "Building operator-bundle image"

Makefile
```diff
@@ -118,11 +119,11 @@ verify-operator-bundle: operator-sdk
...
+	@echo "FUSION=$(FUSION)"
 	hack/build-operator-bundle.sh
 
 operator-index:
-	@echo "Building ocs-operator-index image"
```
malayparida2000, line 125: Can we keep this info like
@echo "Building operator-index image"

deploy/fcs.Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
+LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
+LABEL operators.operatorframework.io.bundle.package.v1=fcs-operator
+LABEL operators.operatorframework.io.bundle.channels.v1=alpha
+LABEL operators.operatorframework.io.bundle.channel.default.v1=alpha
+
+COPY fcs-operator/manifests /manifests/
+COPY fcs-operator/metadata/annotations.yaml /metadata/annotations.yaml
+
+# These are three labels needed to control how the pipeline should handle this container image
```
malayparida2000, line 13: I think we don't need these labels

hack/common.sh
umangachapagain: Moving things around is not in the scope of this PR. It's just allowing a new bundle to co-exist.

deploy/Dockerfile
umangachapagain: If not anything else, these labels remind us of what is required for the release builds. Also, this PR is not a refactor, so I'd like to retain the existing content for consistency. We can always follow up with a refactor PR if needed.

hack/generate-unified-csv.sh
umangachapagain: Could do. But this allows us to change any arg without adding an explicit if..else per arg to change it in the future.

Makefile
umangachapagain: We have those in the scripts. I see no need of repeating it here.

deploy/Dockerfile
malayparida2000: Hm, Okay. Will handle the cleanups later with refactor PRs

malayparida2000 (approved):
The changes look good to me. 
About the other changes I suggested, as Umanga suggested we will take that up in future refactor PRs.

/hold cancel

PR #1919, opened 2023-01-21
Add new file based catalog as sqlite based catalog is deprecated

malayparida2000:
Sqlite-based catalogs and their related subcommands are deprecated.
Support for them will be removed in a future release of opm.
This catalog is built with the ocs-operator bundle and the nooba
operator bundle.


iamniting (approved):
looks good to me
@umangachapagain what do you think?

umangachapagain (commented):
File based catalog is still in alpha and shouldn't be used.
For now we'll continue using sqlite based catalog.

/do-not-merge

iamniting:
> File based catalog is still in alpha and shouldn't be used. For now we'll continue using sqlite based catalog.

@umangachapagain I don't think so, They are supported as far as I know.

Makefile
```diff
@@ -134,6 +136,21 @@ ocs-registry-master:
...
+	${OPM} render --output=yaml ${NOOBAA_BUNDLE_IMG} > catalog/noobaa-bundle.yaml
+	${OPM} validate catalog
+	${IMG_BUILD_CMD} build --platform ${TARGET_OS}/${TARGET_ARCH} -t ${OCS_CATALOG_IMG} -f Dockerfile.catalog .
+
+catalog-push:
+	@echo "Pushing ocs-operator-catalog image"
+	${IMG_BUILD_CMD} push ${OCS_CATALOG_IMG}
+
+opm: 
+	@echo "Ensureing opm tool"
```
umangachapagain, line 151:
```suggestion
	@echo "Ensuring opm tool"
```

Makefile
```diff
@@ -134,6 +136,21 @@ ocs-registry-master:
 	@echo "Building ocs-registry image in appregistry format using master images for all OCS components"
 	hack/build-master-appregistry.sh
 
+catalog-build: opm
+	@echo "Building ocs-operator-catalog"
+	${OPM} render --output=yaml ${OCS_BUNDLE_IMG} > catalog/ocs-bundle.yaml
+	${OPM} render --output=yaml ${NOOBAA_BUNDLE_IMG} > catalog/noobaa-bundle.yaml
```
umangachapagain, line 142: We don't need noobaa. But, we probably need rook.

Makefile
```diff
@@ -134,6 +136,21 @@ ocs-registry-master:
...
+	${OPM} validate catalog
+	${IMG_BUILD_CMD} build --platform ${TARGET_OS}/${TARGET_ARCH} -t ${OCS_CATALOG_IMG} -f Dockerfile.catalog .
+
+catalog-push:
+	@echo "Pushing ocs-operator-catalog image"
+	${IMG_BUILD_CMD} push ${OCS_CATALOG_IMG}
+
+opm:
```
umangachapagain, line 150: Follow similar pattern like below and move this to end of file
https://github.com/red-hat-storage/ocs-operator/blob/2bfec67855b1a73972cbac052dfd12685f92186d/Makefile#L210-L221

Makefile
```diff
@@ -1,3 +1,5 @@
+include hack/makefile-vars.mk
```
umangachapagain, line 1: This is a refactor. Keep it in a separate PR.

umangachapagain (changes requested):

Makefile
iamniting: We do need noobaa to build a catalog as noobaa CSV needs to be installed too

Makefile
iamniting: Umnaga This is a small change I think we can keep it here in the same PR. Maybe separate commit if you insist.

Makefile
malayparida2000: Yes, Nooba is needed for ocs to be installed and storagecluster to get to ready.

Makefile
malayparida2000: This is kind of a refactor PR only, This is a new way of building a catalog and I want to start with using the makefile vars & move away as much as possible from the common.sh format. And I have kept this in a separate commit so I think this can go here.

Makefile
malayparida2000: Instead of end as this is similar to ensuring operator-sdk moved it to below that. These all I will re arrange properly in a makefile refactor PR later.

malayparida2000:
@umangachapagain @iamniting PTAL, I am waiting on this PR to merge to start work on the upcomming refactors

hack/makefile-vars.mk
```diff
@@ -0,0 +1,25 @@
...
+TARGET_OS ?= linux
+TARGET_ARCH ?= amd64
+
+# Tools versions
+OPM_VERSION ?= v1.26.0
+
+# Tools paths
+OUTDIR ?= build/_output
+OUTDIR_TOOLS= ${OUTDIR}/tools
+OPM ?= ${OUTDIR_TOOLS}/opm-${OPM_VERSION}
```
umangachapagain, line 12: OPM vars in this file is shadowed by vars in script. Also, it's confusing to have same vars in multiple places.
Either move these to the script or move the script content to makefile.

hack/makefile-vars.mk
```diff
@@ -0,0 +1,25 @@
...
+REGISTRY_NAMESPACE ?= ocs-dev
+IMG_TAG ?= latest
+
+OCS_BUNDLE_IMG_NAME ?= ocs-operator-bundle
+OCS_CATALOG_IMG_NAME ?= ocs-operator-catalog
+
+OCS_BUNDLE_IMG ?= ${IMG_REGISTRY}/${REGISTRY_NAMESPACE}/${OCS_BUNDLE_IMG_NAME}:${IMG_TAG}
+OCS_CATALOG_IMG ?= ${IMG_REGISTRY}/${REGISTRY_NAMESPACE}/${OCS_CATALOG_IMG_NAME}:${IMG_TAG}
+
+NOOBAA_BUNDLE_IMG ?= quay.io/noobaa/noobaa-operator-bundle:v5.12.0
```
umangachapagain, line 25: If you want these in makefile, remove it from scripts. Let's not have duplicates.

umangachapagain (changes requested):
Since this is a refactor, I'd like to see things cleaned up here rather than in a follow-up PR.

hack/ensure-opm.sh
```diff
@@ -4,7 +4,7 @@ set -e
 
 source hack/common.sh
 
-OPM_VERSION="v1.12.3"
+OPM_VERSION="v1.26.0"
```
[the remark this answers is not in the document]
malayparida2000: Can't be done in this case right now, As The scripts to ensure these tools are still common.sh dependant.
The task of moving all the ensure tools to new format is out of scope of this PR. 
But I need the OPM path during the calling of the catalog build. This looks inconsistent right now, agreed. but have to keep it like this till I refactor the ensure files.

hack/makefile-vars.mk
malayparida2000: As said above, Can't be done in this case right now. I need the vars just here coz I need the path to the OPM tool.
The task of moving all the ensure tools to the new format is out of scope of this PR.
But I need the OPM path during the calling of the catalog build. This looks inconsistent right now, agreed. but have to keep it like this till I refactor the ensure files.

hack/makefile-vars.mk
malayparida2000: As discussed earlier, We have decided to not remove anyhting from the existing scripts as they will cause a chain reaction as these scripts are used by DS build team and also CI environment. As we are in the middle of the refactor we have to keep these as a new copy for the new make targets as I add them 1 by 1. Once we are clear with all our new make targets, We will remove the common.sh vars after informing the DS build team & also making changes in the CI env.

hack/ensure-opm.sh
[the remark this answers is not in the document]
umangachapagain: We don't have to keep it inconsistent if we separate the refactor (which is completely unrelated to file based catalog) from this migration to file based catalog (which is completely unrelated to the refactor). Can't we use the existing patterns to build file based catalog?

hack/makefile-vars.mk
umangachapagain: > We have decided to not remove anyhting from the existing scripts as they will cause a chain reaction as these scripts are used by DS build team and also CI environment.
>
This is exactly why we need to be more thoughtful about the changes we make. Do we really have no other option than making a breaking change?

hack/ensure-opm.sh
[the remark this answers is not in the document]
malayparida2000: Used the existing pattern to make the file-based catalog. The issue is I will have to rearrange all this to fit in the refactor again. That's why I was trying to make this with the new .mk variables. Anyways will bring in the refactor later.

malayparida2000:
@umangachapagain PTAL

umangachapagain (approved):

PR #1920, opened 2023-01-23
fix OCS Dockerfile path in bundle build script

umangachapagain:
Fixes the Dockerfile path in the bundle build script which was broken after commit 50dcce2d5 (#1917 )


iamniting:
cherry-pick release-4.12

PR #1921, opened 2023-01-23
Report versioning errors in StorageCluster's Status

Nikhil-Ladha:
Update StorageCluster status on versioning error. Append an error status condition to StorageCluster status field when an error arises during `versionCheck`.

And, modified the hack scripts to update the VERSION env during the binary build.

Credits to #1447

**NOTE**
This PR is just a rebase with a couple of changes from the original PR


Nikhil-Ladha:
I don't think we can make the "Version" field empty just now, as the downstream docker images don't have the "Version" env set during the build of the standalone `ocs-operator` images. So, if we remove them then the images would have the "Version" field empty in them.

Nikhil-Ladha:
Marking it as draft until I have tested these changes.

Nikhil-Ladha:
Tested it out on the latest 4.99 build and the version field for `storagecluster` is getting updated now.
```
(ocs_env) [nladha@nladha ocs-ci]$ oc get csv -n openshift-storage
NAME                                               DISPLAY                       VERSION                   REPLACES                          PHASE
mcg-operator.v4.99.1-38.custom-nikhil              NooBaa Operator               4.99.1-38.custom-nikhil   mcg-operator.v4.12.0              Succeeded
ocs-operator.v4.99.1-38.custom-nikhil              OpenShift Container Storage   4.99.1-38.custom-nikhil   ocs-operator.v4.12.0              Succeeded
odf-csi-addons-operator.v4.99.1-38.custom-nikhil   CSI Addons                    4.99.1-38.custom-nikhil   odf-csi-addons-operator.v4.12.0   Succeeded
odf-operator.v4.99.1-38.custom-nikhil              OpenShift Data Foundation     4.99.1-38.custom-nikhil                                     Succeeded
(ocs_env) [nladha@nladha ocs-ci]$ oc get storagecluster -n openshift-storage
ocs-storagecluster   14m   Ready              2023-02-06T07:19:00Z   4.99.1
```

Nikhil-Ladha:
If we intend to backport it to older releases we need to make one small change in the build scripts downstream.

Nikhil-Ladha:
@umangachapagain @iamniting can you please review the PR?
TIA :)

controllers/storagecluster/reconcile.go
```diff
@@ -212,6 +212,12 @@ func (r *StorageClusterReconciler) validateStorageClusterSpec(instance *ocsv1.St
 		r.Log.Error(err, "Failed to validate StorageCluster version.", "StorageCluster", klog.KRef(instance.Namespace, instance.Name))
 		r.recorder.ReportIfNotPresent(instance, corev1.EventTypeWarning, statusutil.EventReasonValidationFailed, err.Error())
 		instance.Status.Phase = statusutil.PhaseError
+		instance.Status.Conditions = append(instance.Status.Conditions, conditionsv1.Condition{
```
iamniting, line 215: Actually, you cannot append like that. It is possible that 'conditionsv1.ConditionDegraded' is already a part of the slice, in which case we will have two identical conditions and many more with the same thing because it will be appended on every reconcile.

Nikhil-Ladha: Got it.
Just noticed that we have a utility function `setStatusConditionIfNotPresent` which would handle that, though it was private for some reason. So, I have exported it and used that function to handle the append of status. I think now it should be fine.

Nikhil-Ladha:
How to resolve the permission denied issue for the newly added `hack/unit-test.sh` file?
Also, I have rebased and run the `make gen-latest-csv` cmd and there are not updates in the CSV but it still complains that "it is not the latest csv", any idea?

iamniting:
> How to resolve the permission denied issue for the newly added `hack/unit-test.sh` file? 

chmod +x hack/unit-test.sh

> Also, I have rebased and run the `make gen-latest-csv` cmd and there are not updates in the CSV but it still complains that "it is not the latest csv", any idea?

will look and let you know

Nikhil-Ladha:
> > How to resolve the permission denied issue for the newly added `hack/unit-test.sh` file?
> 
> chmod +x hack/unit-test.sh
> 
> > Also, I have rebased and run the `make gen-latest-csv` cmd and there are not updates in the CSV but it still complains that "it is not the latest csv", any idea?
> 
> will look and let you know

Ok, I seem to have understood the issue with the failing tests.
I think the use of `VERSION` variable in our hack scripts is contradicting with a `VERSION` env variable in the prow env. The later is taking priority and updating the value to `1.18` which seems like the golang version being used.

I am not exactly sure if this could lead to any issue in future, i.e, if we use different values of `VERSION` and `CSV_VERSION` anywhere 🤔

Nikhil-Ladha:
@iamniting @umangachapagain can you please review the PR

hack/build-functest.sh
```diff
@@ -16,8 +16,9 @@ else
 	echo "GINKO binary found at $GINKGO"
 fi
 
+LDFLAGS="-X github.com/red-hat-storage/ocs-operator/version.Version=${CSV_VERSION}"
 
-"${GINKGO}" build "functests/${suite}/"
+"$GOBIN"/ginkgo build --ldflags "${LDFLAGS}" "functests/${suite}/"
```
umangachapagain, line 21:
```suggestion
"${GINKGO}" build --ldflags "${LDFLAGS}" "functests/${suite}/"
```
This value is already set at Line no. 9. No need to change it.

controllers/storagecluster/reconcile.go
```diff
@@ -212,6 +212,12 @@ func (r *StorageClusterReconciler) validateStorageClusterSpec(instance *ocsv1.St
 		r.Log.Error(err, "Failed to validate StorageCluster version.", "StorageCluster", klog.KRef(instance.Namespace, instance.Name))
 		r.recorder.ReportIfNotPresent(instance, corev1.EventTypeWarning, statusutil.EventReasonValidationFailed, err.Error())
 		instance.Status.Phase = statusutil.PhaseError
+		statusutil.SetStatusConditionIfNotPresent(&instance.Status.Conditions, conditionsv1.Condition{
+			Type:    conditionsv1.ConditionDegraded,
+			Status:  corev1.ConditionTrue,
+			Reason:  statusutil.EventReasonValidationFailed,
+			Message: err.Error(),
+		})
```
umangachapagain, line 220: Not sure if Degraded is the right status here. We should have a condition for version mismatch.

Also, no need to export `SetStatusConditionIfNotPresent`. Just define a new public function in statusutil to set version mismatch status.

umangachapagain (changes requested):

controllers/storagecluster/reconcile.go
Nikhil-Ladha: Ack, will do that.

controllers/storagecluster/reconcile.go
```diff
@@ -212,6 +212,9 @@ func (r *StorageClusterReconciler) validateStorageClusterSpec(instance *ocsv1.St
 		r.Log.Error(err, "Failed to validate StorageCluster version.", "StorageCluster", klog.KRef(instance.Namespace, instance.Name))
 		r.recorder.ReportIfNotPresent(instance, corev1.EventTypeWarning, statusutil.EventReasonValidationFailed, err.Error())
 		instance.Status.Phase = statusutil.PhaseError
+		reason := statusutil.EventReasonValidationFailed
+		message := err.Error()
+		statusutil.SetVersionMismatchCondition(&instance.Status.Conditions, reason, message)
```
iamniting, line 217: Think about a case when there is no mismatch in the version you need to make it false in that case. This means you need to set it to false when the version verification is done.

iamniting (changes requested):

controllers/storagecluster/reconcile.go
Nikhil-Ladha: Oh, right. Thanks for the pointer.
Updated the code to set the status to False if the version check is successful.

controllers/util/status.go
```diff
@@ -280,3 +292,23 @@ func MapNoobaaNegativeConditions(conditions *[]conditionsv1.Condition, found *nb
 	}
 
 }
+
+// SetVersionMismatchCondition sets the ConditionReconcileComplete to True and other Conditions
+// to indicate that the reconciliation process has completed successfully.
+func SetVersionMismatchCondition(conditions *[]conditionsv1.Condition, reason string, message string) {
```
iamniting, line 298: instead of doing if else we could get the status as argument IMO

iamniting (changes requested):

controllers/util/status.go
Nikhil-Ladha: I thought that, but looking at other function definitions I opted for this approach.
Will update the function to include the status as an argument

Nikhil-Ladha: Though, I had one doubt if we want to show a "VerionMismatch" status always or not?
As per the current code, the status is updated only if the value is already set and if we want to keep it that way then this type of conditioning is needed.

Nikhil-Ladha: I have tweaked the code a bit, let me know if its better now.

controllers/util/status.go
```diff
@@ -280,3 +294,15 @@ func MapNoobaaNegativeConditions(conditions *[]conditionsv1.Condition, found *nb
 	}
 
 }
+
+// SetVersionMismatchCondition sets the ConditionVersionMismatch to the status passed as argument.
+// It sets the status as True if the status is not already presrnt or has a different status,
+// whereas it sets the status to False only if the status is already present and it is not False
```
iamniting, line 300: The comment needs to be changed as func is changed

controllers/util/status.go
```diff
@@ -239,6 +241,18 @@ func setStatusConditionIfNotPresent(conditions *[]conditionsv1.Condition, condit
 	conditionsv1.SetStatusCondition(conditions, condition)
 }
 
+// update an existing condition only if its already present and the status
+// is different or the condition is not present and we want to add a new status
+func setStatusConditionIfDifferentOrNotPresent(conditions *[]conditionsv1.Condition, condition conditionsv1.Condition) {
```
iamniting, line 246: IMO the condition should be always present Either the state is good or bad. Someone might be looking for the condition specifically. most of the status automation is done based on the conditions themselves. So I think we can just keep the `SetVersionMismatchCondition` func and it will always set the condition without any if condition.

iamniting (changes requested):

controllers/util/status.go
iamniting: pls ignore this one.

controllers/util/status.go
Nikhil-Ladha: Updated the code accordingly, PTAL.

iamniting (approved):
looks good to me. Holding for @umangachapagain  to review.

/hold

controllers/storagecluster/reconcile.go
```diff
@@ -314,7 +321,7 @@ func (r *StorageClusterReconciler) reconcilePhases(
 	}
 
 	// Add conditions if there are none
-	if instance.Status.Conditions == nil {
+	if len(instance.Status.Conditions) == 1 && instance.Status.Conditions[0].Type == ocsv1.ConditionVersionMismatch {
```
umangachapagain, line 324: Won't condition be nil at the beginning?

iamniting: Not really, because we are adding the version mismatch condition before this code executes while checking the spec.

Nikhil-Ladha: @umangachapagain do you think we should go ahead and move it before the `validateClusterSpec` function call or are we good here?

umangachapagain: This is fine.

umangachapagain (approved):

PR #1831, opened 2022-10-02
Add KMS alert

rexagod:
Added StorageCluster collector on OCS exporter to implement KMS connection lost alerts.

Signed-off-by: Pranshu Srivastava <rexagod@gmail.com>

PR #1419, opened 2021-12-01
Notify if KMS connection to the server is lost

rexagod:
Let users know if the connection to the KMS server was lost by including
`KMSServerConnectionStatus` in the `StorageClusterSpec`. This will add a
`KMSServerConnectionStatus` field consisting of the following information:
  - `KMSServerAddress`: The address of the KMS server.
  - `KMSServerConnectionError`: An error message that's thrown when the
    connection is lost.

PR #1923, opened 2023-01-24
Add KMS alert

Nikhil-Ladha:
Added StorageCluster collector on OCS exporter to implement KMS connection lost alerts.

Credits to original PR: #1831

Co-authored-by: Pranshu Srivastava <rexagod@gmail.com>

### DIFF
--- a/controllers/storagecluster/cephcluster.go
+++ b/controllers/storagecluster/cephcluster.go
@@ -174,13 +174,11 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
 				return reconcile.Result{}, err
 			}
 			if kmsConfigMap != nil {
+				if kmsConfigMap.Data["KMS_PROVIDER"] == "vault" {
+					sc.Status.KMSServerConnection.KMSServerAddress = kmsConfigMap.Data["VAULT_ADDR"]
+				}
 				if err = reachKMSProvider(kmsConfigMap); err != nil {
-					if kmsConfigMap.Data["KMS_PROVIDER"] == "vault" {
-						sc.Status.KMSServerConnection = ocsv1.KMSServerConnectionStatus{
-							KMSServerAddress:         kmsConfigMap.Data["VAULT_ADDR"],
-							KMSServerConnectionError: err.Error(),
-						}
-					}
+					sc.Status.KMSServerConnection.KMSServerConnectionError = err.Error()
 					r.Log.Error(err, "Address provided in KMS ConfigMap is not reachable.", "KMSConfigMap", klog.KRef(kmsConfigMap.Namespace, kmsConfigMap.Name))
 					return reconcile.Result{}, err
 				}
--- a/metrics/deploy/prometheus-ocs-rules.yaml
+++ b/metrics/deploy/prometheus-ocs-rules.yaml
@@ -200,5 +200,16 @@ spec:
       expr: |
         ocs_rbd_client_blocklisted{job="ocs-metrics-exporter"} > 0
       for: 10s
+  - name: ocs-encryption-alert.rules
+    rules:
+    - alert: KMSServerConnectionAlert
+      annotations:
+        description: Storage Cluster KMS Server is in un-connected state for more than 5s. Please check KMS config.
+        message: Storage Cluster KMS Server is in un-connected state. Please check KMS config.
+        severity_level: error
+        storage_type: ceph
+      expr: |
+        ocs_storagecluster_kms_connection_status{job="ocs-metrics-exporter"} == 1
+      for: 5s
       labels:
         severity: critical
--- a/metrics/internal/collectors/storage-cluster.go
+++ b/metrics/internal/collectors/storage-cluster.go
@@ -1,0 +1,88 @@
+package collectors
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog"
+
+	v1 "github.com/red-hat-storage/ocs-operator/api/v1"
+	"github.com/red-hat-storage/ocs-operator/metrics/internal/options"
+)
+
+type StorageClusterCollector struct {
+	KMSServerConnectionStatus *prometheus.Desc
+	Informer                  cache.SharedIndexInformer
+}
+
+var _ prometheus.Collector = &StorageClusterCollector{}
+
+func NewStorageClusterCollector(opts *options.Options) *StorageClusterCollector {
+	cl, err := GetOcsClient(opts)
+	if err != nil {
+		klog.Errorf("Unable to get client: %v", err)
+		return nil
+	}
+	lw := cache.NewListWatchFromClient(cl, "storageclusters", metav1.NamespaceAll, fields.Everything())
+	sharedIndexInformer := cache.NewSharedIndexInformer(lw, &v1.StorageCluster{}, 0, cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc})
+	return &StorageClusterCollector{
+		KMSServerConnectionStatus: prometheus.NewDesc(
+			prometheus.BuildFQName(namespace, "storagecluster", "kms_connection_status"),
+			`KMS Connection Status; 0: Connected, 1: Not Connected, 2: KMS not enabled`,
+			[]string{"name", "namespace"},
+			nil,
+		),
+		Informer: sharedIndexInformer,
+	}
+}
+
+func (c *StorageClusterCollector) Run(stopCh <-chan struct{}) {
+	go c.Informer.Run(stopCh)
+}
+
+func (c *StorageClusterCollector) Describe(ch chan<- *prometheus.Desc) {
+	ds := []*prometheus.Desc{
+		c.KMSServerConnectionStatus,
+	}
+	for _, d := range ds {
+		ch <- d
+	}
+}
+
+func (c *StorageClusterCollector) Collect(ch chan<- prometheus.Metric) {
+	scLister := NewStorageClusterLister(c.Informer.GetIndexer())
+	storageClusters := getAllStorageClusters(scLister)
+	if len(storageClusters) > 0 {
+		c.collectKMSConnectionStatuses(ch, storageClusters)
+	}
+}
+
+func getAllStorageClusters(lister StorageClusterLister) []*v1.StorageCluster {
+	storageClusters, err := lister.List(labels.Everything())
+	if err != nil {
+		klog.Errorf("couldn't list StorageCluster: %v", err)
+		return nil
+	}
+	return storageClusters
+}
+
+func (c *StorageClusterCollector) collectKMSConnectionStatuses(ch chan<- prometheus.Metric, storageClusters []*v1.StorageCluster) {
+	for _, storageCluster := range storageClusters {
+		v := 2
+		if storageCluster.Spec.Encryption.Enable || storageCluster.Spec.Encryption.ClusterWide {
+			v = 0
+			if storageCluster.Status.KMSServerConnection.KMSServerConnectionError != "" {
+				v = 1
+			}
+		}
+		ch <- prometheus.MustNewConstMetric(
+			c.KMSServerConnectionStatus,
+			prometheus.GaugeValue,
+			float64(v),
+			storageCluster.Name,
+			storageCluster.Namespace,
+		)
+	}
+}
--- a/metrics/mixin/alerts/encryption-external.libsonnet
+++ b/metrics/mixin/alerts/encryption-external.libsonnet
@@ -1,0 +1,27 @@
+{
+  prometheusAlerts+:: {
+    groups+: [
+      {
+        name: 'external-ocs-encryption-alert.rules',
+        rules: [
+          {
+            alert: 'KMSServerConnectionAlert',
+            expr: |||
+              ocs_storagecluster_kms_connection_status{%(ocsExporterSelector)s} == 1
+            ||| % $._config,
+            'for': $._config.ocsStorageClusterKMSConnectionAlert,
+            labels: {
+              severity: 'critical',
+            },
+            annotations: {
+              message: 'Storage Cluster KMS Server is in un-connected state. Please check KMS config.',
+              description: 'Storage Cluster KMS Server is in un-connected state for more than %s. Please check KMS config.' % $._config.ocsStorageClusterKMSConnectionAlert,
+              storage_type: $._config.cephStorageType,
+              severity_level: 'error',
+            },
+          },
+        ],
+      },
+    ],
+  },
+}
--- a/config/rbac/exporter-role.yaml
+++ b/config/rbac/exporter-role.yaml
@@ -59,6 +59,7 @@ rules:
   - ocs.openshift.io
   resources:
   - storageconsumers
+  - storageclusters
   verbs:
     - get
     - list
--- a/deploy/csv-templates/ics-operator.csv.yaml.in
+++ b/deploy/csv-templates/ics-operator.csv.yaml.in
@@ -199,6 +199,7 @@ spec:
           - ocs.openshift.io
           resources:
           - storageconsumers
+          - storageclusters
           verbs:
           - get
           - list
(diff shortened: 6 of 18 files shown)


Nikhil-Ladha:
Please review the PR @umangachapagain .
TIA :)

metrics/internal/collectors/storage-cluster.go
```diff
@@ -0,0 +1,85 @@
...
+	KMSServerConnectionStatus *prometheus.Desc
+	Informer                  cache.SharedIndexInformer
+}
+
+var _ prometheus.Collector = &StorageClusterCollector{}
+
+func NewStorageClusterCollector(opts *options.Options) *StorageClusterCollector {
+	cl, err := GetOcsClient(opts)
+	if err != nil {
+		klog.Errorf("Unable to get initialize client: %v", err)
```
umangachapagain, line 25:
```suggestion
		klog.Errorf("Unable to get client: %v", err)
```

umangachapagain (changes requested):

umangachapagain (changes requested):
Since this commit was authored by someone else, let's also have a [co-authored-by](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors#creating-co-authored-commits-on-the-command-line) on the commit message.

Nikhil-Ladha:
> Since this commit was authored by someone else, let's also have a [co-authored-by](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors#creating-co-authored-commits-on-the-command-line) on the commit message.

Sure.
I have updated the commit message (and the PR description 😅 ) to have the co-author line.

umangachapagain:
looks good to me. @aruniiird Can you please take a look once?

Nikhil-Ladha:
@aruniiird pinging for review.

jmolmo:
@Nikhil-Ladha : Sorry if i miss something, but if this is a new alert, why it is not included in the ocs alert rules?

metrics/internal/collectors/storage-cluster.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	return storageClusters
+}
+
+func (c *StorageClusterCollector) collectKMSConnectionStatuses(ch chan<- prometheus.Metric, storageClusters []*v1.StorageCluster) {
+	for _, storageCluster := range storageClusters {
+		v := 0
+		if storageCluster.Status.KMSServerConnection.KMSServerConnectionError != "" {
+			v = 1
+		}
```
aruniiird, line 76: Here `v == 0` is connected and `v == 1` means it is not connected.
Need a small clarity, an empty `StorageClusterStatus` object (that is ` v1.StorageClusterStatus{}`) has an empty `KMSServerConnection.KMSServerConnectionError` field, but that will not mean the storagecluster is connected, right? 
StorageCluster Status variable, `KMSServerConnection`, has another field `KMSServerAddress` , don't we have to check whether that field is populated with a valid address (if KMS server is connected).
A follow-up Q is don't we have to check KMS is enabled in `CephCluster`'s spec?

aruniiird (changes requested):
The code looks good.
Need a small clarification.

metrics/internal/collectors/storage-cluster.go
Nikhil-Ladha: > Here v == 0 is connected and v == 1 means it is not connected.
Need a small clarity, an empty StorageClusterStatus object (that is  v1.StorageClusterStatus{}) has an empty 
KMSServerConnection.KMSServerConnectionError field, but that will not mean the storagecluster is connected, right?

Right, I think having an additional `storageCluster.Status.KMSServerConnection != nil` condition would be better than simply checking for the error.

> StorageCluster Status variable, KMSServerConnection, has another field KMSServerAddress , don't we have to check whether that field is populated with a valid address (if KMS server is connected).

I don't think checking that should be a part of the alert, if the address is incorrect then the error is enough for the user to know that something is wrong, and they should check the system.

> A follow-up Q is don't we have to check KMS is enabled in CephCluster's spec?

If we just add a `nil` check for the `KMSServerConnection` that should be enough to consider that KMS is enabled, else the `KMSServerConnection` would be nil.

Nikhil-Ladha:
> @Nikhil-Ladha : Sorry if i miss something, but if this is a new alert, why it is not included in the ocs alert rules?

@jmolmo added a new alert, PTAL and let me know if anything needs to be updated.

metrics/internal/collectors/storage-cluster.go
```diff
@@ -0,0 +1,85 @@
...
+		klog.Errorf("couldn't list StorageCluster: %v", err)
+		return nil
+	}
+	return storageClusters
+}
+
+func (c *StorageClusterCollector) collectKMSConnectionStatuses(ch chan<- prometheus.Metric, storageClusters []*v1.StorageCluster) {
+	for _, storageCluster := range storageClusters {
+		v := 0
+		if storageCluster.Status.KMSServerConnection != *new(v1.KMSServerConnectionStatus) && storageCluster.Status.KMSServerConnection.KMSServerConnectionError != "" {
```
aruniiird, line 74: As `Status.KMSServerConnection` is not a pointer, we only have to check the field `Status.KMSServerConnection.KMSServerAddress` is populated (and may be an extra check of a syntactically valid address) or not, and there is no error message (in `KMSServerConnectionError` field).

```suggestion
		if storageCluster.Status.KMSServerConnection.KMSServerAddress != "" && storageCluster.Status.KMSServerConnection.KMSServerConnectionError == "" {
```

or even better

```suggestion
		if validKMSAddress(storageCluster.Status.KMSServerConnection.KMSServerAddress) && storageCluster.Status.KMSServerConnection.KMSServerConnectionError == "" {
```
where `validKMSAddress()` will check the given argument is a valid address (or not), but this can be a future enhancement/check.

aruniiird (changes requested):
Can we make it a bit simpler? =)

metrics/internal/collectors/storage-cluster.go
Nikhil-Ladha: Opting for the first approach for now.
Though, I was wondering since we are making it more granular to check if KMS is actually enabled or not, should we add a third status named `Not Connected`  in the collector?
Also, I am not exactly sure if we need to add a `prometheusAlerts` for the `kms_connection_status`? Isn't that the job of the collector to pass such alerts to prometheus if the validation fails?

Sorry, I am a bit new to the code and the docs isn't much helpful in this case either so a bit confused.

metrics/deploy/prometheus-ocs-rules.yaml
```diff
@@ -190,3 +190,16 @@ spec:
       for: 15s
       labels:
         severity: critical
+  - name: ocs-encryption-alert.rules
+    rules:
+    - alert: KMSServerConnection
+      annotations:
+        description: Storage Cluster KMS Server is in un-connected state for more than 5s. Please check KMS config.
+        message: Storage Cluster KMS Server is in un-connected state. Please check KMS conifg.
```
jmolmo, line 198: typo: s/conifg/config

metrics/internal/collectors/storage-cluster.go
Nikhil-Ladha: > Anything other than Connected status, that is: other than the metric value ZERO, should be not connected.
Another improvement here is we can have a value of -1 (for the metric) which can refer to KMS not enabled status (but again may not be the current scope).

Currently, I have added 0: Connected, 1: NotConnected, 2: Failure.
Are you suggesting we should probably, update it to `-1: NotConnected, 0: Connected, 1: Failure` ?

jmolmo (changes requested):
I do not know the context of this modification (bug, issue, epic.) So i have the following questions:

- it is needed also to add the alert to "prometheus-ocs-rules-external.yaml" ?
- Do we need to include this alert in the "mixins" alert definition files (services.libsonnet, services-external.libsonnet)?

Nikhil-Ladha:
> I do not know the context of this modification (bug, issue, epic.) So i have the following questions:
> 
> * it is needed also to add the alert to "prometheus-ocs-rules-external.yaml" ?

Seems like yes, it has to be added because we support KMS integration in both Internal/External cluster.
BZ ref: https://bugzilla.redhat.com/show_bug.cgi?id=1944687

> * Do we need to include this alert in the "mixins" alert definition files (services.libsonnet, services-external.libsonnet)?

I have already added this alert under mixins as a new definition having name `encryption.libsonnet`.

jmolmo (approved):

metrics/internal/collectors/storage-cluster.go
```diff
@@ -0,0 +1,88 @@
...
+
+func (c *StorageClusterCollector) collectKMSConnectionStatuses(ch chan<- prometheus.Metric, storageClusters []*v1.StorageCluster) {
+	for _, storageCluster := range storageClusters {
+		v := 1
+		if storageCluster.Status.KMSServerConnection.KMSServerAddress != "" {
+			v = 0
+			if storageCluster.Status.KMSServerConnection.KMSServerConnectionError != "" {
+				v = 2
+			}
+		}
```
aruniiird, line 79: According to the above description (line#33), 0: Connected , 1: Not Connected and 2: Failure
But what is the distinction between `Not Connected` and a `Failure` (failure means it is not connected, right?)
IMHO, we can have THREE states, `Connected`, `Failure` or `Not Connected` and `KMS not enabled`.

Consider this scenario where,
1. `KMSServerAddress` is empty because of a connection failure, that is we have an error message in `KMSServerConnectionError` field.

But as per the above conditions, we will never reach `v = 2` (Failure scenario) and always be in not connected state (which is technically correct) =)

metrics/mixin/alerts/encryption-external.libsonnet
```diff
@@ -0,0 +1,27 @@
+{
```
aruniiird, line 1: Do we require these changes in external encryption file? Do we enable KMS and external mode together?
@umangachapagain ^^^

aruniiird (changes requested):
Some requests...

metrics/mixin/alerts/encryption-external.libsonnet
aruniiird: One of the above comments answer this (we are enabling KMS in internal and external cluster).

metrics/internal/collectors/storage-cluster.go
aruniiird: Nope, the numbers are fine. Added comment in the below review (as I have seen that changes a bit late). Please check

metrics/deploy/prometheus-ocs-rules.yaml
```diff
@@ -190,3 +190,16 @@ spec:
...
+        severity_level: error
+        storage_type: ceph
+      expr: |
+        ocs_storagecluster_kms_connection_status{job="ocs-metrics-exporter"} == 1
+      for: 5s
```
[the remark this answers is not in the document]
Nikhil-Ladha: I gave 5 seconds because that's the timeout we have in the code to check if KMS is reachable or not, [here](https://github.com/red-hat-storage/ocs-operator/blob/main/controllers/storagecluster/cephcluster.go#L177)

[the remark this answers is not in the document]
Nikhil-Ladha: But again, that's for the initial connection so for generating alerts maybe we can keep it as 10s.

metrics/internal/collectors/storage-cluster.go
Nikhil-Ladha: This use case is indeed better, but it seems like we have some limitations as to how code is working as of now.
In the code [here](https://github.com/red-hat-storage/ocs-operator/blob/main/controllers/storagecluster/cephcluster.go#L177), we are updating the `KMSServerAddress`, `KMSServerConnectionError` only when the provider is `vault` and the configmap data field has the `VAULT_ADDR`, else we don't update the `KMSServer*` values and just report an error. This was added in the PR #1419, @umangachapagain do you know as to why we have limited the update only for this particular case? Shouldn't we keep it generic? Like, maybe not update the address for other providers but at least update the `KMSServerConnectionError` value? Else if the KMS connection fails for a provider other than `vault` we have no way to identify and generate an alert.

metrics/mixin/alerts/encryption-external.libsonnet
Nikhil-Ladha: I mean not together but I don't see it mentioned anywhere that we don't support external mode. So, added an alert for external mode as well.

metrics/internal/collectors/storage-cluster.go
Nikhil-Ladha: So, I have updated the code to add address and error status in case of an error irrespective of the KMS provider. For any provider other than `vault` the address field will be empty but the error field will be updated with the respective error.

metrics/internal/collectors/storage-cluster.go
```diff
@@ -0,0 +1,88 @@
...
+
+func (c *StorageClusterCollector) collectKMSConnectionStatuses(ch chan<- prometheus.Metric, storageClusters []*v1.StorageCluster) {
+	for _, storageCluster := range storageClusters {
+		v := 3
+		if storageCluster.Spec.Encryption.Enable || storageCluster.Spec.Encryption.ClusterWide {
+			v = 1
+			if storageCluster.Status.KMSServerConnection.KMSServerConnectionError != "" {
+				v = 2
+			}
+		}
```
aruniiird, line 79: Changes look good @Nikhil-Ladha 
Just to keep the code consistent with other `collectors` (see file: metrics/internal/collectors/ceph-object-store.go or file: metrics/internal/collectors/rbd-mirror.go, where OK state is ZERO) can we make `Connected` state as ZERO and rest of the states according to your choice (something like 0: connected, 1: not-connected/failure, 2: kms not enabled).

PS: In my previous comment, I was just pointing out three different scenarios, sorry if that confused you... =)

aruniiird (changes requested):
Looks good, one final touch...

metrics/internal/collectors/storage-cluster.go
Nikhil-Ladha: Cool, updated the code.

Nikhil-Ladha:
Locally, there are no csv changes even after running the `gen-latest-csv` command, but the test still seems to fail, any idea? 🤔

Nikhil-Ladha:
@aruniiird @umangachapagain can you please add your reviews.
TIA :)

iamniting:
> Locally, there are no csv changes even after running the `gen-latest-csv` command, but the test still seems to fail, any idea? thinking

I reran it and it is passed.

controllers/storagecluster/cephcluster.go
```diff
@@ -175,11 +175,13 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
 			}
 			if kmsConfigMap != nil {
 				if err = reachKMSProvider(kmsConfigMap); err != nil {
```
aruniiird, line 177: As you are any way making changes to this file a few suggestions,
Why add the `Status.KMSServerConnection` details only when there is an error?  A successful connection should populate the address field. (nobody to blame but me =))
Can we populate `Status.KMSServerConnection.KMSServerAddress` before the error and add any error details to `KMSServerConnection.KMSServerConnectionError` if there is an error.

Otherwise `KMSServerAddress` field will only be populated when there is an error...

```suggestion
				if kmsConfigMap.Data["KMS_PROVIDER"] == "vault" {
					sc.Status.KMSServerConnection.KMSServerAddress = kmsConfigMap.Data["VAULT_ADDR"]
				}
				if err = reachKMSProvider(kmsConfigMap); err != nil {
					sc.Status.KMSServerConnection.KMSServerConnectionError = err.Error()
					r.Log.Error(err, "Address provided in KMS ConfigMap is not reachable.", "KMSConfigMap", klog.KRef(kmsConfigMap.Namespace, kmsConfigMap.Name))
```

Nikhil-Ladha: Sure, updated the code as per the suggestion.

aruniiird (approved):
LGTM from my end... @umangachapagain  please take a look

umangachapagain (changes requested):
Please break it into separate commits: metrics changes, generated changes and alert changes. And, add some description about changes in commit message.

Rest looks good to me.

Nikhil-Ladha:
> Please break it into separate commits: metrics changes, generated changes and alert changes. And, add some description about changes in commit message.
> 
> Rest looks good to me.

Done with breaking up the commits (along with co-author :) ), PTAL

umangachapagain (approved):

PR #1927, opened 2023-01-31
Update fusion bundle

umangachapagain:
Changes the name of fusion package to ocs-operator.
This is required for us to support seamless upgrades. OLM does
not support upgrades between 2 different packages.

It also updates base CSV metadata with proper branding contents.

It renames FCS to ICS to be consistent with build references.

Enhances #1917


malayparida2000:
@umangachapagain Inside the deploy folder we should also rename the fcs-operator folder to ics-operator

malayparida2000:
The changes look good
/lgtm

Putting a hold in case you might have something else to add, if not feel free to remove it
/hold

PR #1930, opened 2023-02-02
Bug 2166220: [release-4.12] Backport fusion container storage bundle

umangachapagain:
Backports of #1917 #1927


malayparida2000:
@umangachapagain  I see inside the deploy/csv-templates/crds we have ics as well as fcs. The fcs folder shouldnt be there.
Also some of the commit messages mention fusion container storage, But we later decided it would be ibm container storage. So should we change the commit messages?

umangachapagain:
> @umangachapagain I see inside the deploy/csv-templates/crds we have ics as well as fcs. The fcs folder shouldnt be there. 
>
Yeah I missed deleting that during backport. Will remove it.

> Also some of the commit messages mention fusion container storage, But we later decided it would be ibm container storage. So should we change the commit messages?
>
I am keeping it as it is to mirror the history in main branch. I believe that is the usual practice.

umangachapagain:
/hold
BZ is not fully acked yet.

iamniting (approved):
/lgtm

PR #1932, opened 2023-02-06
Replace various clients in e2e test with a single client with required schemes

malayparida2000:
The Presence of various clients makes the e2e tests complicated. So adding all the required schemes 
to one controller runtime client helps in simplifying the test operations.


iamniting (approved):
/lgtm

PR #1934, opened 2023-02-09
e2e: Use new file based catalog, Create Noobaa sub, Fix broken Must gather path

malayparida2000:
Make necessary changes for the building the functests,
functest ocs operator install to work & gathering of must-gather if e2e test fails.


pkg/deploy-manager/subscription.go
```diff
@@ -126,6 +126,8 @@ func (t *DeployManager) generateClusterObjects(ocsCatalogImage string, subscript
 	ocsOG.SetGroupVersionKind(schema.GroupVersionKind{Group: operatorv1.SchemeGroupVersion.Group, Kind: "OperatorGroup", Version: operatorv1.SchemeGroupVersion.Version})
 
 	co.operatorGroups = append(co.operatorGroups, ocsOG)
+
+	// Catalog Sources
```
iamniting, line 130:
```suggestion
	// Catalog Source
```

iamniting (changes requested):

pkg/deploy-manager/subscription.go
malayparida2000: Although just one Cat Src is there it's the name of the section so I think it should be plural only. Also if you see the other resources are in plural only, Like Namespaces, Subscriptions, Operator groups. The notion is to put how many of them might be in this place and append them all to the list.

iamniting (approved):

PR #1935, opened 2023-02-10
do not use hardcoded ns for ocs-prometheus-rules

iamniting:

### DIFF
--- a/controllers/storagecluster/prometheus.go
+++ b/controllers/storagecluster/prometheus.go
@@ -10,6 +10,7 @@ import (
 	"github.com/imdario/mergo"
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v1"
+	"github.com/red-hat-storage/ocs-operator/controllers/util"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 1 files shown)


controllers/storagecluster/prometheus.go
```diff
@@ -42,20 +41,20 @@ func (r *StorageClusterReconciler) enablePrometheusRules(instance *ocsv1.Storage
...
+func getPrometheusRules(instance *ocsv1.StorageCluster) (*monitoringv1.PrometheusRule, error) {
 	rule := &monitoringv1.PrometheusRule{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       monitoringv1.PrometheusRuleKind,
 			APIVersion: monitoringv1.SchemeGroupVersion.String(),
 		},
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      ruleName,
-			Namespace: ruleNamespace,
+			Namespace: instance.Namespace,
```
umangachapagain, line 52: We should use the namespace of operator here.

umangachapagain (changes requested):

controllers/storagecluster/prometheus.go
iamniting: using operator ns.

controllers/storagecluster/prometheus.go
```diff
@@ -50,7 +50,7 @@ func getPrometheusRules(isExternal bool) (*monitoringv1.PrometheusRule, error) {
 		},
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      ruleName,
-			Namespace: ruleNamespace,
+			Namespace: os.Getenv(util.OperatorNamespaceEnvVar),
```
umangachapagain, line 53: Use `util.GetOperatorNamespace()` for safety. This could fail is env var is not set.

umangachapagain (changes requested):

controllers/storagecluster/prometheus.go
iamniting: We are already doing that in the main.go, So it should be safe. https://github.com/red-hat-storage/ocs-operator/blob/main/main.go#L120

umangachapagain (approved):

PR #1936, opened 2023-02-10
Pass which namespace the e2e functest uses to deploy the ocs operator in

malayparida2000:
With recent changes, we are now able to install & run the OCS-operator on any given namespace and it's no longer limited to just openshift-storage. So exposing a Var INSTALL_NAMESPACE which can be set while calling the tests, to make the installation in any desired namespace.


tools/cluster-deploy/cluster-deploy.go
```diff
@@ -22,10 +23,14 @@ func main() {
 		log.Fatal("--ocs-catalog-image is required")
 	} else if *ocsSubscriptionChannel == "" {
 		log.Fatal("--ocs-subscription-channel is required")
+	} else if *installNamespace == "" {
```
iamniting, line 26: I think this is not required.

tools/cluster-deploy/cluster-deploy.go
```diff
@@ -11,6 +11,7 @@ import (
 var (
 	ocsCatalogSourceImage  = flag.String("ocs-catalog-image", "", "The OCS CatalogSource container image to use in the deployment")
 	ocsSubscriptionChannel = flag.String("ocs-subscription-channel", "", "The subscription channel to receive upgrades from in the deployment")
+	installNamespace       = flag.String("install-namespace", "openshift-storage", "The namespace the operator runs in")
```
iamniting, line 14: Also lets pass the namespace in the shell script as we are doing in the `hack/functest.sh`

iamniting (changes requested):

tools/cluster-deploy/cluster-deploy.go
malayparida2000: It's required, as someone may pass the flag but it will not be set or empty. In that case & an empty namespace will generate runtime segfault errors.

tools/cluster-deploy/cluster-deploy.go
malayparida2000: Is this necessary as on our ocs side cluster-deploy tool we use for 2 things, make gen-latest-deploy-yaml & make cluster-deploy. And on ocs side for us openshift-storage is the only ns we are concerned with right?

iamniting: What is the purpose of adding this argument if we are not passing it? We can pass openshift-storage as the default value. As of now, if someone wanted to change the ns, we don't have a way to do so unless this cluster-deploy is changed.

iamniting (approved):

PR #1939, opened 2023-02-17
cephcluster: add support for read affinity for RBD volumes

Rakshith-R:
This commit enables read affinity for RBD volumes by setting the "CSI_ENABLE_READ_AFFINITY" environment variable to "true". If it is an external cluster, and the "CSI_ENABLE_READ_AFFINITY" variable is not present in rook-ceph-operator-config ConfigMap, only then it is set to false. This provides flexibility to use the configmap enable and modify this feature for external cluster.

refer: https://issues.redhat.com/browse/RHSTOR-4371

### DIFF
--- a/controllers/ocsinitialization/ocsoperator_config.go
+++ b/controllers/ocsinitialization/ocsoperator_config.go
@@ -6,17 +6,13 @@ import (
 
 	configv1 "github.com/openshift/api/config/v1"
 	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v1"
+	"github.com/red-hat-storage/ocs-operator/controllers/util"
+
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
-)
-
-const (
-	// This configmap is purely for the OCS operator to use
-	ocsOperatorConfigName = "ocs-operator-config"
 )
 
 // ensureOCSOperatorConfig ensures that the OCS operator config is present in the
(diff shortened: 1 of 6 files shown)


Rakshith-R:
/wip

I'll modify this pr to use ocsoperatorconfigmap instead.

controllers/storagecluster/cephcluster.go
```diff
@@ -103,6 +109,13 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
 		return reconcile.Result{}, fmt.Errorf("'StorageDeviceSets' should not be initialized in an external CephCluster")
 	}
 
+	err = r.setReadAffinity(sc, !sc.Spec.ExternalStorage.Enable)
```
iamniting, line 112: why do we need to pass a boolean again when we are already passing the storagecluster? we can read the value from storagecluster directly in the func.

iamniting (changes requested):

controllers/storagecluster/cephcluster.go
```diff
@@ -1130,6 +1141,49 @@ func createOrUpdatePrometheusRule(r *StorageClusterReconciler, sc *ocsv1.Storage
...
+	_, err := ctrl.CreateOrUpdate(r.ctx, r.Client, cm, func() error {
+		if cm.Data[readAffinityVarKey] != readAffinityVarVal {
+			cm.Data[readAffinityVarKey] = readAffinityVarVal
+			needsRestart = true
+		}
+		return nil
+	})
+	if err != nil {
+		r.Log.Error(err, fmt.Sprintf("failed to update %q configmap", util.OcsOperatorConfigName),
+			"CephCluster", klog.KRef(sc.Namespace, sc.Name))
```
iamniting, line 1171:
```suggestion
			"StorageCluster", klog.KRef(sc.Namespace, sc.Name))
```

Please change this at other places as well.

iamniting (changes requested):
I guess we are missing the CSV changes. We still need to export this to rook envs.

malayparida2000:
Can we have some the default value false for the key put here https://github.com/red-hat-storage/ocs-operator/blob/main/controllers/ocsinitialization/ocsoperator_config.go#L31. So that rook operator doesn't perceive the variable to be blank  & as it doesn't restart once the storagecluster is created(with the current form the value is not set at the creation of the cm, So once storagecluster reconcile hits your code it will set the value & restart the rook operator pod).

controllers/ocsinitialization/ocsoperator_config.go
```diff
@@ -6,25 +6,21 @@ import (
...
 // ensureOCSOperatorConfig ensures that the OCS operator config is present in the
 // cluster & has the correct values
 func (r *OCSInitializationReconciler) ensureOCSOperatorConfig(initialData *ocsv1.OCSInitialization) error {
 	ocsOperatorConfig := &corev1.ConfigMap{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      ocsOperatorConfigName,
+			Name:      util.OcsOperatorConfigName,
 			Namespace: initialData.Namespace,
 		},
 		Data: map[string]string{
```
malayparida2000, line 26: Let's put a default value here

Rakshith-R:
hmm, weird requesting review from one user seems to dismiss the request from another user. 
ignore the spam.

CI seems all green now

malayparida2000 (approved):
/lgtm

iamniting (approved):

PR #1944, opened 2023-02-18
Update to golang 1.19

malayparida2000:
Update
1. go.mod
2. github actions
3. build Dockerfiles
4. build root image

### DIFF
--- a/.ci-operator.yaml
+++ b/.ci-operator.yaml
@@ -1,4 +1,4 @@
 build_root_image:
-  namespace: openshift
-  name: release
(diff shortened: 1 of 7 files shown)


iamniting (changes requested):
I think we dont need these many commits. These can be converted to 2 commits. one for go 1.19 changes and one for golangci.

malayparida2000:
> I think we dont need these many commits. These can be converted to 2 commits. one for go 1.19 changes and one for golangci.

@iamniting Updated

iamniting (approved):

PR #1947, opened 2023-02-21
change the conditions when cephcluster is in connected state

iamniting:
we are changing the conditions when cephcluster is in connecting state or error state and reflecting the ceph cluster state in the storagecluster conditions accordingly. But once ceph cluster is in connected state we are not changing the conditions at all. Which makes storagecluster conditions out of sync with the cephcluster when it is in connected state. updating the conditions when cephcluster is connected will reflect the actual state.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2172189

### DIFF
--- a/controllers/storagecluster/reconcile.go
+++ b/controllers/storagecluster/reconcile.go
@@ -462,6 +462,10 @@ func (r *StorageClusterReconciler) reconcilePhases(
 		message := ocsv1.ReconcileCompletedMessage
 		statusutil.SetCompleteCondition(&instance.Status.Conditions, reason, message)
 
+		if instance.Spec.ExternalStorage.Enable {
+			statusutil.RemoveExternalCephClusterNegativeConditions(&instance.Status.Conditions)
+		}
+
 		// If no operator whose conditions we are watching reports an error, then it is safe
 		// to set upgradeable to true.
 		if instance.Status.Phase != statusutil.PhaseClusterExpanding {
--- a/controllers/util/status.go
+++ b/controllers/util/status.go
@@ -204,6 +204,16 @@ func MapExternalCephClusterNegativeConditions(conditions *[]conditionsv1.Conditi
 	}
 }
 
+// RemoveExternalCephClusterNegativeConditions removes the External cluster negative conditions
+func RemoveExternalCephClusterNegativeConditions(conditions *[]conditionsv1.Condition) {
+	if conditionsv1.FindStatusCondition(*conditions, ocsv1.ConditionExternalClusterConnecting) != nil {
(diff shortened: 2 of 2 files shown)


Madhu-1 (approved):

Madhu-1:
added hold so that others can take a look at it.

controllers/storagecluster/cephcluster.go
```diff
@@ -266,7 +266,7 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
 	} else {
 		// Interpret CephCluster status and set any negative conditions
 		if sc.Spec.ExternalStorage.Enable {
-			statusutil.MapExternalCephClusterNegativeConditions(&r.conditions, found)
+			statusutil.MapExternalCephClusterConditions(&r.conditions, found)
```
umangachapagain, line 269: If you look at other places, we only set negative conditions here. So existing implementation is fine.

If you want to set positive condition, it is done in reconcile.go at the end of reconcile if there are no negative conditions.

Also IIRC, we made a choice to not have "connected" as a condition in StorageCluster. Instead we want to keep using the existing conditions set by `SetCompleteCondition` utility function.

I see the issue that you are trying to fix, but what is the root cause? We've had this code for a while and I haven't seen any similar issue.

umangachapagain (changes requested):

controllers/storagecluster/cephcluster.go
iamniting: Only a few conditions were set by `SetCompleteCondition`. It does not remove or update the `ExternalClusterConnecting` and `ExternalClusterConnected` conditions. These conditions are only set once. And only in a negative way. As soon as the Ceph cluster is connected, someone needs to make them positive.

umangachapagain: When cluster is connected, we set sc.status.phase to Ready. So after we call `SetCompleteCondition` we can check phase and remove the negative conditions. We already do that for ClusterExpanding phase.

iamniting: Do you really want to remove the conditions, I don't think it is a good practice to remove the conditions. Instead, we should change the status of the conditions.

umangachapagain: I don't want the conditions to change based on deployment mode. It should remain the same. I don't think there's any issue with removing negative conditions.

iamniting: I made the change as you suggested but I still don't think we should remove conditions as we can make them positive.

umangachapagain: We can always change it if it doesn't work out. Current change looks good to me.

umangachapagain (approved):

PR #1952, opened 2023-03-01
remove OCSInitialization controller ref from ocs-operator-config cm

iamniting:
We are upgrading the ocs-operator configmap by adding a storage cluster as an controller, which breaks the upgrade because ocsinitialization is already a controller to it. Before we can add the storagecluster as an controller, we must first remove ocsinitialization as a controller from the owner references.


agarwal-mudit (approved):
/lgtm

PR #1968, opened 2023-03-27
Allow ignoring RBDMirror CR

umangachapagain:
Operator already allows CephRBDMirror metadata to be updated and does not override it on each reconcile.
If this is not enough for some reason, we can now ask the operator to ignore all changes to CephRBDMirror CR.


controllers/storagecluster/cephrbdmirrors.go
```diff
@@ -82,6 +82,11 @@ func (r *StorageClusterReconciler) newCephRbdMirrorInstances(initData *ocsv1.Sto
 
 // ensureCreated ensures that cephRbdMirror resources exist in the desired state.
 func (obj *ocsCephRbdMirrors) ensureCreated(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) (reconcile.Result, error) {
+	reconcileStrategy := ReconcileStrategy(instance.Spec.ManagedResources.CephRBDMirror.ReconcileStrategy)
+	if reconcileStrategy == ReconcileStrategyIgnore {
+		return reconcile.Result{}, nil
```
vbnrh, line 87: This looks like it only stops ocs from reconciling it ? Don't we need to label it as 'do_not_reconcile' for rook to ignore it as well ?

api/v1/storagecluster_types.go
```diff
@@ -217,6 +218,11 @@ type ManageCephToolbox struct {
 	ReconcileStrategy string `json:"reconcileStrategy,omitempty"`
 }
 
+// ManageCephRBDMirror defines how to reconcile Ceph toolbox
```
iamniting, line 221:
```suggestion
// ManageCephRBDMirror defines how to reconcile CephRBDMirror
```

iamniting (changes requested):

iamniting (approved):

api/v1/storagecluster_types.go
```diff
@@ -217,6 +218,11 @@ type ManageCephToolbox struct {
 	ReconcileStrategy string `json:"reconcileStrategy,omitempty"`
 }
 
+// ManageCephRBDMirror defines how to reconcile Ceph RBDMirror
+type ManageCephRBDMirror struct {
+	ReconcileStrategy string `json:"reconcileStrategy,omitempty"`
```
vbnrh, line 223: I think it would be safer if you add few enum values for type safety in the CR

vbnrh (changes requested):

api/v1/storagecluster_types.go
vbnrh: https://book.kubebuilder.io/reference/markers/crd-validation.html

PR #1969, opened 2023-03-27
remove fusion csv generation

umangachapagain:
We do not need to generate this CSV anymore. Removing all related changes.


iamniting (changes requested):
Can you pls also remove the files inside the deploy directory related to the ics?

iamniting (approved):

PR #1972, opened 2023-03-28
do not watch noobaa CRs if SKIP_NOOBAA_CRD_WATCH is set to true

iamniting:
If the CRD is not present, it can prevent the ocs-operator from
starting which can cause problems for MS. Since MS is not going
to install the noobaa CSV anymore, it is better to only watch the
noobaa CRs if the SKIP_NOOBAA_CRD_WATCH is set to false.

MS will set SKIP_NOOBAA_CRD_WATCH env in the sub which will be
populated to the ocs-operator via OLM.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2182041

### DIFF
--- a/controllers/storagecluster/storagecluster_controller.go
+++ b/controllers/storagecluster/storagecluster_controller.go
@@ -154,14 +154,18 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		},
 	)
 
-	return ctrl.NewControllerManagedBy(mgr).
+	builder := ctrl.NewControllerManagedBy(mgr).
 		For(&ocsv1.StorageCluster{}, builder.WithPredicates(scPredicate)).
 		Owns(&cephv1.CephCluster{}).
-		Owns(&nbv1.NooBaa{}).
 		Owns(&corev1.PersistentVolumeClaim{}, builder.WithPredicates(pvcPredicate)).
 		Owns(&appsv1.Deployment{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
 		Owns(&corev1.Service{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
 		Owns(&corev1.ConfigMap{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
-		Watches(&source.Kind{Type: &ocsv1.OCSInitialization{}}, enqueueStorageClusterRequest).
-		Complete(r)
+		Watches(&source.Kind{Type: &ocsv1.OCSInitialization{}}, enqueueStorageClusterRequest)
+
+	if os.Getenv("SKIP_NOOBAA_CRD_WATCH") != "true" {
+		builder.Owns(&nbv1.NooBaa{})
+	}
+
+	return builder.Complete(r)
 }


controllers/storagecluster/reconcile.go
```diff
@@ -281,6 +288,44 @@ func (r *StorageClusterReconciler) validateStorageClusterSpec(instance *ocsv1.St
...
 
+func (r *StorageClusterReconciler) doesCrdExist(crdName string) (bool, error) {
+	crd := extensionsv1.CustomResourceDefinition{}
+	err := r.Client.Get(r.ctx, types.NamespacedName{Name: crdName}, &crd)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return false, nil
+		}
+		return false, err
+	}
```
bindrad, line 299:
```suggestion
	if err := r.Client.Get(r.ctx, types.NamespacedName{Name: crdName}, &crd); err != nil {
		if errors.IsNotFound(err) {
			return false, nil
		}
		return false, err
	}
```

controllers/storagecluster/reconcile.go
```diff
@@ -281,6 +288,44 @@ func (r *StorageClusterReconciler) validateStorageClusterSpec(instance *ocsv1.St
...
+		// crd was not available while starting lets check it
+		exists, err := r.doesCrdExist(crdName)
+		if err != nil {
+			return err
+		}
+		// crd is created restart the operator and add it to watch list
+		if exists {
+			r.Log.Info("New CRD is created, Exiting to getting it added to the watchlist.")
+			os.Exit(1)
+		}
```
bindrad, line 323:
```suggestion
		if exists, err := r.doesCrdExist(crdName); err != nil {
			return err
		} else if exists {
			// crd is created restart the operator and add it to watch list
			r.Log.Info("New CRD is created, Exiting to getting it added to the watchlist.")
			os.Exit(1)
		}
```

bindrad (changes requested):

bindrad:
@iamniting Did you check if any other CRDs are not created as part of OCS Operator bundle but OCS Operator watches them?

iamniting:
> @iamniting Did you check if any other CRDs are not created as part of OCS Operator bundle but OCS Operator watches them?

No other CR is coming from noobaa which we are watching. So we are good.

bindrad (approved):

umangachapagain (changes requested):
I do not like this approach of restarting operator when CRDs are added. If operator requires a CRD, it should be created during deployment. If you want to add/remove watches dynamically, controller has a feature for that. Gating normal operator functions for a very specific use case isn't acceptable.

/do-not-merge

nb-ohad:
@umangachapagain 
> ...controller has a feature for that

FYI, you can call `watch` on a controller even after the start is called, but these new watches will not populate the cache. We have a similar (and even bigger) problem with the new MS agent where offerings are selected dynamically through the lifetime of the operator and could not find any other valid solution to the problem. If there is any we would love to learn about one and I believe Nithin would love that as well. 

The reason that a restart is needed here is that ODF operator has no guarantees to install the noobaa operator before ocs-operator. Is that is solved somehow then there will be no need for a restart.

umangachapagain:
> @umangachapagain
> 
> > ...controller has a feature for that
> 
> FYI, you can call `watch` on a controller even after the start is called, but these new watches will not populate the cache. We have a similar (and even bigger) problem with the new MS agent where offerings are selected dynamically through the lifetime of the operator and could not find any other valid solution to the problem. If there is any we would love to learn about one and I believe Nithin would love that as well.
> 
>
Looks like a bug or potential feature in upstream controller-manager. Are they recommending restarting operators as an acceptable alternative? What other options did we try? I'd be happy to take a look at your test results and provide my comments. There's a 99.99% chance that I am wrong/unreasonable, I'd like to take this opportunity to learn as well.

> The reason that a restart is needed here is that ODF operator has no guarantees to install the noobaa operator before ocs-operator. Is that is solved somehow then there will be no need for a restart.
>
Looks like a request for OLM team. I am sure they have a way to guarantee that dependencies are installed before the operator that needs those. Or atleast a guarantee that operator can start even if the dependencies are not fulfilled.

nb-ohad:
> Looks like a bug or potential feature in upstream controller-manager. 
I agree this should be a potential feature. The way operator SDK is structured right now, a reconciler doesn't even have a way to access the controller that runs it. That design decision was made on purpose (according to docs) to prevent any reconciliation code to cancel the context and as a result "kill" the controller. 

We tried the following
- Capture the controller and push it into the reconciler then call watch - watches did not populate the cache
- Capture the controller and push it into the reconciler then cancel the controller context (not request context) - controller killed but undefined behavior for existing watches.
- Capture the context of the manager and cancel it with the intent to create another manager - many issues.
- kill the pod - the only thing that reliably worked.

We might be able to resolve some of the issues around the non-kill pod options but as of time constraints and the need for a workaround, we went with this approach for now.

> Looks like a request for OLM team.
I agree, but, according to my understanding, the OLM team is not indenting to add new features until OLM 2.0 gets released. The issue here is that ODF operator depends on both noobaa operator and ocs operator, and the current OLM feature set does not allow defining an order relationship between these 2 dependencies. Add to that the fact that the odf-operator is manually managing dependent subscriptions which also complicates things even further as this mechanism does not support an order relationship as well.

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -91,6 +91,10 @@ type StorageClusterReconciler struct {
 	IsNoobaaStandalone bool
 }
 
+func skipNoobaaCrdWatch() bool {
+	return os.Getenv("SKIP_NOOBAA_CRD_WATCH") == "true"
```
sp98, line 95: where is this `SKIP_NOOBAA_CRD_WATCH`  coming from?

sp98: Can it be more generic instead of being specific to Nooba?

umangachapagain: It'd need to be a list of CRDs to allow that. Any generic name would just cause confusion.
We could have another variable for OB/OBC?

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -91,6 +91,10 @@ type StorageClusterReconciler struct {
 	IsNoobaaStandalone bool
 }
 
+func skipNoobaaCrdWatch() bool {
```
umangachapagain, line 94: No need to make this a function, just use it inline.

umangachapagain (changes requested):

controllers/storagecluster/storagecluster_controller.go
iamniting: +1 to @umangachapagain 

This variable will be coming from the sub. Ms will add this variable in the sub and OLM will inject the variable into the operators.

controllers/storagecluster/storagecluster_controller.go
iamniting: removed it

bindrad (approved):

controllers/storagecluster/storagecluster_controller.go
sp98: ok. @umangachapagain Will that be a different PR for OB/OBC?

iamniting: It depends on what we decide on. Mainly it depends on rook how it wants to proceed with the fix.

umangachapagain (approved):

PR #1973, opened 2023-03-29
Construct the topology map instead of just adding to the map

malayparida2000:
We were just adding to the existing topology map any new things that were there but never removing the non-existent node labels. This resulted in a situation where the topology map will still show a node's labels even after the node was deleted. So fixing it by constructing the topology map from scratch in each reconcile so that it will always be in sync with the current state of the cluster.


malayparida2000:
solves : https://bugzilla.redhat.com/show_bug.cgi?id=2102304

iamniting (approved):

malayparida2000:
This is now tested, 
/hold cancel

iamniting (approved):

PR #1987, opened 2023-04-03
[release-4.11] Fix ginko errors in the openshift-ci

iamniting:
openshift-ci is failing for quite a long time on the older branches due to the ginko v2.

We were installing the Ginko binary without the version, which was installing Ginko V2 and making it incompatible with our older code and Golang. Fix it by sticking it to the v1 version.


malayparida2000 (approved):
/lgtm

PR #1993, opened 2023-04-04
Don't require msgr2 port for provider/consumer clusters

malayparida2000:
For consumer/provider MS mode we cannot set global configuration at the csi level as a single consumer will be connected to the multiple provider clusters.

resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2184068#c10

### DIFF
--- a/controllers/storagecluster/cephcluster.go
+++ b/controllers/storagecluster/cephcluster.go
@@ -528,8 +528,8 @@ func getNetworkSpec(sc ocsv1.StorageCluster) rookCephv1.NetworkSpec {
 	// respect both the old way and the new way for enabling HostNetwork
 	networkSpec.HostNetwork = networkSpec.HostNetwork || sc.Spec.HostNetwork
 
-	// Enable the msgr2 port always if it's not an external cluster
-	if !sc.Spec.ExternalStorage.Enable {
+	// Don't require the msgr2 port if it's an provider or external/consumer cluster
+	if !sc.Spec.AllowRemoteStorageConsumers && !sc.Spec.ExternalStorage.Enable {
 		if networkSpec.Connections == nil {
 			networkSpec.Connections = &rookCephv1.ConnectionsSpec{}
 		}
--- a/controllers/storagecluster/ocs_operator_config.go
+++ b/controllers/storagecluster/ocs_operator_config.go
@@ -104,7 +104,8 @@ func (r *StorageClusterReconciler) getClusterID() string {
 
 // getCephFSKernelMountOptions returns the kernel mount options for cephfs
 func getCephFSKernelMountOptions(sc *ocsv1.StorageCluster) string {
-	if sc.Spec.ExternalStorage.Enable {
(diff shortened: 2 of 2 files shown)


controllers/storagecluster/cephcluster.go
```diff
@@ -528,8 +528,8 @@ func getNetworkSpec(sc ocsv1.StorageCluster) rookCephv1.NetworkSpec {
 	// respect both the old way and the new way for enabling HostNetwork
 	networkSpec.HostNetwork = networkSpec.HostNetwork || sc.Spec.HostNetwork
 
-	// Enable the msgr2 port always if it's not an external cluster
-	if !sc.Spec.ExternalStorage.Enable {
+	// Enable the msgr2 port always if it's not an provider or external/consumer cluster
```
parth-gr, line 531: // Enable the msgr2 port always if it's not an provider or external/consumer cluster

//Don't enable msg2 port if it's the provider or external/consumer cluster

Madhu-1:
> This change (adding `sc.Spec.AllowRemoteStorageConsumers` to the condition) only affects provider clusters, but we do not run ceph-csi on provider clusters. How does this solve the attached bug which

@nb-ohad If RequireMsg2 is set to true, Rook will setup the ceph cluster monitor to use the v2 port, which is 3300, ceph will not be listening on v1 ports anymore, and we also need some changes in cephcsi to support v2 ports like setting cephfs kernel mount option and rbd kernel map option. Currently we dont have this support in managed service/

controllers/storagecluster/cephcluster.go
```diff
@@ -528,8 +528,8 @@ func getNetworkSpec(sc ocsv1.StorageCluster) rookCephv1.NetworkSpec {
 	// respect both the old way and the new way for enabling HostNetwork
 	networkSpec.HostNetwork = networkSpec.HostNetwork || sc.Spec.HostNetwork
 
-	// Enable the msgr2 port always if it's not an external cluster
-	if !sc.Spec.ExternalStorage.Enable {
+	// Don't require the msgr2 port if it's an provider or external/consumer cluster
+	if !(sc.Spec.AllowRemoteStorageConsumers || sc.Spec.ExternalStorage.Enable) {
```
nb-ohad, line 532: May I suggest to rewrite the condition in the following way 
```suggestion
	if !sc.Spec.AllowRemoteStorageConsumers && !sc.Spec.ExternalStorage.Enable {
```
To me, this change might convey the condition in a more human-readable way

malayparida2000: done

Madhu-1 (commented):
Changes LGTM. @malayparida2000 have you tested the changes?

malayparida2000:
I dont have an external setup to test. But I tried to simulate it by installing ocs & creating pvc+pod with Not making requiremsgr2 true & passing nothing in the mount option. That worked.

iamniting (approved):

PR #1999, opened 2023-04-07
Correct storage cluster request queuing in OCSInitialization

iamniting:
Fix the issue where OCSInitialization adds its own name to the storage cluster request queue when there is a change in it. Modify the code to enqueue the storage cluster name instead.

### DIFF
--- a/controllers/storagecluster/storagecluster_controller.go
+++ b/controllers/storagecluster/storagecluster_controller.go
@@ -15,6 +15,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/version"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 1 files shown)


controllers/storagecluster/storagecluster_controller.go
```diff
@@ -117,6 +119,31 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		},
 	}
 
+	enqueueStorageClusterRequest := handler.EnqueueRequestsFromMapFunc(
```
umangachapagain, line 122: To limit the number of re-queues (if/when we support multiple clusters), let's only list in the namespace where this ocsinit resource is present and only requeue for ocsinit that are valid, not the ones that are ignored or in error. i.e. `sc.namespace=ocsinit.namespace && ocsinit.status.phase==ready`

umangachapagain (changes requested):

controllers/storagecluster/storagecluster_controller.go
iamniting: done

umangachapagain (approved):

PR #2002, opened 2023-04-10
Set Requiremsgr2 true if Encryption or Compression is enabled

malayparida2000:
/hold

Just for testing


travisn (commented):
From Rook's perspective, the `requireMsgr2: true` field is a no-op if the encryption or compression are enabled. The `requireMsgr2` is only useful for Rook if it is set when encryption and compression are not enabled, since this makes it simpler to transition to enable them later. 

But if we find in testing that no transition is needed such as mon failover as long as we update the msgr2 port for the csi driver and add the needed flags, then perhaps we find this flag is not needed either from OCS operator's perspective.

malayparida2000:
@travisn I have already tested that without any mon failover we can successfully enable encryption just by updating the csi driver to use v2 ports. 

So do you mean if we turn on encryption it automatically implicitly uses v2 port even if requiremsgr2 is not marked as true?

travisn:
> So do you mean if we turn on encryption it automatically implicitly uses v2 port even if requiremsgr2 is not marked as true?

Correct, if encryption or compression are enabled, rook will automatically use the v2 port.

malayparida2000:
I tried to test this. But I see even after enabling encryption, the csi configmap doesn't get updated to v2 ports

malayparida2000:
Closing this PR as the discussion here https://github.com/rook/rook/issues/12063

PR #2004, opened 2023-04-11
Added 'namespace' tag to queries

aruniiird:
This will allow the results to be grouped by namespace and not dependent on any particular one.

### DIFF
--- a/controllers/storagecluster/prometheus/localcephrules.yaml
+++ b/controllers/storagecluster/prometheus/localcephrules.yaml
@@ -11,27 +11,27 @@ spec:
   - name: ceph.rules
     rules:
     - expr: |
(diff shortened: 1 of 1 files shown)


umangachapagain (commented):
Have you verified if the results of both the queries is the same?

aruniiird:
> Have you verified if the results of both the queries is the same?

Yes verified each query (even the alerts, but could not fire each and every alerts, so tested the query without the condition. ie for eg:  'SOME-QUERY' > 0.75, just tested the 'SOME-QUERY' part to see it is giving appropriate results or not)

aruniiird:
Another related change added through PR:https://github.com/red-hat-storage/ocs-operator/pull/2008
@umangachapagain ^

aruniiird:
Good to go, let me know if anything to be changed

umangachapagain (approved):

PR #2023, opened 2023-04-20
Fix the OBC alert labels

aruniiird:
ObjectBucketClaim alerts had wrong label format

Wrong format: `{{$labels.objectbucketclaim}}`
Correct format:  `{{ $labels.objectbucketclaim }}`

### DIFF
--- a/controllers/storagecluster/prometheus/localcephrules.yaml
+++ b/controllers/storagecluster/prometheus/localcephrules.yaml
@@ -122,7 +122,7 @@ spec:
     rules:
     - alert: CephOSDCriticallyFull
       annotations:
-        description: Utilization of storage device {{ $labels.ceph_daemon }} of device_class type {{$labels.device_class}} has crossed 80% on host {{ $labels.hostname }}. Immediately free up some space or add capacity of type {{$labels.device_class}}.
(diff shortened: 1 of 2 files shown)


iamniting:
@aruniiird Is this required for 4.13? If yes do we have a bug?

aruniiird:
> @aruniiird Is this required for 4.13? If yes do we have a bug?

Yes @iamniting , BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2188032

umangachapagain (changes requested):
All the changed alerts are about OBC. Just changing the label to OB is not enough.
Either alert name, message, description everything needs to updated to OB or we need to add objectbucketclaim label.

umangachapagain (commented):
If you just change the first part of expr to `ocs_objectbucketclaim_info * on (namespace, objectbucket) group_left(objectbucketclaim)` it should work i.e. retain the objectbucketclaim label while joining the other half.

aruniiird:
> All the changed alerts are about OBC. Just changing the label to OB is not enough. Either alert name, message, description everything needs to updated to OB or we need to add objectbucketclaim label.

Yes, you are right. The label, `objectbucketclaim`, is already added. Only the formatting was wrong (if you see other jsonnet files for label template). Corrected it

iamniting (approved):

PR #2025, opened 2023-04-21
ms: fix label on storageclassrequest cr

subhamkrai:
Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2188666

### DIFF
--- a/services/provider/server/storageclassrequest.go
+++ b/services/provider/server/storageclassrequest.go
@@ -70,7 +70,7 @@ func (s *storageClassRequestManager) Create(
 			Namespace: s.namespace,
 			Labels: map[string]string{
 				controllers.ConsumerUUIDLabel: consumerUUID,
-				storageClassRequestName:       storageClassRequestName,
+				storageClassRequestNameLabel:  storageClassRequestName,
 			},
 		},
 		Spec: ocsv1alpha1.StorageClassRequestSpec{


subhamkrai:
@Nikhil-Ladha @iamniting

services/provider/server/storageclassrequest.go
```diff
@@ -70,7 +70,7 @@ func (s *storageClassRequestManager) Create(
 			Namespace: s.namespace,
 			Labels: map[string]string{
 				controllers.ConsumerUUIDLabel: consumerUUID,
-				storageClassRequestName:       storageClassRequestName,
+				storageClassRequestName:       storageClassRequestNameLabel,
```
iamniting, line 73: I think it should be vice versa @Madhu-1 WDYT?

iamniting (changes requested):

services/provider/server/storageclassrequest.go
Madhu-1: Yes it should always be key:value  key is the const identifier and value should be the request name

iamniting: @subhamkrai Can you pls change it as Madhu as confirmed it?

subhamkrai: right 😅

Madhu-1 (approved):

iamniting (approved):

Madhu-1:
we shouldnot use `bug 2188666` in the PR title as the fix is getting into main branch it's only required for backporting or else Bugzilla will update the bug once this PR gets merged in to the main branch.

subhamkrai:
> we shouldnot use `bug 2188666` in the PR title as the fix is getting into main branch it's only required for backporting or else Bugzilla will update the bug once this PR gets merged in to the main branch.

okay, got it.

PR #2029, opened 2023-04-24
enable single node cluster deployment

umangachapagain:
For development and demonstration purposes, StorageCluster can be deployed on a single node OpenShift cluster by setting environment variable "SINGLE_NODE" to true in ocs-operator pod.

Single node deployment for StorageCluster allows all pods to be deployed on a single node. This is not supported on a production cluster.

```
$ oc edit subscription ocs-operator
...
spec:
  config:
    env:
    - name: SINGLE_NODE
      value: "true"
...
```

### DIFF
--- a/controllers/storagecluster/cephcluster.go
+++ b/controllers/storagecluster/cephcluster.go
@@ -587,6 +587,9 @@ func getMinDeviceSetReplica(sc *ocsv1.StorageCluster) int {
 	if arbiterEnabled(sc) {
 		return defaults.ArbiterModeDeviceSetReplica
 	}
+	if statusutil.IsSingleNodeDeployment() {
+		return 1
+	}
 	return defaults.DeviceSetReplica
 }
 
@@ -903,7 +906,6 @@ func countAndReplicaOf(ds *ocsv1.StorageDeviceSet) (int, int) {
 }
 
 func newCephDaemonResources(sc *ocsv1.StorageCluster) map[string]corev1.ResourceRequirements {
-
 	custom := sc.Spec.Resources
 	resources := map[string]corev1.ResourceRequirements{
 		"mon": defaults.DaemonResources["mon"],
@@ -1011,7 +1013,7 @@ func generateMonSpec(sc *ocsv1.StorageCluster, nodeCount int) rookCephv1.MonSpec
 
 	return rookCephv1.MonSpec{
 		Count:                getMonCount(nodeCount, false),
-		AllowMultiplePerNode: false,
+		AllowMultiplePerNode: statusutil.IsSingleNodeDeployment(),
 	}
 }
 
@@ -1025,7 +1027,7 @@ func generateMgrSpec(sc *ocsv1.StorageCluster) rookCephv1.MgrSpec {
 	}
 	if sc.Spec.Mgr != nil && sc.Spec.Mgr.EnableActivePassive {
 		spec.Count = 2
-		spec.AllowMultiplePerNode = false
+		spec.AllowMultiplePerNode = statusutil.IsSingleNodeDeployment()
 	}
 
 	return spec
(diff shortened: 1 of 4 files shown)


controllers/storagecluster/cephcluster.go
```diff
@@ -1032,7 +1034,7 @@ func generateMgrSpec(sc *ocsv1.StorageCluster) rookCephv1.MgrSpec {
 	}
 	if sc.Spec.Mgr != nil && sc.Spec.Mgr.EnableActivePassive {
 		spec.Count = 2
```
travisn, line 1036: What about setting the mgr count to 1 if it's single node?

travisn (commented):
Not sure where it's initialized, but the `initData.Status.FailureDomain` should be set to `osd` in the single-node case. This will allow multiple replicas to be set on the pools, when the OSDs are on the same node.

controllers/storagecluster/cephcluster.go
umangachapagain: In a single node setup, we wouldn't enable active/passive mgr right? So, count would still be 1.

umangachapagain:
> Not sure where it's initialized, but the `initData.Status.FailureDomain` should be set to `osd` in the single-node case. This will allow multiple replicas to be set on the pools, when the OSDs are on the same node.

Done. How can I verify that this works as expected?

umangachapagain:
> > 
> > Done. How can I verify that this works as expected?
> 
> If the cluster is all initialized successfully, that would be sufficient. Any replica 2 or 3 pools would be blocked if another failure domain such as zone or host is used. In particular, the rgw pod would not be created because rook would be blocked initializing the rgw pools.
>
@travisn OSDs are going into crashloop and RGW pod is also not getting to ready state. Something is not right.

travisn:
> > > Done. How can I verify that this works as expected?
> > 
> > 
> > If the cluster is all initialized successfully, that would be sufficient. Any replica 2 or 3 pools would be blocked if another failure domain such as zone or host is used. In particular, the rgw pod would not be created because rook would be blocked initializing the rgw pools.
> 
> @travisn OSDs are going into crashloop and RGW pod is also not getting to ready state. Something is not right.

What is the error from the OSDs? Are they just hitting their memory limits? 512MB is very low for OSDs. Can you test without memory limits? 

If the OSDs aren't running, it will block the RGW pools from initializing and then the rgw pod won't be created.

umangachapagain:
> > 
> > @travisn OSDs are going into crashloop and RGW pod is also not getting to ready state. Something is not right.
> 
> What is the error from the OSDs? Are they just hitting their memory limits? 512MB is very low for OSDs. Can you test without memory limits?
> 
> If the OSDs aren't running, it will block the RGW pools from initializing and then the rgw pod won't be created.
>
That was the issue. Removing resource limit from OSDs worked. 
I have updated the sample file accordingly.

@travisn PTAL

travisn:
> > > @travisn OSDs are going into crashloop and RGW pod is also not getting to ready state. Something is not right.
> > 
> > 
> > What is the error from the OSDs? Are they just hitting their memory limits? 512MB is very low for OSDs. Can you test without memory limits?
> > If the OSDs aren't running, it will block the RGW pools from initializing and then the rgw pod won't be created.
> 
> That was the issue. Removing resource limit from OSDs worked. I have updated the sample file accordingly.
> 
> @travisn PTAL

Sounds good that it's working now without the limits. But for consistency, I'd suggest either all the requests/limits should be removed for other components as well, or else OSD resources should be added back to the list, just with higher limits. In minikube, at least Rook upstream works well without any limits. If it's just for testing or demos, not sure why limits are needed.

agarwal-mudit:
@umangachapagain This is nice, we should demo this in the next engineering meeting.

umangachapagain:
@travisn Removed all limits but kept the the resource requests. This helps us bypass ODF defaults for resources which is a lot.

@agarwal-mudit Let's do it.

iamniting (approved):

PR #2048, opened 2023-05-08
[WIP]Auto configure IPFamily in ceph cephcluster

sp98:
Checks cluster Network config:
- Assign IPv6 IPFamily to cephCluster spec in case of single stack IPv6
- Assign IPv4 IPFamily and Dual Stack false in case if the cluster is dual stack
(Dual Stack is not supported downstream. So use IPv4 and Dual Stack false if the cluster is dual stack. References:
https://bugzilla.redhat.com/show_bug.cgi?id=1804290
https://bugzilla.redhat.com/show_bug.cgi?id=2181350#c10 )


controllers/storagecluster/cephcluster.go
```diff
@@ -1205,3 +1217,33 @@ func getMonitoringClient() (*monitoringclient.Clientset, error) {
...
+	if err != nil {
+		return "", false, fmt.Errorf("could not get network config details : %v", err)
+	}
+
+	for _, cidr := range networkConfig.Status.ClusterNetwork {
+		if strings.Count(cidr.CIDR, ":") < 2 {
+			isIPv4 = true
+		} else if strings.Count(cidr.CIDR, ":") >= 2 {
+			isIPv6 = true
+		}
```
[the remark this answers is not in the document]
sp98: I was using this in the beginning. But then saw [this](https://stackoverflow.com/a/48519490/4418897). I will take a look at it again.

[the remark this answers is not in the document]
sp98: Checked again, I feel just checking the colon count might be a easier approach. In `To4` some IPv6 notations are considered as Ipv4 addresses.

agarwal-mudit:
@sp98 any progress on this?

sp98:
> @sp98 any progress on this?

Testing is pending. I'll test and close this by this week.

controllers/storagecluster/cephcluster_test.go
```diff
@@ -1193,3 +1195,71 @@ func TestCephClusterNetworkConnectionsSpec(t *testing.T) {
 		assert.DeepEqual(t, cc.Spec.Network.Connections, testCase.ccSpec.Network.Connections)
 	}
 }
+
+func TestGetIPFamilyConfig(t *testing.T) {
```
malayparida2000, line 1199: @sp98 The unit tests are still failing PTAL

controllers/storagecluster/cephcluster.go
```diff
@@ -203,6 +205,17 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
 		cephCluster.Spec.HealthCheck.DaemonHealth.Monitor.Timeout = "15m"
 	}
 
+	IPFamily, isDualStack, err := getIPFamilyConfig(r.Client)
```
malayparida2000, line 208: We are always getting false for isDualStack var, So is it necessary to be returned by the functions, can't we just set it here?

controllers/storagecluster/cephcluster.go
```diff
@@ -203,6 +205,17 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
...
+	if err != nil {
+		r.Log.Error(err, "failed to get IPFamily of the cluster")
+		return reconcile.Result{}, err
+	}
+
+	if IPFamily != "" {
```
malayparida2000, line 214: When the var IPFamily is empty we don't need to set the vars on the cephcluster at all?

sp98: That's right. IPFamily empty would mean that is a normal IPv4 cluster that we use currently. We don't need to set anything there.

controllers/storagecluster/cephcluster.go
sp98: We are always setting it to `false` because we currently don't support dualStack downstream. Rook supports it upstream though. I've added the details in the `getIPFamilyConfig` function. 

In future if we plan to support DualStack downstream as well, then we should return `true` for isDualStack if the OCP cluster is dual stack.

controllers/storagecluster/cephcluster_test.go
sp98: should be ok now.

malayparida2000 (approved):

controllers/storagecluster/cephcluster.go
```diff
@@ -1198,3 +1211,34 @@ func getMonitoringClient() (*monitoringclient.Clientset, error) {
...
+
+	// Dual Stack is not supported in downstream ODF.
+	// So use IPv4 and DualStack:False in case of dual stack cluster
+	// https://bugzilla.redhat.com/show_bug.cgi?id=1804290
+	// https://bugzilla.redhat.com/show_bug.cgi?id=2181350#c10
+	if isIPv4 && isIPv6 {
+		return rookCephv1.IPv4, false, nil
+	} else if isIPv6 && !isIPv4 { // IPv6 single stack cluster
+		return rookCephv1.IPv6, false, nil
+	}
```
iamniting, line 1241: Can you pls add a comment why we are sending empty if it is ipv4, So that there are no confusions in the future.

controllers/storagecluster/cephcluster.go
iamniting, line 214: Same related comment here as well.

iamniting (changes requested):

controllers/storagecluster/cephcluster.go
sp98: Agreed that it's confusing. I've updated the function to return IPFamily in case of single stack clusters.

controllers/storagecluster/cephcluster.go
sp98: updated with comment.

iamniting (approved):

controllers/storagecluster/cephcluster.go
```diff
@@ -203,6 +205,24 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
 		cephCluster.Spec.HealthCheck.DaemonHealth.Monitor.Timeout = "15m"
 	}
 
+	IPFamily, isDualStack, err := getIPFamilyConfig(r.Client)
```
Madhu-1, line 208: Use the lowercase variable name?

controllers/storagecluster/cephcluster.go
```diff
@@ -1198,3 +1218,30 @@ func getMonitoringClient() (*monitoringclient.Clientset, error) {
...
+		if strings.Count(cidr.CIDR, ":") < 2 {
+			isIPv4 = true
+		} else if strings.Count(cidr.CIDR, ":") >= 2 {
+			isIPv6 = true
+		}
+	}
+
+	if isIPv4 && isIPv6 {
+		return "", true, nil
+	} else if isIPv6 && !isIPv4 { // IPv6 single stack cluster
```
Madhu-1, line 1242:
```suggestion
} else if isIPv6 {
```
The above is enough right? we don't need to check isIPv4 here again

Madhu-1 (commented):
LGTM, small suggestions

controllers/storagecluster/cephcluster.go
sp98: updated.

controllers/storagecluster/cephcluster.go
sp98: Right. First condition should be enough. Update the code.

malayparida2000 (approved):

PR #2053, opened 2023-05-15
[WIP]Always use SSD device class for internal clusters

sp98:
If cluster is not external, then always use SSD as the device class for OSDs

PR is in progress. Need to fix the unit tests and check how it will effect the existing clusters.


malayparida2000:
Changes look good to me, But can you please add some commit message too.

controllers/storagecluster/cephcluster.go
```diff
@@ -782,6 +782,12 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster, serverVersion *version.
 			if ds.DeviceClass != "" {
 				crushDeviceClass = ds.DeviceClass
 			}
+
+			// Use SSD device class for all internal clusters
+			if !sc.Spec.ExternalStorage.Enable {
+				crushDeviceClass = DeviceTypeSSD
+			}
```
Madhu-1, line 789: This always set the value to `ssd` What if someone wants to set the deviceClass to some other name and use it for Blockpool?

travisn: Some open questions that came up in the readout while you were out: 
1. Is the volume claim template exposed in the StorageCluster CR? If so:
   a. If the crushDeviceClass is already set, allow that setting to remain (I believe this answers Madhu's question)
   b. If the crushDeviceClass is not set, default it to ssd
2. What about clusters that configure a metadata device (db/wal)? IIRC the data disk needs to be an hdd for that scenario, although that restriction may not hold for the osd-on-pvc case.
3. How are tuneFastDeviceClass and tuneSlowDeviceClass CephCluster CR settings related?

sp98: Thanks for the details @travisn . I'll check with the PM and get back to you about this.

sp98: Had a discussion with PM and clarified some questions over chat. Conclusion:

1. If the user has specified any deviceType (deviceClass) in the Storagecluster yaml, then use that.

2. If there is no user specified deviceClass, then for all internal clusters, use SSD deviceClass and set `TuneFastDeviceClass` to true.

3. For mixed media setups (data on lower performance disks and metadata/WAL on higher), we don't support HDD in ODF as of now. So we can follow the 2nd point above. Once HDD is supported, we can rely on user providing HDD deviceClass in the storageCluster yaml

@travisn @Madhu-1 Let me know what you think.

travisn: For the TuneFastDeviceClass and TuneSlowDeviceClass properties, let's first evaluate the defaults based on the current implementation, depending on the storage class name or platform. For example, we would always set it to slow for gp2, io1, etc, and always set it to the fast class for azure, etc. So we basically just default to `TuneFastDeviceClass = true` if it's not in the known list of platforms or storage classes. This gives us parity with today's behavior for these fast/slow properties.

sp98: @travisn OCS 
Currently OCS operator uses the following logic to tune for fast vs slow devices:

1. **Based on user provided deviceClass:**
    If the user provides HDD, then we use `TuneSlowDeviceClass` and in case of SSD we use `TuneFastDeviceClass` 

2.  **Based on the storageClass used:**
Provisioner: EBS
StorageClass: gp2
Tuning: TuneSlowDeviceClass

Provisioner: EBS
StorageClass: gp2-csi
Tuning: TuneSlowDeviceClass, 

Provisioner: EBS
StorageClass: io1
Tuning: TuneSlowDeviceClass, 

Provisioner: AzureDisk
StorageClass: managed-premium
Tuning: TuneFastDeviceClass

3. **Based on the platform**:
- TuneFastDeviceClass will be used for following platforms only
oVirt
IBMCloud
Azure

If `1` is satisfied, then we don't check `2` and `3` and so on.

sp98: So if we want to use `SSD` deviceClass (unless user specifies something in StorageCluster yaml), then `2` and `3` will be ignored.

sp98: Also confirmed with the PM that this is not a platform specific request to use SSDs. If user has not provided any deviceclass, then we want to use SSD with tuneFastDeviceClass

travisn: So the default behavior for AWS will change to use SSD with tuneFastDeviceClass? I thought that tuneSlowDeviceClass was specifically for AWS, so would like to just make sure this change is intentional for AWS.

sp98: PM is ok with having SSD and tuneFastDeviceClass as default, unless the user specifies a deviceClass explicitly. 
But I need to check the original implementation and whether this new implementation will have any negative impact.

agarwal-mudit:
@sp98 this is going to be stale soon

controllers/storagecluster/cephcluster.go
sp98: @travisn @Madhu-1 Can you take a look again?  The existing concerns we have about this were answered by the PM. More details about that can be found in the comments of the epic - https://issues.redhat.com/browse/RHSTOR-4326

controllers/storagecluster/cephcluster.go
```diff
@@ -114,6 +114,13 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
...
 		sc.Spec.StorageDeviceSets[i].Config.TuneFastDeviceClass = false
 
+		// Use SSD deviceClass and TuneFastDeviceClass for all internal clusters where deviceClass is not provided by the customer
+		if !sc.Spec.ExternalStorage.Enable && sc.Spec.StorageDeviceSets[i].DeviceClass == "" {
+			sc.Spec.StorageDeviceSets[i].DeviceClass = DeviceTypeSSD
+			sc.Spec.StorageDeviceSets[i].Config.TuneFastDeviceClass = true
+			continue
+		}
+
 		diskSpeed, err := r.checkTuneStorageDevices(ds)
```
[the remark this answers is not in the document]
sp98: I believe yes. Its still needed. For internal setups using custom `deviceClass` (like Managed Services), this new if block will be skipped.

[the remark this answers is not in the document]
travisn: I see now, thanks. So this means for example if an AWS admin wants to set the TuneSlowDeviceClass, they could set the `deviceClass: ssd`, and then the TuneSlowDeviceClass would be set since AWS is not in the slice `TuneFastPlatforms`. Not an intuitive behavior, but I can live with that.

travisn (approved):
/lgtm
/approve

Madhu-1 (approved):
LGTM

controllers/storagecluster/cephcluster.go
```diff
@@ -114,6 +114,13 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
 		sc.Spec.StorageDeviceSets[i].Config.TuneSlowDeviceClass = false
 		sc.Spec.StorageDeviceSets[i].Config.TuneFastDeviceClass = false
 
+		// Use SSD deviceClass and TuneFastDeviceClass for all internal clusters where deviceClass is not provided by the customer
+		if !sc.Spec.ExternalStorage.Enable && sc.Spec.StorageDeviceSets[i].DeviceClass == "" {
```
iamniting, line 118: @sp98 Can you pls mention in the API as well that if the value is empty it will be considered as SSD. or maybe we use default values kubebuilder tags?

iamniting (changes requested):
Can you pls correct the commit msg typo SDD -> SSD

controllers/storagecluster/cephcluster.go
sp98: Updated the API description for `DeviceClass`.

iamniting (approved):

PR #2054, opened 2023-05-18
generate latest yamls ClusterObjectStoreState Alert

weirdwiz:


weirdwiz:
> This should be auto generated from jsonnet using the make target. Why is this a manual change? @weirdwiz
yes, but somehow this file didn't get commited in that PR. it would be good to have a safeguard for this in the CI. current verify-generated-changes don't check for the alerts.

I wanted to make changes in the existing verify makefile target, but the dependency only has files that were generated using controller-gen. Not sure if it's a good place to add. Or do you think I'm missing a piece?

wdyt @umangachapagain

weirdwiz:
@umangachapagain made changes to only run the CI when files change in `metrics/*` directory

https://github.com/red-hat-storage/ocs-operator/pull/2061

umangachapagain (approved):

PR #2057, opened 2023-05-22
Check Connections.Encryption is not nil to avoid panic

malayparida2000:
If Network.Connections is not null, we should check Connections.Encryption is not nill before checking Encryption.Enabled. This issue was found when we try to create a storage cluster with Network.Connections.Compression.Enabled set to true & Connections.Encryption is nil.

https://issues.redhat.com/browse/RHSTOR-4218


malayparida2000:
> Let's create 2 separate PR's one for the fix and one for the rook update. Because if we require backport we can backport the fix and not the whole PR with the rook update

Ok, Keeping this one for the encryption spec check. Creating a separate PR for rook version change.

malayparida2000:
Here is the rook version upgrade PR https://github.com/red-hat-storage/ocs-operator/pull/2077

iamniting (approved):

PR #2058, opened 2023-05-24
Replica 1-Tech Preview- Pass appropriate value to rook via the csv for ceph non-resilient pools

malayparida2000:
When enabling non-resilient/replica-1 pools, it is necessary to
configure CSI by setting the CSI_ENABLE_TOPOLOGY variable to true and
specifying the value of the failure domain key using the
CSI_TOPOLOGY_DOMAIN_LABELS variable.

Previously, users had to manually modify the rook-ceph-operator-config
configmap. Now, the required values are passed automatically as ENV
variables via the CSV from the ocs-operator-config configmap.


iamniting (changes requested):
Can you pls change the commit-msg? As we are always setting the value there should not be any if.

malayparida2000:
> Can you pls change the commit-msg? As we are always setting the value there should not be any if.

Changed

iamniting (approved):

PR #2059, opened 2023-05-25
Allow changing default storageclass names

Nikhil-Ladha:
Added a new field named `StorageClassName` to the CephFilesystems,
CephBlockPools, CephNonResilientPools, CephObjectStores, NFS and Encryption
spec that takes the user defined name of the storage class created for
the respective type of resource.
And, `VirtualizationStorageClassName` for SC created for ceph block pools in
virtualization environment.
Added `MaxLength`, `Pattern` and duplicate values validation on the field.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2112929

### DIFF
--- a/api/v1/storagecluster_types.go
+++ b/api/v1/storagecluster_types.go
@@ -184,11 +184,24 @@ type ManageCephBlockPools struct {
 	ReconcileStrategy    string `json:"reconcileStrategy,omitempty"`
 	DisableStorageClass  bool   `json:"disableStorageClass,omitempty"`
 	DisableSnapshotClass bool   `json:"disableSnapshotClass,omitempty"`
+	// StorageClassName specifies the name of the storage class created for ceph block pools
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$
+	StorageClassName string `json:"storageClassName,omitempty"`
+	// VirtualizationStorageClassName specifies the name of the storage class created for ceph block pools
+	// for virtualization environment
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$
+	VirtualizationStorageClassName string `json:"virtualizationStorageClassName,omitempty"`
 }
 
 // ManageCephNonResilientPools defines how to reconcile ceph non-resilient pools
 type ManageCephNonResilientPools struct {
 	Enable bool `json:"enable,omitempty"`
+	// StorageClassName specifies the name of the storage class created for ceph non-resilient pools
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$
+	StorageClassName string `json:"storageClassName,omitempty"`
 	// ReconcileStrategy and other related fields are not used for now
 	// They can be added once the feature goes to GA
 }
@@ -198,6 +211,10 @@ type ManageCephFilesystems struct {
 	ReconcileStrategy    string `json:"reconcileStrategy,omitempty"`
 	DisableStorageClass  bool   `json:"disableStorageClass,omitempty"`
 	DisableSnapshotClass bool   `json:"disableSnapshotClass,omitempty"`
+	// StorageClassName specifies the name of the storage class created for cephfs
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$
+	StorageClassName string `json:"storageClassName,omitempty"`
 }
 
 // ManageCephObjectStores defines how to reconcile CephObjectStores
@@ -206,6 +223,10 @@ type ManageCephObjectStores struct {
 	DisableStorageClass bool   `json:"disableStorageClass,omitempty"`
 	GatewayInstances    int32  `json:"gatewayInstances,omitempty"`
 	DisableRoute        bool   `json:"disableRoute,omitempty"`
+	// StorageClassName specifies the name of the storage class created for ceph obc's
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$
+	StorageClassName string `json:"storageClassName,omitempty"`
 }
 
 // ManageCephObjectStoreUsers defines how to reconcile CephObjectStoreUsers
@@ -351,6 +372,10 @@ type NFSSpec struct {
 	// Enable specifies whether to enable NFS.
 	// +optional
 	Enable bool `json:"enable,omitempty"`
+	// StorageClassName specifies the name of the storage class created for NFS
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$
+	StorageClassName string `json:"storageClassName,omitempty"`
 }
 
 // MonitoringSpec controls the configuration of resources for exposing OCS metrics
@@ -371,7 +396,11 @@ type EncryptionSpec struct {
 	// +optional
 	ClusterWide bool `json:"clusterWide,omitempty"`
 	// +optional
-	StorageClass         bool                     `json:"storageClass,omitempty"`
+	StorageClass bool `json:"storageClass,omitempty"`
+	// StorageClassName specifies the name of the storage class created for ceph encrypted block pools
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$
+	StorageClassName     string                   `json:"storageClassName,omitempty"`
 	KeyManagementService KeyManagementServiceSpec `json:"kms,omitempty"`
 }
 
--- a/controllers/storagecluster/reconcile.go
+++ b/controllers/storagecluster/reconcile.go
@@ -261,6 +261,17 @@ func (r *StorageClusterReconciler) validateStorageClusterSpec(instance *ocsv1.St
 		return err
 	}
 
+	if err := validateCustomStorageClassNames(instance, r.Log); err != nil {
+		r.Log.Error(err, "Failed to validate custom StorageClassNames.", "StorageCluster", klog.KRef(instance.Namespace, instance.Name))
+		r.recorder.ReportIfNotPresent(instance, corev1.EventTypeWarning, statusutil.EventReasonValidationFailed, err.Error())
+		instance.Status.Phase = statusutil.PhaseError
+		if updateErr := r.Client.Status().Update(context.TODO(), instance); updateErr != nil {
+			r.Log.Error(updateErr, "Could not update StorageCluster.", "StorageCluster", klog.KRef(instance.Namespace, instance.Name))
+			return updateErr
+		}
+		return err
+	}
+
 	return nil
 }
 
@@ -721,5 +732,53 @@ func validateOverprovisionControlSpec(sc *ocsv1.StorageCluster, reqLogger logr.L
 			return fmt.Errorf("missing quotaname")
 		}
 	}
+	return nil
+}
+
+func validateCustomStorageClassNames(sc *ocsv1.StorageCluster, reqLogger logr.Logger) error {
+	scMap := make(map[string]bool)
+	duplicateNames := []string{}
+	if sc.Spec.ManagedResources.CephBlockPools.StorageClassName != "" {
+		if _, ok := scMap[sc.Spec.ManagedResources.CephBlockPools.StorageClassName]; ok {
+			duplicateNames = append(duplicateNames, "CephBlockPools")
+		}
+		scMap[sc.Spec.ManagedResources.CephBlockPools.StorageClassName] = true
+	}
+	if sc.Spec.ManagedResources.CephFilesystems.StorageClassName != "" {
+		if _, ok := scMap[sc.Spec.ManagedResources.CephFilesystems.StorageClassName]; ok {
+			duplicateNames = append(duplicateNames, "CephFilesystems")
+		}
+		scMap[sc.Spec.ManagedResources.CephFilesystems.StorageClassName] = true
+	}
+	if sc.Spec.ManagedResources.CephNonResilientPools.StorageClassName != "" {
+		if _, ok := scMap[sc.Spec.ManagedResources.CephNonResilientPools.StorageClassName]; ok {
+			duplicateNames = append(duplicateNames, "CephNonResilientPools")
+		}
+		scMap[sc.Spec.ManagedResources.CephNonResilientPools.StorageClassName] = true
+	}
+	if sc.Spec.ManagedResources.CephObjectStores.StorageClassName != "" {
+		if _, ok := scMap[sc.Spec.ManagedResources.CephObjectStores.StorageClassName]; ok {
+			duplicateNames = append(duplicateNames, "CephObjectStores")
+		}
+		scMap[sc.Spec.ManagedResources.CephObjectStores.StorageClassName] = true
+	}
+
+	if sc.Spec.NFS != nil && sc.Spec.NFS.Enable && sc.Spec.NFS.StorageClassName != "" {
+		if _, ok := scMap[sc.Spec.NFS.StorageClassName]; ok {
+			duplicateNames = append(duplicateNames, "NFS")
+		}
+		scMap[sc.Spec.NFS.StorageClassName] = true
+	}
+	if sc.Spec.Encryption.StorageClass && sc.Spec.Encryption.KeyManagementService.Enable && sc.Spec.Encryption.StorageClassName != "" {
+		if _, ok := scMap[sc.Spec.Encryption.StorageClassName]; ok {
+			duplicateNames = append(duplicateNames, "Encryption")
+		}
+		scMap[sc.Spec.Encryption.StorageClassName] = true
+	}
+
+	if len(duplicateNames) > 0 {
+		return fmt.Errorf("Duplicate StorageClass name(s) provided: %v", duplicateNames)
+	}
+
 	return nil
 }
--- a/controllers/storagecluster/storageclasses.go
+++ b/controllers/storagecluster/storageclasses.go
@@ -89,6 +89,10 @@ func (obj *ocsStorageClass) ensureDeleted(r *StorageClusterReconciler, instance 
 }
 
 func (r *StorageClusterReconciler) createStorageClasses(sccs []StorageClassConfiguration) error {
+	operatorNamespace, err := util.GetOperatorNamespace()
+	if err != nil {
+		return err
+	}
 	var skippedSC []string
 	for _, scc := range sccs {
 		if scc.reconcileStrategy == ReconcileStrategyIgnore || scc.disable {
@@ -97,7 +101,7 @@ func (r *StorageClusterReconciler) createStorageClasses(sccs []StorageClassConfi
 		sc := scc.storageClass
 
 		switch {
-		case strings.Contains(sc.Name, "-ceph-rbd") && !scc.isClusterExternal:
+		case (strings.Contains(sc.Name, "-ceph-rbd") || strings.Contains(sc.Provisioner, fmt.Sprintf("%s.rbd.csi.ceph.com", operatorNamespace))) && !scc.isClusterExternal:
 			// wait for CephBlockPool to be ready
 			cephBlockPool := cephv1.CephBlockPool{}
 			key := types.NamespacedName{Name: sc.Parameters["pool"], Namespace: sc.Parameters["clusterID"]}
@@ -110,7 +114,7 @@ func (r *StorageClusterReconciler) createStorageClasses(sccs []StorageClassConfi
 				skippedSC = append(skippedSC, sc.Name)
 				continue
 			}
-		case strings.Contains(sc.Name, "-ceph-non-resilient-rbd") && !scc.isClusterExternal:
+		case (strings.Contains(sc.Name, "-ceph-non-resilient-rbd") || sc.Parameters["topologyConstrainedPools"] != "") && !scc.isClusterExternal:
 			// wait for CephBlockPools to be ready
 			cephBlockPools := cephv1.CephBlockPoolList{}
 			err := r.Client.List(context.TODO(), &cephBlockPools, client.InNamespace(sc.Parameters["clusterID"]))
@@ -139,7 +143,7 @@ func (r *StorageClusterReconciler) createStorageClasses(sccs []StorageClassConfi
 				skippedSC = append(skippedSC, sc.Name)
 				continue
 			}
-		case strings.Contains(sc.Name, "-cephfs") && !scc.isClusterExternal:
+		case (strings.Contains(sc.Name, "-cephfs") || strings.Contains(sc.Provisioner, fmt.Sprintf("%s.cephfs.csi.ceph.com", operatorNamespace))) && !scc.isClusterExternal:
 			// wait for CephFilesystem to be ready
 			cephFilesystem := cephv1.CephFilesystem{}
 			key := types.NamespacedName{Name: sc.Parameters["fsName"], Namespace: sc.Parameters["clusterID"]}
@@ -152,7 +156,7 @@ func (r *StorageClusterReconciler) createStorageClasses(sccs []StorageClassConfi
 				skippedSC = append(skippedSC, sc.Name)
 				continue
 			}
-		case strings.Contains(sc.Name, "-nfs"):
+		case strings.Contains(sc.Name, "-nfs") || strings.Contains(sc.Provisioner, fmt.Sprintf("%s.nfs.csi.ceph.com", operatorNamespace)):
 			// wait for CephNFS to be ready
 			cephNFS := cephv1.CephNFS{}
 			key := types.NamespacedName{Name: sc.Parameters["nfsCluster"], Namespace: sc.Parameters["clusterID"]}
--- a/config/crd/bases/ocs.openshift.io_storageclusters.yaml
+++ b/config/crd/bases/ocs.openshift.io_storageclusters.yaml
@@ -480,6 +480,12 @@ spec:
                     type: object
                   storageClass:
                     type: boolean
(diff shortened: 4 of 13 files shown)


Nikhil-Ladha:
@umangachapagain @iamniting please review.

api/v1/storagecluster_types.go
```diff
@@ -105,6 +105,11 @@ type StorageClusterSpec struct {
 	DefaultStorageProfile string `json:"defaultStorageProfile,omitempty"`
 
 	StorageProfiles []StorageProfile `json:"storageProfiles,omitempty"`
+
+	// +kubebuilder:validation:MaxLength=230
+	// StorageClassName is the string to be used as the prefix for
+	// the default storage classes created by ocs-operator.
+	StorageClassName string `json:"storageClassName,omitempty"`
```
Madhu-1, line 112: This should be StorageClassNamePrefix if you are considering it as a prefix.

api/v1/storagecluster_types.go
```diff
@@ -105,6 +105,11 @@ type StorageClusterSpec struct {
 	DefaultStorageProfile string `json:"defaultStorageProfile,omitempty"`
 
 	StorageProfiles []StorageProfile `json:"storageProfiles,omitempty"`
+
+	// +kubebuilder:validation:MaxLength=230
```
Madhu-1, line 109: It's advisable to keep the value low so that you can create a storage class with a longer suffix in the future.

api/v1/storagecluster_types.go
Madhu-1, line 112: If someone changes the prefix after creating the storage cluster, what will be the impact? Should we prevent such operations from happening in the first place?

Nikhil-Ladha: Ack, will update the name

api/v1/storagecluster_types.go
Nikhil-Ladha: Hmm ok, what should be the value in that case in your opinion? Something around 200?

api/v1/storagecluster_types.go
Nikhil-Ladha: In that case, the operator will reconcile and create new storage classes with the updated names.
IMO, we should allow them to change the name if they want to, in that way they can delete older sc's if they are unused and create new one's.

iamniting (changes requested):
@Nikhil-Ladha I don't think we are looking for a prefix solution. If I am not mistaken we should look for exact names. @umangachapagain WDYT?

Nikhil-Ladha:
> @Nikhil-Ladha I don't think we are looking for a prefix solution. If I am not mistaken we should look for exact names. @umangachapagain WDYT?

Updated the code to change exact names of the sc's , PTAL.

api/v1/storagecluster_types.go
```diff
@@ -105,6 +105,9 @@ type StorageClusterSpec struct {
 	DefaultStorageProfile string `json:"defaultStorageProfile,omitempty"`
 
 	StorageProfiles []StorageProfile `json:"storageProfiles,omitempty"`
+
+	// StorageClassNames contains the name of the storage classes created for the respective storage class types
+	StorageClassNames map[string]string `json:"storageClassNames,omitempty"`
```
Madhu-1, line 110: Cannot we add storageclass names to `ManagedResourcesSpec` Spec?

Nikhil-Ladha: Good point, probably we can do that and keep all such details regarding auxiliary resources together.
WDYT @iamniting @umangachapagain ?

Nikhil-Ladha: I gave another look at the API structure, and think this might not be the best option.
We have NFS SC's as well, that are not part of the `ManagedResourcesSpec` field, and the `ManagedResourcesSpec` also contains info about Ceph dashboard, toolbox and couple of other things. So, adding the `StorageClassNames` field to it might not sit well.

Madhu-1: Just a thought ManagedResouceSpec contains different fields for block, file,nfs, etc you can add one more field storageclassname to reach specs to ManageCephBlockPools, ManageCephFilesystems, etc? and for NFS am not sure why it was kept outside of ManagedResouceSpec but you can add storageclassname to the NFS spec as well? 

@iamniting @umangachapagain WDYT?

umangachapagain: I like the idea. Probably better in comparison to using map.

Nikhil-Ladha: Ok, let me update the code accordingly.

Nikhil-Ladha:
Updated the code to add the `StorageClassName` field to the `ManagedResources` properties.

api/v1/storagecluster_types.go
```diff
@@ -206,6 +215,9 @@ type ManageCephObjectStores struct {
 	DisableStorageClass bool   `json:"disableStorageClass,omitempty"`
 	GatewayInstances    int32  `json:"gatewayInstances,omitempty"`
 	DisableRoute        bool   `json:"disableRoute,omitempty"`
+	// StorageClassName specifies the name of the storage class created for ceph obc's
+	// +kubebuilder:validation:MaxLength=253
+	StorageClassName string `json:"storageClassName,omitempty"`
```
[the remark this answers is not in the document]
Nikhil-Ladha: Yes, we need it for the `ceph-rgw` SC.

controllers/storagecluster/reconcile.go
```diff
@@ -716,3 +727,48 @@ func validateOverprovisionControlSpec(sc *ocsv1.StorageCluster, reqLogger logr.L
...
 	return nil
 }
+
+func validateCustomStorageClassNames(sc *ocsv1.StorageCluster) error {
+	scMap := make(map[string]string)
+	if sc.Spec.ManagedResources != (ocsv1.ManagedResourcesSpec{}) {
+		mr := sc.Spec.ManagedResources
+		if mr.CephBlockPools != (ocsv1.ManageCephBlockPools{}) && mr.CephBlockPools.StorageClassName != "" {
+			if _, has := scMap[mr.CephBlockPools.StorageClassName]; has {
+				return fmt.Errorf("duplicate storage class type provided")
```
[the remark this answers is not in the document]
Nikhil-Ladha: Updated.

api/v1/storagecluster_types.go
```diff
@@ -184,11 +184,17 @@ type ManageCephBlockPools struct {
 	ReconcileStrategy    string `json:"reconcileStrategy,omitempty"`
 	DisableStorageClass  bool   `json:"disableStorageClass,omitempty"`
 	DisableSnapshotClass bool   `json:"disableSnapshotClass,omitempty"`
+	// StorageClassName specifies the name of the storage class created for ceph block pools
+	// +kubebuilder:validation:MaxLength=253
```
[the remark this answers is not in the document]
Nikhil-Ladha: That's the maximum length of string allowed for object names in Kubernetes, as per this [doc](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#:~:text=This%20means%20the%20name%20must,no%20more%20than%20253%20characters)

controllers/storagecluster/reconcile.go
```diff
@@ -723,3 +734,48 @@ func validateOverprovisionControlSpec(sc *ocsv1.StorageCluster, reqLogger logr.L
 	}
 	return nil
 }
+
+func validateCustomStorageClassNames(sc *ocsv1.StorageCluster) error {
+	scMap := make(map[string]string)
+	if sc.Spec.ManagedResources != (ocsv1.ManagedResourcesSpec{}) {
```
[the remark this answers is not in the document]
Nikhil-Ladha: Ok, I think `ManagedResources` might never be null if there exists a storagecluster, but the other types like `CephNonResilientPools` and`CephObjectStores` could be null, right? Unless we explicitly enable/set some properties for them, isn't it?

api/v1/storagecluster_types.go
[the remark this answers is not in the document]
iamniting: I think this is not the same as the name, If I am not mistaken name's length is no more than 63 char. Can you pls verify that once in the cluster?

controllers/storagecluster/reconcile.go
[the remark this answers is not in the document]
iamniting: As this is not a pointer, we will be still able to access the variables, If they are not set they will be empty in case of a string.

api/v1/storagecluster_types.go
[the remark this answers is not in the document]
Nikhil-Ladha: Verified, it is same as name.
Also, I checked and we are not generating any other name using the storage class names. Probably, you can double check just to be double sure.

iamniting (changes requested):
A new PR was merged today, Can you pls take care of this SC as well?

api/v1/storagecluster_types.go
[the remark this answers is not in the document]
iamniting: I think 100 char should be enough, @umangachapagain WDYT?

Nikhil-Ladha:
> A new [PR](https://github.com/red-hat-storage/ocs-operator/pull/2076) was merged today, Can you pls take care of this SC as well?

Since, no new field was added for the kubevirt crd, how do you think we should take the SC name as the input from the user?
Maybe under the `CephBlockPools` struct itself named as `KubeVirtStorageClassName`? Since, it is copying the config of `CephBlockPools` and just adding few extra params to the SC created.

Nikhil-Ladha:
> > > A new [PR](https://github.com/red-hat-storage/ocs-operator/pull/2076) was merged today, Can you pls take care of this SC as well?
> > 
> > 
> > Since, no new field was added for the kubevirt crd, how do you think we should take the SC name as the input from the user? Maybe under the `CephBlockPools` struct itself named as `KubeVirtStorageClassName`? Since, it is copying the config of `CephBlockPools` and just adding few extra params to the SC created.
> 
> I think so too

Done, added a new field named `KubeVirtStorageClassName` for the new SC.

api/v1/storagecluster_types.go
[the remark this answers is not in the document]
umangachapagain: Why limit it to 100 if we can allow 253? Let's allow max possible length.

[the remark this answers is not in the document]
iamniting: I am thinking of a case where we may use the storageclass name to generate the other names. So in that case storageclass name should be smaller than the allowed limit right?

controllers/storagecluster/reconcile.go
```diff
@@ -723,3 +734,46 @@ func validateOverprovisionControlSpec(sc *ocsv1.StorageCluster, reqLogger logr.L
 	}
 	return nil
 }
+
+func validateCustomStorageClassNames(sc *ocsv1.StorageCluster) error {
+	scMap := make(map[string]string)
+	if sc.Spec.ManagedResources.CephBlockPools.StorageClassName != "" {
+		if _, has := scMap[sc.Spec.ManagedResources.CephBlockPools.StorageClassName]; has {
+			return fmt.Errorf("duplicate storage class name provided")
```
umangachapagain, line 742: Error message needs more details. Which StorageClass?
Also, better to validate all names before returning error. Let's say I use same name for 5 storageclasses, it should return 1 error and not 4.

controllers/storagecluster/reconcile.go
```diff
@@ -723,3 +734,46 @@ func validateOverprovisionControlSpec(sc *ocsv1.StorageCluster, reqLogger logr.L
 	}
 	return nil
 }
+
+func validateCustomStorageClassNames(sc *ocsv1.StorageCluster) error {
+	scMap := make(map[string]string)
```
umangachapagain, line 739:
```suggestion
	scMap := make(map[string]bool)
```

umangachapagain (changes requested):

umangachapagain (commented):
1. What happens if I provide StorageClass names in an upgraded cluster? Or, if cluster already has default names and I provide custom names?
2. What happens if I change custom names later? 

Tests should cover these basic scenarios.

api/v1/storagecluster_types.go
```diff
@@ -184,11 +184,21 @@ type ManageCephBlockPools struct {
...
+	StorageClassName string `json:"storageClassName,omitempty"`
+	// KubeVirtStorageClassName specifies the name of the storage class created for ceph block pools
+	// for virtualization environment
+	// +kubebuilder:validation:MaxLength=253
+	KubeVirtStorageClassName string `json:"kubeVirtStorageClassName,omitempty"`
```
malayparida2000, line 193: Let's call it something like VirtualizationStorageClassName, As in the future we might create this storageclass for some other crd also apart from kubevirt.

controllers/storagecluster/reconcile.go
Nikhil-Ladha: Updated

controllers/storagecluster/reconcile.go
Nikhil-Ladha: Done

Nikhil-Ladha:
> 1. What happens if I provide StorageClass names in an upgraded cluster? Or, if cluster already has default names and I provide custom names?
> 2. What happens if I change custom names later?
> 
> Tests should cover these basic scenarios.

Added these tests as well.

controllers/storagecluster/storageclasses.go
```diff
@@ -97,7 +97,7 @@ func (r *StorageClusterReconciler) createStorageClasses(sccs []StorageClassConfi
 		sc := scc.storageClass
 
 		switch {
-		case strings.Contains(sc.Name, "-ceph-rbd") && !scc.isClusterExternal:
+		case (strings.Contains(sc.Name, "-ceph-rbd") || strings.Contains(sc.Provisioner, "rbd.csi.ceph.com")) && !scc.isClusterExternal:
```
umangachapagain, line 100: These changes don't look related to changing names.

api/v1/storagecluster_types.go
```diff
@@ -184,11 +184,21 @@ type ManageCephBlockPools struct {
 	ReconcileStrategy    string `json:"reconcileStrategy,omitempty"`
 	DisableStorageClass  bool   `json:"disableStorageClass,omitempty"`
 	DisableSnapshotClass bool   `json:"disableSnapshotClass,omitempty"`
+	// StorageClassName specifies the name of the storage class created for ceph block pools
+	// +kubebuilder:validation:MaxLength=253
+	StorageClassName string `json:"storageClassName,omitempty"`
```
umangachapagain, line 189: We should add more validations to names. Kubernetes has multiple restrictions on naming. We should follow that to not hit into unwanted issue.
Ref: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/

controllers/storagecluster/reconcile.go
```diff
@@ -723,3 +734,57 @@ func validateOverprovisionControlSpec(sc *ocsv1.StorageCluster, reqLogger logr.L
 	}
 	return nil
 }
+
+func validateCustomStorageClassNames(sc *ocsv1.StorageCluster, reqLogger logr.Logger) error {
+	var err error
+	scMap := make(map[string]bool)
+	if sc.Spec.ManagedResources.CephBlockPools.StorageClassName != "" {
+		if _, has := scMap[sc.Spec.ManagedResources.CephBlockPools.StorageClassName]; has {
+			err = fmt.Errorf("Duplicate storage class name provided for CephBlockPools")
```
umangachapagain, line 743: This error will be overridden by errors from other lines. We need to combine all errors and return it in the end.
And to reduce the amount of logs, let's not log it after every check. Let's log it once in the end.

umangachapagain (changes requested):

controllers/storagecluster/storageclasses.go
Nikhil-Ladha: I did this change, because if we use custom names for SC then we would never hit the conditions in the switch the case i.e, to wait for the backend to be ready before creating the SC using them

controllers/storagecluster/reconcile.go
Nikhil-Ladha: Use an array and append the errors to it and log all of them together?

api/v1/storagecluster_types.go
Nikhil-Ladha: In that case, we would have to use a pattern validation like this
```
// StorageClassName specifies the name of the storage class created for cephfs
```
Ref: https://github.com/kubernetes/apimachinery/blob/v0.27.3/pkg/util/validation/validation.go#L215

Does this look fine?

Nikhil-Ladha: Or, should we just use the `validation` pkg from `apimachinery` and validate the name in the `validateStorageClassname` function?

umangachapagain: Pattern validation is good enough.

controllers/storagecluster/reconcile.go
umangachapagain: Yes. Can use errors.Join (needs go 1.20) to merge errors.

Nikhil-Ladha: Ok, updated the code to use an array for concatenating all the errors and logging in the end of the function, as the project still uses go 1.19

controllers/storagecluster/reconcile.go
```diff
@@ -723,3 +734,51 @@ func validateOverprovisionControlSpec(sc *ocsv1.StorageCluster, reqLogger logr.L
 	}
 	return nil
 }
+
+func validateCustomStorageClassNames(sc *ocsv1.StorageCluster, reqLogger logr.Logger) error {
+	scMap := make(map[string]bool)
+	duplicateNames := []string{}
+	if sc.Spec.ManagedResources.CephBlockPools.StorageClassName != "" {
+		if _, has := scMap[sc.Spec.ManagedResources.CephBlockPools.StorageClassName]; has {
+			duplicateNames = append(duplicateNames, fmt.Sprintf("Duplicate storage class name provided for CephBlockPools: %s", sc.Spec.ManagedResources.CephBlockPools.StorageClassName))
```
umangachapagain, line 743:
```suggestion
			duplicateNames = append(duplicateNames, "CephBlockPools"))
```
Just store resources with duplicate storageclass names. Error message is already there in the end, so we can remove duplicates.

controllers/storagecluster/reconcile.go
```diff
@@ -723,3 +734,51 @@ func validateOverprovisionControlSpec(sc *ocsv1.StorageCluster, reqLogger logr.L
...
+		}
+		scMap[sc.Spec.ManagedResources.CephBlockPools.StorageClassName] = true
+	}
+	if sc.Spec.ManagedResources.CephFilesystems.StorageClassName != "" {
+		if _, has := scMap[sc.Spec.ManagedResources.CephFilesystems.StorageClassName]; has {
+			duplicateNames = append(duplicateNames, fmt.Sprintf("Duplicate storage class name provided for CephFilesystems: %s", sc.Spec.ManagedResources.CephFilesystems.StorageClassName))
```
umangachapagain, line 749:
```suggestion
			duplicateNames = append(duplicateNames, "CephFilesystems"))
```

controllers/storagecluster/reconcile.go
```diff
@@ -723,3 +734,51 @@ func validateOverprovisionControlSpec(sc *ocsv1.StorageCluster, reqLogger logr.L
...
+	}
+	if sc.Spec.Encryption.StorageClass && sc.Spec.Encryption.KeyManagementService.Enable && sc.Spec.Encryption.StorageClassName != "" {
+		if _, has := scMap[sc.Spec.Encryption.StorageClassName]; has {
+			duplicateNames = append(duplicateNames, fmt.Sprintf("Duplicate storage class name provided for Encryption: %s", sc.Spec.Encryption.StorageClassName))
+		}
+		scMap[sc.Spec.Encryption.StorageClassName] = true
+	}
+
+	if len(duplicateNames) > 0 {
+		return fmt.Errorf("Duplicate storage class name(s) provided: %v", duplicateNames)
```
umangachapagain, line 780:
```suggestion
		return fmt.Errorf("Duplicate StorageClass name(s) provided for: %v", duplicateNames)
```

Nikhil-Ladha:
Tested out the changes, and everything is working fine.

umangachapagain (changes requested):

controllers/storagecluster/reconcile.go
Nikhil-Ladha: Ahh ok, updated it now. Sorry for the trouble 😓

controllers/storagecluster/reconcile.go
```diff
@@ -723,3 +734,51 @@ func validateOverprovisionControlSpec(sc *ocsv1.StorageCluster, reqLogger logr.L
...
+	}
+	if sc.Spec.ManagedResources.CephFilesystems.StorageClassName != "" {
+		if _, has := scMap[sc.Spec.ManagedResources.CephFilesystems.StorageClassName]; has {
+			duplicateNames = append(duplicateNames, "CephFilesystems")
+		}
+		scMap[sc.Spec.ManagedResources.CephFilesystems.StorageClassName] = true
+	}
+	if sc.Spec.ManagedResources.CephNonResilientPools.StorageClassName != "" {
+		if _, has := scMap[sc.Spec.ManagedResources.CephNonResilientPools.StorageClassName]; has {
+			duplicateNames = append(duplicateNames, "CephNonResilientPools: %s", sc.Spec.ManagedResources.CephNonResilientPools.StorageClassName)
```
umangachapagain, line 755:
```suggestion
			duplicateNames = append(duplicateNames, "CephNonResilientPools")
```

umangachapagain (commented):
/lgtm

umangachapagain:
/hold
for other reviewers

controllers/storagecluster/reconcile.go
```diff
@@ -723,3 +734,51 @@ func validateOverprovisionControlSpec(sc *ocsv1.StorageCluster, reqLogger logr.L
 	}
 	return nil
 }
+
+func validateCustomStorageClassNames(sc *ocsv1.StorageCluster, reqLogger logr.Logger) error {
+	scMap := make(map[string]bool)
+	duplicateNames := []string{}
+	if sc.Spec.ManagedResources.CephBlockPools.StorageClassName != "" {
+		if _, has := scMap[sc.Spec.ManagedResources.CephBlockPools.StorageClassName]; has {
```
Madhu-1, line 742: instead of `has` use `ok` as its a generic key used to check key exists.

controllers/storagecluster/storageclasses.go
Madhu-1: instead of `rbd.csi.ceph.com` check for `<storagecluster-namespace>.rbd.csi-ceph.com`

controllers/storagecluster/storageclasses.go
```diff
@@ -139,7 +139,7 @@ func (r *StorageClusterReconciler) createStorageClasses(sccs []StorageClassConfi
 				skippedSC = append(skippedSC, sc.Name)
 				continue
 			}
-		case strings.Contains(sc.Name, "-cephfs") && !scc.isClusterExternal:
+		case (strings.Contains(sc.Name, "-cephfs") || strings.Contains(sc.Provisioner, "cephfs.csi.ceph.com")) && !scc.isClusterExternal:
```
Madhu-1, line 142: instead of `cephfs.csi.ceph.com` check for `<storagecluster-namespace>.cephfs.csi-ceph.com`

controllers/storagecluster/storageclasses.go
```diff
@@ -152,7 +152,7 @@ func (r *StorageClusterReconciler) createStorageClasses(sccs []StorageClassConfi
 				skippedSC = append(skippedSC, sc.Name)
 				continue
 			}
-		case strings.Contains(sc.Name, "-nfs"):
+		case strings.Contains(sc.Name, "-nfs") || strings.Contains(sc.Provisioner, "nfs.csi.ceph.com"):
```
Madhu-1, line 155: same here as well

Madhu-1 (commented):
LGTM, small nit

controllers/storagecluster/storageclasses.go
Nikhil-Ladha: Ok, for that I think we can use the `clusterID` from the SC for getting the namespace as it contains only the namespace of the storagecluster. Else, we will have to fetch it explicitly, wdyt @iamniting ?

Madhu-1: clusterID cannot be used even though it looks like a namespace name, but it can be any arbitrary name. My suggestion is to use a namespace where the Rook operator is running.

iamniting: @Madhu-1 What problem do you see if we just use `cephfs.csi.ceph.com`?

Nikhil-Ladha: Oh ok, you mean do something like this, right?
```
podList := &corev1.PodList{}
	err := r.Client.List(context.TODO(), podList, client.MatchingLabels{"app": "rook-ceph-operator"})
	if err != nil {
		r.Log.Error(err, "Failed to list rook-ceph-operator pod")
		return err
	}
	// Get the namespace in which the rook operator is running
	scNS := podList.Items[0].Namespace
```

Madhu-1: That can be used based on the assumption that no one is running any csi driver with that prefix for there CSI drivers.

Madhu-1: You can use the namespace of the storagecluster in your case or the ocs-operator namespace can also be used.

iamniting: We can get the namespace easily See this https://github.com/red-hat-storage/ocs-operator/blob/main/main.go#L121-L125

Nikhil-Ladha: Made the update to use the ocs-operator namespace

Madhu-1 (approved):
LGTM

controllers/storagecluster/storageclasses.go
Nikhil-Ladha: Seems like the `OPERATOR_NAMESPACE` env variable is not set in the CI 🤔

Nikhil-Ladha: I think in that case we can add a double check for ns fetch, i.e, if the env variable is not there we can fetch the namespace from the `ocs-operator` namespace.

Nikhil-Ladha: Updated the test to set the `OPERATOR_NAMESPACE` env, looks good now.

Nikhil-Ladha:
@iamniting can you please add your review as well.

controllers/storagecluster/reconcile.go
```diff
@@ -723,3 +734,51 @@ func validateOverprovisionControlSpec(sc *ocsv1.StorageCluster, reqLogger logr.L
...
+	}
+	if sc.Spec.Encryption.StorageClass && sc.Spec.Encryption.KeyManagementService.Enable && sc.Spec.Encryption.StorageClassName != "" {
+		if _, ok := scMap[sc.Spec.Encryption.StorageClassName]; ok {
+			duplicateNames = append(duplicateNames, "Encryption")
+		}
+		scMap[sc.Spec.Encryption.StorageClassName] = true
+	}
+
+	if len(duplicateNames) > 0 {
+		return fmt.Errorf("Duplicate StorageClass name(s) provided for: %v", duplicateNames)
```
iamniting, line 780:
```suggestion
		return fmt.Errorf("Duplicate StorageClass name(s) provided: %v", duplicateNames)
```

iamniting (changes requested):

controllers/storagecluster/reconcile.go
Nikhil-Ladha: Done :)

iamniting (approved):

Nikhil-Ladha:
Seems like the merge is not possible due to the `hold` label, I think we can remove it now @umangachapagain @iamniting

PR #2065, opened 2023-05-31
Always call the ensureOCSOperatorConfig func irrespective of the mode

malayparida2000:
Earlier the ensureOCSOperatorConfig func was being called while creation of the cephcluster, but in Noobaa standalone mode it was not being called, as a cephcluster is not created in that mode. So we have to ensure that the ensureOCSOperatorConfig is called irrespective of the modes. So we are moving it to the reconcile.go file and putting it at the end of reconcilePhases func.

### DIFF
--- a/controllers/storagecluster/reconcile.go
+++ b/controllers/storagecluster/reconcile.go
@@ -295,6 +295,13 @@ func (r *StorageClusterReconciler) reconcilePhases(
 		return reconcile.Result{}, nil
 	}
 
+	// ensure the ocs-operator-config cm exists & has the correct values
+	err := r.ensureOCSOperatorConfig(instance)
+	if err != nil {
+		r.Log.Error(err, "Failed to ensure ocs-operator-config ConfigMap")
+		return reconcile.Result{}, err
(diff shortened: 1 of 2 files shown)


controllers/storagecluster/reconcile.go
```diff
@@ -506,6 +506,13 @@ func (r *StorageClusterReconciler) reconcilePhases(
 		}
 	}
 
+	// ensure the ocs-operator-config cm exists & has the correct values
+	err := r.ensureOCSOperatorConfig(instance)
```
iamniting, line 510: I think it will be better to move this even up in the reconciliation cycle as we might be waiting for status of rook resources and rook wont be running so there wont be a status update for those resources.

iamniting (changes requested):

controllers/storagecluster/reconcile.go
malayparida2000: Moved

iamniting (approved):

iamniting (approved):

PR #2072, opened 2023-06-01
Bug 2211598: [release-4.9] Always update the template parameters & objects to keep it up-to-date

malayparida2000:
Previously once the template existed we were not reconciling it. This resulted in the template using very old rook-ceph images and not having the newly added parameters. We want the template to be updated to use new parameters, any new ceph commands, latest rook-ceph image, etc. Trying to update the whole template object also doesn't work. So we need to update the parameters & objects separately to keep the template up-to-date.


malayparida2000:
4.9 has reached its EOL, The corresponding BZ is also closed now.

PR #2074, opened 2023-06-05
Add namespace controller for ReclaimSpace

Nikhil-Ladha:
Added a new namespace controller that adds a reclaimspace schedule annotation to all the namespaces starting with `openshift-*`, and restarts the csi-addons pod once done.

*NOTE:*
The controller doesn't restart the csi-addons pod if the value of reclaimspace annotation is updated to a different value manully. The pod needs to be restarted manually in that case.

### DIFF
--- a/PROJECT
+++ b/PROJECT
@@ -5,6 +5,12 @@ repo: github.com/red-hat-storage/ocs-operator
 resources:
 -
   controller: true
+  group: core
+  kind: Namespace
+  path: k8s.io/api/core/v1
+  version: v1
+-
+  controller: true
   domain: openshift.io
   group: ocs
   kind: OCSInitialization
--- a/controllers/namespace/namespace_controller.go
+++ b/controllers/namespace/namespace_controller.go
@@ -1,0 +1,210 @@
+package namespace
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"strings"
+	"sync"
+
+	"github.com/go-logr/logr"
+	"github.com/red-hat-storage/ocs-operator/controllers/util"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+// once is a pointer to the sync.Once type
+var once sync.Once
+
+// NamespaceReconciler reconciles the namespaces starting with openshift-*
+// nolint:revive
+type NamespaceReconciler struct {
+	client.Client
+	ctx    context.Context
+	Log    logr.Logger
+	Scheme *runtime.Scheme
+}
+
+// +kubebuilder:rbac:groups=core,resources=pods,verbs=list;delete
+// +kubebuilder:rbac:groups=core,resources=namespaces,verbs=get;list;watch;update;patch
+
+// Reconcile reads that state of the cluster for a Namespace object and makes changes based on the state read
+// i.e,the annotations of the namespace
+// The Controller will requeue the Request to be processed again if the returned error is non-nil or
+// Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
+func (r *NamespaceReconciler) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+
+	prevLogger := r.Log
+	defer func() { r.Log = prevLogger }()
+	r.Log = r.Log.WithValues("Request.Name", request.Name)
+	r.ctx = ctx
+
+	r.Log.Info("Reconciling Namespace.")
+
+	var err error
+	updated := false
+
+	once.Do(func() {
+		updated, err = r.annotateOpenshiftNamespaces()
+	})
+
+	// If it's not the first reconcile and the namespaces were not updated in the once.Do func call
+	if !updated {
+		ns := &corev1.Namespace{}
+		err = r.Client.Get(ctx, request.NamespacedName, ns)
+		if err != nil {
+			// the namespace is probably deleted
+			if errors.IsNotFound(err) {
+				r.Log.Error(err, fmt.Sprintf("Namespace %s does not exist", request.Name))
+				return reconcile.Result{}, nil
+			}
+			return reconcile.Result{}, err
+		}
+
+		// Annotate namespaces starting with openshift-*, but ignore namespace if
+		// it has annotation "'skipReclaimspaceSchedule':'true'" set
+		updated, err = r.annotateNamespaces([]corev1.Namespace{*ns})
+		if err != nil {
+			r.Log.Error(err, fmt.Sprintf("Failed to annotate namespace: %s", ns.Name))
+			return reconcile.Result{}, err
+		}
+	}
+
+	// Restart the csi-addons controller pod if any namespace was annotated
+	if updated {
+		err = r.restartCSIAddonsPod()
+	}
+
+	return reconcile.Result{}, err
+}
+
+// annotateOpenshiftNamespaces annotates all the namespaces starting with openshift-*
+func (r *NamespaceReconciler) annotateOpenshiftNamespaces() (bool, error) {
+	nsList := &corev1.NamespaceList{}
+	err := r.Client.List(r.ctx, nsList)
+	if err != nil {
+		r.Log.Error(err, "Failed to list Namespaces")
+		return false, err
+	}
+
+	openshiftNamespaces := []corev1.Namespace{}
+	for _, ns := range nsList.Items {
+		if strings.HasPrefix(ns.Name, "openshift-") {
+			openshiftNamespaces = append(openshiftNamespaces, ns)
+		}
+	}
+
+	return r.annotateNamespaces(openshiftNamespaces)
+}
+
+// annotateNamespaces annotate the provided namespace but, ignores if
+// shouldSkipAnnotation() returns true
+func (r *NamespaceReconciler) annotateNamespaces(opNs []corev1.Namespace) (bool, error) {
+	updated := false
+	for _, ns := range opNs {
+		if shouldSkipAnnotation(ns) {
+			continue
+		}
+		if ns.Annotations == nil {
+			ns.Annotations = make(map[string]string)
+		}
+		ns.Annotations["reclaimspace.csiaddons.openshift.io/schedule"] = "@weekly"
+
+		err := r.Client.Update(r.ctx, &ns)
+		if err != nil {
+			r.Log.Error(err, fmt.Sprintf("Failed to annotate namespace: %s", ns.Name))
+			return false, err
+		}
+		r.Log.Info(fmt.Sprintf("Successfully annotated namespace: %s", ns.Name))
+		updated = true
+	}
+	return updated, nil
+}
+
+// shouldSkipAnnotation checks if we should skip adding new annotations to the namespace
+func shouldSkipAnnotation(ns corev1.Namespace) bool {
+	if !ns.GetDeletionTimestamp().IsZero() || ns.Annotations["skipReclaimspaceSchedule"] == "true" ||
+		ns.Annotations["reclaimspace.csiaddons.openshift.io/schedule"] != "" {
+		return true
+	}
+	return false
+}
+
+// restartCSIAddonsPod restarts the CSIAddons pod by deleting all of its instances.
+func (r *NamespaceReconciler) restartCSIAddonsPod() error {
+	podList := &corev1.PodList{}
+	listOpts := []client.ListOption{
+		client.InNamespace(os.Getenv(util.OperatorNamespaceEnvVar)),
+		client.MatchingLabels{"app.kubernetes.io/name": "csi-addons"},
+	}
+	err := r.Client.List(r.ctx, podList, listOpts...)
+	if err != nil {
+		r.Log.Error(err, "Failed to get list of pods with label 'csi-addons'")
+		return err
+	}
+
+	// If no csi-addons pod is present
+	if len(podList.Items) == 0 {
+		r.Log.Error(err, "No csi-addons pod found")
+		return nil
+	}
+
+	for _, csiAddonsPod := range podList.Items {
+		err = r.Client.Delete(r.ctx, &csiAddonsPod, client.GracePeriodSeconds(0))
+		if err != nil {
+			r.Log.Error(err, "Failed to delete csi-addons pod")
+			return err
+		}
+	}
+
+	return nil
+}
+
+// SetupWithManager sets up a controller with a manager
+func (r *NamespaceReconciler) SetupWithManager(mgr ctrl.Manager) error {
+
+	predicateFunc := func(obj runtime.Object) bool {
+		instance, ok := obj.(*corev1.Namespace)
+		if !ok {
+			return false
+		}
+
+		// ignore if the namespace's name doesn't start with openshift-*
+		if !strings.HasPrefix(instance.Name, "openshift-") {
+			return false
+		}
+
+		// If the namespace has already been annotated, or skip annotation is set,
+		// or the namespace is marked for deletion
+		if shouldSkipAnnotation(*instance) {
+			return false
+		}
+
+		return true
+	}
+
+	nameSpacePredicate := predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			return predicateFunc(e.Object)
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			return predicateFunc(e.ObjectNew)
+		},
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			// Ignore delete events
+			return false
+		},
+	}
+
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&corev1.Namespace{},
+			builder.WithPredicates(nameSpacePredicate)).
+		Complete(r)
+}
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -144,6 +144,10 @@ rules:
   - namespaces
   verbs:
   - get
+  - list
+  - patch
+  - update
+  - watch
 - apiGroups:
   - ""
   resources:
@@ -152,6 +156,13 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - ""
+  resources:
+  - pods
+  verbs:
+  - delete
+  - list
 - apiGroups:
   - monitoring.coreos.com
   resources:
--- a/controllers/namespace/namespace_controller_test.go
+++ b/controllers/namespace/namespace_controller_test.go
@@ -1,0 +1,77 @@
+package namespace
+
+import (
+	"context"
+	"testing"
+
+	v1 "github.com/red-hat-storage/ocs-operator/api/v1"
+	"github.com/stretchr/testify/assert"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+func getReconciler(t *testing.T) NamespaceReconciler {
+	scheme := createFakeScheme(t)
+	client := fake.NewClientBuilder().WithScheme(scheme).Build()
+	log := logf.Log.WithName("controller_namespace_test")
+
+	return NamespaceReconciler{
+		Scheme: scheme,
+		Client: client,
+		Log:    log,
+	}
+}
+
+func createFakeScheme(t *testing.T) *runtime.Scheme {
+	scheme, err := v1.SchemeBuilder.Build()
+	if err != nil {
+		assert.Fail(t, "unable to build scheme")
+	}
+
+	err = corev1.AddToScheme(scheme)
+	if err != nil {
+		assert.Fail(t, "failed to add corev1 scheme")
+	}
+
+	return scheme
+}
+
+func TestReconcilerImplementsInterface(t *testing.T) {
+	reconciler := NamespaceReconciler{}
+	var i interface{} = &reconciler
+	_, ok := i.(reconcile.Reconciler)
+	assert.True(t, ok)
+}
+
+func TestNamespaceAnnotated(t *testing.T) {
+	ctx := context.TODO()
+	reconciler := getReconciler(t)
+	opNS := &corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "openshift-storage-test",
+		},
+	}
+	request := reconcile.Request{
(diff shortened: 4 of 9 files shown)


Nikhil-Ladha:
> NOTE:
The controller doesn't restart the csi-addons pod if the value of reclaimspace annotation is updated to a different value manully. The pod needs to be restarted manually in that case.

Do let me know, if we need to automate this case as well.

Nikhil-Ladha:
@iamniting @umangachapagain @Madhu-1 PTAL.

controllers/namespace/namespace_controller.go
```diff
@@ -0,0 +1,144 @@
...
+// NamespaceReconciler reconciles the namespaces starting with openshift-*
+// nolint:revive
+type NamespaceReconciler struct {
+	client.Client
+	ctx    context.Context
+	Log    logr.Logger
+	Scheme *runtime.Scheme
+}
+
+// +kubebuilder:rbac:groups=ocs.openshift.io,resources=*,verbs=get;list;watch;create;update;patch;delete
```
iamniting, line 26: Lets not use `*`, Give permissions to the required resources only.

controllers/namespace/namespace_controller.go
```diff
@@ -0,0 +1,144 @@
...
+	// Get all namespaces
+	err := r.List(ctx, ns)
+	if err != nil {
+		// Failed to list all the namespaces in the cluster
+		r.Log.Info("Re-fetching all the namespaces.")
+		return reconcile.Result{}, err
+	}
+
+	// Filter namespaces starting with openshift-*
+	opNS := getOpenshiftNamespaces(ns.Items)
```
iamniting, line 53: IMO we should not get the list of all namespaces and try to annotate all of them together. We should just add annotation only on the one namespace. 

We should use predicates to filter the openshift-* namespaces in the `SetupWithManager` func. 

@umangachapagain What is your opinion ?

Also, I  don't remember if we talked about restarting the CSI addons. I don't think it is a good idea to restart it. CSI-addons should pick up the changes automatically. @Madhu-1 WDYT?

iamniting (changes requested):
@Nikhil-Ladha We also need to update the `PROJECT` file with the new controller. for example https://github.com/red-hat-storage/odf-operator/blob/main/PROJECT#LL24C1-L28C20

controllers/namespace/namespace_controller.go
Nikhil-Ladha: Ack, though I was wondering that this should be changed to `csiaddons.openshift.io`, as we are only accessing the pods from that group and nothing else. WDYT?

controllers/namespace/namespace_controller.go
Nikhil-Ladha: I am not sure how we can handle it using predicates here. AFAIU, predicates monitors create,delete,update and generic events and in this case when the controller will be deployed the namespaces that we want to annotate are already created and no new events will be triggered by them, so how will we be getting an event for each namespace such that we could annotate them?  Please correct me if I am misunderstanding anything here. This is the [ref](https://sdk.operatorframework.io/docs/building-operators/golang/references/event-filtering/) I checked out.

Regarding the csiaddons pod, this was advised by Madhu himself based on our last discussion and the JIRA ticket, Engineering response.

Nikhil-Ladha: Updated the code to use predicates and annotation to only one namespace at a time.
However, I am not sure where could we make use of the `sync.Once` type in the code now, I don't see any repetitive action happening now 🤔 . Could you please help me out with it?

iamniting: Lets meet and sort this.

Nikhil-Ladha: Updated the code as per our discussion, PTAL.

controllers/namespace/namespace_controller.go
```diff
@@ -0,0 +1,194 @@
...
+			r.Log.Error(err, "Failed annotating openshift-* namespaces")
+			return reconcile.Result{}, err
+		}
+	}
+
+	// Restart the csi-addons controller pod if any namespace was annotated
+	if updated || controllerInit {
+		podList := &corev1.PodList{}
+		listOpts := []client.ListOption{
+			client.InNamespace(request.NamespacedName.Namespace),
```
iamniting, line 103: We can not make use of `request.NamespacedName.Namespace`. Instead get the namespace from `OPERATOR_NAMESPACE` env variable.

controllers/namespace/namespace_controller.go
```diff
@@ -0,0 +1,194 @@
...
+	defer func() { r.Log = prevLogger }()
+	r.Log = r.Log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
+	r.ctx = ctx
+
+	r.Log.Info("Reconciling Namespace.", "Namespace", klog.KRef(request.Namespace, request.Name))
+
+	controllerInit = false
+	once.Do(func() {
+		nsList := &corev1.NamespaceList{}
+		err := r.List(ctx, nsList)
```
iamniting, line 58: Can you pls use the `client` variable as well while making client calls?

controllers/namespace/namespace_controller.go
```diff
@@ -0,0 +1,194 @@
...
+			r.Log.Error(err, "Failed to get list of pods with label 'csi-addons'")
+			return reconcile.Result{}, err
+		}
+
+		// If no csi-addons pod is present
+		if len(podList.Items) == 0 {
+			return reconcile.Result{}, nil
+		}
+
+		csiAddonsPod := &podList.Items[0]
```
iamniting, line 117: I think we should restart the whole list, not just first occurence

controllers/namespace/namespace_controller.go
```diff
@@ -0,0 +1,194 @@
...
+		}
+		err := r.Update(r.ctx, &ns)
+		if err != nil {
+			r.Log.Error(err, fmt.Sprintf("Failed to annotate namespace %s", ns.Name))
+			return false, err
+		}
+	}
+
+	return updated, nil
+}
```
iamniting, line 194:
```suggestion
func (r *NamespaceReconciler) annotateOpenshiftNamespaces(opNs []corev1.Namespace) (bool, error) {
	updated := false

	for _, ns := range opNs {
		if shouldSkipAnnotation(ns.Annotations) {
			continue
		}

		if ns.Annotations == nil {
			ns.Annotations = make(map[string]string)
		}

		ns.Annotations["reclaimspace.csiaddons.openshift.io/schedule"] = "@weekly"
		
		updated = true
		

		err := r.Update(r.ctx, &ns)
		if err != nil {
			r.Log.Error(err, fmt.Sprintf("Failed to annotate namespace %s", ns.Name))
			return false, err
		}
	}

	return updated, nil
}
func shouldSkipAnnotation(annotations map[string]string) bool {
	if annotations["skipReclaimspaceSchedule"] == "true" {
		return true
	}

	if annotations["reclaimspace.csiaddons.openshift.io/schedule"] != "" {
		return true
	}

	return false
}

```

iamniting: Also let move this before setup with manager func.

controllers/namespace/namespace_controller.go
```diff
@@ -0,0 +1,194 @@
...
+)
+
+// once is a pointer to the sync.Once type
+var once sync.Once
+
+// funcError contains the error that occurred inside the once.Do() function call
+var funcErr error
+
+// controllerInit is a bool that denotes if the controller is reconciling for the first time
+var controllerInit bool
```
iamniting, line 28: I think we dont need controllerInit at all, It is making things complicated unnecessarily.

controllers/namespace/namespace_controller.go
```diff
@@ -0,0 +1,194 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+// once is a pointer to the sync.Once type
+var once sync.Once
+
+// funcError contains the error that occurred inside the once.Do() function call
+var funcErr error
```
iamniting, line 25: This should not be a global variable.

iamniting (changes requested):

controllers/namespace/namespace_controller.go
```diff
@@ -0,0 +1,194 @@
...
+	prevLogger := r.Log
+	defer func() { r.Log = prevLogger }()
+	r.Log = r.Log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
+	r.ctx = ctx
+
+	r.Log.Info("Reconciling Namespace.", "Namespace", klog.KRef(request.Namespace, request.Name))
+
+	controllerInit = false
+	once.Do(func() {
+		nsList := &corev1.NamespaceList{}
```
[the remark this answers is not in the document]
Nikhil-Ladha: The problem with defining all of the code in the function is that we can't return anything from the func, it has to be a void function.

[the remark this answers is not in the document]
iamniting: We can still do that.

```
var funcErr error
once.Do(func() {
	funcErr := r.newFunc(arguments)
})

if funcErr != nil {
        err := funcErr
        funcErr = nil
        return err
}

```

controllers/namespace/namespace_controller.go
Nikhil-Ladha: It is used to avoid refetching of the namespace name, annotations if the operator is reconciling for the first time.
Or, we are fine with just one extra fetch for the first reconcile? Or, we can handle it in some other way?

iamniting: I think it is okay because that will happen just for once.

controllers/namespace/namespace_controller.go
```diff
@@ -0,0 +1,197 @@
...
+
+// Reconcile reads that state of the cluster for a Namespace object and makes changes based on the state read
+// i.e,the annotations of the namespace
+// The Controller will requeue the Request to be processed again if the returned error is non-nil or
+// Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
+func (r *NamespaceReconciler) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+
+	prevLogger := r.Log
+	defer func() { r.Log = prevLogger }()
+	r.Log = r.Log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
```
iamniting, line 46: We can just pass one value as both will be the same.

controllers/namespace/namespace_controller.go
```diff
@@ -0,0 +1,197 @@
...
+	r.ctx = ctx
+
+	r.Log.Info("Reconciling Namespace.", "Namespace", klog.KRef(request.Namespace, request.Name))
```
iamniting, line 49:
```suggestion
	r.Log.Info("Reconciling Namespace.")
```

controllers/namespace/namespace_controller.go
```diff
@@ -0,0 +1,197 @@
...
+
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	ns := &corev1.Namespace{}
+	err = r.Client.Get(ctx, request.NamespacedName, ns)
+	if err != nil {
+		r.Log.Error(err, fmt.Sprintf("Namespace %s does not exist", request.Name))
+		return reconcile.Result{}, err
```
iamniting, line 66: If it is not found err we should return nil.

controllers/namespace/namespace_controller.go
```diff
@@ -0,0 +1,197 @@
...
+			openshiftNamespaces = append(openshiftNamespaces, ns)
+		}
+	}
+
+	return r.annotateOpenshiftNamespaces(openshiftNamespaces)
+}
+
+// annotateOpenshiftNamespaces annotate the provided namespace starting but, ignores if
+// it has annotation "'skipReclaimspaceSchedule':'true'" set
+func (r *NamespaceReconciler) annotateOpenshiftNamespaces(opNs []corev1.Namespace) (bool, error) {
```
iamniting, line 128:
```suggestion
func (r *NamespaceReconciler) annotateNamespaces(opNs []corev1.Namespace) (bool, error) {
```

controllers/namespace/namespace_controller.go
```diff
@@ -0,0 +1,197 @@
...
+			return reconcile.Result{}, nil
+		}
+
+		for _, csiAddonsPod := range podList.Items {
+			err = r.Client.Delete(ctx, &csiAddonsPod, client.GracePeriodSeconds(0))
+			if err != nil {
+				r.Log.Error(err, "Failed to delete csi-addons pod")
+				return reconcile.Result{}, err
+			}
+		}
```
iamniting, line 101: Let's make this as a separate func, Which can be called twice here and in once.DO, Because your namespace will be already updated here and it won't go into this condition at all.

iamniting (changes requested):

controllers/namespace/namespace_controller.go
Nikhil-Ladha: Yeah, right.
Updated the code to encapsulate this code within a separate function.

controllers/namespace/namespace_controller.go
```diff
@@ -0,0 +1,221 @@
...
+				return reconcile.Result{}, nil
+			}
+			return reconcile.Result{}, err
+		}
+
+		// Annotate namespaces starting with openshift-*, but ignore namespace if
+		// it has annotation "'skipReclaimspaceSchedule':'true'" set
+		updated, err := r.annotateNamespaces([]corev1.Namespace{*ns})
+		if err != nil {
+			r.Log.Error(err, "Failed annotating openshift-* namespaces")
```
iamniting, line 85: pls add the name of the namespace here, Also "namespaces" should be "namespace"

controllers/namespace/namespace_controller.go
```diff
@@ -0,0 +1,221 @@
...
+			if err != nil {
+				return reconcile.Result{}, err
+			}
+		}
+	}
+
+	return reconcile.Result{}, nil
+}
+
+// annotateAllOpenshiftNamespaces annotates all the namespaces starting with openshift-*
```
iamniting, line 101:
```suggestion
// annotateOpenshiftNamespaces annotates all the namespaces starting with openshift-*
```

controllers/namespace/namespace_controller.go
```diff
@@ -0,0 +1,221 @@
...
+	for _, ns := range nsList.Items {
+		if strings.HasPrefix(ns.Name, "openshift-") {
+			openshiftNamespaces = append(openshiftNamespaces, ns)
+		}
+	}
+
+	return r.annotateNamespaces(openshiftNamespaces)
+}
+
+// annotateOpenshiftNamespaces annotate the provided namespace starting but, ignores if
```
iamniting, line 120:
```suggestion
// annotateNamespaces annotate the provided namespace starting but, ignores if
```

controllers/namespace/namespace_controller.go
```diff
@@ -0,0 +1,221 @@
...
+		}
+		ns.Annotations["reclaimspace.csiaddons.openshift.io/schedule"] = "@weekly"
+
+		updated = true
+
+		err := r.Client.Update(r.ctx, &ns)
+		if err != nil {
+			r.Log.Error(err, fmt.Sprintf("Failed to annotate namespace %s", ns.Name))
+			return false, err
+		}
```
iamniting, line 139: Can you pls add a log line on the success as well.

iamniting (changes requested):
@umangachapagain @Madhu-1 Can you pls take a look?

controllers/namespace/namespace_controller.go
Nikhil-Ladha: Ack, done with the changes

controllers/namespace/namespace_controller.go
```diff
@@ -0,0 +1,222 @@
...
+			return reconcile.Result{}, err
+		}
+
+		// Annotate namespaces starting with openshift-*, but ignore namespace if
+		// it has annotation "'skipReclaimspaceSchedule':'true'" set
+		updated, err := r.annotateNamespaces([]corev1.Namespace{*ns})
+		if err != nil {
+			r.Log.Error(err, fmt.Sprintf("Failed to annotate namespace: %s", ns.Name))
+			return reconcile.Result{}, err
+		}
```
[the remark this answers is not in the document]
Nikhil-Ladha: Ack

controllers/namespace/namespace_controller.go
```diff
@@ -0,0 +1,222 @@
...
+	err := r.Client.List(r.ctx, nsList)
+	if err != nil {
+		r.Log.Error(err, "Failed to list Namespaces")
+		return false, err
+	}
+
+	openshiftNamespaces := []corev1.Namespace{}
+	for _, ns := range nsList.Items {
+		if strings.HasPrefix(ns.Name, "openshift-") {
+			openshiftNamespaces = append(openshiftNamespaces, ns)
```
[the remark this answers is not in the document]
Nikhil-Ladha: Ack, aggregated the `deletionTimestamp` check under the `shouldSkipAnnotation` func itself.

controllers/namespace/namespace_controller.go
```diff
@@ -0,0 +1,222 @@
...
+	for _, ns := range opNs {
+		if shouldSkipAnnotation(ns.Annotations) {
+			continue
+		}
+		if ns.Annotations == nil {
+			ns.Annotations = make(map[string]string)
+		}
+		ns.Annotations["reclaimspace.csiaddons.openshift.io/schedule"] = "@weekly"
+
+		updated = true
```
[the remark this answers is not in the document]
Nikhil-Ladha: Done

controllers/namespace/namespace_controller.go
```diff
@@ -0,0 +1,222 @@
...
+}
+
+// shouldSkipAnnotation checks if we should skip adding new annotations to the namespace
+func shouldSkipAnnotation(annotations map[string]string) bool {
+	if annotations["skipReclaimspaceSchedule"] == "true" {
+		return true
+	}
+	if annotations["reclaimspace.csiaddons.openshift.io/schedule"] != "" {
+		return true
+	}
```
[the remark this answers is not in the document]
Nikhil-Ladha: Done

PROJECT
```diff
@@ -3,6 +3,13 @@ layout: go.kubebuilder.io/v2
 projectName: ocs-operator
 repo: github.com/red-hat-storage/ocs-operator
 resources:
+-
+  controller: true
+  domain: k8s.io
```
iamniting, line 8:
```suggestion
```

iamniting (changes requested):

PROJECT
Nikhil-Ladha: Done :)

iamniting:
@Nikhil-Ladha Most of it looks good to me, Let's please test it thoroughly for the last time before merging.

Nikhil-Ladha:
> @Nikhil-Ladha Most of it looks good to me, Let's please test it thoroughly for the last time before merging.

Sure, doing another round of testing.

controllers/namespace/namespace_controller.go
```diff
@@ -0,0 +1,219 @@
...
+
+	return nil
+}
+
+// SetupWithManager sets up a controller with a manager
+func (r *NamespaceReconciler) SetupWithManager(mgr ctrl.Manager) error {
+
+	generationChangedPredicate := predicate.GenerationChangedPredicate{}
+
+	predicateFunc := func(obj runtime.Object) bool {
```
umangachapagain, line 188: Can we add more predicates to skip adding annotations instead of doing it in reconcile? If namespace has the skip annotation or annotation already exists, we don't need to trigger reconcile.

controllers/namespace/namespace_controller.go
```diff
@@ -0,0 +1,219 @@
...
+			return
+		}
+		if updated {
+			err = r.restartCSIAddonsPod()
+		}
+	})
+
+	if err != nil {
+		return reconcile.Result{}, err
+	}
```
Madhu-1, line 66: This can be moved to the end of the Reconcile.

controllers/namespace/namespace_controller.go
```diff
@@ -0,0 +1,219 @@
...
+			return reconcile.Result{}, err
+		}
+
+		// Restart the csi-addons controller pod if any namespace was annotated
+		if updated {
+			err = r.restartCSIAddonsPod()
+			if err != nil {
+				return reconcile.Result{}, err
+			}
+		}
```
Madhu-1, line 95: if you move this outside, you can have one instance of this code to cover both cases 1. In case of Once.Do and for single namespace Reconcile.

Madhu-1 (commented):
small nit, LGTM

controllers/namespace/namespace_controller.go
Nikhil-Ladha: We need it here, so that in case there is an error in `annotateOpenshiftNamespaces` func call, we should error out the reconcile and not proceed further.

Madhu-1: it will get reconciled again and you will add the annotation and restart it?

controllers/namespace/namespace_controller.go
Nikhil-Ladha: Sure, though I have kept the `skipAnnotation` check inside the reconcile as well, for the `once.Do` func call.

controllers/namespace/namespace_controller.go
Nikhil-Ladha: Understood, updated the code

controllers/namespace/namespace_controller.go
Nikhil-Ladha: Ahh, right. Updated the code accordingly.

Nikhil-Ladha:
> @Nikhil-Ladha Most of it looks good to me, Let's please test it thoroughly for the last time before merging.

Done with final round of testing all scenarios work fine.
- All namespaces annotated on first reconcile, and the csi-addons pod restarted
- After the first reconcile, no further reconcile happens because the predicate has been updated to check for existing annotations.
- Annotations are being added to newly created namespaces, and csi-addons pod restarted
- If we delete the annotation from ns, and don't set the skip annotation the annotation is added again by the controller and the pod restarted.
- If we set the skip annotation, no reconcile on the ns happens.
- If we set some other value than, `weekly` then it doesn't reconcile and update to `weekly`.

Hopefully, I have covered all the scenarios.

Madhu-1 (approved):
LGTM

controllers/namespace/namespace_controller.go
```diff
@@ -0,0 +1,210 @@
...
+		if strings.HasPrefix(ns.Name, "openshift-") {
+			openshiftNamespaces = append(openshiftNamespaces, ns)
+		}
+	}
+
+	return r.annotateNamespaces(openshiftNamespaces)
+}
+
+// annotateNamespaces annotate the provided namespace starting but, ignores if
+// it has annotation "'skipReclaimspaceSchedule':'true'" set
```
iamniting, line 108: Can you pls change this comment?

controllers/namespace/namespace_controller.go
```diff
@@ -0,0 +1,210 @@
...
+			continue
+		}
+		if ns.Annotations == nil {
+			ns.Annotations = make(map[string]string)
+		}
+		ns.Annotations["reclaimspace.csiaddons.openshift.io/schedule"] = "@weekly"
+
+		err := r.Client.Update(r.ctx, &ns)
+		if err != nil {
+			r.Log.Error(err, fmt.Sprintf("Failed to annotate namespace %s", ns.Name))
```
iamniting, line 122:
```suggestion
			r.Log.Error(err, fmt.Sprintf("Failed to annotate namespace: %s", ns.Name))
```

iamniting (changes requested):

controllers/namespace/namespace_controller.go
Nikhil-Ladha: Done

iamniting (approved):

PR #2075, opened 2023-06-06
Toggle between OCS Provider Server Service Type

rewantsoni:
Update the storageCluster CR to set the type of service as nodePort or LoadBalancer

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2212773

Different cloud providers needs different annotations to be added in order to bring up the ocs provider server service in private network, hence we might need to create a load balancer externally, hence added a field in storageCluster CR to toggle between NodePort and LoadBalancer Service type

Ref: https://kubernetes.io/docs/concepts/services-networking/service/#loadbalancer

### DIFF
--- a/api/v1/storagecluster_types.go
+++ b/api/v1/storagecluster_types.go
@@ -85,6 +85,12 @@ type StorageClusterSpec struct {
 	// AllowRemoteStorageConsumers Indicates that the OCS cluster should deploy the needed
 	// components to enable connections from remote consumers.
 	AllowRemoteStorageConsumers bool `json:"allowRemoteStorageConsumers,omitempty"`
+
+	// ProviderAPIServerServiceType Indicates the ServiceType for OCS Provider API Server Service.
+	// The supported values are NodePort or LoadBalancer. The default ServiceType is NodePort if the value is empty.
+	// This will only be used when AllowRemoteStorageConsumers is set to true
+	ProviderAPIServerServiceType corev1.ServiceType `json:"providerAPIServerServiceType,omitempty"`
+
 	// EnableCephTools toggles on whether or not the ceph tools pod
 	// should be deployed.
 	// Defaults to false
--- a/controllers/storagecluster/provider_server.go
+++ b/controllers/storagecluster/provider_server.go
@@ -145,6 +145,13 @@ func (o *ocsProviderServer) createDeployment(r *StorageClusterReconciler, instan
 
 func (o *ocsProviderServer) createService(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) (reconcile.Result, error) {
 
+	if instance.Spec.ProviderAPIServerServiceType != "" && instance.Spec.ProviderAPIServerServiceType != corev1.ServiceTypeNodePort &&
+		instance.Spec.ProviderAPIServerServiceType != corev1.ServiceTypeLoadBalancer {
+		err := fmt.Errorf("providerAPIServer only supports service of type %s and %s", corev1.ServiceTypeNodePort, corev1.ServiceTypeLoadBalancer)
+		r.Log.Error(err, "Failed to create/update service, Requested ServiceType is", "ServiceType", instance.Spec.ProviderAPIServerServiceType)
+		return reconcile.Result{}, err
+	}
+
 	desiredService := GetProviderAPIServerService(instance)
 	actualService := &corev1.Service{
 		ObjectMeta: metav1.ObjectMeta{
@@ -306,6 +313,10 @@ func GetProviderAPIServerDeployment(instance *ocsv1.StorageCluster) *appsv1.Depl
 
 func GetProviderAPIServerService(instance *ocsv1.StorageCluster) *corev1.Service {
 
+	if instance.Spec.ProviderAPIServerServiceType == "" {
+		instance.Spec.ProviderAPIServerServiceType = corev1.ServiceTypeNodePort
+	}
+
 	return &corev1.Service{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      ocsProviderServerName,
@@ -325,7 +336,7 @@ func GetProviderAPIServerService(instance *ocsv1.StorageCluster) *corev1.Service
 					TargetPort: intstr.FromString("ocs-provider"),
 				},
 			},
-			Type: corev1.ServiceTypeLoadBalancer,
+			Type: instance.Spec.ProviderAPIServerServiceType,
 		},
 	}
 }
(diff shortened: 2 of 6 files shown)


api/v1/storagecluster_types.go
```diff
@@ -105,6 +105,10 @@ type StorageClusterSpec struct {
 	DefaultStorageProfile string `json:"defaultStorageProfile,omitempty"`
 
 	StorageProfiles []StorageProfile `json:"storageProfiles,omitempty"`
+
+	// EnableLoadBalancerAPIServerService Indicates if the OCS Provider API Server Service should be
+	// of type NodePort or LoadBalancer.
+	EnableLoadBalancerAPIServerService bool `json:"enableLoadBalancerAPIServerService,omitempty"`
```
iamniting, line 111: Making the flag a boolean value can cause issues with upgrades since the existing clusters have the service configured as a load balancer by default, and the flag's value is set to false. This will change the service type from load balancer to nodeport on the existing clusters, which is not desired.

To avoid this, I propose using a string instead of a boolean. By default, if the string is empty, we will create a load balancer service. This approach ensures that we can easily introduce new service types in the future without requiring any modifications to the API.

Furthermore, can you please move this flag right after the AllowRemoteStorageConsumers flag. Additionally, in the comment, mention that this flag will only be considered if AllowRemoteStorageConsumers is set to true.

iamniting (changes requested):

api/v1/storagecluster_types.go
```diff
@@ -85,6 +85,12 @@ type StorageClusterSpec struct {
 	// AllowRemoteStorageConsumers Indicates that the OCS cluster should deploy the needed
 	// components to enable connections from remote consumers.
 	AllowRemoteStorageConsumers bool `json:"allowRemoteStorageConsumers,omitempty"`
+
+	// ProviderAPIServerServiceType Indicates the ServiceType for OCS Provider API Server Service
+	// Supported Values are NodePort or LoadBalancer. The default ServiceType is NodePort
+	// This will also be used when AllowRemoteStorageConsumers is set to true
```
iamniting, line 91:
```suggestion
	// This will only be used when AllowRemoteStorageConsumers is set to true
```

api/v1/storagecluster_types.go
```diff
@@ -85,6 +85,12 @@ type StorageClusterSpec struct {
 	// AllowRemoteStorageConsumers Indicates that the OCS cluster should deploy the needed
 	// components to enable connections from remote consumers.
 	AllowRemoteStorageConsumers bool `json:"allowRemoteStorageConsumers,omitempty"`
+
+	// ProviderAPIServerServiceType Indicates the ServiceType for OCS Provider API Server Service
+	// Supported Values are NodePort or LoadBalancer. The default ServiceType is NodePort
```
iamniting, line 90:
```suggestion
	// Supported Values are NodePort or LoadBalancer. The default ServiceType is NodePort if the value is empty
```

controllers/storagecluster/provider_server.go
```diff
@@ -145,6 +145,12 @@ func (o *ocsProviderServer) createDeployment(r *StorageClusterReconciler, instan
 
 func (o *ocsProviderServer) createService(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) (reconcile.Result, error) {
 
+	if instance.Spec.ProviderAPIServerServiceType == corev1.ServiceTypeClusterIP || instance.Spec.ProviderAPIServerServiceType == corev1.ServiceTypeExternalName {
+		err := fmt.Errorf("service of Type %s is not supported by the operator", instance.Spec.ProviderAPIServerServiceType)
```
iamniting, line 149: Can you pls mention that only 2 types of services are supported for now which are nodeport and loadbalancer.

iamniting (changes requested):

iamniting:
Can you also look at the failures and try to resolve them

api/v1/storagecluster_types.go
```diff
@@ -85,6 +85,12 @@ type StorageClusterSpec struct {
 	// AllowRemoteStorageConsumers Indicates that the OCS cluster should deploy the needed
 	// components to enable connections from remote consumers.
 	AllowRemoteStorageConsumers bool `json:"allowRemoteStorageConsumers,omitempty"`
+
+	// ProviderAPIServerServiceType Indicates the ServiceType for OCS Provider API Server Service
+	// Supported Values are NodePort or LoadBalancer. The default ServiceType is NodePort if the value is empty
+	// This will only be used when AllowRemoteStorageConsumers is set to true
```
iamniting, line 91:
```suggestion
	// ProviderAPIServerServiceType Indicates the ServiceType for OCS Provider API Server Service.
	// The supported Values are NodePort or LoadBalancer. The default ServiceType is NodePort if the value is empty.
	// This will only be used when AllowRemoteStorageConsumers is set to true
```

iamniting (changes requested):

deploy/csv-templates/noobaa-csv.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: operators.coreos.com/v1alpha1
+make geapiVersion: operators.coreos.com/v1alpha1
```
iamniting, line 1: I guess this was added accidentally.

iamniting (changes requested):

deploy/csv-templates/noobaa-csv.yaml
rewantsoni: updated

controllers/storagecluster/provider_server.go
```diff
@@ -145,6 +145,12 @@ func (o *ocsProviderServer) createDeployment(r *StorageClusterReconciler, instan
 
 func (o *ocsProviderServer) createService(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) (reconcile.Result, error) {
 
+	if instance.Spec.ProviderAPIServerServiceType == corev1.ServiceTypeClusterIP || instance.Spec.ProviderAPIServerServiceType == corev1.ServiceTypeExternalName {
```
iamniting, line 148: I believe we should check the services we currently support. This way, we can mitigate the risk associated with the potential addition of new services in the future. By focusing on the ones we already support, we can avoid encountering such problems.

```
if instance.Spec.ProviderAPIServerServiceType != "" &&
    instance.Spec.ProviderAPIServerServiceType != corev1.ServiceTypeNodePort &&
    instance.Spec.ProviderAPIServerServiceType != corev1.ServiceTypeLoadBalancer
```
It is just a suggestion and not a must, Let me know if you would like to do that, otherwise I will approve it.

iamniting (changes requested):

controllers/storagecluster/provider_server.go
rewantsoni: I will make this change and update the PR

iamniting (approved):

PR #2076, opened 2023-06-08
Create a storageclass with rxbounce option if kubevirt crd is present

malayparida2000:
We will watch for the virtualmachines.kubevirt.io crd. Then create a special ceph rbd storageclass for the virtualization
environment. This storageclass will have the rxbounce option to use a bounce buffer during I/O operation. This is mostly targeted to fix issues with windows VMs using ceph rbd. This storageclass has all the same options as the default ceph rbd storageclass with this extra rxbounce option.

### DIFF
--- a/controllers/storagecluster/generate.go
+++ b/controllers/storagecluster/generate.go
@@ -72,6 +72,10 @@ func generateNameForCephBlockPoolSC(initData *ocsv1.StorageCluster) string {
 	return fmt.Sprintf("%s-ceph-rbd", initData.Name)
 }
 
+func generateNameForCephBlockPoolVirtualizationSC(initData *ocsv1.StorageCluster) string {
+	return fmt.Sprintf("%s-ceph-rbd-virtualization", initData.Name)
+}
+
 func generateNameForNonResilientCephBlockPoolSC(initData *ocsv1.StorageCluster) string {
 	return fmt.Sprintf("%s-ceph-non-resilient-rbd", initData.Name)
 }
--- a/controllers/storagecluster/storageclasses.go
+++ b/controllers/storagecluster/storageclasses.go
@@ -12,6 +12,7 @@ import (
 	cephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
 	corev1 "k8s.io/api/core/v1"
 	storagev1 "k8s.io/api/storage/v1"
+	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -280,6 +281,16 @@ func newCephBlockPoolStorageClassConfiguration(initData *ocsv1.StorageCluster) S
 	}
 }
 
+// newCephBlockPoolVirtualizationStorageClassConfiguration generates configuration options for a Ceph Block Pool StorageClass for virtualization environment.
+func newCephBlockPoolVirtualizationStorageClassConfiguration(initData *ocsv1.StorageCluster) StorageClassConfiguration {
+	virtualizationStorageClassConfig := newCephBlockPoolStorageClassConfiguration(initData)
+	virtualizationStorageClassConfig.storageClass.ObjectMeta.Name = generateNameForCephBlockPoolVirtualizationSC(initData)
+	virtualizationStorageClassConfig.storageClass.ObjectMeta.Annotations["description"] = "Provides RWO and RWX Block volumes suitable for Virtual Machine disks"
+	virtualizationStorageClassConfig.storageClass.Parameters["mounter"] = "rbd"
+	virtualizationStorageClassConfig.storageClass.Parameters["mapOptions"] = "krbd:rxbounce"
+	return virtualizationStorageClassConfig
+}
+
 // newNonResilientCephBlockPoolStorageClassConfiguration generates configuration options for a Non-Resilient Ceph Block Pool StorageClass.
 func newNonResilientCephBlockPoolStorageClassConfiguration(initData *ocsv1.StorageCluster) StorageClassConfiguration {
 	persistentVolumeReclaimDelete := corev1.PersistentVolumeReclaimDelete
@@ -394,6 +405,12 @@ func (r *StorageClusterReconciler) newStorageClassConfigurations(initData *ocsv1
 	ret := []StorageClassConfiguration{
 		newCephFilesystemStorageClassConfiguration(initData),
 		newCephBlockPoolStorageClassConfiguration(initData),
+	}
+	// If kubevirt crd is present, we create a specialized rbd storageclass for virtualization environment
+	kvcrd := &extv1.CustomResourceDefinition{}
+	err := r.Client.Get(context.TODO(), types.NamespacedName{Name: "virtualmachines.kubevirt.io", Namespace: ""}, kvcrd)
+	if err == nil {
+		ret = append(ret, newCephBlockPoolVirtualizationStorageClassConfiguration(initData))
 	}
 	if initData.Spec.ManagedResources.CephNonResilientPools.Enable {
 		ret = append(ret, newNonResilientCephBlockPoolStorageClassConfiguration(initData))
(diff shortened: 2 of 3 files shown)


controllers/storagecluster/generate.go
```diff
@@ -72,6 +72,10 @@ func generateNameForCephBlockPoolSC(initData *ocsv1.StorageCluster) string {
 	return fmt.Sprintf("%s-ceph-rbd", initData.Name)
 }
 
+func generateNameForVirtualCephBlockPoolSC(initData *ocsv1.StorageCluster) string {
+	return fmt.Sprintf("%s-ceph-rbd-virtualization", initData.Name)
```
nixpanic, line 76: I prefer this name, but there is also `%s-ceph-non-resilient-rbd` below. Maybe call this `%s-ceph-virtualization-rbd` to match it a little more?

controllers/storagecluster/storageclasses.go
```diff
@@ -281,6 +281,46 @@ func newCephBlockPoolStorageClassConfiguration(initData *ocsv1.StorageCluster) S
...
+func newVirtualizationCephBlockPoolStorageClassConfiguration(initData *ocsv1.StorageCluster) StorageClassConfiguration {
+	persistentVolumeReclaimDelete := corev1.PersistentVolumeReclaimDelete
+	allowVolumeExpansion := true
+	managementSpec := initData.Spec.ManagedResources.CephBlockPools
+	return StorageClassConfiguration{
+		storageClass: &storagev1.StorageClass{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: generateNameForVirtualCephBlockPoolSC(initData),
+				Annotations: map[string]string{
+					"description": "Provides RWO Filesystem volumes, and RWO and RWX Block volumes for virtualization environment",
```
nixpanic, line 295: This can be phased a little differently. Maybe something along the lines of (feel free to adapt)

> Optimized RWX Block volumes for Virtual Machine workloads

Virtual Machines are encouraged to use RWX Block volumes, not RWO Filesystems. When a Filesystem on RBD is used, the option is probably not important either (an image-file on a filesystem has different behaviour). RWX Block volumes allow live-migration of VMs between different workernodes, this is not possible with RWO Filesystem volumes.

nixpanic (commented):
/cc @aglitke @idryomov

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -161,8 +163,18 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&appsv1.Deployment{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
 		Owns(&corev1.Service{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
 		Owns(&corev1.ConfigMap{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
-		Watches(&source.Kind{Type: &ocsv1.OCSInitialization{}}, enqueueStorageClusterRequest)
-
+		Watches(&source.Kind{Type: &ocsv1.OCSInitialization{}}, enqueueStorageClusterRequest).
+		// Watcher for kubevirt virtualmachines crd to detect openshift virtualization installation
+		Watches(
+			&source.Kind{
+				Type: &extv1.CustomResourceDefinition{
```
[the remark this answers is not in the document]
malayparida2000: It's already there https://github.com/red-hat-storage/ocs-operator/blob/fae3a318d2179fbc7d5db1ecdb7ce1c7686b31ab/config/rbac/role.yaml#L8-L17

controllers/storagecluster/generate.go
idryomov: Bike shedding a bit, but imagining a non-resilient virtualization storage class, a natural name for it would be `ceph-non-resilient-rbd-virtualization` (i.e. `virtualization` applies to `rbd`, not to `ceph`).  I like the original better.

controllers/storagecluster/storageclasses.go
```diff
@@ -280,6 +281,46 @@ func newCephBlockPoolStorageClassConfiguration(initData *ocsv1.StorageCluster) S
...
+// newVirtualizationCephBlockPoolStorageClassConfiguration generates configuration options for a Ceph Block Pool StorageClass for virtualization.
+
+func newVirtualizationCephBlockPoolStorageClassConfiguration(initData *ocsv1.StorageCluster) StorageClassConfiguration {
+	persistentVolumeReclaimDelete := corev1.PersistentVolumeReclaimDelete
+	allowVolumeExpansion := true
+	managementSpec := initData.Spec.ManagedResources.CephBlockPools
+	return StorageClassConfiguration{
+		storageClass: &storagev1.StorageClass{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: generateNameForVirtualCephBlockPoolSC(initData),
```
idryomov, line 293: "Virtual" is a very overloaded term.  "Virt" would be a better shorthand here IMO (or just spell it out entirely).

controllers/storagecluster/generate.go
nixpanic: right, I guess a naming convention could be `${clustername}-ceph-${storage_type}-${workload}`

That makes `%s-ceph-non-resilient-rbd-kafka` and `%s-ceph-rbd-postgresql` useful and consistent names too.

:bike:

malayparida2000: Settling on %s-ceph-rbd-virtualization then.

controllers/storagecluster/storageclasses.go
malayparida2000: I had copied the description from the default storageclass. So should we change it?

controllers/storagecluster/storageclasses.go
```diff
@@ -395,6 +436,12 @@ func (r *StorageClusterReconciler) newStorageClassConfigurations(initData *ocsv1
 		newCephFilesystemStorageClassConfiguration(initData),
 		newCephBlockPoolStorageClassConfiguration(initData),
 	}
+	// If openshift virtualization is installed with the kubevirt crd, We create a specialized storageclass for it.
```
aglitke, line 439: I am not familiar with this operator but is this function called from within a reconcile loop?  In the case where kubevirt is installed after this operator will the ceph-rbd-virtualization SC be created?

aglitke: We typically recommend that storage be configured before installing kubevirt so that a default storage class can be set.

malayparida2000: We have added a watch on the kubevirt crd in our controller. So if at any point the kubevirt crd comes up the controller will fire up a reconcile & the storageclass will be immediately created

controllers/storagecluster/storageclasses.go
```diff
@@ -280,6 +281,46 @@ func newCephBlockPoolStorageClassConfiguration(initData *ocsv1.StorageCluster) S
 	}
 }
 
+// newCephBlockPoolVirtualizationStorageClassConfiguration generates configuration options for a Ceph Block Pool StorageClass for virtualization environment.
```
idryomov, line 284: This comment needs updating to match the method name (and the blank line below seems redundant too).

idryomov (approved):
Ack from the RBD perspective

controllers/storagecluster/storageclasses.go
malayparida2000: Done

controllers/storagecluster/storageclasses.go
malayparida2000: Addresed

malayparida2000:
@nixpanic PR is now ready from all perspectives. Please take a final look now.

malayparida2000:
/cc @Madhu-1 
Please review

nixpanic (approved):

controllers/storagecluster/storageclasses.go
```diff
@@ -280,6 +281,45 @@ func newCephBlockPoolStorageClassConfiguration(initData *ocsv1.StorageCluster) S
...
+func newCephBlockPoolVirtualizationStorageClassConfiguration(initData *ocsv1.StorageCluster) StorageClassConfiguration {
```
Madhu-1, line 285: the content of this function is exact copy of `newCephBlockPoolStorageClassConfiguration` can this function call `newCephBlockPoolStorageClassConfiguration` and update the StorageClassConfiguration.storageClass.Parameters to add the new map option?

malayparida2000: Right, We are changing just a different Name, Annotation & 2 extra Parameters. So this seems like a good way for simplification.

Madhu-1 (approved):
LGTM

iamniting (approved):

PR #2078, opened 2023-06-12
storagecluster: set allowVolumeExpansion to true for rbd enc sc

Rakshith-R:
CephCSI supports expansion for rbd encrypted PVCs. This pr sets allowVolumeExpansion to true in
the default encrypted rbd storageclass that
ocs-operator creates.

resolves: https://issues.redhat.com/browse/RHSTOR-4650

/cc @iamniting @agarwal-mudit @malayparida2000 @Madhu-1 @nixpanic


nixpanic:
Does this cause an update of the StorageClass(es) too?

iamniting:
> Does this cause an update of the StorageClass(es) too?

Yes it should do, We are taking care of that for all storage classes [here](https://github.com/red-hat-storage/ocs-operator/blob/main/controllers/storagecluster/storageclasses.go#L188-L203).

iamniting (approved):

malayparida2000:
> > Does this cause an update of the StorageClass(es) too?
> 
> Yes it should do, We are taking care of that for all storage classes [here](https://github.com/red-hat-storage/ocs-operator/blob/main/controllers/storagecluster/storageclasses.go#L188-L203).

@nixpanic I didn't notice we had this in place to update the parameters. So this should be useful for us too, in case we need to update our virtualization storageclass later ref-[RHSTOR-4521](https://issues.redhat.com/browse/RHSTOR-4521)

PR #2081, opened 2023-06-20
Fix alerts not firing issue

aruniiird:
We have new changes regarding ceph metrics,

Addressing the following changes

1. Format change in metadata metrics now has a new 'ceph_daemon' value format eg: 'metric-name': ceph_mon_metadata' OLD-value: 'ceph-mon.a' NEW-value: 'mon.a'

2. A new 'job', 'rook-ceph-exporter', is introduced All the previous metrics has job value 'rook-ceph-mgr' But now some metrics has a new job value 'rook-ceph-exporter'

Why we are adding/duplicating the same alert?
For the backward compatibility: since the new queries have the same name, we can rest assured that either one of them will be fired (according to where or to which version) this file is deployed. Another reason is for easy backports: as the original alert lines where not altered/changed (only addition happened), this file will be easy to backport to older versions.


avanthakkar (commented):
I'll suggest to not add any changes related to `ceph_daemon` label, as there's been a regression detected for the same for `ceph-exporter` and the fix is WIP https://github.com/ceph/ceph/pull/52084

umangachapagain (commented):
If there is a new job label, shouldn't we use that instead of dropping it completely?

aruniiird:
> If there is a new job label, shouldn't we use that instead of dropping it completely?

We don't have to use the new job label, this was a convention used earlier (and that is the reason why we have to do this change). If we drop the job label it should pick the metric from the new job, which now produces the metric.
The situation would have been different if both the jobs produced the same metric, but here in this case all the perf-metrics where shifted/moved from previous job to the new job.

controllers/storagecluster/prometheus/localcephrules.yaml
```diff
@@ -11,10 +11,10 @@ spec:
   - name: ceph.rules
     rules:
     - expr: |
-        kube_node_status_condition{condition="Ready",job="kube-state-metrics",status="true"} * on (node) group_right() max by (node, namespace) (label_replace(ceph_disk_occupation{job="rook-ceph-mgr"},"node","$1","exported_instance","(.*)"))
+        kube_node_status_condition{condition="Ready",job="kube-state-metrics",status="true"} * on (node) group_right() max by (node, namespace) (label_replace(ceph_disk_occupation,"node","$1","exported_instance","(.*)"))
```
umangachapagain, line 14:
```suggestion
        kube_node_status_condition{condition="Ready",job="kube-state-metrics",status="true"} * on (node) group_right() max by (node, namespace) (label_replace(ceph_disk_occupation{job="rook-ceph-exporter"},"node","$1","exported_instance","(.*)"))
```
This is better than completely dropping the job. Let's use it in all places.
We want to add a job filter as we don't want to consider similar metrics from other sources.

umangachapagain (changes requested):

controllers/storagecluster/prometheus/localcephrules.yaml
aruniiird: Made the changes in all the places

umangachapagain (approved):

PR #2083, opened 2023-06-22
Remove quota while creating cephblockpool and remove capacity field from StorageConsumer CR as it is no longer used

rchikatw:
Removing the capacity field in StorageConsumer CR as this is causing problems setting quota and we are not using the capacity feature any more in managed service.


rchikatw:
/ golangci-lint

subhamkrai (commented):
overall LGTM

sp98:
@rchikatw I'm not aware of any discussions that have happened around this. But looking at the BZ https://bugzilla.redhat.com/show_bug.cgi?id=2212333, the issue seems to be default capacity of 1TiB, which is not useful when the cluster has higher capacity. 

was there any offline discussion to completely remove the `capacity`?

rchikatw:
> @rchikatw I'm not aware of any discussions that might have happened around this. But looking at the BZ https://bugzilla.redhat.com/show_bug.cgi?id=2212333, the issue seems to be default capacity of 1TiB, which is not useful when the cluster has higher capacity.
> 
> was there any offline discussion to completely remove the `capacity`?

Yes I had a discussion with @nb-ohad regarding this we have decided to remove this completely

api/v1alpha1/storageconsumer_types.go
```diff
@@ -39,9 +38,6 @@ const (
 
 // StorageConsumerSpec defines the desired state of StorageConsumer
 type StorageConsumerSpec struct {
-	// Capacity is the total quota size allocated to a consumer.
-	Capacity resource.Quantity `json:"capacity"`
```
Madhu-1, line 43: Are we going to have any problem with the upgrades if we remove an entry from CRD? @iamniting can you please confirm?

Madhu-1 (commented):
LGTM, can you please also test the upgrade to ensure we will not have any problems with the upgrade?

api/v1alpha1/storageconsumer_types.go
nb-ohad: I believe you will have an issue if any CR is using it. Usually, a breaking change needs to come with a version update where the k8s engine will translate from the prev version to the next one. 
But, as StorageConsumer is an internal API, and as we do not expect any upgrade from an older version, I think we are ok.

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -339,9 +337,6 @@ func (r *StorageClassRequestReconciler) reconcileCephBlockPool() error {
 					ReplicasPerFailureDomain: 1,
 				},
 				Parameters: r.storageProfile.BlockPoolConfiguration.Parameters,
-				Quotas: rookCephv1.QuotaSpec{
-					MaxSize: &capacity,
-				},
```
nb-ohad, line 344: I don't believe that will handle the removal of existing quotas sections. The reason is that this is internal to CreateOrUpdate which uses merge semantics.

nb-ohad (changes requested):

api/v1alpha1/storageconsumer_types.go
rchikatw: Yes we are ok with this change.

controllers/storageclassrequest/storageclassrequest_controller.go
rchikatw: I have verified this, It's working as expected.

Before Upgrade quota:
...
  quotas:
    maxSize: 1T
...
After upgrade quota:
....
quotas: {}
...

And I have also seen a change in the generation field inside the metadata section,  Changed from 1 -> 2

api/v1alpha1/storageconsumer_types.go
rchikatw: @Madhu-1 I have verified the upgrade path as well and it's working as expected.

controllers/storageclassrequest/storageclassrequest_controller.go
rchikatw: @nb-ohad ^^

services/provider/client/client.go
```diff
@@ -54,15 +54,14 @@ func (cc *OCSProviderClient) Close() {
 
 // OnboardConsumer to validate the consumer and create StorageConsumer
 // resource on the StorageProvider cluster
-func (cc *OCSProviderClient) OnboardConsumer(ctx context.Context, ticket, name, capacity string) (*pb.OnboardConsumerResponse, error) {
+func (cc *OCSProviderClient) OnboardConsumer(ctx context.Context, ticket, name string) (*pb.OnboardConsumerResponse, error) {
```
iamniting, line 57: you need to keep in mind that you will need to make this change in the `ocs-client-operator` as well while making the `OnboardConsumer` request.

iamniting (changes requested):

deploy/ics-operator/manifests/storageconsumer.crd.yaml
```diff
@@ -0,0 +1,80 @@
+---
```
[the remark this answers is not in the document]
rchikatw: Okay.

deploy/csv-templates/crds/ics/ocs.openshift.io_storageconsumers.yaml
```diff
@@ -0,0 +1,81 @@
+---
```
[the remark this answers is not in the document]
rchikatw: Okay.

iamniting (approved):
looks good to me. Holding it for others review.

/hold

controllers/storageclassrequest/storageclassrequest_controller.go
nb-ohad: @rchikatw  if the actual behavior is what you expected, I am not going to object

PR #2086, opened 2023-06-23
update the metric for rgw_health_status

weirdwiz:
Include more conditions in RGW health status, and update the alerts based on the changes

### DIFF
--- a/metrics/internal/collectors/ceph-object-store.go
+++ b/metrics/internal/collectors/ceph-object-store.go
@@ -115,6 +115,30 @@ func (c *CephObjectStoreCollector) collectObjectStoreHealth(cephObjectStores []*
 				cephObjectStore.Name,
 				cephObjectStore.Namespace,
 				cephObjectStore.Status.Info["endpoint"])
+		case cephv1.ConditionConnecting:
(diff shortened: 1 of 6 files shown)


weirdwiz:
oh looks like i pushed a dirty branch
fixing

weirdwiz:
ptal @umangachapagain

metrics/internal/collectors/ceph-object-store.go
```diff
@@ -97,24 +97,48 @@ func getAllObjectStores(lister cephv1listers.CephObjectStoreLister, namespaces [
 func (c *CephObjectStoreCollector) collectObjectStoreHealth(cephObjectStores []*cephv1.CephObjectStore, ch chan<- prometheus.Metric) {
 	for _, cephObjectStore := range cephObjectStores {
 		switch cephObjectStore.Status.Phase {
-		case cephv1.ConditionConnected:
+		case cephv1.ConditionConnecting:
```
umangachapagain, line 100: I don't think we should be changing existing values.

umangachapagain (approved):

PR #2091, opened 2023-07-03
noobaa autoscaler changes, new properties added

naveenpaul1:
**Explain the changes**
Integrate noobaa with Prometheus adaptor for custom metrics autoscaling
This PR will make HPAV2 the default HPA for noobaa installed through OCS,
Autoscaler properties `AutoscalerType` and `PrometheusNamespace` added to Noobaa CR with value `hpav2` and hardcoded Prometheus namespace in openshift; `openshift-monitoring`

**Issues: Fixed #xxx / Gap #xxx**
https://bugzilla.redhat.com/show_bug.cgi?id=2217904

**Testing Instructions:**
Install ODF in cluster
check for the newly created HPA2 autoscaler
update noobaa endpoint min and max pod count in Noobaa CR
Verify new endpoint pod is created/deleted


controllers/storagecluster/noobaa_system_reconciler.go
```diff
@@ -21,6 +21,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
+const (
+	MonitoringNamespace = "openshift-monitoring"
```
iamniting, line 25: @umangachapagain Is it always valid to use this namespace in the core product? I recall that Managed Services prefers to avoid using the default monitoring and wants to have its own. They have the flexibility to deploy it anywhere they want, which is not an issue for us currently because they are not using noobaa. However, if they decide to use it in the future, we will need to modify this. Or shall we do something future-proof now itself?

umangachapagain: It's ok to add it later. PR looks good to me for current requirements.

iamniting (approved):

PR #2097, opened 2023-07-05
Improvements to file based catalog generation process

malayparida2000:
Update opm version to v1.28.0 from v1.26.0
Generate & update the Dockerfile for the catalog image via opm each time Don't keep the rendered config files in the repo
Update noobaa bundle channel entry version in index.yaml


hack/build-operator-catalog.sh
```diff
@@ -13,7 +13,11 @@ echo
 ${OPM} render --output=yaml ${BUNDLE_FULL_IMAGE_NAME} > catalog/ocs-bundle.yaml
 ${OPM} render --output=yaml ${NOOBAA_BUNDLE_FULL_IMAGE_NAME} > catalog/noobaa-bundle.yaml
 ${OPM} validate catalog
-${IMAGE_BUILD_CMD} build -t ${FILE_BASED_CATALOG_FULL_IMAGE_NAME} -f Dockerfile.catalog .
+${OPM} generate dockerfile catalog
 
+mv catalog.Dockerfile Dockerfile.catalog
+${IMAGE_BUILD_CMD} build --no-cache -t ${FILE_BASED_CATALOG_FULL_IMAGE_NAME} -f Dockerfile.catalog .
+
+rm catalog/ocs-bundle.yaml catalog/noobaa-bundle.yaml
```
iamniting, line 21: Let's not remove these files, let them be there on the local machine, You can add them in the git ignore.

iamniting (changes requested):

PR #2099, opened 2023-07-07
Remove unused & unnecessary Dockerfiles, scripts, make targets etc

malayparida2000:
After recent cleanup of openshift-ci & improvements in the build process, we have a lot of unused & unnecessary files left in the repo. This commit removes all of them. The


iamniting (changes requested):

iamniting (approved):

b-ranto:
Hey, could you be more specific about the improvements in the build process? This is breaking the downstream build as we were using the gen-release-csv make target that was removed here.

malayparida2000:
> Hey, could you be more specific about the improvements in the build process? This is breaking the downstream build as we were using the gen-release-csv make target that was removed here.

In upstream we had a mess of dockerfiles in our repo in different places, for different purposes. So we cleaned them up. Also earlier the binaries were being built locally in the host & then being copied to a container image. but now the binaries are also being built in a containerized form. All this is already being done by ds build. 

We were not aware that the gen-release-csv command was being used by the ds build process, so we mistakenly removed it. Have raised a PR to add it back here https://github.com/red-hat-storage/ocs-operator/pull/2107.

PR #2105, opened 2023-07-12
Use new module path for ocs in ldflags

malayparida2000:
Recently we updated the moduele path for ocs to v4 so we need to update the ldflags accordingly. Also ldflags is now a variable in the hack/common.sh file so we need not define it in every script.

### DIFF
--- a/hack/common.sh
+++ b/hack/common.sh
@@ -17,6 +17,7 @@ GO_LINT_IMG="${GO_LINT_IMG:-${GO_LINT_IMG_LOCATION}:${GO_LINT_IMG_TAG}}"
 DEFAULT_CSV_VERSION="4.13.0"
 CSV_VERSION="${CSV_VERSION:-${DEFAULT_CSV_VERSION}}"
 VERSION="${VERSION:-${CSV_VERSION}}"
+LDFLAGS="-X github.com/red-hat-storage/ocs-operator/v4/version.Version=${CSV_VERSION}"
 
 OUTDIR="build/_output"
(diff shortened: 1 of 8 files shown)


hack/cluster-deploy.sh
```diff
@@ -4,7 +4,7 @@ set -e
 
 source hack/common.sh
 
-(cd tools/cluster-deploy/ && go build)
+(cd tools/cluster-deploy/ && go build --ldflags "${LDFLAGS}")
```
iamniting, line 7: I don't think we need this.

hack/common.sh
```diff
@@ -17,6 +17,7 @@ GO_LINT_IMG="${GO_LINT_IMG:-${GO_LINT_IMG_LOCATION}:${GO_LINT_IMG_TAG}}"
 DEFAULT_CSV_VERSION="4.13.0"
 CSV_VERSION="${CSV_VERSION:-${DEFAULT_CSV_VERSION}}"
 VERSION="${VERSION:-${CSV_VERSION}}"
+LDFLAGS="-X github.com/red-hat-storage/ocs-operator/v4/version.Version=$VERSION"
```
iamniting, line 20:
```suggestion
LDFLAGS="-X github.com/red-hat-storage/ocs-operator/v4/version.Version=${CSV_VERSION}"
```

iamniting (changes requested):

iamniting (approved):

iamniting (approved):

PR #2118, opened 2023-07-31
Move autoscale into DesiredState method for upgrade

naveenpaul1:
### Explain the changes
1. make sure noobaa autoscale configuration will added/updated to noobaa CR if noobaa CR missing the configuration

### Issues: Fixed #xxx / Gap #xxx
1. After upgrading to 4.14, MCG endpoint minCount and maxCount in ocs-storagecluster are ignored,
since 4.13 noobaa CR missing autoscaling configuration, after upgrade this configuration was still missing and this created issue

https://bugzilla.redhat.com/show_bug.cgi?id=2227607

### Testing Instructions:

1. install ODF in openshift 4.13 and upgrade it to 4.14
2. update noobaa endpoint maxCount to 2, wait for the second noobaa endpoint pod

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/controllers/storagecluster/noobaa_system_reconciler.go
+++ b/controllers/storagecluster/noobaa_system_reconciler.go
@@ -78,10 +78,6 @@ func (obj *ocsNoobaaSystem) ensureCreated(r *StorageClusterReconciler, sc *ocsv1
 			Labels: nbv1.LabelsSpec{
 				"monitoring": getNooBaaMonitoringLabels(*sc),
 			},
-			Autoscaler: nbv1.AutoscalerSpec{
-				AutoscalerType:      nbv1.AutoscalerTypeHPAV2,
-				PrometheusNamespace: MonitoringNamespace,
-			},
 		},
 	}
 	err = controllerutil.SetControllerReference(sc, nb, r.Scheme)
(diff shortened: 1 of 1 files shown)


controllers/storagecluster/noobaa_system_reconciler.go
```diff
@@ -169,6 +165,13 @@ func (r *StorageClusterReconciler) setNooBaaDesiredState(nb *nbv1.NooBaa, sc *oc
 		// GetDaemonResources()
 		Resources: &endpointResources,
 	}
+	// Add autoscale type for noobaa CR
+	if nb.Spec.Autoscaler.AutoscalerType == "" {
```
iamniting, line 169: Do you want it to change if and only if it is empty? what if it is `keda`? Earlier it looks like we always wanted to have it to `hpav2`.

iamniting (changes requested):

iamniting:
Can you also correct the typo in the commit msg

`move autoscal into DesiredState method for upgrade` -> `move autoscale into DesiredState method for upgrade`

controllers/storagecluster/noobaa_system_reconciler.go
naveenpaul1: Hi @iamniting It can never be `Keda` unless the user updates it manually. Since we are hardcoding `hpav2` AutoscalerType always will have that value. Previous code also not restricting users from updating the AutoscalerType with the value keda. Correct me if am wrong.

iamniting: We have several reconciliation strategies, and one of them involves ignoring certain changes made by the user. In this case, users can update the values manually, and we won't revert them back to the original state. However, if the user decides to switch back to a managed state, we will reconcile the values to their original state. Additionally, if the user changed a value to 'keda' while it was in the ignored state, we are currently not reverting it back, but in my opinion, we should do so.

Can you pls check with Danny or someone if we ever supported changing this value or can it be `keda` in any case?

naveenpaul1: @iamniting, With this release, our plan is to go only with HPAV2 for our side. So I think it's better to avoid the keda in any case. I will update the code.

iamniting (approved):

PR #2122, opened 2023-08-05
Delete completed template job so no need to delete it when rerunning the job

malayparida2000:
Resolves-https://bugzilla.redhat.com/show_bug.cgi?id=1970939

This eliminates the need to delete the older/existing job when
 triggering the job for a second time.

### DIFF
--- a/controllers/storagecluster/job_templates.go
+++ b/controllers/storagecluster/job_templates.go
@@ -12,6 +12,8 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
@@ -56,6 +58,21 @@ func (obj *ocsJobTemplates) ensureCreated(r *StorageClusterReconciler, sc *ocsv1
 
 		if err != nil && !errors.IsAlreadyExists(err) {
 			return reconcile.Result{}, fmt.Errorf("failed to create Template : %v", err.Error())
+		}
+
+		// Delete a template job after the job gets completed
+		// This eliminates the need to delete the older/existing job when triggering the job for a second time
+		job := &batchv1.Job{}
+		err = r.Client.Get(r.ctx, client.ObjectKey{Name: fmt.Sprintf("%s-job", template.Name), Namespace: template.Namespace}, job)
+		if err == nil {
+			if job.Status.Succeeded >= 1 {
+				err = r.Client.Delete(r.ctx, job)
+				if err != nil && !errors.IsNotFound(err) {
+					r.Log.Error(err, "Job couldn't be deleted successfully after completion", "Job", klog.KRef(job.Namespace, job.Name))
+				} else {
+					r.Log.Info("Job deleted successfully after completion", "Job", klog.KRef(job.Namespace, job.Name))
+				}
+			}
 		}
 	}
 


travisn (commented):
/lgtm

controllers/storagecluster/job_templates.go
```diff
@@ -57,6 +58,21 @@ func (obj *ocsJobTemplates) ensureCreated(r *StorageClusterReconciler, sc *ocsv1
...
+		// Delete a template job after the job gets completed
+		// This eliminates the need to delete the older/existing job when triggering the job for a second time
+		job := &batchv1.Job{}
+		err = r.Client.Get(r.ctx, client.ObjectKey{Name: fmt.Sprintf("%s-job", template.Name), Namespace: template.Namespace}, job)
+		if err == nil {
+			if job.Status.Succeeded >= 1 {
+				r.Log.Info(fmt.Sprintf("Job %s is completed", job.Name))
+				err = r.Client.Delete(r.ctx, job)
+				if err != nil {
+					return reconcile.Result{}, fmt.Errorf("failed to delete job %s: %v", template.Name, err.Error())
```
umangachapagain, line 71: If we can't delete the job, should we block rest of the reconcile?
IMO we should try to delete but not return an error.

controllers/storagecluster/job_templates.go
```diff
@@ -57,6 +58,21 @@ func (obj *ocsJobTemplates) ensureCreated(r *StorageClusterReconciler, sc *ocsv1
...
 			return reconcile.Result{}, fmt.Errorf("failed to create Template : %v", err.Error())
 		}
+
+		// Delete a template job after the job gets completed
+		// This eliminates the need to delete the older/existing job when triggering the job for a second time
+		job := &batchv1.Job{}
+		err = r.Client.Get(r.ctx, client.ObjectKey{Name: fmt.Sprintf("%s-job", template.Name), Namespace: template.Namespace}, job)
+		if err == nil {
+			if job.Status.Succeeded >= 1 {
+				r.Log.Info(fmt.Sprintf("Job %s is completed", job.Name))
```
umangachapagain, line 68: This isn't a very helpful message. Let's drop it and log it after we delete it.

controllers/storagecluster/job_templates.go
```diff
@@ -57,6 +58,21 @@ func (obj *ocsJobTemplates) ensureCreated(r *StorageClusterReconciler, sc *ocsv1
...
+				err = r.Client.Delete(r.ctx, job)
+				if err != nil {
+					return reconcile.Result{}, fmt.Errorf("failed to delete job %s: %v", template.Name, err.Error())
+				}
+				r.Log.Info(fmt.Sprintf("Job %s is deleted after completion", job.Name))
```
umangachapagain, line 73: Needs more detail. What is this job related to?
Also it's better to avoid `sprintf` and use structured logging format as used in other places.

malayparida2000: We have job templates, Which can be invoked with oc process command to create jobs to complete some tasks.
The job names will give the context itself. We have two templates ocs-osd-removal & ocs-extend-cluster.

This is how the logs will look like
```
~ $ oc logs ocs-operator-8676bdf8bb-hgr9q | grep "deleted successfully"
{"level":"info","ts":"2023-08-10T10:30:09Z","logger":"controllers.StorageCluster","msg":"Job ocs-osd-removal-job deleted successfully after completion","Request.Namespace":"openshift-storage","Request.Name":"ocs-storagecluster"}
```

controllers/storagecluster/job_templates.go
```diff
@@ -57,6 +58,20 @@ func (obj *ocsJobTemplates) ensureCreated(r *StorageClusterReconciler, sc *ocsv1
 		if err != nil && !errors.IsAlreadyExists(err) {
 			return reconcile.Result{}, fmt.Errorf("failed to create Template : %v", err.Error())
 		}
+
+		// Delete a template job after the job gets completed
```
Madhu-1, line 62: the order of sequence should be like below right?

```
job:=Get()
if err!=nil && !NotPresent{
return nil,err
}
if error==nil && job.Status.Succeeded >= 1{
// delete it and return
} 
err=CreateOrUpdate()
return err
```

malayparida2000: Hi Madhu, actually no. While creating we create the openshift templatest only. When a user wants they create the job by using the oc process command. 

```
oc process -n openshift-storage ocs-osd-removal -p FAILED_OSD_IDS=0 FORCE_OSD_REMOVAL=true |oc create -n openshift-storage -f -
```

Madhu-1: Got it, Thank you

controllers/storagecluster/job_templates.go
```diff
@@ -57,6 +58,20 @@ func (obj *ocsJobTemplates) ensureCreated(r *StorageClusterReconciler, sc *ocsv1
...
+		// This eliminates the need to delete the older/existing job when triggering the job for a second time
+		job := &batchv1.Job{}
+		err = r.Client.Get(r.ctx, client.ObjectKey{Name: fmt.Sprintf("%s-job", template.Name), Namespace: template.Namespace}, job)
+		if err == nil {
+			if job.Status.Succeeded >= 1 {
+				err = r.Client.Delete(r.ctx, job)
+				if err != nil {
+					r.Log.Info("Job " + job.Name + " couldn't be deleted successfully after completion")
+				}
```
travisn, line 71: Line 72 will be mistakenly logged if there is an error
```suggestion
				} else {
```

travisn (changes requested):

controllers/storagecluster/job_templates.go
```diff
@@ -57,6 +58,20 @@ func (obj *ocsJobTemplates) ensureCreated(r *StorageClusterReconciler, sc *ocsv1
...
+
+		// Delete a template job after the job gets completed
+		// This eliminates the need to delete the older/existing job when triggering the job for a second time
+		job := &batchv1.Job{}
+		err = r.Client.Get(r.ctx, client.ObjectKey{Name: fmt.Sprintf("%s-job", template.Name), Namespace: template.Namespace}, job)
+		if err == nil {
+			if job.Status.Succeeded >= 1 {
+				err = r.Client.Delete(r.ctx, job)
+				if err != nil {
+					r.Log.Info("Job " + job.Name + " couldn't be deleted successfully after completion")
```
umangachapagain, line 70:
```suggestion
					r.Log.Info("Job couldn't be deleted successfully after completion.", "Job", klog.KRef(job.Namespace, job.Name))
```

controllers/storagecluster/job_templates.go
```diff
@@ -57,6 +58,20 @@ func (obj *ocsJobTemplates) ensureCreated(r *StorageClusterReconciler, sc *ocsv1
...
+				}
+				r.Log.Info("Job " + job.Name + " deleted successfully after completion")
```
umangachapagain, line 72:
```suggestion
				r.Log.Info("Job deleted successfully after completion.", "Job", klog.KRef(job.Namespace, job.Name))
```

umangachapagain (changes requested):

controllers/storagecluster/job_templates.go
```diff
@@ -57,6 +59,21 @@ func (obj *ocsJobTemplates) ensureCreated(r *StorageClusterReconciler, sc *ocsv1
...
 			return reconcile.Result{}, fmt.Errorf("failed to create Template : %v", err.Error())
 		}
+
+		// Delete a template job after the job gets completed
+		// This eliminates the need to delete the older/existing job when triggering the job for a second time
+		job := &batchv1.Job{}
+		err = r.Client.Get(r.ctx, client.ObjectKey{Name: fmt.Sprintf("%s-job", template.Name), Namespace: template.Namespace}, job)
+		if err == nil {
+			if job.Status.Succeeded >= 1 {
+				err = r.Client.Delete(r.ctx, job)
```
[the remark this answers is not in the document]
malayparida2000: Yes,
https://github.com/red-hat-storage/ocs-operator/blob/58c0a5630cc2511a02f533516bc94cc85ee3fc92/config/rbac/role.yaml#L29-L38

https://github.com/red-hat-storage/ocs-operator/blob/58c0a5630cc2511a02f533516bc94cc85ee3fc92/controllers/storagecluster/reconcile.go#L128

umangachapagain (approved):

PR #2123, opened 2023-08-06
Avoid wrong detection of disk type, such as HDD instead of SSD on vSAN

malayparida2000:
Resolves-https://bugzilla.redhat.com/show_bug.cgi?id=2218116

In some situations, such as vSphere VSAN, the drive media type is detected wrongly, for instance as HDD instead of SSD. For that, we have to force BlueStore to use settings designed for SSDs. We have 2 options for this.

1) bluestore_debug_enforce_settings = "ssd"
but it requires restart

OR

2) bluestore_prefer_deferred_size_hdd = 0
should work right away and no new deferred writes will be enqueued

We are choosing the second option as it does not require osd restart.

### DIFF
--- a/controllers/storagecluster/cephconfig.go
+++ b/controllers/storagecluster/cephconfig.go
@@ -36,6 +36,7 @@ mon_osd_nearfull_ratio = .75
 mon_max_pg_per_osd = 600
 mon_pg_warn_max_object_skew = 0
 mon_data_avail_warn = 15
+bluestore_prefer_deferred_size_hdd = 0
 [osd]
 osd_memory_target_cgroup_limit_ratio = 0.8
 `


controllers/storagecluster/cephconfig.go
```diff
@@ -36,6 +36,7 @@ mon_osd_nearfull_ratio = .75
 mon_max_pg_per_osd = 600
 mon_pg_warn_max_object_skew = 0
 mon_data_avail_warn = 15
+bluestore_prefer_deferred_size_hdd = 0
```
sp98, line 39: Looks like this change will affect all customers whether DR or not.  Is that the expectation?

travisn: Also, since #2053 is creating all the OSDs as SSDs, do we still need this? I'm not clear on the effect of this setting.

agarwal-mudit:
@sp98 @malayparida2000 if https://github.com/red-hat-storage/ocs-operator/pull/2053 enabled the option to create all OSD as SSDs then can be close this PR (and bug)?

sp98:
> @sp98 @malayparida2000 if #2053 enabled the option to create all OSD as SSDs then can be close this PR (and bug)?

#2053 is only for new OSDs.

malayparida2000:
> > @sp98 @malayparida2000 if #2053 enabled the option to create all OSD as SSDs then can be close this PR (and bug)?
> 
> #2053 is only for new OSDs. So based on Adams response :
> 
> ` If device class is SSD then deferred mechanism is turned off.(except a mandatory corner cases)`
> 
> We can still keep these changes for existing OSDs.

Thanks, Santosh for the clarification. I think then we can go ahead & merge these changes soon

iamniting (approved):

PR #2128, opened 2023-08-10
Add multus related annotations to ceph toolbox

Nikhil-Ladha:
Add an annotation to ceph toolbox when multus network provider is set.
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1982721

### DIFF
--- a/controllers/storagecluster/cephtoolbox.go
+++ b/controllers/storagecluster/cephtoolbox.go
@@ -2,10 +2,12 @@ package storagecluster
 
 import (
 	"context"
+	"fmt"
 	"reflect"
 
 	ocsv1 "github.com/red-hat-storage/ocs-operator/v4/api/v1"
 	"github.com/red-hat-storage/ocs-operator/v4/controllers/defaults"
+	"github.com/red-hat-storage/ocs-operator/v4/controllers/util"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -72,6 +74,21 @@ func (r *StorageClusterReconciler) ensureToolsDeployment(sc *ocsv1.StorageCluste
 			return err
 		}
 
+		// Add multus related annotations, if multus provider is set
+		if isMultus(sc.Spec.Network) {
+			net, err := getMultusPublicNetwork(sc)
+			if err != nil {
+				return err
+			}
+			if net != "" {
+				if toolsDeployment.Spec.Template.ObjectMeta.Annotations == nil {
+					toolsDeployment.Spec.Template.ObjectMeta.Annotations = make(map[string]string)
+				}
+				toolsDeployment.Spec.Template.ObjectMeta.Annotations["k8s.v1.cni.cncf.io/networks"] = net
+				toolsDeployment.Spec.Template.Spec.HostNetwork = false
+			}
+		}
+
 		if !isFound {
 			return r.Client.Create(context.TODO(), toolsDeployment)
 		} else if !reflect.DeepEqual(foundToolsDeployment.Spec, toolsDeployment.Spec) {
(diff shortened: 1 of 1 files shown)


Nikhil-Ladha:
@umangachapagain @iamniting can you please take a look

iamniting (changes requested):

controllers/storagecluster/cephtoolbox.go
```diff
@@ -72,6 +74,29 @@ func (r *StorageClusterReconciler) ensureToolsDeployment(sc *ocsv1.StorageCluste
...
+			multusNetNs, err := util.GetOperatorNamespace()
+			if err != nil {
+				return err
+			}
+			multusNetName := ""
+			if sc.Spec.Network.Selectors["public"] != "" {
+				multusNetName = sc.Spec.Network.Selectors["public"]
+			} else if sc.Spec.Network.Selectors["cluster"] != "" {
+				multusNetName = sc.Spec.Network.Selectors["cluster"]
+			}
```
umangachapagain, line 93: Can't we have both? Is it always either "public" or "cluster"?

Nikhil-Ladha: TBH, I am not sure. I was trying to figure out the annotation required and stumbled across this [piece](https://github.com/red-hat-storage/ocs-ci/blob/master/ocs_ci/ocs/utils.py#L839) of code in ocs-ci, and just replicated that here.
Let me ask around someone from rook to confirm this.

controllers/storagecluster/cephtoolbox.go
```diff
@@ -72,6 +74,29 @@ func (r *StorageClusterReconciler) ensureToolsDeployment(sc *ocsv1.StorageCluste
...
+			toolsDeployment.Spec.Template.ObjectMeta.Annotations = map[string]string{
```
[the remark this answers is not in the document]
Nikhil-Ladha: Ack, updated the code.

controllers/storagecluster/cephtoolbox.go
BlaineEXE: Multus can be public, cluster, or both.

BlaineEXE: I would probably suggest a new helper method like this:

```go
func getMultusPublicNetwork(/* ... */) (string, error) {
   err := validateMultusSelectors(sc.Spec.Network.Selectors)
			if err != nil {
				return "", err
			}

			multusNetNs, err := util.GetOperatorNamespace()
			if err != nil {
				return "", err
			}
			multusNetName := ""
			if sc.Spec.Network.Selectors["public"] == "" {
				return "", nil
			}
			if sc.Spec.Network.Selectors["public"] != "" {
				multusNetName = sc.Spec.Network.Selectors["public"]
			}
			return fmt.Sprintf("%s/%s", multusNetNs, multusNetName), nil

}


# ...


net, err := getMultusPublicNetwork(/* ... */)
if err != nil {
	return err
}
if net != "" {
	# TODO: apply annotation to pod
}
```

Nikhil-Ladha: Thanks for the review, and the helper method. Updated the code accordingly.

controllers/storagecluster/cephtoolbox.go
```diff
@@ -99,3 +116,23 @@ func removeDuplicateTolerations(tolerations []corev1.Toleration) []corev1.Tolera
...
+		return "", err
+	}
+	multusNetName := ""
+	if sc.Spec.Network.Selectors["public"] == "" {
+		return "", nil
+	}
+	if sc.Spec.Network.Selectors["public"] != "" {
+		multusNetName = sc.Spec.Network.Selectors["public"]
+	}
+	return fmt.Sprintf("%s/%s", multusNetNs, multusNetName), nil
```
iamniting, line 137:
```suggestion
	if sc.Spec.Network.Selectors["public"] != "" {
		multusNetName = sc.Spec.Network.Selectors["public"]
		return fmt.Sprintf("%s/%s", multusNetNs, multusNetName), nil
	}
		
	return "", nil
```

Nikhil-Ladha: Makes sense, updated.

iamniting (approved):

controllers/storagecluster/cephtoolbox.go
```diff
@@ -99,3 +116,22 @@ func removeDuplicateTolerations(tolerations []corev1.Toleration) []corev1.Tolera
 	}
 	return list
 }
+
+func getMultusPublicNetwork(sc *ocsv1.StorageCluster) (string, error) {
+	err := validateMultusSelectors(sc.Spec.Network.Selectors)
+	if err != nil {
+		return "", err
```
BlaineEXE, line 123: It seems like maybe this is a reflection of ocs-operator's codebase, but I'm curious why most errors are just returned without any additional info (e.g., using %w to wrap errors). Does the ocs-operator team find that this makes debugging failures more difficult?

iamniting: Hello Blaine, I wanted to mention that we aren't currently mandating contributors to encapsulate a message or include additional logs. I believe this is something we should consider implementing in the future as it could enhance the ease of debugging.

PR #2130, opened 2023-08-10
design doc to use optimized OSD store

sp98:
This design doc contains information on how OCS operator can be used to update OSD store for new  ODF customers using Regional Disaster Recovery

### DIFF
--- a/docs/design/disaster-recovery-optimization.md
+++ b/docs/design/disaster-recovery-optimization.md
@@ -1,0 +1,33 @@
+# FEATURE: Regional Disaster Recovery Optimization
+
+## Summary
+
+Implement use of `bluestore-rdr` as the OSD backend store for customers using ODF with Regional Disaster Recovery (RDR).
+
+## Motivation
+
+- Ceph uses BlueStore as a special-purpose storage back end designed specifically for managing data on disk for Ceph OSD workloads. Ceph team is working on improving this backend storage that will help in significantly improving the performance, especially in disaster recovery scenarios. This new OSD backend is known as bluestore-rdr. In the future, other backends will also need to be supported, such as seastore.
+
+## Goals
+
+- Customers should be able to use the optimized OSD store when creating a new ODF cluster
+  
+## Non Goals/Deferred Goals
+
+- Customers should be able to update their existing ODF clusters to use the optimized OSD store
+
+## Proposal
+
+- Enable the optimized OSD backend store for new RDR clusters.
+
+
+### Greenfield
+- If `ocs.openshift.io/clusterIsDisasterRecoveryTarget: true` annotation is available in the `StorageCluster` then configure OSD store type in the `cephCluster` as:
+
+```YAML
+spec:
+  storage:
+    store:
+        type: bluestore-rdr
+```
+- Customers should not be able to remove the optimizations once they are applied. If the annotation is removed `storageCluster` some how, operator should ensure that OSD store optimization settings are not overridden in the `cephCluster` CR


docs/design/disaster-recovery-optimization.md
```diff
@@ -0,0 +1,52 @@
...
+        updateStore: yes-really-update-store
+```
+
+## API Changes
+- `storageCluster` status should indicate if the optimization has been applied successfully. 
+- This is useful for brownfield scenarios where the customers will migrate the OSDs to use the `bluestore-rdr` backend store
+
+```YAML
+status:
+    disasterRecoveryOptimization:
```
travisn, line 44: Why do we need special status for the OSDs? As long as the CephCluster CR stays pending until conversion is done, then the storageCluster CR status would also show as not complete.

travisn (changes requested):

docs/design/disaster-recovery-optimization.md
```diff
@@ -0,0 +1,52 @@
...
+- Enable this optimization for both new (Greenfield) and existing (brownfield) customers.
+
+### Greenfield
+- If `ocs.openshift.io/optimizeDisasterRecovery: true` annotation is available in the `StorageCluster` then configure OSD store type in the `cephCluster` as:
+
+```YAML
+spec:
+  Storage:
+    store:
+        type: bluestore-rdr
```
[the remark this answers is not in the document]
sp98: > If the annotation is set, then spec.mirroring.enabled will also be true, right?

I don't think that would be the case. So user creates two ODF cluster, connects them via ACM and then enables mirroring using a DR policy in MultiClusterOrchestrator.  Its at this point when `spec.mirroring.enabled` is set to `true`. 

`spec.mirroring.enabled` won't be `true` for a new cluster ODF cluster.  So new clusters need to rely on the `annotation` only to decide if they want to use the OSD store optimizations. 

If `spec.mirroring.enabled` is `true`, then it will be a brownfield case. That cluster was already created. Now we can check from the status if optimizations are applied already or not. If not applied, then we can pass `yes-really-update-store` option to the ceph cluster.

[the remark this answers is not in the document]
sp98: @umangachapagain Thoughts on this?

docs/design/disaster-recovery-optimization.md
sp98: This would be useful in case of brownfield cases. The UI could use the status to display some sort of message to the users that the cluster is undergoing optimization (OSD migration).  

This status will be generated from `cephCluster` status. So we can still skip adding this status in the `StorageCluster` and UI can directly query `cephCluster` to get the required information it needs.

docs/design/disaster-recovery-optimization.md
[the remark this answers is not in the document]
umangachapagain: > Do we even need to distinguish between green field and brown field? Seems like the only real difference is if we decide to implement green field first, and then brown field later. Really, the brown field implementation will handle green field as well.
>
Totally agree. Ideally, this should all be hidden behind ocs-operator and handled with a brownfield implementation. However, there were some concerns raised with testing and supporting brownfield in the short term.

So, we're going ahead with greenfield support only. In the long term, we'll handle it as brownfield.

> If the annotation is set, then spec.mirroring.enabled will also be true, right?
>
I agree with @sp98 here. If the annotation is set, it doesn't imply that mirroring is enabled. It only indicates that mirroring might be enabled.

docs/design/disaster-recovery-optimization.md
```diff
@@ -0,0 +1,52 @@
...
+```YAML
+spec:
+  storage:
+    store:
+        type: bluestore-rdr
+```
+
+### Brownfield
+- An existing ODF cluster can be updated to use the new OSD backend store.
+- If `spec.mirroring.enabled` is set to `true`, ODF version is greater than equal to 4.14 and OSD store optimizations are not applied already, then configure OSD store in the `cephCluster` as:
```
umangachapagain, line 28: We can skip version check, since optimization will be applied in whichever version the code is added.
We will never have StorageCluster in 4.13 but operators in 4.14.

docs/design/disaster-recovery-optimization.md
umangachapagain: My initial thought about this special status was that, it can be a list of optimizations applied to the cluster. Since it is mostly going to be done behind the scenes, this could provide information on what has and has not been done on the cluster. This can even be surfaced up to multicluster UI.

Maybe something like
```
status:
    disasterRecoveryOptimizations:
    - OSDCreatedWithNewStore | OSDMigratedToNewStore
    - SomeOtherFutureOptimization
```

I agree that detailed status can be referenced from CephCluster. No point duplicating it.

docs/design/disaster-recovery-optimization.md
```diff
@@ -0,0 +1,52 @@
...
+
+- Ceph uses BlueStore as a special-purpose storage back end designed specifically for managing data on disk for Ceph OSD workloads. Ceph team is working on improving this backend storage that will help in significantly improving the performance, especially in disaster recovery scenarios. This new OSD backend is known as bluestore-rdr. In the future, other backends will also need to be supported, such as seastore.
+
+## Proposal
+
+- Enable the optimized OSD backend store for RDR clusters.
+- Enable this optimization for both new (Greenfield) and existing (brownfield) customers.
+
+### Greenfield
+- If `ocs.openshift.io/optimizeDisasterRecovery: true` annotation is available in the `StorageCluster` then configure OSD store type in the `cephCluster` as:
```
sp98, line 17: @umangachapagain are you ok with the annotation?

umangachapagain: I'd prefer avoiding "optimize" and keeping it generic like "clusterIsDisasterRecoveryReady" or "clusterIsDisasterRecoveryTarget" etc since we might not always be doing optimization operations. But, I don't mind this as well.

umangachapagain: "enabled" would be misleading, since user wants to enable it but might not have done that already.

sp98: got it.

umangachapagain (approved):

PR #2132, opened 2023-08-11
Use optimized osd store for new clusters

sp98:
If `ocs.openshift.io/clusterIsDisasterRecoveryTarget: true` annotation is available in the StorageCluster, then create `cephCluster` using `bluestore-rdr` backend store. 

If storageCluster is some how updated to remove this annotation, then ensure that
`bluestore-rdr` store type is not overridden

This PR also reverts the changes that exposed the OSD store in the storageCluster spec, that is #2116.

### DIFF
--- a/api/v1/storagecluster_types.go
+++ b/api/v1/storagecluster_types.go
@@ -45,7 +45,6 @@ type StorageClusterSpec struct {
 	// Resources follows the conventions of and is mapped to CephCluster.Spec.Resources
 	Resources          map[string]corev1.ResourceRequirements `json:"resources,omitempty"`
 	Encryption         EncryptionSpec                         `json:"encryption,omitempty"`
-	OSDStore           rookCephv1.OSDStore                    `json:"osdStore,omitempty"`
 	StorageDeviceSets  []StorageDeviceSet                     `json:"storageDeviceSets,omitempty"`
 	MonPVCTemplate     *corev1.PersistentVolumeClaim          `json:"monPVCTemplate,omitempty"`
 	MonDataDirHostPath string                                 `json:"monDataDirHostPath,omitempty"`
--- a/controllers/storagecluster/cephcluster.go
+++ b/controllers/storagecluster/cephcluster.go
@@ -72,6 +72,8 @@ const (
 	networkProvider           = "multus"
 	publicNetworkSelectorKey  = "public"
 	clusterNetworkSelectorKey = "cluster"
+	// DisasterRecoveryTargetAnnotation signifies that the cluster is intended to be used for Disaster Recovery
+	DisasterRecoveryTargetAnnotation = "ocs.openshift.io/clusterIsDisasterRecoveryTarget"
 )
 
 const (
@@ -261,6 +263,9 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
 	// Record actual Ceph container image version before attempting update
 	sc.Status.Images.Ceph.ActualImage = found.Spec.CephVersion.Image
 
+	// Prevent removal of any RDR optimizations if they are already applied to the existing cluster spec.
+	cephCluster.Spec.Storage.Store = determineOSDStore(cephCluster.Spec.Storage.Store, found.Spec.Storage.Store)
+
 	// Add it to the list of RelatedObjects if found
 	objectRef, err := reference.GetReference(r.Scheme, found)
 	if err != nil {
@@ -405,6 +410,11 @@ func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, nodeCount int, s
 		MaxLogSize:  &maxLogSize,
 	}
 
+	osdStore := getOSDStoreConfig(sc)
+	if osdStore.Type != "" {
+		reqLogger.Info("osd store settings", osdStore)
+	}
+
 	cephCluster := &rookCephv1.CephCluster{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      generateNameForCephCluster(sc),
@@ -434,6 +444,7 @@ func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, nodeCount int, s
 			},
 			Storage: rookCephv1.StorageScopeSpec{
 				StorageClassDeviceSets: newStorageClassDeviceSets(sc, serverVersion),
+				Store:                  osdStore,
 			},
 			Placement: rookCephv1.PlacementSpec{
 				"all":     getPlacement(sc, "all"),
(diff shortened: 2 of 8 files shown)


api/v1/storagecluster_types.go
```diff
@@ -686,3 +690,14 @@ func (r *StorageCluster) NewToolsDeployment(tolerations []corev1.Toleration) *ap
...
+// OptimizeDisasterRecovery returns true if any RDR optimizations are required
+func (r *StorageCluster) OptimizeDisasterRecovery() bool {
+	if annotation, found := r.GetAnnotations()[DisasterRecoveryOptimizations]; found {
+		if annotation == "true" {
+			return true
+		}
+	}
+
+	return false
+}
```
umangachapagain, line 703: Let's move this to `controllers/storagecluster/cephcluster.go` or to the utility package `controllers/util/`

umangachapagain (changes requested):

api/v1/storagecluster_types.go
sp98: moved it to `controllers/storagecluster/cephcluster.go`

api/v1/storagecluster_types.go
```diff
@@ -29,6 +29,10 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	DisasterRecoveryOptimizations = "ocs.openshift.io/optimizeDisasterRecovery"
+)
```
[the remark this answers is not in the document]
sp98: moved it to `controllers/storagecluster/cephcluster.go`. Didn't find a right place within utils package

controllers/storagecluster/cephcluster.go
```diff
@@ -1255,3 +1259,12 @@ func getIPFamilyConfig(c client.Client) (rookCephv1.IPFamilyType, bool, error) {
 
 	return rookCephv1.IPv4, false, nil
 }
+
+func getOSDStoreConfig(sc *ocsv1.StorageCluster) rookCephv1.OSDStore {
+	osdStore := rookCephv1.OSDStore{}
+	if !sc.Spec.ExternalStorage.Enable && sc.OptimizeDisasterRecovery() {
+		osdStore.Type = string(rookCephv1.StoreTypeBlueStoreRDR)
+	}
```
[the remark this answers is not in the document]
sp98: what might cause the deletion of the annotation, other than a user manually deleting it?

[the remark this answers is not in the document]
sp98: only way I could think of preventing it is to add some validations before the cephCluster is updated and prevent changing osd storetype if `bluestoreRDR` is already applied. But looks hackish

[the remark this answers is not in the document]
umangachapagain: > what might cause the deletion of the annotation, other than a user manually deleting it?
>
Users manually deleting or other automation overriding it.

> only way I could think of preventing it is to add some validations before the cephCluster updated operation and prevent changing osd storetype if bluestoreRDR is already applied. But looks hackish

Yes, that's pretty much what I thought.

```
if osdStore is not bluestoreRDR:
    osdStore = getOSDStoreConfig(sc)
else
   continue
```
This allows to skip the function call when it's not needed, but leaks implementation of OSDStore to calling function.

```
func  getOSDStoreConfig(...) osdStoreConfig{
 if sc.spec.osdStore == blueStoreRDR{
  return sc.spec.osdStore
 }
 return osdStoreConfig{bluestoreRDR}
}
```
Adds an extra function call for every reconcile, but allows encapsulating OSDStoreConfig implementation. It also allows more changes later within the function.

controllers/storagecluster/cephcluster.go
```diff
@@ -405,6 +410,9 @@ func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, nodeCount int, s
 		MaxLogSize:  &maxLogSize,
 	}
 
+	osdStore := getOSDStoreConfig(sc)
+	reqLogger.Info("osd store settings", osdStore)
```
travisn, line 414: For the non-rdr case, will this setting be blank? Perhaps only log this if not blank?

controllers/storagecluster/cephcluster.go
```diff
@@ -1256,3 +1264,31 @@ func getIPFamilyConfig(c client.Client) (rookCephv1.IPFamilyType, bool, error) {
...
+	if annotation, found := sc.GetAnnotations()[DROptimizationAnnotation]; found {
+		if annotation == "true" {
+			return true
+		}
+	}
+
+	return false
+}
+
+func updateRDROptimizations(newOSDStore, existingOSDStore rookCephv1.OSDStore) rookCephv1.OSDStore {
```
travisn, line 1288: How about this method name?
```suggestion
func determineOSDStore(newOSDStore, existingOSDStore rookCephv1.OSDStore) rookCephv1.OSDStore {
```

travisn (changes requested):

controllers/storagecluster/cephcluster.go
sp98: yes. Logging only if not blank.

travisn (commented):
Remind me, why is it that we're using an annotation on the StorageCluster CR instead of a setting on the same CR?

sp98:
> Remind me, why is it that we're using an annotation on the StorageCluster CR instead of a setting on the same CR?

Earlier implementation used the entire osd store spec in the `storageCluster` spec. So user needed to provide `osd-store` type in the spec. 

Current changes hide these implementation details from the spec. Since this is only a one time operation, user can't really go back after applying the new osd store, we decided to keep it an annotation than setting on the CR spec itself.

umangachapagain:
> Remind me, why is it that we're using an annotation on the StorageCluster CR instead of a setting on the same CR?
>
@travisn What OSD store to use is an implementation detail. We do not expect users to know what it means and we do not want them to manage it. Also, once OSD store is set to "bluestore-rdr", we do not want them to revert it. Since ocs-operator will internally handle this configuration, and users have no real control over it, it didn't seem reasonable as a StorageCluster spec. When we support brownfield, this spec will be even more useless and a maintenance overhead.

I suggested making this an annotation as a stopgap, since, in greenfield deployments, we needed a way to know if ODF needs to be configured for DR. And, this is a one time operation, as once cluster is deployed, we do not support changing configurations.
When we have better automation in place including brownfield support, we can just drop this annotation or ignore it without any additional overhead, which wouldn't be possible with a spec change.

Does it make sense? Or, am I overthinking this?

travisn:
> > Remind me, why is it that we're using an annotation on the StorageCluster CR instead of a setting on the same CR?
> 
> @travisn What OSD store to use is an implementation detail. We do not expect users to know what it means and we do not want them to manage it. Also, once OSD store is set to "bluestore-rdr", we do not want them to revert it. Since ocs-operator will internally handle this configuration, and users have no real control over it, it didn't seem reasonable as a StorageCluster spec. When we support brownfield, this spec will be even more useless and a maintenance overhead.
> 
> I suggested making this an annotation as a stopgap, since, in greenfield deployments, we needed a way to know if ODF needs to be configured for DR. And, this is a one time operation, as once cluster is deployed, we do not support changing configurations. When we have better automation in place including brownfield support, we can just drop this annotation or ignore it without any additional overhead, which wouldn't be possible with a spec change.
> 
> Does it make sense? Or, am I overthinking this?

Yes, sounds good to use an annotation for such a short-term setting. Thanks for the background.

umangachapagain (approved):

PR #2150, opened 2023-08-17
external: donot stop rbd and cephfs creation if rgw end.

parth-gr:
… is not reachable

the rbd cephfs and rgw creation is tightly coupled so if the rgw endpoint is not reachable it
doesn't create other to sc as well
So instead of returning the error just log it

BZ: Bug 2213757

### DIFF
--- a/controllers/storagecluster/external_resources.go
+++ b/controllers/storagecluster/external_resources.go
@@ -353,21 +353,22 @@ func (r *StorageClusterReconciler) createExternalStorageClusterResources(instanc
 				scc = newCephBlockPoolStorageClassConfiguration(instance)
 			} else if d.Name == cephRgwStorageClassName {
 				rgwEndpoint := d.Data[externalCephRgwEndpointKey]
-				if err := checkEndpointReachable(rgwEndpoint, 5*time.Second); err != nil {
-					r.Log.Error(err, "RGW endpoint is not reachable.", "RGWEndpoint", rgwEndpoint)
-					return err
-				}
-				extCephObjectStores, err = r.newExternalCephObjectStoreInstances(instance, rgwEndpoint)
-				if err != nil {
-					return err
-				}
-				// rgw-endpoint is no longer needed in the 'd.Data' dictionary,
-				// and can be deleted
-				// created an issue in rook to add `CephObjectStore` type directly in the JSON output
-				// https://github.com/rook/rook/issues/6165
-				delete(d.Data, externalCephRgwEndpointKey)
+				rgwErr := checkEndpointReachable(rgwEndpoint, 5*time.Second)
+				if rgwErr != nil {
+					r.Log.Error(err, "RGW endpoint is not reachable. Will skip creating objectstore and RGW Storage Class", "RGWEndpoint", rgwEndpoint)
+				} else {
+					extCephObjectStores, err = r.newExternalCephObjectStoreInstances(instance, rgwEndpoint)
+					if err != nil {
+						return err
+					}
+					// rgw-endpoint is no longer needed in the 'd.Data' dictionary,
+					// and can be deleted
+					// created an issue in rook to add `CephObjectStore` type directly in the JSON output
+					// https://github.com/rook/rook/issues/6165
+					delete(d.Data, externalCephRgwEndpointKey)
 
-				scc = newCephOBCStorageClassConfiguration(instance)
+					scc = newCephOBCStorageClassConfiguration(instance)
+				}
 			}
 			// now sc is pointing to appropriate StorageClass,
 			// whose parameters have to be updated


iamniting (changes requested):
Can you please change the commit to fit in the 73 char, Also please remove the bug id from the commit-msg, instead add it in the PR description. Having the bug id in the commit-msg creates confusion while backporting a PR.

We will need 2 bugs one for 4.14 and another one for 4.13. We can keep this one for 4.14 and clone it for 4.13.

controllers/storagecluster/external_resources.go
```diff
@@ -354,20 +354,21 @@ func (r *StorageClusterReconciler) createExternalStorageClusterResources(instanc
...
-					return err
+				if err == nil {
+					extCephObjectStores, err = r.newExternalCephObjectStoreInstances(instance, rgwEndpoint)
+					if err != nil {
+						return err
+					}
+					// rgw-endpoint is no longer needed in the 'd.Data' dictionary,
+					// and can be deleted
+					// created an issue in rook to add `CephObjectStore` type directly in the JSON output
+					// https://github.com/rook/rook/issues/6165
```
subhamkrai, line 367:
```suggestion
					// rgw-endpoint is no longer needed in the 'd.Data' dictionary,
					// and can be deleted
					// created an issue in rook to add `CephObjectStore` type directly in the JSON output
					// https://github.com/rook/rook/issues/6165
```
@parth-gr could look at this comment if it still valid?

controllers/storagecluster/external_resources.go
```diff
@@ -354,20 +354,21 @@ func (r *StorageClusterReconciler) createExternalStorageClusterResources(instanc
 			} else if d.Name == cephRgwStorageClassName {
 				rgwEndpoint := d.Data[externalCephRgwEndpointKey]
 				if err := checkEndpointReachable(rgwEndpoint, 5*time.Second); err != nil {
-					r.Log.Error(err, "RGW endpoint is not reachable.", "RGWEndpoint", rgwEndpoint)
-					return err
+					r.Log.Error(err, "RGW endpoint is not reachable. Will not create objectstore and RGW Storage Class", "RGWEndpoint", rgwEndpoint)
```
subhamkrai, line 357:
```suggestion
					r.Log.Error(err, "RGW endpoint is not reachable. Will skip creating objectstore and RGW Storage Class", "RGWEndpoint", rgwEndpoint)
```

subhamkrai (changes requested):

controllers/storagecluster/external_resources.go
parth-gr: Updated

controllers/storagecluster/external_resources.go
parth-gr: Yes look kinda an improvement for the code 

It would need changes in both rook and oc-op

parth-gr:
> Can you please change the commit to fit in the 73 char, Also please remove the bug id from the commit-msg, instead add it in the PR description. Having the bug id in the commit-msg creates confusion while backporting a PR.
> 
> We will need 2 bugs one for 4.14 and another one for 4.13. We can keep this one for 4.14 and clone it for 4.13.

Above one is valid for 4.14 once we have a backport PR I will clone the bug to 4.13
Thanks

controllers/storagecluster/external_resources.go
```diff
@@ -354,20 +354,21 @@ func (r *StorageClusterReconciler) createExternalStorageClusterResources(instanc
 			} else if d.Name == cephRgwStorageClassName {
 				rgwEndpoint := d.Data[externalCephRgwEndpointKey]
 				if err := checkEndpointReachable(rgwEndpoint, 5*time.Second); err != nil {
-					r.Log.Error(err, "RGW endpoint is not reachable.", "RGWEndpoint", rgwEndpoint)
-					return err
+					r.Log.Error(err, "RGW endpoint is not reachable. Will skip creating objectstore and RGW Storage Class", "RGWEndpoint", rgwEndpoint)
```
travisn, line 357: Is the RGW required to be online to create the rgw storage class? If the endpoint is just down temporarily, this means the storage class won't be created as expected. Why not still continue creating the storage class?

parth-gr: Storage class needs the `cephobjectstorename` as a parameter,

So if don't create objectstore at the start because of rgw-endpoint not reachable, then it won't be created later as this code I think execute once at the very start for initializing the external cluster.

And also as we are using the objectstorename to create the SC so I think that would be also needed later for the obc creation internallly

parth-gr: And as we don't send the error to the reconcile I am not sure it will reconcile

controllers/storagecluster/external_resources.go
```diff
@@ -354,20 +354,21 @@ func (r *StorageClusterReconciler) createExternalStorageClusterResources(instanc
...
 				}
-				extCephObjectStores, err = r.newExternalCephObjectStoreInstances(instance, rgwEndpoint)
-				if err != nil {
-					return err
+				if err == nil {
```
travisn, line 359: `err` is actually defined up on line 278, so this is a problem. We want to use the `err` defined on line 356, but that is a different scope. I would suggest using an `else` on line 358 instead of checking `err == nil` on this line.

parth-gr: Updated this for now

controllers/storagecluster/external_resources.go
```diff
@@ -353,21 +353,22 @@ func (r *StorageClusterReconciler) createExternalStorageClusterResources(instanc
 				scc = newCephBlockPoolStorageClassConfiguration(instance)
 			} else if d.Name == cephRgwStorageClassName {
 				rgwEndpoint := d.Data[externalCephRgwEndpointKey]
-				if err := checkEndpointReachable(rgwEndpoint, 5*time.Second); err != nil {
-					r.Log.Error(err, "RGW endpoint is not reachable.", "RGWEndpoint", rgwEndpoint)
-					return err
-				}
-				extCephObjectStores, err = r.newExternalCephObjectStoreInstances(instance, rgwEndpoint)
+				err := checkEndpointReachable(rgwEndpoint, 5*time.Second)
```
travisn, line 356: I thought we discussed to store the rgw error, and then fail the reconcile later around line 363?
```suggestion
				rgwErr = checkEndpointReachable(rgwEndpoint, 5*time.Second)
```

controllers/storagecluster/external_resources.go
travisn: Before we proceed with this change, let's finalize the discussion on the BZ

malayparida2000:
> @travisn I think the approach we discussed yesterday for returning the err while creation of rgw Storage class, Would be blocking this for loop which also creates other resources.
> 
> https://github.com/red-hat-storage/ocs-operator/blob/867fd8f3d54565fb1ff244bf9e4b59de6800f20f/controllers/storagecluster/reconcile.go#L416-L429
> 
> I am not sure why this is not implemented with a go routine for different resources, which would make this as a parallel execution, @malayparida2000 any thoughts

The goroutine idea sounds promising. 
For now it seems like we have 3 options 
1. Return err/requeue and try on reconciling till rgw end point is reachable.
(This will block the creation of other resources like the Cephcluster)
2. Even if endpoint is not reachable create the storageclass anyway
(This will create issues as even though the storage class is present it won't be able to provision)
3. Don't return err or requeue just skip out the rgw storageclass/cephobjectstore
(Whenever there is another reconcile which we don't know when will happen, it will check the status of the endpoint again. if it's availble then create the objectstore & the rgw storageclass)

To me option 3 seems to be the right way.
@iamniting can you please also take a look here.

parth-gr:
The best way the operator should work if returning the error,

Here the more pointers from Travis https://bugzilla.redhat.com/show_bug.cgi?id=2213757#c9
Which says we should re-work the creation of resources, 
Either make them concurrent or at least not return the error till all the resources are called from the for loop,

So basically we would be holding an error messages in the `map` and return the errors only after all the resources get created

malayparida2000:
> The best way the operator should work if returning the error,
> 
> Here the more pointers from Travis https://bugzilla.redhat.com/show_bug.cgi?id=2213757#c9 Which says we should re-work the creation of resources, Either make them concurrent or at least not return the error till all the resources are called from the for loop,
> 
> So basically we would be holding error messages in the `map` and return the errors only after all the resources called ensure created

To correctly address the root cause we have to design a correct resource flow map identifying correctly which resources are independent of each other and can be created simultaneously and which need to happen in order. Looking at the implications of such a big change I think it's right to move this out of 4.14 which Travis has done. We have to bring this up in some discussions where we can have a consensus over this. Probably the odf bi-weekly call will be a good opportunity for this. I will request @parth-gr to please bring this topic up there.

travisn:
Agreed, let's discuss on the approach. A couple thoughts before then...
- Go routines are best to avoid during reconcile for normal reconcile operations. The logging is harder to troubleshoot if multiple things are happening, and it also may require higher pod resource limits since the operator would burst more. 
- Being able to reconcile independent actions will be good, to allow collecting the errors and only failing after all actions have been attempted. Still, it is important to fail the reconcile and retry for the failures where resources could not be created.

malayparida2000:
Hi @parth-gr, do you Plan to work on this? if not can you please close it

malayparida2000:
> > * ing is harder to troubleshoot if multiple things are happening, and it also may require higher pod resource limits since the operator would burst more.
> > * Being able to reconcile independent actions will be good, to allow collecting the errors and only failing after all actions have been attempted. Still, it is important to fail the reconcile and retry for the failures where reso
> 
> We can bring this topic in odf weekly?
> 
> But we need understand it adds value, as currently I see most of the things are in order and dependent

Sure let's bring this again in 1 of the odf-weekly

PR #2152, opened 2023-08-21
metrics: fix storageclass filter in collector

umangachapagain:
Since StorageClass is not namespaced, we do not need to filter them based on namespaces being watched by the metrics exporter. Doing so can cause issues like all the storageclasses being ignored since it's namespace field is empty.

Instead, we filter StorageClass by provisioner, to report metrics for the Ceph based storageclasses only.

### DIFF
--- a/metrics/internal/collectors/cluster-advance-feature-use.go
+++ b/metrics/internal/collectors/cluster-advance-feature-use.go
@@ -1,6 +1,8 @@
 package collectors
 
 import (
+	"strings"
+
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/red-hat-storage/ocs-operator/v4/metrics/internal/options"
 	cephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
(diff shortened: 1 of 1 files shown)


malayparida2000 (approved):
/lgtm

metrics/internal/collectors/cluster-advance-feature-use.go
```diff
@@ -208,18 +210,17 @@ func getAllStorageClasses(
...
-	for _, namespace := range namespaces {
-		for _, eachSC := range allSCs {
-			if eachSC.Namespace == namespace {
-				namespacedSCs = append(namespacedSCs, eachSC)
-			}
+	// Since there can be many StorageClass not related to ceph, we filter
+	// those out based on provisioner value.
+	var odfSC []*storagev1.StorageClass
+	for _, sc := range allSCs {
+		if strings.Contains(sc.Provisioner, "rbd.csi.ceph.com") ||
```
malayparida2000, line 217: Instead of defining these 3 provisioners can't we just take any storage class whose provisioner contains the string "csi.ceph.com"?

umangachapagain: We use this selector everywhere, so I used this for consistency. Also, it's better to be specific about which provisioner we're watching for.

malayparida2000:
@umangachapagain Any reason we are backporting it to 4.12 & 4.13 but not 4.14?

PR #2159, opened 2023-08-23
Update rook version to v1.12.2

malayparida2000:


controllers/storagecluster/cephcluster.go
```diff
@@ -1026,13 +1026,13 @@ func generateStretchClusterSpec(sc *ocsv1.StorageCluster) *rookCephv1.StretchClu
 		if zone == sc.Spec.NodeTopologies.ArbiterLocation {
 			continue
 		}
-		stretchClusterSpec.Zones = append(stretchClusterSpec.Zones, rookCephv1.StretchClusterZoneSpec{
+		stretchClusterSpec.Zones = append(stretchClusterSpec.Zones, rookCephv1.MonZoneSpec{
```
umangachapagain, line 1029: @travisn is this change ok?

malayparida2000: This was necessary due to this https://github.com/rook/rook/pull/12384/commits/2231d73259707e082c24419bf68e4382444e3c6f#diff-46a2e53ea9b825b7c3b65113b63a133da71eb0cd2a2c0e33012c224d02a8eaa1

umangachapagain (changes requested):
Squash the first 2 commits for go deps. It's the same thing so better to keep it together.

controllers/storagecluster/cephcluster.go
malayparida2000: Seems like a rename only, @travisn can confirm

malayparida2000:
> Squash the first 2 commits for go deps. It's the same thing so better to keep it together.

done

controllers/storagecluster/cephcluster.go
travisn: Correct, only the type name changed in a refactoring, but the json representation is the same and is not expected to affect upgrades

umangachapagain (approved):

PR #2160, opened 2023-08-23
Support migrating existing OSDs to bluestore-rdr

sp98:
This PR adds `yes-really-update-store` to `cephCluster` to migrate OSDs to bluestore-rdr` only if mirroring is enabled. 

- Adds `yes-really-update-store` for the first time only if : 
   - Mirroring is enabled.
   - CephCluster is Ready
   - CephCluster status indicates all OSDs are not on bluestore-rdr

- Ensures `yes-really-update-store` flag is not removed if CephCluster status indicates all OSDs are NOT on bluestore-rdr

- Removes `yes-really-update-store` if CephCluster status indicates that all OSDs are on bluestore-rdr and ceph cluster status is `Ready`


controllers/storagecluster/cephcluster.go
```diff
@@ -268,6 +269,13 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
 
 	// Prevent removal of any RDR optimizations if they are already applied to the existing cluster spec.
 	cephCluster.Spec.Storage.Store = determineOSDStore(cephCluster.Spec.Storage.Store, found.Spec.Storage.Store)
+	if osdStoreUpdateRequired(sc, *found) {
+		cephCluster.Spec.Storage.Store.UpdateStore = OSDStoreUpdateConfirmation
+		r.Log.Info("migration of OSDs to bluestore-rdr is required")
+	} else {
+		cephCluster.Spec.Storage.Store.UpdateStore = ""
```
travisn, line 276: This confirmation to update the store will be reverted after the conversion is completed? Seems like we could simplify and just always keep this set when RDR is enabled, since after the conversion is complete it will be a no-op.

sp98: It would save the API call to list all the OSD deployments in the rook operator. 

Do you think we should do that?

or is it ok to always set `yes-really-update-store` ?

sp98: After a discussion in the Rook huddle, we decided to just always enable `bluestore-rdr` and `yes-really-update-store`. Rook takes care of migrating the OSDs and pauses migration if cluster is unhealthy. We don't need to remove `yes-really-update-store` flag, because if all OSDs are already migrated, then it will be no-op

PR #2162, opened 2023-08-23
support using bluestore-rdr for RDR customers

sp98:
If mirroring is enabled, then use bluestore-rdr store and allow migration of OSDs running on bluestore to bluestore-rdr

We always enable `bluestore-rdr` and `yes-really-update-store`. Rook takes care of migrating the OSDs and pauses migration if cluster is unhealthy. We don't need to remove `yes-really-update-store` flag, because if all OSDs are already migrated, then it will be no-op.

### DIFF
--- a/controllers/storagecluster/cephcluster.go
+++ b/controllers/storagecluster/cephcluster.go
@@ -269,6 +269,12 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
 	// Prevent removal of any RDR optimizations if they are already applied to the existing cluster spec.
 	cephCluster.Spec.Storage.Store = determineOSDStore(cephCluster.Spec.Storage.Store, found.Spec.Storage.Store)
 
+	// Use bluestore-rdr if mirrioring is enabled
+	if sc.Spec.Mirroring.Enabled && !sc.Spec.ExternalStorage.Enable {
+		cephCluster.Spec.Storage.Store.Type = string(rookCephv1.StoreTypeBlueStoreRDR)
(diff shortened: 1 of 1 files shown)


controllers/storagecluster/cephcluster.go
```diff
@@ -269,6 +269,12 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
 	// Prevent removal of any RDR optimizations if they are already applied to the existing cluster spec.
 	cephCluster.Spec.Storage.Store = determineOSDStore(cephCluster.Spec.Storage.Store, found.Spec.Storage.Store)
 
+	// Use bluestore-rdr if mirrioring is enabled
+	if sc.Spec.Mirroring.Enabled && !sc.Spec.ExternalStorage.Enable {
+		cephCluster.Spec.Storage.Store.Type = string(rookCephv1.StoreTypeBlueStoreRDR)
```
travisn, line 274: Where is the store type set for green field when the annotation is set? How are these related? I guess the type will just be set to the same value by two different sources?

sp98: Annotation is used only when the new `CephCluster` CR is being created [here](https://github.com/red-hat-storage/ocs-operator/blob/d10e2c8b6a3c4191fcace37c38cc42d1ac8b9f70/controllers/storagecluster/cephcluster.go#L416). 

`yes-really-update-store` only when updating existing `cephCluster` 

>> I guess the type will just be set to the same value by two different sources?
Yes.

travisn (approved):

umangachapagain (approved):

PR #2167, opened 2023-08-25
Bug 2212333:[release-4.12] Removing capacity feature from the storageconsumer CR

rchikatw:


rchikatw:
Closing this PR As we discontinue FaaS development and in the current situation, customers would not see provider/consumer support for 4.12/4.13 so this backport is not needed. Hence Closing the bug.

PR #2168, opened 2023-08-25
Add annotation for multus in toolbox after checking the multus network name for namespace

Nikhil-Ladha:
It is possible that the NetworkAttachmentDefinition for multus network is configured in a different namespace than the one in which the operator is deployed and multus is configured. In that case, the selector (eg: public) value in the storagecluster CR is of the following pattern `NAD_namespace/NAD_name`.

Therefore, check for namespace value in the selector and add annotation to the ceph toolbox accordingly.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1982721


Nikhil-Ladha:
@iamniting @umangachapagain PTAL

controllers/storagecluster/cephtoolbox.go
```diff
@@ -117,21 +119,32 @@ func removeDuplicateTolerations(tolerations []corev1.Toleration) []corev1.Tolera
...
 		return "", err
 	}
 
+	if sc.Spec.Network.Selectors["public"] == "" {
+		return "", nil
+	}
+
+	multusNetName := sc.Spec.Network.Selectors["public"]
+	nsInMultusName := strings.Split(multusNetName, "/")
+	if len(nsInMultusName) > 1 {
```
iamniting, line 134: Can you check that this is == 2
If it is > 2 return an error.

controllers/storagecluster/cephtoolbox.go
```diff
@@ -117,21 +119,32 @@ func removeDuplicateTolerations(tolerations []corev1.Toleration) []corev1.Tolera
...
 	if err != nil {
 		return "", err
 	}
 
+	if sc.Spec.Network.Selectors["public"] == "" {
+		return "", nil
+	}
+
+	multusNetName := sc.Spec.Network.Selectors["public"]
+	nsInMultusName := strings.Split(multusNetName, "/")
```
iamniting, line 133:
```suggestion
	MultusNetNamespacedName := strings.Split(multusNetName, "/")
```

iamniting (changes requested):

controllers/storagecluster/cephtoolbox.go
Nikhil-Ladha: Ack, updated the code

controllers/storagecluster/cephtoolbox.go
```diff
@@ -117,21 +119,34 @@ func removeDuplicateTolerations(tolerations []corev1.Toleration) []corev1.Tolera
...
 		return "", err
 	}
 
-	if sc.Spec.Network.Selectors["public"] != "" {
-		multusNetName := sc.Spec.Network.Selectors["public"]
-		return fmt.Sprintf("%s/%s", multusNetNs, multusNetName), nil
-	}
-
-	return "", nil
+	return fmt.Sprintf("%s/%s", multusNetNs, multusNetName), nil
```
iamniting, line 151: As we are doing the get on the one if it is another namespace should not we do get even if it is in the same namespace?

iamniting (changes requested):

controllers/storagecluster/cephtoolbox.go
Nikhil-Ladha: Ah, right. It would be good to test it for the operator ns as well.
Updated the code accordingly, PTAL.

controllers/storagecluster/cephtoolbox.go
```diff
@@ -117,21 +119,36 @@ func removeDuplicateTolerations(tolerations []corev1.Toleration) []corev1.Tolera
...
 
-	if sc.Spec.Network.Selectors["public"] != "" {
-		multusNetName := sc.Spec.Network.Selectors["public"]
-		return fmt.Sprintf("%s/%s", multusNetNs, multusNetName), nil
+	multusNetName := sc.Spec.Network.Selectors["public"]
+	multusNetNamespacedName := strings.Split(multusNetName, "/")
+	nadNS, nadName := "", ""
+	if len(multusNetNamespacedName) == 2 {
+		nadNS, nadName = multusNetNamespacedName[0], multusNetNamespacedName[1]
+	} else if len(multusNetNamespacedName) < 2 {
```
iamniting, line 137:
```suggestion
	} else if len(multusNetNamespacedName) == 1 {
```

controllers/storagecluster/cephtoolbox.go
```diff
@@ -117,21 +119,36 @@ func removeDuplicateTolerations(tolerations []corev1.Toleration) []corev1.Tolera
...
+		nadNS, err = util.GetOperatorNamespace()
+		if err != nil {
+			return "", err
+		}
+		nadName = multusNetName
+	} else {
+		return "", fmt.Errorf("public selector value: %s, is not as per specification", multusNetName)
```
iamniting, line 144: I think we need to enhance the error msg.

controllers/storagecluster/cephtoolbox.go
```diff
@@ -117,21 +119,36 @@ func removeDuplicateTolerations(tolerations []corev1.Toleration) []corev1.Tolera
...
+	multusNetNamespacedName := strings.Split(multusNetName, "/")
+	nadNS, nadName := "", ""
+	if len(multusNetNamespacedName) == 2 {
+		nadNS, nadName = multusNetNamespacedName[0], multusNetNamespacedName[1]
+	} else if len(multusNetNamespacedName) < 2 {
+		nadNS, err = util.GetOperatorNamespace()
+		if err != nil {
+			return "", err
+		}
+		nadName = multusNetName
```
iamniting, line 142:
```suggestion
		nadNS, nadName = os.Getenv(util.OperatorNamespaceEnvVar), multusNetName
```

controllers/storagecluster/cephtoolbox.go
```diff
@@ -117,21 +119,36 @@ func removeDuplicateTolerations(tolerations []corev1.Toleration) []corev1.Tolera
...
+		return "", nil
 	}
 
-	if sc.Spec.Network.Selectors["public"] != "" {
-		multusNetName := sc.Spec.Network.Selectors["public"]
-		return fmt.Sprintf("%s/%s", multusNetNs, multusNetName), nil
+	multusNetName := sc.Spec.Network.Selectors["public"]
+	multusNetNamespacedName := strings.Split(multusNetName, "/")
+	nadNS, nadName := "", ""
+	if len(multusNetNamespacedName) == 2 {
```
iamniting, line 135: Let's change the order of the if conditions as well. First ==1 and then ==2 and then else.

iamniting (changes requested):

controllers/storagecluster/cephtoolbox.go
Nikhil-Ladha: How about this?
`return "", fmt.Errorf("public network selector for multus: %s, is invalid", multusNetName)`

controllers/storagecluster/cephtoolbox.go
Nikhil-Ladha: Done

controllers/storagecluster/cephtoolbox.go
Nikhil-Ladha: Done

controllers/storagecluster/cephtoolbox.go
Nikhil-Ladha: Ack

controllers/storagecluster/cephtoolbox.go
```diff
@@ -117,21 +120,32 @@ func removeDuplicateTolerations(tolerations []corev1.Toleration) []corev1.Tolera
...
-		return fmt.Sprintf("%s/%s", multusNetNs, multusNetName), nil
+	multusNetName := sc.Spec.Network.Selectors["public"]
+	multusNetNamespacedName := strings.Split(multusNetName, "/")
+	nadNS, nadName := "", ""
+	if len(multusNetNamespacedName) == 1 {
+		nadNS, nadName = os.Getenv(util.OperatorNamespaceEnvVar), multusNetName
+	} else if len(multusNetNamespacedName) == 2 {
+		nadNS, nadName = multusNetNamespacedName[0], multusNetNamespacedName[1]
+	} else {
+		return "", fmt.Errorf("public network selector for multus: %s, is invalid", multusNetName)
```
iamniting, line 141: I think we should add the spec path in the logs so that customers would be aware of what exactly needs to be changed.

iamniting (changes requested):

controllers/storagecluster/cephtoolbox.go
```diff
@@ -117,21 +120,32 @@ func removeDuplicateTolerations(tolerations []corev1.Toleration) []corev1.Tolera
...
+	}
+
+	foundNAD := &nadv1.NetworkAttachmentDefinition{}
+	err = r.Client.Get(context.TODO(), types.NamespacedName{Name: nadName, Namespace: nadNS}, foundNAD)
+	if err != nil {
```
iamniting, line 146: Let's add a log here as well. This is expected to be created by a customer, either create it or alter the spec.

iamniting (changes requested):

controllers/storagecluster/cephtoolbox.go
Nikhil-Ladha: Ok, how about now?

controllers/storagecluster/cephtoolbox.go
```diff
@@ -117,21 +120,33 @@ func removeDuplicateTolerations(tolerations []corev1.Toleration) []corev1.Tolera
...
+	} else if len(multusNetNamespacedName) == 2 {
+		nadNS, nadName = multusNetNamespacedName[0], multusNetNamespacedName[1]
+	} else {
+		return "", fmt.Errorf("spec.Network.Selector.Public value: %s in storagecluster CR is invalid", multusNetName)
+	}
+
+	foundNAD := &nadv1.NetworkAttachmentDefinition{}
+	err = r.Client.Get(context.TODO(), types.NamespacedName{Name: nadName, Namespace: nadNS}, foundNAD)
+	if err != nil {
+		r.Log.Error(err, "Either create the NetworkAttachmentDefinition or alter the storagecluster spec with correct value")
```
iamniting, line 147:
```suggestion
		r.Log.Error(err, "Either create the NetworkAttachmentDefinition or alter the storagecluster Spec.Network.Selectors[\"public\"] with correct value")
```

controllers/storagecluster/cephtoolbox.go
```diff
@@ -117,21 +120,33 @@ func removeDuplicateTolerations(tolerations []corev1.Toleration) []corev1.Tolera
...
-		return fmt.Sprintf("%s/%s", multusNetNs, multusNetName), nil
+	multusNetName := sc.Spec.Network.Selectors["public"]
+	multusNetNamespacedName := strings.Split(multusNetName, "/")
+	nadNS, nadName := "", ""
+	if len(multusNetNamespacedName) == 1 {
+		nadNS, nadName = os.Getenv(util.OperatorNamespaceEnvVar), multusNetName
+	} else if len(multusNetNamespacedName) == 2 {
+		nadNS, nadName = multusNetNamespacedName[0], multusNetNamespacedName[1]
+	} else {
+		return "", fmt.Errorf("spec.Network.Selector.Public value: %s in storagecluster CR is invalid", multusNetName)
```
iamniting, line 141:
```suggestion
		return "", fmt.Errorf("Spec.Network.Selectors[\"public\"] value: %s in storagecluster CR is invalid", multusNetName)
```

iamniting (changes requested):

controllers/storagecluster/cephtoolbox.go
Nikhil-Ladha: Ack, done :)

controllers/storagecluster/cephtoolbox.go
Nikhil-Ladha: Done

iamniting (commented):
The changes look good to me can you test them once and see if all cases are working well

iamniting (approved):
/hold for testing

Nikhil-Ladha:
> The changes look good to me can you test them once and see if all cases are working well

Ok, it seems like the operator doesn't have permissions to access resources outside the `openshift-storage` ns. This is happening due to the recent [change](https://github.com/red-hat-storage/ocs-operator/commit/d98bf7b8f021e8ef1fc68f3cb29cec99c232169b) in operator watch namespace, that limits it to `openshift-storage` only.
If, I  update the value to include the `default` ns as well, then the code works fine (though, I don't think we would like to do that, or can we accommodate the `default` ns only?), but without that the operator is unable to fetch ny resources in any other ns, not necessarily `default`.

WDYT, we should be doing now?

Nikhil-Ladha:
@umangachapagain any suggestions on what to do next?

umangachapagain:
> @umangachapagain any suggestions on what to do next?
>
Can you try creating a new client for the NAD using https://github.com/k8snetworkplumbingwg/network-attachment-definition-client/tree/master/pkg/client ? It should be populated with a different cache with all namespace access.

Nikhil-Ladha:
> 

Thanks, that worked.
The changes look good now, tested out all the scenarios.

iamniting (approved):

PR #2174, opened 2023-09-04
Use client variable to make api calls to k8s cluster

Nikhil-Ladha:
Use the client.Client property of the reconciler object to make client calls instead of using the reconciler object directly.
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2231076


Nikhil-Ladha:
@iamniting @umangachapagain PTAL.

iamniting (approved):

Nikhil-Ladha:
Should we take it for previous versions as well? If so, all of them?

iamniting:
> Should we take it for previous versions as well? If so, all of them?

Nikhil, I don't think that this fix will fix the issue we want to fix here after looking at the logs provided by you. Maybe we should use a new client in the setup with the manager instead of using the existing cache client.

PR #2182, opened 2023-09-06
Make development possible on different os/arch platforms by improving the way tools/binaries are built & fetched

malayparida2000:
https://issues.redhat.com/browse/RHSTOR-4079


hack/source-manifests.sh
```diff
@@ -40,10 +40,10 @@ function dump_noobaa_csv() {
 
 	echo "Dumping Noobaa csv using command: $IMAGE_RUN_CMD --entrypoint=/usr/local/bin/noobaa-operator $NOOBAA_IMAGE $noobaa_dump_csv_cmd"
 	# shellcheck disable=SC2086
-	($IMAGE_RUN_CMD --entrypoint=/usr/local/bin/noobaa-operator "$NOOBAA_IMAGE" $noobaa_dump_csv_cmd) > $NOOBAA_CSV
+	($IMAGE_RUN_CMD --platform="${GOOS}"/"${GOARCH}" --entrypoint=/usr/local/bin/noobaa-operator "$NOOBAA_IMAGE" $noobaa_dump_csv_cmd) > $NOOBAA_CSV
```
iamniting, line 41: Does noobaa push the image for both os and arch, If not, please hardcode this value so that it does not fail if someone tries for those arches where the image is not present

hack/source-manifests.sh
```diff
@@ -58,15 +58,15 @@ function dump_rook_csv() {
 
 	crd_list=$(mktemp)
 	echo "Dumping rook csv using command: $IMAGE_RUN_CMD --entrypoint=cat $ROOK_IMAGE $rook_template_dir/$rook_csv_template"
-	$IMAGE_RUN_CMD --entrypoint=cat "$ROOK_IMAGE" $rook_template_dir/$rook_csv_template > $ROOK_CSV
+	$IMAGE_RUN_CMD --platform="${GOOS}"/"${GOARCH}" --entrypoint=cat "$ROOK_IMAGE" $rook_template_dir/$rook_csv_template > $ROOK_CSV
```
iamniting, line 59: If rook is pushing image for both arches then we can skip this change for the rook image

hack/common.sh
```diff
@@ -27,6 +27,9 @@ OPM="${LOCALBIN}/opm-${OPM_VERSION}"
 GINKGO="${LOCALBIN}/ginkgo"
 GOLANGCI_LINT_VERSION="v1.51.1"
 GOLANGCI_LINT="${LOCALBIN}/golangci-lint"
+SHELLCHECK="${LOCALBIN}/shellcheck"
+SHELLCHECK_VERSION="v0.9.0"
```
iamniting, line 31: Move the version above

iamniting (changes requested):

hack/build-metrics-exporter.sh
```diff
@@ -5,4 +5,6 @@ set -e
 source hack/common.sh
 source hack/docker-common.sh
 
-${IMAGE_BUILD_CMD} build --build-arg="LDFLAGS=${LDFLAGS}" --no-cache -f metrics/Dockerfile -t "${METRICS_EXPORTER_FULL_IMAGE_NAME}" .
+${IMAGE_BUILD_CMD} build  --no-cache \
+    --build-arg="GOOS=${GOOS}" --build-arg="GOARCH=${GOARCH}" --build-arg="LDFLAGS=${LDFLAGS}" \
```
iamniting, line 9: Can we move the build args in the last line, Pls check if it works for both docker and podman

iamniting (changes requested):

iamniting (approved):

PR #2191, opened 2023-09-20
update PROJECT file

iamniting:

### DIFF
--- a/PROJECT
+++ b/PROJECT
@@ -1,42 +1,51 @@
 domain: openshift.io
-layout: go.kubebuilder.io/v2
+layout:
+- go.kubebuilder.io/v3
+plugins:
+  manifests.sdk.operatorframework.io/v2: {}
+  scorecard.sdk.operatorframework.io/v2: {}
 projectName: ocs-operator
-repo: github.com/red-hat-storage/ocs-operator
+repo: github.com/red-hat-storage/ocs-operator/v4
 resources:
--
-  controller: true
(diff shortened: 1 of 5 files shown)


PROJECT
```diff
@@ -1,42 +1,51 @@
 domain: openshift.io
-layout: go.kubebuilder.io/v2
+layout:
+- go.kubebuilder.io/v3
```
malayparida2000, line 3: AFAIK kubebuilder.io/v3 also requires other changes in our project directory & other places.

PROJECT
```diff
@@ -1,42 +1,51 @@
...
+plugins:
+  manifests.sdk.operatorframework.io/v2: {}
+  scorecard.sdk.operatorframework.io/v2: {}
 projectName: ocs-operator
-repo: github.com/red-hat-storage/ocs-operator
+repo: github.com/red-hat-storage/ocs-operator/v4
```
malayparida2000, line 8: I think the repo link won't contain the /v4, going to that url gives a 404 error.

malayparida2000 (changes requested):

malayparida2000 (approved):
/lgtm

PROJECT
iamniting: We sorted this over a call.

PROJECT
iamniting: We sorted this over a call

PR #2200, opened 2023-09-28
Correcting the job names in prometheus metrics

aruniiird:
Earlier we decided that a few metrics will be exposed through job, 'rook-ceph-exporter'. But now the implementation has changed and 'ceph_mon_num_elections' will be the only metric produced by the job. Rest of the metrics will still be coming from 'rook-ceph-mgr' job itself.


aruniiird:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2238400

This issue seems to be show stopper for QE for testing.
@iamniting , @umangachapagain , please take a look

umangachapagain (approved):

PR #2202, opened 2023-09-28
Bug 2241113: [release-4.12]must-gather: collect events in a sorted order

yati1998:
collect events in a sorted order
This commit is to collect the events in
sorted order and also collect yamls and describe
of oc events resource


agarwal-mudit:
@iamniting can you check the failure? Looks like some issue with Nooba, have we merged something recently in 4.12?

PR #2204, opened 2023-10-01
Transforms StorageProfile array into a set of CRs

raaizik:
- Transforms `StorageProfile` array from `StorageCluster`'s spec into a set of immutable CRs
- Reconciliation of `StorageProfile`s is integrated in the `StorageCluster`'s controller
- Skipping `StorageProfile`s with an empty `DeviceClass`


api/v1/storageprofile_types.go
```diff
@@ -0,0 +1,100 @@
...
+// +kubebuilder:validation:XValidation:rule="!has(oldSelf.blockPoolConfiguration) || has(self.blockPoolConfiguration)", message="blockPoolConfiguration is required once set"
+type StorageProfileSpec struct {
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Name is immutable"
+	// +kubebuilder:validation:MaxLength=512
+	// Name of the storage profile.
+	Name string `json:"name"`
```
nb-ohad, line 39: Every k8s resource already manages a name and namespace in its metadata. It is unnecessary to manage a name ourselves

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -223,10 +223,15 @@ func (r *StorageClassRequestReconciler) reconcilePhases() (reconcile.Result, err
...
 
-	for i := range r.storageCluster.Spec.StorageProfiles {
-		profile := &r.storageCluster.Spec.StorageProfiles[i]
+	storageProfileList := &v1.StorageProfileList{}
+	if err := r.list(storageProfileList, client.InNamespace(r.OperatorNamespace)); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	for _, storageProfile := range storageProfileList.Items {
+		profile := storageProfile
```
nb-ohad, line 232:
```suggestion
	for i := range storageProfileList.Items {
		profile := &storageProfileList.Items[i]
```

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -223,10 +223,15 @@ func (r *StorageClassRequestReconciler) reconcilePhases() (reconcile.Result, err
...
 		if profile.Name == profileName {
-			r.storageProfile = profile
+			r.storageProfile = &profile
```
nb-ohad, line 234: It is preferable to not store information related to a specific request on the reconciler as it is persistent between reconcile.

I will suggest to load all storage profiles into a map key by the profile name in the first reconcile and index into it using the name on the request to retrieve the profile for each request

controllers/storagecluster/cephfilesystem.go
```diff
@@ -56,9 +56,9 @@ func (r *StorageClusterReconciler) newCephFilesystemInstances(initData *ocsv1.St
 		}
 	} else {
 		// set deviceClass and parameters from storageProfile
-		for i := range initData.Spec.StorageProfiles {
-			deviceClass := initData.Spec.StorageProfiles[i].DeviceClass
-			parameters := initData.Spec.StorageProfiles[i].SharedFilesystemConfiguration.Parameters
+		for i := range r.StorageProfiles.Items {
+			deviceClass := r.StorageProfiles.Items[i].Spec.DeviceClass
+			parameters := r.StorageProfiles.Items[i].Spec.SharedFilesystemConfiguration.Parameters
```
nb-ohad, line 61:
```suggestion
			spSpec := &r.StorageProfiles.Items[i].Spec
			deviceClass := spSpecDeviceClass
			parameters := spSpec.SharedFilesystemConfiguration.Parameters
```

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -155,6 +155,44 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		},
 	)
 
+	enqueueStorageProfileRequest := handler.EnqueueRequestsFromMapFunc(
```
nb-ohad, line 158: The name `enqueueStorageProfileRequest` does not describe the fact that we are enqueuing a StorageCluster request. Can we change it to: `enqueueFromStorageProfile`? (storagecluster is implicit as this is the controller for storagecluster CRs)

nb-ohad (changes requested):

api/v1/storageprofile_types.go
raaizik: @nb-ohad Question is is that said `name` immutable by default? Think not. Then again maybe renaming SPs can be permitted, and it's the rest of the fields' immutability that should be enforced.

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -155,6 +155,53 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		},
 	)
 
+	enqueueFromStorageProfile := handler.EnqueueRequestsFromMapFunc(
+		func(ctx context.Context, obj client.Object) []reconcile.Request {
```
leelavg, line 159:
```suggestion
		func(_ context.Context, obj client.Object) []reconcile.Request {
```
- unused context

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -155,6 +155,53 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+			spobj, ok := obj.(*ocsv1.StorageProfile)
+			if !ok {
+				return []reconcile.Request{}
+			}
+
+			// todo change according to standards
+			if spobj.Status.Phase != util.PhaseReady {
```
leelavg, line 166: I see the TODO here, question is, shouldn't we reconcile when phase is not ready generally? correct me if I'm reading this statement wrongly.

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -155,6 +155,53 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+			// Get the StorageCluster objects
+			scList := &ocsv1.StorageClusterList{}
+			err := r.List(r.ctx, scList, &client.ListOptions{Namespace: obj.GetNamespace()})
+			if err != nil {
+				r.Log.Error(err, "Unable to list StorageCluster objects")
+				return []reconcile.Request{}
+			}
+
+			// Return name and namespace per StorageProfile object
+			request := []reconcile.Request{}
```
leelavg, line 179: at the start of this func, if the arbitrary obj is of storageprofile type why not reconcile that single obj rather than every profile or it is just an insurance? correct me if I'm wrong here.

api/v1/storageprofile_types.go
nb-ohad: @raaizik `Metadata.Name` is part of a resource identifier and has special meaning for k8s. You cannot change it. As @leelavg mentioned. If you change it and try to save the resource you create a new resource, not update the old one. Also it one of the fields you can select a resource by, where the name in spec (the one you added) does not have that property

api/v1/storagecluster_types.go
```diff
@@ -537,16 +521,38 @@ const (
 	ConditionVersionMismatch conditionsv1.ConditionType = "VersionMismatch"
 )
 
-// List of constants to show different different reconciliation messages and statuses.
+// SCText stores a StorageCluster reconciliation message or status text
```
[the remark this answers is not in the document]
raaizik: @leelavg it's better practice to use enums instead of storing hard coded values, therefore IMO the `.String()` side effect is minor

api/v1/storageprofile_types.go
```diff
@@ -0,0 +1,117 @@
...
+
+// StorageProfileList contains a list of StorageProfile
+type StorageProfileList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           map[string]*StorageProfile `json:"items"`
+}
+
+// SPText stores a StorageProfile reconciliation message or status text
+type SPText int64
```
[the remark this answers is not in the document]
raaizik: @leelavg the usage of `iota` provides:
- Readability
- It carries over the type to spare re-declarations
- Prevents conflicts arising from storing duplicated values

controllers/storagecluster/storagecluster_controller.go
raaizik: Good catch, thanks

api/v1/storagecluster_types.go
```diff
@@ -537,16 +521,38 @@ const (
...
+	case ReconcileCompletedMessage:
+		return "Connected successfully to an external cluster"
+	case ExternalClusterConnected:
+		return "ExternalClusterConnected"
+	case ExternalClusterConnectedMessage:
+		return "Connected successfully to an external cluster"
+
+	}
+	return "unknown"
+}
```
nb-ohad, line 554: Regardless if I agree or disagree with the change, this has nothing to do with the storage profile CR. And if needed should be introduced in a separate unrelated PR

api/v1/storageprofile_types.go
```diff
@@ -0,0 +1,110 @@
...
+
+// StorageProfileStatus defines the observed state of StorageProfile
+type StorageProfileStatus struct {
+	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
+
+	// Phase describes the Phase of StorageProfile
+	// This is used by OLM UI to provide status information
+	// to the user
+	Phase string `json:"phase,omitempty"`
```
nb-ohad, line 59: Please see suggestion above to have this field of type `StorageProfilePhase` instead of a plain old string

api/v1/storageprofile_types.go
```diff
@@ -0,0 +1,110 @@
...
+		return "StorageProfileReconcileFailed"
+	case SPInit:
+		return "StorageProfileInit"
+	case SPReconcileCompleted:
+		return "StorageProfileReconcileCompleted"
+	case SPReconcileCompletedMessage:
+		return "StorageProfile reconcile completed successfully"
+	}
+	return "unknown"
+}
```
nb-ohad, line 106: I can see that there are two concerns coupled here. A Phase value and some text messages. To me, it does not make sense to treat both the same. Messages are strings by nature, while phase can be argued to be an enum (but can be a string as well). Let's keep the 2 concerns separate.

A couple of other remarks:
-  We need to align with the established terminology for Phases. So "Ready" instead of "Completed"
- The actual string values for phases should not include the word `Reconcile` in them (the enum token can) 
- Unknown is not a phase, if you need to keep the field empty just keep it empty
- No reason to have a message that only appears once in the code as a cost (especially as it is a one-time message, not a predefined value)
- For outside facing value (like the status of CR) we should use fall names and not abbreviations or short names, so the text `init` should change to `initializing`
- switch cases have a default case, no need for an explicit return 
- With auto-complete in all of our editor available for us there is no need to got with shorting names, it just make the code less readable (SP prefix -> StorageProfile prefix) 
- Why does an external cluster have to do with storage profile phase/status?

```
const (
	StorageProfilePhase_Init = iota
	StorageProfilePhase_Failed      
	StorageProfilePhase_Ready
)

func (s SCText) StorageProfilePhaseToString() string {
	switch s {
	case StorageProfilePhaseInit:
		return "Initializing"
	case StorageProfilePhaseFailed:
		return "Failed"
	case StorageProfilePhaseReady:
		return "Ready"
	default:
	        return ""
	}
}
```

It also might be that a string subtype will work better than an enum as it doesn't need a conversion from int to string and is still different than just a plain old string and can be defined in the `StorageProfileTypes.go` file as the type of the phase field.

```
// StorageProfilePhase specifies a phase a storage profile reconciliation 
type StorageProfileString string

const (
	StorageProfilePhase_Init  StorageProfileString = "Initializing`
	StorageProfilePhase_Failed  StorageProfileString = Failed
	StorageProfilePhase_Ready StorageProfileString = Ready
)
```

Both approaches works for me

controllers/ocsinitialization/ocsinitialization_controller.go
nb-ohad, line 1: This changes have nothing to do with storage profiles and if needed should be introduced as a separate PR

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -223,14 +223,13 @@ func (r *StorageClassRequestReconciler) reconcilePhases() (reconcile.Result, err
...
-		if profile.Name == profileName {
-			r.storageProfile = profile
-			break
-		}
+	storageProfileList := &v1.StorageProfileList{}
+	if err := r.list(storageProfileList, client.InNamespace(r.storageCluster.GetNamespace())); err != nil {
+		return reconcile.Result{}, err
 	}
 
+	r.storageProfile = storageProfileList.Items[profileName]
```
nb-ohad, line 231: As we discussed in an offline review. The reconciler object state is preserved across calls. As such it does not make sense to save anything on the reconciler that is specific to a single CR instance as it can negatively affect the reconciliation of the next event (for a different `StorageClassRequest` instance)

There are 3 approaches that can be used here instead:
1. Load all of the Profiles into a map with the Metadata.Name as key 
2. Load only the named profile (use `r.get` instead of `r.list`) as a variable and  just pass it into the reconcile functions as parameters
3. Load only the named profile and add it to the context object using `ctx.WithValue` 

For this use case, my recommendation is to go with option 2, it is very straight forward, very readable and because all of out reconcile functions are called from reconcile phases it will not add complexity to the codebase

controllers/storagecluster/cephfilesystem.go
```diff
@@ -41,7 +42,17 @@ func (r *StorageClusterReconciler) newCephFilesystemInstances(initData *ocsv1.St
...
-	if initData.Spec.StorageProfiles == nil {
+	// Load StorageProfile list
+	// Get StorageProfile list in the StorageCluster's ns
+	spList := &ocsv1.StorageProfileList{}
+	spErr := r.Client.List(r.ctx, spList, &client.ListOptions{Namespace: initData.GetNamespace()})
+	if spErr != nil {
+		r.Log.Error(spErr, "Unable to list StorageProfile objects")
+	}
+	// Save StorageProfile list to be reconciled
+	r.StorageProfiles = *spList
```
nb-ohad, line 53: You don't need an intermediate list( and a copy operation), just list into the target variable.
```suggestion
	spErr := r.Client.List(r.ctx, &r.StorageProfiles, &client.ListOptions{Namespace: initData.GetNamespace()})
	if spErr != nil {
		r.Log.Error(spErr, "Unable to list StorageProfile objects")
	}
```

controllers/storagecluster/cephfilesystem.go
```diff
@@ -41,7 +42,17 @@ func (r *StorageClusterReconciler) newCephFilesystemInstances(initData *ocsv1.St
...
+
+	if len(r.StorageProfiles.Items) == 0 {
```
nb-ohad, line 55: I think this is the wrong condition. A Lack of storage profiles does not mean that we are at a standalone deployment. Actually all of our deployments will start with no storage profile and we should not run the code in this then section for provider mode deployments

controllers/storagecluster/reconcile.go
nb-ohad, line 1: This changes have nothing to do with storage profiles and if needed should be introduced as a separate PR

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -92,6 +92,7 @@ type StorageClusterReconciler struct {
 	recorder           *util.EventReporter
 	OperatorCondition  conditions.Condition
 	IsNoobaaStandalone bool
+	StorageProfiles    ocsv1.StorageProfileList
```
nb-ohad, line 95: I think it would be preferable to have this as a pointer and not a value type. It will enable you to load into it without a need for the value (and the entire substruct to be copied). It will also enable us to verify if it was loaded or not using a `nil` value check

```suggestion
	StorageProfiles    *ocsv1.StorageProfileList
```

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -155,6 +156,53 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+						NamespacedName: types.NamespacedName{
+							Namespace: sp.Namespace,
+							Name:      sc.Name,
+						},
+					})
+				}
+
+			}
+
+			return request
```
nb-ohad, line 202: Why do we need a for loop if we know that we are only listing a single item? (the list op should contain a limitation for a single result). Also, why are we listing storage profiles? the event already is for a specific storage profile. 

In addition, this code will re-queue multiple SC events for a single SP event. That just does not make sense. The mapping should be 1-1:

```suggestion
			return []reconcile.Request {
				reconcile.Request {
						NamespacedName: sc.NamespacedName
				},
			}
```

nb-ohad (changes requested):

api/v1/storageprofile_types.go
leelavg: With iota and switch I think you might need (un)marshal method as well or else it'll take base type 🤔.

controllers/storagecluster/storagecluster_controller.go
raaizik: In that case what differentiates a plain `StorageCluster` event from a `StorageProfile` triggered one? Seems to me as though your suggestion implies we're allowing duplicated events (same identifier) in the queue.

controllers/storagecluster/cephfilesystem.go
raaizik: Not accurate. Upon initialization there exists the `defaultStorageProfile`. Meaning the existence of a `StorageCluster` resource necessarily implies that there is at least that one `StorageProfile`. Might be wrong.

controllers/storagecluster/storagecluster_controller.go
raaizik: Reverted `Items` field since to my understanding `List()` successfully retrieves the objects only when provided with plain arrays, so the for loops are back.
```
type StorageProfileList struct {
	metav1.TypeMeta `json:",inline"`
	metav1.ListMeta `json:"metadata,omitempty"`
	Items           []StorageProfile `json:"items"`
}
```

api/v1/storageprofile_types.go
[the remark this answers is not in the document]
raaizik: @leelavg NVM, refactored w/o it

api/v1/storagecluster_types.go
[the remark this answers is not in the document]
raaizik: @leelavg NVM, refactored

controllers/storagecluster/storagecluster_controller.go
raaizik: Removed field

controllers/storagecluster/storagecluster_controller.go
raaizik: Changed

controllers/storagecluster/cephfilesystem.go
raaizik: Changed

api/v1/storageprofile_types.go
raaizik: Refactored

controllers/storagecluster/storagecluster_controller.go
raaizik: I've made phase updates where I saw fit. I didn't see a need to add `[]conditionsv1.Condition` field to `StorageProfileStatus`.

controllers/storagecluster/storagecluster_controller.go
raaizik: Changed

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -20,6 +18,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/version"
 	"k8s.io/client-go/kubernetes"
+	"os"
```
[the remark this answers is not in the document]
raaizik: Removed

controllers/storagecluster/cephfilesystem.go
```diff
@@ -19,59 +20,73 @@ type ocsCephFilesystems struct{}
...
-		// overridden.
-		ret.Spec.DataPools = []cephv1.NamedPoolSpec{
-			{
-				PoolSpec: cephv1.PoolSpec{
-					DeviceClass:   generateDeviceClass(initData),
-					Replicated:    generateCephReplicatedSpec(initData, "data"),
-					FailureDomain: initData.Status.FailureDomain,
+	// Load all StorageProfile objects in the StorageCluster's namespace
+	storageProfiles := &ocsv1.StorageProfileList{}
+	spErr := r.Client.List(r.ctx, storageProfiles, &client.ListOptions{Namespace: initStorageCluster.GetNamespace()})
```
leelavg, line 47: - I don't see a reason for different name
```suggestion
	err := r.Client.List(r.ctx, storageProfiles, &client.ListOptions{Namespace: initStorageCluster.GetNamespace()})
```

api/v1/storageprofile_types.go
```diff
@@ -0,0 +1,103 @@
...
+	StorageProfilePhaseReady  StorageProfilePhase = "Ready"
+)
+
+// StorageProfileMessage stores a StorageProfile reconciliation message
+type StorageProfileMessage string
+
+const (
+	StorageProfilePhaseReadyMessage StorageProfileMessage = "StorageProfile reconcile completed successfully"
+	StorageProfileNotFound          StorageProfileMessage = "no storage profile definition found for storage profile"
+	StorageProfileListError         StorageProfileMessage = "Unable to list StorageProfile objects"
```
leelavg, line 98: - if this is being used as an error it shouldn't be of type `StorageProfileMessage`
- it should be a error value and `errors.Is` can be used to compare the values, something like
``` go
import (
  "errors"
)

var errList = errors.New("Listing")

// some sample func
func do() {
  err := func() error { return errList }()
  if errors.Is(err, errList) { // do what we want for this specific error }
}
```

controllers/storagecluster/reconcile.go
```diff
@@ -448,9 +448,11 @@ func (r *StorageClusterReconciler) reconcilePhases(
 			statusutil.SetErrorCondition(&instance.Status.Conditions, reason, message)
 
 			if instance.Status.Phase != statusutil.PhaseOnboarding {
-				// if the error was due to skipped storage classes
+				// if the error was due to skipped storage classes or
+				// due to StorageProfile list retrieval failure
 				// instead of error phase we will set it to progressing
-				if strings.Contains(returnErr.Error(), storageClassSkippedError) {
+				if strings.Contains(returnErr.Error(), storageClassSkippedError) ||
```
leelavg, line 454: - IMHO, mixing errors and strings isn't a good practice
- Not to refactor existing code, can we at least not follow this in this context and compare errors with `errors.Is`

api/v1/storageprofile_types.go
```diff
@@ -0,0 +1,103 @@
...
+const (
+	StorageProfilePhaseFailed StorageProfilePhase = "Failed"
+	StorageProfilePhaseInit   StorageProfilePhase = "Initializing"
+	StorageProfilePhaseReady  StorageProfilePhase = "Ready"
+)
+
+// StorageProfileMessage stores a StorageProfile reconciliation message
+type StorageProfileMessage string
+
+const (
```
leelavg, line 95: - after a full pass, I realized this message isn't for CR status and as such I didn't feel the need to have this based on it's current usage (only logging)

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -155,6 +155,35 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		func(_ context.Context, obj client.Object) []reconcile.Request {
+			spobj, ok := obj.(*ocsv1.StorageProfile)
+			if !ok {
+				spobj.Status.Phase = ocsv1.StorageProfilePhaseFailed
+				return []reconcile.Request{}
+			}
+
+			// Get the StorageCluster object
+			scList := &ocsv1.StorageClusterList{}
+			err := r.Client.List(r.ctx, scList, &client.ListOptions{Namespace: obj.GetNamespace(), Limit: 1})
```
leelavg, line 168: @nb-ohad I see this block was changed due to your comments, as there are talks to introduce multiple StorageClasses we are creating a hook for ourselves to fix this at a later point of time, wdyt?

config/rbac/storageprofile_editor_role.yaml
```diff
@@ -0,0 +1,24 @@
+# permissions for end users to edit storageprofiles.
```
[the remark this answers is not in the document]
raaizik: @leelavg The `StorageProfile` CR is immutable.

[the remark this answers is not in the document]
raaizik: @leelavg first and foremost, now that it is a standalone CR that's being `Watch`ed by a `StorageCluster`, while without its own controller, it still should have an attached RBAC manifest. 
Second, the `StorageProfile` CR is meant to be immutable, meaning only the following verbs should be included:
```
  verbs:
  - create
  - delete
  - get
  - list
  - watch
```

controllers/storagecluster/cephfilesystem.go
raaizik: @leelavg there's already a variable with the name `err` in this [block](https://github.com/raaizik/ocs-operator/blob/dba040ca8f756a6d8e49c7c1596d75272feb9811/controllers/storagecluster/cephfilesystem.go#L89).

api/v1/storageprofile_types.go
raaizik: @leelavg if you take a closer look, it's being used for other stuff that are crucial.

controllers/storagecluster/storagecluster_controller.go
raaizik: @leelavg you mean `StorageCluster`s. We are aware of that and it was decided to postpone taking it into account for this PR, so it's for now out of scope.

api/v1/storageprofile_types.go
```diff
@@ -0,0 +1,103 @@
...
+	StorageProfilePhaseReadyMessage StorageProfileMessage = "StorageProfile reconcile completed successfully"
```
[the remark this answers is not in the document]
raaizik: @leelavg This message isn't currently logged or used at all, but I thought to keep for when we'll introduce a state message for `StorageProfile`s, which presently as mentioned I didn't see a need to.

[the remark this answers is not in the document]
leelavg: > This message isn't currently logged or used at all,
- yes, I already checked that

config/rbac/storageprofile_editor_role.yaml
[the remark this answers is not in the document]
leelavg: > now that it is a standalone CR  that's being `Watch`ed
- doesn't matter whether it's standalone or not, it is just any other resource and the operator needs to have RBAC for accessing it
- correct me if I missed `kubebuilder` flags in your code for generating RBAC, something like
https://github.com/red-hat-storage/ocs-operator/blob/70fe874aa7654a7294a8278f0f5fece082241020/controllers/storagecluster/reconcile.go#L110-L131

controllers/storagecluster/cephfilesystem.go
leelavg: ack, anyways we are also touching that line, maybe have `err` at both places.

api/v1/storageprofile_types.go
leelavg: I'm aware of that, this comment should be read in conjunction with other comment mentioning comparison of strings as errors.

controllers/storagecluster/storagecluster_controller.go
leelavg: got it.

config/rbac/storageprofile_editor_role.yaml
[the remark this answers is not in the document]
raaizik: @leelavg oh, it's covered [here](https://github.com/raaizik/ocs-operator/blob/dba040ca8f756a6d8e49c7c1596d75272feb9811/controllers/storagecluster/reconcile.go#L110) under: 
```go
// +kubebuilder:rbac:groups=ocs.openshift.io,resources=*,verbs=get;list;watch;create;update;patch;delete
``` 
`StorageProfile` resource was built to belong to the `ocs.openshift.io` [group](https://github.com/raaizik/ocs-operator/blob/8b5aea747034bf9548238dfb080c62f9e0d935d9/config/crd/bases/ocs.openshift.io_storageprofiles.yaml#L10)
That's going to permit `update`s and `patch`es 💭. That being said, it makes perfect sense to keep those resources under the same group.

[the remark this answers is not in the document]
leelavg: Ah! Thanks for pointing it out. Reg updates/patches, I believe the validation that you added to CR should be enough, isn't it?

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -155,6 +155,35 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		},
 	)
 
+	enqueueFromStorageProfile := handler.EnqueueRequestsFromMapFunc(
+		func(_ context.Context, obj client.Object) []reconcile.Request {
+			spobj, ok := obj.(*ocsv1.StorageProfile)
+			if !ok {
+				spobj.Status.Phase = ocsv1.StorageProfilePhaseFailed
```
Madhu-1, line 162: if the object is not of time `ocsv1.StorageProfile` how can we set the status?

config/rbac/storageprofile_editor_role.yaml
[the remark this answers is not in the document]
raaizik: Update: it works:
```
# storageprofiles.ocs.openshift.io "fast-performance-storageprofile" was not valid:
# * spec.deviceClass: Invalid value: "string": DeviceClass is immutable
```

api/v1/storageprofile_types.go
```diff
@@ -0,0 +1,103 @@
...
+// +kubebuilder:validation:XValidation:rule="!has(oldSelf.blockPoolConfiguration) || has(self.blockPoolConfiguration)", message="blockPoolConfiguration is required once set"
+type StorageProfileSpec struct {
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="DeviceClass is immutable"
+	// +kubebuilder:validation:MaxLength=512
+	// DeviceClass is the deviceclass name.
+	DeviceClass string `json:"deviceClass"`
```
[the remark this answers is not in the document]
raaizik: I followed the guide [here](https://kubernetes.io/blog/2022/09/29/enforce-immutability-using-cel/#immutablility-after-first-modification):

> To enforce a field's immutability after its first write, you need to apply the following constraints:
Field must be allowed to be initially unset `+kubebuilder:validation:Optional`

controllers/storagecluster/cephfilesystem.go
```diff
@@ -19,59 +20,73 @@ type ocsCephFilesystems struct{}
...
+			}
 		}
 	} else {
 		// set deviceClass and parameters from storageProfile
-		for i := range initData.Spec.StorageProfiles {
-			deviceClass := initData.Spec.StorageProfiles[i].DeviceClass
-			parameters := initData.Spec.StorageProfiles[i].SharedFilesystemConfiguration.Parameters
+		for i := range storageProfiles.Items {
+			storageProfile := storageProfiles.Items[i]
+			storageProfile.Status.Phase = ocsv1.StorageProfilePhaseInit
```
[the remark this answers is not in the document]
raaizik: Basically this is the placement where the initialization of the `StorageProfile`s occurs. I didn't see anywhere else to place it.

controllers/storagecluster/storagecluster_controller.go
raaizik: But the object is of type `ocsv1.StorageProfile`. Can you clarify?

raaizik: But the object is of type `ocsv1.StorageProfile`. The `StorageProfile` resource is being `watch`ed by `StorageCluster`.

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -155,6 +155,35 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+
+			// Get the StorageCluster object
+			scList := &ocsv1.StorageClusterList{}
+			err := r.Client.List(r.ctx, scList, &client.ListOptions{Namespace: obj.GetNamespace(), Limit: 1})
+			if err != nil {
+				r.Log.Error(err, "Unable to list StorageCluster objects")
+				return []reconcile.Request{}
+			}
+
+			sc := scList.Items[0]
```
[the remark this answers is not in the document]
raaizik: If it reaches this point then I think it's safe to say it is a non-empty list:
> `List` retrieves list of objects for a given namespace and list options. On a successful call, `Items` field in the list will be populated with the result returned from the server.

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -358,10 +357,12 @@ func (r *StorageClassRequestReconciler) reconcileCephBlockPool() error {
 
 	r.setCephResourceStatus(r.cephBlockPool.Name, "CephBlockPool", phase, cephClients)
 
+	storageProfile.Status.Phase = v1.StorageProfilePhaseReady
```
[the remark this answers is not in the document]
raaizik: I initially did not see a need to use `[]conditionsv1.Condition` field for `StorageProfileStatus`, if that's what you're referring to. Although now that you mention it there might be.

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -223,16 +222,16 @@ func (r *StorageClassRequestReconciler) reconcilePhases() (reconcile.Result, err
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      profileName,
+			Namespace: r.storageCluster.Namespace,
+		},
 	}
 
-	if r.storageProfile == nil {
-		return reconcile.Result{}, fmt.Errorf("no storage profile definition found for storage profile %s", profileName)
+	if err := r.get(&storageProfile); err != nil {
+		return reconcile.Result{}, fmt.Errorf("%s: %s", v1.StorageProfileNotFound, profileName)
```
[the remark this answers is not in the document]
raaizik: I thought it can apply to all of them. So just a generic `StorageProfileError` will do?

api/v1/storagecluster_types.go
```diff
@@ -109,22 +109,6 @@ type StorageClusterSpec struct {
 	// DefaultStorageProfile is the default storage profile to use for
 	// the storageclassrequest as StorageProfile is optional.
 	DefaultStorageProfile string `json:"defaultStorageProfile,omitempty"`
```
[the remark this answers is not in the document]
nb-ohad: @Madhu-1 What happens if more then 1 profile is marked with the annotation?

api/v1/storageprofile_types.go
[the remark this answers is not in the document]
nb-ohad: @raaizik So, in the case where a storage class was created without a device set, can you please answer the following questions:

- what will be the behavior of a storage profile reconciliation? 
- which device class will be used? 
- Are we going to reject this storage profile? 
- If we are rejecting, are we going to reflect that in the status section of the profile?

controllers/storageclassrequest/storageclassrequest_controller.go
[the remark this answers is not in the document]
nb-ohad: Why are we updating the status of the `storageProfile` in the `storageClassRequest` controller?
A CR status needs to be updated by the controller that reconciles the desired state (as presented in the CR) to the actual state. In the case of storage profiles that would be the responsibility of the `StorageCluster`'s controller!

The fact that we were able to reconcile the request does not mean that we reconciled the resources that need to be reconciled in order to fulfill this specific storage profile. 

Also, this update is NOT persistent as this controller never saves the storage profile CR status (which is the correct behavior as this is not the storage profile controller). And the line is basically a no-op

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -422,6 +423,8 @@ func (r *StorageClassRequestReconciler) reconcileCephFilesystemSubVolumeGroup()
 
 	r.setCephResourceStatus(r.cephFilesystemSubVolumeGroup.Name, "CephFilesystemSubVolumeGroup", phase, cephClients)
 
+	storageProfile.Status.Phase = v1.StorageProfilePhaseReady
```
nb-ohad, line 426: Same as my last comment about updating the status of the storage profile CR in this controller.

controllers/storagecluster/cephfilesystem.go
nb-ohad, line 47: Can we please use the standard `client` abstraction for options?

```suggestion
	spErr := r.Client.List(r.ctx, storageProfiles, client.InNamespace(initStorageCluster.GetNamespace()))
```

controllers/storagecluster/cephfilesystem.go
```diff
@@ -19,59 +20,73 @@ type ocsCephFilesystems struct{}
...
-				PoolSpec: cephv1.PoolSpec{
-					DeviceClass:   generateDeviceClass(initData),
-					Replicated:    generateCephReplicatedSpec(initData, "data"),
-					FailureDomain: initData.Status.FailureDomain,
+	// Load all StorageProfile objects in the StorageCluster's namespace
+	storageProfiles := &ocsv1.StorageProfileList{}
+	spErr := r.Client.List(r.ctx, storageProfiles, &client.ListOptions{Namespace: initStorageCluster.GetNamespace()})
+	if spErr != nil {
+		r.Log.Error(spErr, fmt.Sprint(ocsv1.StorageProfileListError))
+		return nil, fmt.Errorf("%s", ocsv1.StorageProfileListError)
```
nb-ohad, line 50: This return here is wrong as it has the potential to break ALL non-provider mode deployments. 
Can you please move this list and error check inside the condition that applies only to provider mode deployments?

controllers/storagecluster/cephfilesystem.go
```diff
@@ -19,59 +20,73 @@ type ocsCephFilesystems struct{}
...
+	}
+
+	if storageProfiles == nil {
```
nb-ohad, line 53: This needs to be removed because we need the provider mode logic to run even if we don't find any storage profiles. Mainly as error handling for storage profile reconciliation (including not finding any) should go inside the if that capture the provider mode

controllers/storagecluster/cephfilesystem.go
nb-ohad, line 73: This line is meaningless, as we never persist in the change. If the intent was to mark the fact that we started the reconciliation of a storage profile then we need to persist that change by updating the storage profile CR in the k8s (etcd) database.

Regardless, our reconciliation process is not one that is stretched across multiple reconcile loops (because we always either succeed or fail), so an initializing phase does not make sens as it is guaranteed to be overwritten by the time we exist the reconciliation code. 

I would suggest to remove this phase from the code and from the API (types)

controllers/storagecluster/reconcile.go
```diff
@@ -448,9 +448,11 @@ func (r *StorageClusterReconciler) reconcilePhases(
...
+					strings.Contains(returnErr.Error(), fmt.Sprint(ocsv1.StorageProfileListError)) {
```
nb-ohad, line 455: I believe this addition to be wrong, the point was to skip only if we get a skip error not every error from the reconciliation. If you want an error to be ignored you will have to return a skip error.

controllers/storagecluster/storagecluster_controller.go
nb-ohad: @Madhu-1 I agree with @raaizik on this. If we run this code on a non-sp even, the code should blow as it means that somebody made a coding error not a runtime error. Blowing the code up will make it easier to focus on the fact that you are trying to use a handler that was designed for an SP object on a watch that watches another type of object.

controllers/storagecluster/storagecluster_controller.go
nb-ohad: @leelavg There cannot be two storagecluster CRs in the same namespace and the SP design right now assumes that the SP is related to the SC in the same namespace!

nb-ohad, line 168: Can we please use the abstraction `client` provides over ListOptions?

  ```suggestion
			err := r.Client.List(r.ctx, scList, &client.InNamespace(obj.GetNamespace()).Limit(1))
```

File: controllers/storagecluster/storagecluster_controller.go (hunk 30 above)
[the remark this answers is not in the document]
Comment (nb-ohad) [reply controllers/storagecluster/storagecluster_controller.go:174]: @raaizik I am not sure I agree. It can be that somebody created the SP CR before creating the SC CR. It is a valid configuration as the later creation of the CS CR will guarantee the reconciliation of the  SP CR as well (as they are reconciled by the same controller). 

We can avoid the issue by returning an empty reconcile request list.

```suggestion
			if len(scList) == 0 {
				return []reconcile.Request{}
			}
						
			sc := scList.Items[0]
```

nb-ohad (changes requested):

controllers/storageclassrequest/storageclassrequest_controller.go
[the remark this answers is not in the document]
raaizik: @nb-ohad `v1.StorageProfileNotFound` is not of type string, hence the formatting `%s`.

controllers/storagecluster/reconcile.go
raaizik: This was made to address the missing remediation in a scenario of a failed `StorageCluster` reconciliation due to a failure of some `StorageProfile` or the set of `StorageProfile`s, though in this particular flow apart from the listing error, there aren't any other designated `StorageProfile` errors to check for. Does this equate to _"every error from the reconciliation"_? not sure. It's an added condition to allow re-queuing of the event, meaning nothing but an internal implementation detail, so why does the exact content of its underlying string matter?

controllers/storagecluster/storagecluster_controller.go
raaizik: This is the correct usage:
```suggestion
opts := &client.ListOptions{Limit: 1}
client.InNamespace(obj.GetNamespace()).ApplyToList(opts)
err := r.Client.List(r.ctx, scList, opts)
```

raaizik: Resolving as per @leelavg's comment

controllers/storageclassrequest/storageclassrequest_controller.go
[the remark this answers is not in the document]
nb-ohad: @raaizik All these messages are of type derived from string. 
In this comment https://github.com/red-hat-storage/ocs-operator/pull/2204#discussion_r1359843888 I explain why it does not make sense and even here you can see that it just makes usage more compcomplicated for no reason. 

The bottom line is that these are string messages, and there is no reason to subtype them and lose the string functionality, as subtyping does not add any value to these.

controllers/storagecluster/reconcile.go
nb-ohad: > This was made to address the missing remediation in a scenario of a failed StorageCluster reconciliation due to a failure of some StorageProfile or the set of StorageProfiles, though in this particular flow apart from the listing error, there aren't any other designated StorageProfile errors to check for.

But in the case of storage profile reconciliation error you need to continue reconciliation for that exact position in the code, not to jump one level up and more to the next ensure. Assuming that you agree with my analysis the correct way to handle it is to just not return in case of an error on the code that reconciles storage profiles. just log the message and move on.

controllers/storagecluster/storagecluster_controller.go
nb-ohad: I did make a mistake in my suggestion but you also don't need to run the `ApplyToList` call.
The clean way of using the client abstraction is to have bench one called and passed as a separate param

```suggestion
			err := r.Client.List(r.ctx, scList, &client.Inamespace(obj.GetNamespace()), client.Limit(1))
```

api/v1/storagecluster_types.go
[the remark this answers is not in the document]
Madhu-1: if we have one or more profile, we need to take a decision on what to do, i was comparing this to the storageclass, If we have 2 default storageclasses, nothing will be chosen when PVC is created without storageclass.

controllers/storagecluster/storagecluster_controller.go
Madhu-1: This check will be hit when the typecasting doesn't happen right? I am just going by code reading.

api/v1/storagecluster_types.go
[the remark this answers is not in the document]
nb-ohad: @Madhu-1 Why push ourselves into that corner when we can just have the default be selected by the storage cluster without ambiguity?

controllers/storagecluster/storagecluster_controller.go
nb-ohad: Ok, I understand now. I expected that a type assertion without the `ok` params will be used. That will guarantee that the process panics if the assertions on the code are not correct. In this case, I think that the `ok` param and the `if` are unnecessary and even harmful. Mainly because we want the code to panic if someone has used the watch on the wrong type. What we do not want is for the code to continue to run with no side effect.

@raaizik Can we have the change implemented?

controllers/storagecluster/cephfilesystem.go
```diff
@@ -3,6 +3,7 @@ package storagecluster
 import (
 	"context"
 	"fmt"
+	"sigs.k8s.io/controller-runtime/pkg/client"
```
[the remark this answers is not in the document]
raaizik: Did that, nothing's changed

api/v1/storageprofile_types.go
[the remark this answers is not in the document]
raaizik: Currently this kind of `StorageProfile` will be appended with no `deviceClass`
```
% oc get storageprofile
NAME                                AGE
empty-performance-storageprofile    38m
fast-performance-storageprofile     100m
medium-performance-storageprofile   40m
slow-performance-storageprofile     40m
``` 
`empty-performance-storageprofile`'s spec:
```
Spec:
  Block Pool Configuration:
    Parameters:
      pg_autoscale_mode:  on
      pg_num:             128
      pgp_num:            128
  Shared Filesystem Configuration:
    Parameters:
      pg_autoscale_mode:  on
      pg_num:             128
      pgp_num:            128
Events:                   <none>
``` 
```
Spec:
  Data Pools:
    Device Class:  fast
    Erasure Coded:
      Coding Chunks:  0
      Data Chunks:    0
    Failure Domain:   zone
    Mirroring:
    Name:  fast
    Parameters:
      pg_autoscale_mode:  on
      pg_num:             128
      pgp_num:            128
    Quotas:
    Replicated:
      Replicas Per Failure Domain:  1
      Size:                         3
      Target Size Ratio:            0.49
    Status Check:
      Mirror:
    Device Class:  medium
    Erasure Coded:
      Coding Chunks:  0
      Data Chunks:    0
    Failure Domain:   zone
    Mirroring:
    Name:  medium
    Parameters:
      pg_autoscale_mode:  on
      pg_num:             128
      pgp_num:            128
    Quotas:
    Replicated:
      Replicas Per Failure Domain:  1
      Size:                         3
      Target Size Ratio:            0.49
    Status Check:
      Mirror:
    Device Class:  slow
    Erasure Coded:
      Coding Chunks:  0
      Data Chunks:    0
    Failure Domain:   zone
    Mirroring:
    Name:  slow
    Parameters:
      pg_autoscale_mode:  on
      pg_num:             128
      pgp_num:            128
    Quotas:
    Replicated:
      Replicas Per Failure Domain:  1
      Size:                         3
      Target Size Ratio:            0.49
    Status Check:
      Mirror:
    Erasure Coded:
      Coding Chunks:  0
      Data Chunks:    0
    Failure Domain:   zone
    Mirroring:
    Parameters:
      pg_autoscale_mode:  on
      pg_num:             128
      pgp_num:            128
``` 

Once a `deviceClass` is set, the reconciliation appears to fail:
```
  Warning  ReconcileFailed     28s                    rook-ceph-file-controller  failed to reconcile CephFilesystem "openshift-storage/ocs-storagecluster-cephfilesystem". failed to create filesystem "ocs-storagecluster-cephfilesystem": failed to create filesystem "ocs-storagecluster-cephfilesystem": failed to set pools size: failed to update datapool  "ocs-storagecluster-cephfilesystem-empty": failed to create replicated crush rule "ocs-storagecluster-cephfilesystem-empty": failed to create crush rule ocs-storagecluster-cephfilesystem-empty: exit status 22
```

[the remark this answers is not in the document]
nb-ohad: Ok, so the plan is to just reject this kind of storage profile and not take it into consideration when new requests come up. correct?

controllers/storagecluster/storagecluster_controller.go
raaizik: Done

api/v1/storageprofile_types.go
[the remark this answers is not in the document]
raaizik: Note that the status is set to blank in case of then setting the `deviceClass`, which pretty much leaves the two other phases unused. IMO not the most optimal thing to have, but I did not see spots where such updates are required.

leelavg (approved):

api/v1/storagecluster_types.go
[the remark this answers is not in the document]
Madhu-1: That's true, I was trying to keep both resources independent. If something needs to be set to default it need to be updated in the storageCluster CR not on the StorageProfile CR, this can be marked as resolved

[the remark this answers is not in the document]
nb-ohad: Glad we got to a resolution

controllers/storagecluster/cephfilesystem.go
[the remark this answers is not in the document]
nb-ohad: I don't believe that `fmt` knows where each dependency needs to be placed. It just orders them within the same block of dependency. @raaizik Please move the dependency line to the next block (add a black line above it and remove the one below it) and then run `fmt`

leelavg (approved):

iamniting:
I apologize for joining the conversation later; I was occupied with other matters. I kindly request that the approvers also consider the commit messages when reviewing PRs. It's essential that we refrain from approving PRs when the commit messages are not tidy and well-organized.

leelavg:
Hey, I did check the PR description, seems I missed commit message along the way.

@raaizik could you pls clarify the situation if possible?

nb-ohad:
@leelavg Changing a merged commit is not advisable. Let's not go there. 
We will align with what ever is required in the commit message starting with the next commit

leelavg:
No, I'm not suggesting to change merged commit, I'm asking about what happened to the said fix in the commit msg.

PR #2211, opened 2023-10-11
Push Noobaa reconciler to the end

rohan47:
Noobaa db pod gets stuck in pending state waiting for volume from ceph-rbd storage class which is not available till storageclient is created on the base cluster. StorageClient is not able to connect until Noobaa is disabled and re-enabled. Moving noobaa reconciler at the last solves this issue.

https://bugzilla.redhat.com/show_bug.cgi?id=2243101

### DIFF
--- a/controllers/storagecluster/reconcile.go
+++ b/controllers/storagecluster/reconcile.go
@@ -398,11 +398,11 @@ func (r *StorageClusterReconciler) reconcilePhases(
 				&ocsCephObjectStoreUsers{},
 				&ocsCephRGWRoutes{},
 				&ocsStorageClass{},
-				&ocsNoobaaSystem{},
 				&ocsSnapshotClass{},
 				&ocsJobTemplates{},
 				&ocsCephRbdMirrors{},
 				&ocsClusterClaim{},
+				&ocsNoobaaSystem{},
 			}
 		} else {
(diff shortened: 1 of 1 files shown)


controllers/storagecluster/reconcile.go
```diff
@@ -398,11 +398,11 @@ func (r *StorageClusterReconciler) reconcilePhases(
 				&ocsCephObjectStoreUsers{},
 				&ocsCephRGWRoutes{},
 				&ocsStorageClass{},
-				&ocsNoobaaSystem{},
 				&ocsSnapshotClass{},
 				&ocsJobTemplates{},
 				&ocsCephRbdMirrors{},
 				&ocsClusterClaim{},
+				&ocsNoobaaSystem{},
```
umangachapagain, line 405: Looks like NooBaa was already being reconciled towards the end i.e. after its pre-requisites are met.
Since we don't reconcile StorageClient here and can't control when the first StorageClient gets created, how can we be sure that this isn't working just due to a race condition?

rohan47: Will test again and compare the logs with and without this commit

rohan47:
This PR is no longer required, not facing the issue

PR #2212, opened 2023-10-12
Syncing latest changes from main for ocs-operator

df-build-team:
PR containing the latest commits from main branch

### DIFF
--- a/metrics/internal/collectors/registry.go
+++ b/metrics/internal/collectors/registry.go
@@ -87,7 +87,7 @@ func enableCephBlocklistMirrorStore(opts *options.Options) {
 	go reflector.Run(opts.StopCh)
 }
 
-// RegisterPersistentVolumeAttributesCollector registers PV attribute colletor to registry
+// RegisterPersistentVolumeAttributesCollector registers PV attribute collector to registry
 func RegisterPersistentVolumeAttributesCollector(registry *prometheus.Registry, opts *options.Options) {
 	if !pvStoreEnabled {
 		enablePVStore(opts)


malayparida2000:
@Nikhil-Ladha Can we do something so that the PRs getting raised contain the date range of which changes it's trying to merge? For Ex- Syncing latest changes from main for ocs-operator[DD:MM:YYYY - DD:MM:YYYY]
Or else there will be so many PRs with the exact same name pattern Syncing the latest changes.
@iamniting WDYT?

iamniting:
@Nikhil-Ladha Is it possible to drop the merge commits?

Nikhil-Ladha:
> @Nikhil-Ladha Can we do something so that the PRs getting raised contain the date range of which changes it's trying to merge? For Ex- Syncing latest changes from main for ocs-operator[DD:MM:YYYY - DD:MM:YYYY] Or else there will be so many PRs with the exact same name pattern Syncing the latest changes. @iamniting WDYT?

Why would we want to do that? That will just make the title messier.
Secondly, if suppose this PR is not merged today and we have some new changes in main branch before the pipeline runs tomorrow then. those commits will simply be pushed to this PR itself, as the branch used in the pipeline would be same, so we won't have different PRs everyday.
But, as said, ideally we should not be holding it off and try to merge it daily. I am updating the pipeline to auto-request reviewers to make the process smoother.

Nikhil-Ladha:
> @Nikhil-Ladha Is it possible to drop the merge commits?

I think you are referring to the merge bot commits?
Unfortunately, no. We are not cherry-picking individual commits, rather just fetching the commits not present in this branch and merging them.
I am not even sure that we can pick individual commits and just merge them and create the PR, will have to check.

iamniting:
I like the idea of pushing it to the same branch. This approach ensures that if we decide to hold the pull request for any reason, or in case it causes any issues, we'll have the changes and potential fixes readily available within the same pull request, often the following day and we don't have to close the stale PRs.

PR #2215, opened 2023-10-16
Add new field 'monCount' in storagecluster CR

Nikhil-Ladha:
Added a new field 'monCount' in storagecluster CR that takes in the custom mon count from user i.e, either 3 or 5. Also, updated the `getMonCount` function definition to remove the unused `nodeCount` param.

### DIFF
--- a/controllers/storagecluster/cephcluster.go
+++ b/controllers/storagecluster/cephcluster.go
@@ -435,7 +435,7 @@ func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, nodeCount int, s
 				Image:            cephImage,
 				AllowUnsupported: allowUnsupportedCephVersion(),
 			},
-			Mon:             generateMonSpec(sc, nodeCount),
+			Mon:             generateMonSpec(sc),
 			Mgr:             generateMgrSpec(sc),
 			DataDirHostPath: "/var/lib/rook",
 			DisruptionManagement: rookCephv1.DisruptionManagementSpec{
@@ -696,7 +696,7 @@ func getMgrCount(arbiterMode bool) int {
 	return defaults.DefaultMgrCount
 }
 
-func getMonCount(nodeCount int, arbiter bool) int {
+func getMonCount(monCount int, arbiter bool) int {
 	// return static value if overridden
 	override := os.Getenv(monCountOverrideEnvVar)
 	if override != "" {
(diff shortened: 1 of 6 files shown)


controllers/storagecluster/cephcluster.go
```diff
@@ -708,6 +708,14 @@ func getMonCount(nodeCount int, arbiter bool) int {
 		}
 	}
 
+	if sc != nil && sc.Spec.ManagedResources.CephCluster.MonCount != 0 {
+		if sc.Spec.ManagedResources.CephCluster.MonCount != 3 && sc.Spec.ManagedResources.CephCluster.MonCount != 5 {
+			log.Error(fmt.Errorf("Provided incorrect mon count"), "It is recommended to set mon count as 3 or 5 only")
```
Madhu-1, line 713: You can add this check as kubebuilder tag in the CR?

Nikhil-Ladha: Ah,right. that could be done.
I will update it.

Nikhil-Ladha: Done

controllers/storagecluster/cephcluster.go
```diff
@@ -696,7 +696,7 @@ func getMgrCount(arbiterMode bool) int {
 	return defaults.DefaultMgrCount
 }
 
-func getMonCount(nodeCount int, arbiter bool) int {
+func getMonCount(sc *ocsv1.StorageCluster, arbiter bool) int {
```
Madhu-1, line 699: instead of sc just pass the MountCount?

Madhu-1 (commented):
LGTM

controllers/storagecluster/cephcluster.go
iamniting: +1

iamniting (changes requested):

controllers/storagecluster/cephcluster.go
Nikhil-Ladha: Made the update.

controllers/storagecluster/cephcluster.go
```diff
@@ -708,6 +708,10 @@ func getMonCount(nodeCount int, arbiter bool) int {
 		}
 	}
 
+	if monCount != 0 {
+		return monCount
+	}
```
Madhu-1, line 713: what is the behavior, if arbiter is enabled and someone is setting the MonCount to `3` as well?

if we always want to stay at `5` for arbiter move this check to the end of the function.

Nikhil-Ladha: Right, we would want to stay at 5 if arbiter mode is enabled.
Sorry for the confusion here :/

Madhu-1 (approved):

iamniting (approved):

PR #2217, opened 2023-10-17
Deploy 2 mgr daemons by default, one active and one standby.

malayparida2000:
This significantly improves the reliability and eliminates the possibility of a single point of failure for the mgr service.

Jira-https://issues.redhat.com/browse/RHSTOR-4139


controllers/storagecluster/cephcluster.go
```diff
@@ -1067,16 +1064,13 @@ func generateMonSpec(sc *ocsv1.StorageCluster, nodeCount int) rookCephv1.MonSpec
 
 func generateMgrSpec(sc *ocsv1.StorageCluster) rookCephv1.MgrSpec {
 	spec := rookCephv1.MgrSpec{
-		Count: getMgrCount(arbiterEnabled(sc)),
+		Count:                getMgrCount(),
+		AllowMultiplePerNode: statusutil.IsSingleNodeDeployment(),
```
travisn, line 1068: If it's a single node deployment, how about if we default to a single mgr instead of allowing multiple per node?

api/v1/storagecluster_types.go
```diff
@@ -254,6 +254,7 @@ type ManageCephRBDMirror struct {
 type MgrSpec struct {
 	// EnableActivePassive can be set as true to deploy 2 ceph manager pods, one active and one standby
 	// Ceph will promote the standby mgr when the active mgr goes down due to any reason
+	// +kubebuilder:deprecatedversion:warning="This field has been deprecated and will be removed in future versions. Active passive mgr is now enabled by default."
 	EnableActivePassive bool `json:"enableActivePassive,omitempty"`
```
travisn, line 258: How can the user disable this feature? While we default to 2 mgrs, it still seems valid to allow 1 mgr. Perhaps we add a new setting that is `disableActivePassive`. We still should allow 1 mgr for minimal deployments. For example, in the new profiles, the smallest profile could have a single mgr.

travisn (changes requested):

controllers/storagecluster/cephcluster.go
malayparida2000: Done

api/v1/storagecluster_types.go
```diff
@@ -251,10 +251,12 @@ type ManageCephRBDMirror struct {
 }
 
 // MgrSpec defines the settings for the Ceph Manager
+// +structType:=atomic
```
Madhu-1, line 254: what does this do?

api/v1/storagecluster_types.go
```diff
@@ -251,10 +251,12 @@ type ManageCephRBDMirror struct {
...
 type MgrSpec struct {
-	// EnableActivePassive can be set as true to deploy 2 ceph manager pods, one active and one standby
-	// Ceph will promote the standby mgr when the active mgr goes down due to any reason
+	// EnableActivePassive can be set as true to deploy 2 ceph manager pods(one active and one standby), in any situation no matter the other settings & constraints
 	EnableActivePassive bool `json:"enableActivePassive,omitempty"`
+	// DisableActivePassive can be set as true to deploy only 1 ceph manager pod(it will be always active with no standby), in any situation no matter the other settings & constraints
+	DisableActivePassive bool `json:"disableActivePassive,omitempty"`
```
Madhu-1, line 259: Having two flags can create confusion. what can `EnableActivePassive` be used for both operations?

controllers/storagecluster/cephcluster_test.go
```diff
@@ -295,6 +297,123 @@ func TestNewCephClusterMonData(t *testing.T) {
...
+					{Name: "pg_autoscaler", Enabled: true},
+					{Name: "balancer", Enabled: true},
+				},
+			},
+		},
+	}
+
+	for _, c := range cases {
+		t.Logf("Case: %s\n", c.label)
+		os.Setenv(util.SingleNodeEnvVar, c.isSingleNode)
```
Madhu-1, line 410: use t.SetEnv here

api/v1/storagecluster_types.go
malayparida2000: This basically resets any earlier set value in the struct, So in this case let's say someone had EnableActivePassive as true, Now they mark DisableActivePassive as true, So the earlier flag automatically gets reset.

Ref-https://book.kubebuilder.io/reference/markers/crd-processing#:~:text=string-,specifies%20the%20level%20of%20atomicity%20of%20the%20struct%3B%20i.e.%20whether%20each%20field%20in%20the%20struct%20is%20independent%20of%20the%20others%2C%20or%20all%20fields%20are%20treated%20as%20a%20single%20unit.,-Possible%20values%3A

api/v1/storagecluster_types.go
malayparida2000: I wanted to give the independence to choose the number in all situations.

Let's say in a single node deployment someone wants 2 mgr or in our upcoming resource profiles in a lean profile(it will have 1 mgr), Someone wants 2 mgr. In that case, they can use the EnableActivePassive flag.

Let's say in a normal situation someone wants to use just 1 mgr, in that case they use the DisableActivePassive variable.

If we want to stick to just one var EnableActivePassive we have to make it default true, which I think will create a problem when I implement the resource profile feature as it will not be possible to know if the value has been set manually by a customer or it's just the default.

controllers/storagecluster/cephcluster_test.go
malayparida2000: Done, Thanks.

api/v1/storagecluster_types.go
Madhu-1: As it's an API change we need to be a little careful when adding/removing new Fields.

> I wanted to give the independence to choose the number in all situations.
> 
> Let's say in a single node deployment someone wants 2 mgr or in our upcoming resource profiles in a lean profile(it will have 1 mgr), Someone wants 2 mgr. In that case, they can use the EnableActivePassive flag.
> 

Single Node deployment is for only development/testing purposes.

Are we going to have different mgr counts for different profiles? i assume Lean profile is meant to stick to a set of resources, if they want to change something they will change the profile. 

> Let's say in a normal situation someone wants to use just 1 mgr, in that case they use the DisableActivePassive variable.
> 

They can set the EnableActivePassive to false as well?

> If we want to stick to just one var EnableActivePassive we have to make it default true, which I think will create a problem when I implement the resource profile feature as it will not be possible to know if the value has been set manually by a customer or it's just the default.

lets get to a conclusion on that before making any API change, based on the discussion and feedback we can see do we really need two fields or if we can have only one which is default set to true.

malayparida2000: > i assume Lean profile is meant to stick to a set of resources, if they want to change something they will change the profile.

Travis was suggesting in an earlier comment in lean profile we can have 1 MGR & in Balance profile we can have 2.

> They can set the EnableActivePassive to false as well?
Yes, But then false is the default for a bool value so how do we know if it's manually set by someone or it's the default

> lets get to a conclusion on that before making any API change, based on the discussion and feedback we can see do we really need two fields or if we can have only one which is default set to true.
Sure open to suggestions & discussion to have an optimal outcome

malayparida2000: > if we allow overriding the mgr count wont it cause any problems?

I also have the same question, If we will have profiles for different situations, Do we really need to provide the flexibility to override the mgr numbers? Initially my plan was to just deprecate the EnableActivepassive flag & control the mgr numbers from the profile. 1 in Lean mode, 2 in balanced & performance

travisn: What would give the most flexibility is if the new flag is a `bool*` so that by default (if it's nil) we use our defaults based on profiles, and if the user sets it to `true` or `false` we override the defaults. But I agree this is likely more flexibility than we need. With the introduction of the profiles, we should have enough flexibility already. I can't think why the user would have an opinion about how many mgrs are running. So we could go with your original change to deprecate the old setting and not add a new setting.

controllers/storagecluster/cephcluster_test.go
```diff
@@ -295,6 +296,46 @@ func TestNewCephClusterMonData(t *testing.T) {
...
 
+func TestGenerateMgrSpec(t *testing.T) {
+	cases := []struct {
+		label        string
+		sc           *api.StorageCluster
+		isSingleNode string
+		expectedMgr  cephv1.MgrSpec
+	}{
+		{
+			label: "case 1", // Default case
```
Madhu-1, line 307: instead of case 1 and case 3 , the description can be more details about what you are testing.

controllers/storagecluster/cephcluster_test.go
```diff
@@ -295,6 +296,46 @@ func TestNewCephClusterMonData(t *testing.T) {
 	}
 }
 
+func TestGenerateMgrSpec(t *testing.T) {
+	cases := []struct {
+		label        string
+		sc           *api.StorageCluster
+		isSingleNode string
```
Madhu-1, line 303: use bool?

Madhu-1 (commented):
LGTM

controllers/storagecluster/cephcluster_test.go
malayparida2000: Done

controllers/storagecluster/cephcluster_test.go
malayparida2000: Done

api/v1/storagecluster_types.go
```diff
@@ -254,6 +254,7 @@ type ManageCephRBDMirror struct {
 type MgrSpec struct {
 	// EnableActivePassive can be set as true to deploy 2 ceph manager pods, one active and one standby
 	// Ceph will promote the standby mgr when the active mgr goes down due to any reason
+	// +kubebuilder:deprecatedversion:warning="This field has been deprecated and will be removed in future versions. ActivePassive Mgr is now default."
```
travisn, line 257:
```suggestion
	// +kubebuilder:deprecatedversion:warning="This field has been deprecated and will be removed in future versions. Both the active and passive mgrs are always enabled."
```

travisn (commented):
LGTM, just a suggestion on the deprecated comment.

api/v1/storagecluster_types.go
malayparida2000: Reworded a bit to make it more informative.

api/v1/storagecluster_types.go
```diff
@@ -254,6 +254,7 @@ type ManageCephRBDMirror struct {
 type MgrSpec struct {
 	// EnableActivePassive can be set as true to deploy 2 ceph manager pods, one active and one standby
 	// Ceph will promote the standby mgr when the active mgr goes down due to any reason
+	// +kubebuilder:deprecatedversion:warning="This field has been deprecated and will be removed in future. By default we now have 2 ceph manager pods, one active and one standby ."
```
travisn, line 257: There is no way for the user to change the number of mgrs, right? So it seems better not to mention "by default" since that implies they could change it.

travisn: Although with the basic profile we will allow a single mgr, so this comment would still be valid with the default profile.

iamniting, line 257: Can you remove the extra space at the end of line?

iamniting (changes requested):
Can also correct the `filed` keyword in the commit msg description.

iamniting (approved):

PR #2222, opened 2023-10-18
Provide a flag to mark rbd storageclass as the default for the cluster

malayparida2000:
Ref-https://issues.redhat.com/browse/RHSTOR-2452

### DIFF
--- a/api/v1/storagecluster_types.go
+++ b/api/v1/storagecluster_types.go
@@ -185,11 +185,13 @@ type ManageCephDashboard struct {
 	SSL bool `json:"ssl,omitempty"`
 }
 
-// ManageCephBlockPools defines how to reconcilea CephBlockPools
+// ManageCephBlockPools defines how to reconcile CephBlockPools
 type ManageCephBlockPools struct {
 	ReconcileStrategy    string `json:"reconcileStrategy,omitempty"`
 	DisableStorageClass  bool   `json:"disableStorageClass,omitempty"`
 	DisableSnapshotClass bool   `json:"disableSnapshotClass,omitempty"`
+	// if set to true, the storageClass created for cephBlockPools will be annotated as the default for the whole cluster
+	DefaultStorageClass bool `json:"defaultStorageClass,omitempty"`
 	// StorageClassName specifies the name of the storage class created for ceph block pools
 	// +kubebuilder:validation:MaxLength=253
 	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$
--- a/controllers/storagecluster/storageclasses.go
+++ b/controllers/storagecluster/storageclasses.go
@@ -22,7 +22,8 @@ import (
 )
 
 const (
-	storageClassSkippedError = "some StorageClasses were skipped while waiting for pre-requisites to be met"
+	storageClassSkippedError      = "some StorageClasses were skipped while waiting for pre-requisites to be met"
+	defaultStorageClassAnnotation = "storageclass.kubernetes.io/is-default-class"
 )
 
 // StorageClassConfiguration provides configuration options for a StorageClass.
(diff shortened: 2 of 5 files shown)


controllers/storagecluster/storageclasses.go
```diff
@@ -283,6 +283,10 @@ func newCephBlockPoolStorageClassConfiguration(initData *ocsv1.StorageCluster) S
 		disable:           managementSpec.DisableStorageClass,
 		isClusterExternal: initData.Spec.ExternalStorage.Enable,
 	}
+	if initData.Spec.ManagedResources.CephBlockPools.DefaultSCForCluster {
+		scc.storageClass.Annotations["storageclass.kubernetes.io/is-default-class"] = "true"
```
Madhu-1, line 287: use existing Const from Kubernetes package https://github.com/kubernetes/kubernetes/blob/dc8b57d8a75764e649d641ac77c1539a574d9aed/pkg/volume/util/storageclass.go#L32 ?

api/v1/storagecluster_types.go
```diff
@@ -185,11 +185,13 @@ type ManageCephDashboard struct {
...
 }
 
-// ManageCephBlockPools defines how to reconcilea CephBlockPools
+// ManageCephBlockPools defines how to reconcile CephBlockPools
 type ManageCephBlockPools struct {
 	ReconcileStrategy    string `json:"reconcileStrategy,omitempty"`
 	DisableStorageClass  bool   `json:"disableStorageClass,omitempty"`
 	DisableSnapshotClass bool   `json:"disableSnapshotClass,omitempty"`
+	// if set to true, the storageClass created for cephBlockPools will be annotated as the default for the whole cluster
+	DefaultSCForCluster bool `json:"defaultSCForCluster,omitempty"`
```
Madhu-1, line 194: `DefaultStrageClass` sounds good to keep the SC-related members naming convention similar?

malayparida2000: done

controllers/storagecluster/storageclasses.go
Madhu-1: define a const for it and use it as you are using it in 2 places.

api/v1/storagecluster_types.go
```diff
@@ -185,11 +185,13 @@ type ManageCephDashboard struct {
...
 }
 
-// ManageCephBlockPools defines how to reconcilea CephBlockPools
+// ManageCephBlockPools defines how to reconcile CephBlockPools
 type ManageCephBlockPools struct {
 	ReconcileStrategy    string `json:"reconcileStrategy,omitempty"`
 	DisableStorageClass  bool   `json:"disableStorageClass,omitempty"`
 	DisableSnapshotClass bool   `json:"disableSnapshotClass,omitempty"`
+	// if set to true, the storageClass created for cephBlockPools will be annotated as the default for the whole cluster
+	DefaultStorageClassForCluster bool `json:"defaultStorageClassForCluster,omitempty"`
```
Madhu-1, line 194: ForCluster is the name that can cause confusion as to whether its defaults for the whole OCP cluster or for the storagecluster, my suggestion is to just have `DefaultStorageClass` as the field name

Madhu-1 (commented):
LGTM

controllers/storagecluster/storageclasses.go
```diff
@@ -22,7 +22,8 @@ import (
 )
 
 const (
-	storageClassSkippedError = "some StorageClasses were skipped while waiting for pre-requisites to be met"
+	storageClassSkippedError        = "some StorageClasses were skipped while waiting for pre-requisites to be met"
+	isDefaultStorageClassAnnotation = "storageclass.kubernetes.io/is-default-class"
```
iamniting, line 26: `defaultStorageClassAnnotation` lets rename it

iamniting (changes requested):

controllers/storagecluster/storageclasses.go
malayparida2000: Copied the name from the Kubernets repo https://github.com/kubernetes/kubernetes/blob/dc8b57d8a75764e649d641ac77c1539a574d9aed/pkg/volume/util/storageclass.go#L32

iamniting: I am not sure why are they doing this, But it looks like this is a func variable instead of a const variable.

malayparida2000: Okay, updating

malayparida2000: Done

iamniting (approved):

PR #2223, opened 2023-10-19
Fix Provisioner Namespace Configuration

iamniting:
Previously the storage cluster was deployed within the operator namespace, which was matching the Provisioner namespace and working well. However with the introduction of multiple StorageClusters, using the storagecluster namespace as provisioner namespace is no longer viable. Instead we should use operator namespace as that is the namespace where provisoners are deployed.

### DIFF
--- a/controllers/storagecluster/storageclasses.go
+++ b/controllers/storagecluster/storageclasses.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"os"
 	"reflect"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


controllers/storagecluster/storageclasses.go
```diff
@@ -227,7 +228,7 @@ func newCephFilesystemStorageClassConfiguration(initData *ocsv1.StorageCluster)
 					"description": "Provides RWO and RWX Filesystem volumes",
 				},
 			},
-			Provisioner:   fmt.Sprintf("%s.cephfs.csi.ceph.com", initData.Namespace),
+			Provisioner:   fmt.Sprintf("%s.cephfs.csi.ceph.com", os.Getenv(util.OperatorNamespaceEnvVar)),
```
Madhu-1, line 231: instead of calling os.GetEnv, write a helper function in util which returns the namespace and use it?

iamniting: We already have a func in util. But that also returns an error which I don't want to check here. We are already checking the error in main.
https://github.com/red-hat-storage/ocs-operator/blob/main/main.go#L123

Madhu-1: if you really want you can write a wrapper in util that skips the return error and even caches the operator namespace and return it or you can store it in the reconcile struct as well. calling os.Getenv from the main controller multiple times doesn't look good as this value doesn't get changed anytime during runtime.

iamniting: Yes I am planning to cache it in further changes. I am making small small changes which makes it easier to test the things.

Madhu-1 (approved):

PR #2225, opened 2023-10-20
Ran pending make command and fix obc alerts issue

aruniiird:
Ran make command to update prometheus monitoring yaml files.

### DIFF
--- a/metrics/internal/collectors/object-bucket.go
+++ b/metrics/internal/collectors/object-bucket.go
@@ -12,6 +12,7 @@ import (
 	cephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
 	rookclient "github.com/rook/rook/pkg/client/clientset/versioned"
 	cephv1listers "github.com/rook/rook/pkg/client/listers/ceph.rook.io/v1"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
@@ -143,6 +144,14 @@ func (c *ObjectBucketCollector) getAllObjectBuckets(cephObjectStore cephv1.CephO
 		LabelSelector: selector,
 	}
 
+	// collect the OBC list
+	objectBucketClaimList, err := c.bktclient.ObjectbucketV1alpha1().
+		ObjectBucketClaims(cephObjectStore.Namespace).List(context.TODO(), listOpts)
+	// any error, don't panic
+	if err != nil {
+		// make a non-nil OBC List with Items 'nil'
+		objectBucketClaimList = &libbucket.ObjectBucketClaimList{Items: nil}
+	}
 	objectBucketsList, err := c.bktclient.ObjectbucketV1alpha1().ObjectBuckets().List(context.TODO(), listOpts)
 	if err != nil {
 		klog.Errorf("Couldn't list ObjectBuckets. %v", err)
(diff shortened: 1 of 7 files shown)


jmolmo (approved):

aruniiird:
Adding a screen shot, where OBC name, `aruns-rgw-obc`, is added to the alert (the alert in the image, `MyObcQuotaObjectsAlert`, is created just for test purpose)

Screenshot from 2023-11-03 22-26-20

aruniiird:
@jmolmo , @umangachapagain , can you please take a look?

jmolmo (approved):

metrics/internal/collectors/object-bucket.go
```diff
@@ -153,6 +162,18 @@ func (c *ObjectBucketCollector) getAllObjectBuckets(cephObjectStore cephv1.CephO
...
+			// this for loop will be safely skipped if the 'Items' object is 'nil'
+			for _, obc := range objectBucketClaimList.Items {
+				// add the OBC details into the matching OB's spec.claimref
+				if obc.Spec.ObjectBucketName == ob.Name {
+					ob.Spec.ClaimRef = &corev1.ObjectReference{Name: obc.Name, Namespace: obc.Namespace}
+					// found the obc associated with this ob
+					break
+				}
+			}
+		}
```
umangachapagain, line 176: Isn't this information available by default? If not, in which cases is it not available?
This is a lot of overhead for us and delays metrics generation.

metrics/internal/collectors/object-bucket.go
```diff
@@ -143,6 +144,14 @@ func (c *ObjectBucketCollector) getAllObjectBuckets(cephObjectStore cephv1.CephO
...
 	}
 
+	// collect the OBC list
+	objectBucketClaimList, err := c.bktclient.ObjectbucketV1alpha1().
+		ObjectBucketClaims(cephObjectStore.Namespace).List(context.TODO(), listOpts)
+	// any error, don't panic
+	if err != nil {
+		// make a non-nil OBC List with Items 'nil'
+		objectBucketClaimList = &libbucket.ObjectBucketClaimList{Items: nil}
+	}
```
umangachapagain, line 154: Why not just initialize `objectBucketClaimList` with zero-value and then do the List operation?

umangachapagain (changes requested):

metrics/internal/collectors/object-bucket.go
aruniiird: ClaimRef info is not available.
Following is a debug line added to check the ObjectBucket element collected,
```
[0] ObjectBucket Details        : {TypeMeta:{Kind:ObjectBucket APIVersion:objectbucket.io/v1alpha1} ObjectMeta:{Name:obc-openshift-storage-aruns-rgw-obc GenerateName: Namespace: SelfLink: UID:35aae873-1fb0-46a2-8a22-6075fc9dd5aa ResourceVersion:724028 Generation:1 CreationTimestamp:2023-11-03 05:34:16 +0000 UTC DeletionTimestamp:<nil> DeletionGracePeriodSeconds:<nil> Labels:map[bucket-provisioner:openshift-storage.ceph.rook.io-bucket] Annotations:map[] OwnerReferences:[] Finalizers:[objectbucket.io/finalizer] ManagedFields:[{Manager:rook Operation:Update APIVersion:objectbucket.io/v1alpha1 Time:2023-11-03 05:34:16 +0000 UTC FieldsType:FieldsV1 FieldsV1:{...} Subresource:} {Manager:rook Operation:Update APIVersion:objectbucket.io/v1alpha1 Time:2023-11-03 05:34:16 +0000 UTC FieldsType:FieldsV1 FieldsV1:{"f:status":{".":{},"f:phase":{}}} Subresource:status}]} Spec:{StorageClassName:ocs-storagecluster-ceph-rgw ReclaimPolicy:0xc000913240 ClaimRef:&ObjectReference{Kind:,Namespace:,Name:,UID:,APIVersion:,ResourceVersion:,FieldPath:,} Connection:0xc00095c6c0} Status:{Phase:Bound}}
```

and we can see that the `ClaimRef` (towards the end of the above line) is an empty/default ObjectReference.
Throughout the test, the claimref object was empty and I'm not aware of a scenario where it was available.

metrics/internal/collectors/object-bucket.go
aruniiird: What if the `List()` function returns a `nil, err` value (at line#148)?
Even if we had initialized the `objectBucketClaimList` with a zero value (before the line#148), it would be reset to a `nil` value.
So it is always safer to reset it to a zero (non-nil) value on error.

metrics/internal/collectors/object-bucket.go
umangachapagain: This fix is ok for now but, we need to find a better solution long term.
Either we need to cache these info early or get a fix in OB to include OBC reference.

umangachapagain (approved):

PR #2230, opened 2023-10-25
Fix for metrics exporter pod going in crashbacklloop

rchikatw:
After removing the capacity field from StorageConsumer CR, the dependency on that field in the metrics exporter was apparently not removed.


ezio-auditore (approved):
LGTM

umangachapagain (approved):

PR #2231, opened 2023-10-25
Altering restrictions for StorageProfile API 

raaizik:
- Only `spec` and `deviceClass` are required and the respective rules are changed from `initially unset` to `always set` policy
- This allows full immutability throughout the whole of the CRs' lifecycle
- Removes unused `StorageProfile` phases


.github/workflows/ocs-operator-ci.yaml
```diff
@@ -108,4 +108,4 @@ jobs:
           check_filenames: true
           check_hidden: true
           skip: vendor
-          ignore_words_list: xdescribe,contails,shouldnot
+          ignore_words_list: xdescribe,contails,shouldnot,colletor
```
Madhu-1, line 111: i think its a typo in the current code base which need to be fixed, we should not add it here

api/v1/storageprofile_types.go
```diff
@@ -24,25 +24,25 @@ import (
...
 
-	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Required
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="DeviceClass is immutable"
 	// +kubebuilder:validation:MaxLength=512
 	// DeviceClass is the deviceclass name.
 	DeviceClass string `json:"deviceClass"`
 
-	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Required
```
Madhu-1, line 40: AFAIK this is optional not a mandatory to have it

api/v1/storageprofile_types.go
```diff
@@ -24,25 +24,25 @@ import (
...
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="SharedFilesystemConfiguration is immutable"
 	// configurations to use for cephfilesystem.
 	SharedFilesystemConfiguration SharedFilesystemConfigurationSpec `json:"sharedFilesystemConfiguration,omitempty"`
 
-	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Required
```
Madhu-1, line 45: AFAIK this is optional not a mandatory to have it

api/v1/storageprofile_types.go
```diff
@@ -24,25 +24,25 @@ import (
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
 // StorageProfileSpec defines the desired state of StorageProfile
-// +kubebuilder:validation:XValidation:rule="!has(oldSelf.deviceClass) || has(self.deviceClass)", message="deviceClass is required once set"
-// +kubebuilder:validation:XValidation:rule="!has(oldSelf.sharedFilesystemConfiguration) || has(self.sharedFilesystemConfiguration)", message="sharedFilesystemConfiguration is required once set"
-// +kubebuilder:validation:XValidation:rule="!has(oldSelf.blockPoolConfiguration) || has(self.blockPoolConfiguration)", message="blockPoolConfiguration is required once set"
+// +kubebuilder:validation:XValidation:rule="has(oldSelf.deviceClass) && has(self.deviceClass)", message="deviceClass is required"
+// +kubebuilder:validation:XValidation:rule="has(oldSelf.sharedFilesystemConfiguration) && has(self.sharedFilesystemConfiguration)", message="sharedFilesystemConfiguration is required"
+// +kubebuilder:validation:XValidation:rule="has(oldSelf.blockPoolConfiguration) && has(self.blockPoolConfiguration)", message="blockPoolConfiguration is required"
```
Madhu-1, line 29: i think the earlier check was correct, any specific reason to change it?

api/v1/storageprofile_types.go
raaizik: The idea is to set all fields to required. The issue before was that the fields could've been initially unset and we aim to have them set at all times.

api/v1/storageprofile_types.go
raaizik: See above

api/v1/storageprofile_types.go
raaizik: The old rule applied to a different use case, as mentioned in the previous PR [comment](https://github.com/red-hat-storage/ocs-operator/pull/2204#discussion_r1356840478)

api/v1/storageprofile_types.go
nb-ohad: @Madhu-1 The requirement was for immutable CRs. The current implementation allows you to set values to unset fields after creation. I believe @raaizik fix will achieve that

api/v1/storageprofile_types.go
nb-ohad: https://github.com/red-hat-storage/ocs-operator/pull/2231#discussion_r1371772674

api/v1/storageprofile_types.go
raaizik: @Madhu-1 I think you meant that it's uncertain whether the other fields, apart from `deviceClass`, should be mandatory _outside_ of the fact that currently the values can be initially unset before set, @nb-ohad. I think we made an assumption that they should be mandatory, not sure why though.

nb-ohad: @raaizik I think you are right! I just observed it myself. 
We don't want to require the customer to fill in all of the fields. We just want that the Spec will stay immutable after CR creation. 

I think what we need is a validation rule on the Spec field itself that forces it to be identical to the old spec field

raaizik: @Madhu-1 The CEL rules prevent updates and removals. The difference is [this](https://github.com/red-hat-storage/ocs-operator/pull/2231#discussion_r1371764253)

raaizik: To prove my case, I was able to create an SP without `blockPoolConfiguration` and then edit it to include it! In order to enforce the immutability restriction mentioned, I'd need to do it on a per sub-field basis per each complex type field.

leelavg: - you mean setting validation on `type SharedFilesystemConfigurationSpec struct` and `type BlockPoolConfigurationSpec struct`, that might interfere with StorageCluster Spec, isn't it?
- btw, I don't get why `omitempty` is in the picture, I believe it doesn't help with validations but only for (de/)serialize (or) you want to try this combo with `deviceClass` as optional?

api/v1/storageprofile_types.go
leelavg: - I agree with Madhu, `deviceClass` is required and immutable, everything else is optional but required and immutable after once set.
- btw, current rule on type states, during a transition existing (self) value and new (oldSelf) value should be set which is already achieved with `// +kubebuilder:validation:Required` and so these two rules are redundant
- Only question is, can a complex field be `Immutable once set after modification` as mentioned in another comment (without any side effects to shared CRDs)?

api/v1/storageprofile_types.go
```diff
@@ -24,25 +24,25 @@ import (
...
 
-	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Required
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="SharedFilesystemConfiguration is immutable"
 	// configurations to use for cephfilesystem.
 	SharedFilesystemConfiguration SharedFilesystemConfigurationSpec `json:"sharedFilesystemConfiguration,omitempty"`
 
-	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Required
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="BlockPoolConfiguration is immutable"
```
Madhu-1, line 46: i did the below changes and am getting as below results

```go
 // StorageProfileSpec defines the desired state of StorageProfile
-// +kubebuilder:validation:XValidation:rule="!has(oldSelf.deviceClass) || has(self.deviceClass)", message="deviceClass is required once set"
-// +kubebuilder:validation:XValidation:rule="!has(oldSelf.sharedFilesystemConfiguration) || has(self.sharedFilesystemConfiguration)", message="sharedFilesystemConfiguration is required once set"
-// +kubebuilder:validation:XValidation:rule="!has(oldSelf.blockPoolConfiguration) || has(self.blockPoolConfiguration)", message="blockPoolConfiguration is required once set"
 type StorageProfileSpec struct {
        // INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
        // Important: Run "make" to regenerate code after modifying this file
 
-       // +kubebuilder:validation:Optional
-       // +kubebuilder:validation:XValidation:rule="self == oldSelf",message="DeviceClass is immutable"
       // +kubebuilder:validation:MaxLength=512
+       // +kubebuilder:validation:Required
        // DeviceClass is the deviceclass name.
        DeviceClass string `json:"deviceClass"`
 
        // +kubebuilder:validation:Optional
-       // +kubebuilder:validation:XValidation:rule="self == oldSelf",message="SharedFilesystemConfiguration is immutable"
        // configurations to use for cephfilesystem.
        SharedFilesystemConfiguration SharedFilesystemConfigurationSpec `json:"sharedFilesystemConfiguration,omitempty"`
 
        // +kubebuilder:validation:Optional
-       // +kubebuilder:validation:XValidation:rule="self == oldSelf",message="BlockPoolConfiguration is immutable"
        // configurations to use for  profile specific blockpool.
        BlockPoolConfiguration BlockPoolConfigurationSpec `json:"blockPoolConfiguration,omitempty"`
 }
@@ -67,7 +60,9 @@ type StorageProfile struct {
        metav1.TypeMeta   `json:",inline"`
 
-       Spec   StorageProfileSpec   `json:"spec,omitempty"`
+       // +kubebuilder:validation:Required
+       // +kubebuilder:validation:XValidation:rule="self == oldSelf",message="spec is immutable"
+       Spec   StorageProfileSpec   `json:"spec"`
        Status StorageProfileStatus `json:"status,omitempty"`
 }
```

* Not able to create the SP without Spec or DeviceClass
* Not able to delete or update the DeviceClass once SP is created
* Not able to Update or Delete the sharedFilesystemConfiguration/blockPoolConfiguration once SP is created

Let me know if this is what you are looking for

api/v1/storageprofile_types.go
raaizik: 1. No.
2. It's pretty self explanatory. It literally allows optional fields that are left blank to be omitted (i.e., all the non `deviceClass` fields).

api/v1/storageprofile_types.go
raaizik: @leelavg 

1. No one said otherwise.
2. As mentioned above, I am testing it with new not yet committed changes.
3. It doesn't matter, we aim to eliminate this type of behavior, I was only mentioning it to prove my case to Ohad's point that it can be used on a per _non_ sub-field basis.

api/v1/storageprofile_types.go
```diff
@@ -24,25 +24,25 @@ import (
...
-// +kubebuilder:validation:XValidation:rule="!has(oldSelf.blockPoolConfiguration) || has(self.blockPoolConfiguration)", message="blockPoolConfiguration is required once set"
+// +kubebuilder:validation:XValidation:rule="has(oldSelf.deviceClass) && has(self.deviceClass)", message="deviceClass is required"
+// +kubebuilder:validation:XValidation:rule="has(oldSelf.sharedFilesystemConfiguration) && has(self.sharedFilesystemConfiguration)", message="sharedFilesystemConfiguration is required"
+// +kubebuilder:validation:XValidation:rule="has(oldSelf.blockPoolConfiguration) && has(self.blockPoolConfiguration)", message="blockPoolConfiguration is required"
 type StorageProfileSpec struct {
 	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
 	// Important: Run "make" to regenerate code after modifying this file
 
-	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Required
```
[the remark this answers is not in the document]
raaizik: @leelavg, as mentioned several times, I'm currently making changes and testing them.

api/v1/storageprofile_types.go
raaizik: @Madhu-1 as mentioned, I am making changes to the code! Re-read [comment](https://github.com/red-hat-storage/ocs-operator/pull/2231#discussion_r1371899791)

nb-ohad:
@raaizik Have we tested an approach of just having
```
// +kubebuilder:validation:XValidation:rule="oldSelf.all(key, (key in self && self[key] == oldSelf[key]) || (!(key in oldSelf) && !(key in self)), message="spec is immutable"
type StorageProfileSpec struct {
```
to force immutability of the entire Spec all at one? 
(might syntax might be wrong)

And of course mark the `deviceClass` field as required using normal validation rules:
```
DeviceClass
```

api/v1/storageprofile_types.go
raaizik: @Madhu-1 for the fields `sharedFilesystemConfiguration`and `blockPoolConfiguration` your suggestion allows transitioning from unset to set, or from set to unset, which is not what's required.

raaizik:
@iamniting the flaky check is failing [again](https://github.com/red-hat-storage/ocs-operator/actions/runs/6684612818/job/18162060858?pr=2231)

nb-ohad:
@raaizik The current solution needs a validation rule on each struct, sub-struct (maps), and fields.
This results in code that is very fragile to change and hard to maintain. 
 
 As an alternative, I suggest using the following rule only once above the `type StorageProfileSpec struct` line:
  ```
 // +kubebuilder:validation:XValidation:rule="oldSelf.all(key, (key in self && self[key] == oldSelf[key])) && self.all(key, (key in oldSelf && self[key] == oldSelf[key]))", message="Storage profile spec is immutable"
 ```
This rule will validate that values cannot change for the entire struct, including sub-structs(maps), and fields
 
You can see how it works here in this CEL playground: 
https://playcel.undistro.io/?content=H4sIAAAAAAAAA81STWvcMBD9K1MHkixsN6TbXAw5hGVLKdtkSSghxD2M7VmviDxjJHlTU%2FLfO7JN80Xaa3SwpPl48%2BT3ficlBkzSZA%2B%2BkiMwHsJWN27aADGlR7g5%2B74CcfDt6uIcNuJqDLOMMxZbXpHdpBmDrpJ2pqCFRe9T8L4con6LjsovxpLvfKB6IbwxVeswGOGxMy5M4fjT%2FPGepzD%2FfDI%2FGSK5leJuLWL%2F0Z5bfNZ%2BnLF%2Fx%2BySaUK%2FGkfea60KcHQE12QLqQmC9CIslitYW%2BwqJy2XHzLWkuexKBeyShTIYRHMjhQjP%2FOe6tx20Mi9aloC8c444Zo4RGwda0W1RgXQMznTZ%2B5N2A5zpa6FYfmXHayQqxYrgkOdP5lFJj2bj3DtTCDopHU9tccXwVZHj0U%2FPPXA%2Bq8RNBXP3pQUvfTEatM3vDairCMyHFy2fNC%2FYoe2xTeGY4WGfXjh5RFoOb4%2F9hViLRVRLpCNImDdqBdGYr4xg5QZP7H7DK09vKNuCvEbOUebwf5%2Bv99q7CecnsJY3d8nE00PdnzdPRb%2BByB5%2BAOvCgHoqwMAAA%3D%3D

raaizik:
> @raaizik The current solution needs a validation rule on each struct, sub-struct (maps), and fields. This results in code that is very fragile to change and hard to maintain.


Actually, it's the verified approach according to [this](https://kubernetes.io/blog/2022/09/23/crd-validation-rules-beta/#crd-transition-rules), with emphasis on the second entry (see below). Second, your suggested rule doesn't really differentiate between required and optional fields, @nb-ohad, which is crucial as seen below:
 


> Transition Rule | Purpose
> -- | --
> self == oldSelf | For  a required field, make that field immutable once it is set. For an  optional field, only allow transitioning from unset to set, or from set  to unset.
> (on parent of field) has(self.field) == has(oldSelf.field)on field: self == oldSelf | Make  a field immutable: validate that a field, even if optional, never  changes after the resource is created (for a required field, the  previous rule is simpler).
> 
>

api/v1/storageprofile_types.go
Madhu-1: @raaizik when i tested it i was not able to i was not able to add the sharedFilesystemConfiguration when it was not present , see below output





```yaml
[🎩︎]mrajanna@li-2cfbef4c-22d9-11b2-a85c-a3e4a93c405f ocs-operator $]kubectl get storageprofile -oyaml
apiVersion: v1
items:
  kind: StorageProfile
  metadata:
    creationTimestamp: "2023-10-30T09:18:09Z"
    generation: 1
    labels:
      app.kubernetes.io/created-by: ocs-operator
      app.kubernetes.io/instance: storageprofile-sample
      app.kubernetes.io/managed-by: kustomize
      app.kubernetes.io/name: storageprofile
      app.kubernetes.io/part-of: ocs-operator
    name: storageprofile-sample
    namespace: default
    resourceVersion: "19428"
    uid: cfd7f5cb-6e56-4dbf-8bed-66cd93982278
  spec:
    deviceClass: test
kind: List
metadata:
  resourceVersion: ""
```

When i tried to edit the SP to add the sharedFilesystemConfiguration it failed saying the spec is immutable

```yaml
# Please edit the object below. Lines beginning with a '#' will be ignored,
# and an empty file will abort the edit. If an error occurs while saving this file will be
# reopened with the relevant failures.
#
# storageprofiles.ocs.openshift.io "storageprofile-sample" was not valid:
# * spec: Invalid value: "object": spec is immutable
#
apiVersion: ocs.openshift.io/v1
kind: StorageProfile
metadata:
  creationTimestamp: "2023-10-30T09:18:09Z"
  generation: 1
  labels:
    app.kubernetes.io/created-by: ocs-operator
    app.kubernetes.io/instance: storageprofile-sample
    app.kubernetes.io/managed-by: kustomize
    app.kubernetes.io/name: storageprofile
    app.kubernetes.io/part-of: ocs-operator
  name: storageprofile-sample
  namespace: default
  resourceVersion: "19428"
  uid: cfd7f5cb-6e56-4dbf-8bed-66cd93982278
spec:
  deviceClass: test
  sharedFilesystemConfiguration:
    parameters:
      test: test
```


Am i testing something wrong or if am missing anything please let me know :)

raaizik: @Madhu-1 You're missing rules around `parameters` sub-fields of both `sharedFilesystemConfigurationand` and `blockPoolConfiguration`, that allows for (un)sets post creation. Not critical, my new suggestion uses a part of yours, take a look.

nb-ohad:
> > @raaizik The current solution needs a validation rule on each struct, sub-struct (maps), and fields. This results in code that is very fragile to change and hard to maintain.
> 
> Actually, it's the verified approach according to [this](https://kubernetes.io/blog/2022/09/23/crd-validation-rules-beta/#crd-transition-rules), with emphasis on the second entry (see below). Second, your suggested rule doesn't really differentiate between required and optional fields, @nb-ohad, which is crucial as seen below:
> 
> > Transition Rule
> > Purpose
> > 
> > 
> > 
> > 
> > self == oldSelf
> > For  a required field, make that field immutable once it is set. For an  optional field, only allow transitioning from unset to set, or from set  to unset.
> > 
> > 
> > (on _parent of field_) has(self.field) == has(oldSelf.field) on _field_: self == oldSelf
> > Make  a field immutable: validate that a field, even if optional, never  changes after the resource is created (for a required field, the  previous rule is simpler).

@raaizik 
I understand the reply but my concerns are still the same. 
We should not have the need to micro-manage every field when there is a simple rule that applies to all fields at once
(Imagine someone adding a field tomorrow and forgetting to add the validation rule).

Also, the general rule and the field-specific rules are basically the same. 
The only difference is the way you select the field itself

For the specific field rule, you select the field by having a dedicate rule above it that uses `self` and `oldSelf`. Whilt for the general approach, all the fields are targeted using a single selector: `oldSelf.all(key, (key in self && ...)` and `self.all(key, (key in self && ...)`. But the validation itself is exactly the same.

As for the explanation I just provided, I will vote to use the general approach.
WDYT?

raaizik:
> For the specific field rule, you select the field by having a dedicated rule above it that uses `self` and `oldSelf`. While for the general approach, all the fields are targeted using a single selector: `oldSelf.all(key, (key in self && ...)` and `self.all(key, (key in self && ...)`. But the validation itself is exactly the same.
> 
> As for the explanation I just provided, I will vote to use the general approach. WDYT?

@nb-ohad I get wanting to compress the rule to one-fits-all type of rule, that was of course my initial intuition. Your suggested rule can only wrap map fields, not complex type parent fields. As a matter of fact according to my research I didn't see any way to formulate a generic rule that captures _all_ field types _and_ doesn't require dev maintenance. I'm looking for a way to flatten out the spec field to a map so that it can go through compilation, but even then each inner scope map field would need to be addressed some way or another, so the initial concerns of limited genericity and of being prone to human error by devs remain.

api/v1/storagecluster_types.go
```diff
@@ -112,10 +112,12 @@ type StorageClusterSpec struct {
 }
 
 type SharedFilesystemConfigurationSpec struct {
+	// +kubebuilder:validation:Optional
 	Parameters map[string]string `json:"parameters,omitempty"`
 }
 
 type BlockPoolConfigurationSpec struct {
+	// +kubebuilder:validation:Optional
```
Madhu-1, line 120: I think this is not required because BlockPoolConfigurationSpec is already marked as optional. Same for filesystemconfigspec as well.

raaizik: Probably. Just wanted to be on the safe side of things

metrics/internal/collectors/registry.go
```diff
@@ -87,7 +87,7 @@ func enableCephBlocklistMirrorStore(opts *options.Options) {
 	go reflector.Run(opts.StopCh)
 }
 
-// RegisterPersistentVolumeAttributesCollector registers PV attribute colletor to registry
+// RegisterPersistentVolumeAttributesCollector registers PV attribute collector to registry
```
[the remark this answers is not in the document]
raaizik: Done

Madhu-1 (commented):
LGTM

nb-ohad:
@raaizik @Madhu-1 
The simplified rule suggested here will treat the entire spec as a yaml string. This means that the order of the fields matters! 
Two content identical yaml will fail validation if the top-level spec fields are ordered differently.  

But, because the intent is immutability I am willing to live with this minor issue. 
The side effect will be that even if we try to update the CR with the exact same content to bump the `resourceVersion` (a procedure which has its use cases) we will need to be sensitive to the order of the fields.

As long as we all understa

nb-ohad:
Closed by mistake (by spilling my water nug on the keyboard)
reopened

raaizik:
> @raaizik @Madhu-1 The simplified rule suggested here will treat the entire spec as a yaml string. This means that the order of the fields matters! Two content identical yaml will fail validation if the top-level spec fields are ordered differently.
> 
> But, because the intent is immutability I am willing to live with this minor issue. The side effect will be that even if we try to update the CR with the exact same content to bump the `resourceVersion` (a procedure which has its use cases) we will need to be sensitive to the order of the fields.

No it doesn't, @nb-ohad. Editing with re-ordered fields results in `Edit cancelled, no changes made.` = the validation doesn't fail.

Madhu-1:
> > @raaizik @Madhu-1 The simplified rule suggested here will treat the entire spec as a yaml string. This means that the order of the fields matters! Two content identical yaml will fail validation if the top-level spec fields are ordered differently.
> 
> No, it doesn't matter, @nb-ohad. Editing with re-ordered fields results in `Edit cancelled, no changes made.` = the validation doesn't fail.

Yes this is correct, changing order with kubectl edit `doesn't have any impact, i haven't tested with kube client/controller runtime client.

nb-ohad:
> Yes this is correct, changing order with kubectl edit `

Good to know.

Reading through  CEL documentation, fields should matter. 
I guess the reason it doesn't might be because of implementation details of ETCD. 
One possible explanation is that the usage of gRPC enforces order when encoding the message on the network, so the way you input (order of fields) on the client, does not matter, and when the validation runs the entire document is already normalized.

iamniting:
Hey @raaizik, could you please make sure to use clear and concise commit messages in your pull requests? Additionally, let's avoid including merge commits in the PR.

nb-ohad:
@iamniting I think these got there by mistake. I might have missed them in my last review. The PR should only include the last commit. Let me check how this affects the history of the branch

nb-ohad:
Fortunately for us, the inclusion of these commits here did not affect the branch history.

iamniting:
As far as i can check it does,

https://github.com/red-hat-storage/ocs-operator/commit/d593247acf7f8338cec6615600c6ffb3823e9407
https://github.com/red-hat-storage/ocs-operator/commit/e4e49be93b52e45a3677aac0b981559def06de0a
Screenshot from 2023-11-02 17-36-19

nb-ohad:
@iamniting I don't see duplicate commits

PR #2232, opened 2023-10-26
Use context passed via handler instead of reconciler's context

Nikhil-Ladha:
The reconciler's context can be `nil`, when the handler is executed if the storagecluster CR is not present, as the reconcile loop will not be triggered and the context field will never be initialised.

But, even if it's `nil` the `r.Client.List` operation doesn't always panic because there is a race scenario of the `cache` being synced by the time code execution reaches here. If the cache is not synced, then it waits for the cache to sync before executing the `List` operation and sends a close request via the `Done` channel of `ctx`, which results in a panic as `ctx` is nil in this case.

Therefore, we should not be using the `r.ctx` field and instead use the `context` pass by the handler to execute the `List` operation.

BUG: https://bugzilla.redhat.com/show_bug.cgi?id=2231076

### DIFF
--- a/controllers/storagecluster/storagecluster_controller.go
+++ b/controllers/storagecluster/storagecluster_controller.go
@@ -134,7 +134,7 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 
 			// Get the StorageCluster objects
 			scList := &ocsv1.StorageClusterList{}
-			err := r.Client.List(r.ctx, scList, &client.ListOptions{Namespace: obj.GetNamespace()})
+			err := r.Client.List(context, scList, &client.ListOptions{Namespace: obj.GetNamespace()})
 			if err != nil {
 				r.Log.Error(err, "Unable to list StorageCluster objects")
 				return []reconcile.Request{}


iamniting (approved):

Nikhil-Ladha:
@iamniting shall we backport it to all the previous releases? i.,e upto 4.12?

iamniting:
> @iamniting shall we backport it to all the previous releases? i.,e upto 4.12?

yes we should.

iamniting:
@Nikhil-Ladha We also need this fix for 4.14 as well. we can not backport it to 4.13 before backporting it to 4.14.

Nikhil-Ladha:
> @Nikhil-Ladha We also need this fix for 4.14 as well. we can not backport it to 4.13 before backporting it to 4.14.

Yes, I have created a backport bz for 4.14 as well, but why backporting to 4.14 first is important?

iamniting:
> > @Nikhil-Ladha We also need this fix for 4.14 as well. we can not backport it to 4.13 before backporting it to 4.14.
> 
> Yes, I have created a backport bz for 4.14 as well, but why backporting to 4.14 first is important?

This is crucial because the fix should be incorporated in version 4.14 before backporting it to other versions. Failing to do so would result in a situation where we might have the fix in older releases but not in 4.14.

Nikhil-Ladha:
Got it, I have created backports for all the releases, so we can merge accordingly.
@agarwal-mudit can we take this PR for 4.14 now? Or, do we need to wait for 4.14.1?

agarwal-mudit:
@Nikhil-Ladha we are good to backport this to 4.14, please go ahead with the backport

PR #2238, opened 2023-10-27
[release 4.13] Fix job names in metrics

aruniiird:


avanthakkar (commented):
I don't think we need these changes in 4.13 as we have disabled the `ceph-exporter` in 4.13

PR #2240, opened 2023-10-30
Separate ceph toolbox from ocsinitialization

malayparida2000:
Ref- https://issues.redhat.com/browse/RHSTOR-5012

In 4.11 when we moved the toolbox logic from ocsinitilization to storagecluster. The toolbox should have been associated with the storagecluster itself from the beginning as it's unique to each cephcluster.
So we have already moved all the enable/disable, and reconcile logic to the storagecluster, but still had kept backward compatibility. The plan was that in the future somewhere the compatibility with ocsinitialization would be removed & storagecluster would be the only place for anything related to the ceph toolbox.

With the development for multiple storagecluster happening in 4.15, we need to separate ocsinitialization from the cephtoolboox. This change is needed because there will be only one OCSInitialization CR in the operator's namespace, and no OCSInitialization CR might exist in the namespaces where storage clusters are deployed. Also in case we have multiple storageclusters we might not want the toolbox to be enabled for all of them.

This will require changes in some of the KCS on how to enable the toolbox, How to add tolerations to the toolbox, and some docs and ocs-ci(probably).


api/v1/ocsinitialization_types.go
```diff
@@ -29,11 +29,13 @@ type OCSInitializationSpec struct {
 	// should be deployed.
 	// Defaults to false
 	// +optional
+	// +kubebuilder:deprecatedversion:warning="This field has been deprecated and will be removed in future. The ceph tool box can be enabled by setting the enableCephTools field in the storage cluster CR Spec"
```
iamniting, line 32: I think we should change this message and point to the storagecluster spec which can be used for the same purpose.

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -194,7 +184,6 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&appsv1.Deployment{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
 		Owns(&corev1.Service{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
 		Owns(&corev1.ConfigMap{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
-		Watches(&ocsv1.OCSInitialization{}, enqueueStorageClusterRequest).
```
iamniting, line 197: We should also remove deployment owns from ocsinitialization.

iamniting (changes requested):

iamniting (approved):

PR #2242, opened 2023-10-30
Add resource profiles to control the resourceRequirements for ceph pods

malayparida2000:
```
• Add an storageCluster api field to specify the resource profile.
• The field can only take values "lean", "balanced" and "performance".
• Accordingly, the resourceRequirements for the ceph daemons are set,
  currently only mgr, mon, osd, mds and rgw pods are supported.
• It's checked if custom resourceRequirements are specified in the CR,
  if yes, then those are used. If not, then profiles are used.
• If no resource profile is specified, balanced profile values are used.
```

### DIFF
--- a/api/v1/storagecluster_types.go
+++ b/api/v1/storagecluster_types.go
@@ -43,14 +43,21 @@ type StorageClusterSpec struct {
 	// Placement is optional and used to specify placements of OCS components explicitly
 	Placement rookCephv1.PlacementSpec `json:"placement,omitempty"`
 	// Resources follows the conventions of and is mapped to CephCluster.Spec.Resources
-	Resources          map[string]corev1.ResourceRequirements `json:"resources,omitempty"`
-	Encryption         EncryptionSpec                         `json:"encryption,omitempty"`
-	StorageDeviceSets  []StorageDeviceSet                     `json:"storageDeviceSets,omitempty"`
-	MonPVCTemplate     *corev1.PersistentVolumeClaim          `json:"monPVCTemplate,omitempty"`
-	MonDataDirHostPath string                                 `json:"monDataDirHostPath,omitempty"`
-	Mgr                *MgrSpec                               `json:"mgr,omitempty"`
-	MultiCloudGateway  *MultiCloudGatewaySpec                 `json:"multiCloudGateway,omitempty"`
-	NFS                *NFSSpec                               `json:"nfs,omitempty"`
+	Resources map[string]corev1.ResourceRequirements `json:"resources,omitempty"`
+	// Resource Profile can be used to choose from a set of predefined resource profiles for the ceph daemons.
+	// We have 3 profiles
+	// lean: suitable for clusters with limited resources,
+	// balanced: suitable for most use cases,
+	// performance: suitable for clusters with high amount of resources.
+	// +kubebuilder:validation:Enum=lean;Lean;balanced;Balanced;performance;Performance
+	ResourceProfile    string                        `json:"resourceProfile,omitempty"`
+	Encryption         EncryptionSpec                `json:"encryption,omitempty"`
+	StorageDeviceSets  []StorageDeviceSet            `json:"storageDeviceSets,omitempty"`
+	MonPVCTemplate     *corev1.PersistentVolumeClaim `json:"monPVCTemplate,omitempty"`
+	MonDataDirHostPath string                        `json:"monDataDirHostPath,omitempty"`
+	Mgr                *MgrSpec                      `json:"mgr,omitempty"`
+	MultiCloudGateway  *MultiCloudGatewaySpec        `json:"multiCloudGateway,omitempty"`
+	NFS                *NFSSpec                      `json:"nfs,omitempty"`
 	// Monitoring controls the configuration of resources for exposing OCS metrics
 	Monitoring *MonitoringSpec `json:"monitoring,omitempty"`
 	// Version specifies the version of StorageCluster
--- a/controllers/storagecluster/cephcluster.go
+++ b/controllers/storagecluster/cephcluster.go
@@ -723,7 +723,7 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster, serverVersion *version.
 	for _, ds := range storageDeviceSets {
 		resources := ds.Resources
 		if resources.Requests == nil && resources.Limits == nil {
-			resources = defaults.DaemonResources["osd"]
+			resources = defaults.GetProfileDaemonResources("osd", sc)
 		}
 
 		portable := ds.Portable
@@ -875,7 +875,7 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster, serverVersion *version.
 			ds := rookCephv1.StorageClassDeviceSet{}
 			ds.Name = failureDomainValue
 			ds.Count = 1
-			ds.Resources = defaults.DaemonResources["osd"]
+			ds.Resources = defaults.GetProfileDaemonResources("osd", sc)
 			// passing on existing defaults from existing devcicesets
 			ds.TuneSlowDeviceClass = sc.Spec.StorageDeviceSets[0].Config.TuneSlowDeviceClass
 			ds.TuneFastDeviceClass = sc.Spec.StorageDeviceSets[0].Config.TuneFastDeviceClass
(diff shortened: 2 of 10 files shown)


controllers/storagecluster/resourceprofile.go
```diff
@@ -0,0 +1,199 @@
...
+	resourceProfileChangeStartTime   time.Time
+	resourcePrfoileChangeTimeout     = 30 * time.Minute
+)
+
+const (
+	resourceProfileChangeApplying = "New resource profile is being applied"
+	resourceProfileChangeFailed   = "New resource profile failed to apply, please revert to the last working resource profile"
+)
+
+func (r *StorageClusterReconciler) ensureResourceProfileChangeApplied(sc *ocsv1.StorageCluster) error {
```
travisn, line 28: Why do we need to verify every detail of the resources applied? Can't we just confirm that the reconcile is completed (the CephCluster CR completion status), and perhaps as one extra step confirm that no ceph pods are pending? We should trust that the rook operator will do its job. A simple pod query with the labelSelector `rook_cluster=<namespace>` may be sufficient, then check no pods are in pending status.

malayparida2000: Tried this, Here are the issues

1. When the cephcluster spec is updated it takes a few seconds for rook to mark it as progressing, So in this brief moment both our checks pass i. Cephcluster being Ready, ii, None of the pods with the label are pending. Hence ocs operator thinks the resource profile has been applied successfully. While the truth is nothing has started out yet.

2. Even if we bypass the above problem somehow when rook starts processing the changes, The pod restarts are done one by one and in between two restarts there are brief moments when the cephcluster reports as ready & non of the pods with the label are pending. So OCS can again take it as a false positive & report the new profile has been applied already when it's actually in the halfway stage.

3. When we are trying to do a get on the pods with the label rook_cluster=<namespace> we are also getting crashcollectors, exporters, osd-prepare pods which are not actually being affected by this profile mechanism & if they are pending due to some other reason ocs will still report that it failed to apply the resource profile while it's not the case.

api/v1/storagecluster_types.go
```diff
@@ -171,6 +178,8 @@ type ManagedResourcesSpec struct {
 // ManageCephCluster defines how to reconcile the Ceph cluster definition
 type ManageCephCluster struct {
 	ReconcileStrategy string `json:"reconcileStrategy,omitempty"`
+	// +kubebuilder:validation:Enum=1;2
+	MgrCount int `json:"mgrCount,omitempty"`
```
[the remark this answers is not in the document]
malayparida2000: I think marking something as omitempty makes it optional as well,
ref-https://book.kubebuilder.io/reference/markers.html#:~:text=The%20most%20reliable%20way%20in%201.x%20to%20get%20%2Boptional%20is%20omitempty.

malayparida2000:
/hold
Remove hold after https://github.com/red-hat-storage/ocs-operator/pull/2257 is merged

controllers/storagecluster/resourceprofile.go
travisn: Here is another approach... The CephCluster CR allows setting [labels](https://github.com/rook/rook/blob/master/deploy/examples/cluster.yaml#L202-L208) for different types of daemons. What if you set the name of the profile as a label for `all` components? For example, a label for `odf-profile=lean|balanced|performance`. Then all you need is a single query to see if daemons have the expected profile.

malayparida2000: Tried it, Using the labels looks like a better idea than comparing the ResourceRequirements. Have updated the code. But, instead of all have given labels to specifically mon, mgr, osd on the cephcluster CR & mds on the filesystem CR, rgw on the objectstore CR.

controllers/storagecluster/resourceprofile.go
```diff
@@ -0,0 +1,138 @@
...
+
+	// If we are here, then all the ceph daemons have the correct count of pods with the new resource profile
+	// New resource profile has been applied successfully
+	lastResourceProfile = currentResourceProfile
+	resourceProfileChangeisHappening = false
+	r.Log.Info("Resource profile change applied successfully", "current", currentResourceProfile, "last", lastResourceProfile)
+	return nil
+}
+
+func (r *StorageClusterReconciler) listAndFilterPods(namespace string, labelSelector map[string]string) (*corev1.PodList, error) {
```
iamniting, line 114: Lets change this func name, maybe `getRunningPods`

controllers/storagecluster/resourceprofile.go
```diff
@@ -0,0 +1,138 @@
...
+)
+
+var (
+	lastResourceProfile              string
+	resourceProfileChangeisHappening bool
+	resourceProfileChangeStartTime   time.Time
+)
+
+const (
+	resourceProfileChangeApplying = "New resource profile is being applied"
```
iamniting, line 22: Can you define these as an error?

controllers/storagecluster/resourceprofile.go
```diff
@@ -0,0 +1,138 @@
...
+
+	// Get all the osd pods with the current resource profile label
+	allOsdPodList, err := r.listAndFilterPods(sc.Namespace, map[string]string{"app": "rook-ceph-osd", defaults.ODFResourceProfileKey: currentResourceProfile})
+	if err != nil {
+		return err
+	}
+	for _, storageDeviceSet := range sc.Spec.StorageDeviceSets {
+		// Get the osds for a particular storage device set
+		osdPodList := &corev1.PodList{}
+		pattern := regexp.MustCompile("^" + storageDeviceSet.Name + "-\\d+$")
```
iamniting, line 65: Can you pls explain this regex via adding the comment in the code?

controllers/storagecluster/resourceprofile.go
```diff
@@ -0,0 +1,138 @@
...
+	r.Log.Info("Applying new resource profile", "current", currentResourceProfile, "last", lastResourceProfile)
+
+	// if the resource profile change is taking more than 20 minutes, then say it failed
+	if time.Since(resourceProfileChangeStartTime).Minutes() > resourceProfileChangeTimeout.Minutes() {
+		r.Log.Error(fmt.Errorf(resourceProfileChangeFailed), "current", currentResourceProfile, "last", lastResourceProfile)
+		return fmt.Errorf(resourceProfileChangeFailed)
+	}
+
+	// Verify if expected number of mgr pods with the current resource profile label are running
+	mgrPodList, err := r.listAndFilterPods(sc.Namespace, map[string]string{"app": "rook-ceph-mgr", defaults.ODFResourceProfileKey: currentResourceProfile})
```
iamniting, line 38: I think we can create a common func for verifying these pods.

iamniting (changes requested):

controllers/storagecluster/cephcluster.go
```diff
@@ -461,9 +461,12 @@ func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, nodeCount int, s
 				rookCephv1.KeyOSD: systemNodeCritical,
 			},
 			Resources: newCephDaemonResources(sc),
-			ContinueUpgradeAfterChecksEvenIfNotHealthy: true,
+			ContinueUpgradeAfterChecksEvenIfNotHealthy: !resourceProfileChangeisHappening,
 			LogCollector: logCollector,
 			Labels: rookCephv1.LabelsSpec{
+				rookCephv1.KeyMgr:        rookCephv1.Labels{defaults.ODFResourceProfileKey: sc.Spec.ResourceProfile},
+				rookCephv1.KeyMon:        rookCephv1.Labels{defaults.ODFResourceProfileKey: sc.Spec.ResourceProfile},
+				rookCephv1.KeyOSD:        rookCephv1.Labels{defaults.ODFResourceProfileKey: sc.Spec.ResourceProfile},
```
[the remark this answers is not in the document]
alfonsomthd: @iamniting from the UI perspective, when creating a new cluster the customer will always have to choose a profile and the backend will receive it.
For existing clusters that do not have a `resourceProfile` set yet, we should determine if they were deployed as `minimal` i.e. setting these `ResourceRequirements` (that no longer apply once this feature is merged):
https://github.com/red-hat-storage/odf-console/blob/master/packages/odf/types/common.ts#L42-L74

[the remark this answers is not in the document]
iamniting: @alfonsomthd We should not determine if they are in minimal mode or not as we are setting the values for 5 daemons. There are fewer chances that we will have the same values as the `minimal`. Moreover, if someone has set some values that are not default they should always be treated as custom instead of assigning any profile to them automatically.

If someone has not set any values we should treat it as balanced which is default now.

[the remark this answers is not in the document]
malayparida2000: Hey @alfonsomthd , The minimal mode that was being deployed by the UI was nothing but specifying the daemon resources manually in the storagecluster CR. With this profile mechanism also we respect the manual values added in the CR first ahead of the profiles. So if there are any clusters present that had deployed on that minimal mode from UI their clusters will work fine even after the upgrade as their daemons will still use the values that are present in the CR.

malayparida2000:
@iamniting Along with addressing all your suggestions I have made the 4th commit that was about Ensuring the resource profile changes are applied correctly without any issues, a separate PR here https://github.com/red-hat-storage/ocs-operator/pull/2283. I felt the scope of this PR was getting too big & that was becoming a problem in getting it reviewed. Also as this is a blocker PR for the UI changes making it small with only the necessary things will ensure it getting merged sooner.
Once this PR is merged we can go ahead with the handling the profile change PR.

controllers/storagecluster/cephcluster.go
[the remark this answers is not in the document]
alfonsomthd: @iamniting What I want to point out is that if we deployed as minimal was due to the cluster having very low resources. If now we blindly apply Balanced to that cluster, chances are that we could face pod scheduling issues, right?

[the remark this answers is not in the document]
iamniting: We will not make it balanced. It will still stay as it is. As custom resources were set there.

iamniting (approved):

PR #2249, opened 2023-11-03
storageclassrequest: refactor and unit testing

jarrpa:
This PR is a follow-up to #2228 that addresses some remaining discussion points. First is a refactor to better define the "Initialization" phase of the reconciliation loop. Second is the introduction of unit tests for the controller, as well as a small bug fix that was discovered because of the new tests.

### DIFF
--- a/controllers/storageclassrequest/storageclassrequest_controller.go
+++ b/controllers/storageclassrequest/storageclassrequest_controller.go
@@ -162,14 +162,10 @@ func (r *StorageClassRequestReconciler) SetupWithManager(mgr ctrl.Manager) error
 		Complete(r)
 }
 
-func (r *StorageClassRequestReconciler) reconcilePhases() (reconcile.Result, error) {
-	r.log.Info("Running StorageClassRequest controller in Converged/Provider Mode")
-
-	r.StorageClassRequest.Status.Phase = v1alpha1.StorageClassRequestInitializing
-
+func (r *StorageClassRequestReconciler) initPhase(storageProfile *v1.StorageProfile) error {
 	gvk, err := apiutil.GVKForObject(&v1alpha1.StorageConsumer{}, r.Client.Scheme())
 	if err != nil {
-		return reconcile.Result{}, fmt.Errorf("failed to get gvk for consumer  %w", err)
+		return fmt.Errorf("failed to get gvk for consumer  %w", err)
 	}
 	// reading storageConsumer Name from StorageClassRequest ownerReferences
 	ownerRefs := r.StorageClassRequest.GetOwnerReferences()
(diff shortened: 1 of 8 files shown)


jarrpa:
@nb-ohad @Madhu-1 okay, should be good to go, PTAL.

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -269,6 +264,20 @@ func (r *StorageClassRequestReconciler) reconcilePhases() (reconcile.Result, err
 		r.cephResourcesByName[cephResourceSpec.Name] = cephResourceSpec
 	}
 
+	return nil
+}
+
+func (r *StorageClassRequestReconciler) reconcilePhases() (reconcile.Result, error) {
+	r.log.Info("Running StorageClassRequest controller in Converged/Provider Mode")
```
Madhu-1, line 271:
```suggestion
        r.log.Info("Running StorageClassRequest controller in Provider Mode")
```

Madhu-1 (commented):
LGTM

controllers/storageclassrequest/storageclassrequest_controller_test.go
```diff
@@ -0,0 +1,493 @@
...
+			scrType:         "sharedfilesystem",
+			profileName:     "nope",
+			failureExpected: true,
+		},
+	}
+
+	for _, c := range primaryTestCases {
+		caseCounter++
+		caseLabel := fmt.Sprintf("Case %d: %s", caseCounter, c.label)
+		fmt.Println(caseLabel)
```
leelavg, line 167: why not subtest this and remove print and counter?
`t.Run(caseLabel, func(t * testing.T) {})`

same for remaining tests.

controllers/storageclassrequest/storageclassrequest_controller_test.go
```diff
@@ -0,0 +1,493 @@
...
+
+	r := createFakeReconciler(t)
+	r.Cache = &informertest.FakeInformers{Scheme: r.Scheme}
+	r.Client = fake.NewClientBuilder().WithScheme(r.Scheme).Build()
+
+	req := reconcile.Request{}
+	req.Name = "nope"
+	req.Namespace = r.OperatorNamespace
+	_, err = r.Reconcile(context.TODO(), req)
+	assert.NoError(t, err, caseLabel)
```
leelavg, line 217: I might be wrong, doesn't existing watches stop these requests getting queued in the first place or are we testing something different, like reconciler just runs with no SCR in the store?

controllers/storageclassrequest/storageclassrequest_controller_test.go
```diff
@@ -0,0 +1,493 @@
...
+		{
+			label:            "Request status has CephResource that doesn't exist",
+			expectedPoolName: "test-blockpool",
+			cephResources: []*v1alpha1.CephResourcesSpec{
+				{
+					Name: "test-blockpool",
+					Kind: "CephBlockPool",
+				},
+			},
+		},
```
leelavg, line 332: - when will this condition arise? I'm a but confused as we don't create objects but a pool exist
- is it something like, a failed cleanup?
- same for `TestCephFsSubVolGroup`

leelavg (commented):
should this PR need a BZ since we are past FF 🤔?

nb-ohad:
@jarrpa 
I see that the first thing `initPhase` is doing is finding a storageClassRequest owner of a specific kind. 
This seems like a prime candidate to move into a more generalized util function (in the relevant util file) that can be invoked similar to this:
```  
owners, err : = GetOwnersOfKind(r.StorageClassRequest, &v1alpha1.StorageConsumer{}, r.Client.Scheme())
```
Of course, this util should accept `client.Object` for both the `obj` and `owner` params.

PR #2257, opened 2023-11-07
Add & Change options to provide custom MgrCount, MonCount & no. of CephFSActiveMDS, ObjectStoreGateways

malayparida2000:


controllers/storagecluster/cephcluster.go
```diff
@@ -708,15 +713,13 @@ func getMonCount(monCount int, arbiter bool) int {
...
-	if arbiter {
-		return defaults.ArbiterModeMonCount
-	}
-
+	monCount := sc.Spec.ManagedResources.CephCluster.MonCount
 	if monCount != 0 {
 		return monCount
 	}
-
+	if arbiterEnabled(sc) {
```
travisn, line 720: Let's check for `arbiterEnabled()` before the custom setting. In stretch mode there must be exactly 5 mons and they can't override it.

controllers/storagecluster/cephobjectstores.go
```diff
@@ -186,7 +182,7 @@ func (r *StorageClusterReconciler) newCephObjectStoreInstances(initData *ocsv1.S
 							"service.kubernetes.io/topology-mode":                "Auto",
 						},
 					},
-					Instances: gatewayInstances,
+					Instances: getCephObjectStoreGatewayInstances(initData),
 					Placement: getPlacement(initData, "rgw"),
```
travisn, line 186: What placement is assigned to the RGW? Is there is required pod anti-affinity it could be a problem if they have too high a daemon count, but it's fine if it's just preferred anti-affinity.

travisn (changes requested):

controllers/storagecluster/cephobjectstores.go
malayparida2000: We do actually apply required pod anti-affinity when it's more than 1 rgw, I think it will create a problem if someone specifies more number of rgw daemons than the number of failure domains they have. As in ODF we know we always have minimum 3 failure domains, Should we add maxvalue validator to the filed making it possible to only enter values upto 3?

https://github.com/red-hat-storage/ocs-operator/blob/b5f02792f5c46810e84508d46134608123cd1423/controllers/storagecluster/placement.go#L64-L77

malayparida2000: Or may be we can modify the code to not apply the required anti affinity when the rgw count is more than the failure domains, Also the mdscount*2 is more than the failure domains.

malayparida2000: Updated the code so that if mds or rgw pods are more than the number of failure domains, don't add required pod anti affinity.

controllers/storagecluster/cephcluster.go
```diff
@@ -690,14 +690,24 @@ func getMinimumNodes(sc *ocsv1.StorageCluster) int {
...
 	return defaults.DefaultMgrCount
 }
 
-func getMonCount(monCount int, arbiter bool) int {
+func getMonCount(sc *ocsv1.StorageCluster) int {
+	if arbiterEnabled(sc) {
+		return defaults.ArbiterModeMonCount
+	}
 	// return static value if overridden
 	override := os.Getenv(monCountOverrideEnvVar)
```
travisn, line 712: What is this env var used for? Seems like it's not necessary if the spec allows overriding it?

controllers/storagecluster/placement.go
```diff
@@ -27,6 +27,15 @@ func getPlacement(sc *ocsv1.StorageCluster, component string) rookCephv1.Placeme
 		placement.PodAntiAffinity = &corev1.PodAntiAffinity{}
 	}
 
+	// if number of mds or rgw pods is greater than number of failure domains, then disable requiredDuringScheduling pod anti-affinity
+	failureDomains := len(sc.Status.FailureDomainValues)
+	if (component == "mds" && 2*getActiveMetadataServers(sc) > failureDomains) || (component == "rgw" && getCephObjectStoreGatewayInstances(sc) > failureDomains) {
+		if placement.PodAntiAffinity != nil {
+			placement.PodAntiAffinity.RequiredDuringSchedulingIgnoredDuringExecution = nil
```
travisn, line 34: I wonder if it would be sufficient to just always set preferred anti-affinity. Then we don't need complex calculations like this to figure out if we should have required affinity. I don't see a real need to use required anti-affinity.

controllers/storagecluster/cephcluster.go
malayparida2000: Yeah, I also saw this and tried to dig for more details. But just found a 4 year old PR https://github.com/red-hat-storage/ocs-operator/pull/371, where it was added. There is no other mention of this anywhere else in our docs etc. As we will be providing a better way in the CR itself good to remove this.

controllers/storagecluster/placement.go
malayparida2000: Actually, I saw the preferred anti-affinity is not really that reliable. While trying to test this on my setup, I tried to increase the active MDS count to 3, which would bring 6 pods total. I have 3 nodes, and the pods came up like 3+2+1. So there is a risk of skewing of pods towards a particular node, which someone might not like. So I think we should skip the required anti-affinity only when it's required.

travisn: If there are more daemons than the number of nodes, the preferred anti-affinity does nothing to spread the daemons evenly, that is correct. We would need to use topology spread constraints if we cared about even spread. But at least with preferred anti-affinity, it allows additional daemons to be scheduled. And when there are more daemons than nodes, there is at least one daemon per node. If there is no anti-affinity, then the daemons could all end up on the same node, which is worse than at least one per node.

malayparida2000: So we should be good with removing the required anti affinity all together in case of mds & rgw?

travisn (commented):
LGTM after the CI passes

malayparida2000:
> LGTM after the CI passes

fixed the CI

travisn (approved):
/approve
/lgtm

iamniting (approved):

PR #2258, opened 2023-11-07
Set ROOK_CURRENT_NAMESPACE_ONLY dynamically from the configmap

malayparida2000:
Jira- https://issues.redhat.com/browse/RHSTOR-5007

The option will be set to false if we have multiple storageClusters,
else it will be set to true.
Also Add a field IsMultpleStorageClusters to the StorageClusterReconciler 
This helps in easily setting a lot of options inside the reconciler
based on if there are multiple storage clusters or not.


controllers/storagecluster/reconcile.go
```diff
@@ -171,6 +171,9 @@ func (r *StorageClusterReconciler) Reconcile(ctx context.Context, request reconc
 		r.Log.Error(err, "Failed to get clusters")
 		return reconcile.Result{}, err
 	}
+	if len(r.clusters.GetStorageClusters()) > 1 {
+		r.IsMultpleStorageClusters = true
```
iamniting, line 175: Lets add a defer statement here to reset its value.

iamniting (changes requested):

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -80,19 +80,20 @@ type ImageMap struct {
...
+	serverVersion            *version.Info
+	conditions               []conditionsv1.Condition
+	phase                    string
+	nodeCount                int
+	platform                 *Platform
+	images                   ImageMap
+	recorder                 *util.EventReporter
+	OperatorCondition        conditions.Condition
+	IsNoobaaStandalone       bool
+	IsMultpleStorageClusters bool
```
iamniting, line 95: Can you pls fix the typo? `i`

iamniting (changes requested):
Can you pls fix the typo in the first commit msg as well?

iamniting (approved):

PR #2262, opened 2023-11-08
`StorageProfile` unit tests

raaizik:
### Unit tests covering `StorageProfile` API 

  - [x]  CephFS tests ✔️ 
  - [x]  `StorageClassRequest` tests ✔️
- Related PRs: #2204 #2231 #2274


controllers/storagecluster/initialization_reconciler_test.go
```diff
@@ -236,13 +342,24 @@ func createFakeInitializationStorageClusterReconcilerWithPlatform(t *testing.T,
...
+	statusSubresourceObjs := []client.Object{sc}
+	if storageProfiles != nil {
+		for _, sp := range storageProfiles.Items {
+			statusSubresourceObjs = append(statusSubresourceObjs, &sp)
+		}
+	}
+
 	obj = append(obj, mockNodeList.DeepCopy(), cbp, cfs, cnfs, cnfsbp, cnfssvc, infrastructure, networkConfig)
-	client := fake.NewClientBuilder().WithScheme(scheme).WithRuntimeObjects(obj...).WithStatusSubresource(sc).Build()
+	client := fake.NewClientBuilder().WithScheme(scheme).WithRuntimeObjects(obj...).WithStatusSubresource(statusSubresourceObjs...).Build()
```
raaizik, line 353: I'm wondering whether I've stumped upon a bug here: the only way I was able to get the `StorageProfile`s' status to go through the update via the `StorageCluster`'s reconciler [below](https://github.com/red-hat-storage/ocs-operator/blob/fusion-hci-4.14/controllers/storagecluster/cephfilesystem.go#L71-L85) was to append them along with `StorageCluster` here. Usually it's a one-to-one ratio for resources that have their own controller, e.g. for `OCSInit` that's also being `watch`ed by `StorageCluster`. 
```go
r.Client.Status().Update(r.ctx, &storageProfile)
``` 
@nb-ohad thoughts?

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -406,7 +406,8 @@ func (r *StorageClassRequestReconciler) reconcileCephFilesystemSubVolumeGroup(st
 				break
 			}
 		}
-		if dataPool == nil {
+		// since data pool alone is non-nil upon init, the following covers it
+		if len(dataPool.PoolSpec.DeviceClass) == 0 {
```
nb-ohad, line 410: This fix is unrelated to the unit tests changes. Can we please separate this change into its own PR.

Also, I believe a better fix will be to change link 402 to `var dataPool *rookCephv1.NamedPoolSpec = nil` and have the simpler check `if dataPool == nil` here. 

My reason is that  there is no need to create an instance of the `NamedPoolSpec` struct just for the find loop to override it, or to have the more complex check

nb-ohad (changes requested):

controllers/storageclassrequest/storageclassrequest_controller.go
raaizik: IMO it's related, @nb-ohad:
When I'm testing a "rejected" storage profile, it has a blank device class. Let's assume for the sake of the test that it's also the default storage profile of a storage cluster. What does that tell us? That while reconciling a ceph filesystem sub volume group, it will fail to fetch a matching `cephfilesystem` with the said blank device class since we discard rejected storage profiles from it. It is at that exact point in time that I expect an error to occur.

What you're basically saying is that there can't be a state where a default storage profile of a storage cluster is set to a "rejected' one, but currently nothing prevents it

raaizik: Got it, done

iamniting:
Can you pls resolve the merge conflicts, Also I would like to see only new commits pls rebase as well. Moreover, why do we need to merge this in the fusion hci branch now, why not in the main directly once #2278 is merged?

nb-ohad:
> Can you pls resolve the merge conflicts, Also I would like to see only new commits pls rebase as well. Moreover, why do we need to merge this in the fusion hci branch now, why not in the main directly once #2278 is merged?

@iamniting all of the above is correct. The reason we are waiting is that the needed content is not yet in main (PR not merged yet). When the content is in main, this contribution will be rebased against main which will solve all of the issues you describe.

PR #2264, opened 2023-11-14
StorageCluster spec change and updating Noobaa reconcile

vbnrh:
This commit adds changes to StorageCluster api for supporting additional parameters for Noobaa to use external Postgres server and the parameters are passed to Noobaa CR

https://issues.redhat.com/browse/RHSTOR-4749


api/v1/storagecluster_types.go
```diff
@@ -377,6 +377,31 @@ type MultiCloudGatewaySpec struct {
...
+	PGSecretNamespace string `json:"pgSecretNamespace,omitempty"`
+	// AllowSelfSignedCerts will allow the Postgres server to use self signed certificates to authenticate
+	// +optional
+	AllowSelfSignedCerts bool `json:"allowSelfSignedCerts,omitempty"`
+	// TLSSecret stores the secret name which contains the client side certificates if enabled
+	// +optional
+	TLSSecret string `json:"tlsSecret,omitempty"`
+	// TLSecretNamespace is the namespace where the TLS secret lives
+	// +optional
+	TLSSecretNamespace string `json:"tlsSecretNamespace,omitempty"`
```
Madhu-1, line 404: The namespace mentioned in the spec is the same namespace where storagecluster CR is created or it can be anything?

api/v1/storagecluster_types.go
```diff
@@ -377,6 +377,31 @@ type MultiCloudGatewaySpec struct {
...
+	ExternalPgConfig ExternalPGSpec `json:"externalPgConfig,omitempty"`
+}
+
+type ExternalPGSpec struct {
+	// Enables Noobaa to use external Postgres
+	// +optional
+	Enable bool `json:"enable,omitempty"`
+	// PGSecret stores the secret name which contains connection string of the Postgres server
+	// +optional
+	PGSecret string `json:"pgSecret,omitempty"`
```
Madhu-1, line 392: `PGSecret` to `PGSecretName`

controllers/storagecluster/noobaa_system_reconciler.go
```diff
@@ -172,7 +172,7 @@ func (r *StorageClusterReconciler) setNooBaaDesiredState(nb *nbv1.NooBaa, sc *oc
 	}
 
 	// Override with MCG options specified in the storage cluster spec
-	if sc.Spec.MultiCloudGateway != nil {
+	if sc.Spec.MultiCloudGateway.DbStorageClassName != "" {
```
Madhu-1, line 175: you still need `nil` check as its a pointer

controllers/storagecluster/noobaa_system_reconciler.go
```diff
@@ -193,6 +193,14 @@ func (r *StorageClusterReconciler) setNooBaaDesiredState(nb *nbv1.NooBaa, sc *oc
 		nb.Spec.DisableLoadBalancerService = sc.Spec.MultiCloudGateway.DisableLoadBalancerService
 	}
 
+	if sc.Spec.MultiCloudGateway.ExternalPgConfig.Enable {
```
Madhu-1, line 196: you need `nil` check for `sc.Spec.MultiCloudGateway` here as well

api/v1/storagecluster_types.go
vbnrh: It can be anything. Noobaa takes in a secret reference so name and namespace can be anything.

controllers/storagecluster/noobaa_system_reconciler.go
vbnrh: done

controllers/storagecluster/noobaa_system_reconciler.go
vbnrh: done

api/v1/storagecluster_types.go
vbnrh: done and same with TLSSecret

controllers/storagecluster/noobaa_system_reconciler.go
```diff
@@ -172,7 +172,7 @@ func (r *StorageClusterReconciler) setNooBaaDesiredState(nb *nbv1.NooBaa, sc *oc
 	}
 
 	// Override with MCG options specified in the storage cluster spec
-	if sc.Spec.MultiCloudGateway != nil {
+	if sc.Spec.MultiCloudGateway != nil && sc.Spec.MultiCloudGateway.DbStorageClassName != "" {
```
umangachapagain, line 175: Is this change needed for this feature to work?

vbnrh:
Screenshot 2023-11-14 at 7 45 51 PM
Not sure why it is CI is failing. All changes have already been commited

api/v1/storagecluster_types.go
```diff
@@ -377,6 +377,31 @@ type MultiCloudGatewaySpec struct {
...
+
+	// Allows Noobaa to connect to an external Postgres server
+	// +optional
+	ExternalPgConfig *ExternalPGSpec `json:"externalPgConfig,omitempty"`
+}
+
+type ExternalPGSpec struct {
+	// Enables Noobaa to use external Postgres
+	// +optional
+	Enable bool `json:"enable,omitempty"`
```
umangachapagain, line 389: Can we drop this field and assume that we need external postgres when `PGSecretName` and `PGSecretNamespace` are not nil?
I don't see why I would set `enable:false` and provide connection info or vice-versa.

controllers/storagecluster/noobaa_system_reconciler.go
vbnrh: It is not related to the feature but to avoid the loop since it will not be nil if pgsecrets are not nil

api/v1/storagecluster_types.go
vbnrh: updated. i added the enable to have easy to read checks in future.

api/v1/storagecluster_types.go
Madhu-1: i dont know if we need to support this one or not, i assume we always want the users to make use of the namespace where storagecluster is created /odf is installed.

controllers/storagecluster/noobaa_system_reconciler.go
```diff
@@ -193,6 +193,14 @@ func (r *StorageClusterReconciler) setNooBaaDesiredState(nb *nbv1.NooBaa, sc *oc
 		nb.Spec.DisableLoadBalancerService = sc.Spec.MultiCloudGateway.DisableLoadBalancerService
 	}
 
+	if sc.Spec.MultiCloudGateway != nil && sc.Spec.MultiCloudGateway.ExternalPgConfig != nil && sc.Spec.MultiCloudGateway.ExternalPgConfig.PGSecretName != "" {
+		nb.Spec.ExternalPgSSLRequired = true
+		nb.Spec.ExternalPgSecret = &corev1.SecretReference{Name: sc.Spec.MultiCloudGateway.ExternalPgConfig.PGSecretName, Namespace: sc.Spec.MultiCloudGateway.ExternalPgConfig.PGSecretNamespace}
+		nb.Spec.ExternalPgSSLUnauthorized = sc.Spec.MultiCloudGateway.ExternalPgConfig.AllowSelfSignedCerts
+		if sc.Spec.MultiCloudGateway.ExternalPgConfig.TLSSecretName != "" && sc.Spec.MultiCloudGateway.ExternalPgConfig.TLSSecretNamespace != "" {
+			nb.Spec.ExternalPgSSLSecret = &corev1.SecretReference{Name: sc.Spec.MultiCloudGateway.ExternalPgConfig.TLSSecretName, Namespace: sc.Spec.MultiCloudGateway.ExternalPgConfig.TLSSecretNamespace}
```
Madhu-1, line 201: Do we need a check here to ensure that both name and namespace are set not either one of them?

controllers/storagecluster/noobaa_system_reconciler.go
```diff
@@ -193,6 +193,14 @@ func (r *StorageClusterReconciler) setNooBaaDesiredState(nb *nbv1.NooBaa, sc *oc
 		nb.Spec.DisableLoadBalancerService = sc.Spec.MultiCloudGateway.DisableLoadBalancerService
 	}
 
+	if sc.Spec.MultiCloudGateway != nil && sc.Spec.MultiCloudGateway.ExternalPgConfig != nil && sc.Spec.MultiCloudGateway.ExternalPgConfig.PGSecretName != "" {
+		nb.Spec.ExternalPgSSLRequired = true
+		nb.Spec.ExternalPgSecret = &corev1.SecretReference{Name: sc.Spec.MultiCloudGateway.ExternalPgConfig.PGSecretName, Namespace: sc.Spec.MultiCloudGateway.ExternalPgConfig.PGSecretNamespace}
+		nb.Spec.ExternalPgSSLUnauthorized = sc.Spec.MultiCloudGateway.ExternalPgConfig.AllowSelfSignedCerts
+		if sc.Spec.MultiCloudGateway.ExternalPgConfig.TLSSecretName != "" && sc.Spec.MultiCloudGateway.ExternalPgConfig.TLSSecretNamespace != "" {
```
Madhu-1, line 200: what if someone sets the TLSSecret but forgets to set PGSecret details, should we error out in that case?

Madhu-1 (commented):
Left few questions,LGTM

controllers/storagecluster/noobaa_system_reconciler.go
vbnrh: It will work with either one but just in case added both to for secret ref.

controllers/storagecluster/noobaa_system_reconciler.go
vbnrh: I have changed the check above to see if PGSecret is empty or not. It won't enter this block if it is not set.

controllers/storagecluster/noobaa_system_reconciler.go
```diff
@@ -193,6 +193,14 @@ func (r *StorageClusterReconciler) setNooBaaDesiredState(nb *nbv1.NooBaa, sc *oc
 		nb.Spec.DisableLoadBalancerService = sc.Spec.MultiCloudGateway.DisableLoadBalancerService
 	}
 
+	if sc.Spec.MultiCloudGateway != nil && sc.Spec.MultiCloudGateway.ExternalPgConfig != nil && sc.Spec.MultiCloudGateway.ExternalPgConfig.PGSecretName != "" {
```
umangachapagain, line 196: need to check `PGSecretNamespace!=""` as well.

api/v1/storagecluster_types.go
jackyalbo: let's remove the namspace I will do the same for NooBaa repo as well. We can't mount secrets from different namespaces anyway.

jackyalbo: Also PGSecretNamespace should be removed

controllers/storagecluster/noobaa_system_reconciler.go
vbnrh: done

vbnrh:
Screenshot 2023-11-16 at 2 21 42 PM

umangachapagain (approved):

api/v1/storagecluster_types.go
Madhu-1: @vbnrh will you be deleting these values from the types?

vbnrh: @Madhu-1 @jackyalbo I had failed to see these reviews. I will update the values

malayparida2000:
Nooba uses a slightly older version of rook api 
https://github.com/noobaa/noobaa-operator/blob/2b58b1e61bfff592718abede1824b1efdd6c098c/go.mod#L34
In that version rook had only 1 go.mod in it's repo. Now with the latest rook changes rook has separated their repo a bit, The repo has a go.mod & their pkg dir has another go.mod. So when we are trying to bring in this noobaa version it tries to bring in that older rook version which had only 1 go.mod, But now ocs-operator has both the paths for rook added which creates an ambiguous import.

Just add
github.com/rook/rook v1.12.3-0.20230823002913-59e66edf92f7
to the exclude section in the go.mod with a suitable comment and it will work.

malayparida2000 (approved):
/lgtm

PR #2283, opened 2023-11-23
Ensure resource profile changes are applied correctly without any issues

malayparida2000:
```
• Determine resource profile change by comparing sc.Spec.ResourceProfile
  with sc.Status.LastAppliedResourceProfile. If they are different, then
  a resource profile change is in progress.
• Mark the ContinueUpgradeAfterChecksEvenIfNotHealthy flag on the
  cephCluster CR to false if a resource profile change is in progress.
• During the below process mark the storageCluster phase as Progressing.
  The storageCluster conditions will show "New resource profile is being
  applied".
• Keep checking if any of the pods with the new resource profile
  label are in pending state for more than 2 minutes. If yes, then
  assume that the resource profile change has failed. Mark the 
  storageCluster phase as Error and the storageCluster conditions will 
  show "New resource profile failed to apply, please revert to the last 
  working resource profile".
• Start Checking if the ceph dameons(mgr,mon,osd,mds,rgw) have the
  correct number of running pods where each pod has the current
  resource profile label.
• If the above checks are successful, the new resource profile has
  been applied successfully. Set the resource profile in the status &
  Mark the storageCluster phase to Ready.
```

### DIFF
--- a/controllers/storagecluster/cephcluster.go
+++ b/controllers/storagecluster/cephcluster.go
@@ -461,9 +461,13 @@ func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, nodeCount int, s
 				rookCephv1.KeyOSD: systemNodeCritical,
 			},
 			Resources: newCephDaemonResources(sc),
-			ContinueUpgradeAfterChecksEvenIfNotHealthy: true,
+			// if resource profile change is in progress, then set this flag to false
+			ContinueUpgradeAfterChecksEvenIfNotHealthy: sc.Spec.ResourceProfile == sc.Status.LastAppliedResourceProfile,
 			LogCollector: logCollector,
 			Labels: rookCephv1.LabelsSpec{
+				rookCephv1.KeyMgr:        rookCephv1.Labels{defaults.ODFResourceProfileKey: sc.Spec.ResourceProfile},
+				rookCephv1.KeyMon:        rookCephv1.Labels{defaults.ODFResourceProfileKey: sc.Spec.ResourceProfile},
+				rookCephv1.KeyOSD:        rookCephv1.Labels{defaults.ODFResourceProfileKey: sc.Spec.ResourceProfile},
 				rookCephv1.KeyMonitoring: getCephClusterMonitoringLabels(*sc),
 			},
 			CSI: rookCephv1.CSIDriverSpec{
@@ -1302,4 +1306,13 @@ func isBluestore(store rookCephv1.OSDStore) bool {
 	}
 
 	return false
+}
+
+func (r *StorageClusterReconciler) getOsdCount(sc *ocsv1.StorageCluster) int {
+	storageClassDeviceSets := newStorageClassDeviceSets(sc, r.serverVersion)
+	osdCount := 0
+	for _, ds := range storageClassDeviceSets {
+		osdCount += ds.Count
+	}
+	return osdCount
 }
--- a/controllers/storagecluster/reconcile.go
+++ b/controllers/storagecluster/reconcile.go
@@ -471,6 +471,26 @@ func (r *StorageClusterReconciler) reconcilePhases(
 			return returnRes, nil
 		}
 	}
+	// Process resource profiles only if the cluster is not external or provider mode or noobaa standalone, and if the resource profile has changed
+	if !(instance.Spec.ExternalStorage.Enable || instance.Spec.AllowRemoteStorageConsumers || r.IsNoobaaStandalone) &&
+		(instance.Spec.ResourceProfile != instance.Status.LastAppliedResourceProfile) {
+		err := r.ensureResourceProfileChangeApplied(instance)
+		if err != nil {
+			if err == errResourceProfileChangeApplying {
+				reason := ocsv1.ReconcileFailed
+				message := err.Error()
+				statusutil.SetProgressingCondition(&instance.Status.Conditions, reason, message)
+				instance.Status.Phase = statusutil.PhaseProgressing
+				return reconcile.Result{Requeue: true}, nil
+			} else if err == errResourceProfileChangeFailed {
+				reason := ocsv1.ReconcileFailed
+				message := err.Error()
+				statusutil.SetErrorCondition(&instance.Status.Conditions, reason, message)
+				instance.Status.Phase = statusutil.PhaseError
+			}
+			return reconcile.Result{}, err
+		}
+	}
 	// All component operators are in a happy state.
 	if r.conditions == nil {
 		r.Log.Info("No component operator reported negatively.")
--- a/controllers/storagecluster/resourceprofile.go
+++ b/controllers/storagecluster/resourceprofile.go
@@ -1,0 +1,96 @@
+package storagecluster
+
+import (
+	"errors"
+	"time"
+
+	ocsv1 "github.com/red-hat-storage/ocs-operator/v4/api/v1"
+	"github.com/red-hat-storage/ocs-operator/v4/controllers/defaults"
+	"github.com/red-hat-storage/ocs-operator/v4/controllers/util"
+	corev1 "k8s.io/api/core/v1"
+)
+
+const (
+	resourceProfileChangeApplyingMessage = "New resource profile is being applied"
+	resourceProfileChangeFailedMessage   = "New resource profile failed to apply, please revert to the last working resource profile"
+)
+
+var (
+	errResourceProfileChangeApplying = errors.New(resourceProfileChangeApplyingMessage)
+	errResourceProfileChangeFailed   = errors.New(resourceProfileChangeFailedMessage)
+)
+
+func (r *StorageClusterReconciler) ensureResourceProfileChangeApplied(sc *ocsv1.StorageCluster) error {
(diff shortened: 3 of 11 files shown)


malayparida2000:
The first 3 commit s are of this https://github.com/red-hat-storage/ocs-operator/pull/2242 PR.
Don't merge before that PR.
/hold

malayparida2000:
Removing hold as the mentioned PR is now merged.
/unhold

controllers/storagecluster/cephcluster.go
```diff
@@ -461,9 +461,13 @@ func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, nodeCount int, s
 				rookCephv1.KeyOSD: systemNodeCritical,
 			},
 			Resources: newCephDaemonResources(sc),
-			ContinueUpgradeAfterChecksEvenIfNotHealthy: true,
+			// if resource profile change is in progress, then set this flag to false
+			ContinueUpgradeAfterChecksEvenIfNotHealthy: sc.Spec.ResourceProfile == sc.Status.ResourceProfile,
```
travisn, line 465: The `sc.Status.ResourceProfile` is only updated after the successful reconcile of the new resource profile, right? So this is comparing the new profile with the previous profile (if it changed)?

travisn: Will this cause all upgraded clusters (the first time upgrading to 4.15) to set this to `false`? Let's keep this as `true` if we're applying the default profile and if the applied profile is blank.

controllers/storagecluster/resourceprofile.go
```diff
@@ -0,0 +1,133 @@
...
+		osdPodList := &corev1.PodList{}
+		// This regex will find the osd pods with the deviceSet label value as <storageDeviceSet.Name>-<number>
+		pattern := regexp.MustCompile("^" + storageDeviceSet.Name + "-\\d+$")
+		for _, osdPod := range allOsdPodList.Items {
+			if pattern.MatchString(osdPod.Labels["ceph.rook.io/DeviceSet"]) {
+				osdPodList.Items = append(osdPodList.Items, osdPod)
+			}
+		}
+		// Verify if expected number of osd pods with the current resource profile label are running for a particular storage device set
+		count, replica := countAndReplicaOf(&storageDeviceSet)
```
travisn, line 68: Instead of verifying the OSD pod count with the label, how about just checking if there are any pods that don't have the expected resource profile label? Or a similar question, is why are the OSD pods different from checking the mon or mgr pods?

controllers/storagecluster/cephcluster.go
malayparida2000: Yeah, the thought was that if the values on the status & the spec are not the same that means we are in the process of applying the resource profile that's in the spec.

malayparida2000: No, when someone upgrades to 4.15, They don't have the resource profile either in their spec or status. So two blank values are equal, and the flag would be true only.

controllers/storagecluster/resourceprofile.go
malayparida2000: Hm, Initially I thought about verifying them each storage device set wise, but seems like it's not really required & just adding complexity.

travisn (commented):
/lgtm

controllers/storagecluster/resourceprofile.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/red-hat-storage/ocs-operator/v4/controllers/defaults"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	resourceProfileChangeApplyingMessage = "New resource profile is being applied"
+	resourceProfileChangeFailedMessage   = "New resource profile failed to apply, please revert to the last working resource profile"
+	resourceProfileChangeTimeout         = 20 * time.Minute
```
umangachapagain, line 17: How is this value determined? 20 minutes is a lot of time for pods to restart with new resource requirements. Can we decrease it?

umangachapagain (changes requested):

controllers/storagecluster/resourceprofile.go
malayparida2000: @travisn How much time do you think will be sufficient for a typical cluster, for rook to process all the daemon restarts?

controllers/storagecluster/reconcile.go
```diff
@@ -472,6 +472,26 @@ func (r *StorageClusterReconciler) reconcilePhases(
 			return returnRes, nil
 		}
 	}
+	// Process resource profiles only if the cluster is not external or provider mode or noobaa standalone, and if the resource profile has changed
+	if !(instance.Spec.ExternalStorage.Enable || instance.Spec.AllowRemoteStorageConsumers || r.IsNoobaaStandalone) &&
+		(instance.Spec.ResourceProfile != instance.Status.ResourceProfile) {
+		err = r.ensureResourceProfileChangeApplied(instance)
+		if err != nil {
+			if err == errResourceProfileChangeApplying {
+				reason := ocsv1.ReconcileOngoing
```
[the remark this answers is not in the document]
malayparida2000: In the condition, the message will read "New resource profile is being applied",
So the reason can be something different. As we have reasons related to the word reconcile only https://github.com/red-hat-storage/ocs-operator/blob/1f682da09475c6419ff9f104fecd0caa7197b46d/api/v1/storagecluster_types.go#L559-L561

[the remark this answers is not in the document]
umangachapagain: Those only occur at specific conditions. But, reconcile happens every few seconds.
So we are almost always in "ReconcileOngoing" state, but we are only sometime in "ReconcileFailed" or "ApplyingResourceProfileChanges" state.
Even during cluster expansion, we set condition to "progressing" with message that "cluster is explanding", but we do not define "ReconcileOngoing" as a reason.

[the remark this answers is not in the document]
malayparida2000: Hmm, Then let's put the reason as Reconcile failed, As we are actually waiting for the desired state & it hasn't reached there in this reconcile, and we have to requeue and check again.

controllers/storagecluster/resourceprofile.go
travisn: Timeouts are a challenge because it depends on a number of factors, such as whether images are being pulled, how big the cluster is (number of OSDs), etc. In this case of updating the resources at least we should only be updating existing daemons and not deploying new daemons where images might need to be pulled. What about a more dynamic set of checks such as these?
1. If any daemons are in pending state, consider it failed immediately since it's likely pending due to lack of resources.
2. If no daemons have been updated in some short timeout, then consider it failed. For example, keep track of the number of daemons with the new profile label. If it hasn't increased for two minutes, then consider it failed. 

The first condition should actually be sufficient to catch the cause, so we may not need any timeout checking from the second bullet.

malayparida2000: Have removed the timeout, And implemented another logic to check if any of the coming pods with the resource profile label are stuck in pending for more than 2 mins. PTAL
@travisn @umangachapagain

controllers/storagecluster/resourceprofile.go
```diff
@@ -0,0 +1,93 @@
...
+	lastAppliedResourceProfile := sc.Status.LastAppliedResourceProfile
+	r.Log.Info("Applying new resource profile", "current", currentResourceProfile, "last", lastAppliedResourceProfile)
+
+	// If any of the pods with the current resource profile label are in pending state for more than 2 minutes, then we assume that the resource profile change has failed as the pods are stuck in pending state
+	podList, err := util.GetPodsWithLabels(r.ctx, r.Client, sc.Namespace, map[string]string{defaults.ODFResourceProfileKey: currentResourceProfile})
+	if err != nil {
+		return err
+	}
+	for _, pod := range podList.Items {
+		if pod.Status.Phase == corev1.PodPending && time.Since(pod.CreationTimestamp.Time) > 2*time.Minute {
```
travisn, line 34: If the pod has been pending for less than two minutes, is there a retry again to wait and see if it gets out of the pending state? We don't want to continue if any pod is pending, even if less than two minutes. The two minutes was just suggested as a short timeout to consider the reconcile failed instead of waiting and retrying.

malayparida2000: Ok, Adding a logic to return here itself if it's pending for less than 2 mins.

malayparida2000: I meant if less than 2 min, Return & retry. If more than 2 mins just declare failed.

travisn (commented):
/lgtm assuming the testing goes well to confirm the expected behavior.

malayparida2000:
> /lgtm assuming the testing goes well to confirm the expected behavior.

Testing wise, I have covered profile changing in both up and down directions,
And intentionally setting too high a requirement in performance mode & changing to it, to see the behavior.
Also the recovery from this situation to a previous profile.
Hope this covers the main use cases.

umangachapagain (approved):

PR #2288, opened 2023-11-28
`StorageProfile` unit tests

raaizik:
### Unit tests covering `StorageProfile` API 

  - [x]  CephFS tests ✔️ 
  - [x]  `StorageClassRequest` tests ✔️
- Related PRs: #2204 #2231 #2274

### DIFF
--- a/controllers/storageclassrequest/storageclassrequest_controller_test.go
+++ b/controllers/storageclassrequest/storageclassrequest_controller_test.go
@@ -16,6 +16,7 @@ package storageclassrequest
 import (
 	"context"
 	"fmt"
+	"strings"
 	"testing"
 
 	v1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
@@ -32,6 +33,15 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
+const (
+	pgAutoscaleMode    = "pg_autoscale_mode"
+	pgNum              = "pg_num"
+	pgpNum             = "pgp_num"
+	namespaceName      = "test-ns"
+	deviceClass        = "ssd"
+	storageProfileKind = "StorageProfile"
+)
+
 var fakeStorageProfile = &v1.StorageProfile{
 	ObjectMeta: metav1.ObjectMeta{
 		Name:      "medium",
@@ -40,6 +50,46 @@ var fakeStorageProfile = &v1.StorageProfile{
 	Spec: v1.StorageProfileSpec{
 		DeviceClass: "ssd",
 	},
+}
+
+var validStorageProfile = &v1.StorageProfile{
+	TypeMeta: metav1.TypeMeta{Kind: storageProfileKind},
+	ObjectMeta: metav1.ObjectMeta{
+		Name:      "valid",
+		Namespace: namespaceName,
+	},
+	Spec: v1.StorageProfileSpec{
+		DeviceClass: deviceClass,
+		BlockPoolConfiguration: v1.BlockPoolConfigurationSpec{
+			Parameters: map[string]string{
+				pgAutoscaleMode: "on",
+				pgNum:           "128",
+				pgpNum:          "128",
+			},
+		},
+	},
+	Status: v1.StorageProfileStatus{Phase: ""},
+}
+
+// A rejected StorageProfile is one that is invalid due to having a blank device class field and is set to
+// Rejected in its phase.
+var rejectedStorageProfile = &v1.StorageProfile{
+	TypeMeta: metav1.TypeMeta{Kind: storageProfileKind},
+	ObjectMeta: metav1.ObjectMeta{
+		Name:      "rejected",
+		Namespace: namespaceName,
+	},
+	Spec: v1.StorageProfileSpec{
+		DeviceClass: "",
+		BlockPoolConfiguration: v1.BlockPoolConfigurationSpec{
+			Parameters: map[string]string{
+				pgAutoscaleMode: "on",
+				pgNum:           "128",
+				pgpNum:          "128",
+			},
+		},
+	},
+	Status: v1.StorageProfileStatus{Phase: ""},
 }
 
 var fakeStorageCluster = &v1.StorageCluster{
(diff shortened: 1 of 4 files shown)


controllers/storagecluster/initialization_reconciler_test.go
```diff
@@ -176,22 +281,23 @@ func initStorageClusterResourceCreateUpdateTestWithPlatform(
...
 
 func createFakeInitializationStorageClusterReconciler(t *testing.T, obj ...runtime.Object) StorageClusterReconciler {
 	return createFakeInitializationStorageClusterReconcilerWithPlatform(
-		t, &Platform{platform: configv1.NonePlatformType}, obj...)
+		t, &Platform{platform: configv1.NonePlatformType}, nil, obj...)
 }
 
 func createFakeInitializationStorageClusterReconcilerWithPlatform(t *testing.T,
 	platform *Platform,
+	storageProfiles *api.StorageProfileList,
```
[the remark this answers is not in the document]
raaizik: Well, that would entail sending the `StorageProfile`s to `WithRuntimeObjects` contrary to my implementation that does _not_ treat those as runtime objects of `StorageCluster`. Furthermore, the cost of unpacking would be worse since I'd need to exclude the `StorageProfile`s from the set of runtime objects. So from my POV this exclusion would take place in either case and appending that function argument is the least costly option. Thoughts?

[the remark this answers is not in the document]
jarrpa: > Well, that would entail sending the `StorageProfile`s to `WithRuntimeObjects` contrary to my implementation that does _not_ treat those as runtime objects of `StorageCluster`.

I'm proposing the exact opposite of this, which should be in keeping with the intent of your implementation. In my proposal we would still be *not* sending any StorageProfiles to `.WithRuntimeObjects()` and keeping them in the list to `.WithStatusSubresource()`. I do agree that the proposal is more computationally intensive, but the overhead would be minimal to negligible. This is a unit testing framework, and the addition of a few milliseconds sounds like a worthwhile sacrifice for the sake of cleaner implementations.

Naturally, this is only being suggested because we rely on a `...` parameter, the same reason you're adding a new parameter to begin with.

...all that said, it would probably be even easier to just rename the new parameter to `statusObjs` or something and do the StorageProfiles loop outside this function (since, again, it's only ever used once). Thus:

```suggestion
	statusObj []runtime.Object,
```

Followed by:

```
	statusObj = append(statusObj, sc)
```
Appending to a `nil` slice [works out fine](https://go.dev/play/p/Odak8oxqkcF). From a code clarity standpoint this seems more natural and descriptive than my earlier proposal, and less of a uni-tasker. I still don't like seeing a new parameter be largely useless except for one instance, but I can deal. :laughing:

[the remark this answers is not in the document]
raaizik: I like your 1st suggestion best 😅

controllers/storageclassrequest/storageclassrequest_controller_test.go
```diff
@@ -32,26 +33,74 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
+const (
+	pgAutoscaleMode = "pg_autoscale_mode"
+	pgNum           = "pg_num"
+	pgpNum          = "pgp_num"
+	namespaceName   = "test-ns"
+	deviceClass     = "ssd"
+	kind            = "StorageProfile"
```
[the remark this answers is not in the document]
raaizik: 1. Since these consts resides in a function named `getMockStorageProfiles` that's sole purpose is to generate mock `StorageProfile`s, it's pretty derivative.
2. In order to unit test `StorageProfile`s we need to mock up resources that are a replica of the real world resource. Not sure if you're asking what is the usage of those in general or in the unit tests @rewantsoni

controllers/storageclassrequest/storageclassrequest_controller_test.go
```diff
@@ -32,26 +33,74 @@ import (
...
+				pgAutoscaleMode: "on",
+				pgNum:           "128",
+				pgpNum:          "128",
+			},
+		},
 	},
+	Status: v1.StorageProfileStatus{Phase: ""},
+}
+
+var rejectedStorageProfile = &v1.StorageProfile{
```
rchikatw, line 75: rejectedStorageProfile -> invalidStorageProfile sounds good WDYT?

controllers/storageclassrequest/storageclassrequest_controller_test.go
```diff
@@ -218,6 +267,201 @@ func TestProfileReconcile(t *testing.T) {
...
+				&rookCephv1.CephBlockPool{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "test-valid-blockpool",
+						Namespace: namespaceName,
+						Labels: map[string]string{
+							controllers.StorageConsumerNameLabel: fakeStorageConsumer.Name,
+							controllers.StorageProfileSpecLabel:  validStorageProfile.GetSpecHash(),
+						},
+					}, Spec: rookCephv1.NamedBlockPoolSpec{
+						Name: "spec",
```
rchikatw, line 296: spec -> test-spec ?

controllers/storageclassrequest/storageclassrequest_controller_test.go
```diff
@@ -218,6 +267,201 @@ func TestProfileReconcile(t *testing.T) {
...
+						Name:      "test-rejected-blockpool",
+						Namespace: namespaceName,
+						Labels: map[string]string{
+							controllers.StorageConsumerNameLabel: fakeStorageConsumer.Name,
+							controllers.StorageProfileSpecLabel:  rejectedStorageProfile.GetSpecHash(),
+						},
+					}, Spec: rookCephv1.NamedBlockPoolSpec{
+						Name: "spec",
+						PoolSpec: rookCephv1.PoolSpec{
+							FailureDomain: "zone",
```
rchikatw, line 323: zone -> test-zone?

controllers/storageclassrequest/storageclassrequest_controller_test.go
```diff
@@ -218,6 +267,201 @@ func TestProfileReconcile(t *testing.T) {
...
+		},
+	}
+
+	for _, c := range primaryTestCases {
+		caseCounter++
+		caseLabel := fmt.Sprintf("Case %d: %s", caseCounter, c.label)
+		fmt.Println(caseLabel)
+
+		r := createFakeReconciler(t)
+		if strings.Contains(c.label, "rejected") {
```
rchikatw, line 339: Can we check for specific label instead of contains

controllers/storageclassrequest/storageclassrequest_controller_test.go
```diff
@@ -218,6 +267,201 @@ func TestProfileReconcile(t *testing.T) {
...
+		_, err = r.reconcilePhases()
+		if c.failureExpected {
+			assert.Error(t, err, caseLabel)
+			continue
+		}
+		assert.NoError(t, err, caseLabel)
+
+		assert.Equal(t, c.expectedPoolName, r.cephBlockPool.Name, caseLabel)
+
+		if strings.Contains(c.expectedPoolName, "valid") {
```
rchikatw, line 361: we can check for the exact pool name, right? Any specific reason for contains

controllers/storageclassrequest/storageclassrequest_controller_test.go
```diff
@@ -218,6 +267,201 @@ func TestProfileReconcile(t *testing.T) {
...
+		expectedGroupName string
+		failureExpected   bool
+		createObjects     []runtime.Object
+		cephResources     []*v1alpha1.CephResourcesSpec
+		storageProfile    *v1.StorageProfile
+		cephFs            *rookCephv1.CephFilesystem
+	}{
+		{
+			label:             "valid profile",
+			expectedGroupName: "test-subvolgroup",
```
rchikatw, line 389: test-subvolgroup -> test-subvolumegroup sounds more good

controllers/storageclassrequest/storageclassrequest_controller_test.go
```diff
@@ -218,6 +267,201 @@ func TestProfileReconcile(t *testing.T) {
...
+
+		r.Client = fakeClient.Build()
+		r.StorageClassRequest.Status.CephResources = c.cephResources
+
+		_, err = r.reconcilePhases()
+		if c.failureExpected {
+			assert.Error(t, err, caseLabel)
+			continue
+		}
+		assert.NoError(t, err, caseLabel)
```
rchikatw, line 460: This code is not reachable in case failure, Is this how it should?

controllers/storagecluster/storagecluster_controller_test.go
```diff
@@ -192,6 +192,111 @@ var mockDeviceSets = []api.StorageDeviceSet{
...
+			},
+		},
+	}
+	spblankdeviceclass := &api.StorageProfile{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       kind,
+			APIVersion: apiVersion,
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "blank-performance",
```
rchikatw, line 289: What does blank performance mean? a default or low performance?

controllers/storageclassrequest/storageclassrequest_controller_test.go
raaizik: I wanted to emphasize the fact that invalid SPs are defined as ones that have their phase set to `Rejected` and are in essence rejected.

controllers/storageclassrequest/storageclassrequest_controller_test.go
raaizik: `subvolgroup` is aligned with the naming convention

controllers/storageclassrequest/storageclassrequest_controller_test.go
raaizik: In case `failureExpected` label of the iterated case is set to true then it hits the error assertion, otherwise, etc. What exactly is unreachable?

controllers/storagecluster/storagecluster_controller_test.go
raaizik: A `StorageProfile` is marked as `Rejected` in its phase (also referred to as invalid) when its device class field is blank.

controllers/storageclassrequest/storageclassrequest_controller_test.go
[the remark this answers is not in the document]
rewantsoni: 1. I would still suggest to have it as storageProfileKind 
2. I don't see the usage of the const pgAutoscaleMode, pgNum and pgpNum anywhere in the changes that you made

controllers/storageclassrequest/storageclassrequest_controller_test.go
raaizik: Thought performance and readability wise it's better for cases where there's a unique sub-string

controllers/storageclassrequest/storageclassrequest_controller_test.go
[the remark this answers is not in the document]
raaizik: 1. I don't see the point, sorry. These consts are accessible only in the scope of the said function that handles mock SPs generation
2.   They are used [here](https://github.com/raaizik/ocs-operator/blob/sp-tests-main/controllers/storageclassrequest/storageclassrequest_controller_test.go#L362-L364)

controllers/storageclassrequest/storageclassrequest_controller_test.go
jarrpa: I imagine the strategy is to consistently name the pools on whether they are expected to be valid or invalid/rejected per test case, and just use that as the check. Still, given that you're checking the Phase for the rejected test case below, it could make sense to also verify the Phase in the "valid" cases as well. `if c.expectedPhase == PhaseRejected` as the conditional, for example, then verify the actual values.

controllers/storageclassrequest/storageclassrequest_controller_test.go
raaizik: I think it's aligned with the naming convention

controllers/storageclassrequest/storageclassrequest_controller_test.go
raaizik: I think it's aligned with the naming convention

controllers/storageclassrequest/storageclassrequest_controller_test.go
raaizik: Resolving as per @rchikatw's manual confirmation

controllers/storageclassrequest/storageclassrequest_controller_test.go
raaizik: Resolving as per @rchikatw's manual confirmation

controllers/storageclassrequest/storageclassrequest_controller_test.go
raaizik: Resolving as per @rchikatw's manual confirmation

controllers/storageclassrequest/storageclassrequest_controller_test.go
raaizik: Resolving as per @rchikatw's manual confirmation

controllers/storagecluster/storagecluster_controller_test.go
raaizik: Resolving as per @rchikatw's manual confirmation

controllers/storageclassrequest/storageclassrequest_controller_test.go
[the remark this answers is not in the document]
raaizik: Resolving as per @rewantsoni's manual confirmation

jarrpa (approved):
All right, and now that that's cleaned up, let's FINISH THIS.

/lgtm

PR #2289, opened 2023-11-30
ensure interal cluster is only allowed in the OperatorNamespace 

iamniting:

### DIFF
--- a/controllers/storagecluster/reconcile.go
+++ b/controllers/storagecluster/reconcile.go
@@ -659,6 +659,11 @@ func (r *StorageClusterReconciler) isActiveStorageCluster(instance *ocsv1.Storag
 		return false, nil
 	}
 
+	// Ensure that the internal storageCluster is only allowed in the OperatorNamespace.
+	if !instance.Spec.ExternalStorage.Enable && instance.Namespace != r.OperatorNamespace {
+		return false, nil
+	}
+
 	var storageClusterList []ocsv1.StorageCluster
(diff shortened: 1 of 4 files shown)


umangachapagain (changes requested):
This restriction does not make sense. Why can't I have external mode in OperatorNamespace and internal mode in another namespace?

iamniting:
> This restriction does not make sense. Why can't I have external mode in OperatorNamespace and internal mode in another namespace?

You can still have external in the operator namespace. We are just blocking internal in other namespaces just to be safe if someone creates it, as it is out of our support matrix.

umangachapagain:
> > This restriction does not make sense. Why can't I have external mode in OperatorNamespace and internal mode in another namespace?
> 
> You can still have external in the operator namespace. We are just blocking internal in other namespaces just to be safe if someone creates it, as it is out of our support matrix.
>
I'd prefer not to block it, but better be safe I guess.

Can you please squash all commits? It's all part of same logical change.

iamniting:
> > > This restriction does not make sense. Why can't I have external mode in OperatorNamespace and internal mode in another namespace?
> > 
> > 
> > You can still have external in the operator namespace. We are just blocking internal in other namespaces just to be safe if someone creates it, as it is out of our support matrix.
> 
> I'd prefer not to block it, but better be safe I guess.
> 
> Can you please squash all commits? It's all part of same logical change.

done

umangachapagain (approved):

PR #2297, opened 2023-12-04
Add StorageCluster count and namespace to ClusterClaim

umangachapagain:


iamniting (changes requested):
I think we should also export the details of internal and external clusters.

iamniting (approved):

umangachapagain:
> I think we should also export the details of internal and external clusters.

We are still discussing if ClusterClaims are the right way to surface such information, or if this information is even needed.
For now, we should be ok without it. Eventually we'd want a way to surface more ODF metadata but without overloading ClusterClaims.

PR #2304, opened 2023-12-06
Move API to separate go module

umangachapagain:
Our api/ packages are currently using controller-runtime utilities to register go types to GVK scheme. These utilities are wrappers around k8s.io/apimachinery which we already have dependency on. So, with this change we are getting rid of the extra dependency to keep things maintainable.

This PR also moves ocs-operator API package into it's own go module.


iamniting (changes requested):
I think we should also add the make command similar to the `deps-update` and `verify-deps` or under these commands itself to update the go module inside API directory.

Makefile
```diff
@@ -31,6 +31,8 @@ CONTROLLER_GEN=$(LOCALBIN)/controller-gen
 deps-update:
 	@echo "Running deps-update"
 	go mod tidy && go mod vendor
+	@echo "Running deps-update on api submodule"
```
iamniting, line 34: Can you also update the paths under `hack/verify-dependencies.sh`

iamniting (changes requested):

iamniting (approved):

leelavg:
@umangachapagain / @iamniting will you be creating a followup PR to support repo for multi [modules](https://github.com/golang/tools/blob/master/gopls/doc/workspace.md#go-workspaces-go-118)?

umangachapagain:
> @umangachapagain / @iamniting will you be creating a followup PR to support repo for multi [modules](https://github.com/golang/tools/blob/master/gopls/doc/workspace.md#go-workspaces-go-118) lsp resolutions?
>
Done #2326

PR #2306, opened 2023-12-06
Add alerts for high osd cpu usage

weirdwiz:
This PR adds an alert, OSD is overwhelmed with a lot of performance request. 

There were multiple metrics/values considered to recognise when the cluster is under load. 
1. `util%` for the disks: The value for disk utilization is not very useful for devices with support for multiple queues like ssd and nvme. even if the utilization values are very high, the storage system can "take more work on"

2.  disk latency: we can take a look at an average latency over time, but latency is highly dependent on IO_size

3. CPU usage: on cluster's under high load, the CPU usage of the primary OSD jumps up on a small cluster, increasing the amount of OSDs reduce the cpu usage. 

in the end the PR just adds CPU usage as a factor, with a 35% usage as a threshold, which let's the user know to add more OSDs/increase the cluster size to aleviate the issue

### DIFF
--- a/metrics/deploy/prometheus-ocs-rules.yaml
+++ b/metrics/deploy/prometheus-ocs-rules.yaml
@@ -265,7 +265,7 @@ spec:
         floor((ocs_storage_provider_operator_version>0)/1000) - ignoring(storage_consumer_name) group_right() floor((ocs_storage_client_operator_version>0)/1000) > 1 or floor((ocs_storage_client_operator_version>0)/1000) - ignoring(storage_consumer_name) group_left() floor((ocs_storage_provider_operator_version>0)/1000) >= 1
       labels:
         severity: critical
-  - name: mds-performance-alerts.rules
+  - name: ceph-daemon-performance-alerts.rules
     rules:
     - alert: MDSCacheUsageHigh
       annotations:
@@ -277,3 +277,12 @@ spec:
       for: 5m
       labels:
         severity: critical
+    - alert: OSDCPULoadHigh
+      annotations:
+        description: High CPU usage in the OSD container on pod {{ $labels.pod }}. Please create more OSDs to increase performance
+        message: High CPU usage detected in OSD container on pod {{ $labels.pod}}.
+        severity_level: warning
+      expr: "pod:container_cpu_usage:sum{pod=~\"rook-ceph-osd-.*\"} > 0.35 \n"
+      for: 15m
+      labels:
+        severity: warning
--- a/metrics/mixin/alerts/perf.libsonnet
+++ b/metrics/mixin/alerts/perf.libsonnet
@@ -2,7 +2,7 @@
   prometheusAlerts+:: {
     groups+: [
       {
-        name: 'mds-performance-alerts.rules',
+        name: 'ceph-daemon-performance-alerts.rules',
         rules: [
           {
             alert: 'MDSCacheUsageHigh',
@@ -17,6 +17,21 @@
               message: 'High MDS cache usage for the daemon {{ $labels.ceph_daemon }}.',
               description: 'MDS cache usage for the daemon {{ $labels.ceph_daemon }} has exceeded above 95% of the requested value. Increase the memory request for {{ $labels.ceph_daemon }} pod.',
               severity_level: 'error',
+            }
+          },
+          {
+            alert: 'OSDCPULoadHigh',
+            expr: |||
+              pod:container_cpu_usage:sum{%(osdSelector)s} > 0.35 
+            ||| % $._config,
+            'for': $._config.osdCPULoadHighAlertTime,
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              message: 'High CPU usage detected in OSD container on pod {{ $labels.pod}}.',
+              description: 'High CPU usage in the OSD container on pod {{ $labels.pod }}. Please create more OSDs to increase performance',
+              severity_level: 'warning',
             },
           },
         ],
--- a/metrics/mixin/config.libsonnet
+++ b/metrics/mixin/config.libsonnet
@@ -2,6 +2,7 @@
   _config+:: {
     // Selectors are inserted between {} in Prometheus queries.
     ocsExporterSelector: 'job="ocs-metrics-exporter"',
+    osdSelector: 'pod=~"rook-ceph-osd-.*"',
 
     // Duration to raise various Alerts
     clusterObjectStoreStateAlertTime: '15s',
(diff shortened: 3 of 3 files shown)


metrics/mixin/alerts/perf.libsonnet
```diff
@@ -0,0 +1,26 @@
...
+            alert: 'OSDHighCPULoad',
+            expr: |||
+              pod:container_cpu_usage:sum{%(osdSelector)s} > 0.35 
+            ||| % $._config,
+            'for': '15m',
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              message: 'High CPU usage detected in OSD container on node {{ $labels.pod}}.',
```
sp98, line 17:
```suggestion
              message: 'High CPU usage detected in OSD container on pod {{ $labels.pod}}.',
```

metrics/mixin/alerts/perf.libsonnet
```diff
@@ -0,0 +1,26 @@
...
+              description: 'High CPU usage in the OSD container on node {{ $labels.pod }}. Please create more OSDs to increase performance',
```
sp98, line 18:
```suggestion
              description: 'High CPU usage in the OSD container on pod {{ $labels.pod }}. Please create more OSDs to increase performance',
```

sp98 (changes requested):

weirdwiz:
PTAL @travisn @umangachapagain @aruniiird

metrics/deploy/prometheus-ocs-rules.yaml
```diff
@@ -265,3 +265,14 @@ spec:
...
       labels:
         severity: critical
+  - name: ceph-perf-warnings.rules
+    rules:
+    - alert: OSDHighCPULoad
+      annotations:
+        description: High CPU usage in the OSD container on node {{ $labels.pod }}. Please create more OSDs to increase performance
+        message: High CPU usage detected in OSD container on node {{ $labels.pod}}.
+        severity_level: warning
+      expr: "pod:container_cpu_usage:sum{pod=~\"rook-ceph-osd-.*\"} > 0.35 \n"
```
sp98, line 275: why 0.35?

weirdwiz: i've set it to 35%, because in all of my tests
the OSD cpu usage, hovered around 20%ish. in very very overloaded environments, the osds steadly increased cpu usage till 30-40%

the throughput and latency was very low during the overloaded environment, only increased when i deployed a larger cluster (also resulting in lower CPU usage)

metrics/mixin/alerts/alerts.libsonnet
```diff
@@ -4,3 +4,5 @@
 (import 'blocklist.libsonnet') +
 (import 'encryption.libsonnet') +
 (import 'storage-client.libsonnet')
+(import 'encryption.libsonnet') + 
+(import 'perf.libsonnet')
```
umangachapagain, line 8:
```suggestion
(import 'encryption.libsonnet') +
(import 'storage-client.libsonnet') +
(import 'perf.libsonnet')
```

metrics/mixin/alerts/perf.libsonnet
```diff
@@ -0,0 +1,26 @@
+{
+  prometheusAlerts+:: {
+    groups+: [
+      {
+        name: 'ceph-perf-warnings.rules',
```
umangachapagain, line 5: There's already a new group for perf alerts added by #2309 . Please re-use that.

metrics/mixin/alerts/perf.libsonnet
```diff
@@ -0,0 +1,26 @@
...
+        rules: [
+          {
+            alert: 'OSDHighCPULoad',
```
umangachapagain, line 8:
```suggestion
            alert: 'OSDCPULoadHigh',
```

metrics/mixin/alerts/perf.libsonnet
```diff
@@ -0,0 +1,26 @@
...
+            expr: |||
+              pod:container_cpu_usage:sum{%(osdSelector)s} > 0.35 
+            ||| % $._config,
+            'for': '15m',
```
umangachapagain, line 12: Put this in the config.libsonnet file.

umangachapagain (changes requested):

metrics/mixin/alerts/perf.libsonnet
weirdwiz: it has a group `mds-performance-alerts.rules`, should we rename that
or introduce a new group for osd?

umangachapagain: Make it `ceph-daemon-performance-alerts.rules` maybe.

metrics/deploy/prometheus-ocs-rules.yaml
```diff
@@ -265,3 +265,14 @@ spec:
         floor((ocs_storage_provider_operator_version>0)/1000) - ignoring(storage_consumer_name) group_right() floor((ocs_storage_client_operator_version>0)/1000) > 1 or floor((ocs_storage_client_operator_version>0)/1000) - ignoring(storage_consumer_name) group_left() floor((ocs_storage_provider_operator_version>0)/1000) >= 1
       labels:
         severity: critical
+  - name: ceph-perf-warnings.rules
+    rules:
+    - alert: OSDHighCPULoad
+      annotations:
+        description: High CPU usage in the OSD container on node {{ $labels.pod }}. Please create more OSDs to increase performance
```
jmolmo, line 272: :s/node/pod/

metrics/deploy/prometheus-ocs-rules.yaml
```diff
@@ -265,3 +265,14 @@ spec:
...
+        message: High CPU usage detected in OSD container on node {{ $labels.pod}}.
```
jmolmo, line 273: :s/node/pod/

jmolmo (changes requested):
After addressing suggested changes LGTM.

metrics/deploy/prometheus-ocs-rules.yaml
jmolmo: or regenerate file...

metrics/deploy/prometheus-ocs-rules.yaml
```diff
@@ -265,3 +265,14 @@ spec:
...
+        severity_level: warning
```
jmolmo, line 274: please remember to add the new "runbook_url" annotation with the link to the alert doc file.

jmolmo (commented):
Remember you need also to add the related runbook. (...here the annotation label, in runbooks repo, the doc file.)

weirdwiz:
updated the PR with the changes

metrics/deploy/prometheus-ocs-rules.yaml
umangachapagain: This can be a follow-up PR, since this is a new requirement.

umangachapagain (approved):

PR #2307, opened 2023-12-06
Alert user if ceph metadata server is consuming cpu at threshold point.

manishym:
Raise an alert to scale active MDS pods if the CPU usage of MDS pod is more than 67% for 6 hours.

### DIFF
--- a/metrics/mixin/alerts/perf.libsonnet
+++ b/metrics/mixin/alerts/perf.libsonnet
@@ -34,6 +34,22 @@
               severity_level: 'warning',
             },
           },
+
+          {
+            alert: 'MDSCPUUsageHigh',
+            expr: |||
+              pod:container_cpu_usage:sum{%(mdsSelector)s}/ on(pod) kube_pod_resource_limit{resource='cpu',%(mdsSelector)s} > 0.67
+            ||| % $._config,
+            'for': $._config.mds_cpu_usage_high_threshold_duration,
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              message: 'Ceph metadata server pod ({{ $labels.pod }}) has high cpu usage',
+              description: 'Ceph metadata server pod ({{ $labels.pod }}) has high cpu usage.\nPlease consider increasing the number of active metadata servers,\nit can be done by increasing the number of activeMetadataServers parameter in the StorageCluster CR.',
+              severity_level: 'warning',
+            },
+          },
         ],
       },
     ],
--- a/metrics/mixin/config.libsonnet
+++ b/metrics/mixin/config.libsonnet
@@ -16,6 +16,8 @@
     ocsStorageClusterKMSConnectionAlert: '5s',
     mdsCacheUsageAlertTime: '5m',
     osdCPULoadHighAlertTime: '15m',
+    mdsSelector: 'pod=~"rook-ceph-mds.*"',
+    mds_cpu_usage_high_threshold_duration: '6h',
 
     // Constants
     objectStorageType: 'RGW',
--- a/metrics/deploy/prometheus-ocs-rules.yaml
+++ b/metrics/deploy/prometheus-ocs-rules.yaml
@@ -286,3 +286,16 @@ spec:
       for: 15m
       labels:
         severity: warning
+    - alert: MDSCPUUsageHigh
+      annotations:
+        description: |-
+          Ceph metadata server pod ({{ $labels.pod }}) has high cpu usage.
+          Please consider increasing the number of active metadata servers,
+          it can be done by increasing the number of activeMetadataServers parameter in the StorageCluster CR.
+        message: Ceph metadata server pod ({{ $labels.pod }}) has high cpu usage
+        severity_level: warning
+      expr: |
+        pod:container_cpu_usage:sum{pod=~"rook-ceph-mds.*"}/ on(pod) kube_pod_resource_limit{resource='cpu',pod=~"rook-ceph-mds.*"} > 0.67
+      for: 6h
+      labels:
+        severity: warning


metrics/mixin/alerts/alerts.libsonnet
```diff
@@ -3,4 +3,5 @@
 (import 'services.libsonnet') +
 (import 'blocklist.libsonnet') +
 (import 'encryption.libsonnet') +
-(import 'storage-client.libsonnet')
+(import 'storage-client.libsonnet') +
+(import 'ceph-overload.libsonnet')
```
umangachapagain, line 7: This file does not exist.
Please rebase the commits on top of changes from #2309.

metrics/mixin/alerts/perf.libsonnet
```diff
@@ -0,0 +1,26 @@
+{
+  prometheusAlerts+:: {
+    groups+: [
+      {
+        name: 'ODF-ceph-mds-high-cpu-warnings.rules',
```
umangachapagain, line 5: A new alert group is already defined by #2309 , please put this rule under that group.

metrics/mixin/alerts/perf.libsonnet
```diff
@@ -0,0 +1,26 @@
...
+        rules: [
+          {
+            alert: 'MDS-high-cpu',
```
umangachapagain, line 8:
```suggestion
            alert: 'MDSCPUUsageHigh',
```

metrics/mixin/config.libsonnet
```diff
@@ -13,6 +13,8 @@
     odfPoolMirroringImageHealthCriticalAlertTime: '10s',
     blockedRBDClientAlertTime: '10s',
     ocsStorageClusterKMSConnectionAlert: '5s',
+    mdsSelector: 'pod=~"rook-ceph-mds.*"',
+    mds_cpu_usage_high_threshold_duration: '6h',
```
umangachapagain, line 17: This means alert will be raised only when CPU usage is high for 6 hours. Is that the desired behavior?

metrics/mixin/alerts/perf.libsonnet
```diff
@@ -0,0 +1,26 @@
...
+            expr: |||
+              pod:container_cpu_usage:sum{%(mdsSelector)s}/ on(pod) kube_pod_resource_limit{resource='cpu',%(mdsSelector)s} > 0.67
```
umangachapagain, line 10: `Question:` Why 67%? Is it a randomly selected or is there a reason for it?

metrics/mixin/alerts/perf.libsonnet
```diff
@@ -0,0 +1,26 @@
+{
+  prometheusAlerts+:: {
```
umangachapagain, line 2: Run `make gen-latest-prometheus-rules-yamls` from project root to update the generated rule files.

umangachapagain (changes requested):

metrics/mixin/config.libsonnet
jmolmo: yep.... seems a little bit weird to leave a pod running with more than 67% cpu usage for 6 hours.... If this happens probably the user is going to investigate by himself  why the filesystems affected are so slow before this alert will be raised.
@manishym what is your reasoning behind :
` "raise the alert when mds pod usage >= 67% for more than 6 hours"`

metrics/mixin/alerts/perf.libsonnet
```diff
@@ -0,0 +1,26 @@
...
+              pod:container_cpu_usage:sum{%(mdsSelector)s}/ on(pod) kube_pod_resource_limit{resource='cpu',%(mdsSelector)s} > 0.67
+            ||| % $._config,
+            'for': $._config.mds_cpu_usage_high_threshold_duration,
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              message: 'Ceph metadata server pod ({{ $labels.pod }}) has high cpu usage',
+              description: 'Ceph metadata server pod ({{ $labels.pod }}) has high cpu usage.\nPlease consider increasing the number of active metadata servers,\nit can be done by increasing the number of activeMetadataServers parameter in the StorageCluster CR.',
+              severity_level: 'warning',
```
jmolmo, line 19: It is needed to add  the `runbook__url `annotation with the link to the alert documentation file in runbooks repo.

jmolmo (changes requested):
Please address the suggestions, and do not forget to add the extended documentation file for this new alert in the runbook repo and the related annotation label with the link.
See[ example](https://github.com/openshift/runbooks/blob/master/alerts/openshift-container-storage-operator/CephMonVersionMismatch.md)

metrics/mixin/alerts/perf.libsonnet
manishym: Ack

metrics/mixin/config.libsonnet
manishym: Yes, by default the MDS pods are allocated 3 CPU. 67% means 2 CPUs are busy. I have given 6 hours because if there is some intermittent usage of CPU, like some scan job that reads every file or something, then they don't need multiple active MDS, if it is persistently high, then we should alert. That is the intent.

metrics/mixin/alerts/alerts.libsonnet
manishym: ack

metrics/mixin/alerts/perf.libsonnet
manishym: Yes, by default the MDS pods are allocated 3 CPU. 67% means 2 CPUs are busy. I have given 6 hours because if there is some intermittent usage of CPU, like some scan job that reads every file or something, then they don't need multiple active MDS, if it is persistently high, then we should alert. That is the intent.

metrics/mixin/config.libsonnet
umangachapagain: `Sugegstion`: If that is the case we want to handle, we should probably look at something that says "if CPU usage is high most of the time then alert" instead of "if CPU usage is high for more than x hours then alert".

I still feel 6h is a long time and we should alert early. If as an admin I get this alert and I know I am running a scan job, then I can ignore it. But, we can improve it based on feedback.

metrics/mixin/alerts/perf.libsonnet
```diff
@@ -1,26 +1,43 @@
...
-            labels: {
-              severity: 'critical',
+    groups+:
+      [
+        {
+          name: 'mds-performance-alerts.rules',
+          rules: [
+            {
+              alert: 'MDSCacheUsageHigh',
+              expr: |||
```
umangachapagain, line 10: Is this formatting change caused by a linter? Doesn't look like it's needed here.

manishym:
> Please address the suggestions, and do not forget to add the extended documentation file for this new alert in the runbook repo and the related annotation label with the link. See[ example](https://github.com/openshift/runbooks/blob/master/alerts/openshift-container-storage-operator/CephMonVersionMismatch.md)

I will address runbook changes after Feature Freeze.

metrics/mixin/alerts/perf.libsonnet
umangachapagain: @manishym This comment still needs to be addressed.

metrics/mixin/alerts/perf.libsonnet
umangachapagain: @manishym This is still not fixed.

metrics/mixin/alerts/perf.libsonnet
manishym: Happened while addressing merge conflicts. Fixed it. Thanks.

metrics/mixin/alerts/perf.libsonnet
manishym: @umangachapagain It requires another PR in another repo, I will address it after the feature freeze.

metrics/mixin/alerts/perf.libsonnet
manishym: ack

umangachapagain (approved):

PR #2309, opened 2023-12-07
add alert for mds cache over use

sp98:
**Background:** 

ODF sets requests/limits of the MDS daemon pod to 8Gi. Rook sets the mds_cache_memory_limit is set to 0.5,  that is,  4Gi
The MDS will try to stay under a reservation of this limit by trimming unused metadata in its cache and recalling cached items in the client caches.
It is possible for the MDS to exceed this limit due to slow recall from clients. 
Ceph sends a health warning message when the cache full threshold (mds_health_cache_threshold) of 150%, that is 6Gi in this case, is breached.

Here, we trigger the alert to the user when MDS cache usage breaches 95% of the `mds_cache_memory_limit`, that is, (mds resource request * 0.5)

### DIFF
--- a/metrics/deploy/prometheus-ocs-rules.yaml
+++ b/metrics/deploy/prometheus-ocs-rules.yaml
@@ -265,3 +265,15 @@ spec:
         floor((ocs_storage_provider_operator_version>0)/1000) - ignoring(storage_consumer_name) group_right() floor((ocs_storage_client_operator_version>0)/1000) > 1 or floor((ocs_storage_client_operator_version>0)/1000) - ignoring(storage_consumer_name) group_left() floor((ocs_storage_provider_operator_version>0)/1000) >= 1
       labels:
         severity: critical
+  - name: mds-performance-alerts.rules
+    rules:
+    - alert: MDSCacheUsageHigh
+      annotations:
+        description: MDS cache usage for the daemon {{ $labels.ceph_daemon }} has exceeded above 95% of the requested value. Increase the memory request for {{ $labels.ceph_daemon }} pod.
+        message: High MDS cache usage for the daemon {{ $labels.ceph_daemon }}.
+        severity_level: error
+      expr: |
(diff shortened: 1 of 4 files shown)


sp98:
Tested by reduced the expected value in order to trigger the alert.

Screenshot from 2023-12-07 21-53-30

sp98:
@aruniiird Please review this PR.

metrics/mixin/alerts/mds.libsonnet
```diff
@@ -0,0 +1,26 @@
+{
+  prometheusAlerts+:: {
+    groups+: [
+      {
+        name: 'mds-memory-usage-alerts.rules',
+        rules: [
+          {
+            alert: 'MDSCacheUsage',
+            expr: |||
+              ceph_mds_mem_rss / on(ceph_daemon) group_left(job)(label_replace(kube_pod_container_resource_requests{container="mds", resource="memory"}, "ceph_daemon", "mds.$1", "pod", "rook-ceph-mds-(.*)-(.*)")*.5) > 1
```
aruniiird, line 10: @sp98 , is it possible to give an out put of

`ceph_mds_mem_rss / on(ceph_daemon) group_left(job)(label_replace(kube_pod_container_resource_requests{container="mds", resource="memory"}, "ceph_daemon", "mds.$1", "pod", "rook-ceph-mds-(.*)-(.*)")*.5)`

without any condition, just to verify we have the output, of the expression, as expected

aruniiird, line 10: A minute change for the readability sake
```suggestion
              ceph_mds_mem_rss / on(ceph_daemon) group_left(job)(label_replace(kube_pod_container_resource_requests{container="mds", resource="memory"}, "ceph_daemon", "mds.$1", "pod", "rook-ceph-mds-(.*)-(.*)") * .5) > 1
```

sp98: I don't have a live cluster currently. Give me sometime. I'll past the output here.

aruniiird (approved):
Looks good. LGTM from my side

umangachapagain (approved):

PR #2308, opened 2023-12-07
Onboard clients only if they are on same version as provider - 2

leelavg:
- Operator major and minor version of client and provider should match for onboarding process to begin
- After this commit older clients will fail to get onboarded

PR #2311, opened 2023-12-07
Stop provider upgrade if clients version is lagging

leelavg:
Provider supports client operator at X.Y and X.(Y-1) versions and if any client is already at X.(Y-1) version and provider gets upgraded then the lagging client goes out of support

This PR prohibits such provider upgrades

Minor: use a single existing enqueue function and removes duplicates

[RHSTOR-5073]

### DIFF
--- a/controllers/storagecluster/reconcile.go
+++ b/controllers/storagecluster/reconcile.go
@@ -12,6 +12,7 @@ import (
 	conditionsv1 "github.com/openshift/custom-resource-status/conditions/v1"
 	"github.com/operator-framework/operator-lib/conditions"
 	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
+	ocsv1alpha1 "github.com/red-hat-storage/ocs-operator/api/v4/v1alpha1"
 	"github.com/red-hat-storage/ocs-operator/v4/controllers/util"
 	statusutil "github.com/red-hat-storage/ocs-operator/v4/controllers/util"
 	"github.com/red-hat-storage/ocs-operator/v4/version"
@@ -505,12 +506,34 @@ func (r *StorageClusterReconciler) reconcilePhases(
 		if instance.Status.Phase != statusutil.PhaseClusterExpanding {
 			instance.Status.Phase = statusutil.PhaseReady
 
+			var returnErr error
+			var notUpgradeableReasons, notUpgradeableMessages []string
 			// mark operator upgradeable if and only if all storageclusters are ready
-			if r.clusters.AreOtherStorageClustersReady(instance) {
-				returnErr := r.SetOperatorConditions(message, reason, metav1.ConditionTrue, nil)
-				if returnErr != nil {
-					return reconcile.Result{}, returnErr
+			if !r.clusters.AreOtherStorageClustersReady(instance) {
+				notUpgradeableReasons = append(notUpgradeableReasons, "NotReady")
+				notUpgradeableMessages = append(notUpgradeableMessages, "StorageCluster is not ready")
+			}
+			// check operator upgradeability based on connected clients
+			if instance.Spec.AllowRemoteStorageConsumers {
+				if count, err := getUnsupportedClientsCount(r, instance.Namespace); err != nil {
+					notUpgradeableReasons = append(notUpgradeableReasons, "ODFClients")
+					notUpgradeableMessages = append(notUpgradeableMessages, "Unable to determine status of connected ODF Clients")
+				} else if count != 0 {
+					notUpgradeableReasons = append(notUpgradeableReasons, "ODFClients")
+					notUpgradeableMessages = append(notUpgradeableMessages, fmt.Sprintf("%d connected ODF Client Operators are not up to date", count))
 				}
+			}
+			if len(notUpgradeableMessages) > 0 {
+				// we are not upgradeable
+				returnErr = r.SetOperatorConditions(
+					strings.Join(notUpgradeableMessages, ";"), strings.Join(notUpgradeableReasons, ";"),
+					metav1.ConditionFalse, nil)
+			} else {
+				// we are upgradeable
+				returnErr = r.SetOperatorConditions(message, reason, metav1.ConditionTrue, nil)
+			}
+			if returnErr != nil {
+				return reconcile.Result{}, returnErr
 			}
 		}
 	} else {
@@ -837,4 +860,29 @@ func validateCustomStorageClassNames(sc *ocsv1.StorageCluster) error {
 	}
 
 	return nil
+}
+
+func getUnsupportedClientsCount(r *StorageClusterReconciler, namespace string) (int, error) {
+	scList := &ocsv1alpha1.StorageConsumerList{}
+	err := r.Client.List(r.ctx, scList, client.InNamespace(namespace))
+	if err != nil {
+		r.Log.Error(err, "Failed to list StorageConsumers")
+		return -1, err
+	}
+	var count int
+	providerVersion, _ := semver.Make(version.Version)
+	for idx := range scList.Items {
+		clientVersion, err := semver.Make(scList.Items[idx].Status.Client.OperatorVersion)
+		if err == nil {
+			// provider operator and client operator should be on same version for full compatibility
+			if providerVersion.Major != clientVersion.Major || providerVersion.Minor != clientVersion.Minor {
+				count++
+			}
+		} else {
+			r.Log.Error(err, "Failed to parse client operator version", "StorageConsumer", scList.Items[idx].GetName())
+			count++
+		}
+	}
+
+	return count, nil
 }
--- a/controllers/storagecluster/storagecluster_controller.go
+++ b/controllers/storagecluster/storagecluster_controller.go
@@ -10,6 +10,7 @@ import (
 	conditionsv1 "github.com/openshift/custom-resource-status/conditions/v1"
 	"github.com/operator-framework/operator-lib/conditions"
 	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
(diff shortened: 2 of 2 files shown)


leelavg:
/hold
1. More testing and fixing existing tests in under progress
2. Will unhold when more than one vote is provided
3. Testing done so far is as below
```

=== Provider Operator is at 4.15
# ko get csv -ojsonpath='{range .items[*]}[{.metadata.name}, {.spec.version}]{end}'
[ocs-operator.v4.15.0-79.stable, 4.15.0-79.stable]

=== All connected clients are also at 4.15
# ko get storageconsumer -ojsonpath='{range .items[*]}[{.metadata.name}, {.status.client.operatorVersion}]{"\n"}{end}'
[cons-1, 4.15.1]
[cons-2, 4.15.2]
[cons-3, 4.15.3]

=== Provider can upgrade w/o any issue
# ko get condition -ojsonpath='{range .items[*]}[{.metadata.name}, {.spec.conditions}]{end}' 
[ocs-operator.v4.15.0-79.stable, [{"lastTransitionTime":"2023-12-07T13:57:06Z","message":"Reconcile completed successfully","reason":"ReconcileCompleted","status":"True","type":"Upgradeable"}]]

==== Assume one of the client operator is lagging provider operator by 1 minor version
# ko patch --subresource=status storageconsumers cons-3 --type json -p '[{"op": "replace", "path": "/status/client/operatorVersion", "value": "4.14.3"}]'
storageconsumer.ocs.openshift.io/cons-3 patched

# ko get storageconsumer -ojsonpath='{range .items[*]}[{.metadata.name}, {.status.client.operatorVersion}]{"\n"}{end}'
[cons-1, 4.15.1]
[cons-2, 4.15.2]
[cons-3, 4.14.3]

==== The scenario is surfaced to the Operator Condition and upgrade is prohibited
# ko get condition -ojsonpath='{range .items[*]}[{.metadata.name}, {.spec.conditions}]{end}' 
[ocs-operator.v4.15.0-79.stable, [{"lastTransitionTime":"2023-12-07T13:58:10Z","message":"StorageConsumer \"cons-3\" will become unsupported if upgraded","reason":"ConsumerClientOperatorVersion","status":"False","type":"Upgradeable"}]]

# ko get storagecluster
ocs-storagecluster   30h   Ready              2023-12-06T07:20:31Z   4.15.0

==== Reset client operator version
# ko patch --subresource=status storageconsumers cons-3 --type json -p '[{"op": "replace", "path": "/status/client/operatorVersion", "value": "4.15.3"}]'
storageconsumer.ocs.openshift.io/cons-3 patched

# ko get storageconsumer -ojsonpath='{range .items[*]}[{.metadata.name}, {.status.client.operatorVersion}]{"\n"}{end}'
[cons-1, 4.15.1]
[cons-2, 4.15.2]
[cons-3, 4.15.3]

==== If the connected client operator is upto w/ provider operator upgrade will be unblocked
# ko get condition -ojsonpath='{range .items[*]}[{.metadata.name}, {.spec.conditions}]{end}' 
[ocs-operator.v4.15.0-79.stable, [{"lastTransitionTime":"2023-12-07T13:59:11Z","message":"Reconcile completed successfully","reason":"ReconcileCompleted","status":"True","type":"Upgradeable"}]]
```

leelavg:
depends on #2308 partially

controllers/storagecluster/reconcile.go
```diff
@@ -496,6 +497,38 @@ func (r *StorageClusterReconciler) reconcilePhases(
...
+					if err != nil {
+						r.Log.Error(err, "Failed to parse OCS Version")
+						return reconcile.Result{}, err
+					}
+					for idx := range scList.Items {
+						clientVersion, err := semver.Make(scList.Items[idx].Status.Client.OperatorVersion)
+						if err != nil {
+							r.Log.Error(err, "Failed to parse Client operator version")
+							return reconcile.Result{}, err
+						}
```
nb-ohad, line 517: I am not sure this is the desired way to handle issues with the client version. 
I would prefer to mark that missing information to block the upgrade but not to kill the reconciliation process.  
Of course we will need to log the issue

controllers/storagecluster/reconcile.go
```diff
@@ -496,6 +497,38 @@ func (r *StorageClusterReconciler) reconcilePhases(
...
+						// onboarded client operator can be either on same provider operator version or one less in minor version
+						// we need to stop provider operator upgrades for these lagging client operators only
+						// because if provider operator upgrades, that lagging client operator will become unsupported
+						if providerVersion.Minor-1 == clientVersion.Minor {
```
nb-ohad, line 521: To cover all cases can we just do:
 ` clientVersion.Major < providerVersion.Major || clientVersion.Minor < providerVersion.Minor` to have a stronger check and capture any edge cases

controllers/storagecluster/reconcile.go
```diff
@@ -496,6 +497,38 @@ func (r *StorageClusterReconciler) reconcilePhases(
...
+							message := fmt.Sprintf("StorageConsumer %q will become unsupported if upgraded", scList.Items[idx].GetName())
```
nb-ohad, line 522: I don't think we want a condition for each individual consumer. Also `StorageConsumer` does not have a version it is the client operator that has a version

I would suggest having a flag to capture the existence of a client that is in the wrong version, exit the loop, and add a general condition. Maybe with the message: "One or more of the connected ODF clients are not up to date"

We can even count the number of clients that prevent upgrade in the loop and then have the message: "X connected ODF clients are not up to date""

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -188,6 +189,36 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+			return false
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			if e.ObjectOld == nil {
+				return true
+			}
+			old := e.ObjectOld.(*ocsv1alpha1.StorageConsumer)
+			new := e.ObjectNew.(*ocsv1alpha1.StorageConsumer)
+			return old.Status.Client.OperatorVersion != new.Status.Client.OperatorVersion
+		},
```
nb-ohad, line 203: Why only these 2 functions? we need to listen to all the others to remove the condition

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -188,6 +189,36 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		},
 	)
 
+	storageConsumerPredicate := predicate.Funcs{
```
nb-ohad, line 192: Let's have the name describe the condition itself and not the structure that holds it  
```suggestion
	ocsClientOperatorVersionPredicate := predicate.Funcs{
```

nb-ohad (changes requested):

controllers/storagecluster/reconcile.go
```diff
@@ -496,6 +497,38 @@ func (r *StorageClusterReconciler) reconcilePhases(
 
 			// mark operator upgradeable if and only if all storageclusters are ready
 			if r.clusters.AreOtherStorageClustersReady(instance) {
+				if instance.Spec.AllowRemoteStorageConsumers {
```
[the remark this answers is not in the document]
leelavg: at this point ocs sets the condition to be upgradeable and we want to check the consumers before setting that

controllers/storagecluster/reconcile.go
```diff
@@ -496,6 +497,38 @@ func (r *StorageClusterReconciler) reconcilePhases(
...
+					scList := &ocsv1alpha1.StorageConsumerList{}
+					err := r.Client.List(r.ctx, scList, &client.ListOptions{Namespace: instance.Namespace})
```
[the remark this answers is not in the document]
leelavg: > 1. Can't we the r.List util which already takes into account context?
- I couldn't find any helper here, it was there in deployer repo

controllers/storagecluster/reconcile.go
leelavg: - if we are depending on onboarding check for same versioned client operators we can forgo Major version check
- for Minor version, will change to `lt` from current `eq` match.
---
- if we are not depending on onboarding check, neither existing check nor the suggested one will work, like a client operator might already be ahead of provider operator

controllers/storagecluster/storagecluster_controller.go
leelavg: - the default value of event functions is `true` so we'll indeed get event for other conditions
- based on earlier comments I need to handle remove condition as well now, will do that

controllers/storagecluster/reconcile.go
leelavg: done.

controllers/storagecluster/reconcile.go
leelavg: done.

controllers/storagecluster/reconcile.go
leelavg: done.

controllers/storagecluster/storagecluster_controller.go
leelavg: done.

controllers/storagecluster/reconcile.go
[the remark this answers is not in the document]
leelavg: @nb-ohad I rechecked the ask, we should be updating True only if all conditions are satisfied and so this needs to be nested.

leelavg:
rebased

controllers/storagecluster/reconcile.go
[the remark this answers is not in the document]
leelavg: done.

leelavg:
address comments and retested.

nb-ohad (approved):

PR #2315, opened 2023-12-11
Update direct dependencies like rook,noobaa,openshift,operator framework & others

malayparida2000:


umangachapagain (approved):
/hold
This should be merged later unless this is blocking a feature.

umangachapagain (changes requested):
Squash the first 2 commits, since 1st commit won't work without the 2nd.

umangachapagain (approved):

PR #2317, opened 2023-12-12
set pinning field in CephSubvolumeGroup CR

Nikhil-Ladha:
Updated the CephSubvolumeGroup CR to include the pinning field with a default value of `Distributed: 1`

Ref: https://issues.redhat.com/browse/RHSTOR-5125


controllers/storagecluster/cephfilesystem.go
```diff
@@ -164,30 +163,48 @@ func (obj *ocsCephFilesystems) ensureCreated(r *StorageClusterReconciler, instan
...
-	cephFilesystemSubVolumeGroup := &cephv1.CephFilesystemSubVolumeGroup{ObjectMeta: objectMeta}
-	mutateFn := func() error {
-		cephFilesystemSubVolumeGroup.Spec = cephv1.CephFilesystemSubVolumeGroupSpec{
-			FilesystemName: filesystemName,
+		r.Log.Info("Restoring original CephSubvolumeGroup.", "CephSubvolumeGroup", klog.KRef(cephsvg.Namespace, cephsvg.Name))
+		existingsvg.ObjectMeta.OwnerReferences = cephsvg.ObjectMeta.OwnerReferences
+		existingsvg.Spec = cephsvg.Spec
+		err = r.Client.Update(context.TODO(), existingsvg)
+		if err != nil {
+			r.Log.Error(err, "Unable to update CephSubvolumeGroup.", "CephFileSystem", "CephSubvolumeGroup", klog.KRef(cephsvg.Namespace, cephsvg.Name))
```
iamniting, line 196: I think this is not correct. It should have a key and value. So there should be even arguments after the error and info.

iamniting (changes requested):
Is it tested?

controllers/storagecluster/cephfilesystem.go
Nikhil-Ladha: Oh yes, right.
Copy paste error 😓

umangachapagain (changes requested):
- Break the commits into smaller logical parts.
  - go dependency changes should be first commit.
  - StorageCluster API and CRD changes should be second commit.
  - cephfilesystem changes should be third commit.
- Add clear description of changes to each commit.
- StorageCluster CRD should not be edited manually. Change the api/*_type.go file to make changes.

go.mod
```diff
@@ -30,11 +30,11 @@ require (
 	github.com/prometheus/client_golang v1.17.0
 	github.com/prometheus/client_model v0.5.0
 	github.com/rook/rook v1.12.0-alpha.0
-	github.com/rook/rook/pkg/apis v0.0.0-20231201141116-eacc7e74412e
+	github.com/rook/rook/pkg/apis v0.0.0-20231211195439-c80ea7b64424
```
parth-gr, line 33: can we just update rook dependency for now, as it is making it complicated

controllers/storagecluster/generate.go
```diff
@@ -171,3 +171,8 @@ func generateCephReplicatedSpec(initData *ocsv1.StorageCluster, poolType string)
 func generateStorageQuotaName(storageClassName, quotaName string) string {
 	return fmt.Sprintf("%s-%s", storageClassName, quotaName)
 }
+
+// generateNameForCephSubvolumeGroup function generates a name for CephFilesystemSubVolumeGroup
+func generateNameForCephSubvolumeGroup(filesystemName string) string {
```
parth-gr, line 176: generateNameForCephSubvolumeGroup -> generateNameForDefaultCephSubvolumeGroup

controllers/storagecluster/generate.go
```diff
@@ -171,3 +171,8 @@ func generateCephReplicatedSpec(initData *ocsv1.StorageCluster, poolType string)
...
+	return fmt.Sprintf("%s-subvolumegroup", filesystemName)
```
parth-gr, line 177: lets keep return fmt.Sprintf("%s-csi", filesystemName) for the default one.
Or do  `return fmt.Sprintf("%s-%s", filesystemName, suvgname)`

controllers/storagecluster/cephfilesystem.go
```diff
@@ -164,30 +163,48 @@ func (obj *ocsCephFilesystems) ensureCreated(r *StorageClusterReconciler, instan
 }
 
 func (r *StorageClusterReconciler) createDefaultSubvolumeGroup(filesystemName, filesystemNamespace string, ownerReferences []metav1.OwnerReference) error {
-	// TODO: After fix of rook issue https://github.com/rook/rook/issues/13220 add svg name spec
 
 	existingsvg := &cephv1.CephFilesystemSubVolumeGroup{}
-	err := r.Client.Get(r.ctx, types.NamespacedName{Name: defaultSubvolumeGroupName, Namespace: filesystemNamespace}, existingsvg)
+	defaultPinningValue := 1
```
parth-gr, line 168: maybe set a global variable

controllers/storagecluster/cephfilesystem.go
```diff
@@ -164,30 +163,48 @@ func (obj *ocsCephFilesystems) ensureCreated(r *StorageClusterReconciler, instan
...
+	} else if errors.IsNotFound(err) {
+		r.Log.Info("Creating CephSubvolumeGroup.", "CephSubvolumeGroup", klog.KRef(cephsvg.Namespace, cephsvg.Name))
+		err = r.Client.Create(context.TODO(), cephsvg)
+		if err != nil {
+			r.Log.Error(err, "Unable to create CephSubvolumeGroup.", "CephSubvolumeGroup", klog.KRef(cephsvg.Namespace, cephsvg.Name))
+			return err
 		}
-		return nil
-	}
-	_, err = ctrl.CreateOrUpdate(r.ctx, r.Client, cephFilesystemSubVolumeGroup, mutateFn)
```
parth-gr, line 186: we should use `create or update` what if user want to update the pinning configuration

controllers/storagecluster/cephfilesystem.go
```diff
@@ -164,30 +163,48 @@ func (obj *ocsCephFilesystems) ensureCreated(r *StorageClusterReconciler, instan
...
 		}
-	}
 
-	objectMeta := metav1.ObjectMeta{Name: defaultSubvolumeGroupName, Namespace: filesystemNamespace, OwnerReferences: ownerReferences}
-	cephFilesystemSubVolumeGroup := &cephv1.CephFilesystemSubVolumeGroup{ObjectMeta: objectMeta}
-	mutateFn := func() error {
-		cephFilesystemSubVolumeGroup.Spec = cephv1.CephFilesystemSubVolumeGroupSpec{
-			FilesystemName: filesystemName,
+		r.Log.Info("Restoring original CephSubvolumeGroup.", "CephSubvolumeGroup", klog.KRef(cephsvg.Namespace, cephsvg.Name))
+		existingsvg.ObjectMeta.OwnerReferences = cephsvg.ObjectMeta.OwnerReferences
```
parth-gr, line 192: can we directly call createandupdate from controller runtime as it was before?

parth-gr (changes requested):

go.mod
Nikhil-Ladha: The other updates are part of the rook api updates itself, I have not updated them manually.

controllers/storagecluster/generate.go
Nikhil-Ladha: The function name should be generic, I don't think we should add default in the function name itself

controllers/storagecluster/cephfilesystem.go
Nikhil-Ladha: But, we are using this value only in this function. Why a global variable for that?

controllers/storagecluster/cephfilesystem.go
```diff
@@ -164,22 +162,32 @@ func (obj *ocsCephFilesystems) ensureCreated(r *StorageClusterReconciler, instan
...
-	objectMeta := metav1.ObjectMeta{Name: defaultSubvolumeGroupName, Namespace: filesystemNamespace, OwnerReferences: ownerReferences}
-	cephFilesystemSubVolumeGroup := &cephv1.CephFilesystemSubVolumeGroup{ObjectMeta: objectMeta}
+	cephFilesystemSubVolumeGroup := &cephv1.CephFilesystemSubVolumeGroup{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:            svgName,
+			Namespace:       filesystemNamespace,
+			OwnerReferences: ownerReferences,
+		},
+	}
+	defaultPinningValue := 1
```
parth-gr, line 183: please use the const values if your hardcoding them same for the csi svg

Nikhil-Ladha: We can do this for pinning value because the CR expects a pointer, and we can't pass an address of a const variable.
Will update for csi.

controllers/storagecluster/cephfilesystem.go
```diff
@@ -164,22 +164,32 @@ func (obj *ocsCephFilesystems) ensureCreated(r *StorageClusterReconciler, instan
...
-	// TODO: After fix of rook issue https://github.com/rook/rook/issues/13220 add svg name spec
 
 	existingsvg := &cephv1.CephFilesystemSubVolumeGroup{}
-	err := r.Client.Get(r.ctx, types.NamespacedName{Name: defaultSubvolumeGroupName, Namespace: filesystemNamespace}, existingsvg)
+	svgName := generateNameForCephSubvolumeGroup(filesystemName)
+	err := r.Client.Get(r.ctx, types.NamespacedName{Name: svgName, Namespace: filesystemNamespace}, existingsvg)
 	if err == nil {
 		if existingsvg.DeletionTimestamp != nil {
-			r.Log.Info("Unable to restore subvolumegroup because it is marked for deletion.", "subvolumegroup", klog.KRef(filesystemNamespace, defaultSubvolumeGroupName))
+			r.Log.Info("Unable to restore subvolumegroup because it is marked for deletion.", "subvolumegroup", klog.KRef(filesystemNamespace, existingsvg.Name))
```
parth-gr, line 173: >existingsvg.Name 
this would be null, use the svgName

controllers/storagecluster/cephfilesystem.go
```diff
@@ -164,22 +164,32 @@ func (obj *ocsCephFilesystems) ensureCreated(r *StorageClusterReconciler, instan
...
-	objectMeta := metav1.ObjectMeta{Name: defaultSubvolumeGroupName, Namespace: filesystemNamespace, OwnerReferences: ownerReferences}
-	cephFilesystemSubVolumeGroup := &cephv1.CephFilesystemSubVolumeGroup{ObjectMeta: objectMeta}
+	cephFilesystemSubVolumeGroup := &cephv1.CephFilesystemSubVolumeGroup{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:            svgName,
+			Namespace:       filesystemNamespace,
+			OwnerReferences: ownerReferences,
+		},
+	}
+	defaultPinningValue := 1
```
parth-gr, line 185: maybe a comment to add that this is the deafult value, maybe a top global var variable if you can create as these values are important

controllers/storagecluster/cephfilesystem.go
```diff
@@ -164,22 +164,32 @@ func (obj *ocsCephFilesystems) ensureCreated(r *StorageClusterReconciler, instan
 }
 
 func (r *StorageClusterReconciler) createDefaultSubvolumeGroup(filesystemName, filesystemNamespace string, ownerReferences []metav1.OwnerReference) error {
-	// TODO: After fix of rook issue https://github.com/rook/rook/issues/13220 add svg name spec
 
 	existingsvg := &cephv1.CephFilesystemSubVolumeGroup{}
-	err := r.Client.Get(r.ctx, types.NamespacedName{Name: defaultSubvolumeGroupName, Namespace: filesystemNamespace}, existingsvg)
+	svgName := generateNameForCephSubvolumeGroup(filesystemName)
+	err := r.Client.Get(r.ctx, types.NamespacedName{Name: svgName, Namespace: filesystemNamespace}, existingsvg)
```
parth-gr, line 170: have you tested the changes locally do you see the reconcile of subvolumegroup in rook operator

parth-gr (changes requested):

controllers/storagecluster/cephfilesystem.go
Nikhil-Ladha: If the err is nil, this can't be null.

controllers/storagecluster/cephfilesystem.go
Nikhil-Ladha: Yes

controllers/storagecluster/cephfilesystem.go
Nikhil-Ladha: Added a comment for the same

controllers/storagecluster/cephfilesystem.go
parth-gr: I see make sense

parth-gr (approved):
lgtm

iamniting (approved):

PR #2318, opened 2023-12-12
[RHSTOR-5075] remove StorageClientIncompatibleOperatorVersion alert at Critical level

leelavg:
- This alert is initially raised if client operator is outside of support matrix b/n ODF Provider and ODF Client
- With #2308 and #2311 we are enforcing the onboard and upgrade
- So, this alert can be removed now


leelavg:
/hold

depends on #2308 and #2311 to be merged

nb-ohad:
@leelavg Why are we removing this? It would be helpful if by some edge case, we got to that scenario

leelavg:
> @leelavg Why are we removing this? It would be helpful if by some edge case, we got to that scenario
- ack, I was also in cross-roads to remove this or not, I do agree to have this. I'm closing this PR and corresponding Jira.

PR #2319, opened 2023-12-13
upgraded golangci-lint to 1.54.2

Nikhil-Ladha:
Upgraded golangci-lint to 1.54.2 and resolved the errors reported.


controllers/storagecluster/generate.go
```diff
@@ -111,10 +109,6 @@ func generateNameForCephNetworkFilesystemSC(initData *ocsv1.StorageCluster) stri
 	return fmt.Sprintf("%s-ceph-nfs", initData.Name)
 }
 
-func generateNameForNFSCSIProvisioner(initData *ocsv1.StorageCluster) string {
```
iamniting, line 114: Why are we removing this func. I see this was being used.

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -71,7 +71,7 @@ type StorageConsumerReconciler struct {
 // Note:
 // The Controller will requeue the Request to be processed again if the returned error is non-nil or
 // Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
-func (r *StorageConsumerReconciler) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+func (r *StorageConsumerReconciler) Reconcile(_ context.Context, request reconcile.Request) (reconcile.Result, error) {
```
iamniting, line 74: Lets assign this ctx value to ctx in StorageConsumerReconciler `r.ctx = ctx`

iamniting (changes requested):

iamniting:
@nb-ohad pls take a look at the server code.

controllers/storagecluster/generate.go
Nikhil-Ladha: IT was only used in one place and that too not consuming the parameter passed to it. So, removed it to directly use the return value of `fmt.Sprintf`.
We are doing a similar thing for a function just above it as well.
Ref: https://github.com/red-hat-storage/ocs-operator/blob/fbfe2d9ea0256d9e38ee57ca911c74169cf55f6d/controllers/storagecluster/storageclasses.go#L377

controllers/storageconsumer/storageconsumer_controller.go
Nikhil-Ladha: Ack

umangachapagain (commented):
v1.55.2 is already available. Any reason to not move to that?
Locally other versions did not work for me with go1.21.4.

Nikhil-Ladha:
> v1.55.2 is already available. Any reason to not move to that?

No reason as such, 1.54 was the chronologically first version that supported go1.21 so updated to that.

> Locally other versions did not work for me with go1.21.4.

Interesting, in CI also we are using go1.21.4 and everything seems to work fine 🤔 
Ref: https://github.com/red-hat-storage/ocs-operator/actions/runs/7194272521/job/19594462514?pr=2319

iamniting (approved):

PR #2320, opened 2023-12-13
remove api/ package dependency on controller-runtime

umangachapagain:
Our api/ packages are currently using controller-runtime utilities to register go types to GVK scheme. These utilities are wrappers around k8s.io/apimachinery which we already have dependency on. So, with this change we are getting rid of the extra dependency to keep things maintainable.


umangachapagain:
This is superseded by #2304

PR #2325, opened 2023-12-14
Changed alerts queries to support multicluster mode

aruniiird:
Added needed changes for alerts queries to support multicluster mode

### DIFF
--- a/metrics/deploy/prometheus-ocs-rules-external.yaml
+++ b/metrics/deploy/prometheus-ocs-rules-external.yaml
@@ -73,7 +73,7 @@ spec:
         severity_level: warning
         storage_type: RGW
       expr: |
-        (ocs_objectbucketclaim_info * on (namespace, objectbucket) group_left() (ocs_objectbucket_used_bytes/ocs_objectbucket_max_bytes)) > 0.80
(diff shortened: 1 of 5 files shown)


aruniiird:
Just an FYI
Rules/Alerts under name: `ceph-daemon-performance-alerts.rules`, don't have `managedBy` label

aruniiird:
Added `managedBy` label to alert queries to support multicluster scenario.
@umangachapagain , please take a look

umangachapagain (commented):
> Just an FYI Rules/Alerts under name: `ceph-daemon-performance-alerts.rules`, don't have `managedBy` label
>
All metrics should have that label by default. @aruniiird  Do you see any issues with the query?
I see many other queries that aren't grouped by `managedBy` label. Is it not needed for those queries? Like for standardized metrics queries.

umangachapagain:
@aruniiird Looks like ceph metrics are not working correctly. We inject "managedBy" labels in them.
Let's create an issue for it an investigate.

Kube metrics will not have the labels. But, that shouldn't matter. We should be able to use our labels in the queries.
Let's create another issue and track that as well.

Merging these changes.
Thank you.

umangachapagain (approved):

PR #2326, opened 2023-12-15
Enable multi-module workspace support

umangachapagain:
Follow-up to #2304
This helps devs to work on both modules at the same time. It does not impact builds or anything else.
CI needs to be updated to recognize multi-module.


umangachapagain:
/override ci/prow/ocs-operator-ci
This test is redundant with github-actions test and is being removed by #2328 .

iamniting (approved):

PR #2328, opened 2023-12-15
remove ocs-operator-ci make target

umangachapagain:
ocs-operator-ci is no longer used to execute tests as it is redundant with github action based tests.

CI test removed by openshift/release/pull/46832


iamniting (changes requested):
Umanga, It is still useful for running the tests locally. I would suggest to keep it.

PR #2329, opened 2023-12-18
Send provider operator version in client heartbeat

leelavg:
As a response to client heartbeat we'll send the provider operator version for the client operator to match, this should help the client to decide on it's upgradeability condition.

[RHSTOR-5205]


services/provider/proto/provider.proto
```diff
@@ -161,4 +161,7 @@ message ReportStatusRequest{
     string clientOperatorVersion = 3;
 }
 
-message ReportStatusResponse{}
+message ReportStatusResponse{
+    // Desired version which provider operator wants client operator to match
+    string providerOperatorVersion = 1;
```
nb-ohad, line 166: Can we be more specific here? 
ODFVersion? OCSVersion?

nb-ohad (changes requested):

services/provider/proto/provider.proto
nb-ohad: Actually, I am not sure why the client needs to know anything about provider versions. This should be framed as a desired version not a provider component version.

So maybe `clientDesiredVersion` would work best

leelavg: done.

PR #2339, opened 2023-12-20
Api added for generating the token

rchikatw:
- UX-backed server responsible for serving HTTP requests right now it's only implementing APIs for used for providers in the future it can expanded for any other feature
- UX-backed server has 2 containers running in it 
       1. Proxy container which is responsible for authorizing the user
       2. API server which serves requests via proxy
- Proxy is running as a sidecar container to the HTTP server.


services/uibackend/main.go
```diff
@@ -0,0 +1,82 @@
...
+	// Before signing, we need to hash our message
+	// The hash is what we actually sign
+	msgHash := sha256.New()
+	_, err := msgHash.Write(msg)
+	if err != nil {
+		os.Exit(1)
+	}
+	msgHashSum := msgHash.Sum(nil)
+
+	pemString, err := os.ReadFile("/home/ritesh/ticket/key.pem")
```
rchikatw, line 54: needs to add path for private secret

services/uibackend/main.go
```diff
@@ -0,0 +1,82 @@
...
+
+	Block, _ := pem.Decode(s)
+	k, err := x509.ParsePKCS1PrivateKey(Block.Bytes)
+	if err != nil {
+		panic(err)
+	}
+	return k
+}
+
+func generateToken(w http.ResponseWriter, r *http.Request) {
```
nb-ohad, line 36: Token generation logic be decoupled from http concerns. 
My suggestion is to have a 

`handleOnboardingToken` function to be registered with the endpoint  (in its own file) which then calls an HTTP agnostic generate token function

services/uibackend/main.go
```diff
@@ -0,0 +1,82 @@
...
+	"os"
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/gorilla/mux"
+)
+
+type Onboarding struct {
+	Token string `json:"token"`
+}
```
nb-ohad, line 24: Why do we need this struct? can't we use either an anonymous struct (for the encoding) or the struct the API server use to parse the token?

services/uibackend/main.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+	msgHashSum := msgHash.Sum(nil)
+
+	pemString, err := os.ReadFile("/etc/onboarding-private-key/key")
+
+	// In order to generate the signature, we provide a random number generator,
+	// our private key, the hashing algorithm that we used, and the hash sum
+	// of our message
+
+	signature, err := rsa.SignPKCS1v15(rand.Reader, parseKey(pemString), crypto.SHA256, msgHashSum)
```
nb-ohad, line 60: Why do we need a separate parseKey call here?
An anonymous function will be more readable here as it keeps the context of token generation local to this function.

services/uibackend/main.go
```diff
@@ -0,0 +1,82 @@
...
+	if err != nil {
+		os.Exit(1)
+	}
+
+	sEnc := encodedPayload + "." + base64.StdEncoding.EncodeToString(signature)
```
nb-ohad, line 65:
```suggestion
	sEnc := fmt.Sprintf("%s.%s", encodedPayload, base64.StdEncoding.EncodeToString(signature))
```

services/uibackend/main.go
```diff
@@ -0,0 +1,82 @@
...
+
+	pemString, err := os.ReadFile("/etc/onboarding-private-key/key")
+
+	// In order to generate the signature, we provide a random number generator,
+	// our private key, the hashing algorithm that we used, and the hash sum
+	// of our message
+
+	signature, err := rsa.SignPKCS1v15(rand.Reader, parseKey(pemString), crypto.SHA256, msgHashSum)
+	if err != nil {
+		os.Exit(1)
```
nb-ohad, line 62: Shouldn't we log an error here? also why are you exiting the entire server? shouldn't we just have an HTTP reply with an Interal error code?

services/uibackend/main.go
```diff
@@ -0,0 +1,82 @@
...
+	encodedPayload := base64.StdEncoding.EncodeToString([]byte(payload))
+
+	msg := []byte(payload)
+
+	// Before signing, we need to hash our message
+	// The hash is what we actually sign
+	msgHash := sha256.New()
+	_, err := msgHash.Write(msg)
+	if err != nil {
+		os.Exit(1)
```
nb-ohad, line 50: Shouldn't we log an error here? also why are you exiting the entire server? shouldn't we just have an HTTP reply with an Interal error code?

services/uibackend/main.go
```diff
@@ -0,0 +1,82 @@
...
+type Onboarding struct {
+	Token string `json:"token"`
+}
+
+func parseKey(s []byte) *rsa.PrivateKey {
+
+	Block, _ := pem.Decode(s)
+	k, err := x509.ParsePKCS1PrivateKey(Block.Bytes)
+	if err != nil {
+		panic(err)
```
nb-ohad, line 31: Shouldn't we log an error here? also why are you panicing the entire server? shouldn't we just have an HTTP reply with an Interal error code?

services/uibackend/main.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+	return k
+}
+
+func generateToken(w http.ResponseWriter, r *http.Request) {
+
+	id := uuid.New()
+	now := time.Now()
+	payload := fmt.Sprintf("%s%s%s%d%s", "{\"id\":\"", id.String(), "\", \"expirationDate\":\"", now.Unix()+172800, "\"}")
```
nb-ohad, line 40: Please use a struct and encode it into json instead of trying to create a json using string concat

nb-ohad:
@rchikatw I was not able to locate any code that deals with setting up the server as an HTTPS server. This server needs to be secured which means you should have a certificate generated for it, with the proper annotation on the service to generate the certificate. Then the certificate needs to be mounted into the server pod as a file and used when setting up the https server.

services/uibackend/main.go
rchikatw: Done

deploy/ocs-operator/manifests/ocs-operator.clusterserviceversion.yaml
```diff
@@ -3040,6 +3051,15 @@ spec:
           verbs:
           - use
         serviceAccountName: rook-ceph-system
+      - rules:
```
leelavg, line 3054: Any reason on why these are added twice?

hack/common.sh
```diff
@@ -62,24 +62,34 @@ DEFAULT_OPERATOR_IMAGE_NAME="ocs-operator"
 DEFAULT_OPERATOR_BUNDLE_NAME="ocs-operator-bundle"
 DEFAULT_FILE_BASED_CATALOG_NAME="ocs-operator-catalog"
 DEFAULT_METRICS_EXPORTER_IMAGE_NAME="ocs-metrics-exporter"
+DEFAULT_UX_BACKEND_SERVER_IMAGE_NAME="ux-backend-server"
+DEFAULT_UX_BACKEND_OAUTH_IMAGE_NAME="ux-backend-server"
```
leelavg, line 66: Maybe this can be different?

leelavg (commented):
A lot better! Thx.

hack/common.sh
rchikatw: yes

deploy/ocs-operator/manifests/ocs-operator.clusterserviceversion.yaml
rchikatw: - This came because the role was created from code and yaml file as well. I have removed the code part which i have also added in ocsinitialization controller

rchikatw:
There is one lint error that is not my part of the code, I can fix it if any of maintainer/reviewer suggests me.

services/ux-backend/handler/onboarding.go
```diff
@@ -0,0 +1,103 @@
...
+		w.Write(response)
+
+	default:
+		fmt.Fprintf(w, "Sorry, only POST methods are supported.")
+	}
+}
+
+func getOnboardingToken() ([]byte, error) {
+
+	expirationHour := os.Getenv("ExpirationHour")
```
nb-ohad, line 43: Both the env var name and the variable name do not match the intent. There is no context for what this env var is for and the value is not an exploration date it is the lifetime duration for a given token. Also, env var names should be all capitalized.

```suggestion
	tokenLifetimeInHours := os.Getenv("ONBOARDING_TOKEN_LIFETIME")
```

services/ux-backend/handler/onboarding.go
```diff
@@ -0,0 +1,103 @@
...
+	if expirationHour == "" {
+		expirationHour = "48"
+	}
+
+	expireHour, err := strconv.Atoi(expirationHour)
+	if err != nil {
+		return nil, err
+	}
```
nb-ohad, line 51: Prefered way to read and check the value

```suggestion
	tokenLifetimeInHours, err := strconv.Atoi(os.Getenv("ONBOARDING_TOKEN_LIFETIME"))
	if err != nil {
		tokenLifetimeInHours = 48
	}
```

services/ux-backend/handler/onboarding.go
```diff
@@ -0,0 +1,103 @@
...
+		ID:             uuid.New().String(),
+		ExpirationDate: time.Now().Add(time.Duration(expireHour) * time.Hour).Unix(),
+	}
+
+	payload, err := json.Marshal(data)
+	if err != nil {
+		return nil, err
+	}
+
+	encodedPayload := base64.StdEncoding.EncodeToString(payload)
```
nb-ohad, line 63: I think it will be preferable to use UrlEncoding

services/ux-backend/handler/onboarding.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	// In order to generate the signature, we provide a random number generator,
+	// our private key, the hashing algorithm that we used, and the hash sum
+	// of our message
+	Block, _ := pem.Decode(pemString)
+	privateKey, err := x509.ParsePKCS1PrivateKey(Block.Bytes)
+	if err != nil {
+		return nil, err
+	}
```
nb-ohad, line 85: Why are we reading the private key file and decoding the private key every time we call generated token? 

Preferably load and decode it once into an in-memory scope and reuse it.

services/ux-backend/handler/onboarding.go
```diff
@@ -0,0 +1,103 @@
...
+	onboarding := struct {
+		Token string `json:"token"`
+	}{}
+
+	onboarding.Token = sEnc
+	response, err := json.Marshal(onboarding)
+	if err != nil {
+		return nil, err
+	}
+	return response, nil
```
nb-ohad, line 102: 1. The entire point of an anonymous structure is to use it anonymously 
2. This is a generate token func, it should only care about returning a token, not a wrapper structure. 

```suggestion
	encodedSignature := base64.StdEncoding.EncodeToString(signature)
	return fmt.Sprintf("%s.%s", encodedPayload, encodedSignature)
```

services/ux-backend/handler/onboarding.go
```diff
@@ -0,0 +1,103 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	data := types.OnboardingTicket{
+		ID:             uuid.New().String(),
+		ExpirationDate: time.Now().Add(time.Duration(expireHour) * time.Hour).Unix(),
+	}
+
+	payload, err := json.Marshal(data)
```
nb-ohad, line 58: Data and payload are synonyms, either find better names or just use the struct on location  
```suggestion
        expirationDate := time.Now().Add(time.Duration(expireHour) * time.Hour).Unix()
	payload, err := json.Marshal(types.OnboardingTicket{
		ID:             uuid.New().String(),
		ExpirationDate: expirationDate,
	})
```

services/ux-backend/handler/onboarding.go
nb-ohad, line 1: 1. The folder will hold multiple handlers so lets rename the folder/package to handlers
2. given the go file the name of the resource handler you are implementing not something else (use `onboarding_tokens.go`)

services/ux-backend/main.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	"log"
+	"net/http"
+
+	"github.com/red-hat-storage/ocs-operator/v4/services/ux-backend/handler"
+)
+
+func main() {
+
```
nb-ohad, line 11: You should load and decode the onboarding private key here and abort the server if you cannot.

nb-ohad (changes requested):

services/ux-backend/main.go
```diff
@@ -0,0 +1,15 @@
...
+	http.HandleFunc("/onboarding-tokens", handler.OnboardingTokenHandler)
+	log.Fatal(http.ListenAndServe(":8080", nil))
```
nb-ohad, line 13: The port number should preferably come from an env var UX_BACKEND_PORT

nb-ohad (changes requested):

services/ux-backend/handler/onboarding.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/red-hat-storage/ocs-operator/v4/services/types"
+)
+
+func OnboardingTokenHandler(w http.ResponseWriter, r *http.Request) {
+	switch r.Method {
+	case "POST":
+
+		response, err := getOnboardingToken()
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
```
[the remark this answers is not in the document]
rchikatw: Done

tools/csv-merger/csv-merger.go
```diff
@@ -910,6 +919,118 @@ func copyManifests() {
...
+				"app":                         "ux-backend-server",
+			},
+		},
+		Strategy: appsv1.DeploymentStrategy{Type: appsv1.RecreateDeploymentStrategyType},
+		Template: corev1.PodTemplateSpec{
+			ObjectMeta: metav1.ObjectMeta{
+				Labels: map[string]string{
+					"app.kubernetes.io/component": "ux-backend-server",
+					"app.kubernetes.io/name":      "ux-backend-server",
+					"app.kubernetes.io/version":   "0.0.1",
```
leelavg, line 939: remove this specific label, this doesn't help us w/ anything

tools/csv-merger/csv-merger.go
```diff
@@ -910,6 +919,118 @@ func copyManifests() {
...
+					},
+				},
+				Volumes: []corev1.Volume{
+					{
+						Name: "onboarding-private-key",
+						VolumeSource: corev1.VolumeSource{
+							Secret: &corev1.SecretVolumeSource{
+								SecretName: "onboarding-private-key",
+							},
+						},
```
leelavg, line 1008: if you are already getting this secret mounted by olm, what is the rbac for?

services/ux-backend/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM golang:1.20 as builder
+
+WORKDIR /go/src/github.com/red-hat-storage/ocs-operator
+COPY . .
+USER root
```
leelavg, line 5:
```suggestion
```
- do you need this specifically?

services/ux-backend/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+RUN make build-go
+
+FROM registry.access.redhat.com/ubi9/ubi-minimal
+
+COPY --from=builder /go/src/github.com/red-hat-storage/ocs-operator/build/_output/bin/ux-backend-server /usr/local/bin/ux-backend-server
+
+ENTRYPOINT ["/usr/local/bin/ux-backend-server"]
+
```
leelavg, line 14:
```suggestion
USER operator
```
- pls follow conventions from other Dockerfiles in the repo

services/ux-backend/handlers/onboarding_tokens.go
```diff
@@ -0,0 +1,110 @@
...
+	if err != nil {
+		return "Failed to sign private key", err
+	}
+
+	encodedSignature := base64.StdEncoding.EncodeToString(signature)
+	return fmt.Sprintf("%s.%s", encodedPayload, encodedSignature), nil
+}
+
+func encodeResponse(httpResponse types.HTTPResponse) string {
+	marshaledOnboardingResponse, err := json.Marshal(httpResponse)
```
leelavg, line 105: this may not be only used for onboarding, it's generic, just use `res, err` or better create required methods directly on `type.HTTPResponse`

services/ux-backend/handlers/onboarding_tokens.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/red-hat-storage/ocs-operator/v4/services/types"
+)
+
+func OnboardingTokensHandler(w http.ResponseWriter, r *http.Request) {
+
+	switch r.Method {
+	case "POST":
+		onboardingToken, err := getOnboardingToken()
+		if err != nil {
+			http.Error(w, encodeResponse(types.HTTPResponse{Code: http.StatusMethodNotAllowed, Message: onboardingToken}), http.StatusInternalServerError)
```
leelavg, line 28: shouldn't this be 
```suggestion
			http.Error(w, encodeResponse(types.HTTPResponse{Code: http.StatusMethodNotAllowed, Message: err}), http.StatusInternalServerError)
```

services/ux-backend/handlers/onboarding_tokens.go
```diff
@@ -0,0 +1,110 @@
...
+			return
+		}
+
+		onboarding := struct {
+			Token string `json:"token"`
+		}{}
+
+		onboarding.Token = onboardingToken
```
leelavg, line 36: why can't, pls do format it properly though.
```suggestion
		onboarding := struct {Token string `json:"token"`}{Token: onboardingToken}
```

services/ux-backend/handlers/onboarding_tokens.go
```diff
@@ -0,0 +1,110 @@
...
+		marshaledOnboardingToken, err := json.Marshal(onboarding)
```
leelavg, line 37:
```suggestion
		token, err := json.Marshal(onboarding)
```

services/ux-backend/handlers/onboarding_tokens.go
```diff
@@ -0,0 +1,110 @@
...
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+
+		w.Header().Set("Content-Type", "application/json")
+		//nolint
```
leelavg, line 44: why `nolint`?

services/ux-backend/handlers/onboarding_tokens.go
```diff
@@ -0,0 +1,110 @@
...
+	default:
+		http.Error(w, encodeResponse(types.HTTPResponse{Code: http.StatusMethodNotAllowed, Message: "Only POST request is allowed"}), http.StatusMethodNotAllowed)
+		return
+	}
+}
+
+func getOnboardingToken() (string, error) {
+
+	tokenLifetimeInHours, err := strconv.Atoi(os.Getenv("ONBOARDING_TOKEN_LIFETIME"))
+	if err != nil {
```
leelavg, line 56: could you pls also add minor checks? something like 0 < hr <= 48, we could use negative for lifetime but that might be a security issue if unused.

tools/csv-merger/csv-merger.go
```diff
@@ -910,6 +919,118 @@ func copyManifests() {
...
+	deployment := appsv1.DeploymentSpec{
+		Replicas: &replica,
+		Selector: &metav1.LabelSelector{
+			MatchLabels: map[string]string{
+				"app.kubernetes.io/component": "ux-backend-server",
+				"app.kubernetes.io/name":      "ux-backend-server",
+				"app":                         "ux-backend-server",
+			},
+		},
+		Strategy: appsv1.DeploymentStrategy{Type: appsv1.RecreateDeploymentStrategyType},
```
leelavg, line 933:
```suggestion
		Strategy: appsv1.RecreateDeploymentStrategyType,
```

tools/csv-merger/csv-merger.go
```diff
@@ -910,6 +919,118 @@ func copyManifests() {
...
+				Containers: []corev1.Container{
+					{
+						Name: "ux-backend-server",
+						VolumeMounts: []corev1.VolumeMount{
+							{
+								Name:      "onboarding-private-key",
+								MountPath: "/etc/private-key",
+							},
+						},
+						Image:   *uxBackendServerImage,
```
leelavg, line 953: set `ImagePullPolicy: IfNotPresent` as well

tools/csv-merger/csv-merger.go
```diff
@@ -935,6 +1056,10 @@ func main() {
 		log.Fatal("--crds-directory is required")
 	} else if *outputDir == "" {
 		log.Fatal("--olm-bundle-directory is required")
+	} else if *uxBackendServerImage == "" {
+		log.Fatal("--olm-bundle-directory is required")
+	} else if *uxBackendOauthImage == "" {
+		log.Fatal("--olm-bundle-directory is required")
```
leelavg, line 1062:
```suggestion
	} else if *uxBackendServerImage == "" {
		log.Fatal("--ux-backend-server-image is required")
	} else if *uxBackendOauthImage == "" {
		log.Fatal("--ux-backend-oauth-image is required")
```

tools/csv-merger/csv-merger.go
```diff
@@ -935,6 +1056,10 @@ func main() {
 		log.Fatal("--crds-directory is required")
 	} else if *outputDir == "" {
 		log.Fatal("--olm-bundle-directory is required")
+	} else if *uxBackendServerImage == "" {
+		log.Fatal("--olm-bundle-directory is required")
+	} else if *uxBackendOauthImage == "" {
```
leelavg, line 1061: if default is set, I don't think this value will ever be not present, better remove default as well and just have a comment here saying what image can be used for easy visilbility

tools/csv-merger/csv-merger.go
```diff
@@ -910,6 +919,118 @@ func copyManifests() {
...
+						Ports: []corev1.ContainerPort{
+							{
+								ContainerPort: 8888,
+							},
+						},
+					},
+				},
+				Volumes: []corev1.Volume{
+					{
+						Name: "onboarding-private-key",
```
leelavg, line 1003: afair, we are creating the job conditionally for provider mode and in other installs you might not find this secret created so this pod will fail and so does the csv.

i believe, you need to get the secret when you get a token request.

controllers/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -175,6 +197,8 @@ func (r *OCSInitializationReconciler) SetupWithManager(mgr ctrl.Manager) error {
 
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&ocsv1.OCSInitialization{}).
+		Owns(&corev1.Service{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		Owns(&corev1.Secret{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
```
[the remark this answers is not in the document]
nb-ohad: @leelavg No need. This is an `Owns` watch which means it has a built-in filter for only secrets and services that are owned by the initialization resource.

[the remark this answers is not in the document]
leelavg: Ah! You are correct. Missed it.

nb-ohad (changes requested):

services/ux-backend/handler/onboarding.go
nb-ohad: Agreed (for now)

tools/csv-merger/csv-merger.go
leelavg: ack, didn't know that, thanks.

tools/csv-merger/csv-merger.go
rchikatw: In the code it's already handle if key is not there i will just add the optional

services/ux-backend/main.go
rchikatw: Okay

services/ux-backend/handlers/onboarding_tokens.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	// In order to generate the signature, we provide a random number generator,
+	// our private key, the hashing algorithm that we used, and the hash sum
+	// of our message
+	Block, _ := pem.Decode(pemString)
+	privateKey, err := x509.ParsePKCS1PrivateKey(Block.Bytes)
+	if err != nil {
+		return "Failed to parse private key", err
+	}
```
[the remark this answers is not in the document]
rchikatw: Yes sounds good can we do it after release?

controllers/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -327,3 +351,76 @@ func (r *OCSInitializationReconciler) getEnableNFSKeyValue() string {
...
+	secret := &corev1.Secret{}
+	secret.Name = "ux-proxy-secret"
+	secret.Namespace = initialData.Namespace
+
+	opResult, err := ctrl.CreateOrUpdate(r.ctx, r.Client, secret, func() error {
+
+		if err := ctrl.SetControllerReference(initialData, secret, r.Scheme); err != nil {
+			return err
+		}
+		secret.StringData = map[string]string{"session_secret": strings.Replace(uuid.New().String()+uuid.New().String(), "-", "", -1)}
```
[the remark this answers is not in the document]
bipuladh: `cookie-secret string: the seed string for secure cookies` as per the docs
since it's a seed string I assume having a randomly generated string should be enough,

services/ux-backend/handlers/onboarding_tokens.go
```diff
@@ -0,0 +1,110 @@
...
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+
+		w.Header().Set("Content-Type", "application/json")
+		//nolint
+		w.Write(marshaledOnboardingToken)
+
+	default:
+		http.Error(w, encodeResponse(types.HTTPResponse{Code: http.StatusMethodNotAllowed, Message: "Only POST request is allowed"}), http.StatusMethodNotAllowed)
```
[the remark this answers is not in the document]
rchikatw: This is the format suggested by Leela, and I used it looks like a common format. Hence i used

controllers/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -327,3 +351,76 @@ func (r *OCSInitializationReconciler) getEnableNFSKeyValue() string {
...
+					TargetPort: intstr.IntOrString{
+						Type:   intstr.Int,
+						IntVal: 8888,
+					},
+				},
+			},
+			Selector:        map[string]string{"app": "ux-backend-server"},
+			SessionAffinity: "None",
+			Type:            "ClusterIP",
+		}
```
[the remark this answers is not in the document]
rchikatw: here i care about all the specs, if any field is failed/updated wrong ui may break(not sure on this part).

controllers/ocsinitialization/ocsinitialization_controller.go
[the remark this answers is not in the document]
leelavg: > having a randomly generated string should be enough
- better to add the lifetime as well, like should it be per reconcile or per deployment

services/ux-backend/handlers/onboarding_tokens.go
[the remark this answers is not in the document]
leelavg: sry, I didn't suggest, I just didn't deny.

leelavg (commented):
apart from the http response/error format I'm fine w/ the rest and if possible I want to remove uuid and just use golang stdlib to generate a base64 string.

leelavg:
@nb-ohad Nitin is also fine w/ having the server as part of ocs-operator Dockerfile, seeing the changes I don't think we need a new image and build team wants a discussion if we want to build a new image.

leelavg:
@rchikatw I believe CI is failing in building new image which might need config changes in another repo, I think it's better to build it in same ocs-operator image.

nb-ohad:
@leelavg I never asked for a different image. We can use the ocs-operator one. I think it is the same for the API server
But this PR is not ready yes

tools/csv-merger/csv-merger.go
nb-ohad: Not sure. I might be mistaken. We should test and then decide what to do.

services/ux-backend/Dockerfile
nb-ohad: We should run rootless

leelavg:
> @leelavg I never asked for a different image.
- I was told it was discussed we wanted a different image. Yes, I also don't want a separate image.

controllers/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -327,3 +342,65 @@ func (r *OCSInitializationReconciler) getEnableNFSKeyValue() string {
...
+
+	r.Log.Info("Secret creation succeeded", "Name", desiredSecret.Name)
+
+	return nil
+}
+
+func (r *OCSInitializationReconciler) reconcileUXBackendService(initialData *ocsv1.OCSInitialization) error {
+
+	var err error
+	annotations := map[string]string{"service.alpha.openshift.io/serving-cert-secret-name": "ux-cert-secret"}
```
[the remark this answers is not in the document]
rchikatw: Sure

Makefile
```diff
@@ -9,6 +9,7 @@ CONTROLLER_GEN=$(LOCALBIN)/controller-gen
 .PHONY: \
 	build \
 	gen-protobuf \
+	build-go \
```
leelavg, line 12: Not used.

deploy/deploy-with-olm.yaml
```diff
@@ -23,10 +23,7 @@ metadata:
...
-    mediatype: image/svg+xml
-  image: quay.io/ocs-dev/ocs-operator-catalog:latest
+  image: quay.io/rchikatw/ocs-operator-catalog:latest201
```
leelavg, line 26: Remove local changes

tools/csv-merger/csv-merger.go
```diff
@@ -160,6 +163,10 @@ func unmarshalCSV(filePath string) *csvv1.ClusterServiceVersion {
 				Name:  "ONBOARDING_SECRET_GENERATOR_IMAGE",
 				Value: *ocsContainerImage,
 			},
+			{
```
leelavg, line 166: Isn't ux a separate deployment, do we need this env?

controllers/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -327,3 +344,82 @@ func (r *OCSInitializationReconciler) getEnableNFSKeyValue() string {
...
+
+	b := make([]byte, 20)
+	_, err = rand.Read(b)
+	if err != nil {
+		r.Log.Error(err, "Failed to generate random string for ux-backend secret")
+		return err
+	}
+	session_secret := base64.URLEncoding.EncodeToString(b)[:20]
+
+	secret.StringData = map[string]string{"session_secret": session_secret}
```
leelavg, line 364: minor: session_secret can be inlined

services/ux-backend/main.go
```diff
@@ -0,0 +1,28 @@
...
+
+	klog.Info("Starting ux backend server")
+
+	http.HandleFunc("/onboarding-tokens", handler.OnboardingTokensHandler)
+
+	listenPort, err := strconv.Atoi(os.Getenv("UX_BACKEND_PORT"))
+	if err != nil {
+		listenPort = 8080
+	}
+	klog.Info("ux backend server listening on port ", listenPort)
```
nb-ohad, line 24: Please log differentiate between an empty env var, where it makes sense to default to 8080, and one that has an invalid port, where it makes sense to report and error and shutdown.
```suggestion
	var listenPort int;
	listenPortAsString = os.Getenv("UX_BACKEND_PORT")
	if listenPortAsString == "" {
		klog.Infof("No user-defined server listening port provided, defaulting to %d, defaultListeningPort)
		listenPort = defaultListeningPort
	} else if listenPort, err :=  strconv.Atoi(listenPortAsString); err != nil {
		klog.Errorf("Malformed user-defined listening port: %s. shuting down", listenPortAsString)
		os.Exit(-1)
	}
```

nb-ohad (changes requested):

nb-ohad:
@rchikatw A general note for a common mistake I saw in the last couple of reviews. Pure logic functions that don't hold a state (like `generateOnboardingToken`) shouldn't write directly to the log. instead warp any errors with context (using `fmt.Errorf`) and return them up the call chain until you get to a stateful function that has to deal with a logger. For this PR a good place can be main or the endpoint handler.

tools/csv-merger/csv-merger.go
```diff
@@ -935,6 +1078,11 @@ func main() {
 		log.Fatal("--crds-directory is required")
 	} else if *outputDir == "" {
 		log.Fatal("--olm-bundle-directory is required")
+	} else if *uxBackendServerImage == "" {
```
leelavg, line 1081: Not used, pls remove.

tools/csv-merger/csv-merger.go
```diff
@@ -825,6 +838,18 @@ func injectCSVRelatedImages(r *unstructured.Unstructured) error {
 			"image": *ocsMetricsExporterImage,
 		})
 	}
+	if *uxBackendServerImage != "" {
```
leelavg, line 841: Not used, pls remove.

leelavg (commented):
See my previous comments as well, there is some dead code.

services/ux-backend/handlers/onboarding_tokens.go
```diff
@@ -0,0 +1,131 @@
...
+		Unix()
+
+	payload, err := json.Marshal(types.OnboardingTicket{
+		ID:             uuid.New().String(),
+		ExpirationDate: expirationDate,
+	})
+
+	if err != nil {
+		klog.Error(err, "failed to marshal the payload")
+		return "", err
```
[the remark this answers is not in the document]
rchikatw: make sense

services/ux-backend/handlers/onboarding_tokens.go
```diff
@@ -0,0 +1,131 @@
...
+
+	msgHashSum := msgHash.Sum(nil)
+	signature, err := rsa.SignPKCS1v15(rand.Reader, privateKey, crypto.SHA256, msgHashSum)
+	if err != nil {
+		klog.Error(err, "failed to sign private key")
+		return "", err
+	}
+
+	encodedSignature := base64.StdEncoding.EncodeToString(signature)
+	klog.Info("token is generated and valid for ", tokenLifetimeInHours, " Hours")
```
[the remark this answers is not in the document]
rchikatw: I will remove this and as a security concern we have decided not to log token.

services/ux-backend/handlers/onboarding_tokens.go
```diff
@@ -0,0 +1,131 @@
...
+	return fmt.Sprintf("%s.%s", encodedPayload, encodedSignature), nil
+}
+
+func readAndDecodeOnboardingPrivateKey() (*rsa.PrivateKey, error) {
+
+	pemString, err := os.ReadFile("/etc/private-key/key")
```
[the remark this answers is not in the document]
rchikatw: yes i have mounted there only in the deployment

services/ux-backend/handlers/onboarding_tokens.go
```diff
@@ -0,0 +1,131 @@
...
+	if err != nil {
+		klog.Error(err, "failed to read onboarding private key")
+		return nil, err
```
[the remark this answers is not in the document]
rchikatw: Done

services/ux-backend/handlers/onboarding_tokens.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	// In order to generate the signature, we provide a random number generator,
+	// our private key, the hashing algorithm that we used, and the hash sum
+	// of our message
+	Block, _ := pem.Decode(pemString)
+	privateKey, err := x509.ParsePKCS1PrivateKey(Block.Bytes)
+	if err != nil {
+		klog.Error(err, "failed to parse private key")
+		return nil, err
```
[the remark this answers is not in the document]
rchikatw: Done

services/ux-backend/main.go
rchikatw: right make sense

tools/csv-merger/csv-merger.go
rchikatw: removing all reference of ux_backend_server_image related constants

services/ux-backend/handlers/onboarding_tokens.go
```diff
@@ -0,0 +1,131 @@
...
+	"github.com/red-hat-storage/ocs-operator/v4/services/types"
+	"k8s.io/klog/v2"
+)
+
+func OnboardingTokensHandler(w http.ResponseWriter, r *http.Request) {
+
+	var err error
+	switch r.Method {
+	case "POST":
+		onboardingToken, err := generateOnboardingToken()
```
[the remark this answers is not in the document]
leelavg: An init function w/ pkg level var might fit the purpose.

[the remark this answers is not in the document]
rchikatw: Right now i have added in the main.go, that is good enough i feel

controllers/ocsinitialization/ocsinitialization_controller.go
rchikatw: not needed, made it multiline as ohad requested

rbac/onboarding-proxy-sa.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
```
leelavg, line 1: You should create this file under config/rbac and add it to Kustomization.yml, name should be ux_backend_sa.yaml not the proxy.

I believe if you delete this file and generate bundle this'll not get created unless you add to Kustomization.yml

services/ux-backend/main.go
```diff
@@ -0,0 +1,56 @@
...
+	"strconv"
+
+	handler "github.com/red-hat-storage/ocs-operator/v4/services/ux-backend/handlers"
+	"k8s.io/klog/v2"
+)
+
+var tokenLifetimeInHours int
+var listenPort int
+
+func init() {
```
leelavg, line 17: I didn't mean to push all logic to init, I suggested only for lifetime and that too in handler file. init should be used sparingly.

services/ux-backend/handlers/onboarding_tokens.go
```diff
@@ -0,0 +1,120 @@
...
+
+		onboardingToken, err := generateOnboardingToken(tokenExpirationDate)
+		if err != nil {
+			klog.Error(err, "failed to get onboardig token")
+			w.WriteHeader(http.StatusInternalServerError)
+			w.Header().Set("Content-Type", "text/text")
+			_, err = w.Write([]byte("Failed to generate token"))
+
+			if err != nil {
+				klog.Error(err, "failed write data to response writer")
```
nb-ohad, line 42: By logging the error first you relegate the issue to the last thing in the message. Preferably use:
```suggestion
				klog.Errorf("failed write data to response writer: %v", err)
```

services/ux-backend/handlers/onboarding_tokens.go
```diff
@@ -0,0 +1,120 @@
...
+			return
+		}
+
+		klog.Info("onboarding token generated successfully")
+		w.WriteHeader(http.StatusOK)
+		w.Header().Set("Content-Type", "text/text")
+
+		_, err = w.Write([]byte(onboardingToken))
+		if err != nil {
+			klog.Error(err, "failed write data to response writer")
```
nb-ohad, line 53: By logging the error first you relegate the issue to the last thing in the message. Preferably use:
```suggestion
			klog.Errorf("failed write data to response writer, %v", err)
```

services/ux-backend/handlers/onboarding_tokens.go
```diff
@@ -0,0 +1,120 @@
...
+			return
+		}
+
+	default:
+		klog.Info("Only POST method should be used to send data to this endpoint /onboarding-tokens")
+		w.WriteHeader(http.StatusMethodNotAllowed)
+		w.Header().Set("Content-Type", "text/text")
+		_, err = w.Write([]byte(fmt.Sprintf("Unsupported method : %s", r.Method)))
+		if err != nil {
+			klog.Error(err, "failed write data to response writer")
```
nb-ohad, line 63: By logging the error first you relegate the issue to the last thing in the message. Preferably use:
```suggestion
			klog.Errorf("failed write data to response writer: %v", err)
```

services/ux-backend/handlers/onboarding_tokens.go
```diff
@@ -0,0 +1,120 @@
...
+		}
+		return
+	}
+}
+
+func generateOnboardingToken(tokenExpirationDate int64) (string, error) {
```
nb-ohad, line 69: If you send a `tokenExpirationDate` then you must validate that it is at least one hour into the future as the context of this function needs to guarantee it. As an alternative, and to remove confusion, just send the lifetime instead
```suggestion
unc generateOnboardingToken(tokenLifetimeInHours int) (string, error) {
```

services/ux-backend/main.go
```diff
@@ -0,0 +1,56 @@
...
+		klog.Infof("No user-defined server listening port provided, defaulting to %d ", defaultListeningPort)
+		listenPort = defaultListeningPort
+	} else if listenPort, err = strconv.Atoi(listenPortAsString); err != nil {
+		klog.Errorf("Malformed user-defined listening port: %s. shutting down", listenPortAsString)
+		os.Exit(-1)
+	}
+
+	klog.Info("ux backend server listening on port ", listenPort)
+
+}
```
nb-ohad, line 38: This is not a good use case for init. Init has a finicky invocation time (the first time the package is referenced). Please move everything into `main`. I don't see a use case for `init` for this server.

services/ux-backend/handlers/onboarding_tokens.go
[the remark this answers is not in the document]
nb-ohad: @leelavg @rchikatw 
An init function should be used only in extreme cases where you need to guarantee code loaded at package reference time.  It adds a lot of ambiguity to the order of execution and might change it just because someone had referenced a package something in another module. It is a bad practice to use it unless you are trying to implement something like dependency injection mechanism. 

To add to the above, we are in the main module, there will be no difference between init and just having the code in main (which is always invoked first). 

Let's remove the complexity and have everything back in main.

services/ux-backend/main.go
```diff
@@ -0,0 +1,56 @@
...
+var listenPort int
+
+func init() {
+
+	var err error
+	tokenLifetimeInHours, err = strconv.Atoi(os.Getenv("ONBOARDING_TOKEN_LIFETIME"))
+	if err != nil || tokenLifetimeInHours <= 0 {
+		tokenLifetimeInHours = 48
+	}
+	klog.Info("generated token will be valid for ", tokenLifetimeInHours, " Hours")
```
nb-ohad, line 24: 1. To control the way we present the values use the formatted version of Info -> Infof
```suggestion
	klog.Infof("generated token will be valid for %d hours", tokenLifetimeInHours)
```

2. This message is not very helpful. You do not capture the source of the value (default or user-provided). You do not exit in cases where the user provided a value but it was malformed. Please  take a look at a previous comment about UX_BACKEND_PORT env var for reference

services/ux-backend/handlers/onboarding_tokens.go
```diff
@@ -0,0 +1,120 @@
...
+
+func OnboardingTokensHandler(w http.ResponseWriter, r *http.Request, tokenLifetimeInHours int) {
+
+	var err error
+	switch r.Method {
+	case "POST":
+
+		tokenExpirationDate := time.Now().
+			Add(time.Duration(tokenLifetimeInHours) * time.Hour).
+			Unix()
```
nb-ohad, line 32: This should be done inside the generate token function. IF not the generated token function who have no context about the given date will need to validate the date and return an error (if it is in the past, or less then 1 hour into the future). I always prefer to use delta times when providing information to nested code. This way the only thing that need validation is the sign of the value.

nb-ohad (changes requested):

services/ux-backend/main.go
rchikatw: Done.

services/ux-backend/main.go
rchikatw: Done

services/ux-backend/handlers/onboarding_tokens.go
rchikatw: moved to generate token function

rbac/onboarding-proxy-sa.yaml
rchikatw: I discussed it with leela over DM and we are keeping it here and removing --extra-service-account

rchikatw: i can rename the file as it make sense

services/ux-backend/main.go
```diff
@@ -0,0 +1,58 @@
...
+	tokenLifetimeInHoursAsString := os.Getenv("ONBOARDING_TOKEN_LIFETIME")
+	if tokenLifetimeInHoursAsString == "" {
+		klog.Infof("No user-defined token lifetime provided, defaulting to %d ", defaultTokenLifetimeInHours)
+		tokenLifetimeInHours = defaultTokenLifetimeInHours
+	} else if tokenLifetimeInHours, err = strconv.Atoi(tokenLifetimeInHoursAsString); err != nil {
+		klog.Errorf("Malformed user-defined Token lifetime: %s. shutting down", tokenLifetimeInHoursAsString)
+		os.Exit(-1)
+	}
+
+	klog.Infof("generated token will be valid for %d hours", tokenLifetimeInHours)
```
nb-ohad, line 32:
```suggestion
	klog.Infof("generated tokens will be valid for %d hours", tokenLifetimeInHours)
```

services/ux-backend/main.go
```diff
@@ -0,0 +1,58 @@
...
+
+	defaultListeningPort := 8080
+	listenPortAsString := os.Getenv("UX_BACKEND_PORT")
+	if listenPortAsString == "" {
+		klog.Infof("No user-defined server listening port provided, defaulting to %d ", defaultListeningPort)
+		listenPort = defaultListeningPort
+	} else if listenPort, err = strconv.Atoi(listenPortAsString); err != nil {
+		klog.Errorf("Malformed user-defined listening port: %s. shutting down", listenPortAsString)
+		os.Exit(-1)
+	}
```
nb-ohad, line 42: Having all of this code in the main func context is cumbersome and not scalable.
Can we move this logic into a separate func:
``` func loadAndValidateServerConfig() *serverConfig {}```

Then `serverConfig` can be defined as:
```
type serverConfig struct {
    listenPort int
    tokenLifetimeInHours int
} 
```

This will allow us to extend the config and isolate the approach to the config (maybe in the future we will decide to have all of the config in a config map and not env vars)

**Review by nb-ohad — CHANGES_REQUESTED** [review]

File: services/ux-backend/main.go [hunk 66]
```diff
@@ -0,0 +1,66 @@
...
+
+	var err error
+	defaultTokenLifetimeInHours := 48
+	tokenLifetimeInHoursAsString := os.Getenv("ONBOARDING_TOKEN_LIFETIME")
+	if tokenLifetimeInHoursAsString == "" {
+		klog.Infof("No user-defined token lifetime provided, defaulting to %d ", defaultTokenLifetimeInHours)
+		config.tokenLifetimeInHours = defaultTokenLifetimeInHours
+	} else if config.tokenLifetimeInHours, err = strconv.Atoi(tokenLifetimeInHoursAsString); err != nil {
+		klog.Errorf("Malformed user-defined Token lifetime: %s. shutting down", tokenLifetimeInHoursAsString)
+		os.Exit(-1)
```
nb-ohad, line 30: These kinds of exists should be reserved for the top-most standalone invocation unit. As this is a standalone function it should covey issues by returning errors and the caller should make a decision about error logging, error handling, and program flow. I already mentioned all of these point in a previous PR regarding the generate token func.

To comply please have the following changes:
1. Instead of log, build and return an error using  `fmt.Errorf`
2. On the caller site, log the error and exist.

services/ux-backend/main.go
```diff
@@ -0,0 +1,66 @@
...
+
+	klog.Info("ux backend server listening on port ", config.listenPort)
+	return &config
+}
+
+func main() {
+
+	klog.Info("Starting ux backend server")
+
+	config := loadAndValidateServerConfig()
```
nb-ohad, line 53: `loadAndValidateServerConfig` should return error as well and program flow should be handled at this point in the code.

nb-ohad (changes requested):

rbac/onboarding-proxy-sa.yaml
nb-ohad: The server should not run with the same service account as OCS operator. if we remove this, which service account should we use instead?

leelavg: The premise is if we are mentioning a serviceaccount in csv then olm will create and manage that serviceaccount implictly - this seems to be how ocs operator does

(or)

if we have `--extra-service-account` arg, an explicit file needs to be added and olm will just deploy that sa - this is how client operator does

But, if we have an explicit sa file with no arg supplied to bundle target, then current csv owns sa and during upgrade new csv also tries to own existing sa which is already being owned by old csv and get stuck. I remember seeing this scenario during deployer debugging.

I'm so sry for holding this up, I want to hear that the deployment is happening as expected when we are testing it properly.

leelavg: @rchikatw my initial comment is correct! Pls have role, binding and sa files in config/rbac and use extra-sa arg as well.

PR #2344, opened 2023-12-24
correction of pkg names in services

leelavg:
move types from it's own pkg to services level as this'll be used by multiple services at sub levels.

rename handler pkg to handlers in ux-backend-server services


leelavg:
closing in favor of #2345

PR #2359, opened 2024-01-03
Update go modules to Golang v1.21

umangachapagain:
Updating go modules to Golang v1.21. Also, removing tests for go 1.20 as it might cause false alarms.


iamniting (changes requested):
Lets update the other places as well, like dockerfiles etc.

iamniting (approved):

PR #2362, opened 2024-01-04
Set current mon count value in StorageCluster CR

Nikhil-Ladha:
Added a new field under Status of StorageCluster CR named `CurrentMonCount` that will hold the the mon count set for
the cluster, and this value is updated based on the mon count value set in CephCluster CR

### DIFF
--- a/controllers/storagecluster/cephcluster.go
+++ b/controllers/storagecluster/cephcluster.go
@@ -343,6 +343,11 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
 		}
 	}
 
+	// Update the currentMonCount field in StoragCluster status from the cephCluster CR
+	if !sc.Spec.ExternalStorage.Enable {
+		sc.Status.CurrentMonCount = cephCluster.Spec.Mon.Count
+	}
+
 	// Create the prometheus rules if required by the cephcluster CR
 	if err := createPrometheusRules(r, sc, cephCluster); err != nil {
 		r.Log.Error(err, "Unable to create or update prometheus rules.", "CephCluster", klog.KRef(found.Namespace, found.Name))
--- a/api/v1/storagecluster_types.go
+++ b/api/v1/storagecluster_types.go
@@ -514,6 +514,9 @@ type StorageClusterStatus struct {
 
 	// KMSServerConnection holds the connection state to the KMS server.
 	KMSServerConnection KMSServerConnectionStatus `json:"kmsServerConnection,omitempty"`
+
+	// CurrentMonCount holds the value of ceph mons configured in ceph cluster.
+	CurrentMonCount int `json:"currentMonCount,omitempty"`
 }
 
 // ImagesStatus maps every component image name it's reconciliation status information
--- a/config/crd/bases/ocs.openshift.io_storageclusters.yaml
+++ b/config/crd/bases/ocs.openshift.io_storageclusters.yaml
@@ -6340,6 +6340,10 @@ spec:
                   - type
                   type: object
                 type: array
+              currentMonCount:
+                description: CurrentMonCount holds the value of ceph mons configured
+                  in ceph cluster.
+                type: integer
               externalSecretHash:
                 description: ExternalSecretHash holds the checksum value of external
                   secret data.
--- a/deploy/csv-templates/crds/ocs/ocs.openshift.io_storageclusters.yaml
+++ b/deploy/csv-templates/crds/ocs/ocs.openshift.io_storageclusters.yaml
@@ -6340,6 +6340,10 @@ spec:
                   - type
                   type: object
                 type: array
+              currentMonCount:
+                description: CurrentMonCount holds the value of ceph mons configured
+                  in ceph cluster.
+                type: integer
               externalSecretHash:
                 description: ExternalSecretHash holds the checksum value of external
                   secret data.
--- a/deploy/ocs-operator/manifests/storagecluster.crd.yaml
+++ b/deploy/ocs-operator/manifests/storagecluster.crd.yaml
@@ -6339,6 +6339,10 @@ spec:
                   - type
                   type: object
                 type: array
+              currentMonCount:
+                description: CurrentMonCount holds the value of ceph mons configured
+                  in ceph cluster.
+                type: integer
               externalSecretHash:
                 description: ExternalSecretHash holds the checksum value of external
                   secret data.
--- a/vendor/github.com/red-hat-storage/ocs-operator/api/v4/v1/storagecluster_types.go
+++ b/vendor/github.com/red-hat-storage/ocs-operator/api/v4/v1/storagecluster_types.go
@@ -514,6 +514,9 @@ type StorageClusterStatus struct {
 
 	// KMSServerConnection holds the connection state to the KMS server.
 	KMSServerConnection KMSServerConnectionStatus `json:"kmsServerConnection,omitempty"`
+
+	// CurrentMonCount holds the value of ceph mons configured in ceph cluster.
+	CurrentMonCount int `json:"currentMonCount,omitempty"`
 }
 
 // ImagesStatus maps every component image name it's reconciliation status information


controllers/storagecluster/reconcile.go
```diff
@@ -163,6 +164,11 @@ func (r *StorageClusterReconciler) Reconcile(ctx context.Context, request reconc
 		return reconcile.Result{}, err
 	}
 
+	// Set default mon count value in the StorageCluster CR
+	if sc.Spec.ManagedResources.CephCluster.MonCount == 0 {
```
umangachapagain, line 168: You should also set the default on the API.

umangachapagain (changes requested):

controllers/storagecluster/reconcile.go
Nikhil-Ladha: Done

umangachapagain (changes requested):
Looks good to me.
Please put the generated changes in a separate commit.

Nikhil-Ladha:
> Looks good to me. Please put the generated changes in a separate commit.

Done

controllers/storagecluster/reconcile.go
```diff
@@ -163,6 +164,11 @@ func (r *StorageClusterReconciler) Reconcile(ctx context.Context, request reconc
...
+		sc.Spec.ManagedResources.CephCluster.MonCount = defaults.DefaultMonCount
```
malayparida2000, line 169: @umangachapagain @iamniting AFAIK we from the operator code should not set any fields in the CR Spec from the code, isn't this breaking that? Isn't just setting it from the API not enough?

malayparida2000:
@Nikhil-Ladha Can you share any BZ or something where we can see the requirement for the value to be on the CR for UI to pick up

controllers/storagecluster/reconcile.go
malayparida2000: https://github.com/red-hat-storage/ocs-operator/blob/49da9ae0b3a85da5f141a865520f87995249be90/controllers/storagecluster/cephcluster.go#L717-L726
This will create a confusing state in arbiter mode, where the MonCount will be set to 3 on the storageCluster Cr, but for arbiter mode, we will pass the monCount as 5 to the CephCluster CR. So 3 on StorageCluster but 5 mons in actual and in the CephCluster will be confusing.

Nikhil-Ladha:
> @Nikhil-Ladha Can you share any BZ or something where we can see the requirement & understand exactly why the value needs to be on the CR for UI to pick up

So, this is related to this epic: https://issues.redhat.com/browse/RHSTOR-4474, in which as part of the UI change we are showing a modal that shows the current monCount in the cluster. And, that value is picked up from the CR itself. But, since we are not setting any value by default in the CR, the field stays empty in the UI modal.
Screenshot 2024-01-03 at 3 40 20 PM

controllers/storagecluster/reconcile.go
Nikhil-Ladha: Makes sense, but based on the use case probably it will be better to have the arbiter check in the updated code as well. WDYT?

iamniting:
pls check if there is a possibility using cel to set the default value dynamically based on the arbiter cluster.

https://kubernetes.io/docs/reference/using-api/cel/

malayparida2000:
> > @Nikhil-Ladha Can you share any BZ or something where we can see the requirement & understand exactly why the value needs to be on the CR for UI to pick up
> 
> So, this is related to this epic: https://issues.redhat.com/browse/RHSTOR-4474, in which as part of the UI change we are showing a modal that shows the current monCount in the cluster. And, that value is picked up from the CR itself. But, since we are not setting any value by default in the CR, the field stays empty in the UI modal. Screenshot 2024-01-03 at 3 40 20 PM

Can't this field be picked up from the cephcluster directly? Because the actual mon count is there only.
I think UI can read  the spec.mon.count. This can actually give the exact count of mon present no matter the scenario

Nikhil-Ladha:
> > > @Nikhil-Ladha Can you share any BZ or something where we can see the requirement & understand exactly why the value needs to be on the CR for UI to pick up
> > 
> > 
> > So, this is related to this epic: https://issues.redhat.com/browse/RHSTOR-4474, in which as part of the UI change we are showing a modal that shows the current monCount in the cluster. And, that value is picked up from the CR itself. But, since we are not setting any value by default in the CR, the field stays empty in the UI modal. Screenshot 2024-01-03 at 3 40 20 PM
> 
> Can't this field be picked up from the cephcluster directly? Because the actual mon count is there only. I think UI can read the spec.mon.count. This can actually give the exact count of mon present no matter the scenario

It's not feasible by design. Because, the feature is about updating the mon count in the Storagecluster CR if the env allows 5 or more mons, and which in turn updates the mon count in the cephCluster. We can't take the initial value from ceph Cluster and then update the value in the StorageCluster CR later on.

Nikhil-Ladha:
> pls check if there is a possibility using cel to set the default value dynamically based on the arbiter cluster.
> 
> https://kubernetes.io/docs/reference/using-api/cel/

I couldn't find anything to set the default value dynamically using cel, from what I could get we can use cel for validation purpose only and not to set values to a field.

iamniting:
> > pls check if there is a possibility using cel to set the default value dynamically based on the arbiter cluster.
> > https://kubernetes.io/docs/reference/using-api/cel/
> 
> I couldn't find anything to set the default value dynamically using cel, from what I could get we can use cel for validation purpose only and not to set values to a field.

I think we can set the default value in the spec or status itself and they can pick it up from there itself.

Nikhil-Ladha:
> > > pls check if there is a possibility using cel to set the default value dynamically based on the arbiter cluster.
> > > https://kubernetes.io/docs/reference/using-api/cel/
> > 
> > 
> > I couldn't find anything to set the default value dynamically using cel, from what I could get we can use cel for validation purpose only and not to set values to a field.
> 
> I think we can set the default value in the spec or status using the controller and they can pick it up from there.

Yeah, the current code does that.
I have updated the code to consider the arbiter scenario as well

iamniting:
> > > > pls check if there is a possibility using cel to set the default value dynamically based on the arbiter cluster.
> > > > https://kubernetes.io/docs/reference/using-api/cel/
> > > 
> > > 
> > > I couldn't find anything to set the default value dynamically using cel, from what I could get we can use cel for validation purpose only and not to set values to a field.
> > 
> > 
> > I think we can set the default value in the spec or status using the controller and they can pick it up from there.
> 
> Yeah, the current code does that. I have updated the code to consider the arbiter scenario as well

We are still setting the default value to 3 with tags which we should not.

Nikhil-Ladha:
Done, please check now.

iamniting (changes requested):
Lets create a new func where we can set these kind of values. like `setDynamicDefaults`

controllers/storagecluster/reconcile.go
umangachapagain: Mon count should be read from StorageCluster CR and then written to CephCluster CR.
Validations can be done to ensure valid Mon count is provided.

By default StorageCluster will not have any Mon count. So, internally, we decide to use `ArbiterModeMonCount` or `DefaultMonCount`. But, this is updated in CephCluster and not StorageCluster.

If StorageCluster Mon count is provided, and
1. if we have arbiter mode enabled, then use `ArbiterModeMonCount`.
2. if we have enough zones, then use 5.
3. if none of the above, use 3.

Technically, this shouldn't even be a user option as user does not have much choice here.

umangachapagain: @travisn What do you think of this?

- If we allow users to choose and say they choose MonCount: 3 and enable Arbiter mode, they will see 5 Mons.
- If they choose MonCount: 5 and they don't have 5 zones, they'd have multiple MONs in same zone which isn't desired.
- If they choose MonCount: 3 and disable arbiter but have 5 zones, we expect them to have 5 MONs, which isn't what they desired.

So, it looks like we are throwing a bunch of issues at the user to force them to use a MonCount that we want them to use.

If we still want to make it a user option, we should only allow them to set it to 5 or keep it undefined so we can apply our default.

Nikhil-Ladha: This use case of applying monCount was introduced to tackle the situation where a. user has 5 or more zones and they want to have 5 mons.
Right now, we have an alert configured that would tell them to update the mon count if they have 5 or more zones, but the user can choose to silence that alert. And, continue with 3 mons (if they want to).
This option should be only configured by an user if they have 5 or more zones in their cluster, otherwise it should be controlled by the operator and the appropriate value should be set, be it overriding the user's value.

We have validations in place that doesn't allow the user to set anything else apart from 3,5 for monCount field in the CR.

Nikhil-Ladha:
> Lets create a new func where we can set these kind of values. like `setDynamicDefaults`

Updated the code to have a separate function.

controllers/storagecluster/reconcile.go
```diff
@@ -886,3 +890,14 @@ func getUnsupportedClientsCount(r *StorageClusterReconciler, namespace string) (
 
 	return count, nil
 }
+
+// Set dynamic default values in the StorageCluster CR
+func setDynamicDefaults(sc *ocsv1.StorageCluster) {
+
+	// Set default mon count value in the StorageCluster CR
+	if arbiterEnabled(sc) && sc.Spec.ManagedResources.CephCluster.MonCount != defaults.ArbiterModeMonCount {
```
travisn, line 898: Whenever arbiter is enabled we must always set 5 mons, so I believe this simpler condition will ensure it.
```suggestion
	if arbiterEnabled(sc) {
```

controllers/storagecluster/reconcile.go
travisn: If they desire 5 mons, where is it validated that they have 5 zones?

controllers/storagecluster/reconcile.go
Nikhil-Ladha: The reason I added that extra condition was to avoid resetting the same value again for the next reconcile, a small benefit but worth it I think?

controllers/storagecluster/reconcile.go
Nikhil-Ladha: We don't validate it at ocs-operator level as of now, it's just the `CephMonLowNumber` alert that will be shown to the user on the UI, if they have 5 or more zones.

Nikhil-Ladha: Yeah, that was the plan of the epic for the 4.15 release.

controllers/storagecluster/reconcile.go
Nikhil-Ladha: Ok, got it. I will update the condition.

travisn (approved):
/approve

iamniting (approved):
/hold

iamniting:
@umangachapagain pls take a look and remove hold if looks good.

controllers/storagecluster/reconcile.go
```diff
@@ -886,3 +890,14 @@ func getUnsupportedClientsCount(r *StorageClusterReconciler, namespace string) (
...
+// Set dynamic default values in the StorageCluster CR
+func setDynamicDefaults(sc *ocsv1.StorageCluster) {
+
+	// Set default mon count value in the StorageCluster CR
+	if arbiterEnabled(sc) {
+		sc.Spec.ManagedResources.CephCluster.MonCount = defaults.ArbiterModeMonCount
+	} else if sc.Spec.ManagedResources.CephCluster.MonCount == 0 {
+		sc.Spec.ManagedResources.CephCluster.MonCount = defaults.DefaultMonCount
+	}
+}
```
umangachapagain, line 903: Looking at the implementation again, we already have a function to determine MON count here:
https://github.com/red-hat-storage/ocs-operator/blob/b0f1699d841c248f2ff868ed31e04b83a4708203/controllers/storagecluster/cephcluster.go#L717-L726
This function should be enough for our usecases. We do not need this new function to set Mon count on StorageCluster CR which is user defined.

@Nikhil-Ladha Is the feature not working without this change?

umangachapagain (changes requested):

controllers/storagecluster/reconcile.go
Nikhil-Ladha: > This function should be enough for our usecases. We do not need this new function to set Mon count on StorageCluster CR which is user defined.

Oh yes, seems like we can use this function here as well. Thanks for pointing it out, I will update the code to use this function.

> @Nikhil-Ladha Is the feature not working without this change?

The issue is in the UI, we are seeing a blank field in the UI when the alert is raised.
Shared the screenshot in this [comment](https://github.com/red-hat-storage/ocs-operator/pull/2362#issuecomment-1877091199)

umangachapagain: We can't fix that by updating StorageCluster spec. They need to read it from CephCluster.

Nikhil-Ladha: The reason we need this fix in StorageCluster spec is because the CephCluster spec is updated with the values we have in the StorageCluster CR.
So, when the alert is raised for CephLowMonNumber, we update that value in the StorageCluster CR, and not in the CephCluster CR directly. That's the reason fetching the inital value from CephCluster CR and later updating the StorageCluster CR doesn't seem like the best option here.

malayparida2000: Reading from CephCluster & Updating in StorageCluster may not be convenient for UI but IMO reading it from the CephCluster gives the correct number reliably and will avoid this complicated change.

bipuladh: @malayparida2000 that is not convenient in the UI as well.

umangachapagain: > The reason we need this fix in StorageCluster spec is because the CephCluster spec is updated with the values we have in the StorageCluster CR.
>
This is not always true. Mon count is determined internally unless it is explicitly defined in StorageCluster.
It is an optional field. For all optional fields in StorageCluster, operator decides the corresponding CephCluster value.

> So, when the alert is raised for CephLowMonNumber, we update that value in the StorageCluster CR, and not in the CephCluster CR directly. That's the reason fetching the inital value from CephCluster CR and later updating the StorageCluster CR doesn't seem like the best option here.
>
It may not be the best option, but it is the simplest option we have.
Operators should not edit their own CRs. So, determining Mon count internally and updating it on StorageCluster spec when required, is not an option here.

Alternatively, we can provide the current Mon count as `StorageCluster.Status.CurrentMonCount`. This means, when reconciling CephCluster CR, get the Mon count and write it to `StorageCluster.Status.CurrentMonCount`.

malayparida2000: Providing in the status would be defiently better than the operator editing it's own CR

Nikhil-Ladha: Ok, in that case let's set the value in the Status of the StorageCluster CR, and I think UI can pick the value from there?
cc @bipuladh this should be fine, right?

Nikhil-Ladha: Updated the code to set the value in the Status of StorageCluster CR, under the field `CurrentMonCount`.

controllers/storagecluster/cephcluster.go
```diff
@@ -343,6 +343,9 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
 		}
 	}
 
+	// Update the currentMonCount field in StoragCluster status from the cephCluster CR
+	sc.Status.CurrentMonCount = cephCluster.Spec.Mon.Count
```
umangachapagain, line 347: We should update this only for internal mode. For external mode, this would be undefined.

umangachapagain (changes requested):

controllers/storagecluster/cephcluster.go
Nikhil-Ladha: Ah, sure. Updated the code.

umangachapagain (approved):

Nikhil-Ladha:
@umangachapagain the hold label needs to be removed :)

PR #2366, opened 2024-01-04
added changes to oauth proxy config

rchikatw:
- added changes oauth proxy config to work UI integration smoothly
- removed unused code.


config/rbac/oauth_proxy_role.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: oauth-proxy-role
+rules:
+- apiGroups: ["authentication.k8s.io"]
+  resources:
+  - tokenreviews
+  verbs: ["create"]
```
leelavg, line 9: pls follow the shape of other rbac files, like
```
verbs:
 - create
```
same for the rest

deploy/ocs-operator/manifests/ocs-operator.clusterserviceversion.yaml
```diff
@@ -3293,11 +3306,12 @@ spec:
                 - -https-address=:8888
                 - -http-address=
                 - -email-domain=*
-                - -upstream=https://localhost:8080/onboarding-tokens
+                - -upstream=http://localhost:8080/onboarding-tokens
                 - -tls-cert=/etc/tls/private/tls.crt
                 - -tls-key=/etc/tls/private/tls.key
                 - -cookie-secret-file=/etc/proxy/secrets/session_secret
                 - -openshift-service-account=ux-backend-server
+                - -openshift-delegate-urls={"/":{"resource":"StorageCluster","namespace":"openshift-storage","verb":"get"}}
```
leelavg, line 3314: not `get`, use `create`

tools/csv-merger/csv-merger.go
```diff
@@ -999,11 +999,12 @@ func getUXBackendServerDeployment() appsv1.DeploymentSpec {
 						Args: []string{"-provider=openshift",
 							"-https-address=:8888",
 							"-http-address=", "-email-domain=*",
-							"-upstream=https://localhost:8080/onboarding-tokens",
+							"-upstream=http://localhost:8080/onboarding-tokens",
 							"-tls-cert=/etc/tls/private/tls.crt",
 							"-tls-key=/etc/tls/private/tls.key",
 							"-cookie-secret-file=/etc/proxy/secrets/session_secret",
 							"-openshift-service-account=ux-backend-server",
+							"-openshift-delegate-urls={\"/\":{\"resource\":\"StorageCluster\",\"namespace\":\"openshift-storage\",\"verb\":\"get\"}}",
```
leelavg, line 1007: you could use backticks (\`) rather than escaping quotes like, \`-openshift-....}\`

leelavg (changes requested):

config/rbac/oauth_proxy_role.yaml
rchikatw: if i use create here it's not working as expected i have tried and tested this.

services/ux-backend/main.go
```diff
@@ -40,6 +41,26 @@ func loadAndValidateServerConfig() (*serverConfig, error) {
 		return nil, fmt.Errorf("Malformed user-defined listening port:  %s. shutting down: %v", listenPortAsString, err)
 	}
 
+	runServerWithTLSAsString := os.Getenv("RUN_SERVER_WITH_TLS")
```
nb-ohad, line 44: Lets simplify the env var name 
```suggestion
	tlsEnabledAsString := os.Getenv("TLS_ENABLED")
```

hack/source-manifests.sh
```diff
@@ -69,7 +69,7 @@ function gen_ocs_csv() {
 	pushd config/manager
 	$KUSTOMIZE edit set image ocs-dev/ocs-operator="$OCS_IMAGE"
 	popd
-	$KUSTOMIZE build config/manifests/ocs-operator | $OPERATOR_SDK generate bundle -q --overwrite=false --output-dir deploy/ocs-operator --kustomize-dir config/manifests/ocs-operator --package ocs-operator --version "$CSV_VERSION"
+	$KUSTOMIZE build config/manifests/ocs-operator | $OPERATOR_SDK generate bundle -q --overwrite=false --output-dir deploy/ocs-operator --kustomize-dir config/manifests/ocs-operator --package ocs-operator --version "$CSV_VERSION" --extra-service-accounts=ux-backend-server
```
nb-ohad, line 72: This was removed in the original PR because it was stipulated to break OLM upgrades, have we tested this change with upgrades?

services/ux-backend/main.go
```diff
@@ -40,6 +41,26 @@ func loadAndValidateServerConfig() (*serverConfig, error) {
...
+	if runServerWithTLSAsString == "" {
+		klog.Infof("UX-Backend-server running in insecure mode")
+		config.runServerWithTLS = false
+	} else {
+		runServerWithTLS, err := strconv.ParseBool(runServerWithTLSAsString)
```
nb-ohad, line 49:
```suggestion
		config.tlsEnabled, err := strconv.ParseBool(tlsEnabledAsString)
```

services/ux-backend/main.go
```diff
@@ -40,6 +41,26 @@ func loadAndValidateServerConfig() (*serverConfig, error) {
...
+		if err != nil {
+			return nil, fmt.Errorf("Malformed user-defined run server with TLS:  %s. shutting down: %v", runServerWithTLSAsString, err)
+		}
+		if runServerWithTLS {
+			klog.Infof("UX-Backend-server running in secure mode %t", runServerWithTLS)
+			config.runServerWithTLS = true
+		} else {
+			klog.Infof("UX-Backend-server running in insecure mode")
+			config.runServerWithTLS = false
+		}
```
nb-ohad, line 60: These messages are misleading as they state that we are already running which is wrong. 
```suggestion
		if config.tlsEnabled {
			klog.Info("Server configured to run with TLS")
		} else {
			klog.Info("Server configured to run without TLS")
		}
```

services/ux-backend/main.go
```diff
@@ -58,11 +79,17 @@ func main() {
...
+		log.Fatal(http.ListenAndServeTLS(
+			fmt.Sprintf("%s%d", ":", config.listenPort),
+			"/etc/tls/private/tls.crt",
+			"/etc/tls/private/tls.key",
+			nil,
+		))
+	}
 
-	log.Fatal(http.ListenAndServeTLS(
+	log.Fatal(http.ListenAndServe(
```
nb-ohad, line 91: This should be in an `else`

nb-ohad (changes requested):

hack/source-manifests.sh
rchikatw: If we create sa yaml explicitly we don't have problem with upgrade so created separate yaml for sa

tools/csv-merger/csv-merger.go
```diff
@@ -999,11 +999,12 @@ func getUXBackendServerDeployment() appsv1.DeploymentSpec {
 						Args: []string{"-provider=openshift",
 							"-https-address=:8888",
 							"-http-address=", "-email-domain=*",
-							"-upstream=https://localhost:8080/onboarding-tokens",
+							"-upstream=http://localhost:8080/onboarding-tokens",
```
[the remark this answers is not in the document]
rchikatw: it works for both, Changing it to http://localhost:8080/

services/ux-backend/main.go
rchikatw: in an else I cannot write log.Fatel as it return function and if I do golint will cry

rchikatw: Basically i might get an error something like this (if block ends with call to log.Fatal function, so drop this else and outdent its block)

config/rbac/ux_backend_sa.yaml
```diff
@@ -1,4 +1,4 @@
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  name: ux-backend-server
+  name: ux-backend-server
```
leelavg, line 4: pls add new line

tools/csv-merger/csv-merger.go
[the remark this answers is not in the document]
leelavg: that's weird unless you already had a discussion and found out the reason.

services/ux-backend/main.go
nb-ohad: Ok. then seprate the fatal from the return value
```
	var err error
	addr := fmt.Sprintf("%s%d", ":", config.listenPort)
	if config.tlsEnabled {
		err := http.ListenAndServeTLS(addr,
			"/etc/tls/private/tls.crt",
			"/etc/tls/private/tls.key",
			nil,
		))
	} else {
		err := http.ListenAndServe(addr, nil)
	}
	log.Fatal(err)
```

PR #2369, opened 2024-01-04
UX Backend - Unified handling of env var based configuration

nb-ohad:
Common validation and default value handling for env vars configuration

Signed-off-by: nb-ohad <mitrani.ohad@gmail.com>

### DIFF
--- a/services/ux-backend/main.go
+++ b/services/ux-backend/main.go
@@ -18,37 +18,34 @@ type serverConfig struct {
 	tlsEnabled           bool
 }
 
+func readEnvVar[T any](envVarName string, defaultValue T, parser func(str string) (T, error)) (T, error) {
+	if str := os.Getenv(envVarName); str == "" {
+		klog.Infof("no user-defined %s provided, defaulting to %v", envVarName, defaultValue)
+		return defaultValue, nil
+	} else if value, err := parser(str); err != nil {
+		return *new(T), fmt.Errorf("malformed user-defined %s value %s: %v", envVarName, str, err)
+	} else {
+		return value, nil
+	}
+}
+
 func loadAndValidateServerConfig() (*serverConfig, error) {
 	var config serverConfig
-
 	var err error
-	defaultTokenLifetimeInHours := 48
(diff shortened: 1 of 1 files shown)


nb-ohad:
@rchikatw Can you please review?

services/ux-backend/main.go
```diff
@@ -18,37 +18,34 @@ type serverConfig struct {
 	tlsEnabled           bool
 }
 
+func readEnvVar[T any](envVarName string, defaultValue T, parser func(str string) (T, error)) (T, error) {
+	if str := os.Getenv(envVarName); str == "" {
+		klog.Infof("no user-defined %s provided, defaulting to %v", envVarName, defaultValue)
+		return defaultValue, nil
+	} else if value, err := parser(str); err != nil {
+		return *new(T), fmt.Errorf("malformed user-defined %s value %v: %v", envVarName, value, err)
```
rchikatw, line 26: In case of an error the value always will return 0, so let's not print the value in Errorf.

nb-ohad: 1. It will not return 0 it will return the default value of the type: `0` for int, `false` for bool, `nil` for pointers, etc.
2. There is a bug here. I should point `str` and `value` because I want to show the faulty string. I will fix it.

services/ux-backend/main.go
```diff
@@ -18,37 +18,34 @@ type serverConfig struct {
 	tlsEnabled           bool
 }
 
+func readEnvVar[T any](envVarName string, defaultValue T, parser func(str string) (T, error)) (T, error) {
+	if str := os.Getenv(envVarName); str == "" {
+		klog.Infof("no user-defined %s provided, defaulting to %v", envVarName, defaultValue)
+		return defaultValue, nil
+	} else if value, err := parser(str); err != nil {
+		return *new(T), fmt.Errorf("malformed user-defined %s value %s: %v", envVarName, str, err)
```
leelavg, line 26: I believe you are returning a double pointer here 🤔

leelavg (approved):
/lgtm

PR #2380, opened 2024-01-09
remove rbd-mirror metrics already exported by ceph

weirdwiz:
The PR removes the metrics
```
ocs_rbd_mirror_image_primary_snapshot_timestamp
ocs_rbd_mirror_image_secondary_snapshot_timestamp
ocs_rbd_mirror_image_image_bytes
ocs_rbd_mirror_image_image_snapshot_bytes
```

as corresponding metrics already exist in ceph
```
ceph_rbd_mirror_snapshot_image_local_timestamp
ceph_rbd_mirror_snapshot_image_remote_timestamp
ceph_rbd_mirror_snapshot_image_last_sync_bytes
```

`image_bytes` and snapshot bytes can be calculated by `sync_time` and `sync_bytes` metrics exported by ceph
```
ocs_rbd_mirror_image_image_bytes :  ceph_rbd_mirror_snapshot_image_last_sync_bytes / ceph_rbd_mirror_snapshot_image_last_sync_time
```

### DIFF
--- a/metrics/internal/collectors/rbd-mirror.go
+++ b/metrics/internal/collectors/rbd-mirror.go
@@ -1,7 +1,6 @@
 package collectors
 
 import (
-	"encoding/json"
 	"strings"
 
 	"github.com/prometheus/client_golang/prometheus"
@@ -29,12 +28,8 @@ type RBDMirrorCollector struct {
 	RBDMirrorStore        *internalcache.RBDMirrorStore
 	PersistentVolumeStore *internalcache.PersistentVolumeStore
 	// Metric descriptors
-	MirrorDaemonHealth         *prometheus.Desc
-	ImageStatusState           *prometheus.Desc
-	PrimarySnapshotTimestamp   *prometheus.Desc
-	SecondarySnapshotTimestamp *prometheus.Desc
-	ImageBytes                 *prometheus.Desc
-	ImageSnapshotBytes         *prometheus.Desc
+	MirrorDaemonHealth *prometheus.Desc
+	ImageStatusState   *prometheus.Desc
 }
 
 func NewRBDMirrorCollector(mirrorStore *internalcache.RBDMirrorStore, pvStore *internalcache.PersistentVolumeStore) *RBDMirrorCollector {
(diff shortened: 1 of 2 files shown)


weirdwiz:
ptal @umangachapagain

metrics/internal/collectors/rbd-mirror.go
```diff
@@ -177,25 +145,6 @@ func (c *RBDMirrorCollector) Collect(ch chan<- prometheus.Metric) {
 						klog.Errorf("Failed to unmarshal description of image %q from site %q: %v", image.Name, site.SiteName, err)
 						continue
 					}
-					// LocalSnapshotTimestamp could be unavailable for a while during image resync.
```
umangachapagain, line 180: We can remove the entire if block and delete `internalcache.RBDMirrorPeerSiteDescription{}` structure in cache since we don't use that data anymore. Will save us some space.

https://github.com/red-hat-storage/ocs-operator/blob/68d03b12968375eb97651bc4b74cdf9b102c74ef/metrics/internal/collectors/rbd-mirror.go#L134-L149

metrics/internal/collectors/rbd-mirror.go
```diff
@@ -87,10 +59,6 @@ func (c *RBDMirrorCollector) Describe(ch chan<- *prometheus.Desc) {
 	ds := []*prometheus.Desc{
 		c.MirrorDaemonHealth,
 		c.ImageStatusState,
```
umangachapagain, line 61: Are these metrics not exported by Ceph?

umangachapagain (changes requested):

umangachapagain (changes requested):
- PR description says 
  > image_bytes and snapshot bytes can be calculated by sync_time and sync_bytes 
  metrics exported by ceph
  >
  Can you add example of how they can be calculated?

- We also need to replace the usage of these metrics in alerts.

metrics/internal/collectors/rbd-mirror.go
weirdwiz: No, these metrics are not exported as metrics. Just in the status, as the status is calculated in a different layer than the daemon.

metrics/internal/collectors/rbd-mirror.go
weirdwiz: done

weirdwiz:
> We also need to replace the usage of these metrics in alerts.
all the alerts are based on image state,  that metric is not removed from the exporter

umangachapagain (approved):

umangachapagain (commented):
@weirdwiz DCO check is failing. Please address it.

weirdwiz:
> @weirdwiz DCO check is failing. Please address it.

need to squash the commits, on it

PR #2301, opened 2023-12-05
Token generation for prov/consumer mode

rchikatw:
Creation of a job that is responsible for creating the private and public keys as secret in provider/consumer mode.
- Added watch on a public secret if it's deleted storage cluster will reconcile
- Job is created based on public key,

PR #2385, opened 2024-01-12
fixed re-trigger job issue when public key deleted

rchikatw:
- fixed re-trigger job issue when the public key deleted
- formatted the logger in a proper format

### DIFF
--- a/controllers/storagecluster/provider_server.go
+++ b/controllers/storagecluster/provider_server.go
@@ -16,6 +16,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
@@ -454,6 +455,8 @@ func getOnboardingJobObject(instance *ocsv1.StorageCluster) *batchv1.Job {
 			Namespace: instance.Namespace,
 		},
 		Spec: batchv1.JobSpec{
+			// Eligible to delete automatically when job finishes
+			TTLSecondsAfterFinished: ptr.To(int32(0)),
 			Template: corev1.PodTemplateSpec{
 				Spec: corev1.PodSpec{
 					RestartPolicy:      corev1.RestartPolicyOnFailure,
--- a/controllers/storagecluster/storagecluster_controller.go
+++ b/controllers/storagecluster/storagecluster_controller.go
@@ -164,14 +164,6 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		},
 	}
 
-	onboardingSecretPredicates := builder.WithPredicates(
-		predicate.NewPredicateFuncs(
-			func(client client.Object) bool {
-				return client.GetName() == onboardingTicketPublicKeySecretName
-			},
-		),
-	)
-
 	builder := ctrl.NewControllerManagedBy(mgr).
 		For(&ocsv1.StorageCluster{}, builder.WithPredicates(scPredicate)).
 		Owns(&cephv1.CephCluster{}).
@@ -179,6 +171,7 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&appsv1.Deployment{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
 		Owns(&corev1.Service{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
 		Owns(&corev1.ConfigMap{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		Owns(&corev1.Secret{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
 		Watches(&ocsv1.StorageProfile{}, enqueueStorageClusterRequest).
 		Watches(
 			&extv1.CustomResourceDefinition{
@@ -188,7 +181,6 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			},
 			enqueueStorageClusterRequest,
 		).
-		Watches(&corev1.Secret{}, enqueueStorageClusterRequest, onboardingSecretPredicates).
 		Watches(&ocsv1alpha1.StorageConsumer{}, enqueueStorageClusterRequest, builder.WithPredicates(ocsClientOperatorVersionPredicate))
 	if os.Getenv("SKIP_NOOBAA_CRD_WATCH") != "true" {
 		builder.Owns(&nbv1.NooBaa{})
--- a/onboarding/main.go
+++ b/onboarding/main.go
@@ -4,7 +4,9 @@ import (
 	"crypto/rand"
 	"crypto/rsa"
 	"crypto/x509"
+	"encoding/json"
 	"encoding/pem"
+	"fmt"
 	"os"
 
 	"github.com/red-hat-storage/ocs-operator/v4/controllers/util"
@@ -18,80 +20,115 @@ import (
 )
 
 const (
-	onboardingTicketPublicKeySecretName = "onboarding-ticket-key" //Name of existing public key which is used ocs-operator
-	onboardingPrivateKeySecretName      = "onboarding-private-key"
-	serviceAccountName                  = "onboarding-secret-generator"
+	// Name of existing public key which is used ocs-operator
+	onboardingValidationPublicKeySecretName  = "onboarding-ticket-key"
+	onboardingValidationPrivateKeySecretName = "onboarding-private-key"
+	storageClusterName                       = "ocs-storagecluster"
 )
 
 func main() {
 	clientset, err := newClient()
 	if err != nil {
-		klog.Error(err, "failed to create controller-runtime client")
-		return
-	}
-
-	operatorNamespace, err := util.GetOperatorNamespace()
-	if err != nil {
-		klog.Error(err, "unable to get operator namespace")
+		klog.Errorf("failed to create clientset: %v", err)
 		os.Exit(1)
 	}
 
-	// 1. Check public key secret exist or not
-	_, err = clientset.CoreV1().Secrets(operatorNamespace).Get(context.TODO(), onboardingTicketPublicKeySecretName, metav1.GetOptions{})
+	ctx := context.Background()
+	operatorNamespace, err := util.GetOperatorNamespace()
+	if err != nil {
+		klog.Errorf("unable to get operator namespace: %v", err)
+		os.Exit(1)
+	}
 
-	if err != nil && kerrors.IsNotFound(err) {
-		// Generate RSA key.
-		var err error
-		privateKey, err := rsa.GenerateKey(rand.Reader, 4096)
-		if err != nil {
-			klog.Error(err, "unable to generate private")
-			os.Exit(1)
-		}
+	// Generate RSA key.
+	privateKey, err := rsa.GenerateKey(rand.Reader, 4096)
+	if err != nil {
+		klog.Errorf("unable to generate private: %v", err)
+		os.Exit(1)
+	}
 
-		publicKey := &privateKey.PublicKey
-		// Export the keys to pem string
-		privatePem := convertRsaPrivateKeyAsPemStr(privateKey)
-		publicPem, err := convertRsaPublicKeyAsPemStr(publicKey)
+	publicKey := &privateKey.PublicKey
+	// Export the keys to pem string
+	privatePem := convertRsaPrivateKeyAsPemStr(privateKey)
+	publicPem := convertRsaPublicKeyAsPemStr(publicKey)
 
-		if err != nil {
-			klog.Error(err, "failed to convert public key to pem str")
-			os.Exit(1)
-		}
+	// In situations where there is a risk of one secret being updated and potentially
+	// failing to update another, it is recommended not to rely solely on clientset update mechanisms.
+	// Instead, a safer approach is to delete both secrets and then recreate them simultaneously
+	// to ensure consistency and accuracy of all secrets. By this way it will be easier to diagnose the
+	// issues if one or two secrets do not exist instead of trying to understand if they match
+	err = clientset.CoreV1().
+		Secrets(operatorNamespace).
+		Delete(ctx, onboardingValidationPrivateKeySecretName, metav1.DeleteOptions{})
+	if err != nil && !kerrors.IsNotFound(err) {
+		klog.Errorf("failed to delete private secret: %v", err)
+		os.Exit(1)
+	}
 
-		privateSecret := &corev1.Secret{
-			ObjectMeta: metav1.ObjectMeta{
-				Name:        onboardingPrivateKeySecretName,
-				Namespace:   operatorNamespace,
-				Annotations: map[string]string{"kubernetes.io/service-account.name": serviceAccountName},
-			},
-			Type: "kubernetes.io/service-account-token",
-			StringData: map[string]string{
-				"key": privatePem,
-			},
-		}
+	// Delete public key secret
+	err = clientset.CoreV1().
+		Secrets(operatorNamespace).
+		Delete(ctx, onboardingValidationPublicKeySecretName, metav1.DeleteOptions{})
+	if err != nil && !kerrors.IsNotFound(err) {
+		klog.Errorf("failed to delete public secret: %v", err)
+		os.Exit(1)
+	}
 
-		_, err = clientset.CoreV1().Secrets(operatorNamespace).Create(context.Background(), privateSecret, metav1.CreateOptions{})
+	storageClusterMetadata, err := getStorageClusterMetadata(ctx, operatorNamespace, clientset)
+	if err != nil {
+		klog.Errorf("failed to get storage cluster metadata: %v", err)
+		os.Exit(1)
+	}
 
-		if err != nil {
-			klog.Error(err, "Failed to create private secret.")
-			os.Exit(1)
-		}
-		publicSecret := &corev1.Secret{
-			ObjectMeta: metav1.ObjectMeta{
-				Name:      onboardingTicketPublicKeySecretName,
-				Namespace: operatorNamespace,
-			},
-			StringData: map[string]string{
-				"key": publicPem,
-			},
-		}
+	privateSecret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      onboardingValidationPrivateKeySecretName,
+			Namespace: operatorNamespace,
+			OwnerReferences: []metav1.OwnerReference{
+				{
+					UID:        storageClusterMetadata.UID,
+					APIVersion: storageClusterMetadata.APIVersion,
+					Kind:       storageClusterMetadata.Kind,
+					Name:       storageClusterMetadata.Name,
+				},
+			}},
+		StringData: map[string]string{
+			"key": privatePem,
+		},
+	}
 
-		_, err = clientset.CoreV1().Secrets(operatorNamespace).Create(context.Background(), publicSecret, metav1.CreateOptions{})
-		if err != nil {
-			klog.Error(err, "Failed to create public secret.")
-			os.Exit(1)
-		}
+	_, err = clientset.CoreV1().
+		Secrets(operatorNamespace).
+		Create(ctx, privateSecret, metav1.CreateOptions{})
 
+	if err != nil {
+		klog.Errorf("failed to create private secret: %v", err)
+		os.Exit(1)
+	}
+
+	publicSecret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      onboardingValidationPublicKeySecretName,
+			Namespace: operatorNamespace,
+			OwnerReferences: []metav1.OwnerReference{
+				{
+					UID:        storageClusterMetadata.UID,
+					APIVersion: storageClusterMetadata.APIVersion,
+					Kind:       storageClusterMetadata.Kind,
+					Name:       storageClusterMetadata.Name,
+				},
+			}},
+		StringData: map[string]string{
+			"key": publicPem,
+		},
+	}
+
+	_, err = clientset.CoreV1().
+		Secrets(operatorNamespace).
+		Create(ctx, publicSecret, metav1.CreateOptions{})
+	if err != nil {
+		klog.Errorf("failed to create public secret: %v", err)
+		os.Exit(1)
 	}
 
 }
@@ -99,9 +136,8 @@ func main() {
 func newClient() (*kubernetes.Clientset, error) {
 	config := runtime.GetConfigOrDie()
 	clientset, err := kubernetes.NewForConfig(config)
-
 	if err != nil {
-		klog.Error(err, "failed to get clientset")
+		return nil, err
 	}
 
 	return clientset, nil
@@ -113,12 +149,27 @@ func convertRsaPrivateKeyAsPemStr(privateKey *rsa.PrivateKey) string {
 	return string(privateKeyPem)
 }
 
-func convertRsaPublicKeyAsPemStr(publicKey *rsa.PublicKey) (string, error) {
-	publicKeyBytes, err := x509.MarshalPKIXPublicKey(publicKey)
-	if err != nil {
-		return "", err
-	}
+func convertRsaPublicKeyAsPemStr(publicKey *rsa.PublicKey) string {
+	publicKeyBytes := x509.MarshalPKCS1PublicKey(publicKey)
 	publicKeyPem := pem.EncodeToMemory(&pem.Block{Type: "PUBLIC KEY", Bytes: publicKeyBytes})
+	return string(publicKeyPem)
+}
 
-	return string(publicKeyPem), nil
+func getStorageClusterMetadata(ctx context.Context, operatorNamespace string, clientset *kubernetes.Clientset) (*metav1.PartialObjectMetadata, error) {
+	var storageClusterMetadata metav1.PartialObjectMetadata
+	storageClusterGVKPath := fmt.Sprintf(
+		"/apis/ocs.openshift.io/v1/namespaces/%s/storageclusters/%s",
+		operatorNamespace,
+		storageClusterName,
+	)
+	storageClusterMetadataJSON, err := clientset.RESTClient().Get().AbsPath(storageClusterGVKPath).Do(ctx).Raw()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get storage cluster metadata: %v", err)
+	}
+
+	if err = json.Unmarshal(storageClusterMetadataJSON, &storageClusterMetadata); err != nil {
+		return nil, fmt.Errorf("failed to parse storage cluster metadata response: %v", err)
+	}
+
+	return &storageClusterMetadata, nil
 }
--- a/deploy/ocs-operator/manifests/onboarding-secret-generator-role.yaml
+++ b/deploy/ocs-operator/manifests/onboarding-secret-generator-role.yaml
@@ -11,3 +11,11 @@ rules:
   - get
   - list
   - create
+  - delete
+- apiGroups:
+  - ocs.openshift.io
+  resources:
+  - storageclusters
+  verbs:
+    - get
+    - list
--- a/hack/ticketgen/ticketcheck.go
+++ b/hack/ticketgen/ticketcheck.go
@@ -30,11 +30,10 @@ func main() {
 		panic(err)
 	}
 	pemBlock, _ := pem.Decode(keyfile)
-	key, err := x509.ParsePKIXPublicKey(pemBlock.Bytes)
+	pubKey, err := x509.ParsePKCS1PublicKey(pemBlock.Bytes)
 	if err != nil {
 		panic(err)
 	}
-	pubKey := key.(*rsa.PublicKey)
 
 	fmt.Printf("Reading ticket from: %s\n", *ticketfileStr)
 	ticketData, err := os.ReadFile(*ticketfileStr)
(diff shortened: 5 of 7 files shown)


controllers/storagecluster/provider_server.go
```diff
@@ -448,12 +448,14 @@ func RandomString(l int) string {
...
 func getOnboardingJobObject(instance *ocsv1.StorageCluster) *batchv1.Job {
 
+	ptrToZero := int32(0)
 	return &batchv1.Job{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      onboardingJobName,
 			Namespace: instance.Namespace,
 		},
 		Spec: batchv1.JobSpec{
+			TTLSecondsAfterFinished: &ptrToZero, //Eligible to delete automatically when job completes
```
nb-ohad, line 458: No need to have a local variable. 
This is a common case and I think we had a util for it. If not please add the following util (in the proper file in the utils package)

```
func ToPointer[T any](value T) *T {
    return &T
}
``` 

This will allow you to do
```suggestion
			TTLSecondsAfterFinished: utils.ToPointer(0)
```

onboarding/main.go
```diff
@@ -70,6 +70,8 @@ func main() {
 			},
 		}
 
+		clientset.CoreV1().Secrets(operatorNamespace).Delete(context.Background(), onboardingPrivateKeySecretName, metav1.DeleteOptions{})
```
nb-ohad, line 73: Why do we need this? what is the reason? and why outside the reconcile function?

nb-ohad (changes requested):

onboarding/main.go
nb-ohad: That does not make sense. if the public key is deleted then a reconcile will happen and both keys will be regenerated. I don't see the edge case here

rchikatw: Okay, let me describe it more clearly.
- cluster is up and running and it has both public & private key
- Now if the admin deletes the public key by mistake then what should happen?
- It will reconcile but when we call clientset.CoreV1().Secrets(operatorNamespace).Create on private key it will return an error saying the secret already exists.

nb-ohad: I might be missing something here, but why do you want to call `clientset.CoreV1().Secrets(operatorNamespace).Create` on the secret? 

First, why are you working with clientset and not with a kubebuilder client? 
Second, in case the secret exists you should call `Update` and not `Create. The kube util library has a convenient `CreateOrUpdate` function that does exactly that!  (with the benefit of not actually calling the update if nothing changed)

rchikatw: I have not used the Kubebuilder client because Leela suggested it's a small job bringing the heavyweight controller runtime is not good. Leela mentioned [here](https://github.com/red-hat-storage/ocs-operator/pull/2301#discussion_r1422415217). 
Yes, I agree instead of deleting the secret, I have to update if the secret already exists.

nb-ohad: Ok. But regardless of the client, you will want to replicate the same pattern `CreateOrUpdate` implements

rchikatw: I have addressed the comments, please resolve them?

tools/csv-merger/csv-merger.go
```diff
@@ -1004,8 +1002,9 @@ func getUXBackendServerDeployment() appsv1.DeploymentSpec {
 							"-tls-key=/etc/tls/private/tls.key",
 							"-cookie-secret-file=/etc/proxy/secrets/session_secret",
 							"-openshift-service-account=ux-backend-server",
-							`-openshift-delegate-urls={"/":{"resource":"StorageCluster","namespace":"openshift-storage","verb":"create"}}`,
-							"-openshift-ca=/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"},
+							`-openshift-delegate-urls={"/":{"group":"ocs.openshift.io","resource":"storageclusters","namespace":"openshift-storage","verb":"create"}}`,
+							"-openshift-ca=/var/run/secrets/kubernetes.io/serviceaccount/ca.crt",
+						},
```
jarrpa, line 1007: Is this somehow related to the problem you're solving? If so please add some reasoning to your commit message.

jarrpa (commented):
I made a few comments mostly asking for clarity on a few things. I'd also like to see anything that's not directly related to the problem solution (like the util function and the csv-merger changes) in a separate commit, but if that's not common practice in this part of the project I'm all right with leaving it as one commit.

controllers/util/util.go
```diff
@@ -22,3 +22,7 @@ func DetectDuplicateInStringSlice(slice []string) bool {
 	}
 	return false
 }
+
+func ToPointer[T any](value T) *T {
+	return &value
+}
```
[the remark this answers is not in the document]
jarrpa: Note that, if you end up going forward with it, I intentionally suggested those names so our IDEs can more reliably show autocomplete options for the boolean values. :stuck_out_tongue_winking_eye: Though if you come up with something easier to use, by all means go for it.

controllers/storagecluster/provider_server.go
```diff
@@ -454,6 +454,7 @@ func getOnboardingJobObject(instance *ocsv1.StorageCluster) *batchv1.Job {
 			Namespace: instance.Namespace,
 		},
 		Spec: batchv1.JobSpec{
+			TTLSecondsAfterFinished: util.ToPointer[int32](0), //Eligible to delete automatically when job completes
```
leelavg, line 457: - move comment to line above and not inline
- TTL works for both Job Complete & Failed states, either change the comment to reflect that or just use `job finishes`

leelavg, line 457: Isn't `util.ToPointer(int32(0))` more safer, like ide may throw error w/ `bool(0)` but not for `[bool](0)` 🤔. I believe compiler throws error for both ways though. Pls check, can stay as it is if it's not creating any issue.

onboarding/main.go
```diff
@@ -10,7 +10,7 @@ import (
 	"github.com/red-hat-storage/ocs-operator/v4/controllers/util"
 	"golang.org/x/net/context"
 	corev1 "k8s.io/api/core/v1"
-	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/errors"
```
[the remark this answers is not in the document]
rchikatw: There is no solid reason, I can undo this.

controllers/storagecluster/provider_server.go
rchikatw: i have checked `[bool](0)` compiler is not throwing error for this. We can stay with this

controllers/util/util.go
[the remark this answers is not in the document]
rchikatw: This is the one way of handling and I feel this is unified. No need to create separate functions to just have an autocomplete option in IDE. @nb-ohad Correct me if i am wrong and if you are interested in creating the function in the same way as Jose explained. Let me know

tools/csv-merger/csv-merger.go
rchikatw: There is new BZ raised for this particular issue now and handled in separate PR added the description for it as well. https://github.com/red-hat-storage/ocs-operator/pull/2397

onboarding/main.go
```diff
@@ -18,80 +18,95 @@ import (
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        onboardingPrivateKeySecretName,
+			Namespace:   operatorNamespace,
+			Annotations: map[string]string{"kubernetes.io/service-account.name": serviceAccountName},
+		},
+		Type: "kubernetes.io/service-account-token",
+		StringData: map[string]string{
+			"key": privatePem,
+		},
+	}
```
nb-ohad, line 87: Why do we need the `annotations` and `type` fields?

nb-ohad (changes requested):

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -166,13 +166,12 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 	}
 
-	onboardingSecretPredicates := builder.WithPredicates(
-		predicate.NewPredicateFuncs(
-			func(client client.Object) bool {
-				return client.GetName() == onboardingTicketPublicKeySecretName
-			},
-		),
-	)
+	secretPredicate := predicate.Funcs{
```
leelavg, line 169: pls set other predicates to false (CreateFunc, UpdateFunc)

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -166,13 +166,12 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			// Evaluates to false if the object has been confirmed deleted.
+			return !e.DeleteStateUnknown
```
leelavg, line 172: I do not understand what this field is actually used for even after going through the docs and you don't want to invoke reconcile for every secret deletion (well, nothing usually deletes secrets again & again).

I'd suggest just check the name, `return client.GetName() == onboardingTicketPublicKeySecretName`

onboarding/main.go
```diff
@@ -18,80 +18,95 @@ import (
...
+		os.Exit(1)
+	}
+
+	privateSecret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        onboardingPrivateKeySecretName,
+			Namespace:   operatorNamespace,
+			Annotations: map[string]string{"kubernetes.io/service-account.name": serviceAccountName},
+		},
+		Type: "kubernetes.io/service-account-token",
```
leelavg, line 83: this is a opaque secret, remove `Type` field.

controllers/util/util.go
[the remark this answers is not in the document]
nb-ohad: @rchikatw Leela is correct, I am against adding constants at this point

onboarding/main.go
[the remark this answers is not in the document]
nb-ohad: This is a common pattern for importing k8s errors. We do it in all of our repos and you can even see it in other community operators. Let's leave it as `kerrors`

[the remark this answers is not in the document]
rchikatw: its already undone the changes, Its now kerrors only

controllers/storagecluster/provider_server.go
```diff
@@ -447,13 +447,15 @@ func RandomString(l int) string {
...
-
+	ttlSecondsAfterFinished := int32(0)
 	return &batchv1.Job{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      onboardingJobName,
 			Namespace: instance.Namespace,
 		},
 		Spec: batchv1.JobSpec{
+			// Eligible to delete automatically when job finishes
+			TTLSecondsAfterFinished: &ttlSecondsAfterFinished,
```
nb-ohad, line 458: Please use `utils.ToPointer(0)` instead

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -166,13 +166,17 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 	}
 
-	onboardingSecretPredicates := builder.WithPredicates(
-		predicate.NewPredicateFuncs(
-			func(client client.Object) bool {
-				return client.GetName() == onboardingTicketPublicKeySecretName
-			},
-		),
-	)
+	secretPredicates := predicate.Funcs{
```
nb-ohad, line 169: No reason to have a struct that is not wrapped in a predicate. 

```suggestion
	secretPredicates := builder. WithPredicates(predicate.Funcs{
```

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -166,13 +166,17 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			return e.Object.GetName() == onboardingTicketPublicKeySecretName
+		},
+		CreateFunc: func(e event.CreateEvent) bool {
+			return false
```
nb-ohad, line 174: This is wrong as it will prevent create events on all watched secrets not just the onboarding public key secret

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -166,13 +166,17 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			return false
```
nb-ohad, line 177: This is wrong as it will prevent update events on all watched secrets not just the onboarding public key secret

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -166,13 +166,17 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-	onboardingSecretPredicates := builder.WithPredicates(
-		predicate.NewPredicateFuncs(
-			func(client client.Object) bool {
-				return client.GetName() == onboardingTicketPublicKeySecretName
-			},
-		),
-	)
+	secretPredicates := predicate.Funcs{
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			return e.Object.GetName() == onboardingTicketPublicKeySecretName
```
nb-ohad, line 171: This is wrong as it will prevent delete events on all watched secrets not just the onboarding public key secret
You just need to return `true` here

nb-ohad (changes requested):

nb-ohad:
@rchikatw why aren't we setting a controller reference on the secrets and using an `Own` watch? 
If we don't do it we need to make sure to delete the secrets ourselves!

onboarding/main.go
```diff
@@ -18,80 +21,131 @@ import (
...
+	privatePem := convertRsaPrivateKeyAsPemStr(privateKey)
+	publicPem, err := convertRsaPublicKeyAsPemStr(publicKey)
+
+	if err != nil {
+		klog.Errorf("failed to convert public key to pem str: %v", err)
+		os.Exit(1)
+	}
+
+	sc := ocsv1.StorageClusterList{}
+	instance, err := clientset.RESTClient().Get().AbsPath(
```
leelavg, line 60: Initially we were only using kube native resources (secrets) and so was the comment that client-go is sufficient. 

It is evident now that runtime client abstraction is helpful and so you should be able to freely use it.

nb-ohad: I agree, but please do not do the change now,  we can move back to runtime client in a later PR

onboarding/main.go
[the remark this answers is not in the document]
rchikatw: It's been discussed over DM and we agreed to do it this in separate PR.

onboarding/main.go
rchikatw: I agree.

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -166,13 +166,17 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 	}
 
-	onboardingSecretPredicates := builder.WithPredicates(
-		predicate.NewPredicateFuncs(
-			func(client client.Object) bool {
-				return client.GetName() == onboardingTicketPublicKeySecretName
-			},
-		),
-	)
+	secretPredicates := builder.WithPredicates(predicate.Funcs{
```
leelavg, line 169: why to use `builder.WithPredicates` here itself? better to follow exisitng pattern, this here locks the predicate to be only one iiuc. not mandatory if you aren't going to refresh.

leelavg (approved):

controllers/storagecluster/storagecluster_controller.go
rchikatw: I have added this because ohad said this here, https://github.com/red-hat-storage/ocs-operator/pull/2385#discussion_r1466285752

nb-ohad:
> I made a few comments mostly asking for clarity on a few things. I'd also like to see anything that's not directly related to the problem solution (like the util function and the csv-merger changes) in a separate commit, but if that's not common practice in this part of the project I'm all right with leaving it as one commit.

@jarrpa Regarding the separation into separate PRs. From my point of view, a PR needs to be a complete contribution not to be dependent on some code that needs to be reviewed somewhere else.

Because of that, I would claim that a separate PR is an overkill but I do agree that it makes sense to separate these parts into different commits under the same PR, similar to what I ask when we have vendor changes.

controllers/storagecluster/storagecluster_controller.go
nb-ohad: @rchikatw I was wrong. the `WithPredicates` takes a list of `predicate.Funcs` structure instances with that new knowledge in mind I agree with @leelavg

onboarding/main.go
```diff
@@ -18,80 +21,131 @@ import (
...
+			OwnerReferences: []metav1.OwnerReference{ownerRef},
+		},
+		StringData: map[string]string{
+			"key": publicPem,
+		},
+	}
 
+	_, err = clientset.CoreV1().
+		Secrets(operatorNamespace).
+		Create(context.Background(), publicSecret, metav1.CreateOptions{})
```
nb-ohad, line 145: see previous comment about context

controllers/util/util.go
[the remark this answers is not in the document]
nb-ohad: @jarrpa Can we resolve this?

onboarding/main.go
```diff
@@ -122,3 +158,25 @@ func convertRsaPublicKeyAsPemStr(publicKey *rsa.PublicKey) (string, error) {
...
 	return string(publicKeyPem), nil
 }
+
+func getStorageClusterMetadata(ctx context.Context, operatorNamespace string, clientset *kubernetes.Clientset) metav1.PartialObjectMetadata {
+	var storageClusterMetaDataList metav1.PartialObjectMetadataList
+	storageClusterGVKPath := fmt.Sprintf("/apis/%s/%s/namespaces/%s/%s", "ocs.openshift.io", "v1", operatorNamespace, "storageclusters")
+	instance, err := clientset.RESTClient().Get().AbsPath(storageClusterGVKPath).DoRaw(ctx)
+	if err != nil {
+		klog.Errorf("failed to get storage cluster metadata: %v", err.Error())
+		os.Exit(1)
```
nb-ohad, line 168: Please do not use `os.Exit` inside nested functions. 
Return an error to the caller and act on it on the call site (upper-level main).

onboarding/main.go
```diff
@@ -122,3 +158,25 @@ func convertRsaPublicKeyAsPemStr(publicKey *rsa.PublicKey) (string, error) {
...
+	}
+
+	if err := json.Unmarshal(instance, &storageClusterMetaDataList); err != nil {
+		klog.Errorf("failed to parse storage cluster metadata response: %v", err)
+		os.Exit(1)
```
nb-ohad, line 173: Same as prev comment

onboarding/main.go
```diff
@@ -122,3 +158,25 @@ func convertRsaPublicKeyAsPemStr(publicKey *rsa.PublicKey) (string, error) {
...
+	}
+
+	if len(storageClusterMetaDataList.Items) == 0 {
+		klog.Error("no storage cluster found")
+		os.Exit(1)
```
nb-ohad, line 178: Same as prev comment

onboarding/main.go
```diff
@@ -122,3 +158,25 @@ func convertRsaPublicKeyAsPemStr(publicKey *rsa.PublicKey) (string, error) {
 
 	return string(publicKeyPem), nil
 }
+
+func getStorageClusterMetadata(ctx context.Context, operatorNamespace string, clientset *kubernetes.Clientset) metav1.PartialObjectMetadata {
+	var storageClusterMetaDataList metav1.PartialObjectMetadataList
+	storageClusterGVKPath := fmt.Sprintf("/apis/%s/%s/namespaces/%s/%s", "ocs.openshift.io", "v1", operatorNamespace, "storageclusters")
```
nb-ohad, line 164: Why are we listing all of the storage clusters? we are part of the ocs operator code base and we should have access to the constant that have the hardcoded name of the relevant storage cluster

nb-ohad (changes requested):

onboarding/main.go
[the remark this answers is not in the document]
nb-ohad: @rchikatw  We need a new package name. I agreed to move the package to another place in a separate PR but not keep the existing name

controllers/util/util.go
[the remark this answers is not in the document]
jarrpa: Sure, I'll submit a follow-up PR to fix it.

onboarding/main.go
rchikatw: There is no constant i could find these in the project so i have added these.

onboarding/main.go
```diff
@@ -122,3 +163,23 @@ func convertRsaPublicKeyAsPemStr(publicKey *rsa.PublicKey) (string, error) {
...
 	return string(publicKeyPem), nil
 }
+
+func getStorageClusterMetadata(ctx context.Context, operatorNamespace string, clientset *kubernetes.Clientset) (metav1.PartialObjectMetadata, error) {
+	var storageClusterMetadataList metav1.PartialObjectMetadataList
+	storageClusterGVKPath := fmt.Sprintf("/apis/%s/%s/namespaces/%s/%s", "ocs.openshift.io", "v1", operatorNamespace, "storageclusters")
+	instance, err := clientset.RESTClient().Get().AbsPath(storageClusterGVKPath).DoRaw(ctx)
+	if err != nil {
+		klog.Errorf("failed to get storage cluster metadata: %v", err.Error())
+		return metav1.PartialObjectMetadata{}, fmt.Errorf("failed to get storage cluster metadata: %v", err)
```
nb-ohad, line 173: I did not notice it in prev review but you are returning a copy of the metadata. 
There is no reason to do that, hold the metadata as a pointer and in all of the error cases return nil. 

```suggestion
		return nil, fmt.Errorf("failed to get storage cluster metadata: %v", err)
```

onboarding/main.go
nb-ohad: @rchikatw I don't see any local constant and you still listing all storage clusters. Am I missing something?

rchikatw: Do you want me to get a specific storage cluster with name ocs-storagecluster

rchikatw: If yes,  I have made the change to get a specific storage cluster with the name(ocs-storagecluster), I have not found this constant in ocs-operator so created a local constant.

jarrpa:
Just to note, I poked at the whole pointer-to-value thing and effectively came to a similar conclusion to yours... with one catch: there's already a k8s library for this:

I submitted a draft PR to showcase [its usage](https://github.com/red-hat-storage/ocs-operator/pull/2428/commits/eee8fd59e714326b063ac52174a6c2f1798df955 ). I based it on this PR, so as soon as this one merges I'll mark mine as no longer a draft.

onboarding/main.go
```diff
@@ -101,7 +142,8 @@ func newClient() (*kubernetes.Clientset, error) {
 	clientset, err := kubernetes.NewForConfig(config)
 
 	if err != nil {
-		klog.Error(err, "failed to get clientset")
+		klog.Errorf("failed to create clientset: %v", err)
+		os.Exit(1)
```
leelavg, line 146: better to exit only from main and `return nil, err` here, wdyt?

rchikatw: Make sense, That is how we followed, Will do the change

rchikatw: Done.

controllers/storagecluster/provider_server.go
```diff
@@ -454,6 +454,8 @@ func getOnboardingJobObject(instance *ocsv1.StorageCluster) *batchv1.Job {
 			Namespace: instance.Namespace,
 		},
 		Spec: batchv1.JobSpec{
+			// Eligible to delete automatically when job finishes
+			TTLSecondsAfterFinished: util.ToPointer[int32](0),
```
nb-ohad, line 458: @jarrpa pointed out that k8s vendor already provides a library with similar functionality to `utils.ToPointer`:  https://pkg.go.dev/k8s.io/utils/ptr 

@rchikatw Can we please use that implementation instead and remove our util from the code base?

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -164,13 +164,17 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			return true
+		},
+		CreateFunc: func(e event.CreateEvent) bool {
+			return e.Object.GetName() != onboardingTicketPublicKeySecretName
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			return e.ObjectNew.GetName() != onboardingTicketPublicKeySecretName
+		},
+	}
```
nb-ohad, line 177: @rchikatw Why do we need this? 

If the operator, today or in the future, watches secrets this predicate will filter out these watches for create and update events. That is a bug!

If the underlying reason is that we want to act on events on our secret then we are already doing it with the `Owns` watch so this predicate is not needed anymore.  

If you want to NOT act on private secret events (kind of black listing these events) then you need a different predicate implementation

onboarding/main.go
```diff
@@ -18,90 +20,128 @@ import (
...
 
 }
 
 func newClient() (*kubernetes.Clientset, error) {
 	config := runtime.GetConfigOrDie()
 	clientset, err := kubernetes.NewForConfig(config)
-
 	if err != nil {
-		klog.Error(err, "failed to get clientset")
+		return nil, fmt.Errorf("failed to create clientset: %v", err)
```
nb-ohad, line 144: you are already logging this error on the call side so right now you will get:
` "failed to create clientset: failed to create clientset: <underlying err>` which is confusing. 

As we don't add new information with this message and the fact that we already log it on the call site I would suggest just returning the underlying error in this case.   

```suggestion
		return nil, err

**Comment by nb-ohad** [comment]
@rchikatw What is the reason you choose to use different formats for marshaling of private and public keys? 

You use PKCS1 for the private key:
```	privteKeyBytes := x509.MarshalPKCS1PrivateKey(privateKey)```

And PKIX for the public key:
```	publicKeyBytes, err := x509.MarshalPKIXPublicKey(publicKey)```

File: onboarding/main.go [hunk 28]
```diff
@@ -122,3 +162,18 @@ func convertRsaPublicKeyAsPemStr(publicKey *rsa.PublicKey) (string, error) {
 
 	return string(publicKeyPem), nil
 }
+
+func getStorageClusterMetadata(ctx context.Context, operatorNamespace string, clientset *kubernetes.Clientset) (*metav1.PartialObjectMetadata, error) {
+	var storageClusterMetadata metav1.PartialObjectMetadata
+	storageClusterGVKPath := fmt.Sprintf("/apis/%s/%s/namespaces/%s/%s/%s", "ocs.openshift.io", "v1", operatorNamespace, "storageclusters", storageClusterName)
```
nb-ohad, line 168: 1. This is a very long line, can we please reformat it?
2. Please embed the constant part into the format string. There is no need to have them as parameters.

```suggestion
	storageClusterGVKPath := fmt.Sprintf(
		"/apis/ocs.openshift.io/v1/namespaces/%s/storageclusters/%s", 
		operatorNamespace, 
		storageClusterName,
	)
```

onboarding/main.go
```diff
@@ -122,3 +162,18 @@ func convertRsaPublicKeyAsPemStr(publicKey *rsa.PublicKey) (string, error) {
...
+	instance, err := clientset.RESTClient().Get().AbsPath(storageClusterGVKPath).Do(ctx).Raw()
```
nb-ohad, line 169: `instance` is a meaningless name, it does not convey the content or the intention behind the variable.
lease use something like `storageClusterMetadataJSON` instead.

nb-ohad (changes requested):

rchikatw:
> @rchikatw What is the reason you choose to use different formats for marshaling of private and public keys?
> 
> You use PKCS1 for the private key: ` privteKeyBytes := x509.MarshalPKCS1PrivateKey(privateKey)`
> 
> And PKIX for the public key: ` publicKeyBytes, err := x509.MarshalPKIXPublicKey(publicKey)`

A good catch actually but I never faced any issues while onboarding the application cluster. I did generate the token several times and onboarded the application cluster. But let me try once changing x509.MarshalPKIXPublicKey -> x509.MarshalPKCS1PublicKey

controllers/storagecluster/storagecluster_controller.go
rchikatw: So should we implement blacklisting events on private onboarding secret? or Should i use owns on all secrets like `.Owns(&corev1.Secret{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).` a default predicate

nb-ohad: I think that blacklisting the private key is an overkill, we should not care if it triggers a reconcile or not as that reconcile will be a noop

rchikatw:
> > > @rchikatw What is the reason you choose to use different formats for marshaling of private and public keys?
> > > You use PKCS1 for the private key: ` privteKeyBytes := x509.MarshalPKCS1PrivateKey(privateKey)`
> > > And PKIX for the public key: ` publicKeyBytes, err := x509.MarshalPKIXPublicKey(publicKey)`
> > 
> > 
> > A good catch actually but I never faced any issues while onboarding the application cluster. I did generate the token several times and onboarded the application cluster. But let me try once changing x509.MarshalPKIXPublicKey -> x509.MarshalPKCS1PublicKey
> > I found the reason why I am using PKIX for marshaling the public key, when a new OnboardConsumer RPC call to onboard a new OCS application cluster happens it parses the public key using the ParsePKIXPublicKey so i have used the same for the public key. refer [here](https://github.com/red-hat-storage/ocs-operator/blob/main/services/provider/server/server.go#L407) and there is no function marshell private key using PKIX
> 
> But in that case shouldn't we just align the server code to use PKCS1 as well?
> 
> P.S I just found this https://stackoverflow.com/questions/49876521/golang-x509-marshalpkixpublickey-vs-x509-marshalpkcs1publickey#:~:text=Add%20a%20comment-,1%20Answer,-Sorted%20by%3A
> 
> According to this PCIX was designed to be used for key certificates. It encodes an identity as well as encryption. I think we should align all of our code to use PKCS1. That means both private and public keys

Yes that solves the problem if we used PKCS1 at both places, Let me do that and verify

onboarding/main.go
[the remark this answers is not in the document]
rchikatw: Yes i agree.

PR #2400, opened 2024-01-19
rook: generate rook csv using local script

subhamkrai:
Let's generate rook csv using local bash script that will be inside ocs-operator code and fetch from rook image. It gives us advantage of not depending on rook image for their csv and we can pass branch of commit for specific csv. Also for rook, it doesn't make sense to have csv codes/fiels in rook upstream and the rook csv is used only for the downstream. So, in future rook will remove the csv files from rook image.

### DIFF
--- a/hack/generate-rook-csv.sh
+++ b/hack/generate-rook-csv.sh
@@ -1,0 +1,92 @@
+#!/usr/bin/env bash
+set -e
+
+source hack/ensure-operator-sdk.sh
+
+#############
+# VARIABLES #
+#############
+
+yq="${YQv3:-yq}"
+PLATFORM=$(go env GOARCH)
+
+YQ_CMD_DELETE=("$yq" delete -i)
+YQ_CMD_MERGE_OVERWRITE=("$yq" merge --inplace --overwrite --prettyPrint)
+YQ_CMD_MERGE=("$yq" merge --arrays=append --inplace)
+YQ_CMD_WRITE=("$yq" write --inplace -P)
+CSV_FILE_NAME="deploy/csv-templates/crds/rook/manifests/rook-ceph.clusterserviceversion.yaml"
+CEPH_EXTERNAL_SCRIPT_FILE="https://raw.githubusercontent.com/rook/rook/master/deploy/examples/create-external-cluster-resources.py"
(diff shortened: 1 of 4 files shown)


hack/generate-rook-csv.sh
```diff
@@ -0,0 +1,91 @@
...
+ASSEMBLE_FILE_OCP="deploy/csv-templates/rook_olm_files/metadata-ocp.yaml"
+#############
+# FUNCTIONS #
+#############
+
+function generate_csv() {
+    # If you want to use specific branch, tags or commit has refer to this doc https://github.com/kubernetes-sigs/kustomize/blob/master/examples/remoteBuild.md#remote-directories
+    kubectl kustomize https://github.com/rook/rook//deploy/examples | "$operator_sdk" generate bundle --package="rook-ceph" --output-dir="deploy/csv-templates/crds/rook/" --extra-service-accounts=rook-ceph-system,rook-csi-rbd-provisioner-sa,rook-csi-rbd-plugin-sa,rook-csi-cephfs-provisioner-sa,rook-csi-nfs-provisioner-sa,rook-csi-nfs-plugin-sa,rook-csi-cephfs-plugin-sa,rook-ceph-system,rook-ceph-rgw,rook-ceph-purge-osd,rook-ceph-osd,rook-ceph-mgr,rook-ceph-cmd-reporter
+
+    directory="deploy/csv-templates/crds/rook/manifests"
```
iamniting, line 29: Can we move this to config/rook

iamniting (changes requested):

hack/generate-rook-csv.sh
subhamkrai: since we store rook csv crd in `deploy/csv-templates` I have put it under that. Let me know if you want to put it under `config/rook`. Thanks.

umangachapagain (commented):
This is a maintenance overhead on ocs-operator that we do not want.

travisn:
> This is a maintenance overhead on ocs-operator that we do not want.

Downstream artifacts should never have been included in Rook upstream, now we need to do the right thing for downstream. Let's discuss.

iamniting:
Closing it as we are managing rook CSV in the rook DS

PR #2407, opened 2024-01-22
Enable ceph client and collectors for multicluster case

umangachapagain:
Users can define custom auth via "ocs-metrics-exporter-ceph-auth" secret or fallback to "rook-csi-rbd-provisioner" which has the auth we need. It will work for all modes of deployment.

### DIFF
--- a/controllers/storagecluster/exporter.go
+++ b/controllers/storagecluster/exporter.go
@@ -325,7 +325,10 @@ func deployMetricsExporter(ctx context.Context, r *StorageClusterReconciler, ins
 			Spec: corev1.PodSpec{
 				Containers: []corev1.Container{
 					{
-						Args:    []string{"--namespaces", instance.Namespace},
+						Args: []string{
+							"--namespaces", instance.Namespace,
+							"--ceph-auth-namespace", r.OperatorNamespace,
+						},
 						Command: []string{"/usr/local/bin/metrics-exporter"},
 						Image:   r.images.OCSMetricsExporter,
 						Name:    metricsExporterName,
--- a/controllers/storagecluster/exporter.go
+++ b/controllers/storagecluster/exporter.go
@@ -488,6 +488,12 @@ func updateMetricsExporterClusterRoles(ctx context.Context, r *StorageClusterRec
 
 		currentClusterRole.Rules = []rbacv1.PolicyRule{
 			{
+				APIGroups:     []string{""},
+				Resources:     []string{"configmaps"},
+				Verbs:         []string{"get"},
+				ResourceNames: []string{"rook-ceph-csi-config"},
+			},
+			{
 				APIGroups: []string{""},
 				Resources: []string{"persistentvolumes", "persistentvolumeclaims", "pods", "nodes"},
 				Verbs:     []string{"get", "list", "watch"},
--- a/metrics/internal/cache/pv.go
+++ b/metrics/internal/cache/pv.go
@@ -32,13 +32,14 @@ type PersistentVolumeStore struct {
 	// Store is a map of PV UID to PersistentVolumeAttributes
 	Store map[types.UID]PersistentVolumeAttributes
 	// RBDClientMap is a map of RBD client addresses to the names of the nodes whose images had this client as a watcher
-	RBDClientMap      map[string][]string
-	monitorConfig     cephMonitorConfig
-	kubeClient        clientset.Interface
-	allowedNamespaces []string
+	RBDClientMap         map[string][]string
+	monitorConfig        cephMonitorConfig
+	kubeClient           clientset.Interface
+	cephClusterNamespace string
+	cephAuthNamespace    string
 
 	// Functions to make testing easier
-	initCephFn         func(kubeclient clientset.Interface, allowedNamespaces []string) (cephMonitorConfig, error)
+	initCephFn         func(kubeclient clientset.Interface, cephClusterNamespace, cephAuthNamespace string) (cephMonitorConfig, error)
 	runCephRBDStatusFn func(config *cephMonitorConfig, pool, image string) (Clients, error)
 	// TODO: Use fake k8s client instead
 	getNodeNameForPVFn func(pv *corev1.PersistentVolume, kubeClient clientset.Interface) (string, error)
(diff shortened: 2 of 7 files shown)


metrics/internal/cache/pv.go
```diff
@@ -135,6 +135,10 @@ func (p *PersistentVolumeStore) Add(obj interface{}) error {
 		}
 	}
 
+	if p.monitorConfig.clusterID != pv.Spec.CSI.VolumeAttributes["clusterID"] {
```
umangachapagain, line 138: @Madhu-1 is it safe to assume that `pv.Spec.CSI.VolumeAttributes["clusterID"]` always exists and points to the ceph cluster which backs the PV?

metrics/internal/cache/rbd-mirror.go
```diff
@@ -177,7 +177,7 @@ func initCeph(kubeclient clientset.Interface, allowedNamespaces []string) (cephM
 	}
 	id := "admin"
 
-	configmap, err := kubeclient.CoreV1().ConfigMaps(namespace).Get(context.TODO(), "rook-ceph-csi-config", v1.GetOptions{})
+	configmap, err := kubeclient.CoreV1().ConfigMaps(namespace).Get(context.TODO(), "rook-ceph-mon-endpoints", v1.GetOptions{})
```
umangachapagain, line 180: @Madhu-1 In case of external mode clusters configmap `rook-ceph-csi-config` in external cluster namespace is empty. But, configmap `rook-ceph-mon-endpoints` in same namespace is not empty and has the same details.
However, configmap `rook-ceph-csi-config` in operator namespace has 2 csi config details, 1 for internal mode and 1 for external mode. Is this expected?

For now I have moved to configmap `rook-ceph-mon-endpoints` as it's behavior is most consistent. But, maybe it's an issue that needs to be fixed for rook/ceph-csi.

Madhu-1: @umangachapagain do we have a clusters or some sample data to compare both, i assume `rook-ceph-csi-config` should also contains same details in case of both internal/external mode as well. https://github.com/rook/rook/blob/df04e5cce53f7b4c34cc060877b09cae2a5e4ffc/pkg/operator/ceph/csi/template/cephfs/csi-cephfsplugin-provisioner-dep.yaml#L247-L250 is the configmap mounted to csi pod which contains the mapping details

umangachapagain: CSI is deployed in operator namespace. So, even in case of internal+external, it will use the configmap from operator namespace for both clusters. Since that ConfigMap has needed details, nothing breaks. I am assuming CSI code can handle cases where key `csi-cluster-config-json` has multiple distinct values.

Madhu-1: i think using `rook-ceph-mon-endpoints` might not work if we have different subvolumegroups and radosnamespaces in the cluster, it will contains only the default entry am also not sure who is using that entry from monitor configmap. At just a glance that a dummy entry is monitor configmap which no one uses. you might need to switch back to `rook-ceph-csi-config` for csi operations.

umangachapagain: I am using the configmap only to get the clusterID and MON IPs for each cluster. What's the most reliable way to do that? Something that will work consistently between internal and external modes.
Maybe we need to fix the `rook-ceph-csi-config` behavior.

umangachapagain: @travisn Do you have any suggestions on this?

iamniting (approved):
Looks good to me.
/hold for others

metrics/internal/cache/rbd-mirror.go
```diff
@@ -167,9 +167,12 @@ func initCeph(kubeclient clientset.Interface, allowedNamespaces []string) (cephM
 	if !ok {
 		return cephMonitorConfig{}, fmt.Errorf("failed to get client key from secret in namespace %q", namespace)
 	}
-	id := "admin"
+	id, ok := secret.Data["ceph-user"]
```
umangachapagain, line 170: This provides id as `client.admin`, but earlier we were using `admin` only.
Does both of these work or do we need to strip the `client.` prefix? @travisn @Madhu-1

metrics/internal/cache/rbd-mirror.go
Madhu-1: >I am using the configmap only to get the clusterID and MON IPs for each cluster. What's the most reliable way to do that? Something that will work consistently between internal and external modes.
Maybe we need to fix the rook-ceph-csi-config behavior.

It should be same there wont be any problem because cephcsi uses this configmap in case of both internal/external. can we get some Yaml from both cases to check what exact data you are referring to?

@umangachapagain this doesnt work for the cases where we will have subvolumegroup or the radosnamespace, AFAIK `rook-ceph-mon-endpoints` will not get updated for the same (this might be bug in Rook)

metrics/internal/cache/rbd-mirror.go
Madhu-1: if we have client.x we need to strip `client.` prefix for CLI commands to work or else we will get permission denied error.

Madhu-1 (changes requested):
Requesting changes to ensure this works for subvolumegroup/radosnamespace as well

metrics/internal/cache/rbd-mirror.go
umangachapagain: Ack. Made the change.

metrics/internal/cache/rbd-mirror.go
umangachapagain: I have updated code to use `rook-ceph-csi-config` from the operator namespace which is the updated one and also used by Ceph-CSI.

metrics/internal/cache/pv.go
umangachapagain: @Madhu-1 Is this valid when using radosnamespace and subvolumegroups as well?

metrics/internal/cache/rbd-mirror.go
```diff
@@ -145,31 +137,34 @@ func (s *RBDMirrorStore) WithRBDCommandInput(namespace string) error {
...
+	if err != nil && !apierror.IsNotFound(err) {
+		return cephMonitorConfig{}, err
+	}
+
+	if apierror.IsNotFound(err) {
+		secret, err = kubeclient.CoreV1().Secrets(cephClusterNamespace).Get(context.TODO(), "rook-csi-rbd-provisioner", metav1.GetOptions{})
+		if err != nil {
+			return cephMonitorConfig{}, err
 		}
 	}
```
umangachapagain, line 155: This change allows users to define custom auth via "ocs-metrics-exporter-ceph-auth" secret or fallback to "rook-csi-rbd-provisioner" which has the auth we need. It will work for all modes of deployment.

metrics/internal/cache/rbd-mirror.go
```diff
@@ -145,31 +137,34 @@ func (s *RBDMirrorStore) WithRBDCommandInput(namespace string) error {
...
-	// under any of these circumstances we should not proceed
-	if err != nil || secret == nil || namespace == "" {
-		return cephMonitorConfig{}, fmt.Errorf("failed to get secret in any of these namespaces %+v: %v",
-			allowedNamespaces, err)
+
+	id, ok := secret.Data["userID"]
+	if !ok {
+		return cephMonitorConfig{}, fmt.Errorf("failed to get ceph user id in namespace %q", namespace)
```
Madhu-1, line 159: You might need to log the secret name here as well?

controllers/storagecluster/exporter.go
```diff
@@ -325,7 +325,10 @@ func deployMetricsExporter(ctx context.Context, r *StorageClusterReconciler, ins
 			Spec: corev1.PodSpec{
 				Containers: []corev1.Container{
 					{
-						Args:    []string{"--namespaces", instance.Namespace},
+						Args: []string{
+							"--namespaces", instance.Namespace,
```
Madhu-1, line 329: This works only for the internal/external mode not for both to gether isn't it?

metrics/internal/cache/rbd-mirror.go
umangachapagain: Avoiding secret name in the log. I think that message is enough to hint what is wrong.

controllers/storagecluster/exporter.go
umangachapagain: It is deployed per cluster. So, at a time, exporter works on 1 cluster only (either external or internal).
`--namespace` is storagecluster namespace and `--ceph-auth-namespace` is operator namespace.
So, we can handle all combinations.

Madhu-1 (approved):

iamniting (approved):

PR #2410, opened 2024-01-22
fix false positive error log in metrics-exporter

umangachapagain:
In case CephBlockPool mirroring is not enabled, it is logged as an error by metrics-exporter which is not true. This commit updates the check to ensure that we skip such cases and only log error when mirroring status is available but not in the state we expect.

### DIFF
--- a/metrics/internal/collectors/ceph-block-pool.go
+++ b/metrics/internal/collectors/ceph-block-pool.go
@@ -1,6 +1,8 @@
 package collectors
 
 import (
+	"strings"
+
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/red-hat-storage/ocs-operator/metrics/v4/internal/options"
 	cephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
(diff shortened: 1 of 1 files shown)


metrics/internal/collectors/ceph-block-pool.go
```diff
@@ -108,9 +110,15 @@ func (c *CephBlockPoolCollector) collectMirroringImageHealth(cephBlockPools []*c
 	for _, cephBlockPool := range cephBlockPools {
 		var imageHealth string
 		mirroringStatus := cephBlockPool.Status.MirroringStatus
-		if mirroringStatus != nil {
-			imageHealth = mirroringStatus.Summary.ImageHealth
+		if mirroringStatus == nil || mirroringStatus.Summary == nil {
```
Madhu-1, line 113: @umangachapagain can we check https://github.com/rook/rook/blob/df04e5cce53f7b4c34cc060877b09cae2a5e4ffc/deploy/examples/pool.yaml#L49 her as well as the mirroring need to enabled on the pool to check anything?

umangachapagain: Good catch. Done.

metrics/internal/collectors/ceph-block-pool.go
```diff
@@ -107,10 +109,21 @@ func getAllBlockPools(lister cephv1listers.CephBlockPoolLister, namespaces []str
...
+
+		if !cephBlockPool.Spec.Mirroring.Enabled {
+			return
+		}
+
 		mirroringStatus := cephBlockPool.Status.MirroringStatus
-		if mirroringStatus != nil {
-			imageHealth = mirroringStatus.Summary.ImageHealth
+		if mirroringStatus == nil || mirroringStatus.Summary == nil {
+			return
```
weirdwiz, line 119: if mirroring is enabled, wouldn't we need to show an error when mirroring status is nil?

umangachapagain: I wanted to be optimistic that operator will update it when needed. But, adding an error log here doesn't hurt.
I'll fix it.

Madhu-1 (commented):
LGTM, not approving to get ack from @weirdwiz

iamniting (approved):

PR #2413, opened 2024-01-22
fix PV cache error conditions

umangachapagain:
For cases where PV exists without PVC, or PV/PVC exists but is not used by any Pod etc, we skip caching those PVs instead of returning an error as these neither provide any meaningful data nor have any impact on monitoring.

### DIFF
--- a/metrics/internal/cache/pv.go
+++ b/metrics/internal/cache/pv.go
@@ -181,26 +181,16 @@ func getNodeNameForPV(pv *corev1.PersistentVolume, kubeClient clientset.Interfac
 		return "", nil
 	}
 
-	pvc, err := kubeClient.CoreV1().PersistentVolumeClaims(pv.Spec.ClaimRef.Namespace).Get(context.Background(), pv.Spec.ClaimRef.Name, metav1.GetOptions{})
+	nodeList, err := kubeClient.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{})
 	if err != nil {
-		return "", fmt.Errorf("failed to get PVC %s/%s: %v", pv.Spec.ClaimRef.Namespace, pv.Spec.ClaimRef.Name, err)
+		return "", err
 	}
 
-	if pvc.Spec.VolumeName != pv.Name {
-		return "", fmt.Errorf("persistent volume %s is not bound to claim %s/%s", pv.Name, pv.Spec.ClaimRef.Namespace, pv.Spec.ClaimRef.Name)
-	}
-
-	podList, err := kubeClient.CoreV1().Pods(pvc.Namespace).List(context.Background(), metav1.ListOptions{})
-	if err != nil {
-		return "", fmt.Errorf("failed to list pods in namespace %s: %v", pvc.Namespace, err)
-	}
-
-	for _, pod := range podList.Items {
-		if pod.Spec.Volumes != nil {
-			for _, volume := range pod.Spec.Volumes {
-				if volume.PersistentVolumeClaim != nil && volume.PersistentVolumeClaim.ClaimName == pvc.Name {
-					return pod.Spec.NodeName, nil
-				}
+	uniqueVolumeName := fmt.Sprintf("kubernetes.io/csi/%s^%s", pv.Spec.CSI.Driver, pv.Spec.CSI.VolumeHandle)
+	for _, node := range nodeList.Items {
+		for _, volumeInUse := range node.Status.VolumesInUse {
+			if volumeInUse == corev1.UniqueVolumeName(uniqueVolumeName) {
+				return node.Name, nil
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


metrics/internal/cache/pv.go
```diff
@@ -167,7 +171,7 @@ func (p *PersistentVolumeStore) Add(obj interface{}) error {
 
 func getNodeNameForPV(pv *corev1.PersistentVolume, kubeClient clientset.Interface) (string, error) {
 	if pv.Spec.ClaimRef == nil {
-		return "", fmt.Errorf("persistent volume %s is not bound to any claim", pv.Name)
+		return "", nil
```
Madhu-1, line 174: You can have PV with claimRef which is some PVC already deleted (doesn't exists anymore) but you can have status.Phase =Release but not Bound.


```yaml
[🎩︎]mrajanna@li-2cfbef4c-22d9-11b2-a85c-a3e4a93c405f rbd $]kubectl get pvc
No resources found in default namespace.
[🎩︎]mrajanna@li-2cfbef4c-22d9-11b2-a85c-a3e4a93c405f rbd $]kubectl get pv
NAME                                       CAPACITY   ACCESS MODES   RECLAIM POLICY   STATUS     CLAIM             STORAGECLASS      REASON   AGE
pvc-b00344cb-9c54-4596-bde3-e19447fecf75   1Gi        RWO            Retain           Released   default/rbd-pvc   rook-ceph-block            7m10s
[🎩︎]mrajanna@li-2cfbef4c-22d9-11b2-a85c-a3e4a93c405f rbd $]kubectl get pv -oyaml
apiVersion: v1
items:
- apiVersion: v1
  kind: PersistentVolume
  metadata:
    annotations:
      pv.kubernetes.io/provisioned-by: rook-ceph.rbd.csi.ceph.com
      volume.kubernetes.io/provisioner-deletion-secret-name: rook-csi-rbd-provisioner
      volume.kubernetes.io/provisioner-deletion-secret-namespace: rook-ceph
    creationTimestamp: "2024-01-22T16:42:32Z"
    finalizers:
    - kubernetes.io/pv-protection
    name: pvc-b00344cb-9c54-4596-bde3-e19447fecf75
    resourceVersion: "10934"
    uid: 1d10ed77-3cc9-40dc-9a87-350aa80b422e
  spec:
    accessModes:
    - ReadWriteOnce
    capacity:
      storage: 1Gi
    claimRef:
      apiVersion: v1
      kind: PersistentVolumeClaim
      name: rbd-pvc
      namespace: default
      resourceVersion: "10687"
      uid: b00344cb-9c54-4596-bde3-e19447fecf75
    csi:
      controllerExpandSecretRef:
        name: rook-csi-rbd-provisioner
        namespace: rook-ceph
      driver: rook-ceph.rbd.csi.ceph.com
      fsType: ext4
      nodeStageSecretRef:
        name: rook-csi-rbd-node
        namespace: rook-ceph
      volumeAttributes:
        clusterID: rook-ceph
        imageFeatures: layering
        imageFormat: "2"
        imageName: csi-vol-511a8008-418d-4423-8b31-6c8a8418a896
        journalPool: replicapool
        pool: replicapool
        storage.kubernetes.io/csiProvisionerIdentity: 1705941608727-3282-rook-ceph.rbd.csi.ceph.com
      volumeHandle: 0001-0009-rook-ceph-0000000000000001-511a8008-418d-4423-8b31-6c8a8418a896
    persistentVolumeReclaimPolicy: Retain
    storageClassName: rook-ceph-block
    volumeMode: Filesystem
  status:
    phase: Released
kind: List
metadata:
  resourceVersion: ""
```

umangachapagain: Ah didn't think about that. We must handle that case before listing PVCs.
Good catch again.

weirdwiz (approved):
good idea, will save on the memory
LGTM!

Madhu-1 (commented):
change LGTM , but i would suggest to do below to avoid list operations of the pods in a namespace.

We can List the Node object and check in-use volumes to see if the PVC is on a specific node. @umangachapagain WDYT?

umangachapagain:
> change LGTM , but i would suggest to do below to avoid list operations of the pods in a namespace.
> 
> We can List the Node object and check in-use volumes to see if the PVC is on a specific node. @umangachapagain WDYT?
>
We would avoid listing pods but require listing nodes and then loop through all the in-use volumes in each node.
Current approach requires listing pods in the PVC namespace only and looping through it. I need to see which one performs better for most cases.

Madhu-1:
> > change LGTM , but i would suggest to do below to avoid list operations of the pods in a namespace.
> > We can List the Node object and check in-use volumes to see if the PVC is on a specific node. @umangachapagain WDYT?
> 
> We would avoid listing pods but require listing nodes and then loop through all the in-use volumes in each node. Current approach requires listing pods in the PVC namespace only and looping through it. I need to see which one performs better for most cases.

i assume the number of nodes in the cluster will be less than the number of pods in a namespace but lets see which one works fine.

Madhu-1 (approved):
LGTM

PR #1926, opened 2023-01-31
feat: support key rotation encryption

debjyoti-pandit:

PR #2427, opened 2024-01-29
Add keyRotation spec to the StorageCluster

SanjalKatiyar:
https://issues.redhat.com/browse/RHSTOR-3891

Older PR: https://github.com/red-hat-storage/ocs-operator/pull/1926


controllers/storagecluster/noobaa_system_reconciler.go
```diff
@@ -240,6 +241,9 @@ func (r *StorageClusterReconciler) setNooBaaDesiredState(nb *nbv1.NooBaa, sc *oc
 		}
 	}
 
+	isEnabled, rotationSchedule := util.GetKeyRotationSpec(sc)
+	nb.Spec.Security.KeyManagementService.EnableKeyRotation = isEnabled
+	nb.Spec.Security.KeyManagementService.Schedule = rotationSchedule
```
SanjalKatiyar, line 246: looks like NooBaa is always looking for the spec under KMS: https://github.com/noobaa/noobaa-operator/blob/master/pkg/apis/noobaa/v1alpha1/noobaa_types.go#L254

SanjalKatiyar:
@iamniting @Rakshith-R ptal.

controllers/util/util.go
```diff
@@ -22,3 +26,16 @@ func DetectDuplicateInStringSlice(slice []string) bool {
...
+		schedule := sc.Spec.Encryption.KeyRotation.Schedule
+		if schedule == "" {
+			// default schedule
+			schedule = "@Weekly"
+		}
+		// use key-rotation by default if cluster-wide encryption is opted without KMS
+		return true, schedule
+	}
+	return sc.Spec.Encryption.KeyRotation.Enabled, ""
+}
```
SanjalKatiyar, line 41: when we support KMS, we can update this utility function...

api/v1/storagecluster_types.go
```diff
@@ -447,6 +447,19 @@ type EncryptionSpec struct {
...
+}
+
+// KeyRotationSpec represents the settings for Key Rotation.
+type KeyRotationSpec struct {
+	// Enabled represents whether the key rotation is enabled.
+	// +optional
+	Enabled bool `json:"enabled,omitempty"`
+	// Schedule represents the cron schedule for key rotation.
+	// +optional
+	Schedule string `json:"schedule,omitempty"`
```
Rakshith-R, line 462: `omtiempty` is only helpful when the object is a pointer. We can remove it.

Let's add default to Schedule and keep it as @weekly using kubebuilder annotation too.

Rakshith-R (commented):
just a small nit 
The change looks good to me, Thanks !

cc @baum @travisn

api/v1/storagecluster_types.go
SanjalKatiyar: I did not add default values on purpose, does not make sense to add it even if cluster-wide encryption is not opted by the user...

SanjalKatiyar: instead let `GetKeyRotationSpec` handle it at runtime...

SanjalKatiyar: let me know if you still want to add defaults using markers ?

iamniting (changes requested):

api/v1/storagecluster_types.go
```diff
@@ -447,6 +447,19 @@ type EncryptionSpec struct {
...
+	// KeyRotation defines options for Key Rotation.
+	// +optional
+	KeyRotation KeyRotationSpec `json:"keyRotation,omitempty"`
+}
+
+// KeyRotationSpec represents the settings for Key Rotation.
+type KeyRotationSpec struct {
+	// Enabled represents whether the key rotation is enabled.
+	// +optional
+	Enabled bool `json:"enabled,omitempty"`
```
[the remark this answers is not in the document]
SanjalKatiyar: copied it from rook: https://github.com/rook/rook/blob/master/pkg/apis/ceph.rook.io/v1/types.go#L319 :/
will update here though !

api/v1/storagecluster_types.go
Rakshith-R: > let me know if you still want to add defaults using markers ?

Yes, lets use it so its explicitly visible in the CR.
( even if it is not in use )

SanjalKatiyar: updated...
I did not remove `omtiempty` though, it is harmless and is used for almost every field in the StorageCluster struct...

iamniting (approved):

PR #2437, opened 2024-02-01
storageclassrequest: wait for cache to update before next reconcile

rewantsoni:
status update for storageclassrequest is failing because of stale values in the cache which leads to the creation of multiple svgs, hence we need to wait till the cache is updated before the next reconcile

### DIFF
--- a/controllers/storageclassrequest/storageclassrequest_controller.go
+++ b/controllers/storageclassrequest/storageclassrequest_controller.go
@@ -16,7 +16,9 @@ package storageclassrequest
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"strings"
+	"time"
 
 	"github.com/go-logr/logr"
 	"github.com/google/uuid"
@@ -30,6 +32,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog/v2"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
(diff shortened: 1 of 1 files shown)


controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -131,6 +134,24 @@ func (r *StorageClassRequestReconciler) Reconcile(ctx context.Context, request r
 		return result, statusError
 	}
 
+	// wait for the cache to update so that we don't create multiple resources
```
Madhu-1, line 137: we have waitForsyncCache here https://github.com/red-hat-storage/ocs-operator/blob/13a9dfb60ebd8e539198a33e44ba22751e041c35/controllers/storageclassrequest/storageclassrequest_controller.go#L77-L79 isnt this helping?

rewantsoni: No, the create/update for status is done on the API server, the data is not written to the cache, the cache gets populated via a reflector that watches events from the API server. The apiserver might be generating the status update after we run over WaitForCacheSync.
We read the status in the second reconcile after WaitForCacheSync and it was empty leading to the creation of multiple svg

nb-ohad: @Madhu-1 waitForsyncCache is not doing what you expect it to do by looking at the name alone

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -116,10 +119,72 @@ func (r *StorageClassRequestReconciler) Reconcile(ctx context.Context, request r
...
+			"StorageClassRequest", r.StorageClassRequest.Name)
+	}
+
+	if statusUpdated {
+		err := wait.PollUntilContextTimeout(r.ctx, 100*time.Millisecond, 2*time.Second, true, func(_ context.Context) (done bool, err error) {
+
+			// wait for the cache to update so that we don't create multiple resources
+			cachedStorageClassRequest := &v1alpha1.StorageClassRequest{}
+			cachedStorageClassRequest.Name = request.Name
+			cachedStorageClassRequest.Namespace = request.Namespace
```
nb-ohad, line 171: Why can't we just refetch the resource on `r`? after saving the Status? or even just the resource version?

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -116,10 +119,72 @@ func (r *StorageClassRequestReconciler) Reconcile(ctx context.Context, request r
 
 	result, reconcileError = r.reconcilePhases()
 
-	// Apply status changes to the StorageClassRequest
-	statusError := r.Client.Status().Update(r.ctx, r.StorageClassRequest)
+	statusUpdated := false
```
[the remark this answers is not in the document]
rewantsoni: Right, updated it

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -116,10 +119,72 @@ func (r *StorageClassRequestReconciler) Reconcile(ctx context.Context, request r
...
+				if err := r.get(tempStorageClassRequest); err != nil {
+					if errors.IsNotFound(err) {
+						r.log.Error(err, "StorageClassRequest resource not found. Ignoring since object must be deleted.")
+						return true, err
+					}
+					r.log.Error(err, "Failed to get StorageClassRequest.")
+					return false, err
+				}
+				r.StorageClassRequest.Status.DeepCopyInto(&tempStorageClassRequest.Status)
+				tempStorageClassRequest.DeepCopyInto(r.StorageClassRequest)
```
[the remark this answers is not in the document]
rewantsoni: the tempStorgeClassRequest has the new object (after the conflict error), and the status is stored in r.StorageClassRequest. We need to update the in-memory object to have the same ObjectMeta (Assuming the spec is not changing, as we already have a plan to make spec immutable).
I think I can modify the above code to something like to make sure that ObjectMeta is matching so that we don't have the failure due to conflict in the next reconcile

> tempStorageClassRequest.ObjectMeta.DeepCopyInto(&r.StorageClassRequest.ObjectMeta)

rewantsoni:
Closing this in as this is addressed differently by: https://github.com/red-hat-storage/ocs-operator/pull/2449

PR #2442, opened 2024-02-05
stop CephMonLowNumber alert when MON count is 5

umangachapagain:
Even when number of MONs were increased from 3 to 5, the alert was not stopping. This is because the query did not check if Mon count is already >=5. This PR adds that check to ensure we stop alerting when MON count is 5 or more.


umangachapagain:

This does not test for alerts.

iamniting (approved):

PR #2449, opened 2024-02-07
storageClassRequest: move away from status as source of truth

rewantsoni:
this PR updates the svg to have a consistent naming scheme and fetch the resources via ownerReference instead of looking at the status for the source of truth

### DIFF
--- a/controllers/storageclassrequest/storageclassrequest_controller.go
+++ b/controllers/storageclassrequest/storageclassrequest_controller.go
@@ -15,6 +15,8 @@ package storageclassrequest
 
 import (
 	"context"
+	"crypto/md5"
+	"encoding/hex"
 	"fmt"
 	"strings"
 
@@ -24,6 +26,7 @@ import (
 	v1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
 	"github.com/red-hat-storage/ocs-operator/api/v4/v1alpha1"
 	controllers "github.com/red-hat-storage/ocs-operator/v4/controllers/storageconsumer"
+	"github.com/red-hat-storage/ocs-operator/v4/controllers/util"
 	rookCephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
 	storagev1 "k8s.io/api/storage/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -239,14 +242,33 @@ func (r *StorageClassRequestReconciler) initPhase(storageProfile *v1.StorageProf
 	} else if r.StorageClassRequest.Spec.Type == "sharedfilesystem" {
 		r.cephFilesystemSubVolumeGroup = &rookCephv1.CephFilesystemSubVolumeGroup{}
 		r.cephFilesystemSubVolumeGroup.Namespace = r.OperatorNamespace
-		for _, res := range r.StorageClassRequest.Status.CephResources {
-			if res.Kind == "CephFilesystemSubVolumeGroup" {
-				r.cephFilesystemSubVolumeGroup.Name = res.Name
-				break
-			}
+
+		cephFilesystemSubVolumeGroupList := &rookCephv1.CephFilesystemSubVolumeGroupList{}
+		err := r.Client.List(r.ctx, cephFilesystemSubVolumeGroupList, client.InNamespace(r.OperatorNamespace))
+		if err != nil {
+			return err
 		}
-		if r.cephFilesystemSubVolumeGroup.Name == "" {
-			r.cephFilesystemSubVolumeGroup.Name = fmt.Sprintf("cephfilesystemsubvolumegroup-%s-%s", r.storageConsumer.Name, generateUUID())
+		ownedCephFilesystemSubVolumeGroups := util.Filter(
+			cephFilesystemSubVolumeGroupList.Items,
+			func(item *rookCephv1.CephFilesystemSubVolumeGroup) bool {
+				for i := range item.OwnerReferences {
+					if item.OwnerReferences[i].UID == r.StorageClassRequest.UID {
+						return true
+					}
+				}
+				return false
+			})
+
+		svgItemsLen := len(ownedCephFilesystemSubVolumeGroups)
+		if svgItemsLen == 0 {
+			md5Sum := md5.Sum([]byte(r.StorageClassRequest.Name))
+			r.cephFilesystemSubVolumeGroup.Name = fmt.Sprintf("cephfilesystemsubvolumegroup-%s", hex.EncodeToString(md5Sum[:16]))
+		} else if svgItemsLen == 1 {
+			r.cephFilesystemSubVolumeGroup.Name = ownedCephFilesystemSubVolumeGroups[0].GetName()
+			r.log.V(1).Info(fmt.Sprintf("CephFilesystemSubVolumeGroup found: %s", r.cephFilesystemSubVolumeGroup.Name))
+		} else {
+			return fmt.Errorf(
+				"invalid number of CephFilesystemSubVolumeGroups owned by StorageClassRequest %q: expecting 0-1, found %d", r.StorageClassRequest.Name, svgItemsLen)
 		}
 	}
 
--- a/controllers/storageclassrequest/storageclassrequest_controller_test.go
+++ b/controllers/storageclassrequest/storageclassrequest_controller_test.go
@@ -34,12 +34,13 @@ import (
 )
 
 const (
-	pgAutoscaleMode    = "pg_autoscale_mode"
-	pgNum              = "pg_num"
-	pgpNum             = "pgp_num"
-	namespaceName      = "test-ns"
-	deviceClass        = "ssd"
-	storageProfileKind = "StorageProfile"
+	pgAutoscaleMode        = "pg_autoscale_mode"
+	pgNum                  = "pg_num"
+	pgpNum                 = "pgp_num"
+	namespaceName          = "test-ns"
+	deviceClass            = "ssd"
+	storageProfileKind     = "StorageProfile"
+	storageClassRequestUID = "storageClassRequestUUID"
 )
 
 var fakeStorageProfile = &v1.StorageProfile{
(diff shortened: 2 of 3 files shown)


controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -670,3 +679,19 @@ func generateUUID() string {
...
+		return nil, err
+	}
+
+	for _, cephFilesystemSubVolumeGroup := range cephFilesystemSubVolumeGroupList.Items {
+		if cephFilesystemSubVolumeGroup.OwnerReferences[0].Name == owner.Name {
+			return &cephFilesystemSubVolumeGroup, nil
+		}
+	}
+	return nil, nil
+}
```
nb-ohad, line 697: 1. Let's have this as a func on `r` so there will be no need to pass the client or the context
2. Please use the term `list` instead of `fetch` to align with k8s API terminology 
3. Don't assume a single-owner reference
4. Compare by UUID and not by name as UUID is never reused
5. Don't  return the result, fill in the provided list object

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -239,14 +241,21 @@ func (r *StorageClassRequestReconciler) initPhase(storageProfile *v1.StorageProf
...
-		for _, res := range r.StorageClassRequest.Status.CephResources {
-			if res.Kind == "CephFilesystemSubVolumeGroup" {
-				r.cephFilesystemSubVolumeGroup.Name = res.Name
-				break
-			}
+
+		cephFilesystemSubVolumeGroupList := &rookCephv1.CephFilesystemSubVolumeGroupList{}
+		listOptions := &client.MatchingLabels{
+			controllers.StorageConsumerNameLabel: r.storageConsumer.Name,
+			controllers.StorageProfileSpecLabel:  storageProfile.GetSpecHash(),
```
nb-ohad, line 248: Why do we care about this? we are fetching by owner anyways. What I do think we should filter is by namespace.

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -239,14 +241,21 @@ func (r *StorageClassRequestReconciler) initPhase(storageProfile *v1.StorageProf
...
 		}
-		if r.cephFilesystemSubVolumeGroup.Name == "" {
-			r.cephFilesystemSubVolumeGroup.Name = fmt.Sprintf("cephfilesystemsubvolumegroup-%s-%s", r.storageConsumer.Name, generateUUID())
+		cephFilesystemSubVolumeGroup, err := fetchCephFilesystemSubVolumeGroupByOwnerReference(r.ctx, r.Client, r.StorageClassRequest.ObjectMeta, cephFilesystemSubVolumeGroupList, client.InNamespace(r.OperatorNamespace), listOptions)
```
nb-ohad, line 250: 1. Don't return  
2. Can we have this be more generic and utilize the controller struct?

 ```suggestion
		err := listByOwner(&cephFilesystemSubVolumeGroupList, r.storageClassRequest,  client.InNamespace(r.OperatorNamespace))
```

**Review by nb-ohad — CHANGES_REQUESTED** [review]

File: controllers/storageclassrequest/storageclassrequest_controller.go [hunk 4]
```diff
@@ -239,14 +242,32 @@ func (r *StorageClassRequestReconciler) initPhase(storageProfile *v1.StorageProf
...
-		if r.cephFilesystemSubVolumeGroup.Name == "" {
-			r.cephFilesystemSubVolumeGroup.Name = fmt.Sprintf("cephfilesystemsubvolumegroup-%s-%s", r.storageConsumer.Name, generateUUID())
+		ownedCephFilesystemSubVolumeGroups := util.Filter(cephFilesystemSubVolumeGroupList.Items, func(item *rookCephv1.CephFilesystemSubVolumeGroup) bool {
+			for _, ownerReference := range item.OwnerReferences {
+				if ownerReference.UID == r.StorageClassRequest.UID {
+					return true
+				}
+			}
+			return false
+		})
```
nb-ohad, line 258: 1. Use multiline formatting when the line is long
2. Use pointers when iterating on complex data types to prevent deep copy of the object. Or if you only need to access the item once, refer to it by index 

```suggestion
		ownedCephFilesystemSubVolumeGroups := util.Filter(
			cephFilesystemSubVolumeGroupList.Items, 
			func(item *rookCephv1.CephFilesystemSubVolumeGroup) bool {
				for i := range item.OwnerReferences {
					if item.OwnerReferences[i].UID == r.StorageClassRequest.UID {
						return true
					}
				}
				return false
			}
		)
```

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -239,14 +242,32 @@ func (r *StorageClassRequestReconciler) initPhase(storageProfile *v1.StorageProf
...
+		if svgItemsLen == 0 {
+			md5Sum := md5.Sum([]byte(r.StorageClassRequest.Name))
+			svgNewName := fmt.Sprintf("cephfilesystemsubvolumegroup-%s", hex.EncodeToString(md5Sum[:16]))
+			r.log.V(1).Info("no valid CephFilesystemSubVolumeGroup found, creating new one", "CephFilesystemSubVolumeGroup", svgNewName)
+			r.cephFilesystemSubVolumeGroup.Name = svgNewName
+		} else if svgItemsLen == 1 {
+			r.cephFilesystemSubVolumeGroup.Name = cephFilesystemSubVolumeGroupList.Items[0].GetName()
+			r.log.V(1).Info("valid CephFilesystemSubVolumeGroup found", "CephFilesystemSubVolumeGroup", r.cephFilesystemSubVolumeGroup.Name)
+		} else {
+			return fmt.Errorf("invalid number of cephFilesystemSubVolumeGroups for storage consumer %q and storage profile %q: found %d, expecting 0 or 1", r.storageConsumer.Name, storageProfile.Name, svgItemsLen)
```
nb-ohad, line 270: No need to mention storage profile as this is an SVG not a block pool
```suggestion
			return fmt.Errorf(
				"invalid number of cephFilesystemSubVolumeGroups owned by storage consumer %q: expecting 0-1, found %d, 
				r.storageConsumer.Name, 
				storageProfile.Name, 
				svgItemsLen
			)
```

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -239,14 +242,32 @@ func (r *StorageClassRequestReconciler) initPhase(storageProfile *v1.StorageProf
...
+
+		svgItemsLen := len(ownedCephFilesystemSubVolumeGroups)
+		if svgItemsLen == 0 {
+			md5Sum := md5.Sum([]byte(r.StorageClassRequest.Name))
+			svgNewName := fmt.Sprintf("cephfilesystemsubvolumegroup-%s", hex.EncodeToString(md5Sum[:16]))
+			r.log.V(1).Info("no valid CephFilesystemSubVolumeGroup found, creating new one", "CephFilesystemSubVolumeGroup", svgNewName)
+			r.cephFilesystemSubVolumeGroup.Name = svgNewName
+		} else if svgItemsLen == 1 {
+			r.cephFilesystemSubVolumeGroup.Name = cephFilesystemSubVolumeGroupList.Items[0].GetName()
+			r.log.V(1).Info("valid CephFilesystemSubVolumeGroup found", "CephFilesystemSubVolumeGroup", r.cephFilesystemSubVolumeGroup.Name)
```
nb-ohad, line 268:
```suggestion
			r.log.V(1).Info(fmt.Sprintf("CephFilesystemSubVolumeGroup found: %s",  r.cephFilesystemSubVolumeGroup.Name))
```

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -239,14 +242,32 @@ func (r *StorageClassRequestReconciler) initPhase(storageProfile *v1.StorageProf
...
+				}
+			}
+			return false
+		})
+
+		svgItemsLen := len(ownedCephFilesystemSubVolumeGroups)
+		if svgItemsLen == 0 {
+			md5Sum := md5.Sum([]byte(r.StorageClassRequest.Name))
+			svgNewName := fmt.Sprintf("cephfilesystemsubvolumegroup-%s", hex.EncodeToString(md5Sum[:16]))
+			r.log.V(1).Info("no valid CephFilesystemSubVolumeGroup found, creating new one", "CephFilesystemSubVolumeGroup", svgNewName)
```
nb-ohad, line 264: Not sure why we need this here. You are not creating anything yet

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -239,14 +242,32 @@ func (r *StorageClassRequestReconciler) initPhase(storageProfile *v1.StorageProf
...
+					return true
+				}
+			}
+			return false
+		})
+
+		svgItemsLen := len(ownedCephFilesystemSubVolumeGroups)
+		if svgItemsLen == 0 {
+			md5Sum := md5.Sum([]byte(r.StorageClassRequest.Name))
+			svgNewName := fmt.Sprintf("cephfilesystemsubvolumegroup-%s", hex.EncodeToString(md5Sum[:16]))
```
nb-ohad, line 263: No need for intermediate variables
```suggestion
			r.cephFilesystemSubVolumeGroup.Name := fmt.Sprintf("cephfilesystemsubvolumegroup-%s", hex.EncodeToString(md5Sum[:16]))
```

controllers/util/util.go
```diff
@@ -39,3 +39,15 @@ func GetKeyRotationSpec(sc *ocsv1.StorageCluster) (bool, string) {
...
+func Filter[T any](list []T, f func(item *T) bool) []T {
+	ret := make([]T, 0)
+	for idx := range list {
+		ele := &list[idx]
+		if f(ele) {
+			ret = append(ret, *ele)
+		}
+	}
+	return ret
+}
```
nb-ohad, line 53: Please move the `utils/slices.go` or create if does not exists

PR #2451, opened 2024-02-08
[fusion-hci-4.14] storageClassRequest: move away from status as source of truth

leelavg:
manual cherry-pick of one commit from #2449. the diff from #2449  is we aren't taking the unit test file which doesn't exist in fusion-hci-4.14 branch and different pr introduced that in main.

this also contains two commits from #2249 which was a refactor of function signature and logging update.


leelavg:
either we need to merge this backport pr with only required cherry-picked commits or backport #2249, #2288 and #2449 in their entirety.

PR #2452, opened 2024-02-08
onboarding key generator package rename as it does not implies what it does

rchikatw:
- renaming the package for the onboarding job as it does not imply what it does. Renamed to onboarding-validation-keys-generator


rchikatw:
@leelavg @nb-ohad please review.

controllers/storagecluster/provider_server.go
```diff
@@ -27,11 +27,11 @@ import (
...
 
 const (
-	ocsProviderServerName               = "ocs-provider-server"
-	providerAPIServerImage              = "PROVIDER_API_SERVER_IMAGE"
-	onboardingSecretGeneratorImage      = "ONBOARDING_SECRET_GENERATOR_IMAGE"
-	onboardingJobName                   = "onboarding-secret-generator"
-	onboardingTicketPublicKeySecretName = "onboarding-ticket-key"
+	ocsProviderServerName                   = "ocs-provider-server"
+	providerAPIServerImage                  = "PROVIDER_API_SERVER_IMAGE"
+	rotateOnboardingKeypairsImage           = "ROTATE_ONBOARDING_KEYPAIRS_IMAGE"
```
nb-ohad, line 32: 1. "Key pairs: are two different words as such you should camelcase both. 
2. The terminology here is work. These are not onboarding keys, these are onboarding validation keys.
3. We should not name a component with an action (e.g. "ROTATE") we should name as as a subject.
4. The job technically does not rotate the keys, it regenerates the keys.

With that in mind I suggest calling the component `onboardingValidationKeysGenerator` and all other names should derive from it. 

For example:

```suggestion
	onboardingValidationKeysGeneratorImage           = "ONBOARDING_VALIDATION_KEYS_GENERATOR_IMAGE"
```

controllers/storagecluster/provider_server.go
nb-ohad, line 1: Why is the job reconciliation job under the `provider_server.go` file? 
the way OCS operator is structured each file is responsible for reconciling a single concern.

controllers/storagecluster/provider_server_test.go
nb-ohad, line 1: Same as the comment about `provider_server.go` above

deploy/ocs-operator/manifests/onboarding-secret-generator-binding.yaml
nb-ohad, line 1: This file are not named as we expect

deploy/ocs-operator/manifests/onboarding-secret-generator-role.yaml
nb-ohad, line 1: This file are not named as we expect

deploy/ocs-operator/manifests/onboarding-secret-generator-sa.yaml
nb-ohad, line 1: This file are not named as we expect

rbac/onboarding-secret-generator-binding.yaml
nb-ohad, line 1: This file are not named as we expect

rbac/onboarding-secret-generator-role.yaml
nb-ohad, line 1: This file are not named as we expect

rbac/onboarding-secret-generator-sa.yaml
nb-ohad, line 1: This file are not named as we expect

rotate-onboarding-keys/main.go
nb-ohad, line 1: 1. Naming a package with an action as a name is not a very good practice. Can we call it `onboarding-valiadtion-keys-generator` instead? 

2. The changes in this file are completely separate from the changes in naming. As are These completely different concern we should have 2 separate PRs

jarrpa: Honestly I think it shouldn't be that specific. Something simple like `remote-consumer-onboarding-job` would be ideal, as it also doesn't lock us in to the current mechanism we're using for onboarding, if it ever ends up changing.

jarrpa (changes requested):
I'm in general agreement with Ohad. All the renaming changes have nothing to do with what this PR claims to be doing, and most of them don't make a whole lot of sense per the project's naming conventions. Please move them to a separate PR.

rotate-onboarding-keys/main.go
nb-ohad: @jarrpa, the name you just suggested is misleading. This job has nothing to do with the onboarding process itself. The job responsibility is to generate (and technically also rotate) the public and private keys used in the onboarding process. The reason it is not just another step in ocs operator's reconciliation is because of the need to isolate the write permissions for these secrets   

As a side note, we don't use the "consumer" terminology anymore. we use ocs client

rotate-onboarding-keys/main.go
```diff
@@ -57,90 +56,87 @@ func main() {
...
-	err = clientset.CoreV1().
-		Secrets(operatorNamespace).
-		Delete(ctx, onboardingValidationPrivateKeySecretName, metav1.DeleteOptions{})
+	privateSecret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      onboardingValidationPrivateKeySecretName,
+			Namespace: operatorNamespace,
+		},
+	}
+	err = cl.Delete(ctx, privateSecret, client.GracePeriodSeconds(0))
```
leelavg, line 65: why do we need graceperiod option here? I remember using this only w/ Pods 🤔.

jarrpa: I'm also wondering about this, especially given that it was not being used previously. If it's a known difference in behavior and you're just maintaining previous behaviors, that's perfectly fine.

rchikatw: @leelavg  I have created a new Pull Request (PR) for replacing the native client with runtime. you are Right its only used for pods, I have removed the grace period for deletion of pods in this PR.  please review the new PR and we can discuss any further comments over there. @jarrpa  Just wanted to let you know that the deletion of the secret was also present in the previous PR. Thanks!

onboarding-validation-keys-generator/main.go
leelavg, line 1: I believe the intention was not to only rename but also move this from repo root or was it decided that only rename is required?

rchikatw: It was suggested to move to proper place if there is any. But I can say this is the place where it can fit well. I also discussed with Nitin whether should I move this package to any other location or it is fine to keep it in the repo root. Nitin suggests the current location is correct. @leelavg If you feel this to be moved to any other location please suggest we can think and work on it if it is more suitable. I cannot move it under services as it's not a service.

leelavg: Fair enough. Thanks.

Dockerfile
```diff
@@ -12,7 +12,7 @@ ARG LDFLAGS
 
 RUN GOOS="$GOOS" GOARCH="$GOARCH" go build -ldflags "$LDFLAGS" -tags netgo,osusergo -o ocs-operator main.go
 RUN GOOS="$GOOS" GOARCH="$GOARCH" go build -tags netgo,osusergo -o provider-api services/provider/main.go
-RUN GOOS="$GOOS" GOARCH="$GOARCH" go build -tags netgo,osusergo -o onboarding-secret-generator onboarding/main.go
+RUN GOOS="$GOOS" GOARCH="$GOARCH" go build -tags netgo,osusergo -o onboarding-validation-keypairs-generator onboarding-validation-keys-generator/main.go
```
nb-ohad, line 15: Why do we use 2 different names here?
- onboarding-validation-keypairs-generator
- onboarding-validation-keys-generator

Dockerfile
```diff
@@ -21,7 +21,7 @@ FROM registry.access.redhat.com/ubi9/ubi-minimal
 
 COPY --from=builder workspace/ocs-operator /usr/local/bin/ocs-operator
 COPY --from=builder workspace/provider-api /usr/local/bin/provider-api
-COPY --from=builder workspace/onboarding-secret-generator /usr/local/bin/onboarding-secret-generator
+COPY --from=builder workspace/onboarding-validation-keypairs-generator /usr/local/bin/onboarding-validation-keypairs-generator
```
nb-ohad, line 24: I believe we agreed on `onboarding-validation-keys-generator` and not `onboarding-validation-keypairs-generator`

controllers/storagecluster/provider_server.go
nb-ohad: I am not sure if this is a real concern. We are reconciling a new entity (Job) and the way the OCS operator code is laid out is that each entity gets its own file which implements `ensureCreated` and  `ensureDeleted`. This code breaks that pattern.

With that in mind, I agree this PR is not the place to resolve this issue. @rchikatw Can you please open an issue to capture the need to do this change, attach the URL under this comment, and then I will resolve?

nb-ohad (changes requested):

controllers/storagecluster/provider_server.go
rchikatw: I think I may have misunderstood your comment. Do you mean why I am creating a job in the provider_server.go file?
if yes, in the call, it was decided that we need to create the job when it is deployed in provider mode. The location where I need to add this has also been determined in the call. As per our discussion, I need to add it after the creation of the provider deployment. the place for this should be https://github.com/red-hat-storage/ocs-operator/blob/main/controllers/storagecluster/provider_server.go#L66

nb-ohad: After does not mean in the same file? The `ensure*` functions are ordered. I would have expected an ensure func for the job in its own file scheduled after the provider server ensure function

rchikatw: Okay got it.

Dockerfile
jarrpa: The directory itself can just be `onboarding-validation`. The fact that the artifact produced is an image for generating keypairs is an implementation detail.

controllers/storagecluster/provider_server.go
rchikatw: raised a issue https://github.com/red-hat-storage/ocs-operator/issues/2487

Dockerfile
nb-ohad: The directory should not change as it is the source directory which means it is the name of the go package. And we already deemed `onboarding-validation` as an unfit name for the package in the past.

Why not use something like `onboarding-validation-keys-gen` for the image name?

Dockerfile
nb-ohad: see https://github.com/red-hat-storage/ocs-operator/pull/2452/files#r1510218145

Dockerfile
rchikatw: Make sense to me

leelavg:
@rchikatw pls make sure that you inform build team about Dockerfile changes.

jarrpa:
Hah, I had an earlier review comment that didn't get sent... might as well at least actually publish it.

I don't see why there needs to be a change to the directory name at all, it just looks like introducing needless build changes for the sake of being more verbose. And, as it was, it seemed fine that the directory name itself was kinda vague since the artifacts produced from it are an implementation detail.

rchikatw:
> pls make sure that you inform build team about Dockerfile changes.

Sure...Thanks for pointing that out

PR #2457, opened 2024-02-12
add rpc call for PeerBlockPool

rewantsoni:
Add a new RPC call PeerBlockPool to enable one-way mirroring between the block pools.

The rpc call is initiated from the primary cluster which wants to mirror the pool, the server receives the request and does the following:

1. Enable RBD Mirror Daemon
2. Enable Mirroring on the Block Pool in the request
3. Create the bootstrap secret and set the bootstrap secret ref on the Block Pool in the request.

Depends on: 
1. https://github.com/red-hat-storage/ocs-operator/pull/2480


services/provider/client/client.go
```diff
@@ -207,3 +207,20 @@ func (cc *OCSProviderClient) ReportStatus(ctx context.Context, consumerUUID stri
...
+
+func (cc *OCSProviderClient) PeerBlockPool(ctx context.Context, blockPoolName string, pool, token []byte) (*pb.PeerBlockPoolResponse, error) {
+	if cc.Client == nil || cc.clientConn == nil {
+		return nil, fmt.Errorf("OCS client is closed")
+	}
+
+	req := &pb.PeerBlockPoolRequest{
+		BlockPoolName: blockPoolName,
+		Pool:          pool,
+		Token:         token,
```
leelavg, line 219: should these values be encrypted on wire? these fields are providing less context, token is kinda ok, what should the pool be?

do you anticipate the params for this func increasing, if yes, can this be abstracted by interface like what we did here https://github.com/red-hat-storage/ocs-operator/blob/main/services/provider/interfaces/interfaces.go

services/provider/server/cephrbdmirror.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+func newCephRBDMirrorManager(cl client.Client, namespace string) (*cephRBDMirrorManager, error) {
+	return &cephRBDMirrorManager{
+		client:    cl,
+		namespace: namespace,
+	}, nil
+}
+
+func (c *cephRBDMirrorManager) Create(ctx context.Context) error {
```
leelavg, line 27: any reason in not using creatorupdate from controller util?

services/provider/server/cephblockpool.go
```diff
@@ -0,0 +1,126 @@
...
+func newCephBlockPoolManager(cl client.Client, namespace string) (*cephBlockPoolManager, error) {
+	return &cephBlockPoolManager{
+		client:    cl,
+		namespace: namespace,
+	}, nil
+}
+
+func (c *cephBlockPoolManager) EnableBlockPoolMirroring(ctx context.Context, blockPoolName string) error {
+
+	cephBlockPool, err := c.get(ctx, blockPoolName)
```
leelavg, line 37: you don't need to get the resource, just fill the name, namespace and mutate the desired state in createorupdate.

services/provider/server/server.go
```diff
@@ -690,3 +705,34 @@ func (s *OCSProviderServer) ReportStatus(ctx context.Context, req *pb.ReportStat
 
 	return &pb.ReportStatusResponse{}, nil
 }
+
+// PeerBlockPool RPC call to send the bootstrap secret for the pool
+func (s *OCSProviderServer) PeerBlockPool(ctx context.Context, req *pb.PeerBlockPoolRequest) (*pb.PeerBlockPoolResponse, error) {
```
leelavg, line 710: either make this RPC idempotent or ensure different failure for resource already exists requests.

services/provider/server/cephblockpool.go
```diff
@@ -0,0 +1,126 @@
...
+		return nil
+	})
+	if err != nil {
+		return fmt.Errorf("failed to set bootstrap secret ref on CephBlockPool resource with name %q. %v", blockPoolName, err)
+	}
+
+	return nil
+}
+
+func (c *cephBlockPoolManager) get(ctx context.Context, blockPoolName string) (*rookCephv1.CephBlockPool, error) {
```
leelavg, line 116: func name could be better, like `getBlockPoolByName`?

services/provider/server/cephblockpool.go
```diff
@@ -0,0 +1,126 @@
...
+	if err != nil {
+		return err
+	}
+	secretNameMd5Sum := md5.Sum(secretJSON)
+	secretName := hex.EncodeToString(secretNameMd5Sum[:16])
+
+	gvk, err := apiutil.GVKForObject(cephBlockPool, c.client.Scheme())
+	if err != nil {
+		return fmt.Errorf("failed to get gvk for cephBlockPool %q. %w", cephBlockPool.Name, err)
+	}
```
leelavg, line 73: I don't think you need to extract this info, doesn't controller-runtime has utils for this? there should be some `SetControllerReference` 🤔

services/provider/server/server.go
```diff
@@ -690,3 +705,34 @@ func (s *OCSProviderServer) ReportStatus(ctx context.Context, req *pb.ReportStat
...
+		return nil, status.Errorf(codes.Internal, "Failed to enable mirroring for CephBlockPool resource %s: %v", req.BlockPoolName, err)
+	}
+
+	// create and set secret ref on the blockPool
+	if err := s.cephBlockPoolManager.SetBootstrapSecretRef(
+		ctx,
+		req.BlockPoolName,
+		map[string][]byte{
+			"pool":  req.Pool,
+			"token": req.Token,
```
leelavg, line 730: will you get a request w/ single pool name but different bytes on `pool` and `token`?

services/provider/server/server.go
```diff
@@ -690,3 +705,34 @@ func (s *OCSProviderServer) ReportStatus(ctx context.Context, req *pb.ReportStat
...
+func (s *OCSProviderServer) PeerBlockPool(ctx context.Context, req *pb.PeerBlockPoolRequest) (*pb.PeerBlockPoolResponse, error) {
+
+	if err := s.cephRBDMirrorManager.Create(ctx); err != nil {
+		return nil, status.Errorf(codes.Internal, "Failed to create RBDMirror instance: %v", err)
+	}
+
+	// enable mirroring on blockPool in the req
+	if err := s.cephBlockPoolManager.EnableBlockPoolMirroring(ctx, req.BlockPoolName); err != nil {
+		if kerrors.IsNotFound(err) {
+			return nil, status.Errorf(codes.NotFound, "Failed to find CephBlockPool resource %s: %v", req.BlockPoolName, err)
```
leelavg, line 719: I believe the errors are a bit misleading (same for remaining), basically the function asks for enabling mirroring and if you didn't find the source why are we at this line 🤔?

by source I mean, block pool, at the start why can't we return if we didn't find the blockpool? basically we are checking blockpool existence indirectly three times throughtout this function.

leelavg (commented):
> The rpc call is initiated from the primary cluster which wants to mirror the pool, the server receives the request and does the following
- a bit incomplete (at-least for me), by above do you mean the server is running in "another" cluster (maybe `secondary`?) and primary cluster wants to send over the info to secondary cluster?
- through out app lifecycle, does `primary` and `secondary` change identities and become the opposite? if yes, should we only use `server` and `client` terminology for all operations?


Let me know if you don't want review on draft PRs, I need to review this PR again as I'm not totally upto speed w/ RDR.

rewantsoni:
> * a bit incomplete (at-least for me), by above do you mean the server is running in "another" cluster (maybe secondary?) and primary cluster wants to send over the info to secondary cluster?
Yes right, we will have two clusters, one running the server(secondary) and other cluster as a client who sends the request (StorageClusterPeer; primary). The primary cluster sends over the contents of the bootstrap secret (that is pool and token fields) along with the blockPool Name it wants to mirror.

> * through out app lifecycle, do primary and secondary change identities and become the opposite? if yes, should we only use server and client terminology for all operations?
We are implementing peering in one-way replication, for RDR both primary and secondary clusters will act as both a server and a client, to be more specific, 
StorageClusterPeer on the Primary Cluster will be a client for the API Server on the Secondary Cluster and StorageClusterPeer on the Secondary Cluster will be a client for the API Server on Primary Cluster.


>Let me know if you don't want review on draft PRs, I need to review this PR again as I'm not totally upto speed w/ RDR.

Yes, I would like reviews on the draft PRs I have tested the PR. I need to complete unit tests and uninstall flow (probably as a part of a different PR).

services/provider/server/cephblockpool.go
rewantsoni: But that would mean that if a blockpool doesn't exist, it will create one. We don't want to create the blockpool

services/provider/client/client.go
rewantsoni: I will create a new PR for mTLS. Pool is the base64 encoded value for the blockPoolName, we can remove the blockPoolName and decode the pool when needed, WDYT?

No, I do think the params for PeerBlockPool will increase

services/provider/server/server.go
rewantsoni: Yes, when we have more than 2 clusters, the token will be different for different ceph clusters

services/provider/server/cephrbdmirror.go
rewantsoni: I was thinking about the case when we run MaintenanceMode (this requires the RBD mirror to be scaled down). If a new PeerBlockPool call is received while MaintenacneMode is in progress it and we use CreateOrUpdate, it will interfere with it.

services/provider/server/server.go
rewantsoni: updated to return at the beginning of the function if the blockpool doesn't exist

services/provider/server/cephblockpool.go
```diff
@@ -0,0 +1,112 @@
...
+		return err
+	}
+
+	// create the secret
+	secretJSON, err := json.Marshal(secretData)
+	if err != nil {
+		return err
+	}
+	secretNameMd5Sum := md5.Sum(secretJSON)
+	secretName := hex.EncodeToString(secretNameMd5Sum[:16])
```
rewantsoni, line 64: Instead of deriving a name from the secretInforamion, we could use md5Sum of `blockPoolName-<ClusterID>` which will be passed as a parameter to the grpc req

rewantsoni:
Test Results -
Before:
```
[rewantsoni ] date; oc get cephblockpool -oyaml

Mon Mar  4 19:10:59 IST 2024
apiVersion: v1
items:
- apiVersion: ceph.rook.io/v1
  kind: CephBlockPool
  metadata:
    name: ocs-storagecluster-cephblockpool
    namespace: openshift-storage
  spec:
    enableRBDStats: true
    erasureCoded:
      codingChunks: 0
      dataChunks: 0
    failureDomain: zone
    mirroring: {}
    quotas: {}
    replicated:
      replicasPerFailureDomain: 1
      size: 3
      targetSizeRatio: 0.49
    statusCheck:
      mirror: {}
  status:
    observedGeneration: 1
    phase: Ready
kind: List
metadata:
  resourceVersion: ""
  
[rewantsoni ] kubectl port-forward ocs-provider-server-79658954d9-8zjrg 8881:50051
Forwarding from 127.0.0.1:8881 -> 50051
Forwarding from [::1]:8881 -> 50051

[rewantsoni ] grpcurl -insecure -d '{"secretName":"ocs-storagecluster-cephblockpool-a96546a7-5e86-4d6e-aded","pool":".......=","token":"..........=="}' localhost:8881 provider.OCSProvider/PeerBlockPool
{}
```
After req:
```
[rewantsoni ] date; oc get cephblockpool -oyaml; oc get secret ocs-storagecluster-cephblockpool-a96546a7-5e86-4d6e-aded; oc get cephrbdmirror
Mon Mar  4 19:17:11 IST 2024
apiVersion: v1
items:
- apiVersion: ceph.rook.io/v1
  kind: CephBlockPool
  metadata:
    name: ocs-storagecluster-cephblockpool
    namespace: openshift-storage
  spec:
    enableRBDStats: true
    erasureCoded:
      codingChunks: 0
      dataChunks: 0
    failureDomain: zone
    mirroring:
      enabled: true
      mode: image
      peers:
        secretNames:
        - ocs-storagecluster-cephblockpool-a96546a7-5e86-4d6e-aded
    quotas: {}
    replicated:
      replicasPerFailureDomain: 1
      size: 3
      targetSizeRatio: 0.49
    statusCheck:
      mirror: {}
  status:
    info:
      rbdMirrorBootstrapPeerSecretName: pool-peer-token-ocs-storagecluster-cephblockpool
    mirroringInfo:
      lastChanged: "2024-03-04T13:46:01Z"
      lastChecked: "2024-03-04T13:47:01Z"
      mode: image
      peers:
      - client_name: client.rbd-mirror-peer
        direction: rx-tx
        site_name: a3bdcc3d-a544-40d9-a428-7d0f081ec343
        uuid: b6862733-cdb3-47c4-af2a-fef36a7fd55d
      site_name: 8d75c893-c988-48e5-8f12-77b64f82190b
    mirroringStatus:
      lastChecked: "2024-03-04T13:47:01Z"
      summary:
        daemon_health: OK
        health: OK
        image_health: OK
        states: {}
    observedGeneration: 3
    phase: Ready
    snapshotScheduleStatus: {}
kind: List
metadata:
  resourceVersion: ""
  
NAME                                                       TYPE     DATA   AGE
ocs-storagecluster-cephblockpool-a96546a7-5e86-4d6e-aded   Opaque   2      101s

NAME         PHASE
rbd-mirror   Ready
  ```

File: services/provider/server/cephblockpool.go [hunk 10]
```diff
@@ -0,0 +1,102 @@
...
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type cephBlockPoolManager struct {
+	client    client.Client
+	namespace string
+}
+
+func newCephBlockPoolManager(cl client.Client, namespace string) (*cephBlockPoolManager, error) {
```
rchikatw, line 22: I believe that this task can be completed without any errors in the return type.

services/provider/server/cephblockpool.go
```diff
@@ -0,0 +1,102 @@
...
+	cephBlockPool.Name = blockPoolName
+	cephBlockPool.Namespace = c.namespace
+
+	_, err := ctrl.CreateOrUpdate(ctx, c.client, cephBlockPool, func() error {
+		cephBlockPool.Spec.Mirroring.Enabled = true
+		cephBlockPool.Spec.Mirroring.Mode = "image"
+		return nil
+	})
+	if err != nil {
+		return fmt.Errorf("failed to enable mirroring on CephBlockPool resource with name %q. %v", blockPoolName, err)
```
rchikatw, line 41: return fmt.Errorf("failed to enable mirroring on CephBlockPool resource with name %q: %v", blockPoolName, err)

services/provider/server/cephblockpool.go
```diff
@@ -0,0 +1,102 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to set bootstrap secret ref on CephBlockPool resource with name %q. %v", blockPoolName, err)
+	}
+
+	return nil
+}
+
+func (c *cephBlockPoolManager) GetBlockPoolByName(ctx context.Context, blockPoolName string) (*rookCephv1.CephBlockPool, error) {
+	blockPoolObj := &rookCephv1.CephBlockPool{}
+	err := c.client.Get(ctx, types.NamespacedName{Name: blockPoolName, Namespace: c.namespace}, blockPoolObj)
```
leelavg, line 94: maybe we can use utils to construct the type?
``` suggestion
	blockPool := &rookCephv1.CephBlockPool{}
        blockPool.Name = blockPoolName
        blockPool.Namespace = c.namespace
	err := c.client.Get(ctx, client.ObjectKeyFromObject(blockPool), blockPool)
```
similar at remaining places, it just saves us from constructing a struct.

services/provider/server/cephblockpool.go
```diff
@@ -0,0 +1,102 @@
...
+	bootstrapSecret.Name = secretName
+	bootstrapSecret.Namespace = c.namespace
+
+	_, err = ctrl.CreateOrUpdate(ctx, c.client, bootstrapSecret, func() error {
+		bootstrapSecret.Data = secretData
+		err := ctrl.SetControllerReference(cephBlockPool, bootstrapSecret, c.client.Scheme())
+		if err != nil {
+			return err
+		}
+		return nil
```
leelavg, line 66: just `return ctrl.SetControllerReference....`

services/provider/server/server.go
leelavg, line 710: is there any chance that two clients call this rpc referring same block pool? if yes, the idea is to just fail one client & make it retry?

services/provider/server/cephrbdmirror.go
leelavg: fair enough, does the owner scale up after the maintenance mode?

services/provider/server/cephrbdmirror.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	cephRBDMirrorObj.Name = rBDMirrorName
+	cephRBDMirrorObj.Namespace = c.namespace
+	cephRBDMirrorObj.Spec = rookCephv1.RBDMirroringSpec{Count: 1}
+
+	err = c.client.Create(ctx, cephRBDMirrorObj)
+	if err != nil {
+		return err
+	}
```
leelavg, line 46: maybe always try to create and fail if error is other than `exists`? or do `if !kerrors.IsNotFound(err) { create }`

services/provider/client/client.go
leelavg: can be resolved.

services/provider/server/cephblockpool.go
leelavg: I see change made, can be resolved.

leelavg:
> I would like reviews on the draft PRs
- ack, I usually open all for reviews but if untested I put a `/hold` flag, just saying

can resolved my previous addressed comments, thanks.

services/provider/server/cephblockpool.go
leelavg: this piece of code is removed, isn't it?

rewantsoni: Yes it is removed

services/provider/server/cephrbdmirror.go
rewantsoni: Yes, that would be the responsibility of the controller executing the maintenance mode

services/provider/server/server.go
rewantsoni: Yes, right

leelavg:
@rewantsoni this PR was marked as a dependent on which a comment was made until which invalidates the changes in this PR, let me know if this still need a review, thanks.

services/provider/server/cephrbdmirror.go
```diff
@@ -0,0 +1,43 @@
...
+		namespace: namespace,
+	}, nil
+}
+
+func (c *cephRBDMirrorManager) Create(ctx context.Context) error {
+
+	cephRBDMirror := &rookCephv1.CephRBDMirror{}
+	cephRBDMirror.Name = rBDMirrorName
+	cephRBDMirror.Namespace = c.namespace
+	err := c.client.Get(ctx, client.ObjectKeyFromObject(cephRBDMirror), cephRBDMirror)
```
leelavg, line 30: any reason in making two calls rather than a single create?

services/provider/server/server.go
```diff
@@ -733,3 +748,34 @@ func (s *OCSProviderServer) getOCSSubscriptionChannel(ctx context.Context) (stri
...
+		return nil, status.Errorf(codes.Internal, "Failed to create RBDMirror instance: %v", err)
+	}
+
+	// enable mirroring on blockPool in the req
+	if err := s.cephBlockPoolManager.EnableBlockPoolMirroring(ctx, cephBlockPool); err != nil {
+		return nil, status.Errorf(codes.Internal, "Failed to enable mirroring for CephBlockPool resource %s: %v", req.Pool, err)
+	}
+
+	// create and set secret ref on the blockPool
+	if err := s.cephBlockPoolManager.SetBootstrapSecretRef(
```
leelavg, line 770: It's guaranteed that the call to SetBootstrap will fail first time since the resource version changes due to update in EnableBlock func. I believe we need to get fresh copy every time we want to update I guess.

services/provider/server/cephrbdmirror.go
rewantsoni: We need one rbdMirror for the cluster, hence I am checking if it already exists, and create it only if it doesn't

services/provider/server/server.go
rewantsoni: I will check it again

rewantsoni:
Closing this in favour of https://github.com/red-hat-storage/ocs-operator/pull/2671

PR #2458, opened 2024-02-13
Adds spec changes to enable TLS on external Postgresql server

vbnrh:
This commit adds changes to enable TLS on external postgresql server independent of whether the client passes certificates or not. Certificates are optional requirement and TLS can be enabled without them.

### DIFF
--- a/controllers/storagecluster/noobaa_system_reconciler.go
+++ b/controllers/storagecluster/noobaa_system_reconciler.go
@@ -201,11 +201,16 @@ func (r *StorageClusterReconciler) setNooBaaDesiredState(nb *nbv1.NooBaa, sc *oc
 		if sc.Spec.MultiCloudGateway.ExternalPgConfig != nil && sc.Spec.MultiCloudGateway.ExternalPgConfig.PGSecretName != "" {
 			nb.Spec.ExternalPgSecret = &corev1.SecretReference{Name: sc.Spec.MultiCloudGateway.ExternalPgConfig.PGSecretName, Namespace: sc.Namespace}
 			nb.Spec.ExternalPgSSLUnauthorized = sc.Spec.MultiCloudGateway.ExternalPgConfig.AllowSelfSignedCerts
-			if sc.Spec.MultiCloudGateway.ExternalPgConfig.TLSSecretName != "" {
-				nb.Spec.ExternalPgSSLRequired = true
+			nb.Spec.ExternalPgSSLRequired = sc.Spec.MultiCloudGateway.ExternalPgConfig.EnableTLS
+			if sc.Spec.MultiCloudGateway.ExternalPgConfig.EnableTLS && sc.Spec.MultiCloudGateway.ExternalPgConfig.TLSSecretName != "" {
 				nb.Spec.ExternalPgSSLSecret = &corev1.SecretReference{Name: sc.Spec.MultiCloudGateway.ExternalPgConfig.TLSSecretName, Namespace: sc.Namespace}
 			}
+
+			if !sc.Spec.MultiCloudGateway.ExternalPgConfig.EnableTLS && sc.Spec.MultiCloudGateway.ExternalPgConfig.TLSSecretName != "" {
+				return fmt.Errorf("Failed to create Noobaa system: tlsSecretName is a non-nil value while enableTls is disabled. Please set spec.multiCloudGateway.externalPgConfig.enableTls to true and retry again. enableTls: %v and tlsSecretName: %v", sc.Spec.MultiCloudGateway.ExternalPgConfig.EnableTLS, sc.Spec.MultiCloudGateway.ExternalPgConfig.TLSSecretName)
+			}
 		}
+
 	}
 
 	// Add KMS details to Noobaa spec, only if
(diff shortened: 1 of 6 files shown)


vbnrh:
Relevant UI PR - https://github.com/red-hat-storage/odf-console/pull/1227

vbnrh:
Relevant UI PR: https://github.com/red-hat-storage/odf-console/pull/1230

controllers/storagecluster/noobaa_system_reconciler.go
```diff
@@ -201,8 +201,8 @@ func (r *StorageClusterReconciler) setNooBaaDesiredState(nb *nbv1.NooBaa, sc *oc
 		if sc.Spec.MultiCloudGateway.ExternalPgConfig != nil && sc.Spec.MultiCloudGateway.ExternalPgConfig.PGSecretName != "" {
 			nb.Spec.ExternalPgSecret = &corev1.SecretReference{Name: sc.Spec.MultiCloudGateway.ExternalPgConfig.PGSecretName, Namespace: sc.Namespace}
 			nb.Spec.ExternalPgSSLUnauthorized = sc.Spec.MultiCloudGateway.ExternalPgConfig.AllowSelfSignedCerts
+			nb.Spec.ExternalPgSSLRequired = sc.Spec.MultiCloudGateway.ExternalPgConfig.EnableTLS
 			if sc.Spec.MultiCloudGateway.ExternalPgConfig.TLSSecretName != "" {
```
umangachapagain, line 205: What is the expected behavior when,

1. `enableTLS: false` but user provides `tlsSecretName`?
2. `enableTLS: true`, `allowSelfSignedCerts: false`, `tlsSecretName` is empty?
3. `enableTLS: true`, `allowSelfSignedCerts: true`, `tlsSecretName` is not empty?

jackyalbo: I think this can answer these questions: https://github.com/noobaa/noobaa-operator/blob/master/doc/external-postgres.md#ssl-support
1. externalPgSSLRequired and externalPgSSLUnauthorized are used for the server side, the 1st will force using ssl when talking to the external server, the 2nd will let the user connect to the server even if the client doesn't trust the CA.
2. externalPgSSLSecret will be used for the client side and will be used if the server wants the clients to connect with certificate instead of user/password

I do think that it's not allowed to use client-side secret without enabling TLS, but other than that the rest are fine. 
Hopes that clears it.

umangachapagain: > 
> I do think that it's not allowed to use client-side secret without enabling TLS, but other than that the rest are fine. Hopes that clears it.
>
@vbnrh We should add a validation for this case.

vbnrh: Added validation in the If statement.

Thanks @jackyalbo for the clarification.

controllers/storagecluster/noobaa_system_reconciler.go
```diff
@@ -201,8 +201,8 @@ func (r *StorageClusterReconciler) setNooBaaDesiredState(nb *nbv1.NooBaa, sc *oc
 		if sc.Spec.MultiCloudGateway.ExternalPgConfig != nil && sc.Spec.MultiCloudGateway.ExternalPgConfig.PGSecretName != "" {
 			nb.Spec.ExternalPgSecret = &corev1.SecretReference{Name: sc.Spec.MultiCloudGateway.ExternalPgConfig.PGSecretName, Namespace: sc.Namespace}
 			nb.Spec.ExternalPgSSLUnauthorized = sc.Spec.MultiCloudGateway.ExternalPgConfig.AllowSelfSignedCerts
-			if sc.Spec.MultiCloudGateway.ExternalPgConfig.TLSSecretName != "" {
-				nb.Spec.ExternalPgSSLRequired = true
+			nb.Spec.ExternalPgSSLRequired = sc.Spec.MultiCloudGateway.ExternalPgConfig.EnableTLS
+			if sc.Spec.MultiCloudGateway.ExternalPgConfig.EnableTLS && sc.Spec.MultiCloudGateway.ExternalPgConfig.TLSSecretName != "" {
 				nb.Spec.ExternalPgSSLSecret = &corev1.SecretReference{Name: sc.Spec.MultiCloudGateway.ExternalPgConfig.TLSSecretName, Namespace: sc.Namespace}
 			}
```
umangachapagain, line 207: We need to log error when validation fails or we won't know what went wrong. Even better to return an invalid config error.

umangachapagain (changes requested):

controllers/storagecluster/noobaa_system_reconciler.go
vbnrh: Its not an invalid config if it fails. It can still work with just having the TLS enabled and not have any client side authetication.

umangachapagain: But, if EnableTLS is false and TLSSecretName is provided, then it is an invalid config right? Shouldn't we let the users know about it?

vbnrh: It will be blocked in the UI itself. Only when the checkbox to "Enable TLS/SSL" is selected, then only users are allowed to pass the optional certs. So it will always be true.

Unless we want to add it for CLI purposes, in which case, this should be part of the webhook validations

umangachapagain: We can't depend on UI for validation and webhooks aren't available in ocs-operator.
An error log is the only option we have along with updates to status.conditions to explain what's wrong.

vbnrh: ok done with the changes then

umangachapagain (approved):

PR #2460, opened 2024-02-14
changing native client to runtime client

rchikatw:
- changing native client to runtime client because when the job functionality was implemented it was using only native resources now custom resources are also part of the job.
- Also, I replaced 'os.exit' with 'klog.exit' to make the code cleaner.

### DIFF
--- a/onboarding-validation-keys-generator/main.go
+++ b/onboarding-validation-keys-generator/main.go
@@ -5,7 +5,6 @@ import (
 	"crypto/rsa"
 	"crypto/x509"
 	"encoding/pem"
-	"os"
 
 	v1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
 	"github.com/red-hat-storage/ocs-operator/v4/controllers/util"
@@ -30,22 +29,19 @@ const (
 func main() {
 	cl, err := newClient()
 	if err != nil {
-		klog.Errorf("failed to create clientset: %v", err)
-		os.Exit(1)
+		klog.Exitf("failed to create client: %v", err)
 	}
 
 	ctx := context.Background()
 	operatorNamespace, err := util.GetOperatorNamespace()
 	if err != nil {
-		klog.Errorf("unable to get operator namespace: %v", err)
-		os.Exit(1)
+		klog.Exitf("unable to get operator namespace: %v", err)
 	}
 
 	// Generate RSA key.
 	privateKey, err := rsa.GenerateKey(rand.Reader, 4096)
 	if err != nil {
-		klog.Errorf("unable to generate private: %v", err)
-		os.Exit(1)
+		klog.Exitf("unable to generate private key: %v", err)
 	}
 
 	publicKey := &privateKey.PublicKey
@@ -69,8 +65,7 @@ func main() {
 	privateSecret.Namespace = operatorNamespace
 	err = cl.Delete(ctx, privateSecret)
 	if err != nil && !kerrors.IsNotFound(err) {
-		klog.Errorf("failed to delete private secret: %v", err)
-		os.Exit(1)
+		klog.Exitf("failed to delete private secret: %v", err)
 	}
 
 	// Delete public key secret
@@ -79,13 +74,12 @@ func main() {
 	publicSecret.Namespace = operatorNamespace
 	err = cl.Delete(ctx, publicSecret)
 	if err != nil && !kerrors.IsNotFound(err) {
-		klog.Errorf("failed to delete public secret: %v", err)
-		os.Exit(1)
+		klog.Exitf("failed to delete public secret: %v", err)
 	}
 
 	err = controllerutil.SetOwnerReference(storageCluster, privateSecret, cl.Scheme())
 	if err != nil {
-		klog.Exitf("failed to set owner reference for privateSecret : %v", err)
+		klog.Exitf("failed to set owner reference for private secret: %v", err)
 	}
 
 	privateSecret.StringData = map[string]string{
@@ -94,14 +88,12 @@ func main() {
 
 	err = cl.Create(ctx, privateSecret, &client.CreateOptions{})
 	if err != nil {
-		klog.Errorf("failed to create private secret: %v", err)
-		os.Exit(1)
+		klog.Exitf("failed to create private secret: %v", err)
 	}
 
 	err = controllerutil.SetOwnerReference(storageCluster, publicSecret, cl.Scheme())
 	if err != nil {
-		klog.Errorf("failed to create public secret: %v", err)
-		os.Exit(1)
+		klog.Exitf("failed to set owner reference for public secret: %v", err)
 	}
 	publicSecret.StringData = map[string]string{
 		"key": publicPem,
@@ -109,8 +101,7 @@ func main() {
 
 	err = cl.Create(ctx, publicSecret, &client.CreateOptions{})
 	if err != nil {
-		klog.Errorf("failed to create public secret: %v", err)
-		os.Exit(1)
+		klog.Exitf("failed to create public secret: %v", err)
 	}
 
 }
--- a/onboarding-validation-keys-generator/main.go
+++ b/onboarding-validation-keys-generator/main.go
@@ -4,19 +4,20 @@ import (
 	"crypto/rand"
 	"crypto/rsa"
 	"crypto/x509"
-	"encoding/json"
 	"encoding/pem"
-	"fmt"
 	"os"
 
+	v1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
 	"github.com/red-hat-storage/ocs-operator/v4/controllers/util"
 	"golang.org/x/net/context"
 	corev1 "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/kubernetes"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/klog/v2"
-	runtime "sigs.k8s.io/controller-runtime/pkg/client/config"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
 const (
@@ -27,7 +28,7 @@ const (
 )
 
 func main() {
-	clientset, err := newClient()
+	cl, err := newClient()
 	if err != nil {
 		klog.Errorf("failed to create clientset: %v", err)
 		os.Exit(1)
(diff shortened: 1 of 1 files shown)


rchikatw:
@nb-ohad As per the discussion over the other PR you have suggested keeping this change in seaprate PR. @leelavg

rchikatw:
@jarrpa

onboarding/main.go
```diff
@@ -57,90 +56,87 @@ func main() {
...
-	err = clientset.CoreV1().
-		Secrets(operatorNamespace).
-		Delete(ctx, onboardingValidationPrivateKeySecretName, metav1.DeleteOptions{})
+	privateSecret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      onboardingValidationPrivateKeySecretName,
+			Namespace: operatorNamespace,
+		},
+	}
+	err = cl.Delete(ctx, privateSecret)
```
nb-ohad, line 65: Just a suggestion for shorter code with less nesting:
```suggestion
	privateSecret := &corev1.Secret{}
	privateSecret.Name = onboardingValidationPrivateKeySecretName
	privateSecret.Namespace =  operatorNamespace
	err = cl.Delete(ctx, privateSecret)
```

onboarding/main.go
```diff
@@ -57,90 +56,87 @@ func main() {
...
 	// Delete public key secret
-	err = clientset.CoreV1().
-		Secrets(operatorNamespace).
-		Delete(ctx, onboardingValidationPublicKeySecretName, metav1.DeleteOptions{})
+	publicSecret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      onboardingValidationPublicKeySecretName,
+			Namespace: operatorNamespace,
+		},
+	}
```
nb-ohad, line 77: Similar suggestion as of the prev comment

onboarding/main.go
```diff
@@ -57,90 +56,87 @@ func main() {
...
-		StringData: map[string]string{
-			"key": privatePem,
+	privateSecret.ObjectMeta.OwnerReferences = []metav1.OwnerReference{
+		{
+			UID:        storageClusterMetadata.UID,
+			APIVersion: storageClusterMetadata.APIVersion,
+			Kind:       storageClusterMetadata.Kind,
+			Name:       storageClusterMetadata.Name,
 		},
 	}
```
nb-ohad, line 97: Why don't just use `SetOwnerReference` the same way we use in other places in this code base
(e.g. https://github.com/red-hat-storage/ocs-operator/blob/c8a0b8fb88016821f51605aad025f33552307a8b/controllers/storagecluster/provider_server.go#L140)

onboarding/main.go
```diff
@@ -57,90 +56,87 @@ func main() {
...
-		StringData: map[string]string{
-			"key": publicPem,
+	publicSecret.ObjectMeta.OwnerReferences = []metav1.OwnerReference{
+		{
+			UID:        storageClusterMetadata.UID,
+			APIVersion: storageClusterMetadata.APIVersion,
+			Kind:       storageClusterMetadata.Kind,
+			Name:       storageClusterMetadata.Name,
 		},
 	}
```
nb-ohad, line 115: Same comment as the previous one

onboarding/main.go
```diff
@@ -57,90 +56,87 @@ func main() {
...
 
-func newClient() (*kubernetes.Clientset, error) {
-	config := runtime.GetConfigOrDie()
-	clientset, err := kubernetes.NewForConfig(config)
+func newClient() (client.Client, error) {
+	klog.Info("Setting up k8s client")
+	config, err := config.GetConfig()
 	if err != nil {
-		return nil, err
+		klog.Exitf("Failed to get config: %v", err)
```
nb-ohad, line 132: Don't exist from within a util function. Exit should only be used at the top-level function (e.g. main). Return an error instead

onboarding/main.go
```diff
@@ -57,90 +56,87 @@ func main() {
...
+	}
+	k8sClient, err := client.New(config, client.Options{})
+	if err != nil {
+		klog.Exitf("Failed to create controller-runtime client: %v", err)
```
nb-ohad, line 136: Don't exist from within a util function. Exit should only be used at the top-level function (e.g. main). Return an error instead

onboarding/main.go
```diff
@@ -155,20 +151,15 @@ func convertRsaPublicKeyAsPemStr(publicKey *rsa.PublicKey) string {
...
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "storagecluster",
+			APIVersion: "ocs.openshift.io/v1",
+		},
 	}
-
-	if err = json.Unmarshal(storageClusterMetadataJSON, &storageClusterMetadata); err != nil {
-		return nil, fmt.Errorf("failed to parse storage cluster metadata response: %v", err)
+	if err := client.Get(ctx, types.NamespacedName{Name: storageClusterName, Namespace: operatorNamespace}, &storageClusterMetadata); err != nil {
+		return nil, err
```
nb-ohad, line 162: You are now using a client that can handle custom resources. Why can we just get the entire StorageCluster CR into StorageCluster struct? 

It will simplify the code and will make this entire function redundant

onboarding/main.go
```diff
@@ -27,9 +26,9 @@ const (
 )
 
 func main() {
-	clientset, err := newClient()
+	cl, err := newClient()
 	if err != nil {
-		klog.Errorf("failed to create clientset: %v", err)
+		klog.Errorf("failed to create client: %v", err)
 		os.Exit(1)
```
nb-ohad, line 32: You have access to `klog.Exit` and `klog.Exitf` use them instead of `os.Exit`.
Please apply the change to all instances of `os.Exist` in this file.

nb-ohad (changes requested):

onboarding/main.go
```diff
@@ -57,90 +54,76 @@ func main() {
...
 }
 
-func newClient() (*kubernetes.Clientset, error) {
-	config := runtime.GetConfigOrDie()
-	clientset, err := kubernetes.NewForConfig(config)
+func newClient() (client.Client, error) {
+	klog.Info("Setting up k8s client")
+	scheme := runtime.NewScheme()
+	if err := v1.AddToScheme(scheme); err != nil {
+		klog.Errorf("Failed to add v1 to scheme: %v", err)
```
nb-ohad, line 112: You should return here with an error. There is no point to log here, let the call site log the error

onboarding/main.go
```diff
@@ -57,90 +54,76 @@ func main() {
...
+	}
+	if err := corev1.AddToScheme(scheme); err != nil {
+		klog.Errorf("Failed to add corev1 to scheme: %v", err)
```
nb-ohad, line 115: You should return here with an error. There is no point to log here, let the call site log the error

onboarding/main.go
```diff
@@ -57,90 +54,76 @@ func main() {
...
+	}
+	config, err := config.GetConfig()
 	if err != nil {
-		return nil, err
+		klog.Errorf("Failed to get config: %v", err)
```
nb-ohad, line 119: You should return here with an error. There is no point to log here, let the call site log the error

onboarding/main.go
```diff
@@ -57,90 +54,76 @@ func main() {
...
+	}
+	k8sClient, err := client.New(config, client.Options{Scheme: scheme})
+	if err != nil {
+		klog.Errorf("Failed to create controller-runtime client: %v", err)
```
nb-ohad, line 123: You should return here with an error. There is no point to log here, let the call site log the error

onboarding/main.go
```diff
@@ -155,21 +138,11 @@ func convertRsaPublicKeyAsPemStr(publicKey *rsa.PublicKey) string {
...
-	if err = json.Unmarshal(storageClusterMetadataJSON, &storageClusterMetadata); err != nil {
-		return nil, fmt.Errorf("failed to parse storage cluster metadata response: %v", err)
+func getStorageCluster(ctx context.Context, operatorNamespace string, client client.Client) (*v1.StorageCluster, error) {
+	storageCluster := &v1.StorageCluster{}
+	if err := client.Get(ctx, types.NamespacedName{Name: storageClusterName, Namespace: operatorNamespace}, storageCluster); err != nil {
+		return nil, err
 	}
 
-	return &storageClusterMetadata, nil
+	return storageCluster, nil
```
nb-ohad, line 147: Why do we need this func if the function only executes a single Get command? Let's move the implementation to the call site.

nb-ohad (changes requested):

nb-ohad:
@rchikatw Does it make sense deleting the secrets before loading a storage cluster? isn't it more reasonable to have everything needed in place before applying any persistent changes?

rchikatw:
> @rchikatw Does it make sense deleting the secrets before loading a storage cluster? isn't it more reasonable to have everything needed in place before applying any persistent changes?

I can't think of a situation where this would occur, but to be on the safe side, I can make this change.

leelavg:
@rchikatw  I don't think forcing a retest solves this, if you have already run functests locally the issue might be at CI and better consult larger team as a quick look at current outstanding PRs states no PR is passing the CI from a week.

rchikatw:
> @rchikatw I don't think forcing a retest solves this, if you have already run functests locally the issue might be at CI and better consult larger team as a quick look at current outstanding PRs states no PR is passing the CI from a week.

make functest is running a successful
```
...
Ran 4 of 5 Specs in 388.809 seconds
SUCCESS! -- 4 Passed | 0 Failed | 1 Pending | 0 Skipped
PASS
ritesh:ocs-operator$ 
```

jarrpa (changes requested):
Though I like them, the klog changes seem irrelevant to the PR. At a minimum, move them into their own commit. It would be more ideal to have them as a separate PR, if you can manage it.

Otherwise, client changes look good.

nb-ohad:
> Though I like them, the klog changes seem irrelevant to the PR. At a minimum, move them into their own commit. It would > be more ideal to have them as a separate PR, if you can manage it.

@rchikatw I agree with @jarrpa. 
Let's separate these changes into their own commit then we can merge

jarrpa:
Commit split looks good, please also add a description of the change in the commit message itself. Something like this chunk of the PR description should suffice:

> when the job functionality was implemented it was using only native resources now custom resources are also part of the job.

A related question: why was the base k8s go client used in the first place instead of the runtime client? In general I tend to avoid client-go unless I'm doing something *real* weird, so I'm wondering if there was further rationale to it.

rchikatw:
> Commit split looks good, please also add a description of the change in the commit message itself. Something like this chunk of the PR description should suffice:
> 
> > when the job functionality was implemented it was using only native resources now custom resources are also part of the job.
> 
> A related question: why was the base k8s go client used in the first place instead of the runtime client? In general I tend to avoid client-go unless I'm doing something _real_ weird, so I'm wondering if there was further rationale to it.

In the beginning, we only used the native resources for operations(Get/Delete/Create). Instead of bringing in a heavyweight runtime library, we opted for client-go. No other reason

jarrpa:
> In the beginning, we only used the native resources for operations(Get/Delete/Create). Instead of bringing in a heavyweight runtime library, we opted for client-go. No other reason

Fair enough!

PR #2461, opened 2024-02-14
Added a fix for noobaa-standalone placement configuration

achouhan09:
Currently, default placement is used if noobaa-standalone placement is not found. We want to use noobaa-core if noobaa-standalone placement is not provided or default otherwise.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2260550

### DIFF
--- a/controllers/storagecluster/noobaa_system_reconciler.go
+++ b/controllers/storagecluster/noobaa_system_reconciler.go
@@ -11,6 +11,7 @@ import (
 	"github.com/red-hat-storage/ocs-operator/v4/controllers/util"
 	statusutil "github.com/red-hat-storage/ocs-operator/v4/controllers/util"
 	cephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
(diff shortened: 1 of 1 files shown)


controllers/storagecluster/placement.go
```diff
@@ -18,7 +18,13 @@ func getPlacement(sc *ocsv1.StorageCluster, component string) rookCephv1.Placeme
 	if ok {
 		(&in).DeepCopyInto(&placement)
 	} else {
-		in := defaults.DaemonPlacements[component]
+		// fallback to 'noobaa-core' if 'noobaa-standalone' placement not found, or use default placement.
+		if component == "noobaa-standalone" {
```
iamniting, line 22: I don't think we should add this logic here, This is a generic func, We should check in the calling site and call accordingly.

iamniting (changes requested):

controllers/storagecluster/placement.go
achouhan09: Thanks, sure will update accordingly.

iamniting (approved):

PR #2462, opened 2024-02-14
fix csv generation. 

obnoxxx:
The csv-merger tool contained code intended to replace allegedly unnecessary single quotes with double quotes in component crds.

This could sometimes result in broken csvs where ending quotes were simply removed

This change fixes these breakages by removing the culprit replacement code.

### DIFF
--- a/tools/csv-merger/csv-merger.go
+++ b/tools/csv-merger/csv-merger.go
@@ -461,10 +461,7 @@ func marshallObject(obj interface{}, writer io.Writer, modifyUnstructuredFunc fu
 		return err
 	}
 
-	// fix double quoted strings by removing unneeded single quotes...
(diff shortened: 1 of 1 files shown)


BlaineEXE:
This looks fine to me, and since CI is passing, I think it must be a good sign. However, I think we should get final approval for merge from someone who works more regularly with ocs-operator before merging.

I presume it's fine to merge this into master but that it'll need a BZ to backport to a release version?

b-ranto (approved):
This looks good downstream, I did run the bundle generation downstream script with this PR and it fixed the syntax error we were seeing. I also did not see any issues with double-quoting anywhere.

BlaineEXE (approved):

PR #2469, opened 2024-02-20
UX: implement rotating keys api in ux-backend server

leelavg:
Implement rotating keys by deleting the secret containing onboarding public key. Creation of new private & public key pair when public key is deleted is already implemented in #2385.

The expectation is this new api endpoint be called via CLI or UI and get the existing private & public key pairs rotated if both the keys are being managed by us.

Part of [RHSTOR-5489]

### DIFF
--- a/services/ux-backend/handlers/rotatekeys/handler.go
+++ b/services/ux-backend/handlers/rotatekeys/handler.go
@@ -1,0 +1,62 @@
+package rotatekeys
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	"github.com/red-hat-storage/ocs-operator/v4/services/ux-backend/handlers"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	onboardingValidationPublicKeySecretName = "onboarding-ticket-key"
+)
+
+func HandleMessage(w http.ResponseWriter, r *http.Request, cl client.Client) {
+	switch r.Method {
+	case "POST":
+		handlePost(r.Context(), w, cl)
+	default:
+		handleUnsupportedMethod(w, r)
+	}
+}
+
+func handlePost(ctx context.Context, w http.ResponseWriter, cl client.Client) {
+	klog.Info("POST method on /rotate-keys endpoint is invoked")
+	w.Header().Set("Content-Type", handlers.ContentTypeTextPlain)
+
+	publicKeySecret := &corev1.Secret{}
+	publicKeySecret.Name = onboardingValidationPublicKeySecretName
+	publicKeySecret.Namespace = handlers.GetPodNamespace()
+	err := cl.Delete(ctx, publicKeySecret)
+	if err != nil {
+		klog.Errorf("failed to delete public key secret: %v", err)
+		w.WriteHeader(http.StatusInternalServerError)
+
+		// TODO: should we differentiate b/n secret not found and remaining errors?
+		if _, err = w.Write([]byte("Failed to rotate keys")); err != nil {
+			klog.Errorf("failed to write data to response writer, %v", err)
+		}
+		return
+	}
+
+	klog.Info("onboarding validation keys are rotated successfully")
+	w.WriteHeader(http.StatusOK)
+	if _, err = w.Write([]byte("Successfully rotated keys")); err != nil {
+		klog.Errorf("failed to write data to response writer, %v", err)
+	}
+}
+
+func handleUnsupportedMethod(w http.ResponseWriter, r *http.Request) {
+	klog.Info("Only POST method should be used to send data to this endpoint /rotate-keys")
+	w.WriteHeader(http.StatusMethodNotAllowed)
+	w.Header().Set("Content-Type", handlers.ContentTypeTextPlain)
+	w.Header().Set("Allow", "POST")
+
+	if _, err := w.Write([]byte(fmt.Sprintf("Unsupported method : %s", r.Method))); err != nil {
+		klog.Errorf("failed to write data to response writer: %v", err)
+	}
+}
--- a/services/ux-backend/main.go
+++ b/services/ux-backend/main.go
@@ -10,6 +10,9 @@ import (
 	"k8s.io/klog/v2"
 
 	"github.com/red-hat-storage/ocs-operator/v4/services/ux-backend/handlers/onboardingtokens"
+	"github.com/red-hat-storage/ocs-operator/v4/services/ux-backend/handlers/rotatekeys"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
 )
 
 type serverConfig struct {
@@ -51,6 +54,20 @@ func loadAndValidateServerConfig() (*serverConfig, error) {
 	return &config, nil
 }
 
+func newKubeClient() (client.Client, error) {
+	cfg, err := config.GetConfig()
+	if err != nil {
+		return nil, err
+	}
+
+	newClient, err := client.New(cfg, client.Options{})
+	if err != nil {
+		return nil, err
+	}
+
+	return newClient, nil
+}
+
 func main() {
 
 	klog.Info("Starting ux backend server")
@@ -61,8 +78,18 @@ func main() {
 		klog.Info("shutting down!")
 		os.Exit(-1)
 	}
+
+	cl, err := newKubeClient()
+	if err != nil {
+		klog.Errorf("failed to create kubernetes api client: %v", err)
+		klog.Exit("shutting down!")
+	}
+
 	http.HandleFunc("/onboarding-tokens", func(w http.ResponseWriter, r *http.Request) {
 		onboardingtokens.HandleMessage(w, r, config.tokenLifetimeInHours)
+	})
+	http.HandleFunc("/rotate-keys", func(w http.ResponseWriter, r *http.Request) {
+		rotatekeys.HandleMessage(w, r, cl)
 	})
 
 	klog.Info("ux backend server listening on port ", config.listenPort)
--- a/deploy/ocs-operator/manifests/ocs-operator.clusterserviceversion.yaml
+++ b/deploy/ocs-operator/manifests/ocs-operator.clusterserviceversion.yaml
@@ -3294,6 +3294,10 @@ spec:
                 - name: ONBOARDING_TOKEN_LIFETIME
                 - name: UX_BACKEND_PORT
                 - name: TLS_ENABLED
+                - name: POD_NAMESPACE
+                  valueFrom:
+                    fieldRef:
+                      fieldPath: metadata.namespace
                 image: quay.io/ocs-dev/ocs-operator:latest
                 imagePullPolicy: IfNotPresent
                 name: ux-backend-server
--- a/deploy/ocs-operator/manifests/ux_backend_role.yaml
+++ b/deploy/ocs-operator/manifests/ux_backend_role.yaml
@@ -14,3 +14,11 @@ rules:
   verbs:
     - get
     - list
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  resourceNames:
+  - onboarding-ticket-key
+  verbs:
+  - delete
--- a/rbac/ux_backend_role.yaml
+++ b/rbac/ux_backend_role.yaml
@@ -14,3 +14,11 @@ rules:
   verbs:
     - get
     - list
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  resourceNames:
+  - onboarding-ticket-key
+  verbs:
+  - delete
--- a/services/ux-backend/handlers/common.go
+++ b/services/ux-backend/handlers/common.go
@@ -1,0 +1,23 @@
+package handlers
+
+import (
+	"os"
+)
+
+const (
+	ContentTypeTextPlain = "text/plain"
+)
+
+var namespace string
+
+// returns namespace found in env value, will panic if value is empty
+func GetPodNamespace() string {
+	if namespace != "" {
+		return namespace
+	}
+	if ns := os.Getenv("POD_NAMESPACE"); ns != "" {
+		namespace = ns
+		return namespace
+	}
+	panic("Value for env var 'POD_NAMESPACE' is empty")
+}
--- a/services/ux-backend/handlers/onboardingtokens/handler.go
+++ b/services/ux-backend/handlers/onboardingtokens/handler.go
@@ -16,12 +16,12 @@ import (
 
 	"github.com/google/uuid"
 	"github.com/red-hat-storage/ocs-operator/v4/services"
+	"github.com/red-hat-storage/ocs-operator/v4/services/ux-backend/handlers"
 	"k8s.io/klog/v2"
 )
 
 const (
 	onboardingPrivateKeyFilePath = "/etc/private-key/key"
-	ContentTypeTextPlain         = "text/plain"
 )
 
 func HandleMessage(w http.ResponseWriter, r *http.Request, tokenLifetimeInHours int) {
@@ -37,7 +37,7 @@ func handlePost(w http.ResponseWriter, tokenLifetimeInHours int) {
 	if onboardingToken, err := generateOnboardingToken(tokenLifetimeInHours); err != nil {
 		klog.Errorf("failed to get onboardig token: %v", err)
 		w.WriteHeader(http.StatusInternalServerError)
-		w.Header().Set("Content-Type", ContentTypeTextPlain)
+		w.Header().Set("Content-Type", handlers.ContentTypeTextPlain)
 
 		if _, err := w.Write([]byte("Failed to generate token")); err != nil {
 			klog.Errorf("failed write data to response writer, %v", err)
@@ -45,7 +45,7 @@ func handlePost(w http.ResponseWriter, tokenLifetimeInHours int) {
 	} else {
 		klog.Info("onboarding token generated successfully")
 		w.WriteHeader(http.StatusOK)
-		w.Header().Set("Content-Type", ContentTypeTextPlain)
+		w.Header().Set("Content-Type", handlers.ContentTypeTextPlain)
 
 		if _, err = w.Write([]byte(onboardingToken)); err != nil {
 			klog.Errorf("failed write data to response writer: %v", err)
@@ -56,7 +56,7 @@ func handlePost(w http.ResponseWriter, tokenLifetimeInHours int) {
 func handleUnsupportedMethod(w http.ResponseWriter, r *http.Request) {
 	klog.Info("Only POST method should be used to send data to this endpoint /onboarding-tokens")
 	w.WriteHeader(http.StatusMethodNotAllowed)
-	w.Header().Set("Content-Type", ContentTypeTextPlain)
+	w.Header().Set("Content-Type", handlers.ContentTypeTextPlain)
 	w.Header().Set("Allow", "POST")
 
 	if _, err := w.Write([]byte(fmt.Sprintf("Unsupported method : %s", r.Method))); err != nil {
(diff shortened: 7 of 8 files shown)


leelavg:
/hold

testing in progress.

services/ux-backend/handlers/rotatekeys/handler.go
```diff
@@ -0,0 +1,89 @@
...
+	w.Header().Set("Content-Type", contentTypeTextPlain)
+
+	privateKeySecret := &corev1.Secret{}
+	privateKeySecret.Name = onboardingValidationPrivateKeySecretName
+	privateKeySecret.Namespace = namespace
+	err := cl.Get(ctx, client.ObjectKeyFromObject(privateKeySecret), privateKeySecret)
+	if err != nil {
+		if kerrors.IsNotFound(err) {
+			klog.Info("onboarding private key secret does not exist")
+			w.WriteHeader(http.StatusBadRequest)
```
nb-ohad, line 40: Why is that a `BadRequest`? Is there anything in the request itself that invalidates it? 
To me, this seems more like a `StatusInternalServerError` situation

services/ux-backend/handlers/rotatekeys/handler.go
```diff
@@ -0,0 +1,89 @@
...
+
+			// we cannot rotate keys due to any of:
+			// 1. we are not running in provider mode
+			// 2. keys are not being managed by us via onboarding job
```
nb-ohad, line 44: We should allow the rotation of keys even if we do not manage them. 
That is the automated way to move from customer-owned keys to ODF-owned keys.

services/ux-backend/main.go
```diff
@@ -48,9 +52,30 @@ func loadAndValidateServerConfig() (*serverConfig, error) {
 		return nil, err
 	}
 
+	config.namespace, err = readEnvVar("POD_NAMESPACE", "openshift-storage", func(str string) (string, error) {
+		return str, nil
+	})
+	if err != nil {
+		return nil, err
+	}
```
nb-ohad, line 60: Namespace does not make sense as a server config. It has no default value and is not user-configured. 
Not every piece of information coming from an Env var should be considered a server config.

services/ux-backend/main.go
```diff
@@ -61,9 +86,19 @@ func main() {
...
+		klog.Errorf("failed to create kubernetes api client: %v", err)
+		klog.Exit("shutting down!")
+	}
+
 	http.HandleFunc("/onboarding-tokens", func(w http.ResponseWriter, r *http.Request) {
 		onboardingtokens.HandleMessage(w, r, config.tokenLifetimeInHours)
 	})
+	http.HandleFunc("/rotate-keys", func(w http.ResponseWriter, r *http.Request) {
+		rotatekeys.HandleMessage(w, r, cl, config.namespace)
+	})
```
nb-ohad, line 101: Why send the namespace and not just have it as an exported global? 
Namespace is a property of the deployment itself, are we expecting different values for dif invocations?

nb-ohad (changes requested):

services/ux-backend/handlers/rotatekeys/handler.go
leelavg: Based on another comment, this will cease to be a `BadRequest`. My initial intention is this ep shouldn't be triggered for customer managed public keys.

services/ux-backend/main.go
leelavg: Totally agreed, I also felt this isn't necessarily a server param but we can argue the same for lifetime as that also belongs to a specific endpoint and so just used this config struct.

services/ux-backend/main.go
leelavg: Unfortunately I failed to identify the place where this needs to be exported and so ended up passing as a param.

Let me rethink.

services/ux-backend/handlers/rotatekeys/handler.go
nb-ohad: But why do we cate if the private key exists or not? or even if the public key exists or not?

services/ux-backend/main.go
nb-ohad: > but we can argue the same for lifetime as that also belongs to a specific endpoint and so just used this config struct.

I don't agree. `tokenLifetimeInHours` is user-defined and has a default value. It does not matter if the config is effecting 1 endpoint, 5 endpoints or all endpoints. That is not the criteria.  

This about which type of information will go into a configmap when/if we decide to move the server config into a configmap

services/ux-backend/handlers/rotatekeys/handler.go
leelavg: If public key was customer managed then we should check for private key existence and then only remove the public key.

However, based on your earlier comment and the fact that we also need to generate token for setting native client this endpoint should just delete the public key.

services/ux-backend/main.go
leelavg: > That is not the criteria.
- Yes, what I mentioned above was, I used a wrong criteria to arrive at reusing the server struct for storing namespace which I'll be removing in next refresh.

services/ux-backend/handlers/rotatekeys/handler.go
nb-ohad: I am not sure why we need to delete the public kay and no just trigger the job. What will be the benefit of deleting the public key? It will just requeue a complete reconcile of the storagecluster.

leelavg: I don't think we can trigger job as we aren't using cronjob or it isn't a template.

I understand the intention though and at the same time, what's more alarming, reconciles in quick succession or rotating keys w/o any delays?

services/ux-backend/handlers/rotatekeys/handler.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+func handlePost(ctx context.Context, w http.ResponseWriter, cl client.Client, namespace string) {
+
+	w.Header().Set("Content-Type", contentTypeTextPlain)
+
+	privateKeySecret := &corev1.Secret{}
+	privateKeySecret.Name = onboardingValidationPrivateKeySecretName
+	privateKeySecret.Namespace = namespace
+	err := cl.Get(ctx, client.ObjectKeyFromObject(privateKeySecret), privateKeySecret)
```
rchikatw, line 36: Why aren't we directly deleting the secret option instead of retrieving it? I assume this secret is only present when odf is installed in provider mode. Correct me if I am wrong

leelavg: well, I thought private key will not exist if public key is being managed by user and so this endpoint shouldn't be called.

as per https://github.com/red-hat-storage/ocs-operator/pull/2469#discussion_r1495874806 calling this ep seems to be opting into keys to be managed by odf.

rchikatw: Okay, I understand now. I apologize for missing that comment. So, are you going to remove this check

services/ux-backend/main.go
leelavg: done.

services/ux-backend/handlers/rotatekeys/handler.go
leelavg: done.

services/ux-backend/handlers/rotatekeys/handler.go
```diff
@@ -0,0 +1,89 @@
...
+	switch r.Method {
+	case "POST":
+		handlePost(r.Context(), w, cl, namespace)
+	default:
+		handleUnsupportedMethod(w, r)
+	}
+}
+
+func handlePost(ctx context.Context, w http.ResponseWriter, cl client.Client, namespace string) {
+
```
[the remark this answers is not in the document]
leelavg: done.

services/ux-backend/handlers/rotatekeys/handler.go
leelavg: > I don't think we can trigger job
- awaiting response

services/ux-backend/main.go
leelavg: done.

services/ux-backend/handlers/rotatekeys/handler.go
leelavg: np, done.

leelavg:
/unhold

tested, working as expected
```
# ===== Existing
# ko get secret onboarding-private-key -ojsonpath='{.metadata.creationTimestamp}'
2024-02-28T05:20:07Z

# ko get secret onboarding-ticket-key -ojsonpath='{.metadata.creationTimestamp}'
2024-02-28T05:20:07Z

# ===== Invoked rotation
# ko exec -it deploy/ocs-operator -- bash -c 'curl -H "Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)"  https://ux-backend-proxy:8888/rotate-keys -k -X POST -s'
Successfully rotated keys

# ko logs ux-backend-server-5b7478645c-gbwpn -f
I0228 05:53:09.151963       1 handler.go:34] POST method on /rotate-keys endpoint is invoked
I0228 05:53:09.166107       1 handler.go:65] onboarding keys are rotated successfully

# ===== Rotated
# ko get secret onboarding-private-key -ojsonpath='{.metadata.creationTimestamp}'
2024-02-28T05:53:26Z

# ko get secret onboarding-ticket-key -ojsonpath='{.metadata.creationTimestamp}'
2024-02-28T05:53:26Z
```

services/ux-backend/main.go
```diff
@@ -61,9 +78,19 @@ func main() {
 		klog.Info("shutting down!")
 		os.Exit(-1)
```
rchikatw, line 79: Let's replace all instances of os.exit with klog.Exit. If you disagree, it's okay.

services/ux-backend/handlers/rotatekeys/handler.go
```diff
@@ -0,0 +1,90 @@
...
+		w.WriteHeader(http.StatusInternalServerError)
+
+		// TODO: should we differentiate b/n secret not found and remaining errors?
+		if _, err = w.Write([]byte("Failed to rotate keys")); err != nil {
+			klog.Errorf("failed to write date to response writer, %v", err)
+		}
+		return
+	}
+
+	klog.Info("onboarding keys are rotated successfully")
```
rchikatw, line 65: klog.Info("onboarding validation keys are rotated successfully")

services/ux-backend/handlers/rotatekeys/handler.go
```diff
@@ -0,0 +1,90 @@
...
+	"fmt"
+	"net/http"
+	"os"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
```
rchikatw, line 14: Can we declare this constant globally since it will be needed in multiple handlers?

services/ux-backend/handlers/rotatekeys/handler.go
```diff
@@ -0,0 +1,90 @@
...
+	w.Header().Set("Allow", "POST")
+
+	if _, err := w.Write([]byte(fmt.Sprintf("Unsupported method : %s", r.Method))); err != nil {
+		klog.Errorf("failed to write data to response writer: %v", err)
+	}
+}
+
+// getNamespace returns current namespace that the pod is running
+func getNamespace() (string, error) {
+	f, err := os.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
```
rchikatw, line 85: Can we change the variable name 'f' to 'podNamespace' or any other appropriate name you think?

nb-ohad, line 85: Why are we reading from a file? what was wrong with reading from an envvar? The request from the prev review was to not include this in the server config struct, not to move to read from a file.

services/ux-backend/main.go
nb-ohad: I agree, but unless Leela has introduced new usages of `os.Exit`, this change shouldn't be introduced into this specific PR

services/ux-backend/handlers/rotatekeys/handler.go
nb-ohad: Maybe in a `utils.go` or `constants.go` file directly under the `handlers` package

services/ux-backend/handlers/rotatekeys/handler.go
```diff
@@ -0,0 +1,90 @@
...
+	w.Header().Set("Content-Type", contentTypeTextPlain)
+
+	if namespace == "" {
+		var err error
+		namespace, err = getNamespace()
+		if err != nil {
+			klog.Errorf("failed to get current namespace: %v", err)
+			w.WriteHeader(http.StatusInternalServerError)
+			if _, err = w.Write([]byte("Failed to rotate keys")); err != nil {
+				klog.Errorf("failed to write date to response writer, %v", err)
```
nb-ohad, line 44: I don't really understand this error message. Can you please explain why we are talking about dates here?

services/ux-backend/handlers/rotatekeys/handler.go
```diff
@@ -0,0 +1,90 @@
...
+		if _, err = w.Write([]byte("Failed to rotate keys")); err != nil {
+			klog.Errorf("failed to write date to response writer, %v", err)
+		}
+		return
+	}
+
+	klog.Info("onboarding keys are rotated successfully")
+	w.WriteHeader(http.StatusOK)
+	if _, err = w.Write([]byte("Successfully rotated keys")); err != nil {
+		klog.Errorf("failed to write date to response writer, %v", err)
```
nb-ohad, line 68: Same as the last comment I had about the error message

nb-ohad (changes requested):

services/ux-backend/handlers/rotatekeys/handler.go
nb-ohad: @leelavg I still think we don't need to relay on a reconcile to trigger rotation. But for the sake of progress, let go with the current approach while having you open a tech debt capturing the requested change and having the link attached to this PR

leelavg: done. https://github.com/red-hat-storage/ocs-operator/issues/2485

services/ux-backend/main.go
leelavg: yes, not edited. resolving thread.

services/ux-backend/handlers/rotatekeys/handler.go
leelavg: removed the func, resolving.

services/ux-backend/handlers/rotatekeys/handler.go
leelavg: done. resolving.

services/ux-backend/handlers/rotatekeys/handler.go
leelavg: done.

services/ux-backend/handlers/rotatekeys/handler.go
leelavg: done.

services/ux-backend/handlers/rotatekeys/handler.go
leelavg: thanks. type, fixed.

services/ux-backend/handlers/rotatekeys/handler.go
leelavg: thanks. typo, fixed.

services/ux-backend/handlers/constants.go
```diff
@@ -0,0 +1,8 @@
+package handlers
+
+// this file contains that constants that are common across all handlers
+
+const (
+	ContentTypeTextPlain                    = "text/plain"
+	OnboardingValidationPublicKeySecretName = "onboarding-ticket-key"
```
nb-ohad, line 7: Why do we need this here? Do we have multiple handlers using this constant?

services/ux-backend/handlers/rotatekeys/handler.go
```diff
@@ -0,0 +1,71 @@
...
+	onboardingValidationPublicKeySecretName = "onboarding-ticket-key"
+	podNamespaceEnvVar                      = "POD_NAMESPACE"
+)
+
+var namespace string = func() string {
+	if ns := os.Getenv(podNamespaceEnvVar); ns != "" {
+		return ns
+	}
+	panic(fmt.Sprintf("Value for env var %q is not set", podNamespaceEnvVar))
+}()
```
nb-ohad, line 25: Can we have a func in `handlers/utils.go` (instead of the `constants.go` file, to expose the operator namespace?
We can calc it once and store it in a local var within the handler module/

services/ux-backend/handlers/rotatekeys/handler.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+}
+
+func handlePost(ctx context.Context, w http.ResponseWriter, cl client.Client) {
+	klog.Info("POST method on /rotate-keys endpoint is invoked")
+	w.Header().Set("Content-Type", handlers.ContentTypeTextPlain)
+
+	publicKeySecret := &corev1.Secret{}
+	publicKeySecret.Name = onboardingValidationPublicKeySecretName
+	publicKeySecret.Namespace = namespace
```
nb-ohad, line 42:
```suggestion
	publicKeySecret.Namespace = handlers.GetOperatorNamespace()
```

nb-ohad (changes requested):

services/ux-backend/handlers/constants.go
leelavg: my mistake, missed it.

services/ux-backend/handlers/rotatekeys/handler.go
leelavg: done.

services/ux-backend/handlers/rotatekeys/handler.go
leelavg: done.

jarrpa (commented):
Code looks good. If you get the chance, it would be good to update the commit message of `ux: implement rotating keys api` with a portion of the PR message describing how and when this API call is intended to be used.

services/ux-backend/handlers/constants.go
nb-ohad, line 1: There is no need to have both `constants.go` and `utils.go` under the handlers package 
A constants file only make sense if we have dozens or more exported constants. Right now we have a single exported/public constant.

 Let's fold this file into `utils.go` and rename `utils.go` to `common.go`

services/ux-backend/handlers/utils.go
```diff
@@ -0,0 +1,20 @@
...
+)
+
+var namespace string
+
+// returns namespace found in env value, will panic if value is empty
+func GetOperatorNamespace() string {
+	if namespace != "" {
+		return namespace
+	}
+	if ns := os.Getenv(podNamespaceEnvVar); ns != "" {
```
nb-ohad, line 15: This env var is only used once, in that case, can we please use a literal string instead?
It makes it easier to understand exactly what the code does without looking for the constant definition to read its value.

services/ux-backend/handlers/utils.go
```diff
@@ -0,0 +1,20 @@
...
+
+import (
+	"fmt"
+	"os"
+)
+
+var namespace string
+
+// returns namespace found in env value, will panic if value is empty
+func GetOperatorNamespace() string {
```
nb-ohad, line 11: This is not an operator so why do we call the function `GetOperatorNamespace`? 
```suggestion
func GetPodNamespace() string {
```

nb-ohad (changes requested):

leelavg:
> it would be good to update the commit message
- done.

services/ux-backend/handlers/utils.go
leelavg: yes, done.

services/ux-backend/handlers/constants.go
leelavg: yes, done.

PR #2478, opened 2024-02-23
storageclassrequest: reconcile RADOS Namespace for blockpool request

jarrpa:
Signed-off-by: Jose A. Rivera jarrpa@redhat.com


services/provider/server/server.go
```diff
@@ -543,6 +543,10 @@ func (s *OCSProviderServer) GetStorageClassClaimConfig(ctx context.Context, req
 		}
 	}
 	var extR []*pb.ExternalResource
+
+	var storageClassName string
+	storageClassData := map[string]string{}
```
leelavg, line 548: existing code is inlining the entries, maybe it doesn't hurt to define this map w/ length of 10?

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -168,6 +177,7 @@ func (r *StorageClassRequestReconciler) SetupWithManager(mgr ctrl.Manager) error
 			predicate.GenerationChangedPredicate{},
 		)).
 		Watches(&rookCephv1.CephBlockPool{}, enqueueForOwner).
+		Watches(&rookCephv1.CephBlockPoolRadosNamespace{}, enqueueForOwner).
```
leelavg, line 180: quick check, do we already have rbac for this?

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -238,17 +259,51 @@ func (r *StorageClassRequestReconciler) initPhase(storageProfile *v1.StorageProf
 			// if we found more than one CephBlockPool, we can't determine which one to select, so error out
 			cbpItemsLen := len(cephBlockPoolList.Items)
 			if cbpItemsLen == 0 {
-				cbpNewName := fmt.Sprintf("cephblockpool-%s-%s", r.storageConsumer.Name, generateUUID())
+				cbpNewName := fmt.Sprintf("cephblockpool-profile-%s", storageProfile.Name)
```
leelavg, line 262: am I wrong in assuming that we still need to create storageprofile resource even after this PR gets merged?

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -44,6 +44,7 @@ const (
 	StorageConsumerAnnotation     = "ocs.openshift.io.storageconsumer"
 	StorageRequestAnnotation      = "ocs.openshift.io.storagerequest"
 	StorageCephUserTypeAnnotation = "ocs.openshift.io.cephusertype"
+	StorageProfileLabel           = "ocs.openshift.io/storageprofile"
```
leelavg, line 47: I don't see this being used anywhere in the PR.

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -246,17 +259,51 @@ func (r *StorageClassRequestReconciler) initPhase(storageProfile *v1.StorageProf
...
 				r.cephBlockPool.Name = cephBlockPoolList.Items[0].GetName()
 				r.log.V(1).Info("valid CephBlockPool found", "CephBlockPool", r.cephBlockPool.Name)
 			} else {
-				return fmt.Errorf("invalid number of CephBlockPools for storage consumer %q and storage profile %q: found %d, expecting 0 or 1", r.storageConsumer.Name, storageProfile.Name, cbpItemsLen)
+				return fmt.Errorf("invalid number of CephBlockPools for storage profile %q: found %d, expecting 0 or 1", storageProfile.Name, cbpItemsLen)
+			}
+		}
+
+		// check if a CephBlockPoolRadosNamespace resource exists for the desired storageconsumer and storageprofile.
+		if r.cephRadosNamespace.Name == "" {
```
nb-ohad, line 274: This is the wrong check (see a previous comment), let's remove it.

controllers/storageconsumer/storageconsumer_controller.go
nb-ohad: I believe we agreed to have this as an annotation not as a label.

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -246,17 +259,51 @@ func (r *StorageClassRequestReconciler) initPhase(storageProfile *v1.StorageProf
...
+			if err != nil {
+				return err
+			}
+
+			// if we found no CephBlockPoolRadosNamespaces, generate a new name
+			// if we found only one CephBlockPoolRadosNamespace with our query, we're good
+			// if we found more than one CephBlockPoolRadosNamespace, we can't determine which one to select, so error out
+			rnsItemsLen := len(cephRadosNamespaceList.Items)
+			if rnsItemsLen == 0 {
+				rnsNewName := fmt.Sprintf("cephradosnamespace-%s-%s", r.cephBlockPool.Name, r.storageConsumer.Name)
```
nb-ohad, line 290: With this approach you can hit the name length limit. To solve this we usually use a simple hash for the different parts. So the resulting name will be `cephradosnamespace-<has from blockpool name and storage consumer name>`

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -425,6 +476,41 @@ func (r *StorageClassRequestReconciler) reconcileCephBlockPool(storageProfile *v
 	return nil
 }
 
+func (r *StorageClassRequestReconciler) reconcileRadosNamespace(storageProfile *v1.StorageProfile) error {
+	_, err := ctrl.CreateOrUpdate(r.ctx, r.Client, r.cephRadosNamespace, func() error {
+		if err := r.own(r.cephRadosNamespace); err != nil {
+			return err
+		}
+
+		addLabel(r.cephRadosNamespace, controllers.StorageConsumerNameLabel, r.storageConsumer.Name)
```
nb-ohad, line 485: What is the reason for having this?

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -425,6 +476,41 @@ func (r *StorageClassRequestReconciler) reconcileCephBlockPool(storageProfile *v
...
+		addLabel(r.cephBlockPool, controllers.StorageProfileSpecLabel, storageProfile.GetSpecHash())
+
+		r.cephRadosNamespace.Spec = rookCephv1.CephBlockPoolRadosNamespaceSpec{
+			BlockPoolName: r.cephBlockPool.Name,
```
nb-ohad, line 489: The code is correct, but I was not able to locate where you search for a blockpool named with the profile name on the request (or an alternative profile annotation). This is important because that blockpool should be the one named here as the desired state.

services/provider/server/server.go
jarrpa: Doing that for a `map[string]string{}` kind of defeats the purpose of the map, that just sounds like overhead for adding any new potential fields if the CSI driver ever develops any.

controllers/storageclassrequest/storageclassrequest_controller.go
jarrpa: Anything StorageProfile related will be removed in a subsequent commit. Currently debating whether to do it in another PR or delay this PR further to wait for it.

controllers/storageclassrequest/storageclassrequest_controller.go
jarrpa: I dunno, why weren't we using `Owns` for CephBlockPools in the first place?  I was just following convention since it seems largely inconsequential to the feature behavior.

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -140,20 +139,22 @@ func (r *StorageClassRequestReconciler) Reconcile(ctx context.Context, request r
 
 func (r *StorageClassRequestReconciler) SetupWithManager(mgr ctrl.Manager) error {
 
+	if err := mgr.GetCache().IndexField(
+		context.TODO(),
```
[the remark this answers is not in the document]
jarrpa: By that note, shouldn't we be declaring a global background context at the `main` level?

> Background returns a non-nil, empty [Context](https://pkg.go.dev/context#Context). It is never canceled, has no values, and has no deadline. It is typically used by the main function, initialization, and tests, and as the top-level Context for incoming requests.

controllers/storageclassrequest/storageclassrequest_controller.go
nb-ohad: From my point of view, both need to be using the `Owns` watch. 
The fact that  the block pool watch slipped through our review or that for that specific resource, there are abnormal conditions does not mean we need to repeat it without explaining.

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -230,14 +232,25 @@ func (r *StorageClassRequestReconciler) initPhase(storageProfile *v1.StorageProf
...
 
-		// check if a cephblockpool resource exists for the desired storageconsumer and storageprofile.
+		r.cephRadosNamespace = &rookCephv1.CephBlockPoolRadosNamespace{}
+		r.cephRadosNamespace.Namespace = r.OperatorNamespace
+		for _, res := range r.StorageClassRequest.Status.CephResources {
+			if res.Kind == "CephBlockPoolRadosNamespace" {
+				r.cephRadosNamespace.Name = res.Name
+				break
+			}
+		}
```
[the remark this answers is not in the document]
jarrpa: ACK

controllers/storageclassrequest/storageclassrequest_controller.go
jarrpa: I find it useful in `oc get` statements in my testing.

nb-ohad: Ok good enough reason, can you comment above the label so a future contributor will not remove it when they see it is not in use

controllers/storageclassrequest/storageclassrequest_controller.go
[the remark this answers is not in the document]
jarrpa: I think for the time being, I'll ask to move this to a separate discussion, and leave this PR as-is.

controllers/storageclassrequest/storageclassrequest_controller.go
jarrpa: ....okay so it turns out we are *almost* using `Owns()`, just obliquely. From https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.17.2/pkg/builder#Builder.Owns :

> By default, this is the equivalent of calling Watches(object, handler.EnqueueRequestForOwner([...], ownerType, OnlyControllerOwner())). 

...and indeed enqueueForOwner is almost an exact a wrapper on that. This is likely because we weren't setting OwnerRefs at the time this was developed.  Thus, from a technical perspective, this is *now* viable and I will make this change.

jarrpa:
Address most of the review comments, going to step away for a bit for the afternoon. Still not done with removing storage profiles (getting there...) but pushing what I have thus far just in case.

jarrpa:
Aaaand finally I can consider this ready for review. I still need to do another round of rigorous testing based on all the recent review changes, but for now this is close enough.

**NOTE:** StorageProfiles still remain, but only for SubvolumeGroups. Apart from being out of scope for this PR,  I'll want to have some discussion with folks more familiar with subvolume groups as to how to best determine/set/match device classes.

jarrpa:
I don't know what's causing the e2e-test to fail.... nothing in the operator Pod logs suggests an error of any kind, the tests just time out waiting for the StorageCluster... which it seems to do by watching for the OperatorCondition to be "Upgradable", I think? But the ocs-must-gather doesn't look like it ran, either, so I can't inspect any of the CRs. Any ideas?

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -221,23 +224,50 @@ func (r *StorageClassRequestReconciler) initPhase(storageProfile *v1.StorageProf
...
-				break
-			}
+
+		// check if a CephBlockPoolRadosNamespace resource exists for the desired storageconsumer and storageprofile.
+		cephRadosNamespaceList := &rookCephv1.CephBlockPoolRadosNamespaceList{}
+		err := r.Client.List(
+			r.ctx,
+			cephRadosNamespaceList,
+			client.InNamespace(r.OperatorNamespace),
+			client.MatchingFields{util.OwnerUIDIndexName: string(r.StorageClassRequest.UID)})
```
nb-ohad, line 239: 1. We have a util func on the reconciler to handle common cases, including the injection of the common context. There is no reason to not use it here
2. Code in this controller isolates closing parentheses for function calls on a new line to make future diffs for call alteration easier. 

```suggestion
		err := r.list(
			cephRadosNamespaceList,
			client.InNamespace(r.OperatorNamespace),
			client.MatchingFields{util.OwnerUIDIndexName: string(r.StorageClassRequest.UID)},
		)
```

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -221,23 +224,50 @@ func (r *StorageClassRequestReconciler) initPhase(storageProfile *v1.StorageProf
...
+		// check if a cephblockpool resource exists for the desired storageprofile.
 		if r.cephBlockPool.Name == "" {
 			cephBlockPoolList := &rookCephv1.CephBlockPoolList{}
-			listOptions := &client.MatchingLabels{
-				controllers.StorageConsumerNameLabel: r.storageConsumer.Name,
-				controllers.StorageProfileSpecLabel:  storageProfile.GetSpecHash(),
+			listOptions := []client.ListOption{
+				client.InNamespace(r.OperatorNamespace),
+				&client.MatchingLabels{
+					controllers.StorageProfileLabel: profileName,
```
nb-ohad, line 267: What if the profile name is empty and we need to resort to the default blockpool/ FS datapool?

Right now the profile name is identified by the following code
```
	profileName := r.StorageClassRequest.Spec.StorageProfile
	if profileName == "" {
		profileName = r.storageCluster.Spec.DefaultStorageProfile
	}
```

Which after the design change doesn't make sense as we are deprecating StorageProfle usage and the `DefaultStorageProfile` field

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -246,17 +276,20 @@ func (r *StorageClassRequestReconciler) initPhase(storageProfile *v1.StorageProf
 			// if we found more than one CephBlockPool, we can't determine which one to select, so error out
 			cbpItemsLen := len(cephBlockPoolList.Items)
 			if cbpItemsLen == 0 {
-				cbpNewName := fmt.Sprintf("cephblockpool-%s-%s", r.storageConsumer.Name, generateUUID())
+				cbpNewName := fmt.Sprintf("cephblockpool-profile-%s", profileName)
 				r.log.V(1).Info("no valid CephBlockPool found, creating new one", "CephBlockPool", cbpNewName)
 				r.cephBlockPool.Name = cbpNewName
+				r.cephBlockPool.Spec = r.getDefaultCephBlockPoolSpec()
+				addLabel(r.cephBlockPool, controllers.StorageProfileLabel, profileName)
```
nb-ohad, line 283: This is no according to the design. You are not supposed to create a new block pool, you should fail the reconcile if you cannot find an existing block pool to back up the claim.

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -325,7 +358,11 @@ func (r *StorageClassRequestReconciler) reconcilePhases() (reconcile.Result, err
 				return reconcile.Result{}, err
 			}
 
-			if err := r.reconcileCephBlockPool(&storageProfile); err != nil {
+			if err := r.reconcileCephBlockPool(); err != nil {
+				return reconcile.Result{}, err
+			}
```
nb-ohad, line 363: This call can be removed, this controller should not reconcile any block pools.  
The only thing we might want to keep is this line (from the `reconcileCephBlockPool`) func
```
r.setCephResourceStatus(r.cephBlockPool.Name, "CephBlockPool", phase, nil)
```

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -360,53 +397,53 @@ func (r *StorageClassRequestReconciler) reconcilePhases() (reconcile.Result, err
...
+
+	phase := ""
+	if r.cephBlockPool.Status != nil {
+		phase = string(r.cephBlockPool.Status.Phase)
+	}
+
+	r.setCephResourceStatus(r.cephBlockPool.Name, "CephBlockPool", phase, nil)
+
+	return nil
+}
```
nb-ohad, line 425: This is all dead code, please remove it

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -360,53 +397,53 @@ func (r *StorageClassRequestReconciler) reconcilePhases() (reconcile.Result, err
...
-				FailureDomain: failureDomain,
-				DeviceClass:   deviceClass,
-				Replicated: rookCephv1.ReplicatedSpec{
-					Size:                     3,
-					ReplicasPerFailureDomain: 1,
-				},
-				Parameters: storageProfile.Spec.BlockPoolConfiguration.Parameters,
-			},
+		r.cephRadosNamespace.Spec = rookCephv1.CephBlockPoolRadosNamespaceSpec{
+			BlockPoolName: r.cephBlockPool.Name,
```
nb-ohad, line 436: It will be better to read the block pool name from the source of truth, not from the block pool. 
The source of truth is the `r.StorageClassRequest.Spec.StorageProfile` or if empty, the default blook pool, which in ODF has a fixed and  known name (I belive defined as a constant in some other file)

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -642,6 +679,20 @@ func (r *StorageClassRequestReconciler) reconcileCephClientCephFSNode() error {
...
+			EnableRBDStats: true,
+			FailureDomain:  r.storageCluster.Status.FailureDomain,
+			Replicated: rookCephv1.ReplicatedSpec{
+				Size:                     3,
+				ReplicasPerFailureDomain: 1,
+				TargetSizeRatio:          .49,
+			},
+		},
+	}
+}
```
nb-ohad, line 694: By design, we are not and should not create new block pools as a result of a new request. We cannot as we do not know the user's desired setting for the new block pool (which represents a new storage profile).

Please remove this code

nb-ohad (changes requested):

controllers/storageclassrequest/storageclassrequest_controller.go
jarrpa: Done.

controllers/storageclassrequest/storageclassrequest_controller.go
jarrpa: The default block pool doesn't not have a fixed name, it is generated from the StorageCluster name. And if this is a legitimate provider-mode deployment there shouldn't be any default block pool at all. Per your requests elsewhere, if the StorageClassRequest has no specified profile and the StorageCluster has no default profile, we will fail.

controllers/storageclassrequest/storageclassrequest_controller.go
jarrpa: The more I worked on it, the more this design decision looks like a mistake. But sure, I'll remove it.

controllers/storageclassrequest/storageclassrequest_controller.go
jarrpa: Not gonna bother.

controllers/storageclassrequest/storageclassrequest_controller.go
jarrpa: Done.

controllers/storageclassrequest/storageclassrequest_controller.go
jarrpa: The default block pool does not have a constant name, it's generated based on the StorageCluster name. And as I've already said, per my understanding of not only the existing provider mode behavior but the new design, there shouldn't be a default block pool existing at all, nor am I being allowed to create it if it doesn't exist.

controllers/storageclassrequest/storageclassrequest_controller.go
jarrpa: That's literally what the default block pool is for, but sure, done.

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -171,11 +172,13 @@ func (r *StorageClassRequestReconciler) SetupWithManager(mgr ctrl.Manager) error
...
 	enqueueForOwner := handler.EnqueueRequestForOwner(mgr.GetScheme(), mgr.GetRESTMapper(), &v1alpha1.StorageClassRequest{})
+	generationChangedPredicate := builder.WithPredicates(predicate.GenerationChangedPredicate{})
+
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&v1alpha1.StorageClassRequest{}, builder.WithPredicates(
 			predicate.GenerationChangedPredicate{},
 		)).
-		Watches(&rookCephv1.CephBlockPool{}, enqueueForOwner).
+		Owns(&rookCephv1.CephBlockPoolRadosNamespace{}, generationChangedPredicate).
 		Watches(&rookCephv1.CephFilesystemSubVolumeGroup{}, enqueueForOwner).
```
rewantsoni, line 182: 1. Should we remove the watch over cephBlockPool? Can a blockpool be deleted if rados namespace exists? 
2. Why are we Owning RadosNamespace and Watching SubVolumeGroup? shouldn't we either Own both of them or Watch both of them?

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -221,52 +224,74 @@ func (r *StorageClassRequestReconciler) initPhase(storageProfile *v1.StorageProf
...
+		// if we found more than one CephBlockPoolRadosNamespace, we can't determine which one to select, so error out
+		rnsItemsLen := len(cephRadosNamespaceList.Items)
+		if rnsItemsLen == 0 {
+			rnsNewName := fmt.Sprintf("cephradosnamespace-%s-%s", r.StorageClassRequest.Spec.StorageProfile, r.storageConsumer.Name)
+			r.log.V(1).Info("no valid CephBlockPoolRadosNamespace found, creating new one", "CephBlockPoolRadosNamespace", rnsNewName)
+			r.cephRadosNamespace.Name = rnsNewName
+		} else if rnsItemsLen == 1 {
+			cephRns := cephRadosNamespaceList.Items[0]
+			r.cephBlockPool.Name = cephRns.Spec.BlockPoolName
+			r.cephRadosNamespace = &cephRns
```
rewantsoni, line 254: We don't need a new variable here

controllers/storageclassrequest/storageclassrequest_controller.go
nb-ohad: @jarrpa 
> The default block pool doesn't have a fixed name, it is generated from the StorageCluster name

Even if it does not have a fixed name it has a predictable name that we can use. 

> And if this is a legitimate provider-mode deployment 

There is no such a thing anymore, the point is to converge in implementation and on storage administration behavior

> Per your requests elsewhere, if the StorageClassRequest has no specified profile and the StorageCluster has no default profile, we will fail.

After these changes `default profile = default pool`, we will always have a default pool so if a profile name was not given you should be able to fulfill the request 100% of the time. The other comment mentioned a case when a user specify an explicit storage profile name, but we don't have a pool with that name.  

On a more general note:
Block pools are capturing storage concerns/requirements (which are defined by the env and the user). Namespaces are capturing access requirements (which are defined by the provider's multi-tenancy model). The entire point of using namespaces and moving away from creating new pools on behalf of the user it to give back the responsibilities of storage administration to the user while keeping access administration inside ODF (Provider mode) automation.

controllers/storageclassrequest/storageclassrequest_controller.go
nb-ohad: Please see the last part of my comment here:
https://github.com/red-hat-storage/ocs-operator/pull/2478#discussion_r1530134814

Not agreeing with a design decision does not mean the design is a mistake. In this specific case, I can guess that you are missing the underlying point of moving away from managing storage concerns in provider mode.

controllers/storageclassrequest/storageclassrequest_controller.go
nb-ohad: > That's literally what the default block pool is for, but sure, done.

I think you misunderstood me. I agree with your statement, but that is also my point. The default block pool is already there. we should not start creating additional "default" block pools, we should utilize the one that already exists.

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -221,52 +224,74 @@ func (r *StorageClassRequestReconciler) initPhase(storageProfile *v1.StorageProf
...
+			r.log.V(1).Info("valid CephBlockPoolRadosNamespace found", "CephBlockPoolRadosNamespace", r.cephRadosNamespace.Name)
+		} else {
+			return fmt.Errorf("invalid number of CephBlockPoolRadosNamespaces for storage consumer %q and storage profile %q: found %d, expecting 0 or 1", r.storageConsumer.Name, profileName, rnsItemsLen)
 		}
 
-		// check if a cephblockpool resource exists for the desired storageconsumer and storageprofile.
+		// check if a cephblockpool resource exists for the desired storageprofile.
 		if r.cephBlockPool.Name == "" {
```
nb-ohad, line 261: How can we get here without a name? 
If the request has a storage profile defined then we have a name. 
If not we already have the code above to set the profile name as the default block pool name. 

This means we always have a name and the code under the if will never execute.

nb-ohad (changes requested):

controllers/storageclassrequest/storageclassrequest_controller.go
jarrpa: 1. We should, so I did. And no, Rook-Ceph keeps a finalizer on the CephBlockPool CR until (among other conditions) there are no more RADOS namespaces defined for the pool.
2. See this (resolved) comment: https://github.com/red-hat-storage/ocs-operator/pull/2478#discussion_r1521870114 Has to do with the recent-ish change to using OwnerRefs.

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -221,52 +224,74 @@ func (r *StorageClassRequestReconciler) initPhase(storageProfile *v1.StorageProf
...
+		if err != nil {
+			return err
+		}
+
+		// if we found no CephBlockPoolRadosNamespaces, generate a new name
+		// if we found only one CephBlockPoolRadosNamespace with our query, we're good
+		// if we found more than one CephBlockPoolRadosNamespace, we can't determine which one to select, so error out
+		rnsItemsLen := len(cephRadosNamespaceList.Items)
+		if rnsItemsLen == 0 {
+			rnsNewName := fmt.Sprintf("cephradosnamespace-%s-%s", r.StorageClassRequest.Spec.StorageProfile, r.storageConsumer.Name)
```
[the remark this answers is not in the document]
jarrpa: Using hashed names is annoying, since it just makes it harder for humans to actually do meaningful troubleshooting, though I understand the technical limitations that lead to us using them. I believe that I had done the character counting such that it would take an almost max-length profile name string to cause issues, but I'm not about to do that calculation again. I'll just hash the names. 

And the empty-name edge case is a holdover from before you told me to gut the CephBlockPool reconciliation, as there would always be a StorageProfile defined by this point. Fixed.

controllers/storageclassrequest/storageclassrequest_controller.go
jarrpa: Holdover from deleted code. Fixed.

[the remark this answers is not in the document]
jarrpa: Holdover from deleted code. Fixed.

controllers/storageclassrequest/storageclassrequest_controller.go
jarrpa: Holdover from deleted code. Fixed.

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -221,52 +224,53 @@ func (r *StorageClassRequestReconciler) initPhase(storageProfile *v1.StorageProf
 
 	// check request status already contains the name of the resource. if not, add it.
 	if r.StorageClassRequest.Spec.Type == "blockpool" {
+		// initialize in-memory structs
+		r.cephRadosNamespace = &rookCephv1.CephBlockPoolRadosNamespace{}
+		r.cephRadosNamespace.Namespace = r.OperatorNamespace
 		r.cephBlockPool = &rookCephv1.CephBlockPool{}
+		r.cephBlockPool.Name = profileName
 		r.cephBlockPool.Namespace = r.OperatorNamespace
```
nb-ohad, line 232: There is no real need to have a block pool object on the reconciler. We are not loading it, creating it, or updating it.
In all cases, the name is derived from the profile name on the input (StoargeClassRequest), or if not provided from from the default ODF entities. 
So we don't even need to hold the name on the reconciler object.  
```suggestion
```

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -360,53 +364,26 @@ func (r *StorageClassRequestReconciler) reconcilePhases() (reconcile.Result, err
...
-				FailureDomain: failureDomain,
-				DeviceClass:   deviceClass,
-				Replicated: rookCephv1.ReplicatedSpec{
-					Size:                     3,
-					ReplicasPerFailureDomain: 1,
-				},
-				Parameters: storageProfile.Spec.BlockPoolConfiguration.Parameters,
-			},
+		r.cephRadosNamespace.Spec = rookCephv1.CephBlockPoolRadosNamespaceSpec{
+			BlockPoolName: r.cephBlockPool.Name,
```
nb-ohad, line 376:
```suggestion
		blockPoolName := r.StorageClassRequest.Spec.StorageProfile
		if blockPoolName == "" {
				blockPoolName = <code to get the name of ODF's default block pool>
		} 
		r.cephRadosNamespace.Spec = rookCephv1.CephBlockPoolRadosNamespaceSpec{
			BlockPoolName: blockPoolName,
```

nb-ohad (changes requested):

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -221,52 +224,53 @@ func (r *StorageClassRequestReconciler) initPhase(storageProfile *v1.StorageProf
...
-				r.cephBlockPool.Name = cephBlockPoolList.Items[0].GetName()
-				r.log.V(1).Info("valid CephBlockPool found", "CephBlockPool", r.cephBlockPool.Name)
-			} else {
-				return fmt.Errorf("invalid number of CephBlockPools for storage consumer %q and storage profile %q: found %d, expecting 0 or 1", r.storageConsumer.Name, storageProfile.Name, cbpItemsLen)
-			}
+			md5Sum := md5.Sum([]byte(r.StorageClassRequest.Name))
+			rnsNewName := fmt.Sprintf("cephradosnamespace-%s", hex.EncodeToString(md5Sum[:16]))
+			r.log.V(1).Info("no valid CephBlockPoolRadosNamespace found, creating new one", "CephBlockPoolRadosNamespace", rnsNewName)
+			r.cephRadosNamespace.Name = rnsNewName
+			r.cephRadosNamespace.Spec.BlockPoolName = profileName
```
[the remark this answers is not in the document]
nb-ohad: On second thought, the block pool name is part of the sepc and needs to be edited inside the `createOrUpdate` if not it might get overwritten by the `Get` operation.  
Let's remove this line
```suggestion
```

PR #2479, opened 2024-02-26
skip csi-addons pod restart for new namespaces

Nikhil-Ladha:
The csi-addons pod shouldn't be restarted when new namespaces starting with "openshift-" are created. We should only annotate the namespace and skip the csi-addons pod restart.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2257259

### DIFF
--- a/controllers/namespace/namespace_controller.go
+++ b/controllers/namespace/namespace_controller.go
@@ -23,6 +23,9 @@ import (
 // once is a pointer to the sync.Once type
 var once sync.Once
 
+// performCSIAddonsPodRestart a bool type to check if the csi-Addons pod should be restarted
+var performCSIAddonsPodRestart bool
+
 // NamespaceReconciler reconciles the namespaces starting with openshift-*
 // nolint:revive
 type NamespaceReconciler struct {
@@ -78,7 +81,7 @@ func (r *NamespaceReconciler) Reconcile(ctx context.Context, request reconcile.R
 	}
 
 	// Restart the csi-addons controller pod if any namespace was annotated
-	if updated {
+	if performCSIAddonsPodRestart && updated {
 		err = r.restartCSIAddonsPod()
 	}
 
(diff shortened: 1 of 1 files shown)


controllers/namespace/namespace_controller.go
```diff
@@ -23,6 +23,9 @@ import (
 // once is a pointer to the sync.Once type
 var once sync.Once
 
+// skipCSIAddonsPodRestart a bool type to check if the csi-Addons should be restarted
+var skipCSIAddonsPodRestart bool
```
iamniting, line 27: Lets rename this variable to opposite
`triggerCSIAddonsPodRestart`
`performCSIAddonsPodRestart`
`executeCSIAddonsPodRestart`

iamniting (changes requested):

controllers/namespace/namespace_controller.go
Nikhil-Ladha: Ak, updated to `performCSIAddonsPodRestart`

controllers/namespace/namespace_controller.go
```diff
@@ -78,7 +81,7 @@ func (r *NamespaceReconciler) Reconcile(ctx context.Context, request reconcile.R
 	}
 
 	// Restart the csi-addons controller pod if any namespace was annotated
-	if updated {
+	if !performCSIAddonsPodRestart && updated {
```
iamniting, line 84: not should be removed now.

iamniting (changes requested):

controllers/namespace/namespace_controller.go
```diff
@@ -23,6 +23,9 @@ import (
 // once is a pointer to the sync.Once type
 var once sync.Once
 
+// performCSIAddonsPodRestart a bool type to check if the csi-Addons pod should be restarted
+var performCSIAddonsPodRestart bool
```
Madhu-1, line 27: as its global should we take lock. not required now but if we enable max-concurrent-reconciles we will get into problems

Nikhil-Ladha: I think we are fine for now, we can update it later on when we enable max-concurrent-reconciles.

controllers/namespace/namespace_controller.go
Nikhil-Ladha: Done

controllers/namespace/namespace_controller.go
```diff
@@ -192,9 +195,11 @@ func (r *NamespaceReconciler) SetupWithManager(mgr ctrl.Manager) error {
 
 	nameSpacePredicate := predicate.Funcs{
 		CreateFunc: func(e event.CreateEvent) bool {
+			performCSIAddonsPodRestart = true
```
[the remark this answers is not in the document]
Nikhil-Ladha: Yeah, updated it now

Nikhil-Ladha:
The openshift cluster installation is failing for the bundle-e2e test, seems like an issue with aws or so :/

Madhu-1 (commented):
LGTM

iamniting (approved):

PR #2480, opened 2024-02-26
cephblockpool: allow updation of mirroring field from different components

rewantsoni:
When ocs-operator reconciles the cephblockpool object, it resets the mirroring field irrespective of whether mirroring is being used in the storageCluster API. This PR aims to modify the storageCluster API and allow us to enable mirroring on cephblockpool from StorageClusterPeer/API Server (required for RDR for Provider Mode).

After this PR,
When the mirroring object is nil, other components can set/unset the mirroringSpec on the cephBlockPool.
When the mirroring object is not nil, mirroring on the pools will be set based on the value of MirroringSpec in the storageCluster.

### DIFF
--- a/controllers/storagecluster/cephblockpools.go
+++ b/controllers/storagecluster/cephblockpools.go
@@ -46,7 +46,7 @@ func (r *StorageClusterReconciler) newCephBlockPoolInstances(initData *ocsv1.Sto
 	poolName := generateNameForCephBlockPool(initData)
 	poolNamespace := initData.Namespace
 
-	if initData.Spec.Mirroring.Enabled {
+	if initData.Spec.Mirroring != nil && initData.Spec.Mirroring.Enabled {
 		mirroringSpec.Enabled = true
 		mirroringSpec.Mode = "image"
 		mirroringPeerSpec := r.addPeerSecretsToCephBlockPool(initData, poolName, poolNamespace)
@@ -160,7 +160,14 @@ func (obj *ocsCephBlockPools) ensureCreated(r *StorageClusterReconciler, instanc
 
 			r.Log.Info("Restoring original CephBlockPool.", "CephBlockPool", klog.KRef(cephBlockPool.Namespace, cephBlockPool.Name))
 			existing.ObjectMeta.OwnerReferences = cephBlockPool.ObjectMeta.OwnerReferences
+
+			existingMirroringSpec := existing.Spec.Mirroring
 			existing.Spec = cephBlockPool.Spec
+
+			// If mirroring is not set on storageCluster instance, don't update it
+			if instance.Spec.Mirroring == nil {
+				existing.Spec.Mirroring = existingMirroringSpec
+			}
 			err = r.Client.Update(context.TODO(), &existing)
 			if err != nil {
 				r.Log.Error(err, "Failed to update CephBlockPool.", "CephBlockPool", klog.KRef(cephBlockPool.Namespace, cephBlockPool.Name))
--- a/api/v1/storagecluster_types.go
+++ b/api/v1/storagecluster_types.go
@@ -85,7 +85,9 @@ type StorageClusterSpec struct {
 	Arbiter ArbiterSpec `json:"arbiter,omitempty"`
 	// Mirroring specifies data mirroring configuration for the storage cluster.
 	// This configuration will only be applied to resources managed by the operator.
-	Mirroring MirroringSpec `json:"mirroring,omitempty"`
+	// +nullable
+	// +optional
+	Mirroring *MirroringSpec `json:"mirroring,omitempty"`
 	// OverprovisionControl specifies the allowed hard-limit PVs overprovisioning relative to
 	// the effective usable storage capacity.
 	OverprovisionControl []OverprovisionControlSpec `json:"overprovisionControl,omitempty"`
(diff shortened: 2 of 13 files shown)


controllers/storagecluster/cephblockpools.go
```diff
@@ -156,7 +156,18 @@ func (obj *ocsCephBlockPools) ensureCreated(r *StorageClusterReconciler, instanc
...
+			existing.Spec.Replicated = cephBlockPool.Spec.Replicated
+			existing.Spec.EnableRBDStats = cephBlockPool.Spec.EnableRBDStats
+			// update the mirroring field only when it is not nil
+			if instance.Spec.Mirroring != nil {
+				existing.Spec.Mirroring = cephBlockPool.Spec.Mirroring
+			}
+			// update the name only if it is NFS blockpool
+			if cephBlockPool.Name == generateNameForCephNFSBlockPool(instance) && instance.Spec.NFS != nil && instance.Spec.NFS.Enable {
+				existing.Spec.Name = cephBlockPool.Spec.Name
+			}
```
leelavg, line 170: if `existing.Spec` is being replaced by all these conditional entries, what happens if some new field gets added to `cephBlockPool` and we failed to update that in here?

but again, this is only true if the initial assumption is, this piece doesn't require an update if rook api gets updated.

controllers/storagecluster/cephblockpools.go
leelavg, line 1: I only see code changes wrt api changes, however may I know who sets this (mirroringSpec) in the first place?

if that's from UI and has a default, can the field be nullable?

rewantsoni: MirroringSpec is [set](https://github.com/red-hat-storage/odf-multicluster-orchestrator/blob/main/addons/agent_mirrorpeer_controller.go#L354) by the MCO operator

umangachapagain (commented):
Default CephBlockPool is managed by StorageCluster. So, it is expected that all changes to default pool is made through it. We do not allow other components to change our managed resources by design. We should not have 2 ocs-operator APIs managing the same resource.

leelavg:
@rewantsoni wrt https://github.com/red-hat-storage/ocs-operator/pull/2480#pullrequestreview-1947991624 should this PR be reviewed or not? if not, pls mention what alternatives were considered to move forward, thanks.

rewantsoni:
For RDR, we will allow users to use both default block pools and any additional blockpool created by them. The default blockpool is created and managed by the storageCluster whereas the user will create any additional blockpool from the UI. 

The Mirroring field needs to be updated from both StorageClusterPeer and OCS API Server (Provider API Server) ref [design](https://docs.google.com/document/d/1QEZn64-03zQikm7xbPGBvv7IWXVdbmk7S76XqQ02lro/edit?usp=sharing) doc for more details. If we don't allow the updation of the mirroring field from a different entity, we would need duplication of code and storageClusterPeer and OCS API Server would need to update storageCluster CR which won't be a good design either. As a trade-off, we are allowing an update of the cephblockpool mirroring field from different components.

@leelavg The PR can be reviewed, I have synced with Umanga and updated the PR.

controllers/storagecluster/cephblockpools.go
```diff
@@ -159,8 +159,14 @@ func (obj *ocsCephBlockPools) ensureCreated(r *StorageClusterReconciler, instanc
 			}
 
 			r.Log.Info("Restoring original CephBlockPool.", "CephBlockPool", klog.KRef(cephBlockPool.Namespace, cephBlockPool.Name))
-			existing.ObjectMeta.OwnerReferences = cephBlockPool.ObjectMeta.OwnerReferences
```
leelavg, line 162: is this removal due to both LHS & RHS being same?

controllers/storagecluster/cephblockpools.go
```diff
@@ -159,8 +159,14 @@ func (obj *ocsCephBlockPools) ensureCreated(r *StorageClusterReconciler, instanc
...
+
+			existingMirroringSpec := existing.Spec.Mirroring
 			existing.Spec = cephBlockPool.Spec
+
+			// If mirroring is not set on storageCluster instance, don't update it
+			if instance.Spec.Mirroring == nil {
+				existing.Spec.Mirroring = existingMirroringSpec
+			}
```
leelavg, line 169: this is a bit confusing, is this what you are trying to achieve?
```suggestion
			// If mirroring is not set on storageCluster instance, don't update it
			if instance.Spec.Mirroring == nil {
				cephBlockPool.Spec.Mirroring = nil
			}
			existing.Spec = cephBlockPool.Spec
```

controllers/storagecluster/cephblockpools.go
rewantsoni: this was done by mistake, I will revert it. Thanks for pointing it out

controllers/storagecluster/cephblockpools.go
rewantsoni: I want to allow the updation of the mirroring field on the cephBlockPool instance from other components if the mirroring field is not set on storageCluster CR.

Initially, I was updating each field under spec one by one, if a new field is added it might be missed here and lead to bugs. For that, I am creating a temporary copy of the mirroring field from cephblockpool CR and mirroring is not set on storageCluster CR I am setting the value to the temporary variable. So, it will not update the mirroring field on cephblockpool when we don't enable mirroring from storageCluster CR.

The suggestion you made, will update the mirroring field on cephblockpool to be nil if mirroring is not set on storageCluster CR

PR #2481, opened 2024-02-28
Remove active mds solution from alert

manishym:
Running more than one active MDS is not tested as of now.


agarwal-mudit (approved):
/lgtm

PR #2488, opened 2024-03-01
creation of onboarding job should be part seaprate ensure function

rchikatw:
- creation of onboarding job should be part seaprate ensure function

### DIFF
--- a/controllers/storagecluster/onboarding_validation_keys_generator_job.go
+++ b/controllers/storagecluster/onboarding_validation_keys_generator_job.go
@@ -1,0 +1,107 @@
+package storagecluster
+
+import (
+	"fmt"
+	"os"
+
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
+	"github.com/red-hat-storage/ocs-operator/v4/controllers/util"
+)
+
+const (
+	onboardingValidationKeysGeneratorImage   = "ONBOARDING_VALIDATION_KEYS_GENERATOR_IMAGE"
+	onboardingValidationKeysGeneratorJobName = "onboarding-validation-keys-generator"
+	onboardingValidationPublicKeySecretName  = "onboarding-ticket-key"
+)
+
+var onboardingJobSpec = batchv1.JobSpec{
+	// Eligible to delete automatically when job finishes
+	TTLSecondsAfterFinished: ptr.To(int32(0)),
+	Template: corev1.PodTemplateSpec{
+		Spec: corev1.PodSpec{
+			RestartPolicy:      corev1.RestartPolicyOnFailure,
+			ServiceAccountName: onboardingValidationKeysGeneratorJobName,
+			Containers: []corev1.Container{
+				{
+					Name:    onboardingValidationKeysGeneratorJobName,
+					Image:   os.Getenv(onboardingValidationKeysGeneratorImage),
+					Command: []string{"/usr/local/bin/onboarding-validation-keys-gen"},
+					Env: []corev1.EnvVar{
+						{
+							Name:  util.OperatorNamespaceEnvVar,
+							Value: os.Getenv(util.OperatorNamespaceEnvVar),
+						},
+					},
+				},
+			},
+		},
+	},
+}
+
+type onboardingValidationKeysGeneratorJob struct{}
+
+var _ resourceManager = &onboardingValidationKeysGeneratorJob{}
+
+func (o *onboardingValidationKeysGeneratorJob) ensureCreated(r *StorageClusterReconciler, storagecluster *ocsv1.StorageCluster) (reconcile.Result, error) {
+
+	if !storagecluster.Spec.AllowRemoteStorageConsumers {
+		r.Log.Info("Spec.AllowRemoteStorageConsumers is disabled, skipping onboarding validation key generator job creation")
+		return reconcile.Result{}, nil
+	}
+
+	if res, err := o.createJob(r, storagecluster.Namespace); err != nil {
+		return reconcile.Result{}, err
+	} else if !res.IsZero() {
+		return res, nil
+	}
+
+	return reconcile.Result{}, nil
+}
+
+func (o *onboardingValidationKeysGeneratorJob) createJob(r *StorageClusterReconciler, namespace string) (reconcile.Result, error) {
+	var err error
+	r.Log.Info("Spec.AllowRemoteStorageConsumers is enabled. Creating Onboarding validation key generator job")
+	if os.Getenv(onboardingValidationKeysGeneratorImage) == "" {
+		err = fmt.Errorf("OnboardingSecretGeneratorImage env var is not set")
+		r.Log.Error(err, "No value set for env variable")
+		return reconcile.Result{}, err
+	}
+	publicKeySecret := &corev1.Secret{}
+	publicKeySecret.Name = onboardingValidationPublicKeySecretName
+	publicKeySecret.Namespace = namespace
+	actualSecret := &corev1.Secret{}
+	// Creating the job only if public key is not found
+	err = r.Client.Get(r.ctx, client.ObjectKeyFromObject(publicKeySecret), actualSecret)
+	if kerrors.IsNotFound(err) {
+		onboardingSecretGeneratorJob := &batchv1.Job{}
+		onboardingSecretGeneratorJob.Name = onboardingValidationKeysGeneratorJobName
+		onboardingSecretGeneratorJob.Namespace = namespace
+		onboardingSecretGeneratorJob.Spec = *onboardingJobSpec.DeepCopy()
+		err = r.Client.Create(r.ctx, onboardingSecretGeneratorJob)
+		if err != nil && !kerrors.IsAlreadyExists(err) {
+			r.Log.Error(err, "failed to create onboarding validation key generator job")
+			return reconcile.Result{}, err
+		}
+
+	}
+	return reconcile.Result{}, nil
+}
+
+func (o *onboardingValidationKeysGeneratorJob) ensureDeleted(r *StorageClusterReconciler, storagecluster *ocsv1.StorageCluster) (reconcile.Result, error) {
+	onboardingSecretGeneratorJob := &batchv1.Job{}
+	onboardingSecretGeneratorJob.Name = onboardingValidationKeysGeneratorJobName
+	onboardingSecretGeneratorJob.Namespace = storagecluster.Namespace
+	if err := r.Client.Delete(r.ctx, onboardingSecretGeneratorJob); err != nil && !kerrors.IsNotFound(err) {
+		r.Log.Info("Failed to delete secret %s: %v", onboardingSecretGeneratorJob.Name, err)
+		return reconcile.Result{}, err
+	}
+
+	return reconcile.Result{}, nil
+}
--- a/controllers/storagecluster/provider_server_test.go
+++ b/controllers/storagecluster/provider_server_test.go
@@ -345,7 +345,6 @@ func createSetupForOcsProviderTest(t *testing.T, allowRemoteStorageConsumers boo
 	}
 
 	os.Setenv(providerAPIServerImage, "fake-image")
-	os.Setenv(onboardingValidationKeysGeneratorImage, "fake-image")
 	os.Setenv(util.WatchNamespaceEnvVar, "openshift-storage")
 
 	return r, instance
--- a/controllers/storagecluster/provider_server.go
+++ b/controllers/storagecluster/provider_server.go
@@ -11,13 +11,10 @@ import (
 
 	"go.uber.org/multierr"
 	appsv1 "k8s.io/api/apps/v1"
-	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
-	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
@@ -26,14 +23,12 @@ import (
 	"github.com/red-hat-storage/ocs-operator/v4/controllers/defaults"
 	"github.com/red-hat-storage/ocs-operator/v4/controllers/util"
 	"github.com/red-hat-storage/ocs-operator/v4/services/provider/server"
+	"k8s.io/utils/ptr"
 )
 
 const (
-	ocsProviderServerName                    = "ocs-provider-server"
-	providerAPIServerImage                   = "PROVIDER_API_SERVER_IMAGE"
-	onboardingValidationKeysGeneratorImage   = "ONBOARDING_VALIDATION_KEYS_GENERATOR_IMAGE"
-	onboardingValidationKeysGeneratorJobName = "onboarding-validation-keys-generator"
-	onboardingValidationPublicKeySecretName  = "onboarding-ticket-key"
+	providerAPIServerImage = "PROVIDER_API_SERVER_IMAGE"
+	ocsProviderServerName  = "ocs-provider-server"
 
 	ocsProviderServicePort     = int32(50051)
 	ocsProviderServiceNodePort = int32(31659)
@@ -66,12 +61,6 @@ func (o *ocsProviderServer) ensureCreated(r *StorageClusterReconciler, instance 
 	}
 
 	if res, err := o.createDeployment(r, instance); err != nil {
-		return reconcile.Result{}, err
-	} else if !res.IsZero() {
-		return res, nil
-	}
-
-	if res, err := o.createJob(r, instance); err != nil {
 		return reconcile.Result{}, err
 	} else if !res.IsZero() {
 		return res, nil
@@ -460,66 +449,6 @@ func RandomString(l int) string {
 	}
 
 	return string(bytes)
-}
-
-func getOnboardingJobObject(instance *ocsv1.StorageCluster) *batchv1.Job {
-
-	return &batchv1.Job{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      onboardingValidationKeysGeneratorJobName,
-			Namespace: instance.Namespace,
-		},
-		Spec: batchv1.JobSpec{
-			// Eligible to delete automatically when job finishes
-			TTLSecondsAfterFinished: ptr.To(int32(0)),
-			Template: corev1.PodTemplateSpec{
-				Spec: corev1.PodSpec{
-					RestartPolicy:      corev1.RestartPolicyOnFailure,
-					ServiceAccountName: onboardingValidationKeysGeneratorJobName,
-					Containers: []corev1.Container{
-						{
-							Name:    onboardingValidationKeysGeneratorJobName,
-							Image:   os.Getenv(onboardingValidationKeysGeneratorImage),
-							Command: []string{"/usr/local/bin/onboarding-validation-keys-gen"},
-							Env: []corev1.EnvVar{
-								{
-									Name:  util.OperatorNamespaceEnvVar,
-									Value: os.Getenv(util.OperatorNamespaceEnvVar),
-								},
-							},
-						},
-					},
-				},
-			},
-		},
-	}
-}
-
-func (o *ocsProviderServer) createJob(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) (reconcile.Result, error) {
-	var err error
-	if os.Getenv(onboardingValidationKeysGeneratorImage) == "" {
-		err = fmt.Errorf("OnboardingSecretGeneratorImage env var is not set")
-		r.Log.Error(err, "No value set for env variable")
-
-		return reconcile.Result{}, err
-	}
-
-	actualSecret := &corev1.Secret{}
-	// Creating the job only if public is not found
-	err = r.Client.Get(context.Background(), types.NamespacedName{Name: onboardingValidationPublicKeySecretName,
-		Namespace: instance.Namespace}, actualSecret)
-
-	if errors.IsNotFound(err) {
-		onboardingSecretGeneratorJob := getOnboardingJobObject(instance)
-		err = r.Client.Create(context.Background(), onboardingSecretGeneratorJob)
-	}
-	if err != nil {
-		r.Log.Error(err, "failed to create/ensure secret")
-		return reconcile.Result{}, err
-	}
-
-	r.Log.Info("Job is running as desired")
-	return reconcile.Result{}, nil
 }
 
 func (o *ocsProviderServer) updateClientConfigMap(r *StorageClusterReconciler, namespace string) error {
--- a/controllers/storagecluster/reconcile.go
+++ b/controllers/storagecluster/reconcile.go
@@ -392,6 +392,7 @@ func (r *StorageClusterReconciler) reconcilePhases(
 			objs = []resourceManager{
 				&ocsProviderServer{},
 				&storageClient{},
+				&onboardingValidationKeysGeneratorJob{},
 				&backingStorageClasses{},
 				&ocsTopologyMap{},
 				&ocsStorageQuota{},
--- a/controllers/storagecluster/uninstall_reconciler.go
+++ b/controllers/storagecluster/uninstall_reconciler.go
@@ -319,6 +319,7 @@ func (r *StorageClusterReconciler) deleteResources(sc *ocsv1.StorageCluster) (re
 
 	objs := []resourceManager{
 		&ocsExternalResources{},
+		&onboardingValidationKeysGeneratorJob{},
 		&ocsNoobaaSystem{},
 		&storageClient{},
 		&ocsProviderServer{},


jarrpa (approved):
/lgtm

jarrpa (changes requested):
Sorry about that early LGTM.... please remove the extraneous merge commit, looks like there should only be one commit in this PR.

controllers/storagecluster/onboarding_validation_keys_generator_job.go
```diff
@@ -0,0 +1,114 @@
...
+									Value: os.Getenv(util.OperatorNamespaceEnvVar),
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+}
```
nb-ohad, line 87: Why do we need a function for this? 
Let's have the fixed struct as a var at the package scope (at the top of the file)

controllers/storagecluster/onboarding_validation_keys_generator_job.go
```diff
@@ -0,0 +1,114 @@
...
+			return res, nil
+		}
+	}
+
+	return reconcile.Result{}, nil
+}
+
+func (o *onboardingValidationKeysGeneratorJob) ensureDeleted(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) (reconcile.Result, error) {
+
+	onboardingSecretGeneratorJob := getOnboardingJobObject(instance)
```
nb-ohad, line 45: You don't need the entire object, just the `name` and `namespace`

controllers/storagecluster/onboarding_validation_keys_generator_job.go
```diff
@@ -0,0 +1,114 @@
...
+	err := r.Client.Delete(context.TODO(), onboardingSecretGeneratorJob)
+
```
nb-ohad, line 47:
```suggestion
```

controllers/storagecluster/onboarding_validation_keys_generator_job.go
```diff
@@ -0,0 +1,114 @@
...
+	if err != nil && errors.IsNotFound(err) {
```
nb-ohad, line 48: I think you meant 
```suggestion
	if err != nil && !errors.IsNotFound(err) {
```

controllers/storagecluster/onboarding_validation_keys_generator_job.go
```diff
@@ -0,0 +1,114 @@
...
+		r.Log.Info("Failed to delete resource", "Kind", onboardingSecretGeneratorJob.GetObjectKind(), "Name", onboardingSecretGeneratorJob.GetName())
```
nb-ohad, line 49: Why do we need to log the Kind? we know this is a job by the context of the code

controllers/storagecluster/onboarding_validation_keys_generator_job.go
```diff
@@ -0,0 +1,114 @@
...
+		onboardingSecretGeneratorJob := getOnboardingJobObject(instance)
+		err = r.Client.Create(context.Background(), onboardingSecretGeneratorJob)
+	}
+	if err != nil {
+		r.Log.Error(err, "failed to create/ensure secret")
+		return reconcile.Result{}, err
+	}
+
+	r.Log.Info("Job is running as desired")
+	return reconcile.Result{}, nil
```
nb-ohad, line 113: This code is the implementation of the ensure func and the additional indirection does not add anything.  Let's move the code there.

controllers/storagecluster/onboarding_validation_keys_generator_job.go
```diff
@@ -0,0 +1,114 @@
...
+	if os.Getenv(onboardingSecretGeneratorImage) == "" {
+		err = fmt.Errorf("OnboardingSecretGeneratorImage env var is not set")
+		r.Log.Error(err, "No value set for env variable")
+
+		return reconcile.Result{}, err
+	}
+
+	actualSecret := &corev1.Secret{}
+	// Creating the job only if public is not found
+	err = r.Client.Get(context.Background(), types.NamespacedName{Name: onboardingTicketPublicKeySecretName,
```
nb-ohad, line 100: We should not create a new context for each client operation. 
Please use the context tracking the entire reconciliation. I believe it can be found in `r.ctx`

controllers/storagecluster/onboarding_validation_keys_generator_job.go
```diff
@@ -0,0 +1,114 @@
...
+		Namespace: instance.Namespace}, actualSecret)
+
+	if errors.IsNotFound(err) {
+		onboardingSecretGeneratorJob := getOnboardingJobObject(instance)
+		err = r.Client.Create(context.Background(), onboardingSecretGeneratorJob)
```
nb-ohad, line 105: Same as last comment about context reuse

controllers/storagecluster/onboarding_validation_keys_generator_job.go
```diff
@@ -0,0 +1,114 @@
...
+
+func (o *onboardingValidationKeysGeneratorJob) ensureCreated(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) (reconcile.Result, error) {
+
+	if instance.Spec.AllowRemoteStorageConsumers {
+		r.Log.Info("onboarding validation key generator job: Spec.AllowRemoteStorageConsumers is enabled")
+		if res, err := o.createJob(r, instance); err != nil {
+			return reconcile.Result{}, err
+		} else if !res.IsZero() {
+			return res, nil
+		}
```
nb-ohad, line 37: Is there a reason to not use the `createOrUpdate` pattern?

nb-ohad (changes requested):

controllers/storagecluster/onboarding_validation_keys_generator_job.go
jarrpa: In general this is an established pattern in ocs-operator for a variety of reasons, from having to calculate certain field values to obnoxious things like int pointers (which, yes, is now moot with the use of `ptr`). It also follows the pattern of "getting" an object rather than directly referencing is from somewhere.

While *here* it's not strictly needed since no logic needs to happen, I'd rather see the function present anyway for the sake of future changes that might require actual computation.

controllers/storagecluster/onboarding_validation_keys_generator_job.go
jarrpa: Fairly standard practice across the project, makes it easier to skim the logs.

controllers/storagecluster/onboarding_validation_keys_generator_job.go
```diff
@@ -0,0 +1,114 @@
...
+const (
+	providerAPIServerImage              = "PROVIDER_API_SERVER_IMAGE"
+	onboardingSecretGeneratorImage      = "ONBOARDING_SECRET_GENERATOR_IMAGE"
+	onboardingJobName                   = "onboarding-secret-generator"
+	onboardingTicketPublicKeySecretName = "onboarding-ticket-key"
+)
+
+type onboardingValidationKeysGeneratorJob struct{}
+
+func (o *onboardingValidationKeysGeneratorJob) ensureCreated(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) (reconcile.Result, error) {
```
leelavg, line 29: I'm just thinking whether we are successful if we created a job or ensured public keys 🤔? I see two issues:

1. Every successful reconcile basically rotates keys and if we are failing in some other reconciler after this, then we are potentially rotating keys until storagecluster stabilizes, even with that due to some spurious reconciles we will potentially be rerunning this job every minute, is that a good thing?
2. Assuming everything is stable and we aren't getting any reconciles (most ideal case) how do we cope up from job failed to create/rotate keys?

I don't think these questions are particularly related to the refactor, they just came up since we are using this feature from some time.

Unlike the status reporter job at client side, w/ current job we are closely coupled w/ outcome of the job, isn't it?

controllers/storagecluster/onboarding_validation_keys_generator_job.go
```diff
@@ -0,0 +1,114 @@
...
+			},
+		},
+	}
+}
+
+func (o *onboardingValidationKeysGeneratorJob) createJob(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) (reconcile.Result, error) {
+	var err error
+	if os.Getenv(onboardingSecretGeneratorImage) == "" {
+		err = fmt.Errorf("OnboardingSecretGeneratorImage env var is not set")
+		r.Log.Error(err, "No value set for env variable")
```
leelavg, line 93: is this a recoverable error? shouldn't this be panic? Or do we want to add a longer path for getting current ocs-operator image and use that?

rchikatw: I have not seen anywhere adding panic in the reconcile function, I can set the default path if it's empty but from where can I get the ocs-operator image path?

controllers/storagecluster/onboarding_validation_keys_generator_job.go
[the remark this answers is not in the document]
rchikatw: I believe the job create function is unnecessary, so I am removing it and moving it ensure created function

controllers/storagecluster/onboarding_validation_keys_generator_job.go
rchikatw: At this location, we are successful if we create the job, While the job responsibility is to rotate the key.

Regarding 1. We are not rotating the key every time. Instead, we are checking for the existence of the public key. Therefore, even if it reconciles, the public key will exist the next time, and a job won't be created.
Regarding 2. I don't see a reason for the job to fail. If it did fail, there must be something wrong with the environment, and we need to address it.

controllers/storagecluster/onboarding_validation_keys_generator_job.go
```diff
@@ -0,0 +1,103 @@
...
+	onboardingSecretGeneratorJob := &batchv1.Job{}
+	onboardingSecretGeneratorJob.ObjectMeta.Name = onboardingValidationKeysGeneratorJobName
+	onboardingSecretGeneratorJob.ObjectMeta.Namespace = instance.Namespace
+	err := r.Client.Delete(r.ctx, onboardingSecretGeneratorJob)
+	if err != nil && errors.IsNotFound(err) {
+		r.Log.Info("Failed to delete resource", "Kind", onboardingSecretGeneratorJob.GetObjectKind(), "Name", onboardingSecretGeneratorJob.GetName())
+		return reconcile.Result{}, nil
+	}
+
+	return reconcile.Result{}, err
```
rewantsoni, line 69: We are we returning nil if the error was found? maybe we want something like:
```suggestion
	if err != nil && !errors.IsNotFound(err) {
		r.Log.Info("Failed to delete resource", "Kind", onboardingSecretGeneratorJob.GetObjectKind(), "Name", onboardingSecretGeneratorJob.GetName())
		return reconcile.Result{}, err
	}

	return reconcile.Result{}, nil
```

controllers/storagecluster/provider_server_test.go
```diff
@@ -321,7 +321,6 @@ func createSetupForOcsProviderTest(t *testing.T, allowRemoteStorageConsumers boo
 	}
 
 	os.Setenv(providerAPIServerImage, "fake-image")
-	os.Setenv(onboardingValidationKeysGeneratorImage, "fake-image")
```
rewantsoni, line 324: Are you going to add unit-tests in a separate PR?

controllers/storagecluster/onboarding_validation_keys_generator_job.go
```diff
@@ -0,0 +1,103 @@
...
+
+type onboardingValidationKeysGeneratorJob struct{}
+
+var _ resourceManager = &onboardingValidationKeysGeneratorJob{}
+
+func (o *onboardingValidationKeysGeneratorJob) ensureCreated(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) (reconcile.Result, error) {
+
+	if instance.Spec.AllowRemoteStorageConsumers {
+		var err error
+		r.Log.Info("onboarding validation key generator job: Spec.AllowRemoteStorageConsumers is enabled")
```
rewantsoni, line 33: Could we reformat the message to something like:
```suggestion
		r.Log.Info("Spec.AllowRemoteStorageConsumers is enabled. Creating Onboarding validation key generator job")
```

controllers/storagecluster/onboarding_validation_keys_generator_job.go
```diff
@@ -0,0 +1,103 @@
...
+	onboardingValidationPublicKeySecretName  = "onboarding-ticket-key"
+)
+
+type onboardingValidationKeysGeneratorJob struct{}
+
+var _ resourceManager = &onboardingValidationKeysGeneratorJob{}
+
+func (o *onboardingValidationKeysGeneratorJob) ensureCreated(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) (reconcile.Result, error) {
+
+	if instance.Spec.AllowRemoteStorageConsumers {
```
rewantsoni, line 31: A small suggestion, could we return first if instance.Spec.AllowRemoteStorageConsumers is not enabled and then create the Job? Something like:
```
if !instance.Spec.AllowRemoteStorageConsumers {
//return
}
//create job
```

rewantsoni (changes requested):

controllers/storagecluster/provider_server_test.go
rchikatw: I missed adding that, I will add it in the same PR itself

controllers/storagecluster/onboarding_validation_keys_generator_job.go
nb-ohad: @jarrpa 
An established pattern does not mean it is a good pattern. In addition, I don't agree with the statement. We have other controllers inside ocs-operator that do not use this approach. Instead, they use a `templates` module with global vars for all of the static stuff. 


Regarding the `getting` part, from a design perspective it collides with the design behind a reconciliation process. 
With reconciliation, you want to set the desired state of an existing object, not get a new object. 
I can point you to more then one example  where I saw the following pattern 
```
field_i_need_to_preserve0 := oldObject.field0
field_i_need_to_preserve1 := oldObject.field1
field_i_need_to_preserve2 := oldObject.field2
newObj := GetNewObject()
new Obj.DeepCopyInto(oldObj)
oldObject.field0 = field_i_need_to_preserve0
oldObject.field1 = field_i_need_to_preserve1
oldObject.field2 = field_i_need_to_preserve2
```

It shows that the person writing the code put more emphasis on having a "clean" `GetNewObject` than actually solving the problem at hand.

My point is that a blanket statement like this is the pattern we were using is not a good indication id the code fit the needs in this specific instance.

controllers/storagecluster/onboarding_validation_keys_generator_job.go
nb-ohad: In that case why just the Kind and not the entire GVK, why not the namespace name? 
our logs then to bloat. When I see a preference for one type of information while others are ignored it raises the question of standard.  

Add to that the this log message is at a place in the code where we know exactly the type we are referring to, it is not a generic function handling multiple types, so the log record should be much much simpler and should include the underlying error (which is the most important part of the log message)
```suggestion
		r.Log.Infof("Failed to delete secret %s: %v\m", onboardingSecretGeneratorJob.Name, err)
```

controllers/storagecluster/onboarding_validation_keys_generator_job.go
nb-ohad: I don't understand the answer, can you elaborate?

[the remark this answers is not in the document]
nb-ohad: @leelavg There is no one happy path here. if you look at the statements there are 3 distinct branches

controllers/storagecluster/onboarding_validation_keys_generator_job.go
```diff
@@ -0,0 +1,114 @@
...
+			TTLSecondsAfterFinished: ptr.To(int32(0)),
+			Template: corev1.PodTemplateSpec{
+				Spec: corev1.PodSpec{
+					RestartPolicy:      corev1.RestartPolicyOnFailure,
+					ServiceAccountName: onboardingJobName,
+					Containers: []corev1.Container{
+						{
+							Name:    onboardingJobName,
+							Image:   os.Getenv(onboardingSecretGeneratorImage),
+							Command: []string{"/usr/local/bin/onboarding-secret-generator"},
```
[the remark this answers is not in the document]
nb-ohad: That approach has a lot of limitations. I would suggest having the image name provided by an envvar similar to how we inject other components' images

controllers/storagecluster/onboarding_validation_keys_generator_job.go
nb-ohad: @leelavg The key point is point #1 in @rchikatw's answer. We will not rotate the key if even one reconcile already executed this code

controllers/storagecluster/onboarding_validation_keys_generator_job.go
nb-ohad: No need, this is the common practice, we are expecting the build process to provide an image path for downstream

controllers/storagecluster/onboarding_validation_keys_generator_job.go
jarrpa: I don't exactly remember why this is the case, but it is the project standard to use `r.Log.Info` with value-pairs of "Kind" and "Name". Nowhere else do we use `r.Log.Infof` in a controller. Without a deeper investigation as to why we're not using `Infof` to begin with, I don't think it's fair to arbitrarily start using it here.

controllers/storagecluster/onboarding_validation_keys_generator_job.go
```diff
@@ -0,0 +1,105 @@
...
+)
+
+type onboardingValidationKeysGeneratorJob struct{}
+
+var _ resourceManager = &onboardingValidationKeysGeneratorJob{}
+
+func (o *onboardingValidationKeysGeneratorJob) ensureCreated(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) (reconcile.Result, error) {
+
+	if !instance.Spec.AllowRemoteStorageConsumers {
+		r.Log.Info("Spec.AllowRemoteStorageConsumers is disabled, Job is not created")
```
nb-ohad, line 32:
```suggestion
		r.Log.Info("Spec.AllowRemoteStorageConsumers is disabled, skipping onboarding validation key generator job creation")
```

controllers/storagecluster/onboarding_validation_keys_generator_job.go
```diff
@@ -0,0 +1,105 @@
...
+	r.Log.Info("Spec.AllowRemoteStorageConsumers is enabled. Creating Onboarding validation key generator job")
+	if os.Getenv(onboardingValidationKeysGeneratorImage) == "" {
+		err = fmt.Errorf("OnboardingSecretGeneratorImage env var is not set")
+		r.Log.Error(err, "No value set for env variable")
+		return reconcile.Result{}, err
+	}
+	actualSecret := &corev1.Secret{}
+	// Creating the job only if public is not found
+	err = r.Client.Get(r.ctx, types.NamespacedName{Name: onboardingValidationPublicKeySecretName,
+		Namespace: instance.Namespace}, actualSecret)
```
nb-ohad, line 45: 1. Why name the secret `actualSecret`? this name does not describe anything.
2. I am suggesting an alternative to the same operation that I believe is more readable. 

```suggestion
	 publicKeySecret := &corev1.Secret{}
	 publicKeySecret.Name = onboardingValidationPublicKeySecretName
	actualSecret.Namespace = instance.Namespace
	// Creating the job only if public is not found
	err = r.Client.Get(r.ctx, client.ObjectKeyFromObject(publicKeySecret), publicKeySecret)
```

controllers/storagecluster/onboarding_validation_keys_generator_job.go
```diff
@@ -0,0 +1,105 @@
...
+
```
nb-ohad, line 46: Please don't leave empty lines between an operation and error handling for that operation
```suggestion
```

controllers/storagecluster/onboarding_validation_keys_generator_job.go
```diff
@@ -0,0 +1,105 @@
...
+	if errors.IsNotFound(err) {
+		onboardingSecretGeneratorJob := getOnboardingJobObject(instance)
+		err = r.Client.Create(r.ctx, onboardingSecretGeneratorJob)
+		if err != nil {
+			r.Log.Error(err, "failed to create/ensure secret")
```
nb-ohad, line 51: You are not creating the secret here:
```suggestion
			r.Log.Error(err, "failed to create onboarding validation key generator job")
```

controllers/storagecluster/onboarding_validation_keys_generator_job.go
```diff
@@ -0,0 +1,105 @@
...
+			return reconcile.Result{}, err
+		}
+
+	}
+
+	return reconcile.Result{}, nil
+}
+
+func (o *onboardingValidationKeysGeneratorJob) ensureDeleted(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) (reconcile.Result, error) {
+
```
nb-ohad, line 61:
```suggestion
```

controllers/storagecluster/onboarding_validation_keys_generator_job.go
```diff
@@ -0,0 +1,105 @@
...
+	onboardingSecretGeneratorJob := &batchv1.Job{}
+	onboardingSecretGeneratorJob.ObjectMeta.Name = onboardingValidationKeysGeneratorJobName
+	onboardingSecretGeneratorJob.ObjectMeta.Namespace = instance.Namespace
```
nb-ohad, line 64:
```suggestion
	onboardingSecretGeneratorJob.Name = onboardingValidationKeysGeneratorJobName
	onboardingSecretGeneratorJob.Namespace = instance.Namespace
```

controllers/storagecluster/onboarding_validation_keys_generator_job.go
```diff
@@ -0,0 +1,105 @@
...
+	return reconcile.Result{}, nil
+}
+
+func getOnboardingJobObject(instance *ocsv1.StorageCluster) *batchv1.Job {
+
+	return &batchv1.Job{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      onboardingValidationKeysGeneratorJobName,
+			Namespace: instance.Namespace,
+		},
```
nb-ohad, line 80: Name and namesapce should not be part of the template. They are external information added to the constant information of the template.

controllers/storagecluster/onboarding_validation_keys_generator_job.go
```diff
@@ -0,0 +1,105 @@
...
+		return reconcile.Result{}, err
+	}
+	actualSecret := &corev1.Secret{}
+	// Creating the job only if public is not found
+	err = r.Client.Get(r.ctx, types.NamespacedName{Name: onboardingValidationPublicKeySecretName,
+		Namespace: instance.Namespace}, actualSecret)
+
+	if errors.IsNotFound(err) {
+		onboardingSecretGeneratorJob := getOnboardingJobObject(instance)
+		err = r.Client.Create(r.ctx, onboardingSecretGeneratorJob)
```
nb-ohad, line 49: Why is this `create` and not `CreateOrUpdate`? What happens if the job is already running?

nb-ohad (changes requested):

controllers/storagecluster/onboarding_validation_keys_generator_job.go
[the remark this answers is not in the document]
leelavg: it was there when I commented, like creating job inside an if checking allowconsumer field. it's changed now, can be resolved.

controllers/storagecluster/onboarding_validation_keys_generator_job.go
leelavg: yes, thanks for clarification, can be resolved.

controllers/storagecluster/provider_server_test.go
rchikatw: I will add the unit-test in different PR

controllers/storagecluster/onboarding_validation_keys_generator_job.go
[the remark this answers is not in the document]
rchikatw: The image is already added as env,  the command for the container we add directly there is no need for env.

controllers/storagecluster/onboarding_validation_keys_generator_job.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+	return reconcile.Result{}, nil
+}
+
+func (o *onboardingValidationKeysGeneratorJob) ensureDeleted(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) (reconcile.Result, error) {
+	onboardingSecretGeneratorJob := &batchv1.Job{}
+	onboardingSecretGeneratorJob.Name = onboardingValidationKeysGeneratorJobName
+	onboardingSecretGeneratorJob.Namespace = instance.Namespace
+	err := r.Client.Delete(r.ctx, onboardingSecretGeneratorJob)
+	if err != nil && !errors.IsNotFound(err) {
```
leelavg, line 103:
```suggestion
	if err := r.Client.Delete(r.ctx, onboardingSecretGeneratorJob); err != nil && !errors.IsNotFound(err) {
```

controllers/storagecluster/onboarding_validation_keys_generator_job.go
```diff
@@ -0,0 +1,109 @@
...
+		if err != nil {
+			r.Log.Error(err, "failed to create onboarding validation key generator job")
+			return reconcile.Result{}, err
+		}
+
+	}
+	return reconcile.Result{}, nil
+}
+
+func (o *onboardingValidationKeysGeneratorJob) ensureDeleted(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) (reconcile.Result, error) {
```
leelavg, line 98:
```suggestion
func (o *onboardingValidationKeysGeneratorJob) ensureDeleted(r *StorageClusterReconciler, storagecluster *ocsv1.StorageCluster) (reconcile.Result, error) {
```

controllers/storagecluster/onboarding_validation_keys_generator_job.go
```diff
@@ -0,0 +1,109 @@
+package storagecluster
+
+import (
+	"fmt"
+	"os"
+
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
```
leelavg, line 9:
```suggestion
	kerrors "k8s.io/apimachinery/pkg/api/errors"
```
just not to collide w/ stdlib pkg.

controllers/storagecluster/onboarding_validation_keys_generator_job.go
```diff
@@ -0,0 +1,109 @@
...
+		onboardingSecretGeneratorJob.Name = onboardingValidationKeysGeneratorJobName
+		onboardingSecretGeneratorJob.Namespace = instance.Namespace
+		onboardingSecretGeneratorJob.Spec = *onboardingJobSpec.DeepCopy()
+		err = r.Client.Create(r.ctx, onboardingSecretGeneratorJob)
+		if err != nil {
+			r.Log.Error(err, "failed to create onboarding validation key generator job")
+			return reconcile.Result{}, err
+		}
+
+	}
```
leelavg, line 94: you are failing silently here if you fail to get the secret as there's no `err != nil`

controllers/storagecluster/onboarding_validation_keys_generator_job.go
```diff
@@ -0,0 +1,109 @@
...
+	if res, err := o.createJob(r, instance); err != nil {
+		return reconcile.Result{}, err
+	} else if !res.IsZero() {
+		return res, nil
+	}
+
+	return reconcile.Result{}, nil
+}
+
+func (o *onboardingValidationKeysGeneratorJob) createJob(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) (reconcile.Result, error) {
```
leelavg, line 69:
```suggestion
func (o *onboardingValidationKeysGeneratorJob) createJob(r *StorageClusterReconciler, namespace string) (reconcile.Result, error) {
```

controllers/storagecluster/onboarding_validation_keys_generator_job.go
```diff
@@ -0,0 +1,109 @@
...
+			},
+		},
+	},
+}
+
+type onboardingValidationKeysGeneratorJob struct{}
+
+var _ resourceManager = &onboardingValidationKeysGeneratorJob{}
+
+func (o *onboardingValidationKeysGeneratorJob) ensureCreated(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) (reconcile.Result, error) {
```
leelavg, line 53:
```suggestion
func (o *onboardingValidationKeysGeneratorJob) ensureCreated(r *StorageClusterReconciler, storagecluster *ocsv1.StorageCluster) (reconcile.Result, error) {
```

controllers/storagecluster/onboarding_validation_keys_generator_job.go
```diff
@@ -0,0 +1,109 @@
...
+	} else if !res.IsZero() {
+		return res, nil
+	}
+
+	return reconcile.Result{}, nil
+}
+
+func (o *onboardingValidationKeysGeneratorJob) createJob(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) (reconcile.Result, error) {
+	var err error
+	r.Log.Info("Spec.AllowRemoteStorageConsumers is enabled. test ritesh Creating Onboarding validation key generator job")
```
leelavg, line 71: missed removing this log line?

controllers/storagecluster/onboarding_validation_keys_generator_job.go
```diff
@@ -0,0 +1,109 @@
...
+	if os.Getenv(onboardingValidationKeysGeneratorImage) == "" {
+		err = fmt.Errorf("OnboardingSecretGeneratorImage env var is not set")
+		r.Log.Error(err, "No value set for env variable")
+		return reconcile.Result{}, err
+	}
+	publicKeySecret := &corev1.Secret{}
+	publicKeySecret.Name = onboardingValidationPublicKeySecretName
+	publicKeySecret.Namespace = instance.Namespace
+	actualSecret := &corev1.Secret{}
+	// Creating the job only if public is not found
```
leelavg, line 81:
```suggestion
	// Creating the job only if public key is not found
```

controllers/storagecluster/onboarding_validation_keys_generator_job.go
```diff
@@ -0,0 +1,109 @@
...
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
+	"github.com/red-hat-storage/ocs-operator/v4/controllers/util"
+	"k8s.io/klog/v2"
```
leelavg, line 16: doesn't the reconciler has a logger?

controllers/storagecluster/onboarding_validation_keys_generator_job.go
```diff
@@ -0,0 +1,109 @@
...
+	actualSecret := &corev1.Secret{}
+	// Creating the job only if public is not found
+	err = r.Client.Get(r.ctx, client.ObjectKeyFromObject(publicKeySecret), actualSecret)
+	if errors.IsNotFound(err) {
+		onboardingSecretGeneratorJob := &batchv1.Job{}
+		onboardingSecretGeneratorJob.Name = onboardingValidationKeysGeneratorJobName
+		onboardingSecretGeneratorJob.Namespace = instance.Namespace
+		onboardingSecretGeneratorJob.Spec = *onboardingJobSpec.DeepCopy()
+		err = r.Client.Create(r.ctx, onboardingSecretGeneratorJob)
+		if err != nil {
```
leelavg, line 89: should you skip checking `AlreadyExist` error as you may get many reconciles before storagecluster gets stable or other route is set the job failure policy to `Never` and recreate the job every time which is wasteful of resources i guess.

controllers/storagecluster/onboarding_validation_keys_generator_job.go
rchikatw: yes removed

controllers/storagecluster/onboarding_validation_keys_generator_job.go
rchikatw: i didn't get this point. What you are expecting here?

rchikatw:
I believe this pull request is outdated. I will create a new one if needed.

PR #2495, opened 2024-03-05
Use podman cp over podman run cat/ls when fetching CSVs for Rook

GingerGeek:
When building a pipeline ([okd-project/okd-operator-pipeline](https://github.com/okd-project/okd-operator-pipeline/tree/ocs/base/tekton.dev/repositories/container-storage)) to build and assemble the storage operators, we had issues with building OCS due to running of containers-in-a-pod as part of the pipeline.

Running a container requires additional privileges that we are trying to avoid giving to our Tekton pipeline and in this case it seemed using `podman cp` would provide a slightly simpler solution.


GingerGeek:
Thanks, I also noted that the Rook CRDs currently in the repo all have CRLF line endings. The `cp` command will replace them with LF but not sure if that should be included in this PR or not?

Nikhil-Ladha:
We won't be needing this PR, once we are set to take this PR in https://github.com/red-hat-storage/ocs-operator/pull/2492

GingerGeek:
It's not a blocker no, especially if you're planning to switch away. We currently hold this PR [as a patch](https://github.com/okd-project/okd-operator-pipeline/blob/ocs/base/tekton.dev/repositories/container-storage/operator/0001-Prefer-create-and-cp-over-cat-and-ls-manipulation.patch) on our build pipeline for this repository. cc/ @SupremeMortal

jarrpa:
@GingerGeek Will PR #2492 still work for you if it's using `operator-sdk run bundle`? https://sdk.operatorframework.io/docs/cli/operator-sdk_run_bundle/ You may want to verify this.

jarrpa:
For what it's work, I have some dev patches where I"m using kustomize to deploy rook-ceph-operator from within this repo, and moving away from using `podman` directly would be a step backwards for me. I like not using OLM where I don't have to. :stuck_out_tongue_closed_eyes:

GingerGeek:
I'm not sure, if there's an underlying use of a running container then it would likely fail in our current pipelines as podman-in-okd-pods is non-trivial if you want to maintain non-privileged running.

We may have to bite the bullets and just start running builds in privileged contexts. 

It's worth mentioning that this method of us taking a project like this and running the build ourself is a somewhat stop-gap measure to help bootstrap operators-on-okd.

Our overall goal would be for the OKD catalog to be a first-class destination for any upstream operators that package into Openshift. So ie you folks on this project would have CI/CD to build operator bundles and then submit that into our catalog. Happy to partake in any community process that would help make that happen

PR #2500, opened 2024-03-07
Various small fixes found during development

jarrpa:
Because I tend to do things slightly weird I found a few small bits of cleanup here and there, as well as one odd bug. I don't remember how I hit it, but I've been holding this in a working branch for a while, so here ya go. The change in the CSV is purely cosmetic, and shouldn't affect the downstream builds.

### DIFF
--- a/controllers/util/k8sutil.go
+++ b/controllers/util/k8sutil.go
@@ -32,8 +32,6 @@ const (
 	// These are the keys in the ocs-operator-config configmap
 	ClusterNameKey              = "CSI_CLUSTER_NAME"
 	RookCurrentNamespaceOnlyKey = "ROOK_CURRENT_NAMESPACE_ONLY"
-	EnableReadAffinityKey       = "CSI_ENABLE_READ_AFFINITY"
-	CephFSKernelMountOptionsKey = "CSI_CEPHFS_KERNEL_MOUNT_OPTIONS"
 	EnableTopologyKey           = "CSI_ENABLE_TOPOLOGY"
 	TopologyDomainLabelsKey     = "CSI_TOPOLOGY_DOMAIN_LABELS"
 	EnableNFSKey                = "ROOK_CSI_ENABLE_NFS"
--- a/deploy/ocs-operator/manifests/ocs-operator.clusterserviceversion.yaml
+++ b/deploy/ocs-operator/manifests/ocs-operator.clusterserviceversion.yaml
@@ -56,7 +56,7 @@ metadata:
       ]
     capabilities: Deep Insights
     categories: Storage
-    containerImage: quay.io/ocs-dev/ocs-operator:4.15.0
+    containerImage: quay.io/ocs-dev/ocs-operator:latest
     description: Red Hat OpenShift Container Storage provides hyperconverged storage
       for applications within an OpenShift cluster.
     external.features.ocs.openshift.io/export-script: |-
--- a/tools/csv-merger/csv-merger.go
+++ b/tools/csv-merger/csv-merger.go
@@ -18,7 +18,6 @@ import (
 	"github.com/operator-framework/api/pkg/lib/version"
 	csvv1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/red-hat-storage/ocs-operator/v4/controllers/util"
-	ocsversion "github.com/red-hat-storage/ocs-operator/v4/version"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
@@ -609,7 +608,7 @@ func generateUnifiedCSV() *csvv1.ClusterServiceVersion {
 		}
 		ocsCSV.Annotations["createdAt"] = time.Now().In(loc).Format("2006-01-02 15:04:05")
 	}
-	ocsCSV.Annotations["containerImage"] = "quay.io/ocs-dev/ocs-operator:" + ocsversion.Version
+	ocsCSV.Annotations["containerImage"] = *ocsContainerImage
 	ocsCSV.Annotations["capabilities"] = "Deep Insights"
 	ocsCSV.Annotations["categories"] = "Storage"
 	ocsCSV.Annotations["operators.operatorframework.io/operator-type"] = "non-standalone"


controllers/storagecluster/storagecluster_controller.go
```diff
@@ -131,6 +132,7 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	enqueueStorageClusterRequest := handler.EnqueueRequestsFromMapFunc(
 		func(context context.Context, obj client.Object) []reconcile.Request {
 			// Get the StorageCluster objects
+			r.ctx = ctrllog.IntoContext(context, r.Log)
```
nb-ohad, line 135: What are the benefits of adding the logger to the context? If you want to control the logger of each reconcile iteration there is a controller option to provide a function to create a logger for each processed event and store it on context sent to the reconcile call for that event.

controllers/storagecluster/provider_server.go
```diff
@@ -496,9 +496,12 @@ func (o *ocsProviderServer) createJob(r *StorageClusterReconciler, instance *ocs
...
 		onboardingSecretGeneratorJob := getOnboardingJobObject(instance)
-		err = r.Client.Create(context.Background(), onboardingSecretGeneratorJob)
-	}
-	if err != nil {
+		jobErr := r.Client.Create(context.Background(), onboardingSecretGeneratorJob)
+		if jobErr != nil && !errors.IsAlreadyExists(jobErr) {
+			r.Log.Error(jobErr, "failed to create onboarding job")
+			return reconcile.Result{}, jobErr
+		}
+	} else if err != nil {
```
nb-ohad, line 504: We want to fail if it already exists as we don't have guarantees about the state and content of the secrets when the job was originally invoked. 

Before this change, the reconciliation will fail and the next reconciliation will either create the job or fail again. Eventually, a job will be created where we have exact context and guarantees of its outcome.

nb-ohad (changes requested):

controllers/storagecluster/provider_server.go
```diff
@@ -496,9 +496,12 @@ func (o *ocsProviderServer) createJob(r *StorageClusterReconciler, instance *ocs
 
 	if errors.IsNotFound(err) {
 		onboardingSecretGeneratorJob := getOnboardingJobObject(instance)
-		err = r.Client.Create(context.Background(), onboardingSecretGeneratorJob)
-	}
-	if err != nil {
+		jobErr := r.Client.Create(context.Background(), onboardingSecretGeneratorJob)
+		if jobErr != nil && !errors.IsAlreadyExists(jobErr) {
```
leelavg, line 500: I believe you might have hit this during very quick reconciles that matches the length of job duration and I agree we don't want to keep on queuing reconciles and at the same time we are unsure whether the job status is completed due to failure or success.

So, I'd like to settle for a middle ground if you will, send a reconcile with delay if job already exists? then the issue will be, who'll ensure updating this if the job responsibilities increases and I'd argue that's better than not getting a reconcile at all.

At the end, everything needs to stabilize unless the garbage collector controller give up and not removing pods, basically not respecting ttl (happens more often than we think) and storagecluster will not progress.

Sry for being a li'l exhaustive and if every solution has edge cases I believe we want to take action once we get some regression results from testing and settle on fixing the case w/ most probability, wdyt?

controllers/util/k8sutil.go
```diff
@@ -32,8 +32,6 @@ const (
 	// These are the keys in the ocs-operator-config configmap
 	ClusterNameKey              = "CSI_CLUSTER_NAME"
 	RookCurrentNamespaceOnlyKey = "ROOK_CURRENT_NAMESPACE_ONLY"
-	EnableReadAffinityKey       = "CSI_ENABLE_READ_AFFINITY"
-	CephFSKernelMountOptionsKey = "CSI_CEPHFS_KERNEL_MOUNT_OPTIONS"
```
leelavg, line 36: are these leftovers of some previous commits?

deploy/ocs-operator/manifests/ocs-operator.clusterserviceversion.yaml
```diff
@@ -56,7 +56,7 @@ metadata:
       ]
     capabilities: Deep Insights
     categories: Storage
-    containerImage: quay.io/ocs-dev/ocs-operator:4.15.0
+    containerImage: quay.io/ocs-dev/ocs-operator:latest
```
leelavg, line 59: why should this be `latest`?

tools/csv-merger/csv-merger.go
```diff
@@ -609,7 +608,7 @@ func generateUnifiedCSV() *csvv1.ClusterServiceVersion {
 		}
 		ocsCSV.Annotations["createdAt"] = time.Now().In(loc).Format("2006-01-02 15:04:05")
 	}
-	ocsCSV.Annotations["containerImage"] = "quay.io/ocs-dev/ocs-operator:" + ocsversion.Version
+	ocsCSV.Annotations["containerImage"] = *ocsContainerImage
```
leelavg, line 611: why is this change required?

controllers/storagecluster/storagecluster_controller.go
jarrpa: I've had this particular patch for a very, very long time. It looks like the issue I was running into (there was a `nil` pointer reference to `r.ctx` somewhere, I don't remember now) was resolved in a different place by PR #1748 . I can probably safely remove it.

controllers/storagecluster/provider_server.go
jarrpa: Are you proposing a requeue? If so then yeah that makes sense.

controllers/storagecluster/provider_server.go
jarrpa: So, really, the "happy path" behavior is `errors.IsNotFound(err)` and we should have an error if `err == nil`?

jarrpa: I believe I ran into this while I was hacking on the Job and it had gone into a crash loop. Does this mean that if a version of the Job was already created, we want to keep failing the reconcile loop until either the garbage collector or an admin deletes the extant Job?

controllers/util/k8sutil.go
jarrpa: Yup: https://github.com/red-hat-storage/ocs-operator/pull/2294

deploy/ocs-operator/manifests/ocs-operator.clusterserviceversion.yaml
jarrpa: Because we're still trying to claim this is an upstream repo, and the `main` branch should by default build to `latest`, *not* the "latest release version". Our downstream pipelines always explicitly set this variable anyway, so that should be of no consequence.

tools/csv-merger/csv-merger.go
jarrpa: This is what produced the change in `deploy/ocs-operator/manifests/ocs-operator.clusterserviceversion.yaml`. Since it's in an annotation it is largely cosmetic, so there's no strict "requirement" for it. But it's still technically wrong, so why not?

controllers/storagecluster/provider_server.go
leelavg: Sry, by the end I meant, there isn't any fix known yet and we wait for regression tests to see what edge case is more frequent and fix that.

controllers/storagecluster/provider_server.go
nb-ohad: The retention time of the job is set to 0, so it should delete itself immediately after completion. 
I acknowledge that this approach is also lacking. We are looking at alternatives, making sure that the job itself will recognize the need to iterate again without involving a full new reconcile for storagecluster. But that is out of the scope of this PR.

controllers/storagecluster/provider_server.go
nb-ohad: @leelavg @jarrpa Please take a look at this comment: https://github.com/red-hat-storage/ocs-operator/pull/2500#discussion_r1521399363

We already have Ritesh looking at the problem from a different angle

controllers/storagecluster/provider_server.go
jarrpa: Okay, I can take this out of the PR and hold on to it in my dev branches until a different solution presents itself. If there's a place where this discussion happening, feel free to loop me in on it.

controllers/storagecluster/provider_server.go
jarrpa: ACK

controllers/storagecluster/storagecluster_controller.go
iamniting: @jarrpa I think you are referring to https://github.com/red-hat-storage/ocs-operator/pull/2232

iamniting (changes requested):

controllers/storagecluster/storagecluster_controller.go
jarrpa: Sure seems so! I'll remove this, then.

deploy/ocs-operator/manifests/ocs-operator.clusterserviceversion.yaml
jarrpa: Sure, but this is an annotation. Do you need *this* for the odf-operator repo, but not the actual images in the Deployment spec?

iamniting: I would like to have the actual tag at both the places.

jarrpa: That's the thing, this *is* the actual tag. There's a difference between the CSV version and the tag used for our container images. The CSV version needs to be a valid semantic version, the tag does not. And indeed, in the CSV itself all the container images use a default tag of `latest`. This is an annotation that's supposed to be advertising the container image of the operator, so why lie?

iamniting: you are correct. While building and pushing the images I export the version tag which creates and push the images with the version.

jarrpa: So.... if that's the case, doing this would also keep this field up to date with whatever tag you're using at build time rather than having to manually update it at random intervals. Can I update this one string field to actually reflect the container image being referenced in the CSV Deployments?

controllers/storagecluster/provider_server.go
jarrpa: Done.

controllers/storagecluster/provider_server.go
jarrpa: Done.

deploy/ocs-operator/manifests/ocs-operator.clusterserviceversion.yaml
jarrpa: I'll not again, this is a *generated* change from the line changed in the csv-merger tool.

iamniting: yes lets have the actual container image version

jarrpa: Thank you! Could I get an Approved review?

iamniting (approved):

PR #2501, opened 2024-03-07
Create replica-1 sc for external mode if topology details are provided

malayparida2000:
In external mode if replica-1 is enabled & topology details are provided then create a storage class for replica-1. We construct the TopologyConstrainedPools from the topology details provided in the external cluster details secret.

### DIFF
--- a/controllers/storagecluster/external_resources.go
+++ b/controllers/storagecluster/external_resources.go
@@ -30,6 +30,7 @@ const (
 	cephFsStorageClassName                = "cephfs"
 	cephRbdStorageClassName               = "ceph-rbd"
 	cephRbdRadosNamespaceStorageClassName = "ceph-rbd-rados-namespace"
+	cephRbdTopologyStorageClassName       = "ceph-rbd-topology"
 	cephRgwStorageClassName               = "ceph-rgw"
 	externalCephRgwEndpointKey            = "endpoint"
 	cephRgwTLSSecretKey                   = "ceph-rgw-tls-cert"
@@ -371,6 +372,7 @@ func (r *StorageClusterReconciler) createExternalStorageClusterResources(instanc
 
 		case "StorageClass":
 			var scc StorageClassConfiguration
+			var err error
 			if d.Name == cephFsStorageClassName {
 				scc = newCephFilesystemStorageClassConfiguration(instance)
 				enableRookCSICephFS = true
@@ -380,6 +382,13 @@ func (r *StorageClusterReconciler) createExternalStorageClusterResources(instanc
 				scc = newCephBlockPoolStorageClassConfiguration(instance)
 				// update the storageclass name to rados storagesclass name
 				scc.storageClass.Name = fmt.Sprintf("%s-%s", instance.Name, d.Name)
+			} else if d.Name == cephRbdTopologyStorageClassName {
+				scc = newNonResilientCephBlockPoolStorageClassConfiguration(instance)
+				scc.storageClass.Parameters["topologyConstrainedPools"], err = getTopologyConstrainedPoolsExternalMode(d.Data)
+				if err != nil {
+					r.Log.Error(err, "Failed to get topologyConstrainedPools from external mode secret.", "StorageClass", klog.KRef(instance.Namespace, d.Name))
+					return err
+				}
 			} else if d.Name == cephRgwStorageClassName {
 				rgwEndpoint := d.Data[externalCephRgwEndpointKey]
 				if err := checkEndpointReachable(rgwEndpoint, 5*time.Second); err != nil {
--- a/controllers/storagecluster/storageclasses.go
+++ b/controllers/storagecluster/storageclasses.go
@@ -528,3 +528,44 @@ func getTopologyConstrainedPools(initData *ocsv1.StorageCluster) string {
 	}
 	return string(topologyConstrainedPoolsStr)
 }
+
+// getTopologyConstrainedPoolsExternalMode constructs the topologyConstrainedPools string for external mode from the data map
+func getTopologyConstrainedPoolsExternalMode(data map[string]string) (string, error) {
+	type topologySegment struct {
+		DomainLabel string `json:"domainLabel"`
+		DomainValue string `json:"value"`
+	}
+	// TopologyConstrainedPool stores the pool name and a list of its associated topology domain values.
+	type topologyConstrainedPool struct {
+		PoolName       string            `json:"poolName"`
+		DomainSegments []topologySegment `json:"domainSegments"`
+	}
+	var topologyConstrainedPools []topologyConstrainedPool
+
+	domainLabel := data["topologyFailureDomainLabel"]
+	domainValues := strings.Split(data["topologyFailureDomainValues"], ",")
+	poolNames := strings.Split(data["topologyPools"], ",")
+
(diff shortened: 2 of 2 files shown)


malayparida2000:
Depends on-https://github.com/rook/rook/pull/13821

controllers/storagecluster/external_resources.go
```diff
@@ -30,6 +30,7 @@ const (
 	cephFsStorageClassName                = "cephfs"
 	cephRbdStorageClassName               = "ceph-rbd"
 	cephRbdRadosNamespaceStorageClassName = "ceph-rbd-rados-namespace"
+	cephRbdTopologyStorageClassName       = "ceph-rbd-topology"
```
parth-gr, line 33: can we just use the name from the json blob

controllers/storagecluster/external_resources.go
```diff
@@ -380,6 +381,9 @@ func (r *StorageClusterReconciler) createExternalStorageClusterResources(instanc
 				scc = newCephBlockPoolStorageClassConfiguration(instance)
 				// update the storageclass name to rados storagesclass name
 				scc.storageClass.Name = fmt.Sprintf("%s-%s", instance.Name, d.Name)
+			} else if d.Name == cephRbdTopologyStorageClassName && instance.Spec.ManagedResources.CephNonResilientPools.Enable {
```
parth-gr, line 384: > instance.Spec.ManagedResources.CephNonResilientPools.Enable

Why we need this setting to get enabled for external mode?

controllers/storagecluster/external_resources.go
```diff
@@ -380,6 +381,9 @@ func (r *StorageClusterReconciler) createExternalStorageClusterResources(instanc
...
+				scc = newNonResilientCephBlockPoolStorageClassConfiguration(instance)
```
parth-gr, line 385: will the replicated pool will be also set, which is passed by "pool" in json

parth-gr (changes requested):

controllers/storagecluster/external_resources.go
malayparida2000: This is the key we will look for the json blob to get the details of the storageclass, exactly same process as we do for the other storageclasses

controllers/storagecluster/external_resources.go
malayparida2000: Hm, depends on the usage flow you want. If you want the storageclass to be created whenever the json has the details for it we can skip the flag for external mode.

controllers/storagecluster/external_resources.go
malayparida2000: Yes as of now that field will be set as AFAIK that's a required field. You were working on some changes to remove that field as a requirement when topologyConstrainedPools are specified, do let me know the progress of that and we can accordingly decide.

controllers/storagecluster/external_resources.go
parth-gr: Exactly we dont need another option to enable it, its totally on the ceph admin to enable it

controllers/storagecluster/external_resources.go
parth-gr: yes correct but currently we need it to set

controllers/storagecluster/storageclasses.go
```diff
@@ -528,3 +528,38 @@ func getTopologyConstrainedPools(initData *ocsv1.StorageCluster) string {
 	}
 	return string(topologyConstrainedPoolsStr)
 }
+
```
iamniting, line 531: Can you pls add a comment explaining the func.

controllers/storagecluster/storageclasses.go
```diff
@@ -528,3 +528,38 @@ func getTopologyConstrainedPools(initData *ocsv1.StorageCluster) string {
...
+	type topologyConstrainedPool struct {
+		PoolName       string            `json:"poolName"`
+		DomainSegments []topologySegment `json:"domainSegments"`
+	}
+	var topologyConstrainedPools []topologyConstrainedPool
+
+	domainLabel := data["topologyFailureDomainLabel"]
+	domainValues := strings.Split(data["topologyFailureDomainValues"], ",")
+	poolNames := strings.Split(data["topologyPools"], ",")
+
```
iamniting, line 547: Can you pls check if we can construct the structure directly from the json? using marshal/unmarshal etc.

controllers/storagecluster/external_resources.go
```diff
@@ -380,6 +381,9 @@ func (r *StorageClusterReconciler) createExternalStorageClusterResources(instanc
...
+				scc.storageClass.Parameters["topologyConstrainedPools"] = getTopologyConstrainedPoolsExternalMode(d.Data)
```
iamniting, line 386: @parth-gr I think it is better to have this constructed and passed in the JSON blob itself instead of building it in the ocs-operator.

iamniting (changes requested):

controllers/storagecluster/external_resources.go
parth-gr: @iamniting I would like to use the power of go code for this case as it is more straightforward, in the python script we have to create a template like we did it in import shell script, https://github.com/rook/rook/pull/13821/files#diff-17c14009bbe3863a3d47d44e3648b462b4269bb8a98f38492bc597321d053816 

if you have any easy maintainable way in python please let me know.

parth-gr: moreover script is just to extract/scrape information, styling would be done in the next level of code.

iamniting: okay in that case we will do the magic here itself.

controllers/storagecluster/storageclasses.go
```diff
@@ -528,3 +528,39 @@ func getTopologyConstrainedPools(initData *ocsv1.StorageCluster) string {
...
+	// TopologyConstrainedPool stores the pool name and a list of its associated topology domain values.
+	type topologyConstrainedPool struct {
+		PoolName       string            `json:"poolName"`
+		DomainSegments []topologySegment `json:"domainSegments"`
+	}
+	var topologyConstrainedPools []topologyConstrainedPool
+
+	domainLabel := data["topologyFailureDomainLabel"]
+	domainValues := strings.Split(data["topologyFailureDomainValues"], ",")
+	poolNames := strings.Split(data["topologyPools"], ",")
```
parth-gr, line 547: for debugging add print here and validate what are the values, plus then try to print `topologyConstrainedPools` after the for loop before marshaling

malayparida2000:
Tested the topologyConstarinedPools format from the json given by Parth from the rook PR. It looks good.
```
[
  {
    "poolName": "replica1a",
    "domainSegments": [
      {
        "domainLabel": "hostname",
        "value": "minikube"
      }
    ]
  },
  {
    "poolName": "replica1b",
    "domainSegments": [
      {
        "domainLabel": "hostname",
        "value": "minikube-m02"
      }
    ]
  },
  {
    "poolName": "replica1c",
    "domainSegments": [
      {
        "domainLabel": "hostname",
        "value": "minikube-m03"
      }
    ]
  }
]
```

iamniting (approved):
/hold for rook PR to get merged.

PR #2504, opened 2024-03-11
RDR: exporting client info

raaizik:
Handles exporting the incoming data found in the on-boarded consumer's request into the consumer's status, that in turn will populate the ODF info CM's `ConnectedClient`s via the `StorageCluster`'s reconciler (addressed as part of #2440).


api/v1alpha1/storageconsumer_types.go
```diff
@@ -72,6 +72,10 @@ type ClientStatus struct {
 	PlatformVersion string `json:"platformVersion"`
 	// StorageClient Operator Version
 	OperatorVersion string `json:"operatorVersion"`
+	// ClusterID is the id of the openshift cluster
+	ClusterID string `json:"clusterId"`
+	// StorageClientNamespacedName is the name and namespace of the StorageClient
+	StorageClientNamespacedName string `json:"storageClientNamespacedName"`
```
nb-ohad, line 78: This field is defined inside the consumer status inside the client's section so I think the `StorageClient` prefix is an overkill. I would suggest calling this field `NamespacedName` so accessing it would look like:
`consumer.Status.Client.NamespacedName`

services/provider/interfaces/interfaces.go
```diff
@@ -6,6 +6,8 @@ type StorageClientStatus interface {
 	// no client is using it
 	GetPlatformVersion() string
 	GetOperatorVersion() string
+	GetClusterId() string
+	GetStorageClientNamespacedName() string
```
nb-ohad, line 10: I don't think we need to report this information on the status message as it is always constant and we are already decided to report these in the onboard message

services/provider/interfaces/interfaces.go
```diff
@@ -16,6 +18,8 @@ type StorageClientOnboarding interface {
 	GetOnboardingTicket() string
 	GetConsumerName() string
 	GetClientOperatorVersion() string
+	GetClusterId() string
+	GetStorageClientNamespacedName() string
```
nb-ohad, line 22: As of a similar reason to a previous comment, I would suggest the following rename:
 ```suggestion
	     GetNamespacedName() string
```

Also, aren't we missing the compliment `SetNamespacedName`?

File: services/provider/proto/provider.proto [hunk 4]
```diff
@@ -49,6 +49,10 @@ message OnboardConsumerRequest{
     string consumerName = 2;
     // clientOperatorVersion is the semver version of ocs-client-operator
     string clientOperatorVersion = 3;
+    // clusterID is the id of the openshift cluster
+    string clusterId = 4;
+    // storageClientNamespacedName is the name and namespace of the StorageClient
+    string storageClientNamespacedName = 5;
```
nb-ohad, line 55: The same suggestion about naming also applies to the proto definition here

services/provider/proto/provider.proto
```diff
@@ -159,6 +163,8 @@ message ReportStatusRequest{
     string storageConsumerUUID = 1;
     string clientPlatformVersion = 2;
     string clientOperatorVersion = 3;
+    string clusterId = 4;
+    string storageClientNamespacedName = 5;
```
nb-ohad, line 167: I don't see a need to report these as part of the status request (they are constant and static)

services/provider/server/consumer.go
```diff
@@ -215,6 +217,8 @@ func (c *ocsConsumerManager) UpdateConsumerStatus(ctx context.Context, id string
 	consumerObj.Status.LastHeartbeat = metav1.Now()
 	consumerObj.Status.Client.PlatformVersion = status.GetPlatformVersion()
 	consumerObj.Status.Client.OperatorVersion = status.GetOperatorVersion()
+	consumerObj.Status.Client.ClusterID = status.GetClusterId()
+	consumerObj.Status.Client.StorageClientNamespacedName = status.GetStorageClientNamespacedName()
```
nb-ohad, line 221: Continuing from my last comment: 
```suggestion
```

nb-ohad (changes requested):

services/provider/pb/storageclient.go
```diff
@@ -42,3 +42,13 @@ func (o *OnboardConsumerRequest) SetClientOperatorVersion(version string) ifaces
...
+
+func (o *OnboardConsumerRequest) SetClusterID(clusterID string) ifaces.StorageClientOnboarding {
+	o.ClusterId = clusterID
+	return o
+}
+
+func (o *OnboardConsumerRequest) SetNamespacedName(namespacedName string) ifaces.StorageClientOnboarding {
+	o.NamespacedName = namespacedName
+	return o
+}
```
nb-ohad, line 54: Aren't we missing implementations? I believe that without them  `OnboardConsumerRequest` would not satisfy the 
`ifaces.StorageClientOnboarding` interface.

nb-ohad (changes requested):

services/provider/pb/storageclient.go
raaizik: All the getters are in [provider.pb.go](https://github.com/raaizik/ocs-operator/blob/rdr-odfinfo-server/services/provider/pb/provider.pb.go#L122-L155)

raaizik: No, that's where the existing ones were and I just added mine. Can move all of them to this file.

services/provider/server/consumer.go
```diff
@@ -79,6 +79,8 @@ func (c *ocsConsumerManager) Create(ctx context.Context, onboard ifaces.StorageC
 		Status: ocsv1alpha1.StorageConsumerStatus{
 			Client: ocsv1alpha1.ClientStatus{
 				OperatorVersion: onboard.GetClientOperatorVersion(),
+				ClusterID:       onboard.GetClusterId(),
+				NamespacedName:  onboard.GetNamespacedName(),
```
leelavg, line 83: 1. there is an existing bug here, updating the status here doesn't stick because the storageconsumer controller is probably resetting it, yet to be investigated, however operatorversion is refreshed by hearbeat mechanism and so this was a bit safer for operatorversion but not for the additional params
2. if by mistake the status is cleaned additional params will get erased w/ no way of retrieving them and so I feel status section for these constant fields isn't a good idea

services/provider/pb/storageclient.go
raaizik: @leelavg that's incorrect. The `make build` command executes `gen-protobuf` and nothing of this sort got auto-generated, hence why I needed to manually implement it. There's no sense, nor is it recommended, in editing the pb file, that's why I suggested moving them to this file.

api/v1alpha1/storageconsumer_types.go
```diff
@@ -72,6 +72,10 @@ type ClientStatus struct {
 	PlatformVersion string `json:"platformVersion"`
 	// StorageClient Operator Version
 	OperatorVersion string `json:"operatorVersion"`
+	// ClusterID is the id of the openshift cluster
+	ClusterID string `json:"clusterId"`
+	// NamespacedName is the name and namespace of the StorageClient
+	NamespacedName string `json:"namespacedName"`
```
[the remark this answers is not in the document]
raaizik: 1. There are several kinds of cluster identifiers. Extracting a _possibly_ embedded one from a `StorageConsumer`'s name instead of just having it as a standalone field seems like an unnecessary detour. 
2. It's already been decided. The dynamic fields you mentioned are reported as part of heartbeat mechanism. If you take a closer look, the above static fields are added **only** during the creation of `StorageConsumer`, and not via its update function (which is the actual "refresh").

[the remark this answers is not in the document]
nb-ohad: @leelavg We are not considering naming conventions as source a of truth. We used the cluster ID in the name because it allowed us to easily create a distinct name for each client. But now we need  the cluster id as a piece of information so we should provide it explicitly, exactly as @raaizik did

services/provider/interfaces/interfaces.go
```diff
@@ -16,8 +16,12 @@ type StorageClientOnboarding interface {
 	GetOnboardingTicket() string
 	GetConsumerName() string
 	GetClientOperatorVersion() string
+	GetClusterId() string
```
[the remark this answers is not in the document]
nb-ohad: I think we should align with what pb generator outputs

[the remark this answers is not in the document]
leelavg: I meant to change proto from `clusterId` to `clusterID`, asked here because the current func convention isn't consistent `GetClusterId()` and `SetClusterID()`

services/provider/server/consumer.go
raaizik: Whether it'd stick or not in the long run shouldn't be of any concern here since those fields' values are queried at the very moment a `StorageConsumer` resource gets created (it's watched by `StorageCluster`). From that point forward it's set in the CM.

nb-ohad: @leelavg I am not sure about #1, but you do have a point with #2. 
From the start, I pushed toward NOT sending the data again on every heartbeat (and only doing it once as part of the onboarding), but that didn't takes into consideration that status should not be a source of truth. 

With that in mind, maybe we should move all of these fields to the heartbeat instead of the onboarding message. @raaizik what do you think?

services/provider/interfaces/interfaces.go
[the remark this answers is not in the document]
nb-ohad: +1 if we are introducing the change in the proto.

api/v1alpha1/storageconsumer_types.go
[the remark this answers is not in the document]
nb-ohad: @leelavg Can this be resolved?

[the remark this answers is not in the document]
leelavg: I asked for extracting the id due to it's nature being in constant throughtout the lifecycle of storageconsumer as there's no other construct which provides such durability of the value.

> If you take a closer look, the above static fields are added only during the creation of StorageConsumer, and not via its update function (which is the actual "refresh").
- pls excuse, the answer itself surfaces the issue here, ie, `added only during the creation`, however as it is tracked in another thread, closing this.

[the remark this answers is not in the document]
leelavg: Can be resolved.

services/provider/interfaces/interfaces.go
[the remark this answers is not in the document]
raaizik: Annoyingly enough linting fails with the lowercase, so changed to uppercase.

services/provider/proto/provider.proto
```diff
@@ -159,6 +159,10 @@ message ReportStatusRequest{
     string storageConsumerUUID = 1;
     string clientPlatformVersion = 2;
     string clientOperatorVersion = 3;
+  // clusterID is the id of the openshift cluster
+  string clusterID = 4;
+  // NamespacedName is the name and namespace of the StorageClient
+  string NamespacedName = 5;
```
nb-ohad, line 165: I think there is an issue with indentation here (take a look at line 161 and above for reference)

nb-ohad (changes requested):

PR #2508, opened 2024-03-14
disable holder pods in greenfield clusters

BlaineEXE:
When deploying new StorageClusters, ocs-operator should apply the new Rook operator config `CSI_REMOVE_HOLDER_PODS: "true"`. This reflects the new config and default value that Rook specifies in example manifests here: https://github.com/rook/rook/pull/13890

This new value should be applied to the `rook-ceph-operator-config` only during installation of new (greenfield) StorageClusters.

It is required that ocs-operator **should not** update this value in existing (brownfield) clusters.

### DIFF
--- a/controllers/ocsinitialization/ocsinitialization_controller.go
+++ b/controllers/ocsinitialization/ocsinitialization_controller.go
@@ -273,6 +273,7 @@ func (r *OCSInitializationReconciler) ensureOcsOperatorConfigExists(initialData 
 		util.EnableTopologyKey:           r.getEnableTopologyKeyValue(),
 		util.TopologyDomainLabelsKey:     r.getTopologyDomainLabelsKeyValue(),
 		util.EnableNFSKey:                r.getEnableNFSKeyValue(),
+		util.DisableHolderKey:            "true", // disable holder pods for all greenfield clusters
 	}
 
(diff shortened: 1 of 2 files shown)


malayparida2000 (changes requested):
Hey Blaine, here the added value just went to the ocs-operator-config CM. We also need to add them as env variables to the rook-csv in the csv-merger https://github.com/red-hat-storage/ocs-operator/blob/c65c08c6ff670afe39890ffda5117d2ab7749ca1/tools/csv-merger/csv-merger.go#L306(so that it reads the value from the configmap). 

But that process is also undergoing a change right now. https://github.com/red-hat-storage/ocs-operator/pull/2492#discussion_r1513928135. 

Also if we add the value this way it will be also applicable to existing clusters. We have to think of some hack to avoid setting the value in existing clusters.

iamniting (changes requested):
@BlaineEXE We currently lack the mechanism to determine whether the cluster is a new one or an upgraded one. How do you think we can achieve what we wanted to do here.

BlaineEXE:
@malayparida2000 I see I made a mistake here. Thanks for catching that.

I'm not sure how ocs-operator handles the CSV, so I am not quite sure what to do here. I'll close this PR, and I think I will need some help accomplishing the goals here.

PR #2507, opened 2024-03-14
Update rook api and image to latest version and also updated the default service account in toolbox pod

Nikhil-Ladha:
- Updated rook api, image to latest version
- We are now using latest downstream rook image, i.e, red-hat-storage/rook since we are now actively maintaining the downstream fork and there are some downstream only changes that is required for the ocs-operator to run properly.
- Updated the code to use VolumeClaimTemplate instead of PersistentVolumeClaim
- Added default service account for toolbox pod, as the rook scc is now updated and the default service account is removed.

PR #2511, opened 2024-03-15
toolbox: add default service account to toolbox

parth-gr:
currently scc are updated to remove the k8s default service account, so toolbox should start with rook-ceph-default service account

### DIFF
--- a/api/v1/storagecluster_types.go
+++ b/api/v1/storagecluster_types.go
@@ -683,7 +683,8 @@ func (r *StorageCluster) NewToolsDeployment(tolerations []corev1.Toleration) *ap
 					},
 				},
 				Spec: corev1.PodSpec{
-					DNSPolicy: corev1.DNSClusterFirstWithHostNet,
+					DNSPolicy:          corev1.DNSClusterFirstWithHostNet,
+					ServiceAccountName: "rook-ceph-default",
 					Containers: []corev1.Container{
 						{
 							Name:    name,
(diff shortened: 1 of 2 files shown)


jarrpa (approved):
/lgtm

jarrpa:
Looks like the e2e-test is failing at deploying the toolbox pod. Should definitely look into it.

parth-gr:
@iamniting how to understand what is failing in the CI, can we look at the CI cluster somehow?

And I believe we would have the latest csv change to create the default service account in the ci https://github.com/red-hat-storage/rook/pull/589

parth-gr:
@jarrpa I cant find the CI failing because of toolbox, it looks like a different issue,
Plus we need the update of the rook image in the OCS CI which is coming from https://github.com/red-hat-storage/ocs-operator/pull/2507
So they both are dependent to eachother

jarrpa:
If it's a strict dependency then it should be marked as such in the PR description, and we gotta wait. Something like this that messes with RBAC should not be merged without the e2e-tests passing, it could destabilize the downstream builds.

parth-gr:
> If it's a strict dependency then it should be marked as such in the PR description, and we gotta wait. Something like this that messes with RBAC should not be merged without the e2e-tests passing, it could destabilize the downstream builds.

Sure I believe we can merge the two PR.

parth-gr:
Closing as the api change PR has the above change

PR #2513, opened 2024-03-19
Add CSI_REMOVE_HOLDER_PODS key and its value in the ocs configmap

iamniting:
set the CSI_REMOVE_HOLDER_PODS to true for new users on version 4.16 onwards and set it to false for older/upgraded user to give them time to migrate manually.

with version 4.17 we will make the CSI_REMOVE_HOLDER_PODS to true for all customers older and new both.

If the configmap is being created for the first time we are considering it as a new install.

Jira: https://issues.redhat.com/browse/RHSTOR-5340?focusedId=24357013&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-24357013

### DIFF
--- a/controllers/ocsinitialization/ocsinitialization_controller.go
+++ b/controllers/ocsinitialization/ocsinitialization_controller.go
@@ -345,9 +345,23 @@ func (r *OCSInitializationReconciler) ensureOcsOperatorConfigExists(initialData 
 			Name:      util.OcsOperatorConfigName,
 			Namespace: initialData.Namespace,
 		},
-		Data: ocsOperatorConfigData,
 	}
 	opResult, err := ctrl.CreateOrUpdate(r.ctx, r.Client, ocsOperatorConfig, func() error {
+
+		// If the configmap is being created for the first time, set the entry for
+		// CSI_REMOVE_HOLDER_PODS to "true". This configuration is applied for new clusters
+		// starting from ODF version 4.16 onwards. For old or upgraded clusters,
+		// it's initially set to "false", allowing users time to manually migrate from holder pods.
+		// In ODF version 4.17, we will universally set it to "true" for all users.
+
+		if ocsOperatorConfig.CreationTimestamp.IsZero() {
+			ocsOperatorConfigData[util.CsiRemoveHolderPodsKey] = "true"
+		} else if ocsOperatorConfig.Data[util.CsiRemoveHolderPodsKey] == "" {
+			ocsOperatorConfigData[util.CsiRemoveHolderPodsKey] = "false"
+		} else if ocsOperatorConfig.Data[util.CsiRemoveHolderPodsKey] != "" {
+			ocsOperatorConfigData[util.CsiRemoveHolderPodsKey] = ocsOperatorConfig.Data[util.CsiRemoveHolderPodsKey]
+		}
+
 		if !reflect.DeepEqual(ocsOperatorConfig.Data, ocsOperatorConfigData) {
 			r.Log.Info("Updating ocs-operator-config configmap")
 			ocsOperatorConfig.Data = ocsOperatorConfigData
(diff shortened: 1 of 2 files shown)


iamniting:
@subhamkrai once this PR is merged can you pls read this key from the ocs-operator-config configmap in the rook deployment?

controllers/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -345,9 +345,23 @@ func (r *OCSInitializationReconciler) ensureOcsOperatorConfigExists(initialData
 			Name:      util.OcsOperatorConfigName,
 			Namespace: initialData.Namespace,
 		},
-		Data: ocsOperatorConfigData,
```
malayparida2000, line 348: I guess the data section was there do that when the CM is created the data section is not null, will removing this not create any issue?

malayparida2000 (changes requested):
The bundle separation PR https://github.com/red-hat-storage/ocs-operator/pull/2492 is not yet merged. So it will be appropriate to add the change for reading this as env variable to the csv in the csv-merger. This will make sure after that PR is merged all the mentioned env vars are properly migrated to the rook ds repo csv.

controllers/ocsinitialization/ocsinitialization_controller.go
iamniting: No, it does not create any issues, I have tested it.

subhamkrai:
> @subhamkrai once this PR is merged can you pls read this key from the ocs-operator-config configmap in the rook deployment?

suree

iamniting:
> The bundle separation PR #2492 is not yet merged. So it will be appropriate to add the change for reading this as env variable to the csv in the csv-merger. This will make sure after that PR is merged all the mentioned env vars are properly migrated to the rook ds repo csv.

I am not doing that because if we do so we will have merge conflicts in the other PR where we are removing the rook CSV.

controllers/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -345,9 +345,23 @@ func (r *OCSInitializationReconciler) ensureOcsOperatorConfigExists(initialData
...
+		// starting from ODF version 4.16 onwards. For old or upgraded clusters,
+		// it's initially set to "false", allowing users time to manually migrate from holder pods.
+		// In ODF version 4.17, we will universally set it to "true" for all users.
+
+		if ocsOperatorConfig.CreationTimestamp.IsZero() {
+			ocsOperatorConfigData[util.CsiRemoveHolderPodsKey] = "true"
+		} else if ocsOperatorConfig.Data[util.CsiRemoveHolderPodsKey] == "" {
+			ocsOperatorConfigData[util.CsiRemoveHolderPodsKey] = "false"
+		} else if ocsOperatorConfig.Data[util.CsiRemoveHolderPodsKey] != "" {
+			ocsOperatorConfigData[util.CsiRemoveHolderPodsKey] = ocsOperatorConfig.Data[util.CsiRemoveHolderPodsKey]
```
subhamkrai, line 362:
```suggestion
		} else if ocsOperatorConfig.Data[util.CsiRemoveHolderPodsKey] != "" {
			ocsOperatorConfigData[util.CsiRemoveHolderPodsKey] = ocsOperatorConfig.Data[util.CsiRemoveHolderPodsKey]
```

do we need this else if?

iamniting: yes we do, this is the third case where we want to keep the value that was set. If we don't do that it will always remove it.

malayparida2000:
> > The bundle separation PR #2492 is not yet merged. So it will be appropriate to add the change for reading this as env variable to the csv in the csv-merger. This will make sure after that PR is merged all the mentioned env vars are properly migrated to the rook ds repo csv.
> 
> I am not doing that because if we do so we will have merge conflicts in the other PR where we are removing the rook CSV.

That should be okay right? I just want to make sure that all the env vars defined in the csv merger get transferred directly to the rook ds. If we don't want the merge conflict, IMO we should wait till the bundle separation PR merges and then merge this.

controllers/ocsinitialization/ocsinitialization_controller.go
subhamkrai: since we are setting the same value, where are we removing it if we don't set?

iamniting: actually if i do not use this 3rd case the value will be removed as when the variable was created this key was not part of the map and it will be replaced.

subhamkrai (approved):

PR #2516, opened 2024-03-20
Set values in ocs-operator-config CM for replica-1 in external mode

malayparida2000:
In external mode the replica-1 feature is not enabled via the
storageCluster CR. So detect the feature is enabled by checking if the
replica-1 storageClass exists. In external mode the failure domain is
not set in the storageCluster CR, so determine the failure domain key
by using the parameter of the storageClass. Also add a watch for
storageClass in the ocsinitialization controller to detect the
storageClass creation. Accordingly set the values in the CM.

### DIFF
--- a/controllers/ocsinitialization/ocsinitialization_controller.go
+++ b/controllers/ocsinitialization/ocsinitialization_controller.go
@@ -14,6 +14,7 @@ import (
 	rookCephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
 	"gopkg.in/yaml.v2"
 	corev1 "k8s.io/api/core/v1"
+	storagev1 "k8s.io/api/storage/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -208,6 +209,23 @@ func (r *OCSInitializationReconciler) SetupWithManager(mgr ctrl.Manager) error {
 				},
 			),
 		).
+		// Watcher for storageClass required to update values related to replica-1
+		// in ocs-operator-config configmap, if storageClass changes
+		Watches(
+			&storagev1.StorageClass{},
+			handler.EnqueueRequestsFromMapFunc(
+				func(context context.Context, obj client.Object) []reconcile.Request {
+					// Only reconcile if the storageClass has topologyConstrainedPools set
+					sc := obj.(*storagev1.StorageClass)
+					if sc.Parameters["topologyConstrainedPools"] != "" {
+						return []reconcile.Request{{
+							NamespacedName: InitNamespacedName(),
+						}}
+					}
+					return []reconcile.Request{}
+				},
+			),
+		).
 		// Watcher for rook-ceph-operator-config cm
 		Watches(
 			&corev1.ConfigMap{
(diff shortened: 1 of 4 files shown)


controllers/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -425,6 +441,35 @@ func (r *OCSInitializationReconciler) getEnableNFSKeyValue() string {
...
+		DomainSegments []topologySegment `json:"domainSegments"`
+	}
+	var topologyConstrainedPools []topologyConstrainedPool
+
+	tcPools := sc.Parameters["topologyConstrainedPools"]
+	err := json.Unmarshal([]byte(tcPools), &topologyConstrainedPools)
+	if err != nil {
+		return ""
+	}
+	failuredomain := topologyConstrainedPools[0].DomainSegments[0].DomainLabel
```
iamniting, line 460: Can you add the check here before accessing the index?

controllers/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -403,11 +412,18 @@ func (r *OCSInitializationReconciler) getEnableTopologyKeyValue() string {
 
 func (r *OCSInitializationReconciler) getTopologyDomainLabelsKeyValue() string {
 
-	// return value from the internal storagecluster as failureDomain is set only in internal cluster
 	for _, sc := range r.clusters.GetStorageClusters() {
-		if !sc.Spec.ExternalStorage.Enable && sc.Status.FailureDomainKey != "" {
+		if !sc.Spec.ExternalStorage.Enable {
```
iamniting, line 416: lets add a check here if it is replica 1 then only pass it.

iamniting (changes requested):

controllers/ocsinitialization/ocsinitialization_controller.go
parth-gr: exactly and I don't know how the ocsinitialization will know that. As we can get the info only from the json blob
So we are adding it with `(r *OCSInitializationReconciler) Reconcile(`  does it calls after `func (r *StorageClusterReconciler) Reconcile(ctx`, or they can be in parallel.

parth-gr: if they are in parallel we need to separately do the unmarshal of the secret, or we should keep this update with the normal `externalreource` call  where we create the storageclass

malayparida2000: ocs initialization will know about replica-1 by the presence of the replica-1 storageclass, and accordingly, it will set the values.

iamniting (approved):

controllers/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -391,22 +393,39 @@ func (r *OCSInitializationReconciler) ensureOcsOperatorConfigExists(initialData
...
 }
 
+// In case of multiple storageClusters when replica-1 is enabled for both an internal and an external cluster, different failure domain keys can lead to complications.
+// To prevent this, when gathering information for the external cluster, ensure that the failure domain is specified to match that of the internal cluster (sc.Status.FailureDomain).
 func (r *OCSInitializationReconciler) getTopologyDomainLabelsKeyValue() string {
 
-	// return value from the internal storagecluster as failureDomain is set only in internal cluster
 	for _, sc := range r.clusters.GetStorageClusters() {
-		if !sc.Spec.ExternalStorage.Enable && sc.Status.FailureDomainKey != "" {
+		if !sc.Spec.ExternalStorage.Enable && sc.Spec.ManagedResources.CephNonResilientPools.Enable {
```
parth-gr, line 418: we arent setting this filled for external cluster storagecluster, so this will be always false

parth-gr (changes requested):

controllers/ocsinitialization/ocsinitialization_controller.go
malayparida2000: The case you are referring to is of an internal cluster, the external cluster case is in the next else if, also we have added a note.

parth-gr (approved):
lgtm

malayparida2000:
/hold cancel
Tested the code & added a watch for the storageclass as necessary. Now works as expected.

controllers/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -208,6 +210,23 @@ func (r *OCSInitializationReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 		).
+		// Watcher for storageClass required to update values related to replica-1
+		// in ocs-operator-config configmap, if storageClass changes
+		Watches(
+			&storagev1.StorageClass{},
+			handler.EnqueueRequestsFromMapFunc(
+				func(context context.Context, obj client.Object) []reconcile.Request {
+					// Only reconcile if the storageClass is external & has topologyConstrainedPools set
+					sc := obj.(*storagev1.StorageClass)
+					if strings.Contains(sc.Name, "external") && sc.Parameters["topologyConstrainedPools"] != "" {
```
iamniting, line 221: Can we pls remove the first condition? There might be customers who do not have an external name in the storageclasses. They may have created the storage cluster from the CLI.

iamniting (changes requested):

iamniting (approved):

PR #2519, opened 2024-03-21
controllers: do not skip cephfs creation in provider mode

leelavg:
we do want to create default cephfs in provider mode however the svg is left untouched as they will be created on demand based on requests from consumer.


leelavg:
generated diff isn't easy to read, the main change is removal of if condition and else block.

PR #2518, opened 2024-03-21
provider: check for RADOS namespace CephResource

jarrpa:
Signed-off-by: Jose A. Rivera <jarrpa@redhat.com>

PR #2517, opened 2024-03-20
controllers: remove storageprofiles from storageclassrequest 

jarrpa:
Signed-off-by: Jose A. Rivera <jarrpa@redhat.com>

PR #2523, opened 2024-03-25
label ceph svg correctly if datapool name is empty

leelavg:
when the datapool in cephfs is empty (aka default) the constructed resource label value ends with hyphen(`-`) which isn't valid. This commit adds the suffix only if the datapool isn't empty (ref #2517)

add rbac for provider-sa to query ceph radosnamespace (ref #2518)


controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -416,7 +416,11 @@ func (r *StorageClassRequestReconciler) reconcileCephFilesystemSubVolumeGroup()
 		// This label is required to set the dataPool on the CephFS
 		// storageclass so that each PVC created from CephFS storageclass can
 		// use correct dataPool backed by deviceclass.
-		addLabel(r.cephFilesystemSubVolumeGroup, v1alpha1.CephFileSystemDataPoolLabel, fmt.Sprintf("%s-%s", cephFilesystem.Name, dataPool.Name))
+		labelValue := cephFilesystem.Name
+		if dataPool.Name != "" {
+			labelValue = fmt.Sprintf("%s-%s", labelValue, dataPool.Name)
+		}
+		addLabel(r.cephFilesystemSubVolumeGroup, v1alpha1.CephFileSystemDataPoolLabel, labelValue)
```
Madhu-1, line 423: we should not set the label if there is no data pool for the filesystem and we are going to use default one, and also https://github.com/red-hat-storage/ocs-operator/blob/b2c823679e4bd7c633483df193af80be11d0aced/services/provider/server/server.go#L650 this should be inside if check where set `pool` only if there is a Label set or set the default dataPool name always.

deploy/ocs-operator/manifests/provider-role.yaml
```diff
@@ -15,6 +15,7 @@ rules:
   - ceph.rook.io
   resources:
   - cephfilesystemsubvolumegroups
+  - cephblockpoolradosnamespaces
```
Madhu-1, line 18: remove blockpool RBAC like Create/Delete and add this at the top of the function before `Reconcile` like below and let these gets auto generated.

// +kubebuilder:rbac:groups=ceph.rook.io,resources=cephblockpools,verbs=get;list;watch;create;update;delete

Madhu-1 (changes requested):

controllers/storageclassrequest/storageclassrequest_controller.go
leelavg: > we should not set the label if there is no data pool for the filesystem
- ack.

> this should be inside if check where set pool only if there is a Label 
- at client side the value is just copied w/o any checks? should this be in a `if` as non-existence of label returns an empty value?

deploy/ocs-operator/manifests/provider-role.yaml
leelavg: I don't think so, rbac for provider-sa should be hand written.

Madhu-1: Yes correct but the controller need to be adjusted for the above RBAC but it can be a separate PR.

controllers/storageclassrequest/storageclassrequest_controller.go
leelavg: done.

deploy/ocs-operator/manifests/provider-role.yaml
leelavg: > but it can be a separate PR.
- no, will include in this only, it's hard to track many, I saw StorageClaim got ready after for block when I updated the rbac for provider-sa, for controller I believe there's a large net like `can do anything for all resources in ceph.rook.io`.

will recheck though.

Madhu-1 (approved):
LGTM

deploy/ocs-operator/manifests/provider-role.yaml
leelavg: verified, rbac for controllers already exist here
https://github.com/red-hat-storage/ocs-operator/blob/b2c823679e4bd7c633483df193af80be11d0aced/config/rbac/role.yaml#L39-L51

leelavg: done.

Madhu-1 (commented):
LGTM

leelavg:
rebased to pick CI changes.

leelavg:
rebased.

PR #2465, opened 2024-02-19
Added support for Prometheus

ezio-auditore:
Story link : [RHSTOR-5367](https://issues.redhat.com/browse/RHSTOR-5367)
Things added in this PR

- Prometheus instance
- Prometheus Service
- AlertManager instance
- Prometheus behing TLS
- Check for ROSA HCP platform before setting up monitoring resources

PR #2528, opened 2024-03-25
Added support for Prometheus

iamniting:
replica of #2465

just for testing and debugging purpose


iamniting:
Original PR is merged.

PR #2537, opened 2024-03-28
provider: avoid CSI secret name collisions 

jarrpa:
Previously, when fetching a storage claim config, a remote consumer was
given information for creating the Ceph CSI Secrets that was identical
to the secret names on the provider side. Running a consumer in the same
namespace as the provider introduced namespace collisions.

Hilariously, by virtue of being the same information, things just kinda
worked out anyway. Still, this is likely going to produce massive
problems later, so better to fix it now.


services/provider/server/server.go
```diff
@@ -590,16 +597,6 @@ func (s *OCSProviderServer) GetStorageClassClaimConfig(ctx context.Context, req
...
 
-			nodeCephClientSecret, _, err := s.getCephClientInformation(ctx, cephRes.CephClients["node"])
-			if err != nil {
-				return nil, status.Error(codes.Internal, err.Error())
-			}
-
-			provisionerCephClientSecret, _, err := s.getCephClientInformation(ctx, cephRes.CephClients["provisioner"])
-			if err != nil {
-				return nil, status.Error(codes.Internal, err.Error())
-			}
```
leelavg, line 601: just so I understand correctly, if ceph clients doesn't exist (or created) rather than failing here, we are pushing the handling to client side?

taking block as an example, suppose we fail to reconcile rbdprovisioner and rbdnode secrets for cephclient-node will not exist, is it fine?

only a question, not requesting a mandatory change.

leelavg: My point is a bit different, the clients are reconciled sequentially and in the refactor you are generating the secret names even w/o checking the Ceph client is created or not.

Switch might not help if the status for Ceph client isn't updated at all, then also you'll update storageclasses w/ not existing secret names.

The repercussion is choosing b/n these, if storageclass is created at client end we also have referenced secret created (existing) vs even if storageclass is created at client end we aren't guaranteed that referenced secret is also created (refractor).

Am I making sense?

jarrpa: ...not, I'm not following you. This function is just generating information to send to the consumer. So *prior* to anything being sent to the consumer we're already validating the existence of the CephClients, and this PR does not change that behavior. 

Before this PR, for a given StorageClassRequest, we loop through `Status.CephResources` and either 1. Validate the existence of a CephClient, inspect its existing provider-side Secret, and then forward that information for the consumer to create a Secret on its side with a new (generated) name. Or 2. For any rados namespace or subvolgroup, loop through its associated CephClients, inspect their Status for the existence of a Secret name.

However, at this point we no longer care about 2 since we're not telling the consumer to create Secrets based on the provider-side Secret names, and 1 takes care of ensuring the provider-side Secrets exist for both CephClients.

Am I still missing the point? Sorry if I am, it's been a rough morning.

leelavg: Just rechecked, the loop gets triggered only after phase is ready and so the scenario I mention will not be hit.

Only a minor comment, could you pls construct the name after the status check?

jarrpa: Sure, done.

leelavg:
LGTM.

services/provider/server/server.go
```diff
@@ -553,13 +557,18 @@ func (s *OCSProviderServer) GetStorageClassClaimConfig(ctx context.Context, req
 	var storageClassName string
 	storageClassData := map[string]string{}
 
+	storageClassRequestHash := strings.TrimPrefix(storageClassRequest.Name, "storageclassrequest-")
```
nb-ohad, line 560: I think it is a bad idea to rely on naming conventions for required information.  

What is the exact information stored on the name that you are trying to access? and do we have a way to rebuild it directly here?

services/provider/server/server.go
```diff
@@ -553,13 +557,18 @@ func (s *OCSProviderServer) GetStorageClassClaimConfig(ctx context.Context, req
...
+	nodeSecretName := scrCephClientCsiSecretName("node", storageClassRequestHash)
+	provisionerSecretName := scrCephClientCsiSecretName("provisioner", storageClassRequestHash)
```
nb-ohad, line 562: Why are we generating names for client types that we are not sure would even have secrets provided for? 
The source of truth should be the resource list, not the code.

services/provider/server/server.go
nb-ohad, line 601: This check is important! even if you don't need the information again you should check the existence of the "node" and "provisioner" entries into the `cephRes` before generating names. This solves 2 problems with:

1. The assumption for the existence of specific resources on the res list. 
2. The pre generation of names based on code constants as opposed to the data-driven approach from the current code.

nb-ohad (changes requested):

services/provider/server/server.go
jarrpa: We are already verifying that the Secret information exists in the `CephClient` case. I don't know why the design decision was made, but each CephClient has two entries in the resource list, once at the top level and again within each given blockpool or subvol entry. And they're always the same.

services/provider/server/server.go
jarrpa: ..... we are sure, though. Both StorageClasses are using CSI provisioners, which means they will have the requisite Secrets. I'll also add that we are generating the resource list to send to the consumer from this point, so it's not like we're just passing a 1-to-1 list of extant resources.

services/provider/server/server.go
jarrpa: The name generation is from [here](https://github.com/red-hat-storage/ocs-operator/blob/main/services/provider/server/storageclassrequest.go#L36-L53) and it's already being called [here](https://github.com/red-hat-storage/ocs-operator/blob/main/services/provider/server/storageclassrequest.go#L160) and being passed to us from [here](https://github.com/red-hat-storage/ocs-operator/blob/main/services/provider/server/server.go#L525). Given that the SCR name is namespaced, it's safe to assume that `ceph-client-<type>-<hash>` will also be unique within the same scope as long as `<type>` isn't duplicated. Which per our implementation it isn't.

Also, best I know I'm just implementing what was decided in the meeting on Thursday. Let me know if this was incorrect.

jarrpa: Changes addressed following a discussion elsewhere.

services/provider/server/server.go
```diff
@@ -393,6 +393,10 @@ func (s *OCSProviderServer) getCephClientInformation(ctx context.Context, name s
 	return cephClient.Status.Info["secretName"], cephClient.Annotations[controllers.StorageCephUserTypeAnnotation], nil
 }
 
+func scrCephClientCsiSecretName(secretType, suffix string) string {
```
nb-ohad, line 396: Shouldn't this be called `storageClaimCepCSISecretName` mainly because it produces a name for a secret reconciled by the storage claim, not the request?

services/provider/server/server.go
```diff
@@ -553,13 +557,22 @@ func (s *OCSProviderServer) GetStorageClassClaimConfig(ctx context.Context, req
 	var storageClassName string
 	storageClassData := map[string]string{}
 
+	storageClassRequestHash := getStorageClassRequestHash(req.StorageConsumerUUID, req.StorageClassClaimName)
+	nodeSecretName := scrCephClientCsiSecretName("node", storageClassRequestHash)
+	provisionerSecretName := scrCephClientCsiSecretName("provisioner", storageClassRequestHash)
+
+	storageClassData["csi.storage.k8s.io/provisioner-secret-name"] = provisionerSecretName
+	storageClassData["csi.storage.k8s.io/node-stage-secret-name"] = nodeSecretName
+	storageClassData["csi.storage.k8s.io/controller-expand-secret-name"] = provisionerSecretName
```
nb-ohad, line 566: I was under the impression we are not moving any more code outside the switch case and that we are going to refactor the code inside the switch case into its own function in the next commit. I agreed to keep the  `nodeSecretName` and `provisionerSecretName` based on that. 

This change is the exact opposite of that agreement. What changed?

services/provider/server/storageclassrequest.go
```diff
@@ -32,24 +32,32 @@ func newStorageClassRequestManager(cl client.Client, namespace string) (*storage
...
+func getStorageClassRequestHash(consumerUUID, storageClassClaimName string) string {
 	var s struct {
-		StorageConsumerUUID     string `json:"storageConsumerUUID"`
-		StorageClassRequestName string `json:"storageClassRequestName"`
+		StorageConsumerUUID string `json:"storageConsumerUUID"`
+
+		// This field was named before a small refactor to clarify
+		// what it actually represents, but is being kept as-is so as
+		// to not change the generative logic.
+		StorageClassClaimName string `json:"storageClassRequestName"`
```
nb-ohad, line 44: Shouldn't this be called `StorageClaimName`? we moved away from StorageClassClaim API and replaced it with StorageClaim API

services/provider/server/storageclassrequest.go
```diff
@@ -32,24 +32,32 @@ func newStorageClassRequestManager(cl client.Client, namespace string) (*storage
 	}, nil
 }
 
-// getStorageClassRequestName generates a name for a StorageClassRequest resource.
-func getStorageClassRequestName(consumerUUID, storageClassRequestName string) string {
+// getStorageClassRequestHash generates a hash for a StorageClassRequest based
+// on the MD5 hash of the StorageClassClaim name and storageConsumer UUID.
+func getStorageClassRequestHash(consumerUUID, storageClassClaimName string) string {
```
nb-ohad, line 37: If we are doing this change then you need to change the call sites on `storageClassRequestManager` that refer to the same param as `StorageClassRequestName` which I agree is misleading

nb-ohad (changes requested):

services/provider/server/server.go
jarrpa: Since I'm the one that's doing the follow-on PR, I'm organizing it in a way that makes sense for me to then move it elsewhere.

services/provider/server/storageclassrequest.go
jarrpa: I didn't think it was worth the effort to unify that *everywhere* in the codebase on the provider side. That's another follow-up PR, if anything.
image

services/provider/server/server.go
jarrpa: Sure, done.

services/provider/server/storageclassrequest.go
jarrpa: Fair, done.

services/provider/server/server.go
```diff
@@ -43,11 +43,11 @@ import (
...
 const (
-	TicketAnnotation             = "ocs.openshift.io/provider-onboarding-ticket"
-	ProviderCertsMountPoint      = "/mnt/cert"
-	onboardingTicketKeySecret    = "onboarding-ticket-key"
-	storageClassRequestNameLabel = "ocs.openshift.io/storageclassrequest-name"
-	notAvailable                 = "N/A"
+	TicketAnnotation           = "ocs.openshift.io/provider-onboarding-ticket"
+	ProviderCertsMountPoint    = "/mnt/cert"
+	onboardingTicketKeySecret  = "onboarding-ticket-key"
+	storageClassClaimNameLabel = "ocs.openshift.io/storageclassclaim-name"
```
Madhu-1, line 49:
```suggestion
storageClaimNameLabel = "ocs.openshift.io/storageclaim-name"
```
it should be storageClaimNameLabel?

services/provider/server/storageclassrequest.go
Madhu-1: change `storageClassClaimName` to ` storageClaimName` ?

services/provider/server/storageclassrequest.go
Madhu-1: As the final results are changes we don't need to keep `storageClassRequestName` in the json we can just change it to `storageclaimName`. WDYT?

Madhu-1 (commented):
LGTM, can we switch from storageclassclaim to storageclaim as we are changing things respect to Marhsal/Unmarshal to generate names as we also need to handle migration in this PR or are we going to handle that in followup PR?

leelavg:
> LGTM, can we switch from storageclassclaim to storageclaim as we are changing things respect to Marhsal/Unmarshal to generate names as we also need to handle migration in this PR or are we going to handle that in followup PR?
- I believe there'll be two followup PRs (mostly I'll be the one working on it), one that changes gRPC which renames StorageClassClaims to StorageClaims and another that changes StorageClassRequests to StorageRequests.

jarrpa:
@Madhu-1: To all your `StrorageClassClaim` to `StorageClaim` comments: no. That should be handled by a larger PR that comes and does that across the entire code base.

@leelavg Neither! The follow-up PR will be to refactor the logic in the `FulfillStorageClassClaim` function and has nothing to do with the renaming.

services/provider/server/server.go
nb-ohad: ok

services/provider/server/server.go
nb-ohad: I think this is more dangerous (and might have side effects) and regardless should not be part of this PR

nb-ohad:
> @nb-ohad @Madhu-1 wrt reviews could we pls consider this PR as bug fix only, unless I'm missing something which needs more changes ie, update secret names correctly and ntg more?

I agree with a small caveat. If you are renaming existing names or adding new vars/fields with new names you should use the correct terminology with this new stuff. This does not mean you need to go and rename the entirety of the existing codebase

leelavg:
to author & reviewers, may I know the status quo of this PR? if it makes sense, can we resist the renames not to ride on a bug fix?

yes, we renamed storageclassclaim api to storageclaim api in client side and storageclassrequest gets the fields indirectly from client via gRPC and gRPC methods aren't updated yet, can we not do some part here and some part in another PR?

I understand the refactor of func to get hash and can we still go w/ storageclassclaim and during gRPC update I'll do all the changes in one go?

leelavg:
Superseded by #2571, thanks for working on this!

PR #2538, opened 2024-03-29
make provider mode deployment w/ CI and manual testing

leelavg:
update ocs-client-operator bundle for 4.16 and make ocs-client-operator install position independent if deployed along w/ ocs-operator by explicitly informing ocs-client-op not to deploy csi.

add an example for deploying provider mode in single node cluster.

### DIFF
--- a/deploy/deploy-with-olm.yaml
+++ b/deploy/deploy-with-olm.yaml
@@ -7,6 +7,14 @@ metadata:
   name: openshift-storage
 spec: {}
 ---
+apiVersion: v1
+data:
+  DEPLOY_CSI: "false"
+kind: ConfigMap
+metadata:
+  name: ocs-client-operator-config
+  namespace: openshift-storage
+---
 apiVersion: operators.coreos.com/v1
 kind: OperatorGroup
 metadata:
--- a/hack/install-ocs-client.sh
+++ b/hack/install-ocs-client.sh
@@ -14,6 +14,18 @@ else
     oc create ns "$INSTALL_NAMESPACE"
 fi
 
+# Ensure position independent make targets in release CI, explicitly setting the values ensures client-op doesn't deploy CSI
+# when storagecluster is configured for remoteconsumers, controllers set this value to "true"
+cat <<EOF | oc create -f -
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: ocs-client-operator-config
+  namespace: openshift-storage
+data:
+  DEPLOY_CSI: "false"
+EOF
+
 "$OPERATOR_SDK" run bundle "$OCS_CLIENT_BUNDLE_FULL_IMAGE_NAME" --timeout=10m --security-context-config restricted -n "$INSTALL_NAMESPACE" --index-image "$CSI_ADDONS_CATALOG_FULL_IMAGE_NAME"
 
 oc wait --timeout=5m --for condition=Available -n "$INSTALL_NAMESPACE" deployment ocs-client-operator-controller-manager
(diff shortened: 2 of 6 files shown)


leelavg:
@rchikatw & @rewantsoni ptal. thanks.

hack/install-ocs-client.sh
```diff
@@ -16,5 +16,17 @@ fi
...
+# Ensure position independent make targets in release CI, explicitly setting the values ensures client-op doesn't deploy CSI
+# when storagecluster is configured for remoteconsumers, controllers set this value to "true"
+cat <<EOF | oc create -f -
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: ocs-client-operator-config
+  namespace: openshift-storage
+data:
+  DEPLOY_CSI: "false"
```
iamniting, line 28: @leelavg Who is responsible for creating this configmap in the actual cluster?

If it is going to be ocs-operator what if ocs-operator comes up after ocs-client-operator?

deploy/deploy-with-olm.yaml
```diff
@@ -41,6 +41,17 @@ spec:
   source: ocs-catalogsource
   sourceNamespace: openshift-marketplace
 ---
+# Setting this value in the configmap ensures CSI isn't deployed by client
```
iamniting, line 44: Can you pls add this to the go code? This is a generated file.

iamniting (changes requested):

hack/install-ocs-client.sh
leelavg: this is created as part of ocs-client-op bundle.

deploy/deploy-with-olm.yaml
leelavg: ya, I'm currently doing it, was wondering why there was no cm of ocs for rook is created 🤔, maybe no once is using deploy-with-olm & deploy-manager anymore?

iamniting: This is only for the tests on the local cluster, Ci uses install scripts.

hack/install-ocs-client.sh
iamniting: So who will change its value when only ocs-client-operator will be deployed?

leelavg: only ocs-client-operator will not be deployed when run in CI for ocs-operator. When ocs-client-operator is alone deployed it looks as StorageCluster CRD to deploy CSI or not, however during CI we can't guarantee which comes first and so this is only an insurance.

deploy/deploy-with-olm.yaml
leelavg: so deploy-manager is used to generate deploy-with-olm.yaml, also deploy ocs along w/ dependents (where does make install-* scripts are used), also part of functests.

iamniting: I did not get it, Can you pls rephrase it?

leelavg: I was trying to understand what tasks does deploy-manager carry out.

leelavg:
/hold

CI failed for 2nd time, will investigate the failure.

leelavg:
/test ocs-operator-bundle-e2e-aws

didn't observe anything that could cause the failure.

leelavg:
/unhold

I see CI passed in the cluster
```
Ran 4 of 5 Specs in 413.919 seconds
SUCCESS! -- 4 Passed | 0 Failed | 1 Pending | 0 Skipped
PASS
```

iamniting (approved):

PR #2539, opened 2024-04-01
controllers: do not cache all the CRDs in the cluster

leelavg:
first argument to `Watches` is only used to refer the `Kind` of the resource to watch, existing mechanism will list & cache all the resources and send reconciles for all events for all CRDs.

this commit solves above by indicating `manager` to only listen for `virtualmachines` CRD if and when it exists and also create metadata cache for that specific CRD as we are interested only to get a reconcile event.


leelavg:
updated commit msg to pass commitlint.

leelavg:
Closing due to no reviews.

PR #2543, opened 2024-04-04
remove ceph resorces forcefully when client is deleted

rchikatw:
Epic: https://issues.redhat.com/browse/RHSTOR-5698
- In the case of HCP, a cluster can be deleted on the fly, and PVs can remain on the provider's side, so we need to remove them forcefully.
- What API do internally
  1. This API takes the name of the storage consumer as input, adds an annotation to the storage consumer resource, and deletes the storage consumer.
  2. Storage Consumer is responsible for reconciling based on the added annotation and it then adds annotation back to the storage request.
  3. Storage Request is responsible for reconciling based on the added annotation and adding an annotation to Ceph resources. That forcefully cleans the resources.

### DIFF
--- a/controllers/storageclassrequest/storageclassrequest_controller.go
+++ b/controllers/storageclassrequest/storageclassrequest_controller.go
@@ -45,6 +45,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
+const (
+	RookCephResourceForceDeleteAnnotation = "rook.io/force-deletion"
+)
+
 // StorageClassRequestReconciler reconciles a StorageClassRequest object
 // nolint:revive
 type StorageClassRequestReconciler struct {
@@ -79,7 +83,6 @@ func (r *StorageClassRequestReconciler) Reconcile(ctx context.Context, request r
 	if ok := r.Cache.WaitForCacheSync(ctx); !ok {
 		return reconcile.Result{}, fmt.Errorf("cache sync failed")
 	}
-
 	r.log = ctrllog.FromContext(ctx, "StorageClassRequest", request)
 	r.ctx = ctrllog.IntoContext(ctx, r.log)
 	r.log.Info("Reconciling StorageClassRequest.")
@@ -118,8 +121,10 @@ func (r *StorageClassRequestReconciler) Reconcile(ctx context.Context, request r
 
 	result, reconcileError = r.reconcilePhases()
 
-	// Apply status changes to the StorageClassRequest
-	statusError := r.Client.Status().Update(r.ctx, r.StorageClassRequest)
+	controllerutil.AddFinalizer(r.StorageClassRequest, v1alpha1.StorageClassRequestFinalizer)
+
+	// Apply status changes and finalizer to the StorageClassRequest
+	statusError := r.Client.Update(r.ctx, r.StorageClassRequest)
 	if statusError != nil {
 		r.log.Info("Failed to update StorageClassRequest status.")
 	}
@@ -174,7 +179,7 @@ func (r *StorageClassRequestReconciler) SetupWithManager(mgr ctrl.Manager) error
 
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&v1alpha1.StorageClassRequest{}, builder.WithPredicates(
-			predicate.GenerationChangedPredicate{},
+			predicate.Or(predicate.GenerationChangedPredicate{}, predicate.AnnotationChangedPredicate{}),
 		)).
 		Owns(&rookCephv1.CephBlockPoolRadosNamespace{}).
 		Watches(&rookCephv1.CephFilesystemSubVolumeGroup{}, enqueueForOwner).
@@ -333,6 +338,8 @@ func (r *StorageClassRequestReconciler) reconcilePhases() (reconcile.Result, err
 		}
 
 	} else {
+		controllerutil.RemoveFinalizer(r.StorageClassRequest, v1alpha1.StorageClassRequestFinalizer)
+		r.log.Info("finalizer removed successfully")
 		r.StorageClassRequest.Status.Phase = v1alpha1.StorageClassRequestDeleting
 	}
 	return reconcile.Result{}, nil
@@ -351,6 +358,12 @@ func (r *StorageClassRequestReconciler) reconcileRadosNamespace() error {
 		blockPoolName := r.StorageClassRequest.Spec.StorageProfile
 		if blockPoolName == "" {
 			blockPoolName = fmt.Sprintf("%s-cephblockpool", r.storageCluster.Name)
+		}
+		if r.StorageClassRequest.Annotations[controllers.StorageRequestForceDeleteAnnotation] == "true" {
+			if r.cephRadosNamespace.Annotations == nil {
+				r.cephRadosNamespace.Annotations = make(map[string]string)
+			}
+			r.cephRadosNamespace.Annotations[RookCephResourceForceDeleteAnnotation] = "true"
 		}
 		r.cephRadosNamespace.Spec = rookCephv1.CephBlockPoolRadosNamespaceSpec{
 			BlockPoolName: blockPoolName,
@@ -428,6 +441,13 @@ func (r *StorageClassRequestReconciler) reconcileCephFilesystemSubVolumeGroup() 
 
 		r.cephFilesystemSubVolumeGroup.Spec = rookCephv1.CephFilesystemSubVolumeGroupSpec{
 			FilesystemName: cephFilesystem.Name,
+		}
+
+		if r.StorageClassRequest.Annotations[controllers.StorageRequestForceDeleteAnnotation] == "true" {
+			if r.cephFilesystemSubVolumeGroup.Annotations == nil {
+				r.cephFilesystemSubVolumeGroup.Annotations = make(map[string]string)
+			}
+			r.cephFilesystemSubVolumeGroup.Annotations[RookCephResourceForceDeleteAnnotation] = "true"
 		}
 		return nil
 	})
--- a/services/ux-backend/handlers/cleanupconsumer/handler.go
+++ b/services/ux-backend/handlers/cleanupconsumer/handler.go
@@ -1,0 +1,83 @@
+package cleanupconsumer
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	ocsv1alpha1 "github.com/red-hat-storage/ocs-operator/api/v4/v1alpha1"
+	"github.com/red-hat-storage/ocs-operator/v4/services/ux-backend/handlers"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func HandleMessage(w http.ResponseWriter, r *http.Request, cl client.Client) {
+	switch r.Method {
+	case "DELETE":
+		consumerName := r.URL.Query().Get("consumerName")
+		handleDelete(r.Context(), w, cl, consumerName)
+	default:
+		handleUnsupportedMethod(w, r)
+	}
+}
+
+func handleDelete(ctx context.Context, w http.ResponseWriter, cl client.Client, consumerName string) {
+	klog.Info("Delete method on /cleanup-consumer endpoint is invoked")
+	w.Header().Set("Content-Type", handlers.ContentTypeTextPlain)
+	if message, err := removeStorageConsumer(ctx, cl, consumerName); err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		w.Header().Set("Content-Type", handlers.ContentTypeTextPlain)
+
+		if _, err := w.Write([]byte("Failed to delete resources")); err != nil {
+			klog.Errorf("failed write data to response writer, %v", err)
+		}
+	} else {
+		w.WriteHeader(http.StatusOK)
+		w.Header().Set("Content-Type", handlers.ContentTypeTextPlain)
+
+		if _, err = w.Write([]byte(message)); err != nil {
+			klog.Errorf("failed write data to response writer: %v", err)
+		}
+	}
+
+}
+
+func handleUnsupportedMethod(w http.ResponseWriter, r *http.Request) {
+	klog.Info("Only Delete method should be used to send data to this endpoint /cleanup-consumer")
+	w.WriteHeader(http.StatusMethodNotAllowed)
+	w.Header().Set("Content-Type", handlers.ContentTypeTextPlain)
+	w.Header().Set("Allow", "DELETE")
+
+	if _, err := w.Write([]byte(fmt.Sprintf("Unsupported method : %s", r.Method))); err != nil {
+		klog.Errorf("failed to write data to response writer: %v", err)
+	}
+}
+
+func removeStorageConsumer(ctx context.Context, cl client.Client, consumerName string) (string, error) {
+	storageconsumer := &ocsv1alpha1.StorageConsumer{}
+	storageconsumer.Name = consumerName
+	storageconsumer.Namespace = handlers.GetPodNamespace()
+	var err error
+	err = cl.Get(ctx, client.ObjectKeyFromObject(storageconsumer), storageconsumer)
+	if err != nil {
+		return "", fmt.Errorf("failed to get storage consumer with name: %v", err)
+	}
+
+	if storageconsumer.Annotations == nil {
+		storageconsumer.Annotations = make(map[string]string)
+	}
+	storageconsumer.Annotations[handlers.RookCephResourceForceDeleteAnnotation] = "true"
+	// Clean up of resource is part of respective controller
+	err = cl.Update(ctx, storageconsumer)
+	if err != nil {
+		return "", fmt.Errorf("failed to update storage consumer: %v", err)
+	}
+
+	// Trigger delete for cleaning up resources
+	err = cl.Delete(ctx, storageconsumer)
+	if err != nil {
+		return "", fmt.Errorf("failed to delete storage consumer: %v", err)
+	}
+
+	return "Successfully deleted the storage consumer", nil
+}
--- a/services/ux-backend/handlers/common.go
+++ b/services/ux-backend/handlers/common.go
@@ -5,7 +5,8 @@ import (
 )
 
 const (
-	ContentTypeTextPlain = "text/plain"
+	ContentTypeTextPlain                  = "text/plain"
+	RookCephResourceForceDeleteAnnotation = "ocs.openshift.io.storageconsumer/force-deletion"
 )
 
 var namespace string
--- a/services/ux-backend/main.go
+++ b/services/ux-backend/main.go
@@ -7,8 +7,11 @@ import (
 	"os"
 	"strconv"
 
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/klog/v2"
 
+	ocsv1alpha1 "github.com/red-hat-storage/ocs-operator/api/v4/v1alpha1"
+	"github.com/red-hat-storage/ocs-operator/v4/services/ux-backend/handlers/cleanupconsumer"
 	"github.com/red-hat-storage/ocs-operator/v4/services/ux-backend/handlers/onboardingtokens"
 	"github.com/red-hat-storage/ocs-operator/v4/services/ux-backend/handlers/rotatekeys"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -55,12 +58,16 @@ func loadAndValidateServerConfig() (*serverConfig, error) {
 }
 
 func newKubeClient() (client.Client, error) {
+	scheme := runtime.NewScheme()
+	if err := ocsv1alpha1.AddToScheme(scheme); err != nil {
+		return nil, err
+	}
 	cfg, err := config.GetConfig()
 	if err != nil {
 		return nil, err
 	}
 
-	newClient, err := client.New(cfg, client.Options{})
+	newClient, err := client.New(cfg, client.Options{Scheme: scheme})
 	if err != nil {
 		return nil, err
 	}
@@ -90,6 +97,9 @@ func main() {
 	})
 	http.HandleFunc("/rotate-keys", func(w http.ResponseWriter, r *http.Request) {
 		rotatekeys.HandleMessage(w, r, cl)
+	})
+	http.HandleFunc("/cleanup-consumer", func(w http.ResponseWriter, r *http.Request) {
+		cleanupconsumer.HandleMessage(w, r, cl)
 	})
 
 	klog.Info("ux backend server listening on port ", config.listenPort)
--- a/controllers/storageconsumer/storageconsumer_controller.go
+++ b/controllers/storageconsumer/storageconsumer_controller.go
@@ -41,12 +41,14 @@ import (
 )
 
 const (
-	StorageConsumerAnnotation     = "ocs.openshift.io.storageconsumer"
-	StorageRequestAnnotation      = "ocs.openshift.io.storagerequest"
-	StorageCephUserTypeAnnotation = "ocs.openshift.io.cephusertype"
-	StorageProfileLabel           = "ocs.openshift.io/storageprofile"
-	ConsumerUUIDLabel             = "ocs.openshift.io/storageconsumer-uuid"
-	StorageConsumerNameLabel      = "ocs.openshift.io/storageconsumer-name"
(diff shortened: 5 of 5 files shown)


services/ux-backend/handlers/removecephclient/handler.go
```diff
@@ -0,0 +1,125 @@
...
+				subVolumeGroup.Annotations["ceph.rook.io/force-deletion"] = "true"
+
+				err := cl.Update(ctx, subVolumeGroup)
+				if err != nil {
+					return "", fmt.Errorf("failed to annotate CephFilesystemSubVolumeGroup, %v", err)
+				}
+			}
+		}
+	}
+	err = cl.Delete(ctx, storageconsumer)
```
rchikatw, line 119: Since deleting a storage consumer triggers a delete event on storage class request, so we don't need to SCR mannually.

services/ux-backend/handlers/removecephclient/handler.go
```diff
@@ -0,0 +1,125 @@
...
+				radosNamespace := &cephv1.CephBlockPoolRadosNamespace{}
+				err = cl.Get(context.Background(), types.NamespacedName{Name: cephResourceSpec.Name,
+					Namespace: scr.Namespace}, radosNamespace)
+				if err != nil {
+					return "", fmt.Errorf("failed to get CephBlockPoolRadosNamespace, %v", err)
+				}
+				if radosNamespace.Annotations == nil {
+					radosNamespace.Annotations = make(map[string]string)
+				}
+				radosNamespace.Annotations["ceph.rook.io/force-deletion"] = "true"
```
rchikatw, line 94: I have confirmed with Santosh Pillai and included this annotation.

services/ux-backend/handlers/removecephclient/handler.go
```diff
@@ -0,0 +1,125 @@
...
+	switch r.Method {
+	case "POST":
+		handlePost(r.Context(), w, cl, consumerName)
+	default:
+		handleUnsupportedMethod(w, r)
+	}
+}
+
+func handlePost(ctx context.Context, w http.ResponseWriter, cl client.Client, consumerName string) {
+	klog.Info("POST method on /delete-ceph-client endpoint is invoked")
```
rewantsoni, line 26: Is it remove-ceph-client or delete-ceph-client? I saw that in main we used remove-ceph-client, let's stick to one name

services/ux-backend/handlers/removecephclient/handler.go
```diff
@@ -0,0 +1,125 @@
...
+	if _, err := w.Write([]byte(fmt.Sprintf("Unsupported method : %s", r.Method))); err != nil {
+		klog.Errorf("failed to write data to response writer: %v", err)
+	}
+}
+
+func removeCephZombieResources(ctx context.Context, cl client.Client, consumerName string) (string, error) {
+
+	storageconsumer := &ocsv1alpha1.StorageConsumer{}
+	err := cl.Get(context.Background(), types.NamespacedName{Name: consumerName,
+		Namespace: handlers.GetPodNamespace()}, storageconsumer)
```
rewantsoni, line 63: 1. We can use the ctx passed in the function
2. Maybe we can use utils to construct the type? it just saves us from constructing a struct

```suggestion
	storageconsumer := &ocsv1alpha1.StorageConsumer{}
	storageconsumer.Name = consumerName
	storageconsumer.Namespace = handlers.GetPodNamespace()
	err := cl.Get(context.Background(), client.ObjectKeyFromObject(storageconsumer), storageconsumer)
```

services/ux-backend/handlers/removecephclient/handler.go
```diff
@@ -0,0 +1,125 @@
...
+	if err != nil {
+		return "", fmt.Errorf("failed to get storage consumer with name: %s:%v", storageconsumer.Name, err)
+	}
+
+	storageclassrequests := &ocsv1alpha1.StorageClassRequestList{}
+	listOpts := []client.ListOption{
+		client.InNamespace(handlers.GetPodNamespace()),
+		client.MatchingLabels{"ocs.openshift.io/storageconsumer-uuid": string(storageconsumer.UID)},
```
rewantsoni, line 71: could we define the label in a const file somewhere and use it so that we don't have copies?

services/ux-backend/handlers/removecephclient/handler.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	if len(storageclassrequests.Items) == 0 {
+		return "", fmt.Errorf("no storageclassrequests present")
+	}
+	for _, scr := range storageclassrequests.Items {
+		for _, cephResourceSpec := range scr.Status.CephResources {
+			if cephResourceSpec.Kind == "CephBlockPoolRadosNamespace" {
+				radosNamespace := &cephv1.CephBlockPoolRadosNamespace{}
+				err = cl.Get(context.Background(), types.NamespacedName{Name: cephResourceSpec.Name,
```
rewantsoni, line 86: 1. We can use the ctx passed in the function
2. Maybe we can use utils to construct the type? it just saves us from constructing a struct

rewantsoni (changes requested):
I had a couple of questions:
1. If we are adding an annotation will it trigger a reconcile for the storageClassRequest and remove the annotation from the storageClassRequest controller?
2. If we are adding an annotation to the ceph CRs will it also delete the ceph CR or just the underlying resources in ceph?

services/ux-backend/handlers/removecephclient/handler.go
rchikatw: Yes, remove-ceph-client make sense.

services/ux-backend/main.go
```diff
@@ -91,6 +102,10 @@ func main() {
 	http.HandleFunc("/rotate-keys", func(w http.ResponseWriter, r *http.Request) {
 		rotatekeys.HandleMessage(w, r, cl)
 	})
+	http.HandleFunc("/remove-ceph-client", func(w http.ResponseWriter, r *http.Request) {
+		consumerName := r.URL.Query().Get("consumerName")
```
leelavg, line 106: let the `HandleMessage` extract the consumer name from the path.

services/ux-backend/handlers/removecephclient/handler.go
```diff
@@ -0,0 +1,125 @@
...
+
+	if _, err := w.Write([]byte(fmt.Sprintf("Unsupported method : %s", r.Method))); err != nil {
+		klog.Errorf("failed to write data to response writer: %v", err)
+	}
+}
+
+func removeCephZombieResources(ctx context.Context, cl client.Client, consumerName string) (string, error) {
+
+	storageconsumer := &ocsv1alpha1.StorageConsumer{}
+	err := cl.Get(context.Background(), types.NamespacedName{Name: consumerName,
```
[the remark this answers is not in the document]
rchikatw: Okay.

services/ux-backend/main.go
```diff
@@ -91,6 +102,10 @@ func main() {
 	http.HandleFunc("/rotate-keys", func(w http.ResponseWriter, r *http.Request) {
 		rotatekeys.HandleMessage(w, r, cl)
 	})
+	http.HandleFunc("/remove-ceph-client", func(w http.ResponseWriter, r *http.Request) {
```
leelavg, line 105: we are cleaning up consumer and to do so we need to remove some resources which are internal and may not be cephclient only, can the ep be `/cleanup-consumer`?

services/ux-backend/handlers/removecephclient/handler.go
rchikatw: Okay.

services/ux-backend/handlers/removecephclient/handler.go
rchikatw: Okay

services/ux-backend/handlers/removecephclient/handler.go
```diff
@@ -0,0 +1,125 @@
...
+	default:
+		handleUnsupportedMethod(w, r)
+	}
+}
+
+func handlePost(ctx context.Context, w http.ResponseWriter, cl client.Client, consumerName string) {
+	klog.Info("POST method on /delete-ceph-client endpoint is invoked")
+	w.Header().Set("Content-Type", handlers.ContentTypeTextPlain)
+
+	if message, err := removeCephZombieResources(ctx, cl, consumerName); err != nil {
```
leelavg, line 29: there's no need to have the naming `Zombie` in code? this is usually attributed to process where child is waiting to be terminated, in our context we are more like cleaning up orphans assuming the storageclient which created consumer isn't reachable anymore (or probably doesn't exist).

we can just do `removeStorageConsumer`.

services/ux-backend/handlers/removecephclient/handler.go
```diff
@@ -0,0 +1,125 @@
...
+		w.WriteHeader(http.StatusInternalServerError)
+		w.Header().Set("Content-Type", handlers.ContentTypeTextPlain)
+
+		if _, err := w.Write([]byte("Failed to delete resources")); err != nil {
+			klog.Errorf("failed write data to response writer, %v", err)
+		}
+	} else {
+		klog.Info(message)
```
leelavg, line 37: there's no need to just log what func returned as the intention of the method is consumer removal you can log that when the func returned w/o errors.

services/ux-backend/handlers/removecephclient/handler.go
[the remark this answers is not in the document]
rchikatw: Yes, I need to verify if the delete event has been triggered. I am not sure how to do that. Will check and add the code just before deleting the storage consumer.

rchikatw:
> I had a couple of questions:
> 
> 1. If we are adding an annotation will it trigger a reconcile for the storageClassRequest and remove the annotation from the storageClassRequest controller?
> 2. If we are adding an annotation to the ceph CRs will it also delete the ceph CR or just the underlying resources in ceph?

1. Adding annotation won't trigger deletion of storage class request. Once the annotation is added  then I am calling delete on the storage consumer. Storage consumer is the owner of storageclassrequest it will take care of deleting that.
2. Good question I am not sure on what all resources it will delete let me check with Santosh Pillai and get back to you

services/ux-backend/handlers/removecephclient/handler.go
```diff
@@ -0,0 +1,128 @@
...
+	storageclassrequests := &ocsv1alpha1.StorageClassRequestList{}
+	listOpts := []client.ListOption{
+		client.InNamespace(handlers.GetPodNamespace()),
+		client.MatchingLabels{controllers.ConsumerUUIDLabel: string(storageconsumer.UID)},
+	}
+	err = cl.List(ctx, storageclassrequests, listOpts...)
+
+	if err != nil {
+		return "", fmt.Errorf("failed to get list of storageclassrequests, %v", err)
+	}
```
rewantsoni, line 78:
```suggestion
	if err = cl.List(ctx, storageclassrequests, listOpts...); err != nil {
		return "", fmt.Errorf("failed to get list of storageclassrequests, %v", err)
	}
```

services/ux-backend/handlers/removecephclient/handler.go
```diff
@@ -0,0 +1,128 @@
...
+	w.WriteHeader(http.StatusMethodNotAllowed)
+	w.Header().Set("Content-Type", handlers.ContentTypeTextPlain)
+	w.Header().Set("Allow", "POST")
+
+	if _, err := w.Write([]byte(fmt.Sprintf("Unsupported method : %s", r.Method))); err != nil {
+		klog.Errorf("failed to write data to response writer: %v", err)
+	}
+}
+
+func removeCephZombieResources(ctx context.Context, cl client.Client, consumerName string) (string, error) {
```
leelavg, line 59: should a http request perform all these operations or let the storageconsumer know it's time for force removal? My reasoning is:

if everything is successful the entry will be removed from UI, however this can't be guaranteed just by placing annotations on ceph resources, the backend controllers should work on cleanup those resources.

suppose if we didn't annotate svg but did for rns still we failed the cleanup but did partially, so why not reduce the work for the backend by just placing an annotation on storageconsumer and modify storageconsumer controller to perform the cleanup?

moreover the client here isn't backed by a cache and every http call will fetch these resources from api server. wdyt?

services/ux-backend/handlers/removecephclient/handler.go
```diff
@@ -0,0 +1,125 @@
+package removecephclient
```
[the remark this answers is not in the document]
rchikatw: Yes make sense

services/ux-backend/handlers/removecephclient/handler.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/red-hat-storage/ocs-operator/v4/services/ux-backend/handlers"
+	cephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func HandleMessage(w http.ResponseWriter, r *http.Request, cl client.Client, consumerName string) {
+	switch r.Method {
+	case "POST":
```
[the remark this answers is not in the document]
rchikatw: Correct, I will do the change

services/ux-backend/main.go
rchikatw: Yes instead of using remove, cleanup sounds better

services/ux-backend/handlers/removecephclient/handler.go
rchikatw: Yes idea looks better than the one I implemented, Do you think we should add the same annotation that we are adding on SVG/RNS or should we introduce a new one?

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -118,8 +121,10 @@ func (r *StorageClassRequestReconciler) Reconcile(ctx context.Context, request r
 
 	result, reconcileError = r.reconcilePhases()
 
-	// Apply status changes to the StorageClassRequest
-	statusError := r.Client.Status().Update(r.ctx, r.StorageClassRequest)
+	controllerutil.AddFinalizer(r.StorageClassRequest, v1alpha1.StorageClassRequestFinalizer)
+
+	// Apply status changes and finalizer to the StorageClassRequest
+	statusError := r.Client.Update(r.ctx, r.StorageClassRequest)
```
nb-ohad, line 127: You should not update that status using `r.Client.Update`. Please revert the change.

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -118,8 +121,10 @@ func (r *StorageClassRequestReconciler) Reconcile(ctx context.Context, request r
 
 	result, reconcileError = r.reconcilePhases()
 
-	// Apply status changes to the StorageClassRequest
-	statusError := r.Client.Status().Update(r.ctx, r.StorageClassRequest)
+	controllerutil.AddFinalizer(r.StorageClassRequest, v1alpha1.StorageClassRequestFinalizer)
```
nb-ohad, line 124: This should be one of the first steps in the reconcile process - inside reconcilePhases

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -333,6 +338,8 @@ func (r *StorageClassRequestReconciler) reconcilePhases() (reconcile.Result, err
 		}
 
 	} else {
+		controllerutil.RemoveFinalizer(r.StorageClassRequest, v1alpha1.StorageClassRequestFinalizer)
+		r.log.Info("finalizer removed successfully")
```
nb-ohad, line 342: 1. I don't understand why are you removing the finalizer immediately. if there is no sync condition that we need to wait for then we should not have added a finalizer in the first place.   

2. The pattern needs to be
```
if controllerutil.RemoveFinalizer(r.StorageClassRequest, v1alpha1.StorageClassRequestFinalizer) {
       r.Client.Update(....)
}

File: controllers/storageclassrequest/storageclassrequest_controller.go [hunk 19]
```diff
@@ -352,6 +359,12 @@ func (r *StorageClassRequestReconciler) reconcileRadosNamespace() error {
 		if blockPoolName == "" {
 			blockPoolName = fmt.Sprintf("%s-cephblockpool", r.storageCluster.Name)
 		}
+		if r.StorageClassRequest.Annotations[controllers.StorageRequestForceDeleteAnnotation] == "true" {
+			if r.cephRadosNamespace.Annotations == nil {
+				r.cephRadosNamespace.Annotations = make(map[string]string)
+			}
+			r.cephRadosNamespace.Annotations[RookCephResourceForceDeleteAnnotation] = "true"
+		}
```
nb-ohad, line 367: There is an add annotation util that does all of this. No need to write this logic yourself.

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -429,6 +442,13 @@ func (r *StorageClassRequestReconciler) reconcileCephFilesystemSubVolumeGroup()
 		r.cephFilesystemSubVolumeGroup.Spec = rookCephv1.CephFilesystemSubVolumeGroupSpec{
 			FilesystemName: cephFilesystem.Name,
 		}
+
+		if r.StorageClassRequest.Annotations[controllers.StorageRequestForceDeleteAnnotation] == "true" {
+			if r.cephFilesystemSubVolumeGroup.Annotations == nil {
+				r.cephFilesystemSubVolumeGroup.Annotations = make(map[string]string)
+			}
+			r.cephFilesystemSubVolumeGroup.Annotations[RookCephResourceForceDeleteAnnotation] = "true"
+		}
```
nb-ohad, line 451: There is an add annotation util that does all of this. No need to write this logic yourself.

services/ux-backend/handlers/cleanupconsumer/handler.go
nb-ohad, line 1: This is a delete operation, not a cleanup operation.  please call the handle `deleteconsumer`

services/ux-backend/handlers/cleanupconsumer/handler.go
```diff
@@ -0,0 +1,83 @@
...
+	case "DELETE":
+		consumerName := r.URL.Query().Get("consumerName")
+		handleDelete(r.Context(), w, cl, consumerName)
+	default:
+		handleUnsupportedMethod(w, r)
+	}
+}
+
+func handleDelete(ctx context.Context, w http.ResponseWriter, cl client.Client, consumerName string) {
+	klog.Info("Delete method on /cleanup-consumer endpoint is invoked")
```
nb-ohad, line 25: This log does not make sense. what is a `/cleanup-consumer` resource?

services/ux-backend/handlers/cleanupconsumer/handler.go
```diff
@@ -0,0 +1,83 @@
...
+	w.Header().Set("Content-Type", handlers.ContentTypeTextPlain)
+	if message, err := removeStorageConsumer(ctx, cl, consumerName); err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		w.Header().Set("Content-Type", handlers.ContentTypeTextPlain)
+
+		if _, err := w.Write([]byte("Failed to delete resources")); err != nil {
```
nb-ohad, line 31:
```suggestion
		if _, err := w.Write([]byte("Failed to delete storage consumer")); err != nil {
```

services/ux-backend/handlers/cleanupconsumer/handler.go
```diff
@@ -0,0 +1,83 @@
...
+			klog.Errorf("failed write data to response writer, %v", err)
+		}
+	} else {
+		w.WriteHeader(http.StatusOK)
+		w.Header().Set("Content-Type", handlers.ContentTypeTextPlain)
+
+		if _, err = w.Write([]byte(message)); err != nil {
+			klog.Errorf("failed write data to response writer: %v", err)
```
nb-ohad, line 39:
```suggestion
			klog.Errorf("failed to write data to response writer: %v", err)
```

services/ux-backend/handlers/cleanupconsumer/handler.go
```diff
@@ -0,0 +1,83 @@
...
+		}
+	}
+
+}
+
+func handleUnsupportedMethod(w http.ResponseWriter, r *http.Request) {
+	klog.Info("Only Delete method should be used to send data to this endpoint /cleanup-consumer")
```
nb-ohad, line 46: Again, the endpoint name does not make sense

services/ux-backend/handlers/cleanupconsumer/handler.go
```diff
@@ -0,0 +1,83 @@
...
+
+	if storageconsumer.Annotations == nil {
+		storageconsumer.Annotations = make(map[string]string)
+	}
+	storageconsumer.Annotations[handlers.RookCephResourceForceDeleteAnnotation] = "true"
+	// Clean up of resource is part of respective controller
+	err = cl.Update(ctx, storageconsumer)
+	if err != nil {
+		return "", fmt.Errorf("failed to update storage consumer: %v", err)
+	}
```
nb-ohad, line 74: Why are you doing all of that? 
You just need to invoke a delete operation on the StorageConsumer CR after checking that we meet the last heartbeat requirements - a check that is missing here.

services/ux-backend/handlers/common.go
```diff
@@ -5,7 +5,8 @@ import (
 )
 
 const (
-	ContentTypeTextPlain = "text/plain"
+	ContentTypeTextPlain                  = "text/plain"
+	RookCephResourceForceDeleteAnnotation = "ocs.openshift.io.storageconsumer/force-deletion"
```
nb-ohad, line 9: ux-backend should not use this annotation, please remove

services/ux-backend/main.go
```diff
@@ -91,6 +98,9 @@ func main() {
 	http.HandleFunc("/rotate-keys", func(w http.ResponseWriter, r *http.Request) {
 		rotatekeys.HandleMessage(w, r, cl)
 	})
+	http.HandleFunc("/cleanup-consumer", func(w http.ResponseWriter, r *http.Request) {
```
nb-ohad, line 101: There is no resource named `clenup-consumer`. The name for the resource should make sense:
`/storageconsumers/<consumer_id>`

services/ux-backend/handlers/cleanupconsumer/handler.go
```diff
@@ -0,0 +1,83 @@
...
+	ocsv1alpha1 "github.com/red-hat-storage/ocs-operator/api/v4/v1alpha1"
+	"github.com/red-hat-storage/ocs-operator/v4/services/ux-backend/handlers"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func HandleMessage(w http.ResponseWriter, r *http.Request, cl client.Client) {
+	switch r.Method {
+	case "DELETE":
+		consumerName := r.URL.Query().Get("consumerName")
```
nb-ohad, line 17: consumer name should come from the path not from the query. 
The resource we are trying to delete is: `/sotrageconsumers/<consumer_id>`

nb-ohad:
@rchikatw Thinking about this endpoint again. Why do we need it in the first place? Why can't the UI call delete on the storageconsumer resource?

services/ux-backend/handlers/removecephclient/handler.go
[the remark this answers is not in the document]
nb-ohad: The package name should be the endpoint name. As we talk about `storageconsumers` here as the target resource  the endpoint in question should be storageconsumers as the entire point of the operation is to invoke delete on a storageconsumer at path `/storageconsumers/<storage_consumer>`

services/ux-backend/main.go
nb-ohad: I think there is a misconception here. With rest-based APIs the path indicates a resource or a ser of resources.
There is no resource in the system that is named `cleanup-consumer`

leelavg:
> @rchikatw Thinking about this endpoint again. Why do we need it in the first place? Why can't the UI call delete on the storageconsumer resource? The current implementation is basically just a proxy to k8s delete anyways.
- so is the earlier `rotate-keys` endpoint, Bipul also raised the same concern, the only reason that we are still going w/ these endpoints is to support API calls from other operators (aka HCI) which may not use UI.

leelavg:
superseded by #2563

leelavg:
superseded by #2563

PR #2561, opened 2024-04-15
BUG 2267067: rbd metrics set for provider-client cluster

Mrudraia:


ezio-auditore (changes requested):
Please close if no more relevant

PR #2563, opened 2024-04-17
allow cleanup of ceph resources when storageclassrequest is deleted 

leelavg:
rook cleans up underlying ceph resources when an annotation is set during deletion and we set that annotation when storageclassrequest is marked for deletion.

### DIFF
--- a/controllers/storageclassrequest/storageclassrequest_controller.go
+++ b/controllers/storageclassrequest/storageclassrequest_controller.go
@@ -45,6 +45,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
+const (
+	storageClassRequestFinalizer = "ocs.openshift.io.storageclassrequest"
+	forceDeletionAnnotationKey   = "ceph.rook.io/force-deletion"
+)
+
 // StorageClassRequestReconciler reconciles a StorageClassRequest object
 // nolint:revive
 type StorageClassRequestReconciler struct {
@@ -293,6 +298,11 @@ func (r *StorageClassRequestReconciler) reconcilePhases() (reconcile.Result, err
 	r.StorageClassRequest.Status.Phase = v1alpha1.StorageClassRequestCreating
 
 	if r.StorageClassRequest.GetDeletionTimestamp().IsZero() {
+		if controllerutil.AddFinalizer(r.StorageClassRequest, storageClassRequestFinalizer) {
+			if err := r.update(r.StorageClassRequest); err != nil {
+				return reconcile.Result{}, fmt.Errorf("failed to add finalizer: %v", err)
+			}
+		}
 		if r.StorageClassRequest.Spec.Type == "blockpool" {
 
 			if err := r.reconcileCephClientRBDProvisioner(); err != nil {
@@ -334,6 +344,9 @@ func (r *StorageClassRequestReconciler) reconcilePhases() (reconcile.Result, err
 
 	} else {
 		r.StorageClassRequest.Status.Phase = v1alpha1.StorageClassRequestDeleting
+		if err := r.deletionPhase(); err != nil {
+			return reconcile.Result{}, err
+		}
 	}
 	return reconcile.Result{}, nil
 }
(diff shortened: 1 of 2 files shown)


leelavg:
Obsoletes #2543

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -218,11 +233,82 @@ func (r *StorageConsumerReconciler) setCephResourceStatus(name string, kind stri
...
+		if err := r.annotateCephResources(); err != nil {
+			return err
+		}
+	}
+
+	if controllerutil.RemoveFinalizer(r.storageConsumer, storageConsumerFinalizer) {
+		if err := r.update(r.storageConsumer); err != nil {
+			return fmt.Errorf("failed to remove finalizer: %v", err)
+		}
+	}
```
Madhu-1, line 248: should we wait till the rook CR's are deleted before removing the finalizers and deleting the consumer CR?

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -218,11 +233,82 @@ func (r *StorageConsumerReconciler) setCephResourceStatus(name string, kind stri
...
+
+func (r *StorageConsumerReconciler) annotateCephResources() error {
+
+	scrList := &ocsv1alpha1.StorageClassRequestList{}
+	if err := r.list(scrList,
+		client.InNamespace(r.storageConsumer.Namespace),
+		client.MatchingFields{ConsumerUUIDLabel: string(r.storageConsumer.UID)}); err != nil {
+		return fmt.Errorf("failed to list storageclassrequests that are referring storageconsumer: %v", err)
+	}
+	for idx := range scrList.Items {
```
Madhu-1, line 261: instead of doing this can you list CephBlockPoolRadosNamespace and CephFilesystemSubVolumeGroup based on the label? i think we add the consumer name on these CR's as label

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -218,11 +233,82 @@ func (r *StorageConsumerReconciler) setCephResourceStatus(name string, kind stri
...
+		scr := &scrList.Items[idx]
+		for idx := range scr.Status.CephResources {
+			resource := scr.Status.CephResources[idx]
+			switch resource.Kind {
+			case "CephBlockPoolRadosNamespace":
+				rns := &rookCephv1.CephBlockPoolRadosNamespace{}
+				rns.Name = resource.Name
+				rns.Namespace = r.storageConsumer.Namespace
```
Madhu-1, line 269: use r.OperatorNamespace as its used in most places for ceph resources creation.

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -218,11 +233,82 @@ func (r *StorageConsumerReconciler) setCephResourceStatus(name string, kind stri
...
+				if err := r.get(rns); err != nil {
+					return fmt.Errorf("failed to get CephBlockPoolRadosNamespace: %v", err)
```
Madhu-1, line 271: should handle not found error as well?

nb-ohad:
LGTM

controllers/storageconsumer/storageconsumer_controller.go
leelavg: reimplemented a straight forward way, pls re-review. thanks.

controllers/storageconsumer/storageconsumer_controller.go
leelavg: reimplemented a straight forward way, pls re-review. thanks.

controllers/storageconsumer/storageconsumer_controller.go
leelavg: reimplemented a straight forward way, pls re-review. thanks.

controllers/storageconsumer/storageconsumer_controller.go
leelavg: reimplemented a straight forward way, pls re-review. thanks.

leelavg:
> should we wait till the rook CR's are deleted before
@Madhu-1 I believe this needs explanation, I'm not waiting here as the delete event for rook comes after setting this annotation and so it's rooks responsibility to cleanup and I don't think we can anything if deletion is stuck unless we are planning on sending alerts or events.

Madhu-1 (approved):

Madhu-1:
> * I believe this needs explanation, I'm not waiting here as the delete event for rook comes after setting this annotation and so it's rooks responsibility to cleanup and I don't think we can anything if deletion is stuck unless we are planning on sending alerts or events.

my only concern is with corner cases what if we get request to create the same storageclassrequest , if that is covered or if we don't hit it, we are good 👍🏻

leelavg:
> request to create the same storageclassrequest
- ack, that'll not happen, the name is already predefined and k8s guarantees no duplicated resources since writes directly hits k8s server rather than cache.

Madhu-1:
> > request to create the same storageclassrequest
> 
> * ack, that'll not happen, the name is already predefined and k8s guarantees no duplicated resources since writes directly hits k8s server rather than cache.

what i meant was we are not waiting for Rook resource deletion and we are deleting the storageclassrequest, if we get a request to create it again with the same name we will try to create ceph resources which are in the deletion phase and it will fail in the same state until those CR's are deleted. i hope this wont happen and its a corner case as well.

controllers/storageconsumer/storageconsumer_controller.go
leelavg: resolving after confirming offline.

controllers/storageconsumer/storageconsumer_controller.go
leelavg: resolving after confirming offline.

controllers/storageconsumer/storageconsumer_controller.go
leelavg: resolving after confirming offline.

leelavg:
> i hope this wont happen and its a corner case as well.
- makes sense, will await testing results.

leelavg:
rebased, CI should pass now.

leelavg:

Overiding because previous run was success and the rerun trigerred due to merge of a PR which doesn't effect current PR
```
[1780862628051030016](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/red-hat-storage_ocs-operator/2563/pull-ci-red-hat-storage-ocs-operator-main-ocs-operator-bundle-e2e-aws/1780862628051030016)	#[2563](https://github.com/red-hat-storage/ocs-operator/pull/2563)
([cecc3b9](https://github.com/red-hat-storage/ocs-operator/pull/2563/commits/cecc3b9a7ea748efee05334b2ab0743e76f0603c)) by [leelavg](https://github.com/leelavg)	
13:05:16
1h21m41s	SUCCESS
```

controllers/storageclassrequest/storageclassrequest_controller.go
```diff
@@ -45,6 +45,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
+const (
+	storageClassRequestFinalizer = "ocs.openshift.io.storageclassrequest"
+	forceDeletionAnnotationKey   = "ceph.rook.io/force-deletion"
```
rchikatw, line 50: Later, It got changed to **rook.io/force-deletion**

leelavg: thanks for the updated, raised https://github.com/red-hat-storage/ocs-operator/pull/2573

PR #2570, opened 2024-04-19
change storagerequest spec.type to be either block or sharedfile

leelavg:
change the request types from `sharedfilesystem` to `sharedfile` and `blockpool` to `block`, at the other end storageclaims correspond to this naming and when these are listed if names doesn't corresponding it'll not be a good experience.

the new names are at the correct abstraction level wrt the type of storage that is being claimed.

### DIFF
--- a/api/v1alpha1/storagerequest_types.go
+++ b/api/v1alpha1/storagerequest_types.go
@@ -24,7 +24,7 @@ import (
 
 // StorageRequestSpec defines the desired state of StorageRequest
 type StorageRequestSpec struct {
-	//+kubebuilder:validation:Enum=blockpool;sharedfilesystem
+	//+kubebuilder:validation:Enum=block;sharedfile
 	Type             string `json:"type"`
 	EncryptionMethod string `json:"encryptionMethod,omitempty"`
 	StorageProfile   string `json:"storageProfile,omitempty"`
--- a/controllers/storagerequest/storagerequest_controller.go
+++ b/controllers/storagerequest/storagerequest_controller.go
@@ -213,7 +213,7 @@ func (r *StorageRequestReconciler) initPhase() error {
 	}
 
 	// check request status already contains the name of the resource. if not, add it.
-	if r.StorageRequest.Spec.Type == "blockpool" {
+	if r.StorageRequest.Spec.Type == "block" {
 		// initialize in-memory structs
 		r.cephRadosNamespace = &rookCephv1.CephBlockPoolRadosNamespace{}
 		r.cephRadosNamespace.Namespace = r.OperatorNamespace
@@ -243,7 +243,7 @@ func (r *StorageRequestReconciler) initPhase() error {
 		} else {
 			return fmt.Errorf("invalid number of CephBlockPoolRadosNamespaces for storage consumer %q: found %d, expecting 0 or 1", r.storageConsumer.Name, rnsItemsLen)
 		}
-	} else if r.StorageRequest.Spec.Type == "sharedfilesystem" {
+	} else if r.StorageRequest.Spec.Type == "sharedfile" {
 		r.cephFilesystemSubVolumeGroup = &rookCephv1.CephFilesystemSubVolumeGroup{}
 		r.cephFilesystemSubVolumeGroup.Namespace = r.OperatorNamespace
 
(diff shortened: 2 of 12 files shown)


api/v1alpha1/storagerequest_types.go
```diff
@@ -24,7 +24,7 @@ import (
 
 // StorageRequestSpec defines the desired state of StorageRequest
 type StorageRequestSpec struct {
-	//+kubebuilder:validation:Enum=blockpool;sharedfilesystem
+	//+kubebuilder:validation:XValidation:rule="self.lowerAscii()=='block'||self.lowerAscii()=='sharedfile'",message="value should be either 'sharedfile' or 'block' (case insensitive)"
```
nb-ohad, line 27: No need to complicate stuff with the mention of case insensitive in the message.  
```suggestion
	//+kubebuilder:validation:XValidation:rule="self.lowerAscii()=='block'||self.lowerAscii()=='sharedfile'",message="value should be either 'sharedfile' or 'block'"
```

controllers/storagerequest/storagerequest_controller.go
```diff
@@ -213,7 +213,7 @@ func (r *StorageRequestReconciler) initPhase() error {
 	}
 
 	// check request status already contains the name of the resource. if not, add it.
-	if r.StorageRequest.Spec.Type == "blockpool" {
+	if strings.ToLower(r.StorageRequest.Spec.Type) == "block" {
```
nb-ohad, line 216: You are repeating string manipulation logic in multiple places. Instead of codifying the implementation in each code site use a util func to codify the intent, so if the implementation changes we will only need to change it once

```suggestion
	if isBlockRequest(r.StorageRequest.Spec.Type) {
```

controllers/storagerequest/storagerequest_controller.go
```diff
@@ -243,7 +243,7 @@ func (r *StorageRequestReconciler) initPhase() error {
 		} else {
 			return fmt.Errorf("invalid number of CephBlockPoolRadosNamespaces for storage consumer %q: found %d, expecting 0 or 1", r.storageConsumer.Name, rnsItemsLen)
 		}
-	} else if r.StorageRequest.Spec.Type == "sharedfilesystem" {
+	} else if strings.ToLower(r.StorageRequest.Spec.Type) == "sharedfile" {
```
nb-ohad, line 246: For completeness I will add this suggestion to the last one 

```suggestion
	} else if isSharedFileRequest(r.StorageRequest.Spec.Type) {
```

nb-ohad (changes requested):

api/v1alpha1/storagerequest_types.go
leelavg: done.

controllers/storagerequest/storagerequest_controller.go
leelavg: done.

controllers/storagerequest/storagerequest_controller.go
leelavg: done.

nb-ohad (changes requested):

nb-ohad:
@leelavg Looking at the PR again, I don't understand why are we even allowing variance in the StorageRequest CR? This is not the customer-facing CR and it is completely driven by values coming from the API server which uses a fix value gRPC enum. 

Given the above, we should align here on a single letter casing as the actual values - would suggest "block" and "sharedfile" for simplicity

leelavg:
> Given the above, we should align here on a single letter casing as the actual values - would suggest "block" and "sharedfile" for simplicity
- done.

leelavg:
Pls override the CI, it was already passed twice and rebase shouldn't effect the content in this PR
```
[1782242722539638784](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/red-hat-storage_ocs-operator/2570/pull-ci-red-hat-storage-ocs-operator-main-ocs-operator-bundle-e2e-aws/1782242722539638784)	#[2570](https://github.com/red-hat-storage/ocs-operator/pull/2570)
([33eef39](https://github.com/red-hat-storage/ocs-operator/pull/2570/commits/33eef397af676cd8cd9d0395096362478e15d9e4)) by [leelavg](https://github.com/leelavg)	
08:29:17
1h13m28s	SUCCESS
```

PR #2575, opened 2024-04-22
use existing suffix hash from storagerequest name for ceph resources

leelavg:
storagerequest suffix is already a hash of storageconsumer uuid and storageclaim name sent by the client and hashing again the hash value although guarantee uniqueness it's redundant.

other ways considered are, 1) directly using the uuid of storagerequest, however it creates extra work for migration since the migration may happen from ceph resources to k8s resources. 2) generate the hash again in storagerequest controller, however by that time we can only access storageconsumer uuid but not the storageclaim name send by the client.

what this PR adds is consistent names of resources created by storagerequest controller and also aligns w/ current refactor of ceph client secret names.

### DIFF
--- a/controllers/storagerequest/storagerequest_controller.go
+++ b/controllers/storagerequest/storagerequest_controller.go
@@ -15,8 +15,6 @@ package storagerequest
 
 import (
 	"context"
-	"crypto/md5"
-	"encoding/hex"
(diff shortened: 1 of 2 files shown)


leelavg:
need to be re-assessed at a later point.

PR #2579, opened 2024-04-23
Add waitTimeoutForHealthyOSDInMinutes field in the storagecluster CR

Nikhil-Ladha:
Add new waitTimeoutForHealthyOSDInMinutes field under `ManageCephCluster` in the storagecluster CR, to allow users to override defaults in the CephCluster.

JIRA: https://issues.redhat.com/browse/RHSTOR-5756

### DIFF
--- a/api/v1/storagecluster_types.go
+++ b/api/v1/storagecluster_types.go
@@ -18,6 +18,7 @@ package v1
 
 import (
 	"os"
+	"time"
 
 	nbv1 "github.com/noobaa/noobaa-operator/v5/pkg/apis/noobaa/v1alpha1"
 	quotav1 "github.com/openshift/api/quota/v1"
@@ -176,6 +177,14 @@ type ManageCephCluster struct {
 	MgrCount int `json:"mgrCount,omitempty"`
 	// +kubebuilder:validation:Enum=3;5
 	MonCount int `json:"monCount,omitempty"`
+	// WaitTimeoutForHealthyOSDInMinutes defines the time the operator would wait before an OSD can be stopped for upgrade or restart.
+	// If `continueUpgradeAfterChecksEvenIfNotHealthy` is `false` and the timeout exceeds and OSD is not ok to stop, then the operator
+	// would skip upgrade for the current OSD and proceed with the next one.
+	// If `continueUpgradeAfterChecksEvenIfNotHealthy` is `true`, then operator would continue with the upgrade of an OSD even if its
+	// not ok to stop after the timeout.
+	// This timeout won't be applied if `skipUpgradeChecks` is `true`.
+	// The default wait timeout is 10 minutes.
+	WaitTimeoutForHealthyOSDInMinutes time.Duration `json:"waitTimeoutForHealthyOSDInMinutes,omitempty"`
 }
 
 // ManageCephConfig defines how to reconcile the Ceph configuration
--- a/config/crd/bases/ocs.openshift.io_storageclusters.yaml
+++ b/config/crd/bases/ocs.openshift.io_storageclusters.yaml
@@ -737,6 +737,19 @@ spec:
                         type: integer
                       reconcileStrategy:
                         type: string
+                      waitTimeoutForHealthyOSDInMinutes:
+                        description: WaitTimeoutForHealthyOSDInMinutes defines the
+                          time the operator would wait before an OSD can be stopped
+                          for upgrade or restart. If `continueUpgradeAfterChecksEvenIfNotHealthy`
+                          is `false` and the timeout exceeds and OSD is not ok to
+                          stop, then the operator would skip upgrade for the current
+                          OSD and proceed with the next one. If `continueUpgradeAfterChecksEvenIfNotHealthy`
+                          is `true`, then operator would continue with the upgrade
+                          of an OSD even if its not ok to stop after the timeout.
+                          This timeout won't be applied if `skipUpgradeChecks` is
+                          `true`. The default wait timeout is 10 minutes.
+                        format: int64
+                        type: integer
                     type: object
                   cephConfig:
                     description: ManageCephConfig defines how to reconcile the Ceph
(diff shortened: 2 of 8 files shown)


api/v1/storagecluster_types.go
```diff
@@ -176,6 +177,12 @@ type ManageCephCluster struct {
 	MgrCount int `json:"mgrCount,omitempty"`
 	// +kubebuilder:validation:Enum=3;5
 	MonCount int `json:"monCount,omitempty"`
+	// WaitTimeoutForHealthyOSDInMinutes defines the time the operator would wait before an OSD can be stopped for upgrade or restart.
+	// If the timeout exceeds and OSD is not ok to stop, then the operator would skip upgrade for the current OSD and proceed with the next one
+	// if `continueUpgradeAfterChecksEvenIfNotHealthy` is `false`. If `continueUpgradeAfterChecksEvenIfNotHealthy` is `true`, then operator would
+	// continue with the upgrade of an OSD even if its not ok to stop after the timeout. This timeout won't be applied if `skipUpgradeChecks` is `true`.
+	// The default wait timeout is 10 minutes.
```
iamniting, line 184: It is a bit confusing, Can we pls simplify it?

iamniting (changes requested):

api/v1/storagecluster_types.go
Nikhil-Ladha: Updated it now.

iamniting (approved):

api/v1/storagecluster_types.go
```diff
@@ -176,6 +177,14 @@ type ManageCephCluster struct {
...
 	// +kubebuilder:validation:Enum=3;5
 	MonCount int `json:"monCount,omitempty"`
+	// WaitTimeoutForHealthyOSDInMinutes defines the time the operator would wait before an OSD can be stopped for upgrade or restart.
+	// If `continueUpgradeAfterChecksEvenIfNotHealthy` is `false` and the timeout exceeds and OSD is not ok to stop, then the operator
+	// would skip upgrade for the current OSD and proceed with the next one.
+	// If `continueUpgradeAfterChecksEvenIfNotHealthy` is `true`, then operator would continue with the upgrade of an OSD even if its
+	// not ok to stop after the timeout.
+	// This timeout won't be applied if `skipUpgradeChecks` is `true`.
+	// The default wait timeout is 10 minutes.
+	WaitTimeoutForHealthyOSDInMinutes time.Duration `json:"waitTimeoutForHealthyOSDInMinutes,omitempty"`
```
Madhu-1, line 187: @Nikhil-Ladha this doesnt seems to be correct if we expect the input in Minutes this should be `unit` not time.Duration.  if i pass here `1` its not considered as 1 minutes it is considered as 1 nano second. please let me know if am wrong

Madhu-1: cc @travisn

Nikhil-Ladha: The reason I set it as time.Duration was because rook uses time.Duration as the type for this API.
Also, we multiply by `time.Minute` before setting the default value in the CR.

Madhu-1: we are doing it only for default not for the user value, we should have done it for both

Nikhil-Ladha: Hmm, yeah we can update this.
Though, I am not sure if we should be updating it for previous backports as well? We can mention in the KCS for now and update it for 4.17 only maybe?
cc @agarwal-mudit @travisn

travisn: @Madhu-1 @Nikhil-Ladha This is really a Rook bug, it should have been defined as an `int` instead of a `time.Duration`, to avoid this confusion. But now we are stuck with the type. 

If the setting is applied, Rook will internally multiply by `time.Minute` [here](https://github.com/rook/rook/blob/48b657099d3c5b55b1ffcc76475b39abf5c5ce77/pkg/operator/ceph/cluster/osd/osd.go#L213). So anytime this is applied on the CephCluster CR it will be treated as minutes. If the user sets this in the StorageCluster CR it looks like it will behave as expected.

The problem comes when OCS operator sets the default of `10 * time.minute` on the CephCluster CR. Rook will again multiply by `time.Minute`, causing the default value to be incorrect. We need to get this fixed. Is this already in a release?

Madhu-1: Thanks @travisn , @agarwal-mudit @Nikhil-Ladha ^^ we need to get it fixed i assume this is not in any release yet as the backport PRs are still under testing phase @petr-balogh can confirm the same.

travisn: I just commented on https://bugzilla.redhat.com/show_bug.cgi?id=2276694, let's move the discussion to the BZ to get a fix, thanks!

petr-balogh: @travisn `4.15.3-5` where this BZ is targeted is under testing, so it still can be fixed and respin the build.
4.16 is not GAed so I guess also no issue there.

The weird is that when we during CLI Deployment we provide storageCluster CR with this value , that it's not reflected, any idea why?

travisn: > The weird is that when we during CLI Deployment we provide storageCluster CR with this value , that it's not reflected, any idea why?

@Nikhil-Ladha Could you look at this?

travisn: For the fix of the main issue, I'd suggest OCS operator shouldn't set a default, and would only set the value in the CephCluster CR if a value is set in the StorageCluster CR.

travisn: > For the fix of the main issue, I'd suggest OCS operator shouldn't set a default, and would only set the value in the CephCluster CR if a value is set in the StorageCluster CR.

And I see it's already done in #2639, thanks @Nikhil-Ladha! :)

Nikhil-Ladha: Already sent a PR for this #2639 .

Nikhil-Ladha: We have removed the default from the main branch for 4.17, but it is not done for previous releases.
So, just creating backport PRs.

PR #2584, opened 2024-04-24
Bug 2213757: external: separate rgw endpoint validation from cephfs and rbd storage…

parth-gr:
…class

currently if the rgw endpoint is not reachable
all the storageclass creation fails
Fix will make the rgw endpoint validation separate and will create rbd and cephfs storageclasses if rgw endpoints validation fails


iamniting (changes requested):
Can we pls have the PR in the main branch first, otherwise it will be missing in the 4.17 and so on.

PR #2588, opened 2024-04-29
Adding empty label selector for prom service monitor selector

ezio-auditore:
Introduced in #2465 the service monitor selector in the Prometheus configuration was marked wrongly,hence it was scraping metrics from only k8s-service-monitor. Adding an empty service monitor selector scrapes metrics from all Service monitors  present in the namespace.
List of Service Monitors present in the namespace
```
k8s-metrics-service-monitor                      
noobaa-mgmt-service-monitor                    
ocs-metrics-exporter                            
odf-operator-controller-manager-metrics-monitor  
rook-ceph-exporter                              
rook-ceph-mgr                                    
s3-service-monitor                                
```

### DIFF
--- a/templates/prometheus.go
+++ b/templates/prometheus.go
@@ -10,12 +10,6 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
-var serviceMonitorSelector = metav1.LabelSelector{
-	MatchLabels: map[string]string{
-		"app": "odf-prometheus",
-	},
-}
-
 var ruleSelector = metav1.LabelSelector{
 	MatchLabels: map[string]string{
 		"prometheus": "rook-prometheus",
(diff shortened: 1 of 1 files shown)


templates/prometheus.go
```diff
@@ -31,7 +25,7 @@ var (
 var PrometheusSpecTemplate = promv1.PrometheusSpec{
 	CommonPrometheusFields: promv1.CommonPrometheusFields{
 		ServiceAccountName:     "prometheus-k8s",
-		ServiceMonitorSelector: &serviceMonitorSelector,
+		ServiceMonitorSelector: &metav1.LabelSelector{},
```
umangachapagain, line 28: Why select all the ServiceMonitors? We already have labels that is used by OCP monitoring. We could re-use the same here.

ezio-auditore: @umangachapagain can you be more specific. Which labels? I can see all the service monitors have different labels

umangachapagain: ok, I re-checked. OCP is using NamespaceSelector with empty ServiceMonitorSelector. 
This implementation is fine.

umangachapagain (approved):

PR #2592, opened 2024-05-01
hack: define proper 'GOARCH' when building ARM64 images

synarete:
Use local host CPU architecture and pass it as GOARCH to image-build command. Needed when building images on non x86_64 images (e.g., AWS/Graviton ARM64, [RHSTOR-2044]).

### DIFF
--- a/hack/build-metrics-exporter.sh
+++ b/hack/build-metrics-exporter.sh
@@ -7,5 +7,5 @@ source hack/docker-common.sh
 
 pushd metrics
 ${IMAGE_BUILD_CMD} build --no-cache -f Dockerfile -t "${METRICS_EXPORTER_FULL_IMAGE_NAME}" . \
-    --build-arg="GOOS=${GOOS}" --build-arg="GOARCH=${GOARCH}" --build-arg="LDFLAGS=${LDFLAGS}"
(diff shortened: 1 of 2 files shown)


iamniting (changes requested):
As far as I remember we did this intentionally to build the images for multiple archs from a single place.

@malayparida2000 can correct me if I am wrong here.

malayparida2000:
Yes, this was done intentionally as ther's no assurity that the target arch is always the host arch. Whatever is the target arch can be defined by setting the GOARCH value accordingly.

synarete:
> As far as I remember we did this intentionally to build the images for multiple archs from a single place.
> 
> @malayparida2000 can correct me if I am wrong here.

Assuming I have ARM64 (AWS/Graviton) vm at hand, how should one build the `ocs-operator` image from source for this host arch ?

synarete:
> Yes, this was done intentionally as ther's no assurity that the target arch is always the host arch. Whatever is the target arch can be defined by setting the GOARCH value accordingly.

@malayparida2000 How should one build ocs-operator image for arm64 from source? That is, what is the proper command line? (I have both arm64 and x86_64 vms).

synarete:
OK, I now get it. in that case, we may ignore this PR (I will convert it to draft).

PR #2596, opened 2024-05-03
Make all alerts/rules compatible with multicluster mode

aruniiird:
Added 'managedBy' label to all the expressions to avoid any conflicting results.
These changes make sure that we don't have any 'PrometheusRuleFailures' in a multicluster mode scenario

### DIFF
--- a/controllers/storagecluster/prometheus/localcephrules.yaml
+++ b/controllers/storagecluster/prometheus/localcephrules.yaml
@@ -19,19 +19,19 @@ spec:
   - name: telemeter.rules
     rules:
     - expr: |
-        count by (namespace) (ceph_osd_metadata{job="rook-ceph-mgr"})
+        count by (namespace, managedBy) (ceph_osd_metadata{job=~"rook-ceph-mgr|rook-ceph-mgr-external"})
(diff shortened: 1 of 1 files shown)


umangachapagain (commented):
- Are these rules tested on a multicluster setup?
- Is there any difference in output after this change?

aruniiird:
> * Are these rules tested on a multicluster setup?


Tested all the queries on a normal (single mode) internal cluster.
Will check on a multicluster scenario.

> * Is there any difference in output after this change?

Many queries, which are changed here, won't be working properly in a multicluster mode. So now, with the changes, all the queries are supposed to give proper output (without triggering any `PrometheusAlertFailure` errors).
All the output will be backward compatible as all the existing fields will still be there.

aruniiird:
> * Are these rules tested on a multicluster setup?

Now all the query changes are tested with multicluster setup and with single aws cluster.
No issues found

> 
> * Is there any difference in output after this change?

Now, where ever possible, we are getting the external cluster results as well. An example screenshot is given below,

image

umangachapagain (approved):

PR #2599, opened 2024-05-08
Changes to onboarding token API to add quota parameter

rchikatw:
- Removed ID from embedded token as it's not used anywhere
- Added a new quota as URL parameter


rchikatw:
Closing this PR due to local branch issue.

PR #2600, opened 2024-05-08
Changes to the onboarding token api to include quota in the body

rchikatw:
- Passing nil quota to function means unlimited quota
- Quota will be sent as HTTP request body
- Sample request: 
-`{
    "value": 1,
    "unit": "Pi"
}`  or no body
- Valid types are [Gi,Ti,Pi]
- Its been decided that quota will be saved always in Gib hence logic for converting in Gib is added
- Test performed:
Send empty body or value as zero  decoded json looks like this
`{"id":"4ae06e2f-cd04-41b1-b903-81a5947df11d","expirationDate":"1719058928"}`
With valid value it looks like this
`{"id":"64f3458e-bf91-4dc5-b9ac-13cd8ccb6daa","expirationDate":"1719059027","storageQuotaInGiB":1}`
- Test cases for handler will be added as part of different PR

### DIFF
--- a/controllers/util/provider.go
+++ b/controllers/util/provider.go
@@ -19,15 +19,20 @@ import (
 
 // GenerateOnboardingToken generates a token valid for a duration of "tokenLifetimeInHours".
 // The token content is predefined and signed by the private key which'll be read from supplied "privateKeyPath".
-func GenerateOnboardingToken(tokenLifetimeInHours int, privateKeyPath string) (string, error) {
+// The storageQuotaInGiB is optional, and it is used to limit the storage of PVC in the application cluster.
+func GenerateOnboardingToken(tokenLifetimeInHours int, privateKeyPath string, storageQuotaInGiB *uint) (string, error) {
 	tokenExpirationDate := time.Now().
 		Add(time.Duration(tokenLifetimeInHours) * time.Hour).
 		Unix()
 
-	payload, err := json.Marshal(services.OnboardingTicket{
+	ticket := services.OnboardingTicket{
 		ID:             uuid.New().String(),
 		ExpirationDate: tokenExpirationDate,
-	})
+	}
+	if storageQuotaInGiB != nil {
+		ticket.StorageQuotaInGiB = *storageQuotaInGiB
+	}
+	payload, err := json.Marshal(ticket)
 	if err != nil {
 		return "", fmt.Errorf("failed to marshal the payload: %v", err)
 	}
--- a/services/types.go
+++ b/services/types.go
@@ -1,6 +1,7 @@
 package services
 
 type OnboardingTicket struct {
-	ID             string `json:"id"`
-	ExpirationDate int64  `json:"expirationDate,string"`
+	ID                string `json:"id"`
+	ExpirationDate    int64  `json:"expirationDate,string"`
+	StorageQuotaInGiB uint   `json:"storageQuotaInGiB,omitempty"`
 }
--- a/services/ux-backend/handlers/onboardingtokens/handler.go
+++ b/services/ux-backend/handlers/onboardingtokens/handler.go
@@ -1,29 +1,62 @@
 package onboardingtokens
 
 import (
+	"encoding/json"
 	"fmt"
 	"net/http"
 
 	"github.com/red-hat-storage/ocs-operator/v4/controllers/util"
 	"github.com/red-hat-storage/ocs-operator/v4/services/ux-backend/handlers"
 	"k8s.io/klog/v2"
+	"k8s.io/utils/ptr"
 )
 
 const (
 	onboardingPrivateKeyFilePath = "/etc/private-key/key"
 )
 
+var unitToGib = map[string]uint{
+	"Gi": 1,
+	"Ti": 1024,
+	"Pi": 1024 * 1024,
+}
+
 func HandleMessage(w http.ResponseWriter, r *http.Request, tokenLifetimeInHours int) {
 	switch r.Method {
 	case "POST":
-		handlePost(w, tokenLifetimeInHours)
+		handlePost(w, r, tokenLifetimeInHours)
 	default:
 		handleUnsupportedMethod(w, r)
 	}
 }
 
-func handlePost(w http.ResponseWriter, tokenLifetimeInHours int) {
-	if onboardingToken, err := util.GenerateOnboardingToken(tokenLifetimeInHours, onboardingPrivateKeyFilePath); err != nil {
+func handlePost(w http.ResponseWriter, r *http.Request, tokenLifetimeInHours int) {
+	var storageQuotaInGiB *uint
+	// When ContentLength is 0 that means request body is empty and
+	// storage quota is unlimited
+	var err error
+	if r.ContentLength != 0 {
+		var quota = struct {
+			Value uint   `json:"value"`
+			Unit  string `json:"unit"`
+		}{}
+		if err = json.NewDecoder(r.Body).Decode(&quota); err != nil {
+			http.Error(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+
+		if quota.Value == 0 {
+			http.Error(w, fmt.Sprintf("invalid value sent in request body, value should be greater than 0: %v", quota.Value), http.StatusBadRequest)
+			return
+		}
+		unitAsGiB, ok := unitToGib[quota.Unit]
+		if !ok {
+			http.Error(w, fmt.Sprintf("invalid Unit type sent in request body, Valid types are [Gi,Ti,Pi]: %v", quota.Unit), http.StatusBadRequest)
+			return
+		}
+		storageQuotaInGiB = ptr.To(unitAsGiB * quota.Value)
+	}
+	if onboardingToken, err := util.GenerateOnboardingToken(tokenLifetimeInHours, onboardingPrivateKeyFilePath, storageQuotaInGiB); err != nil {
 		klog.Errorf("failed to get onboardig token: %v", err)
 		w.WriteHeader(http.StatusInternalServerError)
 		w.Header().Set("Content-Type", handlers.ContentTypeTextPlain)
--- a/controllers/storagecluster/storageclient.go
+++ b/controllers/storagecluster/storageclient.go
@@ -31,7 +31,7 @@ func (s *storageClient) ensureCreated(r *StorageClusterReconciler, storagecluste
 	storageClient.Name = storagecluster.Name
 	_, err := controllerutil.CreateOrUpdate(r.ctx, r.Client, storageClient, func() error {
 		if storageClient.Status.ConsumerID == "" {
-			token, err := util.GenerateOnboardingToken(tokenLifetimeInHours, onboardingPrivateKeyFilePath)
+			token, err := util.GenerateOnboardingToken(tokenLifetimeInHours, onboardingPrivateKeyFilePath, nil)
 			if err != nil {
 				return fmt.Errorf("unable to generate onboarding token: %v", err)
 			}


services/types.go
```diff
@@ -3,4 +3,5 @@ package services
 type OnboardingTicket struct {
 	ID             string `json:"id"`
 	ExpirationDate int64  `json:"expirationDate,string"`
+	Quota          string `json:"quota"`
```
Madhu-1, line 6: where does this quota get validated with the existing quota in ceph? or we don't have any plan to consider the ceph quota and users can set whatever quota they want?

nb-ohad:
@rchikatw Why use 0 or unlimited/unbound? why not have it a pointer and use nil instead?

controllers/util/provider.go
```diff
@@ -19,14 +19,15 @@ import (
 
 // GenerateOnboardingToken generates a token valid for a duration of "tokenLifetimeInHours".
 // The token content is predefined and signed by the private key which'll be read from supplied "privateKeyPath".
-func GenerateOnboardingToken(tokenLifetimeInHours int, privateKeyPath string) (string, error) {
+func GenerateOnboardingToken(tokenLifetimeInHours int, privateKeyPath, quota string) (string, error) {
```
nb-ohad, line 22: Quota shouldn't be a string, it should be wither a number (assuming a specific unit) or a size object.
Also it should preferably be optional (a pointer) so we can generate a token that does not have a quota

nb-ohad (changes requested):

controllers/util/provider.go
rchikatw: I have considered this as a string because in ClusterResourceQuota we also need to mention the unit. It will be better if we get that from the UI itself.

nb-ohad: Even it that case this should not be a string but a k8s quantity object

rchikatw: Okay will do the changes according

services/types.go
Madhu-1: which means we can use quota more than the ceph cluster on all the consumer clusters.

nb-ohad: The quota on the onboarding token cannot be validated as generating an onboarding token is a customer-owned operation. What we can do we can decide to not allow onboarding using this token if the quota does not make sense

Madhu-1: that make sense, this is only to generate the token and we are going to have followup PR to block such tokens with quota, please mark this as resolved as i don't have permission to do it.

nb-ohad: @rchikatw  please take note to issue a follow-up PR with the needed changes.

controllers/util/provider.go
```diff
@@ -15,18 +15,21 @@ import (
 
 	"github.com/google/uuid"
 	"github.com/red-hat-storage/ocs-operator/v4/services"
+	"k8s.io/apimachinery/pkg/api/resource"
 )
 
 // GenerateOnboardingToken generates a token valid for a duration of "tokenLifetimeInHours".
 // The token content is predefined and signed by the private key which'll be read from supplied "privateKeyPath".
-func GenerateOnboardingToken(tokenLifetimeInHours int, privateKeyPath string) (string, error) {
+// The quantity is optional, and it is used to limit the storage of PVC in the application cluster.
```
leelavg, line 23: so nil value of `quantity` is a way for saying no restrictions in consumption?

services/ux-backend/handlers/onboardingtokens/handler.go
```diff
@@ -16,14 +18,35 @@ const (
 func HandleMessage(w http.ResponseWriter, r *http.Request, tokenLifetimeInHours int) {
 	switch r.Method {
 	case "POST":
-		handlePost(w, tokenLifetimeInHours)
+		type Message struct {
+			Quota string `json:"quota"`
+		}
+		var quota Message
```
leelavg, line 24: 1. I don't think we should be processing the body before sending it to `handlePost` cause in the switch you are only deciding the method
2. No need to create a type if we are not going to initialize it, can get away w/ var. Unrelated one, for encoding & decoding you only need fields to be exported not necessarily the struct.
```suggestion
		var quota = struct { Quota string `json:"quota"`}{}
```

services/ux-backend/handlers/onboardingtokens/handler.go
```diff
@@ -16,14 +18,35 @@ const (
...
+		if err != nil {
+			if err.Error() == "EOF" {
+				handlePost(w, tokenLifetimeInHours, nil)
+				return
+			}
+			http.Error(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+
+		quantity, err := resource.ParseQuantity(quota.Quota)
```
leelavg, line 35: `quota.Quota` will be empty for older clients which breaks backward compat unless `resource.ParseQuantity` resistant to empty strings. is mandating it a requirement during onboard?

services/ux-backend/handlers/onboardingtokens/handler.go
```diff
@@ -16,14 +18,35 @@ const (
 func HandleMessage(w http.ResponseWriter, r *http.Request, tokenLifetimeInHours int) {
 	switch r.Method {
 	case "POST":
-		handlePost(w, tokenLifetimeInHours)
+		type Message struct {
+			Quota string `json:"quota"`
+		}
+		var quota Message
+		err := json.NewDecoder(r.Body).Decode(&quota)
+		if err != nil {
```
leelavg, line 26: nit, can perform err handling in the same statement/line.

services/ux-backend/handlers/onboardingtokens/handler.go
rchikatw: This change is while generation of Token only. Older clients are already connected. they don't need to worry about quota for them.

services/ux-backend/handlers/onboardingtokens/handler.go
```diff
@@ -52,3 +59,26 @@ func handleUnsupportedMethod(w http.ResponseWriter, r *http.Request) {
...
+		Quota string `json:"quota"`
+	}{}
+	err := json.NewDecoder(r.Body).Decode(&quota)
+	if err != nil && err.Error() != "EOF" {
+		return nil, err
+	}
+
+	// When length is 0 that means either request body is empty or
+	// quota parameter is not passed in that case quota is considerred
+	// to be unlimited
```
leelavg, line 74: no, we already established that empty body represents unlimited quota and better not to club that w/ empty value for quota.

btw, as mentioned in the comment, I don't think `quota` is a param.

services/ux-backend/handlers/onboardingtokens/handler.go
```diff
@@ -16,14 +18,19 @@ const (
...
-		handlePost(w, tokenLifetimeInHours)
+		handlePost(w, tokenLifetimeInHours, r)
 	default:
 		handleUnsupportedMethod(w, r)
 	}
 }
 
-func handlePost(w http.ResponseWriter, tokenLifetimeInHours int) {
-	if onboardingToken, err := util.GenerateOnboardingToken(tokenLifetimeInHours, onboardingPrivateKeyFilePath); err != nil {
+func handlePost(w http.ResponseWriter, tokenLifetimeInHours int, r *http.Request) {
```
leelavg, line 27: could you pls swap 2nd & 3rd param?

services/ux-backend/handlers/onboardingtokens/handler.go
leelavg: yes, good one. resolve.

services/ux-backend/handlers/onboardingtokens/handler.go
rchikatw: Done

services/ux-backend/handlers/onboardingtokens/handler.go
rchikatw: Yes, correct, we are only checking with the request body only. I missed updating this comment.

services/ux-backend/handlers/onboardingtokens/handler.go
```diff
@@ -52,3 +59,25 @@ func handleUnsupportedMethod(w http.ResponseWriter, r *http.Request) {
...
+	if len(quota.StorageQuota) == 0 {
+		return nil, nil
+	}
+	quantity, err := resource.ParseQuantity(quota.StorageQuota)
+	if err != nil {
+		return nil, fmt.Errorf("invalid StorageQuota value sent in request body: %v", err)
+	}
+	return &quantity, nil
+
+}
```
nb-ohad, line 83: I am not sure about the logic here. If the UI is passing us a quantity I would expect to see 2 fields in the body. One for the unit and one for the size. But if there is an assumption about the unit then there is no reason to parse it as a quantity, it should be just an int. 

Anything else just add unnecessary complexity to the system

nb-ohad (changes requested):

services/ux-backend/handlers/onboardingtokens/handler.go
```diff
@@ -16,14 +18,19 @@ const (
...
+		handlePost(w, r, tokenLifetimeInHours)
 	default:
 		handleUnsupportedMethod(w, r)
 	}
 }
 
-func handlePost(w http.ResponseWriter, tokenLifetimeInHours int) {
-	if onboardingToken, err := util.GenerateOnboardingToken(tokenLifetimeInHours, onboardingPrivateKeyFilePath); err != nil {
+func handlePost(w http.ResponseWriter, r *http.Request, tokenLifetimeInHours int) {
+	quantity, err := getQuantity(r)
```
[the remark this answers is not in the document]
rchikatw: Yes from the UI design slides user can select both size and unit.

controllers/util/provider.go
```diff
@@ -15,18 +15,21 @@ import (
...
+func GenerateOnboardingToken(tokenLifetimeInHours int, privateKeyPath string, quantity *resource.Quantity) (string, error) {
 	tokenExpirationDate := time.Now().
 		Add(time.Duration(tokenLifetimeInHours) * time.Hour).
 		Unix()
 
 	payload, err := json.Marshal(services.OnboardingTicket{
 		ID:             uuid.New().String(),
 		ExpirationDate: tokenExpirationDate,
+		StorageQuota:   quantity,
```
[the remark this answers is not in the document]
rchikatw: After marshal quantity will look like `<size><Unit>` ex: `1Gi.` Is it unmarshalabe? -> Yes it.

Why do we need a complex type for quantity? In your previous comment you suggested to using k8s quantity Refer: https://github.com/red-hat-storage/ocs-operator/pull/2600#discussion_r1597427959

Advantages of using quantity:  
1. we can validate quota unit is valid (using parseQuantity function) no need to add manual checks for the unit
2. You have also suggested using a pointer so we can generate a token that does not have a quota. Refer https://github.com/red-hat-storage/ocs-operator/pull/2600#discussion_r1593849621

If we don't want to use complex data type then we should not allow the user to select the unit in the UI and make quantity int that should solve our problem not using complex data type.

Let me know which approach should I use k8s quantity or int

services/ux-backend/handlers/onboardingtokens/handler.go
rchikatw: UI will be passed as a string in the body with the format `1Gi/1Mi/1Ti` to indicate quantity. I will decode and parse it to validate the correctness of the passed value. Invalid values will look like `1/1gi/1ti`.

services/types.go
```diff
@@ -3,4 +3,5 @@ package services
 type OnboardingTicket struct {
 	ID             string `json:"id"`
 	ExpirationDate int64  `json:"expirationDate,string"`
+	StorageQuota   *int   `json:"storageQuota,omitempty"`
```
nb-ohad, line 6: The content of the token was designed to be viewable by the end user. Let's be more specific with the name to remove any kind of ambiguity 
```suggestion
	StorageQuotaInGiB   *int   `json:"storageQuotaInGiB,omitempty"`
```

services/ux-backend/handlers/onboardingtokens/handler.go
```diff
@@ -52,3 +58,40 @@ func handleUnsupportedMethod(w http.ResponseWriter, r *http.Request) {
...
+		StorageQuota int    `json:"storageQuota"`
+		Unit         string `json:"unit"`
+	}{}
+	err := json.NewDecoder(r.Body).Decode(&quota)
+	if err != nil {
+		// When error is EOF that means request body is empty and quota
+		// is unlimited
+		if err.Error() == "EOF" {
+			return nil, nil
+		}
```
nb-ohad, line 73: But it does not mean has you can get an EOF even if you have a broken JSON, or a partial JSON. 
EOF means you got to the end of the data stream before a complete json have been read.

If you want to check for an empty body, check the length of the bytes slice/array before sending it to the decode function

services/ux-backend/handlers/onboardingtokens/handler.go
```diff
@@ -52,3 +58,40 @@ func handleUnsupportedMethod(w http.ResponseWriter, r *http.Request) {
 		klog.Errorf("failed write data to response writer: %v", err)
 	}
 }
+
+func getAllocatedQuotaInGib(r *http.Request) (*int, error) {
+	var quota = struct {
+		StorageQuota int    `json:"storageQuota"`
+		Unit         string `json:"unit"`
+	}{}
```
nb-ohad, line 66: Do we really need a custom struct? can't we use the known quantity struct to decode into?

services/ux-backend/handlers/onboardingtokens/handler.go
```diff
@@ -52,3 +58,40 @@ func handleUnsupportedMethod(w http.ResponseWriter, r *http.Request) {
 		klog.Errorf("failed write data to response writer: %v", err)
 	}
 }
+
+func getAllocatedQuotaInGib(r *http.Request) (*int, error) {
+	var quota = struct {
+		StorageQuota int    `json:"storageQuota"`
```
nb-ohad, line 64: Using the word quota inside an object that is already defined as quota creates confusion. 
A quota is a quantity that is built from a unit and a value
```suggestion
		Value int    `json:"value"`
```

services/ux-backend/handlers/onboardingtokens/handler.go
```diff
@@ -52,3 +58,40 @@ func handleUnsupportedMethod(w http.ResponseWriter, r *http.Request) {
...
+	switch quota.Unit {
+	case "Gi":
+		quotaInGib = quota.StorageQuota * 1
+	case "Ti":
+		quotaInGib = quota.StorageQuota * 1024
+	case "Pi":
+		quotaInGib = quota.StorageQuota * 1024 * 1024
+	default:
+		return nil, fmt.Errorf("invalid Unit type sent in request body, Valid types are [Gi,Ti,Pi]: %v", quota.Unit)
+	}
```
nb-ohad, line 94: wouldn't a fixed map be better here?
So you will have:

unitAsGiB, ok := UnitToGib[qouta.Unit] 
if !ok { 
    handle error
}
return toPtr(quota.Value * unitAsGiB)

nb-ohad (changes requested):

services/ux-backend/handlers/onboardingtokens/handler.go
rchikatw: Sure, I can verify that the content length is 0.

services/ux-backend/handlers/onboardingtokens/handler.go
rchikatw: Yes we need a custom quantity to decode.

services/ux-backend/handlers/onboardingtokens/handler.go
```diff
@@ -52,3 +59,35 @@ func handleUnsupportedMethod(w http.ResponseWriter, r *http.Request) {
...
+		return nil, fmt.Errorf("invalid Value sent in request body: %v", quota.Value)
+	}
+	if quota.Unit == "" {
+		return nil, fmt.Errorf("invalid Unit type sent in request body, Valid types are [Gi,Ti,Pi]: %v", quota.Unit)
+	}
+	UnitToGib := map[string]int{
+		"Gi": 1,
+		"Ti": 1024,
+		"Pi": 1048576,
+	}
```
nb-ohad, line 87: Why create this map again and again for each call? You can move it to file scope to solve that.  And make sure to have the name starting with a lowercase letter to make it private

services/ux-backend/handlers/onboardingtokens/handler.go
```diff
@@ -52,3 +59,35 @@ func handleUnsupportedMethod(w http.ResponseWriter, r *http.Request) {
...
+	if quota.Value <= 0 {
+		return nil, fmt.Errorf("invalid Value sent in request body: %v", quota.Value)
+	}
+	if quota.Unit == "" {
+		return nil, fmt.Errorf("invalid Unit type sent in request body, Valid types are [Gi,Ti,Pi]: %v", quota.Unit)
+	}
+	UnitToGib := map[string]int{
+		"Gi": 1,
+		"Ti": 1024,
+		"Pi": 1048576,
```
nb-ohad, line 86: Please communicate how you got to the number  1048576. There are 2 options:
```suggestion
		"Pi": 1024 * 1024,
```

or 

```suggestion
		"Pi": 1048576, // 1024 * 1024
```

services/ux-backend/handlers/onboardingtokens/handler.go
```diff
@@ -52,3 +59,35 @@ func handleUnsupportedMethod(w http.ResponseWriter, r *http.Request) {
 		klog.Errorf("failed write data to response writer: %v", err)
 	}
 }
+
+func getQuotaInGib(r *http.Request) (*int, error) {
+	// When ContentLength is 0 that means request body is empty and
+	// storage quota is unlimited
+	if r.ContentLength == 0 {
+		return nil, nil
+	}
```
nb-ohad, line 68: This function should only handle a single concern: Covert a string containing a quota to a number

As such I expect to have a string passed into the function and not an HTTP request object.
This also means that the func should assume that a quota is codified in the string so the return value type is `int` and not `*int`

The handling of the request should be moved to the calling site and this fun should only be invoked if a  value was extracted from the request

services/types.go
```diff
@@ -1,6 +1,7 @@
 package services
 
 type OnboardingTicket struct {
-	ID             string `json:"id"`
-	ExpirationDate int64  `json:"expirationDate,string"`
+	ID                string `json:"id"`
+	ExpirationDate    int64  `json:"expirationDate,string"`
+	StorageQuotaInGiB *int   `json:"storageQuotaInGiB,omitempty"`
```
nb-ohad, line 6: There is no need to have this as a pointer. As of the "omitempty" tag a value of 0 will not be marshaled and 0 has no meaningful value as a quota

controllers/util/provider.go
```diff
@@ -19,14 +19,16 @@ import (
...
 		Add(time.Duration(tokenLifetimeInHours) * time.Hour).
 		Unix()
 
 	payload, err := json.Marshal(services.OnboardingTicket{
-		ID:             uuid.New().String(),
-		ExpirationDate: tokenExpirationDate,
+		ID:                uuid.New().String(),
+		ExpirationDate:    tokenExpirationDate,
+		StorageQuotaInGiB: storageQuotaInGiB,
 	})
```
nb-ohad, line 32:
```suggestion
	ticket := services.OnboardingTicket{
		ID:                uuid.New().String(),
		ExpirationDate:    tokenExpirationDate,
	})
	if storageQuotaInGiB != nil {
		ticket.StorageQuotaInGiB = *storageQuotaInGiB
	}
	payload, err := json.Marshal(ticket)
	```

**Review by nb-ohad — CHANGES_REQUESTED** [review]

File: services/ux-backend/handlers/onboardingtokens/handler.go (hunk 19 above)
Comment (rchikatw) [reply services/ux-backend/handlers/onboardingtokens/handler.go]: As the map creation moved to file scope, it doesn't make sense to have a function to read from the map, so I removed this function.

File: controllers/util/provider.go [hunk 22]
```diff
@@ -19,15 +19,20 @@ import (
...
 		Unix()
 
-	payload, err := json.Marshal(services.OnboardingTicket{
+	ticket := services.OnboardingTicket{
 		ID:             uuid.New().String(),
 		ExpirationDate: tokenExpirationDate,
-	})
+	}
+	if storageQuotaInGiB != nil {
+		ticket.StorageQuotaInGiB = *storageQuotaInGiB
```
leelavg, line 33: I think in the absence of body, this will always be Zero, is that fine?
```
var val int // gets a zero value
gen(_, _, ptr.To(val)) // you are always passing a pointer and so the 3rd param to generate isn't nil ever
```

I believe below could work, pls cross check
```
var val *int
t := unit * q.V
val = &t
gen(_, _, val)
```

rchikatw: Yes, in the absence of a body, this will always be Zero.

You are correct. During my testing, it worked because after marshaling, the field was not being passed due to the omitempty tag being added at the struct level. I will make the changes as you suggested because that makes more sense to me.

leelavg: > due to the omitempty tag being added at the struct level
- ya, that could be fine as well and so I don't mind change this then unless other reviewers also want to change this.

nb-ohad, line 33: @rchikatw You decided to use an *int as opposed to *uint. You need to validate that the passed value is greater than 0 to eliminate cases where someone invoked with 
0 or a negative number

nb-ohad: Please consider we do need to support nil in cases where a quota is not provided. We already defined that a quota with value of 0 (or negative value) is invalid and that a omitted quota should be nil

services/types.go
```diff
@@ -1,6 +1,7 @@
 package services
 
 type OnboardingTicket struct {
-	ID             string `json:"id"`
-	ExpirationDate int64  `json:"expirationDate,string"`
+	ID                string `json:"id"`
+	ExpirationDate    int64  `json:"expirationDate,string"`
+	StorageQuotaInGiB int    `json:"storageQuotaInGiB,omitempty"`
```
nb-ohad, line 6: Is there a reason why we should not use a `uint` for quota values?

services/ux-backend/handlers/onboardingtokens/handler.go
```diff
@@ -13,17 +14,49 @@ const (
...
+		handlePost(w, r, tokenLifetimeInHours)
 	default:
 		handleUnsupportedMethod(w, r)
 	}
 }
 
-func handlePost(w http.ResponseWriter, tokenLifetimeInHours int) {
-	if onboardingToken, err := util.GenerateOnboardingToken(tokenLifetimeInHours, onboardingPrivateKeyFilePath); err != nil {
+func handlePost(w http.ResponseWriter, r *http.Request, tokenLifetimeInHours int) {
+	var storageQuotaInGiB *int
```
nb-ohad, line 33: Shouldn't this be `*uint`?

services/ux-backend/handlers/onboardingtokens/handler.go
```diff
@@ -13,17 +14,49 @@ const (
...
+			return
+		}
+
+		unitAsGiB, ok := unitToGib[quota.Unit]
+		if !ok {
+			http.Error(w, fmt.Sprintf("invalid Unit type sent in request body, Valid types are [Gi,Ti,Pi]: %v", quota.Unit), http.StatusBadRequest)
+			return
+		}
+		val := unitAsGiB * quota.Value
+		storageQuotaInGiB = &val
```
nb-ohad, line 57: No need for intermiddate var
```suggestion
		storageQuotaInGiB = ptr.To(unitAsGiB * quota.Value)
```

nb-ohad (changes requested):

controllers/util/provider.go
rchikatw: I am validating against negative numbers here https://github.com/red-hat-storage/ocs-operator/pull/2600/files#diff-5575cae66180001e0066d01f86d13ae4a7137f5424f09f4af76abfefc5eb8dd7R46 but using `*unit` no need to add this condition.

services/types.go
rchikatw: There is no such reason i have added condition to validate against negative number so used `int` will remove it and use uint

controllers/util/provider.go
rchikatw: Verified with 0 as well nil it working as expected.

nb-ohad: Quota need to be validate to be greater then 0 which is not forced by uint

nb-ohad (changes requested):

PR #2606, opened 2024-05-14
Migrate osds to bluestore

sp98:
EPIC: https://issues.redhat.com/browse/RHSTOR-5722

We need to migrate `bluestore-rdr` OSDs back to `bluestore` in 4.17 because both are merged in RHCS 8.0.  Following changes are added in the PR: 

**Greenfield**: Don't use bluestore-rdr store type in `cephCluster` because its not valid for 4.17 (RHCS 8.0)

**Brownfield**: If an existing cluster is migrating to 4.17, then update the OSD store type to from `bluestore-rdr` to `bluestore` and also add `yes-really-update-store` to the `cephCluster` to start the OSD migration to `bluestore`. 

Remove `isDROptmized` flag from `clusterclaims` as clusters will be DR optimized by default.


controllers/storagecluster/clusterclaims.go
```diff
@@ -102,15 +101,10 @@ func (obj *ocsClusterClaim) ensureCreated(r *StorageClusterReconciler, instance
...
-	var isDROptimized = "false"
-	// Set isDROptmized to "false" in case of external clusters as we currently don't have to way to determine
-	// if external cluster OSDs are using bluestore-rdr
-	if !instance.Spec.ExternalStorage.Enable {
-		isDROptimized, err = creator.getIsDROptimized()
-		if err != nil {
-			r.Log.Error(err, "failed to get cephcluster status. retrying again")
-			return reconcile.Result{}, err
-		}
+	var isDROptimized = "true"
```
sp98, line 104: @umangachapagain 
This PR is for the [EPIC](https://issues.redhat.com/browse/RHSTOR-5722)

I don't think we we would be need `IsDROptimized` in the clusterClaims anymore?  
Should we remove it completely or just keep it to `true`?  (All 4.17 clusters will be DR optimized by default)

umangachapagain: Remove it.

sp98: removed it.

controllers/storagecluster/cephcluster.go
```diff
@@ -453,7 +445,7 @@ func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, kmsConfigMap *co
 			},
 			Storage: rookCephv1.StorageScopeSpec{
 				StorageClassDeviceSets:       newStorageClassDeviceSets(sc),
-				Store:                        osdStore,
+				Store:                        rookCephv1.OSDStore{},
```
iamniting, line 448: Can not we just remove this field instead of having an empty struct?

iamniting (changes requested):

controllers/storagecluster/cephcluster.go
sp98: removed.

iamniting (approved):

PR #2613, opened 2024-05-15
Add upgrade reliability params in storagecluster CR

Nikhil-Ladha:
Add upgrade reliability parameters to storagecluster CR
and overriding the default values in cephCluster CR if the
values are defined in the storagecluster CR

JIRA: https://issues.redhat.com/browse/RHSTOR-5734


iamniting (approved):
Don't we need to make any checks in the ceph cluster status? Otherwise changes looks good to me. 

/hold for Umanga reviews.

Nikhil-Ladha:
> Don't we need to make any checks in the ceph cluster status? Otherwise changes looks good to me.

From what I understand, no we don't need to make any status changes as such. If at all needed, rook will be taking care of it. It reads the cephCluster CR and does other related operations based on that and will update the Status where and all needed.
@travisn please correct me if I am wrong.

controllers/storagecluster/cephcluster.go
```diff
@@ -491,6 +491,18 @@ func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, kmsConfigMap *co
 		},
 	}
 
+	if sc.Spec.ManagedResources.CephCluster.ContinueUpgradeAfterChecksEvenIfNotHealthy {
```
travisn, line 494: It should be same result even if you don't check these conditions, and always set the bool. 
```suggestion
```

travisn (commented):
> > Don't we need to make any checks in the ceph cluster status? Otherwise changes looks good to me.
> 
> From what I understand, no we don't need to make any status changes as such. If at all needed, rook will be taking care of it. It reads the cephCluster CR and does other related operations based on that and will update the Status where and all needed. @travisn please correct me if I am wrong.

Agreed, no status checks needed here, let's keep it simple and just pass through to Rook

controllers/storagecluster/cephcluster.go
Nikhil-Ladha: So, in that case we will have to define default values in the CR, and that will lead to those fields present in the storagecluster CR by default and visible to everyone, which we don't want (as per the last discussion).
That's the reason I added this condition to only update, if present in SC CR

travisn: Perhaps that only applies to the `bool?` type? A `bool` already has a default value of false, so I don't see how this will change customer visibility.

Nikhil-Ladha: Right, agreed.
In the epic, description for `ContinueUpgradeAfterChecksEvenIfNotHealthy` the default value mentioned is `true`, but I think that's not the case of go by the usage of the property? If so, then we can remove the conditionals but, if that's not the case then we would have to define `default: true` in the spec and that would make the property visible in the CR after creation. So, just need a clarity on that.

travisn: Actually, `ContinueUpgradeAfterChecksEvenIfNotHealthy` is a special case since its default value is defined up on line 475. So for this setting we would need to define the StorageCluster as a `bool?` and only override the default if explicitly set. So we just as well take the same approach and use `bool?` for the other two settings as well.

Nikhil-Ladha: I have defined all of them as bool types only, and override the default if set in the StorageCluster CR.
So, are you suggesting that I should just do something like this:
```
         cephCluster := &rookCephv1.CephCluster {
                      .....
                      .....
                      WaitTimeoutForHealthyOSDInMinutes: getWaitTimeoutForHealthOSD(sc),
                      SkipUpgradeChecks: sc.Spec.ManagedResources.CephCluster.SkipUpgradeChecks
                      UpgradeOSDRequiresHealthyPGs: sc.Spec.ManagedResources.CephCluster.UpgradeOSDRequiresHealthyPGs
		},
	}

	if sc.Spec.ManagedResources.CephCluster.ContinueUpgradeAfterChecksEvenIfNotHealthy {
		cephCluster.Spec.ContinueUpgradeAfterChecksEvenIfNotHealthy = sc.Spec.ManagedResources.CephCluster.ContinueUpgradeAfterChecksEvenIfNotHealthy
	}
```

travisn: Clarification... I meant to define a `bool*` in the StorageCluster CRD definition. The problem for a `bool` is that we can't tell the difference between the user setting it to `false`, or if it is not specified at all. With a `bool*`, the check would be something like:
```
if sc.Spec.ManagedResources.CephCluster.ContinueUpgradeAfterChecksEvenIfNotHealthy != nil {
		cephCluster.Spec.ContinueUpgradeAfterChecksEvenIfNotHealthy = *sc.Spec.ManagedResources.CephCluster.ContinueUpgradeAfterChecksEvenIfNotHealthy
	}
```

Looks good to set the other two values above since they already have `false` for a default.

Nikhil-Ladha: Made the changes.

controllers/storagecluster/cephcluster.go
```diff
@@ -488,9 +488,15 @@ func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, kmsConfigMap *co
 				},
 			},
 			WaitTimeoutForHealthyOSDInMinutes: getWaitTimeoutForHealthOSD(sc),
+			SkipUpgradeChecks:                 sc.Spec.ManagedResources.CephCluster.SkipUpgradeChecks,
```
travisn, line 491: nit: How about moving line 475 down next to these other upgrade settings?

travisn (commented):
LGTM, just a tiny suggestion!

travisn (commented):
/lgtm

controllers/storagecluster/cephcluster.go
```diff
@@ -445,6 +445,7 @@ func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, kmsConfigMap *co
 				ManagePodBudgets:                 true,
 				ManageMachineDisruptionBudgets:   false,
 				MachineDisruptionBudgetNamespace: "openshift-machine-api",
+				OSDMaintenanceTimeout:            getOsdMaintenanceTimeout(sc),
```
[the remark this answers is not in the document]
Nikhil-Ladha: I was thinking this would allow us to set a different default under ocs-operator itself, if needed and we won't have to change the rook default.
But, if you suggest we let rook only take care of the defaults then I will remove the default setting from here.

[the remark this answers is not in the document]
Nikhil-Ladha: Updated the code to remove the default setting from ocs-operator.

controllers/defaults/defaults.go
```diff
@@ -54,7 +52,4 @@ var (
 	// ArbiterReplicasPerFailureDomain is the default replica count in the failure domain when arbiter is enabled
 	// This maps to the ReplicasPerFailureDomain in the CephReplicatedSpec when creating the CephBlockPools
 	ArbiterReplicasPerFailureDomain = 2
-	// DefaultWaitTimeoutForHealthyOSD is the default time for which the operator would wait before an OSD can be stopped
-	// for an upgrade or restart
-	DefaultWaitTimeoutForHealthyOSD = 10 * time.Minute
```
Nikhil-Ladha, line 59: We had these in previous release but I guess that should be fine, it's not breaking anything.

controllers/storagecluster/cephcluster.go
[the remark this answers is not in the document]
travisn: If there is a need for a separate default, that would be fine to declare it here, it just didn't seem necessary. At the same time, I don't expect the Rook default to change either.

travisn (commented):
/lgtm

Nikhil-Ladha:
@iamniting are we good to merge?
Or, still waiting for Umanga's review?

iamniting (approved):
/unhold

PR #2615, opened 2024-05-16
Set deviceClasses to avoid replicated pool spreading PGs across all OSDs

malayparida2000:
BZ-https://bugzilla.redhat.com/show_bug.cgi?id=2274175

Set the deviceClasses of the pools to avoid wrong data placement 

Earlier when replica-1 was getting enabled on an existing cluster the pool was setting it's deviceClass to replicated but there were no osds with this deviceClass as osd's during normal creation were given SSD deviceClass if blank, and now they can't change. To avoid & solve this problem we have to intelligently set the deviceClass for the pools, Depending on whether replica-1 is enabled & the number of deviceClasses found in the cephCluster CR status.

To do this we added a DefaultCephDeviceClass field in StorageCluster Status which will  hold the default ceph device class to be used for the pools. Depending  on whether replica-1 is enabled & the number of deviceClasses found in  the status this value is determined and set in the status.

### DIFF
--- a/controllers/storagecluster/cephblockpools.go
+++ b/controllers/storagecluster/cephblockpools.go
@@ -61,7 +61,7 @@ func (r *StorageClusterReconciler) newCephBlockPoolInstances(initData *ocsv1.Sto
 			},
 			Spec: cephv1.NamedBlockPoolSpec{
 				PoolSpec: cephv1.PoolSpec{
-					DeviceClass:    generateDeviceClass(initData),
+					DeviceClass:    initData.Status.DefaultCephDeviceClass,
 					FailureDomain:  getFailureDomain(initData),
 					Replicated:     generateCephReplicatedSpec(initData, "data"),
 					EnableRBDStats: true,
@@ -113,7 +113,7 @@ func (r *StorageClusterReconciler) newCephBlockPoolInstances(initData *ocsv1.Sto
 				Spec: cephv1.NamedBlockPoolSpec{
 					Name: ".nfs",
 					PoolSpec: cephv1.PoolSpec{
-						DeviceClass:    generateDeviceClass(initData),
+						DeviceClass:    initData.Status.DefaultCephDeviceClass,
 						FailureDomain:  getFailureDomain(initData),
 						Replicated:     generateCephReplicatedSpec(initData, "data"),
 						EnableRBDStats: true,
--- a/controllers/storagecluster/cephcluster.go
+++ b/controllers/storagecluster/cephcluster.go
@@ -266,6 +266,11 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
 		return reconcile.Result{}, nil
 	}
 
+	// Set the default CephDeviceClass in the StorageCluster status(not needed for external clusters)
+	if !sc.Spec.ExternalStorage.Enable && found.Status.CephStorage != nil {
+		sc.Status.DefaultCephDeviceClass = determineDefaultCephDeviceClass(found.Status.CephStorage.DeviceClasses, sc.Spec.ManagedResources.CephNonResilientPools.Enable, sc.Status.FailureDomainValues)
+	}
+
 	// Record actual Ceph container image version before attempting update
 	sc.Status.Images.Ceph.ActualImage = found.Spec.CephVersion.Image
 
@@ -847,10 +852,6 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster) []rookCephv1.StorageCla
 			annotations := map[string]string{
 				"crushDeviceClass": crushDeviceClass,
 			}
-			// if Non-Resilient Pools are enabled then change the existing osd crushDeviceClass to "replicated"
-			if sc.Spec.ManagedResources.CephNonResilientPools.Enable {
-				annotations["crushDeviceClass"] = "replicated"
-			}
 			// Annotation crushInitialWeight is an optional, explicit weight to set upon OSD's init (as float, in TiB units).
 			// ROOK & Ceph do not want any (optional) Ti[B] suffix, so trim it here.
 			// If not set, Ceph will define OSD's weight based on its capacity.
@@ -1340,4 +1341,28 @@ func getOsdCount(sc *ocsv1.StorageCluster) int {
 		osdCount += ds.Count
 	}
 	return osdCount
+}
+
+func determineDefaultCephDeviceClass(foundDeviceClasses []rookCephv1.DeviceClasses, isReplica1 bool, replica1DeviceClasses []string) string {
+	// If device classes are found in status
+	if len(foundDeviceClasses) != 0 {
+		// If replica-1 is not enabled return the first device class
+		if !isReplica1 {
+			return foundDeviceClasses[0].Name
+		}
+		// If replica-1 is enabled return the first one that is not a replica1 device class
+		if len(foundDeviceClasses) != 0 {
+			replica1DeviceClassMap := make(map[string]bool)
+			for _, deviceClass := range replica1DeviceClasses {
+				replica1DeviceClassMap[deviceClass] = true
+			}
+			for _, deviceClass := range foundDeviceClasses {
+				if found := replica1DeviceClassMap[deviceClass.Name]; !found {
+					return deviceClass.Name
+				}
+			}
+		}
+	}
+	// By default return "ssd"
+	return "ssd"
 }
--- a/controllers/storagecluster/generate.go
+++ b/controllers/storagecluster/generate.go
@@ -44,14 +44,6 @@ func generateNameForCephBlockPool(initData *ocsv1.StorageCluster) string {
 	return fmt.Sprintf("%s-cephblockpool", initData.Name)
 }
 
-func generateDeviceClass(initData *ocsv1.StorageCluster) string {
-	if initData.Spec.ManagedResources.CephNonResilientPools.Enable {
-		return "replicated"
-	}
-	// if Non-Resilient pools are not enabled then keep the device class blank
-	return ""
-}
-
 func generateNameForNonResilientCephBlockPool(initData *ocsv1.StorageCluster, failureDomainValue string) string {
 	return fmt.Sprintf("%s-cephblockpool-%s", initData.Name, failureDomainValue)
 }
--- a/metrics/vendor/github.com/red-hat-storage/ocs-operator/api/v4/v1/storagecluster_types.go
+++ b/metrics/vendor/github.com/red-hat-storage/ocs-operator/api/v4/v1/storagecluster_types.go
@@ -569,6 +569,9 @@ type StorageClusterStatus struct {
 	// Images holds the image reconcile status for all images reconciled by the operator
 	Images ImagesStatus `json:"images,omitempty"`
 
+	// DefaultCephDeviceClass holds the default ceph device class to be used for the pools
+	DefaultCephDeviceClass string `json:"defaultCephDeviceClass,omitempty"`
+
 	// KMSServerConnection holds the connection state to the KMS server.
 	KMSServerConnection KMSServerConnectionStatus `json:"kmsServerConnection,omitempty"`
 
--- a/api/v1/storagecluster_types.go
+++ b/api/v1/storagecluster_types.go
@@ -569,6 +569,9 @@ type StorageClusterStatus struct {
 	// Images holds the image reconcile status for all images reconciled by the operator
 	Images ImagesStatus `json:"images,omitempty"`
 
+	// DefaultCephDeviceClass holds the default ceph device class to be used for the pools
+	DefaultCephDeviceClass string `json:"defaultCephDeviceClass,omitempty"`
+
 	// KMSServerConnection holds the connection state to the KMS server.
 	KMSServerConnection KMSServerConnectionStatus `json:"kmsServerConnection,omitempty"`
 
--- a/config/crd/bases/ocs.openshift.io_storageclusters.yaml
+++ b/config/crd/bases/ocs.openshift.io_storageclusters.yaml
@@ -8026,6 +8026,10 @@ spec:
                 description: CurrentMonCount holds the value of ceph mons configured
                   in ceph cluster.
                 type: integer
+              defaultCephDeviceClass:
+                description: DefaultCephDeviceClass holds the default ceph device
+                  class to be used for the pools
+                type: string
               externalSecretHash:
                 description: ExternalSecretHash holds the checksum value of external
                   secret data.
--- a/controllers/storagecluster/cephcluster_test.go
+++ b/controllers/storagecluster/cephcluster_test.go
@@ -1584,3 +1584,140 @@ func TestEnsureUpgradeReliabilityParams(t *testing.T) {
 	assert.Equal(t, 20*time.Minute, expected.Spec.WaitTimeoutForHealthyOSDInMinutes)
 	assert.Equal(t, 45*time.Minute, expected.Spec.DisruptionManagement.OSDMaintenanceTimeout)
 }
+
+func TestDetermineDefaultCephDeviceClass(t *testing.T) {
+	cases := []struct {
+		label                 string
+		foundDeviceClasses    []rookCephv1.DeviceClasses
+		isReplica1            bool
+		replica1DeviceClasses []string
+		expectedDeviceClass   string
+	}{
+		{
+			label:                 "Case 1: Replica 1 not enabled & no DeviceClass in status",
+			foundDeviceClasses:    []rookCephv1.DeviceClasses{},
+			isReplica1:            false,
+			replica1DeviceClasses: []string{},
+			expectedDeviceClass:   "ssd",
+		},
+		{
+			label: "Case 2: Replica 1 not enabled & 1 DeviceClass is in status",
+			foundDeviceClasses: []rookCephv1.DeviceClasses{
+				{
+					Name: "gold",
+				},
+			},
+			isReplica1:            false,
+			replica1DeviceClasses: []string{},
+			expectedDeviceClass:   "gold",
+		},
+		{
+			label: "Case 3: Replica 1 not enabled & more than 1 DeviceClass is in status",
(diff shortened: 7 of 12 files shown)


malayparida2000:
Test results

default cephblockpool spec section with the deviceclass SSD
```
spec:
  application: ""
  deviceClass: ssd
  enableRBDStats: true
  erasureCoded:
    codingChunks: 0
    dataChunks: 0
  failureDomain: zone
  mirroring: {}
  quotas: {}
  replicated:
    replicasPerFailureDomain: 1
    size: 3
    targetSizeRatio: 0.49
  statusCheck:
    mirror: {}
```
    
Ceph osd tree, as we can see default pool data is replicated amon default osd's and no data is spread to replica-1 osds.
```
sh-5.1$ ceph osd df tree
ID   CLASS       WEIGHT    REWEIGHT  SIZE    RAW USE  DATA     OMAP     META     AVAIL    %USE  VAR   PGS  STATUS  TYPE NAME                                           
 -1              12.00000         -  12 TiB   29 GiB   28 GiB    6 KiB  413 MiB   12 TiB  0.23  1.00    -          root default                                        
 -5              12.00000         -  12 TiB   29 GiB   28 GiB    6 KiB  413 MiB   12 TiB  0.23  1.00    -              region us-east-1                                
-10               4.00000         -   4 TiB  9.7 GiB  9.5 GiB    2 KiB  174 MiB  4.0 TiB  0.24  1.00    -                  zone us-east-1a                             
 -9               2.00000         -   2 TiB  9.6 GiB  9.5 GiB    1 KiB   91 MiB  2.0 TiB  0.47  1.99    -                      host ocs-deviceset-gp3-csi-2-data-0gwjwd
  1         ssd   2.00000   1.00000   2 TiB  9.6 GiB  9.5 GiB    1 KiB   91 MiB  2.0 TiB  0.47  1.99  102      up                  osd.1                               
-46               2.00000         -   2 TiB   86 MiB  2.4 MiB    1 KiB   83 MiB  2.0 TiB  0.00  0.02    -                      host us-east-1a-data-0n4xg8             
  4  us-east-1a   2.00000   1.00000   2 TiB   86 MiB  2.4 MiB    1 KiB   83 MiB  2.0 TiB  0.00  0.02   27      up                  osd.4                               
-14               4.00000         -   4 TiB  9.6 GiB  9.5 GiB    2 KiB  118 MiB  4.0 TiB  0.23  1.00    -                  zone us-east-1b                             
-13               2.00000         -   2 TiB  9.6 GiB  9.5 GiB    1 KiB   91 MiB  2.0 TiB  0.47  1.99    -                      host ocs-deviceset-gp3-csi-0-data-0ksc9p
  2         ssd   2.00000   1.00000   2 TiB  9.6 GiB  9.5 GiB    1 KiB   91 MiB  2.0 TiB  0.47  1.99  104      up                  osd.2                               
-41               2.00000         -   2 TiB   29 MiB  2.3 MiB    1 KiB   26 MiB  2.0 TiB  0.00  0.01    -                      host us-east-1b-data-0tdpl5             
  3  us-east-1b   2.00000   1.00000   2 TiB   29 MiB  2.3 MiB    1 KiB   26 MiB  2.0 TiB  0.00  0.01   25      up                  osd.3                               
 -4               4.00000         -   4 TiB  9.6 GiB  9.5 GiB    2 KiB  122 MiB  4.0 TiB  0.23  1.00    -                  zone us-east-1c                             
 -3               2.00000         -   2 TiB  9.6 GiB  9.5 GiB    1 KiB   91 MiB  2.0 TiB  0.47  1.99    -                      host ocs-deviceset-gp3-csi-1-data-0hrvml
  0         ssd   2.00000   1.00000   2 TiB  9.6 GiB  9.5 GiB    1 KiB   91 MiB  2.0 TiB  0.47  1.99  107      up                  osd.0                               
-51               2.00000         -   2 TiB   33 MiB  2.3 MiB    1 KiB   30 MiB  2.0 TiB  0.00  0.01    -                      host us-east-1c-data-0vssjc             
  5  us-east-1c   2.00000   1.00000   2 TiB   33 MiB  2.3 MiB    1 KiB   30 MiB  2.0 TiB  0.00  0.01   22      up                  osd.5                               
                              TOTAL  12 TiB   29 GiB   28 GiB  9.5 KiB  413 MiB   12 TiB  0.23                                                                         
MIN/MAX VAR: 0.01/1.99  STDDEV: 0.23
```

controllers/storagecluster/generate.go
```diff
@@ -45,10 +45,25 @@ func generateNameForCephBlockPool(initData *ocsv1.StorageCluster) string {
...
+	if len(foundDeviceClasses) == 0 {
+		if initData.Spec.ManagedResources.CephNonResilientPools.Enable {
+			return "replicated"
+		}
+		// if Non-Resilient pools are not enabled then keep the device class blank
+		return ""
+	} else if len(foundDeviceClasses) == 1 {
+		return foundDeviceClasses[0].Name
+	} else if len(foundDeviceClasses) == len(replica1DeviceClasses)+1 {
+		deviceClassMap := make(map[string]bool)
```
travisn, line 57: This naming will help for readability
```suggestion
		replica1DeviceClassMap := make(map[string]bool)
```

controllers/storagecluster/generate.go
```diff
@@ -45,10 +45,25 @@ func generateNameForCephBlockPool(initData *ocsv1.StorageCluster) string {
...
+		for _, deviceClass := range replica1DeviceClasses {
+			deviceClassMap[deviceClass.Name] = true
+		}
+		for _, deviceClass := range foundDeviceClasses {
```
travisn, line 61: How about adding a comment that clarifies we must not return a replica 1 device class?

controllers/storagecluster/generate.go
```diff
@@ -45,10 +45,25 @@ func generateNameForCephBlockPool(initData *ocsv1.StorageCluster) string {
 }
 
 func generateDeviceClass(initData *ocsv1.StorageCluster) string {
```
travisn, line 47: How about adding a unit test to cover this new logic?

travisn (changes requested):

controllers/storagecluster/generate.go
malayparida2000: Although nice but as this function involves runtime updates in the cephcluster status it would be a lot of error to write a unit test for this, as it's a small bug fix I would not prefer spending a considerable amount of effort just for adding the unit test for this.

travisn: Ok, I assumed it would be a small unit test for this one method.

malayparida2000: No, it actually depends on the populating of cephcluster status field with the deviceclasses.

controllers/storagecluster/generate.go
```diff
@@ -45,10 +45,26 @@ func generateNameForCephBlockPool(initData *ocsv1.StorageCluster) string {
...
-	if initData.Spec.ManagedResources.CephNonResilientPools.Enable {
-		return "replicated"
+	if len(foundDeviceClasses) == 0 {
```
iamniting, line 48: I don't think this condition will help us much, eventually, it will go in the next else if condition in the next reconcile once the status is set.

controllers/storagecluster/generate.go
```diff
@@ -45,10 +45,26 @@ func generateNameForCephBlockPool(initData *ocsv1.StorageCluster) string {
...
+		if initData.Spec.ManagedResources.CephNonResilientPools.Enable {
+			return "replicated"
+		}
+		// if Non-Resilient pools are not enabled then keep the device class blank
+		return ""
+	} else if len(foundDeviceClasses) == 1 {
```
iamniting, line 54: What if we have more than one replicated/normal device classes

iamniting (approved):

iamniting (changes requested):
Approved mistakenly

controllers/storagecluster/cephblockpools.go
```diff
@@ -16,6 +17,8 @@ import (
 
 type ocsCephBlockPools struct{}
 
+var replica1DeviceClasses []rookCephv1.DeviceClasses
```
travisn, line 20: Can we do this without a package var? A local var is preferred. For example, if there are multiple ceph clusters (one internal and one external cluster), there will be confusion about this package var being used for both.

malayparida2000: When the reconcile request comes for an external storagecluster we are not creating the pools in that case, so the concerned function generateDeviceClass() which would use these variables is never called so there won't be any confusion.

controllers/storagecluster/cephcluster.go
```diff
@@ -266,6 +266,14 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
 		return reconcile.Result{}, nil
 	}
 
+	if found.Status.CephStorage != nil {
+		cephDeviceClasses := []string{}
+		for _, deviceClass := range found.Status.CephStorage.DeviceClasses {
+			cephDeviceClasses = append(cephDeviceClasses, deviceClass.Name)
+		}
+		sc.Status.CephDeviceClasses = cephDeviceClasses
```
travisn, line 274: Why are the device classes duplicated in the StorageCluster CR status? Can the CephCluster status be used directly instead?

malayparida2000: Hm, actually there is no way to pass the cephcluster status to the generateDeviceClass function, So either we have to use a package level var to share the information or do this. Also keeping the status on the storagecluster status make it possible to write unit tests for this, so I thought of this approach.

travisn: Anytime a CRD spec or status exposes information it is considered API that users may consume. We don't want to put info in the status as an internal implementation detail if the user never needs to know it. In other words, let's avoid exposing unnecessary API information.

Something that may be useful for the user to be aware of, is which is the default device class name. So instead of including all device classes in this StorageCluster status to evaluate later during reconcile, how about we immediately evaluate the default storage class name and set it here? We have all the parameters needed at this point to determine the default storage class, right?

metrics/vendor/github.com/red-hat-storage/ocs-operator/api/v4/v1/storagecluster_types.go
```diff
@@ -569,6 +569,9 @@ type StorageClusterStatus struct {
 	// Images holds the image reconcile status for all images reconciled by the operator
 	Images ImagesStatus `json:"images,omitempty"`
 
+	// CephDeviceClasses holds the device classes available in the ceph cluster
+	CephDeviceClasses []string `json:"cephDeviceClasses,omitempty"`
```
travisn, line 573: Per previous comment, how about this?
```suggestion
	DefaultDeviceClass string `json:"defaultDeviceClass,omitempty"`
```

controllers/storagecluster/cephcluster.go
malayparida2000: Thanks for the suggestions, I have updated the PR based on this. Do have another look.

controllers/storagecluster/cephcluster.go
```diff
@@ -1341,3 +1345,31 @@ func getOsdCount(sc *ocsv1.StorageCluster) int {
...
+			replica1DeviceClassMap[deviceClass] = true
+		}
+		for _, deviceClass := range foundDeviceClasses {
+			if found := replica1DeviceClassMap[deviceClass.Name]; !found {
+				return deviceClass.Name
+			}
+		}
+	}
+	// If device classes are not found, return "replicated"
+	return "replicated"
```
travisn, line 1374: How about if we always default to `ssd`? We will only use `replicated` for existing clusters that find it above.
```suggestion
	return "ssd"
```

controllers/storagecluster/generate.go
```diff
@@ -45,11 +45,7 @@ func generateNameForCephBlockPool(initData *ocsv1.StorageCluster) string {
 }
 
 func generateDeviceClass(initData *ocsv1.StorageCluster) string {
```
travisn, line 47: Maybe no need for this helper method anymore if you an directly use `initData.Status.DefaultCephDeviceClass`, but keeping it is find too if you want

controllers/storagecluster/cephcluster.go
```diff
@@ -266,6 +266,10 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
 		return reconcile.Result{}, nil
 	}
 
+	if found.Status.CephStorage != nil {
+		sc.Status.DefaultCephDeviceClass = determineDefaultCephDeviceClass(found.Status.CephStorage.DeviceClasses, sc.Spec.ManagedResources.CephNonResilientPools.Enable, sc.Status.FailureDomainValues)
```
travisn, line 270: Are we sure the `sc.Status.FailureDomainValues` will already be set at this point? If manual testing looks good, then it must be.

controllers/storagecluster/cephcluster.go
```diff
@@ -1341,3 +1345,31 @@ func getOsdCount(sc *ocsv1.StorageCluster) int {
 	}
 	return osdCount
 }
+
+func determineDefaultCephDeviceClass(foundDeviceClasses []rookCephv1.DeviceClasses, isReplica1 bool, replica1DeviceClasses []string) string {
```
travisn, line 1349: `sc.Status.FailureDomainValues` only contains replica 1 device classes? Those don't seem like the same thing to me, but maybe it's just a naming issue.

controllers/storagecluster/cephcluster.go
malayparida2000: We can change it to ssd, but then we also need to remove this https://github.com/red-hat-storage/ocs-operator/blob/b18b234e14f6182c858e64bffd3e627b1a5350b3/controllers/storagecluster/cephcluster.go#L850-L853

controllers/storagecluster/generate.go
malayparida2000: Yeah let's just remove this func

controllers/storagecluster/cephcluster.go
malayparida2000: Yes see here https://github.com/red-hat-storage/ocs-operator/blob/b18b234e14f6182c858e64bffd3e627b1a5350b3/controllers/storagecluster/reconcile.go#L393.
The failuredomain and such thing are set first before we reach cephcluster reconcilation.

controllers/storagecluster/cephcluster.go
malayparida2000: Yes for each failure domain value we have created one osd with the deviceclass as the failuredomainvalue.

controllers/storagecluster/cephcluster.go
travisn: Ok good catch, seems like we anyway want to set that annotation to the DefaultDeviceClass.

malayparida2000: Modified

controllers/storagecluster/cephcluster.go
```diff
@@ -847,10 +852,6 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster) []rookCephv1.StorageCla
 			annotations := map[string]string{
 				"crushDeviceClass": crushDeviceClass,
```
travisn, line 853: In a previously-created cluster that set `replicated`, will it continue using `replicated` here? Or why not set the `DefaultDeviceClass` here, unless overridden by the user?

malayparida2000: Yes as once set the crushDeviceClass doesn't change so it won't have a problem. We shouldn't use the defaultDeviceClass as it kinda depends on the cephCluster CR status to determine that, and this crushDeviceClass annotation is decided before that, also if it's set it can't be changed further.

travisn (approved):
/approve
/lgtm

malayparida2000:
```
2024-06-10T07:34:24.303690686Z time="2024-06-10T07:34:24Z" level=info msg="creating bucket nb.1718003653158.makestoragegreatagain.com" sys=openshift-storage/noobaa
2024-06-10T07:34:24.669995832Z time="2024-06-10T07:34:24Z" level=error msg="got error when trying to create bucket nb.1718003653158.makestoragegreatagain.com. error: RequestError: send request failed\ncaused by: Put \"https://s3.us-east-2.amazonaws.com/nb.1718003653158.makestoragegreatagain.com\": tls: failed to verify certificate: x509: certificate signed by unknown authority" sys=openshift-storage/noobaa
2024-06-10T07:34:24.669995832Z time="2024-06-10T07:34:24Z" level=info msg="SetPhase: temporary error during phase \"Configuring\"" sys=openshift-storage/noobaa
2024-06-10T07:34:24.670049595Z time="2024-06-10T07:34:24Z" level=warning msg="⏳ Temporary Error: RequestError: send request failed\ncaused by: Put \"https://s3.us-east-2.amazonaws.com/nb.1718003653158.makestoragegreatagain.com\": tls: failed to verify certificate: x509: certificate signed by unknown authority" sys=openshift-storage/noobaa
2024-06-10T07:34:24.678836665Z time="2024-06-10T07:34:24Z" level=info msg="UpdateStatus: Done generation 1" sys=openshift-storage/noobaa
2024-06-10T07:34:24.679320223Z time="2024-06-10T07:34:24Z" level=info msg="Update event detected for noobaa (openshift-storage), queuing Reconcile"
```

The noobaa problem keeps failing the e2e test again n again

malayparida2000:
I did extensive testing for the PR with the below scenarios
Case 1: Cluster was created with old code, then upgraded to new code
Case 2: Cluster was created & replica 1 was enabled on the old code, then upgraded to new code
Case 3: Cluster was created with replica-1 with old code, then upgraded to new code
Case 4: Cluster gets created & replica-1 gets enabled both on new code
Case 5: Cluster is created from scratch with replica-1 with new code
here is the complete testing results https://hackmd.io/@Yh4a4hAATcW2BNYBJVSx4w/ryLe_4kHR

Happy to state that the code works for all the above cases & solves the bug.

iamniting (approved):

PR #2618, opened 2024-05-17
do not check for storageconsumer when storagerequest is being deleted

leelavg:
there is a possibility that getting storageconsumer from cache would work even when the resource is deleted however if the event gets requeued then we will always hit failure while trying to get deleted resource after a cache sync.

with this commit we only check presence of storageconsumer if storagerequest is not being deleted as the deletion of storagerequest is generally triggered via owner references and storageconsumer is owner of storagerequest.

fix for https://bugzilla.redhat.com/show_bug.cgi?id=2280813

### DIFF
--- a/controllers/storagerequest/storagerequest_controller.go
+++ b/controllers/storagerequest/storagerequest_controller.go
@@ -208,10 +208,6 @@ func (r *StorageRequestReconciler) initPhase() error {
 		return fmt.Errorf("no storage consumer owner ref on the storage class request")
 	}
 
-	if err := r.get(r.storageConsumer); err != nil {
-		return err
-	}
-
 	// check request status already contains the name of the resource. if not, add it.
 	if r.StorageRequest.Spec.Type == "block" {
 		// initialize in-memory structs
@@ -303,6 +299,12 @@ func (r *StorageRequestReconciler) reconcilePhases() (reconcile.Result, error) {
 				return reconcile.Result{}, fmt.Errorf("failed to add finalizer: %v", err)
 			}
 		}
+
+		// we are loading the storageconsumer only to confirm it's presence
+		if err := r.get(r.storageConsumer); err != nil {
+			return reconcile.Result{}, err
+		}
+
 		if r.StorageRequest.Spec.Type == "block" {
 
 			if err := r.reconcileCephClientRBDProvisioner(); err != nil {


leelavg:
An alternate implementation is not doing a get on storageconsumer from storagerequest controller at all, that also seemed clean but didn't find strong reasons to prefer one over other.

controllers/storagerequest/storagerequest_controller.go
```diff
@@ -208,7 +208,8 @@ func (r *StorageRequestReconciler) initPhase() error {
 		return fmt.Errorf("no storage consumer owner ref on the storage class request")
 	}
 
-	if err := r.get(r.storageConsumer); err != nil {
+	// unless storage request is marked for deletion we need to verify presence of storageconsumer
+	if err := r.get(r.storageConsumer); (errors.IsNotFound(err) && r.StorageRequest.DeletionTimestamp.IsZero()) || err != nil {
```
nb-ohad, line 212: The condition described by this in is very hard to understand, it conflates between 2 different cases, one where there is a non-not-found error and one where there is no error. 

There is even a subtle bug here where we can return a nil from this func without running the entire reconcile (where we should return an error) in the case that the timestamp is non-zero.

To add to the above, I think that ignoring a storage request that is in deletion is wrong as you might need the reconcile to happen fully in order to finalize the deletion process.  

I would suggest rethinking the approach or at least re-write the condition (maybe split it) in a way that it will be easier to understand and reevaluate

nb-ohad (changes requested):

controllers/storagerequest/storagerequest_controller.go
leelavg: > it conflates between 2 different cases
- my intention is only one statement. Get storageconsumer unless storagerequest is set for deletion

> we can return a nil from this func without running the entire reconcile 
- we are not doing it

> I think that ignoring a storage request that is in deletion is wrong
- I'm not doing that

> at least re-write the condition (maybe split it)
- ack, as mentioned in the commit msg, the cond is "return err if (consumer is not found when storagerequest exists) or (err not nil)"

As commented earlier, I think there is no need to get the storageconsumer in storagerequest controller.

leelavg: refreshed.

PR #2619, opened 2024-05-20
Fix storagecluster version in odfinfoconfig unit test

Nikhil-Ladha:
Update storagcluster version in odfinfoconfig test to 4.17

### DIFF
--- a/controllers/storagecluster/initialization_reconciler_test.go
+++ b/controllers/storagecluster/initialization_reconciler_test.go
@@ -3,12 +3,13 @@ package storagecluster
 import (
 	"context"
 	"fmt"
+	"os"
+	"testing"
+
 	"github.com/blang/semver/v4"
 	oprverion "github.com/operator-framework/api/pkg/lib/version"
 	opv1a1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	ocsversion "github.com/red-hat-storage/ocs-operator/v4/version"
-	"os"
-	"testing"
 
 	"github.com/imdario/mergo"
 	nbv1 "github.com/noobaa/noobaa-operator/v5/pkg/apis/noobaa/v1alpha1"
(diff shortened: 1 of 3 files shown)


iamniting (approved):

raaizik (approved):

raaizik:
Pasting relevant [comment](https://github.com/red-hat-storage/ocs-operator/pull/2440#discussion_r1606657764). I tend to agree as mentioned in my first comment this var's value is better off as a dummy.

malayparida2000:
The e2e is failing due to ocs-client-operator installation failing

Nikhil-Ladha:
> Pasting relevant [comment](https://github.com/red-hat-storage/ocs-operator/pull/2440#discussion_r1606657764). @Nikhil-Ladha, I tend to agree as as mentioned in my first comment this data is better off as a dummy.

Technically, it is not a dummy value. But, a hardcoded value that would change in each release.

raaizik:
> > Pasting relevant [comment](https://github.com/red-hat-storage/ocs-operator/pull/2440#discussion_r1606657764). @Nikhil-Ladha, I tend to agree as as mentioned in my first comment this data is better off as a dummy.
> 
> Technically, it is not a dummy value. But, a hardcoded value that would change in each release.

But IMO this type of data should be treated as a mock in tests and not in a way that stores its real value.

Nikhil-Ladha:
> > > Pasting relevant [comment](https://github.com/red-hat-storage/ocs-operator/pull/2440#discussion_r1606657764). @Nikhil-Ladha, I tend to agree as as mentioned in my first comment this data is better off as a dummy.
> > 
> > 
> > Technically, it is not a dummy value. But, a hardcoded value that would change in each release.
> 
> But IMO this type of data should be treated as a mock in tests and not in a way that stores its real value.

It should be, but in the part of code that this test is testing out, we can't set version to a mock value, since it is comparing the value with an actual storagecluster that is created in the test env.

PR #2620, opened 2024-05-20
Adds metadata to external ceph resources

raaizik:
## Changes
~~Adds metadata field to storage claim response proto message that contains a mirroring flag of `CephBlockPool` to be read directly from the `StorageClaim`'s controller which will in turn label the `StorageClass` with the required Ramen labels that allow RBD mirroring.~~

Adds metadata in the form of labels and annotations to the ceph external resource struct. 

[RHSTOR-5754](https://issues.redhat.com/browse/RHSTOR-5754)


services/provider/server/server.go
```diff
@@ -593,6 +593,12 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
 				return nil, status.Errorf(codes.Internal, "failed to get %s CephBlockPoolRadosNamespace. %v", cephRes.Name, err)
 			}
 
+			cbp := &rookCephv1.CephBlockPool{}
+			err = s.client.Get(ctx, types.NamespacedName{Name: cephRes.Name, Namespace: s.namespace}, rns)
+			if err != nil {
+				return nil, status.Errorf(codes.Internal, "failed to get %s CephBlockPool. %v", cephRes.Name, err)
+			}
+
```
rewantsoni, line 601: Not sure what you are trying to achieve here, you are creating a cpb var but you are getting the radosnamespace. 
Also, we should not depend on one cephBlockPool to decide if we are enabling mirroring or not. There might be a case where we might need to blacklist some blockpools for mirroring.

raaizik: No, I'm tagging the CBPRN class with a flag that will signal to StorageClaim's controller whether mirroring's enabled for said class. 

> There might be a case where we might need to blacklist some blockpools for mirroring.

Care to elaborate? What indicates a blacklisted pool?

rewantsoni:
```
err = s.client.Get(ctx, types.NamespacedName{Name: cephRes.Name, Namespace: s.namespace}, rns)
```
You are getting the rns here, not the cbl, is this intended?

services/provider/server/server.go
```diff
@@ -611,10 +617,14 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
 				rbdStorageClassData["encryptionKMSID"] = storageRequest.Spec.EncryptionMethod
 			}
 
+			rbdStorageClassMetaData := map[string]string{
+				"mirroring": fmt.Sprintf("%v", cbp.Spec.Mirroring.Enabled),
+			}
```
rewantsoni, line 622: If I am not wrong this will always be false, as you are haven't fetched the cbp

raaizik: > You are getting the rns here, not the cbl, is this intended?

CBPRN and it's respective CPB carry the same name.

services/provider/server/server.go
raaizik: Discussed over DMs:

- CBP is the only source to fetch the mirroring flag from.
- In case of blacklisted pools we can have their flag indicating that override the mirroring flag.

rewantsoni:
> ## Changes
> Adds metadata field to storage claim response proto message that contains a mirroring flag of `CephBlockPool` to be read directly from the `StorageClaim`'s controller which will in turn label the `StorageClass` with the required Ramen labels that allow RBD mirroring.
> 
> [RHSTOR-5754](https://issues.redhat.com/browse/RHSTOR-5754)

I went through the PR and this works in the case of storageClaim of type block, if I am not wrong we also need to label the storageClass created by storageClaim of type filesystem as well right? How are we handling that?

raaizik:
> > ## Changes
> > Adds metadata field to storage claim response proto message that contains a mirroring flag of `CephBlockPool` to be read directly from the `StorageClaim`'s controller which will in turn label the `StorageClass` with the required Ramen labels that allow RBD mirroring.
> > [RHSTOR-5754](https://issues.redhat.com/browse/RHSTOR-5754)
> 
> I went through the PR and this works in the case of storageClaim of type block, if I am not wrong we also need to label the storageClass created by storageClaim of type filesystem as well right? How are we handling that?

For `Volume*SnapshotClass`, the SID can be applied for backed classes of **_both_** types. 
For `Volume*ReplicationClass`, the RepID can only be applied when there are two peered EPs. Can add a Replication flag (that I'd need to figure out how to populate) along with the Mirroring flag. @rewantsoni

services/provider/proto/provider.proto
```diff
@@ -153,6 +153,8 @@ message StorageClaimConfigRequest{
 message StorageClaimConfigResponse{
     // ExternalResource holds the configuration data of external storage cluster
     repeated ExternalResource externalResource = 1;
+    // MetaData contains additional info regarding the external cluster resources
+    bytes MetaData = 2;
```
leelavg, line 157: no need to Title case the field as generated proto already gives us an exported field.

leelavg, line 157: after going through PR, why not `bool enableMirroring = 2;`? why do we need to convert bool (`cbp.Spec.Mirroring.Enabled`) into string (`%v`) and then encode to bytes (`json.Marshal`)

The default proto value will be `false` anyways, unless `false` is also a valid value like after enabling you could disable we don't need to use bool pointer.

raaizik: See first link in the description of the attached Jira for design details, @leelavg. We're looking at a feature that would allow us to add _multiple_ fields that are considered as metadata. Once this PR's WIP tag is removed, there will be two flags appended to the metadata field.

raaizik: Could be. I will take this into account.

nb-ohad (changes requested):

rewantsoni:
LGTM

services/provider/proto/provider.proto
```diff
@@ -73,6 +73,10 @@ message ExternalResource {
     string kind = 2;
     // Data contains the contents of the external cluster resource
     bytes Data = 3;
+    // Labels contains the labels of the external cluster resource
+    bytes Labels = 4;
+    // Annotations contains the annotations of the external cluster resource
+    bytes Annotations = 5;
```
nb-ohad, line 79: Why bytes? Labels and annotations have a known structure of a string key-value pair list.
Let do proper encoding for it

nb-ohad (changes requested):

nb-ohad:
LGTM, 
@leelavg, as our protobuf expert, can you please also take a look?

nb-ohad:
> Sooner than later we might want to move proto definitions and client svc into it's own go module as I see huge deps ending up in client-op in each update, wdyt @nb-ohad?

That might or might not be a future target. But that does not directly relate to this change. So can we merge?

PR #2621, opened 2024-05-21
controllers: limit reconciliations to relevant events

iamniting:
Noobaa and CephCluster CRs' status.conditions[].lastHeartbeatTime and status.conditions[].lastTransitionTime etc change frequently, causing unnecessary StorageCluster reconciliations. Implement logic to filter out these frequent, non-essential changes and trigger reconciliations only on relevant events.

Also add the owns for other objects as noobaa and cephcluster will trigger the recociles only for the relevant events. Previously, even though these objects were not owned by the controller, reconciliations were triggered due to status updates of the NooBaa and CephCluster CRs. Now the other objects should send thier own recocile requests because noobaa and ceph clkuster wont trigger the unnecessary reconciles and these objects can not rely on noobaa and ceph cluster.

Co-authored-by: Malay Kumar Parida <mparida@redhat.com>

Bug: https://bugzilla.redhat.com/show_bug.cgi?id=2280595

We are required to backport this till 4.14

### DIFF
--- a/controllers/ocsinitialization/ocsinitialization_controller.go
+++ b/controllers/ocsinitialization/ocsinitialization_controller.go
@@ -253,7 +253,7 @@ func (r *OCSInitializationReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		},
 	)
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&ocsv1.OCSInitialization{}).
+		For(&ocsv1.OCSInitialization{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
 		Owns(&corev1.Service{}).
 		Owns(&corev1.Secret{}).
 		Owns(&promv1.Prometheus{}).
@@ -271,6 +271,7 @@ func (r *OCSInitializationReconciler) SetupWithManager(mgr ctrl.Manager) error {
 					}}
 				},
 			),
+			builder.WithPredicates(predicate.GenerationChangedPredicate{}),
 		).
 		// Watcher for storageClass required to update values related to replica-1
 		// in ocs-operator-config configmap, if storageClass changes
--- a/controllers/storagecluster/storagecluster_controller.go
+++ b/controllers/storagecluster/storagecluster_controller.go
@@ -7,7 +7,11 @@ import (
 	"reflect"
 
 	"github.com/go-logr/logr"
+	"github.com/google/go-cmp/cmp"
+	volumesnapshotv1 "github.com/kubernetes-csi/external-snapshotter/client/v7/apis/volumesnapshot/v1"
 	nbv1 "github.com/noobaa/noobaa-operator/v5/pkg/apis/noobaa/v1alpha1"
+	routev1 "github.com/openshift/api/route/v1"
+	templatev1 "github.com/openshift/api/template/v1"
 	conditionsv1 "github.com/openshift/custom-resource-status/conditions/v1"
 	"github.com/operator-framework/operator-lib/conditions"
 	ocsclientv1a1 "github.com/red-hat-storage/ocs-client-operator/api/v1alpha1"
@@ -17,6 +21,7 @@ import (
 	cephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	storagev1 "k8s.io/api/storage/v1"
 	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -144,14 +149,81 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		},
 	)
 
-	builder := ctrl.NewControllerManagedBy(mgr).
+	noobaaIgnoreTimeUpdatePredicate := predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			if e.ObjectOld == nil || e.ObjectNew == nil {
+				return false
+			}
+			oldObj := e.ObjectOld.(*nbv1.NooBaa)
+			newObj := e.ObjectNew.(*nbv1.NooBaa)
+
+			ignorePaths := func(path cmp.Path) bool {
+				switch path.String() {
+				case "ObjectMeta.ManagedFields",
+					"ObjectMeta.ResourceVersion",
+					"Status.Conditions.LastHeartbeatTime",
+					"Status.Conditions.LastTransitionTime":
+					return true
+				}
+				return false
+			}
+			diff := cmp.Diff(
+				oldObj, newObj,
+				cmp.FilterPath(ignorePaths, cmp.Ignore()),
+			)
+
+			return diff != ""
+		},
+	}
+
+	cephClusterIgnoreTimeUpdatePredicate := predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			if e.ObjectOld == nil || e.ObjectNew == nil {
+				return false
+			}
+			oldObj := e.ObjectOld.(*cephv1.CephCluster)
+			newObj := e.ObjectNew.(*cephv1.CephCluster)
+
+			ignorePaths := func(path cmp.Path) bool {
+				switch path.String() {
+				case "ObjectMeta.ManagedFields",
+					"ObjectMeta.ResourceVersion",
+					"Status.CephStatus.LastChecked",
+					"Status.CephStatus.Capacity.LastUpdated",
+					"Status.Conditions.LastHeartbeatTime",
+					"Status.Conditions.LastTransitionTime":
+					return true
+				}
+				return false
+			}
+			diff := cmp.Diff(
+				oldObj, newObj,
+				cmp.FilterPath(ignorePaths, cmp.Ignore()),
+			)
+
+			return diff != ""
+		},
+	}
+
+	build := ctrl.NewControllerManagedBy(mgr).
 		For(&ocsv1.StorageCluster{}, builder.WithPredicates(scPredicate)).
-		Owns(&cephv1.CephCluster{}).
+		Owns(&cephv1.CephCluster{}, builder.WithPredicates(cephClusterIgnoreTimeUpdatePredicate)).
+		Owns(&cephv1.CephBlockPool{}).
+		Owns(&cephv1.CephFilesystem{}).
+		Owns(&cephv1.CephFilesystemSubVolumeGroup{}).
+		Owns(&cephv1.CephNFS{}).
+		Owns(&cephv1.CephObjectStore{}).
+		Owns(&cephv1.CephObjectStoreUser{}).
+		Owns(&cephv1.CephRBDMirror{}).
 		Owns(&corev1.PersistentVolumeClaim{}, builder.WithPredicates(pvcPredicate)).
 		Owns(&appsv1.Deployment{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
 		Owns(&corev1.Service{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
 		Owns(&corev1.ConfigMap{}, builder.MatchEveryOwner, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
 		Owns(&corev1.Secret{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		Owns(&routev1.Route{}).
+		Owns(&templatev1.Template{}).
+		Watches(&storagev1.StorageClass{}, enqueueStorageClusterRequest).
+		Watches(&volumesnapshotv1.VolumeSnapshotClass{}, enqueueStorageClusterRequest).
 		Watches(&ocsclientv1a1.StorageClient{}, enqueueStorageClusterRequest).
 		Watches(&ocsv1.StorageProfile{}, enqueueStorageClusterRequest).
 		Watches(
@@ -163,9 +235,10 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			enqueueStorageClusterRequest,
 		).
 		Watches(&ocsv1alpha1.StorageConsumer{}, enqueueStorageClusterRequest, builder.WithPredicates(storageConsumerStatusPredicate))
+
 	if os.Getenv("SKIP_NOOBAA_CRD_WATCH") != "true" {
-		builder.Owns(&nbv1.NooBaa{})
+		build.Owns(&nbv1.NooBaa{}, builder.WithPredicates(noobaaIgnoreTimeUpdatePredicate))
 	}
 
-	return builder.Complete(r)
+	return build.Complete(r)
 }
--- a/go.mod
+++ b/go.mod
@@ -12,6 +12,7 @@ require (
 	github.com/ceph/ceph-csi/api v0.0.0-20240322131550-063319f6e516
 	github.com/ghodss/yaml v1.0.1-0.20220118164431-d8423dcdf344
 	github.com/go-logr/logr v1.4.1
+	github.com/google/go-cmp v0.6.0
 	github.com/google/uuid v1.6.0
 	github.com/imdario/mergo v0.3.16
 	github.com/k8snetworkplumbingwg/network-attachment-definition-client v1.6.0
@@ -78,7 +79,6 @@ require (
 	github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da // indirect
 	github.com/golang/protobuf v1.5.4 // indirect
 	github.com/google/gnostic-models v0.6.8 // indirect
-	github.com/google/go-cmp v0.6.0 // indirect
 	github.com/google/gofuzz v1.2.0 // indirect
 	github.com/google/pprof v0.0.0-20230510103437-eeec1cb781c3 // indirect
 	github.com/hashicorp/errwrap v1.1.0 // indirect


controllers/storagecluster/storagecluster_controller.go
```diff
@@ -144,28 +148,91 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+			}
+			oldObj := e.ObjectOld.(*nbv1.NooBaa)
+			newObj := e.ObjectNew.(*nbv1.NooBaa)
+
+			for _, cond := range oldObj.Status.Conditions {
+				cond.LastHeartbeatTime = metav1.Time{}
+				cond.LastTransitionTime = metav1.Time{}
+			}
+
+			for _, cond := range oldObj.Status.Conditions {
```
parth-gr, line 164: instead of using a `for loop` we do have a better and faster comparison using resourceQtycompare,

cmpopts is package provide by go here can use it https://pkg.go.dev/github.com/google/go-cmp/cmp/cmpopts#IgnoreFields

```
resourceQtyComparer = cmpopts.IgnoreFields(corev1.NodeCondition{}, "LastHeartbeatTime")
	diff1 := cmp.Diff(objOld.Spec, objNew.Spec, resourceQtyComparer)
```

For more details refer to https://github.com/rook/rook/pull/14186, we have just included the similar checks for the node reconcile

parth-gr (changes requested):

controllers/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -253,13 +253,13 @@ func (r *OCSInitializationReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-		Owns(&corev1.ConfigMap{}).
-		Owns(&promv1.Alertmanager{}).
-		Owns(&promv1.ServiceMonitor{}).
+		For(&ocsv1.OCSInitialization{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		Owns(&corev1.Service{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		Owns(&corev1.Secret{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		Owns(&promv1.Prometheus{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		Owns(&corev1.ConfigMap{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		Owns(&promv1.Alertmanager{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		Owns(&promv1.ServiceMonitor{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
```
nb-ohad, line 262: @iamniting I am not sure generation will change for metadata changes, like labels and annotation (I am quite sure the opposite is correct)

nb-ohad (changes requested):

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -144,28 +148,91 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		Watches(&ocsclientv1a1.StorageClient{}, enqueueStorageClusterRequest, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		Watches(&ocsv1.StorageProfile{}, enqueueStorageClusterRequest, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
 		Watches(
 			&extv1.CustomResourceDefinition{
 				ObjectMeta: metav1.ObjectMeta{
 					Name: "virtualmachines.kubevirt.io",
 				},
 			},
 			enqueueStorageClusterRequest,
+			builder.WithPredicates(predicate.GenerationChangedPredicate{}),
```
nb-ohad, line 229: @iamniting I am not sure generation will change for metadata changes, like labels and annotation (I am quite sure the opposite is correct)

nb-ohad (changes requested):

controllers/storagecluster/storagecluster_controller.go
iamniting: Done

controllers/ocsinitialization/ocsinitialization_controller.go
iamniting:
```
// GenerationChangedPredicate implements a default update predicate function on Generation change.                                                                                            
//                                                                              
// This predicate will skip update events that have no change in the object's metadata.generation field.
// The metadata.generation field of an object is incremented by the API server when writes are made to the spec field of an object.
// This allows a controller to ignore update events where the spec is unchanged, and only the metadata and/or status fields are changed.
//                                                                              
// For CustomResource objects the Generation is only incremented when the status subresource is enabled.
//                                                                              
// Caveats:                                                                     
//                                                                              
// * The assumption that the Generation is incremented only on writing to the spec does not hold for all APIs.
// E.g For Deployment objects the Generation is also incremented on writes to the metadata.annotations field.
// For object types other than CustomResources be sure to verify which fields will trigger a Generation increment when they are written to.
//                                                                              
// * With this predicate, any update events with writes only to the status field will not be reconciled.
// So in the event that the status block is overwritten or wiped by someone else the controller will not self-correct to restore the correct status.
```

This is a comment from the pkg.

nb-ohad: Yes, the above proves my point. 
``` This predicate will skip update events that have no change in the object's metadata.generation field.
// The metadata.generation field of an object is incremented by the API server when writes are made to the spec field of an object.
// This allows a controller to ignore update events where the spec is unchanged, and only the metadata and/or status fields are changed.```

Comment (iamniting) [reply controllers/ocsinitialization/ocsinitialization_controller.go:262]: Why do we care about the status and metadata of all the objects?

We are worried about the ceph cluster and noobaa status and for those, I have not added GenerationChangedPredicate.

File: controllers/ocsinitialization/ocsinitialization_controller.go [hunk 4]
```diff
@@ -253,13 +253,13 @@ func (r *OCSInitializationReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 	)
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&ocsv1.OCSInitialization{}).
-		Owns(&corev1.Service{}).
-		Owns(&corev1.Secret{}).
-		Owns(&promv1.Prometheus{}).
-		Owns(&corev1.ConfigMap{}).
-		Owns(&promv1.Alertmanager{}).
-		Owns(&promv1.ServiceMonitor{}).
+		For(&ocsv1.OCSInitialization{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
```
leelavg, line 256: this controller is for ocsinit, I don't think this should be filtered via predicate.

controllers/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -253,13 +253,13 @@ func (r *OCSInitializationReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		Owns(&corev1.Service{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
```
leelavg, line 257: service probably has some interested annotations like a serving cert etc for which if removed we want to reconcile, basically I believe we can't add this predicate w/o checking how the dependents are being used in the controller code.

controllers/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -288,6 +289,7 @@ func (r *OCSInitializationReconciler) SetupWithManager(mgr ctrl.Manager) error {
 					return []reconcile.Request{}
 				},
 			),
+			builder.WithPredicates(predicate.GenerationChangedPredicate{}),
```
leelavg, line 292: I don't believe sc has generation field.

controllers/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -331,7 +335,7 @@ func (r *OCSInitializationReconciler) SetupWithManager(mgr ctrl.Manager) error {
 					}}
 				},
 			),
-			builder.WithPredicates(prometheusPredicate),
+			builder.WithPredicates(predicate.GenerationChangedPredicate{}, prometheusPredicate),
```
leelavg, line 338: if possible, let's try to be consistent, could you pls move the predicate to end like being done for all other watches?

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -7,7 +7,12 @@ import (
 	"reflect"
 
 	"github.com/go-logr/logr"
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
```
leelavg, line 11: was tidy already run? I don't see these mod jumping to `require` section w/o `indirect` flag.

parth-gr, line 11: fix the action `cannot find module providing package github.com/google/go-cmp/cmp/cmpopts: import lookup disabled by -mod=vendor`

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -144,28 +150,77 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	}
+
+	cephClusterIgnoreTimeUpdatePredicate := predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			if e.ObjectOld == nil || e.ObjectNew == nil {
+				return false
+			}
+			oldObj := e.ObjectOld.(*cephv1.CephCluster)
+			newObj := e.ObjectNew.(*cephv1.CephCluster)
+
```
parth-gr, line 178: have you tested passing multiple `IgnoreFields` at the same time?

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -144,28 +150,77 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+				cmpopts.IgnoreFields(conditionsv1.Condition{}, "LastHeartbeatTime", "LastTransitionTime"),
+			)
+
+			return diff != ""
+		},
+	}
+
+	build := ctrl.NewControllerManagedBy(mgr).
 		For(&ocsv1.StorageCluster{}, builder.WithPredicates(scPredicate)).
-		Owns(&cephv1.CephCluster{}).
```
leelavg, line 149: are we sure that no component under storagecluster reconcilers are interested label/annotation changes?

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -144,28 +150,77 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+			if e.ObjectOld == nil || e.ObjectNew == nil {
+				return false
+			}
+			oldObj := e.ObjectOld.(*nbv1.NooBaa)
+			newObj := e.ObjectNew.(*nbv1.NooBaa)
+
+			diff := cmp.Diff(
+				oldObj, newObj,
+				cmpopts.IgnoreFields(metav1.ObjectMeta{}, "ResourceVersion"),
+				cmpopts.IgnoreFields(conditionsv1.Condition{}, "LastHeartbeatTime", "LastTransitionTime"),
```
leelavg, line 164: quick check, are ignored fields directly part of Condition struct? does this `cmpopts` recurse into sub types?

leelavg:
Small suggestion, can we only fix the known entities that are queuing reconciles rather than including that with a blanket change? Rationale is, in current context, it helps a lot fixing the main problem and only debug side effects directly related to that (if we hit any)?

controllers/storagecluster/storagecluster_controller.go
iamniting: Done

controllers/storagecluster/storagecluster_controller.go
iamniting: Yes i did

controllers/ocsinitialization/ocsinitialization_controller.go
iamniting: It should be otherwise it will trigger 2 reconcile for the same thing, one for the actual change and one for the status update in that reconcile.

controllers/storagecluster/storagecluster_controller.go
iamniting: Just did that

controllers/storagecluster/storagecluster_controller.go
iamniting: Yes i have tested it, It works as expected.

controllers/ocsinitialization/ocsinitialization_controller.go
iamniting: removed it.

controllers/ocsinitialization/ocsinitialization_controller.go
leelavg: unless ocsinit status has timestamps or not ordered arrays among other things, a single extra reconcile is fine as status for a desired state is almost constant.

controllers/ocsinitialization/ocsinitialization_controller.go
iamniting: removed it.

controllers/ocsinitialization/ocsinitialization_controller.go
iamniting: removed it.

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -144,14 +150,61 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+			)
+
+			return diff != ""
+		},
+	}
+
+	build := ctrl.NewControllerManagedBy(mgr).
```
leelavg, line 189: what's the intention of new `Owns` in this PR? is `storagecluster` already a owner of these resources?

controllers/storagecluster/storagecluster_controller.go
iamniting: You can try it out.

```
package main

import (
	"fmt"

	"github.com/google/go-cmp/cmp"
	"github.com/google/go-cmp/cmp/cmpopts"
	conditionsv1 "github.com/openshift/custom-resource-status/conditions/v1"
)

type NooBaa struct {
	Spec              NoobaaSpec   `json:"spec,omitempty"`
	Status            NooBaaStatus `json:"status,omitempty"`
}

type NoobaaSpec struct {
}

type NooBaaStatus struct {
	Conditions []conditionsv1.Condition `json:"conditions,omitempty"  patchStrategy:"merge" patchMergeKey:"type"`
}

func main() {
	n1 := &NooBaa{
			Name:            "n1",
			ResourceVersion: "1",
		},
		Status: NooBaaStatus{
			Conditions: []conditionsv1.Condition{
				{
					LastHeartbeatTime:  metav1.Now(),
					LastTransitionTime: metav1.Now(),
					Message:            "Hello1",
				},
			},
		},
	}

	n2 := &NooBaa{
			Name:            "n2",
			ResourceVersion: "2",
		},
		Status: NooBaaStatus{
			Conditions: []conditionsv1.Condition{
				{
					LastHeartbeatTime:  metav1.Now(),
					LastTransitionTime: metav1.Now(),
					Message:            "Hello2",
				},
			},
		},
	}

	diff := cmp.Diff(
		n1, n2,
		cmpopts.IgnoreFields(metav1.ObjectMeta{}, "ResourceVersion"),
		cmpopts.IgnoreFields(conditionsv1.Condition{}, "LastHeartbeatTime", "LastTransitionTime"),
	)
	if diff != "" {
		fmt.Println(diff)
	}
}

```

controllers/storagecluster/storagecluster_controller.go
iamniting: The storagecluster is already a owner of these resources.

Previously, any changes to the NooBaa and Ceph clusters would trigger the reconciliation process. This is no longer happening. Before, even if other custom resources (CRs) were changed, the reconciliation would still run because of the NooBaa and Ceph clusters. Now, we need to ensure that changes to other CRs still trigger the reconciliation process.

controllers/ocsinitialization/ocsinitialization_controller.go
iamniting: Here is a yaml of it.
```apiVersion: v1
items:
  kind: OCSInitialization
  metadata:
    creationTimestamp: "2024-05-21T14:12:19Z"
    generation: 1
    name: ocsinit
    namespace: openshift-storage
    resourceVersion: "157119"
    uid: 4537ce40-5856-4418-9599-c75987b71511
  spec: {}
  status:
    conditions:
    - lastHeartbeatTime: "2024-05-21T14:26:07Z"
      lastTransitionTime: "2024-05-21T14:12:19Z"
      message: Reconcile completed successfully
      reason: ReconcileCompleted
      status: "True"
      type: ReconcileComplete
    - lastHeartbeatTime: "2024-05-21T14:26:07Z"
      lastTransitionTime: "2024-05-21T14:12:19Z"
      message: Reconcile completed successfully
      reason: ReconcileCompleted
      status: "True"
      type: Available
    - lastHeartbeatTime: "2024-05-21T14:26:07Z"
      lastTransitionTime: "2024-05-21T14:12:19Z"
      message: Reconcile completed successfully
      reason: ReconcileCompleted
      status: "False"
      type: Progressing
    - lastHeartbeatTime: "2024-05-21T14:26:07Z"
      lastTransitionTime: "2024-05-21T14:12:19Z"
      message: Reconcile completed successfully
      reason: ReconcileCompleted
      status: "False"
      type: Degraded
    - lastHeartbeatTime: "2024-05-21T14:26:07Z"
      lastTransitionTime: "2024-05-21T14:12:19Z"
      message: Reconcile completed successfully
      reason: ReconcileCompleted
      status: "True"
      type: Upgradeable
    phase: Ready
    rookCephOperatorConfig: {}
    sCCsCreated: true
kind: List
metadata:
  resourceVersion: ""
```

controllers/ocsinitialization/ocsinitialization_controller.go
nb-ohad: Because there are resources where our reconciliation adds labels/annotations and we want an event whenever these are touched. Trying to track which one we do and which we don't, using a different code path (using the annotation vs not using the annotation) will fail very fast when any other developer adds an annotation to an existing resource (as part of some current or future change)  but will not know that they need to change the Watch!

controllers/storagecluster/storagecluster_controller.go
malayparida2000: The generation change predicate been removed

controllers/ocsinitialization/ocsinitialization_controller.go
iamniting: I removed `GenerationChangedPredicate`, so we should be good now.

controllers/storagecluster/storagecluster_controller.go
iamniting: removed it.

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -144,14 +150,63 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+			if e.ObjectOld == nil || e.ObjectNew == nil {
+				return false
+			}
+			oldObj := e.ObjectOld.(*nbv1.NooBaa)
+			newObj := e.ObjectNew.(*nbv1.NooBaa)
+
+			diff := cmp.Diff(
+				oldObj, newObj,
+				cmpopts.IgnoreFields(metav1.ObjectMeta{}, "ManagedFields", "ResourceVersion"),
+				cmpopts.IgnoreFields(conditionsv1.Condition{}, "LastHeartbeatTime", "LastTransitionTime"),
```
nb-ohad, line 164: @iamniting is there a way to specify the exact field path to ignore? for example `.status.LastHeartbeatTime`, Right now this code might ignore fields we don't want to ignore in cases where we will introduce fields with similar names

iamniting: There is no way to ignore the Path, But there are fewer chances of that as we are passing the exact struct where we would like to ignore the field.

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -144,14 +150,63 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+
+	cephClusterIgnoreTimeUpdatePredicate := predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			if e.ObjectOld == nil || e.ObjectNew == nil {
+				return false
+			}
+			oldObj := e.ObjectOld.(*cephv1.CephCluster)
+			newObj := e.ObjectNew.(*cephv1.CephCluster)
+
+			diff := cmp.Diff(
```
travisn, line 179: Why does OCS operator need to reconcile if anything changed in the CephCluster status? Normally I would expect reconcile is only needed when the `spec` is updated. Otherwise we're going to be chasing exceptions in this list every time another cephcluster status is updated too frequently.

travisn: And did something change in OCS operator recently about this? Or the reconciles weren't noticed as much previously?

controllers/storagecluster/storagecluster_controller.go
iamniting: We can do that found a way to do so.

controllers/storagecluster/storagecluster_controller.go
iamniting: We are updating the storage cluster status based on the ceph cluster status, that's why we are watching for the status.

controllers/storagecluster/storagecluster_controller.go
iamniting: Done

controllers/storagecluster/storagecluster_controller.go
iamniting: Things are not changed in the ocs. We got an customer case where ocs-operator is consuming around 107 gb of space for logs in there loki bucket.

travisn: got it thanks

controllers/ocsinitialization/ocsinitialization_controller.go
leelavg: > unless ocsinit status has timestamps
- based on above, it has, can be closed.

controllers/storagecluster/storagecluster_controller.go
leelavg: yes, can be resolved.

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -144,14 +149,76 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+				oldObj, newObj,
+				cmp.FilterPath(ignorePaths, cmp.Ignore()),
+			)
+
+			return diff != ""
+		},
+	}
+
+	build := ctrl.NewControllerManagedBy(mgr).
```
leelavg, line 203: is `builder` conflicting w/ existing names? why to rename to `build`?

malayparida2000: Yes the package name is builder already so while adding the predicates there is a conflict here 
https://github.com/red-hat-storage/ocs-operator/pull/2621/files#diff-12ce985080537239d98b7ebf7e0b97bb5382604bf08ebc9208ac5e0fed055533R235

leelavg:
not able to find the log in successful run to verify it's working or not, based on a failed run I still see reconciles getting logged every minute in the last ~15/119 events (basically once stabilized I guess)

log is from https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/red-hat-storage_ocs-operator/2621/pull-ci-red-hat-storage-ocs-operator-main-ocs-operator-bundle-e2e-aws/1793114562673774592/artifacts/ocs-operator-bundle-e2e-aws/gather-extra/artifacts/pods/openshift-storage_ocs-operator-f6fd7455d-jxpkb_ocs-operator.log

```
[root@forge tmp]# grep 'Reconciling StorageCluster' ocs.log | jq -r .ts | tail -n20
2024-05-22T03:51:39Z
2024-05-22T03:51:41Z
2024-05-22T03:51:42Z
2024-05-22T03:52:30Z
2024-05-22T03:53:30Z
2024-05-22T03:54:31Z
2024-05-22T03:55:32Z
2024-05-22T03:56:33Z
2024-05-22T03:57:34Z
2024-05-22T03:58:35Z
2024-05-22T03:59:36Z
2024-05-22T04:00:37Z
2024-05-22T04:01:38Z
2024-05-22T04:02:39Z
2024-05-22T04:03:40Z
2024-05-22T04:04:41Z
2024-05-22T04:05:42Z
2024-05-22T04:06:43Z
2024-05-22T04:07:44Z
2024-05-22T04:08:45Z
[root@forge tmp]# grep 'Reconciling StorageCluster' ocs.log | jq -r .ts | wc -l
119
```

iamniting:
> not able to find the log in successful run to verify it's working or not, based on a failed run I still see reconciles getting logged every minute in the last ~15/119 events (basically once stabilized I guess, from 3:52 in below log)
> 
> log is from https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/red-hat-storage_ocs-operator/2621/pull-ci-red-hat-storage-ocs-operator-main-ocs-operator-bundle-e2e-aws/1793114562673774592/artifacts/ocs-operator-bundle-e2e-aws/gather-extra/artifacts/pods/openshift-storage_ocs-operator-f6fd7455d-jxpkb_ocs-operator.log
> 
> ```
> [root@forge tmp]# grep 'Reconciling StorageCluster' ocs.log | jq -r .ts | tail -n20
> 2024-05-22T03:51:39Z
> 2024-05-22T03:51:41Z
> 2024-05-22T03:51:42Z
> 2024-05-22T03:52:30Z
> 2024-05-22T03:53:30Z
> 2024-05-22T03:54:31Z
> 2024-05-22T03:55:32Z
> 2024-05-22T03:56:33Z
> 2024-05-22T03:57:34Z
> 2024-05-22T03:58:35Z
> 2024-05-22T03:59:36Z
> 2024-05-22T04:00:37Z
> 2024-05-22T04:01:38Z
> 2024-05-22T04:02:39Z
> 2024-05-22T04:03:40Z
> 2024-05-22T04:04:41Z
> 2024-05-22T04:05:42Z
> 2024-05-22T04:06:43Z
> 2024-05-22T04:07:44Z
> 2024-05-22T04:08:45Z
> [root@forge tmp]# grep 'Reconciling StorageCluster' ocs.log | jq -r .ts | wc -l
> 119
> ```

They are every minute due to the cephcluster being updated its capacity field every minute. Earlier they were very frequent.

leelavg:
> They are every minute due to the cephcluster being updated its capacity field every minute. Earlier they were very frequent.
- Ack, unfortunately the worst case that I was aware of is reconciling every minute 😓.
- As the fix can't be realised easily, is there anyway to quantify `very frequent` and provide some kind of metric to take in this PR? Maybe, perform above filtering on other failed PRs but run similarly in CI to compare reconciles. (iirc, controller-runtime itself exports a metric on how many times it requeued events) 

Nevertheless, if what I'm asking makes less sense, I could standby for others to approve PR.

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -144,14 +149,76 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+				return false
+			}
+			oldObj := e.ObjectOld.(*cephv1.CephCluster)
+			newObj := e.ObjectNew.(*cephv1.CephCluster)
+
+			ignorePaths := func(path cmp.Path) bool {
+				switch path.String() {
+				case "ObjectMeta.ManagedFields", "ObjectMeta.ResourceVersion",
+					"Status.CephStatus.LastChecked", "Status.CephStatus.Capacity.LastUpdated",
+					"Status.Conditions.LastHeartbeatTime", "Status.Conditions.LastTransitionTime":
```
nb-ohad, line 189: Can we please format this with one case per line to help with readability
```suggestion
				case "ObjectMeta.ManagedFields", 
					"ObjectMeta.ResourceVersion",
					"Status.CephStatus.LastChecked", 
					"Status.CephStatus.Capacity.LastUpdated",
					"Status.Conditions.LastHeartbeatTime", 
					"Status.Conditions.LastTransitionTime":
```

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -144,14 +149,76 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+			if e.ObjectOld == nil || e.ObjectNew == nil {
+				return false
+			}
+			oldObj := e.ObjectOld.(*nbv1.NooBaa)
+			newObj := e.ObjectNew.(*nbv1.NooBaa)
+
+			ignorePaths := func(path cmp.Path) bool {
+				switch path.String() {
+				case "ObjectMeta.ManagedFields", "ObjectMeta.ResourceVersion",
+					"Status.Conditions.LastHeartbeatTime", "Status.Conditions.LastTransitionTime":
```
nb-ohad, line 163: Can we please format this with one case per line to help with readability
```suggestion
				case "ObjectMeta.ManagedFields", 
					"ObjectMeta.ResourceVersion",
					"Status.Conditions.LastHeartbeatTime", 
					"Status.Conditions.LastTransitionTime":
```

nb-ohad (changes requested):

controllers/storagecluster/storagecluster_controller.go
iamniting: Done

controllers/storagecluster/storagecluster_controller.go
iamniting: Done

iamniting:
> @iamniting Are we sure the path for an array item looks like this? `"Status.Conditions.LastTransitionTime"` shouldn't it need a moniker for the item?

I have tested it, It works like this only.

nb-ohad:
LGTM - This seems good from my side, not merging to allow others to comment.

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -144,28 +148,91 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-		Owns(&corev1.PersistentVolumeClaim{}, builder.WithPredicates(pvcPredicate)).
+		Owns(&cephv1.CephCluster{}, builder.WithPredicates(cephClusterIgnoreTimeUpdatePredicate)).
+		Owns(&cephv1.CephBlockPool{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		Owns(&cephv1.CephFilesystem{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		Owns(&cephv1.CephFilesystemSubVolumeGroup{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		Owns(&cephv1.CephNFS{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		Owns(&cephv1.CephObjectStore{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		Owns(&cephv1.CephObjectStoreUser{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		Owns(&cephv1.CephRBDMirror{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		Owns(&corev1.PersistentVolumeClaim{}, builder.WithPredicates(predicate.GenerationChangedPredicate{}, pvcPredicate)).
```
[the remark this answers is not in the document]
Madhu-1: i think this can be removed as well as we don't have any PVC Controller as of now. we removed it long back but we need to see if any other functionality depends on this one before removing it.

PR #2630, opened 2024-05-28
Bug 2279928:[release-4.16] exporter: add metric and alert if a system has OSD based on LVM

weirdwiz:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2279928


umangachapagain (approved):

weirdwiz:
ah, right
sent a new patch

PR #2635, opened 2024-05-30
use "upmap-read" as default balancer mode

sp98:
This PR changes the default mode for the balancer module from "upmap" to "upmap-read". This new mode was added in Squid due to the balancer module not balancing the reads (BZ - https://bugzilla.redhat.com/show_bug.cgi?id=1870804)


travisn:
Is this option already in the downstream builds, or we need to wait on this a bit?

iamniting (approved):

PR #2639, opened 2024-05-30
Bug 2276694: [release-4.16] Remove default value of waitTimeoutForHealthyOSDInMinutes

Nikhil-Ladha:
Remove default value of waitTimeoutForHealthyOSDInMinutes from ocs-operator, and only update the SC CR, if a value is provided by the user.
Reference PR that added this code: #2579

### DIFF
--- a/controllers/storagecluster/cephcluster.go
+++ b/controllers/storagecluster/cephcluster.go
@@ -490,8 +490,11 @@ func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, serverVersion *v
 					KernelMountOptions: getCephFSKernelMountOptions(sc),
 				},
 			},
-			WaitTimeoutForHealthyOSDInMinutes: getWaitTimeoutForHealthOSD(sc),
 		},
+	}
+
+	if sc.Spec.ManagedResources.CephCluster.WaitTimeoutForHealthyOSDInMinutes != 0 {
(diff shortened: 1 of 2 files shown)


controllers/storagecluster/cephcluster.go
```diff
@@ -490,10 +490,13 @@ func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, serverVersion *v
 					KernelMountOptions: getCephFSKernelMountOptions(sc),
 				},
 			},
-			WaitTimeoutForHealthyOSDInMinutes: getWaitTimeoutForHealthOSD(sc),
 		},
 	}
 
+	if sc.Spec.ManagedResources.CephCluster.WaitTimeoutForHealthyOSDInMinutes != 0 {
```
travisn, line 496: This will be set in every case on the CephCluster CR whenever this setting is on the StorageCluster CR, right? From [Petr's comment](https://github.com/red-hat-storage/ocs-operator/pull/2579#discussion_r1620875815), he saw it wasn't always applied previously, but perhaps this will fix it as well?

Nikhil-Ladha: Yes, this should set it always whenever a change is applied on the storageCluster CR.
Infact, it should have been previously as well 🤔

travisn (commented):
/approve
/lgtm

Nikhil-Ladha:
Tested the change, and everything is working fine.

iamniting (approved):

PR #2661, opened 2024-06-09
Sends `VolumeGroupSnapshotClass` to client

raaizik:
Sends `VolumeGroupSnapshotClass` as an external resource to the client via the storage claim response.
Must be merged prior to [ocs-client-operator/pull/168](https://github.com/red-hat-storage/ocs-client-operator/pull/168).
[RHSTOR-5795](https://issues.redhat.com/browse/RHSTOR-5795)

### DIFF
--- a/services/provider/server/server.go
+++ b/services/provider/server/server.go
@@ -611,19 +611,24 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
 				rbdStorageClassData["encryptionKMSID"] = storageRequest.Spec.EncryptionMethod
 			}
 
-			extR = append(extR, &pb.ExternalResource{
-				Name: "ceph-rbd",
-				Kind: "StorageClass",
-				Data: mustMarshal(rbdStorageClassData),
-			})
-
-			extR = append(extR, &pb.ExternalResource{
-				Name: "ceph-rbd",
-				Kind: "VolumeSnapshotClass",
-				Data: mustMarshal(map[string]string{
-					"clusterID": rbdStorageClassData["clusterID"],
-					"csi.storage.k8s.io/snapshotter-secret-name": provisionerSecretName,
-				})})
+			extR = append(extR,
+				&pb.ExternalResource{
+					Name: "ceph-rbd",
+					Kind: "StorageClass",
+					Data: mustMarshal(rbdStorageClassData),
+				},
+				&pb.ExternalResource{
+					Name: "ceph-rbd",
+					Kind: "VolumeSnapshotClass",
+					Data: mustMarshal(map[string]string{
+						"csi.storage.k8s.io/snapshotter-secret-name": provisionerSecretName,
+					})},
+				&pb.ExternalResource{
+					Name: "ceph-rbd",
+					Kind: "VolumeGroupSnapshotClass",
+					Data: mustMarshal(map[string]string{
+						"csi.storage.k8s.io/group-snapshotter-secret-name": provisionerSecretName,
+					})})
 
 		case "CephFilesystemSubVolumeGroup":
 			subVolumeGroup := &rookCephv1.CephFilesystemSubVolumeGroup{}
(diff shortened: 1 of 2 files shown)


leelavg (approved):
Take a look at Jira, this PR is standalone right even if we use or not use the feature, like, creating resource in the client side will not have any side-effects & dependencies?

leelavg:
/unhold

for a sec I thought ocs-client-op repo and put a hold.

services/provider/server/server.go
```diff
@@ -664,6 +672,14 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
 					"clusterID": getSubVolumeGroupClusterID(subVolumeGroup),
 					"csi.storage.k8s.io/snapshotter-secret-name": provisionerSecretName,
 				})})
+
+			extR = append(extR, &pb.ExternalResource{
+				Name: "cephfs",
+				Kind: "VolumeGroupSnapshotClass",
+				Data: mustMarshal(map[string]string{
+					"clusterID": getSubVolumeGroupClusterID(subVolumeGroup),
+					"csi.storage.k8s.io/snapshotter-secret-name": provisionerSecretName,
```
Madhu-1, line 681: `"csi.storage.k8s.io/snapshotter-secret-name"` to `"csi.storage.k8s.io/group-snapshotter-secret-name"`, @raaizik we need to rethink about sending this information from provider to consumer cluster because this information already exists on the consumer cluster like clusterID it decided on the consumer cluster, not on the provider cluster and same with the secret as well. we need to break a few things like this. cc @nb-ohad

services/provider/server/server.go
```diff
@@ -664,6 +672,14 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
 					"clusterID": getSubVolumeGroupClusterID(subVolumeGroup),
 					"csi.storage.k8s.io/snapshotter-secret-name": provisionerSecretName,
 				})})
+
+			extR = append(extR, &pb.ExternalResource{
+				Name: "cephfs",
+				Kind: "VolumeGroupSnapshotClass",
+				Data: mustMarshal(map[string]string{
+					"clusterID": getSubVolumeGroupClusterID(subVolumeGroup),
```
nb-ohad, line 680: Can we please calculate the cluster-ID once, store it in a var, and use it where necessary?

services/provider/server/server.go
nb-ohad, line 681: Each append has the potential to grow the capacity of the baking storage of the slice. That means coping of the entire slice data. 

I would suggest refactoring the code so we will have a single append operation:
```
extR = append(
    extR, 
    &pb.ExternalResource{
    },
    &pb.ExternalResource{
    },
    &pb.ExternalResource{
    },
)
```

nb-ohad (changes requested):

services/provider/server/server.go
```diff
@@ -625,6 +625,14 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
 					"csi.storage.k8s.io/snapshotter-secret-name": provisionerSecretName,
 				})})
 
+			extR = append(extR, &pb.ExternalResource{
+				Name: "ceph-rbd",
+				Kind: "VolumeGroupSnapshotClass",
+				Data: mustMarshal(map[string]string{
+					"clusterID": rbdStorageClassData["clusterID"],
```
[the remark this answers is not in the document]
raaizik: This is how `clusterID` is  generated for `VolumeSnapshotClass` of both types, as well. Are you suggesting I change that? Why?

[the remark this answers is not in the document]
raaizik: Also, this is not "data from another resource", it holds the namespace of `OCSProviderServer`

services/provider/server/server.go
raaizik: @Madhu-1 I agree, although I am not sure how to go about this. Let's look at `ClusterID`: for storage classes of type cephFS, it seems like the server generates it with data pulled from `CephFilesystemSubVolumeGroup`. For that of type cephRBD, it stores in it the namespace of `OCSProviderServer`. Now, on the client both of those get overridden with `storageClaimHash`. Which side is correct?

Madhu-1: it doesnt make sense to send the information from provider to consumer which is solely decided on the consumer cluster not on the provider cluster. https://github.com/red-hat-storage/ocs-client-operator/blob/3165a1b7db9b73661099f89005ed271690234240/controllers/storageclaim_controller.go#L373-L379 this is complete decided on the consumer, whatever the information sent from provider its not used anywhere in the provider cluster.

raaizik: That is obvious, my question was which side is correct, not what is the eventual value that's picked up. Asking because it looks too [elaborate](https://github.com/red-hat-storage/ocs-operator/blob/main/services/provider/server/server.go#L428-L437) to be coincidental.

Madhu-1: not sending it on the provider but setting it up at the consumer seems correct to me.

raaizik: @nb-ohad 

> on the client the clusterID of both of those gets overridden with storageClaimHash

raaizik: Final decision is to remove `clusterID` calculation here. @Madhu-1 regarding `csi.storage.k8s.io/snapshotter-secret-name`: it isn't assigned in the consumer, only `csi.storage.k8s.io/snapshotter-secret-namespace` is.

services/provider/server/server.go
```diff
@@ -611,19 +611,24 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
...
+					Name: "ceph-rbd",
+					Kind: "VolumeSnapshotClass",
+					Data: mustMarshal(map[string]string{
+						"csi.storage.k8s.io/snapshotter-secret-name": provisionerSecretName,
+					})},
+				&pb.ExternalResource{
+					Name: "ceph-rbd",
+					Kind: "VolumeGroupSnapshotClass",
+					Data: mustMarshal(map[string]string{
+						"csi.storage.k8s.io/snapshotter-secret-name": provisionerSecretName,
```
Madhu-1, line 630: https://github.com/kubernetes-csi/external-snapshotter/blob/c72f087f751abb285a90c4f5bb02df9014d2bc19/pkg/utils/util.go#L62-L63 is the key for groupsnapshot secret

services/provider/server/server.go
```diff
@@ -644,26 +649,30 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
...
+					Name: "cephfs",
+					Kind: "VolumeSnapshotClass",
+					Data: mustMarshal(map[string]string{
+						"csi.storage.k8s.io/snapshotter-secret-name": provisionerSecretName,
+					})},
+				&pb.ExternalResource{
+					Name: "cephfs",
+					Kind: "VolumeGroupSnapshotClass",
+					Data: mustMarshal(map[string]string{
+						"csi.storage.k8s.io/snapshotter-secret-name": provisionerSecretName,
```
Madhu-1, line 674: https://github.com/kubernetes-csi/external-snapshotter/blob/c72f087f751abb285a90c4f5bb02df9014d2bc19/pkg/utils/util.go#L62-L63 is the key for groupsnapshot secret

Madhu-1 (changes requested):

Madhu-1 (approved):
LGTM

nb-ohad:
@raaizik can you please check why the e2e test failed?

PR #2663, opened 2024-06-11
Bug 2291321: [release-4.15] Set deviceClasses to avoid replicated pool spreading PGs across all OSDs

malayparida2000:
This is a manual backport of https://github.com/red-hat-storage/ocs-operator/pull/2615

BZ-https://bugzilla.redhat.com/show_bug.cgi?id=2274175

Set the deviceClasses of the pools to avoid wrong data placement

Earlier when replica-1 was getting enabled on an existing cluster the pool was setting it's deviceClass to replicated but there were no osds with this deviceClass as osd's during normal creation were given SSD deviceClass if blank, and now they can't change. To avoid & solve this problem we have to intelligently set the deviceClass for the pools, Depending on whether replica-1 is enabled & the number of deviceClasses found in the cephCluster CR status.

To do this we added a DefaultCephDeviceClass field in StorageCluster Status which will hold the default ceph device class to be used for the pools. Depending on whether replica-1 is enabled & the number of deviceClasses found in the status this value is determined and set in the status.

### DIFF
--- a/api/v1/storagecluster_types.go
+++ b/api/v1/storagecluster_types.go
@@ -534,6 +534,9 @@ type StorageClusterStatus struct {
 	// Images holds the image reconcile status for all images reconciled by the operator
 	Images ImagesStatus `json:"images,omitempty"`
 
+	// DefaultCephDeviceClass holds the default ceph device class to be used for the pools
+	DefaultCephDeviceClass string `json:"defaultCephDeviceClass,omitempty"`
+
 	// KMSServerConnection holds the connection state to the KMS server.
 	KMSServerConnection KMSServerConnectionStatus `json:"kmsServerConnection,omitempty"`
 
--- a/config/crd/bases/ocs.openshift.io_storageclusters.yaml
+++ b/config/crd/bases/ocs.openshift.io_storageclusters.yaml
@@ -6834,6 +6834,10 @@ spec:
                 description: CurrentMonCount holds the value of ceph mons configured
                   in ceph cluster.
                 type: integer
+              defaultCephDeviceClass:
+                description: DefaultCephDeviceClass holds the default ceph device
+                  class to be used for the pools
+                type: string
               externalSecretHash:
                 description: ExternalSecretHash holds the checksum value of external
                   secret data.
--- a/controllers/storagecluster/cephblockpools.go
+++ b/controllers/storagecluster/cephblockpools.go
@@ -61,7 +61,7 @@ func (r *StorageClusterReconciler) newCephBlockPoolInstances(initData *ocsv1.Sto
 			},
 			Spec: cephv1.NamedBlockPoolSpec{
 				PoolSpec: cephv1.PoolSpec{
-					DeviceClass:    generateDeviceClass(initData),
+					DeviceClass:    initData.Status.DefaultCephDeviceClass,
 					FailureDomain:  getFailureDomain(initData),
 					Replicated:     generateCephReplicatedSpec(initData, "data"),
 					EnableRBDStats: true,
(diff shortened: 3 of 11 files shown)


iamniting (approved):

travisn:
Are we also backporting #2694? This fix isn't complete without that, right?

iamniting:
@malayparida2000 Can you pls take a look at Travis request.

malayparida2000:
> Are we also backporting #2694? This fix isn't complete without that, right?

My understanding was the issue which #2694 solves was not caused by #2615 and they were independent problems.
My reason for believing this was that the BZ for #2615 was raised way back for ODF 4.13. The original fix for that BZ is still on master branch & not yet verified. If you belive this fix on it's own is not useful enough we have to block this fix till the time the BZ of #2694 is verified on 4.17, then backported to 4.16. After that we have to take both the backports together on 4.15.

travisn:
> > Are we also backporting #2694? This fix isn't complete without that, right?
> 
> My understanding was the issue which #2694 solves was not caused by #2615 and they were independent problems. My reason for believing this was that the BZ for #2694 [Bug 2253013](https://bugzilla.redhat.com/show_bug.cgi?id=2253013) was raised way back for ODF 4.13. The original fix for that BZ is still on 4.17 & not yet verified. If you belive this fix on it's own is not useful enough we have to block this fix till the time the BZ of #2694 is verified on 4.17, then backported to 4.16. After that we have to take both the backports together on 4.15.

Let's wait to backport this until we can verify #2694 and #2615 together, which will fix both [2253013](https://bugzilla.redhat.com/show_bug.cgi?id=2253013) and [2274175](https://bugzilla.redhat.com/show_bug.cgi?id=2274175). While the data pools were fixed with #2615, the metadata pools would still be spreading their metadata across all the OSDs. It's true that #2615 is the majority of the fix, but we need #2694 as well for the full fix.

malayparida2000:
> > > Are we also backporting #2694? This fix isn't complete without that, right?
> > 
> > 
> > My understanding was the issue which #2694 solves was not caused by #2615 and they were independent problems. My reason for believing this was that the BZ for #2694 [Bug 2253013](https://bugzilla.redhat.com/show_bug.cgi?id=2253013) was raised way back for ODF 4.13. The original fix for that BZ is still on 4.17 & not yet verified. If you belive this fix on it's own is not useful enough we have to block this fix till the time the BZ of #2694 is verified on 4.17, then backported to 4.16. After that we have to take both the backports together on 4.15.
> 
> Let's wait to backport this until we can verify #2694 and #2615 together, which will fix both [2253013](https://bugzilla.redhat.com/show_bug.cgi?id=2253013) and [2274175](https://bugzilla.redhat.com/show_bug.cgi?id=2274175). While the data pools were fixed with #2615, the metadata pools would still be spreading their metadata across all the OSDs. It's true that #2615 is the majority of the fix, but we need #2694 as well for the full fix.

ack, Keeping the PR on hold then & updating the BZ accordingly to reflect that this BZ is blocked by that other BZ.

travisn:
In my investigation for **, we also need these fixes if we're going to merge this back to 4.15:
- Skip updating crush rules for stretch clusters: https://github.com/rook/rook/pull/14447
- Skip crush rule update when not needed: https://github.com/rook/rook/pull/13772

These were already in the 4.16 branch, where we are testing #2694 for the 4.16.1 blocker.

malayparida2000:
> In my investigation for [2303080](https://bugzilla.redhat.com/show_bug.cgi?id=2303080), we also need these fixes if we're going to merge this back to 4.15:
> 
> * [pool: Skip updating crush rules for stretch clusters rook/rook#14447](https://github.com/rook/rook/pull/14447)
> * [pool: Skip crush rule update when not needed rook/rook#13772](https://github.com/rook/rook/pull/13772)
> 
> These were already in the 4.16 branch, where we are testing #2694 for the 4.16.1 blocker.

I see, So is rook planning to backport the fixes to 4.15?

travisn:
> I see, So is rook planning to backport the fixes to 4.15?

With all these PRs, it feels risky to backport. Let's discuss more the risk of not fixing this in 4.15.z.

malayparida2000:
Closing as the attached BZ is now closed.

PR #2665, opened 2024-06-14
BUG 2292114: api: changed storageConsumer status fields to optional

rohan47:
Changed the status fields for storageconsumers to optional

### DIFF
--- a/api/v1alpha1/storageconsumer_types.go
+++ b/api/v1alpha1/storageconsumer_types.go
@@ -72,15 +72,24 @@ type StorageConsumerStatus struct {
 // ClientStatus is the information pushed from connected storage client
 type ClientStatus struct {
 	// StorageClient Platform Version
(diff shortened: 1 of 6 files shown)


leelavg:
@rohan47 pls also add marker `+optional` on all these fields. I know the testing works, adding this will *not* result (can cross check as well) in a changed CRD and so further testing isn't required. 

Ref: https://github.com/kubernetes/community/blob/master/contributors%2Fdevel%2Fsig-architecture%2Fapi-conventions.md

leelavg:
@rohan47 I looked at kubebuilder docs to see if there is an option to set the optional at struct level, seems there is one https://book.kubebuilder.io/reference/markers.html#difference-between--optional-and--kubebuildervalidationoptional, mind giving it a go?

leelavg (approved):

rohan47:
The changes are tested.

PR #2668, opened 2024-06-18
Adding storage quota field to spec of storage consumer

rchikatw:
- Adding spec field to storage consumer CR
- Making it optional because for the consumer whos is having unlimited quota this field will not be set

### DIFF
--- a/api/v1alpha1/storageconsumer_types.go
+++ b/api/v1alpha1/storageconsumer_types.go
@@ -40,6 +40,9 @@ const (
 type StorageConsumerSpec struct {
 	// Enable flag ignores a reconcile if set to false
 	Enable bool `json:"enable,omitempty"`
+	// StorageQuotaInGiB describes quota for the consumer
+	// +optional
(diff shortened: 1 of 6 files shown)


rchikatw:
@leelavg @nb-ohad Please Review

api/v1alpha1/storageconsumer_types.go
```diff
@@ -40,6 +40,9 @@ const (
 type StorageConsumerSpec struct {
 	// Enable flag ignores a reconcile if set to false
 	Enable bool `json:"enable,omitempty"`
+	// StorageQuotaInGi describes quota for the consumer
+	// +optional
+	StorageQuotaInGi int `json:"storageQuotaInGi,omitempty"`
```
nb-ohad, line 45: The term is GigaBytes (Gib)
```suggestion
	StorageQuotaInGiB int `json:"storageQuotaInGiB,omitempty"`
```

nb-ohad (changes requested):

PR #2669, opened 2024-06-19
changes to onboarding token api to include role in body

rewantsoni:
Update the onboarding Ticket to include an OnboardingSubjectRole - ocs-client, ocs-peer
1. ocs-client will be used to onboard a Client
2. ocs-peer role will be used to onboard a StorageCluster

Add a new endpoint to generate the peer onboarding token.

### DIFF
--- a/controllers/util/provider.go
+++ b/controllers/util/provider.go
@@ -17,10 +17,31 @@ import (
 	"github.com/red-hat-storage/ocs-operator/v4/services"
 )
 
-// GenerateOnboardingToken generates a token valid for a duration of "tokenLifetimeInHours".
+// GenerateClientOnboardingToken generates a ocs-client token valid for a duration of "tokenLifetimeInHours".
 // The token content is predefined and signed by the private key which'll be read from supplied "privateKeyPath".
 // The storageQuotaInGiB is optional, and it is used to limit the storage of PVC in the application cluster.
-func GenerateOnboardingToken(tokenLifetimeInHours int, privateKeyPath string, storageQuotaInGiB *uint) (string, error) {
+func GenerateClientOnboardingToken(tokenLifetimeInHours int, privateKeyPath string, storageQuotainGib *uint) (string, error) {
+	tokenExpirationDate := time.Now().
+		Add(time.Duration(tokenLifetimeInHours) * time.Hour).
+		Unix()
+
+	ticket := services.OnboardingTicket{
+		ID:                uuid.New().String(),
+		ExpirationDate:    tokenExpirationDate,
+		SubjectRole:       services.ClientRole,
+		StorageQuotaInGiB: storageQuotainGib,
+	}
+
+	token, err := encodeAndSignOnboardingToken(privateKeyPath, ticket)
+	if err != nil {
+		return "", err
+	}
+	return token, nil
+}
+
+// GeneratePeerOnboardingToken generates a ocs-peer token valid for a duration of "tokenLifetimeInHours".
+// The token content is predefined and signed by the private key which'll be read from supplied "privateKeyPath".
+func GeneratePeerOnboardingToken(tokenLifetimeInHours int, privateKeyPath string) (string, error) {
 	tokenExpirationDate := time.Now().
 		Add(time.Duration(tokenLifetimeInHours) * time.Hour).
 		Unix()
@@ -28,10 +49,18 @@ func GenerateOnboardingToken(tokenLifetimeInHours int, privateKeyPath string, st
 	ticket := services.OnboardingTicket{
 		ID:             uuid.New().String(),
 		ExpirationDate: tokenExpirationDate,
+		SubjectRole:    services.PeerRole,
 	}
-	if storageQuotaInGiB != nil {
-		ticket.StorageQuotaInGiB = *storageQuotaInGiB
+	token, err := encodeAndSignOnboardingToken(privateKeyPath, ticket)
+	if err != nil {
+		return "", err
 	}
+	return token, nil
+}
+
+// encodeAndSignOnboardingToken generates a token from the ticket.
+// The token content is predefined and signed by the private key which'll be read from supplied "privateKeyPath".
+func encodeAndSignOnboardingToken(privateKeyPath string, ticket services.OnboardingTicket) (string, error) {
 	payload, err := json.Marshal(ticket)
 	if err != nil {
 		return "", fmt.Errorf("failed to marshal the payload: %v", err)
--- a/services/provider/server/server.go
+++ b/services/provider/server/server.go
@@ -116,7 +116,13 @@ func (s *OCSProviderServer) OnboardConsumer(ctx context.Context, req *pb.Onboard
 		return nil, status.Errorf(codes.InvalidArgument, "onboarding ticket is not valid. %v", err)
 	}
 
-	storageConsumerUUID, err := s.consumerManager.Create(ctx, req, int(onboardingTicket.StorageQuotaInGiB))
+	if onboardingTicket.SubjectRole != services.ClientRole {
+		err := fmt.Errorf("unsupported ticket role for consumer %q, found %s, expected %s", req.ConsumerName, onboardingTicket.SubjectRole, services.ClientRole)
+		klog.Error(err)
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+
+	storageConsumerUUID, err := s.consumerManager.Create(ctx, req, int(*onboardingTicket.StorageQuotaInGiB))
 	if err != nil {
 		if !kerrors.IsAlreadyExists(err) && err != errTicketAlreadyExists {
 			return nil, status.Errorf(codes.Internal, "failed to create storageConsumer %q. %v", req.ConsumerName, err)
@@ -487,8 +493,17 @@ func decodeAndValidateTicket(ticket string, pubKey *rsa.PublicKey) (*services.On
 		return nil, fmt.Errorf("failed to unmarshal onboarding ticket message. %v", err)
 	}
 
-	if ticketData.StorageQuotaInGiB > math.MaxInt {
-		return nil, fmt.Errorf("invalid value sent in onboarding ticket, storage quota should be greater than 0 and less than %v: %v", math.MaxInt, ticketData.StorageQuotaInGiB)
+	switch ticketData.SubjectRole {
+	case services.ClientRole:
+		if ticketData.StorageQuotaInGiB != nil {
+			quota := *ticketData.StorageQuotaInGiB
+			if quota > math.MaxInt {
+				return nil, fmt.Errorf("invalid value sent in onboarding ticket, storage quota should be greater than 0 and less than %v: %v", math.MaxInt, quota)
+			}
+		}
+	case services.PeerRole:
+	default:
+		return nil, fmt.Errorf("invalid onboarding ticket subject role")
 	}
 
 	signature, err := base64.StdEncoding.DecodeString(ticketArr[1])
--- a/services/types.go
+++ b/services/types.go
@@ -1,7 +1,15 @@
 package services
 
+type OnboardingSubjectRole string
+
+const (
+	ClientRole OnboardingSubjectRole = "ocs-client"
+	PeerRole   OnboardingSubjectRole = "ocs-peer"
+)
+
 type OnboardingTicket struct {
-	ID                string `json:"id"`
-	ExpirationDate    int64  `json:"expirationDate,string"`
-	StorageQuotaInGiB uint   `json:"storageQuotaInGiB,omitempty"`
+	ID                string                `json:"id"`
+	ExpirationDate    int64                 `json:"expirationDate,string"`
+	SubjectRole       OnboardingSubjectRole `json:"subjectRole"`
+	StorageQuotaInGiB *uint                 `json:"storageQuotaInGiB,omitempty"`
 }
--- a/services/ux-backend/main.go
+++ b/services/ux-backend/main.go
@@ -7,9 +7,10 @@ import (
 	"os"
 	"strconv"
 
-	"k8s.io/klog/v2"
+	"github.com/red-hat-storage/ocs-operator/v4/services/ux-backend/handlers/onboarding/clienttokens"
+	"github.com/red-hat-storage/ocs-operator/v4/services/ux-backend/handlers/onboarding/peertokens"
 
-	"github.com/red-hat-storage/ocs-operator/v4/services/ux-backend/handlers/onboardingtokens"
+	"k8s.io/klog/v2"
 )
 
 type serverConfig struct {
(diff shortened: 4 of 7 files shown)


services/ux-backend/handlers/onboardingtokens/handler.go
```diff
@@ -16,14 +19,33 @@ const (
...
+			http.Error(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+		if strings.ToLower(body.Role) == string(services.MirroringPeerRole) {
+			role = services.MirroringPeerRole
+		} else if strings.ToLower(body.Role) == string(services.ClientRole) || body.Role == "" {
+			// to ensure backward compatibility, if the role is empty, we assume it to be client
+			role = services.ClientRole
+		} else {
+			http.Error(w, fmt.Sprintf("invalid Role sent in request body, Valid types are [mirroring-peer,client]: %v", body.Role), http.StatusBadRequest)
```
rchikatw, line 44: Valid types -> Valid roles

rchikatw:
PR looks good to me.

controllers/util/provider.go
```diff
@@ -20,18 +20,21 @@ import (
...
 	tokenExpirationDate := time.Now().
 		Add(time.Duration(tokenLifetimeInHours) * time.Hour).
 		Unix()
 
 	ticket := services.OnboardingTicket{
 		ID:             uuid.New().String(),
 		ExpirationDate: tokenExpirationDate,
+		Role:           role,
 	}
+
```
nb-ohad, line 33:
```suggestion
```

services/types.go
```diff
@@ -1,7 +1,16 @@
 package services
 
+type Role string
+
+const (
+	ClientRole        Role = "client"
+	MirroringPeerRole Role = "mirroring-peer"
```
nb-ohad, line 7: The rule is not related to mirroring. 
As for the agreed architecture, there are 2 rules: `ocs-client` and `ocs-peer`,

Please align everything with the given terminology

services/types.go
```diff
@@ -1,7 +1,16 @@
...
+)
+
 type OnboardingTicket struct {
 	ID                string `json:"id"`
 	ExpirationDate    int64  `json:"expirationDate,string"`
 	StorageQuotaInGiB uint   `json:"storageQuotaInGiB,omitempty"`
+	// Role can be either client or mirroring-peer
+	Role Role `json:"role,string"`
```
nb-ohad, line 15: The rule is not for the token, it is for the subject. 
```suggestion
	SubjectRole Role `json:"subjectRole,string"`
```

services/types.go
```diff
@@ -1,7 +1,16 @@
 package services
 
+type Role string
```
nb-ohad, line 3: This is a very generic name and might create collision or confusion. 
Please rename to something more meaningful and more specific. 
```suggestion
type OnboardingSubjectRole string
```

services/ux-backend/handlers/onboardingtokens/handler.go
```diff
@@ -33,31 +35,42 @@ func HandleMessage(w http.ResponseWriter, r *http.Request, tokenLifetimeInHours
...
+				return
+			}
+			unitAsGiB, ok := unitToGib[body.Unit]
+			if !ok {
+				http.Error(w, fmt.Sprintf("invalid Unit type sent in request body, Valid types are [Gi,Ti,Pi]: %v", body.Unit), http.StatusBadRequest)
+				return
+			}
+			storageQuotaInGiB = ptr.To(unitAsGiB * body.Value)
+		} else {
+			http.Error(w, fmt.Sprintf("invalid Role sent in request body, Valid roles are [mirroring-peer,client]: %v", body.Role), http.StatusBadRequest)
```
nb-ohad, line 69: We should not do that. This will allow anyone to generate a peer onboarding token.

nb-ohad (changes requested):

services/types.go
```diff
@@ -1,7 +1,16 @@
...
+const (
+	ClientRole OnboardingSubjectRole = "ocs-client"
+	PeerRole   OnboardingSubjectRole = "ocs-peer"
+)
+
 type OnboardingTicket struct {
 	ID                string `json:"id"`
 	ExpirationDate    int64  `json:"expirationDate,string"`
 	StorageQuotaInGiB uint   `json:"storageQuotaInGiB,omitempty"`
+	// OnboardingSubjectRole can be either ocs-client or ocs-peer
```
nb-ohad, line 14: If you are adding a comment please separate using a blank line
```suggestion

	// OnboardingSubjectRole can be either ocs-client or ocs-peer
```

services/ux-backend/handlers/onboardingtokens/handler.go
```diff
@@ -33,32 +36,44 @@ func HandleMessage(w http.ResponseWriter, r *http.Request, tokenLifetimeInHours
...
 	// storage quota is unlimited
 	var err error
 	if r.ContentLength != 0 {
-		var quota = struct {
-			Value uint   `json:"value"`
-			Unit  string `json:"unit"`
+		var body = struct {
+			SubjectRole string `json:"subjectRole"`
+			Value       uint   `json:"value"`
+			Unit        string `json:"unit"`
```
nb-ohad, line 48: These does not belong on the same level

services/ux-backend/handlers/onboardingtokens/handler.go
```diff
@@ -33,32 +36,44 @@ func HandleMessage(w http.ResponseWriter, r *http.Request, tokenLifetimeInHours
...
+				return
+			}
+			unitAsGiB, ok := unitToGib[body.Unit]
+			if !ok {
+				http.Error(w, fmt.Sprintf("invalid Unit type sent in request body, Valid types are [Gi,Ti,Pi]: %v", body.Unit), http.StatusBadRequest)
+				return
+			}
+			storageQuotaInGiB = ptr.To(unitAsGiB * body.Value)
+		} else {
+			http.Error(w, fmt.Sprintf("invalid OnboardingSubjectRole sent in request body, Valid roles are [mirroring-peer,client]: %v", body.SubjectRole), http.StatusBadRequest)
```
nb-ohad, line 70: The flow is structured incorrectly. The current structure allows sending peer role with quota fields and the quota fields are just ignored. This is the wrong behavior. I would expect a bad request response instead.

I would suggest using a switch case based on the role and delegate each of the 2 cases into separate functions. In addition, you could use the "default" case to return an error for an unsupported  role

nb-ohad (changes requested):

services/provider/server/server.go
```diff
@@ -470,7 +470,7 @@ func getSubVolumeGroupClusterID(subVolumeGroup *rookCephv1.CephFilesystemSubVolu
 	return hex.EncodeToString(hash[:16])
 }
 
-func decodeAndValidateTicket(ticket string, pubKey *rsa.PublicKey) (*services.OnboardingTicket, error) {
+func decodeAndValidateTicket(ticket string, pubKey *rsa.PublicKey, expectedRole services.OnboardingSubjectRole) (*services.OnboardingTicket, error) {
```
nb-ohad, line 473: Decode should not be passed an expected role. The role comes from the payload itself. If after decoding you don't get a ticket the meet your need you can fail the request upward but you should not send an expected role into this function.

services/provider/server/server.go
```diff
@@ -487,8 +487,13 @@ func decodeAndValidateTicket(ticket string, pubKey *rsa.PublicKey) (*services.On
 		return nil, fmt.Errorf("failed to unmarshal onboarding ticket message. %v", err)
 	}
 
-	if ticketData.StorageQuotaInGiB > math.MaxInt {
-		return nil, fmt.Errorf("invalid value sent in onboarding ticket, storage quota should be greater than 0 and less than %v: %v", math.MaxInt, ticketData.StorageQuotaInGiB)
+	if expectedRole == services.ClientRole {
```
nb-ohad, line 490: This is the wrong assumption to make. There should be a switch statement here based on the decoded role

services/types.go
```diff
@@ -1,7 +1,18 @@
...
+)
+
 type OnboardingTicket struct {
-	ID                string `json:"id"`
-	ExpirationDate    int64  `json:"expirationDate,string"`
-	StorageQuotaInGiB uint   `json:"storageQuotaInGiB,omitempty"`
+	ID             string `json:"id"`
+	ExpirationDate int64  `json:"expirationDate,string"`
+
+	// SubjectRole specifies the role
```
nb-ohad, line 14:
```suggestion
```

services/types.go
```diff
@@ -1,7 +1,18 @@
...
+	SubjectRole OnboardingSubjectRole `json:"subjectRole"`
+
```
nb-ohad, line 16:
```suggestion
```

services/ux-backend/handlers/clientonboardingtokens/handler.go
nb-ohad, line 1: You cannot rename this file. The file name is supposed to represent the name of the endpoint

services/ux-backend/handlers/clientonboardingtokens/handler.go
```diff
@@ -74,10 +76,11 @@ func handlePost(w http.ResponseWriter, r *http.Request, tokenLifetimeInHours int
 			klog.Errorf("failed write data to response writer: %v", err)
 		}
 	}
+
 }
 
 func handleUnsupportedMethod(w http.ResponseWriter, r *http.Request) {
-	klog.Info("Only POST method should be used to send data to this endpoint /onboarding-tokens")
+	klog.Info("Only POST method should be used to send data to this endpoint /client-onboarding-tokens")
```
nb-ohad, line 83: You cannot change the endpoint, we need to keep the endpoint for backport compatibility

services/ux-backend/main.go
```diff
@@ -61,8 +62,15 @@ func main() {
 		klog.Info("shutting down!")
 		os.Exit(-1)
 	}
+	//TODO: deprecate in future release
 	http.HandleFunc("/onboarding-tokens", func(w http.ResponseWriter, r *http.Request) {
-		onboardingtokens.HandleMessage(w, r, config.tokenLifetimeInHours)
+		clientonboardingtokens.HandleMessage(w, r, config.tokenLifetimeInHours)
```
nb-ohad, line 67: This will result in wrong error handling because the code inside assumes a `client-onboarding-tokens` endpoint

services/ux-backend/main.go
```diff
@@ -61,8 +62,15 @@ func main() {
...
+	})
+	http.HandleFunc("/client-onboarding-tokens", func(w http.ResponseWriter, r *http.Request) {
+		clientonboardingtokens.HandleMessage(w, r, config.tokenLifetimeInHours)
+	})
+	http.HandleFunc("/peer-onboarding-tokens", func(w http.ResponseWriter, r *http.Request) {
+		peeronboardingtokens.HandleMessage(w, r, config.tokenLifetimeInHours)
```
nb-ohad, line 73: I am not sure it makes sense that the 2 different onboarding tokens are different REST resources. 
In my mind, it makes more sense that these are 2 sub-resource of a single resource type. 
So in that case the endpoint should be modeled as:

`/onboarding-tokens/client-tokens/` and `/onboarding-tokens/peer-tokens/`
In this case, the mapping of the endpoint `/onboarding-tokens` is to `onboarding-tokens/client-tokens`

This also means that you need to change the folder accordingly

nb-ohad (changes requested):

services/ux-backend/main.go
```diff
@@ -61,7 +60,8 @@ func main() {
 		klog.Info("shutting down!")
 		os.Exit(-1)
 	}
-	http.HandleFunc("/onboarding-tokens", func(w http.ResponseWriter, r *http.Request) {
+
+	http.HandleFunc("/onboarding-tokens/{token-type}", func(w http.ResponseWriter, r *http.Request) {
```
rchikatw, line 64: Please verify without passing {token-type} to confirm if it's working as expected. I'm asking because generating a token from the UI will not pass this test.

services/ux-backend/handlers/onboardingtokens/peertokens/handler.go
```diff
@@ -0,0 +1,54 @@
...
+	switch r.Method {
+	case "POST":
+		handlePost(w, r, tokenLifetimeInHours)
+	default:
+		handleUnsupportedMethod(w, r)
+	}
+}
+
+func handlePost(w http.ResponseWriter, _ *http.Request, tokenLifetimeInHours int) {
+
```
rchikatw, line 25: remove this line

services/ux-backend/main.go
rewantsoni: Tried a few examples locally and for maintaing backward compatibilty, we have two options,

1. We either manually trim the prefix and match the path
2. Have two endpoints one `/onboarding-tokens/{tokenType}` and the other being `/onboarding-tokens`

rchikatw: how about adding query parameters these can be optional.

rewantsoni: Had a discussion with Ohad offline and decided to keep two different endpoints for client and peer onboarding

rchikatw: okay, you can mark this as resolved

services/provider/server/server.go
```diff
@@ -487,8 +493,16 @@ func decodeAndValidateTicket(ticket string, pubKey *rsa.PublicKey) (*services.On
...
+	switch ticketData.SubjectRole {
+	case services.ClientRole:
+		if ticketData.StorageQuotaInGiB != nil {
+			quota := *ticketData.StorageQuotaInGiB
+			if quota > math.MaxInt {
+				return nil, fmt.Errorf("invalid value sent in onboarding ticket, storage quota should be greater than 0 and less than %v: %v", math.MaxInt, quota)
+			}
+		}
+	default:
+		return nil, fmt.Errorf("invalid onboarding ticket subject role")
```
nb-ohad, line 505: You are missing an (empty) case for the peer role. Without it, a ticket with a peer role will just go to the default case and will return an error

services/ux-backend/handlers/onboardingtokens/clienttoken/handler.go
nb-ohad, line 1: resource names (which are reflected in the file name) should be in plural not singular
Please rename `clienttoken` into `clienttokens`

services/ux-backend/handlers/onboardingtokens/clienttoken/handler.go
```diff
@@ -1,4 +1,4 @@
-package onboardingtokens
+package clienttoken
```
nb-ohad, line 1: Package name should be `clienttokens`

services/ux-backend/handlers/onboardingtokens/peertoken/handler.go
nb-ohad, line 1: resource names (which are reflected in the file name) should be in plural not singular
Please rename `peertoken` into `peertokens`

services/ux-backend/handlers/onboardingtokens/peertoken/handler.go
```diff
@@ -0,0 +1,55 @@
+package peertoken
```
nb-ohad, line 1: package name shouldbe peertokens

services/ux-backend/main.go
```diff
@@ -61,8 +62,19 @@ func main() {
...
+		w.Header().Set("Deprecation", "true") // Standard "Deprecation" header
+		w.Header().Set("Link", "/onboarding-tokens/client-token; rel=\"alternate\"")
+		clienttoken.HandleMessage(w, r, config.tokenLifetimeInHours)
+	})
+	http.HandleFunc("/onboarding-tokens/client-token", func(w http.ResponseWriter, r *http.Request) {
+		clienttoken.HandleMessage(w, r, config.tokenLifetimeInHours)
+	})
+	http.HandleFunc("/onboarding-tokens/peer-token", func(w http.ResponseWriter, r *http.Request) {
+		peertoken.HandleMessage(w, r, config.tokenLifetimeInHours)
 	})
```
nb-ohad, line 78: As of the new path design, onboardingtokens becomes more of a subject then a resource. Maybe we should rename the new paths to `/onboarding/client-tokens` and `/onboarding/peer-tokens`. It will also help in removing the confusion around the old deprecated path.

nb-ohad (changes requested):

services/provider/server/server.go
rewantsoni: I will be adding the case related to the peer role while implementing the rpc calls that we have for StorageClusterPeer, I don't think we should add it beforehand

nb-ohad: This pr intreduces the ability to generate peer tokens which means peer tokens should be decodable (as they are already encodable). Let's have the additional case to support it here

PR #2672, opened 2024-06-24
Adding changes to decode/Unmarshal quota from token

rchikatw:
- Code changes involve decoding/unmarshalling the token when the client is onboarding.
- Added the new test cases
- I have modified the test case to function with the current changes

### DIFF
--- a/services/provider/server/consumer.go
+++ b/services/provider/server/consumer.go
@@ -54,7 +54,7 @@ func newConsumerManager(ctx context.Context, cl client.Client, namespace string)
 }
 
 // Create creates a new storageConsumer resource, updates the consumer cache and returns the storageConsumer UID
-func (c *ocsConsumerManager) Create(ctx context.Context, onboard ifaces.StorageClientOnboarding) (string, error) {
+func (c *ocsConsumerManager) Create(ctx context.Context, onboard ifaces.StorageClientOnboarding, storageQuotaInGiB int) (string, error) {
 	ticket := onboard.GetOnboardingTicket()
 	name := onboard.GetConsumerName()
 	c.mutex.RLock()
@@ -74,7 +74,8 @@ func (c *ocsConsumerManager) Create(ctx context.Context, onboard ifaces.StorageC
 			},
 		},
 		Spec: ocsv1alpha1.StorageConsumerSpec{
-			Enable: false,
+			Enable:            false,
+			StorageQuotaInGiB: storageQuotaInGiB,
 		},
 		Status: ocsv1alpha1.StorageConsumerStatus{
 			Client: ocsv1alpha1.ClientStatus{
--- a/services/provider/server/server.go
+++ b/services/provider/server/server.go
@@ -11,6 +11,7 @@ import (
 	"encoding/json"
 	"encoding/pem"
 	"fmt"
+	"math"
 	"net"
 	"strconv"
 	"strings"
@@ -106,12 +107,13 @@ func (s *OCSProviderServer) OnboardConsumer(ctx context.Context, req *pb.Onboard
 		return nil, status.Errorf(codes.Internal, "failed to get public key to validate onboarding ticket for consumer %q. %v", req.ConsumerName, err)
 	}
 
-	if err := validateTicket(req.OnboardingTicket, pubKey); err != nil {
+	onboardingTicket, err := decodeAndValidateTicket(req.OnboardingTicket, pubKey)
+	if err != nil {
 		klog.Errorf("failed to validate onboarding ticket for consumer %q. %v", req.ConsumerName, err)
 		return nil, status.Errorf(codes.InvalidArgument, "onboarding ticket is not valid. %v", err)
 	}
 
-	storageConsumerUUID, err := s.consumerManager.Create(ctx, req)
+	storageConsumerUUID, err := s.consumerManager.Create(ctx, req, int(onboardingTicket.StorageQuotaInGiB))
 	if err != nil {
 		if !kerrors.IsAlreadyExists(err) && err != errTicketAlreadyExists {
 			return nil, status.Errorf(codes.Internal, "failed to create storageConsumer %q. %v", req.ConsumerName, err)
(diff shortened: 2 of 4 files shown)


services/types.go
```diff
@@ -1,6 +1,7 @@
 package services
 
 type OnboardingTicket struct {
-	ID             string `json:"id"`
-	ExpirationDate int64  `json:"expirationDate,string"`
+	ID                string `json:"id"`
+	ExpirationDate    int64  `json:"expirationDate,string"`
+	StorageQuotaInGiB uint   `json:"storageQuotaInGiB,omitempty"`
```
rchikatw, line 6: Once this pull request (https://github.com/red-hat-storage/ocs-operator/pull/2600/files) is merged, I will rebase this. Then there won't be any changes in the structure.

rchikatw:
@nb-ohad @leelavg Please review.

services/provider/server/consumer.go
```diff
@@ -225,3 +238,23 @@ func (c *ocsConsumerManager) UpdateConsumerStatus(ctx context.Context, id string
...
+	}
+
+	var ticketData services.OnboardingTicket
+	err = json.Unmarshal(message, &ticketData)
+	if err != nil {
+		return 0, fmt.Errorf("failed to unmarshal onboarding ticket %v", err)
+	}
+
+	return int(ticketData.StorageQuotaInGiB), nil
+}
```
nb-ohad, line 260: This will better suit us as a general `parseOnboardingTicket` function, then in the call site you will be able to fetch any piece of information from the ticket. 

You can even move it to the services package and have it as
```go 
ticket := services.DecodeOnboardingTicket(ticketString)
```

services/provider/server/server.go
```diff
@@ -117,6 +117,10 @@ func (s *OCSProviderServer) OnboardConsumer(ctx context.Context, req *pb.Onboard
 			return nil, status.Errorf(codes.Internal, "failed to create storageConsumer %q. %v", req.ConsumerName, err)
 		}
 
+		if err == errTicketInvalidStorageQuota {
+			return nil, status.Errorf(codes.InvalidArgument, "failed to create storageConsumer %q. %v", req.ConsumerName, err)
```
nb-ohad, line 121: Then using the InvalidArgument code the message should describe the argument in question and not a general message.

nb-ohad (changes requested):

services/provider/server/consumer.go
rchikatw: Make sense, we don't have the util package in the server folder. i will create and add a function.

nb-ohad: Can we have it here? https://github.com/red-hat-storage/ocs-operator/blob/main/controllers/util/provider.go

This file already deals with concerns regarding onboarding tokens/tickets

services/provider/server/server.go
```diff
@@ -117,6 +117,10 @@ func (s *OCSProviderServer) OnboardConsumer(ctx context.Context, req *pb.Onboard
 			return nil, status.Errorf(codes.Internal, "failed to create storageConsumer %q. %v", req.ConsumerName, err)
 		}
 
+		if err == errTicketInvalidStorageQuota {
+			return nil, status.Errorf(codes.InvalidArgument, "onboarding ticket with quota is not valid %q", req.OnboardingTicket)
```
nb-ohad, line 121: Use `invalid` instead of "not valid" and use "contains" instead of "with" 
```suggestion
			return nil, status.Errorf(codes.InvalidArgument, "onboarding ticket contains an invalid storage quota value" %q", req.OnboardingTicket)
```

nb-ohad (changes requested):

services/provider/server/consumer.go
rchikatw: Sure

controllers/util/provider.go
```diff
@@ -77,3 +78,23 @@ func readAndDecodePrivateKey(privateKeyPath string) (*rsa.PrivateKey, error) {
...
+	message, err := base64.StdEncoding.DecodeString(ticketArr[0])
+	if err != nil {
+		return nil, fmt.Errorf("failed to decode onboarding ticket: %v", err)
+	}
+
+	var ticketData services.OnboardingTicket
+	err = json.Unmarshal(message, &ticketData)
+	if err != nil {
+		return nil, fmt.Errorf("failed to unmarshal onboarding ticket %v", err)
+	}
```
nb-ohad, line 97: It seems decoding is already a part of the onboarding process here:
https://github.com/red-hat-storage/ocs-operator/blob/main/services/provider/server/server.go#L440-L454 which is the correct place to decode. 

I suggest renaming that method to `decodeAndValidateTicket` and return the decoded ticket as well

services/provider/server/consumer.go
```diff
@@ -83,7 +85,15 @@ func (c *ocsConsumerManager) Create(ctx context.Context, onboard ifaces.StorageC
...
 	}
 
-	err := c.client.Create(ctx, consumerObj)
+	onboardingTicket, err := util.DecodeOnboardingTicket(ticket)
+	if err != nil {
+		klog.Warningf("onboarding ticket has invalid quota %v :", err.Error())
+		return "", errTicketInvalidStorageQuota
+	}
+
+	consumerObj.Spec.StorageQuotaInGiB = int(onboardingTicket.StorageQuotaInGiB)
```
nb-ohad, line 94: Why are we decoding a ticket inside the storage consumer manager create method? We should accept the quota as a parameter for this method. 

I would expect the existing onboarding flow to already decode the ticket in order to validate it. We should use the decoded ticket from that step to get the quota.

nb-ohad (changes requested):

controllers/util/provider.go
rchikatw: Okay.

services/provider/server/consumer.go
rchikatw: Yes, I was not sure, so I checked with Leela and chose this approach. Leela said, "Just try any one you like; we can discuss it over PR."

services/provider/server/consumer.go
```diff
@@ -83,6 +84,10 @@ func (c *ocsConsumerManager) Create(ctx context.Context, onboard ifaces.StorageC
 		},
 	}
 
+	if onboardingTicket != nil {
+		consumerObj.Spec.StorageQuotaInGiB = int(onboardingTicket.StorageQuotaInGiB)
+	}
+
```
nb-ohad, line 90: Instead of an if outside of the struct you can do:
```
Spec: ocsv1alpha1.StorageConsumerSpec{
			Enable: false,
			StorageQuotaInGiB: ptr.Deref(onboardingTicket.StorageQuotaInGiB, 0),
		},
```

services/provider/server/consumer.go
```diff
@@ -54,7 +55,7 @@ func newConsumerManager(ctx context.Context, cl client.Client, namespace string)
 }
 
 // Create creates a new storageConsumer resource, updates the consumer cache and returns the storageConsumer UID
-func (c *ocsConsumerManager) Create(ctx context.Context, onboard ifaces.StorageClientOnboarding) (string, error) {
+func (c *ocsConsumerManager) Create(ctx context.Context, onboard ifaces.StorageClientOnboarding, onboardingTicket *services.OnboardingTicket) (string, error) {
```
nb-ohad, line 58: I am not sure it will benefit us to have the Consumer Manager be aware of the internal structure of the ticket. Up until this change, the ticket string was treated as an opaque string. 

```suggestion
func (c *ocsConsumerManager) Create(ctx context.Context, onboard ifaces.StorageClientOnboarding, int StorageQuotaInGiB) (string, error) {
```

nb-ohad (changes requested):

services/provider/server/server.go
```diff
@@ -106,12 +106,13 @@ func (s *OCSProviderServer) OnboardConsumer(ctx context.Context, req *pb.Onboard
...
 
-	if err := validateTicket(req.OnboardingTicket, pubKey); err != nil {
+	onboardingTicket, err := decodeAndValidateTicket(req.OnboardingTicket, pubKey)
+	if err != nil {
 		klog.Errorf("failed to validate onboarding ticket for consumer %q. %v", req.ConsumerName, err)
 		return nil, status.Errorf(codes.InvalidArgument, "onboarding ticket is not valid. %v", err)
 	}
 
-	storageConsumerUUID, err := s.consumerManager.Create(ctx, req)
+	storageConsumerUUID, err := s.consumerManager.Create(ctx, req, int(onboardingTicket.StorageQuotaInGiB))
```
nb-ohad, line 115: Why do we need that cast to int? Is the type of value on the ticket different than the type of the value on the consumer?

rchikatw: In `OnboardingTicket` we have it as uint and in storageconsumer spec we have it as int

PR #2671, opened 2024-06-20
Add RPC calls for onboarding/offboarding secondary provider

rewantsoni:
Add the RPC calls for onboarding/offboarding secondary providers, these include
1. OnboardMirrorPeer
2. OffboardMirrorPeer
3. AcknowledgeMirrorPeerOnboarding
4. GetMirroringInfo

Part of [RHSTOR-4886](https://issues.redhat.com/browse/RHSTOR-4886)

PR #2675, opened 2024-06-25
Implement the server side rpc calls for peering

rewantsoni:
Add server side implementation for RPC calls for

1. PeerStorageCluster

Depends on:
1. https://github.com/red-hat-storage/ocs-operator/pull/2669
2. https://github.com/red-hat-storage/ocs-operator/pull/2671
3. https://github.com/red-hat-storage/ocs-operator/pull/2844
4. https://github.com/red-hat-storage/ocs-operator/pull/2843

(review only the last commit)


services/provider/server/server.go
```diff
@@ -722,3 +734,135 @@ func (s *OCSProviderServer) getOCSSubscriptionChannel(ctx context.Context) (stri
...
+
+	representationID, err := s.storageClusterPeerManager.Create(ctx, req.OnboardingTicket, req.StorageClusterPeerName)
+	if err != nil {
+		if !kerrors.IsAlreadyExists(err) && err != errTicketAlreadyExistsStorageClusterPeer {
+			return nil, status.Errorf(codes.Internal, "failed to create StorageClusterPeer representation %q. %v", req.StorageClusterPeerName, err)
+		}
+
+		representation, err := s.storageClusterPeerManager.GetByName(ctx, req.StorageClusterPeerName)
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "Failed to get storageConsumer. %v", err)
```
rchikatw, line 758: failed to retrieve the storage consumer. Did you mean storageClusterPeer instead?

services/provider/server/server.go
```diff
@@ -893,6 +900,80 @@ func extractMonitorIps(data string) ([]string, error) {
...
+	}
+	for i := range storageClusterList.Items {
+		if storageClusterList.Items[i].UID == uid {
+			return &storageClusterList.Items[i], nil
+		}
+	}
+	return nil, fmt.Errorf("storage cluster with uid %q not found", uid)
+}
+
+func decodeToken(ticket string) (*services.OnboardingTicket, error) {
```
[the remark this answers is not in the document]
rewantsoni: created a different function to decode and resuing that for decodeAndValidateTicket

services/provider/server/server.go
```diff
@@ -893,6 +900,80 @@ func extractMonitorIps(data string) ([]string, error) {
...
+	err = s.storageClusterPeerManager.UpdateStorageClusterPeerStatus(ctx, storageClusterPeer, req.StorageClusterUID)
+	if err != nil {
+		klog.Error(err)
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+
+	return &pb.PeerStorageClusterResponse{StorageClusterUID: string(storageCluster.UID)}, nil
+}
+
+func (s *OCSProviderServer) findStorageClusterWithUID(ctx context.Context, uid types.UID) (*ocsv1.StorageCluster, error) {
```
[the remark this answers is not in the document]
rewantsoni: As there is a provider-server per storageCluster, we can store the storageCluster info which creating the struct

services/provider/server/storageclusterpeer.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+func newStorageClusterPeerManager(cl client.Client, namespace string) (*storageClusterPeerManager, error) {
+	return &storageClusterPeerManager{
+		client:    cl,
+		namespace: namespace,
+	}, nil
+}
+
+func (s *storageClusterPeerManager) FindStorageClusterPeerWithStorageClusterID(ctx context.Context, storageClusterUID types.UID) (*ocsv1.StorageClusterPeer, error) {
```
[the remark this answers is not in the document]
rewantsoni: updating it to be peerStorageClusterUID

services/provider/server/server.go
```diff
@@ -893,6 +900,80 @@ func extractMonitorIps(data string) ([]string, error) {
...
+		return nil, status.Errorf(codes.InvalidArgument, "onboarding ticket is not valid. %v", err)
+	}
+
+	if onboardingToken.SubjectRole != services.PeerRole {
+		err := fmt.Errorf("unsupported ticket role for StorageCluster %q, found %s, expected %s", req.StorageClusterUID, onboardingToken.SubjectRole, services.ClientRole)
+		klog.Error(err)
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+
+	storageCluster, err := s.findStorageClusterWithUID(ctx, onboardingToken.StorageClusterUID)
```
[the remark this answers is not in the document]
rewantsoni: We need to return the local storageCluster UID in the response and for that we need to storageCluster obj

services/provider/server/server.go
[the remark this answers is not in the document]
rewantsoni: using the utils function to get storageCluster, the comment is no longer relevant. resolving it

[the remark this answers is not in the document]
rewantsoni: using the utils function to get storageCluster, the comment is no longer relevant. resolving it

services/provider/server/server.go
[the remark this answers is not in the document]
rewantsoni: Using the util function to get the storageCluster, this comment is no longer relevant, resolving it

services/provider/server/server.go
```diff
@@ -925,6 +941,51 @@ func extractMonitorIps(data string) ([]string, error) {
...
+
+	klog.Infof("Found StorageClusterPeer %s for PeerStorageCluster", storageClusterPeer.Name)
+
+	//TODO: Should we store the onboarding token in storageClusterPeer?
+
+	err = s.storageClusterPeerManager.UpdateStorageClusterPeerStatus(ctx, storageClusterPeer, req.StorageClusterUID)
+	if err != nil {
+		klog.Error(err)
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
```
nb-ohad, line 988: This logic seems reverse to me. To prevent external interventions where an entity that is not the controller of a CR trying to update the status of a CR, we decided that the caller site will be the one to retrieve the information from the target and store the cluster ID of the remote on its own peer as part of the reconcile logic. 
The above seems to be the opposite, as if the caller is trying to push its storage cluster ID to the target

nb-ohad (changes requested):

services/provider/server/server.go
```diff
@@ -925,6 +941,51 @@ func extractMonitorIps(data string) ([]string, error) {
 	return ips, nil
 }
 
-func (s *OCSProviderServer) PeerStorageCluster(_ context.Context, _ *pb.PeerStorageClusterRequest) (*pb.PeerStorageClusterResponse, error) {
-	return &pb.PeerStorageClusterResponse{}, nil
+func (s *OCSProviderServer) PeerStorageCluster(ctx context.Context, req *pb.PeerStorageClusterRequest) (*pb.PeerStorageClusterResponse, error) {
+
+	pubKey, err := s.getOnboardingValidationKey(ctx)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to get public key to validate onboarding ticket for StorageCluster %q. %v", req.StorageClusterUID, err)
```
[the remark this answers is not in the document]
rewantsoni: No, every storage Cluster will have it's own onboarding validation key

services/provider/server/storageclusterpeer.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func newStorageClusterPeerManager(cl client.Client, namespace string) (*storageClusterPeerManager, error) {
+	return &storageClusterPeerManager{
+		client:    cl,
+		namespace: namespace,
+	}, nil
+}
+
+func (s *storageClusterPeerManager) FindStorageClusterPeerWithStorageClusterID(ctx context.Context, peerStorageClusterUID types.UID) (*ocsv1.StorageClusterPeer, error) {
```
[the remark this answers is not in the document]
rewantsoni: The linter fails with Uid suffix, 
```
  Error: services/provider/server/storageclusterpeer.go:27:81: var-naming: method parameter storageClusterUid should be storageClusterUID (revive)
```
Modifying it to, 
```
  func (s *storageClusterPeerManager) GetByStorageClusterUID(ctx context.Context, storageClusterUID types.UID) (*ocsv1.StorageClusterPeer, error) {
```
until linter issue is resolved

services/provider/server/server.go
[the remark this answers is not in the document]
nb-ohad: How can that be true if the function that returns the pub key does not take any kind of input?

services/provider/server/storageclusterpeer.go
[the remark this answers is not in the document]
nb-ohad: I am not sure the linter complained about `GetByStorageClusterUid`, the error seems to only about storageClusterUid. In that case please change to

```suggestion
func (s *storageClusterPeerManager) GetByStorageClusterUid(ctx context.Context, storageClusterUID types.UID) (*ocsv1.StorageClusterPeer, error) {
```

services/provider/server/server.go
```diff
@@ -584,6 +581,25 @@ func decodeAndValidateTicket(ticket string, pubKey *rsa.PublicKey) (*services.On
...
+		return nil, nil, fmt.Errorf("failed to decode onboarding ticket message: %v", err)
+	}
+
+	signature, err := base64.StdEncoding.DecodeString(ticketArr[1])
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to decode onboarding ticket signature: %v", err)
+	}
+
+	return message, signature, nil
+}
```
nb-ohad, line 601: If decode becomes its own method then there is no point in decodeAndValidae anymore as the validation part is used once. In that case,I would like for you to embed the validation part in on the call site.

services/provider/server/server.go
```diff
@@ -925,6 +941,48 @@ func extractMonitorIps(data string) ([]string, error) {
...
+
+	storageClusterPeer, err := s.storageClusterPeerManager.GetByStorageClusterUID(ctx, types.UID(req.StorageClusterUID))
+	if err != nil {
+		klog.Error(err)
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+
+	klog.Infof("Found StorageClusterPeer %s for PeerStorageCluster", storageClusterPeer.Name)
+
+	//TODO: Should we store the onboarding token in storageClusterPeer?
```
nb-ohad, line 982: I believe this TODO needs to go away

services/provider/server/server.go
```diff
@@ -925,6 +941,48 @@ func extractMonitorIps(data string) ([]string, error) {
...
+		return nil, status.Errorf(codes.Internal, "failed to get storageCluster. %v", err)
+	}
+
+	if storageCluster.UID != onboardingToken.StorageCluster {
+		return nil, status.Errorf(codes.InvalidArgument, "storageClusterUID specified on the onboarding token does not match any existing storage cluster")
+	}
+
+	klog.Infof("Found StorageCluster %s for PeerStorageCluster", storageCluster.Name)
+
+	storageClusterPeer, err := s.storageClusterPeerManager.GetByStorageClusterUID(ctx, types.UID(req.StorageClusterUID))
```
nb-ohad, line 974: You are passing the wrong storage cluster UID, you need to send the UID of the local storage cluster
```suggestion
	storageClusterPeer, err := s.storageClusterPeerManager.GetByStorageClusterUID(ctx, storageCluster.UID)
```

services/provider/server/server.go
```diff
@@ -925,6 +941,48 @@ func extractMonitorIps(data string) ([]string, error) {
...
+	if err != nil {
+		klog.Error(err)
+		return nil, status.Error(codes.InvalidArgument, err.Error())
```
nb-ohad, line 977: This should not be an Invalid argument error as the action was not derived from any argument from the request.

services/provider/server/server.go
```diff
@@ -925,6 +941,48 @@ func extractMonitorIps(data string) ([]string, error) {
...
+	}
+
+	klog.Infof("Found StorageClusterPeer %s for PeerStorageCluster", storageClusterPeer.Name)
+
+	//TODO: Should we store the onboarding token in storageClusterPeer?
+	if storageClusterPeer.Status.State == ocsv1.StorageClusterPeerStatePending || storageClusterPeer.Status.State == ocsv1.StorageClusterPeerStatePeered {
```
nb-ohad, line 983: The design doc for the feature stated `Peering` while here we use `Pending`. Which one is the correct one?

services/provider/server/server.go
```diff
@@ -584,6 +581,25 @@ func decodeAndValidateTicket(ticket string, pubKey *rsa.PublicKey) (*services.On
...
+
+	message, err := base64.StdEncoding.DecodeString(ticketArr[0])
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to decode onboarding ticket message: %v", err)
+	}
+
+	signature, err := base64.StdEncoding.DecodeString(ticketArr[1])
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to decode onboarding ticket signature: %v", err)
+	}
```
nb-ohad, line 598: Why decode the signature always, only a single call site actually cares about the signature.
It seems to me the original `decodeAndValidate` was the current approach and should be used when you need to decode and validate, and for the second call site you just want a `readStorageClusterUid` method that does not need to read the signature or validate

services/provider/server/storageclusterpeer.go
```diff
@@ -0,0 +1,49 @@
...
+		err = json.Unmarshal(message, &ticketData)
+		if err != nil {
+			return nil, fmt.Errorf("failed to unmarshal onboarding ticket message. %v", err)
+		}
+
+		if ticketData.StorageCluster == storageClusterUID {
+			return &storageClusterPeerList.Items[i], nil
+		}
+	}
+	return nil, fmt.Errorf("StorageClusterPeer linked to StorageCluster with uid %q not found", storageClusterUID)
```
nb-ohad, line 48: Please read the UID from the status and not from the ticket.

nb-ohad (changes requested):

services/provider/server/server.go
rewantsoni: Based on alogrithm, we are deriving from the req, hence it should be InvalidArgument

services/provider/server/server.go
rewantsoni: Pending is correct, it was decide during review of the API PR that Pending would make more sense

services/provider/server/storageclusterpeer.go
rewantsoni: Status won't be available till this call is a success from the controller. It will cause a deadlock where API server waits for status and the controller populates the status only after sucess of this rpc call

services/provider/server/server.go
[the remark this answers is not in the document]
rewantsoni: Opened a tech debt for it https://github.com/red-hat-storage/ocs-operator/issues/2881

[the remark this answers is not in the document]
nb-ohad: Thank you, resolving

services/provider/server/server.go
```diff
@@ -551,6 +558,11 @@ func decodeAndValidateTicket(ticket string, pubKey *rsa.PublicKey) (*services.On
 		return nil, fmt.Errorf("failed to unmarshal onboarding ticket message. %v", err)
 	}
 
+	signature, err := base64.StdEncoding.DecodeString(ticketArr[1])
+	if err != nil {
+		return nil, fmt.Errorf("failed to decode onboarding ticket %s signature: %v", ticketData.ID, err)
+	}
+
```
nb-ohad, line 565: Why move this code here?  it makes sense to validate anything we can about the content of the token preforming the second decode operation 
Can we leave this function without any changes?

services/provider/server/server.go
```diff
@@ -925,6 +932,36 @@ func extractMonitorIps(data string) ([]string, error) {
...
+		return nil, err
+	}
+
+	klog.Infof("Found StorageClusterPeer %s for PeerStorageCluster", storageClusterPeer.Name)
+
+	if storageClusterPeer.Status.State == ocsv1.StorageClusterPeerStatePending || storageClusterPeer.Status.State == ocsv1.StorageClusterPeerStatePeered {
+		return &pb.PeerStorageClusterResponse{}, nil
+	}
+
+	return nil, status.Errorf(codes.NotFound, "Peer Cluster is not reachable")
```
nb-ohad, line 966: 1. Please perfer the "successful response" as the last return in the function. A failed result should not be the last/default thing to return. 

2. Not reachable is not the same as not found. 

```suggestion
	if storageClusterPeer.Status.State != ocsv1.StorageClusterPeerStatePending && storageClusterPeer.Status.State != ocsv1.StorageClusterPeerStatePeered {
		return nil, status.Errorf(codes.NotFound, "Cannot find a storage cluster peer that meets all criterias")
	}

	return &pb.PeerStorageClusterResponse{}, nil

```

services/provider/server/server.go
```diff
@@ -925,6 +932,36 @@ func extractMonitorIps(data string) ([]string, error) {
...
+	if onboardingToken.SubjectRole != services.PeerRole {
+		err := fmt.Errorf("invalid onboarding ticket for %q, expecting role %s found role %s", req.StorageClusterUID, services.PeerRole, onboardingToken.SubjectRole)
+		klog.Error(err)
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+
+	storageClusterPeer, err := s.storageClusterPeerManager.GetByPeerStorageClusterUID(ctx, types.UID(req.StorageClusterUID))
+	if err != nil {
+		klog.Error(err)
+		return nil, err
```
nb-ohad, line 957: In a GRPC endpoint implementation, we should not blindly forward errors from other modules, mainly because you want to fully control the code that is being returned

services/provider/server/storageclusterpeer.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+	for i := range storageClusterPeerList.Items {
+		storageClusterPeer := &storageClusterPeerList.Items[i]
+		if storageClusterPeer.Status.PeerInfo != nil {
+			if storageClusterUID == types.UID(storageClusterPeer.Status.PeerInfo.StorageClusterUid) {
+				return storageClusterPeer, nil
+			}
+		}
+	}
+	return nil, status.Error(codes.NotFound, fmt.Sprintf("StorageClusterPeer linked to StorageCluster with uid %q not found", storageClusterUID))
```
nb-ohad, line 40: There is not need to return a `codes.###` error. This is an internal module and preferably we should just return normal message-based errors or a custom type that this module defines

services/provider/server/storageclusterpeer.go
```diff
@@ -0,0 +1,41 @@
...
+		client:    cl,
+		namespace: namespace,
+	}, nil
+}
+
+func (s *storageClusterPeerManager) GetByPeerStorageClusterUID(ctx context.Context, storageClusterUID types.UID) (*ocsv1.StorageClusterPeer, error) {
+	storageClusterPeerList := &ocsv1.StorageClusterPeerList{}
+	err := s.client.List(ctx, storageClusterPeerList, client.InNamespace(s.namespace))
+	if err != nil {
+		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("failed to list StorageClusterPeer: %v", err))
```
nb-ohad, line 30: There is not need to return a codes.### error. This is an internal module and preferably we should just return normal message-based errors or a custom type that this module defines

nb-ohad (changes requested):

services/provider/server/storageclusterpeer.go
```diff
@@ -0,0 +1,39 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to list StorageClusterPeer: %v", err)
+	}
+	for i := range storageClusterPeerList.Items {
+		storageClusterPeer := &storageClusterPeerList.Items[i]
+		if storageClusterPeer.Status.PeerInfo != nil {
+			if storageClusterUID == types.UID(storageClusterPeer.Status.PeerInfo.StorageClusterUid) {
+				return storageClusterPeer, nil
+			}
+		}
```
nb-ohad, line 36: No need for all of the nesting nesting

```suggestion
		peerInfo :=  storageClusterPeerList.Items[i].Status.PeerInfo
		if peerInfo != nil &&  types.UID(peerInfo.storageClusterUID) == storageClusterUID {
			return storageClusterPeer, nil
		}
```

services/provider/server/server.go
```diff
@@ -925,6 +932,37 @@ func extractMonitorIps(data string) ([]string, error) {
...
+	pubKey, err := s.getOnboardingValidationKey(ctx)
+	if err != nil {
+		klog.Errorf("failed to get public key to validate peer onboarding ticket %v", err)
+		return nil, status.Errorf(codes.Internal, "failed to validate peer onboarding ticket")
+	}
+
+	onboardingToken, err := decodeAndValidateTicket(req.OnboardingToken, pubKey)
+	if err != nil {
+		klog.Errorf("Invalid onboarding token. %v", err)
+		return nil, status.Errorf(codes.InvalidArgument, "onboarding ticket is invalid")
```
nb-ohad, line 946: Please use consistent phrasing
```suggestion
		return nil, status.Errorf(codes.InvalidArgument, "invalid onboarding ticket")
```

nb-ohad (changes requested):

PR #2676, opened 2024-06-26
Sending desired state of provider config

rchikatw:
- Sending the desired state of provider config
- If this config does not match trigger a reconciliation from status reporter that will be part of a different PR
- Also this is helpful when edit storage quota is made on a particular client.
- Testing is pending will add results once it's done
- Made changes in test cases as per the code change


services/provider/proto/provider.proto
```diff
@@ -171,4 +171,6 @@ message ReportStatusRequest{
 message ReportStatusResponse{
     // Contains subscription channel of provider operator for client operator to match
     string desiredClientOperatorChannel = 1;
+    // StorageQuotaInGiB is quota available for the client
+	int32 storageQuotaInGiB = 2;
```
rchikatw, line 175: Formatter issue appears only in the GitHub UI, not sure why. It looks fine in IDEs.

services/provider/server/server.go
```diff
@@ -700,12 +700,18 @@ func (s *OCSProviderServer) ReportStatus(ctx context.Context, req *pb.ReportStat
 		return nil, status.Errorf(codes.Internal, "Failed to update lastHeartbeat payload in the storageConsumer resource: %v", err)
 	}
 
+	stroageConsumer, err := s.consumerManager.Get(ctx, req.StorageConsumerUUID)
```
alfonsomthd, line 703:
```suggestion
	storageConsumer, err := s.consumerManager.Get(ctx, req.StorageConsumerUUID)
```

services/provider/server/server.go
```diff
@@ -747,12 +729,20 @@ func (s *OCSProviderServer) ReportStatus(ctx context.Context, req *pb.ReportStat
...
+	clusterResourceQuotaSpec := s.getClusterResourceQuotaSpec(storageConsumer)
 	channelName, err := s.getOCSSubscriptionChannel(ctx)
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "Failed to construct status response: %v", err)
 	}
 
-	return &pb.ReportStatusResponse{DesiredClientOperatorChannel: channelName}, nil
+	desiredProviderConfigHash := util.GetMD5Hash(fmt.Sprintf("%v%s", clusterResourceQuotaSpec, channelName))
+
+	return &pb.ReportStatusResponse{DesiredProviderConfigHash: desiredProviderConfigHash}, nil
```
rewantsoni, line 745: The status-reporter is using the `channelName` to set an annotation on storageClient, if we pass a hash we might need to modify that

services/provider/server/server.go
```diff
@@ -747,12 +729,20 @@ func (s *OCSProviderServer) ReportStatus(ctx context.Context, req *pb.ReportStat
...
+	}
+
+	clusterResourceQuotaSpec := s.getClusterResourceQuotaSpec(storageConsumer)
 	channelName, err := s.getOCSSubscriptionChannel(ctx)
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "Failed to construct status response: %v", err)
 	}
 
-	return &pb.ReportStatusResponse{DesiredClientOperatorChannel: channelName}, nil
+	desiredProviderConfigHash := util.GetMD5Hash(fmt.Sprintf("%v%s", clusterResourceQuotaSpec, channelName))
```
rewantsoni, line 743: should we use map[string]string and then get a hash?

rewantsoni, line 743: If we also include the hash for mon-endpoints, we could remove the [GetStorageConfig](https://github.com/red-hat-storage/ocs-client-operator/blob/main/service/status-report/main.go#L121-L143) call from status-reporter and update this endpoint in storageClient controller along with creating/Updating [clusterResourceQuota](https://github.com/red-hat-storage/ocs-client-operator/blob/main/internal/controller/storageclient_controller.go#L202-L214).

This will allow us to call only`ReportStatus` from status-reporter.

WDYT @nb-ohad @leelavg ?

rewantsoni (changes requested):

services/provider/server/server.go
leelavg: I kinda saw the fine print of the implementation in https://docs.google.com/document/d/1zJr_IWH9YeX1LukYjfh68duzyIfb7bdn3t8rI7UbQwQ/edit?disco=AAABSUNF_fE, however I believe code here isn't matching the expectations.

Let me take this opportunity to have a terser explanation of conversations that we had:

t0:
client tries to onboard onto provider, here we could send client identifiable static data (ex: clusterid, custername) to the provider and potentially remove the msg passing of this static data many times.

t1:
once the client is onboarded, the heartbeat kicks in, the original intention of heartbeat is whether client & provider is discoverable to each other or not and send any data that changes over time (mostly versions)

as a result, provider could maintain (or generate on the fly) the dynamic info that it wants to send to client (suppose only the channel name, quota, mons), however even the dynamic info is static for a longer period of time and so we want to take hash of all the dynamic info and send as heartbeat response to client.

now heartbeat just updates the hash onto storageclient status (preferably) and corresponding controller would refetch data if the status is updated.

I do not want to expand more, pls excuse if above doesn't strike anything, we can discuss here/meet and close this at the earliest.

-------

> If we also include the hash for mon-endpoints, we could remove the [GetStorageConfig](https://github.com/red-hat-storage/ocs-client-operator/blob/main/service/status-report/main.go#L121-L143) call
- I see we are sending more info than mons but client is only consuming mons and so there is not need to remove the RPC

> This will allow us to call onlyReportStatus from status-reporter.
- I get this, I believe this stemmed from having ceph-csi-config as the sink to the data received from more than one RPC call
- due to ceph-csi-op which demarcates the responsibilities of what data goes where and I would like the current RPCs to be unchanged, give it some more time on how the integration helps the overall flow.


Nevertheless, open to be corrected.

services/provider/server/server.go
rchikatw: yes i will take care in next updated commit.

services/provider/server/server.go
rchikatw: Yes, that's a good suggestion. I will use a map, which will also be helpful in sending external resources.

rchikatw: PR is updated now, Got reviews so soon which haven't expected to happen :-). @leelavg yes doing the changes as you explained the above.

services/provider/server/server.go
```diff
@@ -373,33 +373,26 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
 		extR = append(extR, &pb.ExternalResource{
 			Name: "QuotaForConsumer",
 			Kind: "ClusterResourceQuota",
-			Data: mustMarshal(clusterResourceQuotaSpec),
+			Data: mustMarshal(s.getClusterResourceQuotaSpec(consumerResource)),
 		})
 
 	}
 
+	desiredClientConfigMap, err := s.getDesiredClientConfigMap(ctx, consumerResource)
```
nb-ohad, line 385: Please replace this with inlining the data section of the desired config map inside the append call. 
(see all other places where we use append of external resources)

services/provider/server/server.go
```diff
@@ -373,33 +373,26 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to prepare desired config map %v", err)
+	}
+
+	extR = append(extR, &pb.ExternalResource{
+		Name: "DesiredClientConfig",
+		Kind: "ConfigMap",
+		Data: mustMarshal(desiredClientConfigMap),
+	})
```
nb-ohad, line 394: What is this new config map used for? If the intern is to send the hash this is the wrong mechanism. The has is a metadata of the entire operation, and should not be treated as another external resource.

services/provider/server/server.go
```diff
@@ -769,3 +765,43 @@ func (s *OCSProviderServer) getOCSSubscriptionChannel(ctx context.Context) (stri
...
+				int64(consumerResource.Spec.StorageQuotaInGiB),
+				resource.Giga,
+			)},
+		},
+	}
+
+	return *clusterResourceQuotaSpec
+}
+
+func (s *OCSProviderServer) getDesiredClientConfigMap(ctx context.Context, consumerResource *ocsv1alpha1.StorageConsumer) (map[string]string, error) {
```
nb-ohad, line 793: 1. This does not return a config map, and should not return one so the name does not fit the function. 

2. You cannot use a map for this, a map has no greatness on order so hashing it might not result in the same hash even if it has the same keys and values. Please use an array of key/value pairs, and make sure that you have a consistent order for the items in the array. 

3. Nowhere in the code there is a need to read into the content of the return value of this map/array. Any and all code outside of this function only cares about the hash of the important part of the consumer's data. Please restructure your code to reflect this - the function should return the hash value and should be named accordingly

services/provider/server/server.go
```diff
@@ -769,3 +765,43 @@ func (s *OCSProviderServer) getOCSSubscriptionChannel(ctx context.Context) (stri
 	}
 	return subscription.Spec.Channel, nil
 }
+
+func (s *OCSProviderServer) getClusterResourceQuotaSpec(consumerResource *ocsv1alpha1.StorageConsumer) quotav1.ClusterResourceQuotaSpec {
```
nb-ohad, line 769: There is no reason for this function to be attached to the `ProviderServer` object. Please make it a stateless standalone function.

services/provider/proto/provider.proto
```diff
@@ -169,6 +169,7 @@ message ReportStatusRequest{
 }
 
 message ReportStatusResponse{
-    // Contains subscription channel of provider operator for client operator to match
-    string desiredClientOperatorChannel = 1;
+    // Contains hash of desired state of client
+    string desiredClientConfigHash = 1;
```
nb-ohad, line 173: 1. Client config is a broad term that does not mean anything specifically and as such can create confusion. Add to that the fact that it does not contain a hash of the entire client configuration which means the name is also misleading.

2. There is no need to surface the fact that this is calculated as a hash as part of the API message definition. From the API perspective, it is an opaque value.

nb-ohad (changes requested):

services/provider/proto/provider.proto
rchikatw: what I can think of right now is `desiredClientState` wdyt?

services/provider/server/server.go
```diff
@@ -373,33 +373,26 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
-					resource.Giga,
-				)},
-			},
-		}
 
 		extR = append(extR, &pb.ExternalResource{
 			Name: "QuotaForConsumer",
 			Kind: "ClusterResourceQuota",
-			Data: mustMarshal(clusterResourceQuotaSpec),
+			Data: mustMarshal(s.getClusterResourceQuotaSpec(consumerResource)),
```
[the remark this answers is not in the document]
rchikatw: This is been used in 2 places now 
1. Here
2. While calculating the hash

services/provider/server/server.go
rchikatw: even with different orders, the md5 hash returns the same hash I have verified that but I agree using array. I will do the change in updated commit.

services/provider/proto/provider.proto
```diff
@@ -169,6 +169,7 @@ message ReportStatusRequest{
 }
 
 message ReportStatusResponse{
-    // Contains subscription channel of provider operator for client operator to match
-    string desiredClientOperatorChannel = 1;
+    // Contains hash of desired client state
+    string desiredClientState = 1;
```
leelavg, line 173: we are sending a representation calculated by provider and client couldn't validate it unlike the existing response like, send the channel and that will be stored on some resource.

this can just be `providerState`.

services/provider/server/server.go
```diff
@@ -390,33 +390,28 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get subscription channel %v", err)
+	}
+
+	extR = append(extR, &pb.ExternalResource{
+		Name: "operator-subs-channels",
+		Kind: "Subscription",
+		Data: mustMarshal(map[string]string{
+			"ocs-operator": channelName,
+		}),
```
leelavg, line 412: why don't you just send the subscription spec?

services/provider/proto/provider.proto
```diff
@@ -169,6 +169,7 @@ message ReportStatusRequest{
 }
 
 message ReportStatusResponse{
-    // Contains subscription channel of provider operator for client operator to match
-    string desiredClientOperatorChannel = 1;
```
leelavg, line 173: Probably No.1 rule in Proto, do not reuse field tags, reserve if old tags are being deprecated.

We have one relaxation that atm we need to support clients lagging by 1 minor version and this update breaks the compatibility.

services/provider/server/server.go
```diff
@@ -799,12 +794,25 @@ func (s *OCSProviderServer) ReportStatus(ctx context.Context, req *pb.ReportStat
...
 		return nil, status.Errorf(codes.Internal, "Failed to construct status response: %v", err)
 	}
+	var arr = make([]string, 0)
+	arr = append(arr, "DesiredClientOperatorChannel", channelName)
+	if storageConsumer.Spec.StorageQuotaInGiB > 0 {
+		clusterResourceQuotaSpec := getClusterResourceQuotaSpec(storageConsumer)
+		arr = append(arr, "ClusterResourceQuotaSpec", fmt.Sprintf("%v", clusterResourceQuotaSpec))
+	}
+
+	desiredClientStateMd5 := util.GetMD5Hash(fmt.Sprintf("%v", arr))
```
leelavg, line 813: why should this code exist here? we only wanted a hash and this doesn't need to happen inline in this response.

what I'm thinking is, have a struct with methods that collects runtime required data points and also returns the hash, wdyt?

controllers/util/provider.go
```diff
@@ -77,3 +79,8 @@ func readAndDecodePrivateKey(privateKeyPath string) (*rsa.PrivateKey, error) {
 	}
 	return privateKey, nil
 }
+
+func GetMD5Hash(text string) string {
+	hash := md5.Sum([]byte(text))
+	return hex.EncodeToString(hash[:])
+}
```
nb-ohad, line 86: This is not a provider-specific util, it is just an MD5 hashing function. Please move it to `util.go`

services/provider/proto/provider.proto
nb-ohad, line 173: You cannot remove this, we need to deprecate the field for a couple of releases and only then remove it.
You can add a `DEPRECATED: ` keyword at the start of the comment above the field

services/provider/proto/provider.proto
nb-ohad, line 173: `desiredClientState` is kind of misleading, as I would expect that this will derive from the state itself. 
Let me suggest a different name here: `DesiredConfigHash`.
The hash part in the name is important because it is not obvious that the value does not contain the state/data/config

Also, why is the reporter request designated a client/consumer ID? If not, how do we know which client this configuration belongs to?

services/provider/server/server.go
```diff
@@ -390,33 +390,28 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
-					resource.Giga,
-				)},
-			},
-		}
 
 		extR = append(extR, &pb.ExternalResource{
 			Name: "QuotaForConsumer",
 			Kind: "ClusterResourceQuota",
-			Data: mustMarshal(clusterResourceQuotaSpec),
+			Data: mustMarshal(getClusterResourceQuotaSpec(consumerResource)),
```
nb-ohad, line 397: Why is this change necessary?

services/provider/server/server.go
nb-ohad, line 412: If the kind on the response is Subscription, I would expect to have a SubscriptionSpec sent back to the client.

But the main question is why this change is part of this PR in the first place. Let's leave any changes that are not tied directly to the hash mechanism for a follow-up PR

services/provider/server/server.go
```diff
@@ -799,12 +794,25 @@ func (s *OCSProviderServer) ReportStatus(ctx context.Context, req *pb.ReportStat
...
+	storageConsumer, err := s.consumerManager.Get(ctx, req.StorageConsumerUUID)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "Failed to get storageConsumer resource: %v", err)
+	}
+
 	channelName, err := s.getOCSSubscriptionChannel(ctx)
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "Failed to construct status response: %v", err)
 	}
+	var arr = make([]string, 0)
```
nb-ohad, line 806: 1. It is better to use any when constructing the value to hash. So there will be no need to convert things to 
string
2. The structure needs to be stable, so don't use one item as a key and one item as a value. 
```suggestion 
	var arr = []any{}
```

services/provider/server/server.go
```diff
@@ -799,12 +794,25 @@ func (s *OCSProviderServer) ReportStatus(ctx context.Context, req *pb.ReportStat
...
+	arr = append(arr, "DesiredClientOperatorChannel", channelName)
+	if storageConsumer.Spec.StorageQuotaInGiB > 0 {
+		clusterResourceQuotaSpec := getClusterResourceQuotaSpec(storageConsumer)
+		arr = append(arr, "ClusterResourceQuotaSpec", fmt.Sprintf("%v", clusterResourceQuotaSpec))
+	}
```
nb-ohad, line 811: 1. You are hashing too much, there is no reason to hash the entire spec. You need only the value for the quota which comes directly from the storage consumer spec
2. You should hash the quota even if it is not set. The data to create the ash should be predictable and not conditional

nb-ohad, line 811: As of the last 3 comments, please change the code to:
```suggestion

	var arr = []any {
		channelName,
		storageConsumer.Spec.StorageQuotaInGiB,
	}

```

services/provider/server/server.go
```diff
@@ -836,3 +844,27 @@ func extractMonitorIps(data string) ([]string, error) {
...
+		Quota: corev1.ResourceQuotaSpec{
+			Hard: corev1.ResourceList{"requests.storage": *resource.NewScaledQuantity(
+				int64(consumerResource.Spec.StorageQuotaInGiB),
+				resource.Giga,
+			)},
+		},
+	}
+
+	return *clusterResourceQuotaSpec
+}
```
nb-ohad, line 870: As of my last 3 or 4 comments, this refactor is not needed

nb-ohad (changes requested):

leelavg (commented):
LGTM

services/provider/proto/provider.proto
```diff
@@ -77,8 +77,10 @@ message ExternalResource {
 
 // StorageConfigResponse holds the response for the GetStorageConfig API request
 message StorageConfigResponse{
-     // ExternalResource holds the configuration data of external storage cluster
-     repeated ExternalResource externalResource = 1;
+    // ExternalResource holds the configuration data of external storage cluster
+    repeated ExternalResource externalResource = 1;
+    // Contains hash of desired config
+    string desiredConfigHash = 2;
```
rchikatw, line 83: fields are 'optional' by default from proto3

services/provider/server/server.go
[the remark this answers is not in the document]
nb-ohad: @RishabhKodes We should not use the spec in the hash calculation. Most of the spec is static structure, we only want to use the variable info in the hash which in this case is the value for the quota

services/provider/server/server.go
nb-ohad: There are different types of hash algo, the fact the md5 is no susceptible to order changes does not mean that other hash algo aren't and we should not assume that we will keep on using md5 in the future

services/provider/proto/provider.proto
```diff
@@ -169,6 +171,9 @@ message ReportStatusRequest{
 }
 
 message ReportStatusResponse{
-    // Contains subscription channel of provider operator for client operator to match
+    // DEPRECATED: Contains subscription channel of provider operator for client operator to match
     string desiredClientOperatorChannel = 1;
```
nb-ohad, line 175: @leelavg will there be a way to get this field removed in the future?

services/provider/proto/provider.proto
rchikatw: resolving as i got LGTM in the comment of pr

services/provider/proto/provider.proto
rchikatw: resolving as i got LGTM in the comment of pr

services/provider/proto/provider.proto
```diff
@@ -169,6 +171,9 @@ message ReportStatusRequest{
 }
 
 message ReportStatusResponse{
-    // Contains subscription channel of provider operator for client operator to match
+    // DEPRECATED: Contains subscription channel of provider operator for client operator to match
```
Madhu-1, line 174: `[deprecated = true];` need to be added in front of the field to mark is deprecated.

rchikatw: Done as per the doc https://protobuf.dev/programming-guides/proto3/ so resolving.

services/provider/server/server.go
```diff
@@ -799,12 +806,27 @@ func (s *OCSProviderServer) ReportStatus(ctx context.Context, req *pb.ReportStat
...
+
 	channelName, err := s.getOCSSubscriptionChannel(ctx)
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "Failed to construct status response: %v", err)
 	}
 
-	return &pb.ReportStatusResponse{DesiredClientOperatorChannel: channelName}, nil
+	desiredClientConfigHash := getDesiredClientConfigHash(channelName, storageConsumer)
+
+	return &pb.ReportStatusResponse{DesiredClientOperatorChannel: channelName, DesiredConfigHash: desiredClientConfigHash}, nil
```
nb-ohad, line 821:
```suggestion
	return &pb.ReportStatusResponse{
		DesiredClientOperatorChannel: channelName, 
		DesiredConfigHash: desiredClientConfigHash,
	}, nil
```

services/provider/server/server.go
rchikatw: got LGTM in comments and so resolving the conv.

services/provider/proto/provider.proto
rchikatw: got LGTM in comments and so resolving the conv.

services/provider/server/server.go
rchikatw: yes done the change

services/provider/server/server.go
rchikatw: have done the changes accordingly and got LGTM from leela in comments hence resolving

services/provider/server/server.go
```diff
@@ -799,12 +806,27 @@ func (s *OCSProviderServer) ReportStatus(ctx context.Context, req *pb.ReportStat
...
+}
+
+func getDesiredClientConfigHash(channelName string, storageConsumer *ocsv1alpha1.StorageConsumer) string {
+	var arr = []any{
+		channelName,
+		storageConsumer.Spec.StorageQuotaInGiB,
+	}
+	return util.GetMD5Hash(fmt.Sprintf("%v", arr))
```
nb-ohad, line 829: using `Sprintf` with the `%v` uses space as a delimiter, this creates the following serialization bug:

both `[]any{1, "ab cd", true}` and `[]any{"1 ab cd true"}` are getting serialized to the exact same string `[a ab true]`

nb-ohad (changes requested):

services/provider/server/server.go
rchikatw: good catch, Done the change

controllers/util/util.go
```diff
@@ -81,3 +86,13 @@ func AddLabel(obj metav1.Object, key string, value string) bool {
 	}
 	return false
 }
+
+func GetMD5Hash(text any) string {
```
nb-ohad, line 90: The argument name cannot be called text if you can accept `any`. Also, this function calculates the hash so a rename we'll needed
```suggestion
func calcMD5Hash(value any) string {
```

nb-ohad (changes requested):

controllers/util/util.go
rchikatw: Done

services/provider/proto/provider.proto
leelavg: @nb-ohad yes it's possible afair, there is a `reserved ` field which takes tag numbers and throw error if used & also tells the users about tags that shouldn't be used.

leelavg:
@rchikatw I don't remember confighash in configresponse msg when I soft approved, was it added later and discussed as well?

rchikatw:
Yes 

> @rchikatw I don't remember confighash in configresponse msg when I soft approved, was it added later and discussed as well?

Over DM'd discussed with Ohad and its suggested that: send the hash as part of the config information as it's new info and not a resource so it should be the root field.
**Reason**: StorageClient controller is responsible for setting the status (ConsumerConfig- {hash,timestamp}) and we don't want to calculate hash at the client so the only way we need to send this as new info(root field).

About trigger reconcile: The status reporter will set annotation on the client and will be using `CreateOrUpdate` on PartialMetadata of storage client if something is changed then it will take care of itself to reconciling.

PR #2677, opened 2024-07-01
add new cr and controller for peering

rewantsoni:
add a new CR and controller for peering 

used the command: operator-sdk create api --group ocs --version v1 --kind StorageClusterPeer --resource --controller
to create the CR and controller


api/v1/storageclusterpeer_types.go
```diff
@@ -0,0 +1,73 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+type StorageClusterPeerState string
+
+const (
+	StorageClusterPeerFinalizer = "storageclusterpeer.ocs.openshift.io"
```
rchikatw, line 29: I don't see the finalizer setting in the PR. It would be better to remove it if we're not using it.

api/v1/storageclusterpeer_types.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+// StorageClusterPeerSpec defines the desired state of StorageClusterPeer
+type StorageClusterPeerSpec struct {
+
+	// APIServerEndpoint is the URI of the ocs api server
+	APIServerEndpoint string `json:"APIServerEndpoint"`
```
rchikatw, line 36: How about making `APIServerEndpoint` to `apiServerEndpoint` in JSON?

api/v1/storageclusterpeer_types.go
```diff
@@ -0,0 +1,73 @@
...
+	// OnboardingTicket holds an identity information required for provider to onboard.
+	OnboardingTicket string `json:"onboardingTicket"`
+}
+
+// StorageClusterPeerStatus defines the observed state of StorageClusterPeer
+type StorageClusterPeerStatus struct {
+
+	// Phase describes the Phase of StorageClusterPeer
+	// This is used by OLM UI to provide status information to the user
+	Phase StorageClusterPeerState `json:"phase,omitempty"`
```
umangachapagain, line 47: What are the possible values here?
Will this show if the connection to the peer cluster is lost?
Will this show when an existing peer needs to be re-boarded?

rewantsoni: For now, we can start with Initializing, Onboarding, Offboarding, Connected and Deleting
This will not show if the connection is lost.
I don't think we would need to re-board the existing peer

api/v1/storageclusterpeer_types.go
nb-ohad, line 36: Why do we need a "server" in the name? why not just APIEndpoint? 
Also inconsistency in the Json tag

api/v1/storageclusterpeer_types.go
nb-ohad: I am noe sure you want phase here, more and more operators are moving away from this pattern. 
My suggestion is to have it empty for now and fill it in as the implementation goes forward and you identify meaningful information to add to the status

nb-ohad (changes requested):

api/v1/storageclusterpeer_types.go
rewantsoni: I have removed the phase for now and will fill in the status with meaningful information as the implementation goes forward

api/v1/storageclusterpeer_types.go
rewantsoni: update it to "apiEndpoint"

api/v1/storageclusterpeer_types.go
```diff
@@ -0,0 +1,83 @@
...
+// RemoteClusterSpec specifies the spec required for the remote cluster
+type RemoteClusterSpec struct {
+	// APIEndpoint is the URI of the ODF api server
+	ApiEndpoint string `json:"apiEndpoint"`
+
+	// OnboardingTicket holds an identity information required by the local ODF cluster to onboard.
+	OnboardingTicket string `json:"onboardingTicket"`
+
+	// ClusterName holds the NamespacedName of the Remote ODF Cluster
+	ClusterName string `json:"clusterName"`
```
nb-ohad, line 36: houldn't the type be NamespecedName and not a string?

api/v1/storageclusterpeer_types.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+// MirrorBlockPoolSpec enables setting up of mirroring for blockPools in the same namespace.
+type MirrorBlockPoolSpec struct {
```
nb-ohad, line 40:
```suggestion
type BlockPoolMirroringSpec struct {
```

api/v1/storageclusterpeer_types.go
```diff
@@ -0,0 +1,83 @@
...
+	// LabelSelector is used to select blockPools by label
+	LabelSelector *metav1.LabelSelector `json:"labelSelector"`
```
nb-ohad, line 42:
```suggestion
	Selector *metav1.LabelSelector `json:"labelSelector"`
```

api/v1/storageclusterpeer_types.go
```diff
@@ -0,0 +1,83 @@
...
+
+	// RemoteCluster specifies the spec required for the remote cluster
+	RemoteCluster RemoteClusterSpec `json:"remoteCluster"`
+
+	// LocalClusterName holds the Name of the local ODF Cluster
+	LocalClusterName corev1.LocalObjectReference `json:"localClusterName"`
+
+	// MirrorBlockPools enables setting up of mirroring for the Ceph Block Pools in the same namespace.
+	//+optional
+	MirrorBlockPools *MirrorBlockPoolSpec `json:"mirrorBlockPools,omitempty"`
```
nb-ohad, line 56:
```suggestion
	BlockPoolMirroring * BlockPoolMirroringSpec `json:"blockPoolMirroring,omitempty"`
```

nb-ohad (changes requested):

api/v1/storageclusterpeer_types.go
rewantsoni: done

api/v1/storageclusterpeer_types.go
```diff
@@ -0,0 +1,87 @@
...
+// RemoteClusterSpec specifies the spec required for the remote cluster
+type RemoteClusterSpec struct {
+	// APIEndpoint is the URI of the ODF api server
+	ApiEndpoint string `json:"apiEndpoint"`
+
+	// OnboardingTicket holds an identity information required by the local ODF cluster to onboard.
+	OnboardingTicket string `json:"onboardingTicket"`
+
+	// ClusterName holds the namespacedName of the Remote ODF Cluster
+	ClusterName NamespacedName `json:"clusterName"`
```
nb-ohad, line 40: Lets add the stoargeclsuter context into the name, just in case we will need another field in the future for kubecluster

```suggestion
	// ClusterName holds the namespacedName of the remote storage clsuter
	StorgeClusterName NamespacedName `json:"storageClusterName"`
```

api/v1/storageclusterpeer_types.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// BlockPoolMirroringSpec enables setting up of mirroring for blockPools in the same namespace.
+type BlockPoolMirroringSpec struct {
+	// LabelSelector is used to select blockPools by label
+	Selector *metav1.LabelSelector `json:"selector"`
```
nb-ohad, line 46: Why a pointer? If we are already opting into block pool mirroring

api/v1/storageclusterpeer_types.go
```diff
@@ -0,0 +1,87 @@
...
+// StorageClusterPeerSpec defines the desired state of StorageClusterPeer
+type StorageClusterPeerSpec struct {
+
+	// RemoteCluster specifies the spec required for the remote cluster
+	RemoteCluster RemoteClusterSpec `json:"remoteCluster"`
+
+	// LocalClusterName holds the Name of the local ODF Cluster
+	LocalClusterName corev1.LocalObjectReference `json:"localClusterName"`
+
+	// BlockPoolMirroring enables setting up of mirroring for the Ceph Block Pools in the same namespace.
```
nb-ohad, line 58: Let's make the comment more meanigfull
```suggestion
	// BlockPoolMirroring indicates ceph block mirroring between block pool on the local and remote clusters
```

api/v1/storageclusterpeer_types.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// StorageClusterPeerSpec defines the desired state of StorageClusterPeer
+type StorageClusterPeerSpec struct {
+
+	// RemoteCluster specifies the spec required for the remote cluster
+	RemoteCluster RemoteClusterSpec `json:"remoteCluster"`
+
+	// LocalClusterName holds the Name of the local ODF Cluster
+	LocalClusterName corev1.LocalObjectReference `json:"localClusterName"`
```
nb-ohad, line 56: Let's try to align the spec structure between RemoteCluster and LocalCluster
 ```suggestion
	LocalCluster LocalClusterSpec `json:"localCluster"`
```

```
type LocalClusterSpec struct {
    Name: corev1.LocalObjectReference `json:"name"`
}
```

nb-ohad (changes requested):

rewantsoni:
looked at the ci failures, it is because noobaa is stuck at configuring, re-testing again.
/retest-required

PR #2678, opened 2024-07-01
storageClusterPeer: add logic for storageclusterpeer controller

rewantsoni:
add logic for storage cluster peer controller

Depends on:

1. https://github.com/red-hat-storage/ocs-operator/pull/2671
2. https://github.com/red-hat-storage/ocs-operator/pull/2677
3. https://github.com/red-hat-storage/ocs-operator/pull/2843
4. https://github.com/red-hat-storage/ocs-operator/pull/2671

(review only the last two commits)

### DIFF
--- a/api/v1/storageclusterpeer_types.go
+++ b/api/v1/storageclusterpeer_types.go
@@ -26,8 +26,10 @@ import (
 type StorageClusterPeerState string
 
 const (
-	StorageClusterPeerStatePending StorageClusterPeerState = "Pending"
-	StorageClusterPeerStatePeered  StorageClusterPeerState = "Peered"
+	StorageClusterPeerStateInitializing StorageClusterPeerState = "Initializing"
+	StorageClusterPeerStatePending      StorageClusterPeerState = "Pending"
+	StorageClusterPeerStatePeered       StorageClusterPeerState = "Peered"
+	StorageClusterPeerStateFailed       StorageClusterPeerState = "Failed"
 )
 
 type PeerInfo struct {
--- a/controllers/storagecluster/reconcile.go
+++ b/controllers/storagecluster/reconcile.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	error1 "errors"
 	"fmt"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"strings"
 	"time"
 
@@ -403,6 +404,10 @@ func (r *StorageClusterReconciler) reconcilePhases(
 		return reconcile.Result{}, nil
 	}
 
+	if res, err := r.ownStorageClusterPeersInNamespace(instance); err != nil || !res.IsZero() {
+		return reconcile.Result{}, err
+	}
+
 	// in-memory conditions should start off empty. It will only ever hold
 	// negative conditions (!Available, Degraded, Progressing)
 	r.conditions = nil
@@ -786,6 +791,29 @@ func (r *StorageClusterReconciler) isStorageClusterNotIgnored(
 	}
 
 	return true
+}
+
+func (r *StorageClusterReconciler) ownStorageClusterPeersInNamespace(instance *ocsv1.StorageCluster) (reconcile.Result, error) {
+	storageClusterPeerList := &ocsv1.StorageClusterPeerList{}
+	err := r.Client.List(r.ctx, storageClusterPeerList, client.InNamespace(instance.Namespace))
+	if err != nil {
+		return reconcile.Result{}, fmt.Errorf("failed to list storageClusterPeer: %w", err)
+	}
+	for i := range storageClusterPeerList.Items {
+		scp := &storageClusterPeerList.Items[i]
+		lenOwners := len(scp.OwnerReferences)
+		err := controllerutil.SetOwnerReference(instance, scp, r.Scheme)
+		if err != nil {
+			return reconcile.Result{}, fmt.Errorf("failed to set owner reference on StorageClusterPeer %v: %w", scp.Name, err)
+		}
+		if lenOwners != len(scp.OwnerReferences) {
+			err = r.Client.Update(r.ctx, scp)
+			if err != nil {
+				return reconcile.Result{}, fmt.Errorf("failed to persist StorageCluster owner ref on StorageClusterPeer %v: %w", scp.Name, err)
+			}
+		}
+	}
+	return reconcile.Result{}, nil
 }
 
 // Checks whether a string is contained within a slice
--- a/controllers/storageclusterpeer/storageclusterpeer_controller.go
+++ b/controllers/storageclusterpeer/storageclusterpeer_controller.go
@@ -18,13 +18,28 @@ package storageclusterpeer
 
 import (
 	"context"
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"google.golang.org/grpc/codes"
+	"strings"
+	"time"
 
+	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
+	providerClient "github.com/red-hat-storage/ocs-operator/services/provider/api/v4/client"
+	"github.com/red-hat-storage/ocs-operator/v4/controllers/util"
+	"github.com/red-hat-storage/ocs-operator/v4/services"
+
+	"github.com/go-logr/logr"
+	"google.golang.org/grpc/status"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
-
-	v1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 )
 
 // StorageClusterPeerReconciler reconciles a StorageClusterPeer object
@@ -32,32 +47,149 @@ import (
 type StorageClusterPeerReconciler struct {
 	client.Client
 	Scheme *runtime.Scheme
-}
 
-//+kubebuilder:rbac:groups=ocs.openshift.io,resources=storageclusterpeers,verbs=get;list;watch;create;update;patch;delete
-//+kubebuilder:rbac:groups=ocs.openshift.io,resources=storageclusterpeers/status,verbs=get;update;patch
-//+kubebuilder:rbac:groups=ocs.openshift.io,resources=storageclusterpeers/finalizers,verbs=update
-
-// Reconcile is part of the main kubernetes reconciliation loop which aims to
-// move the current state of the cluster closer to the desired state.
-// TODO(user): Modify the Reconcile function to compare the state specified by
-// the StorageClusterPeer object against the actual cluster state, and then
-// perform operations to make the cluster state reflect the state specified by
-// the user.
-//
-// For more details, check Reconcile and its Result here:
-// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.13.0/pkg/reconcile
-func (r *StorageClusterPeerReconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Result, error) {
-	_ = log.FromContext(ctx)
-
-	// TODO(user): your logic here
-
-	return ctrl.Result{}, nil
+	log logr.Logger
+	ctx context.Context
 }
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *StorageClusterPeerReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&v1.StorageClusterPeer{}).
+		For(&ocsv1.StorageClusterPeer{}).
+		Watches(&ocsv1.StorageCluster{}, &handler.EnqueueRequestForObject{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
 		Complete(r)
+}
+
+//+kubebuilder:rbac:groups=ocs.openshift.io,resources=storageclusterpeers,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=ocs.openshift.io,resources=storageclusterpeers/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=ocs.openshift.io,resources=storageclusterpeers/finalizers,verbs=update
+//+kubebuilder:rbac:groups=ocs.openshift.io,resources=storageclusters,verbs=get;list;watch
+
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
+// move the current state of the cluster closer to the desired state.
+func (r *StorageClusterPeerReconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
+	var err error
+	r.ctx = ctx
+	r.log = log.FromContext(ctx, "StorageClient", request)
+	r.log.Info("Reconciling StorageClusterPeer.")
+
+	// Fetch the StorageClusterPeer instance
+	storageClusterPeer := &ocsv1.StorageClusterPeer{}
+	storageClusterPeer.Name = request.Name
+	storageClusterPeer.Namespace = request.Namespace
+
+	if err = r.get(storageClusterPeer); err != nil {
+		if k8serrors.IsNotFound(err) {
+			r.log.Info("StorageClusterPeer resource not found. Ignoring since object must have been deleted.")
+			return ctrl.Result{}, nil
+		}
+		r.log.Error(err, "Failed to get StorageClusterPeer.")
+		return ctrl.Result{}, err
+	}
+
+	if storageClusterPeer.Status.State == ocsv1.StorageClusterPeerStatePeered {
+		return ctrl.Result{}, nil
+	}
+
+	result, reconcileError := r.reconcileStates(storageClusterPeer)
+
+	// Apply status changes
+	statusError := r.Client.Status().Update(r.ctx, storageClusterPeer)
+	if statusError != nil {
+		r.log.Info("Could not update StorageClusterPeer status.")
+	}
+
+	// Reconcile errors have higher priority than status update errors
+	if reconcileError != nil {
+		return result, reconcileError
+	} else if statusError != nil {
+		return result, statusError
+	}
+
+	return result, nil
+}
+
+func (r *StorageClusterPeerReconciler) reconcileStates(storageClusterPeer *ocsv1.StorageClusterPeer) (ctrl.Result, error) {
+	storageCluster := &ocsv1.StorageCluster{}
+	storageCluster.Namespace = storageClusterPeer.Namespace
+
+	// Fetch StorageCluster
+	owner := util.FindOwnerRefByKind(storageClusterPeer, "StorageCluster")
+
+	if owner == nil {
+		storageClusterPeer.Status.State = ocsv1.ReconcileFailed
+		return ctrl.Result{}, fmt.Errorf("failed to find StorgeCluster owning the StorageClusterPeer")
+	}
+
+	storageCluster.Name = owner.Name
+
+	if err := r.get(storageCluster); client.IgnoreNotFound(err) != nil {
+		r.log.Error(err, "Error fetching StorageCluster for StorageClusterPeer found in the same namespace.")
+		storageClusterPeer.Status.State = ocsv1.ReconcileFailed
+		return ctrl.Result{}, err
+	} else if k8serrors.IsNotFound(err) {
+		r.log.Error(err, "Cannot find a StorageCluster for StorageClusterPeer in the same namespace.")
+		storageClusterPeer.Status.State = ocsv1.ReconcileFailed
+		return ctrl.Result{}, nil
+	}
+
+	storageClusterPeer.Status.State = ocsv1.StorageClusterPeerStateInitializing
+
+	// Read StorageClusterUID from ticket
+	ticketArr := strings.Split(string(storageClusterPeer.Spec.OnboardingToken), ".")
+	if len(ticketArr) != 2 {
+		storageClusterPeer.Status.State = ocsv1.ReconcileFailed
+		r.log.Error(fmt.Errorf("invalid ticket"), "Invalid onboarding ticket. Does not conform to expected ticket structure")
+		return ctrl.Result{}, nil
+	}
+	message, err := base64.StdEncoding.DecodeString(ticketArr[0])
+	if err != nil {
+		storageClusterPeer.Status.State = ocsv1.ReconcileFailed
+		r.log.Error(err, "failed to decode onboarding ticket")
+		return ctrl.Result{}, nil
+	}
+	var ticketData services.OnboardingTicket
+	err = json.Unmarshal(message, &ticketData)
+	if err != nil {
+		storageClusterPeer.Status.State = ocsv1.ReconcileFailed
+		r.log.Error(err, "onboarding ticket message is not a valid JSON.")
+		return ctrl.Result{}, nil
+	}
+
+	if storageClusterPeer.Status.PeerInfo == nil {
+		storageClusterPeer.Status.PeerInfo = &ocsv1.PeerInfo{StorageClusterUid: string(ticketData.StorageCluster)}
+	}
+
+	ocsClient, err := providerClient.NewProviderClient(r.ctx, storageClusterPeer.Spec.ApiEndpoint, time.Second*10)
+	if err != nil {
+		storageClusterPeer.Status.State = ocsv1.ReconcileFailed
+		return ctrl.Result{}, fmt.Errorf("failed to create a new provider client: %v", err)
+	}
+	defer ocsClient.Close()
+
+	storageClusterPeer.Status.State = ocsv1.StorageClusterPeerStatePending
+
+	_, err = ocsClient.PeerStorageCluster(r.ctx, storageClusterPeer.Spec.OnboardingToken, string(storageCluster.UID))
+	if err != nil {
+		r.log.Error(err, fmt.Sprintf("failed to Peer Storage Cluster, reason: %v.", err))
+		st, ok := status.FromError(err)
+		if !ok {
+			storageClusterPeer.Status.State = ocsv1.ReconcileFailed
+			r.log.Error(fmt.Errorf("invalid code"), "failed to extract an HTTP status code from error")
+			return ctrl.Result{}, fmt.Errorf("failed to extract an HTTP status code from error")
+		}
+		if st.Code() == codes.InvalidArgument {
+			storageClusterPeer.Status.State = ocsv1.ReconcileFailed
+			return ctrl.Result{}, nil
+		}
+		storageClusterPeer.Status.State = ocsv1.ReconcileFailed
+	} else {
+		storageClusterPeer.Status.State = ocsv1.StorageClusterPeerStatePeered
+	}
+	return ctrl.Result{}, nil
+}
+
+func (r *StorageClusterPeerReconciler) get(obj client.Object) error {
+	key := client.ObjectKeyFromObject(obj)
+	return r.Client.Get(r.ctx, key, obj)
 }
--- a/controllers/util/k8sutil.go
+++ b/controllers/util/k8sutil.go
@@ -3,6 +3,7 @@ package util
 import (
 	"context"
 	"fmt"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"os"
 	"strings"
 
@@ -201,4 +202,14 @@ func NewK8sClient(scheme *runtime.Scheme) (client.Client, error) {
 	}
 
 	return k8sClient, nil
+}
+
+func FindOwnerRefByKind(obj client.Object, kind string) *v1.OwnerReference {
+	owners := obj.GetOwnerReferences()
+	for i := range owners {
+		if owners[i].Kind == kind {
+			return &owners[i]
+		}
+	}
+	return nil
 }
--- a/controllers/storagecluster/storagecluster_controller.go
+++ b/controllers/storagecluster/storagecluster_controller.go
@@ -245,7 +245,8 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Watches(&storagev1.StorageClass{}, enqueueStorageClusterRequest).
 		Watches(&volumesnapshotv1.VolumeSnapshotClass{}, enqueueStorageClusterRequest).
 		Watches(&ocsv1.StorageProfile{}, enqueueStorageClusterRequest).
-		Watches(&ocsv1alpha1.StorageConsumer{}, enqueueStorageClusterRequest, builder.WithPredicates(storageConsumerStatusPredicate))
+		Watches(&ocsv1alpha1.StorageConsumer{}, enqueueStorageClusterRequest, builder.WithPredicates(storageConsumerStatusPredicate)).
+		Watches(&ocsv1.StorageClusterPeer{}, enqueueStorageClusterRequest, builder.WithPredicates(predicate.GenerationChangedPredicate{}))
 
 	if os.Getenv("SKIP_NOOBAA_CRD_WATCH") != "true" {
 		build.Owns(&nbv1.NooBaa{}, builder.WithPredicates(noobaaIgnoreTimeUpdatePredicate))
--- a/metrics/vendor/github.com/red-hat-storage/ocs-operator/api/v4/v1/storageclusterpeer_types.go
+++ b/metrics/vendor/github.com/red-hat-storage/ocs-operator/api/v4/v1/storageclusterpeer_types.go
@@ -26,8 +26,10 @@ import (
 type StorageClusterPeerState string
 
 const (
-	StorageClusterPeerStatePending StorageClusterPeerState = "Pending"
-	StorageClusterPeerStatePeered  StorageClusterPeerState = "Peered"
+	StorageClusterPeerStateInitializing StorageClusterPeerState = "Initializing"
+	StorageClusterPeerStatePending      StorageClusterPeerState = "Pending"
+	StorageClusterPeerStatePeered       StorageClusterPeerState = "Peered"
+	StorageClusterPeerStateFailed       StorageClusterPeerState = "Failed"
 )
 
 type PeerInfo struct {
--- a/metrics/vendor/github.com/red-hat-storage/ocs-operator/v4/controllers/util/k8sutil.go
+++ b/metrics/vendor/github.com/red-hat-storage/ocs-operator/v4/controllers/util/k8sutil.go
@@ -3,6 +3,7 @@ package util
 import (
 	"context"
 	"fmt"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"os"
 	"strings"
 
@@ -201,4 +202,14 @@ func NewK8sClient(scheme *runtime.Scheme) (client.Client, error) {
 	}
 
 	return k8sClient, nil
+}
+
+func FindOwnerRefByKind(obj client.Object, kind string) *v1.OwnerReference {
+	owners := obj.GetOwnerReferences()
+	for i := range owners {
+		if owners[i].Kind == kind {
+			return &owners[i]
+		}
+	}
+	return nil
 }
--- a/vendor/github.com/red-hat-storage/ocs-operator/api/v4/v1/storageclusterpeer_types.go
+++ b/vendor/github.com/red-hat-storage/ocs-operator/api/v4/v1/storageclusterpeer_types.go
@@ -26,8 +26,10 @@ import (
 type StorageClusterPeerState string
 
 const (
-	StorageClusterPeerStatePending StorageClusterPeerState = "Pending"
-	StorageClusterPeerStatePeered  StorageClusterPeerState = "Peered"
+	StorageClusterPeerStateInitializing StorageClusterPeerState = "Initializing"
+	StorageClusterPeerStatePending      StorageClusterPeerState = "Pending"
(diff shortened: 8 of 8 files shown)


config/rbac/role.yaml
```diff
@@ -60,6 +60,13 @@ rules:
   - cephrbdmirrors
   verbs:
   - '*'
+- apiGroups:
+  - ceph.rook.io
+  resources:
+  - cephblockpools
+  verbs:
+  - list
```
rchikatw, line 68: verify once we might need get as well

controllers/util/k8sutil.go
```diff
@@ -42,6 +42,9 @@ const (
 
 	// This is the name for the OwnerUID FieldIndex
 	OwnerUIDIndexName = "ownerUID"
+
+	// CephBlockPoolBlackListAnnotation is used to disable cephBlockPool mirroring for RDR
+	CephBlockPoolBlackListAnnotation = "ocs.openshift.io/disable-mirroring"
```
rchikatw, line 47: CephBlockPoolBlackListAnnotation -> CephBlockPoolDisableMirroringAnnotation WDYT?

leelavg, line 47: Possible to use `allow/reject` synonyms?

api/v1/storageclusterpeer_types.go
```diff
@@ -25,6 +25,21 @@ import (
...
+	// StorageClusterPeerOnboarding represents Onboarding state of storageClusterPeer
+	StorageClusterPeerOnboarding StorageClusterPeerState = "Onboarding"
+	// StorageClusterPeerConnected represents Onboarding state of storageClusterPeer
+	StorageClusterPeerConnected StorageClusterPeerState = "Connected"
+	// StorageClusterPeerOffboarding represents Onboarding state of storageClusterPeer
+	StorageClusterPeerOffboarding StorageClusterPeerState = "Offboarding"
+	// StorageClusterPeerFailed represents Failed state of StorageClusterPeerState
+	StorageClusterPeerFailed StorageClusterPeerState = "Failed"
+	// StorageClusterPeerDeleting represents Deleting state of StorageClusterPeerState
+	StorageClusterPeerDeleting StorageClusterPeerState = "Deleting"
```
leelavg, line 40: let's use only the things that we are using, in client-op as well we weren't using many and they just add up needing to look what every state entails.

I'm still confused about phase vs state vs status vs conditions, probably this might be a good time to discuss on this (team meeting maybe) and follow a single approach?

controllers/storageclusterpeer/storageclusterpeer_controller.go
```diff
@@ -18,46 +18,488 @@ package storageclusterpeer
...
+		CreateFunc: func(createEvent event.CreateEvent) bool {
+			// Check if the annotation is not present
+			if createEvent.Object.GetAnnotations() == nil {
+				return false
+			}
+			if _, exists := createEvent.Object.GetAnnotations()[util.CephBlockPoolBlackListAnnotation]; !exists {
+				return false
+			}
+			return true
+		},
```
leelavg, line 89: why not use a value for annotation and change predicates to
``` go
CreateFunc: func(e event.CreateEvent) bool {
 return e.Object.GetAnnotations()[annotation] != ""
}
```

controllers/storageclusterpeer/storageclusterpeer_controller.go
```diff
@@ -18,46 +18,488 @@ package storageclusterpeer
...
-// the user.
-//
-// For more details, check Reconcile and its Result here:
-// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.13.0/pkg/reconcile
-func (r *StorageClusterPeerReconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Result, error) {
-	_ = log.FromContext(ctx)
+func (r *StorageClusterPeerReconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
+	var err error
+	r.ctx = ctrllog.IntoContext(ctx, r.Log)
+	r.Log.Info("Reconciling StorageClusterPeer.")
```
leelavg, line 143: in the logs, could you pls check in this line, how many times the context is being logged? I wasn't able to find the offending line and probably think supplied param `ctx` already has object and again `ctrllog.IntoContext` duplicates it.

controllers/storageclusterpeer/storageclusterpeer_controller.go
```diff
@@ -18,46 +18,488 @@ package storageclusterpeer
...
+		if errors.IsNotFound(err) {
+			r.Log.Info("StorageClusterPeer resource not found. Ignoring since object must be deleted.")
+			return reconcile.Result{}, nil
+		}
+		r.Log.Error(err, "Failed to get StorageClusterPeer.")
+		return reconcile.Result{}, err
+	}
+
+	// Don't Reconcile the StorageClusterPeer if it is in failed state
+	if r.storageClusterPeer.Status.Phase == v1.StorageClusterPeerFailed {
```
leelavg, line 160: where are you setting the phase to be failed?

config/rbac/role.yaml
rewantsoni: Tested it, we don't need get

controllers/util/k8sutil.go
rewantsoni: Updated it to use `CephBlockPoolForbidMirroringLabel = "ocs.openshift.io/forbid-mirroring"`

api/v1/storageclusterpeer_types.go
rewantsoni: Updated the status to have `RemoteState` which will denote the onboarding State for the storageClusterPeer.
We can add BlockPoolMirroringState to denote the state of BlockPoolMirroring, in the future as we add more operations for cross odf communication, we could add more states to denote the state of those operations

controllers/storageclusterpeer/storageclusterpeer_controller.go
rewantsoni: removed this

api/v1/storageclusterpeer_types.go
leelavg: I think my comment is misunderstood, I mean, I see a lot of States here are all these states being used in the code? If not, let's have only what we are using.

rewantsoni: I have updated the state to be more meaningful and reflect what they are depicting

controllers/storageclusterpeer/storageclusterpeer_controller.go
```diff
@@ -18,46 +18,520 @@ package storageclusterpeer
...
+			return reconcile.Result{}, nil
+		}
+		r.log.Error(err, "Failed to get StorageClusterPeer.")
+		return reconcile.Result{}, err
+	}
 
-	// TODO(user): your logic here
+	// Fetch local StorageCluster
+	r.storageCluster = &v1.StorageCluster{}
+	r.storageCluster.Name = r.storageClusterPeer.Spec.LocalCluster.Name.Name
```
leelavg, line 168: bit of a mouthful what does localobjectref do?

controllers/storageclusterpeer/storageclusterpeer_controller.go
```diff
@@ -18,46 +18,520 @@ package storageclusterpeer
...
+			return reconcile.Result{}, nil
+		}
+		r.log.Error(err, "Failed to get StorageCluster.")
+		return reconcile.Result{}, err
+	}
 
-	return ctrl.Result{}, nil
+	r.remoteStorageClusterName = types.NamespacedName{
+		Name:      r.storageClusterPeer.Spec.RemoteCluster.StorageClusterName.Name,
+		Namespace: r.storageClusterPeer.Spec.RemoteCluster.StorageClusterName.Namespace}
```
leelavg, line 181: ...Name.Name{,space} isn't reading well, possible to trim the `Name` suffix?

controllers/storageclusterpeer/storageclusterpeer_controller.go
```diff
@@ -18,46 +18,520 @@ package storageclusterpeer
...
-	return ctrl.NewControllerManagedBy(mgr).
-		For(&v1.StorageClusterPeer{}).
-		Complete(r)
+func (r *StorageClusterPeerReconciler) reconcilePhases() (ctrl.Result, error) {
+	ocsClient, err := r.newExternalClusterClient()
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	defer ocsClient.Close()
+
```
leelavg, line 204: the layout of this func seems a bit off, it is taking a decision around deletion at both top & inner level, possible to separate the deletion flow? Or did you try that already and hit issues?

controllers/storageclusterpeer/storageclusterpeer_controller.go
```diff
@@ -18,46 +18,520 @@ package storageclusterpeer
...
+	if r.storageClusterPeer.Status.RemoteState == "" || r.storageClusterPeer.Status.RemoteState == v1.StorageClusterPeerRemoteStateInitializing {
+		return r.onboardRemoteCluster(ocsClient)
+	} else if r.storageClusterPeer.Status.RemoteState == v1.StorageClusterPeerRemoteStateOnboarding {
+		return r.acknowledgeOnboarding(ocsClient)
+	}
+	return reconcile.Result{}, nil
+}
+
+// newExternalClusterClient returns the *providerClient.OCSProviderClient
```
leelavg, line 259: comments that are duplicating code could be removed.

controllers/storageclusterpeer/storageclusterpeer_controller.go
```diff
@@ -18,46 +18,520 @@ package storageclusterpeer
...
+		if enableMirroring {
+			cephBlockPool.Spec.Mirroring.Enabled = true
+			cephBlockPool.Spec.Mirroring.Mode = "image"
+		} else {
+			cephBlockPool.Spec.Mirroring.Enabled = false
+			cephBlockPool.Spec.Mirroring.Mode = ""
+		}
+
+		err := r.update(&cephBlockPool)
+		if err != nil {
```
leelavg, line 395: could be clubbed.

controllers/storagecluster/reconcile.go
```diff
@@ -371,6 +371,17 @@ func (r *StorageClusterReconciler) reconcilePhases(
 		// The object is marked for deletion
 		instance.Status.Phase = statusutil.PhaseDeleting
 
+		storageClusterPeerList := ocsv1.StorageClusterPeerList{}
+		err := r.List(ctx, &storageClusterPeerList, client.InNamespace(instance.Namespace))
+		if err != nil {
+			return reconcile.Result{}, err
+		}
```
leelavg, line 378: can be in same line.

controllers/storagecluster/reconcile.go
```diff
@@ -371,6 +371,17 @@ func (r *StorageClusterReconciler) reconcilePhases(
...
+		if len(storageClusterPeerList.Items) != 0 {
+			err := fmt.Errorf("waiting for %v StorageClusterPeer attached to StorageCluster to be deleted before proceeding", len(storageClusterPeerList.Items))
+			r.Log.Info(err.Error())
```
leelavg, line 381: error logging is error handling and don't do it twice, I suppose this log line will appear twice, once in this Info and another as error in reconciler.

controllers/storageclusterpeer/storageclusterpeer_controller.go
```diff
@@ -18,46 +18,176 @@ package storageclusterpeer
...
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
 
-	v1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
+const (
+	storageClusterPeerFinalizer = "storageclusterpeer.ocs.openshift.io"
 )
 
 // StorageClusterPeerReconciler reconciles a StorageClusterPeer object
 // nolint:revive
```
leelavg, line 43: nolint marker can be removed now?

controllers/storageclusterpeer/storageclusterpeer_controller.go
```diff
@@ -18,46 +18,176 @@ package storageclusterpeer
...
+	var err error
+	r.ctx = ctx
+	r.log = log.FromContext(ctx, "StorageClient", request)
+	r.log.Info("Reconciling StorageClusterPeer.")
+
+	// Fetch StorageCluster(s)
+	storageClusterList := &ocsv1.StorageClusterList{}
+	err = r.list(storageClusterList, client.InNamespace(request.Namespace))
+	if err != nil {
+		r.log.Error(err, "StorageCluster for StorageClusterPeer found in the same namespace.")
```
leelavg, line 78: is this error correct? if err is not nil then the query is failed and ntg to deduce form that, isn't it?

controllers/storageclusterpeer/storageclusterpeer_controller.go
```diff
@@ -18,46 +18,176 @@ package storageclusterpeer
...
+		return ctrl.Result{}, err
+	}
+
+	if len(storageClusterList.Items) != 1 {
```
leelavg, line 82: I recently came to know if there is more that 1 storagecluster in same ns one of the CRs will have phase ignored, should we continue to use the other healthy one?

@malayparida2000 what is expectation from user and operator side when two storageclusters exist in same ns? operator continues to reconcile healthy one and user should delete the extra one?

controllers/storageclusterpeer/storageclusterpeer_controller.go
```diff
@@ -18,46 +18,176 @@ package storageclusterpeer
...
+		err := fmt.Errorf("invalid number of StorageCluster found")
+		r.log.Error(err, "invalid number of StorageCluster(s) found, expected 1, found %v", len(storageClusterList.Items))
```
leelavg, line 84: repeated errors here?

controllers/storageclusterpeer/storageclusterpeer_controller.go
```diff
@@ -18,46 +18,176 @@ package storageclusterpeer
...
+
+	r.storageCluster = &storageClusterList.Items[0]
+
+	// Fetch the StorageClusterPeer instance
+	r.storageClusterPeer = &ocsv1.StorageClusterPeer{}
+	r.storageClusterPeer.Name = request.Name
+	r.storageClusterPeer.Namespace = request.Namespace
+
+	if err = r.get(r.storageClusterPeer); err != nil {
+		if errors.IsNotFound(err) {
```
leelavg, line 96: possible to rename errors to k8serrors as not to mistake this with stdlib?

controllers/storageclusterpeer/storageclusterpeer_controller.go
rewantsoni: the query can also fail if there is no storageCluster in the namespace (IsNotFoundError)

controllers/storageclusterpeer/storageclusterpeer_controller.go
rewantsoni: updated to use the util function

controllers/storageclusterpeer/storageclusterpeer_controller.go
rewantsoni: This is becasue the struct name starts with same as package name

leelavg: are you sure? the names are different, isn't it?

controllers/storageclusterpeer/storageclusterpeer_controller.go
leelavg: yes, that's why I'm saying the error statement is incorrect.

controllers/storageclusterpeer/storageclusterpeer_controller.go
rewantsoni: Yes, the struct name is StorageClusterPeerReconciler and package name is storageclusterpeer

controllers/storageclusterpeer/storageclusterpeer_controller.go
rewantsoni: Yes missed it, updating

controllers/storageclusterpeer/storageclusterpeer_controller.go
leelavg: Ooh, you mean `starts with.... `, missed it, I don't get what linter is trying to enforce though 😅. Resolve by adding a comment in code as well.

controllers/storageclusterpeer/storageclusterpeer_controller.go
```diff
@@ -18,46 +18,160 @@ package storageclusterpeer
...
+func (r *StorageClusterPeerReconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
+	var err error
+	r.ctx = ctx
+	r.log = log.FromContext(ctx, "StorageClient", request)
+	r.log.Info("Reconciling StorageClusterPeer.")
+
+	// Fetch StorageCluster(s)
+	r.storageCluster, err = util.GetStorageClusterInNamespace(r.ctx, r.Client, request.Namespace)
+	if err != nil && k8serrors.IsNotFound(err) {
+		r.log.Error(err, "StorageCluster for StorageClusterPeer found in the same namespace.")
```
nb-ohad, line 78: Is there a reason we need to load the storage cluster in this controller? The peer object represents the remote storage cluster not the local one.

controllers/storageclusterpeer/storageclusterpeer_controller.go
```diff
@@ -18,46 +18,160 @@ package storageclusterpeer
...
+	}
+
+	// Fetch the StorageClusterPeer instance
+	r.storageClusterPeer = &ocsv1.StorageClusterPeer{}
+	r.storageClusterPeer.Name = request.Name
+	r.storageClusterPeer.Namespace = request.Namespace
+
+	if err = r.get(r.storageClusterPeer); err != nil {
+		if k8serrors.IsNotFound(err) {
+			r.log.Info("StorageClusterPeer resource not found. Ignoring since object must be deleted.")
```
nb-ohad, line 89:
```suggestion
			r.log.Info("StorageClusterPeer resource not found. Ignoring since object must have been deleted.")
```

controllers/storageclusterpeer/storageclusterpeer_controller.go
```diff
@@ -18,46 +18,160 @@ package storageclusterpeer
...
+	if !r.storageClusterPeer.GetDeletionTimestamp().IsZero() {
+		//TODO: Removing PeerOCS Call
+
+		if controllerutil.RemoveFinalizer(r.storageClusterPeer, storageClusterPeerFinalizer) {
+			r.log.Info("removing finalizer from StorageClusterPeer.", "StorageClusterPeer", r.storageClusterPeer.Name)
+			if err := r.update(r.storageClusterPeer); err != nil {
+				r.log.Info("Failed to remove finalizer from StorageClusterPeer", "StorageClusterPeer", r.storageClusterPeer.Name)
+				return reconcile.Result{}, fmt.Errorf("failed to remove finalizer from StorageClient: %v", err)
+			}
+		}
```
nb-ohad, line 130: Why do we have a finalizer if there are no clean-up steps implemented here?
If the intent is to have this as scaffolding for future de-peering steps I would say remove it an introduce it together with the de-peering code.

controllers/storageclusterpeer/storageclusterpeer_controller.go
```diff
@@ -18,46 +18,160 @@ package storageclusterpeer
...
+	}
+
+	if controllerutil.AddFinalizer(r.storageClusterPeer, storageClusterPeerFinalizer) {
+		r.log.Info("Finalizer not found for StorageClusterPeer. Adding finalizer.", "StorageClusterPeer", r.storageClusterPeer.Name)
+		if err := r.update(r.storageClusterPeer); err != nil {
+			return reconcile.Result{}, fmt.Errorf("failed to update StorageClusterPeer: %v", err)
+		}
+	}
```
nb-ohad, line 138: As of the last comment, I don't see a reason for a finalizer at this point in time.

controllers/storageclusterpeer/storageclusterpeer_controller.go
```diff
@@ -18,46 +18,160 @@ package storageclusterpeer
...
+	if reconcileErr != nil {
+		err = reconcileErr
+	} else if statusErr != nil {
+		err = statusErr
+	}
+
+	return result, err
+}
+
+func (r *StorageClusterPeerReconciler) reconcilePhases() (ctrl.Result, error) {
```
nb-ohad, line 113: Why do we need reconcilePhases? it seems we are reconciling a single entity in a single pass.

controllers/storageclusterpeer/storageclusterpeer_controller.go
```diff
@@ -18,46 +18,160 @@ package storageclusterpeer
...
+
+	if controllerutil.AddFinalizer(r.storageClusterPeer, storageClusterPeerFinalizer) {
+		r.log.Info("Finalizer not found for StorageClusterPeer. Adding finalizer.", "StorageClusterPeer", r.storageClusterPeer.Name)
+		if err := r.update(r.storageClusterPeer); err != nil {
+			return reconcile.Result{}, fmt.Errorf("failed to update StorageClusterPeer: %v", err)
+		}
+	}
+
+	if r.storageClusterPeer.Status.State != ocsv1.StorageClusterPeerStatePeered {
+		return r.peerOCS(ocsClient)
```
nb-ohad, line 141: The heart of the reconcile is the peering algorithm, so it should be implemented in the reconcile func. There is no point in having another indirection

controllers/storageclusterpeer/storageclusterpeer_controller.go
```diff
@@ -18,46 +18,160 @@ package storageclusterpeer
...
 
 	return ctrl.Result{}, nil
 }
 
-// SetupWithManager sets up the controller with the Manager.
-func (r *StorageClusterPeerReconciler) SetupWithManager(mgr ctrl.Manager) error {
-	return ctrl.NewControllerManagedBy(mgr).
-		For(&v1.StorageClusterPeer{}).
-		Complete(r)
+func (r *StorageClusterPeerReconciler) newExternalClusterClient() (*providerClient.OCSProviderClient, error) {
```
nb-ohad, line 147: The term ExternalCluster is problematic because it has another meaning in OCS (related to external mode)

controllers/storageclusterpeer/storageclusterpeer_controller.go
```diff
@@ -18,46 +18,160 @@ package storageclusterpeer
...
+
+	ocsProviderClient, err := providerClient.NewProviderClient(
+		r.ctx, r.storageClusterPeer.Spec.ApiEndpoint, time.Second*10)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create a new provider client: %v", err)
+	}
+
+	return ocsProviderClient, nil
```
nb-ohad, line 155: This function does not have any logic at all, it just call a single `New` on providerClient, and returns the result or the error. I don't see a reason for this indirection, just instantiate the provider client directly at the call site.

controllers/storageclusterpeer/storageclusterpeer_controller.go
```diff
@@ -18,46 +18,160 @@ package storageclusterpeer
...
+}
+
+func (r *StorageClusterPeerReconciler) peerOCS(ocsClient *providerClient.OCSProviderClient) (ctrl.Result, error) {
+	r.storageClusterPeer.Status.State = ocsv1.StorageClusterPeerStatePending
```
nb-ohad, line 159: This step is not attached to the peering process but to the establishing of the connection. Please move the line just after you opened a successful connection.

controllers/storageclusterpeer/storageclusterpeer_controller.go
```diff
@@ -18,46 +18,160 @@ package storageclusterpeer
...
+	response, err := ocsClient.PeerStorageCluster(r.ctx, r.storageClusterPeer.Spec.OnboardingToken, string(r.storageCluster.UID))
```
nb-ohad, line 160: You cannot read the UID from the storage cluster object. What if the storage cluster gets deleted and another one gets created. The peer needs to be specified for a specific storage cluster instance

controllers/storageclusterpeer/storageclusterpeer_controller.go
```diff
@@ -18,46 +18,160 @@ package storageclusterpeer
...
+	if err != nil {
+		r.log.Error(err, fmt.Sprintf("failed to Peer Storage Cluster, code: %v.", status.Code(err)))
+		return ctrl.Result{}, err
+	}
+	r.storageClusterPeer.Status.PeerInfo.StorageClusterUid = response.StorageClusterUID
```
nb-ohad, line 165: PeerInfo should be a pointer. If peering is not completed there should not be a substruct on the API status.
With that in mind, the above line has the potential to throw a nil access error.

controllers/storageclusterpeer/storageclusterpeer_controller.go
```diff
@@ -18,46 +18,160 @@ package storageclusterpeer
...
+	r.storageClusterPeer.Status.State = ocsv1.StorageClusterPeerStatePeered
+	return ctrl.Result{}, nil
+}
+
+func (r *StorageClusterPeerReconciler) get(obj client.Object) error {
+	key := client.ObjectKeyFromObject(obj)
+	return r.Client.Get(r.ctx, key, obj)
+}
+
+func (r *StorageClusterPeerReconciler) update(obj client.Object, opts ...client.UpdateOption) error {
```
nb-ohad, line 175: Why do we need update? we are not updating any resources (only our local status).

nb-ohad (changes requested):

controllers/storageclusterpeer/storageclusterpeer_controller.go
rewantsoni: To prevent that from happening, we have the code to block the deletion of storageCluster until all the peers have been removed.

controllers/storagecluster/reconcile.go
```diff
@@ -375,6 +375,14 @@ func (r *StorageClusterReconciler) reconcilePhases(
...
 		instance.Status.Phase = statusutil.PhaseDeleting
 
+		storageClusterPeerList := ocsv1.StorageClusterPeerList{}
+		if err := r.List(ctx, &storageClusterPeerList, client.InNamespace(instance.Namespace)); err != nil {
+			return reconcile.Result{}, err
+		}
+		if len(storageClusterPeerList.Items) != 0 {
+			return reconcile.Result{}, fmt.Errorf("waiting for %v StorageClusterPeer attached to StorageCluster to be deleted before proceeding", len(storageClusterPeerList.Items))
+		}
+
```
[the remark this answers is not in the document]
rewantsoni: This is to prevent a new storageCluster created from being peered.

controllers/storageclusterpeer/storageclusterpeer_controller.go
rewantsoni: This is for the rpc call, we need to send the local storageCluster UID. Also, we have a peer created and no storageCluster it doesn't make sense to go ahead with the peering process

controllers/storagecluster/reconcile.go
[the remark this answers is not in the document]
nb-ohad: You cannot prevent the creation of a storage cluster from a controller, if you really want to prevent that you will need a webhook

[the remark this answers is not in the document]
rewantsoni: updated. Owning storageClusterPeer via ownerRef in storageCluster reconciler

controllers/storagecluster/reconcile.go
```diff
@@ -403,6 +404,19 @@ func (r *StorageClusterReconciler) reconcilePhases(
...
+	if err != nil {
+		return reconcile.Result{}, fmt.Errorf("failed to list storageClusterPeer: %w", err)
+	}
+	for i := range storageClusterPeerList.Items {
+		scp := &storageClusterPeerList.Items[i]
+		err := controllerutil.SetControllerReference(instance, scp, r.Scheme)
+		if err != nil {
+			return reconcile.Result{}, fmt.Errorf("failed to set controller reference on StorageClusterPeer %v: %w", scp.Name, err)
+		}
+	}
```
nb-ohad, line 418: You are setting the controller ref on an in-memory object but you never update k8s with the resource update.

controllers/storageclusterpeer/storageclusterpeer_controller.go
```diff
@@ -18,46 +18,176 @@ package storageclusterpeer
...
+	if storageClusterPeer.Status.State == ocsv1.StorageClusterPeerStatePeered {
+		return ctrl.Result{}, nil
+	}
+
+	result, reconcileError := r.reconcileStates(storageClusterPeer)
+
+	// Apply status changes
+	statusError := r.Client.Status().Update(r.ctx, storageClusterPeer)
+	if statusError != nil {
+		r.log.Info("Could not update StorageClusterPeer status.", "StorageClusterPeer", klog.KRef(storageClusterPeer.Namespace, storageClusterPeer.Name))
```
nb-ohad, line 99: You have already added the req object to all log messages (see line 73). there is no point in repeating the namespace and name on this message.

controllers/storageclusterpeer/storageclusterpeer_controller.go
```diff
@@ -18,46 +18,176 @@ package storageclusterpeer
...
+	storageCluster := &ocsv1.StorageCluster{}
+	storageCluster.Namespace = storageClusterPeer.Namespace
+
+	// Fetch StorageCluster
+	for i := range storageClusterPeer.OwnerReferences {
+		if storageClusterPeer.OwnerReferences[i].Kind == "StorageCluster" {
+			storageCluster.Name = storageClusterPeer.OwnerReferences[i].Name
+			break
+		}
+	}
```
nb-ohad, line 122: Can we make this into a general k8s util? `FindOwnerRefByKind`? It will have to be a generic util that is constant on the metav1.Object interface

controllers/storageclusterpeer/storageclusterpeer_controller.go
```diff
@@ -18,46 +18,176 @@ package storageclusterpeer
...
+
+	if err := r.get(storageCluster); err != nil {
+		r.log.Error(err, "failed to fetch StorageCluster for StorageClusterPeer found in the same namespace.")
+		if k8serrors.IsNotFound(err) {
+			storageClusterPeer.Status.State = ocsv1.ReconcileFailed
+			return ctrl.Result{Requeue: false}, nil
+		}
+		storageClusterPeer.Status.State = ocsv1.ReconcileFailed
+		return ctrl.Result{}, err
+	}
```
nb-ohad, line 132: There are 2 distinct cases here, treat them as such 
```suggestion
	if err := r.get(storageCluster); client.IgnoreNotFound(err) != nil {
		r.log.Error(err, "Error fetching StorageCluster for StorageClusterPeer found in the same namespace.")
		storageClusterPeer.Status.State = ocsv1.ReconcileFailed
		return ctrl.Result{}, err
	} else if k8serror.IsNotFound(err) {
		r.log.Error(err, "Cannot find a StorageCluster for StorageClusterPeer in the same namespace.")
		storageClusterPeer.Status.State = ocsv1.ReconcileFailed
		return ctrl.Result{}, nil
	}
```

controllers/storageclusterpeer/storageclusterpeer_controller.go
```diff
@@ -18,46 +18,176 @@ package storageclusterpeer
...
+
+	storageClusterPeer.Status.State = ocsv1.StorageClusterPeerStateInitializing
+
+	// Read StorageClusterUID from ticket
+	ticketArr := strings.Split(string(storageClusterPeer.Spec.OnboardingToken), ".")
+	if len(ticketArr) != 2 {
+		storageClusterPeer.Status.State = ocsv1.ReconcileFailed
+		r.log.Error(fmt.Errorf("invalid ticket"), "onboarding ticket has been tampered")
```
nb-ohad, line 140: You do not know that it was tempered with or if it was produced wrongfully, or even if we have a bug in the token generation code under some conditions.
```suggestion
		r.log.Error(fmt.Errorf("invalid ticket"), "Invalid onboarding ticket. Does not conform to expected ticket structure")
```

controllers/storageclusterpeer/storageclusterpeer_controller.go
```diff
@@ -18,46 +18,176 @@ package storageclusterpeer
...
+		return ctrl.Result{Requeue: false}, nil
+	}
+	message, err := base64.StdEncoding.DecodeString(ticketArr[0])
+	if err != nil {
+		storageClusterPeer.Status.State = ocsv1.ReconcileFailed
+		r.log.Error(err, "failed to decode onboarding ticket")
+		return ctrl.Result{Requeue: false}, nil
```
nb-ohad, line 147: Requeue is false by default
```suggestion
		return ctrl.Result{}, nil
```

controllers/storageclusterpeer/storageclusterpeer_controller.go
```diff
@@ -18,46 +18,176 @@ package storageclusterpeer
...
+	}
+	var ticketData services.OnboardingTicket
+	err = json.Unmarshal(message, &ticketData)
+	if err != nil {
+		storageClusterPeer.Status.State = ocsv1.ReconcileFailed
+		r.log.Error(err, "onboarding ticket message is not a valid JSON.")
+		return ctrl.Result{Requeue: false}, nil
```
nb-ohad, line 154: Same as last comment
```suggestion
		return ctrl.Result{}, nil
```

controllers/storageclusterpeer/storageclusterpeer_controller.go
```diff
@@ -18,46 +18,176 @@ package storageclusterpeer
...
+	storageClusterPeer.Status.State = ocsv1.StorageClusterPeerStatePending
+
+	_, err = ocsClient.PeerStorageCluster(r.ctx, storageClusterPeer.Spec.OnboardingToken, string(storageCluster.UID))
+	if err != nil {
+		r.log.Error(err, fmt.Sprintf("failed to Peer Storage Cluster, reason: %v.", err))
+		st, ok := status.FromError(err)
+		if !ok {
+			storageClusterPeer.Status.State = ocsv1.ReconcileFailed
+			r.log.Info("invalid code return from the rpc call")
+			return ctrl.Result{Requeue: false}, nil
```
nb-ohad, line 177: `false` is already the default
```suggestion
			return ctrl.Result{}, nil
```

nb-ohad, line 177: If you failed to get the status from the error, why do you assume that a requeue will not help?

controllers/storageclusterpeer/storageclusterpeer_controller.go
```diff
@@ -18,46 +18,176 @@ package storageclusterpeer
...
+		}
+		if st.Code() == codes.InvalidArgument {
+			storageClusterPeer.Status.State = ocsv1.ReconcileFailed
+			return ctrl.Result{Requeue: false}, nil
```
nb-ohad, line 181: false is already the default
```suggestion
			return ctrl.Result{}, nil
```

nb-ohad (changes requested):

controllers/storageclusterpeer/storageclusterpeer_controller.go
rewantsoni: Right, missed this. We should requeue if we failed to get status from the error

controllers/storageclusterpeer/storageclusterpeer_controller.go
```diff
@@ -18,46 +18,175 @@ package storageclusterpeer
...
+	storageClusterPeer.Status.State = ocsv1.StorageClusterPeerStatePending
+
+	_, err = ocsClient.PeerStorageCluster(r.ctx, storageClusterPeer.Spec.OnboardingToken, string(storageCluster.UID))
+	if err != nil {
+		r.log.Error(err, fmt.Sprintf("failed to Peer Storage Cluster, reason: %v.", err))
+		st, ok := status.FromError(err)
+		if !ok {
+			storageClusterPeer.Status.State = ocsv1.ReconcileFailed
+			r.log.Info("invalid code return from the rpc call")
+			return ctrl.Result{}, fmt.Errorf("invalid code return from the rpc call")
```
nb-ohad, line 176: Why mention the RPC call here? The issue encountered is that the error is not a StatusCode-based error. I would suggest changing the message to something like `Failed to extract an HTTP status code from error` 

Also, you should log that as an Error not as an info

nb-ohad (changes requested):

controllers/storagecluster/reconcile.go
```diff
@@ -403,6 +404,26 @@ func (r *StorageClusterReconciler) reconcilePhases(
...
+		scp := &storageClusterPeerList.Items[i]
+		lenOwners := len(scp.OwnerReferences)
+		err := controllerutil.SetOwnerReference(instance, scp, r.Scheme)
+		if err != nil {
+			return reconcile.Result{}, fmt.Errorf("failed to set owner reference on StorageClusterPeer %v: %w", scp.Name, err)
+		}
+		if lenOwners != len(scp.OwnerReferences) {
+			err = r.Client.Update(r.ctx, scp)
+			if err != nil {
+				return reconcile.Result{}, fmt.Errorf("failed to update StorageClusterPeer %v: %w", scp.Name, err)
```
nb-ohad, line 422: Your message is very generic and basically just say that an update has failed. That same context comes from the internal error returned from the client. It will be preferable if your wrapping error would offer additional context
```suggestion
				return reconcile.Result{}, fmt.Errorf("Failed to persist StorageCluster owner ref on StorageClusterPeer %v: %w", scp.Name, err)
```

nb-ohad (changes requested):

controllers/storageclusterpeer/storageclusterpeer_controller.go
```diff
@@ -18,46 +18,176 @@ package storageclusterpeer
...
+	storageCluster := &ocsv1.StorageCluster{}
+	storageCluster.Namespace = storageClusterPeer.Namespace
+
+	// Fetch StorageCluster
+	owner, err := util.FindOwnerRefByKind(storageClusterPeer, "StorageCluster")
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	storageCluster.Name = owner.Name
```
nb-ohad, line 122: owner might be nil

nb-ohad (changes requested):

PR #2682, opened 2024-07-02
Update ocs-operator direct dependencies like rook,noobaa,openshift & others

malayparida2000:


api/go.mod
```diff
@@ -1,14 +1,16 @@
 module github.com/red-hat-storage/ocs-operator/api/v4
 
-go 1.21
+go 1.22.0
+
+toolchain go1.22.4
```
iamniting, line 5: why do we need a seperate toolchain version

iamniting (changes requested):

api/go.mod
malayparida2000: It was coming up automatically after the dependency upgrades, now removed after seeing this in go documentation here https://go.dev/doc/toolchain.

> If the toolchain line is omitted, the module or workspace is considered to have an implicit toolchain goV line, where V is the Go version from the go line.
> 
> For example, a go.mod that says go 1.21.0 with no toolchain line is interpreted as if it had a toolchain go1.21.0 line.

iamniting (approved):

iamniting (approved):

malayparida2000:
https://github.com/kubernetes-sigs/controller-tools/issues/888#issuecomment-2163073485

Github action with golang 1.22 fail with Controller-gen versions below v0.13.0.

iamniting (approved):

PR #2683, opened 2024-07-02
Sending hash for storage quota

rchikatw:
- When clients request storage client config, Send a hash of the storage quota.
- Removed duplicate imports
- Renamed existing GetStorageConfig -> GetStorageClientConfig
- Renamed request/response type
- Removed unused code
- Removed test case which is no longer useful

### DIFF
--- a/api/v1alpha1/storageconsumer_types.go
+++ b/api/v1alpha1/storageconsumer_types.go
@@ -40,6 +40,9 @@ const (
 type StorageConsumerSpec struct {
 	// Enable flag ignores a reconcile if set to false
 	Enable bool `json:"enable,omitempty"`
+	// StorageQuotaInGiB describes quota for the consumer
+	// +optional
+	StorageQuotaInGiB int `json:"storageQuotaInGiB,omitempty"`
 }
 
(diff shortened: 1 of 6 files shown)


services/provider/proto/provider.proto
```diff
@@ -13,8 +13,8 @@ service OCSProvider {
   rpc OnboardConsumer (OnboardConsumerRequest)
   returns (OnboardConsumerResponse) {}
   // GetStorageConfig RPC call to generate the json config for connecting to storage provider cluster
-  rpc GetStorageConfig(StorageConfigRequest)
-  returns (StorageConfigResponse){}
+  rpc GetStorageClientConfig(StorageClientRequest)
```
leelavg, line 16: pls recheck in client-op, this is still being used, only some of the info that it is sending isn't being used.

however, I like being explicit in the naming and so for backwards compat, create a new RPC and link old to new RPC.

this could change the implementation, will review after the change.

rchikatw: Just looked at the client operator code and noticed that RPC is being used in 2 places to retrieve mon endpoints. I will add a new RPC.

PR #2685, opened 2024-07-03
Sending encoded bytes for storage quota

rchikatw:
- When clients request storage client config, Send encoded bytes of the cluster resource quota spec.
- Send the bytes only if storage quota is applicable for the client
- Testing results added https://github.com/red-hat-storage/ocs-client-operator/pull/182-

### DIFF
--- a/services/provider/server/server.go
+++ b/services/provider/server/server.go
@@ -18,6 +18,7 @@ import (
 	"time"
 
 	"github.com/blang/semver/v4"
+	quotav1 "github.com/openshift/api/quota/v1"
 	"github.com/red-hat-storage/ocs-operator/api/v4/v1alpha1"
 	ocsv1alpha1 "github.com/red-hat-storage/ocs-operator/api/v4/v1alpha1"
 	controllers "github.com/red-hat-storage/ocs-operator/v4/controllers/storageconsumer"
@@ -25,6 +26,8 @@ import (
 	pb "github.com/red-hat-storage/ocs-operator/v4/services/provider/pb"
 	ocsVersion "github.com/red-hat-storage/ocs-operator/v4/version"
 	rookCephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	opv1a1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/red-hat-storage/ocs-operator/v4/services"
@@ -369,6 +372,34 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
 		}),
 	})
 
+	if consumerResource.Spec.StorageQuotaInGiB > 0 {
+		clusterResourceQuotaSpec := &quotav1.ClusterResourceQuotaSpec{
+			Selector: quotav1.ClusterResourceQuotaSelector{
+				LabelSelector: &metav1.LabelSelector{
+					MatchExpressions: []metav1.LabelSelectorRequirement{
+						{
+							Key:      string(consumerResource.UID),
+							Operator: metav1.LabelSelectorOpDoesNotExist,
+						},
+					},
+				},
+			},
+			Quota: corev1.ResourceQuotaSpec{
+				Hard: corev1.ResourceList{"requests.storage": *resource.NewScaledQuantity(
+					int64(consumerResource.Spec.StorageQuotaInGiB),
+					resource.Giga,
+				)},
+			},
+		}
+
+		extR = append(extR, &pb.ExternalResource{
+			Name: "QuotaForConsumer",
+			Kind: "ClusterResourceQuota",
+			Data: mustMarshal(clusterResourceQuotaSpec),
+		})
+
+	}
+
 	return extR, nil
 }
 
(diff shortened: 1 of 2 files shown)


services/provider/server/server.go
```diff
@@ -367,6 +370,38 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
 		}),
 	})
 
+	if consumerResource.Spec.StorageQuotaInGiB > 0 {
+		hardLimit := resource.MustParse(fmt.Sprintf("%d%s", consumerResource.Spec.StorageQuotaInGiB, "Gi"))
+		storageQuota := &quotav1.ClusterResourceQuota{
+			ObjectMeta: metav1.ObjectMeta{Name: "ocs-storagecluster-quota"},
```
rchikatw, line 376: Please suggest a name for the ClusterResourceQuota as well

services/provider/server/server.go
```diff
@@ -367,6 +370,38 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
+			Spec: quotav1.ClusterResourceQuotaSpec{
+				Selector: quotav1.ClusterResourceQuotaSelector{
+					LabelSelector: &metav1.LabelSelector{
+						MatchExpressions: []metav1.LabelSelectorRequirement{
+							{
+								Key:      "volume.kubernetes.io/storage-provisioner",
```
rchikatw, line 382: I need to verify this and change accordingly

services/provider/server/server.go
```diff
@@ -367,6 +370,38 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
+					},
+				},
+				Quota: corev1.ResourceQuotaSpec{
+					Hard: corev1.ResourceList{"requests.storage": hardLimit},
+				},
+			},
+		}
+
+		extR = append(extR, &pb.ExternalResource{
+			Name: "QuotaForConsumer",
```
rchikatw, line 396: Please suggest a name for the ClusterResourceQuota as well

rchikatw:
@leelavg please review

services/provider/server/server.go
```diff
@@ -367,6 +370,38 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
+			Kind: "ClusterResourceQuota",
+			Data: mustMarshal(map[string]string{
+				"spec": fmt.Sprintf("%+v\n", storageQuota),
```
rchikatw, line 399: instead of sending the entire spec for ClusterResourceQuota we can send only a value for `StorageQuotaInGiB` and construct the ClusterResourceQuota on the client side. I am sending this as this is the initial PR and need your inputs

rchikatw: Sending the entire resource so that the client does not need to create or set anything, just create it.

services/provider/server/server.go
```diff
@@ -367,6 +370,37 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
+			},
+		}
+
+		extR = append(extR, &pb.ExternalResource{
+			Name: "QuotaForConsumer",
+			Kind: "ClusterResourceQuota",
+			Data: mustMarshal(map[string]string{
+				"clusterResourceQuota": fmt.Sprintf("%+v\n", clusterResourceQuota),
+			}),
+		})
```
nb-ohad, line 401: Taking all of the comments above into account the entire thing should become much simpler:

```suggestion
		extR = append(extR, &pb.ExternalResource{
			Name: "QuotaForConsumer",
			Kind: "ClusterResourceQuota",
			Data: mustMarshal(quotav1.ClusterResourceQuotaSpec{
				Quota: corev1.ResourceQuotaSpec{
					Hard: corev1.ResourceList{
						"requests.storage": resource.NewScaledQuantity(
							consumerResource.Spec.StorageQuotaInGiB,
							resource.Giga,
						),
					},
				},
			},),
		})
```

nb-ohad (changes requested):

services/provider/server/server.go
```diff
@@ -367,6 +370,35 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
+				LabelSelector: &metav1.LabelSelector{
+					MatchExpressions: []metav1.LabelSelectorRequirement{
+						{
+							Key:      "dummy",
+							Values:   []string{"dummy"},
+							Operator: metav1.LabelSelectorOpNotIn,
+						},
+					},
+				},
+			},
```
nb-ohad, line 385: Why do we need a selector here and why is the value called "dummy"?

rchikatw: ClusterResourceQuotaSelector is used to select projects.  At least one of LabelSelector or AnnotationSelector is mandatory. There is no option to select all projects. this is the one way to mitigate as of now.

I have discussed this with Leela as well and we 3 options as of now.
1. Same as above "guess" some `key-vaule` may not exist and apply the selector
2. Edit Project template if there is any in the cluster to automatically add an `annotation/label` when new project is created [project.config.openshift.io/cluster](https://red-hat-storage.github.io/ocs-training/training/ocs4/ocs4-cluster-storage-quotas.html#_creating_a_project_template_that_includes_storage_annotation)
3. Extend the webhook in the client operator to trap namespace creation and add annotation/label

PR #2686, opened 2024-07-03
Reverts deprecated ClusterClaims for odf-info

raaizik:
## Changes

- Re-adds a single `ClusterClaim` that stores the NamespacedName of the odf-info CM so that MCO will be able to fetch it.
- Skips uninstall/delete so this cluster claim will live as long as the operator does.
- [RHSTOR-5117](https://issues.redhat.com/browse/RHSTOR-5117)
- #2440 
- Depends on #2712 

## TODO

- [x] Simplify by moving `ClusterClaim` to be reconciled in `OCSInitialization` 
- [x] Delete the cluster claim once the odf-info CM is removed


controllers/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -169,6 +170,13 @@ func (r *OCSInitializationReconciler) Reconcile(ctx context.Context, request rec
 		return reconcile.Result{}, err
 	}
 
+	clusterClaim := &ocsClusterClaim{}
+	reconcileResult, err := clusterClaim.ensureCreated(r, instance)
+	if err != nil {
+		r.Log.Error(err, "Failed to ensure odf-info namespacedname ClusterClaim")
+		return reconcileResult, err
+	}
+
```
nb-ohad, line 179: This is not a full solution, as we need to trigger an event whenever an external change is forced on the claim (like someone deleting it). This means we need to add a watch as part of `SetupWithManager` but you cannot watch a type if the CRD does not exist, this means we need code to conditionally watch and another watch for the creation of the CRD itself to issue a pod restart (via panic) for cases where the CRD is introduced into the cluster after the operator is already running.

This use case, of the need to watch a CR of a CRD that might or might not exist on a cluster, becomes more and more common. We might want to consider implementation as part of utils to be used as a common implementation.

controllers/ocsinitialization/clusterclaims.go
```diff
@@ -0,0 +1,137 @@
...
+	if err != nil {
+		creator.Logger.Error(err, "Failed to get kubeconfig for ClusterClaim client.")
+		return reconcile.Result{}, err
+	}
+
+	client, err := clusterclientv1alpha1.NewForConfig(kubeconfig)
+	if err != nil {
+		creator.Logger.Error(err, "Failed to create ClusterClaim client.")
+		return reconcile.Result{}, err
+	}
```
nb-ohad, line 91: Is there really a need to create a new non-cached client? Why can't we use the client which is already installed on the reconciler object? 

As a general note, we should only use non-cached clients in cases where there is underlying reason to do so.

controllers/ocsinitialization/clusterclaims.go
```diff
@@ -0,0 +1,137 @@
...
+		creator.Logger.Info("clusterclaim unchanged", "ClusterClaim", OdfInfoNamespacedNameClaimName)
+	}
+
+	existingClaim.Spec = cc.Spec
+	_, err = client.ClusterV1alpha1().ClusterClaims().Update(creator.Context, existingClaim, metav1.UpdateOptions{})
+	if err != nil {
+		creator.Logger.Error(err, "failed to update clusterclaim", "ClusterClaim", OdfInfoNamespacedNameClaimName)
+		return reconcile.Result{}, err
+	}
+	creator.Logger.Info("updated clusterclaim", "ClusterClaim", cc.Name)
```
nb-ohad, line 126: This entire code can be replaced by a simple `CreateOrUpdate` call which will handle internally the differentiation between `Create`, `Update` and noop

nb-ohad (changes requested):

controllers/ocsinitialization/ocsinitialization_controller.go
raaizik: > We might want to consider implementation as part of utils to be used as a common implementation.

I've already added it [here](https://github.com/red-hat-storage/ocs-client-operator/pull/168). I just need to convert it to a util function.

vbnrh (commented):
@raaizik Why does the claim need to be removed manually ? Can we not set it to be removed alongside StorageCluster?

vbnrh (commented):
Can we add another claim which will indicate which clusters are client and provider for the UI ?

raaizik:
> @raaizik Why does the claim need to be removed manually ? Can we not set it to be removed alongside StorageCluster?

The cluster claims used to be reconciled via storage cluster because they contained information about it. This is no longer the case, plus the cluster claim needs to be coupled with the operator's lifecycle where it is created once and removed manually (since there's no uninstall for OCSInit).

vbnrh:
> > @raaizik Why does the claim need to be removed manually ? Can we not set it to be removed alongside StorageCluster?
> 
> @vbnrh The cluster claims used to be reconciled via storage cluster because they contained information about it. This is no longer the case, plus the cluster claim needs to be coupled with the operator's lifecycle where it is created once and removed manually (since there's no uninstall for OCSInit).

If the StorageCluster is uninstalled at a later point of time, the CM will still be present on the hub.  How are we going to indicate that cluster is not available for RDR on hub ? Are we planning or have we added the StorageCluster's status on the CM ?

raaizik:
> > > @raaizik Why does the claim need to be removed manually ? Can we not set it to be removed alongside StorageCluster?
> > 
> > 
> > @vbnrh The cluster claims used to be reconciled via storage cluster because they contained information about it. This is no longer the case, plus the cluster claim needs to be coupled with the operator's lifecycle where it is created once and removed manually (since there's no uninstall for OCSInit).
> 
> If the StorageCluster is uninstalled at a later point of time, the CM will still be present on the hub. How are we going to indicate that cluster is not available for RDR on hub ? Are we planning or have we added the StorageCluster's status on the CM ?

@vbnrh Whenever a StorageCluster is uninstalled, its respective key in the CM gets removed. If that key was the last one to get uninstalled, then the whole CM gets removed.

controllers/ocsinitialization/clusterclaims.go
```diff
@@ -0,0 +1,104 @@
...
+)
+
+const (
+	ClusterClaimCRDName = "clusterclaims.cluster.open-cluster-management.io"
+)
+
+var (
+	ClusterClaimGroup              = "odf"
+	OdfInfoNamespacedNameClaimName = fmt.Sprintf("odfinfo.%s.openshift.io", ClusterClaimGroup)
+)
```
umangachapagain, line 27:
```suggestion
const (
	ClusterClaimCRDName = "clusterclaims.cluster.open-cluster-management.io"
	OdfInfoNamespacedNameClaimName = "odfinfo.odf.openshift.io"
)
```

controllers/ocsinitialization/clusterclaims.go
```diff
@@ -0,0 +1,104 @@
...
+		Context: ctx,
+		Client:  r.Client,
+		OcsInit: instance,
+	}
+
+	operatorNamespace, err := creator.getOperatorNamespace()
+	if err != nil {
+		r.Log.Error(err, "failed to get operator's namespace. retrying again")
+		return reconcile.Result{}, err
+	}
```
umangachapagain, line 72: No need of this. Simply fetch the operator namespace and proceed. This pattern was used earlier when there were a lot of things to fetch and create.

controllers/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -252,7 +263,16 @@ func (r *OCSInitializationReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-	return ctrl.NewControllerManagedBy(mgr).
+
+	clusterClaimCrdPredicate := predicate.Funcs{
+		CreateFunc: func(e event.TypedCreateEvent[client.Object]) bool {
+			panic("ClusterClaim CRD was found. Restarting pod to initiate creation")
+		},
+		DeleteFunc: func(e event.TypedDeleteEvent[client.Object]) bool {
+			panic("ClusterClaim CRD was found. Restarting pod to initiate deletion")
+		},
+	}
```
umangachapagain, line 274: I'd like to avoid setting panic to restart the process as much as possible. There are ways to set watches dynamically, we should try those first.

controllers/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -74,6 +77,7 @@ type OCSInitializationReconciler struct {
 // +kubebuilder:rbac:groups="monitoring.coreos.com",resources={alertmanagers,prometheuses},verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups="monitoring.coreos.com",resources=servicemonitors,verbs=get;list;watch;update;patch;create;delete
 // +kubebuilder:rbac:groups=operators.coreos.com,resources=clusterserviceversions,verbs=get;list;watch;delete;update;patch
+// +kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=clusterclaims,verbs=get;list;watch;create;update;delete
```
umangachapagain, line 80: Since we are not deleting clusterclaims, can we remove the rbac for it?

controllers/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -169,6 +173,13 @@ func (r *OCSInitializationReconciler) Reconcile(ctx context.Context, request rec
 		return reconcile.Result{}, err
 	}
 
+	clusterClaim := &ocsClusterClaim{}
+	reconcileResult, err := clusterClaim.ensureCreated(r, instance)
+	if err != nil {
+		r.Log.Error(err, "Failed to ensure odf-info namespacedname ClusterClaim")
+		return reconcileResult, err
+	}
```
umangachapagain, line 181: We don't need to use this pattern. We can just create a "ensureClusteClaimExists" function and call it directly.

umangachapagain (changes requested):

controllers/ocsinitialization/ocsinitialization_controller.go
nb-ohad: @umangachapagain They don't work because they do not re-popolate the cache. The only reliable way we found was to panic out. 

@raaizik I would not use panic, I will use `os.exit` with a specific code. That way we can warp the operator invocation with a small bash command that invokes the operator process if it sees that particular exit code. 
Doing it that way will prevent unnecessary pos restarts

controllers/ocsinitialization/ocsinitialization_controller.go
nb-ohad: If the intent is to make sure it exists, we can have an `IsClusterClaimExist` function that returns a bool. 
Using an error to indicate a valid result (does not exists) is a semantic mistake

umangachapagain: `IsClusterClaimExist` is a check that returns `True` or `False`. What we want is to take action to `EnsureClusterClaimExists` and return an error if failed. That is semantically correct.

```
func ensureClusterClaimExists() error {
  if IsClusterClaimExist(){
     return nil
  }
  err := create.clusterclaim()
  return err
}
```

nb-ohad:
@umangachapagain @raaizik 
If the ODF inf config map is created when the first `storagecluster` is created and gets deleted when the last storage cluster is deleted (all as part of the storagecluster controller), why do we tie the claim's lifecycle to `ocsinitialization`, shouldn't we treat the claim and the ConfigMap as a pair? created together and deleted together?

vbnrh:
@nb-ohad @umangachapagain QE tests for MDR and RDR is getting blocked as this PR is not in 4.17

Please take a look ASAP

raaizik:
> @nb-ohad @umangachapagain QE tests for MDR and RDR is getting blocked as this PR is not in 4.17
> 
> Please take a look ASAP

It depends on #2712

controllers/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -333,8 +350,42 @@ func (r *OCSInitializationReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-		).
-		Complete(r)
+		)
+	return ocsInitializationController.Complete(r)
+}
+
+func (r *OCSInitializationReconciler) ensureClusterClaimExists() error {
+	if !r.availableCrds[ClusterClaimCrdName] || len(r.clusters.GetStorageClusters()) < 1 {
+		return nil
+	}
```
nb-ohad, line 360: 1. Can we move the CRD check outside to the call site so we would not invoke this function at all if the CRD are not avail?

2. We should not rely on the number of clusters. It can change over time and this means you will have to also manage the deletion and recreation according to storage cluster changes. It is an undesired behavior.

What is preventing us from having the cluster claim at all times?

controllers/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -333,8 +350,42 @@ func (r *OCSInitializationReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	cc := &v1alpha1.ClusterClaim{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: util.OdfInfoNamespacedNameClaimName,
+		},
+	}
+
+	_, err = controllerutil.CreateOrUpdate(r.ctx, r.Client, cc, func() error {
+		cc.Spec.Value = OdfInfoNamespacedName
+		return nil
+	})
```
nb-ohad, line 381: Who and how we are going to handle deletion of this cluster claim?

controllers/storagecluster/odfinfoconfig.go
```diff
@@ -114,6 +115,12 @@ func (obj *odfInfoConfig) ensureDeleted(r *StorageClusterReconciler, storageClus
 			r.Log.Error(err, "Failed to delete odf-info config map.", "ConfigMap", client.ObjectKeyFromObject(odfInfoConfigMap))
 			return reconcile.Result{}, fmt.Errorf("failed to delete odf-info %v: %v", odfInfoConfigMap.Name, err)
 		}
+		cc := &v1alpha1.ClusterClaim{}
+		cc.Name = util.OdfInfoNamespacedNameClaimName
+		if err = r.Client.Delete(r.ctx, cc); client.IgnoreNotFound(err) != nil {
+			r.Log.Error(err, "Failed to delete odf-info cluster claim.", "ClusterClaim", client.ObjectKeyFromObject(cc))
+			return reconcile.Result{}, fmt.Errorf("failed to delete odf-info cluster claim %v: %v", cc.Name, err)
+		}
```
nb-ohad, line 123: If ocsinitalization controller is responsible for creating the cluster claim and owns its lifecycle then it is the one who should delete it. having delete code here means that cluster claim deletion can get unsync with the initialization CR lifetime

nb-ohad (changes requested):

controllers/ocsinitialization/ocsinitialization_controller.go
raaizik: In odf-info's `ensureDeleted`

raaizik: Decided on manual deletion for now

nb-ohad:
LGTM
@iamniting @umangachapagain Do you want to take a look before merging?

umangachapagain (approved):

PR #2694, opened 2024-07-11
Fix pg_autoscaler won't scale due to overlapping roots

malayparida2000:
There was no CephBlockPool defined for the .mgr pool, therefore it was using the default crush rule (with no deviceClass specified) The CephObjectStore & CephFilesystem metadataPool does not define the deviceClass (it is only set in the dataPools).

To fix this Create a CephBlockPool CR for the .mgr pool with the same poolSpec as the default CephBlockPool. Also set the deviceClass for the metadataPool in CephObjectStore and CephFilesystem.

BZ-https://bugzilla.redhat.com/show_bug.cgi?id=2253013

### DIFF
--- a/controllers/storagecluster/cephblockpools.go
+++ b/controllers/storagecluster/cephblockpools.go
@@ -44,6 +44,8 @@ func (r *StorageClusterReconciler) addPeerSecretsToCephBlockPool(initData *ocsv1
 func (r *StorageClusterReconciler) newCephBlockPoolInstances(initData *ocsv1.StorageCluster) ([]*cephv1.CephBlockPool, error) {
 	var mirroringSpec cephv1.MirroringSpec
 	poolName := generateNameForCephBlockPool(initData)
+	// This name is used by UI to hide this pool from the list of CephBlockPools
+	builtinMgrPoolName := "builtin-mgr"
 	poolNamespace := initData.Namespace
 
 	if initData.Spec.Mirroring.Enabled {
@@ -66,6 +68,20 @@ func (r *StorageClusterReconciler) newCephBlockPoolInstances(initData *ocsv1.Sto
 					Replicated:     generateCephReplicatedSpec(initData, "data"),
 					EnableRBDStats: true,
 					Mirroring:      mirroringSpec,
+				},
+			},
+		},
+		{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      builtinMgrPoolName,
+				Namespace: poolNamespace,
+			},
+			Spec: cephv1.NamedBlockPoolSpec{
+				Name: ".mgr",
+				PoolSpec: cephv1.PoolSpec{
+					DeviceClass:   initData.Status.DefaultCephDeviceClass,
+					FailureDomain: getFailureDomain(initData),
+					Replicated:    generateCephReplicatedSpec(initData, "metadata"),
 				},
 			},
 		},
--- a/controllers/storagecluster/cephfilesystem.go
+++ b/controllers/storagecluster/cephfilesystem.go
@@ -79,6 +79,9 @@ func (r *StorageClusterReconciler) newCephFilesystemInstances(initStorageCluster
 		}
 	}
 
+	// set device class for metadata pool from the default data pool
+	ret.Spec.MetadataPool.DeviceClass = ret.Spec.DataPools[0].PoolSpec.DeviceClass
+
 	err := controllerutil.SetControllerReference(initStorageCluster, ret, r.Scheme)
 	if err != nil {
 		r.Log.Error(err, "Unable to set Controller Reference for CephFileSystem.", "CephFileSystem", klog.KRef(ret.Namespace, ret.Name))
--- a/controllers/storagecluster/cephblockpools_test.go
+++ b/controllers/storagecluster/cephblockpools_test.go
@@ -171,8 +171,8 @@ func assertCephNFSBlockPool(t *testing.T, reconciler StorageClusterReconciler, c
 	expectedAf, err := reconciler.newCephBlockPoolInstances(cr)
 	assert.NoError(t, err)
 
-	assert.Equal(t, len(expectedAf[1].OwnerReferences), 1)
+	assert.Equal(t, len(expectedAf[2].OwnerReferences), 1)
 
-	assert.Equal(t, expectedAf[1].ObjectMeta.Name, actualNFSBlockPool.ObjectMeta.Name)
-	assert.Equal(t, expectedAf[1].Spec, actualNFSBlockPool.Spec)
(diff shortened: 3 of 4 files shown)


controllers/storagecluster/cephblockpools.go
```diff
@@ -44,6 +44,7 @@ func (r *StorageClusterReconciler) addPeerSecretsToCephBlockPool(initData *ocsv1
 func (r *StorageClusterReconciler) newCephBlockPoolInstances(initData *ocsv1.StorageCluster) ([]*cephv1.CephBlockPool, error) {
 	var mirroringSpec cephv1.MirroringSpec
 	poolName := generateNameForCephBlockPool(initData)
+	builtinMgrPoolName := "builtin-mgr"
```
alfonsomthd, line 47: @malayparida2000 @travisn We'll use the pool name in the UI in order to filter out this pool and hide it in the UI, so we have to ensure that this name won't change over time.
cc @SanjalKatiyar

SanjalKatiyar: 1. Can we rely on some unique label which can be added to any pool which needs to be hidden on UI ??
2. Or, can we add a check on `Pool.Spec.Name === .mgr` for filtering in UI (assuming that it's less likely to get changed in future) ??

^ it might be okay to move forward with hard coded name as well (as for now only this pool needs to be hidden), so it's more of a suggestion/open-question.

alfonsomthd: Let's go wit the more stable/reliable identifier: be it `metadata.name`, `spec.name` or another.

malayparida2000: As of now we have just this one pool which needs to be hidden so `metadata.name` sounds like a reasonable filter to me

SanjalKatiyar: Then plz add a comment before the pool name explaining that name is used by UI... so that someone do not rename it 2-3 release from now by mistake or anything.

More generic approach (labels) is something we can revisit on need per basis in future, as UI can't keep on expanding list of hardcoded names.

malayparida2000: Added a comment.

controllers/storagecluster/cephblockpools.go
```diff
@@ -69,6 +71,22 @@ func (r *StorageClusterReconciler) newCephBlockPoolInstances(initData *ocsv1.Sto
...
+				Namespace: poolNamespace,
+			},
+			Spec: cephv1.NamedBlockPoolSpec{
+				Name: ".mgr",
+				PoolSpec: cephv1.PoolSpec{
+					DeviceClass:    initData.Status.DefaultCephDeviceClass,
+					FailureDomain:  getFailureDomain(initData),
+					Replicated:     generateCephReplicatedSpec(initData, "data"),
+					EnableRBDStats: true,
+					Mirroring:      mirroringSpec,
```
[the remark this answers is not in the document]
malayparida2000: Was looking to get this answer by @travisn about what all specs we need to keep

controllers/storagecluster/cephblockpools.go
malayparida2000: I think rook requires the name of the pool to be exactly this if I am not wrong https://github.com/rook/rook/blob/master/deploy/examples/pool-builtin-mgr.yaml

travisn: The `metadata.name` of the pool is not strictly important for this pool. As a special pool that starts with a `.`, there is actually a `spec.name` field that specifies the required `.mgr` pool name. The `metadata.name` for the pool could be anything we want, but keeping it set to a constant `builtin-mgr` is my vote, for simplicity and consistency with upstream, since we want to hide from the UI anyway.

controllers/storagecluster/cephblockpools.go
```diff
@@ -69,6 +71,22 @@ func (r *StorageClusterReconciler) newCephBlockPoolInstances(initData *ocsv1.Sto
...
+				Name:      builtinMgrPoolName,
+				Namespace: poolNamespace,
+			},
+			Spec: cephv1.NamedBlockPoolSpec{
+				Name: ".mgr",
+				PoolSpec: cephv1.PoolSpec{
+					DeviceClass:    initData.Status.DefaultCephDeviceClass,
+					FailureDomain:  getFailureDomain(initData),
+					Replicated:     generateCephReplicatedSpec(initData, "data"),
+					EnableRBDStats: true,
```
travisn, line 85: no need for rbd stats enabled on this pool
```suggestion
					EnableRBDStats: false,
```

controllers/storagecluster/cephblockpools.go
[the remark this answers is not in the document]
travisn: no need for the mirroring spec, we will never need to enable mirroring on it, and `enabled: false` is the default for the mirroring spec.
```suggestion
```

controllers/storagecluster/cephfilesystem.go
```diff
@@ -79,6 +79,9 @@ func (r *StorageClusterReconciler) newCephFilesystemInstances(initStorageCluster
 		}
 	}
 
+	// set device class for metadata pool from the default data pool
```
travisn, line 82:
```suggestion
	// set device class for metadata pool from the default metadata pool
```

controllers/storagecluster/cephblockpools.go
```diff
@@ -69,6 +71,22 @@ func (r *StorageClusterReconciler) newCephBlockPoolInstances(initData *ocsv1.Sto
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      builtinMgrPoolName,
+				Namespace: poolNamespace,
+			},
+			Spec: cephv1.NamedBlockPoolSpec{
+				Name: ".mgr",
+				PoolSpec: cephv1.PoolSpec{
+					DeviceClass:    initData.Status.DefaultCephDeviceClass,
+					FailureDomain:  getFailureDomain(initData),
+					Replicated:     generateCephReplicatedSpec(initData, "data"),
```
travisn, line 84: This is considered more of a metadata pool, rather than a data pool. Importantly, we don't want to set the `targetSizeRatio`, which is done inside `generateCephReplicatedSpec()` for data pools.
```suggestion
					Replicated:     generateCephReplicatedSpec(initData, "metadata"),
```

travisn (changes requested):

controllers/storagecluster/cephfilesystem.go
malayparida2000: By this comment I meant we are fetching the deviceClass to put in the metadata pool from the 1st data pool in the CephFS data pools list.

travisn (approved):

iamniting (approved):

PR #2695, opened 2024-07-11
Adding endpoint key for Thales enterprise KMS (KMIP)

aruniiird:
Add the missing KMS endpoint keys for Thales and Azure KMS configs to the known list of `kmsProviderAddressKeyMap`.

### DIFF
--- a/controllers/storagecluster/cephcluster.go
+++ b/controllers/storagecluster/cephcluster.go
@@ -180,14 +180,17 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
 				// reset the KMS connection's error field,
 				// it will be anyway set if there is an error
 				sc.Status.KMSServerConnection.KMSServerConnectionError = ""
-				if kmsConfigMap.Data["KMS_PROVIDER"] == "vault" {
-					sc.Status.KMSServerConnection.KMSServerAddress = kmsConfigMap.Data["VAULT_ADDR"]
-				} else if kmsConfigMap.Data["KMS_PROVIDER"] == AzureKSMProvider {
-					sc.Status.KMSServerConnection.KMSServerAddress = kmsConfigMap.Data["AZURE_VAULT_URL"]
-				} else if kmsConfigMap.Data["KMS_PROVIDER"] == ThalesKMSProvider {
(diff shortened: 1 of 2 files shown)


controllers/storagecluster/cephcluster.go
```diff
@@ -181,14 +181,17 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
...
-					sc.Status.KMSServerConnection.KMSServerAddress = kmsConfigMap.Data["KMIP_ENDPOINT"]
-				}
-				if err = reachKMSProvider(kmsConfigMap); err != nil {
+				// get the current KMS provider name
+				currKMSProvider := kmsConfigMap.Data[KMSProviderKey]
+				// according to the KMS provider, get the corresponding address/endpoint key
+				// and use it to set 'KMSServerAddress' status
+				kmsAddrssKey := kmsProviderAddressKeyMap[currKMSProvider]
+				sc.Status.KMSServerConnection.KMSServerAddress = kmsConfigMap.Data[kmsAddrssKey]
+				// if the KMS connection address is empty, log it as an error and continue
```
iamniting, line 190: @aruniiird you comment says to continue, Dont you want to continue to check the `reachKMSProvider(kmsConfigMap)`

iamniting (changes requested):
Can you pls fix the first commit msg as well

`This caused the 'reachKMSProvider'
function, which checks only the known address keys, to ignore the
provided address.
`

controllers/storagecluster/cephcluster.go
aruniiird: @iamniting ,
if the `KMSServerAddress` is empty, that means the endpoint address, which we are trying to reach, is empty, then we don't have to check/call `reachKMSProvider(kmsConfigMap)` function (as there is no address to check/reach).

> // if the KMS connection address is empty, log it as an error and continue

By logging error and continue, what I meant was to just print the error message (previously we didn't have any indication on why KMS is not working) and we continue to create the new CephCluster.

Modified the commit message, please check.

iamniting (approved):

iamniting:
Do we want this code in the 4.17 branch? Do you have a bug for this?

PR #2696, opened 2024-07-11
Always enforce the TSCs even if the placement is specified by the user

malayparida2000:
Currently TSCs are enforced only if there is no other placement spec defined for the osds & osd-prepares. Without the TSCs, the OSDs will rarely be balanced. So we should always enforce the TSCs even if the placement is specified by the user.
BZ-https://bugzilla.redhat.com/show_bug.cgi?id=2254035#c16

### DIFF
--- a/controllers/defaults/placements.go
+++ b/controllers/defaults/placements.go
@@ -45,8 +45,30 @@ var (
 			Tolerations: []corev1.Toleration{
 				getOcsToleration(),
 			},
+			// The first TSC is a hard constraint which restricts the movement of OSDs between failure domain
+			// The second TSC is a soft constraint which restricts the movement of OSDs between hosts
+			// The topology key in the first TSC is set to empty string, it should be updated to the failure domain key
 			TopologySpreadConstraints: []corev1.TopologySpreadConstraint{
-				getTopologySpreadConstraintsSpec(1, []string{osdLabelSelector}),
+				{
+					MaxSkew:           1,
+					TopologyKey:       "",
+					WhenUnsatisfiable: "DoNotSchedule",
+					LabelSelector: &metav1.LabelSelector{
+						MatchLabels: map[string]string{
+							appLabelSelectorKey: osdLabelSelector,
+						},
+					},
+				},
+				{
+					MaxSkew:           1,
+					TopologyKey:       corev1.LabelHostname,
+					WhenUnsatisfiable: "ScheduleAnyway",
+					LabelSelector: &metav1.LabelSelector{
+						MatchLabels: map[string]string{
+							appLabelSelectorKey: osdLabelSelector,
+						},
+					},
+				},
 			},
 		},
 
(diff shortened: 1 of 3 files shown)


malayparida2000:
TSC on osd

```
topologySpreadConstraints:
  - labelSelector:
      matchLabels:
        app: rook-ceph-osd
    maxSkew: 1
    topologyKey: topology.kubernetes.io/zone
    whenUnsatisfiable: DoNotSchedule
  - labelSelector:
      matchLabels:
        app: rook-ceph-osd
    maxSkew: 1
    topologyKey: kubernetes.io/hostname
    whenUnsatisfiable: ScheduleAnyway
```

TSCs on osd-prepare
```
topologySpreadConstraints:
  - labelSelector:
      matchExpressions:
      - key: app
        operator: In
        values:
        - rook-ceph-osd
        - rook-ceph-osd-prepare
    maxSkew: 1
    topologyKey: topology.kubernetes.io/zone
    whenUnsatisfiable: DoNotSchedule
  - labelSelector:
      matchExpressions:
      - key: app
        operator: In
        values:
        - rook-ceph-osd
        - rook-ceph-osd-prepare
    maxSkew: 1
    topologyKey: kubernetes.io/hostname
    whenUnsatisfiable: ScheduleAnyway
```

controllers/defaults/placements.go
```diff
@@ -45,17 +45,71 @@ var (
...
 				getOcsToleration(),
 			},
+			// The first TSC is a hard constraint which restricts the movement of OSDs between failure domain
+			// The second TSC is a soft constraint which restricts the movement of OSDs between hosts
+			// The topology key in the first TSC is set to empty string, it should be updated to the failure domain key
 			TopologySpreadConstraints: []corev1.TopologySpreadConstraint{
-				getTopologySpreadConstraintsSpec(1, []string{osdLabelSelector}),
+				{
+					MaxSkew:           1,
+					TopologyKey:       "",
```
travisn, line 54: What if `flexibleScaling: true`? Seems like in that case we should change the host-based TSC to required and not have a second TSC.

controllers/storagecluster/cephcluster.go
```diff
@@ -800,42 +794,25 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster) []rookCephv1.StorageCla
...
-					in := getPlacement(sc, "osd-prepare")
-					(&in).DeepCopyInto(&preparePlacement)
-				}
-
-				if len(topologyKeyValues) >= getMinDeviceSetReplica(sc) {
-					// Hard constraints are set in OSD placement for portable volumes with rack failure domain
-					// domain as there is no node affinity in PVs. This restricts the movement of OSDs
-					// between failure domain.
-					if portable && !strings.Contains(topologyKey, "zone") {
-						addStrictFailureDomainTSC(&placement, topologyKey)
```
travisn, line 816: `addStrictFailureDomainTSC()` added the required TSC previously, right? So where was the gap that we are trying to fix? This was ignored if the user set the placement? Does that mean we just need to call this method outside the check for `noPlacement`?

malayparida2000: No, the strict failure domain just adds another TSC(hard one on the failure domain level) to the osd placement when the failure domain is rack based. Our actual aim is to enforce the TSCs always even when user has specified some placement spec.

controllers/defaults/placements.go
malayparida2000: Hm, right. We have basically 3 types of failure domains & we need to think of proper spec for each time.
As far as I have understood I am summarizing the requirement below, do correct me if I am wrong anywhere.

```
If the failure domain is zone
We need soft constraints on hostname key, We don't want to put hard constraint on failure domain level.

For rack failure domain
We need hard constraint on rack level, with soft constraint on hostname level. 

For hostname failure domain(flexible scaling)
We need hard constraint on hostname level
```

travisn: > If the failure domain is zone
> We need soft constraints on hostname key, We don't want to put hard constraint on failure domain level.

Why wouldn't we put a hard constraint on the zone level? I thought we always needed a hard constraint for the top level failure domain.

malayparida2000: That was my expectation too, but if you see here in the old code(which was written around 4 years back), Adding hard constraint on zone level is specifically avoided which I am not sure why

https://github.com/red-hat-storage/ocs-operator/blob/main/controllers/storagecluster/cephcluster.go#L815

malayparida2000: I found this little discussion where you raised the same question https://github.com/red-hat-storage/ocs-operator/pull/1043#discussion_r573147451

malayparida2000: My opinion would be keeping a consistent behaviour in this way
1. Hard constraint on the failure domain level
2. soft constraint on the hostname level

travisn: Wow nice old thread! :) Ok sounds good to keep it as previously with the soft constraint on zones as described there.

travisn (commented):
The logic looks good. Can you just confirm when you could test it for the different failure domains? All the placement changes still just make me nervous until it's tested.

malayparida2000:
Replaced by #2720

PR #2697, opened 2024-07-15
Adding StorageQuotaUtilizationRatio field for storage consumer api

rchikatw:
- This field will be useful in showing the Storage quota utilization of client in the UI
- Value for this will be set from status section of cluster resource quota resource of client 
```
status:
  total:
    used:
      requests.storage: 15Gi
```
- status reporter is responsible for setting the value for StorageQuotaUtilization field


api/v1alpha1/storageconsumer_types.go
```diff
@@ -90,6 +90,10 @@ type ClientStatus struct {
 	// Name is the name of connected storageclient
 	// +optional
 	Name string `json:"name,omitempty"`
+
+	// UsedCapacityInGiB is the used capacity of connected storageclient
+	// +optional
+	UsedCapacityInGiB string `json:"usedCapacityInGiB,omitempty"`
```
leelavg, line 96: what resources on client side provides this value?

rchikatw: The value for this will be available in status section of `ClusterResourceQuota` CR. https://docs.openshift.com/container-platform/4.15/rest_api/schedule_and_quota_apis/clusterresourcequota-quota-openshift-io-v1.html#status 
We have used a section that will display the used capacity from the allocated capacity.
Ex:
```
status:
  total:
    hard:
      requests.storage: 17Gi
    used:
      requests.storage: 15Gi

```

rchikatw: I submitted this PR as a work in progress because I accidentally set the type as a string instead of an int. I'm currently working on fixing it.

rchikatw:
@nb-ohad can i get review on this.

api/v1alpha1/storageconsumer_types.go
nb-ohad: @rchikatw The name used capacity seems misleading here. `Used` represents a portion of something, but the name does not represent the "what". So when I see the term used capacity there is a question of used of what. Is it from total cluster capacity? from the quota? from ODF available storage? 

Because this is supposed to reflect the storage quota utilization, I will prefer if we will call it something like:
`StorageQoutaUtilizationInGB and maybe we should consider to report it as a ratio and not as a number

WDYT?

rchikatw: I have updated the name to `StorageQuotaUtilizationInGiB`. I don't think using ratio is a good choice for displaying storage utilization in the UI. The UI would have to perform the calculation to display it in the column(In GiB). The ratio would be useful only for triggering alerts, so it would be better to keep that logic in the backend.

nb-ohad: Doesn't the UI need to convert that back into a ration? they are going to create a filled bar out of the value, and they already have the quota itself, so converting a ratio back into a quota is just 1 multiplication. I would even say that displaying a ratio makes more sense to me then displaying a value. We can have a discussion with the UX eng and then settle on a representation of the value.

bipuladh: The bar will need to report values also. On hover we could show the actual values.

rchikatw: As per our chat discussion, we have decided to store the ratio in the storage consumer status section. The UI team will determine whether to display it as a bar or in GIB based on the UI requirements.

api/v1alpha1/storageconsumer_types.go
```diff
@@ -90,6 +90,10 @@ type ClientStatus struct {
 	// Name is the name of connected storageclient
 	// +optional
 	Name string `json:"name,omitempty"`
+
+	// StorageQuotaUtilizationRatio is the ratio of utilized quota of connected client
+	// +optional
+	StorageQuotaUtilizationRatio string `json:"storageQuotaUtilizationRatio,omitempty"`
```
nb-ohad, line 96: Why a string? we want a ratio which means a floating point number

services/provider/api/interfaces/interfaces.go
```diff
@@ -9,12 +9,14 @@ type StorageClientStatus interface {
 	GetClusterID() string
 	GetClusterName() string
 	GetClientName() string
+	GetStorageQuotaUtilizationRatio() string
```
nb-ohad, line 12: Why a string? we want a ratio which means a floating point number

services/provider/api/interfaces/interfaces.go
```diff
@@ -9,12 +9,14 @@ type StorageClientStatus interface {
...
 
 	SetPlatformVersion(string) StorageClientStatus
 	SetOperatorVersion(string) StorageClientStatus
 	SetClusterID(string) StorageClientStatus
 	SetClusterName(string) StorageClientStatus
 	SetClientName(string) StorageClientStatus
+	SetStorageQuotaUtilizationRatio(string) StorageClientStatus
```
nb-ohad, line 19: Why a string? we want a ratio which means a floating point number

services/provider/proto/provider.proto
```diff
@@ -171,6 +171,8 @@ message ReportStatusRequest{
     string clusterName = 5;
     // clientName is the name of the connected storageclient
     string clientName = 6;
+    // storageQuotaUtilizationRatio is the ratio of utilized quota of connected client
+    string storageQuotaUtilizationRatio = 7;
```
nb-ohad, line 175: Why a string? we want a ratio which means a floating point number

nb-ohad (changes requested):

services/provider/api/interfaces/interfaces.go
rchikatw: The reason for choosing the string for this is i saw the platform version is string even when its floating so i made it string. Now i checked its values and saw it has multiple decimal points in it(like `4.16.0 / 4.17.0`). I will change to `float32`

api/v1alpha1/storageconsumer_types.go
rchikatw: The reason for choosing the string for this is i saw the platform version is string even when its floating so i made it string. Now i checked its values and saw it has multiple decimal points in it(like `4.16.0 / 4.17.0`). I will change to `float32`

services/provider/proto/provider.proto
rchikatw: The reason for choosing the string for this is i saw the platform version is string even when its floating so i made it string. Now i checked its values and saw it has multiple decimal points in it(like `4.16.0 / 4.17.0`). I will change to `float32`

PR #2698, opened 2024-07-16
Migrate odf-info structs to API folder

raaizik:
## Changes

Converts odf-info structs into its own API
#2440


nb-ohad:
A couple of points:
* As suggested by @umangachapagain in an offline discussion. A better name for the file containing the new type would be `odfinfo_config.go`. Mainly because the `_type` suffix is usually reserved for k8s API types.  
* This is a new API and we have no indication of its stability. It is better we have it in the v1alpha1 folder
* The second commit seems unrelated to this change and I would request to separate it into its own PR

raaizik:
> A couple of points:
> 
>     * As suggested by @umangachapagain in an offline discussion. A better name for the file containing the new type would be `odfinfo_config.go`. Mainly because the `_type` suffix is usually reserved for k8s API types.
> 
>     * This is a new API and we have no indication of its stability. It is better we have it in the v1alpha1 folder
> 
>     * The second commit seems unrelated to this change and I would request to separate it into its own PR

Done @nb-ohad

PR #2702, opened 2024-07-18
external: enable log rotation and log collection for external mode

parth-gr:
with 4.17 csi pods will be collecting logs by the
log rtoation setting, so even if its external mode turn on the log collector setting

### DIFF
--- a/controllers/storagecluster/cephcluster.go
+++ b/controllers/storagecluster/cephcluster.go
@@ -192,15 +192,9 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
 					return reconcile.Result{}, err
 				}
 			}
-			cephCluster, err = newCephCluster(sc, r.images.Ceph, kmsConfigMap, r.Log)
-			if err != nil {
-				return reconcile.Result{}, err
-			}
+			cephCluster = newCephCluster(sc, r.images.Ceph, kmsConfigMap, r.Log)
 		} else {
-			cephCluster, err = newCephCluster(sc, r.images.Ceph, nil, r.Log)
-			if err != nil {
-				return reconcile.Result{}, err
-			}
+			cephCluster = newCephCluster(sc, r.images.Ceph, nil, r.Log)
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


malayparida2000:
Changes look good but please take a look at the failing unit test due to the change of function signature of newExternalCephCluster function.

parth-gr:
> Changes look good but please take a look at the failing unit test due to the change of function signature of newExternalCephCluster function.

Done

malayparida2000 (approved):
/lgtm

controllers/storagecluster/cephcluster.go
```diff
@@ -605,11 +608,22 @@ func getNetworkSpec(sc ocsv1.StorageCluster) rookCephv1.NetworkSpec {
 	return networkSpec
 }
 
-func newExternalCephCluster(sc *ocsv1.StorageCluster, monitoringIP, monitoringPort string) *rookCephv1.CephCluster {
+func newExternalCephCluster(sc *ocsv1.StorageCluster, monitoringIP, monitoringPort string) (*rookCephv1.CephCluster, error) {
 	labels := map[string]string{
 		"app": sc.Name,
 	}
 
+	maxLogSize, err := resource.ParseQuantity("500Mi")
```
iamniting, line 616: I think we should use must parse, As this is set in the controller itself and can not be overridden by the user. Same for the internal mode as well.

iamniting (changes requested):

controllers/storagecluster/cephcluster.go
parth-gr: Okay, make sense

controllers/storagecluster/cephcluster.go
```diff
@@ -411,10 +411,7 @@ func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, kmsConfigMap *co
 		"app": sc.Name,
 	}
 
-	maxLogSize, err := resource.ParseQuantity("500Mi")
-	if err != nil {
-		return &rookCephv1.CephCluster{}, fmt.Errorf("Failed to parse maxLogSize for log rotate. %v", err)
-	}
+	maxLogSize := resource.MustParse("500Mi")
```
iamniting, line 414: Can you also remove the error part from the func signature, It is not used anywhere else.

iamniting (changes requested):

controllers/storagecluster/cephcluster.go
parth-gr: Okay done

iamniting (changes requested):
Can you look at the failures?

parth-gr:
> Can you look at the failures?

Fixed, there were more places where the cephcluster was called

controllers/storagecluster/cephcluster.go
```diff
@@ -192,12 +192,12 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
 					return reconcile.Result{}, err
 				}
 			}
-			cephCluster, err = newCephCluster(sc, r.images.Ceph, kmsConfigMap, r.Log)
+			cephCluster = newCephCluster(sc, r.images.Ceph, kmsConfigMap, r.Log)
```
iamniting, line 195: Can you remove 3 lines under this, They are not required now

```
if err != nil {				
}
```

iamniting (changes requested):

iamniting (approved):

PR #2707, opened 2024-07-20
provider-server: send info of kernelMountOptions for cephfs to client when encryption in transit is enabled

rohan47:
add kernel mount option ms_mode=secure to cephfs storageclass data when encryption in transit is enabled

### DIFF
--- a/controllers/storagecluster/csi_options.go
+++ b/controllers/storagecluster/csi_options.go
@@ -5,33 +5,6 @@ import (
 	rookCephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
 )
 
-// getCephFSKernelMountOptions returns the kernel mount options for CephFS based on the spec on the StorageCluster
-func getCephFSKernelMountOptions(sc *ocsv1.StorageCluster) string {
-	// If Encryption is enabled, Always use secure mode
-	if sc.Spec.Network != nil && sc.Spec.Network.Connections != nil &&
-		sc.Spec.Network.Connections.Encryption != nil && sc.Spec.Network.Connections.Encryption.Enabled {
-		return "ms_mode=secure"
-	}
-
-	// If Encryption is not enabled, but Compression or RequireMsgr2 is enabled, use prefer-crc mode
-	if sc.Spec.Network != nil && sc.Spec.Network.Connections != nil &&
-		((sc.Spec.Network.Connections.Compression != nil && sc.Spec.Network.Connections.Compression.Enabled) ||
-			sc.Spec.Network.Connections.RequireMsgr2) {
-		return "ms_mode=prefer-crc"
-	}
-
-	// Network spec always has higher precedence even in the External or Provider cluster. so they are checked first above
-
-	// None of Encryption, Compression, RequireMsgr2 are enabled on the StorageCluster
-	// If it's an External or Provider cluster, We don't require msgr2 by default so no mount options are needed
-	if sc.Spec.ExternalStorage.Enable || sc.Spec.AllowRemoteStorageConsumers {
-		return "ms_mode=legacy"
-	}
-	// If none of the above cases apply, We set RequireMsgr2 true by default on the cephcluster
-	// so we need to set the mount options to prefer-crc
-	return "ms_mode=prefer-crc"
-}
-
 // getReadAffinityyOptions returns the read affinity options based on the spec on the StorageCluster.
 func getReadAffinityOptions(sc *ocsv1.StorageCluster) rookCephv1.ReadAffinitySpec {
 	if sc.Spec.CSI != nil && sc.Spec.CSI.ReadAffinity != nil {
--- a/services/provider/server/server.go
+++ b/services/provider/server/server.go
@@ -19,6 +19,7 @@ import (
 
 	"github.com/blang/semver/v4"
 	quotav1 "github.com/openshift/api/quota/v1"
+	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
 	"github.com/red-hat-storage/ocs-operator/api/v4/v1alpha1"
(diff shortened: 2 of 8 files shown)


services/provider/server/server.go
```diff
@@ -769,3 +776,25 @@ func (s *OCSProviderServer) getOCSSubscriptionChannel(ctx context.Context) (stri
 	}
 	return subscription.Spec.Channel, nil
 }
+
+func (s *OCSProviderServer) getCephfsKernelMountOptions(ctx context.Context) (string, error) {
```
Madhu-1, line 780: we already have https://github.com/red-hat-storage/ocs-operator/blob/9200d553079cbe217c19193f671f05d825cc3fb3/controllers/storagecluster/csi_options.go#L9-L33 just adjust it as per the requirement?

rohan47:
Testing the code changes

rohan47:
Tested the code changes

Csi config after the changes

```yaml
oc get cm ceph-csi-configs -o yaml
apiVersion: v1
data:
  config.json: '[{"clusterID":"81fd58def37a35d1b4d4bee7ba35c3c3","storageClientID":"fb3e4f11-a38b-4859-ad18-1e8d579eb72b",
"monitors":["10.0.16.90:3300","10.0.61.203:3300","10.0.59.136:3300"],"rbd":{"radosNamespace":"cephradosnamespace-f6e94668817e4eefeb19b7a402fd947e"}},{"clusterID":"fe09c94a50c72d033da43478c21b8314",
"storageClientID":"fb3e4f11-a38b-4859-ad18-1e8d579eb72b","monitors":["10.0.16.90:3300","10.0.61.203:3300","10.0.59.136:3300"],
"cephFS":{"subvolumeGroup":"cephfilesystemsubvolumegroup-38beca789f5bc10e556b9832fd1070f6",
"kernelMountOptions":"ms_mode=secure"}}]'
kind: ConfigMap
metadata:
  creationTimestamp: "2024-07-22T19:24:10Z"
  name: ceph-csi-configs
  namespace: openshift-storage
  ownerReferences:
  - apiVersion: v1
    blockOwnerDeletion: true
    controller: true
    kind: ConfigMap
    name: ocs-client-operator-config
    uid: 664e804f-fc77-4aef-9387-a3c39dcb0b00
  resourceVersion: "3135818"
  uid: 7dee1733-974c-48f3-9cc7-7dbab17aee62
```

services/provider/server/server.go
```diff
@@ -687,6 +694,15 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
 				"csi.storage.k8s.io/controller-expand-secret-name": provisionerSecretName,
 			}
 
+			storageClusters, err := util.GetStorageClustersInNamespace(ctx, s.client, s.namespace)
+			if err != nil || len(storageClusters.Items) == 0 {
+				return nil, status.Errorf(codes.Internal, "failed to get storage cluster %v", err)
+			}
+			kernelMountOptions := util.GetCephFSKernelMountOptions(&storageClusters.Items[0])
+			if kernelMountOptions == kernelMountOptionSecure {
+				cephfsStorageClassData[kernelMountOptionsKey] = kernelMountOptionSecure
```
nb-ohad, line 703: We should not piggyback csi config info onto the storage class resource. 
CSI config needs to be returned as its own resource

nb-ohad (changes requested):

services/provider/server/server.go
```diff
@@ -687,6 +694,15 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
 				"csi.storage.k8s.io/controller-expand-secret-name": provisionerSecretName,
 			}
 
+			storageClusters, err := util.GetStorageClustersInNamespace(ctx, s.client, s.namespace)
+			if err != nil || len(storageClusters.Items) == 0 {
+				return nil, status.Errorf(codes.Internal, "failed to get storage cluster %v", err)
+			}
+			kernelMountOptions := util.GetCephFSKernelMountOptions(&storageClusters.Items[0])
```
leelavg, line 701: how are you getting `ms_mode=secure` as mentioned in the test comment as the function will return `ms_mode=legacy` if remote consumers are allowed? Is it due to `sc.Spec.Network.Connections.Encryption.Enabled`?

leelavg (commented):
In the generated ceph config, shouldn't it also contain mons w/ `6789` port for ceph client to communicate w/o encryption?

services/provider/server/server.go
rohan47: Yes if `sc.Spec.Network.Connections.Encryption.Enabled` then it will return ms_mode=secure

rohan47:
> In the generated ceph config, shouldn't it also contain mons w/ `6789` port for ceph client to communicate w/o encryption?

In Case EiT is enabled on storagecluster all the clients are supposed to use 3300 right as mons will only listen on secure port?

controllers/storagecluster/csi_options.go
leelavg, line 1: https://github.com/red-hat-storage/ocs-operator/pull/2707#issuecomment-2244796375

continuing the discussion on PR rather than as a comment. So, does the kernel client on client cluster continues to offer services for PVCs which were created before enabling EiT?

rohan47: for 4.17 deployments the decision to use EiT will be taken when the provider is deployed right and we don't have to worry about clients connecting on unsecure port

leelavg: Well, rather than asking the specifics, let me put it this way, w/ this PR does the PVCs created on 4.16 still work?

rohan47: No, It won't work
 4.16-based deployments will not support EiT even after they upgrade to 4.17 right?

leelavg:
included in #2713 pls check & confirm.

PR #2708, opened 2024-07-22
use "upmap-read" as default balancer mode

sp98:
This new mode was added in Squid due to the balancer module not balancing the reads.
(BZ - https://bugzilla.redhat.com/show_bug.cgi?id=1870804)


iamniting (approved):

idryomov (commented):
`upmap-read` mode of the balancer isn't supported by the kernel client, so this needs to be reverted I'm afraid.

(There was a recent change on the RADOS side which actually enforced this, see https://github.com/ceph/ceph/pull/57776.  Prior to 18.2.4 and 19.1.1 all sorts of incompatibility checks were missing in this area.)

idryomov:
This is being reverted in #2747.

PR #2711, opened 2024-07-24
Expose storageCluster flag to allow updating of the OSD device classes

malayparida2000:
Earlier the device class of an OSD was determined at provisioning and
it was not possible to update thereafter. In some scenarios the admin
may want to force update the device class to a new value.
To avoid accidental updates, rook requires first setting
allowDeviceClassUpdate in the storage spec of the cephcluster for which
we are exposing a new flag at
Spec.ManagedResources.CephCluster.AllowDeviceClassUpdate.

context-https://github.com/rook/rook/pull/14457


controllers/storagecluster/cephcluster.go
```diff
@@ -445,6 +445,7 @@ func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, kmsConfigMap *co
 				Interval: &metav1.Duration{Duration: 30 * time.Second},
 			},
 			Storage: rookCephv1.StorageScopeSpec{
+				AllowDeviceClassUpdate:       true,
```
travisn, line 448: If all went well with #2615 and #2694 to properly set the correct device classes on the OSDs and pools, this will be a no-op and no changes will be needed to the OSD device classes. So I'm having second thoughts about always setting this to true. If there are any bugs in the device classes of the OSDs, it could cause unexpected data movement or pools to break. Instead, I am thinking we should set this to `false` by default and expose a storageCluster setting that allows the admin to set it to true if they do actually want to force update the OSD device class. I would expect only support would need to temporarily enable that setting if they do find an issue with the device classes.

malayparida2000: Hm, That's actually true the bug attached to 2615 has been verified after testing but 2694 is still under testing, should we wait till 2694 is also tested & verified?

malayparida2000: Ack, have updated the PR now.

iamniting (approved):
/lgtm
/hold for Travis review

Nikhil-Ladha (commented):
@malayparida2000 better to have the generated crds, manifests change in a seperate commit?

malayparida2000:
> @malayparida2000 better to have the generated crds, manifests change in a seperate commit?

Yeah kind off but as it's already LGTMed, I will let that pass to avoid another push & review.

PR #2712, opened 2024-07-24
Available CRDs check feature

raaizik:
## Changes

- Adds a new controller that reacts whenever a create/delete event has been enqueued for a specific CRD that's become newly available. 
- Note: this PR introduces only the structure of the feature without any specific CRDs which will be added from their respective PRs (e.g., `ClusterClaim`, `VolumeGroupSnapshotClass` and `DRClusterConfig`).

### DIFF
--- a/controllers/storagecluster/reconcile.go
+++ b/controllers/storagecluster/reconcile.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	error1 "errors"
 	"fmt"
+	"os"
+	"reflect"
 	"strings"
 	"time"
 
@@ -159,6 +161,17 @@ func (r *StorageClusterReconciler) Reconcile(ctx context.Context, request reconc
 		return reconcile.Result{}, err
 	}
 
+	var err error
+	availableCrds, err := statusutil.MapCRDAvailability(ctx, r.Client, statusutil.GetCrds()...)
+	if err != nil {
+		r.Log.Error(err, "error fetching the CRDs")
+		return reconcile.Result{}, err
+	}
+	if !reflect.DeepEqual(availableCrds, r.AvailableCrds) {
+		r.Log.Info("CustomResourceDefinitions created/deleted. Restarting process.")
+		os.Exit(42)
+	}
+
 	if err := r.validateStorageClusterSpec(sc); err != nil {
 		return reconcile.Result{}, err
 	}
@@ -166,7 +179,6 @@ func (r *StorageClusterReconciler) Reconcile(ctx context.Context, request reconc
 	r.IsNoobaaStandalone = sc.Spec.MultiCloudGateway != nil &&
 		ReconcileStrategy(sc.Spec.MultiCloudGateway.ReconcileStrategy) == ReconcileStrategyStandalone
 
-	var err error
 	r.clusters, err = statusutil.GetClusters(ctx, r.Client)
 	if err != nil {
 		r.Log.Error(err, "Failed to get clusters")
--- a/controllers/storagecluster/storagecluster_controller.go
+++ b/controllers/storagecluster/storagecluster_controller.go
@@ -96,6 +96,7 @@ type StorageClusterReconciler struct {
 	IsMultipleStorageClusters bool
 	clusters                  *util.Clusters
 	OperatorNamespace         string
+	AvailableCrds             map[string]bool
 }
 
 // SetupWithManager sets up a controller with manager
@@ -211,6 +212,31 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		},
 	}
 
+	crdPredicate := predicate.Funcs{
+		CreateFunc: func(e event.TypedCreateEvent[client.Object]) bool {
+			crdAvailable, keyExist := r.AvailableCrds[e.Object.GetName()]
+			if keyExist && !crdAvailable {
+				r.Log.Info("CustomResourceDefinition %s was Created.", e.Object.GetName())
+				return true
+			}
+			return false
+		},
+		DeleteFunc: func(e event.TypedDeleteEvent[client.Object]) bool {
+			crdAvailable, keyExist := r.AvailableCrds[e.Object.GetName()]
+			if keyExist && crdAvailable {
+				r.Log.Info("CustomResourceDefinition %s was Deleted.", e.Object.GetName())
+				return true
+			}
+			return false
+		},
+		UpdateFunc: func(e event.TypedUpdateEvent[client.Object]) bool {
+			return false
+		},
+		GenericFunc: func(e event.TypedGenericEvent[client.Object]) bool {
+			return false
+		},
+	}
+
 	build := ctrl.NewControllerManagedBy(mgr).
 		For(&ocsv1.StorageCluster{}, builder.WithPredicates(scPredicate)).
 		Owns(&cephv1.CephCluster{}, builder.WithPredicates(cephClusterIgnoreTimeUpdatePredicate)).
@@ -228,6 +254,7 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&corev1.Secret{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
 		Owns(&routev1.Route{}).
 		Owns(&templatev1.Template{}).
+		Watches(&extv1.CustomResourceDefinition{}, enqueueStorageClusterRequest, builder.WithPredicates(crdPredicate)).
 		Watches(&storagev1.StorageClass{}, enqueueStorageClusterRequest).
 		Watches(&volumesnapshotv1.VolumeSnapshotClass{}, enqueueStorageClusterRequest).
 		Watches(&ocsclientv1a1.StorageClient{}, enqueueStorageClusterRequest).
--- a/controllers/util/k8sutil.go
+++ b/controllers/util/k8sutil.go
@@ -11,6 +11,7 @@ import (
 	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
 	corev1 "k8s.io/api/core/v1"
 	storagev1 "k8s.io/api/storage/v1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -43,6 +44,10 @@ const (
 	// This is the name for the OwnerUID FieldIndex
 	OwnerUIDIndexName = "ownerUID"
 )
+
+func GetCrds() []string {
+	return []string{}
+}
 
 // GetWatchNamespace returns the namespace the operator should be watching for changes
 func GetWatchNamespace() (string, error) {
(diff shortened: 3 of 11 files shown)


iamniting (changes requested):
/hold

Why do we need this PR? I really discourage having such changes without the proper discussion with the maintainers.

Also, I would highly encourage having meaningful commit title and message. The current one does not make any sense.

Screenshot from 2024-07-25 11-08-27

iamniting (changes requested):
Could you please separate this commit into two commits?

* The first commit should include the controller changes.
* The second commit should include the script changes.

I will help test the first commit without the script. I believe we should be sorted with just the first commit.

umangachapagain (changes requested):
This feels like a complex hack to cause a panic without actually using `panic()`.
How did we finalize on this approach?

iamniting:
> This feels like a complex hack to cause a panic without actually using `panic()`. How did we finalize on this approach?

@umangachapagain We do not need this approach at all, Me and Rewant tested the solution without any script and fixed the bugs we had and it works fine. They will soon update the PR.

iamniting (changes requested):
```
- At the start each reconcile iteration, revalidate which CRD are now
 available. If a CRD of interest is now avail, panic the op
```

Can you pls rephrase this statement in the commit message, This is not correct. Also please update the commit title.

controllers/util/k8sutil.go
```diff
@@ -44,6 +45,8 @@ const (
 	OwnerUIDIndexName = "ownerUID"
 )
 
+var CRDList []string
```
iamniting, line 48: Can we rename this variable and please add a comment on what does it holds.

controllers/crd/crd_controller.go
```diff
@@ -0,0 +1,73 @@
...
+	Client        client.Client
+	ctx           context.Context
+	Log           logr.Logger
+	AvailableCrds map[string]bool
+}
+
+// Reconcile compares available CRDs maps following either a Create or Delete event
+func (r *CustomResourceDefinitionReconciler) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+	r.ctx = ctx
+	r.Log.Info("Reconciling CustomResourceDefinition.", "CRD", klog.KRef(request.Namespace, request.Name))
```
iamniting, line 30: I think only name is enough here.

iamniting (changes requested):

controllers/util/k8sutil.go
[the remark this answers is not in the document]
raaizik: > Note: this PR introduces only the structure of the feature without any specific CRDs which will be added from their respective PRs (e.g., ClusterClaim, VolumeGroupSnapshotClass and DRClusterConfig).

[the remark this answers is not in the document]
iamniting: We also need this to be tested with an empty list. We should not assume that it will work.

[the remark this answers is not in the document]
rewantsoni: Tested with with and without entries in CRDList and it works

controllers/crd/crd_controller.go
```diff
@@ -0,0 +1,73 @@
...
+
+	var err error
+	availableCrds, err := util.MapCRDAvailability(ctx, r.Client, util.CRDList...)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	if !reflect.DeepEqual(availableCrds, r.AvailableCrds) {
```
nb-ohad, line 37: I do not believe a map guarantees the order of keys if the 2 maps contain different keys. DeepEqual might fail even if both maps contain the exact same keys.  

Please move to a "len check + loop on items" for checking equality.

controllers/crd/crd_controller.go
```diff
@@ -0,0 +1,73 @@
...
+		r.Log.Info("CustomResourceDefinitions created/deleted. Restarting process.")
+		panic("CustomResourceDefinitions created/deleted. Restarting process.")
```
nb-ohad, line 39: I don't believe panic will allow us to detect the reason for the exit from outside the operator. use `os.Exit` with a specific exit code.

controllers/util/k8sutil.go
nb-ohad, line 48: Global singleton non-cost, non-atomic state inside a utils file is a red flag. Why do we need this?

nb-ohad (changes requested):
Maybe I am missing something, but the current code will panic the entire pod! 
Why have we removed the code that recycles the process without exiting failing the pod?

nb-ohad:
@iamniting @umangachapagain 
This approach it not the approach that was discussed. We need to talk about this before merging this one.

nb-ohad:
/hold
Adding a hold until we have a proper discussion and agreement on the approach

controllers/storagecluster/reconcile.go
```diff
@@ -159,14 +161,24 @@ func (r *StorageClusterReconciler) Reconcile(ctx context.Context, request reconc
...
+	var err error
+	availableCrds, err := statusutil.MapCRDAvailability(ctx, r.Client, statusutil.GetCrds()...)
+	if err != nil {
+		r.Log.Error(err, "error fetching the CRDs")
+		return reconcile.Result{}, err
+	}
+	if !reflect.DeepEqual(availableCrds, r.AvailableCrds) {
+		r.Log.Info("CustomResourceDefinitions created/deleted. Restarting process.")
+		os.Exit(42)
+	}
```
nb-ohad, line 173: Available CRD is a map. I am not sure if there is an order guarantee on keys which means that DeepEqual might fail even if all the keys and values are the same. I would suggest changing the check to `maps.Equal`

controllers/storagecluster/storagecluster_controller.go
nb-ohad, line 1: Where is the conditional watch that checks the CRD exists before adding the watch?

controllers/storagecluster/reconcile.go
```diff
@@ -159,14 +161,24 @@ func (r *StorageClusterReconciler) Reconcile(ctx context.Context, request reconc
...
 
+	var err error
+	availableCrds, err := statusutil.MapCRDAvailability(ctx, r.Client, statusutil.GetCrds()...)
+	if err != nil {
+		r.Log.Error(err, "error fetching the CRDs")
+		return reconcile.Result{}, err
+	}
+	if !reflect.DeepEqual(availableCrds, r.AvailableCrds) {
+		r.Log.Info("CustomResourceDefinitions created/deleted. Restarting process.")
+		os.Exit(42)
```
nb-ohad, line 172: Can we please have the exit code in a constant with a name that explains the usage? 
In addition, a comment about this line will help

nb-ohad (changes requested):

hack/crdavail.sh
```diff
@@ -0,0 +1,11 @@
...
+
+RESTART_EXIT_CODE=42
+
+while true; do
+    ./usr/local/bin/ocs-operator --enable-leader-election --health-probe-bind-address=:8081
+    EXIT_CODE=$?
+    if [ $EXIT_CODE -ne $RESTART_EXIT_CODE ]; then
+      exit $EXIT_CODE
+    fi
+done
```
nb-ohad, line 11: There is a missing newline here and git is complaining

hack/crdavail.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+RESTART_EXIT_CODE=42
+
+while true; do
+    ./usr/local/bin/ocs-operator --enable-leader-election --health-probe-bind-address=:8081
```
leelavg, line 6: we are effectively making this a sub-process rather than a PID 1 and I'm not confident enough in the shell handling signals vs controller-runtime.

I could see good amount of comments on this PR, in brief, may I know why did we backoff from kubelet restarting us?

controllers/util/k8sutil.go
```diff
@@ -44,6 +45,10 @@ const (
 	OwnerUIDIndexName = "ownerUID"
 )
 
+func GetCrds() []string {
+	return []string{}
```
leelavg, line 49: is my understanding correct that in upcoming PRs we register the CRDNames in this and conditionally watch on them in the setupwithmanager?

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -211,6 +212,31 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+			return false
+		},
+		DeleteFunc: func(e event.TypedDeleteEvent[client.Object]) bool {
+			crdAvailable, keyExist := r.AvailableCrds[e.Object.GetName()]
+			if keyExist && crdAvailable {
+				r.Log.Info("CustomResourceDefinition %s was Deleted.", e.Object.GetName())
+				return true
+			}
+			return false
+		},
```
leelavg, line 231: is there a possibility that CRD deletion is less than our operator lifetime?

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -228,6 +254,7 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&corev1.Secret{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
 		Owns(&routev1.Route{}).
 		Owns(&templatev1.Template{}).
+		Watches(&extv1.CustomResourceDefinition{}, enqueueStorageClusterRequest, builder.WithPredicates(crdPredicate)).
```
leelavg, line 257: we already made a mistake (with Virt) in watching all CRDs increasing the memory footprint, my mistake in not pushing hard on this fix https://github.com/red-hat-storage/ocs-operator/pull/2539, seeing this PR we don't want to watch all CRDs until & unless necessary.

Clubbed w/ 2539 we could do something like below as we are only interested in the metadata of CRDs while mapping CRDs.
```
crd := &metav1.PartialObjectMetadata{}
crd.SetGroupVersionKind(extv1.SchemeGroupVersion.WithKind("CustomResourceDefinition"))
crd.Name = <CRDName>
client.Get(<ctx>, <ns-name>, crd)
```

leelavg (commented):
I believe it might've been discussed, just out of curiosity, why can't the operation be offloaded to odf-op? odf-op could set env and ocs-op correspondingly sets up the watch? I didn't think of all the possibilities though.

controllers/storagecluster/reconcile.go
rewantsoni: I tried to create a map, added keys in a different order, and compared them using DeepEqual and it worked. maps.Equal produces same results
https://go.dev/play/p/uFkmp57xEIp

controllers/storagecluster/storagecluster_controller.go
rewantsoni: Added it in https://github.com/red-hat-storage/ocs-operator/pull/2745

controllers/storagecluster/reconcile.go
rewantsoni: Added it with https://github.com/red-hat-storage/ocs-operator/pull/2745

hack/crdavail.sh
rewantsoni: Done

controllers/util/k8sutil.go
rewantsoni: Added it with https://github.com/red-hat-storage/ocs-operator/pull/2745

controllers/storagecluster/storagecluster_controller.go
rewantsoni: Although it is very unlikely but it is possible, and if they happens it will lead to a pod restart

controllers/storagecluster/storagecluster_controller.go
rewantsoni: Looked at this and I couldn't find a way to specify multiple names in the cache using field selector, But I have updated it to use PartialObjectMetadata here: https://github.com/red-hat-storage/ocs-operator/pull/2745

PR #2713, opened 2024-07-25
provider: send csi-op APIs spec as part of gRPC response

leelavg:
client-operator needs two pieces of data among other things to configure csi-op (new) which inturn reconciles CSI.

The data required is Monitor IPs and names of CephBlockPoolRadosNamespace, CephFileSystemSubvolumeGroup. With this PR, the former will be sent via StorageConfig and latter names are via StorageClaimConfig, this helps client-operator to directly use these values rather than gluing the data from different gRPC responses.


leelavg:
I'll split into commits when we are close to merge.

go.mod
```diff
@@ -6,9 +6,12 @@ replace github.com/red-hat-storage/ocs-operator/api/v4 => ./api
 
 replace github.com/red-hat-storage/ocs-operator/metrics/v4 => ./metrics
 
+replace github.com/ceph/ceph-csi-operator => github.com/nb-ohad/ceph-csi-operator v0.0.0-20240731113052-89481456292a
+
 require (
 	github.com/RHsyseng/operator-utils v1.4.13
 	github.com/blang/semver/v4 v4.0.0
+	github.com/ceph/ceph-csi-operator v0.0.0-00010101000000-000000000000
```
nb-ohad, line 14: I believe there is no need to load the entire `ceph-csi-operator` package as a dependency but take only the API sub-packge

services/provider/proto/provider.proto
```diff
@@ -73,6 +73,8 @@ message ExternalResource {
     string kind = 2;
     // Data contains the contents of the external cluster resource
     bytes Data = 3;
+    // ApiVersion contains Group and Version of the resource
+    string apiVersion = 4;
```
nb-ohad, line 77: If you are introducing this field in this PR you should add the correct value for all payloads who are sending `ExternalResource`

services/provider/server/server.go
```diff
@@ -275,6 +277,18 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
 		})})
 
+	monIps, err := extractMonitorIps(configmap.Data["data"])
+	if err != nil {
+		return nil, fmt.Errorf("failed to extract monitor IPs from configmap %s: %v", monConfigMap, err)
+	}
+
+	extR = append(extR, &pb.ExternalResource{
+		ApiVersion: csiopv1a1.GroupVersion.Identifier(),
+		Kind:       "CephConnection",
```
nb-ohad, line 287: This becomes very verbose, and with the introduction of `ApiVersion` I can see how a copy-paste error can happen very easily.  I suggest having something like a `NewExternalResourceFor(object, client.object)` util to help in the correct reconstruction of the `ExternalResource` object.

This new function can use the schema to extract `APIVersion` and `Kind` information about the object and can extract the object name and data from the object itself.

nb-ohad (changes requested):

go.mod
leelavg: moved to api module as it is available now.

services/provider/proto/provider.proto
leelavg: Removed from this PR and I'll introduce that as a separate one.

services/provider/server/server.go
leelavg: Will be part of different PR.

services/provider/server/server.go
```diff
@@ -681,12 +707,31 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
...
+				},
+			}
+			storageClusters := &ocsv1.StorageClusterList{}
+			if err := s.client.List(ctx, storageClusters, client.InNamespace(s.namespace), client.Limit(1)); err != nil {
+				return nil, status.Errorf(codes.Internal, "failed to get storage cluster: %v", err)
+			}
+			if len(storageClusters.Items) < 1 {
+				return nil, status.Errorf(codes.Internal, "storagecluster does not exist")
+			}
+			kernelMountOptions := util.GetCephFSKernelMountOptions(&storageClusters.Items[0])
```
nb-ohad, line 728: How do you know that storage cluster 0 is the correct one to take the configuration from?

services/provider/server/server.go
```diff
@@ -681,12 +707,31 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
...
+			// TODO: do we want to always send the options?
+			if kernelMountOptions == "ms_mode=secure" {
+				parts := strings.Split(kernelMountOptions, "=")
+				clientProfileSpec.CephFs.KernelMountOptions = map[string]string{parts[0]: parts[1]}
+			}
```
nb-ohad, line 733:
```suggestion
			if kernelMountOptions == "ms_mode=secure" {
				clientProfileSpec.CephFs.KernelMountOptions = map[string]string{"ms_mode": "secure"}
			}
```

services/provider/server/server.go
```diff
@@ -681,12 +707,31 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
...
+			storageClusters := &ocsv1.StorageClusterList{}
+			if err := s.client.List(ctx, storageClusters, client.InNamespace(s.namespace), client.Limit(1)); err != nil {
+				return nil, status.Errorf(codes.Internal, "failed to get storage cluster: %v", err)
+			}
+			if len(storageClusters.Items) < 1 {
+				return nil, status.Errorf(codes.Internal, "storagecluster does not exist")
+			}
+			kernelMountOptions := util.GetCephFSKernelMountOptions(&storageClusters.Items[0])
+			// TODO: do we want to always send the options?
+			if kernelMountOptions == "ms_mode=secure" {
```
nb-ohad, line 730: what happens if the `kernelMountOptions` includes more then "ms_mode"? 
Your check will fail and we will not send it to the client.

services/provider/server/server.go
```diff
@@ -681,12 +707,31 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
...
 				"csi.storage.k8s.io/provisioner-secret-name":       provisionerSecretName,
 				"csi.storage.k8s.io/node-stage-secret-name":        nodeSecretName,
 				"csi.storage.k8s.io/controller-expand-secret-name": provisionerSecretName,
 			}
 
+			clientProfileSpec := &csiopv1a1.ClientProfileSpec{
+				CephFs: &csiopv1a1.CephFsConfigSpec{
+					SubVolumeGroup: cephRes.Name,
+				},
+			}
```
nb-ohad, line 720: Let's build that after we have calculated the kernel mount options (or even hardcoded inside the ExternalResource definition) so there will be no need to edit the object we created.

nb-ohad (changes requested):

services/provider/server/server.go
leelavg: In our namespace we are only supporting 1 storagecluster.

services/provider/server/server.go
leelavg: I could trim but we were getting that string from static code.

services/provider/server/server.go
leelavg: we will not be editing if there is no mountoption, I can move the same to after calculation.

services/provider/server/server.go
nb-ohad: So why do you need to list?

services/provider/server/server.go
nb-ohad: A map can be nil, so I don't understand the issue. 
Start be defining it as nil, then have your checks, then overwrite the nil if nececary, then build the struct with whatever is in the map's variable

services/provider/server/server.go

services/provider/server/server.go

services/provider/server/server.go

services/provider/server/server.go
```diff
@@ -681,12 +707,25 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
...
+				return nil, status.Errorf(codes.Internal, "failed to get storage cluster: %v", err)
+			}
+			if len(storageClusters.Items) != 1 {
+				return nil, status.Errorf(codes.Internal, "expecting one single storagecluster to exist")
+			}
+			kernelMountOptions := map[string]string{}
+			for _, option := range strings.Split(util.GetCephFSKernelMountOptions(&storageClusters.Items[0]), ",") {
+				parts := strings.Split(option, "=")
+				kernelMountOptions[parts[0]] = parts[1]
+			}
```
nb-ohad, line 727: If there are not kernel mount options you want to keep the map as a nil map and not as a empty map. 
```suggestion
			var kernelMountOptions map[string]string = nil
			for _, option := range strings.Split(util.GetCephFSKernelMountOptions(&storageClusters.Items[0]), ",") {
				if kernelMountOptions == nil {
					kernelMountOptions = map[string]string{} 
				}
				parts := strings.Split(option, "=")
				kernelMountOptions[parts[0]] = parts[1]
			}
```

nb-ohad (changes requested):

services/provider/server/server.go

services/provider/server/server.go
```diff
@@ -681,12 +707,28 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
...
 
+			storageClusters := &ocsv1.StorageClusterList{}
+			if err := s.client.List(ctx, storageClusters, client.InNamespace(s.namespace), client.Limit(2)); err != nil {
+				return nil, status.Errorf(codes.Internal, "failed to get storage cluster: %v", err)
+			}
+			if len(storageClusters.Items) != 1 {
+				return nil, status.Errorf(codes.Internal, "expecting one single storagecluster to exist")
+			}
+			var kernelMountOptions map[string]string
+			for _, option := range strings.Split(util.GetCephFSKernelMountOptions(&storageClusters.Items[0]), ",") {
```
leelavg, line 724: @Madhu-1 this is what I was asking, we get ms_mode either as legacy (allowingcobsumers) or secure (encryption) and so do we want to set the option only for encryption?

Madhu-1: GetCephFSKernelMountOptions returns ms_mode with 3 options, secure, prefer-crc, or legacy based on the storagecluser.

services/provider/server/server.go
```diff
@@ -681,12 +707,28 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
...
+				if kernelMountOptions == nil {
+					kernelMountOptions = map[string]string{}
+				}
+				parts := strings.Split(option, "=")
```
Madhu-1, line 728: we need if check for parts length and i don't understand if check for kernelMountOptions above, when that can happen?

nb-ohad: @Madhu-1 Are we assuming that the mount options will be returned malformed?

nb-ohad: Regarding the check. I have asked to ensure that we will serialize a nil map and not an empty one in case of no mount options. This means that we need to start with a nil map and allocate a map only if we encounter a mount option 

Please see my comment here: https://github.com/red-hat-storage/ocs-operator/pull/2713#discussion_r1713733642

Madhu-1: it won't be but if the function changed to something else or a bug, this assumption breaks with panic.

nb-ohad: If the function changes in a way that breaks the format of the KMO then it should panic so the dev would notice and fix it. If the format changes this code should change as well. Returning an error at runtime will not be visible until runtime which is too late

Madhu-1:
Lgtm

PR #2718, opened 2024-07-29
Update provider-api-server to be deployed in internal and provider mode

rewantsoni:
Allow provider-server to be deployed in intenal and provider mode

Why?
This is a part of RDR for Provider Mode RHSTOR-4886.
With this epic, we are introducing rpc calls for odf to odf communication, the RPC's we have for this epic are limited to Onboarding a Storage Cluster on different Openshift clusters and Getting the Mirroring Info required for setting up RDR (refer #2671).

This can also be used to replace the mechanism we have in the internal mode for setting up Mirroring for blockpool using MCO's Agents.

Hence moving the ocs-provider-server deployment to be deployed in both modes to enable both client to provider communication and cross odf communication.

The PR does the following:
1. Create service, deployment, onboarding job for both modes
2. Update the variable from watchnamespace to podnamespace
3. Remove hardcoded name for storagecluster
4. Move client configmap in storageclient
5. Update the unit tests

### DIFF
--- a/controllers/storagecluster/provider_server.go
+++ b/controllers/storagecluster/provider_server.go
@@ -1,13 +1,9 @@
 package storagecluster
 
 import (
-	"context"
 	"fmt"
-	"maps"
-	"math/rand"
 	"os"
 	"sort"
-	"strconv"
 	"time"
 
 	"go.uber.org/multierr"
@@ -40,47 +36,22 @@ const (
 	ocsProviderServiceNodePort = int32(31659)
 
 	ocsProviderCertSecretName = ocsProviderServerName + "-cert"
-
-	ocsClientConfigMapName = "ocs-client-operator-config"
-	deployCSIKey           = "DEPLOY_CSI"
-	manageNoobaaSubKey     = "manageNoobaaSubscription"
 )
 
 type ocsProviderServer struct{}
 
 func (o *ocsProviderServer) ensureCreated(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) (reconcile.Result, error) {
 
-	if !instance.Spec.AllowRemoteStorageConsumers {
-		r.Log.Info("Spec.AllowRemoteStorageConsumers is disabled")
-		return o.ensureDeleted(r, instance)
+	if res, err := o.createService(r, instance); err != nil || !res.IsZero() {
+		return res, err
 	}
 
-	r.Log.Info("Spec.AllowRemoteStorageConsumers is enabled. Creating Provider API resources")
-
-	if err := o.createSecret(r, instance); err != nil {
-		return reconcile.Result{}, err
+	if res, err := o.createDeployment(r, instance); err != nil || !res.IsZero() {
+		return res, err
 	}
 
-	if res, err := o.createService(r, instance); err != nil {
-		return reconcile.Result{}, err
-	} else if !res.IsZero() {
-		return res, nil
-	}
-
-	if res, err := o.createDeployment(r, instance); err != nil {
-		return reconcile.Result{}, err
-	} else if !res.IsZero() {
-		return res, nil
-	}
-
-	if res, err := o.createJob(r, instance); err != nil {
-		return reconcile.Result{}, err
-	} else if !res.IsZero() {
-		return res, nil
-	}
-
-	if err := o.updateClientConfigMap(r, instance.Namespace); err != nil {
-		return reconcile.Result{}, err
+	if res, err := o.createJob(r, instance); err != nil || !res.IsZero() {
+		return res, err
 	}
 
 	return reconcile.Result{}, nil
@@ -103,11 +74,10 @@ func (o *ocsProviderServer) ensureDeleted(r *StorageClusterReconciler, instance 
 	var finalErr error
 
 	for _, resource := range []client.Object{
-		GetProviderAPIServerSecret(instance),
 		GetProviderAPIServerService(instance),
 		GetProviderAPIServerDeployment(instance),
 	} {
-		err := r.Client.Delete(context.TODO(), resource)
+		err := r.Client.Delete(r.ctx, resource)
 
 		if err != nil && !errors.IsNotFound(err) {
 			r.Log.Error(err, "Failed to delete resource", "Kind", resource.GetObjectKind(), "Name", resource.GetName())
@@ -127,7 +97,7 @@ func (o *ocsProviderServer) createDeployment(r *StorageClusterReconciler, instan
 
 	var finalErr error
 
-	for _, env := range []string{providerAPIServerImage, util.WatchNamespaceEnvVar} {
+	for _, env := range []string{providerAPIServerImage} {
 		if _, ok := os.LookupEnv(env); !ok {
 			multierr.AppendInto(&finalErr, fmt.Errorf("ENV var %s not found", env))
 		}
@@ -145,13 +115,10 @@ func (o *ocsProviderServer) createDeployment(r *StorageClusterReconciler, instan
 		},
 	}
 
-	_, err := controllerutil.CreateOrUpdate(
-		context.TODO(), r.Client, actualDeployment,
-		func() error {
-			actualDeployment.Spec = desiredDeployment.Spec
-			return controllerutil.SetOwnerReference(instance, actualDeployment, r.Client.Scheme())
-		},
-	)
+	_, err := controllerutil.CreateOrUpdate(r.ctx, r.Client, actualDeployment, func() error {
+		actualDeployment.Spec = desiredDeployment.Spec
+		return controllerutil.SetOwnerReference(instance, actualDeployment, r.Client.Scheme())
+	})
 	if err != nil && !errors.IsAlreadyExists(err) {
 		r.Log.Error(err, "Failed to create/update deployment", "Name", desiredDeployment.Name)
 		return reconcile.Result{}, err
@@ -190,24 +157,21 @@ func (o *ocsProviderServer) createService(r *StorageClusterReconciler, instance 
 		},
 	}
 
-	_, err := controllerutil.CreateOrUpdate(
-		context.TODO(), r.Client, actualService,
-		func() error {
-			desiredService.Spec.ClusterIP = actualService.Spec.ClusterIP
-			desiredService.Spec.IPFamilies = actualService.Spec.IPFamilies
+	_, err := controllerutil.CreateOrUpdate(r.ctx, r.Client, actualService, func() error {
+		desiredService.Spec.ClusterIP = actualService.Spec.ClusterIP
+		desiredService.Spec.IPFamilies = actualService.Spec.IPFamilies
 
-			if actualService.Annotations == nil {
-				actualService.Annotations = map[string]string{}
-			}
+		if actualService.Annotations == nil {
+			actualService.Annotations = map[string]string{}
+		}
 
-			for key, value := range desiredService.Annotations {
-				actualService.Annotations[key] = value
-			}
+		for key, value := range desiredService.Annotations {
+			actualService.Annotations[key] = value
+		}
 
-			actualService.Spec = desiredService.Spec
-			return controllerutil.SetOwnerReference(instance, actualService, r.Client.Scheme())
-		},
-	)
+		actualService.Spec = desiredService.Spec
+		return controllerutil.SetOwnerReference(instance, actualService, r.Client.Scheme())
+	})
 	if err != nil {
 		r.Log.Error(err, "Failed to create/update service", "Name", desiredService.Name)
 		return reconcile.Result{}, err
@@ -270,7 +234,7 @@ func (o *ocsProviderServer) getWorkerNodesInternalIPAddresses(r *StorageClusterR
 
 	nodes := &corev1.NodeList{}
 
-	err := r.Client.List(context.TODO(), nodes)
+	err := r.Client.List(r.ctx, nodes)
 	if err != nil {
 		r.Log.Error(err, "Failed to list nodes")
 		return nil, err
@@ -293,28 +257,6 @@ func (o *ocsProviderServer) getWorkerNodesInternalIPAddresses(r *StorageClusterR
 	sort.Strings(nodeAddresses)
 
 	return nodeAddresses, nil
-}
-
-func (o *ocsProviderServer) createSecret(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) error {
-
-	desiredSecret := GetProviderAPIServerSecret(instance)
-	actualSecret := &corev1.Secret{}
-
-	err := r.Client.Get(context.TODO(), client.ObjectKeyFromObject(desiredSecret), actualSecret)
-
-	if err != nil && errors.IsNotFound(err) {
-		err = r.Client.Create(context.TODO(), desiredSecret)
-		if err != nil {
-			r.Log.Error(err, "Failed to create secret", "Name", desiredSecret.Name)
-			return err
-		}
-		r.Log.Info("Secret creation succeeded", "Name", desiredSecret.Name)
-	} else if err != nil {
-		r.Log.Error(err, "Failed to get secret", "Name", desiredSecret.Name)
-		return err
-	}
-
-	return nil
 }
 
 func (o *ocsProviderServer) ensureDeploymentReplica(actual, desired *appsv1.Deployment) error {
@@ -354,16 +296,12 @@ func GetProviderAPIServerDeployment(instance *ocsv1.StorageCluster) *appsv1.Depl
 							Command: []string{"/usr/local/bin/provider-api"},
 							Env: []corev1.EnvVar{
 								{
-									Name:  util.WatchNamespaceEnvVar,
-									Value: os.Getenv(util.WatchNamespaceEnvVar),
-								},
-								{
-									Name:  "STORAGE_CLUSTER_NAME",
-									Value: instance.Name,
-								},
-								{
-									Name:  "STORAGE_CLUSTER_UID",
-									Value: string(instance.UID),
+									Name: util.PodNamespaceEnvVar,
+									ValueFrom: &corev1.EnvVarSource{
+										FieldRef: &corev1.ObjectFieldSelector{
+											FieldPath: "metadata.namespace",
+										},
+									},
 								},
 							},
 							Ports: []corev1.ContainerPort{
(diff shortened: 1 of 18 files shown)


nb-ohad:
@rewantsoni Please provide a proper PR title and description

iamniting:
Can the provider handle requests for multiple storage clusters?

Do we want to run only one ocs-provider-server deployment in case of the multiple storagecluster?

If yes then we can have it as part of the ocs initialization otherwise it should be part of the storagecluster reconciliation.

iamniting (changes requested):
Can you pls add deprication warning in the storage cluster API in first commit?

Also pls separate the generated changes in the 3rd commit and rename the deployment to the `ocs-server`.

rewantsoni:
> Can the provider handle requests for multiple storage clusters?

Yes, the server should be able to handle multiple storage clusters, but the client will onboard to only the storageCluster in the same namespace as the provider-server
> Do we want to run only one ocs-provider-server deployment in case of the multiple storagecluster?

Yes, one provider-server for multiple storageClusters
> If yes then we can have it as part of the ocs initialization otherwise it should be part of the storagecluster reconciliation.

I have added the deployment as a part of csv and secret, service as part of ocsinit

iamniting (changes requested):
Can you pls resolve merge conflict

controllers/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -554,6 +580,175 @@ func getFailureDomainKeyFromStorageClassParameter(sc *storagev1.StorageClass) st
...
+
+	return endpoint
+}
+
+// getWorkerNodesInternalIPAddresses return slice of Internal IPAddress of worker nodes
+func (r *OCSInitializationReconciler) getWorkerNodesInternalIPAddresses() ([]string, error) {
+
+	nodes := &corev1.NodeList{}
+
+	err := r.Client.List(context.TODO(), nodes)
```
iamniting, line 705: lets use r.context

controllers/storagecluster/provider_server.go
```diff
@@ -1,552 +1,54 @@
...
-		return res, nil
-	}
-
-	if err := o.updateClientConfigMap(r, instance.Namespace); err != nil {
-		return reconcile.Result{}, err
-	}
-
-	return reconcile.Result{}, nil
-}
+	// handle deletion of the secret, deployment and the service
```
iamniting, line 20: Can you pls add comments why are we deleting it?

tools/csv-merger/csv-merger.go
```diff
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
+	"github.com/red-hat-storage/ocs-operator/v4/services/provider/server"
```
iamniting, line 7: Pls fix this import

iamniting (changes requested):

controllers/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -554,6 +574,177 @@ func getFailureDomainKeyFromStorageClassParameter(sc *storagev1.StorageClass) st
...
+	}
+
+	actualSecret := &corev1.Secret{}
+	// Creating the job only if public is not found
+	err := r.Client.Get(r.ctx, types.NamespacedName{Name: util.OnboardingValidationPublicKeySecretName,
+		Namespace: initialData.Namespace}, actualSecret)
+
+	if errors.IsNotFound(err) {
+		onboardingSecretGeneratorJob := getOnboardingJobObject(initialData)
+		err = r.Client.Create(context.Background(), onboardingSecretGeneratorJob)
```
leelavg, line 738:
```suggestion
		err = r.Client.Create(r.ctx, onboardingSecretGeneratorJob)
```

controllers/storagecluster/provider_server.go
```diff
@@ -1,552 +1,61 @@
...
-			}
-		}
-	}
-
-	sort.Strings(nodeAddresses)
-
-	return nodeAddresses, nil
-}
-
-func (o *ocsProviderServer) createSecret(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) error {
```
leelavg, line 298: I see this wasn't moved to ocsinit, what was this secret used for?

controllers/storagecluster/storageclient.go
```diff
@@ -59,3 +70,29 @@ func (s *storageClient) ensureDeleted(r *StorageClusterReconciler, storagecluste
...
+	if err := r.Client.Get(r.ctx, client.ObjectKeyFromObject(clientConfig), clientConfig); err != nil {
+		r.Log.Error(err, "failed to get ocs client configmap")
+		return err
+	}
+
+	existingData := maps.Clone(clientConfig.Data)
+	if clientConfig.Data == nil {
+		clientConfig.Data = map[string]string{}
+	}
+	clientConfig.Data[deployCSIKey] = "true"
```
leelavg, line 88: I see noobaasub key isn't carried over, intentional?

leelavg (commented):
client-op needs an update after this gets merged, isn't it?

controllers/storagecluster/provider_server.go
rewantsoni: I didn't find any use for this secret, and I am not sure why we had it. Nitin mentioned that it was used for the old way of getting tokens

controllers/storagecluster/storageclient.go
rewantsoni: The other PR was not merged when I raised this PR, thanks for pointing it out adding it

rewantsoni:
> client-op needs an update after this gets merged, isn't it?

@leelavg I don't think we need any update in client o/p after merging this, let me know if I missed anything.

leelavg:
Looked again, yes, no update required.

controllers/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -604,6 +624,177 @@ func getFailureDomainKeyFromStorageClassParameter(sc *storagev1.StorageClass) st
 	}
 }
 
+func (r *OCSInitializationReconciler) reconcileOCSServerService(initialData *ocsv1.OCSInitialization) (reconcile.Result, error) {
+	/*
+		OCS API Server ServiceType follows the following level of priority:
+			1. Value set in OCSInitialization
+			2. Value set in InternalStorageCluster
+		This helps to maintain backward compatibility
+	*/
```
iamniting, line 633: // The ServiceType for the OCS API Server is determined based on the following priority:
// 1. The value set in OCSInitialization.
// 2. The value set in InternalStorageCluster.
// This order ensures backward compatibility.

iamniting (changes requested):
Most of it looks good to me Can you please move the ocs server-related code to its own file inside the ocs intiallization controller?

controllers/ocsinitialization/ocsinitialization_controller.go
rewantsoni: done

controllers/util/k8sutil.go
```diff
@@ -45,6 +45,16 @@ const (
 	OwnerUIDIndexName = "ownerUID"
 
 	OdfInfoNamespacedNameClaimName = "odfinfo.odf.openshift.io"
+	OCSInitName                    = "ocsinit"
```
iamniting, line 48: Pls add a comment here that this is a ocsintialization CR names.

go.mod
```diff
@@ -38,6 +38,7 @@ require (
 	google.golang.org/grpc v1.63.3
 	google.golang.org/protobuf v1.34.2
 	gopkg.in/ini.v1 v1.67.0
+	gopkg.in/yaml.v2 v2.4.0
```
iamniting, line 41: Pls add these changes in the respective commit itself.

controllers/ocsinitialization/ocsserver.go
```diff
@@ -0,0 +1,162 @@
+package ocsinitialization
```
iamniting, line 1: Lets rename this file to ocs_server

iamniting (changes requested):

leelavg (commented):
LGTM.

controllers/ocsinitialization/ocsserver.go
rewantsoni: done

controllers/util/k8sutil.go
rewantsoni: done

go.mod
rewantsoni: done

controllers/ocsinitialization/onboarding.go
```diff
@@ -0,0 +1,73 @@
...
+	// Creating the job only if public is not found
+	err := r.Client.Get(r.ctx, types.NamespacedName{Name: util.OnboardingValidationPublicKeySecretName,
+		Namespace: initialData.Namespace}, actualSecret)
+
+	if errors.IsNotFound(err) {
+		onboardingSecretGeneratorJob := getOnboardingJobObject(initialData)
+		err = r.Client.Create(r.ctx, onboardingSecretGeneratorJob)
+	}
+	if err != nil {
+		r.Log.Error(err, "failed to create/ensure secret")
```
iamniting, line 35: failed to create/ensure job

controllers/ocsinitialization/onboarding.go
```diff
@@ -0,0 +1,73 @@
...
+
+func (r *OCSInitializationReconciler) ensureOnboardingJob(initialData *ocsv1.OCSInitialization) error {
+	if os.Getenv(util.OnboardingValidationKeysGeneratorImage) == "" {
+		err := fmt.Errorf("OnboardingSecretGeneratorImage env var is not set")
+		r.Log.Error(err, "No value set for env variable")
+		return err
+	}
+
+	actualSecret := &corev1.Secret{}
+	// Creating the job only if public is not found
```
iamniting, line 26: // Create the job only if the secret is not found

iamniting (changes requested):

obnoxxx:
@nb-ohad  wrote:

> @rewantsoni Please provide a proper PR title and description

I think the description is leaking too much internal downstream product information to this public project.

Shouldn't we try to avoid this?

iamniting (approved):
looks good to me.

/hold for @nb-ohad review.

nb-ohad:
/hold 

@rewantsoni 
This PR introduces architecture changes to the way we handle the API service and to the way we are acting on onboarding actions. Looking at the code I think these changes are unnecessary, complicated, and are pushing us in the wrong direction. 
I would like to have more discussion on this

nb-ohad:
Adding to my last comment: 
The API server is not part of the orchestrator, it is part of the application being deployed. Assuming multiple StorageClusters deployed over multiple namespaces, I would expect to see an API server in each of these namespaces, the same way I would expect to see any other orchestrated payload.

We also need to consider that each of these might be configured differently

controllers/storagecluster/reconcile.go
```diff
@@ -429,6 +429,7 @@ func (r *StorageClusterReconciler) reconcilePhases(
 		// for external cluster, we have a different set of ensure functions
 		// preserve list order
 		objs = []resourceManager{
+			&ocsProviderServer{},
```
rewantsoni, line 432: adding it here so that we deploy ocs-provider-server in external mode

controllers/storagecluster/provider_server.go
```diff
@@ -354,16 +296,12 @@ func GetProviderAPIServerDeployment(instance *ocsv1.StorageCluster) *appsv1.Depl
...
-									Name:  util.WatchNamespaceEnvVar,
-									Value: os.Getenv(util.WatchNamespaceEnvVar),
-								},
-								{
-									Name:  "STORAGE_CLUSTER_NAME",
-									Value: instance.Name,
-								},
-								{
-									Name:  "STORAGE_CLUSTER_UID",
-									Value: string(instance.UID),
```
nb-ohad, line 366: Why are we removing these?

controllers/storagecluster/reconcile.go
nb-ohad: Why do we need to run this in external mode?

controllers/util/k8sutil.go
```diff
@@ -47,6 +50,15 @@ const (
...
 )
 
+// GetPodNamespace returns the namespace in which the pod is deployed
+func GetPodNamespace() (string, error) {
+	ns, found := os.LookupEnv(PodNamespaceEnvVar)
+	if !found {
+		return "", fmt.Errorf("%s must be set", PodNamespaceEnvVar)
+	}
+	return ns, nil
+}
```
nb-ohad, line 60: 1. Why do we need to have an os look on every call? 
2. A look-up into an env var that is mandatory to the operator's operation should not be able to return an error. An error means rescheduling the event. That does not make sense because there is no way for a value to change in the next reconciliation. It is preferable to panic, as an extreme indication that there is a coding error somewhere. 
3. For a mandatory value, an empty string will be considered an errornous value as well


```suggestion
podNamespace := (func() {
	if ns := os.GetEnv(PodNamespaceEnvVar); ns != "" {
		panic fmt.Errorf("%s must be set", PodNamespaceEnvVar)
	} else {
		return ns
	}
})()
func GetPodNamespace() (string) {
	return podNamespace
}
```

onboarding-validation-keys-generator/main.go
```diff
@@ -77,7 +78,7 @@ func main() {
 		klog.Exitf("failed to delete public secret: %v", err)
 	}
 
-	err = controllerutil.SetOwnerReference(storageCluster, privateSecret, cl.Scheme())
+	err = controllerutil.SetOwnerReference(&storageClusterList.Items[0], privateSecret, cl.Scheme())
```
nb-ohad, line 81: Drop this change in favor of last comment change

onboarding-validation-keys-generator/main.go
```diff
@@ -91,7 +92,7 @@ func main() {
 		klog.Exitf("failed to create private secret: %v", err)
 	}
 
-	err = controllerutil.SetOwnerReference(storageCluster, publicSecret, cl.Scheme())
+	err = controllerutil.SetOwnerReference(&storageClusterList.Items[0], publicSecret, cl.Scheme())
```
nb-ohad, line 95: Drop this change in favor of last comment change

onboarding-validation-keys-generator/main.go
```diff
@@ -49,10 +47,13 @@ func main() {
...
-	err = cl.Get(ctx, types.NamespacedName{Name: storageClusterName, Namespace: operatorNamespace}, storageCluster)
+	storageClusterList := &v1.StorageClusterList{}
+	err = cl.List(ctx, storageClusterList, client.InNamespace(operatorNamespace))
 	if err != nil {
-		klog.Exitf("failed to get storage cluster: %v", err)
+		klog.Exitf("unable to list storageCluster(s) in %v namespace, %v", operatorNamespace, err)
+	}
+	if len(storageClusterList.Items) != 1 {
+		klog.Exitf("unexpected number of storageCluster(s) found in %v namespace, expected: 1 actual: %v", operatorNamespace, len(storageClusterList.Items))
 	}
```
nb-ohad, line 57:
```suggestion
		klog.Exitf("unexpected number of storageCluster(s) found in %v namespace, expected: 1 actual: %v", operatorNamespace, len(storageClusterList.Items))
	}
	storageCluster := &storageClusterList[0]
```

nb-ohad (changes requested):

controllers/storagecluster/provider_server.go
rewantsoni: These are not used in the server hence removing the unused code

controllers/util/k8sutil.go
rewantsoni: I agree and I am going to raise a separate PR for this, as we have other helper functions similar to this that read the env variable for OperatorNamespace, WatchNamespace

controllers/storagecluster/reconcile.go
rewantsoni: I think we decided that provider API server will be running in all modes

controllers/storagecluster/storageclient.go
```diff
@@ -59,3 +72,30 @@ func (s *storageClient) ensureDeleted(r *StorageClusterReconciler, storagecluste
...
+	clientConfig.Name = ocsClientConfigMapName
+	clientConfig.Namespace = namespace
+
+	if err := r.Client.Get(r.ctx, client.ObjectKeyFromObject(clientConfig), clientConfig); err != nil {
+		r.Log.Error(err, "failed to get ocs client configmap")
+		return err
+	}
+
+	existingData := maps.Clone(clientConfig.Data)
+	if clientConfig.Data == nil {
```
rchikatw, line 87: how about using `utils.AddAnnotation` ?

controllers/util/k8sutil.go
```diff
@@ -23,6 +23,9 @@ const (
 	// this value is empty if the operator is running with clusterScope.
 	WatchNamespaceEnvVar = "WATCH_NAMESPACE"
 
+	// PodNamespaceEnvVar is the env variable for the pod namespace
+	PodNamespaceEnvVar = "POD_NAMESPACE"
```
rchikatw, line 27: is there any reason we are not using `WatchNamespaceEnvVar`?

rewantsoni: WatchNamespaceEnvVar is a list of Namespace (from what I recall comma-separated list) the operator should be watching for changes. We don't want the namespace in the onboarding job and provider server to be set as watchNamespace.

We could use OperatorNamespace but this will break if storageCluster is deployed in a different namespace from the operator. Hence using PodNamespace will ensure that these components are deployed in the operator namespace and can get Resources (ex: storageCluster for onboarding job) from the same namespace as it is deployed.

controllers/util/k8sutil.go
nb-ohad: I would prefer we should not start having `os.Getenv` all over the codebase. The point of the util is to abstract the way/location we are getting the pod namespace from and to enforce its existence and  I believe it is the correct approach. 

To solve the issue you presented in your last comment we can decide that the env var is not mandatory for all code paths (which seems to be the case) and remove the panic (leaving you with an empty pod namespace in some cases). Then enforce the existence on a `GetPodNamespace` call. This way the assumption is that if you called `GetPodNamespace` the coder must make sure the env var was set.


The change will look like 
```golang
var podNamespace =  os.GetEnv(PodNamespaceEnvVar)

func GetPodNamespace() (string) {
	if podNamespace != "" {
		panic fmt.Errorf("%s must be set", PodNamespaceEnvVar)
	}
	return podNamespace
}
```

nb-ohad:
LGTM

rewantsoni:
resolved conflicts

nb-ohad:
@iamniting Can you please take a look as well?

controllers/storagecluster/storageclient.go
```diff
@@ -2,19 +2,27 @@ package storagecluster
 
 import (
 	"fmt"
+	"strconv"
 
 	ocsclientv1a1 "github.com/red-hat-storage/ocs-client-operator/api/v1alpha1"
 	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
 	"github.com/red-hat-storage/ocs-operator/v4/controllers/util"
 
+	"golang.org/x/exp/maps"
```
iamniting, line 11: We are not using this library in the existing code, We are only using `maps`, pls change this.

iamniting (changes requested):

rewantsoni:
moved unit-tests into a separate commit

iamniting (approved):

PR #2719, opened 2024-07-30
Disable client operator's noobaa subscription management

bernerhat:
Set a configuration for client operator's configmap to not manage noobaa's subscription while in provider mode

### DIFF
--- a/controllers/storagecluster/provider_server.go
+++ b/controllers/storagecluster/provider_server.go
@@ -7,6 +7,7 @@ import (
 	"math/rand"
 	"os"
 	"sort"
+	"strconv"
 	"time"
 
 	"go.uber.org/multierr"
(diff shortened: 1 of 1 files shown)


controllers/storagecluster/provider_server.go
```diff
@@ -42,6 +42,7 @@ const (
 
 	ocsClientConfigMapName = "ocs-client-operator-config"
 	deployCSIKey           = "DEPLOY_CSI"
+	manageNoobaaSub        = "MANAGE_NOOBAA_SUB"
```
nb-ohad, line 45: Why do we use all caps + snake case for configmap keys?

nb-ohad, line 45: Lets not use short names in keys 
```suggestion
	manageNoobaaSub        = "MANAGE_NOOBAA_SUBSCRIPTION"
```

nb-ohad (changes requested):

controllers/storagecluster/provider_server.go
bernerhat: was following the same pattern that was used for CSI above, it can be change imo.

Nikhil-Ladha:
Do we want this change in 4.17 branch? If so, please create a bug and backport the fix.
Thanks.

PR #168, opened 2019-09-27
bugfix - updated name spec of managed resources

umangachapagain:
Signed-off-by: Umanga Chapagain <chapagainumanga@gmail.com>

PR #2727, opened 2024-08-01
Sends and labels `V*ReplicationClass`, `V*SnapClass`, `StorageClass` to client

raaizik:
## Changes

- Sends two VRCs (one for image flattening) and one VGRC. Depends on #2620 
- Labels V*RepClass, V*SnapClass and StorageClass

[RHSTOR-5753](https://issues.redhat.com/browse/RHSTOR-5753), [RHSTOR-5794](https://issues.redhat.com/browse/RHSTOR-5794), [RHSTOR-5754](https://issues.redhat.com/browse/RHSTOR-5754)

### DIFF
--- a/services/provider/server/server.go
+++ b/services/provider/server/server.go
@@ -58,6 +58,10 @@ const (
 	onboardingTicketKeySecret = "onboarding-ticket-key"
 	storageRequestNameLabel   = "ocs.openshift.io/storagerequest-name"
 	notAvailable              = "N/A"
+
+	ramenDRStorageIDKey     = "ramendr.openshift.io/storageID"
+	ramenDRReplicationIDKey = "ramendr.openshift.io/replicationid"
+	ramenDRFlattenModeKey   = "replication.storage.openshift.io/flatten-mode"
 )
 
 const (
@@ -656,9 +660,24 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
 			return nil, status.Error(codes.Internal, msg)
 		}
 	}
+
+	// fetch storage cluster peer to indicate whether replication is enabled
+	scp := &ocsv1.StorageClusterPeerList{}
+	if err = s.client.List(ctx, scp, client.InNamespace(s.namespace)); err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to get StorageClusterPeerList. %v", err)
+	}
+	replicationEnabled := len(scp.Items) > 1
+	var replicationID string
+	if replicationEnabled {
+		replicationID = util.CalculateMD5Hash(req.StorageClaimName)
+	}
+
 	var extR []*pb.ExternalResource
 
 	storageRequestHash := getStorageRequestHash(req.StorageConsumerUUID, req.StorageClaimName)
+	// SID for RamenDR
+	storageID := storageRequestHash
+
 	for _, cephRes := range storageRequest.Status.CephResources {
 		switch cephRes.Kind {
 		case "CephClient":
@@ -690,13 +709,12 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
 			})
 
 		case "CephBlockPoolRadosNamespace":
-
 			rns := &rookCephv1.CephBlockPoolRadosNamespace{}
 			err = s.client.Get(ctx, types.NamespacedName{Name: cephRes.Name, Namespace: s.namespace}, rns)
 			if err != nil {
 				return nil, status.Errorf(codes.Internal, "failed to get %s CephBlockPoolRadosNamespace. %v", cephRes.Name, err)
 			}
-
+			mirroringEnabled := rns.Spec.Mirroring != nil
 			provisionerSecretName := storageClaimCephCsiSecretName("provisioner", storageRequestHash)
 			nodeSecretName := storageClaimCephCsiSecretName("node", storageRequestHash)
 			rbdStorageClassData := map[string]string{
@@ -717,22 +735,48 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
 
 			extR = append(extR,
 				&pb.ExternalResource{
-					Name: "ceph-rbd",
-					Kind: "StorageClass",
-					Data: mustMarshal(rbdStorageClassData),
+					Name:   "ceph-rbd",
+					Kind:   "StorageClass",
+					Data:   mustMarshal(rbdStorageClassData),
+					Labels: getExternalResourceLabels("StorageClass", mirroringEnabled, false, replicationID, storageID),
 				},
 				&pb.ExternalResource{
 					Name: "ceph-rbd",
 					Kind: "VolumeSnapshotClass",
 					Data: mustMarshal(map[string]string{
 						"csi.storage.k8s.io/snapshotter-secret-name": provisionerSecretName,
-					})},
+					}),
+					Labels: getExternalResourceLabels("VolumeSnapshotClass", mirroringEnabled, false, replicationID, storageID)},
 				&pb.ExternalResource{
 					Name: "ceph-rbd",
 					Kind: "VolumeGroupSnapshotClass",
 					Data: mustMarshal(map[string]string{
 						"csi.storage.k8s.io/group-snapshotter-secret-name": provisionerSecretName,
-					})},
+					}),
+					Labels: getExternalResourceLabels("VolumeGroupSnapshotClass", mirroringEnabled, false, replicationID, storageID)},
+				&pb.ExternalResource{
+					Name: "ceph-rbd",
+					Kind: "VolumeReplicationClass",
+					Data: mustMarshal(map[string]string{
+						"replication.storage.openshift.io/replication-secret-name": provisionerSecretName,
+						"mirroringMode": "snapshot",
+					}),
+					Labels: getExternalResourceLabels("VolumeReplicationClass", mirroringEnabled, false, replicationID, storageID),
+					Annotations: map[string]string{
+						"replication.storage.openshift.io/is-default-class": "true",
+					},
+				},
+				&pb.ExternalResource{
+					Name: "ceph-rbd-flatten",
+					Kind: "VolumeReplicationClass",
+					Data: mustMarshal(map[string]string{
+						"replication.storage.openshift.io/replication-secret-name": provisionerSecretName,
+						"mirroringMode": "snapshot",
+						"flattenMode":   "force",
+					}),
+					Labels:      getExternalResourceLabels("VolumeReplicationClass", mirroringEnabled, true, replicationID, storageID),
+					Annotations: map[string]string{},
+				},
 				&pb.ExternalResource{
 					Kind: "ClientProfile",
 					Name: "ceph-rbd",
@@ -777,9 +821,10 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
 
 			extR = append(extR,
 				&pb.ExternalResource{
-					Name: "cephfs",
-					Kind: "StorageClass",
-					Data: mustMarshal(cephfsStorageClassData),
+					Name:   "cephfs",
+					Kind:   "StorageClass",
+					Data:   mustMarshal(cephfsStorageClassData),
+					Labels: getExternalResourceLabels("StorageClass", replicationEnabled, false, replicationID, storageID),
 				},
 				&pb.ExternalResource{
 					Name: cephRes.Name,
@@ -792,13 +837,17 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
 					Kind: "VolumeSnapshotClass",
 					Data: mustMarshal(map[string]string{
 						"csi.storage.k8s.io/snapshotter-secret-name": provisionerSecretName,
-					})},
+					}),
+					Labels: getExternalResourceLabels("VolumeSnapshotClass", replicationEnabled, false, replicationID, storageID),
+				},
 				&pb.ExternalResource{
 					Name: "cephfs",
 					Kind: "VolumeGroupSnapshotClass",
 					Data: mustMarshal(map[string]string{
 						"csi.storage.k8s.io/group-snapshotter-secret-name": provisionerSecretName,
-					})},
+					}),
+					Labels: getExternalResourceLabels("VolumeGroupSnapshotClass", replicationEnabled, false, replicationID, storageID),
+				},
 				&pb.ExternalResource{
 					Kind: "ClientProfile",
 					Name: "cephfs",
@@ -815,6 +864,33 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
 	klog.Infof("successfully returned the storage class claim %q for %q", req.StorageClaimName, req.StorageConsumerUUID)
 	return &pb.StorageClaimConfigResponse{ExternalResource: extR}, nil
 
+}
+
+func getExternalResourceLabels(kind string, isMirroringReplicationEnabled bool, isFlattenMode bool,
+	replicationID string, storageID string) map[string]string {
+	labels := make(map[string]string)
+	switch kind {
+	case "VolumeReplicationClass", "VolumeGroupReplicationClass":
+		if isMirroringReplicationEnabled {
+			labels[ramenDRReplicationIDKey] = replicationID
+			if isFlattenMode {
+				labels[ramenDRFlattenModeKey] = "force"
+			}
+			labels[ramenDRStorageIDKey] = storageID
+		}
+	case "VolumeSnapshotClass", "VolumeGroupSnapshotClass":
+		if isMirroringReplicationEnabled {
+			labels[ramenDRStorageIDKey] = storageID
+		}
+	case "StorageClass":
+		if isMirroringReplicationEnabled {
+			labels[ramenDRStorageIDKey] = storageID
+		}
+
+	default:
+		panic(fmt.Sprintf("unknown storage class kind %q", kind))
+	}
+	return labels
 }
 
 // ReportStatus rpc call to check if a consumer can reach to the provider.
--- a/services/provider/server/server_test.go
+++ b/services/provider/server/server_test.go
@@ -29,9 +29,11 @@ import (
 )
 
 type externalResource struct {
-	Kind string `json:"kind"`
-	Data any    `json:"data"`
-	Name string `json:"name"`
+	Kind        string            `json:"kind"`
+	Data        any               `json:"data"`
+	Name        string            `json:"name"`
+	Labels      map[string]string `json:"labels"`
+	Annotations map[string]string `json:"annotations"`
 }
 
 var serverNamespace = "openshift-storage"
@@ -310,7 +312,10 @@ func TestGetExternalResources(t *testing.T) {
 			data, err := json.Marshal(mockResoruce.Data)
 			assert.NoError(t, err)
 			assert.Equal(t, string(extResource.Data), string(data))
+
 		}
+		assert.Equal(t, extResource.GetLabels(), mockResoruce.Labels)
+		assert.Equal(t, extResource.GetAnnotations(), mockResoruce.Annotations)
 		assert.Equal(t, extResource.Kind, mockResoruce.Kind)
 		assert.Equal(t, extResource.Name, mockResoruce.Name)
 	}
@@ -348,7 +353,8 @@ func TestGetExternalResources(t *testing.T) {
 			assert.NoError(t, err)
 			assert.Equal(t, string(extResource.Data), string(data))
 		}
-
+		assert.Equal(t, extResource.GetLabels(), mockResoruce.Labels)
+		assert.Equal(t, extResource.GetAnnotations(), mockResoruce.Annotations)
 		assert.Equal(t, extResource.Kind, mockResoruce.Kind)
 		assert.Equal(t, extResource.Name, mockResoruce.Name)
 	}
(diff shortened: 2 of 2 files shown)


rewantsoni (changes requested):
For Volume\*ReplicationClass we are sending only the parameters field like we did for Volume\*SnapshotClass and StorageClass, Should we send the entire spec for it or maybe the entire object that will allow us to add labels/annotations required (ramen labels, reclaimspace annotation) from the provider itself and the client would be responsible for adding missing fields like ProvisionerName and namespace for secrets and then creating/upating it.

@nb-ohad WDYT?

services/provider/server/server.go
```diff
@@ -666,7 +666,42 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
...
+				&pb.ExternalResource{
+					Name: "ceph-rbd",
+					Kind: "VolumeGroupReplicationClass",
+					Data: mustMarshal(map[string]string{
+						"replication.storage.openshift.io/group-replication-secret-name": provisionerSecretName,
+						"replication.storage.openshift.io/is-default-class":              "true",
+					}),
+				},
+				&pb.ExternalResource{
+					Name: "ceph-rbd-image-flattening",
```
Rakshith-R, line 696:
```suggestion
					Name: "ceph-rbd-flatten",
```

Please make sure everything matches with https://github.com/red-hat-storage/odf-multicluster-orchestrator/blob/b3df88c23b06d80515f27a36f656296e2d413681/controllers/drpolicy_controller.go#L212

raaizik: I notice the name's suffix is a hashed scheduling interval. Since I am watching the new DRClusterConfig resource (that provides the sched intervals) from the client op, I'd need to override this field there. But I guess it doesn't matter on your end because all you care about is that the client op sends the correct name.

Rakshith-R (commented):
`"replication.storage.openshift.io/is-default-class": "true",`
This should be added as a annotation 

and 
`"replication.storage.openshift.io/flatten-mode": "force",`

This as a label

hope both of the above will be handled accordingly in consumer side.

Other than the above, everything else looks good to me.

raaizik:
> `"replication.storage.openshift.io/is-default-class": "true",` This should be added as a annotation
> 
> and `"replication.storage.openshift.io/flatten-mode": "force",`
> 
> This as a label
> 
> hope both of the above will be handled accordingly in consumer side.
> 
> Other than the above, everything else looks good to me.

I'm aware and it will.

nb-ohad:
> For Volume*ReplicationClass we are sending only the parameters field like we did for Volume*SnapshotClass and StorageClass, Should we send the entire spec for it or maybe the entire object that will allow us to add labels/annotations required (ramen labels, reclaimspace annotation) from the provider itself and the client would be responsible for adding missing fields like ProvisionerName and namespace for secrets and then creating/upating it.
> 
> @nb-ohad WDYT?

I will start with the later proposal first:
> maybe the entire object that will allow us to add labels/annotations required 

We do not want to do that, an object contains not just desired state but also status and some metadata that is local to the provider cluster. In internal discussions, we already acknowledged the need to send labels and annotations but not as part of the serialized spec we send (the data field of the ExternalResource struct). The plan is to add it to the metadata of the `ExternalResource` object, similar to the kind and name. Until that is implemented we will keep on having logic in the client to add the missing metadata. 

> Should we send the entire spec
This depends on the structure of the type, we originally used `data` because that was the desired state for configmaps and secrets. If the `Volume*ReplicationClass` has a formal spec field it will make sense to serialize it instead of a `map[string]string`. 
As a matter of fact, we are already doing something similar for the `ClusterResourceQuota` CR (https://github.com/red-hat-storage/ocs-operator/blob/main/services/provider/server/server.go#L376-L399)

rewantsoni:
/hold
Holding as per https://github.com/red-hat-storage/ocs-operator/pull/2620#discussion_r1749185716
As the comment requests to add metadata (labels and annotation) and we are sending the labels and annotations both in the Data(Spec) field.

raaizik:
> /hold Holding as per [#2620 (comment)](https://github.com/red-hat-storage/ocs-operator/pull/2620#discussion_r1749185716) As the comment requests to add metadata (labels and annotation) and we are sending the labels and annotations both in the Data(Spec) field.

Can unhold now 👍

rewantsoni (changes requested):
Should we consider sending the spec of Volume*ReplicationClass instead of spec.parameter from the provider server?

raaizik:
> Should we consider sending the spec of Volume*ReplicationClass instead of spec.parameter from the provider server?

I think we've had a previous discussion regarding this or similar. IMO the server should send only the info that's required and not the entire spec.

services/provider/server/server_test.go
```diff
@@ -1074,6 +1136,10 @@ func TestOCSProviderServerGetStorageClaimConfig(t *testing.T) {
 		data, err := json.Marshal(mockResoruce.Data)
 		assert.NoError(t, err)
 		assert.Equal(t, string(extResource.Data), string(data))
+		assert.NoError(t, err)
+		assert.True(t, reflect.DeepEqual(extResource.Labels, mockResoruce.Labels))
+		assert.NoError(t, err)
+		assert.True(t, reflect.DeepEqual(extResource.Annotations, mockResoruce.Annotations))
```
rewantsoni, line 1142:
```suggestion
		assert.True(t, reflect.DeepEqual(extResource.Labels, mockResoruce.Labels))
		assert.True(t, reflect.DeepEqual(extResource.Annotations, mockResoruce.Annotations))
```

services/provider/server/server_test.go
```diff
@@ -350,7 +353,10 @@ func TestGetExternalResources(t *testing.T) {
 			data, err := json.Marshal(mockResoruce.Data)
 			assert.NoError(t, err)
 			assert.Equal(t, string(extResource.Data), string(data))
+
 		}
+		assert.True(t, reflect.DeepEqual(extResource.Labels, mockResoruce.Labels))
+		assert.True(t, reflect.DeepEqual(extResource.Annotations, mockResoruce.Annotations))
```
rewantsoni, line 359: We could use assert.Equal() it uses reflect.DeepEqual internally

services/provider/server/server.go
```diff
@@ -745,6 +774,56 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
...
+					Name: "ceph-rbd-flatten",
+					Kind: "VolumeReplicationClass",
+					Data: mustMarshal(map[string]string{
+						"replication.storage.openshift.io/replication-secret-name": provisionerSecretName,
+						"mirroringMode": "snapshot",
+						"flattenMode":   "force",
+					}),
+					Labels: getExternalResourceLabels("VolumeReplicationClass", replicationEnabled, mirroringEnabled, true, ID, SID),
+					Annotations: map[string]string{
+						"replication.storage.openshift.io/is-default-class": "true",
```
rewantsoni, line 799: from what I recall, we don't require this annotation for flatten VRC

services/provider/server/server.go
```diff
@@ -745,6 +774,56 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
...
+					Name: "ceph-rbd-flatten",
+					Kind: "VolumeGroupReplicationClass",
+					Data: mustMarshal(map[string]string{
+						"replication.storage.openshift.io/group-replication-secret-name": provisionerSecretName,
+						"mirroringMode": "snapshot",
+						"flattenMode":   "force",
+					}),
+					Labels: getExternalResourceLabels("VolumeGroupReplicationClass", replicationEnabled, mirroringEnabled, true, ID, SID),
+					Annotations: map[string]string{
+						"replication.storage.openshift.io/is-default-class": "true",
```
rewantsoni, line 824: from what I recall, we don't require this annotation for flatten VGRC

services/provider/server/server.go
```diff
@@ -832,6 +911,39 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
 
 }
 
+func getExternalResourceLabels(kind string, isReplicationEnabled bool, isMirroringEnabled bool, isFlattenMode bool,
```
rewantsoni, line 914: We don't need to differentiate between the Replication and Mirroring flag, we can have one mirroring flag (which will check the cephblockpool) and add labels based on that.

services/provider/server/server.go
```diff
@@ -668,10 +669,32 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
...
+		storageClaimName, err := json.Marshal(req.StorageClaimName)
+		if err != nil {
+			klog.Errorf("failed to marshal a name for a storage claim request based on %v. %v", s, err)
+			panic("failed to marshal storage class request name")
+		}
+		md5Sum := md5.Sum(storageClaimName)
+		ID = hex.EncodeToString(md5Sum[:16])
+	} // todo storageID in provider is planned to be string(r.storageClaim.UID)
+	// SID for RamenDR
+	SID := req.StorageConsumerUUID
```
rewantsoni, line 690: StorageID is used to match V*C to StorageClass, we don't need it to be storageClaim UID, we can have it as storageRequest UID as well.

services/provider/server/server_test.go
raaizik: Both `labels` and `annotations` are of type map, which means that in order to use `assert.Equal()` I'd need to convert from map to []byte which is not straightforward. Therefore I've chosen this assertion instead so it's compact as a one-liner.

services/provider/server/server.go
raaizik: I'm pretty sure it is needed. @Rakshith-R

services/provider/server/server.go
raaizik: See [reply](https://github.com/red-hat-storage/ocs-operator/pull/2727#discussion_r1771283490)

services/provider/server/server.go
raaizik: We've already agreed that it'll be `string(r.storageClaim.UID)` in [#168](https://github.com/red-hat-storage/ocs-client-operator/pull/168). Eventually the value that will be assigned to this ID is the one from the client, so it actually doesn't matter what value is chosen on this end.

services/provider/server/server.go
raaizik: From my perspective, there's no other way to handle these flags. Replication is `Vol*RepClass` specific, while Mirroring is aimed for both `Vol*RepClass` and `Vol*SnapClass`.

services/provider/server/server.go
rewantsoni: As the concept for storageClaim might go away in the future because of convergence, we should pivot and not use it

services/provider/server/server.go
rewantsoni: MCO is creating two VRC's one with flatten mode and one without flatten.
The one that is without flatten mode has the annotation the other one doesn't. See [here](https://github.com/red-hat-storage/odf-multicluster-orchestrator/blob/b3df88c23b06d80515f27a36f656296e2d413681/controllers/drpolicy_controller.go#L170-L195) and [here](https://github.com/red-hat-storage/odf-multicluster-orchestrator/blob/b3df88c23b06d80515f27a36f656296e2d413681/controllers/drpolicy_controller.go#L212-L226)
We are setting the annotations to be empty when creating the flatten mode VRC.

Also the same in mentioned in the design doc [Changes required](https://docs.google.com/document/d/1W9rnZNF82oUlIX7c0JJJrNDwx7pJfLU5X3suC-1jfqs/edit#heading=h.kf5l0za50hwu)

services/provider/server/server.go
rewantsoni: See [reply](https://github.com/red-hat-storage/ocs-operator/pull/2727#discussion_r1771526698)

services/provider/server/server.go
rewantsoni: Your replication flag depends on storageClusterPeer, whereas the mirroring flag is derived from CephBlockPool.

Enabling mirroring on the cephblockpool is the responsibility of the storageClusterPeer. Isn't both pointing to the same field eventually?

I think we should add the labels for rbd-based V*RC/V*SC and SC based only on the mirroring field. It is possible that a particular blockpool is not selected for mirroring from StorageClusterPeer CR. See [here](https://github.com/red-hat-storage/ocs-operator/blob/main/api/v1/storageclusterpeer_types.go#L50-L53)

I was going through the PR and I am not able to find the labels applied for CephFS-based V*SC and SC. Do we not need them?

services/provider/server/server.go
raaizik: Even so, the final value assigned would be determined on the client.

rewantsoni: Why do we need to determine the label value on the client if we can do it from the provider itself?

services/provider/server/server.go
rewantsoni: >I am not aware of such thing. How does it fetch this flag exactly? Is it merged already?

This was discussed in the design discussion. Each StorageClusterPeer will enable mirroring on the cephblockpool based on the label selector.

For rbd we can use the mirroring field on the cephblockpool to add ramenRelated labels, we don't need another replication flag explicitly.

raaizik: I think you're misunderstanding. I am using the Mirroring flag solely for the StorageID label. It's unrelated to RepID.

services/provider/server/server.go
raaizik: We could, I'm saying with the current state of things this is what would happen. After this PR will get merged, the value assigned from here will be propagated to the client.

services/provider/server/server.go
rewantsoni: Why do we need two flags for the same set of ramen labels? From what I understand we are creating a differentiation which we don't need. It will just complicate the things. 

As an example what if we have 2 cephblockpool a and b the storageclusterpeer is configured only for blockpool a.

There are two clients 1 and 2. 1 uses blockpool a, 2 uses blockpool b.

If we keep the flags separate,

For client 1 we will enable both replication and storage id label

For client 2 we will enable only replication id label

In the second case we don't require any ramen related labels to be added as that blockpool is not configured for mirroring. 

Hence we don't need to check for both storageclusterpeer and cephblockpool. Checking only the cephblockpool will work

services/provider/server/server.go
rewantsoni: Yes, I agree that we could but I think as storageID is an arbitrary value used by ramen to cross-link the V*RC/V*SC and SC, the hash of storageRequest UID perfectly fits our use case. I don't think we need to change it further on the client. Let's wait for others to review this.

services/provider/server/server.go
raaizik: > Each StorageClusterPeer will enable mirroring on the cephblockpool based on the label selector.

It doesn't make much sense to me to retrieve this flag by checking each SCP instead of a single CBP.

rewantsoni: For rbd,
StorageClass will created from a radosnamesapce on a blockpool. From my perspective to populate the mirroring and replication flag, we should check if the mirroring is enabled for radosnamespace [spec.mirroring ](https://github.com/rook/rook/pull/14701)

For cephfs,
Since there is not need to perform any actions for this on ceph level unlike rbd, we should enable the mirroring and replication flag for this as soon as we have a storageClusterPeer present

services/provider/server/server.go
```diff
@@ -690,13 +705,12 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
...
 
 		case "CephBlockPoolRadosNamespace":
-
 			rns := &rookCephv1.CephBlockPoolRadosNamespace{}
 			err = s.client.Get(ctx, types.NamespacedName{Name: cephRes.Name, Namespace: s.namespace}, rns)
 			if err != nil {
 				return nil, status.Errorf(codes.Internal, "failed to get %s CephBlockPoolRadosNamespace. %v", cephRes.Name, err)
 			}
-
+			mirroringEnabled := len(rns.Spec.Mirroring.Mode) > 0
```
rewantsoni, line 713: Could you change it to If rns.Spec.Mirroring is not nil then mirroring is enabled

services/provider/server/server.go
```diff
@@ -817,6 +877,35 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
 
 }
 
+func getExternalResourceLabels(kind string, isMirroringReplicationEnabled bool, isFlattenMode bool,
+	replicationID string, storageID string) map[string]string {
+	labels := make(map[string]string)
+	switch kind {
+	case "VolumeReplicationClass", "VolumeGroupReplicationClass":
+		if isMirroringReplicationEnabled {
+			replicationID := replicationID
```
rewantsoni, line 886: I think this is redundant

rewantsoni (changes requested):

services/provider/server/server.go
raaizik: I think this is more precise check (it checks both)

services/provider/server/server.go
raaizik: This is the flag for setting up those IDs

services/provider/server/server.go
```diff
@@ -666,7 +666,36 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
...
+					Kind: "VolumeReplicationClass",
+					Data: mustMarshal(map[string]string{
+						"replication.storage.openshift.io/replication-secret-name": provisionerSecretName,
+					}),
+				},
+				&pb.ExternalResource{
+					Name: "ceph-rbd-image-flattening",
+					Kind: "VolumeReplicationClass",
+					Data: mustMarshal(map[string]string{
+						"replication.storage.openshift.io/replication-secret-name": provisionerSecretName,
```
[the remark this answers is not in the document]
rewantsoni: @raaizik I think this was missed

rewantsoni (changes requested):

rewantsoni:
LGTM

services/provider/server/server.go
```diff
@@ -792,13 +856,17 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
 					Kind: "VolumeSnapshotClass",
 					Data: mustMarshal(map[string]string{
 						"csi.storage.k8s.io/snapshotter-secret-name": provisionerSecretName,
-					})},
+					}),
+					Labels: getExternalResourceLabels("VolumeGroupReplicationClass", replicationEnabled, false, replicationID, storageID),
+				},
```
rewantsoni, line 861:
```suggestion
					Labels: getExternalResourceLabels("VolumeSnapshotClass", replicationEnabled, false, replicationID, storageID),
				},
```

services/provider/server/server.go
```diff
@@ -792,13 +856,17 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
...
 				&pb.ExternalResource{
 					Name: "cephfs",
 					Kind: "VolumeGroupSnapshotClass",
 					Data: mustMarshal(map[string]string{
 						"csi.storage.k8s.io/group-snapshotter-secret-name": provisionerSecretName,
-					})},
+					}),
+					Labels: getExternalResourceLabels("VolumeGroupReplicationClass", replicationEnabled, false, replicationID, storageID),
```
rewantsoni, line 868:
```suggestion
					Labels: getExternalResourceLabels("VolumeGroupSnapshotClass", replicationEnabled, false, replicationID, storageID),
```

rewantsoni (changes requested):

services/provider/server/server.go
raaizik: The naming will be handled on consumer's end. Could you please resolve? @Rakshith-R

Rakshith-R: I think i already mentioned it looks good to me here 
https://github.com/red-hat-storage/ocs-operator/pull/2727#pullrequestreview-2224881050

services/provider/server/server.go
```diff
@@ -717,22 +731,71 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
...
+					Name: "ceph-rbd-flatten",
+					Kind: "VolumeGroupReplicationClass",
+					Data: mustMarshal(map[string]string{
+						"replication.storage.openshift.io/group-replication-secret-name": provisionerSecretName,
+						"mirroringMode": "snapshot",
+						"flattenMode":   "force",
+					}),
+					Labels:      getExternalResourceLabels("VolumeGroupReplicationClass", mirroringEnabled, true, replicationID, storageID),
+					Annotations: map[string]string{},
+				},
```
Madhu-1, line 798: we dont have voluemegroupreplicationclass in 4.18 isnt it? can we remove this and add it later?

services/provider/server/server.go
```diff
@@ -792,13 +856,17 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
...
+					Labels: getExternalResourceLabels("VolumeSnapshotClass", replicationEnabled, false, replicationID, storageID),
+				},
 				&pb.ExternalResource{
 					Name: "cephfs",
 					Kind: "VolumeGroupSnapshotClass",
 					Data: mustMarshal(map[string]string{
 						"csi.storage.k8s.io/group-snapshotter-secret-name": provisionerSecretName,
-					})},
+					}),
+					Labels: getExternalResourceLabels("VolumeGroupSnapshotClass", replicationEnabled, false, replicationID, storageID),
```
Madhu-1, line 868: IMO we should use consts for all class kinds, we are passing it as string in multiple places.

services/provider/server/server.go
```diff
@@ -817,6 +885,33 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
...
+		if isMirroringReplicationEnabled {
+			labels["ramendr.openshift.io/replicationid"] = replicationID
+			if isFlattenMode {
+				labels["replication.storage.openshift.io/flatten-mode"] = "force"
+			}
+			labels["ramendr.openshift.io/storageID"] = storageID
+		}
+	case "VolumeSnapshotClass", "VolumeGroupSnapshotClass":
+		if isMirroringReplicationEnabled {
+			labels["ramendr.openshift.io/storageID"] = storageID
```
Madhu-1, line 902: can we have a const for the label keys?

services/provider/server/server.go
raaizik: If you mean replace it in all places not just the function call -- I did that at first, but @nb-ohad opposed to the idea

Madhu-1: Yes we can easily make mistake when using string in multiple places. we need to have a var or const for it.

Madhu-1 (approved):
LGTM except https://github.com/red-hat-storage/ocs-operator/pull/2727/files#r1830834991

services/provider/server/server.go
raaizik: @rewantsoni

PR #2680, opened 2024-07-02
Added support for ocs provider server to fetch noobaa client resources

ezio-auditore:
1. Added an annotation `is-local-client:true` for local storage client
2. Create a Noobaa Account CR for auth_token generation of remote noobaa client
    a.https://github.com/noobaa/noobaa-operator/pull/1389
4. Assemble noobaa remote auth token and management address to be sent via `GetStorageConfig`
5. Updated test cases

Part of : [RHSTOR-5187](https://issues.redhat.com//browse/RHSTOR-5187)

PR #2733, opened 2024-08-07
Added support for ocs provider server to fetch noobaa client resources

bernerhat:
Continuation PR for #2680

### DIFF
--- a/controllers/storageconsumer/storageconsumer_controller.go
+++ b/controllers/storageconsumer/storageconsumer_controller.go
@@ -21,9 +21,11 @@ import (
 	"crypto/md5"
 	"encoding/hex"
 	"encoding/json"
+	"fmt"
 
 	"github.com/go-logr/logr"
 	"github.com/red-hat-storage/ocs-operator/api/v4/v1alpha1"
+	"github.com/red-hat-storage/ocs-operator/v4/controllers/util"
 	rookCephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -36,6 +38,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
+	nbv1 "github.com/noobaa/noobaa-operator/v5/pkg/apis/noobaa/v1alpha1"
 	ocsv1alpha1 "github.com/red-hat-storage/ocs-operator/api/v4/v1alpha1"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
@@ -59,12 +62,14 @@ type StorageConsumerReconciler struct {
 	storageConsumer         *ocsv1alpha1.StorageConsumer
 	cephClientHealthChecker *rookCephv1.CephClient
 	namespace               string
+	noobaaAccount           *nbv1.NooBaaAccount
 }
 
 //+kubebuilder:rbac:groups=ocs.openshift.io,resources=storageconsumers,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=ceph.rook.io,resources=cephclients,verbs=get;list;watch;create;update;delete
 //+kubebuilder:rbac:groups=ocs.openshift.io,resources=storageconsumers/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=ocs.openshift.io,resources=storagerequests,verbs=get;list;
+// +kubebuilder:rbac:groups=noobaa.io,resources=noobaaaccounts,verbs=get;list;watch;create;update;delete
 
 // Reconcile reads that state of the cluster for a StorageConsumer object and makes changes based on the state read
 // and what is in the StorageConsumer.Spec
@@ -127,6 +132,10 @@ func (r *StorageConsumerReconciler) initReconciler(request reconcile.Request) {
 	r.cephClientHealthChecker = &rookCephv1.CephClient{}
 	r.cephClientHealthChecker.Name = GenerateHashForCephClient(r.storageConsumer.Name, "global")
 	r.cephClientHealthChecker.Namespace = r.namespace
+
+	r.noobaaAccount = &nbv1.NooBaaAccount{}
+	r.noobaaAccount.Name = "noobaa-remote-" + r.storageConsumer.Name
+	r.noobaaAccount.Namespace = r.storageConsumer.Namespace
 }
 
 func (r *StorageConsumerReconciler) reconcilePhases() (reconcile.Result, error) {
@@ -142,6 +151,10 @@ func (r *StorageConsumerReconciler) reconcilePhases() (reconcile.Result, error) 
 	if r.storageConsumer.GetDeletionTimestamp().IsZero() {
 
 		if err := r.reconcileCephClientHealthChecker(); err != nil {
+			return reconcile.Result{}, err
+		}
+
+		if err := r.reconcileNoobaaAccount(); err != nil {
 			return reconcile.Result{}, err
 		}
 
@@ -201,6 +214,25 @@ func (r *StorageConsumerReconciler) reconcileCephClientHealthChecker() error {
 	}
 
 	r.setCephResourceStatus(r.cephClientHealthChecker.Name, "CephClient", phase, nil)
+
+	return nil
+}
+
+func (r *StorageConsumerReconciler) reconcileNoobaaAccount() error {
+	_, err := ctrl.CreateOrUpdate(r.ctx, r.Client, r.noobaaAccount, func() error {
+		if err := r.own(r.noobaaAccount); err != nil {
+			return err
+		}
+		// the following annotation will enable noobaa-operator to create a auth_token secret based on this account
+		util.AddAnnotation(r.noobaaAccount, "remote-operator", "true")
+		return nil
+	})
+	if err != nil {
+		return fmt.Errorf("failed to create noobaa account for storageConsumer %v: %v", r.storageConsumer.Name, err)
+	}
+
+	phase := string(r.noobaaAccount.Status.Phase)
+	r.setCephResourceStatus(r.noobaaAccount.Name, "NooBaaAccount", phase, nil)
 
 	return nil
 }
--- a/rbac/provider-role.yaml
+++ b/rbac/provider-role.yaml
@@ -68,3 +68,19 @@ rules:
   verbs:
   - get
   - list
+- apiGroups:
+  - noobaa.io
+  resources:
+  - noobaaaccounts
+  verbs:
+  - get
+  - list
+  - create
+  - delete
+- apiGroups:
+  - route.openshift.io
+  resources:
+  - routes
+  verbs:
+  - get
+  - list
--- a/rbac/provider-role.yaml
+++ b/rbac/provider-role.yaml
@@ -69,15 +69,6 @@ rules:
   - get
   - list
 - apiGroups:
-  - noobaa.io
-  resources:
-  - noobaaaccounts
-  verbs:
-  - get
-  - list
-  - create
-  - delete
-- apiGroups:
   - route.openshift.io
   resources:
   - routes
--- a/services/provider/server/consumer.go
+++ b/services/provider/server/consumer.go
@@ -4,12 +4,9 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"strings"
 	"sync"
 
-	nbv1 "github.com/noobaa/noobaa-operator/v5/pkg/apis/noobaa/v1alpha1"
 	ocsv1alpha1 "github.com/red-hat-storage/ocs-operator/api/v4/v1alpha1"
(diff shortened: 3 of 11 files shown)


leelavg:
> Continuation PR for #2680
- should #2680 be also in for this PR?

services/provider/server/consumer.go
```diff
@@ -226,3 +229,47 @@ func (c *ocsConsumerManager) UpdateConsumerStatus(ctx context.Context, id string
...
 }
+
+func (c *ocsConsumerManager) CreateNoobaaAccount(ctx context.Context, id string) error {
+
+	consumerObj, err := c.Get(ctx, id)
+	if err != nil {
+		return err
+	}
+	consumerClusterID := strings.TrimPrefix(consumerObj.Name, "storageconsumer-")
+	if consumerClusterID != "" && len(consumerClusterID) == 0 {
```
leelavg: checking length would be enough, isn't it?

services/provider/server/server.go
```diff
@@ -146,7 +148,10 @@ func (s *OCSProviderServer) AcknowledgeOnboarding(ctx context.Context, req *pb.A
 		}
 		return nil, status.Errorf(codes.Internal, "Failed to update the storageConsumer. %v", err)
 	}
-
+	// create noobaa account CR
+	if err := s.consumerManager.CreateNoobaaAccount(ctx, req.StorageConsumerUUID); err != nil {
```
leelavg: there is a possibility (if client-op restarts) that this endpoint could be called more than once and we need to handle a resource is already created corresponding to same request.

services/provider/server/consumer.go
```diff
@@ -226,3 +229,47 @@ func (c *ocsConsumerManager) UpdateConsumerStatus(ctx context.Context, id string
...
+		return fmt.Errorf("failed to get clusterID from consumerResource Name: %s %v", consumerObj.Name, err)
+	}
+
+	noobaaAccountName := fmt.Sprintf("noobaa-remote-%s", consumerClusterID)
+	nbAccountObj := &nbv1.NooBaaAccount{}
+	nbAccountObj.Name = noobaaAccountName
+	nbAccountObj.Namespace = consumerObj.Namespace
+	// the following annotation will enable noobaa-operator to create a auth_token secret based on this account
+	util.AddAnnotation(nbAccountObj, "remote-operator", "true")
```
leelavg: i see this annotation is not conditional but the comment says this is to track something, what is the differentiation you are trying to make?

services/provider/server/server.go
leelavg: if creation of NoobaaAccount isn't conditional why can't the whole operation be offloaded to StorageConsumer controller? I believe this is what Ohad mentioned to me once.

thinking about it, as a side-effect/requirement to support StorageConsumer a couple of Rook resources are created and likewise if NB is also a requirement the controller is a good place. Deletion could be done via OwnerRef and not in a stepped phase, wdyt?

services/provider/server/server.go
```diff
@@ -400,6 +416,59 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
 
 	}
 
+	// Fetch noobaa remote secret and management address and append to extResources
+	noobaaOperatorSecret := &v1.Secret{}
+	clusterID := strings.TrimPrefix(consumerResource.Name, "storageconsumer-")
+	if clusterID != "" && len(clusterID) == 0 {
+		return nil, fmt.Errorf("failed to get clusterID from consumerResource Name: %s %v", consumerResource.Name, err)
+	}
```
leelavg: this block is already repeated twice. client clusterid now can be directly queried from status section of storageconsumer.

services/provider/server/server_test.go
```diff
@@ -24,9 +26,9 @@ import (
 )
 
 type externalResource struct {
-	Kind string            `json:"kind"`
-	Data map[string]string `json:"data"`
-	Name string            `json:"name"`
+	Kind string `json:"kind"`
+	Data []byte `json:"data"`
```
leelavg, line 30: you are replacing one dependency w/ another, in test this could be `Data any json:"data"` as the mustMarshal is already a generic function and the result is `[]byte` that can be used w/o any conversion in protobuf.

leelavg (commented):
after seeing the pattern I believe it might be good to have this in storageconsumer controller and may I know what is the diff b/n this & the referred PR?

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -127,6 +131,10 @@ func (r *StorageConsumerReconciler) initReconciler(request reconcile.Request) {
 	r.cephClientHealthChecker = &rookCephv1.CephClient{}
 	r.cephClientHealthChecker.Name = GenerateHashForCephClient(r.storageConsumer.Name, "global")
 	r.cephClientHealthChecker.Namespace = r.namespace
+
+	r.noobaaAccount = &nbv1.NooBaaAccount{}
+	r.noobaaAccount.Name = "noobaa-remote-" + r.storageConsumer.Status.Client.ClusterID
```
leelavg, line 136: there is a subtle bug here, the status will be available here only after atleast a single heartbeat. you may need to set this based on the presence of the ID during reconcile directly.

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -157,13 +165,30 @@ func (r *StorageConsumerReconciler) reconcilePhases() (reconcile.Result, error)
...
 
 	return reconcile.Result{}, nil
 }
 
+func (r *StorageConsumerReconciler) reconcileNoobaaAccount() error {
+	_, err := ctrl.CreateOrUpdate(r.ctx, r.Client, r.noobaaAccount, func() error {
+		// the following annotation will enable noobaa-operator to create a auth_token secret based on this account
+		util.AddAnnotation(r.noobaaAccount, "remote-operator", "true")
+		err := r.own(r.noobaaAccount)
+		return err
```
leelavg, line 184: I don't see a reason to split this, maybe `return r.own(r.noobaaAccount)`?

services/provider/server/server.go
```diff
@@ -417,6 +425,51 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
+	clusterID := consumerResource.Status.Client.ClusterID
+	noobaaOperatorSecret := &v1.Secret{}
+	noobaaOperatorSecretName := fmt.Sprintf("noobaa-remote-join-secret-%s", clusterID)
+	err = s.client.Get(ctx, types.NamespacedName{Name: noobaaOperatorSecretName, Namespace: s.namespace}, noobaaOperatorSecret)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get %s secret. %v", noobaaOperatorSecretName, err)
+	}
+
+	authToken, ok := noobaaOperatorSecret.Data["auth_token"]
+	if !ok || len(authToken) == 0 {
```
leelavg, line 438: does this happen anytime that secret is created but data will not exist?

services/provider/server/server.go
```diff
@@ -417,6 +425,51 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
+		return nil, fmt.Errorf("failed to get noobaa-mgmt route. %v", err)
+	}
+	if noobaMgmtRoute.Status.Ingress == nil || len(noobaMgmtRoute.Status.Ingress) == 0 {
+		return nil, fmt.Errorf("no Ingress available in noobaa-mgmt route")
+	}
+
+	noobaaMgmtAddress := noobaMgmtRoute.Status.Ingress[0].Host
+	if noobaaMgmtAddress == "" {
+		return nil, fmt.Errorf("no Host found in noobaa-mgmt route Ingress")
+	}
```
leelavg, line 454: isn't it possible to just create a secret w/ required info from Noobaa itself and send that only?

services/provider/server/server_test.go
leelavg, line 1: is this test file rebased?

services/provider/server/server.go
```diff
@@ -244,6 +244,14 @@ func newClient() (client.Client, error) {
...
 		return nil, fmt.Errorf("failed to add ocsv1 to scheme. %v", err)
 	}
+	err = routev1.AddToScheme(scheme)
+	if err != nil {
+		return nil, fmt.Errorf("failed to add routev1 to scheme. %v", err)
+	}
+	err = nbapis.AddToScheme(scheme)
+	if err != nil {
+		return nil, fmt.Errorf("failed to add nbapis to scheme. %v", err)
+	}
```
leelavg, line 254: you aren't querying noobaa api's isn't it?

rbac/provider-role.yaml
```diff
@@ -68,3 +68,19 @@ rules:
...
   - list
+- apiGroups:
+  - noobaa.io
+  resources:
+  - noobaaaccounts
+  verbs:
+  - get
+  - list
+  - create
+  - delete
```
leelavg, line 79: based on code, is this still required? do we already have rbac for secrets?

services/provider/server/server.go
nb-ohad: Why does it matter? adding protection against this case will not harm us

services/provider/server/server.go
nb-ohad: No, Noobaa don't create a join secret, it creates an account secret and we need to translate it into a join secret format.

services/provider/server/server.go
nb-ohad: Lets change "nbapi" to "noobaa api"

services/provider/server/server.go
```diff
@@ -417,6 +425,51 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
 	}
 
+	// Fetch noobaa remote secret and management address and append to extResources
+	clusterID := consumerResource.Status.Client.ClusterID
+	noobaaOperatorSecret := &v1.Secret{}
+	noobaaOperatorSecretName := fmt.Sprintf("noobaa-remote-join-secret-%s", clusterID)
+	err = s.client.Get(ctx, types.NamespacedName{Name: noobaaOperatorSecretName, Namespace: s.namespace}, noobaaOperatorSecret)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get %s secret. %v", noobaaOperatorSecretName, err)
+	}
```
nb-ohad, line 435:
```suggestion
	noobaaOperatorSecret := &v1.Secret{}
	noobaaOperatorSecret.Name = fmt.Sprintf("noobaa-remote-join-secret-%s", clusterID)
	noobaaOperatorSecret.Namespace = s.namespace
	
	if err := s.client.Get(ctx, client.ObjectKeyFromObject(noobaaOperatorSecret), noobaaOperatorSecret); err != nil {
		return nil, fmt.Errorf("failed to get %s secret. %v", noobaaOperatorSecretName, err)
	}
```

services/provider/server/server.go
```diff
@@ -417,6 +425,51 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
+	authToken, ok := noobaaOperatorSecret.Data["auth_token"]
+	if !ok || len(authToken) == 0 {
+		return nil, fmt.Errorf("auth_token not found in %s secret", noobaaOperatorSecretName)
+	}
+
+	noobaMgmtRoute := &routev1.Route{}
+	err = s.client.Get(ctx, types.NamespacedName{Name: "noobaa-mgmt", Namespace: s.namespace}, noobaMgmtRoute)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get noobaa-mgmt route. %v", err)
+	}
```
nb-ohad, line 446:
```suggestion
	noobaMgmtRoute := &routev1.Route{}
	noobaMgmtRoute.Name = "noobaa-mgmt"
	noobaMgmtRoute.Namespace = s.namespce
	
	if err = s.client.Get(ctx, client.ObjectKeyFromObject(noobaMgmtRoute). noobaMgmtRoute); err != nil {
		return nil, fmt.Errorf("failed to get noobaa-mgmt route. %v", err)
	}
```

services/provider/server/server.go
```diff
@@ -417,6 +425,51 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
+	if noobaMgmtRoute.Status.Ingress == nil || len(noobaMgmtRoute.Status.Ingress) == 0 {
+		return nil, fmt.Errorf("no Ingress available in noobaa-mgmt route")
+	}
+
+	noobaaMgmtAddress := noobaMgmtRoute.Status.Ingress[0].Host
```
nb-ohad, line 451: Are we assuming we should always use the first ingress host? Don't we need to search for it by name?

nb-ohad (changes requested):

services/provider/server/server.go
nb-ohad: The name will be identical, so there is no way we will create 2 separate accounts.

services/provider/server/consumer.go
nb-ohad: It is conditional, but we need the condition to be true in our case as we ant the information for the auth secret

services/provider/server/server.go
nb-ohad: That change was done. resolving

PR #2743, opened 2024-08-09
PoC: Dymanically add watch for clusterclaim CRD

umangachapagain:


controllers/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -264,14 +310,21 @@ func (r *OCSInitializationReconciler) Reconcile(ctx context.Context, request rec
 
 // SetupWithManager sets up a controller with a manager
 func (r *OCSInitializationReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	cluster, err := cluster.New(mgr.GetConfig(), func(options *cluster.Options) {
+		options.Scheme = mgr.GetScheme()
+	})
```
iamniting, line 315: How about we fetch it in main and pass it to all controllers and it can be used in a storagecluster controller as well.

controllers/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -63,6 +68,8 @@ func InitNamespacedName() types.NamespacedName {
 // nolint:revive
 type OCSInitializationReconciler struct {
 	client.Client
+	cluster           cluster.Cluster
```
iamniting, line 71: Can you add a comment on what does it holds?

iamniting (changes requested):

controllers/ocsinitialization/ocsinitialization_controller.go
leelavg, line 315: @umangachapagain it's been sometime that I looked into controller-runtime (during ODFMS), pls excuse if I'm a bit wrong here.

You are using only Scheme from manager but not the cluster which is auto created and has a cache attached to it. The controller that is being saved on reconciler could probably be configured against managers' cluster cache but may not be against this newly created cluster cache.

If above is correct, we may want to use a single cache across all controllers?

controllers/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -152,6 +159,45 @@ func (r *OCSInitializationReconciler) Reconcile(ctx context.Context, request rec
...
+			},
+		}
+		if err := r.controller.Watch(source.Kind[client.Object](r.cluster.GetCache(),
+			&apiextensionsv1.CustomResourceDefinition{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: ClusterClaimCrdName,
+				},
+			}, crdHandler, crdPredicate)); err != nil {
+			return reconcile.Result{}, fmt.Errorf("unable to watch CRD")
+		}
```
nb-ohad, line 198: @umangachapagain This will not achieve the desired result. You cannot just add watches to a controller after the manager started, the cache will become incoherent and inconsistent. The only way to do it within the process is to close all controllers stop and reset the manager then start the manager again. What you are doing here can only work if the client we use is cacheless, but then the informers will not work properly and there will be a runtime penalty for lists and gets

nb-ohad (changes requested):

PR #2744, opened 2024-08-12
RequiredDropCapabilities set to ALL for rookcephcsi scc

ShravaniVangur:
In line with the principle of least privilege, permissions should be granted with only the minimal access necessary to perform the required tasks. Previously, the security context constraints for rook-ceph-csi did not have RequiredDropCapabilities set. This commit ensures that it is now set to "ALL".

**Output:** 
`oc describe scc rook-ceph-csi | grep " Required Drop Capabilities"
  Required Drop Capabilities:			ALL`


Nikhil-Ladha:
Please remove the bug name from the PR title.
And, a description to the commit.
Also, prefer not to capitalise first letter of the commit title.

Nikhil-Ladha:
Also, can you please explain a bit more in the PR description as to why was this change needed?

iamniting (approved):

controllers/ocsinitialization/sccs.go
```diff
@@ -66,6 +67,7 @@ func newRookCephCSISCC(namespace string) *secv1.SecurityContextConstraints {
 	}
 
 	scc, _ := cephcsi.NewSecurityContextConstraints(rookValues)
+	scc.RequiredDropCapabilities = []corev1.Capability{"ALL"}
```
Madhu-1, line 70: ideally, this change should have gone into cephcsi not here (but this is also acceptable as a short-term solution), Have we tested this change and ensured that no regression in PVC workflow? if yes, we can get this in.

travisn: Agreed, this should be done in cephcsi so upstream gets the change as well. Can we do that now? For 4.17 we still have time. 

Anytime privileges are reduced, it can easily cause a regression. We will need confirmation that all relevant scenarios have been tested.

iamniting: @ShravaniVangur Can you pls test all required scenarios with this change then you can implement the change in the upstream APIS and bring them here in the ocs.

ShravaniVangur: Sure will make these changes in cephcsi. 

The following have been tested for with the changes in this PR: 
- Successful deployment of storagecluster. 
- Creation of PVCs for cephfs and rbd following which pods were created with the respective pvcs.
- Edited the PVC size.
- Taking a snapshot.
- Cloned the PVC.
At each point the provisioner/plugin logs were checked to ensure success of the action. 

@Madhu-1  are there any other PVC tests to be done in upstream or are the above-mentioned checks in downstream enough?

Madhu-1: Thanks for testing above, This should be good enough

iamniting (changes requested):
Pls remove first and second commit and just have the 3rd one only. Also remove the conflicts

iamniting (changes requested):
Can you pls update the vendor changes as well?

iamniting (approved):

iamniting (changes requested):
pls look at the commitlint action and fix commit msg

PR #2747, opened 2024-08-13
remove upmap-read balancer mode

sp98:
upmap-read mode of the balancer isn't supported by the kernel client. There was a recent change on the RADOS side which actually enforced this, see ceph/ceph#57776. Prior to 18.2.4 and 19.1.1 all sorts of incompatibility checks were missing in this area. So we need to revert upmap-read mode for now.


idryomov (approved):
Ack from the Ceph perspective

agarwal-mudit (approved):
/lgtm

PR #2750, opened 2024-08-16
sending subs as part of ext resource

rchikatw:
- Sending the subscription is part of external resource 
- Depends on https://github.com/red-hat-storage/ocs-operator/pull/2676

### DIFF
--- a/services/provider/server/server.go
+++ b/services/provider/server/server.go
@@ -180,11 +180,11 @@ func (s *OCSProviderServer) GetStorageConfig(ctx context.Context, req *pb.Storag
 			return nil, status.Errorf(codes.Internal, "failed to get external resources. %v", err)
 		}
 
-		channelName, err := s.getOCSSubscriptionChannel(ctx)
+		subscriptionSpec, err := s.getOCSSubscriptionSpec(ctx)
 		if err != nil {
 			return nil, status.Errorf(codes.Internal, "Failed to construct status response: %v", err)
 		}
-		desiredClientConfigHash := getDesiredClientConfigHash(channelName, consumerObj)
+		desiredClientConfigHash := getDesiredClientConfigHash(subscriptionSpec.Channel, consumerObj)
 
 		klog.Infof("successfully returned the config details to the consumer.")
 		return &pb.StorageConfigResponse{
@@ -426,6 +426,17 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
 		})
 
 	}
+
+	subscriptionSpec, err := s.getOCSSubscriptionSpec(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get subscription %v", err)
+	}
+
+	extR = append(extR, &pb.ExternalResource{
+		Name: "ocs-operator",
+		Kind: "Subscription",
+		Data: mustMarshal(subscriptionSpec),
+	})
 
 	return extR, nil
 }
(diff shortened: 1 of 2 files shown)


services/provider/server/server.go
```diff
@@ -799,27 +810,24 @@ func (s *OCSProviderServer) ReportStatus(ctx context.Context, req *pb.ReportStat
 		return nil, status.Errorf(codes.Internal, "Failed to update lastHeartbeat payload in the storageConsumer resource: %v", err)
 	}
 
-	channelName, err := s.getOCSSubscriptionChannel(ctx)
-	if err != nil {
-		return nil, status.Errorf(codes.Internal, "Failed to construct status response: %v", err)
-	}
-
-	return &pb.ReportStatusResponse{DesiredClientOperatorChannel: channelName}, nil
+	// This will be updated as part of GetStorageConfig RPC call,
```
leelavg, line 813: "This will be updated as part of GetStorageConfig RPC call".... only in newer clients but upgraded provider will start sending empty channel name to older clients and they may never get updated.

rchikatw: make sense.

leelavg: resolve.

services/provider/server/server.go
```diff
@@ -180,11 +180,11 @@ func (s *OCSProviderServer) GetStorageConfig(ctx context.Context, req *pb.Storag
 			return nil, status.Errorf(codes.Internal, "failed to get external resources. %v", err)
 		}
 
-		channelName, err := s.getOCSSubscriptionChannel(ctx)
+		subscriptionSpec, err := s.getOCSSubscriptionSpec(ctx)
```
nb-ohad, line 183: We should not create an entire spec to get the channel name. preferably have a func to get the channel name then use it here and when creating the spec.

services/provider/server/server.go
```diff
@@ -427,6 +427,17 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
 
 	}
 
+	subscriptionSpec, err := s.getOCSSubscriptionSpec(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get subscription %v", err)
+	}
+
+	extR = append(extR, &pb.ExternalResource{
+		Name: "ocs-operator",
```
nb-ohad, line 436: This name does not make sense, are we going to create/update a subscription called "ocs-operator" on the client side? This should be the name of the subscription you want to create/update on the client side

services/provider/server/server.go
```diff
@@ -814,15 +825,15 @@ func (s *OCSProviderServer) ReportStatus(ctx context.Context, req *pb.ReportStat
 		return nil, status.Errorf(codes.Internal, "Failed to get storageConsumer resource: %v", err)
 	}
 
-	channelName, err := s.getOCSSubscriptionChannel(ctx)
+	subscriptionSpec, err := s.getOCSSubscriptionSpec(ctx)
```
nb-ohad, line 828: We should not create an entire spec to get the channel name. preferably have a func to get the channel name then use it here and when creating the spec.

services/provider/server/server.go
```diff
@@ -427,6 +427,17 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
 
 	}
 
+	subscriptionSpec, err := s.getOCSSubscriptionSpec(ctx)
```
nb-ohad, line 430: Why are we getting the ocs-operator subscription when we need to send the subscription for the client?

nb-ohad (changes requested):

services/provider/server/server.go
rchikatw: This is the existing code, and that's what they are sending in the heartbeat. I have used the same function. I guess ocs-operator is the base for it, and that's the reason they are sending it in the existing code.

rchikatw:
Right now, it needs more time to rethink. We might need to rework after the 4.18 changes. Hence, dropping this change for now and will work accordingly after 4.18 changes.

PR #2754, opened 2024-08-21
kms: skip checking azure kms endpoint reachability

sp98:
Vault URL provided in the KMS settings does not have a port number. As a result, endpoint reachable check in `checkEndpointReachable` function fails.  . This only works when we provide the default port of 443 and along with the URL. For now, we can simply avoid checking azure vault URL for endpoint reachability.

This fixes BZ - https://bugzilla.redhat.com/show_bug.cgi?id=2303619 which is a blocker for Azure customers using KMS


aruniiird (approved):
Thanks @sp98 . LGTM from my side

iamniting (approved):

PR #2761, opened 2024-08-26
Bug 2307232: Allow osd crush reweight

parth-gr:
From 4.17 there is setting introduced for upstream user, to allow if they want to add cutom weights,

For downstream always setting it true


parth-gr:
Rook latest is bringing k8s 1.31.0. That's forcing us to bring up controller runtime to 1.19.0 due to some dependancies. After that controller runtime 1.19.0 runs into issue with OLM 1.28.0 as they are using k8 1.30 something. Normally this doesn't happen but k8s 1.31.0 has some breaking issues coming from 1.30, the same with controller runtime 1.19.0

controllers/storagecluster/cephcluster.go
```diff
@@ -451,6 +451,7 @@ func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, kmsConfigMap *co
 			},
 			Storage: rookCephv1.StorageScopeSpec{
 				AllowDeviceClassUpdate:       sc.Spec.ManagedResources.CephCluster.AllowDeviceClassUpdate,
```
parth-gr, line 453: @travisn this is set from the ceph cluster

>AllowOsdCrushWeightUpdate we always wanted true, right?

travisn (commented):
/lgtm
/approve

iamniting (approved):

PR #2762, opened 2024-08-26
mds: fix mds placement

parth-gr:
mds pod created from UI and default mds pod
can be created at same node

### DIFF
--- a/controllers/defaults/placements.go
+++ b/controllers/defaults/placements.go
@@ -76,7 +76,7 @@ var (
 			},
 			PodAntiAffinity: &corev1.PodAntiAffinity{
 				PreferredDuringSchedulingIgnoredDuringExecution: []corev1.WeightedPodAffinityTerm{
-					getWeightedPodAffinityTerm(100, "rook-ceph-mds"),
+					// left the selector value empty as it will be updated later in the getPlacement()
 				},
 			},
 		},
@@ -158,7 +158,7 @@ func getTopologySpreadConstraintsSpec(maxSkew int32, valueLabels []string) corev
 }
 
 func getWeightedPodAffinityTerm(weight int32, selectorValue ...string) corev1.WeightedPodAffinityTerm {
-	WeightedPodAffinityTerm := corev1.WeightedPodAffinityTerm{
+	return corev1.WeightedPodAffinityTerm{
 		Weight: weight,
 		PodAffinityTerm: corev1.PodAffinityTerm{
 			LabelSelector: &metav1.LabelSelector{
@@ -173,7 +173,24 @@ func getWeightedPodAffinityTerm(weight int32, selectorValue ...string) corev1.We
 			TopologyKey: corev1.LabelHostname,
 		},
 	}
-	return WeightedPodAffinityTerm
+}
+
+func GetMdsWeightedPodAffinityTerm(weight int32, selectorValue ...string) corev1.WeightedPodAffinityTerm {
+	return corev1.WeightedPodAffinityTerm{
+		Weight: weight,
+		PodAffinityTerm: corev1.PodAffinityTerm{
+			LabelSelector: &metav1.LabelSelector{
+				MatchExpressions: []metav1.LabelSelectorRequirement{
+					{
+						Key:      "rook_file_system",
+						Operator: metav1.LabelSelectorOpIn,
+						Values:   selectorValue,
+					},
+				},
+			},
+			TopologyKey: corev1.LabelHostname,
+		},
+	}
 }
 
 func getPodAffinityTerm(selectorValue ...string) corev1.PodAffinityTerm {
--- a/controllers/storagecluster/placement.go
+++ b/controllers/storagecluster/placement.go
@@ -20,6 +20,14 @@ func getPlacement(sc *ocsv1.StorageCluster, component string) rookCephv1.Placeme
 	} else {
 		in := defaults.DaemonPlacements[component]
 		(&in).DeepCopyInto(&placement)
+		// label rook_file_system is added to the mds pod using rook operator
+		if component == "mds" {
+			placement.PodAntiAffinity = &corev1.PodAntiAffinity{
+				PreferredDuringSchedulingIgnoredDuringExecution: []corev1.WeightedPodAffinityTerm{
+					defaults.GetMdsWeightedPodAffinityTerm(100, generateNameForCephFilesystem(sc)),
+				},
+			}
+		}
 	}
 
 	// ignore default PodAntiAffinity mon placement when arbiter is enabled
--- a/controllers/storagecluster/placement_test.go
+++ b/controllers/storagecluster/placement_test.go
@@ -375,9 +375,9 @@ func TestGetPlacement(t *testing.T) {
 									LabelSelector: &metav1.LabelSelector{
 										MatchExpressions: []metav1.LabelSelectorRequirement{
 											{
-												Key:      "app",
+												Key:      "rook_file_system",
 												Operator: metav1.LabelSelectorOpIn,
-												Values:   []string{"rook-ceph-mds"},
+												Values:   []string{"storage-test-cephfilesystem"},
 											},
 										},
 									},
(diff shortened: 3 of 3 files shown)


controllers/defaults/placements.go
```diff
@@ -176,6 +176,25 @@ func getWeightedPodAffinityTerm(weight int32, selectorValue ...string) corev1.We
...
 }
 
+func getMdsWeightedPodAffinityTerm(weight int32, selectorValue ...string) corev1.WeightedPodAffinityTerm {
+	WeightedPodAffinityTerm := corev1.WeightedPodAffinityTerm{
+		Weight: weight,
+		PodAffinityTerm: corev1.PodAffinityTerm{
+			LabelSelector: &metav1.LabelSelector{
+				MatchExpressions: []metav1.LabelSelectorRequirement{
+					{
+						Key:      "app.kubernetes.io/part-of",
```
travisn, line 186: This label seems more clear for the intention that we want antiaffinity to the same filesystem
```suggestion
						Key:      "rook_file_system",
```

controllers/defaults/placements.go
```diff
@@ -76,7 +76,7 @@ var (
 			},
 			PodAntiAffinity: &corev1.PodAntiAffinity{
 				PreferredDuringSchedulingIgnoredDuringExecution: []corev1.WeightedPodAffinityTerm{
-					getWeightedPodAffinityTerm(100, "rook-ceph-mds"),
+					getMdsWeightedPodAffinityTerm(100, "ocs-storagecluster-cephfilesystem"),
```
travisn, line 79: We need to pass the name of the actual filesystem instead of the hard coded value, right?

travisn (changes requested):

controllers/defaults/placements.go
parth-gr: this one is only for the default case

parth-gr: UI/CLI needs to create a CR with there filesystem value

travisn: IIRC the name of the cephfilesystem CR is: `<storageCluster>-cephfilesystem`, and the name of the `storageCluster` could be different than the default if the CR is created directly by the user instead of the UI. If that's the case, we can't hard code the filesystem name here.

parth-gr: correct updated

controllers/storagecluster/cephfilesystem.go
```diff
@@ -27,6 +27,7 @@ func (r *StorageClusterReconciler) newCephFilesystemInstances(initStorageCluster
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      generateNameForCephFilesystem(initStorageCluster),
 			Namespace: initStorageCluster.Namespace,
+			Labels:    cephv1.Labels{"rook_file_system": generateNameForCephFilesystem(initStorageCluster)},
```
travisn, line 30: Why add this label to the filesystem CR? It's the same as the name of the CR, so it seems unnecessary.

controllers/storagecluster/placement.go
```diff
@@ -22,6 +22,11 @@ func getPlacement(sc *ocsv1.StorageCluster, component string) rookCephv1.Placeme
 		(&in).DeepCopyInto(&placement)
 	}
 
+	// default value for mds pod anti affinity is set to the filesystem name
+	if component == "mds" && placement.PodAffinity != nil {
```
iamniting, line 26: What if `placement.PodAffinity` is nil, The value will always be `ocs-storagecluster-cephfilesystem` . Is that okay?

iamniting (changes requested):

controllers/storagecluster/placement.go
parth-gr: I think how ocs operator appends it wont be ever null.
I have added this check because of the test cases we have in the test file.

controllers/storagecluster/cephfilesystem.go
parth-gr: The podAffinity looks for this label of the pod, that why we need it

travisn: The pod affinity looks on the pod labels, not the cephcluster CR labels

controllers/storagecluster/placement.go
iamniting: I will be nil. The default set `PodAntiAffinity` not `PodAffinity` so it will be nil if we are using defaults. Moreover, in the case of customer-specified tolerations, we never know.

parth-gr: actually I need to check `PodAntiAffinity `

iamniting: Now the default case is solved. 

What about the customer given tolerations? Do we need a value if the tolerations are given by the customer and `PodAntiAffinity` is not set? What if customer sets more than one item in the list?

parth-gr: I think it will work like the default case, and later if there is a user input we have a later checks for that in the same function,

I will look again into it if nothing breaks

parth-gr: This function will take care of all the caseshttps://github.com/red-hat-storage/ocs-operator/blob/main/controllers/storagecluster/placement.go#L115
And we should always add the deafult values as per design, as in what ways we want to restrict,
If we wanna override this function should be changed,
Which been called after I have set the deafult values

controllers/defaults/placements.go
```diff
@@ -176,6 +182,25 @@ func getWeightedPodAffinityTerm(weight int32, selectorValue ...string) corev1.We
...
+func getMdsWeightedPodAffinityTerm(weight int32, selectorValue ...string) corev1.WeightedPodAffinityTerm {
+	WeightedPodAffinityTerm := corev1.WeightedPodAffinityTerm{
+		Weight: weight,
+		PodAffinityTerm: corev1.PodAffinityTerm{
+			LabelSelector: &metav1.LabelSelector{
+				MatchExpressions: []metav1.LabelSelectorRequirement{
+					{
+						Key:      "rook_file_system",
+						Operator: metav1.LabelSelectorOpIn,
+						Values:   selectorValue,
```
iamniting, line 194: One comment here as well pls.

controllers/defaults/placements.go
```diff
@@ -76,7 +79,10 @@ var (
 			},
 			PodAntiAffinity: &corev1.PodAntiAffinity{
 				PreferredDuringSchedulingIgnoredDuringExecution: []corev1.WeightedPodAffinityTerm{
-					getWeightedPodAffinityTerm(100, "rook-ceph-mds"),
+					// the selector `MdsNodeAffinitySelector` value will be changed later in the getPlacement(),
+					// to the actual value of the filesystem,
+					// do not append the list of values here as later it will just replace the first value
```
iamniting, line 84: We can append it, Pls mention we are using the 0 index to change the value. So 0 index should remain 0.

controllers/storagecluster/placement.go
```diff
@@ -20,6 +20,12 @@ func getPlacement(sc *ocsv1.StorageCluster, component string) rookCephv1.Placeme
 	} else {
 		in := defaults.DaemonPlacements[component]
 		(&in).DeepCopyInto(&placement)
+		// default value for mds pod anti affinity is set to the filesystem name
+		// label rook_file_system is added to the mds pod using rook operator
+		// should be set before appendNodeRequirements() so to preserve the default values
```
iamniting, line 25:
```suggestion
```

No need of this line as in the `appendNodeRequirements` we don't do anything with the `PodAntiAffinity`

iamniting (changes requested):

controllers/defaults/placements.go
```diff
@@ -176,6 +182,26 @@ func getWeightedPodAffinityTerm(weight int32, selectorValue ...string) corev1.We
...
+func getMdsWeightedPodAffinityTerm(weight int32, selectorValue ...string) corev1.WeightedPodAffinityTerm {
+	WeightedPodAffinityTerm := corev1.WeightedPodAffinityTerm{
+		Weight: weight,
+		PodAffinityTerm: corev1.PodAffinityTerm{
+			LabelSelector: &metav1.LabelSelector{
+				MatchExpressions: []metav1.LabelSelectorRequirement{
+					{
+						Key:      "rook_file_system",
+						Operator: metav1.LabelSelectorOpIn,
+						// the selector `MdsNodeAffinitySelector` value will be changed later in the getPlacement(),
```
iamniting, line 194:
```suggestion
                                       // the selector value will be changed later in the getPlacement(),
					// to the actual value of the filesystem,
					// the zero index is used to set the value of the selector, which will be changed later in the getPlacement()```

**Review by iamniting — CHANGES_REQUESTED** [review]

File: controllers/defaults/placements.go [hunk 9]
```diff
@@ -176,6 +182,26 @@ func getWeightedPodAffinityTerm(weight int32, selectorValue ...string) corev1.We
 	return WeightedPodAffinityTerm
 }
 
+func getMdsWeightedPodAffinityTerm(weight int32, selectorValue ...string) corev1.WeightedPodAffinityTerm {
+	WeightedPodAffinityTerm := corev1.WeightedPodAffinityTerm{
```
travisn, line 186: can return this directly, no need for a var
```suggestion
	return corev1.WeightedPodAffinityTerm{
```

controllers/storagecluster/placement.go
```diff
@@ -20,6 +20,11 @@ func getPlacement(sc *ocsv1.StorageCluster, component string) rookCephv1.Placeme
 	} else {
 		in := defaults.DaemonPlacements[component]
 		(&in).DeepCopyInto(&placement)
+		// default value for mds pod anti affinity is set to the filesystem name
+		// label rook_file_system is added to the mds pod using rook operator
+		if component == "mds" && placement.PodAntiAffinity != nil {
```
travisn, line 25: Can we just directly assign the antiaffinity here? It seems so awkward to add a placeholder that will be updated later. This way we just directly generate it when needed. 
```suggestion
		if component == "mds" {
			placement.PodAntiAffinity = getMdsWeightedPodAffinityTerm(100, generateNameForCephFilesystem(sc))
		}
```

parth-gr: done

controllers/defaults/placements.go
parth-gr: Now we dont need it

iamniting (approved):
looks good to me
/hold for travis review

travisn (approved):
/lgtm

PR #2766, opened 2024-08-29
use qualified name upto ".svc" for endpoint svc in native storageclient

leelavg:
cluster proxies are configured not to proxy in cluster traffic and since we were using only the service name the connection is being proxied, using a qualified name upto ".svc" or ".svc.cluster.local" which is already configured under `.spec.noProxy` in `proxy/cluster` resource doesn't proxy the connection.


leelavg:
/cherry-pick release-4.16
https://bugzilla.redhat.com/show_bug.cgi?id=2308445

PR #2773, opened 2024-09-02
update deps bundle images

iamniting:


Nikhil-Ladha (approved):
/lgtm

hack/common.sh
```diff
@@ -50,8 +50,8 @@ OCS_CSV="$OUTDIR_TEMPLATES/ocs-operator.csv.yaml.in"
 # and we should continue to use dowsntream image only due to few
 # downstream only changes present in rook downstream fork.
 LATEST_ROOK_IMAGE="quay.io/ocs-dev/rook-ceph:vrelease-4.17-91cc5780b"
-LATEST_NOOBAA_CORE_IMAGE="quay.io/noobaa/noobaa-core:master-20240820"
-LATEST_NOOBAA_DB_IMAGE="docker.io/centos/postgresql-12-centos8"
+LATEST_NOOBAA_CORE_IMAGE="quay.io/noobaa/noobaa-core:master-20240901"
+LATEST_NOOBAA_DB_IMAGE="quay.io/sclorg/postgresql-15-c9s"
```
malayparida2000, line 54: Just adding this link here for future reference
https://github.com/noobaa/noobaa-operator/blob/master/pkg/options/options.go#L73-L75

malayparida2000 (approved):

malayparida2000 (approved):
/lgtm

hack/install-ocs-operator.sh
```diff
@@ -33,28 +33,39 @@ data:
...
-  name: ocs-client-operator-config
-  namespace: openshift-storage
-data:
-  DEPLOY_CSI: "false"
-EOF
+patch_ocs_client_operator_config_configmap() {
+    while true; do
+        if oc get cm ocs-client-operator-config -n openshift-storage; then
+            oc patch cm ocs-client-operator-config -n openshift-storage --type merge -p '{"data":{"DEPLOY_CSI":"false"}}'
+            sleep 10
```
leelavg, line 40: is any external entity resetting it? if yes, it can do so after sleep as well, if the api returned successfully after a patch then we could directly test the value isn't it?

hack/install-ocs-operator.sh
```diff
@@ -33,28 +33,39 @@ data:
...
+            value=$(oc get cm ocs-client-operator-config -n openshift-storage -o custom-columns=:data.DEPLOY_CSI --no-headers)
```
leelavg, line 41: any reason in not using json? `oc get cm ocs-client-operator-config -nopenshift-storage -ojsonpath={.data.DEPLOY_CSI}`?

Nikhil-Ladha (approved):
/lgtm

Nikhil-Ladha:
/hold
for above reviews to be addressed

hack/install-ocs-operator.sh
```diff
@@ -33,28 +33,39 @@ data:
...
+            if [[ "$value" == "false" ]]; then
+                break
+            fi
+        fi
+        sleep 10
```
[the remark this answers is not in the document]
iamniting: It was not working after patching once. One sleep is for waiting after patch and one sleep is for waiting for CM availability.

hack/install-ocs-operator.sh
iamniting: Both do the same work.

hack/install-ocs-operator.sh
iamniting: If we directly check the value without sleep, we may get the value that we set, and the external entity has not reset it yet. I added this after testing only.

leelavg: May I know what made you pick `10s`? I don't think sleep is a good protection against external entity and is csv-controller the one changing fields? Did you check managed-fields on cm before coming up w/ sleep?

I'm just being extra cautious at the expense of delaying the merge as this is a high frequency script, pls excuse.

hack/install-ocs-operator.sh
leelavg: I understand the outcome will be same , however a li'l surprised w/ the choice and interested to know the reasoning.

Basically the current directive is creating single unnamed column w/ single row and again explicitly cutting the column name rather than directly picking the value of interest.

But fine if this is just a matter of preference.

hack/install-ocs-operator.sh
iamniting: I believe it is CSV controller but I am not sure. I used 10 second just like that and tested it multiple times and it worked.

hack/install-ocs-operator.sh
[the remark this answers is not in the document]
leelavg: > one sleep is for waiting for CM availability.
- makes sense, pls resolve.

hack/install-ocs-operator.sh
leelavg: > multiple times and it worked.
- ack, basing my vote only on this.

leelavg:
LGTM

PR #2776, opened 2024-09-02
disabling rook mgr module in CephCluster

ShravaniVangur:
Sometimes, during installation with ODF, the rook mgr module in ceph cluster is being enabled. This pr ensures that it is disabled during installation. 

For example:
`sh-4.4$ ceph mgr module ls | grep rook`
`rook                  on`     

With this change:
`sh-4.4$ ceph mgr module ls | grep rook`
`rook                  -`

### DIFF
--- a/controllers/storagecluster/cephcluster.go
+++ b/controllers/storagecluster/cephcluster.go
@@ -1101,6 +1101,7 @@ func generateMgrSpec(sc *ocsv1.StorageCluster) rookCephv1.MgrSpec {
 		Modules: []rookCephv1.Module{
 			{Name: "pg_autoscaler", Enabled: true},
 			{Name: "balancer", Enabled: true},
+			{Name: "rook", Enabled: false},
 		},
 	}
 
--- a/controllers/storagecluster/cephcluster_test.go
+++ b/controllers/storagecluster/cephcluster_test.go
@@ -306,6 +306,7 @@ func TestGenerateMgrSpec(t *testing.T) {
 				Modules: []rookCephv1.Module{
 					{Name: "pg_autoscaler", Enabled: true},
 					{Name: "balancer", Enabled: true},
+					{Name: "rook", Enabled: false},
 				},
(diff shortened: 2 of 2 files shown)


controllers/storagecluster/cephcluster.go
```diff
@@ -1101,6 +1101,7 @@ func generateMgrSpec(sc *ocsv1.StorageCluster) rookCephv1.MgrSpec {
 		Modules: []rookCephv1.Module{
 			{Name: "pg_autoscaler", Enabled: true},
 			{Name: "balancer", Enabled: true},
+			{Name: "rook", Enabled: false},
```
Nikhil-Ladha, line 1104: Are we planning to keep it disabled permanently? I doubt 🤔

ShravaniVangur: Technically it is supposed to be disabled when it is installed. But due to some issues as observed in this bz: https://bugzilla.redhat.com/show_bug.cgi?id=2300312#add_comment the rook mgr module is enabled initially. This just makes sure that on installation it remains disabled. Can be enabled through the rook toolbox post installation.

> Are we planning to keep it disabled permanently? I doubt 🤔

Nikhil-Ladha: Understood, but what if in future we would want to enable the module for some case? By disabling it this way, there would be clash between a user manually enabling it through toolbox and the operator reconciling it to disable.
@travisn is it guaranteed that in future as well we don't see any use case for rook-mgr module in ODF d/s?

travisn: The `rook` mgr module is really only used by the upstream dashboard, which we don't use downstream. There is not currently a scenario to enable this module in the downstream. When it is enabled (though we are not sure the cause), it causes errors as we have seen in a couple customer clusters.

travisn (commented):
/lgtm

PR #2715, opened 2024-07-29
Read secrets for onboarding-token validation

Mrudraia:
This PR reads the secrets instead of reading the secrets from the volume mounts. 
whenever the new onboarding secrets are created, it takes more time to read the secrets from the volume mounts,
The user clicks the rotate onboarding  keys, the kubernetes still uses the old  public, private keys , the new keys are mounted later,  So this PR will read the secrets directly from the kubernetes secrets.

PR #2569, opened 2024-04-19
remove rotation of keys endpoint from ux-backend

leelavg:
existing mechanism is only working as a proxy for deletion of secret w/ no added value, so deleting this endpoint and UI will be wired to directly perform the operator.

ref: #2469

PR #2790, opened 2024-09-10
Read secrets for onboarding-token validation

mrudraia1:
This PR is a copy PR of https://github.com/red-hat-storage/ocs-operator/pull/2715

I faced git account issues, I was not able to recover my account, so created a new PR with new account

This PR reads the secrets instead of reading the secrets from the volume mounts.
whenever the new onboarding secrets are created, it takes more time to read the secrets from the volume mounts,
The user clicks the rotate onboarding keys, the kubernetes still uses the old public, private keys , the new keys are mounted later, So this PR will read the secrets directly from the kubernetes secrets.

### DIFF
--- a/controllers/storagecluster/storageclient.go
+++ b/controllers/storagecluster/storageclient.go
@@ -6,6 +6,7 @@ import (
 	ocsclientv1a1 "github.com/red-hat-storage/ocs-client-operator/api/v1alpha1"
 	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
 	"github.com/red-hat-storage/ocs-operator/v4/controllers/util"
+
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
@@ -31,7 +32,7 @@ func (s *storageClient) ensureCreated(r *StorageClusterReconciler, storagecluste
 	storageClient.Name = storagecluster.Name
 	_, err := controllerutil.CreateOrUpdate(r.ctx, r.Client, storageClient, func() error {
 		if storageClient.Status.ConsumerID == "" {
-			token, err := util.GenerateOnboardingToken(tokenLifetimeInHours, onboardingPrivateKeyFilePath, nil)
+			token, err := util.GenerateClientOnboardingToken(tokenLifetimeInHours, onboardingPrivateKeyFilePath, nil)
 			if err != nil {
 				return fmt.Errorf("unable to generate onboarding token: %v", err)
 			}
--- a/controllers/storagecluster/storageclient.go
+++ b/controllers/storagecluster/storageclient.go
@@ -2,12 +2,16 @@ package storagecluster
 
 import (
 	"fmt"
+	"maps"
+	"strconv"
 
 	ocsclientv1a1 "github.com/red-hat-storage/ocs-client-operator/api/v1alpha1"
 	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
 	"github.com/red-hat-storage/ocs-operator/v4/controllers/util"
 
+	corev1 "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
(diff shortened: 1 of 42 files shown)


services/ux-backend/main.go
```diff
@@ -7,12 +7,17 @@ import (
...
 	"github.com/red-hat-storage/ocs-operator/v4/services/ux-backend/handlers/onboardingtokens"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
 )
 
 type serverConfig struct {
+	client.Client
```
leelavg, line 20: where is this being used?

services/ux-backend/main.go
```diff
@@ -7,12 +7,17 @@ import (
 	"os"
 	"strconv"
 
-	"k8s.io/klog/v2"
-
+	v1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
```
leelavg, line 10: pls change to `ocsv1`, we generally use `v1` for native apis, even that I'm not seeing lately.

controllers/util/util.go
```diff
@@ -5,6 +5,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const onboardingValidationPrivateKeySecretName = "onboarding-private-key"
```
leelavg, line 8: could you pls move this const to the file where it is being used?

controllers/util/onboarding_secrets.go
```diff
@@ -0,0 +1,41 @@
...
+	"crypto/x509"
+	"encoding/pem"
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func ReadPrivateKey(cl client.Client) (*rsa.PrivateKey, error) {
```
leelavg, line 15: why can't this function be in existing `provider.go` itself?

leelavg, line 15: we aren't reading the key, we are getting it from API and parsing it as well, name `GetParsedPrivateKey` is more apt.

controllers/util/onboarding_secrets.go
```diff
@@ -0,0 +1,41 @@
...
+	klog.Info("Getting the Pem key")
+	ctx := context.Background()
+
+	operatorNamespace, err := GetOperatorNamespace()
```
leelavg, line 19: get the namespace as param rather than reading it everytime, I see this is being used in two places where you can already have ns.
1. in reconciler, already has it
2. in handler, pls initialize ns once and read from it rather than reading from env everytime, ref #2569

leelavg (commented):
much better, can merge it at the earliest.

controllers/storagecluster/storageclient.go
```diff
@@ -27,11 +24,17 @@ func (s *storageClient) ensureCreated(r *StorageClusterReconciler, storagecluste
 		return s.ensureDeleted(r, storagecluster)
 	}
 
+	privateKey, err := util.ReadPrivateKey(r.Client)
+	if err != nil {
+		r.Log.Error(err, "Unable to get privatekey")
+		return reconcile.Result{}, nil
```
rchikatw, line 30: you might want to return an error here.

controllers/util/onboarding_secrets.go
mrudraia1: In Provider.go, their will be import issues for package
"crypto", "crypto/rand", "crypto/rsa". ReadPrivateKey function require "crypto/rsa", "crypto/x509". So for import conflicts, I created an new file.

controllers/util/util.go
mrudraia1: ack

services/ux-backend/main.go
mrudraia1: ack

services/ux-backend/main.go
mrudraia1: removed, it is no were called

controllers/util/onboarding_secrets.go
mrudraia1: ok

controllers/storagecluster/storageclient.go
mrudraia1: ack

controllers/util/onboarding_secrets.go
mrudraia1: ack

controllers/util/onboarding_secrets.go
mrudraia1: Ack, deleted the onboarding_secrets.go and function called in provider.go

leelavg:
We add `Bug <num>:[branch]` to PR summary when we are back porting to that branch but not when targeting `main`

mrudraia1:
reopened new PR 2827. closing this PR due to some conflicting files and commits

PR #2791, opened 2024-09-11
extend the onboarding struct to make it modular

rewantsoni:
Extend the onboarding struct have clients field nested instead of on the root. This will provide us the option to have onboarding ticket for different roles.

### DIFF
--- a/services/provider/server/server.go
+++ b/services/provider/server/server.go
@@ -110,13 +110,13 @@ func (s *OCSProviderServer) OnboardConsumer(ctx context.Context, req *pb.Onboard
 		return nil, status.Errorf(codes.Internal, "failed to get public key to validate onboarding ticket for consumer %q. %v", req.ConsumerName, err)
 	}
 
-	onboardingTicket, err := decodeAndValidateTicket(req.OnboardingTicket, pubKey)
+	onboardingTicket, err := decodeAndValidateTicket(req.OnboardingTicket, pubKey, services.ClientRole)
 	if err != nil {
 		klog.Errorf("failed to validate onboarding ticket for consumer %q. %v", req.ConsumerName, err)
 		return nil, status.Errorf(codes.InvalidArgument, "onboarding ticket is not valid. %v", err)
 	}
 
-	storageConsumerUUID, err := s.consumerManager.Create(ctx, req, int(onboardingTicket.StorageQuotaInGiB))
+	storageConsumerUUID, err := s.consumerManager.Create(ctx, req, int(onboardingTicket.SubjectRole.ClientOptions.StorageQuotaInGiB))
 	if err != nil {
 		if !kerrors.IsAlreadyExists(err) && err != errTicketAlreadyExists {
 			return nil, status.Errorf(codes.Internal, "failed to create storageConsumer %q. %v", req.ConsumerName, err)
@@ -470,7 +470,7 @@ func getSubVolumeGroupClusterID(subVolumeGroup *rookCephv1.CephFilesystemSubVolu
 	return hex.EncodeToString(hash[:16])
 }
 
-func decodeAndValidateTicket(ticket string, pubKey *rsa.PublicKey) (*services.OnboardingTicket, error) {
+func decodeAndValidateTicket(ticket string, pubKey *rsa.PublicKey, expectedRole services.OnboardingSubjectRole) (*services.OnboardingTicket, error) {
 	ticketArr := strings.Split(string(ticket), ".")
 	if len(ticketArr) != 2 {
 		return nil, fmt.Errorf("invalid ticket")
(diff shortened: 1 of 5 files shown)


rewantsoni:
ci failure is due to noobaa stuck in connecting phase
```
[rewantsoni tmp] k get noobaa       
NAME     S3-ENDPOINTS   STS-ENDPOINTS   SYSLOG-ENDPOINTS   IMAGE                                        PHASE        AGE
noobaa                                                     quay.io/noobaa/noobaa-core:master-20240901   Connecting   2m41s
```

services/ux-backend/handlers/onboardingtokens/handler.go
```diff
@@ -32,9 +33,9 @@ func HandleMessage(w http.ResponseWriter, r *http.Request, tokenLifetimeInHours
 }
 
 func handlePost(w http.ResponseWriter, r *http.Request, tokenLifetimeInHours int) {
-	var storageQuotaInGiB *uint
-	// When ContentLength is 0 that means request body is empty and
-	// storage quota is unlimited
+
+	var roleSpec services.OnboardingSubjectRoleSpec
+	roleSpec.Role = services.ClientRole
```
leelavg, line 38: setting fields at multiple places forces us to look for where all these are set, better to move this to line 59 only where you are setting the options?

services/ux-backend/handlers/onboardingtokens/handler.go
```diff
@@ -55,9 +56,9 @@ func handlePost(w http.ResponseWriter, r *http.Request, tokenLifetimeInHours int
 			http.Error(w, fmt.Sprintf("invalid Unit type sent in request body, Valid types are [Gi,Ti,Pi]: %v", quota.Unit), http.StatusBadRequest)
 			return
 		}
-		storageQuotaInGiB = ptr.To(unitAsGiB * quota.Value)
+		roleSpec.ClientOptions = services.ClientRoleOptions{StorageQuotaInGiB: unitAsGiB * quota.Value}
 	}
-	if onboardingToken, err := util.GenerateOnboardingToken(tokenLifetimeInHours, onboardingPrivateKeyFilePath, storageQuotaInGiB); err != nil {
+	if onboardingToken, err := util.GenerateOnboardingToken(tokenLifetimeInHours, onboardingPrivateKeyFilePath, roleSpec); err != nil {
```
leelavg, line 61: if `contentLength` is 0 then you'll be setting empty roleSpec (after doing above change), is that still fine?

services/types.go
```diff
@@ -1,7 +1,24 @@
 package services
 
+type OnboardingSubjectRole string
+
+type OnboardingSubjectRoleSpec struct {
+	Role          OnboardingSubjectRole `json:"role"`
```
leelavg, line 6: you are trying to make `role` as mandatory but the original could be empty quota, thought of having `OnboardingSubjectRoleSpec` as a pointer in parent struct?

services/provider/server/server.go
```diff
@@ -487,8 +487,18 @@ func decodeAndValidateTicket(ticket string, pubKey *rsa.PublicKey) (*services.On
 		return nil, fmt.Errorf("failed to unmarshal onboarding ticket message. %v", err)
 	}
 
-	if ticketData.StorageQuotaInGiB > math.MaxInt {
-		return nil, fmt.Errorf("invalid value sent in onboarding ticket, storage quota should be greater than 0 and less than %v: %v", math.MaxInt, ticketData.StorageQuotaInGiB)
+	if ticketData.SubjectRole.Role != expectedRole {
+		return nil, fmt.Errorf("unsupported role sent in onboarding ticket %s, expectedRole is %s", ticketData.SubjectRole.Role, expectedRole)
+	}
```
leelavg, line 492: isn't this repeating the default case of below switch?

leelavg (commented):
to me the proposed API seems more opinionated than required, my assumptions are
1. quota will always be >0 if set
2. not all tickets will have quota
3. tickets can be of two explicit types: no role, client and peer role

this PR is trying to club `no role` and `client role`, is that the expectation? this resulted in mandating a role for local client which could be no different than supplying an explicit nil.

I propose for each of roles to be distinct even though being verbose it doesn't force quota in normal case and accept pointer in func signature for backward compat? Think of some automation outside of odf generating tokens and doesn't want newer futures and current API forces to set client role.

wdyt? pls do ping other reviewers before acting on the review if required. open to be corrected as always.

services/ux-backend/handlers/onboardingtokens/handler.go
rewantsoni: Since we have only one role up to this point, I am always setting the role as clientRole to avoid further changes to console. With 4.18, we will explicitly set the role while calling this endpoint

services/ux-backend/handlers/onboardingtokens/handler.go
rewantsoni: If that is done, the ticket will be generated for an empty role if the body is empty, which will always lead to failure while onboarding since we cannot verify the role

services/types.go
rewantsoni: `original could be empty quota` Could you elaborate on what you mean by original could be empty quota?

services/provider/server/server.go
rewantsoni: Yes, I am expecting this function to be called while onboarding (onboarding clients and onboarding peers). This check is required so that we don't enter the peer role ticket while onboarding the client and vice-versa.

I could remove the default case in the switch as the code would never reach it.

rewantsoni:
> to me the proposed API seems more opinionated than required, my assumptions are
> 
> 1. quota will always be >0 if set
> 2. not all tickets will have quota
> 3. tickets can be of two explicit types: no role, client and peer role
> 
> this PR is trying to club `no role` and `client role`, is that the expectation? this resulted in mandating a role for local client which could be no different than supplying an explicit nil.
> 
> I propose for each of roles to be distinct even though being verbose it doesn't force quota in normal case and accept pointer in func signature for backward compat? Think of some automation outside of odf generating tokens and doesn't want newer futures and current API forces to set client role (I understand you are setting the default but that has a different comment).
> 
> wdyt? pls do ping other reviewers before acting on the review if required. open to be corrected as always.

In the Onboarding struct
1. storageQuotaInGiB can be >=0, =0 in case of unlimited quota
2. Right, we don't expect all tickets to have a quota
3. the ticket will always have a role, either ocs-client or ocs-peer

I think If we have an external automation generating tokens, we should have them specify the role of the ticket they want to generate (for client or peer) just like they specify the quota they need. Or, for backward compatibility of such automation, we can always set the default role to ocs-client.

leelavg:
> the ticket will always have a role, either ocs-client or ocs-peer
- Ack, if that is the requirement then the PR lgtm.

rewantsoni:
closing it as not required

PR #2792, opened 2024-09-11
Bug 2311468: [release-4.17] Re-adds DR Optimized check feature

raaizik:
## Changes

- This change re-adds the isDROptimized check feature as it's now required in 4.17 which will be using Ceph 7.
- Skipping the tests for this never before tested feature with a flag for now. Will be tested in follow up PR. 

Fix for [BZ-2311468](https://bugzilla.redhat.com/show_bug.cgi?id=2311468).


malayparida2000 (changes requested):
Please confirm if this is an issue isolated to only 4.17 & not affecting the main. If not open the fix for the main branch first & after it's merged then only create the backport.

sp98:
> Please confirm if this is an issue isolated to only 4.17 & not affecting the main. If not open the fix for the main branch first & after it's merged then only create the backport.Pleas

@malayparida2000 We only plan to fix this in 4.17 since 4.17 is now targeting Ceph 7 instead of ceph 8.

umangachapagain (approved):

malayparida2000 (approved):

PR #2793, opened 2024-09-11
Use msgrv2 always for all ODF modes(internal,external,provider)

malayparida2000:
Initially in 4.13 when encryption in-transit was implemented we started
to use msgrv2 for internal mode only. External mode & Provider mode were
excluded.

For external mode the concern was the availability of msgrv2
port always on the RHCS cluster. But now we have reached ceph v18 &
msgrv2 is being by default on since ceph Nautilus 14.2. So we
can now use mgsr2 for external mode.

For provider mode the concern was availability of the required kernel
versions on old client clusters. But now that concern also seems no
longer valid. So we can now use msgrv2 for provider mode as well.

Also as we would always use msgrv2, we can now simplify the
GetCephFSKernelMountOptions function to remove the conditional checks.
If encryption is enabled use "secure" otherwise "prefer-crc".

*This change can still break external mode/provider mode CI. But as we are just getting started with 4.18 this seems like a safe time to test this out & fix any follow-up issues if they come.*

### DIFF
--- a/controllers/storagecluster/cephcluster.go
+++ b/controllers/storagecluster/cephcluster.go
@@ -605,13 +605,11 @@ func getNetworkSpec(sc ocsv1.StorageCluster) rookCephv1.NetworkSpec {
 	// respect both the old way and the new way for enabling HostNetwork
 	networkSpec.HostNetwork = networkSpec.HostNetwork || sc.Spec.HostNetwork
 
-	// If it's not an external and not a provider cluster always require msgr2
-	if !sc.Spec.AllowRemoteStorageConsumers && !sc.Spec.ExternalStorage.Enable {
-		if networkSpec.Connections == nil {
-			networkSpec.Connections = &rookCephv1.ConnectionsSpec{}
-		}
-		networkSpec.Connections.RequireMsgr2 = true
+	if networkSpec.Connections == nil {
+		networkSpec.Connections = &rookCephv1.ConnectionsSpec{}
 	}
+	// Always require msgr2 for all modes of ODF
+	networkSpec.Connections.RequireMsgr2 = true
 
 	return networkSpec
 }
--- a/controllers/storagecluster/cephcluster_test.go
+++ b/controllers/storagecluster/cephcluster_test.go
@@ -1290,7 +1290,9 @@ func TestCephClusterNetworkConnectionsSpec(t *testing.T) {
 			},
 			ccSpec: rookCephv1.ClusterSpec{
 				Network: rookCephv1.NetworkSpec{
-					Connections: &rookCephv1.ConnectionsSpec{},
+					Connections: &rookCephv1.ConnectionsSpec{
+						RequireMsgr2: true,
+					},
 				},
 			},
 		},
@@ -1308,6 +1310,7 @@ func TestCephClusterNetworkConnectionsSpec(t *testing.T) {
 			ccSpec: rookCephv1.ClusterSpec{
 				Network: rookCephv1.NetworkSpec{
 					Connections: &rookCephv1.ConnectionsSpec{
+						RequireMsgr2: true,
 						Encryption: &rookCephv1.EncryptionSpec{
 							Enabled: true,
 						},
@@ -1329,6 +1332,7 @@ func TestCephClusterNetworkConnectionsSpec(t *testing.T) {
 			ccSpec: rookCephv1.ClusterSpec{
 				Network: rookCephv1.NetworkSpec{
 					Connections: &rookCephv1.ConnectionsSpec{
+						RequireMsgr2: true,
 						Compression: &rookCephv1.CompressionSpec{
 							Enabled: true,
 						},
@@ -1353,6 +1357,7 @@ func TestCephClusterNetworkConnectionsSpec(t *testing.T) {
 			ccSpec: rookCephv1.ClusterSpec{
 				Network: rookCephv1.NetworkSpec{
 					Connections: &rookCephv1.ConnectionsSpec{
+						RequireMsgr2: true,
 						Encryption: &rookCephv1.EncryptionSpec{
 							Enabled: true,
 						},
@@ -1380,6 +1385,7 @@ func TestCephClusterNetworkConnectionsSpec(t *testing.T) {
 			ccSpec: rookCephv1.ClusterSpec{
 				Network: rookCephv1.NetworkSpec{
 					Connections: &rookCephv1.ConnectionsSpec{
+						RequireMsgr2: true,
 						Encryption: &rookCephv1.EncryptionSpec{
 							Enabled: false,
 						},
--- a/controllers/util/csi_options.go
+++ b/controllers/util/csi_options.go
@@ -13,22 +13,7 @@ func GetCephFSKernelMountOptions(sc *ocsv1.StorageCluster) string {
 		return "ms_mode=secure"
 	}
 
-	// If Encryption is not enabled, but Compression or RequireMsgr2 is enabled, use prefer-crc mode
-	if sc.Spec.Network != nil && sc.Spec.Network.Connections != nil &&
-		((sc.Spec.Network.Connections.Compression != nil && sc.Spec.Network.Connections.Compression.Enabled) ||
-			sc.Spec.Network.Connections.RequireMsgr2) {
-		return "ms_mode=prefer-crc"
-	}
-
-	// Network spec always has higher precedence even in the External or Provider cluster. so they are checked first above
-
-	// None of Encryption, Compression, RequireMsgr2 are enabled on the StorageCluster
-	// If it's an External or Provider cluster, We don't require msgr2 by default so no mount options are needed
-	if sc.Spec.ExternalStorage.Enable || sc.Spec.AllowRemoteStorageConsumers {
-		return "ms_mode=legacy"
-	}
-	// If none of the above cases apply, We set RequireMsgr2 true by default on the cephcluster
-	// so we need to set the mount options to prefer-crc
+	// If encryption is not enabled, use prefer-crc mode
 	return "ms_mode=prefer-crc"
 }
 
--- a/services/provider/server/server_test.go
+++ b/services/provider/server/server_test.go
@@ -769,7 +769,7 @@ func TestOCSProviderServerGetStorageClaimConfig(t *testing.T) {
 					CephFs: &csiopv1a1.CephFsConfigSpec{
 						SubVolumeGroup: "cephFilesystemSubVolumeGroup",
 						KernelMountOptions: map[string]string{
-							"ms_mode": "legacy",
+							"ms_mode": "prefer-crc",
 						},
 					},
 				},


travisn (commented):
LGTM, assuming we can confirm that RHCS will always have the msgrv2 port available.

travisn:
@parth-gr Can you also confirm if the external cluster script collects the v2 endpoint (port 3300) when detecting the mons? Commonly I would expect both v1 and v2 to be available, so we just want to start relying exclusively on the v2 port.

leelavg (commented):
Does the existing PVs/VSCs continue to work if ms_mode is changed to prefer_crc from legacy, before 4.17 we (provider) weren't using this option and so it was default.

LGTM and I'd like a quick opinion from @Madhu-1, thanks.

malayparida2000:
> Does the existing PVs/VSCs continue to work if ms_mode is changed to prefer_crc from legacy, before 4.17 we (provider) weren't using this option and so it was default.
> 
> LGTM and I'd like a quick opinion from @Madhu-1, thanks.

Yes, existing volumes would continue to use the legacy v1 port without any issues. The mount option is utilised only during volume mounting process in the beginning. So any new volumes that would be mounted after this change will use the prefer-crc mount option.

controllers/storagecluster/cephcluster.go
```diff
@@ -605,13 +605,11 @@ func getNetworkSpec(sc ocsv1.StorageCluster) rookCephv1.NetworkSpec {
...
-	if !sc.Spec.AllowRemoteStorageConsumers && !sc.Spec.ExternalStorage.Enable {
-		if networkSpec.Connections == nil {
-			networkSpec.Connections = &rookCephv1.ConnectionsSpec{}
-		}
-		networkSpec.Connections.RequireMsgr2 = true
+	if networkSpec.Connections == nil {
+		networkSpec.Connections = &rookCephv1.ConnectionsSpec{}
 	}
+	// Always require msgr2 for all modes of ODF
+	networkSpec.Connections.RequireMsgr2 = true
```
parth-gr, line 612: v2 for external mode is controlled by 
`--v2-port-enable` flag, we should not explicitly set here,
Or if there is a need to always make it true we should first do at python script.


Current flow is, 
1) User select v2 port in the python script
2) Script extracts v2 mon ports
3) Internally in rook if  there is v2 detected it will turn cluster:connection to true 
https://github.com/rook/rook/blob/58a0c834bf5bce6782de5c387f3dac732393ef33/pkg/operator/ceph/cluster/cluster_external.go#L126

parth-gr (changes requested):

parth-gr:
> Yes, existing volumes would continue to use the legacy v1 port without any issues. The mount option is utilised only during volume mounting process in the beginning.

Does that means v2 port can still reconcile/watch/read the PVC that is created by v1 port??

Madhu-1:
@malayparida2000 are we sure that RHCS doesn't support running only on v1 ports? if its supported we also need to support the same, if not we are good, can we get this confirmation?

controllers/storagecluster/cephcluster.go
malayparida2000: @parth-gr is there any reason why v2 mon ports are not extracted by default if they are available?
Also please look at this BZ here https://bugzilla.redhat.com/show_bug.cgi?id=2180921#c2, This cam for external mode when we tried to use msgrv2 at that point.

malayparida2000:
> @malayparida2000 are we sure that RHCS doesn't support running only on v1 ports? if its supported we also need to support the same, if not we are good, can we get this confirmation?

By support, I have to check with the support team may be. But yes on a RHCS cluster msgrv2 port can be disabled if wanted by the customer, so technically a RHCS cluster can be of v1 port only.

malayparida2000:
> > Yes, existing volumes would continue to use the legacy v1 port without any issues. The mount option is utilised only during volume mounting process in the beginning.
> 
> Does that means v2 port can still mount/remount the PVC that is created by v1 port??

Yes if remounting happens then the volume will now mount using the v2 port

controllers/storagecluster/cephcluster.go
malayparida2000: I don't think we can do that in rook upstream. Is there any hard requirement on the rook version & the Ceph provider clusters's ceph versions? If not upstream user might be running some old version of ceph release. 
@travisn wdyt?

Madhu-1:
> > @malayparida2000 are we sure that RHCS doesn't support running only on v1 ports? if its supported we also need to support the same, if not we are good, can we get this confirmation?
> 
> By support, I have to check with the support team may be. But yes on a RHCS cluster msgrv2 port can be disabled if wanted by the customer, so technically a RHCS cluster can be of v1 port only.

we need to check on this one before making this default.

controllers/storagecluster/cephcluster.go
travisn: I think the behavior we want is:
- When we run the script downstream on the RHCS cluster, if v2 port is available, always capture the v2 port. If no v2 port is available then return the v1 port.
   - In upstream, we still want to default to the v1 port since we don't know if the clients have a recent kernel. 
   - The difference between upstream/downstream hopefully can be automated, so users don't need to be aware of a new flag that they might forget to use.
- If we import the v2 port, then OCS operator can set the `requireMsgr2: true`. If it's still possible downstream to import the v1 port, then we can't set that. But if we can confirm it's not supported for RHCS, then we could always assume the v2 port is required for downstream.

malayparida2000: @parth-gr @travisn  do we really need a separate flag to extract the v2 ports?
I noticed the mons have the same address for both v1 & v2, so if we have the address of the mon for v1, for v2 shouldn't it be simply done by replacing port 6789 with port 3300?

```
sh-5.1$ ceph mon dump
epoch 4
fsid 053e09a7-bf58-4312-a3b9-2d05b5723fbf
last_changed 2023-04-30T20:56:08.009439+0000
created 2023-04-30T20:37:36.381975+0000
min_mon_release 17 (quincy)
election_strategy: 1
0: [v2:172.30.15.98:3300/0,v1:172.30.15.98:6789/0] mon.a
1: [v2:172.30.230.232:3300/0,v1:172.30.230.232:6789/0] mon.b
2: [v2:172.30.44.179:3300/0,v1:172.30.44.179:6789/0] mon.c
dumped monmap epoch 4
```

parth-gr: @malayparida2000 we need a separate flag to identify what user is requesting.
Till now in standalone cluster there can 4 conditions,
> mons only have v1 port
> mons only have v2 port
> mons have v1 and v2 port
> custom mons from user(not supported but there a BZ opened)

gguglielmetti: Hi, on latest RHCS (7.1) by default mon listen to both v1 (6789) and v2 (3300) ports, but, as previously stated, a customer can have very old clients and/or decide (even if not recommended) to enable v1 only.
So, for external cluster, I agree with @travisn approach

Ejdesgaard: iirc. v2+v1 notation were introduced when msgrv2 was introduced and it's still the default today `[v2:mon1:3300/0,v1:mon1:6789/0]`

This should allow us to just probe for v2, against each mon, and, cluster-wide, fall-back to v1, as long as features from v2 are not wished / required.

malayparida2000:
Just an update on the discussion here,
Me & Travis did get in touch with Federico(Ceph PM) & Orit. Their opinion is on the RHCS cluster since nautilus(14.2) the msgrv2 port is by default available & in use. There is no reasonable general reason for v2 port being turned off/unavailable on RHCS cluster. So we should be safe in making the v2 port default for ODF. We are also in touch with Illya about this & would soon make a decision on this.

travisn:
> Just an update on the discussion here, Me & Travis did get in touch with Federico(Ceph PM) & Orit. Their opinion is on the RHCS cluster since nautilus(14.2) the msgrv2 port is by default available & in use. There is no reasonable general reason for v2 port being turned off/unavailable on RHCS cluster. So we should be safe in making the v2 port default for ODF. We are also in touch with Illya about this & would soon make a decision on this.

Agreed, let's go ahead with this requirement to always use the v2 port for downstream, including for external clusters.

We just need to ensure the mon endpoints that are imported are also setting v2 ports. I believe this is already something that Rook enforces during the cluster reconcile. @parth-gr Could you test to confirm that if `RequireMsgr2` is set in the cephcluster CR for the external cluster scenario, we replace any v1 ports with v2 ports as part of the rook cluster reconcile?

parth-gr:
> Agreed, let's go ahead with this requirement to always use the v2 port for downstream, including for external clusters.
We just need to ensure the mon endpoints that are imported are also setting v2 ports. I believe this is already something that Rook enforces during the cluster reconcile. @parth-gr Could you test to confirm that if RequireMsgr2 is set in the cephcluster CR for the external cluster scenario, we replace any v1 ports with v2 ports as part of the rook cluster reconcile?

For the external cluster, we check if the leader endpoint provided by the Python script has v2 if it contains a v2 port rook internally update it `RequireMsgr2`: true https://github.com/rook/rook/blob/58a0c834bf5bce6782de5c387f3dac732393ef33/pkg/operator/ceph/cluster/cluster_external.go#L119

So we should first default the collection of the v2 port from the Python script and we may also add the CR spec `RequireMsgr2` to `true`

travisn:
> > Agreed, let's go ahead with this requirement to always use the v2 port for downstream, including for external clusters.
> > We just need to ensure the mon endpoints that are imported are also setting v2 ports. I believe this is already something that Rook enforces during the cluster reconcile. @parth-gr Could you test to confirm that if RequireMsgr2 is set in the cephcluster CR for the external cluster scenario, we replace any v1 ports with v2 ports as part of the rook cluster reconcile?
> 
> For the external cluster, we check if the leader endpoint provided by the Python script has v2 if it contains a v2 port rook internally update it `RequireMsgr2`: true https://github.com/rook/rook/blob/58a0c834bf5bce6782de5c387f3dac732393ef33/pkg/operator/ceph/cluster/cluster_external.go#L119
> 
> So we should first default the collection of the v2 port from the Python script and we may also add the CR spec `RequireMsgr2` to `true`

@parth-gr My question is a bit different, here are a few points to consider:
1. Upstream we don't want to default to v2, only for downstream
2. OCS operator (with this PR) is planning to set `RequireMsgr2: true` for external mode (it was already set for internal mode), so there is no need for Rook to automatically set requireMsgr2. 
3. The code you linked is perfect for upstream, but I think we need to add another check such that if `RequireMsgr2: true`, force the v2 port for the external mon endpoints.

malayparida2000:
> > > Agreed, let's go ahead with this requirement to always use the v2 port for downstream, including for external clusters.
> > > We just need to ensure the mon endpoints that are imported are also setting v2 ports. I believe this is already something that Rook enforces during the cluster reconcile. @parth-gr Could you test to confirm that if RequireMsgr2 is set in the cephcluster CR for the external cluster scenario, we replace any v1 ports with v2 ports as part of the rook cluster reconcile?
> > 
> > 
> > For the external cluster, we check if the leader endpoint provided by the Python script has v2 if it contains a v2 port rook internally update it `RequireMsgr2`: true https://github.com/rook/rook/blob/58a0c834bf5bce6782de5c387f3dac732393ef33/pkg/operator/ceph/cluster/cluster_external.go#L119
> > So we should first default the collection of the v2 port from the Python script and we may also add the CR spec `RequireMsgr2` to `true`
> 
> @parth-gr My question is a bit different, here are a few points to consider:
> 
> 1. Upstream we don't want to default to v2, only for downstream
> 2. OCS operator (with this PR) is planning to set `RequireMsgr2: true` for external mode (it was already set for internal mode), so there is no need for Rook to automatically set requireMsgr2.
> 3. The code you linked is perfect for upstream, but I think we need to add another check such that if `RequireMsgr2: true`, force the v2 port for the external mon endpoints.

To add we also might need to check,
4. For upgraded clusters, When requireMsgr2 is true, we have to make sure the rok-ceph-csi-config CM changes the mon ports to v2 ports like it does for internal mode.

controllers/storagecluster/cephcluster.go
nb-ohad, line 612: I am not sure we can blindly change that for provider mode. What happens to existing deployments?

malayparida2000: For existing deployments the mons after the upgrade would have the v1 port still available, so there would be no problem for the existing volumes that are already mounted. For any new volumes that are created after the upgrade, the volume would use the v2 port.

parth-gr:
> @parth-gr My question is a bit different, here are a few points to consider:
> 
> 1. Upstream we don't want to default to v2, only for downstream
> 2. OCS operator (with this PR) is planning to set `RequireMsgr2: true` for external mode (it was already set for internal mode), so there is no need for Rook to automatically set requireMsgr2.
> 3. The code you linked is perfect for upstream, but I think we need to add another check such that if `RequireMsgr2: true`, force the v2 port for the external mon endpoints.


For downstream we can use the `RequireMsgr2: true` info, but the problem is with extracting the data,
How do we make sure the downstream python script always scrapes the v2 port

travisn:
> For downstream we can use the `RequireMsgr2: true` info, but the problem is with extracting the data, How do we make sure the downstream python script always scrapes the v2 port

To capture our discussion:
- We will make the change in Rook to enforce the v2 port for external clusters if requiremsgr2 is true
- The external python script will be changed in the downstream-only branch to only use the v2 port.

controllers/storagecluster/cephcluster.go
nb-ohad: @malayparida2000  I am not sure that will work out of the box, don't we need to inform the CSI running on the clients that we moved to v2? If so how do we do that in provider mode? the CSI is running on many different clusters

malayparida2000:
> > For downstream we can use the `RequireMsgr2: true` info, but the problem is with extracting the data, How do we make sure the downstream python script always scrapes the v2 port
> 
> To capture our discussion:
> 
> * We will make the change in Rook to enforce the v2 port for external clusters if requiremsgr2 is true
> * The external python script will be changed in the downstream-only branch to only use the v2 port.

From Slack by Parth
```
From two changes we discussed,
I think we don't the rook upstream change
As we always check for the cluster.Spec.RequireMsgr2() https://github.com/rook/rook/blob/master/pkg/operator/ceph/cluster/cluster_external.go#L126
If it set from the CR still it will work with the same check
```
@parth-gr this should mean this PR is now no longer blocked from external mode perspective, right?

malayparida2000:
> @malayparida2000 can we do a quick testing?? (End-to-end) And after this change for sure it wont block [red-hat-storage/rook#748](https://github.com/red-hat-storage/rook/pull/748)

Sounds good to do a test, will do it.

malayparida2000:
> > @malayparida2000 can we do a quick testing?? (End-to-end) And after this change for sure it wont block [red-hat-storage/rook#748](https://github.com/red-hat-storage/rook/pull/748)
> 
> Sounds good to do a test, will do it.

Currently, there is some infra issue with the availability of external mode setups from QE & anyways, the scenario we are concerned about is the behavior of the cluster upon upgrade from 4.17 to 4.18 which will be easy to test upon this change is officially part of the build. And anyway, I have to document the whole process as well, so I will do the e2e testing of the flow after the PR is merged & part of the build.

malayparida2000:
The PR is now LGTM from both the Internal Mode & External Mode perspective.
@nb-ohad / @leelavg  Can you please ack LGTM from the Provider-Client Mode perspective so we can finally merge this?

controllers/storagecluster/cephcluster.go
leelavg: > would now get the new info from the provider cluster
- this config will be pushed to clients right after provider gets upgraded but not when the client gets upgraded
- nevertheless, based on all the info presented even 4.17 clients will have v2 endpoint which should be supported by kernel on client clusters

> don't we need to inform the CSI running on the clients that we moved to v2
- two changes will happen after provider upgrade, new v2 endpoints and mount option `ms_mode=prefer-crc` will be pushed to client clusters and IIUC they can be changed on the fly w/o any issues for both existing and new workloads.

leelavg:
LGTM, since Ohad has an unresolved comment need to wait for that to be resolved.

malayparida2000:
Ohad will be on PTO till the 21st of Oct. This is an important change & with more delay to its merge we are losing precious time we might need to make further changes based on the result of this. This PR also blocks other stories, thereby blocking the Epic's progress. As we have acks from all Povs including Provider mode from Leela, We should go ahead and merge this.
/cc @travisn

rewantsoni:
@malayparida2000 We missed rebasing it against main and running make deps-update, there are changes in metrics/vendor/ from this PR in my local

malayparida2000:
> @malayparida2000 We missed rebasing it against main and running make deps-update, there are changes in metrics/vendor/ from this PR in my local

Didn't understand, How does this PR affect the vendor folder?

PR #2815, opened 2024-09-23
Set default CephDeviceClass in the status only if it's not set yet

malayparida2000:
The sc.Status.DefaultCephDeviceClass helps us to set the deviceClasses on the pool on the CRs. Till now we are trying to determine the field in each reconcile loop. But with the introduction of multiple deviceClasses support in day-2, we can't always determine it as the logic for the determination is that it chooses the first non replica-1 deviceClass from the list of deviceClasses on the cephCluster CR status. If we determine it in each reconcile loop, we are running the risk of going back & forth between the multiple non replica-1 deviceClasses.

Although the current ceph logic returns the list in a consistent order so the problem shouldn't appear today, but any change in this logic might introduce future bugs. as we have the assurance that any additional deviceClass will be added in day-2 only so we can safely set the default deviceClass in the status only if it's not set yet & not touch it there after.

Ref-https://issues.redhat.com/browse/RHSTOR-6444

### DIFF
--- a/controllers/storagecluster/cephcluster.go
+++ b/controllers/storagecluster/cephcluster.go
@@ -264,11 +264,6 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
 		return reconcile.Result{}, nil
 	}
 
-	// Set the default CephDeviceClass in the StorageCluster status(not needed for external clusters)
-	if !sc.Spec.ExternalStorage.Enable && found.Status.CephStorage != nil {
-		sc.Status.DefaultCephDeviceClass = determineDefaultCephDeviceClass(found.Status.CephStorage.DeviceClasses, sc.Spec.ManagedResources.CephNonResilientPools.Enable, sc.Status.FailureDomainValues)
-	}
-
 	// Record actual Ceph container image version before attempting update
 	sc.Status.Images.Ceph.ActualImage = found.Spec.CephVersion.Image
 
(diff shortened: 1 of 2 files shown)


controllers/storagecluster/cephcluster.go
```diff
@@ -349,6 +344,11 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
 		}
 	}
 
+	// Set the default CephDeviceClass in the StorageCluster status if it's not set yet (not needed for external clusters)
+	if !sc.Spec.ExternalStorage.Enable && sc.Status.DefaultCephDeviceClass == "" && found.Status.CephStorage != nil {
```
parth-gr, line 348: I thought ocs-operator hardcode the value somewhere for the device class that is set to storagecluster,
Can we match the list of device class from ceph with the default value, and set it in the status output

controllers/storagecluster/cephcluster.go
```diff
@@ -1335,6 +1335,6 @@ func determineDefaultCephDeviceClass(foundDeviceClasses []rookCephv1.DeviceClass
 			}
 		}
 	}
-	// By default return "ssd"
-	return "ssd"
+	// if no device classes are found in status return empty string
+	return ""
```
travisn, line 1339: Looks good to not have a default unless one has been found on the CephCluster CR. As far as past impact of this `ssd` default, what was the behavior when the user set a deviceClass explicitly in the StorageCluster CR? Were we setting this default `ssd` incorrectly? Just wondering if we had a bug there.

malayparida2000: Yeah, it was returning this value by default, but as soon as a deviceclass would be found on the cephcluster CR status, it would change to the actual deviceclass. So I don't think it was a bug but yeah it was misleading though.

controllers/storagecluster/cephcluster.go
malayparida2000: Didn't fully understood, are you asking about setting the deviceclass on the osd's if they are not specified on storagecluster?

travisn (commented):
/lgtm

controllers/storagecluster/cephcluster.go
parth-gr: So form 4.17 we always specify the device class, correct can we not just use that value from the storagecluster CR (if user has set or default value)

parth-gr: @malayparida2000

malayparida2000: > form 4.17 we always specify the device class

I think that will be if they are created from the UI, we also need to think of what if the storagecluster is created from CLI without the deviceclass specified in the storage device set.

parth-gr: got it, I thought it was from ocs operator

PR #2817, opened 2024-09-24
rbd: Allow user to disable key rotation

black-dragon74:
This patch allows user to disable automatic
key rotation by annotating StorageCluster
with `keyrotation.csiaddons.openshift.io/enable=false`


controllers/storagecluster/storageclasses.go
```diff
@@ -281,6 +283,7 @@ func newCephBlockPoolStorageClassConfiguration(initData *ocsv1.StorageCluster) S
 	persistentVolumeReclaimDelete := corev1.PersistentVolumeReclaimDelete
 	allowVolumeExpansion := true
 	managementSpec := initData.Spec.ManagedResources.CephBlockPools
+	disableKeyRotation := !util.IsAnnotationTruthy(initData, keyRotationEnableAnnotation)
```
rchikatw, line 286: I believe we should use the variable name 'isKeyRotationEnabled'. What do you think?

controllers/storagecluster/storageclasses.go
```diff
@@ -367,6 +374,10 @@ func newNonResilientCephBlockPoolStorageClassConfiguration(initData *ocsv1.Stora
 		},
 		isClusterExternal: initData.Spec.ExternalStorage.Enable,
 	}
+	if disableKeyRotation {
```
rchikatw, line 377: The same piece of code is in 2 places. I think we can refactor it into a single method and reuse it. it.

controllers/storagecluster/storageclasses.go
black-dragon74: I decided to go with `disableKeyRotation` as it conveys the intent more clearly. If we go with enable we will have to reverse the checks.

The goal is to add the annotation on SC only when KR is disabled.

WDYT?

rchikatw: What I mean here is something like this:
	isKeyRotationEnabled := util.IsAnnotationTruthy(initData, keyRotationEnableAnnotation)
Later you do the logic based on this var, 
The main reason to use the value as is in the variable and perform the logic accordingly. Both ways are the same, but it's your choice which one you prefer.

controllers/storagecluster/storageclasses.go
black-dragon74: Done

controllers/storagecluster/storageclasses.go
```diff
@@ -27,6 +27,8 @@ const (
 
 	//storage class driver name prefix
 	storageclassDriverNamePrefix = "openshift-storage"
+
```
nb-ohad, line 30: No need for a separating line, both const are private.

controllers/storagecluster/storageclasses.go
```diff
@@ -314,6 +316,9 @@ func newCephBlockPoolStorageClassConfiguration(initData *ocsv1.StorageCluster) S
 	if initData.Spec.ManagedResources.CephBlockPools.DefaultStorageClass {
 		scc.storageClass.Annotations[defaultStorageClassAnnotation] = "true"
 	}
+	util.If(!util.IsAnnotationTruthy(initData, keyRotationEnableAnnotation), func() {
+		util.AddAnnotation(scc.storageClass, keyRotationEnableAnnotation, "false")
+	})
 	return scc
```
nb-ohad, line 322: Don't we need to issue an update comment for the storagecluster to apply the new annotation?

controllers/util/util.go
```diff
@@ -103,3 +105,21 @@ func AssertEqual[T comparable](actual T, expected T, exitCode int) {
...
+
+// IsAnnotationTruthy returns true if the annotation is present
+// and has a truthy value
+func IsAnnotationTruthy(obj metav1.Object, key string) bool {
+	annotations := obj.GetAnnotations()
+
+	if val, found := annotations[key]; found {
+		return strings.ToLower(val) == "true"
+	}
+	return false
```
nb-ohad, line 117: The entire function just equal to
```
return obj.GetAnnotations()(key) == "true"
```

And it that case we don't need a util at all, just use the above line in the call sites

controllers/util/util.go
```diff
@@ -103,3 +105,21 @@ func AssertEqual[T comparable](actual T, expected T, exitCode int) {
...
+}
+
+// Execute the provided function if the condition is true.
+func If(cond bool, fn func()) {
+	if cond {
+		fn()
+	}
+}
```
nb-ohad, line 125: This util does not make sense. What would be a good reason to call a function and pass another function into it instead of just putting a normal if in place? 

Usually one creates an `IfElse` utils for cases where you want to treat the `if` as an expression (not as a statement).  In that case the implementation will be something like

```
func IfElse[T any](cond bool, trueVal, falseVal T) T {
    if cond {
       return trueVal
    } else {
        return falseVal
    }
}
```

To enable you doing stuff like:
```
instance := someStruct {
      ...
      b: util.IfElse(someCond, Val1, val2),
      ...
}
```

But this is not the case here and the implementation above does not provide any value. 
Please remove it

nb-ohad (changes requested):

controllers/storagecluster/storageclasses.go
black-dragon74: `newCephBlockPoolStorageClassConfiguration` generates config for a new SC that is going to be created. Upon creation, it will have the annotation present.

controllers/storagecluster/storageclasses.go
```diff
@@ -314,6 +316,9 @@ func newCephBlockPoolStorageClassConfiguration(initData *ocsv1.StorageCluster) S
 	if initData.Spec.ManagedResources.CephBlockPools.DefaultStorageClass {
 		scc.storageClass.Annotations[defaultStorageClassAnnotation] = "true"
 	}
+	util.If(!util.IsAnnotationTruthy(initData, keyRotationEnableAnnotation), func() {
+		util.AddAnnotation(scc.storageClass, keyRotationEnableAnnotation, "false")
+	})
```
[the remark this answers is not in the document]
black-dragon74: This is exactly how it was initially. It was suggested to have a util to prevent duplication in an earlier review so I added it :D

controllers/util/util.go
black-dragon74: > This util does not make sense.

Couldn't agree more :)

controllers/storagecluster/storageclasses.go
```diff
@@ -366,6 +370,10 @@ func newNonResilientCephBlockPoolStorageClassConfiguration(initData *ocsv1.Stora
 		},
 		isClusterExternal: initData.Spec.ExternalStorage.Enable,
 	}
+	if !(initData.GetAnnotations()[keyRotationEnableAnnotation] == "true") {
```
Madhu-1, line 373:
```suggestion
initData.GetAnnotations()[keyRotationEnableAnnotation] == "false")
```
How about making it simple check and expect only false value?

Madhu-1 (commented):
I think we dont have code that updates the storageclass, can you please check/test what happens if we add/remove the annotation?

controllers/storagecluster/storageclasses.go
black-dragon74: Done.

controllers/storagecluster/storageclasses.go
```diff
@@ -366,6 +371,10 @@ func newNonResilientCephBlockPoolStorageClassConfiguration(initData *ocsv1.Stora
 		},
 		isClusterExternal: initData.Spec.ExternalStorage.Enable,
 	}
+	if initData.GetAnnotations()[keyRotationEnableAnnotation] == "false" {
+		util.AddAnnotation(scc.storageClass, keyRotationEnableAnnotation, "false")
```
Madhu-1, line 375: if the annotation is removed on the storagecluster same need to be removed on the SC as well right?

black-dragon74: The StorageClass that we return from this function replaces the existing one. And we only add the annotation to this NEW StorageClass if we find it on StorageCluster.

If the annotation is removed, the new SC won't have it present and since the annotations will be different between this NEW SC and existing SC, the existing SC (with disable annotation) will be replaced with this new SC (which does not have the annotation set)

controllers/storagecluster/storageclasses.go
```diff
@@ -217,7 +218,8 @@ func (r *StorageClusterReconciler) createStorageClasses(sccs []StorageClassConfi
 			if existing.DeletionTimestamp != nil {
 				return fmt.Errorf("failed to restore StorageClass  %s because it is marked for deletion", existing.Name)
 			}
-			if !reflect.DeepEqual(sc.Parameters, existing.Parameters) {
+			if !reflect.DeepEqual(sc.Parameters, existing.Parameters) ||
+				!reflect.DeepEqual(sc.Annotations, existing.Annotations) {
```
Madhu-1, line 222: Checking if this creates the problem when we add annotation on the SC like for DR or not (which am not sure), @ShyamsundarR @umangachapagain will there be any problem if we override the annotations on the SC?

Madhu-1 (commented):
LGTM

controllers/storagecluster/storageclasses.go
umangachapagain: As long as we merge all the annotations and not remove it, we should be fine.

ShyamsundarR: Recreating (or updating) the StorageClass should preserve existing annotations and labels, so if the change is overwriting these then we would lose information, which should not happen. (I have not looked at code, answering based on the comment here)

black-dragon74: @nb-ohad I recall you saying somewhere that the SCs created by the ocs-operator are managed and user is not supposed to make edits to it. Would like to know your thoughts on this?

ShyamsundarR: The spec section and name (or related metadata) should be retained by the operator. User additions like annotations and labels (and even finalizers) should not be overwritten.

black-dragon74: In cases where the params do not differ, I have updated the PR to update the annotations instead of recreating the SC.

cc: @Madhu-1

controllers/storagecluster/storageclasses.go
```diff
@@ -232,6 +234,16 @@ func (r *StorageClusterReconciler) createStorageClasses(sccs []StorageClassConfi
 					r.Log.Info("Failed to create StorageClass.", "StorageClass", klog.KRef(sc.Namespace, sc.Name))
 					return err
 				}
+				scRecreated = true
+			}
+			if !scRecreated {
```
ShyamsundarR, line 239: I would recreate the SC in line 232 with the annotations and labels already present. The labels are also important.

If the reconcile fails to update the SC annotations in this line, a future reconcile will not have the copy of the existing annotations etc. and hence this will never be set on the recreated SC.

black-dragon74: The SC being recreated on line 232 already has the required annotations and labels if any. 

If it is not recreated (same params), only then we try to add the annotation by updating the SC.

This is what you meant, right?

ShyamsundarR: If SC is not recreated you are updating annotations to `existing` SC and updating it, this is fine.

I am still missing how an SC that is recreated (due to parameters differing) preserves annotations and labels from `existing`? IOW this [comment](https://github.com/red-hat-storage/ocs-operator/pull/2817#discussion_r1829111436)

controllers/storagecluster/storageclasses.go
```diff
@@ -232,6 +234,16 @@ func (r *StorageClusterReconciler) createStorageClasses(sccs []StorageClassConfi
...
+				util.AddAnnotation(existing, keyRotationEnableAnnotation, sc.GetAnnotations()[keyRotationEnableAnnotation])
+
+				err = r.Client.Update(context.TODO(), existing)
```
Madhu-1, line 242: can you please check if we have RBAC to update the SC?

controllers/storagecluster/storageclasses.go
Madhu-1: We need to preserve the annotation/labels from the SC before delete and reapply that before creating.

Madhu-1 (changes requested):

controllers/storagecluster/storageclasses.go
black-dragon74: I would say doing so is out of scope of this PR. The SC was being recreated earlier the same way. 

As far as key rotation is concerned, we consider storage cluster as source of truth. We add/remove annotations to SC accordingly. WDYT?

Madhu-1: Sure, let's open a tracker for it.

black-dragon74: Perfect! I'd gladly follow this up as an enhancement. Any suggestions specific to KR on this PR?

Madhu-1: https://github.com/red-hat-storage/ocs-operator/blob/99fdb662050ddbd412e0a240c477e84ea9dab25d/config/rbac/role.yaml#L312-L321 RBAC comment i mentioned below need to be addressed, we dont have update access

black-dragon74: @Madhu-1 done!

controllers/storagecluster/storageclasses.go
```diff
@@ -232,6 +234,16 @@ func (r *StorageClusterReconciler) createStorageClasses(sccs []StorageClassConfi
 					r.Log.Info("Failed to create StorageClass.", "StorageClass", klog.KRef(sc.Namespace, sc.Name))
 					return err
 				}
+				scRecreated = true
+			}
+			if !scRecreated {
+				util.AddAnnotation(existing, keyRotationEnableAnnotation, sc.GetAnnotations()[keyRotationEnableAnnotation])
```
Madhu-1, line 240: @black-dragon74 here we need to get the annotation from storagecluster and set it/Remove it right?

black-dragon74: The annotation will be present on `sc` (the new storageclass) only when it is present on the storagecluster. Hence here we are reading from the `sc` instead of reading from the `initData` (storagecluster) again.

Madhu-1: what about the case when the annotation is removed from the storagecluster? i didnt see the code where we are removing the annotation from SC, Are we setting the key and empty value? If thats the case, i would suggest to remove the key or set it to true explicitly

black-dragon74: Yes, `util.AddAnnotation` will set the value of the annotation to an empty string. Let's change it to:

```go
krEnabled := sc.GetAnnotations()[keyRotationEnableAnnotation]
if krEnabled != "false" {
	krEnabled = "true"
}
util.AddAnnotation(existing, keyRotationEnableAnnotation, krEnabled)
```
This is more explicit and won't set an empty value.

black-dragon74: Done!

Madhu-1 (approved):

iamniting (changes requested):

controllers/storagecluster/storageclasses.go
```diff
@@ -27,6 +27,7 @@ const (
 
 	//storage class driver name prefix
 	storageclassDriverNamePrefix = "openshift-storage"
+	keyRotationEnableAnnotation  = "keyrotation.csiaddons.openshift.io/enable"
```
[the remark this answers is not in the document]
black-dragon74: Sure.

controllers/storagecluster/storageclasses.go
```diff
@@ -232,6 +234,20 @@ func (r *StorageClusterReconciler) createStorageClasses(sccs []StorageClassConfi
 					r.Log.Info("Failed to create StorageClass.", "StorageClass", klog.KRef(sc.Namespace, sc.Name))
 					return err
 				}
+				scRecreated = true
+			}
+			if !scRecreated {
+				// Delete existing key rotation annotation and set it on sc only when it is false
+				delete(existing.Annotations, keyRotationEnableAnnotation)
+				if krState := sc.GetAnnotations()[keyRotationEnableAnnotation]; krState == "false" {
```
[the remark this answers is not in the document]
black-dragon74: Annotation is added on the `sc` only when it is present on the storagecluster. 
```go
if initData.GetAnnotations()[keyRotationEnableAnnotation] == "false" {
    util.AddAnnotation(scc.storageClass, keyRotationEnableAnnotation, "false")
}
```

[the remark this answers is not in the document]
iamniting: I am resolving this as this is same as Madhu comment.

controllers/storagecluster/storageclasses.go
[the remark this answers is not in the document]
iamniting: Also pls move this annotation to the default package, So that it can be used by other controllers as well when required.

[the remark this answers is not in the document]
black-dragon74: Done. PTAL now :)

iamniting (approved):

black-dragon74:
> @black-dragon74 Have you tested it? If not pls add a hold on the PR and test it before it gets merged.

I tested an older version of this PR. The core implementation did not change. I am putting a hold, will cancel it once I test it out.

/hold

iamniting:
> > @black-dragon74 Have you tested it? If not pls add a hold on the PR and test it before it gets merged.
> 
> I tested an older version of this PR. The core implementation did not change. I am putting a hold, will cancel it once I test it out.
> 
> /hold

Thanks

black-dragon74:
Tested with the latest set of changes. Everything is working as expected.

/hold cancel

PR #2822, opened 2024-09-26
Refactoring of 'replaceToken' code which changes PrometheusRules

aruniiird:
Added more details to 'replaceToken' struct which give more details about where the change is going to be reflected in the rule. Added more usecases to tests.


controllers/storagecluster/cephcluster.go
```diff
@@ -1150,32 +1150,35 @@ func createPrometheusRules(r *StorageClusterReconciler, sc *ocsv1.StorageCluster
...
 	specifiedBackfillFullRatio := sc.Spec.ManagedResources.CephCluster.BackfillFullRatio
 	specifiedFullRatio := sc.Spec.ManagedResources.CephCluster.FullRatio
 
 	if specifiedNearFullRatio != nil {
 		replaceTokens = append(replaceTokens,
-			createReplaceToken("", "", "75%", fmt.Sprintf("%.2f%%", *specifiedNearFullRatio*100)),
-			createReplaceToken("", "", "0.75", fmt.Sprintf("%f", *specifiedNearFullRatio)))
+			createReplaceToken("cluster-utilization-alert.rules", "CephClusterNearFull",
+				"75%", fmt.Sprintf("%0.0f%%", *specifiedNearFullRatio*100), AnnotationRuleSection, DescriptionKey),
+			createReplaceToken("cluster-utilization-alert.rules", "CephClusterNearFull",
```
malayparida2000, line 1166: 75 % & 0.75 needs to be replaced in CephClusterNearFull & CephOSDNearFull both

controllers/storagecluster/cephcluster.go
```diff
@@ -1150,32 +1150,35 @@ func createPrometheusRules(r *StorageClusterReconciler, sc *ocsv1.StorageCluster
...
+				"0.75", fmt.Sprintf("%0.2f", *specifiedNearFullRatio), ExprRuleSection, ""))
 	}
 	if specifiedBackfillFullRatio != nil {
 		replaceTokens = append(replaceTokens,
-			createReplaceToken("", "", "80%", fmt.Sprintf("%.2f%%", *specifiedBackfillFullRatio*100)),
-			createReplaceToken("", "", "0.80", fmt.Sprintf("%f", *specifiedBackfillFullRatio)))
+			createReplaceToken("cluster-utilization-alert.rules", "CephClusterCriticallyFull",
+				"80%", fmt.Sprintf("%0.0f%%", *specifiedBackfillFullRatio*100), AnnotationRuleSection, DescriptionKey),
```
malayparida2000, line 1172: 80%/0.80 needs to replaced in CephClusterCriticallyFull & CephOSDCriticallyFull.

controllers/storagecluster/cephcluster.go
```diff
@@ -1150,32 +1150,35 @@ func createPrometheusRules(r *StorageClusterReconciler, sc *ocsv1.StorageCluster
...
+			createReplaceToken("cluster-utilization-alert.rules", "CephClusterCriticallyFull",
+				"0.80", fmt.Sprintf("%0.2f", *specifiedBackfillFullRatio), ExprRuleSection, ""))
 	}
 	if specifiedFullRatio != nil {
 		replaceTokens = append(replaceTokens,
-			createReplaceToken("", "", "85%", fmt.Sprintf("%.2f%%", *specifiedFullRatio*100)),
-			createReplaceToken("", "", "0.85", fmt.Sprintf("%f", *specifiedFullRatio)))
+			createReplaceToken("cluster-utilization-alert.rules", "CephClusterReadOnly",
+				"85%", fmt.Sprintf("%0.0f%%", *specifiedFullRatio*100), AnnotationRuleSection, DescriptionKey),
```
malayparida2000, line 1179: 85% needs to be replaced in CephClusterNearFull, CephClusterCriticallyFull, CephClusterReadOnly

malayparida2000 (changes requested):
The initial idea of not specifying rule names for the osd/cluster utilization numbers was due to the fact that they were being changed across multiple rules, and we wanted the value to be replaced in all the rules. So if we want to cover it with particular rules we have to be careful to cover them in all the rules they appear.

controllers/storagecluster/cephcluster.go
aruniiird: For `CephOSDNearFull` alert , are we replacing it with same `*specifiedNearFullRatio` value?
Or do we have to add a separate set of fields under `ManageCephCluster` struct (similar to the fields we currently added like `NearFullRatio`, `BackfillFullRatio` and `FullRatio`)?

controllers/storagecluster/cephcluster.go
aruniiird: Same question as above, but regarding `CephOSDCriticallyFull` alert, are we going to use the same `*BackfillFullRatio` value or create new fields for CephOSD* alerts?

controllers/storagecluster/cephcluster.go
aruniiird: Done

malayparida2000:
Hi @aruniiird do you Plan to work on this? if not can you please close it

controllers/storagecluster/cephcluster.go
aruniiird: Done

aruniiird:
@malayparida2000 , Addressed the comments/concerns. Can you please take another look?

PR #2823, opened 2024-09-27
Fix metrics params for k8s service monitor

ezio-auditore:
Currently the params are generated in k8sservice monitor to federate in-cluster kube metrics as below which is wrong. This patch fixes the correct string params.
```
params:
      match[]:
      - '{__name__=''kube_node_status_condition''}'
      - '{__name__=''kube_persistentvolume_info''}'
      - '{__name__=''kube_storageclass_info''}'
      - '{__name__=''kube_persistentvolumeclaim_info''}'
      - '{__name__=''kube_deployment_spec_replicas''}'
      - '{__name__=''kube_pod_status_phase''}'
      - '{__name__=''kubelet_volume_stats_capacity_bytes''}'
      - '{__name__=''kubelet_volume_stats_used_bytes''}'
      - '{__name__=''node_disk_read_time_seconds_total''}'
      - '{__name__=''node_disk_write_time_seconds_total''}'
      - '{__name__=''node_disk_reads_completed_total''}'
      - '{__name__=''node_disk_writes_completed_total''}'
```

Also fixed tech debt to fetch kube-rbac-proxy image from environment variables

### DIFF
--- a/templates/prometheus.go
+++ b/templates/prometheus.go
@@ -2,6 +2,7 @@ package templates
 
 import (
 	"fmt"
+	"os"
 
 	promv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	"github.com/red-hat-storage/ocs-operator/v4/controllers/defaults"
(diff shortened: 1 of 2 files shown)


templates/prometheus.go
```diff
@@ -29,9 +30,8 @@ var PrometheusSpecTemplate = promv1.PrometheusSpec{
 		ListenLocal:            true,
 		Resources:              defaults.MonitoringResources["prometheus"],
 		Containers: []corev1.Container{{
-			Name: "kube-rbac-proxy",
-			// Tech-debt to include kube-rbac-proxy image from environment
-			Image: "gcr.io/kubebuilder/kube-rbac-proxy:v0.13.0",
+			Name:  "kube-rbac-proxy",
+			Image: os.Getenv("KUBE_RBAC_PROXY_IMAGE"),
```
umangachapagain, line 34: I don't remember if we ensure anywhere that this env is set?

ezio-auditore: We do ensure this env is set here 
https://github.com/ezio-auditore/ocs-operator/blob/6defc783a9f091af1d9ab6662c84b7bc2df8b98f/controllers/storagecluster/storagecluster_controller.go#L46

umangachapagain: ah ok.

umangachapagain (approved):

PR #2826, opened 2024-09-30
fix panic in provider server while onboarding

rewantsoni:
Fix provider server panic while onboarding Consumers when the value of the storageQuota in the onboardingTicket is set to nil

Introduced with: https://github.com/red-hat-storage/ocs-operator/pull/2669

### DIFF
--- a/services/provider/server/server.go
+++ b/services/provider/server/server.go
@@ -11,6 +11,7 @@ import (
 	"encoding/json"
 	"encoding/pem"
 	"fmt"
+	"k8s.io/utils/ptr"
 	"math"
 	"net"
 	"slices"
@@ -119,6 +120,7 @@ func (s *OCSProviderServer) OnboardConsumer(ctx context.Context, req *pb.Onboard
 		klog.Errorf("failed to validate onboarding ticket for consumer %q. %v", req.ConsumerName, err)
 		return nil, status.Errorf(codes.InvalidArgument, "onboarding ticket is not valid. %v", err)
 	}
+	storageQuotaInGiB := ptr.Deref(onboardingTicket.StorageQuotaInGiB, 0)
 
 	if onboardingTicket.SubjectRole != services.ClientRole {
 		err := fmt.Errorf("unsupported ticket role for consumer %q, found %s, expected %s", req.ConsumerName, onboardingTicket.SubjectRole, services.ClientRole)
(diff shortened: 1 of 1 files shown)


leelavg (commented):
pls mention which PR introduced this, not able to recall, thanks.

services/provider/server/consumer.go
```diff
@@ -74,8 +74,13 @@ func (c *ocsConsumerManager) Create(ctx context.Context, onboard ifaces.StorageC
 			},
 		},
 		Spec: ocsv1alpha1.StorageConsumerSpec{
-			Enable:            false,
-			StorageQuotaInGiB: storageQuotaInGiB,
+			Enable: false,
+			StorageQuotaInGiB: func() int {
+				if storageQuotaInGiB != nil {
+					return int(*storageQuotaInGiB)
```
leelavg, line 80: converting uint to int here, yes, we may not have larger values where we overflow but why not using `uint(*storageQuotaInGiB)`?

leelavg: ouch, I think there was some discussion which made this field in token struct as uint, seems non consistent with only a benefit of reducing one extra check for signedness.

Anyways, can't change that now.

nb-ohad: What is the issue with converting the unit to int? We are not expecting very large numbers here.
How does moving the conversion from outside the function into the function solves the problem?

nb-ohad (changes requested):

rewantsoni:
>What is the issue with converting the unit to int? We are not expecting very large numbers here.
How does moving the conversion from outside the function into the function solves the problem?

@nb-ohad The issue is with converting the pointer to unit to int. If the storageQuotaInGib is nil it will panic. I moved everything inside a inline function to make it clean and return 0 as storageQuotaInGib if the value is nil (we could have the same check outside the function as well)

nb-ohad:
@rewantsoni You should not convert a pointer. You should convert the value is not the same thing. 
Converting a pointer means that you are trying to reference the same underlying value but at the same time interpreting it as something else. That is wrong in this case. 

You should convert the value (outside of the function), and send the converted value into the function.

nb-ohad:
A good fix for the bug will be
```
if size := onboardingTicket.StorageQuotaInGiBl size == nil || *size > math.MaxInt {
    // return some error
} 
storageConsumerUUID, err := s.consumerManager.Create(ctx, req, int(*size))

PR #2827, opened 2024-09-30
Read secrets for client-onboarding-token-validation

mrudraia1:
This PR reads the secrets instead of reading the secrets from the volume mounts.
whenever the new onboarding secrets are created, it takes more time to read the secrets from the volume mounts,
The user clicks the rotate onboarding keys, the kubernetes still uses the old public, private keys , the new keys are mounted later, So this PR will read the secrets directly from the kubernetes secrets.

### DIFF
--- a/controllers/storagecluster/storageclient.go
+++ b/controllers/storagecluster/storageclient.go
@@ -17,8 +17,7 @@ import (
 )
 
 const (
-	tokenLifetimeInHours         = 48
-	onboardingPrivateKeyFilePath = "/etc/private-key/key"
+	tokenLifetimeInHours = 48
 
 	ocsClientConfigMapName = "ocs-client-operator-config"
 	manageNoobaaSubKey     = "manageNoobaaSubscription"
@@ -47,7 +46,11 @@ func (s *storageClient) ensureCreated(r *StorageClusterReconciler, storagecluste
 	storageClient.Name = storagecluster.Name
 	_, err := controllerutil.CreateOrUpdate(r.ctx, r.Client, storageClient, func() error {
 		if storageClient.Status.ConsumerID == "" {
-			token, err := util.GenerateClientOnboardingToken(tokenLifetimeInHours, onboardingPrivateKeyFilePath, nil, storagecluster.UID)
+			privateKey, err := util.LoadOnboardingValidationPrivateKey(r.ctx, r.Client, r.OperatorNamespace)
+			if err != nil {
+				return fmt.Errorf("unable to get Parsed Private Key: %v", err)
+			}
+			token, err := util.GenerateClientOnboardingToken(tokenLifetimeInHours, privateKey, nil, storagecluster.UID)
 			if err != nil {
 				return fmt.Errorf("unable to generate onboarding token: %v", err)
 			}
--- a/controllers/util/provider.go
+++ b/controllers/util/provider.go
@@ -1,6 +1,7 @@
 package util
 
 import (
+	"context"
 	"crypto"
 	"crypto/rand"
 	"crypto/rsa"
@@ -10,19 +11,22 @@ import (
 	"encoding/json"
 	"encoding/pem"
 	"fmt"
-	"os"
 	"time"
 
 	"github.com/red-hat-storage/ocs-operator/v4/services"
 
 	"github.com/google/uuid"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
+
+const onboardingValidationPrivateKeySecretName = "onboarding-private-key"
 
 // GenerateClientOnboardingToken generates a ocs-client token valid for a duration of "tokenLifetimeInHours".
 // The token content is predefined and signed by the private key which'll be read from supplied "privateKeyPath".
 // The storageQuotaInGiB is optional, and it is used to limit the storage of PVC in the application cluster.
-func GenerateClientOnboardingToken(tokenLifetimeInHours int, privateKeyPath string, storageQuotainGib *uint, storageClusterUID types.UID) (string, error) {
+func GenerateClientOnboardingToken(tokenLifetimeInHours int, privateKey *rsa.PrivateKey, storageQuotainGib *uint, storageClusterUID types.UID) (string, error) {
 	tokenExpirationDate := time.Now().
 		Add(time.Duration(tokenLifetimeInHours) * time.Hour).
 		Unix()
@@ -35,7 +39,7 @@ func GenerateClientOnboardingToken(tokenLifetimeInHours int, privateKeyPath stri
 		StorageCluster:    storageClusterUID,
 	}
 
-	token, err := encodeAndSignOnboardingToken(privateKeyPath, ticket)
+	token, err := encodeAndSignOnboardingToken(privateKey, ticket)
 	if err != nil {
 		return "", err
 	}
@@ -44,7 +48,7 @@ func GenerateClientOnboardingToken(tokenLifetimeInHours int, privateKeyPath stri
 
 // GeneratePeerOnboardingToken generates a ocs-peer token valid for a duration of "tokenLifetimeInHours".
 // The token content is predefined and signed by the private key which'll be read from supplied "privateKeyPath".
-func GeneratePeerOnboardingToken(tokenLifetimeInHours int, privateKeyPath string, storageClusterUID types.UID) (string, error) {
+func GeneratePeerOnboardingToken(tokenLifetimeInHours int, privateKey *rsa.PrivateKey, storageClusterUID types.UID) (string, error) {
 	tokenExpirationDate := time.Now().
 		Add(time.Duration(tokenLifetimeInHours) * time.Hour).
 		Unix()
@@ -55,7 +59,7 @@ func GeneratePeerOnboardingToken(tokenLifetimeInHours int, privateKeyPath string
 		SubjectRole:    services.PeerRole,
 		StorageCluster: storageClusterUID,
 	}
-	token, err := encodeAndSignOnboardingToken(privateKeyPath, ticket)
+	token, err := encodeAndSignOnboardingToken(privateKey, ticket)
 	if err != nil {
 		return "", err
 	}
@@ -64,7 +68,7 @@ func GeneratePeerOnboardingToken(tokenLifetimeInHours int, privateKeyPath string
 
 // encodeAndSignOnboardingToken generates a token from the ticket.
 // The token content is predefined and signed by the private key which'll be read from supplied "privateKeyPath".
-func encodeAndSignOnboardingToken(privateKeyPath string, ticket services.OnboardingTicket) (string, error) {
+func encodeAndSignOnboardingToken(privateKey *rsa.PrivateKey, ticket services.OnboardingTicket) (string, error) {
 	payload, err := json.Marshal(ticket)
 	if err != nil {
 		return "", fmt.Errorf("failed to marshal the payload: %v", err)
@@ -77,11 +81,6 @@ func encodeAndSignOnboardingToken(privateKeyPath string, ticket services.Onboard
 	_, err = msgHash.Write(payload)
 	if err != nil {
 		return "", fmt.Errorf("failed to hash onboarding token payload: %v", err)
-	}
-
-	privateKey, err := readAndDecodePrivateKey(privateKeyPath)
-	if err != nil {
-		return "", fmt.Errorf("failed to read and decode private key: %v", err)
 	}
 
 	msgHashSum := msgHash.Sum(nil)
@@ -97,16 +96,30 @@ func encodeAndSignOnboardingToken(privateKeyPath string, ticket services.Onboard
 	return fmt.Sprintf("%s.%s", encodedPayload, encodedSignature), nil
 }
 
-func readAndDecodePrivateKey(privateKeyPath string) (*rsa.PrivateKey, error) {
-	pemString, err := os.ReadFile(privateKeyPath)
+func LoadOnboardingValidationPrivateKey(ctx context.Context, cl client.Client, namespace string) (*rsa.PrivateKey, error) {
+	privateSecret := &corev1.Secret{}
+	privateSecret.Name = onboardingValidationPrivateKeySecretName
+	privateSecret.Namespace = namespace
+
+	err := cl.Get(ctx, client.ObjectKeyFromObject(privateSecret), privateSecret)
 	if err != nil {
-		return nil, fmt.Errorf("failed to read private key: %v", err)
+		return nil, fmt.Errorf("failed to get private secret: %v", err)
 	}
 
-	Block, _ := pem.Decode(pemString)
-	privateKey, err := x509.ParsePKCS1PrivateKey(Block.Bytes)
+	privateSecretKey := privateSecret.Data["key"]
+	if len(privateSecretKey) == 0 {
+		return nil, fmt.Errorf("No data found in secret")
+	}
+
+	block, rest := pem.Decode(privateSecretKey)
+	if len(rest) > 0 {
+		return nil, fmt.Errorf("PEM block not found in private key: %s", rest)
+	}
+
+	privateKey, err := x509.ParsePKCS1PrivateKey(block.Bytes)
 	if err != nil {
 		return nil, fmt.Errorf("failed to parse private key: %v", err)
 	}
+
 	return privateKey, nil
 }
--- a/deploy/ocs-operator/manifests/onboarding-validation-keys-generator-role.yaml
+++ b/deploy/ocs-operator/manifests/onboarding-validation-keys-generator-role.yaml
@@ -19,3 +19,10 @@ rules:
     verbs:
       - get
       - list
+  - apiGroups:
+      - ocs.openshift.io
+    resources:
+      - storageclusters/finalizers
+    verbs:
+      - update
+
--- a/metrics/vendor/github.com/red-hat-storage/ocs-operator/v4/controllers/util/provider.go
+++ b/metrics/vendor/github.com/red-hat-storage/ocs-operator/v4/controllers/util/provider.go
@@ -1,6 +1,7 @@
 package util
 
 import (
+	"context"
 	"crypto"
 	"crypto/rand"
 	"crypto/rsa"
@@ -10,19 +11,22 @@ import (
 	"encoding/json"
 	"encoding/pem"
 	"fmt"
-	"os"
 	"time"
 
 	"github.com/red-hat-storage/ocs-operator/v4/services"
 
 	"github.com/google/uuid"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
+
+const onboardingValidationPrivateKeySecretName = "onboarding-private-key"
 
 // GenerateClientOnboardingToken generates a ocs-client token valid for a duration of "tokenLifetimeInHours".
 // The token content is predefined and signed by the private key which'll be read from supplied "privateKeyPath".
 // The storageQuotaInGiB is optional, and it is used to limit the storage of PVC in the application cluster.
-func GenerateClientOnboardingToken(tokenLifetimeInHours int, privateKeyPath string, storageQuotainGib *uint, storageClusterUID types.UID) (string, error) {
+func GenerateClientOnboardingToken(tokenLifetimeInHours int, privateKey *rsa.PrivateKey, storageQuotainGib *uint, storageClusterUID types.UID) (string, error) {
 	tokenExpirationDate := time.Now().
 		Add(time.Duration(tokenLifetimeInHours) * time.Hour).
 		Unix()
@@ -35,7 +39,7 @@ func GenerateClientOnboardingToken(tokenLifetimeInHours int, privateKeyPath stri
 		StorageCluster:    storageClusterUID,
 	}
 
-	token, err := encodeAndSignOnboardingToken(privateKeyPath, ticket)
+	token, err := encodeAndSignOnboardingToken(privateKey, ticket)
 	if err != nil {
 		return "", err
 	}
@@ -44,7 +48,7 @@ func GenerateClientOnboardingToken(tokenLifetimeInHours int, privateKeyPath stri
 
 // GeneratePeerOnboardingToken generates a ocs-peer token valid for a duration of "tokenLifetimeInHours".
 // The token content is predefined and signed by the private key which'll be read from supplied "privateKeyPath".
-func GeneratePeerOnboardingToken(tokenLifetimeInHours int, privateKeyPath string, storageClusterUID types.UID) (string, error) {
+func GeneratePeerOnboardingToken(tokenLifetimeInHours int, privateKey *rsa.PrivateKey, storageClusterUID types.UID) (string, error) {
 	tokenExpirationDate := time.Now().
 		Add(time.Duration(tokenLifetimeInHours) * time.Hour).
 		Unix()
@@ -55,7 +59,7 @@ func GeneratePeerOnboardingToken(tokenLifetimeInHours int, privateKeyPath string
 		SubjectRole:    services.PeerRole,
 		StorageCluster: storageClusterUID,
 	}
-	token, err := encodeAndSignOnboardingToken(privateKeyPath, ticket)
+	token, err := encodeAndSignOnboardingToken(privateKey, ticket)
 	if err != nil {
 		return "", err
 	}
@@ -64,7 +68,7 @@ func GeneratePeerOnboardingToken(tokenLifetimeInHours int, privateKeyPath string
 
 // encodeAndSignOnboardingToken generates a token from the ticket.
 // The token content is predefined and signed by the private key which'll be read from supplied "privateKeyPath".
-func encodeAndSignOnboardingToken(privateKeyPath string, ticket services.OnboardingTicket) (string, error) {
+func encodeAndSignOnboardingToken(privateKey *rsa.PrivateKey, ticket services.OnboardingTicket) (string, error) {
 	payload, err := json.Marshal(ticket)
 	if err != nil {
 		return "", fmt.Errorf("failed to marshal the payload: %v", err)
@@ -77,11 +81,6 @@ func encodeAndSignOnboardingToken(privateKeyPath string, ticket services.Onboard
 	_, err = msgHash.Write(payload)
 	if err != nil {
 		return "", fmt.Errorf("failed to hash onboarding token payload: %v", err)
-	}
-
-	privateKey, err := readAndDecodePrivateKey(privateKeyPath)
-	if err != nil {
-		return "", fmt.Errorf("failed to read and decode private key: %v", err)
 	}
 
 	msgHashSum := msgHash.Sum(nil)
@@ -97,16 +96,30 @@ func encodeAndSignOnboardingToken(privateKeyPath string, ticket services.Onboard
 	return fmt.Sprintf("%s.%s", encodedPayload, encodedSignature), nil
 }
 
-func readAndDecodePrivateKey(privateKeyPath string) (*rsa.PrivateKey, error) {
-	pemString, err := os.ReadFile(privateKeyPath)
+func LoadOnboardingValidationPrivateKey(ctx context.Context, cl client.Client, namespace string) (*rsa.PrivateKey, error) {
+	privateSecret := &corev1.Secret{}
+	privateSecret.Name = onboardingValidationPrivateKeySecretName
+	privateSecret.Namespace = namespace
+
+	err := cl.Get(ctx, client.ObjectKeyFromObject(privateSecret), privateSecret)
 	if err != nil {
-		return nil, fmt.Errorf("failed to read private key: %v", err)
+		return nil, fmt.Errorf("failed to get private secret: %v", err)
 	}
 
-	Block, _ := pem.Decode(pemString)
-	privateKey, err := x509.ParsePKCS1PrivateKey(Block.Bytes)
+	privateSecretKey := privateSecret.Data["key"]
+	if len(privateSecretKey) == 0 {
+		return nil, fmt.Errorf("No data found in secret")
+	}
+
+	block, rest := pem.Decode(privateSecretKey)
+	if len(rest) > 0 {
+		return nil, fmt.Errorf("PEM block not found in private key: %s", rest)
+	}
+
+	privateKey, err := x509.ParsePKCS1PrivateKey(block.Bytes)
 	if err != nil {
 		return nil, fmt.Errorf("failed to parse private key: %v", err)
 	}
+
 	return privateKey, nil
 }
--- a/onboarding-validation-keys-generator/main.go
+++ b/onboarding-validation-keys-generator/main.go
@@ -81,7 +81,7 @@ func main() {
 		klog.Exitf("failed to delete public secret: %v", err)
 	}
 
-	err = controllerutil.SetOwnerReference(storageCluster, privateSecret, cl.Scheme())
+	err = controllerutil.SetControllerReference(storageCluster, privateSecret, cl.Scheme())
 	if err != nil {
 		klog.Exitf("failed to set owner reference for private secret: %v", err)
 	}
@@ -95,7 +95,7 @@ func main() {
 		klog.Exitf("failed to create private secret: %v", err)
 	}
 
-	err = controllerutil.SetOwnerReference(storageCluster, publicSecret, cl.Scheme())
+	err = controllerutil.SetControllerReference(storageCluster, publicSecret, cl.Scheme())
 	if err != nil {
 		klog.Exitf("failed to set owner reference for public secret: %v", err)
 	}
--- a/services/ux-backend/handlers/onboarding/clienttokens/handler.go
+++ b/services/ux-backend/handlers/onboarding/clienttokens/handler.go
@@ -14,10 +14,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-const (
-	onboardingPrivateKeyFilePath = "/etc/private-key/key"
-)
-
 var unitToGib = map[string]uint{
 	"Gi": 1,
 	"Ti": 1024,
@@ -37,13 +33,13 @@ func handlePost(w http.ResponseWriter, r *http.Request, tokenLifetimeInHours int
 	var storageQuotaInGiB *uint
 	// When ContentLength is 0 that means request body is empty and
 	// storage quota is unlimited
-	var err error
+
 	if r.ContentLength != 0 {
 		var quota = struct {
 			Value uint   `json:"value"`
 			Unit  string `json:"unit"`
 		}{}
-		if err = json.NewDecoder(r.Body).Decode(&quota); err != nil {
+		if err := json.NewDecoder(r.Body).Decode(&quota); err != nil {
 			http.Error(w, err.Error(), http.StatusBadRequest)
 			return
 		}
@@ -66,7 +62,19 @@ func handlePost(w http.ResponseWriter, r *http.Request, tokenLifetimeInHours int
 		return
 	}
 
-	if onboardingToken, err := util.GenerateClientOnboardingToken(tokenLifetimeInHours, onboardingPrivateKeyFilePath, storageQuotaInGiB, storageCluster.UID); err != nil {
+	klog.Info("Loading onboarding validation private Key")
+	privateKey, err := util.LoadOnboardingValidationPrivateKey(r.Context(), cl, namespace)
+	if err != nil {
+		http.Error(w, fmt.Sprintf("Failed loading onboarding validation private key: %v", err), http.StatusBadRequest)
+		return
+	}
+
+	if onboardingToken, err := util.GenerateClientOnboardingToken(
+		tokenLifetimeInHours,
+		privateKey,
+		storageQuotaInGiB,
+		storageCluster.UID,
+	); err != nil {
 		klog.Errorf("failed to get onboarding token: %v", err)
 		w.WriteHeader(http.StatusInternalServerError)
 		w.Header().Set("Content-Type", handlers.ContentTypeTextPlain)
--- a/services/ux-backend/handlers/onboarding/peertokens/handler.go
+++ b/services/ux-backend/handlers/onboarding/peertokens/handler.go
@@ -11,10 +11,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-const (
-	onboardingPrivateKeyFilePath = "/etc/private-key/key"
-)
-
 func HandleMessage(w http.ResponseWriter, r *http.Request, tokenLifetimeInHours int, cl client.Client, namespace string) {
 	switch r.Method {
 	case "POST":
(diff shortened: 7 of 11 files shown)


controllers/storagecluster/storageclient.go
```diff
@@ -44,7 +44,11 @@ func (s *storageClient) ensureCreated(r *StorageClusterReconciler, storagecluste
 	storageClient.Name = storagecluster.Name
 	_, err := controllerutil.CreateOrUpdate(r.ctx, r.Client, storageClient, func() error {
 		if storageClient.Status.ConsumerID == "" {
-			token, err := util.GenerateClientOnboardingToken(tokenLifetimeInHours, onboardingPrivateKeyFilePath, nil)
+			privateKey, err := util.GetParsedPrivateKey(r.Client, r.OperatorNamespace)
```
rewantsoni, line 47: instead of OperatorNamespace we should be using storageCluster Namespace as private key pair is generated per storageCluster and it is created in the same namespace as the storageCluster

controllers/storagecluster/storageclient.go
```diff
@@ -44,7 +44,11 @@ func (s *storageClient) ensureCreated(r *StorageClusterReconciler, storagecluste
 	storageClient.Name = storagecluster.Name
 	_, err := controllerutil.CreateOrUpdate(r.ctx, r.Client, storageClient, func() error {
 		if storageClient.Status.ConsumerID == "" {
-			token, err := util.GenerateClientOnboardingToken(tokenLifetimeInHours, onboardingPrivateKeyFilePath, nil)
```
rewantsoni, line 47: let's remove onboardingPrivateKeyFilePath if it is not used anymore

controllers/util/provider.go
```diff
@@ -10,17 +11,21 @@ import (
...
 
 	"github.com/google/uuid"
 	"github.com/red-hat-storage/ocs-operator/v4/services"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const onboardingValidationPrivateKeySecretName = "onboarding-private-key"
+
```
rewantsoni, line 24: I think this is also defined in onboarding-ticket-generator, could we make sure that we are not defining at only one place

controllers/util/provider.go
```diff
@@ -93,16 +93,24 @@ func encodeAndSignOnboardingToken(privateKeyPath string, ticket services.Onboard
 	return fmt.Sprintf("%s.%s", encodedPayload, encodedSignature), nil
 }
 
-func readAndDecodePrivateKey(privateKeyPath string) (*rsa.PrivateKey, error) {
-	pemString, err := os.ReadFile(privateKeyPath)
+func GetParsedPrivateKey(cl client.Client, ns string) (*rsa.PrivateKey, error) {
+	klog.Info("Getting the Pem key")
+	ctx := context.Background()
```
rewantsoni, line 98: Let's pass the context from outside

controllers/util/provider.go
```diff
@@ -93,16 +93,24 @@ func encodeAndSignOnboardingToken(privateKeyPath string, ticket services.Onboard
 	return fmt.Sprintf("%s.%s", encodedPayload, encodedSignature), nil
 }
 
-func readAndDecodePrivateKey(privateKeyPath string) (*rsa.PrivateKey, error) {
-	pemString, err := os.ReadFile(privateKeyPath)
+func GetParsedPrivateKey(cl client.Client, ns string) (*rsa.PrivateKey, error) {
+	klog.Info("Getting the Pem key")
```
rewantsoni, line 97: we can log at the call site

services/ux-backend/handlers/common.go
```diff
@@ -1,5 +1,21 @@
...
+func GetPodNamespace() string {
+	if namespace != "" {
+		return namespace
+	}
+	if ns := os.Getenv("OPERATOR_NAMESPACE"); ns != "" {
+		namespace = ns
+		return namespace
+	}
+	panic("Value for env var 'POD_NAMESPACE' is empty")
+}
```
rewantsoni, line 21: This function is already defined in utils let's use that

services/ux-backend/handlers/onboarding/clienttokens/handler.go
```diff
@@ -23,20 +20,29 @@ var unitToGib = map[string]uint{
...
 }
 
-func handlePost(w http.ResponseWriter, r *http.Request, tokenLifetimeInHours int) {
+func handlePost(w http.ResponseWriter, r *http.Request, tokenLifetimeInHours int, cl client.Client) {
 	var storageQuotaInGiB *uint
 	// When ContentLength is 0 that means request body is empty and
 	// storage quota is unlimited
 	var err error
+
+	ns := handlers.GetPodNamespace()
```
rewantsoni, line 38: Let's pass the podNamespace from outside along with the client

services/ux-backend/handlers/onboarding/peertokens/handler.go
```diff
@@ -8,23 +8,29 @@ import (
...
 		handleUnsupportedMethod(w, r)
 	}
 }
 
-func handlePost(w http.ResponseWriter, _ *http.Request, tokenLifetimeInHours int) {
-	if onboardingToken, err := util.GeneratePeerOnboardingToken(tokenLifetimeInHours, onboardingPrivateKeyFilePath); err != nil {
+func handlePost(w http.ResponseWriter, _ *http.Request, tokenLifetimeInHours int, cl client.Client) {
+	var err error
+
+	ns := handlers.GetPodNamespace()
```
rewantsoni, line 26: Let's pass the podNamespace from outside along with the client

services/ux-backend/main.go
```diff
@@ -94,3 +104,25 @@ func main() {
 	log.Fatal(err)
 
 }
+
+func newClient() (client.Client, error) {
+	klog.Info("Setting up k8s client")
+	scheme := runtime.NewScheme()
+	if err := ocsv1.AddToScheme(scheme); err != nil {
+		return nil, err
+	}
```
rewantsoni, line 113: are we fetching any resources created by ocsv1? do we need to add it to the scheme?

tools/csv-merger/csv-merger.go
```diff
@@ -674,6 +670,14 @@ func getUXBackendServerDeployment() appsv1.DeploymentSpec {
 								Name:  "TLS_ENABLED",
 								Value: os.Getenv("TLS_ENABLED"),
 							},
+							{
+								Name: util.OperatorNamespaceEnvVar,
```
rewantsoni, line 674: Use PodNamespaceEnvVar instead of OperatorNamespaceEnvVar

rewantsoni (changes requested):

controllers/storagecluster/storageclient.go
leelavg: I don't think so, could you pls expand more?

rewantsoni: Onboarding job is created in the same [namespace](https://github.com/red-hat-storage/ocs-operator/blob/main/controllers/storagecluster/provider_server.go#L385) as the storageCluster. The job generates the private/pub key pair in the same [namespace](https://github.com/red-hat-storage/ocs-operator/blob/main/onboarding-validation-keys-generator/main.go#L34) as the storageCluster is created. If the storageCluster is created in a different namespace from operator namespace, it will fail at this condition. So it is better if we use storageCluster namespace instead of Operator namespace

mrudraia1: ack, changed to storagecluster namespace

controllers/storagecluster/storageclient.go
mrudraia1: ack

controllers/util/provider.go
mrudraia1: onboardingValidationPrivateKeySecretName is used at two places >
1) provider.go
2) onboarding-validation-keys-generator

controllers/util/provider.go
mrudraia1: ack

controllers/util/provider.go
mrudraia1: ack

services/ux-backend/handlers/common.go
mrudraia1: removed the dublicate code, used fro utils

services/ux-backend/handlers/onboarding/clienttokens/handler.go
mrudraia1: ack

services/ux-backend/handlers/onboarding/peertokens/handler.go
mrudraia1: ack

tools/csv-merger/csv-merger.go
mrudraia1: ack

services/ux-backend/main.go
mrudraia1: client gets failed, if i did not added the ocsv1 schema, the same is followed in onboarding-validation-key-generator main.go

rewantsoni: the onboarding-validation-key-generator requires us to access the storageCluster hence the ocsv1 scheme is added to the client we create there. The ux-backend for now only fetches the secret. I don't think we need the ocsv1 scheme here

mrudraia1: ack, removed the ocsv1 schema

mrudraia1:
I have tested this PR, with the latest 4.18 build. I could see the keys getting exchanged when the rotate signing key is clicked in storageclient page.

services/ux-backend/handlers/onboarding/clienttokens/handler.go
```diff
@@ -23,20 +21,29 @@ var unitToGib = map[string]uint{
...
 }
 
-func handlePost(w http.ResponseWriter, r *http.Request, tokenLifetimeInHours int) {
+func handlePost(w http.ResponseWriter, r *http.Request, tokenLifetimeInHours int, cl client.Client, ns string) {
 	var storageQuotaInGiB *uint
 	// When ContentLength is 0 that means request body is empty and
 	// storage quota is unlimited
 	var err error
+
+	ctx := context.TODO()
```
leelavg, line 39: use context from http.Request.

services/ux-backend/handlers/onboarding/clienttokens/handler.go
```diff
@@ -23,20 +21,29 @@ var unitToGib = map[string]uint{
...
 	case "POST":
-		handlePost(w, r, tokenLifetimeInHours)
+		handlePost(w, r, tokenLifetimeInHours, cl, ns)
 	default:
 		handleUnsupportedMethod(w, r)
 	}
 }
 
-func handlePost(w http.ResponseWriter, r *http.Request, tokenLifetimeInHours int) {
+func handlePost(w http.ResponseWriter, r *http.Request, tokenLifetimeInHours int, cl client.Client, ns string) {
```
leelavg, line 33: other params have types associated with them but not `ns`, pls use full name `namespace`, same in other places as well.

leelavg:
> I have tested this PR, with the latest 4.18 build. I could see the keys getting exchanged when the rotate signing key is clicked in storageclient page.
- ack, only minor comments now. @nb-ohad possible for a final review?

controllers/storagecluster/storageclient.go
```diff
@@ -42,9 +41,14 @@ func (s *storageClient) ensureCreated(r *StorageClusterReconciler, storagecluste
 
 	storageClient := &ocsclientv1a1.StorageClient{}
 	storageClient.Name = storagecluster.Name
+	storageClusterNamespace := storagecluster.Namespace
```
nb-ohad, line 44: Why are we assuming storage cluster namespace and not operator namespace or pod namespace?

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -225,7 +225,7 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&appsv1.Deployment{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
 		Owns(&corev1.Service{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
 		Owns(&corev1.ConfigMap{}, builder.MatchEveryOwner, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
-		Owns(&corev1.Secret{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		Owns(&corev1.Secret{}, builder.MatchEveryOwner, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
```
nb-ohad, line 228: Why do we need a MatchEveryOwner here? Isn't StorageCluster the controller owner of the secret we want to watch?

controllers/util/provider.go
```diff
@@ -10,17 +11,20 @@ import (
...
-	"os"
 	"time"
 
 	"github.com/google/uuid"
 	"github.com/red-hat-storage/ocs-operator/v4/services"
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const onboardingValidationPrivateKeySecretName = "onboarding-private-key"
```
nb-ohad, line 22: There is no need for a const if it is only used once and is local for this file.

controllers/util/provider.go
```diff
@@ -93,16 +92,22 @@ func encodeAndSignOnboardingToken(privateKeyPath string, ticket services.Onboard
...
+	privateSecret.Namespace = ns
+
+	err := cl.Get(ctx, client.ObjectKeyFromObject(privateSecret), privateSecret)
 	if err != nil {
-		return nil, fmt.Errorf("failed to read private key: %v", err)
+		return nil, fmt.Errorf("failed to get private secret: %v", err)
 	}
 
-	Block, _ := pem.Decode(pemString)
+	Block, _ := pem.Decode(privateSecret.Data["key"])
```
nb-ohad, line 105: Don't assume the secret has a data section or the correct map key. You need to check that first and return an error or else you will send a nil value to decode.

Also, why are we ignoring decode errors?

controllers/util/provider.go
```diff
@@ -93,16 +92,22 @@ func encodeAndSignOnboardingToken(privateKeyPath string, ticket services.Onboard
 	return fmt.Sprintf("%s.%s", encodedPayload, encodedSignature), nil
 }
 
-func readAndDecodePrivateKey(privateKeyPath string) (*rsa.PrivateKey, error) {
-	pemString, err := os.ReadFile(privateKeyPath)
+func GetParsedPrivateKey(ctx context.Context, cl client.Client, ns string) (*rsa.PrivateKey, error) {
```
nb-ohad, line 95: "Private Key" is a very generic name. When talking about the entire code base, mentioning "private key" does not help identify what key we are talking about. Please rename it to `LoadOnboardingValidationPrivateKey`

services/ux-backend/handlers/onboarding/clienttokens/handler.go
nb-ohad: @mrudraia1 Please address the above

services/ux-backend/handlers/onboarding/clienttokens/handler.go
nb-ohad: @mrudraia1 Please address the above

services/ux-backend/handlers/onboarding/clienttokens/handler.go
```diff
@@ -23,20 +21,29 @@ var unitToGib = map[string]uint{
...
 	var err error
+
+	ctx := context.TODO()
+	privateKey, err := util.GetParsedPrivateKey(ctx, cl, ns)
+	klog.Info("Getting the Pem key")
+	if err != nil {
+		http.Error(w, fmt.Sprintf("Failed to get private key: %v", err), http.StatusBadRequest)
+		return
+	}
+
```
nb-ohad, line 46: Please load he key only after finishing with all of the checks. 
As always it is preferable to run costly load/get operation as close as possible to the code that uses the result

services/ux-backend/main.go
```diff
@@ -94,3 +106,22 @@ func main() {
 	log.Fatal(err)
 
 }
+
+func newClient() (client.Client, error) {
```
nb-ohad, line 110: Please be more specific with your function names
```suggestion
func newK8sClient() (client.Client, error) {
```

services/ux-backend/main.go
```diff
@@ -94,3 +106,22 @@ func main() {
...
+	if err != nil {
+		return nil, err
+	}
+	k8sClient, err := client.New(config, client.Options{Scheme: scheme})
+	if err != nil {
+		return nil, err
+	}
+
+	return k8sClient, nil
+}
```
nb-ohad, line 127: I believe we already have other places that create a new client. 
Please reuse the same code instead of rolling out your own

services/ux-backend/main.go
```diff
@@ -63,18 +68,25 @@ func main() {
 		os.Exit(-1)
 	}
 
+	cl, err := newClient()
+	if err != nil {
+		klog.Exitf("failed to create client: %v", err)
+	}
+
+	ns := util.GetPodNamespace()
```
nb-ohad, line 76: This contradicts other changes made in some other files. 
Sometimes you assume the secret is in the pod namespace and sometimes you assume it is in the storage cluster namespace (see prev review comment), we need to be consistent

services/ux-backend/main.go
nb-ohad, line 110: In this file, all private function goes above main, not below. 
Let's stay consistent

nb-ohad (changes requested):

controllers/storagecluster/storageclient.go
mrudraia1: https://github.com/red-hat-storage/ocs-operator/pull/2827#discussion_r1782099611

services/ux-backend/main.go
mrudraia1: The newClient() function is used in onboarding-validation-keys-generator. I have re-used the same code, not modified anything.

services/ux-backend/main.go
mrudraia1: ack

mrudraia1: ack

services/ux-backend/handlers/onboarding/clienttokens/handler.go
mrudraia1: ack

services/ux-backend/handlers/onboarding/clienttokens/handler.go
mrudraia1: ack

services/ux-backend/handlers/onboarding/clienttokens/handler.go
mrudraia1: ack

controllers/storagecluster/storageclient.go
mrudraia1: changed to operator namespace , ack

controllers/storagecluster/storagecluster_controller.go
mrudraia1: controller owner reference set to secrets, ack

services/ux-backend/main.go
mrudraia1: ack, using the operator-namespace in storageclient.go

controllers/util/provider.go
mrudraia1: renamed the function with suggested name

controllers/util/provider.go
mrudraia1: The onboardingValidationPrivateKeySecretName is used at 2 places
1) onboarding-validation-key-generator 
2) provider.go

So instead of variable it is used as const.

deepsm007:
infra issue
/test images

controllers/storagecluster/storagecluster_controller.go
nb-ohad: This comment was not addressed as tall!  You are matching by controller ownership, it does not make sense to match every owner

controllers/util/provider.go
```diff
@@ -97,16 +96,26 @@ func encodeAndSignOnboardingToken(privateKeyPath string, ticket services.Onboard
...
 	if err != nil {
-		return nil, fmt.Errorf("failed to read private key: %v", err)
+		return nil, fmt.Errorf("failed to get private secret: %v", err)
 	}
 
-	Block, _ := pem.Decode(pemString)
-	privateKey, err := x509.ParsePKCS1PrivateKey(Block.Bytes)
-	if err != nil {
-		return nil, fmt.Errorf("failed to parse private key: %v", err)
+	if privateSecret.Data != nil {
```
nb-ohad, line 109: You don't need this check: Maps in golang, even when not initialized, support the key access operator. 
```suggestion
```

controllers/util/provider.go
```diff
@@ -97,16 +96,26 @@ func encodeAndSignOnboardingToken(privateKeyPath string, ticket services.Onboard
...
+			privateKey, err := x509.ParsePKCS1PrivateKey(Block.Bytes)
+			if err != nil {
+				return nil, fmt.Errorf("failed to parse private key: %v", err)
+			}
+			return privateKey, nil
+		}
 	}
-	return privateKey, nil
+
+	return nil, fmt.Errorf("No data found in secret")
```
nb-ohad, line 120: I would prefer, from code readability and extensibility concerns, if we first handle error conditions and only at the end have the return that is not erroneous. This way you don't need nested returns and extending the function in the future with more checks become easier 

```suggestion
		privateSecretKey, ok := privateSecret.Data["key"]; 
		if !ok {
			return nil, fmt.Errorf("No data found in secret")
		}

		Block, _ := pem.Decode(privateSecretKey)
		privateKey, err := x509.ParsePKCS1PrivateKey(Block.Bytes)
		if err != nil {
			return nil, fmt.Errorf("failed to parse private key: %v", err)
		}

		return privateKey, nil

```

services/ux-backend/handlers/onboarding/clienttokens/handler.go
```diff
@@ -66,7 +63,14 @@ func handlePost(w http.ResponseWriter, r *http.Request, tokenLifetimeInHours int
...
 
-	if onboardingToken, err := util.GenerateClientOnboardingToken(tokenLifetimeInHours, onboardingPrivateKeyFilePath, storageQuotaInGiB, storageCluster.UID); err != nil {
+	privateKey, err := util.LoadOnboardingValidationPrivateKey(ctx, cl, namespace)
+	klog.Info("Getting the Pem key")
+	if err != nil {
+		http.Error(w, fmt.Sprintf("Failed to get private key: %v", err), http.StatusBadRequest)
+		return
+	}
+
+	if onboardingToken, err := util.GenerateClientOnboardingToken(tokenLifetimeInHours, privateKey, storageQuotaInGiB, storageCluster.UID); err != nil {
```
nb-ohad, line 73: This line is too long, please split over multi-lines
```suggestion
	if onboardingToken, err := util.GenerateClientOnboardingToken(
		tokenLifetimeInHours, 
		privateKey, 
		storageQuotaInGiB, 
		storageCluster.UID,
	); err != nil {
```

services/ux-backend/handlers/onboarding/peertokens/handler.go
```diff
@@ -25,14 +22,21 @@ func HandleMessage(w http.ResponseWriter, r *http.Request, tokenLifetimeInHours
...
-	if onboardingToken, err := util.GeneratePeerOnboardingToken(tokenLifetimeInHours, onboardingPrivateKeyFilePath, storageCluster.UID); err != nil {
+	ctx := context.TODO()
+	privateKey, err := util.LoadOnboardingValidationPrivateKey(ctx, cl, namespace)
+	klog.Info("Getting the Pem key")
+	if err != nil {
+		http.Error(w, fmt.Sprintf("Failed to get private key: %v", err), http.StatusBadRequest)
+		return
+	}
+
+	if onboardingToken, err := util.GeneratePeerOnboardingToken(tokenLifetimeInHours, privateKey, storageCluster.UID); err != nil {
```
nb-ohad, line 39: Line too long, see last comment for breaking into multi lines

services/ux-backend/handlers/onboarding/peertokens/handler.go
```diff
@@ -25,14 +22,21 @@ func HandleMessage(w http.ResponseWriter, r *http.Request, tokenLifetimeInHours
...
 func handlePost(w http.ResponseWriter, r *http.Request, tokenLifetimeInHours int, cl client.Client, namespace string) {
-
 	storageCluster, err := util.GetStorageClusterInNamespace(r.Context(), cl, namespace)
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
 
-	if onboardingToken, err := util.GeneratePeerOnboardingToken(tokenLifetimeInHours, onboardingPrivateKeyFilePath, storageCluster.UID); err != nil {
+	ctx := context.TODO()
```
nb-ohad, line 31: 1. You should not use a TODO context but a background context
2. If there is a need to create a context for handling a request it should be created at the topmost level of the request handling and passed down

services/ux-backend/main.go
```diff
@@ -87,10 +87,10 @@ func main() {
 		// Set the Deprecation header
 		w.Header().Set("Deprecation", "true") // Standard "Deprecation" header
 		w.Header().Set("Link", "/onboarding/client-tokens; rel=\"alternate\"")
-		clienttokens.HandleMessage(w, r, config.tokenLifetimeInHours, cl, namespace)
+		clienttokens.HandleMessage(r.Context(), w, r, config.tokenLifetimeInHours, cl, namespace)
```
nb-ohad, line 90: You should not use the same context for main and for specific request handling. Each request should initialize its own background context that will be used during the entire request.

services/ux-backend/main.go
```diff
@@ -87,10 +87,10 @@ func main() {
...
 	})
 	http.HandleFunc("/onboarding/client-tokens", func(w http.ResponseWriter, r *http.Request) {
-		clienttokens.HandleMessage(w, r, config.tokenLifetimeInHours, cl, namespace)
+		clienttokens.HandleMessage(r.Context(), w, r, config.tokenLifetimeInHours, cl, namespace)
```
nb-ohad, line 93: Same as last comment

nb-ohad (changes requested):

controllers/storagecluster/storagecluster_controller.go
mrudraia1: Owner reference were set during creation of secrets in on-boarding job, when the rotate signing key is clicked the re-consiliation is not happening, The onboarding-token needs to be created which is failing. 

Any suggestion for this issue.

services/ux-backend/handlers/onboarding/clienttokens/handler.go
```diff
@@ -66,7 +63,19 @@ func handlePost(w http.ResponseWriter, r *http.Request, tokenLifetimeInHours int
 		return
 	}
 
-	if onboardingToken, err := util.GenerateClientOnboardingToken(tokenLifetimeInHours, onboardingPrivateKeyFilePath, storageQuotaInGiB, storageCluster.UID); err != nil {
+	privateKey, err := util.LoadOnboardingValidationPrivateKey(ctx, cl, namespace)
+	klog.Info("Getting the Pem key")
+	if err != nil {
+		http.Error(w, fmt.Sprintf("Failed to get private key: %v", err), http.StatusBadRequest)
+		return
+	}
```
nb-ohad, line 71: 1. It does not help to log after the action, you should log before the action 
2. Why should this context even know that we are using a `pem` key?
3. The error message should be more descriptive and needs to give more context

```suggestion
	klog.Info("Loading onboarding validation private Key")
	if err := privateKey, err := util.LoadOnboardingValidationPrivateKey(ctx, cl, namespace); err != nil {
		http.Error(w, fmt.Sprintf("Failed loading onboarding validation private: %v", err), http.StatusBadRequest)
		return
	}
```

services/ux-backend/handlers/onboarding/peertokens/handler.go
```diff
@@ -25,14 +22,25 @@ func HandleMessage(w http.ResponseWriter, r *http.Request, tokenLifetimeInHours
 }
 
 func handlePost(w http.ResponseWriter, r *http.Request, tokenLifetimeInHours int, cl client.Client, namespace string) {
-
+	ctx := context.Background()
```
nb-ohad, line 25: Similar to the client token handler, context need to be passed form the outside

nb-ohad (changes requested):

services/ux-backend/main.go
```diff
@@ -94,15 +96,18 @@ func main() {
...
 	})
 	http.HandleFunc("/onboarding/client-tokens", func(w http.ResponseWriter, r *http.Request) {
-		clienttokens.HandleMessage(w, r, config.tokenLifetimeInHours, cl, namespace)
+		ctx := context.Background()
+		clienttokens.HandleMessage(ctx, w, r, config.tokenLifetimeInHours, cl, namespace)
 	})
 	http.HandleFunc("/onboarding/peer-tokens", func(w http.ResponseWriter, r *http.Request) {
-		peertokens.HandleMessage(w, r, config.tokenLifetimeInHours, cl, namespace)
+		ctx := context.Background()
+		peertokens.HandleMessage(ctx, w, r, config.tokenLifetimeInHours, cl, namespace)
```
leelavg, line 110: as already mentioned in one of earlier comments, http.Request already provides a context and use that inside the handler but not create the ctx everytime as being done here.

nb-ohad: @leelavg @mrudraia1 
Let me clarify. Evey request should handle its own context. If the req itself provide a context we should use that, if not we should create a new context at the topmost level of the request and pass it down the call chain

leelavg: > Every request should handle its own context.
- yes, that's what I mentioned earlier in another comment.

mrudraia1: Do, I need to change this to original like

peertokens.HandleMessage( r.context,  .......)

nb-ohad: ok

controllers/storagecluster/storagecluster_controller.go
nb-ohad: If there is a controller owner ref on the secret and then someone deletes that secret then a delete event will be sent here. And because the item had an owner ref (with controller set) then a reconcile request will be queued. Now if that is not happening there can be a bug in a couple of places:
- The predicate might be the one the drops the event (should not, but might be)
- The reconcile is initiated but the reconcile code might not re-create the secret (might happen because of stale cache)
- Are we sure we are adding a controller reference on the secret? if not and we are only adding an owner ref then that is your bug

@mrudraia1 You need to identify the issue, and only after you fully understand what is going on you can suggest a fix.

rewantsoni: StorageCluster is the owner of the secrets, not the controller owner [here](https://github.com/red-hat-storage/ocs-operator/blob/main/onboarding-validation-keys-generator/main.go#L84)

controllers/util/provider.go
```diff
@@ -97,16 +96,30 @@ func encodeAndSignOnboardingToken(privateKeyPath string, ticket services.Onboard
...
+	}
+
+	privateSecretKey := privateSecret.Data["key"]
+	if len(privateSecretKey) == 0 {
+		return nil, fmt.Errorf("No data found in secret")
+	}
+
+	Block, err1 := pem.Decode(privateSecretKey)
+	if err1 != nil {
+		return nil, fmt.Errorf("Failed to decode private key: %v", err1)
```
leelavg, line 116: lower case `Block` and as per `pem.Decode` func signature the second return value isn't err but a slice of bytes which should be empty.
```suggestion
	block, rest := pem.Decode(privateSecretKey)
	if len(rest) > 0 {
		return nil, fmt.Errorf("PEM block not found in private key: %s", rest)
```

services/ux-backend/handlers/onboarding/clienttokens/handler.go
```diff
@@ -66,7 +62,19 @@ func handlePost(w http.ResponseWriter, r *http.Request, tokenLifetimeInHours int
 		return
 	}
 
-	if onboardingToken, err := util.GenerateClientOnboardingToken(tokenLifetimeInHours, onboardingPrivateKeyFilePath, storageQuotaInGiB, storageCluster.UID); err != nil {
+	klog.Info("Loading onboarding validation private Key")
+	privateKey, err := util.LoadOnboardingValidationPrivateKey(r.Context(), cl, namespace)
+	if err != nil {
+		http.Error(w, fmt.Sprintf("Failed loading onboarding validation private: %v", err), http.StatusBadRequest)
```
leelavg, line 68:
```suggestion
		http.Error(w, fmt.Sprintf("Failed loading onboarding validation private key: %v", err), http.StatusBadRequest)
```

services/ux-backend/handlers/onboarding/peertokens/handler.go
```diff
@@ -25,14 +21,24 @@ func HandleMessage(w http.ResponseWriter, r *http.Request, tokenLifetimeInHours
...
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
 
-	if onboardingToken, err := util.GeneratePeerOnboardingToken(tokenLifetimeInHours, onboardingPrivateKeyFilePath, storageCluster.UID); err != nil {
+	klog.Info("Loading onboarding validation private Key")
+	privateKey, err := util.LoadOnboardingValidationPrivateKey(r.Context(), cl, namespace)
+	if err != nil {
+		http.Error(w, fmt.Sprintf("Failed loading onboarding validation private: %v", err), http.StatusBadRequest)
```
leelavg, line 33:
```suggestion
		http.Error(w, fmt.Sprintf("Failed loading onboarding validation private key: %v", err), http.StatusBadRequest)
```

leelavg (commented):
pls run `make deps-update` to update vendor.

controllers/storagecluster/storagecluster_controller.go
leelavg: > Owner reference were set during creation of secrets in on-boarding job, when the rotate signing key is clicked the re-consiliation is not happening, The onboarding-token needs to be created which is failing.
- ouch, so the issue all along is storagecluster not getting a reconcile event when rotate keys is invoked (ie, deletion of secrets) but not the delay b/n rotation and kubelet updating the secret as diagnosed in https://bugzilla.redhat.com/show_bug.cgi?id=2311012#c5?

@rchikatw was there a discussion w/ @mrudraia1 along these lines when he started to work on this? @mrudraia1 did you observe the same behavior from when the work started on this PR, end of July and in latest revision I don't see this change which basically means we are just refactoring the code w/o any observable behavior and not fixing the issue.

services/ux-backend/handlers/onboarding/clienttokens/handler.go
mrudraia1: Ack

services/ux-backend/handlers/onboarding/peertokens/handler.go
mrudraia1: Ack

controllers/util/provider.go
mrudraia1: pem.Decode  returns byte, changed as suggested by Leela

leelavg: I also mentioned to lowercase `Block` as it doesn't have any extra significance.

mrudraia1: ack

controllers/util/provider.go
```diff
@@ -97,16 +96,26 @@ func encodeAndSignOnboardingToken(privateKeyPath string, ticket services.Onboard
...
+		return nil, fmt.Errorf("failed to get private secret: %v", err)
+	}
+
+	privateSecretKey, ok := privateSecret.Data["key"]
+	if !ok {
+		return nil, fmt.Errorf("No data found in secret")
 	}
 
-	Block, _ := pem.Decode(pemString)
+	Block, _ := pem.Decode(privateSecretKey)
```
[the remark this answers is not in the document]
mrudraia1: Ack

controllers/storagecluster/storagecluster_controller.go
mrudraia1: @leelavg earlier secret keys are read from the volume mounts , which is now changed to read directly by reading the secrets. 

OwnerReference of the secrets changed to ControllerReference.

onboarding-validation-keys-generator/main.go
```diff
@@ -81,7 +81,7 @@ func main() {
 		klog.Exitf("failed to delete public secret: %v", err)
 	}
 
-	err = controllerutil.SetOwnerReference(storageCluster, privateSecret, cl.Scheme())
+	err = controllerutil.SetControllerReference(storageCluster, privateSecret, cl.Scheme())
```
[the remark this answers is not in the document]
mrudraia1: Added the permissions for ControllerReference

controllers/storagecluster/storagecluster_controller.go
mrudraia1: yes, know with this PR build I tested, I can observe the 
1) creation of  new on-boarding_token immediately after the clicking of Rotate signing key.
2) earlier I was using MatchEveryOwner to watch the secrets, Know the ControllerOwner is set to watch the resources.

deploy/ocs-operator/manifests/onboarding-validation-keys-generator-role.yaml
leelavg, line 1: Proposed rbac will provide update access to storagecluster which isn't good, use a new section for only update on finalizers and the job doesn't has controller and watch isn't required.

mrudraia1: Ack

controllers/storagecluster/storagecluster_controller.go
rchikatw: I am not aware of any discussion made to change the predicate, And not sure why we are changing this OwnerRef -> ControllerRef: [Here](https://github.com/red-hat-storage/ocs-operator/pull/2827/files#diff-5c4a3e6001667acda632743674c8d1463952f5d5d7d35a23c60db6ed8340dd0aR84) And because of this change you might be seeing job creation is not happening when keys are rotated this is my just a guess. AFAIK I know a job is created when keys are rotated and even QE team tested this.

onboarding-validation-keys-generator/main.go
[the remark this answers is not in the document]
rchikatw: I believe Rewant mentioned adding additional permissions, not removing existing ones, which suggests Owner and Controller are needed. @rewantsoni am I right here?

deploy/ocs-operator/manifests/onboarding-validation-keys-generator-role.yaml
```diff
@@ -19,3 +19,10 @@ rules:
     verbs:
       - get
       - list
+  - apiGroups:
+      - ocs.openshift.io
+    resources:
+      - storageclusters/finalizers
+    verbs:
+      - update
```
nb-ohad, line 27: Why does the validator service account need to be able to update the finalizer of the storage cluster?

onboarding-validation-keys-generator/main.go
nb-ohad, line 84: Having the token generator add a controller ref for storage cluster is an antipattern. Why are we doing this? What prevents us having code in the storage cluster controller that take ownership on these secrets?

onboarding-validation-keys-generator/main.go
```diff
@@ -95,7 +95,7 @@ func main() {
 		klog.Exitf("failed to create private secret: %v", err)
 	}
 
-	err = controllerutil.SetOwnerReference(storageCluster, publicSecret, cl.Scheme())
+	err = controllerutil.SetControllerReference(storageCluster, publicSecret, cl.Scheme())
```
nb-ohad, line 98: Same as last comment

metrics/vendor/github.com/red-hat-storage/ocs-operator/v4/controllers/util/provider.go
nb-ohad, line 1: How come we now have a need to copy this file as well. This is supposed to be an implementation util file not an api backing file

nb-ohad (changes requested):
I believe you are missing code in the storage cluster controller to watch the secrets

deploy/ocs-operator/manifests/onboarding-validation-keys-generator-role.yaml
mrudraia1: The following error was observed 

Failed to create private secret: secrets "onboarding-private-key" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: , <nil>

so added update permission for the resource storageclusters/finalizers .

PR #2832, opened 2024-10-02
use a default backing store for noobaa acount

leelavg:
noobaa account creation requires a default backing store to be specified which will be used to create new buckets for connected clients.


leelavg:
as this is a blocker for current release we are hardcoding the name of the default backing store which is tested manually and solves the issue. This has to be replaced w/ runtime finding of name.

leelavg:
/cherrypick release-4.17
https://bugzilla.redhat.com/show_bug.cgi?id=2314200

rewantsoni:
@leelavg shouldn't we also watch the Noobaa account [here](https://github.com/red-hat-storage/ocs-operator/blob/main/controllers/storageconsumer/storageconsumer_controller.go#L286-L296)?

iamniting (approved):

PR #2837, opened 2024-10-03
Include Encryption In-Transit state as part of the desired state hash sent to clients

malayparida2000:
When in-transit encryption is enabled/disabled the kernel mount option for cephFS needs to be updated between prefer-crc/secure. So the desired state hash needs to include the EiT state, so that if the EiT state is changed the desired state hash will change and the client will reconcile to get the updated mount option.

### DIFF
--- a/services/provider/server/server.go
+++ b/services/provider/server/server.go
@@ -11,13 +11,14 @@ import (
 	"encoding/json"
 	"encoding/pem"
 	"fmt"
-	"k8s.io/utils/ptr"
 	"math"
 	"net"
 	"slices"
 	"strconv"
 	"strings"
 	"time"
+
+	"k8s.io/utils/ptr"
 
 	"github.com/blang/semver/v4"
 	nbv1 "github.com/noobaa/noobaa-operator/v5/pkg/apis/noobaa/v1alpha1"
@@ -193,7 +194,17 @@ func (s *OCSProviderServer) GetStorageConfig(ctx context.Context, req *pb.Storag
 		if err != nil {
 			return nil, status.Errorf(codes.Internal, "Failed to construct status response: %v", err)
 		}
-		desiredClientConfigHash := getDesiredClientConfigHash(channelName, consumerObj)
+
+		storageCluster, err := s.getStorageCluster(ctx)
+		if err != nil {
+			return nil, err
+		}
+
+		desiredClientConfigHash := getDesiredClientConfigHash(
+			channelName,
+			consumerObj,
+			isEncryptionInTransitEnabled(storageCluster.Spec.Network),
+		)
 
 		klog.Infof("successfully returned the config details to the consumer.")
 		return &pb.StorageConfigResponse{
@@ -751,15 +762,12 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
 				"csi.storage.k8s.io/controller-expand-secret-name": provisionerSecretName,
 			}
 
-			storageClusters := &ocsv1.StorageClusterList{}
-			if err := s.client.List(ctx, storageClusters, client.InNamespace(s.namespace), client.Limit(2)); err != nil {
-				return nil, status.Errorf(codes.Internal, "failed to get storage cluster: %v", err)
-			}
-			if len(storageClusters.Items) != 1 {
-				return nil, status.Errorf(codes.Internal, "expecting one single storagecluster to exist")
+			storageCluster, err := s.getStorageCluster(ctx)
+			if err != nil {
+				return nil, err
 			}
 			var kernelMountOptions map[string]string
-			for _, option := range strings.Split(util.GetCephFSKernelMountOptions(&storageClusters.Items[0]), ",") {
+			for _, option := range strings.Split(util.GetCephFSKernelMountOptions(storageCluster), ",") {
 				if kernelMountOptions == nil {
 					kernelMountOptions = map[string]string{}
 				}
@@ -847,7 +855,16 @@ func (s *OCSProviderServer) ReportStatus(ctx context.Context, req *pb.ReportStat
 		return nil, status.Errorf(codes.Internal, "Failed to construct status response: %v", err)
 	}
 
-	desiredClientConfigHash := getDesiredClientConfigHash(channelName, storageConsumer)
+	storageCluster, err := s.getStorageCluster(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	desiredClientConfigHash := getDesiredClientConfigHash(
+		channelName,
+		storageConsumer,
+		isEncryptionInTransitEnabled(storageCluster.Spec.Network),
+	)
 
 	return &pb.ReportStatusResponse{
 		DesiredClientOperatorChannel: channelName,
@@ -855,10 +872,11 @@ func (s *OCSProviderServer) ReportStatus(ctx context.Context, req *pb.ReportStat
 	}, nil
 }
 
-func getDesiredClientConfigHash(channelName string, storageConsumer *ocsv1alpha1.StorageConsumer) string {
+func getDesiredClientConfigHash(channelName string, storageConsumer *ocsv1alpha1.StorageConsumer, encryptionInTransit bool) string {
 	var arr = []any{
 		channelName,
 		storageConsumer.Spec.StorageQuotaInGiB,
+		encryptionInTransit,
 	}
 	return util.CalculateMD5Hash(arr)
 }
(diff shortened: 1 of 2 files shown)


malayparida2000:
Should go after https://github.com/red-hat-storage/ocs-operator/pull/2793

services/provider/server/server.go
```diff
@@ -870,17 +879,29 @@ func (s *OCSProviderServer) ReportStatus(ctx context.Context, req *pb.ReportStat
...
+	storageCluster, err := s.getStorageCluster(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	inTransitEncryptionEnabled := false
+	if storageCluster.Spec.Network != nil && storageCluster.Spec.Network.Connections != nil &&
+		storageCluster.Spec.Network.Connections.Encryption != nil && storageCluster.Spec.Network.Connections.Encryption.Enabled {
+		inTransitEncryptionEnabled = true
+	}
```
nb-ohad, line 891: This pattern repeats itself. I would suggest to replace the `getStorageCluster` utils with a `IsEncryptionInTransitEnabled()` function and call it in both places

services/provider/server/server.go
```diff
@@ -901,6 +922,28 @@ func (s *OCSProviderServer) getOCSSubscriptionChannel(ctx context.Context) (stri
...
+		}
+	}
+
+	if len(storageClusters) > 1 {
+		return nil, status.Errorf(codes.Internal, "found more than one storage cluster")
+	}
+	if len(storageClusters) == 0 {
+		return nil, status.Errorf(codes.NotFound, "no storage cluster found")
+	}
+	return &storageClusters[0], nil
```
nb-ohad, line 944: 1. Don't use the `range` value capture, it will create an in-memory copy of the storage cluster - It is a very big structure

2. There is no need to create a new slice and start appending to it

```suggestion
	var found *ocsv1.StorageCluster = nil
	for i := range storageClusterList.Items {
		cluster := &storageClusterList.Items[i]
		if cluster.Spec.ExternalStorage.Enable {
			continue
		} 
		if found != nil {
			return nil, status.Errorf(codes.Internal, "found more than one storage cluster")
		}
		found = cluster
	}
	if found == nil {
		return nil, status.Errorf(codes.NotFound, "no storage cluster found")
	}
	return found
```

services/provider/server/server.go
```diff
@@ -870,17 +879,29 @@ func (s *OCSProviderServer) ReportStatus(ctx context.Context, req *pb.ReportStat
...
 	}, nil
 }
 
-func getDesiredClientConfigHash(channelName string, storageConsumer *ocsv1alpha1.StorageConsumer) string {
+func getDesiredClientConfigHash(channelName string, inTransitEncryptionEnabled bool, storageConsumer *ocsv1alpha1.StorageConsumer) string {
 	var arr = []any{
 		channelName,
+		inTransitEncryptionEnabled,
 		storageConsumer.Spec.StorageQuotaInGiB,
 	}
```
nb-ohad, line 906: Please add the new var to the end of the list
```suggestion
	var arr = []any{
		channelName,
		storageConsumer.Spec.StorageQuotaInGiB,
		inTransitEncryptionEnabled,
	}
```

nb-ohad (changes requested):

services/provider/server/server.go
malayparida2000: Kept the getStorageCluster func as it's being used at another place too, created a separate isEncryptionInTransitEnabled() taking only the networkSpec.

services/provider/server/server.go
malayparida2000: Thanks for the detailed skeleton of the function, I have updated the function accordingly.

services/provider/server/server_test.go
```diff
@@ -275,6 +275,11 @@ func TestGetExternalResources(t *testing.T) {
 	ocsSubscription.Spec = ocsSubscriptionSpec
 	assert.NoError(t, client.Create(ctx, ocsSubscription))
 
+	storageCluster := &ocsv1.StorageCluster{}
+	storageCluster.Name = "test-storagecluster"
+	storageCluster.Namespace = serverNamespace
+	assert.NoError(t, client.Create(ctx, storageCluster))
```
rohan47, line 281: Is this the test for encryption in transit is enabled?

malayparida2000: No this just creates a storagecluster, as I am adding a util function in our code path which expects a storagecluster otherwise it will return an error. This is just to prevent that.

nb-ohad:
@malayparida2000 Please let the reviewers resolve review comments threads. 
A reviewer must validate the changes made after you fix/change anything. We usually use the resolve state of a thread as confirmation that the change was made or the conversation resulted in a resolution that satisfies the reviewer, so the correct person to resolve a review comment thread is the reviewer after confirmation

services/provider/server/server.go
```diff
@@ -847,18 +852,24 @@ func (s *OCSProviderServer) ReportStatus(ctx context.Context, req *pb.ReportStat
...
+	desiredClientConfigHash := getDesiredClientConfigHash(channelName, storageConsumer, isEncryptionInTransitEnabled(storageCluster.Spec.Network))
 
 	return &pb.ReportStatusResponse{
 		DesiredClientOperatorChannel: channelName,
 		DesiredConfigHash:            desiredClientConfigHash,
 	}, nil
 }
 
-func getDesiredClientConfigHash(channelName string, storageConsumer *ocsv1alpha1.StorageConsumer) string {
+func getDesiredClientConfigHash(channelName string, storageConsumer *ocsv1alpha1.StorageConsumer, inTransitEncryptionEnabled bool) string {
```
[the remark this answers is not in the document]
malayparida2000: Have addressed all the conversations along with this one, PTAL.

malayparida2000:
/cc @leelavg 
Since Ohad will be on PTO until the 21st of October, could you please take up the responsibility for reviewing this PR? It's crucial that we have sufficient time after merging these changes to conduct end-to-end testing for the epic, identify any faults, and fix them. The current delay in the review process is causing us to lose valuable time. Additionally, I need to document the end-to-end workflow for all three modes, so your help with the review & merging the changes would be greatly appreciated.

services/provider/server/server.go
```diff
@@ -878,6 +895,36 @@ func (s *OCSProviderServer) getOCSSubscriptionChannel(ctx context.Context) (stri
 	return subscription.Spec.Channel, nil
 }
 
+func (s *OCSProviderServer) getStorageCluster(ctx context.Context) (*ocsv1.StorageCluster, error) {
```
leelavg, line 898: the implementation seems to be a roundabout way to me, in a namespace there'll only be 1 storagecluster isn't it? you could limit the list by 2 and anything that isn't 1 is an error scenario.

moreover from 4.18 api-server is always deployed and checking storagecluster for external enabled, the condition is implicitly missing out internal only scenario and I suggest storagecluster w/ allowremotestorageclusters atm is more suitable for the condition here.

leelavg:
> help with the review & merging
- I can only review but don't have merge rights

> 21st of October
- can drop a note to review on 20th as well as it's working day for Ohad

malayparida2000:
> I can only review but don't have merge rights
Right, But if you ack LGTM, I can get it merged.

> can drop a note to review on 20th as well as it's working day for Ohad
He told he is on PTO till 21st, so he should be back on 22nd only.

services/provider/server/server.go
```diff
@@ -193,7 +193,17 @@ func (s *OCSProviderServer) GetStorageConfig(ctx context.Context, req *pb.Storag
 		if err != nil {
 			return nil, status.Errorf(codes.Internal, "Failed to construct status response: %v", err)
 		}
-		desiredClientConfigHash := getDesiredClientConfigHash(channelName, consumerObj)
+
+		storageCluster, err := s.getStorageCluster(ctx)
+		if err != nil {
+			return nil, err
```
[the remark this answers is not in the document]
malayparida2000: The error the getStorageCluster is returning is already formated in gRPC conventions so we can directly return it here instead of reformating it again.

services/provider/server/server.go
malayparida2000: Acc to how the current implimenetation is if there are multiple storageClusters, only one of them is handled, the others are just moved into Phase: Ignored. So I don't think erroring out ACC to the lits's size would be a good indicator.  

Agree, specifically looking for cluster with allowremotestorageclusters seems to be the best filter.

malayparida2000: Updated the func

leelavg: > into Phase: Ignored
- didn't know that, thanks!

> I don't think erroring out ACC to the lits's size would be a good choice.
- fair enough, then you should also be filtering out storagecluster that is set w/ ignored phase isn't it?

services/provider/server/server.go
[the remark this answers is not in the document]
leelavg: ack, resolve comment.

services/provider/server/server.go
malayparida2000: Right, Added a check for that.

iamniting (approved):
looks good to me
/hold only for @leelavg as @nb-ohad is not available

malayparida2000:
/unhold
Resolving the hold as Leela is also on PTO, And this is required for testing in next week's build, Any follow on suggestions if there can be resolved with follow on PR.

Issue #2829, opened 2024-09-30
Remove healthchecker cephclient from storageconsumer CR status


ezio-auditore:
With the removal of the healthchecker cephclient from being created in https://github.com/red-hat-storage/ocs-operator/pull/2828 the same should be reflected in the StorageConsumer CR status as well.
```
cephResources:
  - kind: CephClient
    name: d2ee5d8373e686f9a69ec7add3360eff
    status: Ready
```

PR #2838, opened 2024-10-03
Remove healthchecker ceph client

ezio-auditore:
Solves: https://github.com/red-hat-storage/ocs-operator/issues/2829

### DIFF
--- a/controllers/storageconsumer/consumer_test.go
+++ b/controllers/storageconsumer/consumer_test.go
@@ -26,8 +26,6 @@ import (
 	routev1 "github.com/openshift/api/route/v1"
 	v1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
 	ocsv1alpha1 "github.com/red-hat-storage/ocs-operator/api/v4/v1alpha1"
-	"github.com/red-hat-storage/ocs-operator/v4/controllers/util"
-	rookCephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
 	"github.com/stretchr/testify/assert"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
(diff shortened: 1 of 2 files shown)


rewantsoni (changes requested):

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -59,11 +59,10 @@ type StorageConsumerReconciler struct {
 	Log    logr.Logger
 	Scheme *runtime.Scheme
 
-	ctx                     context.Context
-	storageConsumer         *ocsv1alpha1.StorageConsumer
-	cephClientHealthChecker *rookCephv1.CephClient
```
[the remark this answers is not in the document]
ezio-auditore: Updated

rewantsoni:
The changes look good, just wondering if we should update the name for setCephResourceStatus method and `CephResources` field in the StorageConsumer Status to reflect that they can be any kind of `Resources` (NoobaaAccount or any other fields we might want to add in future)

controllers/storageconsumer/consumer_test.go
```diff
@@ -179,18 +147,13 @@ func TestCephName(t *testing.T) {
 				{
 					Kind:  "CephClient",
 					Name:  "rbd",
-					Phase: "Error",
+					Phase: "Ready",
```
Madhu-1, line 150: we have only Noobaa in the status nothing else, right? why do we have ceph clients in the status field?

ezio-auditore: Removed all cephclients

ezio-auditore:
> The changes look good, just wondering if we should update the name for setCephResourceStatus method and `CephResources` field in the StorageConsumer Status to reflect that they can be any kind of `Resources` (NoobaaAccount or any other fields we might want to add in future)

@rewantsoni Yes I agree the name should be changed from CephResources to GenericResources or some other name to clearly justify the need and the purpose of the field.

Madhu-1 (commented):
LGTM.

Madhu-1:
> > The changes look good, just wondering if we should update the name for setCephResourceStatus method and `CephResources` field in the StorageConsumer Status to reflect that they can be any kind of `Resources` (NoobaaAccount or any other fields we might want to add in future)
> 
> @rewantsoni Yes I agree the name should be changed from CephResources to GenericResources or some other name to clearly justify the need and the purpose of the field.

This might need API change as API is tied to the CephResource name

iamniting (approved):

PR #2841, opened 2024-10-07
Adjust Node Label Conditions Based on Full Label Name

OdedViner:
The OCS operator searches for the `topology.kubernetes.io/zone` label key using a string containing logic. 
If the customer has a node with a label that includes the string `zone`. 
The OCS operator can read another label containing 'zone' instead of topology.kubernetes.io/zone.

### DIFF
--- a/controllers/storagecluster/topology.go
+++ b/controllers/storagecluster/topology.go
@@ -100,7 +100,7 @@ func setFailureDomain(sc *ocsv1.StorageCluster) {
 	// If sufficient zones are available then we select zone as the failure domain
 	topologyMap := sc.Status.NodeTopologies
 	for label, labelValues := range topologyMap.Labels {
-		if strings.Contains(label, "zone") {
+		if label == corev1.LabelZoneFailureDomainStable || label == labelZoneFailureDomainWithoutBeta {
 			if (len(labelValues) >= 2 && arbiterEnabled(sc)) || (len(labelValues) >= 3) {
 				failureDomain = "zone"
 			}
@@ -135,7 +135,7 @@ func determinePlacementRack(
 	targetAZ := ""
 	for label, value := range node.Labels {
 		for _, key := range validTopologyLabelKeys {
-			if strings.Contains(label, key) && strings.Contains(label, "zone") {
+			if strings.Contains(label, key) && (label == corev1.LabelZoneFailureDomainStable || label == labelZoneFailureDomainWithoutBeta) {
 				targetAZ = value
 				break
 			}
@@ -159,7 +159,7 @@ func determinePlacementRack(
 					if n.Name == nodeName {
 						for label, value := range n.Labels {
 							for _, key := range validTopologyLabelKeys {
-								if strings.Contains(label, key) && strings.Contains(label, "zone") && value == targetAZ {
+								if strings.Contains(label, key) && (label == corev1.LabelZoneFailureDomainStable || label == labelZoneFailureDomainWithoutBeta) && value == targetAZ {
 									validRack = true
 									break
 								}


controllers/storagecluster/topology.go
```diff
@@ -100,7 +100,7 @@ func setFailureDomain(sc *ocsv1.StorageCluster) {
 	// If sufficient zones are available then we select zone as the failure domain
 	topologyMap := sc.Status.NodeTopologies
 	for label, labelValues := range topologyMap.Labels {
-		if strings.Contains(label, "zone") {
+		if label == corev1.LabelZoneFailureDomainStable {
```
malayparida2000, line 103: label would not directly match like this, We need to first parse the label on the node to get the key part & then do the matching.
As the label on the node would in the form like
`
topology.kubernetes.io/zone=zone1 or topology.rook.io/rack=rack1 or kubernetes.io/hostname=node1
`
We need to parse the label first to get the key part of the label like 
`topology.kubernetes.io/zone or topology.rook.io/rack or kubernetes.io/hostname
`
And then we need to match them.

controllers/storagecluster/topology.go
```diff
@@ -135,7 +135,7 @@ func determinePlacementRack(
 	targetAZ := ""
 	for label, value := range node.Labels {
 		for _, key := range validTopologyLabelKeys {
-			if strings.Contains(label, key) && strings.Contains(label, "zone") {
+			if strings.Contains(label, key) && label == corev1.LabelZoneFailureDomainStable {
```
malayparida2000, line 138: For zone, we need to check `topology.kubernetes.io/zone` corev1.LabelTopologyZone
https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesiozone

For host we need to check `kubernetes.io/hostname` corev1. LabelHostname
https://kubernetes.io/docs/reference/labels-annotations-taints/#kubernetesiohostname

For rack we need to check "topology.rook.io/rack"

malayparida2000 (changes requested):

controllers/storagecluster/topology.go
OdedViner: image
Based on my debug, the label and label values are separated.

controllers/storagecluster/topology.go
OdedViner: Why don't we check `hostname` and  `rack`  in the original code?

controllers/storagecluster/topology.go
malayparida2000: Oh, Sorry for the confusion, you are write in the node topology map the label contains just the key part.
found an example here 
```
nodeTopologies:
      labels:
        kubernetes.io/hostname:
        - k8sbm-1283800.ln.fw.gs.com
        - k8sbm-1283803.ln.fw.gs.com
        - k8sbm-1283804.ln.fw.gs.com
        - k8sbm-1283805.ln.fw.gs.com
        - k8sbm-1431842.k8s.gs.com
        - k8sbm-1431848.k8s.gs.com
        - k8sbm-1433161.k8s.gs.com
        - k8sbm-1433163.k8s.gs.com
        - k8sbm-1433165.k8s.gs.com
        - k8sbm-1433166.k8s.gs.com
        - k8sbm-1433178.k8s.gs.com
        - k8sbm-1433184.k8s.gs.com
        - k8sbm-1433185.k8s.gs.com
        - k8sbm-1433321.k8s.gs.com
        - k8sbm-1433322.k8s.gs.com
        - k8sbm-1433333.k8s.gs.com
        topology.rook.io/rack:
        - rack0
        - rack1
        - rack2
```

controllers/storagecluster/topology.go
malayparida2000: That's due to the way failure domain are determined, We by default assume it to be rack, Then hostname if flexible scaling is true, if not at the end we check for zone.

iamniting (approved):

PR #2842, opened 2024-10-07
provider: update csi secret name to be consistent

rewantsoni:
For RDR, the metadata for pv,pvc are  backed up and restored on the secondary cluster as it is, if the csi secrets with the same name are not present, it will cause an error. Hence we should keep the name for csi secret consistent and based on the storageClass(storageClaim) hash instead of consumerUID and storageClaimHash

### DIFF
--- a/services/provider/server/server.go
+++ b/services/provider/server/server.go
@@ -647,7 +647,7 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
 	}
 	var extR []*pb.ExternalResource
 
-	storageRequestHash := getStorageRequestHash(req.StorageConsumerUUID, req.StorageClaimName)
+	storageRequestHash := util.CalculateMD5Hash(req.StorageClaimName)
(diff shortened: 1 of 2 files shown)


services/provider/server/server.go
```diff
@@ -647,7 +647,7 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
 	}
 	var extR []*pb.ExternalResource
 
-	storageRequestHash := getStorageRequestHash(req.StorageConsumerUUID, req.StorageClaimName)
+	storageRequestHash := util.CalculateMD5Hash(req.StorageClaimName)
```
nb-ohad, line 650: @rewantsoni We cannot apply this change, it breaks backport compatibility with existing deployments.

nb-ohad (changes requested):

nb-ohad:
@rewantsoni Can we please close this PR, the change is not an acceptable change as it breaks backport compatibility

PR #2843, opened 2024-10-08
api: update the storageClusterPeer api

rewantsoni:
Update the storageClusterPeer API to contain only onboardingToken and the endpoint


leelavg (commented):
/lgtm
based on updated design

api/v1/storageclusterpeer_types.go
```diff
@@ -17,57 +17,37 @@ limitations under the License.
...
-// BlockPoolMirroringSpec enables setting up of mirroring for blockPools in the same namespace.
-type BlockPoolMirroringSpec struct {
-	// Selector is used to select blockPools by label
-	Selector metav1.LabelSelector `json:"selector"`
-}
+const (
+	// StorageClusterPeerRemoteStateInitializing represents Initializing state of StorageClusterPeerState
+	StorageClusterPeerRemoteStateInitializing StorageClusterPeerState = "Initializing"
+	// StorageClusterPeerRemoteStatePeering represents Peering state of storageClusterPeer
+	StorageClusterPeerRemoteStatePeering StorageClusterPeerState = "Peering"
```
nb-ohad, line 32: Why peering? are we doing anything active or are we just waiting for a response? if we are just waiting then the name should be called "Pending"

api/v1/storageclusterpeer_types.go
```diff
@@ -17,57 +17,37 @@ limitations under the License.
...
-}
-
-// BlockPoolMirroringSpec enables setting up of mirroring for blockPools in the same namespace.
-type BlockPoolMirroringSpec struct {
-	// Selector is used to select blockPools by label
-	Selector metav1.LabelSelector `json:"selector"`
-}
+const (
+	// StorageClusterPeerRemoteStateInitializing represents Initializing state of StorageClusterPeerState
+	StorageClusterPeerRemoteStateInitializing StorageClusterPeerState = "Initializing"
```
nb-ohad, line 30: I am not sure what Initializing means in the context of our peering algorithm

api/v1/storageclusterpeer_types.go
```diff
@@ -17,57 +17,37 @@ limitations under the License.
...
-	//+optional
-	BlockPoolMirroring *BlockPoolMirroringSpec `json:"blockPoolMirroring,omitempty"`
+	// OnboardingToken holds an identity information required by the local ODF cluster to onboard.
+	OnboardingToken string `json:"onboardingToken"`
 }
 
 // StorageClusterPeerStatus defines the observed state of StorageClusterPeer
 type StorageClusterPeerStatus struct {
+	State                   StorageClusterPeerState `json:"state,omitempty"`
+	RemoteStorageClusterUID string                  `json:"remoteStorageClusterUID,omitempty"`
```
nb-ohad, line 50: Please remove the terminology of "remote". The established terminology we agreed to is "peer".
```suggestion
	PeerStorageClusterUID string                  `json:"peerStorageClusterUID,omitempty"`
```

nb-ohad, line 50: Can we have this as a struct `PeerInfo`?
This will allow us to return more information about the peer we might need for one reason or another.
Some candidates for such information are:
- StorageClusterName
- ClusterName
- ClusterID 
etc.

nb-ohad (changes requested):

api/v1/storageclusterpeer_types.go
rewantsoni: removed

api/v1/storageclusterpeer_types.go
```diff
@@ -17,57 +17,37 @@ limitations under the License.
...
+	StorageClusterPeerStatePending StorageClusterPeerState = "Pending"
+	StorageClusterPeerStatePeered  StorageClusterPeerState = "Peered"
+)
 
-// BlockPoolMirroringSpec enables setting up of mirroring for blockPools in the same namespace.
-type BlockPoolMirroringSpec struct {
-	// Selector is used to select blockPools by label
-	Selector metav1.LabelSelector `json:"selector"`
+type PeerInfo struct {
+	State                 StorageClusterPeerState `json:"state,omitempty"`
```
nb-ohad, line 34: Why is `State` under PeerInfo? Isn't that the state of the entire peering operation? it should go under the Status section.

api/v1/storageclusterpeer_types.go
```diff
@@ -17,57 +17,37 @@ limitations under the License.
...
+	PeerStorageClusterUID string                  `json:"peerStorageClusterUID,omitempty"`
```
nb-ohad, line 35: You are already under the peer info section, there is no need to prefix the field with the word peer, as the field is already accessed as `Status.PeerInfo.StorageClusterUid`

nb-ohad, line 35: Abbreviations should capitalize only the first letter. This means UID -> Uid

api/v1/storageclusterpeer_types.go
```diff
@@ -17,57 +17,37 @@ limitations under the License.
...
-	// BlockPoolMirroring indicates ceph block mirroring between block pool on the local and remote clusters
-	//+optional
-	BlockPoolMirroring *BlockPoolMirroringSpec `json:"blockPoolMirroring,omitempty"`
+	// OnboardingToken holds an identity information required by the local ODF cluster to onboard.
+	OnboardingToken string `json:"onboardingToken"`
 }
 
 // StorageClusterPeerStatus defines the observed state of StorageClusterPeer
 type StorageClusterPeerStatus struct {
+	PeerInfo PeerInfo `json:"peerInfo"`
```
nb-ohad, line 50: I think you are missing your status field here

nb-ohad (changes requested):

PR #2844, opened 2024-10-08
onboardingTicket: add uid to the ticket

rewantsoni:
onboarding ticket should contain the storageCluster UID the token was generated for and to fetch the storageCluster for UID, we fetch the storage Cluster from operator namespace

### DIFF
--- a/controllers/util/k8sutil.go
+++ b/controllers/util/k8sutil.go
@@ -13,8 +13,11 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	storagev1 "k8s.io/api/storage/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
 )
 
 const (
@@ -165,4 +168,37 @@ func GenerateNameForNonResilientCephBlockPoolSC(initData *ocsv1.StorageCluster) 
 		return initData.Spec.ManagedResources.CephNonResilientPools.StorageClassName
 	}
 	return fmt.Sprintf("%s-ceph-non-resilient-rbd", initData.Name)
+}
+
+func GetStorageClusterInNamespace(ctx context.Context, cl client.Client, namespace string) (*ocsv1.StorageCluster, error) {
+	storageClusterList := &ocsv1.StorageClusterList{}
+	err := cl.List(ctx, storageClusterList, client.InNamespace(namespace), client.Limit(1))
+	if err != nil {
+		return nil, fmt.Errorf("unable to list storageCluster(s) in namespace %s: %v", namespace, err)
+	}
+
+	if len(storageClusterList.Items) == 0 {
+		return nil, fmt.Errorf("no storageCluster found in namespace %s", namespace)
+	}
+	if storageClusterList.Items[0].Status.Phase == PhaseIgnored {
+		return nil, fmt.Errorf("storageCluster with Phase 'Ignored' found. Please delete the storageCluster to proceed")
+	}
+
+	return &storageClusterList.Items[0], nil
+}
+
+func NewK8sClient(scheme *runtime.Scheme) (client.Client, error) {
+	klog.Info("Setting up k8s client")
+
+	config, err := config.GetConfig()
+	if err != nil {
+		return nil, err
+	}
+
+	k8sClient, err := client.New(config, client.Options{Scheme: scheme})
+	if err != nil {
+		return nil, err
+	}
+
+	return k8sClient, nil
 }
--- a/controllers/util/provider.go
+++ b/controllers/util/provider.go
@@ -13,14 +13,16 @@ import (
 	"os"
 	"time"
 
-	"github.com/google/uuid"
 	"github.com/red-hat-storage/ocs-operator/v4/services"
+
+	"github.com/google/uuid"
+	"k8s.io/apimachinery/pkg/types"
 )
 
 // GenerateClientOnboardingToken generates a ocs-client token valid for a duration of "tokenLifetimeInHours".
 // The token content is predefined and signed by the private key which'll be read from supplied "privateKeyPath".
 // The storageQuotaInGiB is optional, and it is used to limit the storage of PVC in the application cluster.
-func GenerateClientOnboardingToken(tokenLifetimeInHours int, privateKeyPath string, storageQuotainGib *uint) (string, error) {
+func GenerateClientOnboardingToken(tokenLifetimeInHours int, privateKeyPath string, storageQuotainGib *uint, storageClusterUID types.UID) (string, error) {
 	tokenExpirationDate := time.Now().
 		Add(time.Duration(tokenLifetimeInHours) * time.Hour).
 		Unix()
@@ -30,6 +32,7 @@ func GenerateClientOnboardingToken(tokenLifetimeInHours int, privateKeyPath stri
 		ExpirationDate:    tokenExpirationDate,
 		SubjectRole:       services.ClientRole,
 		StorageQuotaInGiB: storageQuotainGib,
+		StorageCluster:    storageClusterUID,
 	}
 
 	token, err := encodeAndSignOnboardingToken(privateKeyPath, ticket)
@@ -41,7 +44,7 @@ func GenerateClientOnboardingToken(tokenLifetimeInHours int, privateKeyPath stri
 
 // GeneratePeerOnboardingToken generates a ocs-peer token valid for a duration of "tokenLifetimeInHours".
 // The token content is predefined and signed by the private key which'll be read from supplied "privateKeyPath".
-func GeneratePeerOnboardingToken(tokenLifetimeInHours int, privateKeyPath string) (string, error) {
+func GeneratePeerOnboardingToken(tokenLifetimeInHours int, privateKeyPath string, storageClusterUID types.UID) (string, error) {
 	tokenExpirationDate := time.Now().
 		Add(time.Duration(tokenLifetimeInHours) * time.Hour).
 		Unix()
@@ -50,6 +53,7 @@ func GeneratePeerOnboardingToken(tokenLifetimeInHours int, privateKeyPath string
 		ID:             uuid.New().String(),
 		ExpirationDate: tokenExpirationDate,
 		SubjectRole:    services.PeerRole,
+		StorageCluster: storageClusterUID,
 	}
 	token, err := encodeAndSignOnboardingToken(privateKeyPath, ticket)
 	if err != nil {
--- a/onboarding-validation-keys-generator/main.go
+++ b/onboarding-validation-keys-generator/main.go
@@ -5,15 +5,16 @@ import (
 	"crypto/rsa"
 	"crypto/x509"
 	"encoding/pem"
-	v1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
+
+	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
 	"github.com/red-hat-storage/ocs-operator/v4/controllers/util"
+
 	"golang.org/x/net/context"
 	corev1 "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
@@ -24,9 +25,18 @@ const (
 )
 
 func main() {
-	cl, err := newClient()
+
+	scheme := runtime.NewScheme()
+	if err := corev1.AddToScheme(scheme); err != nil {
+		klog.Exitf("failed to add corev1 to scheme. %v", err)
+	}
+	if err := ocsv1.AddToScheme(scheme); err != nil {
+		klog.Exitf("failed to add ocsv1 to scheme. %v", err)
+	}
+
+	cl, err := util.NewK8sClient(scheme)
 	if err != nil {
-		klog.Exitf("failed to create client: %v", err)
+		klog.Exitf("failed to create kube client: %v", err)
 	}
 
 	ctx := context.Background()
@@ -44,15 +54,10 @@ func main() {
 	privatePem := convertRsaPrivateKeyAsPemStr(privateKey)
 	publicPem := convertRsaPublicKeyAsPemStr(publicKey)
 
-	storageClusterList := &v1.StorageClusterList{}
-	err = cl.List(ctx, storageClusterList, client.InNamespace(namespace))
+	storageCluster, err := util.GetStorageClusterInNamespace(ctx, cl, namespace)
 	if err != nil {
-		klog.Exitf("unable to list storageCluster(s) in %v namespace, %v", namespace, err)
+		klog.Exitf("failed to get storage cluster: %v", err)
 	}
-	if len(storageClusterList.Items) != 1 {
-		klog.Exitf("unexpected number of storageCluster(s) found in %v namespace, expected: 1 actual: %v", namespace, len(storageClusterList.Items))
-	}
-	storageCluster := &storageClusterList.Items[0]
 
 	// In situations where there is a risk of one secret being updated and potentially
 	// failing to update another, it is recommended not to rely solely on clientset update mechanisms.
@@ -103,27 +108,6 @@ func main() {
 		klog.Exitf("failed to create public secret: %v", err)
 	}
 
-}
-
-func newClient() (client.Client, error) {
-	klog.Info("Setting up k8s client")
-	scheme := runtime.NewScheme()
-	if err := v1.AddToScheme(scheme); err != nil {
-		return nil, err
-	}
-	if err := corev1.AddToScheme(scheme); err != nil {
-		return nil, err
-	}
-	config, err := config.GetConfig()
-	if err != nil {
-		return nil, err
-	}
-	k8sClient, err := client.New(config, client.Options{Scheme: scheme})
-	if err != nil {
-		return nil, err
-	}
-
-	return k8sClient, nil
 }
 
 func convertRsaPrivateKeyAsPemStr(privateKey *rsa.PrivateKey) string {
--- a/services/types.go
+++ b/services/types.go
@@ -1,5 +1,7 @@
 package services
 
+import "k8s.io/apimachinery/pkg/types"
+
 type OnboardingSubjectRole string
 
 const (
@@ -12,4 +14,5 @@ type OnboardingTicket struct {
 	ExpirationDate    int64                 `json:"expirationDate,string"`
 	SubjectRole       OnboardingSubjectRole `json:"subjectRole"`
 	StorageQuotaInGiB *uint                 `json:"storageQuotaInGiB,omitempty"`
+	StorageCluster    types.UID             `json:"storageCluster"`
 }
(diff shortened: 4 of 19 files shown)


mrudraia1:
I faced storagecluster UID error while testing the PR 2827. The PR 2827 can be tested after this PR is merged.

services/ux-backend/handlers/onboarding/peertokens/handler.go
```diff
@@ -1,30 +1,73 @@
...
-func handlePost(w http.ResponseWriter, _ *http.Request, tokenLifetimeInHours int) {
-	if onboardingToken, err := util.GeneratePeerOnboardingToken(tokenLifetimeInHours, onboardingPrivateKeyFilePath); err != nil {
+func handlePost(w http.ResponseWriter, r *http.Request, tokenLifetimeInHours int, cl client.Client) {
+	if r.ContentLength == 0 {
+		http.Error(w, "Request body is empty, expected StorageCluster NamespacedName", http.StatusBadRequest)
+		return
+	}
+
+	var body = struct {
+		StorageClusterNamespacedName string `json:"storageClusterNamespacedName"`
```
nb-ohad, line 37: Why so we need the namespaced name? Don't we just need a namespace?

services/ux-backend/handlers/onboarding/peertokens/handler.go
```diff
@@ -1,30 +1,73 @@
...
+	if err != nil {
+		http.Error(
+			w,
+			fmt.Sprintf("failed to find StorageCluster with NamespacedName %s: %v", body.StorageClusterNamespacedName, err),
+			http.StatusBadRequest,
+		)
+		return
+	}
+
+	if onboardingToken, err := util.GeneratePeerOnboardingToken(tokenLifetimeInHours, onboardingPrivateKeyFilePath, string(storageCluster.UID)); err != nil {
```
nb-ohad, line 70: Can't we accept UID as the parameter type for the function? It makes more sense that we would like to keep the param as typed as possible

services/ux-backend/handlers/onboarding/peertokens/handler.go
```diff
@@ -1,30 +1,73 @@
...
+
+	err := cl.Get(r.Context(), client.ObjectKeyFromObject(storageCluster), storageCluster)
+	if err != nil {
+		http.Error(
+			w,
+			fmt.Sprintf("failed to find StorageCluster with NamespacedName %s: %v", body.StorageClusterNamespacedName, err),
+			http.StatusBadRequest,
+		)
+		return
+	}
```
nb-ohad, line 68: This entire logic here needs to be duplicated to the client onboarding token handler. I would suggest to have this as a util of sorts to be invoked from both places

services/ux-backend/main.go
```diff
@@ -94,3 +104,26 @@ func main() {
...
+		return nil, err
+	}
+
+	client, err := client.New(config, client.Options{Scheme: scheme})
+	if err != nil {
+		return nil, err
+	}
+
+	return client, nil
+}
```
nb-ohad, line 129: 1. I believe we already have this code in the code base somewhere else. 
I would prefer to have a shared code invoked from all locations. 

2. In this file all private function resides above the main function, can we please be consistency

nb-ohad (changes requested):

services/ux-backend/handlers/onboarding/peertokens/handler.go
rewantsoni: I was thinking from a user perspective, it would make more sense to specify the storageCluster Name they want to onboard to instead of the namespace of the storageCluster. Since storageCluster is a namespace resource, we can have storageClusters with the same name in a different namespace hence chose to go with storageCluster namespacedName to identify the storageCluster user wants to onboard to

nb-ohad: These endpoints are mainly for the UI to invoke, the UI already has the namespace context and does not need to ask the user for anything. Also, a namespace name as part of a user ask can create mistaken input, where the named resource does not exist in the namespace.

controllers/util/client.go
leelavg, line 1: since we have atleast two clients, gRPC and Kube better rename this file to be more accurate? `k8sclient.go`?

controllers/util/k8sutil.go
```diff
@@ -166,3 +166,15 @@ func GenerateNameForNonResilientCephBlockPoolSC(initData *ocsv1.StorageCluster)
 	}
 	return fmt.Sprintf("%s-ceph-non-resilient-rbd", initData.Name)
 }
+
+func GetStorageClusterInNamespace(ctx context.Context, cl client.Client, storageClusterNamespace string) (*ocsv1.StorageCluster, error) {
```
leelavg, line 170: function name already says `Namespace` and last arg is repeating the same, remove a duplicate from one of these? `GetStorageCluster` or `namespace string`?

onboarding-validation-keys-generator/main.go
```diff
@@ -44,7 +54,7 @@ func main() {
 	privatePem := convertRsaPrivateKeyAsPemStr(privateKey)
 	publicPem := convertRsaPublicKeyAsPemStr(publicKey)
 
-	storageClusterList := &v1.StorageClusterList{}
+	storageClusterList := &ocsv1.StorageClusterList{}
 	err = cl.List(ctx, storageClusterList, client.InNamespace(namespace))
 	if err != nil {
 		klog.Exitf("unable to list storageCluster(s) in %v namespace, %v", namespace, err)
```
leelavg, line 60: can be replaced with newly created function `util.GetStorageCluster`?

services/ux-backend/handlers/onboarding/peertokens/handler.go
```diff
@@ -8,23 +8,31 @@ import (
...
 
-func handlePost(w http.ResponseWriter, _ *http.Request, tokenLifetimeInHours int) {
-	if onboardingToken, err := util.GeneratePeerOnboardingToken(tokenLifetimeInHours, onboardingPrivateKeyFilePath); err != nil {
+func handlePost(w http.ResponseWriter, r *http.Request, tokenLifetimeInHours int, cl client.Client, namespace string) {
+
+	storageCluster, err := util.GetStorageClusterInNamespace(r.Context(), cl, namespace)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
```
leelavg, line 33: could you pls follow the existing way of error handling across the service?

services/ux-backend/main.go
```diff
@@ -63,6 +67,21 @@ func main() {
 		os.Exit(-1)
 	}
 
+	namespace := util.GetPodNamespace()
```
leelavg, line 70: we are starting w/ supporting a single storagecluster that exists in operator ns?

services/ux-backend/main.go
```diff
@@ -63,6 +67,21 @@ func main() {
...
+
+	scheme := runtime.NewScheme()
+	if err := corev1.AddToScheme(scheme); err != nil {
+		klog.Exitf("failed to add corecv2 to scheme. %v", err)
```
leelavg, line 74:
```suggestion
		klog.Exitf("failed to add corev1 to scheme. %v", err)
```

services/ux-backend/main.go
```diff
@@ -63,6 +67,21 @@ func main() {
...
+	}
+	if err := ocsv1.AddToScheme(scheme); err != nil {
+		klog.Exitf("failed to add ocsv1 to scheme. %v", err)
+	}
+
+	cl, err := util.NewK8sClient(scheme)
+	if err != nil {
+		klog.Exitf("failed to create client: %v", err)
```
leelavg, line 82:
```suggestion
		klog.Exitf("failed to create kube client: %v", err)
```

onboarding-validation-keys-generator/main.go
```diff
@@ -24,7 +25,16 @@ const (
 )
 
 func main() {
-	cl, err := newClient()
+
+	scheme := runtime.NewScheme()
+	if err := corev1.AddToScheme(scheme); err != nil {
+		klog.Exitf("failed to add corecv2 to scheme. %v", err)
```
leelavg, line 31:
```suggestion
		klog.Exitf("failed to add corev1 to scheme. %v", err)
```

services/ux-backend/main.go
rewantsoni: Yes

controllers/util/k8sutil.go
leelavg: thinking again, suggestions may just show up function name and in that scenario as we are using in multiple places, `GetStorageClusterByNamespace` would be good?

controllers/util/k8sutil.go
```diff
@@ -166,3 +166,15 @@ func GenerateNameForNonResilientCephBlockPoolSC(initData *ocsv1.StorageCluster)
 	}
 	return fmt.Sprintf("%s-ceph-non-resilient-rbd", initData.Name)
 }
+
+func GetStorageClusterInNamespace(ctx context.Context, cl client.Client, namespace string) (*ocsv1.StorageCluster, error) {
+	storageClusterList := &ocsv1.StorageClusterList{}
+	err := cl.List(ctx, storageClusterList, client.InNamespace(namespace))
```
nb-ohad, line 172: Can you please cap this list operation to 1 result?

controllers/util/provider.go
```diff
@@ -41,15 +43,16 @@ func GenerateClientOnboardingToken(tokenLifetimeInHours int, privateKeyPath stri
...
 		Unix()
 
 	ticket := services.OnboardingTicket{
-		ID:             uuid.New().String(),
-		ExpirationDate: tokenExpirationDate,
-		SubjectRole:    services.PeerRole,
+		ID:                uuid.New().String(),
+		ExpirationDate:    tokenExpirationDate,
+		SubjectRole:       services.PeerRole,
+		StorageClusterUID: storageClusterUID,
```
nb-ohad, line 55: 1. Please cap only the first letter of an abbreviation 
2. Why do we need to call this StorageClusterUID and not just StorageCluster in the context of the ticket?

controllers/util/provider.go
```diff
@@ -41,15 +43,16 @@ func GenerateClientOnboardingToken(tokenLifetimeInHours int, privateKeyPath stri
 
 // GeneratePeerOnboardingToken generates a ocs-peer token valid for a duration of "tokenLifetimeInHours".
 // The token content is predefined and signed by the private key which'll be read from supplied "privateKeyPath".
-func GeneratePeerOnboardingToken(tokenLifetimeInHours int, privateKeyPath string) (string, error) {
+func GeneratePeerOnboardingToken(tokenLifetimeInHours int, privateKeyPath string, storageClusterUID types.UID) (string, error) {
```
nb-ohad, line 46: Please cap only the first letter of an abbreviation 
```suggestion
func GeneratePeerOnboardingToken(tokenLifetimeInHours int, privateKeyPath string, storageClusterUid types.UID) (string, error) {
```

services/types.go
```diff
@@ -12,4 +14,5 @@ type OnboardingTicket struct {
 	ExpirationDate    int64                 `json:"expirationDate,string"`
 	SubjectRole       OnboardingSubjectRole `json:"subjectRole"`
 	StorageQuotaInGiB *uint                 `json:"storageQuotaInGiB,omitempty"`
+	StorageClusterUID types.UID             `json:"storageClusterUID"`
```
nb-ohad, line 17: Only capital the first latter of an abbreviations
```suggestion
	StorageClusterUid types.UID             `json:"storageClusterUid"`
```

services/ux-backend/handlers/onboarding/peertokens/handler.go
nb-ohad: @rewantsoni You are in the context of handling the message, please follow the proper way to write into the response stream (w.WriteHeader), and use a proper HTTP response code

controllers/util/k8sclient.go
```diff
@@ -0,0 +1,24 @@
+package util
+
+import (
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+)
+
+func NewK8sClient(scheme *runtime.Scheme) (client.Client, error) {
```
nb-ohad, line 10: 1. There is no need to have a new util file. This is definitely a k8s util function.
2. Moving this into util is a good move, but you still need to consolidate all the calla (for client creation) across the repo

nb-ohad (changes requested):

nb-ohad (commented):
@rewantsoni I see the changes to the token done for the `peertokens` handler, but where are the changes to the `peerclient` handler? 
The introduction of a storagecluster UID into the token need to happen there as well (it should be a mandatory field)

rewantsoni:
>@rewantsoni You are in the context of handling the message, please follow the proper way to write into the response stream (w.WriteHeader), and use a proper HTTP response code
@nb-ohad http.Error is a helper function that sets the error code on the header and sets the response. See [here](https://pkg.go.dev/net/http#Error)

rewantsoni:
>@rewantsoni You are in the context of handling the message, please follow the proper way to write into the response stream (w.WriteHeader), and use a proper HTTP response code
@nb-ohad http.Error is a helper function that sets the error code on the header and sets the response. See [here](https://pkg.go.dev/net/http#Error)

controllers/util/k8sutil.go
rewantsoni: I was going through https://github.com/red-hat-storage/ocs-operator/pull/2837, and from what I understood we can have multiple storageClusters, but the other storageCluster's phase will be ignored. Modified the function there to GetStorageClusterInNamespace

controllers/util/provider.go
rewantsoni: golang-lint is failing with this change

controllers/util/k8sutil.go
nb-ohad: You cannot have multi-storage clusters in a single namespace, and this lists only inside the namespace. So my ask adding `client.Count(1)` is meaningful

malayparida2000: We can have multple storageclusters inside a single namespace, Just that except one all else would be in phase ignored. Limiting the list to 1 result can easily give us the wrong result with a phasIgnored storageCluster. I would suggest listing all the items & then filtering from that list based on the requirement is the better approach.

nb-ohad: @malayparida2000 I understand what you are saying, but that is a scenario we should not address this way. That is a user error, and if we want to solve it we should introduce a webhook to prevent the creation of a second storage cluster. 

Listing with a count of (1) prevents the deep copy operations of all other storage classes from the cache into the resulting list, in the case of storage cluster it is a very very big structure (even if it is mostly not filled) and we should avoid the deep copy.

controllers/util/provider.go
nb-ohad: I think it is a golang lint issue, @iamniting might be able to give some advice here, or suggest a separate fix for this.
In the meantime, we can merge this (assuming no other issues) without this change.

services/ux-backend/handlers/onboarding/clienttokens/handler.go
```diff
@@ -59,7 +60,13 @@ func handlePost(w http.ResponseWriter, r *http.Request, tokenLifetimeInHours int
 		storageQuotaInGiB = ptr.To(unitAsGiB * quota.Value)
 	}
 
-	if onboardingToken, err := util.GenerateClientOnboardingToken(tokenLifetimeInHours, onboardingPrivateKeyFilePath, storageQuotaInGiB); err != nil {
+	storageCluster, err := util.GetStorageClusterInNamespace(r.Context(), cl, namespace)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
```
nb-ohad, line 67: This cannot be an `http.StatusBadRequest` because the failure has nothing to do with the content of the request, which also means that the user cannot change the request and clear the check. 
For these kinds of issues, I would suggest using `http.StatusInternalServerError`

services/ux-backend/handlers/onboarding/peertokens/handler.go
nb-ohad, line 33: Same as my last comment. This cannot be a `http.StatusBadRequest`

nb-ohad (changes requested):

controllers/util/k8sutil.go
rewantsoni: @malayparida2000 @nb-ohad I will open an issue about it, so that we can create a webhook for it after the feature freeze, for now, we can list with a count of (1) and ignore the others

rewantsoni: Issue: https://github.com/red-hat-storage/ocs-operator/issues/2877

leelavg: pls excuse for pitching in and correct me if I'm not understanding it properly. Agreed, it's an user error but if it did happen, count of (1) has a min of 50% chance of going wrong.

until we make the storagecluster in a ns singleton I don't think we incur a lot of mem usage if runtime does a deep copy to provide us as the stack size can go upto MBs and last time I checked storagecluster is around 2KB (without managed fields) but w/ pointer and all it'll more than that but not in a 2 digit MBs.

however, I agree w/ Ohad since we are returning a pointer to caller and storagecluster escapes to heap so does the list that is queried in this function increasing GC and heap usage, but kube clients outside of manager aren't backed up by cache and so every list call hits the api server (with the limit if set).

imo, we should code for correctness and list all storageclusters to filter out ignored one's unless we make storagecluster a singleton before merging this PR.

malayparida2000: I completely agree with Leela. Whether or not we decide to enforce having a single StorageCluster as mandatory is a separate discussion. However, currently, limiting the results to just one could lead to incorrect outcomes if we actually have more than one StorageCluster. So we can not limit the result to count(1) as of now.

Regarding the idea of creating a webhook to prevent additional StorageCluster creation, let's consider the practical implications. In most cases, a customer would either have a single StorageCluster as intended, or they might mistakenly create a second one. It's unlikely that a customer would intentionally create hundreds of ignored StorageCluster CRs to overwhelm our cache. Given this, the overhead of implementing a webhook to restrict additional StorageCluster creation might outweigh listine an extra StorageCluster in the "ignored" phase if one exists due to a customer error.

nb-ohad: @leelavg The point is that the probability of that happening is very low and we will inform that on the log and will instruct the user/support to delete the other StorageCluster. The code as it is now just ignores the additional storage cluster which is the wrong thing to do! It should be an error. 

now, considering the amount of places and flows where this method is invoked and the count of deep copies this method will create for a very big struct (storage cluster). I am asking us to treat an additional storage cluster as an error not as thing that is just ignored

@rewantsoni please make that change as requested

controllers/util/k8sutil.go
```diff
@@ -166,3 +169,33 @@ func GenerateNameForNonResilientCephBlockPoolSC(initData *ocsv1.StorageCluster)
...
+	err := cl.List(ctx, storageClusterList, client.InNamespace(namespace), client.Limit(1))
+	if err != nil {
+		return nil, fmt.Errorf("unable to list storageCluster(s) in namespace %s: %v", namespace, err)
+	}
+
+	if len(storageClusterList.Items) == 0 {
+		return nil, fmt.Errorf("no storageCluster found in namespace %s", namespace)
+	}
+
+	return &storageClusterList.Items[0], nil
```
nb-ohad, line 184: Coming from the discussion we had on the other comment thread. You need to check that the storage cluster that you fetched is not in ignored state before proceeding. And if it is then you should return an error saying that this SC need to be deleted.

nb-ohad (changes requested):

controllers/util/k8sutil.go
nb-ohad: @malayparida2000 
> Given this, the overhead of implementing a webhook to restrict additional StorageCluster creation might outweigh listine an extra StorageCluster in the "ignored" phase if one exists due to a customer error.

We can deliberate about the nececity for a webhook at a later time, that has nothing to do with the fact that the system is misconfigured and we just ignoring it.

PR #2846, opened 2024-10-09
Add option to enable encryption for particular deviceset & Add KMS details when any of the deviceSet is encrypted.

malayparida2000:
Add api field under StorageDeviceSet spec to encrypt the deviceSet & Add KMS details to places when any of the DeviceSet is encrypted. Earlier we were adding KMS details to CephCluster, CephObjectStore, Noobaa & metrics when Cluster wide encryption was enabled. But now we are adding KMS details to these places when any of the DeviceSet
is encrypted.

https://issues.redhat.com/browse/RHSTOR-6484


malayparida2000:
Have to consider 
https://github.com/red-hat-storage/ocs-operator/blob/0b8482ea3e379ddf214deec0edf3b121bf24fc18/controllers/storagecluster/cephobjectstores.go#L48
https://github.com/red-hat-storage/ocs-operator/blob/0b8482ea3e379ddf214deec0edf3b121bf24fc18/controllers/storagecluster/noobaa_system_reconciler.go#L235
https://github.com/red-hat-storage/ocs-operator/blob/0b8482ea3e379ddf214deec0edf3b121bf24fc18/controllers/util/util.go#L45

api/v1/storagecluster_types.go
```diff
@@ -402,6 +402,10 @@ type StorageDeviceSet struct {
 	DataPVCTemplate     corev1.PersistentVolumeClaim  `json:"dataPVCTemplate"`
 	MetadataPVCTemplate *corev1.PersistentVolumeClaim `json:"metadataPVCTemplate,omitempty"`
 	WalPVCTemplate      *corev1.PersistentVolumeClaim `json:"walPVCTemplate,omitempty"`
+
+	// Whether to encrypt the deviceSet
+	// +optional
+	Encrypted bool `json:"encrypted,omitempty"`
```
parth-gr, line 408: we need a pointer here

controllers/storagecluster/cephcluster.go
```diff
@@ -882,7 +883,7 @@ func newStorageClassDeviceSets(sc *ocsv1.StorageCluster) []rookCephv1.StorageCla
 				Portable:            portable,
 				TuneSlowDeviceClass: ds.Config.TuneSlowDeviceClass,
 				TuneFastDeviceClass: ds.Config.TuneFastDeviceClass,
-				Encrypted:           sc.Spec.Encryption.Enable || sc.Spec.Encryption.ClusterWide,
+				Encrypted:           sc.Spec.Encryption.Enable || sc.Spec.Encryption.ClusterWide || ds.Encrypted,
```
parth-gr, line 886: we need a fallback mechanism here,
We don't have just trust `true`

If the deviceset is set to `true` or `false` we should use it.
If not then we should see the storagecluster top-level `Encryption` values

controllers/storagecluster/cephobjectstores.go
```diff
@@ -42,10 +43,10 @@ func (obj *ocsCephObjectStores) ensureCreated(r *StorageClusterReconciler, insta
 	}
 	var cephObjectStores []*cephv1.CephObjectStore
 	// Add KMS details to cephObjectStores spec, only if
-	// cluster-wide encryption is enabled
-	// ie, sc.Spec.Encryption.ClusterWide/sc.Spec.Encryption.Enable is True
+	// cluster-wide encryption is enabled or any of the device set is encrypted
+	// ie, sc.Spec.Encryption.ClusterWide/sc.Spec.Encryption.Enable is True or any deviceSet is encrypted
 	// and KMS ConfigMap is available
-	if instance.Spec.Encryption.Enable || instance.Spec.Encryption.ClusterWide {
+	if instance.Spec.Encryption.Enable || instance.Spec.Encryption.ClusterWide || util.IsAnyDeviceSetEncrypted(instance.Spec.StorageDeviceSets) {
```
parth-gr, line 49: We dont have objectstore at day-2?
But here we can also do a fallback mechanism

controllers/storagecluster/noobaa_system_reconciler.go
```diff
@@ -230,10 +230,10 @@ func (r *StorageClusterReconciler) setNooBaaDesiredState(nb *nbv1.NooBaa, sc *oc
 
 	// Add KMS details to Noobaa spec, only if
 	// KMS is enabled, along with
-	// ClusterWide encryption OR in a StandAlone Noobaa cluster mode
+	// ClusterWide encryption/any deviceSet Encryption OR in a StandAlone Noobaa cluster mode
 	// PS: sc.Spec.Encryption.Enable field is deprecated and added for backward compatibility
 	if sc.Spec.Encryption.KeyManagementService.Enable &&
-		(sc.Spec.Encryption.Enable || sc.Spec.Encryption.ClusterWide || r.IsNoobaaStandalone) {
+		(sc.Spec.Encryption.Enable || sc.Spec.Encryption.ClusterWide || util.IsAnyDeviceSetEncrypted(sc.Spec.StorageDeviceSets) || r.IsNoobaaStandalone) {
```
parth-gr, line 236: same

parth-gr (changes requested):

controllers/storagecluster/cephcluster.go
malayparida2000: But if someone has specified cluster wide encryption shouldn't that have more precedence over what the individual devicseSet is specifying?

malayparida2000: Added the new logic as discussed in the call

controllers/storagecluster/cephobjectstores.go
malayparida2000: As discussed this and other places where just KMS information is being added we don't need to respect the deviecSet setting first.

controllers/storagecluster/cephcluster.go
```diff
@@ -167,10 +167,11 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
...
 	} else {
 		// Add KMS details to CephCluster spec, only if
-		// cluster-wide encryption is enabled
-		// ie, sc.Spec.Encryption.ClusterWide/sc.Spec.Encryption.Enable is True
+		// cluster-wide encryption is enabled or any of the device set is encrypted
+		// ie, sc.Spec.Encryption.ClusterWide/sc.Spec.Encryption.Enable is True or any device is encrypted
 		// and KMS ConfigMap is available
-		if sc.Spec.Encryption.Enable || sc.Spec.Encryption.ClusterWide {
+
+		if sc.Spec.Encryption.Enable || sc.Spec.Encryption.ClusterWide || util.IsAnyDeviceSetEncrypted(sc.Spec.StorageDeviceSets) {
```
travisn, line 174: This check for multiple conditions is repeated in several places. Could this be refactored to a helper method?

malayparida2000: Done

SanjalKatiyar:
cc @GowthamShanmugam @bipuladh FYI

parth-gr:
@malayparida2000 can we do a quick testing with these changes, 
At least on a LSO based cluster

As their will be a UI impact for LSO based clusters 

@SanjalKatiyar I have talked with Nikhil to support this in the UX

controllers/util/util.go
```diff
@@ -103,3 +104,19 @@ func AssertEqual[T comparable](actual T, expected T, exitCode int) {
...
+func IsClusterOrDeviceSetEncrypted(sc *ocsv1.StorageCluster) bool {
+	// If cluster-wide encryption is enabled
+	if sc.Spec.Encryption.Enable || sc.Spec.Encryption.ClusterWide {
+		return true
+	}
+
+	// If any device set is encrypted
+	for _, deviceSet := range sc.Spec.StorageDeviceSets {
+		if deviceSet.Encrypted != nil && *deviceSet.Encrypted {
+			return true
```
parth-gr, line 117: can we just `return *deviceSet.Encrypted` and do not check for &&

controllers/storagecluster/cephcluster.go
```diff
@@ -989,6 +990,13 @@ func countAndReplicaOf(ds *ocsv1.StorageDeviceSet) (int, int) {
 	return count, replica
 }
 
+func isDeviceSetToBeEncrypted(sc *ocsv1.StorageCluster, ds ocsv1.StorageDeviceSet) bool {
+	if ds.Encrypted != nil {
+		return *ds.Encrypted
+	}
+	return sc.Spec.Encryption.Enable || sc.Spec.Encryption.ClusterWide
```
SanjalKatiyar, line 997: if cluster is already marked for "clusterwide" encryption during day-1, do we still allow to override a particular future deviceSet(s) from being encrypted ?? cc @malayparida2000 @parth-gr

controllers/util/util.go
malayparida2000: That would be wrong as, If one deviceSet has false in the Encrypted field and the loop encounters that first it would return false from there & consequentially KMS details would not be added. But we expect to return true even if any one DeviceSet has Encrypted True so that KMS details are added.

controllers/storagecluster/cephcluster.go
parth-gr: Yes, we can change it when we are adding day-2 device sets,

But once encryption is set on a device set we cant change in 4.18

SanjalKatiyar: ack !

parth-gr: cc @SanjalKatiyar only we change for new device class device sets

And not possible for same device class device sets

SanjalKatiyar: cc @vbnrh FYI.

travisn (approved):

PR #2847, opened 2024-10-09
controller: update the odf-info to include client identification info

rewantsoni:
To uniquely indentify a storageConsumer from outside a cluster we need StorageClusterUID and ClientID (storageConsumerUID). Add this information to the odf-info configMap so that MCO or any other component can easily identify the consumer from these two information.


rewantsoni:
Failure due to Noobaa, re-testing 
/retest-required

malayparida2000:
/lgtm
As only rebased with upstream w/o any changes.

PR #2851, opened 2024-10-14
Clear the affinity on Noobaa if it was set previously to handle upgrades

malayparida2000:
We are adding affinity to Noobaa if not in noobaa-standalone mode or if placement is specified. But we are not clearing the affinity if the affinity was set previously. This is breaking upgrade.

https://bugzilla.redhat.com/show_bug.cgi?id=2314636

### DIFF
--- a/controllers/storagecluster/noobaa_system_reconciler.go
+++ b/controllers/storagecluster/noobaa_system_reconciler.go
@@ -161,10 +161,15 @@ func (r *StorageClusterReconciler) setNooBaaDesiredState(nb *nbv1.NooBaa, sc *oc
 	placement := getPlacement(sc, component)
 
 	nb.Spec.Tolerations = placement.Tolerations
-	// if we are "noobaa-standalone" and placement is not set - don't set affinity
+
 	if !r.IsNoobaaStandalone || ok {
+		// Add affinity if not in noobaa-standalone mode or if placement is specified
 		nb.Spec.Affinity = &corev1.Affinity{NodeAffinity: placement.NodeAffinity}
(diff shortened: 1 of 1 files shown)


Nikhil-Ladha (approved):
/lgtm

controllers/storagecluster/noobaa_system_reconciler.go
```diff
@@ -161,10 +161,15 @@ func (r *StorageClusterReconciler) setNooBaaDesiredState(nb *nbv1.NooBaa, sc *oc
...
 	nb.Spec.Tolerations = placement.Tolerations
-	// if we are "noobaa-standalone" and placement is not set - don't set affinity
+
 	if !r.IsNoobaaStandalone || ok {
+		// Add affinity if not in noobaa-standalone mode or if placement is specified
 		nb.Spec.Affinity = &corev1.Affinity{NodeAffinity: placement.NodeAffinity}
+	} else if nb.Spec.Affinity != nil {
+		// Clear the affinity if it was set previously to handle upgrades
+		nb.Spec.Affinity = nil
 	}
```
Nikhil-Ladha, line 171: Wondering, if we can simply set to nil, as anyway we would enter else only if it's non-standalone.
Hence, we can avoid an extra condition check here 🤔

malayparida2000: Yeah initially I had kept it like that, But I thought it might be complicated to understand later, so I kept the extra check for better understanding.

agarwal-mudit (approved):

PR #2853, opened 2024-10-15
change float32 to float64 in storageconsumer api

rchikatw:
- the reason for this change is native type size of our machine is 64bit


rchikatw:
@nb-ohad can i get review on this

PR #2856, opened 2024-10-17
services: remove Getter as they generated as part of proto

rchikatw:
- remove Getter as they are generated as part of proto
- not removing GetPlatformVersion() & GetOperatorVersion() as they are already used. Leela will take care of removing this.

### DIFF
--- a/services/provider/api/interfaces/interfaces.go
+++ b/services/provider/api/interfaces/interfaces.go
@@ -6,11 +6,6 @@ type StorageClientStatus interface {
 	// no client is using it
 	GetPlatformVersion() string
 	GetOperatorVersion() string
-	GetClusterID() string
-	GetClusterName() string
-	GetClientName() string
-	GetClientID() string
-	GetStorageQuotaUtilizationRatio() float64
 
 	SetPlatformVersion(string) StorageClientStatus
 	SetOperatorVersion(string) StorageClientStatus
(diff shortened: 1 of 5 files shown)


services/provider/server/consumer.go
```diff
@@ -207,7 +208,7 @@ func (c *ocsConsumerManager) Get(ctx context.Context, id string) (*ocsv1alpha1.S
 	return consumerObj, nil
 }
 
-func (c *ocsConsumerManager) UpdateConsumerStatus(ctx context.Context, id string, status ifaces.StorageClientStatus) error {
+func (c *ocsConsumerManager) UpdateConsumerStatus(ctx context.Context, id string, status *pb.ReportStatusRequest) error {
```
leelavg, line 211: why is this being changed? the whole point was to abstract ReportStatusRequest as it keeps on increasing and keep on satisfying StorageClientStatus interface.

services/provider/api/interfaces/interfaces.go
```diff
@@ -6,10 +6,6 @@ type StorageClientStatus interface {
 	// no client is using it
 	GetPlatformVersion() string
 	GetOperatorVersion() string
-	GetClusterID() string
-	GetClusterName() string
-	GetClientName() string
-	GetStorageQuotaUtilizationRatio() float64
```
leelavg, line 12: can you change the comment to
```
These two getters are redirects for generated protobuf rpcs' and only add new getters here if you want to use different names which is not encouraged as these two are here only for backward compatibility

you can find setters at services/provider/api/storageclient.go and generated getters at services/provider/api/provider.pb.go
```

rchikatw:
Based on latest discussion, these will remain unchanged. Therefore, I am closing the pull request.

PR #2861, opened 2024-10-17
exporter: add alert for storagequotautilizationration > 0.8

rchikatw:
- Trigger alert when quota utilization reaches 80%

### DIFF
--- a/metrics/deploy/prometheus-ocs-rules.yaml
+++ b/metrics/deploy/prometheus-ocs-rules.yaml
@@ -354,6 +354,16 @@ spec:
         floor((ocs_storage_provider_operator_version>0)/1000) - ignoring(storage_consumer_name) group_right() floor((ocs_storage_client_operator_version>0)/1000) > 1 or floor((ocs_storage_client_operator_version>0)/1000) - ignoring(storage_consumer_name) group_left() floor((ocs_storage_provider_operator_version>0)/1000) >= 1
       labels:
         severity: critical
+    - alert: StorageQuotaUtilizationThresholdReached
+      annotations:
+        description: ODF client ({{ $labels.client_name }}) - ({{ $labels.client_cluster_name }}) storage consumption has surpassed 80% of the configured quota, this might result in issues in the creation of new PV/PVCs on the target cluster.
+        message: ODF client ({{ $labels.odf_client_name }}) storage consumption has surpassed 80% of the configured quota, this might result in issues in the creation of new PV/PVCs on the target cluster.
+        runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/openshift-container-storage-operator/StorageQuotaUtilizationThresholdReached.md
+        severity_level: warning
+      expr: |
+        ocs_storage_client_storage_quota_utilization_ratio > 0.8
+      labels:
+        severity: warning
   - name: storage-cluster-alerts.rules
     rules:
     - alert: CephMonLowNumber
--- a/metrics/internal/collectors/storageconsumer.go
+++ b/metrics/internal/collectors/storageconsumer.go
@@ -19,12 +19,13 @@ import (
 var _ prometheus.Collector = &StorageConsumerCollector{}
 
 type StorageConsumerCollector struct {
-	Informer                cache.SharedIndexInformer
-	StorageConsumerMetadata *prometheus.Desc
-	LastHeartbeat           *prometheus.Desc
-	ProviderOperatorVersion *prometheus.Desc
-	ClientOperatorVersion   *prometheus.Desc
-	AllowedNamespace        string
+	Informer                     cache.SharedIndexInformer
+	StorageConsumerMetadata      *prometheus.Desc
+	LastHeartbeat                *prometheus.Desc
+	StorageQuotaUtilizationRatio *prometheus.Desc
+	ProviderOperatorVersion      *prometheus.Desc
+	ClientOperatorVersion        *prometheus.Desc
+	AllowedNamespace             string
 }
 
 func NewStorageConsumerCollector(opts *options.Options) *StorageConsumerCollector {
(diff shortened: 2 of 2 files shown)


rchikatw:
@leelavg can i get review on this.

metrics/deploy/prometheus-ocs-rules.yaml
```diff
@@ -354,6 +354,16 @@ spec:
         floor((ocs_storage_provider_operator_version>0)/1000) - ignoring(storage_consumer_name) group_right() floor((ocs_storage_client_operator_version>0)/1000) > 1 or floor((ocs_storage_client_operator_version>0)/1000) - ignoring(storage_consumer_name) group_left() floor((ocs_storage_provider_operator_version>0)/1000) >= 1
       labels:
         severity: critical
+    - alert: StorageQuotaUtilizationThresholdReached
+      annotations:
+        description: Storage Client ({{ $labels.storage_consumer_name }}) has reached more than 80% of storage. Client might face issue while creating new pvc more than the quota allotted.
```
nb-ohad, line 359: 1. We don't call the clients "Storage Clients" but "ODF Clients" 
2. The storage consumer name is not a good identifier for anything, please use the client name instead, maybe in conjunction with the cluster name (all of these should be available on the consumer's status section)
3. Please use the terminology of consumption instead of reaching. Maybe something like
``` ODF client XXXXXX storage consumption has surpassed 80% of the configured quota, this might result in issues in the creation of new PV/PVCs on the target cluster```

File: metrics/deploy/prometheus-ocs-rules.yaml [hunk 2]
```diff
@@ -354,6 +354,16 @@ spec:
...
+        message: Storage Client ({{ $labels.storage_consumer_name }}) has reached more than 80% of storage. Client might face issue while creating new pvc more than the quota allotted.
```
nb-ohad, line 360: Same as the prev comment

metrics/internal/collectors/storageconsumer.go
```diff
@@ -60,6 +61,12 @@ func NewStorageConsumerCollector(opts *options.Options) *StorageConsumerCollecto
 			[]string{"storage_consumer_name"},
 			nil,
 		),
+		StorageQuotaUtilizationRatio: prometheus.NewDesc(
+			prometheus.BuildFQName("ocs", "storage_client", "storage_quota_utilization_ratio"),
+			`StorageQuotaUtilizationRatio of OCS Storage Client`,
```
nb-ohad, line 66:
```suggestion
			`StorageQuotaUtilizationRatio of ODF Storage Client`,
```

metrics/internal/collectors/storageconsumer.go
```diff
@@ -137,6 +145,10 @@ func (c *StorageConsumerCollector) collectStorageConsumersMetadata(storageConsum
 			prometheus.GaugeValue,
 			float64(encodeVersion(storageConsumer.Status.Client.OperatorVersion)),
 			storageConsumer.Name)
+
+		ch <- prometheus.MustNewConstMetric(c.StorageQuotaUtilizationRatio,
+			prometheus.GaugeValue, storageConsumer.Status.Client.StorageQuotaUtilizationRatio,
+			storageConsumer.Name)
```
nb-ohad, line 151: this is a bad identification for an alert, the consumer name is mainly a hash and we do not display it in the UI. Use the storage client name in conjunction with the cluster name for the user to be able to identify the client.

nb-ohad (changes requested):

metrics/internal/collectors/storageconsumer.go
```diff
@@ -60,6 +61,12 @@ func NewStorageConsumerCollector(opts *options.Options) *StorageConsumerCollecto
 			[]string{"storage_consumer_name"},
 			nil,
 		),
+		StorageQuotaUtilizationRatio: prometheus.NewDesc(
+			prometheus.BuildFQName("ocs", "storage_client", "storage_quota_utilization_ratio"),
+			`StorageQuotaUtilizationRatio of ODF Storage Client`,
+			[]string{"odf_client_name"},
```
aruniiird, line 67: from line#151, we understand that the `odf_client_name` label is populated by concatenating client_name and cluster_name. Question is why not we add two individual labels (something like "client_name" and "client_cluster_name") and populate them separately? (a multiple label example can be seen here with `StorageConsumerMetadata` description)

rchikatw: There is no specific reason for using multiple labels instead of using a single label. Do you suggest not to use concat?

aruniiird: No, nothing restricting us to add one single label concatenating multiple values (from other labels), if you prefer that way it is perfectly fine. Raised this question, because we have the option to add multiple labels individually and their combination determines the uniqueness (so why restrict our self by using a single label) (=

Another plus point (of using multiple labels) is about ease of understanding. For example,
 if you have three labels like `namespace`, `cluster-name` and `host-name`, it is much easier to understand the metric object with these three distinct labels, rather than adding a single label with all the values concatenated.

Above example with separate labels
```
Name                  namespace            host             clustername                 Value
metric-abc             my-ns           10.9.31.109         my_cluster_1                    1
```

Same example with concatenated metric label
```
Name                      my_unique_label                         Value
metric-abc          my_cluster_1_my-ns_10.9.31.109                  1
```

As a general thumb-rule, prefer the concatenated values if we are creating a new label by combining some internal values (which is not exposed to the user/admin), but try to create separate labels if the given labels are known ones (like namespace, cluster-name etc).

aruniiird (approved):
Thanks for the changes.
LGTM

umangachapagain (commented):
lgtm

umangachapagain (approved):

malayparida2000:
doing manual merge as prow seems to have glitched out on this PR.

PR #2863, opened 2024-10-18
scale down ocs-client-operator deployments in CI

iamniting:
In the ocs-operator CI, since the ODF operator is not running scale down the ocs-client-operator deployments via install script itself.


leelavg:
I relooked at the issues and raised #2865, ptal, thanks.

iamniting:
PR is not required anymore.

PR #2870, opened 2024-11-04
rpc: add rpc calls to create the maintenance mode reconciler in ocs-operator

rewantsoni:
During Failover action for DR, the ramen operator creates the MaintenanceMode CR on the Client clusters.
The MaintenancMode action requires us to scale down the rbd mirror daemon created on the provider.
Create rpc calls so that maintenance mode can be created on the providers and scale down the rbd mirror deployment


services/provider/proto/provider.proto
```diff
@@ -90,6 +93,12 @@ message StorageConfigResponse{
     repeated ExternalResource externalResource = 1;
     // Contains hash of desired config
     string desiredConfigHash = 2;
+
+    message StorageStatus {
+      string FailoverMaintenanceMode = 1;
```
rewantsoni, line 98: Should we have enum for FailoverMaintenanceMode?

controllers/util/k8sutil.go
```diff
@@ -53,7 +53,8 @@ const (
 	OdfInfoNamespacedNameClaimName      = "odfinfo.odf.openshift.io"
 	ExitCodeThatShouldRestartTheProcess = 42
 
-	BlockPoolMirroringTargetIDAnnotation = "ocs.openshift.io/mirroring-target-id"
+	BlockPoolMirroringTargetIDAnnotation     = "ocs.openshift.io/mirroring-target-id"
+	RequestFailoverMaintenanceModeAnnotation = "ocs.openshift.io/request-failover-maintenance-mode"
```
nb-ohad, line 57: At this point in time we decided not to differentiate between the modes and even if we do we will do that via the value of the annotation not the key
```suggestion
	RequestMaintenanceModeAnnotation = "ocs.openshift.io/request-maintenance-mode"
```

services/provider/api/client/client.go
```diff
@@ -227,3 +227,19 @@ func (cc *OCSProviderClient) PeerStorageCluster(ctx context.Context, onboardingT
 
 	return cc.Client.PeerStorageCluster(apiCtx, req)
 }
+
+func (cc *OCSProviderClient) RequestMaintenanceMode(ctx context.Context, consumerUUID string, requestFailover bool) (*pb.RequestMaintenanceModeResponse, error) {
```
nb-ohad, line 231: As agreed we should not use 'failover' terminology 
```suggestion
func (cc *OCSProviderClient) RequestMaintenanceMode(ctx context.Context, consumerUUID string, enable bool) (*pb.RequestMaintenanceModeResponse, error) {
```

services/provider/proto/provider.proto
```diff
@@ -90,6 +93,12 @@ message StorageConfigResponse{
     repeated ExternalResource externalResource = 1;
     // Contains hash of desired config
     string desiredConfigHash = 2;
+
+    message StorageStatus {
```
nb-ohad, line 97: StorageStatus is a bad name as it indicates that the only thing that gets here is a status and only about the storage  parts of the system, we need a message to hold general/global information that we want to pass

Maybe use `SystemAttributes` as the message name? @rewantsoni WDYT?

services/provider/proto/provider.proto
nb-ohad, line 98: As agreed in our discussion, we should not use Failover or any specific reason for the MM at this point. 
I would say this need to be something like `bool SystemInMaintannceMode` or something similar

services/provider/server/server.go
```diff
@@ -1079,3 +1090,54 @@ func (s *OCSProviderServer) PeerStorageCluster(ctx context.Context, req *pb.Peer
...
+		util.AddAnnotation(consumerObj, util.RequestFailoverMaintenanceModeAnnotation, "")
+	} else {
+		delete(consumerObj.GetAnnotations(), util.RequestFailoverMaintenanceModeAnnotation)
+	}
+
+	err = s.consumerManager.Update(ctx, consumerObj)
+	if err != nil {
+		klog.Error(err)
+		return nil, status.Errorf(codes.Internal, "Failed to update storageConsumer resource")
+	}
```
nb-ohad, line 1112: Should we always update? even if there is no change?

services/provider/server/server.go
```diff
@@ -1079,3 +1090,54 @@ func (s *OCSProviderServer) PeerStorageCluster(ctx context.Context, req *pb.Peer
...
+
+	return &pb.RequestMaintenanceModeResponse{}, nil
+}
+
+func (s *OCSProviderServer) getMaintenanceModeStatus(ctx context.Context) string {
```
nb-ohad, line 1117: What is this for?

nb-ohad (changes requested):

services/provider/proto/provider.proto
rewantsoni: Yes, `SystemAttributes` would a better suited name

services/provider/server/server.go
rewantsoni: This function is used to get if system is in MaintenanceMode, updated the name and return bool

services/provider/server/server.go
rewantsoni: No, added logic to update when there is change

nb-ohad (changes requested):

services/provider/server/server.go
```diff
@@ -1079,3 +1083,51 @@ func (s *OCSProviderServer) PeerStorageCluster(ctx context.Context, req *pb.Peer
...
+			return nil, status.Errorf(codes.Internal, "Failed to update storageConsumer resource")
+		}
+	}
+
+	return &pb.RequestMaintenanceModeResponse{}, nil
+}
+
+func (s *OCSProviderServer) isSystemInMaintenanceMode(ctx context.Context) bool {
+	// TODO: Refactor to use presence of cephRBDMirror CR
+	// Looking for the rbd mirror deployment until rook ensures that the deployment is deleted
```
rewantsoni, line 1115: Opened an issue for this in Rook https://github.com/rook/rook/issues/15031

nb-ohad: Thanks

PR #2871, opened 2024-11-04
provider: fix issue in cluster resource quota

rchikatw:
- Creating a resource of cluster resource quota on the client side does not limit storage usage for the default namespace by default.
- Typically, application users do not have access to the default namespace; however, if an application admin attempts to create a Persistent Volume Claim (PVC) in the default namespace, the quota does not restrict this action.
- BZ for this: https://bugzilla.redhat.com/show_bug.cgi?id=2318045

### DIFF
--- a/controllers/storagecluster/prometheus/localcephrules.yaml
+++ b/controllers/storagecluster/prometheus/localcephrules.yaml
@@ -11,10 +11,10 @@ spec:
   - name: ceph.rules
     rules:
     - expr: |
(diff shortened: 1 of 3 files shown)


rchikatw:
@nb-ohad @leelavg please review

leelavg:
@rchikatw pls expand on description as it says what was happening but doesn't mention what this PR does to fix that?

The PR seems to be trying to not track consumption in namespace with name "default", shouldn't the fix for the BZ be an addition in documentation that PVC's in "default" namespace isn't counted towards quota calculations? Internally, an explanation why this can't be done in "default" namespace be added to BZ, wdyt?

PR #2865, opened 2024-10-21
remove dependency of ocs-client-operator CRDs during manager setup

leelavg:
use the recently introduced CRD checks feature for removing dependency of ocs-client-operator CRDs by storagecluster controller.

PR #2872, opened 2024-11-04
Fix broken watch for CRDs on storageCluster controller

malayparida2000:
The watch had builder.WithPredicates twice, which was causing the watch to not work as expected. Thus even after the CRD was created the controller was not able to detect it and queue a reconcile. Now the predicates are combined under one builder.WithPredicates.


leelavg (approved):
Thanks for the fix, bug introduced in #2865. My testing might have worked due to a side effect. Needs a backport I guess?

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -235,8 +235,6 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			builder.WithPredicates(
```
leelavg, line 235: or should this use `util.ComposePredicates`? I don't remember does the default one allows any match or all matches.

malayparida2000:
https://issues.redhat.com/browse/DFBUGS-726,
@ShravaniVangur is now looking into the Bug, if necessary she would issue a fix.

PR #2874, opened 2024-11-04
provider: enable csi omap generator for client

rewantsoni:
send a configMap from provider to client to enable csi-omap generator required for DR on client cluster

### DIFF
--- a/controllers/storageclusterpeer/storageclusterpeer_controller.go
+++ b/controllers/storageclusterpeer/storageclusterpeer_controller.go
@@ -27,6 +27,10 @@ import (
 	v1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
 )
 
+const (
+	OcsClientPeerConfigMapName = "ocs-client-peer-mapping"
+)
+
 // StorageClusterPeerReconciler reconciles a StorageClusterPeer object
 // nolint:revive
 type StorageClusterPeerReconciler struct {
--- a/services/provider/server/server.go
+++ b/services/provider/server/server.go
@@ -11,6 +11,7 @@ import (
 	"encoding/json"
 	"encoding/pem"
 	"fmt"
+	"github.com/red-hat-storage/ocs-operator/v4/controllers/storageclusterpeer"
 	"math"
 	"net"
 	"slices"
@@ -224,10 +225,20 @@ func (s *OCSProviderServer) GetStorageConfig(ctx context.Context, req *pb.Storag
 			return nil, err
 		}
 
+		clientMappingConfig, err := s.getPeerClientMapping(ctx)
+		if err != nil {
+			return nil, err
+		}
+		clientIsMapped := false
+		if clientMappingConfig != nil && clientMappingConfig[req.StorageConsumerUUID] != "" {
+			clientIsMapped = true
+		}
+
 		desiredClientConfigHash := getDesiredClientConfigHash(
 			channelName,
 			consumerObj,
 			isEncryptionInTransitEnabled(storageCluster.Spec.Network),
+			clientIsMapped,
 		)
 
 		klog.Infof("successfully returned the config details to the consumer.")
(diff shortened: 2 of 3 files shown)


services/provider/server/server.go
```diff
@@ -404,6 +410,24 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
+	if err != nil {
+		return nil, err
+	}
+	enableOmap := "false"
+	if len(storageClusterPeers.Items) > 0 {
+		enableOmap = "true"
+	}
+
+	extR = append(extR, &pb.ExternalResource{
+		Name: "ceph-csi-sidecar-config",
```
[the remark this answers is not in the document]
rewantsoni: how does `ceph-csi-config` sound?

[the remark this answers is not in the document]
Madhu-1: That is already used in cephcsi oeprator, you need to make it client+cephcsi `ocs-ceph-csi-config` something like that

Madhu-1 (commented):
LGTM

controllers/storageclusterpeer/storageclusterpeer_controller.go
```diff
@@ -27,6 +27,10 @@ import (
 	v1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
 )
 
+const (
+	ClientMappingConfigMapName = "client-mapping-config"
```
nb-ohad, line 31: 'Client' and 'Mapping' are very generic terms when you consider all of ODF and not just RDR. Please choose a better name for the config map and the var

services/provider/server/server.go
```diff
@@ -224,10 +225,17 @@ func (s *OCSProviderServer) GetStorageConfig(ctx context.Context, req *pb.Storag
 			return nil, err
 		}
 
+		clientMappingConfig, err := s.getClientMapping(ctx)
+		if err != nil {
+			return nil, err
+		}
+		_, hasClientID := clientMappingConfig.Data[req.StorageConsumerUUID]
```
nb-ohad, line 232: I remember the design was a mapping of client names not consumer uuids

services/provider/server/server.go
```diff
@@ -224,10 +225,17 @@ func (s *OCSProviderServer) GetStorageConfig(ctx context.Context, req *pb.Storag
 			return nil, err
 		}
 
+		clientMappingConfig, err := s.getClientMapping(ctx)
+		if err != nil {
+			return nil, err
+		}
```
nb-ohad, line 231: why would we fail this function if there is not storage client id mapping config?

services/provider/server/server.go
```diff
@@ -418,6 +426,21 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
 
 	}
 
+	//Enable CSI-OMAP is storageClusterPeer is present
+	clientMappingConfig, err := s.getClientMapping(ctx)
+	if err != nil {
+		return nil, err
+	}
```
nb-ohad, line 433: Same as last comment

services/provider/server/server.go
```diff
@@ -418,6 +426,21 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
+	_, hasClientID := clientMappingConfig.Data[string(consumerResource.UID)]
```
nb-ohad, line 434: This is not good enough, what if the mapping is broken? For example the key exists but not a value
I would suggest doing this instead

```suggestion
	clientIsMapped := clientMappingConfig.Data[string(consumerResource.UID)] != ""
```

services/provider/server/server.go
```diff
@@ -418,6 +426,21 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
+
+	extR = append(extR, &pb.ExternalResource{
+		Name: "ocs-ceph-csi-config",
+		Kind: "ConfigMap",
+		Data: mustMarshal(map[string]string{
+			"CSI_ENABLE_OMAP_GENERATOR": strconv.FormatBool(hasClientID),
+		}),
+	})
```
nb-ohad, line 442: What is this ocs-ceph-csi-config? I believe that ceph csi operator is using a CR for that

services/provider/server/server.go
```diff
@@ -910,6 +941,17 @@ func (s *OCSProviderServer) getOCSSubscriptionChannel(ctx context.Context) (stri
 	return subscription.Spec.Channel, nil
 }
 
+func (s *OCSProviderServer) getClientMapping(ctx context.Context) (*corev1.ConfigMap, error) {
```
nb-ohad, line 944: What return a config map and not just `map[string]string`. You are not interested in the resource, just in the mapping.

Also, please rename this func to something more specific

nb-ohad (changes requested):

services/provider/server/server.go
rewantsoni: From provider we are sending the storageConsumerUIDs, and MCO updates this configmap with local ClientID to remote ClientID

services/provider/server/server.go
rewantsoni: This function is failing in cases expect isNotFound

services/provider/server/server.go
rewantsoni: This function is failing in cases expect isNotFound

rewantsoni:
Closing this as we pivoted from the design

PR #2875, opened 2024-11-04
rpc: add new rpc call for mirroring

rewantsoni:
Add two rpc calls required for setting up mirroring for blockpool and radosnamespace
1. GetBlockPoolsInfo
2. GetClientInfo


services/provider/server/server.go
```diff
@@ -52,11 +52,12 @@ import (
...
-	ProviderCertsMountPoint   = "/mnt/cert"
-	onboardingTicketKeySecret = "onboarding-ticket-key"
-	storageRequestNameLabel   = "ocs.openshift.io/storagerequest-name"
-	notAvailable              = "N/A"
+	TicketAnnotation                 = "ocs.openshift.io/provider-onboarding-ticket"
+	ProviderCertsMountPoint          = "/mnt/cert"
+	onboardingTicketKeySecret        = "onboarding-ticket-key"
+	storageRequestNameLabel          = "ocs.openshift.io/storagerequest-name"
+	notAvailable                     = "N/A"
+	rbdMirrorBootstrapPeerSecretName = "rbdMirrorBootstrapPeerSecretName"
```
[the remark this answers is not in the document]
rewantsoni: After mirroring is enabled, the mirroringToken is stored and saved in a k8s secret, the name of the secret is specified in the blockpool status with a key. See [here](https://github.com/rook/rook/blob/7822e6b19952a2a93eaeead225fb42570c196741/Documentation/CRDs/Block-Storage/ceph-block-pool-crd.md?plain=1#L116-L120)

services/provider/server/server.go
```diff
@@ -966,3 +967,95 @@ func (s *OCSProviderServer) PeerStorageCluster(ctx context.Context, req *pb.Peer
...
+
+func (s *OCSProviderServer) GetBlockPoolsInfo(ctx context.Context, req *pb.GetBlockPoolsInfoRequest) (*pb.GetBlockPoolsInfoResponse, error) {
+	storageCluster, err := util.GetStorageClusterInNamespace(ctx, s.client, s.namespace)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to get storageCluster. %v", err)
+	}
+
+	if storageCluster.UID != types.UID(req.StorageClusterUID) {
+		return nil, status.Errorf(codes.InvalidArgument, "storageClusterUID specified on the req does not match any existing storage cluster")
+	}
```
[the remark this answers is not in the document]
rewantsoni: In the future when we move to ocs-provider deployed in operator namespace, we need to send the storageClusterUID and verify the storageCluster exists, but yes we don't need to load the storagecluster, we could have a util function for it

services/provider/server/server.go
```diff
@@ -1042,3 +1043,143 @@ func (s *OCSProviderServer) PeerStorageCluster(ctx context.Context, req *pb.Peer
...
+			klog.Errorf("failed to get consumer with id %v: %v", req.ClientIDs[i], err)
+			clientError = append(clientError,
+				&pb.StorageClientInfoError{
+					ClientID: req.ClientIDs[i],
+					Code:     pb.ErrorCode_Internal,
+					Message:  err.Error(),
+				},
+			)
+			continue
+		}
```
nb-ohad, line 1063: Why is not found an error?

services/provider/server/server.go
```diff
@@ -1042,3 +1043,143 @@ func (s *OCSProviderServer) PeerStorageCluster(ctx context.Context, req *pb.Peer
 
 	return &pb.PeerStorageClusterResponse{}, nil
 }
+
+func (s *OCSProviderServer) GetStorageClientsInfo(ctx context.Context, req *pb.StorageClientsInfoRequest) (*pb.StorageClientsInfoResponse, error) {
+	var clientsInfo []*pb.ClientInfo
+	var clientError []*pb.StorageClientInfoError
```
nb-ohad, line 1049: Why do we need these? why not fill in the response directly?

services/provider/server/server.go
```diff
@@ -1042,3 +1043,143 @@ func (s *OCSProviderServer) PeerStorageCluster(ctx context.Context, req *pb.Peer
...
+			klog.Errorf("failed to find StorgeCluster owning the StorageConsumer %v", req.ClientIDs[i])
+			clientError = append(clientError,
+				&pb.StorageClientInfoError{
+					ClientID: req.ClientIDs[i],
+					Code:     pb.ErrorCode_Internal,
+					Message:  "failed to find StorgeCluster owning the StorageConsumer",
+				},
+			)
+			continue
+		}
```
nb-ohad, line 1076: Why is this an error? this is equivalent to not found

services/provider/server/server.go
```diff
@@ -1042,3 +1043,143 @@ func (s *OCSProviderServer) PeerStorageCluster(ctx context.Context, req *pb.Peer
...
+
+		if owner.UID != types.UID(req.StorageClusterUID) {
+			klog.Errorf("storageCluster specified on the req does not own the Client")
+			clientError = append(clientError,
+				&pb.StorageClientInfoError{
+					ClientID: req.ClientIDs[i],
+					Code:     pb.ErrorCode_InvalidArgument,
```
nb-ohad, line 1083: Where are these codes coming form?

services/provider/server/server.go
```diff
@@ -1042,3 +1043,143 @@ func (s *OCSProviderServer) PeerStorageCluster(ctx context.Context, req *pb.Peer
...
+			continue
+		}
+
+		rnsList := &rookCephv1.CephBlockPoolRadosNamespaceList{}
+		err = s.client.List(
+			ctx,
+			rnsList,
+			client.InNamespace(s.namespace),
+			client.MatchingLabels{controllers.StorageConsumerNameLabel: consumer.Name},
+			client.Limit(1),
```
nb-ohad, line 1096: You cannot limit to 1, limit to 2, and make sure you only get 1. A consumer might have 2 (or more) if someone created a second (or a third)  claim

services/provider/server/server.go
```diff
@@ -1042,3 +1043,143 @@ func (s *OCSProviderServer) PeerStorageCluster(ctx context.Context, req *pb.Peer
...
+			klog.Error(err)
+			clientError = append(clientError,
+				&pb.StorageClientInfoError{
+					ClientID: req.ClientIDs[i],
+					Code:     pb.ErrorCode_Internal,
+					Message:  "rados namespace for the client not found",
+				},
+			)
+			continue
+		}
```
nb-ohad, line 1108: Not found is not considered an error. it just means we cannot return a rados namespace name

services/provider/server/server.go
```diff
@@ -1042,3 +1043,143 @@ func (s *OCSProviderServer) PeerStorageCluster(ctx context.Context, req *pb.Peer
...
+			klog.Errorf("no rados namespace found for StorageClient %s", req.ClientIDs[i])
+			clientError = append(clientError,
+				&pb.StorageClientInfoError{
+					ClientID: req.ClientIDs[i],
+					Code:     pb.ErrorCode_Internal,
+					Message:  "rados namespace for the client not found: %v",
+				},
+			)
+			continue
+		}
```
nb-ohad, line 1119: Same as last comment.

services/provider/server/server.go
```diff
@@ -1042,3 +1043,143 @@ func (s *OCSProviderServer) PeerStorageCluster(ctx context.Context, req *pb.Peer
...
+			klog.Error(err)
+			blockPoolError = append(blockPoolError,
+				&pb.BlockPoolInfoError{
+					BlockPoolName: cephBlockPool.Name,
+					Code:          pb.ErrorCode_Internal,
+					Message:       "failed to get CephBlockPool",
+				},
+			)
+			continue
+		}
```
nb-ohad, line 1145: don't group not found with the other errors

services/provider/server/server.go
nb-ohad, line 1: I am adding a general note here.
Not finding a requested resource or information for a resource on the provider should not be considered an error when it comes to a batch operation.

nb-ohad (changes requested):

services/provider/server/server.go
rewantsoni: The initiator for this call is requesting the clinetInfo for the clientID, if the client is not present we should return an error to let the initiator of this call know that there is an error fetch clientInfo

services/provider/server/server.go
rewantsoni: I didn't understand how we could fill in the response directly. We need to somehow aggregate it before passing it in the response

services/provider/server/server.go
rewantsoni: This is an error because for the clientID in the req there was no storageCluster owning the consumer. So I am aggregating it into the error Info struct and then sending it back at the end. It will continue reconcile for other clientIDs

services/provider/server/server.go
rewantsoni: I have defined them in the proto

services/provider/server/server.go
rewantsoni: This scenario is not supported right? Do we want to restrict from code if there is an additional radosnamespace?

services/provider/server/server.go
rewantsoni: Here, there is an error fetching the radosnamespace, maybe I need to change the error Message

services/provider/server/server.go
rewantsoni: If it is not found, since the default value of the string is "", it will always return something. If we don't want to return an error then, we should make radosnamespace field a pointer Have something below this in clientInfo or from the initiator consider "" as an invalid response and requeue
```
message raodsnamespace{
  value = 1;
}
message clientInfo{
  string ClientID
  raodsnamespace raodsnamespace
}
```

services/provider/server/server.go
nb-ohad: If a case is not supported then it is an error, but you should not just ignore the fact it can exists

services/provider/server/server.go
nb-ohad: If a client has not issued a rados namespace yet (because the other controller is broken or just delayed) it does not mean the client does not exists or that we should not return info for it

services/provider/server/server.go
nb-ohad: see reply here: https://github.com/red-hat-storage/ocs-operator/pull/2875#discussion_r1838102551

services/provider/server/server.go
```diff
@@ -1042,3 +1043,130 @@ func (s *OCSProviderServer) PeerStorageCluster(ctx context.Context, req *pb.Peer
...
+		if len(rnsList.Items) > 1 {
+			response.Error = append(response.Error,
+				&pb.StorageClientInfoError{
+					ClientID: req.ClientIDs[i],
+					Code:     pb.ErrorCode_Internal,
+					Message:  "invalid number of radosnamespace found for the Client",
+				},
+			)
```
nb-ohad, line 1109: Missing a continue here (with a log message)

services/provider/server/server.go
```diff
@@ -1042,3 +1043,130 @@ func (s *OCSProviderServer) PeerStorageCluster(ctx context.Context, req *pb.Peer
...
+		} else if len(rnsList.Items) == 1 {
+			response.ClientsInfo = append(response.ClientsInfo, &pb.ClientInfo{ClientID: req.ClientIDs[i], RadosNamespace: rnsList.Items[0].Name})
+		} else {
+			response.ClientsInfo = append(response.ClientsInfo, &pb.ClientInfo{ClientID: req.ClientIDs[i]})
+		}
```
nb-ohad, line 1114: 1. There is no need to have 2 append statements
2. You should treat the "good" case as a for loop scoped code

```suggestion
		} 

		clientInfo := pb.ClientInfo{ClientID: req.ClientIDs[i]
		if len(len(rnsList.Items) != 0 {
			clientInfo.RadosNamesapce = rnsList[0].Name
		}
		response.ClientsInfo = append(response.ClientsInfo, &clientInfo)
	
```

services/provider/server/server.go
```diff
@@ -1042,3 +1043,130 @@ func (s *OCSProviderServer) PeerStorageCluster(ctx context.Context, req *pb.Peer
...
+				response.Error = append(response.Error,
+					&pb.BlockPoolInfoError{
+						BlockPoolName: cephBlockPool.Name,
+						Code:          pb.ErrorCode_Internal,
+						Message:       err.Error(),
+					},
+				)
+			}
+			continue
+		}
```
nb-ohad, line 1139: No need for nesting and please log a message into the log for a continue

services/provider/server/server.go
```diff
@@ -1042,3 +1043,130 @@ func (s *OCSProviderServer) PeerStorageCluster(ctx context.Context, req *pb.Peer
...
+			secret.Namespace = s.namespace
+			err := s.client.Get(ctx, client.ObjectKeyFromObject(secret), secret)
+			if err != nil {
+				errMsg := fmt.Sprintf(
+					"failed to get bootstrap secret %s for CephBlockPool %s: %v",
+					cephBlockPool.Status.Info[mirroringTokenKey],
+					cephBlockPool.Name,
+					err,
+				)
+				klog.Error(errMsg)
```
nb-ohad, line 1157: Not finding the secret should not be treated as an error from the perspective of the RPC call. It might be an error for a controller but not for the server

nb-ohad (changes requested):

services/provider/server/server.go
```diff
@@ -1124,6 +1125,134 @@ func (s *OCSProviderServer) RequestMaintenanceMode(ctx context.Context, req *pb.
...
+		if kerrors.IsNotFound(err) {
+			klog.Infof("blockpool %v not found", cephBlockPool.Name)
+			continue
+		} else if err != nil {
+			klog.Errorf("failed to get blockpool %v: %v", cephBlockPool.Name, err)
+			response.Errors = append(response.Errors,
+				&pb.BlockPoolInfoError{
+					BlockPoolName: cephBlockPool.Name,
+					Code:          pb.ErrorCode_Internal,
+					Message:       err.Error(),
```
nb-ohad, line 1215: Let's keep the internal of the issue on the provider side
```suggestion
					Message:       "failed loading block pool information",
```

services/provider/server/server.go
```diff
@@ -1124,6 +1125,134 @@ func (s *OCSProviderServer) RequestMaintenanceMode(ctx context.Context, req *pb.
...
+				},
+			)
+		}
+
+		var mirroringToken string
+
```
nb-ohad, line 1221: There is no point in adding a space here
```suggestion
```

nb-ohad (changes requested):

PR #2751, opened 2024-08-19
Mark enableCrushUpdates as true for all the pools created

malayparida2000:
With recent changes in rook, without this flag rook operator would not
be able to update crush rules by default. This was giving rise
to the issue again where the mgr logs would complain about
"won't scale due to overlapping roots" & pool's PG won't scale beyond 1.

PR #2876, opened 2024-11-05
ux: add a new expandStorage handler

parth-gr:
update the ux backend server to use new expandstorage handler

For end to end UI testing image used for backend server: http://quay.io/paarora/rook-ceph:v637

UI will send the form 
```
type AttachStorageForm = {
  storageClassForOSDs: string
  enableEncryption: boolean
  storage: "1"
  replica: int
  count: int
  strogaeClusterName: string
  poolDetails: { 
    volumeType: 'Filesystem' | 'Block'
    poolName: string
    dataProtectionPolicy: '2Way' | '3Way'
    enableCompression: boolean
    filesystemName: string
    failureDomain: string 
  }
  storageClassDetails: {
    reclaimPolicy: 'Retain' | 'Delete'
    name: string 
    volumeBindingMode: 'Immediate' | 'WaitForFirstConsumer'
    enableStorageClassEncryption : bool
    encryptionKMSID: string
  }
}
```

### DIFF
--- a/services/ux-backend/handlers/expandstorage/handler.go
+++ b/services/ux-backend/handlers/expandstorage/handler.go
@@ -1,0 +1,298 @@
+package expandstorage
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"strconv"
+
+	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
+	storagecluster "github.com/red-hat-storage/ocs-operator/v4/controllers/storagecluster"
+	"github.com/red-hat-storage/ocs-operator/v4/services/ux-backend/handlers"
+	cephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
+	corev1 "k8s.io/api/core/v1"
+	storagev1 "k8s.io/api/storage/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func HandleMessage(w http.ResponseWriter, r *http.Request, client client.Client, namespace string) {
+	switch r.Method {
+	case "POST":
+		handlePost(w, r, client, namespace)
+	default:
+		handleUnsupportedMethod(w, r)
+	}
+}
+
+func handlePost(w http.ResponseWriter, r *http.Request, client client.Client, namespace string) {
+	// When ContentLength is 0 that means request body is empty
+	var err error
+	if r.ContentLength == 0 {
+		klog.Errorf("body in the request is required")
+		http.Error(w, "body in the request is required", http.StatusBadRequest)
+		return
+	}
+
+	type poolDetails struct {
+		VolumeType           string `json:"volumeType"`
+		PoolName             string `json:"poolName"`
+		DataProtectionPolicy int    `json:"dataProtectionPolicy"`
+		EnableCompression    bool   `json:"enableCompression"`
+		FilesystemName       string `json:"filesystemName"`
+		FailureDomain        string `json:"failureDomain"`
+	}
+
+	type storageClassDetails struct {
+		ReclaimPolicy                string `json:"reclaimPolicy"`
+		Name                         string `json:"name"`
+		VolumeBindingMode            string `json:"volumeBindingMode"`
+		EnableStorageClassEncryption bool   `json:"enableStorageClassEncryption"`
+		EncryptionKMSID              string `json:"encryptionKMSID"`
+	}
+
+	var ExpandStorage = struct {
+		StorageClassForOSDs string              `json:"storageClassForOSDs"`
+		EnableEncryption    bool                `json:"enableEncryption"`
+		Storage             string              `json:"storage"`
+		Replica             int                 `json:"replica"`
+		Count               int                 `json:"count"`
+		StorageClusterName  string              `json:"storageClusterName"`
+		PoolDetails         poolDetails         `json:"poolDetails"`
+		StorageClassDetails storageClassDetails `json:"storageClassDetails"`
+	}{}
+
+	if err = json.NewDecoder(r.Body).Decode(&ExpandStorage); err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	storageCluster := &ocsv1.StorageCluster{}
+	err = client.Get(r.Context(), types.NamespacedName{Name: ExpandStorage.StorageClusterName, Namespace: namespace}, storageCluster)
+	if err != nil {
+		klog.Errorf("failed to get storageCluster: %v", err)
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	// update storageCluster
+	updateStorageCluster(w, r, client, ExpandStorage.StorageClassForOSDs, ExpandStorage.EnableEncryption, ExpandStorage.Storage, ExpandStorage.Count, ExpandStorage.Replica, storageCluster)
+
+	if ExpandStorage.PoolDetails.VolumeType == "block" {
+		klog.Info("Creating cephBlockPool and storageClass")
+		// create cephBlockPool
+		createCephBlockPool(w, r, client, ExpandStorage.PoolDetails.PoolName, ExpandStorage.StorageClassForOSDs, ExpandStorage.PoolDetails.DataProtectionPolicy, ExpandStorage.PoolDetails.EnableCompression, namespace, ExpandStorage.PoolDetails.FailureDomain, storageCluster.Spec.Arbiter.Enable)
+
+		// create storageClass
+		createCephBlockPoolStorageClass(w, r, client, ExpandStorage.StorageClassDetails.Name, ExpandStorage.PoolDetails.PoolName, ExpandStorage.StorageClassDetails.ReclaimPolicy, ExpandStorage.StorageClassDetails.VolumeBindingMode, ExpandStorage.StorageClassDetails.EnableStorageClassEncryption, ExpandStorage.StorageClassDetails.EncryptionKMSID, namespace)
+	} else if ExpandStorage.PoolDetails.VolumeType == "filesystem" {
+		klog.Info("Creating cephFilesystem dataPool and storageClass")
+		// create cephFilesystem dataPool
+		createCephFilesystemDataPool(w, r, client, ExpandStorage.PoolDetails.PoolName, ExpandStorage.StorageClassForOSDs, ExpandStorage.PoolDetails.DataProtectionPolicy, ExpandStorage.PoolDetails.EnableCompression, ExpandStorage.PoolDetails.FailureDomain, storageCluster)
+
+		// create storageClass
+		createCephFilesystemStorageClass(w, r, client, ExpandStorage.StorageClassDetails.Name, ExpandStorage.PoolDetails.PoolName, ExpandStorage.StorageClassDetails.ReclaimPolicy, ExpandStorage.StorageClassDetails.VolumeBindingMode, ExpandStorage.PoolDetails.FilesystemName, namespace)
+	} else {
+		klog.Errorf("invalid volumeType: %s", ExpandStorage.PoolDetails.VolumeType)
+		http.Error(w, "invalid volumeType", http.StatusBadRequest)
+		return
+	}
+
+}
+
+func updateStorageCluster(w http.ResponseWriter, r *http.Request, client client.Client, storageClassForOSDs string, enableEncryption bool, storage string, count, replica int, storageCluster *ocsv1.StorageCluster) {
+	klog.Infof("Updating storageCluster %q", storageCluster.Name)
+	storageQty := resource.MustParse(storage)
+	volumeMode := corev1.PersistentVolumeBlock
+	deviceSet := ocsv1.StorageDeviceSet{
+		Name:        storageClassForOSDs,
+		Count:       count,
+		Replica:     replica,
+		Portable:    false,
+		Encrypted:   &enableEncryption,
+		DeviceClass: storageClassForOSDs,
+		DataPVCTemplate: corev1.PersistentVolumeClaim{
+			Spec: corev1.PersistentVolumeClaimSpec{
+				AccessModes:      []corev1.PersistentVolumeAccessMode{"ReadWriteOnce"},
+				StorageClassName: &storageClassForOSDs,
+				VolumeMode:       &volumeMode,
+				Resources: corev1.VolumeResourceRequirements{
+					Requests: corev1.ResourceList{
+						corev1.ResourceStorage: storageQty,
+					},
+				},
+			},
+		},
+	}
+
+	storageCluster.Spec.StorageDeviceSets = append(storageCluster.Spec.StorageDeviceSets, deviceSet)
+	err := client.Update(r.Context(), storageCluster)
+	if err != nil {
+		klog.Errorf("failed to update storageCluster: %q", storageCluster.Name)
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+}
+
+func createCephBlockPool(w http.ResponseWriter, r *http.Request, client client.Client, poolName, storageClassForOSDs string, dataProtectionPolicy int, enableCompression bool, namespace, failureDomain string, arbiter bool) {
+	compression := "none"
+	if enableCompression {
+		compression = "aggressive"
+	}
+	replicasPerFailureDomain := 1
+	if arbiter {
+		replicasPerFailureDomain = 2
+	}
+
+	cephBlockPool := &cephv1.CephBlockPool{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      poolName,
+			Namespace: namespace,
+		},
+		Spec: cephv1.NamedBlockPoolSpec{
+			PoolSpec: cephv1.PoolSpec{
+				FailureDomain:  failureDomain,
+				EnableRBDStats: true,
+				DeviceClass:    storageClassForOSDs,
+				Replicated: cephv1.ReplicatedSpec{
+					Size:                     uint(dataProtectionPolicy),
+					RequireSafeReplicaSize:   true,
+					ReplicasPerFailureDomain: uint(replicasPerFailureDomain),
+				},
+				Parameters: map[string]string{
+					"compression_mode": compression,
+				},
+				EnableCrushUpdates: true,
+			},
+		},
+	}
+
+	err := client.Create(r.Context(), cephBlockPool)
+	if err != nil {
+		klog.Errorf("failed to create cephBlockPool: %v", err)
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+}
+
+func createCephBlockPoolStorageClass(w http.ResponseWriter, r *http.Request, client client.Client, storageClassName, poolName, reclaimPolicy, volumeBindingMode string, enableEncryption bool, encryptionKMSID, namespace string) {
+	allowVolumeExpansion := true
+	storageClass := &storagev1.StorageClass{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: storageClassName,
+			Annotations: map[string]string{
+				"description": "Provides RWO Filesystem volumes, and RWO and RWX Block volumes",
+				"reclaimspace.csiaddons.openshift.io/schedule": "@weekly",
+			},
+		},
+		Provisioner:       storagecluster.RbdDriverName,
+		ReclaimPolicy:     (*corev1.PersistentVolumeReclaimPolicy)(&reclaimPolicy),
+		VolumeBindingMode: (*storagev1.VolumeBindingMode)(&volumeBindingMode),
+		// AllowVolumeExpansion is set to true to enable expansion of OCS backed Volumes
+		AllowVolumeExpansion: &allowVolumeExpansion,
+		Parameters: map[string]string{
+			"clusterID":                 namespace,
+			"pool":                      poolName,
+			"imageFeatures":             "layering,deep-flatten,exclusive-lock,object-map,fast-diff",
+			"csi.storage.k8s.io/fstype": "ext4",
+			"imageFormat":               "2",
+			"encrypted":                 strconv.FormatBool(enableEncryption),
+			"csi.storage.k8s.io/provisioner-secret-name":            "rook-csi-rbd-provisioner",
+			"csi.storage.k8s.io/provisioner-secret-namespace":       namespace,
+			"csi.storage.k8s.io/node-stage-secret-name":             "rook-csi-rbd-node",
+			"csi.storage.k8s.io/node-stage-secret-namespace":        namespace,
+			"csi.storage.k8s.io/controller-expand-secret-name":      "rook-csi-rbd-provisioner",
+			"csi.storage.k8s.io/controller-expand-secret-namespace": namespace,
+		},
+	}
+	if enableEncryption {
+		storageClass.Parameters["encryptionKMSID"] = encryptionKMSID
+	}
+
+	err := client.Create(r.Context(), storageClass)
+	if err != nil {
+		klog.Errorf("failed to create storageClass: %v", err)
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+}
+
+func createCephFilesystemDataPool(w http.ResponseWriter, r *http.Request, client client.Client, poolName, storageClassForOSDs string, dataProtectionPolicy int, enableCompression bool, failureDomain string, storageCluster *ocsv1.StorageCluster) {
+	compression := "none"
+	if enableCompression {
+		compression = "aggressive"
+	}
+
+	datapool := cephv1.NamedPoolSpec{
+		Name: poolName,
+		PoolSpec: cephv1.PoolSpec{
+			FailureDomain: failureDomain,
+			DeviceClass:   storageClassForOSDs,
+			Replicated: cephv1.ReplicatedSpec{
+				Size:                   uint(dataProtectionPolicy),
+				RequireSafeReplicaSize: true,
+			},
+			Parameters: map[string]string{
+				"compression_mode": compression,
+			},
+			EnableCrushUpdates: true,
+		},
+	}
+
+	storageCluster.Spec.ManagedResources.CephFilesystems.AdditionalDataPools = append(storageCluster.Spec.ManagedResources.CephFilesystems.AdditionalDataPools, datapool)
+	err := client.Update(r.Context(), storageCluster)
+	if err != nil {
+		klog.Errorf("failed to update storageCluster: %q", storageCluster.Name)
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+}
+
+func createCephFilesystemStorageClass(w http.ResponseWriter, r *http.Request, client client.Client, storageClassName, poolName, reclaimPolicy, volumeBindingMode string, filesystemName, namespace string) {
+	allowVolumeExpansion := true
+	storageClass := &storagev1.StorageClass{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: storageClassName,
+			Annotations: map[string]string{
+				"description": "Provides RWO and RWX Filesystem volumes",
+			},
+		},
+		Provisioner:       storagecluster.CephFSDriverName,
+		ReclaimPolicy:     (*corev1.PersistentVolumeReclaimPolicy)(&reclaimPolicy),
+		VolumeBindingMode: (*storagev1.VolumeBindingMode)(&volumeBindingMode),
+		// AllowVolumeExpansion is set to true to enable expansion of OCS backed Volumes
+		AllowVolumeExpansion: &allowVolumeExpansion,
+		Parameters: map[string]string{
+			"clusterID": namespace,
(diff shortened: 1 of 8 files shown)


services/ux-backend/handlers/expandstorage/handler.go
```diff
@@ -0,0 +1,108 @@
...
+				FailureDomain: "host",
+				DeviceClass:   lsoStorageClassName,
+				Replicated: cephv1.ReplicatedSpec{
+					Size:                   3,
+					RequireSafeReplicaSize: true,
+				},
+			},
+		},
+	}
+	_, err := ctrl.CreateOrUpdate(context.TODO(), client, cephBlockPool, mutate)
```
rchikatw, line 91: you have missed adding the rbac as well you can add it here: https://github.com/red-hat-storage/ocs-operator/tree/main/config/rbac

services/ux-backend/handlers/expandstorage/handler.go
```diff
@@ -0,0 +1,108 @@
...
+
+	client, err := client.New(ctrl.GetConfigOrDie(), client.Options{})
+	if err != nil {
+		klog.Errorf("failed to create client: %v", err)
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+	mutate := func() error { return nil }
+
+	// update storageCluster
```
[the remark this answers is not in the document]
parth-gr: Will update

services/ux-backend/handlers/expandstorage/handler.go
```diff
@@ -0,0 +1,108 @@
...
+		PoolDetails         poolDetails         `json:"poolDetails"`
+		StorageClassDetails storageClassDetails `json:"storageClassDetails"`
+	}{}
+
+	if err = json.NewDecoder(r.Body).Decode(&ExpandStorage); err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	client, err := client.New(ctrl.GetConfigOrDie(), client.Options{})
```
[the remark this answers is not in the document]
parth-gr: I can use the same client function, once that PR get merged, 
No need to create two clients,

Will open a future re-factor PR

services/ux-backend/handlers/expandstorage/handler.go
```diff
@@ -0,0 +1,287 @@
...
+		DataProtectionPolicy int    `json:"dataProtectionPolicy"`
+		EnableCompression    bool   `json:"enableCompression"`
+		FilesystemName       string `json:"filesystemName"`
+	}
+
+	type storageClassDetails struct {
+		ReclaimPolicy                string `json:"reclaimPolicy"`
+		Name                         string `json:"name"`
+		VolumeBindingMode            string `json:"volumeBindingMode"`
+		EnableStorageClassEncryption bool   `json:"enableStorageClassEncryption"`
```
SanjalKatiyar, line 53: So far from UI it is mandatory to use KMS with StorageClass encryption. Here I can see `EnableStorageClassEncryption` is of type `bool`, so we are certain that we don't want KMS here ??

Will Ceph-CSI use K8s Secret instead to store keys ??

services/ux-backend/handlers/expandstorage/handler.go
```diff
@@ -0,0 +1,287 @@
...
+	_, err = ctrl.CreateOrUpdate(r.Context(), client, storageCluster, mutate)
+	if err != nil {
+		klog.Errorf("failed to update storageCluster: %v", err)
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+}
+
+func createCephBlockPool(w http.ResponseWriter, r *http.Request, client client.Client, poolName, lsoStorageClassName string, mutate func() error, dataProtectionPolicy int, enableCompression bool, namespace string) {
+	compressiom := "none"
```
SanjalKatiyar, line 134:
```suggestion
	compression := "none"
```

services/ux-backend/handlers/expandstorage/handler.go
```diff
@@ -0,0 +1,287 @@
...
+		compressiom = "aggressive"
+	}
+	cephBlockPool := &cephv1.CephBlockPool{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      poolName,
+			Namespace: namespace,
+		},
+		Spec: cephv1.NamedBlockPoolSpec{
+			PoolSpec: cephv1.PoolSpec{
+				FailureDomain: "host",
```
SanjalKatiyar, line 145: also need to set `enableCrushUpdates: true,` spec...

Ref:
https://github.com/rook/rook/pull/14322
https://github.com/red-hat-storage/ocs-operator/pull/2751

services/ux-backend/handlers/expandstorage/handler.go
```diff
@@ -0,0 +1,287 @@
...
+	_, err := ctrl.CreateOrUpdate(r.Context(), client, storageClass, mutate)
+	if err != nil {
+		klog.Errorf("failed to create storageClass: %v", err)
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+}
+
+func createCephFilesystem(w http.ResponseWriter, r *http.Request, client client.Client, poolName, lsoStorageClassName string, mutate func() error, dataProtectionPolicy int, enableCompression bool, filesystemName, namespace string) {
+	compressiom := "none"
```
SanjalKatiyar, line 205:
```suggestion
	compression := "none"
```

services/ux-backend/handlers/expandstorage/handler.go
SanjalKatiyar, line 145: will FD always be "host", just to confirm that we are sure ??

asking because UI reads it from the StorageCluster CR's status and set the value accordingly, instead of hardcoding "host" always.

services/ux-backend/handlers/expandstorage/handler.go
```diff
@@ -0,0 +1,287 @@
...
+	if err != nil {
+		klog.Errorf("failed to get cephFilesystem: %v", err)
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	datapool := cephv1.NamedPoolSpec{
+		Name: poolName,
+		PoolSpec: cephv1.PoolSpec{
+			FailureDomain: "host",
```
SanjalKatiyar, line 221: same FD hardcoding related question here as well...

services/ux-backend/handlers/expandstorage/handler.go
```diff
@@ -0,0 +1,287 @@
...
+		Provisioner:       storagecluster.CephFSDriverName,
+		ReclaimPolicy:     (*corev1.PersistentVolumeReclaimPolicy)(&reclaimPolicy),
+		VolumeBindingMode: (*storagev1.VolumeBindingMode)(&volumeBindingMode),
+		// AllowVolumeExpansion is set to true to enable expansion of OCS backed Volumes
+		AllowVolumeExpansion: &allowVolumeExpansion,
+		Parameters: map[string]string{
+			"clusterID":  namespace,
+			"fsName":     filesystemName,
+			"pool":       poolName,
+			"encryption": strconv.FormatBool(enableEncryption),
```
SanjalKatiyar, line 261: just want to point out that UI doesn't support encryption for CephFS prov classes yet... I hope Eran is aware that this feature does not follow existing behaviour, if not then please bring this up with him once.

services/ux-backend/handlers/expandstorage/handler.go
parth-gr: We will use KMS itself the cluster-wide settings,
This field `EnableStorageClassEncryption` can set per device set encryption, whether a device set needs encryption or not,

services/ux-backend/handlers/expandstorage/handler.go
parth-gr: We can get from UI then

services/ux-backend/handlers/expandstorage/handler.go
SanjalKatiyar: IIUC field `enableEncryption` is for deviceSet encryption and here we could use cluster-wide details.

but I am confirming about `enableStorageClassEncryption` field, which is for StorageClass encryption (not OSD/deviceSet), Ceph-CSI uses different ConfigMap than Rook for KMS.

services/ux-backend/handlers/expandstorage/handler.go
malayparida2000: Yes, we will need that fiag in a new block pool that is created. Sanjal has referenced the correct PRs, There was a UI change as well to add this option to any ned block pool getting created via UI

parth-gr: Added

services/ux-backend/handlers/expandstorage/handler.go
parth-gr: I am specifying this using bool setting  https://github.com/rook/rook/blob/master/deploy/examples/csi/rbd/storageclass.yaml#L51

can UI also provide https://github.com/rook/rook/blob/master/deploy/examples/csi/rbd/storageclass.yaml#L56?

SanjalKatiyar: We need to display the same KMS form for this feature that UI currently displays on "Create StorageClass" page.

After that it's a design question that who will create/patch the csi-kms ConfigMap with KMS details. On StorageClass creation page (day-2) and during initial deployment (day-1) UI creates it... but here we are introducing new API endpoint, so whether this API should do it now or UI should do it and pass the `encryptionKMSID` in the payload.

SanjalKatiyar: cc @vbnrh

services/ux-backend/handlers/expandstorage/handler.go
SanjalKatiyar: then enable encryption checkbox should be hidden or disabled then filesystem is selected cc @vbnrh

services/ux-backend/handlers/expandstorage/handler.go
parth-gr: Added the field `encryptionKMSID`

services/ux-backend/handlers/expandstorage/handler.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+	_, err := ctrl.CreateOrUpdate(context.TODO(), client, cephBlockPool, mutateFunc)
+	if err != nil {
+		klog.Errorf("failed to create cephBlockPool: %v", err)
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+}
+
+func createCephBlockPoolStorageClass(w http.ResponseWriter, client client.Client, storageClassName, poolName, reclaimPolicy, volumeBindingMode string, enableEncryption bool, encryptionKMSID string, namespace string) {
```
travisn, line 175:
```suggestion
func createCephBlockPoolStorageClass(w http.ResponseWriter, client client.Client, storageClassName, poolName, reclaimPolicy, volumeBindingMode string, enableEncryption bool, encryptionKMSID, namespace string) {
```

services/ux-backend/handlers/expandstorage/handler.go
```diff
@@ -0,0 +1,302 @@
...
+	type storageClassDetails struct {
+		ReclaimPolicy                string `json:"reclaimPolicy"`
+		Name                         string `json:"name"`
+		VolumeBindingMode            string `json:"volumeBindingMode"`
+		EnableStorageClassEncryption bool   `json:"enableStorageClassEncryption"`
+		EncryptionKMSID              string `json:"encryptionKMSID"`
+	}
+
+	var ExpandStorage = struct {
+		LsoStorageClassName string              `json:"lsoStorageClassName"`
```
travisn, line 60: Today our only scenario might be for LSO, but it's good to keep the naming more generic in case other scenarios are added in the future. If this would be confused with the storage class name we are generating, perhaps we could name it something like `StorageClassForOSDs`.
```suggestion
		StorageClassName string              `json:"storageClassName"`
```

services/ux-backend/handlers/expandstorage/handler.go
```diff
@@ -0,0 +1,302 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      poolName,
+			Namespace: namespace,
+		},
+		Spec: cephv1.NamedBlockPoolSpec{
+			PoolSpec: cephv1.PoolSpec{
+				FailureDomain: failureDomain,
+				DeviceClass:   lsoStorageClassName,
+				Replicated: cephv1.ReplicatedSpec{
+					Size:                   uint(dataProtectionPolicy),
```
travisn, line 153: Is this feature enabled for stretch clusters? The pool spec is a bit different in that case, where we need to set `replicasPerFailureDomain: 2`, and the `subFailureDomain`

travisn (changes requested):

services/ux-backend/handlers/expandstorage/handler.go
SanjalKatiyar: @travisn by any chance Rook takes care of this automatically ??
Asking because UI never sets this on day-2 BlockPools.
Also, I didn't find any reference of it in OCS operator for day-1 BlockPool (unless I missed, @malayparida2000 can u plz confirm ??)

parth-gr: Correct we don't have this enabled for `BlockPools` UI currently, so it is not a common scenario,
We can have this added once we have Blockpool UI using it,

malayparida2000: On the default cephblockpool that ocs-operator creates the Size & the Replica per failure domain field are set accordingly if it's a stretch cluster. https://github.com/red-hat-storage/ocs-operator/blob/c079ab81fd02767ddccc8c24bd324e6c4df89090/controllers/storagecluster/generate.go#L135-L148.

malayparida2000: If extra blockpools are being created via UI those would not be reconciled by ocs operator. So adding the options during creation as Travis suggests seems to be a good idea, but not sure why UI didn't have this configured till now

parth-gr: Added a arbiter check

parth-gr:
@rchikatw @travisn @malayparida2000 @SanjalKatiyar @rewantsoni ready for the approval?

services/ux-backend/handlers/expandstorage/handler.go
```diff
@@ -0,0 +1,310 @@
...
+	updateStorageCluster(w, r, client, ExpandStorage.StorageClassForOSDs, ExpandStorage.EnableEncryption, ExpandStorage.Storage, ExpandStorage.Count, ExpandStorage.Replica, storageCluster)
+
+	if ExpandStorage.PoolDetails.VolumeType == "block" {
+		// create cephBlockPool
+		createCephBlockPool(w, r, client, ExpandStorage.PoolDetails.PoolName, ExpandStorage.StorageClassForOSDs, ExpandStorage.PoolDetails.DataProtectionPolicy, ExpandStorage.PoolDetails.EnableCompression, namespace, ExpandStorage.PoolDetails.FailureDomain, storageCluster.Spec.Arbiter.Enable)
+
+		// create storageClass
+		createCephBlockPoolStorageClass(w, r, client, ExpandStorage.StorageClassDetails.Name, ExpandStorage.PoolDetails.PoolName, ExpandStorage.StorageClassDetails.ReclaimPolicy, ExpandStorage.StorageClassDetails.VolumeBindingMode, ExpandStorage.StorageClassDetails.EnableStorageClassEncryption, ExpandStorage.StorageClassDetails.EncryptionKMSID, namespace)
+	} else {
+		// create cephFilesystem
```
malayparida2000, line 92: // create cephFilesystem dataPool

malayparida2000 (commented):
mostly LGTM, just these 2 neat suggestions

services/ux-backend/handlers/expandstorage/handler.go
```diff
@@ -0,0 +1,310 @@
...
+		StorageClassDetails storageClassDetails `json:"storageClassDetails"`
+	}{}
+
+	if err = json.NewDecoder(r.Body).Decode(&ExpandStorage); err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	storageCluster := &ocsv1.StorageCluster{}
+	err = client.Get(r.Context(), types.NamespacedName{Name: ExpandStorage.StorageClusterName, Namespace: namespace}, storageCluster)
```
rewantsoni, line 75: Is this only supported for storagecluster in operator namespace?

services/ux-backend/handlers/expandstorage/handler.go
```diff
@@ -0,0 +1,310 @@
...
+		},
+	}
+	if enableEncryption {
+		storageClass.Parameters["encryptionKMSID"] = encryptionKMSID
+	}
+
+	mutateFunc := func() error {
+		return nil
+	}
+	_, err := ctrl.CreateOrUpdate(r.Context(), client, storageClass, mutateFunc)
```
rewantsoni, line 220: Since storageClass params cannot be updated after it is created once, would it make sense to use create here? and if the params differ then delete and create it again?

services/ux-backend/handlers/expandstorage/handler.go
```diff
@@ -0,0 +1,310 @@
...
+			"csi.storage.k8s.io/node-stage-secret-namespace":        namespace,
+			"csi.storage.k8s.io/controller-expand-secret-name":      "rook-csi-cephfs-provisioner",
+			"csi.storage.k8s.io/controller-expand-secret-namespace": namespace,
+		},
+	}
+
+	mutateFunc := func() error {
+		return nil
+	}
+	_, err := ctrl.CreateOrUpdate(r.Context(), client, storageClass, mutateFunc)
```
rewantsoni, line 293: Since storageClass params cannot be updated after it is created once, would it make sense to use create here? and if the params differ then delete and create it again?

services/ux-backend/main.go
```diff
@@ -76,6 +79,12 @@ func main() {
 	if err := ocsv1.AddToScheme(scheme); err != nil {
 		klog.Exitf("failed to add ocsv1 to scheme. %v", err)
 	}
+	if err := cephv1.AddToScheme(scheme); err != nil {
+		klog.Exitf("failed to add ocsv1 to scheme. %v", err)
```
rewantsoni, line 83:
```suggestion
		klog.Exitf("failed to add cephv1 to scheme. %v", err)
```

services/ux-backend/main.go
```diff
@@ -76,6 +79,12 @@ func main() {
...
+	}
+	if err := storagev1.AddToScheme(scheme); err != nil {
+		klog.Exitf("failed to add ocsv1 to scheme. %v", err)
```
rewantsoni, line 86:
```suggestion
		klog.Exitf("failed to add storagev1 to scheme. %v", err)
```

rewantsoni (changes requested):

SanjalKatiyar:
LGTM.

services/ux-backend/handlers/expandstorage/handler.go
parth-gr: Yes we do run storage cluster on same namespace as per this feature

malayparida2000 (approved):
/lgtm

parth-gr:
cc @iamniting PTAL

Madhu-1 (commented):
It looks like this design is more specific to LSO, shouldnt we try to make the API or code more generic instead of targeting only LSO (ocs-operator/backend server)?

services/ux-backend/handlers/expandstorage/handler.go
```diff
@@ -0,0 +1,310 @@
...
+		return
+	}
+
+	type poolDetails struct {
+		VolumeType           string `json:"volumeType"`
+		PoolName             string `json:"poolName"`
+		DataProtectionPolicy int    `json:"dataProtectionPolicy"`
+		EnableCompression    bool   `json:"enableCompression"`
+		FilesystemName       string `json:"filesystemName"`
+		FailureDomain        string `json:"failureDomain"`
```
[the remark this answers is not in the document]
parth-gr: As per the design we will use the name of the (LSO storageclass === Device Class)

services/ux-backend/handlers/expandstorage/handler.go
```diff
@@ -0,0 +1,310 @@
...
+		http.Error(w, "body in the request is required", http.StatusBadRequest)
+		return
+	}
+
+	type poolDetails struct {
+		VolumeType           string `json:"volumeType"`
+		PoolName             string `json:"poolName"`
+		DataProtectionPolicy int    `json:"dataProtectionPolicy"`
+		EnableCompression    bool   `json:"enableCompression"`
+		FilesystemName       string `json:"filesystemName"`
```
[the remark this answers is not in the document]
parth-gr: We support adding multiple dataPools for a fileSystem

services/ux-backend/handlers/expandstorage/handler.go
```diff
@@ -0,0 +1,310 @@
...
+		FailureDomain        string `json:"failureDomain"`
+	}
+
+	type storageClassDetails struct {
```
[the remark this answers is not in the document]
parth-gr: It is common for both, 
Only encryption will be not be used for cephfs

services/ux-backend/handlers/expandstorage/handler.go
```diff
@@ -0,0 +1,310 @@
...
+		ReclaimPolicy                string `json:"reclaimPolicy"`
+		Name                         string `json:"name"`
+		VolumeBindingMode            string `json:"volumeBindingMode"`
+		EnableStorageClassEncryption bool   `json:"enableStorageClassEncryption"`
+		EncryptionKMSID              string `json:"encryptionKMSID"`
+	}
+
+	var ExpandStorage = struct {
+		StorageClassForOSDs string              `json:"storageClassForOSDs"`
+		EnableEncryption    bool                `json:"enableEncryption"`
```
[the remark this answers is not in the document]
parth-gr: We are adding support for per-device-set OSD encryption.
SO some device Sets can have encryption and some dont

services/ux-backend/handlers/expandstorage/handler.go
```diff
@@ -0,0 +1,310 @@
...
+		klog.Info("Creating cephBlockPool and storageClass")
+		// create cephBlockPool
+		createCephBlockPool(w, r, client, ExpandStorage.PoolDetails.PoolName, ExpandStorage.StorageClassForOSDs, ExpandStorage.PoolDetails.DataProtectionPolicy, ExpandStorage.PoolDetails.EnableCompression, namespace, ExpandStorage.PoolDetails.FailureDomain, storageCluster.Spec.Arbiter.Enable)
+
+		// create storageClass
+		createCephBlockPoolStorageClass(w, r, client, ExpandStorage.StorageClassDetails.Name, ExpandStorage.PoolDetails.PoolName, ExpandStorage.StorageClassDetails.ReclaimPolicy, ExpandStorage.StorageClassDetails.VolumeBindingMode, ExpandStorage.StorageClassDetails.EnableStorageClassEncryption, ExpandStorage.StorageClassDetails.EncryptionKMSID, namespace)
+	} else if ExpandStorage.PoolDetails.VolumeType == "filesystem" {
+		klog.Info("Creating cephFilesystem dataPool and storageClass")
+		// create cephFilesystem dataPool
+		createCephFilesystemDataPool(w, r, client, ExpandStorage.PoolDetails.PoolName, ExpandStorage.StorageClassForOSDs, ExpandStorage.PoolDetails.DataProtectionPolicy, ExpandStorage.PoolDetails.EnableCompression, ExpandStorage.PoolDetails.FailureDomain, storageCluster)
```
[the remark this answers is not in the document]
parth-gr: SubvolumeGroup currently is not supported, we can look it for provider mode,

We are adding new dataPools on the same filesystem to increase the storage.

parth-gr:
> It looks like this design is more specific to LSO, shouldnt we try to make the API or code more generic instead of targeting only LSO (ocs-operator/backend server)?

Thanks, Removed any specific instance of LSO

services/ux-backend/handlers/expandstorage/handler.go
```diff
@@ -0,0 +1,310 @@
...
+	}
+
+}
+
+func updateStorageCluster(w http.ResponseWriter, r *http.Request, client client.Client, lsoStorageClassName string, enableEncryption bool, storage string, count, replica int, storageCluster *ocsv1.StorageCluster) {
+	klog.Infof("Updating storageCluster %q", storageCluster.Name)
+	storageQty := resource.MustParse(storage)
+	volumeMode := corev1.PersistentVolumeBlock
+	deviceSet := ocsv1.StorageDeviceSet{
+		Name:        lsoStorageClassName,
```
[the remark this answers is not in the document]
parth-gr: Yes currently the feature is only for LSO storageclass

parth-gr:
@Madhu-1 @iamniting @rewantsoni can we get a final approval, 
We would  like to test the end-to-end changes with UI using the build,  https://github.com/red-hat-storage/odf-console/pull/1662, 
Currently, we are just able to test it by manually editing csv's and images

services/ux-backend/handlers/expandstorage/handler.go
```diff
@@ -0,0 +1,310 @@
...
+			},
+		},
+	}
+
+	mutateFunc := func() error {
+		storageCluster.Spec.StorageDeviceSets = append(storageCluster.Spec.StorageDeviceSets, deviceSet)
+		return nil
+	}
+
+	_, err := ctrl.CreateOrUpdate(r.Context(), client, storageCluster, mutateFunc)
```
iamniting, line 137: Why do we need a `CreateOrUpdate`? I believe just update is enough.

services/ux-backend/handlers/expandstorage/handler.go
```diff
@@ -0,0 +1,310 @@
...
+	}
+
+	mutateFunc := func() error {
+		storageCluster.Spec.ManagedResources.CephFilesystems.AdditionalDataPools = append(storageCluster.Spec.ManagedResources.CephFilesystems.AdditionalDataPools, datapool)
+		return nil
+	}
+
+	_, err := ctrl.CreateOrUpdate(r.Context(), client, storageCluster, mutateFunc)
+	if err != nil {
+		klog.Errorf("failed to update storageCluster: %v", err)
```
iamniting, line 260: Can you pls correct this err msg

iamniting (changes requested):
Can you pls separate manual changes and generated changes?

Also, if the pool and storage class already exist, should we not send a response indicating that the request has already been fulfilled or something similar? So that the requester/UI already knows that things are in place already.

parth-gr:
> Can you pls separate manual changes and generated changes?

Added a separate commit

> Also, if the pool and storage class already exist, should we not send a response indicating that the request has already been fulfilled or something similar? So that the requester/UI already knows that things are in place already.

This endpoint is only called when there is new LSO storageclass is used, that means we always need to create the pools and storageclass
PLus UI currently checks the ok status with 200 status code

services/ux-backend/handlers/expandstorage/handler.go
```diff
@@ -0,0 +1,298 @@
...
+				"compression_mode": compression,
+			},
+			EnableCrushUpdates: true,
+		},
+	}
+
+	storageCluster.Spec.ManagedResources.CephFilesystems.AdditionalDataPools = append(storageCluster.Spec.ManagedResources.CephFilesystems.AdditionalDataPools, datapool)
+	err := client.Update(r.Context(), storageCluster)
+	if err != nil {
+		klog.Errorf("failed to update storageCluster: %q", storageCluster.Name)
```
iamniting, line 248: Pls fix this msg.

iamniting (changes requested):
Can you have below 3 files in first commit and others in second commit

`controllers/storagecluster/storageclasses.go`
`services/ux-backend/handlers/expandstorage/handler.go`
`services/ux-backend/main.go`

iamniting:
> > Can you pls separate manual changes and generated changes?
> 
> Added a separate commit
> 
> > Also, if the pool and storage class already exist, should we not send a response indicating that the request has already been fulfilled or something similar? So that the requester/UI already knows that things are in place already.
> 
> This endpoint is only called when there is new LSO storageclass is used, that means we always need to create the pools and storageclass PLus UI currently checks the ok status with 200 status code

The names are passed by the UI. If the same names exist we should send succeed msg IMO. But let's hear from @SanjalKatiyar

SanjalKatiyar:
> The names are passed by the UI. If the same names exist we should send succeed msg IMO. But let's hear from @SanjalKatiyar

UI has validations in place so this should not happen for happy-path, but in case it happens IMO it's okay to show error.
K8s API server also returns error if we try to create an existing CR (same name/namespace). Also, it can happen that specs/parameters of already existing pool/class are different than what user wants now in new pool/class, so better to show error and let user use new name instead.

parth-gr:
> > The names are passed by the UI. If the same names exist we should send succeed msg IMO. But let's hear from @SanjalKatiyar
> 
> UI has validations in place so this should not happen for happy-path, but in case it happens IMO it's okay to show error. K8s API server also returns error if we try to create an existing CR (same name/namespace). Also, it can happen that specs/parameters of already existing pool/class are different than what user wants now in new pool/class, so better to show error and let user use new name instead.

As UI is sending a unique name, I don't this so we need to re-validate,
It's a single endpoint call which already loaded with three/four API operation,
We should not overload it.

If the blockpool wasn't created anyway user will see the error in the UI.

services/ux-backend/handlers/expandstorage/handler.go
iamniting: We sorted it over a call.

PR #2878, opened 2024-11-06
controllers: add a new controller to setup mirroring

rewantsoni:
A new controller to enable mirroring for cephblockpool and radosnamespace

Depends On:
1. https://github.com/red-hat-storage/ocs-operator/pull/2875
2. https://github.com/red-hat-storage/ocs-operator/pull/2675


controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,409 @@
...
+	r.log = log.FromContext(ctx, "Mirroring Controller", request)
+	r.log.Info("Reconciling Mirroring Controller.")
+
+	r.clientMappingConfig = &corev1.ConfigMap{}
+	r.clientMappingConfig.Name = request.Name
+	r.clientMappingConfig.Namespace = request.Namespace
+
+	if err = r.get(r.clientMappingConfig); err != nil {
+		if k8serrors.IsNotFound(err) {
+			r.log.Info(fmt.Sprintf("ConfigMap %s not found. Ignoring since object must be deleted.", r.clientMappingConfig.Name))
```
umangachapagain, line 94: Do not use sprintf when logging. Use key-value pairs instead.

umangachapagain (changes requested):

umangachapagain (commented):
Having comments on code would definitely make it much easier for us to review it. This is such a big change and doesn't make any sense to me.

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,448 @@
...
+		context.TODO(),
+		&ocsv1alpha1.StorageConsumer{},
+		"clientID",
+		util.StorageConsumerClientIDIndexFieldFunc,
+	); err != nil {
+		return fmt.Errorf("unable to set up FieldIndexer on StorageConsumer for clientID: %v", err)
+	}
+
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&corev1.ConfigMap{}, builder.WithPredicates(util.NamePredicate(storageclusterpeer.StorageClientMappingConfigName))).
```
nb-ohad, line 85: Please split into multiple lines

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,448 @@
...
+//+kubebuilder:rbac:groups=core,resources=configmaps/finalizers,verbs=update
+//+kubebuilder:rbac:groups=core,resources=secrets,verbs=get;list;watch;create;update;delete
+// +kubebuilder:rbac:groups=ceph.rook.io,resources=cephrbdmirrors,verbs=get;list;watch;create;update;delete
+
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
+// move the current state of the cluster closer to the desired state.
+func (r *MirroringReconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
+	var err error
+	r.ctx = ctx
+	r.log = log.FromContext(ctx, "Mirroring Controller", request)
```
nb-ohad, line 102: I believe the controller name is already set up by the framework

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,448 @@
...
+	r.log.Info("Reconciling Mirroring Controller.")
```
nb-ohad, line 103: No need to repeat the controller name again in the message

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,448 @@
...
+		"clientID",
+		util.StorageConsumerClientIDIndexFieldFunc,
+	); err != nil {
+		return fmt.Errorf("unable to set up FieldIndexer on StorageConsumer for clientID: %v", err)
+	}
+
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&corev1.ConfigMap{}, builder.WithPredicates(util.NamePredicate(storageclusterpeer.StorageClientMappingConfigName))).
+		Watches(&ocsv1.StorageClusterPeer{}, &handler.EnqueueRequestForObject{}).
+		Watches(&ocsv1alpha1.StorageConsumer{}, &handler.EnqueueRequestForObject{}).
```
nb-ohad, line 87: By design, this controller is supposed to handle configmap requests. we should not allow other object kind to be enqueued

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,448 @@
...
+	r.clientMappingConfig = &corev1.ConfigMap{}
+	r.clientMappingConfig.Name = request.Name
+	r.clientMappingConfig.Namespace = request.Namespace
+
+	if err = r.get(r.clientMappingConfig); err != nil {
+		if k8serrors.IsNotFound(err) {
+			r.log.Info("ClientMappingConfig %s not found. Ignoring since object must be deleted.")
+			return ctrl.Result{}, nil
+		}
+		r.log.Error(err, "Failed to get ConfigMap.", "ConfigMap", r.clientMappingConfig.Name)
```
nb-ohad, line 114: One thing that can help here is to add the Config Name to the log context so it will be part of every log message
Also, use the full NamespacedName form the request for that

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,448 @@
...
+			if err := r.update(r.clientMappingConfig); err != nil {
+				r.log.Info(
+					"Failed to remove finalizer from ClientMappingConfig",
+					"ClientMappingConfig",
+					r.clientMappingConfig.Name,
+				)
+				return ctrl.Result{}, fmt.Errorf("failed to remove finalizer from ClientMappingConfig: %v", err)
+			}
+		}
+	}
```
nb-ohad, line 139: Handle deletion after normal pat

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,448 @@
...
+	if !r.clientMappingConfig.GetDeletionTimestamp().IsZero() {
+		if res, err := r.disableBlockPoolMirroring(); err != nil || !res.IsZero() {
+			return res, err
+		}
+
+		if controllerutil.RemoveFinalizer(r.clientMappingConfig, mirroringFinalizer) {
+			r.log.Info(
+				"removing finalizer from ClientMappingConfig.",
+				"ClientMappingConfig",
+				r.clientMappingConfig.Name,
```
nb-ohad, line 128: As mentioned in a prev comment, this information should already be part logger on the context

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,448 @@
...
+					"ClientMappingConfig",
+					r.clientMappingConfig.Name,
+				)
+				return ctrl.Result{}, fmt.Errorf("failed to remove finalizer from ClientMappingConfig: %v", err)
+			}
+		}
+	}
+
+	if len(r.clientMappingConfig.Data) < 1 {
+		return ctrl.Result{}, nil
```
nb-ohad, line 142: I think that you need to disable mirroring in this case and remove the finalizer

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,448 @@
...
+	}
+
+	if controllerutil.AddFinalizer(r.clientMappingConfig, mirroringFinalizer) {
+		r.log.Info(
+			"Finalizer not found for ClientMappingConfig. Adding finalizer.",
+			"ClientMappingConfig",
+			r.clientMappingConfig.Name,
```
nb-ohad, line 149: Same as prev comment

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,448 @@
...
+		)
+		if err := r.update(r.clientMappingConfig); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to update ClientMappingConfig: %v", err)
+		}
+	}
+
+	// Find the StorageClusterPeer from OwnerRef
+	owner := util.FindOwnerRefByKind(r.clientMappingConfig, "StorageClusterPeer")
```
nb-ohad, line 157: StorageClusterPeer in not an owner of client mapping config

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,448 @@
...
+		r.log.Error(err, "Failed to get StorageClusterPeer.")
+		return ctrl.Result{}, err
+	}
+
+	if r.storageClusterPeer.Status.State != ocsv1.StorageClusterPeerStatePeered {
+		return ctrl.Result{}, fmt.Errorf(
+			"waiting for StorageClusterPeer %s to be in Peered state",
+			r.storageClusterPeer.Name,
+		)
+	}
```
nb-ohad, line 181: This is part of the reconciliation, which means it needs to reside inside the reconcile phases (at a sub func)

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,448 @@
...
+
+	for i := range cephBlockPoolsList.Items {
+		cephBlockPool := &cephBlockPoolsList.Items[i]
+
+		cephBlockPool.Spec.Mirroring = rookCephv1.MirroringSpec{}
+		err := r.update(cephBlockPool)
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to disable mirroring for CephBlockPool %q: %v", cephBlockPool.Name, err)
+		}
```
nb-ohad, line 222:
```suggestion
		if err := r.update(cephBlockPool); err != nil {
			return ctrl.Result{}, fmt.Errorf(
				"failed to disable mirroring for CephBlockPool %q: %v", 
				cephBlockPool.Name, 
				err,
			)
```

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,448 @@
...
+	}
+
+	if res, err := r.reconcileRadosNamespaceMirroring(ocsClient); err != nil || !res.IsZero() {
+		return res, err
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func (r *MirroringReconciler) disableBlockPoolMirroring() (ctrl.Result, error) {
```
nb-ohad, line 208: Can't this be just another value of the normal reconciliation of the block pool?

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,448 @@
...
+		cephBlockPool.Spec.Mirroring = rookCephv1.MirroringSpec{}
+		err := r.update(cephBlockPool)
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to disable mirroring for CephBlockPool %q: %v", cephBlockPool.Name, err)
+		}
+	}
+	return ctrl.Result{}, nil
+}
+
+func (r *MirroringReconciler) reconcileRBDMirrorDaemon() error {
```
nb-ohad, line 227: 1. You are not reconciling the daemon, you are reconciling the CR
2. RBD -> Rbd
```suggestion
func (r *MirroringReconciler) reconcileRbdMirror() error {
```

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,448 @@
...
+
+	storageConsumers := &ocsv1alpha1.StorageConsumerList{}
+	if err := r.list(
+		storageConsumers,
+		client.MatchingFields{"annotations": util.RequestMaintenanceModeAnnotation},
+	); err != nil {
+		return fmt.Errorf("failed to list StorageConsumer(s): %v", err)
+	}
+
+	if len(storageConsumers.Items) > 0 {
```
nb-ohad, line 240: This is not a good check, it will always be true (the length will always be greater then 1)

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,448 @@
...
+	}
+	selector = selector.Add(*blockPoolLabelSelectorRequirement)
+
+	cephBlockPoolsList := &rookCephv1.CephBlockPoolList{}
+	err = r.list(cephBlockPoolsList, client.InNamespace(r.storageClusterPeer.Namespace), client.MatchingLabelsSelector{Selector: selector})
+	if err != nil {
+		return nil, fmt.Errorf("failed to list CephBlockPools: %w", err)
+	}
+	return cephBlockPoolsList, nil
+}
```
nb-ohad, line 426: There is no point in creating this complicated function for listing.
Preferably create one for the selector only

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,448 @@
...
+			"pool":  []byte(blockPoolName),
+			"token": []byte(mirroringToken),
+		}
+		return nil
+	})
+	if err != nil {
+		return fmt.Errorf("failed to create/update bootstrap secret: %w", err)
+	}
+	return nil
+}
```
nb-ohad, line 448: This logic runs only once. Why do we need a separate func?

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,448 @@
...
+	return r.Client.Update(r.ctx, obj, opts...)
+}
+
+func (r *MirroringReconciler) delete(obj client.Object, opts ...client.DeleteOption) error {
+	return r.Client.Delete(r.ctx, obj, opts...)
+}
+
+func (r *MirroringReconciler) own(obj client.Object) error {
+	return controllerutil.SetControllerReference(r.storageClusterPeer, obj, r.Scheme)
+}
```
nb-ohad, line 410: An r.Own util is created to have an object owned by the input CR of this controller. This is not the case here. And this controller should not mark ownership for another controller's CR

nb-ohad (changes requested):

controllers/mirroring/mirroring_controller.go
rewantsoni: Yes, we need to disable mirroring but why do we need to remove finalizer in this case?

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,448 @@
...
+	ctx                 context.Context
+	clientMappingConfig *corev1.ConfigMap
+	storageClusterPeer  *ocsv1.StorageClusterPeer
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *MirroringReconciler) SetupWithManager(mgr ctrl.Manager) error {
+
+	if err := mgr.GetCache().IndexField(
+		context.TODO(),
```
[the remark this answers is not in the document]
rewantsoni: I will pass the context from main like we do while creating other indexer

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,477 @@
...
+		if err := r.own(clientMappingConfig, rbdMirror); err != nil {
+			return err
+		}
+		rbdMirror.Spec.Count = 1
+		return nil
+	})
+	if err != nil {
+		r.log.Error(err, "Failed to create/update the CephRBDMirror", "CephRBDMirror", rbdMirror)
+		return err
+	}
```
nb-ohad, line 265: 1. Reverse the order, create first delete second. 
2. The condition for creating vs deleting should be more encompassing to include all known scenarios

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,477 @@
...
+	selector = selector.Add(*blockPoolLabelSelectorRequirement)
+
+	cephBlockPoolsList := &rookCephv1.CephBlockPoolList{}
+	if err = r.list(
+		cephBlockPoolsList,
+		client.InNamespace(clientMappingConfig.Namespace),
+		client.MatchingLabelsSelector{Selector: selector},
+	); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to list cephBlockPools: %v", err)
+	}
```
nb-ohad, line 295: I prefer if you list everything and then skip the labeled one in your for loop. This will create a more maintainable and readable code

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,477 @@
...
+
+	var blockPoolsList []string
```
nb-ohad, line 297: There is no need to have a list of names, you can extract it from the map using `maps.Keys`

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,477 @@
...
+			cephBlockPool.Spec.Mirroring = rookCephv1.MirroringSpec{}
+		}
+
+		if err = r.update(&cephBlockPool); err != nil {
+			return ctrl.Result{}, fmt.Errorf(
+				"failed to update mirroring for CephBlockPool %v: %w",
+				cephBlockPool.Name,
+				err,
+			)
+		}
```
nb-ohad, line 319: This code need to only be applied to block pools that also exists on the peer

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,477 @@
...
+	for i := range response.BlockPoolsInfo {
+		blockPoolName := response.BlockPoolsInfo[i].BlockPoolName
+
+		mirroringToken := response.BlockPoolsInfo[i].MirroringToken
+		if mirroringToken == "" {
+			return ctrl.Result{}, fmt.Errorf("failed to fetch mirroring token for the blockPool")
+		}
+
+		mirroringSecret := &corev1.Secret{}
+		mirroringSecret.Name = fmt.Sprintf("%s-%s", "peer", blockPoolName)
```
nb-ohad, line 341: Let's find a better name that reflects the content

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,477 @@
...
+		make a map of peerClientIDtoRadosNamespace
+		list all radosNamespace and for each,
+			find out which consumer does it belong to
+			find the localClientID and use map to get radosnamespce
+			if radosnamespace is empty, continue
+			else enable mirroring
+	*/
+
+	peerClientIDs := []string{}
+	nameToStorageConsumer := map[string]*ocsv1alpha1.StorageConsumer{}
```
nb-ohad, line 404: Offering a better name `StorageConsumerByName`

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,477 @@
...
+		}
+		nameToStorageConsumer[storageConsumers.Items[0].Name] = &storageConsumers.Items[0]
+		peerClientIDs = append(peerClientIDs, peerClientID)
+	}
+
+	response, err := ocsClient.GetStorageClientsInfo(r.ctx, storageClusterPeer.Status.PeerInfo.StorageClusterUid, peerClientIDs)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	peerClientIDToRadosNamespace := map[string]string{}
```
nb-ohad, line 426:
```suggestion
	remoteNamespaceByClientId = map[string]string{}
```

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,477 @@
...
+	}
+
+	for i := range radosNamespaceList.Items {
+		rns := &radosNamespaceList.Items[i]
+		consumer := nameToStorageConsumer[rns.GetLabels()[controllers.StorageConsumerNameLabel]]
+		if consumer == nil {
+			continue
+		}
+		clientID := consumer.Status.Client.ID
+		radosNamespace := peerClientIDToRadosNamespace[clientMappingConfig.Data[clientID]]
```
nb-ohad, line 444:
```
remoteClientId := clientMappingConfig.Data[consumer.Status.Client.ID]
remoteNamespace := remoteNamespaceByClientId[remoteClientId]
```

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,477 @@
...
+		if radosNamespace == "" {
+			continue
+		}
+		rns.Spec.Mirroring = &rookCephv1.RadosNamespaceMirroring{
+			RemoteNamespace: ptr.To(radosNamespace),
+			Mode:            "image",
+		}
+		if err := r.update(rns); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to update radosnamespace %s", rns.Name)
+		}
```
nb-ohad, line 454: We need to add logic to disable the mirroring

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,477 @@
...
+	storageClusterPeer.Namespace = clientMappingConfig.Namespace
+
+	if err := r.get(storageClusterPeer); err != nil {
+		if k8serrors.IsNotFound(err) {
+			r.log.Info("StorageClusterPeer resource not found. Ignoring since object must be deleted.")
+			return ctrl.Result{}, nil
+		}
+		r.log.Error(err, "Failed to get StorageClusterPeer.")
+		return ctrl.Result{}, err
+	}
```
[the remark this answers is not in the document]
rewantsoni: I will open a bug to track it

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,478 @@
...
+	return r.reconcilePhases(clientMappingConfig)
+}
+
+func (r *MirroringReconciler) reconcilePhases(clientMappingConfig *corev1.ConfigMap) (ctrl.Result, error) {
+
+	disableMirroring := false
+	disableMirroring = disableMirroring || !clientMappingConfig.DeletionTimestamp.IsZero()
+	disableMirroring = disableMirroring || clientMappingConfig.Data == nil || len(clientMappingConfig.Data) < 1
```
nb-ohad, line 170: 1. Prefer the positive case for the negative case, it removes negations from the main code path
2. The var name should reflect the state, not the request

```suggestion
	shouldMirror := clientMappingConfig.DeletionTimestamp.IsZero() &&
		clientMappingConfig.Data == nil && 
		len(clientMappingConfig.Data) < 1
```

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,478 @@
...
+			r.log.Info("Finalizer not found for ConfigMap. Adding finalizer.")
+			if err := r.update(clientMappingConfig); err != nil {
+				return ctrl.Result{}, fmt.Errorf("failed to update ConfigMap: %v", err)
+			}
+		}
+	}
+
+	// Fetch the StorageClusterPeer instance
+	storageClusterPeerList := &ocsv1.StorageClusterPeerList{}
+	if err := r.list(storageClusterPeerList, client.InNamespace(clientMappingConfig.Namespace)); err != nil {
```
nb-ohad, line 183: cap your list operation at 2 and make this list op multiline as the existing line is already quite long

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,478 @@
...
+
+	storageClusterPeer := &storageClusterPeerList.Items[0]
+
+	if storageClusterPeer.Status.State != ocsv1.StorageClusterPeerStatePeered {
+		r.log.Info(
+			"waiting for StorageClusterPeer to be in Peered state",
+			"StorageClusterPeer",
+			storageClusterPeer.Name,
+		)
+		return ctrl.Result{Requeue: true}, nil
```
nb-ohad, line 199: This will create an immediate requeue, I would suggest to add a requeue backoff (maybe 1-5 seconds) to give some time for other process to complete

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,478 @@
...
+		})
+		if err != nil {
+			r.log.Error(err, "Failed to create/update the CephRBDMirror", "CephRBDMirror", rbdMirror)
+			return err
+		}
+	} else {
+		if err := r.delete(rbdMirror); err != nil {
+			return fmt.Errorf("failed to delete CephRBDMirror: %v", err)
+		}
+	}
```
nb-ohad, line 274: 1. Prefer the positive case for the negative case, it removes negations from the main code path
2. use a variable with a proper name to convey the intent (better than using comments) 

```
	maintenanceModeRequested := len(storageConsumers.Items) >= 1
	if mirroringEnabled || !maintenanceModeRequested {
		// Create or update	
	} else 
		// Delete 
	}
```

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,478 @@
...
+		cephBlockPoolsList,
+		client.InNamespace(clientMappingConfig.Namespace),
+	); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to list cephBlockPools: %v", err)
+	}
+
+	blockPoolByName := map[string]*rookCephv1.CephBlockPool{}
+	for i := range cephBlockPoolsList.Items {
+		cephBlockPool := &cephBlockPoolsList.Items[i]
+		if _, ok := cephBlockPool.GetLabels()[util.ForbidMirroringLabel]; !ok {
```
nb-ohad, line 297: We should skip if we find the label, not the other way around
```suggestion
		if _, ok := cephBlockPool.GetLabels()[util.ForbidMirroringLabel]; ok {
```

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,478 @@
...
+			continue
+		}
+		blockPoolByName[cephBlockPoolsList.Items[i].Name] = &cephBlockPoolsList.Items[i]
```
nb-ohad, line 300: You have a local var, use it.
```suggestion
		blockPoolByName[cephBlockPool.Name] = cephBlockPool
```

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,478 @@
...
+	}
+
+	// fetch BlockPoolsInfo
+	response, err := ocsClient.GetBlockPoolsInfo(
+		r.ctx,
+		storageClusterPeer.Status.PeerInfo.StorageClusterUid,
```
nb-ohad, line 306: Shouldn't we first check if we have a peerInfo with a storageClusterUid value before we send a request. Maybe we even need this to this check as the first check in the function

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,478 @@
...
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to get CephBlockPool(s) info from Peer: %w", err)
+	}
+
+	for i := range response.BlockPoolsInfo {
+		blockPoolName := response.BlockPoolsInfo[i].BlockPoolName
+		cephBlockPool := blockPoolByName[blockPoolName]
+
+		mirroringToken := response.BlockPoolsInfo[i].MirroringToken
+		secretName := fmt.Sprintf("mirroring-token-%s", blockPoolName)
```
nb-ohad, line 318: 1. This might create a name overflow because we have no control on the length of the block pool name so maybe we need to hash it, or at least cap the name length
2. Lets use `RBD` in the name

```suggestion
		secretName := fmt.Sprintf("rbd-mirroring-token-%s", hashOf(blockPoolName))
```

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,478 @@
...
+					}
+
+					if !slices.Contains(cephBlockPool.Spec.Mirroring.Peers.SecretNames, secretName) {
+						cephBlockPool.Spec.Mirroring.Peers.SecretNames = append(
+							cephBlockPool.Spec.Mirroring.Peers.SecretNames, secretName)
+					}
+				}
+			}
+			return nil
+		})
```
nb-ohad, line 364: It does not make sense to have this code running if we don't get a mirroring token

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,478 @@
...
+
+		_, err = controllerutil.CreateOrUpdate(r.ctx, r.Client, cephBlockPool, func() error {
+			util.AddAnnotation(
+				cephBlockPool,
+				util.BlockPoolMirroringTargetIDAnnotation,
+				response.BlockPoolsInfo[i].BlockPoolID,
+			)
+
+			if disableMirroring {
+				cephBlockPool.Spec.Mirroring = rookCephv1.MirroringSpec{}
```
nb-ohad, line 348: I think this check needs to be at a higher scope as there is no reason to run all of the code up until this point if we are disabling mirroring

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,478 @@
...
+	storageConsumerByName := map[string]*ocsv1alpha1.StorageConsumer{}
+	for localClientID, peerClientID := range clientMappingConfig.Data {
+		// for internal mode, we need only blockPool mirroring, hence skipping this for the special key "internal"
+		if localClientID == internalKey {
+			continue
+		}
+		// Check if the storageConsumer with the ClientID exists, this is a fancy get operation as
+		// there will be only one consumer with the clientID
+		storageConsumers := &ocsv1alpha1.StorageConsumerList{}
+		if err := r.list(storageConsumers, client.MatchingFields{clientIDIndexName: localClientID}); err != nil {
```
nb-ohad, line 405: Please cap the list to 2 and make the this a multiline

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,478 @@
...
+		if len(storageConsumers.Items) != 1 {
+			return ctrl.Result{}, fmt.Errorf("expected 1 StorageConsumer but got %v", len(storageConsumers.Items))
+		}
+		storageConsumerByName[storageConsumers.Items[0].Name] = &storageConsumers.Items[0]
+		peerClientIDs = append(peerClientIDs, peerClientID)
+	}
+
+	response, err := ocsClient.GetStorageClientsInfo(
+		r.ctx,
+		storageClusterPeer.Status.PeerInfo.StorageClusterUid,
```
nb-ohad, line 417: Shouldn't we first check if we have a peerInfo with a storageClusterUid value before we send a request. Maybe we even need this to this check as the first check in the function

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,478 @@
...
+	}
+
+	for i := range radosNamespaceList.Items {
+		rns := &radosNamespaceList.Items[i]
+		consumer := storageConsumerByName[rns.GetLabels()[controllers.StorageConsumerNameLabel]]
+		if consumer == nil {
+			continue
+		}
+		remoteClientID := clientMappingConfig.Data[consumer.Status.Client.ID]
+		remoteNamespace := remoteNamespaceByClientID[remoteClientID]
```
nb-ohad, line 441: We should not assume that there is a value in `client.ID` on the consumer's status

nb-ohad (changes requested):

controllers/mirroring/mirroring_controller.go
```diff
@@ -0,0 +1,529 @@
...
+		clientInfo := response.ClientsInfo[i]
+		remoteNamespaceByClientID[clientInfo.ClientID] = clientInfo.RadosNamespace
+	}
+
+	radosNamespaceList := &rookCephv1.CephBlockPoolRadosNamespaceList{}
+	if err = r.list(radosNamespaceList, client.InNamespace(storageClusterPeer.Namespace)); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to list CephBlockPoolRadosNamespace(s): %w", err)
+	}
+
+	for i := range radosNamespaceList.Items {
```
nb-ohad, line 485: Within the loop, you should handle error differently, as we do not want an issue with one block pool to stop the reconciliation of all other block pools

nb-ohad (changes requested):

PR #2879, opened 2024-11-07
controllers: send the client profile mapping required for DR

rewantsoni:
This mapping is sent only if rns is enabled for mirroring

- remote poolID is fetched from GetBlockPoolInfo rpc, which is then written to a blockpool as an annotation to be read by provider server
- local poolID is fetched from cephBlockPool status


services/provider/server/server.go
```diff
@@ -52,18 +52,21 @@ import (
...
+	notAvailable                     = "N/A"
+	rbdMirrorBootstrapPeerSecretName = "rbdMirrorBootstrapPeerSecretName"
 )
 
 const (
 	monConfigMap = "rook-ceph-mon-endpoints"
 	monSecret    = "rook-ceph-mon"
 )
 
+const peerPoolIDConfigMapName = "peer-pool-id"
```
rchikatw, line 68: How about moving this above after line 60.

deploy/ocs-operator/manifests/provider-role.yaml
```diff
@@ -16,6 +16,7 @@ rules:
     resources:
       - cephfilesystemsubvolumegroups
       - cephblockpoolradosnamespaces
+      - cephblockpools
     verbs:
       - get
       - list
```
rchikatw, line 22: For the Ceph block pool, you only need to get access. I am not sure Is it fair to add a new entry for the same API group?

services/provider/server/server.go
```diff
@@ -422,6 +422,45 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
 
 	}
 
+	cbpList := &rookCephv1.CephBlockPoolList{}
+	err = s.client.List(ctx, cbpList, client.InNamespace(s.namespace))
```
nb-ohad, line 426: List only the ones with the annotation set, you have no interest in any other ceph block pool

services/provider/server/server.go
```diff
@@ -422,6 +422,45 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to list cephBlockPools in namespace. %v", err)
+	}
+	if len(cbpList.Items) > 0 {
+		blockPoolMapping := []csiopv1a1.BlockPoolIdPair{}
+		for i := range cbpList.Items {
+			cephBlockPool := cbpList.Items[i]
+			if cephBlockPool.GetAnnotations()[util.BlockPoolMirroringTargetIDAnnotation] != "" {
```
nb-ohad, line 434: There is no point in this check. See my comment about listing only the blockpools with the annotation

services/provider/server/server.go
```diff
@@ -422,6 +422,45 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
+				blockPoolMapping = append(
+					blockPoolMapping,
+					[]string{
+						strconv.Itoa(cephBlockPool.Status.PoolID),
+						cephBlockPool.GetAnnotations()[util.BlockPoolMirroringTargetIDAnnotation],
+					},
+				)
```
nb-ohad, line 441:
```suggestion
				blockPoolMapping[i][0] := strconv.Itoa(cephBlockPool.Status.PoolID),
				blockPoolMapping[i][1] := cephBlockPool.GetAnnotations()
```

services/provider/server/server.go
```diff
@@ -422,6 +422,45 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
 
 	}
 
+	cbpList := &rookCephv1.CephBlockPoolList{}
+	err = s.client.List(ctx, cbpList, client.InNamespace(s.namespace))
+	if err != nil {
+		return nil, fmt.Errorf("failed to list cephBlockPools in namespace. %v", err)
+	}
+	if len(cbpList.Items) > 0 {
+		blockPoolMapping := []csiopv1a1.BlockPoolIdPair{}
```
nb-ohad, line 431: You know the length of this list ahead of time, preallocate it so you can safely index into it

services/provider/server/server.go
```diff
@@ -422,6 +422,45 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
+					blockPoolMapping,
+					[]string{
+						strconv.Itoa(cephBlockPool.Status.PoolID),
+						cephBlockPool.GetAnnotations()[util.BlockPoolMirroringTargetIDAnnotation],
+					},
+				)
+			}
+		}
+
+		if len(blockPoolMapping) > 0 {
```
nb-ohad, line 445: This check does no make sense. Even if we don't have block pool mapping we still need client profile mapping

services/provider/server/server.go
```diff
@@ -422,6 +422,45 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
+			clientName := consumerResource.Status.Client.Name
+			clientProfileName := util.CalculateMD5Hash(fmt.Sprintf("%s-ceph-rbd", clientName))
+			extR = append(extR, &pb.ExternalResource{
+				Name: consumerResource.Status.Client.Name,
+				Kind: "ClientProfileMapping",
+				Data: mustMarshal(&csiopv1a1.ClientProfileMappingSpec{
+					Mappings: []csiopv1a1.MappingsSpec{
+						{
+							LocalClientProfile:  clientProfileName,
+							RemoteClientProfile: clientProfileName,
```
nb-ohad, line 455: How can the local and remote be the same name?

services/provider/server/server.go
```diff
@@ -423,6 +420,48 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to list cephBlockPools in namespace. %v", err)
+	}
+	blockPoolMapping := []csiopv1a1.BlockPoolIdPair{}
+	for i := range cbpList.Items {
+		cephBlockPool := cbpList.Items[i]
+		if cephBlockPool.GetAnnotations()[util.BlockPoolMirroringTargetIDAnnotation] != "" {
+			blockPoolMapping = append(
+				blockPoolMapping,
+				[]string{
```
nb-ohad, line 434: Please use the proper type here:
```suggestion
				csiopv1a1.BlockPoolIdPair{
```

services/provider/server/server.go
```diff
@@ -423,6 +420,48 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
+					strconv.Itoa(cephBlockPool.Status.PoolID),
+					cephBlockPool.GetAnnotations()[util.BlockPoolMirroringTargetIDAnnotation],
+				},
+			)
+		}
+	}
+
+	if len(blockPoolMapping) != 0 {
```
nb-ohad, line 442: Length cannot be smaller then 0
```suggestion
	if len(blockPoolMapping) > 0 {
```

services/provider/server/server.go
```diff
@@ -423,6 +420,48 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
+		/*
+			This is an assumption, and which should go away when we move to converged, and we remove storageClaims
```
nb-ohad, line 444:
```suggestion
			This is an assumption and should go away when deprecating the StorageClaim API
```

deploy/ocs-operator/manifests/provider-role.yaml
rewantsoni: We need list access as we are listing the blockpools with the refactored code

nb-ohad: There is a need for both get and list access for ceph block pools

PR #2883, opened 2024-11-08
rbdmirror: add a blockpool name label in the rados namespace cr

parth-gr:
add a blockpool name in the lable so UI can watch for the rados namespace that belongs to the particular blockpool

### DIFF
--- a/controllers/storagerequest/storagerequest_controller.go
+++ b/controllers/storagerequest/storagerequest_controller.go
@@ -48,6 +48,7 @@ import (
 const (
 	storageRequestFinalizer    = "ocs.openshift.io.storagerequest"
 	forceDeletionAnnotationKey = "rook.io/force-deletion"
+	blockPoolNameLabel         = "ocs.openshift.io/cephblockpool-name"
 )
 
 // StorageRequestReconciler reconciles a StorageRequest object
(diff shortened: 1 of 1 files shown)


rewantsoni (changes requested):
I was wondering if ocs-operator is the right place to add this label and if adding the label from Rook would make more sense. If Rook adds the label then, irrespective of the controller that creates the CR, we will always have the label. As an example, for external mode we might also require monitoring and might have to add the label on the CR from where we are creating radosNS CR for external mode. WDYT?

parth-gr:
> I was wondering if ocs-operator is the right place to add this label and if adding the label from Rook would make more sense. If Rook adds the label then, irrespective of the controller that creates the CR, we will always have the label. As an example, for external mode we might also require monitoring and might have to add the label on the CR from ting radosnamespacce where we are creating radosNS CR for external mode. WDYT?

we are only doing mirroring in provider mode for the rados namespace, in future, we can change it to rook but for now, I think we can be specific

controllers/storagerequest/storagerequest_controller.go
```diff
@@ -48,6 +48,7 @@ import (
 const (
 	storageRequestFinalizer    = "ocs.openshift.io.storagerequest"
 	forceDeletionAnnotationKey = "rook.io/force-deletion"
+	blockPoolNameLabel         = "blockPoolName"
```
rewantsoni, line 51: If we are having this here, could we have something like `ocs.openshift.io/cephblockpool-name`

rewantsoni:
> we are only doing mirroring in provider mode for the rados namespace, in future, we can change it to rook but for now, I think we can be specific

If it is a small fix and we can have it from the rook, I suggest making changes to the rook so that we don't have to change it in the future.

parth-gr:
> > we are only doing mirroring in provider mode for the rados namespace, in future, we can change it to rook but for now, I think we can be specific
> 
> If it is a small fix and we can have it from the rook, I suggest making changes to the rook so that we don't have to change it in the future.

@travisn should we do at rook?
As this label would be unnecessary for other deployments, but it wont hurt

malayparida2000 (approved):
/lgtm

iamniting (approved):

PR #2887, opened 2024-11-11
Fix replication flag assignment

raaizik:
When there are at least two connected peers, the replication flag needs to be enabled for the correct labeling to take place.


rewantsoni:
/lgtm
cc: @iamniting

iamniting (approved):

PR #2891, opened 2024-11-13
client: use storageclient k8s id as the client id

rewantsoni:
Use storageClient k8s UID for clientID instead of storageConsumerUID


rewantsoni:
@nb-ohad Opened a issue for having them in onboarding: https://github.com/red-hat-storage/ocs-operator/issues/2894

PR #2895, opened 2024-11-13
saving quota in GiB instead GB

rchikatw:
-  PVCs are in Gib and quota is in GB because of which pvc creation is restricted when the quota reaches around 95%
- Fix for https://issues.redhat.com/browse/DFBUGS-559

### DIFF
--- a/services/provider/server/server.go
+++ b/services/provider/server/server.go
@@ -61,6 +61,7 @@ const (
 	ramenDRStorageIDKey     = "ramendr.openshift.io/storageID"
 	ramenDRReplicationIDKey = "ramendr.openshift.io/replicationid"
 	ramenDRFlattenModeKey   = "replication.storage.openshift.io/flatten-mode"
+	oneGibInBytes           = 1024 * 1024 * 1024
 )
 
 const (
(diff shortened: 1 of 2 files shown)


rchikatw:
@nb-ohad @leelavg please review

services/provider/server/server.go
```diff
@@ -407,9 +407,9 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
 				},
 			},
 			Quota: corev1.ResourceQuotaSpec{
-				Hard: corev1.ResourceList{"requests.storage": *resource.NewScaledQuantity(
-					int64(consumerResource.Spec.StorageQuotaInGiB),
-					resource.Giga,
+				Hard: corev1.ResourceList{"requests.storage": *resource.NewQuantity(
+					int64(consumerResource.Spec.StorageQuotaInGiB*1024*1024*1024),
```
leelavg, line 411: reading only this line seems a bit confusing, did it occur to you that if only this line is read, it feels like we are multiplying a value which has GiB in it's name by 1024^3 resulting in something like Exbibyte?

maybe it only occurred to me, since the conversion is correct, approving it.

leelavg (approved):

services/provider/server/server.go
leelavg: resolve.

PR #2905, opened 2024-11-21
provider: send vrcs with fixed replication schedule

rewantsoni:
Send the entire spec of VRC that includes schedule, provisioner, clusterID from provider

Sending a VRC with a fixed schedule is temporary for feature freeze, we will work on moving it to dynamic schedules after feature freeze


controllers/util/storageclass.go
nb-ohad, line 1: The file name is `storageclass.go` but the content is only csi driver names, shouldn't this go under `utils/csi.go` or something similar (if already exists)?

services/provider/server/server.go
```diff
@@ -814,24 +816,36 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
 					}),
 					Labels: getExternalResourceLabels("VolumeGroupSnapshotClass", mirroringEnabled, false, replicationID, storageID)},
 				&pb.ExternalResource{
-					Name: "ceph-rbd",
+					Name: req.StorageClaimName,
```
nb-ohad, line 819: Like all other classes, our VRCs should have a fixed name and not one that depends on the request

services/provider/server/server.go
```diff
@@ -814,24 +816,36 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
...
+						Provisioner: util.RbdDriverName,
 					}),
 					Labels: getExternalResourceLabels("VolumeReplicationClass", mirroringEnabled, false, replicationID, storageID),
 					Annotations: map[string]string{
 						"replication.storage.openshift.io/is-default-class": "true",
 					},
 				},
 				&pb.ExternalResource{
-					Name: "ceph-rbd-flatten",
+					Name: fmt.Sprintf("%s-flatten", req.StorageClaimName),
```
nb-ohad, line 837: Like all other classes, our VRCs should have a fixed name and not one that depends on the request

services/provider/server/server_test.go
```diff
@@ -658,12 +661,17 @@ func TestOCSProviderServerGetStorageClaimConfig(t *testing.T) {
 					"ramendr.openshift.io/storageID": "8d40b6be71600457b5dec219d2ce2d4c",
 				},
 			},
-			"ceph-rbd-volumereplicationclass": {
-				Name: "ceph-rbd",
+			"block-pool-claim-volumereplicationclass": {
```
nb-ohad, line 664: Please keep the original name which is the name of the VRC

nb-ohad (changes requested):

services/provider/server/server.go
```diff
@@ -731,6 +731,8 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
 	replicationEnabled := len(scp.Items) > 0
 	replicationID := req.StorageClaimName
 
+	clientProfile := util.CalculateMD5Hash(req.StorageClaimName)
```
leelavg, line 734: this could become problematic when we refactor the responsibility of provider & client side code, better be consistent and let client-op set this.

leelavg (commented):
either expand the existing comment on why this is temporary or the PR description, ie, a bit of technical details.

services/provider/server/server.go
rewantsoni: It was already decided that we are going to set clusterID from the provider

leelavg (approved):

services/provider/server/server.go
```diff
@@ -816,9 +818,15 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
 				&pb.ExternalResource{
 					Name: "ceph-rbd",
 					Kind: "VolumeReplicationClass",
-					Data: mustMarshal(map[string]string{
-						"replication.storage.openshift.io/replication-secret-name": provisionerSecretName,
-						"mirroringMode": "snapshot",
+					Data: mustMarshal(&replicationv1alpha1.VolumeReplicationClassSpec{
```
iamniting, line 821: Is the client operator parsing this info now instead of a map?

rewantsoni: Yes, the client operator is parsing based on the the kind we send

PR #2906, opened 2024-11-22
provider: generate vrcs name to align with internal mode

rewantsoni:


services/provider/server/server.go
```diff
@@ -64,6 +64,10 @@ const (
 	oneGibInBytes           = 1024 * 1024 * 1024
 	monConfigMap            = "rook-ceph-mon-endpoints"
 	monSecret               = "rook-ceph-mon"
+
+	RBDVolumeReplicationClassNameTemplate        = "rbd-volumereplicationclass-%v"
+	RBDFlattenVolumeReplicationClassNameTemplate = "rbd-flatten-volumereplicationclass-%v"
```
nb-ohad, line 69: No need for these, they are only used once and you can easily just inline them

nb-ohad (changes requested):

PR #2909, opened 2024-11-26
controllers: fix mirroring controller

rewantsoni:
1. Update the mirroring finalizer to the right name
2. fix radosnamespace mirroring


nb-ohad:
\lgtm

PR #2911, opened 2024-11-28
controllers: fix dynamic crd predicate

leelavg:
if a single predicate reject the event then the runtime doesn't queue the reconcile for that controller and for dynamic crd feature multiple crds should be checked for event and at the same time presence of a single matching crd with corresponding predicate functions should initiate the reconcile.

this commit uses helper function in predicate and creates the required logical conditions and refactor the crd predicate to directly take bools based on crd availability.

### DIFF
--- a/controllers/storagecluster/storagecluster_controller.go
+++ b/controllers/storagecluster/storagecluster_controller.go
@@ -233,12 +233,18 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			&extv1.CustomResourceDefinition{},
 			enqueueStorageClusterRequest,
 			builder.WithPredicates(
-				util.NamePredicate(VirtualMachineCrdName),
-				util.CrdCreateAndDeletePredicate(&r.Log, VirtualMachineCrdName, r.AvailableCrds[VirtualMachineCrdName]),
-			),
-			builder.WithPredicates(
-				util.NamePredicate(StorageClientCrdName),
-				util.CrdCreateAndDeletePredicate(&r.Log, StorageClientCrdName, r.AvailableCrds[StorageClientCrdName]),
+				predicate.Or(
+					util.NamePredicate(VirtualMachineCrdName),
+					util.NamePredicate(StorageClientCrdName),
+				),
+				util.EventTypePredicate(
+					// the values in the below map are filled before controller watches are setup and these conditions
+					// will just be evaluated to a boolean by the time builder completes setting up watches.
+					!r.AvailableCrds[VirtualMachineCrdName] || !r.AvailableCrds[StorageClientCrdName],
+					false,
+					true,
+					false,
+				),
 			),
 			builder.OnlyMetadata,
 		).
--- a/controllers/ocsinitialization/ocsinitialization_controller.go
+++ b/controllers/ocsinitialization/ocsinitialization_controller.go
@@ -355,7 +355,12 @@ func (r *OCSInitializationReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			enqueueOCSInit,
 			builder.WithPredicates(
 				util.NamePredicate(ClusterClaimCrdName),
-				util.CrdCreateAndDeletePredicate(&r.Log, ClusterClaimCrdName, r.AvailableCrds[ClusterClaimCrdName]),
+				util.EventTypePredicate(
+					!r.AvailableCrds[ClusterClaimCrdName],
+					false,
+					true,
+					false,
+				),
 			),
 			builder.OnlyMetadata,
 		)
--- a/controllers/util/predicates.go
+++ b/controllers/util/predicates.go
@@ -1,10 +1,8 @@
 package util
 
 import (
-	"fmt"
 	"reflect"
 
-	"github.com/go-logr/logr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"sigs.k8s.io/controller-runtime/pkg/event"
@@ -81,25 +79,23 @@ func NamePredicate(name string) predicate.Predicate {
 	})
 }
 
-func CrdCreateAndDeletePredicate(log *logr.Logger, crdName string, crdExists bool) predicate.Predicate {
+// EventTypePredicate return a predicate to filter events based on their
+// respective event type. This helper allows for the selection of multiple
+// types resulting in a predicate that can filter in more than a single event
+// type
+func EventTypePredicate(create, update, del, generic bool) predicate.Predicate {
 	return predicate.Funcs{
 		CreateFunc: func(_ event.CreateEvent) bool {
-			if !crdExists {
-				log.Info(fmt.Sprintf("CustomResourceDefinition %s was Created.", crdName))
-			}
-			return !crdExists
-		},
-		DeleteFunc: func(_ event.DeleteEvent) bool {
-			if crdExists {
-				log.Info(fmt.Sprintf("CustomResourceDefinition %s was Deleted.", crdName))
-			}
-			return crdExists
+			return create
 		},
 		UpdateFunc: func(_ event.UpdateEvent) bool {
-			return false
+			return update
+		},
+		DeleteFunc: func(_ event.DeleteEvent) bool {
+			return del
 		},
 		GenericFunc: func(_ event.GenericEvent) bool {
-			return false
+			return generic
 		},
 	}
 }
--- a/metrics/vendor/github.com/red-hat-storage/ocs-operator/v4/controllers/util/predicates.go
+++ b/metrics/vendor/github.com/red-hat-storage/ocs-operator/v4/controllers/util/predicates.go
@@ -1,10 +1,8 @@
 package util
 
 import (
-	"fmt"
 	"reflect"
 
-	"github.com/go-logr/logr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"sigs.k8s.io/controller-runtime/pkg/event"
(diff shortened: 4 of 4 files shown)


leelavg:
/hold

for testing.

leelavg:
@rewantsoni ptal.

leelavg:
> if a single predicate reject the event then the runtime doesn't queue the reconcile 
- sample https://github.com/kubernetes-sigs/controller-runtime/blob/8e44a4307f70a4d5f4c5aaf0869e2bec7167f4ab/pkg/internal/source/event_handler.go#L83-L87
``` go
	for _, p := range e.predicates {
		if !p.Create(c) {
			return
		}
	}
```
similar for other events can be found in same file.

rewantsoni:
We have moved to [EventTypePredicate](https://github.com/red-hat-storage/ocs-client-operator/pull/264/files#diff-48091e6567647fd1b847536d158deac0116296d67d20f8f43efeb37bdf86118bR21) in ocs-client op to make it more generic, could we align it here as well?

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -233,12 +233,10 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-				util.NamePredicate(VirtualMachineCrdName),
-				util.CrdCreateAndDeletePredicate(&r.Log, VirtualMachineCrdName, r.AvailableCrds[VirtualMachineCrdName]),
-			),
-			builder.WithPredicates(
-				util.NamePredicate(StorageClientCrdName),
-				util.CrdCreateAndDeletePredicate(&r.Log, StorageClientCrdName, r.AvailableCrds[StorageClientCrdName]),
+				util.ComposePredicates(
+					util.CrdCreateAndDeletePredicate(&r.Log, VirtualMachineCrdName, r.AvailableCrds[VirtualMachineCrdName]),
+					util.CrdCreateAndDeletePredicate(&r.Log, StorageClientCrdName, r.AvailableCrds[StorageClientCrdName]),
+				),
```
rewantsoni, line 239: We could allow the NamePredicate to accept multiple CRD names and instead of using the ComposePredicates, we could use predicates.Or, which does a logical Or of the predicates. 
Also we should move to more generic EventTypePredicate as mentioned in the previous comment
```suggestion
				util.NamePredicate(VirtualMachineCrdName,StorageClientCrdName),
				predicate.Or(
					util.EventTypePredicate(
						!r.AvailableCrds[VirtualMachineCrdName],
						false,
						r.AvailableCrds[VirtualMachineCrdName],
						false,
					),
					util.EventTypePredicate(
						!r.AvailableCrds[StorageClientCrdName],
						false,
						r.AvailableCrds[StorageClientCrdName],
						false,
					),
				),
```

rewantsoni (changes requested):

controllers/storagecluster/storagecluster_controller.go
leelavg: Then the reconcile can get trigerred for wrong reason, like `. Or` isn't scoped for the object which trigerred the event.

Basically we want an And b/n event obj name & event type (create/update/delete/generic etc)

rewantsoni:
```
				predicate.Or(
					predicate.And(
						util.NamePredicate(VirtualMachineCrdName),
						util.EventTypePredicate(
							!r.AvailableCrds[VirtualMachineCrdName],
							false,
							r.AvailableCrds[VirtualMachineCrdName],
							false,
						),
					),
					predicate.And(
						util.NamePredicate(StorageClientCrdName),
						util.EventTypePredicate(
							!r.AvailableCrds[StorageClientCrdName],
							false,
							r.AvailableCrds[StorageClientCrdName],
							false,
						),
					),
				),
```
I would prefer this over CrdCreateAndDeletePredicate, but I would let the other maintainers decide the right approach

leelavg: much better, meets the requirement, will change.

leelavg: done.

rewantsoni: ack, can resolve

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -233,12 +233,26 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			&extv1.CustomResourceDefinition{},
 			enqueueStorageClusterRequest,
 			builder.WithPredicates(
-				util.NamePredicate(VirtualMachineCrdName),
-				util.CrdCreateAndDeletePredicate(&r.Log, VirtualMachineCrdName, r.AvailableCrds[VirtualMachineCrdName]),
-			),
-			builder.WithPredicates(
-				util.NamePredicate(StorageClientCrdName),
-				util.CrdCreateAndDeletePredicate(&r.Log, StorageClientCrdName, r.AvailableCrds[StorageClientCrdName]),
+				predicate.Or(
```
malayparida2000, line 236: Not able to understand the calling pattern of this function. Can you please add few lines of comment here to explain what exactly we are trying to achieve here, will be helpful for future changes as well

leelavg: > Not able to understand the calling pattern of this function
- may I know if the issue was understood based on the PR description? let me know if below makes sense and I'll add the same as comment.

requirement:
1. if the required CRDs state changes we want to reconcile and the interested state is if there are created or deleted
2. the function accepts booleans corresponding to create, update, delete and generic. we always supply false to update & generic as we don't need reconcile for that
3. now we want to get reconcile if
a. if crd didn't exist at start of controller and now it became availabe i.e, options for create event
b. if crd exist at start of controller and now it is deleted ie., option for delete event
4. point 3 is for a specific crd and we want the same for multiple CRDs and so the condition is `OR( (name AND event) ... )`

malayparida2000: Thanks for the detailed explanation, I understand now.

leelavg:
/unhold

Loaned a cluster which already had provider-client setup and didn't test for virt sc creation but tested for restart when the crd state changes as follows

```
disregard the slight time diff, seems to be time skew b/n my lab machine and ocp in another region.
================================= when crd is installed, ocs-op restarts

> date --utc && k create -f virt-crd.yaml
Wed Dec  4 06:41:35 UTC 2024
customresourcedefinition.apiextensions.k8s.io/virtualmachines.kubevirt.io created

[...]

{"level":"info","ts":"2024-12-04T06:40:48Z","msg":"Running StorageRequest controller in Converged/Provider Mode","controller":"storagerequest","controllerGroup":"ocs.openshift.io","controllerKind":"StorageRequest","StorageRequest":{"name":"storagerequest-64967a74c3cc43e7148b8ca9c978de2d","namespace":"openshift-storage"},"namespace":"openshift-storage","name":"storagerequest-64967a74c3cc43e7148b8ca9c978de2d","reconcileID":"3f501dbd-0859-48f8-99f7-0340febc8aaa","StorageRequest":{"name":"storagerequest-64967a74c3cc43e7148b8ca9c978de2d","namespace":"openshift-storage"}}
{"level":"info","ts":"2024-12-04T06:40:53Z","logger":"cmd","msg":"Go Version: go1.22.10"}
{"level":"info","ts":"2024-12-04T06:40:53Z","logger":"cmd","msg":"Go OS/Arch: linux/amd64"}
{"level":"info","ts":"2024-12-04T06:40:53Z","logger":"cmd","msg":"Cluster is running on OpenShift."}

================================= when crd is deleted, ocs-op restarts

> date --utc && k delete -f virt-crd.yaml
Wed Dec  4 06:43:44 UTC 2024
customresourcedefinition.apiextensions.k8s.io "virtualmachines.kubevirt.io" deleted

[...]

{"level":"info","ts":"2024-12-04T06:42:48Z","msg":"Running StorageRequest controller in Converged/Provider Mode","controller":"storagerequest","controllerGroup":"ocs.openshif
t.io","controllerKind":"StorageRequest","StorageRequest":{"name":"storagerequest-64967a74c3cc43e7148b8ca9c978de2d","namespace":"openshift-storage"},"namespace":"openshift-sto
rage","name":"storagerequest-64967a74c3cc43e7148b8ca9c978de2d","reconcileID":"d2a74831-d025-4fda-afcc-3cabe2c613ab","StorageRequest":{"name":"storagerequest-64967a74c3cc43e71
48b8ca9c978de2d","namespace":"openshift-storage"}}
{"level":"info","ts":"2024-12-04T06:43:01Z","logger":"cmd","msg":"Go Version: go1.22.10"}
{"level":"info","ts":"2024-12-04T06:43:01Z","logger":"cmd","msg":"Go OS/Arch: linux/amd64"}
{"level":"info","ts":"2024-12-04T06:43:01Z","logger":"cmd","msg":"Cluster is running on OpenShift."}
{"level":"info","ts":"2024-12-04T06:43:01Z","logger":"cmd","msg":"OCSInitialization resource already exists"}
```

leelavg:
rebased and added a comment.

nb-ohad:
@leelavg we do not want to couple to predicate, each one should stay separate. The event predicate is a generic predicate to filter events based on event type. If the relationship between the predicates is wrong (and AND operation) let's fix that and not change the predicate themselves. Same goes for the order

leelavg:
> let's fix that and not change the predicate themselves
- the fix is to change the predicate, older predicate function is updated in client-op and we reusing it here as well.

nb-ohad:
Looking at the code again. You are using the tools the wrong way which complicates things. 

Your condition is logically 2 distinct conditions:
- filter for a fixed list of CRD names 
- filter events for creation only if CRD is does not exists or it is a delete event

This boils down to filtering first on names, then filtering on event types:
```
predicate.Or(
    util.NamePredicate(crd1Name),
    util.NamePredicaet(crd2Name),
),
util.EventTypePredicate(
    !crd1Available || !crd2Available2,
    false,
    true,
    false,
)
```

leelavg:
> Can you please describe your filtering logic in words first 
- I mentioned the "why" in pr description and "what" is being done in the code comments.

> Your condition is logically 2 distinct conditions:
filter for a fixed list of CRD names
filter events for creation only if CRD is does not exists or it is a delete event
- I don't think so, the event (create/delete) for a crd (name) is one distinct set. With the suggestion, I need the change for addition or deletion on watching new CRD in two places

> or it is a delete event
- not for all delete events, if the required crd didn't exist when we started the controller we don't want a reconcile if a delete event is trigerred on that

nb-ohad:
>I don't think so, the event (create/delete) for a crd (name) is one distinct set. With the suggestion, I need the change for addition or deletion on watching new CRD in two places

No it doesn't, the queued event does not care 

> not for all delete events, if the required crd didn't exist when we started the controller we don't want a reconcile if a delete event is trigerred on that

This is why we have the named filter to only trigger for named CRDs. Add to that the fact that you cannot get a DELETE event for a CR that does not exists from the start

leelavg:
> This boils down to filtering first on names, then filtering on event types:
- took the suggestion based on an offline discussion, thanks.

leelavg:
https://github.com/red-hat-storage/ocs-operator/pull/2911#issuecomment-2516932679
/unhold based on above

leelavg:
/unhold
based on https://github.com/red-hat-storage/ocs-operator/pull/2911#issuecomment-2516932679

PR #2914, opened 2024-12-02
provider: send mirrorEnabled info for each client

rewantsoni:
Extend SystemAttributes to send to the client if it has been mirrorEnabled, currently this will be used to enable omap info on the client clusters


services/provider/server/server.go
```diff
@@ -1266,3 +1281,17 @@ func (s *OCSProviderServer) isSystemInMaintenanceMode(ctx context.Context) (bool
...
+	clientMappingConfig.Name = util.StorageClientMappingConfigName
+	clientMappingConfig.Name = s.namespace
+	err := s.client.Get(ctx, client.ObjectKeyFromObject(clientMappingConfig), clientMappingConfig)
+	if client.IgnoreNotFound(err) != nil {
+		return false, err
+	} else if kerrors.IsNotFound(err) {
+		return false, nil
+	}
+
+	return clientMappingConfig.Data != nil && clientMappingConfig.Data[consumer.Status.Client.ID] != "", nil
```
nb-ohad, line 1296: No need for the first check. In go a nil map can still be queried 
```suggestion
	return clientMappingConfig.Data[consumer.Status.Client.ID] != "", nil
```

services/provider/server/server.go
```diff
@@ -1266,3 +1281,17 @@ func (s *OCSProviderServer) isSystemInMaintenanceMode(ctx context.Context) (bool
...
+func (s *OCSProviderServer) isConsumerMirrorEnabled(ctx context.Context, consumer *ocsv1alpha1.StorageConsumer) (bool, error) {
+	clientMappingConfig := &corev1.ConfigMap{}
+	clientMappingConfig.Name = util.StorageClientMappingConfigName
+	clientMappingConfig.Name = s.namespace
+	err := s.client.Get(ctx, client.ObjectKeyFromObject(clientMappingConfig), clientMappingConfig)
+	if client.IgnoreNotFound(err) != nil {
+		return false, err
+	} else if kerrors.IsNotFound(err) {
+		return false, nil
+	}
```
nb-ohad, line 1294:
```suggestion
	if err := s.client.Get(ctx, client.ObjectKeyFromObject(clientMappingConfig), clientMappingConfig); err != nil {
		return false, client.IgnoreNotFound(err)
	}
```

nb-ohad (changes requested):

PR #2918, opened 2024-12-03
provider: fix ramen labels

rewantsoni:
The PR does the following:
1. ramen label for storageid was misspelled as storageID 
2. add the maintenancemode label to the VRC
3. always add the labels for ramen


ShyamsundarR (commented):
Label names, values and resources that need to be labelled are correct, LGTM!

PR #2919, opened 2024-12-04
DFBUGS-976: monitoring: mark mirroring enabled if spec.Mirroring is !nil

weirdwiz:
consider mirroring enabled if spec.Mirroring for
CephBlockPoolRadosNamespace is not nil


weirdwiz:
ptal @umangachapagain @aruniiird @rewantsoni

umangachapagain (approved):

PR #2924, opened 2024-12-05
filesystem: correct the pool name in the ux backend handler

parth-gr:
In the expandstorage handler pool name is incorrect,

Rook adds the filesystemName prefix to every datapool, which was missing in the storageclass parameter
https://github.com/rook/rook/blob/44e96c8ab67a27121354a04b121b7f16b6a38894/pkg/operator/ceph/file/filesystem.go#L321

### DIFF
--- a/services/ux-backend/handlers/expandstorage/handler.go
+++ b/services/ux-backend/handlers/expandstorage/handler.go
@@ -268,7 +268,7 @@ func createCephFilesystemStorageClass(w http.ResponseWriter, r *http.Request, cl
 		Parameters: map[string]string{
 			"clusterID": namespace,
 			"fsName":    filesystemName,
-			"pool":      poolName,
+			"pool":      fmt.Sprintf("%s-%s", filesystemName, poolName),
 			"csi.storage.k8s.io/provisioner-secret-name":            "rook-csi-cephfs-provisioner",
 			"csi.storage.k8s.io/provisioner-secret-namespace":       namespace,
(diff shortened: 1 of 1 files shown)


services/ux-backend/handlers/expandstorage/handler.go
```diff
@@ -268,7 +268,7 @@ func createCephFilesystemStorageClass(w http.ResponseWriter, r *http.Request, cl
 		Parameters: map[string]string{
 			"clusterID": namespace,
 			"fsName":    filesystemName,
-			"pool":      poolName,
+			"pool":      fmt.Sprintf("%s-%s", filesystemName, poolName),
```
Madhu-1, line 271: is pool name mandatory when we expand the storage? if not Rook appends different value with integers

parth-gr: Yaa in the UI the pool name is mandatory,
cc @SanjalKatiyar to keep me honest

Madhu-1: in that case lets make it mandatory here as well

SanjalKatiyar: yeah, UI has it as a mandatory field.

iamniting (approved):
looks good to me, holding for others review. 

/hold

SanjalKatiyar (commented):
LGTM

PR #2936, opened 2024-12-18
testing the waters for convergence

leelavg:
intention of this PR is to have storageclasses and volumesnapshotclasses be working for creation, expansion, deletion of storage in below scenarios and more importantly to reason about what all codepaths need to be changed while starting the actual work as whole design for this effort can't be documented w/o testing the waters!
1. upgraded internal mode cluster
2. upgraded provider mode cluster
3. new install

new content that'll not be part of this PR:
1. webhook defaulting the params of storageclass
2. any adapters for UI to call into backend
3. distribution of storageclasses

existing content that is left out:
1. external mode
2. removal of internal storageclasses reconciliation
3. handling various storage features: nfs, keyrotation, topology/affinity
4. upstream CI script changes to always install client-op
5. rdr specifics

Any unknowns on becoming knows will be added to one of above subheads.


leelavg:
testing will follow.

PR #2937, opened 2024-12-19
Unset CephBlockPool mirroring if Mirroring spec is nil on SC

vbnrh:
https://issues.redhat.com/browse/DFBUGS-1185

### DIFF
--- a/controllers/storagecluster/cephblockpools.go
+++ b/controllers/storagecluster/cephblockpools.go
@@ -94,12 +94,14 @@ func (o *ocsCephBlockPools) reconcileCephBlockPool(r *StorageClusterReconciler, 
 		cephBlockPool.Spec.PoolSpec.Replicated = generateCephReplicatedSpec(storageCluster, "data")
 		cephBlockPool.Spec.PoolSpec.EnableRBDStats = true
 
-		if storageCluster.Spec.Mirroring != nil {
-			if storageCluster.Spec.Mirroring.Enabled {
+		// Since provider mode handles mirroring, we only need to handle for internal mode
+		if storageCluster.Annotations["ocs.openshift.io/deployment-mode"] != "provider" {
+			if storageCluster.Spec.Mirroring != nil && storageCluster.Spec.Mirroring.Enabled {
 				cephBlockPool.Spec.PoolSpec.Mirroring.Enabled = true
 				cephBlockPool.Spec.PoolSpec.Mirroring.Mode = "image"
 				cephBlockPool.Spec.PoolSpec.Mirroring.Peers = o.addPeerSecretsToCephBlockPool(r, storageCluster, cephBlockPool.Name, cephBlockPool.Namespace)
 			} else {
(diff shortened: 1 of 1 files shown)


rewantsoni:
/hold 
This will cause problems with mirroring controller. As for provider mode, we will try to enable mirroring from the mirroring controller and the storageCluster controller will try to disable the mirroring.

rewantsoni:
What we could do is,
```
if the storageCluster does not have provider mode annotation ( ocs.openshift.io/deployment-mode!= "provider")
  if storageCluster.Spec.Mirroring !=nil && storageCluster.Spec.Mirroring.Enabled
     enable mirroring and set the peer secrets
  else 
     disable mirroring
```
@umangachapagain @vbnrh WDYT?
This way, MCO could disable mirroring by setting the MirroringSpec as empty and the storageCluster controller would disable mirroring/use this field in non-provider mode.

And with converged when we move to use the mirroring controller to manage mirroring/secret exchange, we could remove enabling mirroring from the storageCluster controller (cephblockpool.go)

controllers/storagecluster/cephblockpools.go
```diff
@@ -257,12 +257,19 @@ func (o *ocsCephBlockPools) reconcileNonResilientCephBlockPool(r *StorageCluster
...
 
-			if storageCluster.Spec.Mirroring != nil {
-				if storageCluster.Spec.Mirroring.Enabled {
+			deploymentType, ok := storageCluster.Annotations["ocs.openshift.io/deployment-mode"]
+			if !ok {
+				return fmt.Errorf("deployment mode annotation not found")
+			}
+
+			// Since provider mode handles mirroring, we only need to handle for converged mode
+			if deploymentType != "provider" {
```
rewantsoni, line 266:
```suggestion
			// Since provider mode handles mirroring, we only need to handle for internal mode
			if storageCluster.Annotations["ocs.openshift.io/deployment-mode"] != "provider" {
```

rewantsoni (changes requested):
There are other places in this file where we are enabling mirroring, please make sure that we follow the same in other places as well

controllers/storagecluster/cephblockpools.go
vbnrh: Annotation is guaranteed to be there in all cases ?

rewantsoni: No, it will be present only in provider mode, but if we query an empty map with a key that is not present it will return an empty value

rewantsoni:
Were you able to verify it on a cluster?

rewantsoni:
It would be good if you could verify it so we don't miss any edge cases. Maybe add the annotation to the internal Mode cluster and check that storageCluster doesn't reconcile the mirroring field. Anyway, I will un-hold the PR.
/unhold

leelavg:
@rewantsoni pls be aware that most if not all the traces of provider mode will be removed and this specific annotation will also go away starting in #2936 (4.19)

vbnrh:
@umangachapagain @leelavg @rewantsoni Can we merge this ? Tested it, it looks good to me.

umangachapagain (approved):

PR #2938, opened 2025-01-02
remove MDSCacheUsageHigh prometheus alert

sp98:
This is related to - https://issues.redhat.com/browse/DFBUGS-368 

Customers are complaining about erroneous MDS cache usage alerts. Ceph team suggested that `ceph_mds_mem_rss` might not be the right metric to capture this cache usage. So this alert needs to looked at again. For the time being, we can just remove this alert while looking for a better solution.  Decision to remove this was discussed in Weekly meeting.

Details about the discussion with the Ceph team:

`rss` is not right metric for cache warning. `mds_co_bytes` is the correct metric but its not exposed by Ceph.
MDS pod memory is 8Gi set by ODF is low. (Not related to this issue but just a general observation)
Possible Resolution:

Operator should be able to read the MDS cache warning from the Ceph health and raise a promethus alert.
If MDS pod memory usage breaches 5/10/15% of the allocated memory size (8Gi), raise a prometheus alert.

### DIFF
--- a/metrics/deploy/prometheus-ocs-rules.yaml
+++ b/metrics/deploy/prometheus-ocs-rules.yaml
@@ -367,19 +367,6 @@ spec:
         severity: info
   - name: ceph-daemon-performance-alerts.rules
     rules:
-    - alert: MDSCacheUsageHigh
-      annotations:
-        description: MDS cache usage for the daemon {{ $labels.ceph_daemon }} has
-          exceeded above 95% of the requested value. Increase the memory request for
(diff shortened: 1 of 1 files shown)


aruniiird (approved):
LGTM
AFAIK, this is the only place from which the alert has to be removed (as far as this repo is concerned)

malayparida2000 (approved):
/lgtm
/cc @iamniting

iamniting (approved):

PR #2955, opened 2025-01-08
external: fix sc name for rados namespace #2955

parth-gr:
Currently, the sc name is hardcoded,
in future if we support multiple rados namespace per tenant we need a variable name 
updating sc names to contain rados-namespace name
rook pr: https://github.com/rook/rook/pull/15243

### DIFF
--- a/controllers/storagecluster/external_resources.go
+++ b/controllers/storagecluster/external_resources.go
@@ -25,15 +25,15 @@ import (
 )
 
 const (
-	externalClusterDetailsSecret          = "rook-ceph-external-cluster-details"
-	externalClusterDetailsKey             = "external_cluster_details"
-	cephFsStorageClassName                = "cephfs"
-	cephRbdStorageClassName               = "ceph-rbd"
-	cephRbdRadosNamespaceStorageClassName = "ceph-rbd-rados-namespace"
-	cephRbdTopologyStorageClassName       = "ceph-rbd-topology"
-	cephRgwStorageClassName               = "ceph-rgw"
-	externalCephRgwEndpointKey            = "endpoint"
-	cephRgwTLSSecretKey                   = "ceph-rgw-tls-cert"
+	externalClusterDetailsSecret                = "rook-ceph-external-cluster-details"
+	externalClusterDetailsKey                   = "external_cluster_details"
+	cephFsStorageClassName                      = "cephfs"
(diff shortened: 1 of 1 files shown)


iamniting (approved):

controllers/storagecluster/external_resources.go
```diff
@@ -375,7 +375,7 @@ func (r *StorageClusterReconciler) createExternalStorageClusterResources(instanc
 				enableRookCSICephFS = true
 			} else if d.Name == cephRbdStorageClassName {
 				scc = newCephBlockPoolStorageClassConfiguration(instance)
-			} else if d.Name == cephRbdRadosNamespaceStorageClassName {
+			} else if strings.Contains(d.Name, cephRbdRadosNamespaceStorageClassName) { // ceph-rbd-rados-namespace-<radosNamespaceName>
```
Madhu-1, line 378: using contains might be a bad idea isn't it? what if I have 2 Radosnamespace with same name but one with different prefix?

parth-gr: So this prefix will be constant, and name should be always different/unique,
We will be passing the storageclass name from the python script, https://github.com/rook/rook/pull/15243
And this check is to just make sure it class this `else if` case if its has the prefix name.

Madhu-1: instead of contains use HasPrefix you want to check for specific prefix?

controllers/storagecluster/external_resources.go
```diff
@@ -375,7 +375,7 @@ func (r *StorageClusterReconciler) createExternalStorageClusterResources(instanc
 				enableRookCSICephFS = true
 			} else if d.Name == cephRbdStorageClassName {
 				scc = newCephBlockPoolStorageClassConfiguration(instance)
-			} else if d.Name == cephRbdRadosNamespaceStorageClassName {
+			} else if strings.HasPrefix(d.Name, cephRbdRadosNamespaceStorageClassName) { // ceph-rbd-rados-namespace-<radosNamespaceName>
```
Madhu-1, line 378: please update the variable name to `cephRbdRadosNamespaceStorageClassNamePrefix` for better code readbility.

parth-gr: Done

PR #2957, opened 2025-01-09
controllers: disable mirroring when peer cluster is down

rewantsoni:
allow the mirroring to be disabled when the peer site is down

Bug: https://issues.redhat.com/browse/DFBUGS-1258


controllers/mirroring/mirroring_controller.go
```diff
@@ -209,13 +209,16 @@ func (r *MirroringReconciler) reconcilePhases(clientMappingConfig *corev1.Config
 		return ctrl.Result{RequeueAfter: 3 * time.Second}, nil
 	}
 
+	errorOccurred := false
+
 	ocsClient, err := providerClient.NewProviderClient(r.ctx, storageClusterPeer.Spec.ApiEndpoint, util.OcsClientTimeout)
 	if err != nil {
-		return ctrl.Result{}, fmt.Errorf("failed to create a new provider client: %v", err)
+		r.log.Error(err, "failed to create a new provider client")
```
leelavg, line 216: if grpcClient isn't required b/n when it's created and used why can't the creation be moved to usage site?

controllers/mirroring/mirroring_controller.go
```diff
@@ -209,13 +209,16 @@ func (r *MirroringReconciler) reconcilePhases(clientMappingConfig *corev1.Config
...
+		errorOccurred = true
+	}
+	if ocsClient != nil {
+		defer ocsClient.Close()
 	}
-	defer ocsClient.Close()
```
leelavg, line 216: where is the connection being closed now? leaking connections perhaps?

controllers/mirroring/mirroring_controller.go
rewantsoni: agreed, I will move creation to usage site

rewantsoni:
/unhold 
Tested it on the QE Cluster when the site is down the, Failover was completed. The cephRBDMirror CR was deleted when the MaintenanceMode was created.

controllers/mirroring/mirroring_controller.go
```diff
@@ -211,13 +210,16 @@ func (r *MirroringReconciler) reconcilePhases(clientMappingConfig *corev1.Config
...
 
+	errorOccurred := false
+
 	ocsClient, err := providerClient.NewProviderClient(r.ctx, storageClusterPeer.Spec.ApiEndpoint, util.OcsClientTimeout)
 	if err != nil {
-		return ctrl.Result{}, fmt.Errorf("failed to create a new provider client: %v", err)
+		r.log.Error(err, "failed to create a new provider client")
+		errorOccurred = true
+	}
+	if ocsClient != nil {
```
leelavg, line 220: is there a possibility that err is nil and still ocsClient isn't created or else no need for this check.

controllers/mirroring/mirroring_controller.go
```diff
@@ -31,10 +28,12 @@ import (
 
 	"github.com/go-logr/logr"
 	rookCephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
+	"golang.org/x/exp/maps"
```
leelavg, line 31: check if you can remove this, all exp of maps is already in stdlib afair.

controllers/mirroring/mirroring_controller.go
rewantsoni: Since I am not returning the error in the previous if block, we need to check if ocsClient is nil or not before closing it

controllers/mirroring/mirroring_controller.go
rewantsoni: Tried removing this and it was throwing error

controllers/mirroring/mirroring_controller.go
rewantsoni: done

iamniting (approved):

PR #2963, opened 2025-01-10
osd: add tunefastdeviceclass: true for all ssd disk

parth-gr:
with exandstorage ux-backend handler we will be
adding ssd disk always,
So by default make the deviceSet spec, `deviceType: ssd`


malayparida2000 (approved):
/lgtm

iamniting (approved):

PR #2965, opened 2025-01-12
Adjust Node Label Conditions Based on Full Label Name on strechcluster

OdedViner:


api/v1/topologymap.go
```diff
@@ -76,3 +79,19 @@ func (m *NodeTopologyMap) GetKeyValues(topologyKey string) (string, []string) {
 
 	return topologyKey, values
 }
+
+// GetKeyValues returns a node label matching the topologyKey and all values for StrechCluster
+// for that label across all storage nodes
```
Nikhil-Ladha, line 84: Please update the comment as per the working of this function.

api/v1/topologymap.go
```diff
@@ -76,3 +79,19 @@ func (m *NodeTopologyMap) GetKeyValues(topologyKey string) (string, []string) {
...
+func (m *NodeTopologyMap) GetKeyValuesStrechCluster() (string, []string) {
+	values := []string{}
+	topologyKey := ""
+	for label, labelValues := range m.Labels {
+		if label == corev1.LabelZoneFailureDomainStable || label == labelZoneFailureDomainWithoutBeta {
+			topologyKey = label
+			values = labelValues
+			break
+		}
+	}
```
Nikhil-Ladha, line 94: This means for stretch cluster we can't have a different failure zone? Like, `rack`, `host`?

Nikhil-Ladha (changes requested):

api/v1/topologymap.go
Nikhil-Ladha: We should update the `GetKeysValues` function to do an equality check with valid labels that we define in the `validTopologyLabelKeys` and return the key, values accordingly.

Nikhil-Ladha: @malayparida2000 @iamniting what do you guys think?

api/v1/topologymap.go
```diff
@@ -17,9 +17,12 @@ limitations under the License.
 package v1
 
 import (
+	corev1 "k8s.io/api/core/v1"
 	"strings"
 )
 
+const labelZoneFailureDomainWithoutBeta = "failure-domain.kubernetes.io/zone"
```
malayparida2000, line 24: Wasn't this already defined somewhere in the codebase?

malayparida2000 (commented):
@OdedViner Can you add some details on what's the problem with the existing code? And what's the exact issue. It's difficult to understand the motive for the change without knowing the root cause you are trying to fix.

OdedViner:
> @OdedViner Can you add some details on what's the problem with the existing code? And what's the exact issue. It's difficult to understand the motive for the change without knowing the root cause you are trying to fix.

Root Cause:
When incorrect labels are matched, it updates the NodeTopologies in the StorageCluster spec with incorrect zone information. This directly impacts the stretch cluster settings in the CephCluster spec by adding invalid zone information, causing the Rook operator to fail.

Impact:
1.The storage cluster misidentifies node topology information, leading to incorrect zone configuration.
2.In the stretch cluster setup, this results in the wrong zone labels being applied, disrupting the intended fault domain configuration.
3.Ultimately, the Rook operator fails to reconcile the CephCluster, breaking the deployment process.

Proposed Solution: To address this issue, I introduced a new function, GetKeyValuesStrechCluster, which specifically checks for relevant and valid topology keys (corev1.LabelZoneFailureDomainStable and labelZoneFailureDomainWithoutBeta). This ensures only the intended labels are considered, preventing incorrect matches and maintaining the integrity of the stretch cluster configuration.
The function isolates the logic required for stretch cluster scenarios, leaving the existing GetKeyValues behavior unchanged for other use cases. This makes the fix targeted and minimizes the risk of regressions elsewhere.

Why This Fix Is Necessary:
The root cause of the issue lies in the overly broad match logic (strings.Contains), which needs to be refined for stretch cluster use cases.
Without this fix, incorrect labels (e.g., topology.kubernetes.io/zone-principal) will continue to disrupt the stretch cluster settings.
The new function ensures correctness by limiting the keys checked to those explicitly relevant for stretch cluster configurations.
Let me know if more clarification or details are needed!


@malayparida2000 @Nikhil-Ladha @iamniting @sp98 
Is the stretch cluster failure domain always set to "zone"?

api/v1/topologymap.go
OdedViner: The variable is defined in another package within the project, but it is a private variable (starts with a lowercase letter). Since it cannot be accessed outside its package.

malayparida2000:
> @malayparida2000 @Nikhil-Ladha @iamniting @sp98 Is the stretch cluster failure domain always set to "zone"?
I belive so https://docs.redhat.com/en/documentation/red_hat_openshift_data_foundation/4.13/html/configuring_openshift_data_foundation_disaster_recovery_for_openshift_workloads/introduction-to-stretch-cluster-disaster-recovery_stretch-cluster#applying-topology-zone-labels-to-ocp-nodes_stretch-cluster 
@travisn can you please correct me?

api/v1/topologymap.go
malayparida2000: I saw our docs & went as far back as [OCS 4.7.](https://red-hat-storage.github.io/ocs-training/training/ocs4/ocs4-metro-stretched.html) And I can confirm 2 things.
1. Stretch cluster always uses failure domain zone
2. For zone failure domain always `topology.kubernetes.io/zone` label is used. I don't see any need to search for any other labels such as `failure-domain.kubernetes.io/zone`.

travisn: Agreed with Malay on all these points. `failure-domain.kubernetes.io/zone` is an old deprecated label for topology that was removed from K8s many releases ago, we should not use it anymore.

api/v1/topologymap.go
travisn: Stretch always uses zones, just to confirm in this thread in addition to comments from others that confirmed the same.

api/v1/topologymap.go
OdedViner: done

api/v1/topologymap.go
OdedViner: done

malayparida2000:
@OdedViner This fix selectively addresses the case of stretch cluster only, while the mis match of labels can still happen to a normal cluster if the customer uses an improperly formatted label. I would suggest fixing the issue properly for all cases.
Ref-https://github.com/red-hat-storage/ocs-operator/pull/2965#issuecomment-2587063571
We should refactor the parts like how topologymap is built, failure domain is determined, keyvalues are fetched so that we finally fix this issue as a whole.

malayparida2000:
The fix looks good now, @OdedViner Can you please test this with a cluster & let us know the results here.

OdedViner:
> The fix looks good now, @OdedViner Can you please test this with a cluster & let us know the results here.

The private image seems to have resolved the issue

Procedure:
1.Deploy arbiter cluster:
ocs-ci path: 
deployment/vsphere/upi_1az_rhcos_vsan_lso_vmdk_3m_6w_arbiter.yaml [4.18.0-109.stable image]

2.Add "topology.kubernetes.io/zone-principal=true"  label to all worker nodes
```
oc label nodes compute-0 topology.kubernetes.io/zone-principal=true
oc label nodes compute-1 topology.kubernetes.io/zone-principal=true
oc label nodes compute-2 topology.kubernetes.io/zone-principal=true
oc label nodes compute-3 topology.kubernetes.io/zone-principal=true
oc label nodes compute-4 topology.kubernetes.io/zone-principal=true
oc label nodes compute-5 topology.kubernetes.io/zone-principal=true
```
3. Reset ocs-operaotr pod 

4.Check Storagecluster status:

    Message:                  Error while reconciling: CephCluster.ceph.rook.io "ocs-storagecluster-cephcluster" is invalid: spec.mon: Invalid value: "object": stretchCluster zones must be equal to 3
  
   Failure Domain Key:         topology.kubernetes.io/zone-principal
```
$ oc describe storagecluster
Status:
  Conditions:
    Last Heartbeat Time:      2025-01-21T11:54:01Z
    Last Transition Time:     2025-01-21T11:54:01Z
    Message:                  Version check successful
    Reason:                   VersionMatched
    Status:                   False
    Type:                     VersionMismatch
    Last Heartbeat Time:      2025-01-21T16:04:41Z
    Last Transition Time:     2025-01-21T15:29:45Z
    Message:                  Error while reconciling: CephCluster.ceph.rook.io "ocs-storagecluster-cephcluster" is invalid: spec.mon: Invalid value: "object": stretchCluster zones must be equal to 3
    Reason:                   ReconcileFailed
    Status:                   False
    Type:                     ReconcileComplete
    Last Heartbeat Time:      2025-01-21T15:29:06Z
    Last Transition Time:     2025-01-21T15:09:32Z
    Message:                  Reconcile completed successfully
    Reason:                   ReconcileCompleted
    Status:                   True
    Type:                     Available
    Last Heartbeat Time:      2025-01-21T15:29:06Z
    Last Transition Time:     2025-01-21T15:19:39Z
    Message:                  Reconcile completed successfully
    Reason:                   ReconcileCompleted
    Status:                   False
    Type:                     Progressing
    Last Heartbeat Time:      2025-01-21T15:29:06Z
    Last Transition Time:     2025-01-21T15:09:32Z
    Message:                  Reconcile completed successfully
    Reason:                   ReconcileCompleted
    Status:                   False
    Type:                     Degraded
    Last Heartbeat Time:      2025-01-21T15:29:06Z
    Last Transition Time:     2025-01-21T15:19:39Z
    Message:                  Reconcile completed successfully
    Reason:                   ReconcileCompleted
    Status:                   True
    Type:                     Upgradeable
  Current Mon Count:          5
  Default Ceph Device Class:  ssd
  Failure Domain:             zone
  Failure Domain Key:         topology.kubernetes.io/zone-principal
```
5.Create private image based relase-4.18 branch
```
$ export REGISTRY_NAMESPACE=oviner
$ export IMAGE_TAG=nitin-test
$ make ocs-operator
$ podman push quay.io/$REGISTRY_NAMESPACE/ocs-operator:$IMAGE_TAG
```
image location: 
quay.io/oviner/ocs-operator:nitin-test

7.Change csv image
```
$ oc edit csv ocs-operator.v4.18.0-109.stable

                image: quay.io/oviner/ocs-operator:nitin-test
                imagePullPolicy: Always
                name: ocs-operator
```
8. Manual remove some entries from the StorageCluster CR status.
```
oc patch storagecluster ocs-storagecluster -n openshift-storage --type=json --subresource=status --patch '[

{ "op": "remove", "path": "/status/failureDomain" }
]'
oc patch storagecluster ocs-storagecluster -n openshift-storage --type=json --subresource=status --patch '[

{ "op": "remove", "path": "/status/failureDomainValues"}
]'
oc patch storagecluster ocs-storagecluster -n openshift-storage --type=json --subresource=status --patch '[

{ "op": "remove", "path": "/status/failureDomainKey" }
]'
oc patch storagecluster ocs-storagecluster -n openshift-storage --type=json --subresource=status --patch '[

{ "op": "remove", "path": "/status/nodeTopologies" }
]'
```
9. Reset ocs-operaotr pod 
```
$ oc delete pod ocs-operator-74c9fd8477-kp88x 
pod "ocs-operator-74c9fd8477-kp88x" deleted
```
10.Check storagecluster status 
```
Current Mon Count:          5
  Default Ceph Device Class:  ssd
  Failure Domain:             zone
  Failure Domain Key:         topology.kubernetes.io/zone
  Failure Domain Values:
    data-1
    data-2
  Images:
    Ceph:
      Actual Image:   registry.redhat.io/rhceph/rhceph-8-rhel9@sha256:665278442f848fe2ac12be483b8778e672f35325daf6b90977ac12c3165bcabc
      Desired Image:  registry.redhat.io/rhceph/rhceph-8-rhel9@sha256:665278442f848fe2ac12be483b8778e672f35325daf6b90977ac12c3165bcabc
    Noobaa Core:
      Actual Image:   registry.redhat.io/odf4/mcg-core-rhel9@sha256:35f7f2c455823b04f77f664d0acf0a5759d6c6601e9ae0ad11bb00b97ed8cf1c
      Desired Image:  registry.redhat.io/odf4/mcg-core-rhel9@sha256:35f7f2c455823b04f77f664d0acf0a5759d6c6601e9ae0ad11bb00b97ed8cf1c
    Noobaa DB:
      Actual Image:   registry.redhat.io/rhel9/postgresql-15@sha256:44a08b83a6c50714b52f4cf1c3476bc16b66faec21dd9a9bc07d1be5f97b8150
      Desired Image:  registry.redhat.io/rhel9/postgresql-15@sha256:44a08b83a6c50714b52f4cf1c3476bc16b66faec21dd9a9bc07d1be5f97b8150
  Kms Server Connection:
  Node Topologies:
    Labels:
      kubernetes.io/hostname:
        compute-0
        compute-1
        compute-2
        compute-3
        compute-4
        compute-5
      topology.kubernetes.io/zone:
        data-1
        data-2
      topology.kubernetes.io/zone-principal:
        true
  Phase:  Ready
```
```
$ oc describe csv ocs-operator.v4.18.0-109.stable | grep -i oviner -C 3
                  Value From:
                    Field Ref:
                      Field Path:   metadata.namespace
                Image:              quay.io/oviner/ocs-operator:nitin-test
                Image Pull Policy:  Always
                Name:               ocs-operator
                Readiness Probe:
```

malayparida2000 (approved):
/lgtm

iamniting (approved):

PR #2966, opened 2025-01-15
add support for enabling clusterwide encryption as day-2 operation

sp98:
On existing ODF clusters, customers can enable `clusterwide` encryption by enabling `spec.Encryption.ClusterWide` to `true`.  Rook will destroy the OSDs are recreate them with same OSD ID after enabling encryption.


controllers/storagecluster/cephcluster.go
```diff
@@ -1399,3 +1407,17 @@ func determineDefaultCephDeviceClass(foundDeviceClasses []rookCephv1.DeviceClass
 	// if no device classes are found in status return empty string
 	return ""
 }
+
+// isEncrptionSettingUpdated checks whether ecryption was enabled or disabled for the storageClassDeviceSet.
+func isEncrptionSettingUpdated(newDeviceSet, existingDeviceSet []rookCephv1.StorageClassDeviceSet) bool {
+	if len(newDeviceSet) != len(existingDeviceSet) {
+		return false
```
iamniting, line 1414: I believe this is an unknown state instead of false, Is sending false safe here?

iamniting (changes requested):

controllers/storagecluster/cephcluster.go
sp98: This will happen in case of cluster expansion when the storagedeviceclass sets won't of the same size in the new and existing cluster.  Sending false should be safe here, as in the next reconcile the lengths would match.

iamniting: Will the list be in the status of the CR in the same order as spec?

sp98: Good question. I don't think its guaranteed and code should not reply on this that.  I've updated the changes. Looks like we don't need to compare both the new and old storageclassdeviceSets at all.  Just comparing the new `clusterWide` encryption setting with the existing StorageClassDeviceSet would be enough. Let me know what you think.

controllers/storagecluster/cephcluster.go
```diff
@@ -1399,3 +1407,14 @@ func determineDefaultCephDeviceClass(foundDeviceClasses []rookCephv1.DeviceClass
 	// if no device classes are found in status return empty string
 	return ""
 }
+
+// isEncrptionSettingUpdated checks whether ecryption was enabled or disabled by comparing the new clusterWide encryption setting
```
iamniting, line 1411:
```suggestion
// isEncrptionSettingUpdated checks whether encryption was enabled or disabled by comparing the new clusterWide encryption setting of the storagecluster
```

controllers/storagecluster/cephcluster.go
```diff
@@ -1399,3 +1407,14 @@ func determineDefaultCephDeviceClass(foundDeviceClasses []rookCephv1.DeviceClass
...
+// with the encryption setting on existing StorageClassDeviceSets.
```
iamniting, line 1412:
```suggestion
// with the encryption setting on existing StorageClassDeviceSets of the cephcluster.
```

iamniting (changes requested):

iamniting (approved):

PR #2970, opened 2025-01-20
Avoid reconciling rook-ceph-operator-config via the external controller

iamniting:

https://issues.redhat.com/browse/DFBUGS-1357

### DIFF
--- a/controllers/ocsinitialization/ocsinitialization_controller.go
+++ b/controllers/ocsinitialization/ocsinitialization_controller.go
@@ -445,6 +445,9 @@ func (r *OCSInitializationReconciler) ensureRookCephOperatorConfigExists(initial
 			return err // nolint:revive
 		}
 
+		// TODO: remove this in the next release, look at commit msg for more info
+		delete(rookCephOperatorConfig.Data, "ROOK_CSI_ENABLE_CEPHFS")
+
 		return nil
 	})
 
--- a/controllers/storagecluster/external_resources.go
+++ b/controllers/storagecluster/external_resources.go
@@ -36,11 +36,6 @@ const (
 	cephRgwTLSSecretKey                         = "ceph-rgw-tls-cert"
 )
 
-const (
-	rookCephOperatorConfigName = "rook-ceph-operator-config"
-	rookEnableCephFSCSIKey     = "ROOK_CSI_ENABLE_CEPHFS"
-)
-
 // store the name of the rados-namespace
 var radosNamespaceName string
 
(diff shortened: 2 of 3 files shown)


parth-gr (commented):
looks good,

@travisn @Madhu-1  by this PR external mode will always have cephfs csi-provisioners.
plus more details on commit message

malayparida2000 (approved):
/lgtm,
/hold in case we need more eyes on this before merge.

Madhu-1:
> looks good,
> 
> @travisn @Madhu-1 by this PR external mode will always have cephfs csi-provisioners. plus more details on commit message

wont this be a change in the behavior on the upgraded/new cluster where only RBD is in use not cephfs? Are we fine with it and also do we need any documentation change for it?

Madhu-1:
> > wont this be a change in the behavior on the upgraded/new cluster where only RBD is in use not cephfs? Are we fine with it and also do we need any documentation change for it?
> 
> We are thinking to update the ocs-operator config map with this config value as a next step but need a discussion on that. We would be deleting it from rook config with this PR and adding it to the ocs operator config in the same release, wdyt?

IMO as long as the behavior remains the same we are good.

travisn:
> > > wont this be a change in the behavior on the upgraded/new cluster where only RBD is in use not cephfs? Are we fine with it and also do we need any documentation change for it?
> > 
> > 
> > We are thinking to update the ocs-operator config map with this config value as a next step but need a discussion on that. We would be deleting it from rook config with this PR and adding it to the ocs operator config in the same release, wdyt?
> 
> IMO as long as the behavior remains the same we are good.

If the user desires to disable the CephFS CSI driver, now how do they do it? It seems this PR is forcing the cephfs csi driver to be enabled and the user is not allowed to override it? The configmap settings should be controlled by the user, but maybe I'm missing something.

parth-gr:
>If the user desires to disable the CephFS CSI driver, now how do they do it? It seems this PR is forcing the cephfs csi driver to be enabled and the user is not allowed to override it? The configmap settings should be controlled by the user, but maybe I'm missing something.

It correct with this PR we have force enabled the cephfs driver,
But now we are thinking of enabling disabling the cephfs driver from user input using the ocs operator config map and rook operator env variable...

controllers/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -445,6 +445,9 @@ func (r *OCSInitializationReconciler) ensureRookCephOperatorConfigExists(initial
 			return err // nolint:revive
 		}
 
+		// TODO: remove this in the next release, look at commit msg for more info
+		delete(rookCephOperatorConfig.Data, "ROOK_CSI_ENABLE_CEPHFS")
```
parth-gr, line 449: wont the fail if the key is not present @iamniting

iamniting: no it wont.

PR #2979, opened 2025-01-28
Allow specifying poolSpec for default block, file, object pools from storageCluster

malayparida2000:
Ref-https://issues.redhat.com/browse/RHSTOR-5690

With this change the poolSpec for the default cephBlockPool, default cephFS data pool and default cephObjectStore data pool can be configured via the storageCluster CR. The pool spec is directly passed to the ceph CRs & only necessary fields are checked if set or not. Rest of the fields are not touched. Although not all the fields are configurable, but most fields are.

This will allow users to configure the target_size_ratio for these pools via the storageCluster CR.


controllers/storagecluster/generate.go
```diff
@@ -136,14 +136,32 @@ func generateNameForCephRbdMirror(initData *ocsv1.StorageCluster) string {
...
-func generateCephReplicatedSpec(initData *ocsv1.StorageCluster, poolType string) cephv1.ReplicatedSpec {
+func generateCephReplicatedSpec(initData *ocsv1.StorageCluster, poolType string, storageType string) cephv1.ReplicatedSpec {
 	crs := cephv1.ReplicatedSpec{}
 
 	crs.Size = getCephPoolReplicatedSize(initData)
 	crs.ReplicasPerFailureDomain = uint(getReplicasPerFailureDomain(initData))
-	//lint:ignore ST1017 required to compare it directly
-	if "data" == poolType {
-		crs.TargetSizeRatio = .49
+	if poolType == "data" {
```
iamniting, line 144: Can you pls create const variables for `block`, `file` etc and use them at all places?

controllers/storagecluster/generate.go
```diff
@@ -136,14 +136,32 @@ func generateNameForCephRbdMirror(initData *ocsv1.StorageCluster) string {
...
+		defaultTargetSizeRatio := 0.49
+		var definedSize float64
+
+		switch storageType {
+		case "block":
```
iamniting, line 149: dont we need a case for the nfs and the API changes as well?

iamniting (changes requested):

controllers/storagecluster/generate.go
malayparida2000: Acc to the design doc discussion we don't need to expose for nfs block pool or replica-1 block pool.

malayparida2000: https://docs.google.com/document/d/1BOag-Hm9vpmYnpW76MstJXBdDCyAlYQf0rexpdKZnCA/edit?disco=AAABUdNdABs
But just to confirm again, @travisn do we need to expose a field for the nfs block pool? or may be use the same value in the nfs blokcpool as the default block pool? or the nfs block pool should stay at 0.49?

controllers/storagecluster/generate.go
malayparida2000: Changed

controllers/storagecluster/cephblockpools.go
```diff
@@ -199,7 +199,7 @@ func (o *ocsCephBlockPools) reconcileNFSCephBlockPool(r *StorageClusterReconcile
 		cephBlockPool.Spec.PoolSpec.DeviceClass = storageCluster.Status.DefaultCephDeviceClass
 		cephBlockPool.Spec.EnableCrushUpdates = true
 		cephBlockPool.Spec.PoolSpec.FailureDomain = getFailureDomain(storageCluster)
-		cephBlockPool.Spec.PoolSpec.Replicated = generateCephReplicatedSpec(storageCluster, "data")
+		cephBlockPool.Spec.PoolSpec.Replicated = generateCephReplicatedSpec(storageCluster, poolTypeData, storageTypeBlock)
```
iamniting, line 202: dont we need a separate one here for nfs at it was earlier?

malayparida2000: Have asked Travis on what we need to do for nfs here
https://github.com/red-hat-storage/ocs-operator/pull/2979#discussion_r1935423042

controllers/storagecluster/generate.go
travisn: For NFS, are you referring to the `.nfs` CephBlockPool? That is just a metadata pool, so it doesn't need the targetSizeRatio to be set. NFS uses a data pool from CephFS.

controllers/storagecluster/generate.go
```diff
@@ -136,19 +136,40 @@ func generateNameForCephRbdMirror(initData *ocsv1.StorageCluster) string {
...
 	return crs
 }
 
+func getTargetSizeRatio(initData *ocsv1.StorageCluster, storageType string) float64 {
+	defaultTargetSizeRatio := 0.49
+	var definedRatio float64
+
+	switch storageType {
+	case storageTypeBlock:
+		definedRatio = initData.Spec.ManagedResources.CephBlockPools.PoolSpec.Replicated.TargetSizeRatio
```
travisn, line 157: Actually, the PoolSpec.TargetSizeRatio was deprecated in rook a long time ago, but we have kept it for backward compatibility. Instead, let's use the parameters:
```suggestion
		definedRatio = initData.Spec.ManagedResources.CephBlockPools.PoolSpec.Parameters["target_size_ratio"]
```

travisn (changes requested):

controllers/storagecluster/generate.go
malayparida2000: Ack, @travisn Even though the field was deprecated there doesn't seem to any comment/kubebuilder warning in rook suggesting that. I think that can be added in rook APIs.

malayparida2000: @travisn Another question if both Replicated.TargetSizeRatio & Parameters["target_size_ratio"] are specified then which one has higher priority?

malayparida2000: I see the Replicated.TargetSizeRatio is given precedence ahead of the parameter.
https://github.com/rook/rook/blob/3fcc7682f2c3b02f6c9fed2184665ac6f0ce945b/pkg/daemon/ceph/client/pool.go#L296
And I also see the Replicated.TargetSizeRatio field being used here to determine if TragetSizeRatio is defined or not here https://github.com/rook/rook/blob/master/pkg/apis/ceph.rook.io/v1/pool.go#L39

travisn: Yes, it's true the targetSizeRatio would have higher precedence, it was just implemented that way to easily support changing the implementation to use the parameters, and maintain backward compatibility. But you're right, we are long overdue to mark it as deprecated, it has been like this for many releases.

malayparida2000: With current rook behaviour we can't shift to parameters right now as it is, As if we move to the parameters the values left on the replicated.TargetSizeRatio would still have precedence.

travisn: Oh I see now the point. Actually, just go ahead and keep using the replicated.TargetSizeRatio. We can keep Rook behavior the same and not bother to deprecate the setting, and it will keep working smoothly for downstream with your current approach.

travisn (approved):
/lgtm

malayparida2000:
1. Allowed pass-through of poolSpec so that in the future new field support can be added seamlessly
2. In stretch cluster Size & replicas per failure domain can't be changed.
3. Changed '.nfs' cephblockpool to be a metadata Pool ref- https://github.com/red-hat-storage/ocs-operator/pull/2979#discussion_r1936368884
@travisn PTAL

api/v1/storagecluster_types.go
```diff
@@ -294,6 +296,8 @@ type ManageCephObjectStores struct {
 	// +kubebuilder:validation:MaxLength=253
 	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$
 	StorageClassName string `json:"storageClassName,omitempty"`
+	// DataPoolSpec specifies the pool specification for the default cephObjectStore data pool
+	DataPoolSpec rookCephv1.PoolSpec `json:"dataPoolSpec,omitempty"`
```
travisn, line 300: How about if we make these nullable? Then it's simple to know if the user specified a spec or not, and we can write to the log to make it more obvious for troubleshooting, in case the user applied unsupported settings.
```suggestion
	DataPoolSpec *rookCephv1.PoolSpec `json:"dataPoolSpec,omitempty"`
```

controllers/storagecluster/cephblockpools.go
```diff
@@ -88,10 +88,30 @@ func (o *ocsCephBlockPools) reconcileCephBlockPool(r *StorageClusterReconciler,
...
-		cephBlockPool.Spec.PoolSpec.Replicated = generateCephReplicatedSpec(storageCluster, "data")
+		// Pass the poolSpec from the storageCluster CR
+		// Set default values in the poolSpec if necessary
+		cephBlockPool.Spec.PoolSpec = storageCluster.Spec.ManagedResources.CephBlockPools.PoolSpec
+
+		poolSpec := &cephBlockPool.Spec.PoolSpec
+		if poolSpec.DeviceClass == "" {
+			poolSpec.DeviceClass = storageCluster.Status.DefaultCephDeviceClass
+		}
+		poolSpec.EnableCrushUpdates = true
```
travisn, line 99: Any reason to disallow them from overriding this setting? I can't think of a user scenario, but perhaps support could run into some condition where they want to override it.

malayparida2000: Couldn't do with these booleans as there is no way to tell if a boolean field is either just unset or explicitly set to false. If this was a bool ptr we could have done the check.

travisn: oh right, nevermind, it doesn't help even if the parent spec is nullable

api/v1/storagecluster_types.go
malayparida2000: resolving this as oer the discussion on the other thread

controllers/storagecluster/cephblockpools.go
iamniting, line 99: Can we move fields we are not checking before setting the value before all if statements?

controllers/storagecluster/cephobjectstores.go
```diff
@@ -209,6 +204,28 @@ func (r *StorageClusterReconciler) newCephObjectStoreInstances(initData *ocsv1.S
 			obj.Spec.Gateway.HostNetwork = initData.Spec.ManagedResources.CephObjectStores.HostNetwork
 		}
 
+		// Set default values in the data pool spec if necessary
```
iamniting, line 207: It appears same as above, Can we create a func and use it.

iamniting (changes requested):

iamniting (approved):

PR #2980, opened 2025-01-28
Set bulk flag to true on intended pools for performance gain

malayparida2000:
The bulk flag makes the autoscaler start with the max number of PGs, and the autoscaler then decreases the pg count only if the PG usage starts to skew too much. This could improve performance for ODF users due to great amount of parallelism during read/write on large clusters.

We are going to set the bulk flag for now on the default cephBlockPool & replica-1 cephBlockpools, default cephFS metadata pool & data pools, default cephObjectStore metadata pool & data pool.

The flag is set to true only during new pool creation, as setting it on existing pools can cause data movement.

Ref-https://issues.redhat.com/browse/RHSTOR-6774

### DIFF
--- a/controllers/storagecluster/cephblockpools.go
+++ b/controllers/storagecluster/cephblockpools.go
@@ -2,6 +2,7 @@ package storagecluster
 
 import (
 	"fmt"
+	"strconv"
 
 	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
 	"github.com/red-hat-storage/ocs-operator/v4/controllers/util"
@@ -90,11 +91,20 @@ func (o *ocsCephBlockPools) reconcileCephBlockPool(r *StorageClusterReconciler, 
 	_, err = ctrl.CreateOrUpdate(r.ctx, r.Client, cephBlockPool, func() error {
 		// Pass the poolSpec from the storageCluster CR
 
+		existingPoolSpec := cephBlockPool.Spec.PoolSpec
 		cephBlockPool.Spec.PoolSpec = storageCluster.Spec.ManagedResources.CephBlockPools.PoolSpec
 
 		// Set default values in the poolSpec as necessary
 		setDefaultDataPoolSpec(&cephBlockPool.Spec.PoolSpec, storageCluster)
 		cephBlockPool.Spec.PoolSpec.EnableRBDStats = true
+
+		// The bulk flag is set to true only during new pool creation, as setting it on existing pools can cause data movement.
+		if cephBlockPool.CreationTimestamp.IsZero() {
+			setBulkFlagParameter(&cephBlockPool.Spec.PoolSpec.Parameters)
+		}
+
+		// Ensures the bulk flag set during new pool creation is not removed during updates.
+		preserveBulkFlagParameter(existingPoolSpec.Parameters, &cephBlockPool.Spec.PoolSpec.Parameters)
 
 		// Since provider mode handles mirroring, we only need to handle for internal mode
 		if storageCluster.Annotations["ocs.openshift.io/deployment-mode"] != "provider" {
@@ -246,16 +256,33 @@ func (o *ocsCephBlockPools) reconcileNonResilientCephBlockPool(r *StorageCluster
 			poolSpec.DeviceClass = failureDomainValue
 			poolSpec.EnableCrushUpdates = true
 			poolSpec.FailureDomain = getFailureDomain(storageCluster)
+			existingParameters := poolSpec.Parameters
 			poolSpec.Parameters = storageCluster.Spec.ManagedResources.CephNonResilientPools.Parameters
 			if poolSpec.Parameters == nil {
 				poolSpec.Parameters = make(map[string]string)
 			}
-			if _, ok := poolSpec.Parameters["pg_num"]; !ok {
-				poolSpec.Parameters["pg_num"] = "16"
+
+			// The bulk flag is set to true only during new pool creation, as setting it on existing pools can cause data movement.
+			if cephBlockPool.CreationTimestamp.IsZero() {
+				setBulkFlagParameter(&poolSpec.Parameters)
 			}
-			if _, ok := poolSpec.Parameters["pgp_num"]; !ok {
-				poolSpec.Parameters["pgp_num"] = "16"
+
+			// Ensures the bulk flag set during new pool creation is not removed during updates.
+			preserveBulkFlagParameter(existingParameters, &poolSpec.Parameters)
+
+			// set the pg_num & pgp_num parameters only when bulk flag is not set or false
+			if bulk, ok := poolSpec.Parameters["bulk"]; !ok {
+				b, err := strconv.ParseBool(bulk)
+				if err != nil || !b {
+					if _, ok := poolSpec.Parameters["pg_num"]; !ok {
+						poolSpec.Parameters["pg_num"] = "16"
+					}
+					if _, ok := poolSpec.Parameters["pgp_num"]; !ok {
+						poolSpec.Parameters["pgp_num"] = "16"
+					}
+				}
 			}
+
 			poolSpec.Replicated = cephv1.ReplicatedSpec{
 				Size:                   1,
 				RequireSafeReplicaSize: false,
--- a/controllers/storagecluster/cephobjectstores.go
+++ b/controllers/storagecluster/cephobjectstores.go
@@ -140,6 +140,11 @@ func (r *StorageClusterReconciler) createCephObjectStores(cephObjectStores []*ce
 			r.Log.Info("Restoring original CephObjectStore.", "CephObjectStore", klog.KRef(cephObjectStore.Namespace, cephObjectStore.Name))
 			existing.ObjectMeta.OwnerReferences = cephObjectStore.ObjectMeta.OwnerReferences
 			cephObjectStore.ObjectMeta = existing.ObjectMeta
+
+			// Ensures the bulk flag set during new pool creation is not removed during updates.
+			preserveBulkFlagParameter(existing.Spec.MetadataPool.Parameters, &cephObjectStore.Spec.MetadataPool.Parameters)
+			preserveBulkFlagParameter(existing.Spec.DataPool.Parameters, &cephObjectStore.Spec.DataPool.Parameters)
+
 			err = r.Client.Update(context.TODO(), cephObjectStore)
 			if err != nil {
 				r.Log.Error(err, "Failed to update CephObjectStore.", "CephObjectStore", klog.KRef(cephObjectStore.Namespace, cephObjectStore.Name))
@@ -147,6 +152,11 @@ func (r *StorageClusterReconciler) createCephObjectStores(cephObjectStores []*ce
 			}
 		case errors.IsNotFound(err):
 			r.Log.Info("Creating CephObjectStore.", "CephObjectStore", klog.KRef(cephObjectStore.Namespace, cephObjectStore.Name))
+
+			// The bulk flag is set to true only during new pool creation, as setting it on existing pools can cause data movement.
+			setBulkFlagParameter(&cephObjectStore.Spec.MetadataPool.Parameters)
+			setBulkFlagParameter(&cephObjectStore.Spec.DataPool.Parameters)
+
 			err = r.Client.Create(context.TODO(), cephObjectStore)
 			if err != nil {
 				r.Log.Error(err, "Failed to create CephObjectStore.", "CephObjectStore", klog.KRef(cephObjectStore.Namespace, cephObjectStore.Name))
--- a/controllers/storagecluster/cephobjectstores_test.go
+++ b/controllers/storagecluster/cephobjectstores_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"testing"
 
+	"github.com/imdario/mergo"
 	configv1 "github.com/openshift/api/config/v1"
 	api "github.com/red-hat-storage/ocs-operator/api/v4/v1"
 	"github.com/red-hat-storage/ocs-operator/v4/controllers/defaults"
@@ -11,6 +12,8 @@ import (
 	cephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
 	"github.com/stretchr/testify/assert"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
@@ -47,6 +50,13 @@ func TestCephObjectStores(t *testing.T) {
 
 func assertCephObjectStores(t *testing.T, reconciler StorageClusterReconciler, cr *api.StorageCluster, request reconcile.Request) {
 	expectedCos, err := reconciler.newCephObjectStoreInstances(cr, nil)
+	expectedCos[0].Spec.MetadataPool.Parameters = map[string]string{
+		"bulk": "true",
+	}
+	expectedCos[0].Spec.DataPool.Parameters = map[string]string{
+		"bulk": "true",
+	}
+
 	assert.NoError(t, err)
 
 	actualCos := &cephv1.CephObjectStore{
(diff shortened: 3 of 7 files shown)


controllers/storagecluster/cephobjectstores.go
```diff
@@ -173,6 +173,7 @@ func (r *StorageClusterReconciler) newCephObjectStoreInstances(initData *ocsv1.S
 					EnableCrushUpdates: true,
 					FailureDomain:      initData.Status.FailureDomain,
 					Replicated:         generateCephReplicatedSpec(initData, "data"),
+					Parameters:         map[string]string{"bulk": "true"},
 				},
 				MetadataPool: cephv1.PoolSpec{
```
travisn, line 178: @BlaineEXE Should we be setting the bulk flag on the object and filesystem metadata pools, in addition to the data pools?

BlaineEXE: Anthony and Kyle seemed to be saying yes when we chatted. It sounds a bit counterintuitive to me as well. My best guess is that the performance gains from increased parallelism must end up being a bigger positive effect than the performance loss from having info split into more chunks.

BlaineEXE (commented):
Are there plans to allow this to be configured via any means? I trust Kyle and Anthony implicitly, but we still don't have a measure of how the `bulk` flag impacts ODF specifically. Further, we don't know whether ODF's common 3-OSD/4-OSD clusters will be impacted positively or negatively by this change. My vague understanding is that 3-/4-OSD clusters are much less common for RHCS than for ODF, and my intuition says that cluster size may play an important part in the performance impacts.

I would recommend having some mechanism for enabling/disabling this behavior. We don't have to document it for users, but I think it will still be useful to us internally. That configurability will make it easier for the performance team to do A:B testing to quantify the performance changes, and it will allow us an easy way to disable this change close to ODF release if we have to by changing the "business logic" instead of the code. It'll also give us the ability to disable/enable it for customer environments in cases where that is imporant to them.

Travis is also asking an important question [here](https://github.com/red-hat-storage/ocs-operator/pull/2980/files#r1932623529). His and my intuition both suggest that the RGW metadata pools might respond to the bulk flag differently than other pools. Thus, I would also suggest allowing the CephObjectStore metadata pools have bulk enabled/disabled independently from the rest of the cluster. This is essentially for the same reasons above.

Here's my guess about what we will find with `bulk` flag performance testing: 
- I think we will find that the bulk flag has no impact on 3-OSD clusters on the whole
- I suspect there will be noticeable positive impact on clusters with 9-12 or more OSDs
- There is a risk we find S3 metadata operation slowdown with 3-OSD clusters, possibly resulting in minor S3 I/O loss until the cluster has 6-9 OSDs, where we will break even and then see gradual improvements

controllers/storagecluster/cephblockpools.go
```diff
@@ -246,16 +256,33 @@ func (o *ocsCephBlockPools) reconcileNonResilientCephBlockPool(r *StorageCluster
...
+			if cephBlockPool.CreationTimestamp.IsZero() {
+				setBulkFlagParameter(&poolSpec.Parameters)
 			}
-			if _, ok := poolSpec.Parameters["pgp_num"]; !ok {
-				poolSpec.Parameters["pgp_num"] = "16"
+
+			// Ensures the bulk flag set during new pool creation is not removed during updates.
+			preserveBulkFlagParameter(existingParameters, &poolSpec.Parameters)
+
+			// set the pg_num & pgp_num parameters only when bulk flag is not set or false
```
malayparida2000, line 273: @travisn whats the behaviour of ceph when bulk is set to true as well as pg_num, pgp_num are set to 16? Do we have to selectively exclude setting these 2 parameters when bulk flag is true?

malayparida2000:
some basic testing on 4.19 fresh cluster.

Bulk flag presence on pools
```
~ $ oc get cephblockpool ocs-storagecluster-cephblockpool -o yaml | yq '.spec.parameters'

bulk: "true"

~ $ oc get cephfilesystem ocs-storagecluster-cephfilesystem -o yaml | yq '.spec.dataPools[].parameters'

bulk: "true"

~ $ oc get cephfilesystem ocs-storagecluster-cephfilesystem -o yaml | yq '.spec.metadataPool.parameters'

bulk: "true"

sh-5.1$ ceph osd pool autoscale-status
POOL                                          SIZE  TARGET SIZE  RATE  RAW CAPACITY   RATIO  TARGET RATIO  EFFECTIVE RATIO  BIAS  PG_NUM  NEW PG_NUM  AUTOSCALE  BULK   
.mgr                                        576.5k                3.0         6144G  0.0000                                  1.0       1              on         False  
ocs-storagecluster-cephblockpool            66071k                3.0         6144G  0.3333        0.4900           0.3333   1.0      32              on         True   
ocs-storagecluster-cephfilesystem-metadata  25586                 3.0         6144G  0.3333        0.4900           0.3333   4.0     128              on         True   
ocs-storagecluster-cephfilesystem-data0         0                 3.0         6144G  0.3333        0.4900           0.3333   1.0      32              on         True   
```

malayparida2000:
Wrote around 110GB(10*100MB, 10*1GB, 10*10GB) of data each on rbd & cephFS

```
sh-5.1$ ceph osd pool autoscale-status
POOL                                          SIZE  TARGET SIZE  RATE  RAW CAPACITY   RATIO  TARGET RATIO  EFFECTIVE RATIO  BIAS  PG_NUM  NEW PG_NUM  AUTOSCALE  BULK   
.mgr                                        576.5k                3.0         6144G  0.0000                                  1.0       1              on         False  
ocs-storagecluster-cephblockpool            108.5G                3.0         6144G  0.3333        0.4900           0.3333   1.0      32              on         True   
ocs-storagecluster-cephfilesystem-metadata  162.7M                3.0         6144G  0.3333        0.4900           0.3333   4.0     128              on         True   
ocs-storagecluster-cephfilesystem-data0     108.3G                3.0         6144G  0.3333        0.4900           0.3333   1.0      32              on         True 
```

malayparida2000:
> Wrote around 110GB(10_100MB, 10_1GB, 10*10GB) of data each on rbd & cephFS
> 
> ```
> sh-5.1$ ceph osd pool autoscale-status
> POOL                                          SIZE  TARGET SIZE  RATE  RAW CAPACITY   RATIO  TARGET RATIO  EFFECTIVE RATIO  BIAS  PG_NUM  NEW PG_NUM  AUTOSCALE  BULK   
> .mgr                                        576.5k                3.0         6144G  0.0000                                  1.0       1              on         False  
> ocs-storagecluster-cephblockpool            108.5G                3.0         6144G  0.3333        0.4900           0.3333   1.0      32              on         True   
> ocs-storagecluster-cephfilesystem-metadata  162.7M                3.0         6144G  0.3333        0.4900           0.3333   4.0     128              on         True   
> ocs-storagecluster-cephfilesystem-data0     108.3G                3.0         6144G  0.3333        0.4900           0.3333   1.0      32              on         True 
> ```

@travisn Why do we have a BIAS of 4 for the cephFS metadata pool? As a result it ends up having 4X amount of PGs.

travisn:
> @travisn Why do we have a BIAS of 4 for the cephFS metadata pool? As a result it ends up having 4X amount of PGs.

This must be a default that is set for the cephfs metadata pool. @batrick Is this expected? I didn't realize either that the cephfs metadata pool should have so many PGs.

controllers/storagecluster/cephblockpools.go
```diff
@@ -90,12 +91,21 @@ func (o *ocsCephBlockPools) reconcileCephBlockPool(r *StorageClusterReconciler,
...
 		// Pass the poolSpec from the storageCluster CR
 
+		existingPoolSpec := cephBlockPool.Spec.PoolSpec
 		cephBlockPool.Spec.PoolSpec = storageCluster.Spec.ManagedResources.CephBlockPools.PoolSpec
 
 		// Set default values in the poolSpec as necessary
 		setDefaultDataPoolSpec(&cephBlockPool.Spec.PoolSpec, storageCluster)
 		cephBlockPool.Spec.PoolSpec.EnableRBDStats = true
 
+		// The bulk flag is set to true only during new pool creation, as setting it on existing pools can cause data movement.
```
BlaineEXE, line 101: IMO, it would be helpful to add a note here that explains that this condition should only be possible when the CephBlockPool doesn't already exist, meaning this struct is initialized only in code.

controllers/storagecluster/cephobjectstores_test.go
```diff
@@ -47,6 +47,13 @@ func TestCephObjectStores(t *testing.T) {
...
 func assertCephObjectStores(t *testing.T, reconciler StorageClusterReconciler, cr *api.StorageCluster, request reconcile.Request) {
 	expectedCos, err := reconciler.newCephObjectStoreInstances(cr, nil)
+	expectedCos[0].Spec.MetadataPool.Parameters = map[string]string{
+		"bulk": "true",
+	}
+	expectedCos[0].Spec.DataPool.Parameters = map[string]string{
+		"bulk": "true",
+	}
+
 	assert.NoError(t, err)
```
BlaineEXE, line 57: Is there a test that checks that `bulk` isn't applied to already-existing stores? That is also important behavior.

BlaineEXE (commented):
I only see a single unit test check for this, regarding CephObjectStores. IMO, it would be best to check that `bulk` is set on new objects and **not** set on existing objects for all resource kinds.

BlaineEXE:
Any response to this? https://github.com/red-hat-storage/ocs-operator/pull/2980#pullrequestreview-2582566300

malayparida2000:
> I would recommend having some mechanism for enabling/disabling this behavior.

Now all the metadata/data pool's poolSpec are configurable via the storageCluster CR so the bulk can be added/removed from the poolSpec parameters on demand by the customers. This can achive the A/B testing perf team might want to do.

> Travis is also asking an important question [here](https://github.com/red-hat-storage/ocs-operator/pull/2980/files#r1932623529).
The RGW & CephFS metadata pools would have the bulk flag by default with this change.

controllers/storagecluster/cephobjectstores_test.go
malayparida2000: Will add soon

malayparida2000: Added detailed unit tests

controllers/storagecluster/cephobjectstores.go
mmgaggle: The objects are striped based on rbd order or rgw / cephfs stripe size, so having more PGs doesn't translate to smaller rados objects, it just means we have more targets (pgs) to distribute objects across, and more targets (pgs) to do work.

controllers/storagecluster/cephblockpools.go
malayparida2000: Did some testing on a replica-1 cluster & found that when bulk flag is true the settings related to pg_num, pgp_num are ignored.

 ```
~ $  oc get cephblockpool ocs-storagecluster-cephblockpool-us-east-1a -o yaml | yq '.spec.parameters'
bulk: "true"
pg_num: "16"
pgp_num: "16"
```

```
sh-5.1$ ceph osd pool autoscale-status
POOL                                           SIZE  TARGET SIZE  RATE  RAW CAPACITY   RATIO  TARGET RATIO  EFFECTIVE RATIO  BIAS  PG_NUM  NEW PG_NUM  AUTOSCALE  BULK   
.mgr                                         576.5k                3.0         6144G  0.0000                                  1.0       1              on         False  
ocs-storagecluster-cephblockpool             65219k                3.0         6144G  0.3333        0.4900           0.3333   1.0      32              on         True   
ocs-storagecluster-cephfilesystem-metadata   39776                 3.0         6144G  0.3333        0.4900           0.3333   4.0     128              on         True   
ocs-storagecluster-cephfilesystem-data0          0                 3.0         6144G  0.3333        0.4900           0.3333   1.0      32              on         True   
ocs-storagecluster-cephblockpool-us-east-1a     19                 1.0         2048G  0.0000                                  1.0     128              on         True   
ocs-storagecluster-cephblockpool-us-east-1b     19                 1.0         2048G  0.0000                                  1.0     128              on         True   
ocs-storagecluster-cephblockpool-us-east-1c     19                 1.0         2048G  0.0000                                  1.0     128              on         True   

```

controllers/storagecluster/cephobjectstores.go
malayparida2000: Thanks @mmgaggle for such a detailed analysis.

malayparida2000:
@travisn @BlaineEXE Can you please review the changes
Also With Kyle's suggestions here https://github.com/red-hat-storage/ocs-operator/pull/2980#discussion_r1983487751
Should we also look at the `mon_target_pg_per_osd` parameter.

BlaineEXE:
> @travisn @BlaineEXE Can you please review the changes Also With Kyle's suggestions here [#2980 (comment)](https://github.com/red-hat-storage/ocs-operator/pull/2980#discussion_r1983487751) Should we also look at the `mon_target_pg_per_osd` parameter.

Anthony suggested setting `mon_max_pg_per_osd` which ocs-operator already does, but Kyle's response was the first I heard of `mon_target_pg_per_osd`. I'd love to see Kyle and Anthony weigh in here on how important this is for bulk/autoscale behavior. 

https://github.com/red-hat-storage/ocs-operator/blob/7a7f0c532755bc7f6bbaf8fbfaa6bab452340195/controllers/storagecluster/cephconfig.go#L37

I also think this value may risk data movement in existing clusters if we are overwriting it in the `rook-config-override` configmap. Does ocs-operator only set that configmap once during first instantiation, or does it update it every ODF release?

I think it makes sense to me to not set `pg_num`/`pgp_num` to 16 when we are using the `bulk` flag. I think setting them to 16 is what Kyle is referring to as "paying in advance", and I'm assuming we did this to prevent these from getting super low numbers with the non-bulk autoscale behavior. But I don't feel certain my interpretation is correct.

controllers/storagecluster/cephblockpools.go
mmgaggle: I think my preference would be to employ a strategy like the one outlined here that uses `target_ratio` on each pool instead of defining the number of PGs. I don't think `bulk` will do anything without either `target_ratio` or `target-size`, and if you supplied one of those with a static number of PGs the autoscaler would adjust the quantity.

https://github.com/red-hat-storage/ocs-operator/pull/2980#discussion_r1983487751

malayparida2000:
> Does ocs-operator only set that configmap once during first instantiation, or does it update it every ODF release?

Yes currently the behaviour is ocs operator reconciles the configmap & keeps it updated.

> I think it is safe to say that ODF should update the bootstrapped rook-config-override to use mon_target_pg_per_osd = 200 as long as we can do so only for new ODF clusters -- we don't want existing clusters to be effected.

Right we are already setting the `mon_max_pg_per_osd` to 600, so for that we don't need to do anything more. I will work on a separate PR to set `mon_target_pg_per_osd = 200`. 

> your change to make it so pg_num/pgp_num are not set to 16 when we are using the bulk flag makes sense

That change is actually only for replica-1 ceph block pools, as earlier these pools had issues with auto scaling so we used to set the PG Nums to 16. Now with bulk flag they won't be necessary anymore.

controllers/storagecluster/cephblockpools.go
malayparida2000: Right @mmgaggle. This specifying the PG Num as 16 was only for specific kind of replica-1 ceph block pools where we had issues with auto scaling.
For all the data pools we create we specify the target size ratio as 0.49

travisn (approved):
Looks good to set the bulk flag on all the new pools, and avoid it for the existing pools.

mmgaggle:
If the number of pgs an OSD is participating is > mon_max_pg_per_osd then there will be a warning. This is distinct from mon_target_pg_per_osd which informs the number of PGs the PG autoscaler should aim to have per OSD.

malayparida2000:
> If the number of pgs an OSD is participating is > `mon_max_pg_per_osd` then there will be a warning. This is distinct from `mon_target_pg_per_osd` which informs the number of PGs the PG autoscaler should aim to have per OSD.

@mmgaggle We havebeen setting `mon_max_pg_per_osd` to 600 since a long time now. https://github.com/red-hat-storage/ocs-operator/blob/7a7f0c532755bc7f6bbaf8fbfaa6bab452340195/controllers/storagecluster/cephconfig.go#L37.
Do we need to change that value for 'bulk flag now?

malayparida2000:
/unhold
Queries regarding `mon_max_pg_per_osd` & `mon_target_pg_per_osd` can be continued to discussed after the PR is merged.

malayparida2000:
Opened a new PR for the mon_target_pg_per_osd value https://github.com/red-hat-storage/ocs-operator/pull/3080

PR #2989, opened 2025-01-29
provider: add missing fields for vgsc

rewantsoni:
The PR does the following:

1. Add the missing required fields for VolumeGroupSnapshotClass 
2. Update the name to align with internal mode.

### DIFF
--- a/services/provider/server/server.go
+++ b/services/provider/server/server.go
@@ -3,6 +3,7 @@ package server
 import (
 	"context"
 	"crypto"
+	"crypto/md5"
 	"crypto/rsa"
 	"crypto/sha256"
 	"crypto/x509"
@@ -734,7 +735,8 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
 		}
 	}
 
-	clientProfile := util.CalculateMD5Hash(req.StorageClaimName)
+	clientProfilemd5Sum := md5.Sum([]byte(req.StorageClaimName))
+	clientProfile := hex.EncodeToString(clientProfilemd5Sum[:])
 
 	var extR []*pb.ExternalResource
 
@@ -817,13 +819,15 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
 					}),
 				},
 				&pb.ExternalResource{
-					Name: "ceph-rbd",
+					Name: fmt.Sprintf("%s-groupsnapclass", req.StorageClaimName),
 					Kind: "VolumeGroupSnapshotClass",
 					Labels: map[string]string{
 						ramenDRStorageIDLabelKey: storageID,
 					},
 					Data: mustMarshal(map[string]string{
 						"csi.storage.k8s.io/group-snapshotter-secret-name": provisionerSecretName,
+						"clusterID": clientProfile,
+						"pool":      rns.Spec.BlockPoolName,
 					}),
 				},
 				&pb.ExternalResource{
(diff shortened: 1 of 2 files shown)


services/provider/server/server.go
```diff
@@ -817,13 +817,15 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
 					}),
 				},
 				&pb.ExternalResource{
-					Name: "ceph-rbd",
+					Name: fmt.Sprintf("%s-groupsnapclass", req.StorageClaimName),
```
Madhu-1, line 820: @rewantsoni what is the reason for name change? can we keep it similar to what we have today for SC/VSClass etc?

services/provider/server/server.go
```diff
@@ -937,13 +939,15 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
...
-					Name: "cephfs",
+					Name: fmt.Sprintf("%s-groupsnapclass", req.StorageClaimName),
 					Kind: "VolumeGroupSnapshotClass",
 					Labels: map[string]string{
 						ramenDRStorageIDLabelKey: storageID,
 					},
 					Data: mustMarshal(map[string]string{
 						"csi.storage.k8s.io/group-snapshotter-secret-name": provisionerSecretName,
+						"fsName":    subVolumeGroup.Spec.FilesystemName,
+						"clusterID": clientProfile,
```
Madhu-1, line 950: do we set clusterID as clientProfile for other resouces as well?

services/provider/server/server.go
rewantsoni: I am sending the exact name for VGSC here so that on client side we can create it. This name is inline with what we have in internal mode

services/provider/server/server.go
rewantsoni: We have started setting it for VolumeReplicationClass with RDR since with convergence the clientProfile name will be sent from the provider.

services/provider/server/server.go
rewantsoni: We have done something similar for VolumeReplicationClass where we are deciding the actual name on the provider side and sending it

services/provider/server/server.go
```diff
@@ -734,7 +735,8 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
 		}
 	}
 
-	clientProfile := util.CalculateMD5Hash(req.StorageClaimName)
+	clientProfilemd5Sum := md5.Sum([]byte(req.StorageClaimName))
+	clientProfile := hex.EncodeToString(clientProfilemd5Sum[:])
```
rewantsoni, line 739: client op uses a different hash function, from what we have at CalculateMD5Hash, hence using this to make sure the clientProfileName is same that we have on the client side

services/provider/server/server.go
```diff
@@ -734,7 +735,8 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
 		}
 	}
 
-	clientProfile := util.CalculateMD5Hash(req.StorageClaimName)
```
leelavg, line 737: where else was this being used in this func? I see new code is added using it but no change in existing code, is that the expectation?

rewantsoni: This was also being used in VolumeReplicationClass, This probably is a bug as well that was missed. 
If we have a different clusterID in VolumeReplicationClass it might cause problems. Yes, the expectation is that the clusterID should be the same for VRC, VGSC, VSC, and SC and should be the clientProfileName

leelavg: ack, so this changes implicitly fixes other bug?

rewantsoni: Yes right

leelavg: resolve.

services/provider/server/server.go
rewantsoni: Resolving since Madhu cannot resolve the comments

services/provider/server/server.go
rewantsoni: Resolving since Madhu cannot resolve the comments

Madhu-1:
LGTM

services/provider/server/server.go
```diff
@@ -937,13 +941,15 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
 					}),
 				},
 				&pb.ExternalResource{
-					Name: "cephfs",
+					Name: fmt.Sprintf("%s-groupsnapclass", req.StorageClaimName),
```
iamniting, line 944: How is it matching with the internal mode, as it does not include `ceph-rbd` in the name? Are we relying on the claim name to have that in its name?

iamniting (changes requested):

services/provider/server/server.go
rewantsoni: The default storageClaim name for internal modes are "ocs-storagecluster-ceph-rbd" and "ocs-storagecluster-cephfs"

iamniting (approved):

PR #2997, opened 2025-01-31
external: fix rados

parth-gr:


parth-gr:
Fix in https://github.com/red-hat-storage/ocs-operator/pull/3001

PR #2999, opened 2025-02-03
Rename hostname key to host in supportedLabels to resolve mismatch

iamniting:
The failure domain is determined as "host", but the supportedLabels list used "hostname". This mismatch caused the topologyKey in the spec to be empty during topology generation, as the "host" key was not present. Updated the key in supportedLabels to "host" to resolve the issue.

It was introduced in https://github.com/red-hat-storage/ocs-operator/pull/2965


malayparida2000 (approved):
/lgtm

PR #3001, opened 2025-02-03
Fix issue with the ceph block pool rados namespace storage class

iamniting:

Bug: https://issues.redhat.com/browse/DFBUGS-1512

### DIFF
--- a/controllers/storagecluster/external_resources.go
+++ b/controllers/storagecluster/external_resources.go
@@ -214,13 +214,6 @@ func (r *StorageClusterReconciler) newExternalCephObjectStoreInstances(
 func (obj *ocsExternalResources) ensureCreated(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) (reconcile.Result, error) {
 
 	// rhcs external mode
-	data, err := r.retrieveExternalSecretData(instance)
-	if err != nil {
-		r.Log.Error(err, "Failed to retrieve external secret resources.")
-		return reconcile.Result{}, err
-	}
-	externalOCSResources[instance.UID] = data
-
 	extSecretChecksum, err := r.getExternalSecretDataChecksum(instance)
 	if err != nil {
 		r.Log.Error(err, "Failed to get checksum of external secret data.")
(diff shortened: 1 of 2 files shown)


controllers/storagecluster/reconcile.go
```diff
@@ -459,9 +465,9 @@ func (r *StorageClusterReconciler) reconcilePhases(
 		// for external cluster, we have a different set of ensure functions
 		// preserve list order
```
parth-gr, line 466: @aruniiird please review as we are changing the order

controllers/storagecluster/reconcile.go
```diff
@@ -415,6 +415,12 @@ func (r *StorageClusterReconciler) reconcilePhases(
 		return reconcile.Result{}, err
 	}
 
+	if instance.Spec.ExternalStorage.Enable {
+		if err := r.setExternalOCSResourcesData(instance); err != nil {
+			return reconcile.Result{}, err
```
parth-gr, line 420: can we also log the error?

controllers/storagecluster/reconcile.go
```diff
@@ -415,6 +415,12 @@ func (r *StorageClusterReconciler) reconcilePhases(
 		return reconcile.Result{}, err
 	}
 
+	if instance.Spec.ExternalStorage.Enable {
```
parth-gr, line 418: can we move to  above 
`objs = []resourceManager{`

controllers/storagecluster/external_resources.go
```diff
@@ -248,6 +241,20 @@ func (obj *ocsExternalResources) ensureDeleted(_ *StorageClusterReconciler, _ *o
 	return reconcile.Result{}, nil
 }
 
+// setExternalOCSResourcesData retrieves data from the secret and stores it in the externalOCSResources global variable.
+func (r *StorageClusterReconciler) setExternalOCSResourcesData(instance *ocsv1.StorageCluster) error {
```
parth-gr, line 245: check if the reconciliation its get updated if we change the JSON

parth-gr (changes requested):

controllers/storagecluster/reconcile.go
iamniting: yes i can do that.

controllers/storagecluster/reconcile.go
iamniting: Actually i have kept it separate so that the lists logic wont get clutter.

controllers/storagecluster/external_resources.go
iamniting: yes this func is being called always which means it will get updated.

parth-gr (approved):

controllers/storagecluster/reconcile.go
aruniiird: Order part, we really have to test it out in a real cluster

iamniting: We did a test and it was passed.

PR #3005, opened 2025-02-04
update the odf-info api to contain the exported endpoint address

rewantsoni:
The ocs-provider-server endpoint will not always be reachable from the other side as the user might have configured the networking differently. To address this, we need to make sure that the user can specify the exported endpoint as an annotation to the storageCluster with the key `ocs.openshift.io/ocs-api-endpoint-exported-address` and we add and pass the info to the odf-info config map to be used by MCO in case of establishing DR.


rewantsoni:
Testing:
```
[rewantsoni aws] k get storagecluster ocs-storagecluster -oyaml | yq '.metadata.annotations'
ocs.openshift.io/deployment-mode: provider
ocs.openshift.io/ocs-api-endpoint-exported-address: c4.ocs-provider-server.openshift-storage.svc.clusterset.local:50051
uninstall.ocs.openshift.io/cleanup-policy: delete
uninstall.ocs.openshift.io/mode: graceful
[rewantsoni aws] k get cm odf-info -oyaml
apiVersion: v1
data:
  openshift-storage_ocs-storagecluster.config.yaml: |
    version: 4.18.0-109.stable
    deploymentType: internal
    clients:
    - name: ocs-storagecluster
      clusterId: d05ab998-a9a2-4dbf-99d1-50d2a40df699
      clientId: 53d390a5-3984-4aad-9722-e96f4d7c9851
    storageCluster:
      namespacedName:
        namespace: openshift-storage
        name: ocs-storagecluster
      storageProviderEndpoint: 172.30.11.31:50051
      cephClusterFSID: 96e87814-46e3-4124-9343-f14c452aa653
      storageClusterUID: 81d635aa-a654-4bee-8e55-47adcbb9117a
      annotations:
        ocs.openshift.io/ocs-api-endpoint-exported-address: c4.ocs-provider-server.openshift storage.svc.clusterset.local:50051
    storageSystemName: ocs-storagecluster-storagesystem
kind: ConfigMap
metadata:
  creationTimestamp: "2025-01-20T07:50:01Z"
  name: odf-info
  namespace: openshift-storage
  resourceVersion: "18041749"
  uid: b5e60618-11a2-43f8-8288-feaf1219442c
  ````

File: api/v1alpha1/odfinfoconfig.go [hunk 1]
```diff
@@ -18,6 +18,8 @@ package v1alpha1
 
 import "k8s.io/apimachinery/pkg/types"
 
+const OcsApiEndpointExportedAddress = "ocs.openshift.io/ocs-api-endpoint-exported-address"
```
umangachapagain, line 21: How about this?
```suggestion
const OcsApiEndpointExportedAddress = "ocs.openshift.io/api-public-address"
```

controllers/storagecluster/odfinfoconfig.go
```diff
@@ -137,6 +137,12 @@ func getOdfInfoData(r *StorageClusterReconciler, storageCluster *ocsv1.StorageCl
 		return "", err
 	}
 
+	annotation := map[string]string{}
+	ocsApiEndpointExportedAddress := storageCluster.GetAnnotations()[ocsv1a1.OcsApiEndpointExportedAddress]
```
umangachapagain, line 141: What is the maximum length of string supported by annotation? Would it be enough to capture the address?

rewantsoni: Yes, I don't think there is a restriction on the value of the annotation key. We capture the onboarding ticket on storageConsumer CR that is 800+ characters

api/v1alpha1/odfinfoconfig.go
rewantsoni: I think we should keep ocs in the key, maybe `ocs.openshift.io/ocs-api-public-address`?
I will wait for suggestions from @nb-ohad and then push the final annotation key.

api/v1alpha1/odfinfoconfig.go
```diff
@@ -31,6 +33,7 @@ type InfoStorageCluster struct {
 	StorageProviderEndpoint string               `yaml:"storageProviderEndpoint"`
```
umangachapagain, line 33: Should this be removed?

api/v1alpha1/odfinfoconfig.go
umangachapagain: You already have `ocs.openshift.io/` in the name. What good does the additional `ocs-` do? It's redundant.

nb-ohad: I agree we should drop the ocs part but then we get api-endpoint which is kind of confusing. 
Can we change this to `ocs.openshift.io/api-server-exported-address` instead?

nb-ohad, line 21: Please add the suffix `AnnotationName` to your var name

api/v1alpha1/odfinfoconfig.go
nb-ohad: Let's keeps this for now until we are sure no one else (including Fusion) is using this field.

controllers/storagecluster/odfinfoconfig.go
nb-ohad: I think length restriction only apply to labels, not annotations

controllers/storagecluster/odfinfoconfig.go
```diff
@@ -137,6 +137,12 @@ func getOdfInfoData(r *StorageClusterReconciler, storageCluster *ocsv1.StorageCl
...
+	if ocsApiEndpointExportedAddress != "" {
+		annotation[ocsv1a1.OcsApiEndpointExportedAddress] = ocsApiEndpointExportedAddress
+	}
```
nb-ohad, line 144: Please copy all annotations from the storage cluster under the OCS domain. We should not have special code to just handle this annotation. That was the main point behind choosing an exporting mechanism

nb-ohad (changes requested):

controllers/storagecluster/odfinfoconfig.go
```diff
@@ -137,6 +137,13 @@ func getOdfInfoData(r *StorageClusterReconciler, storageCluster *ocsv1.StorageCl
 		return "", err
 	}
 
+	annotation := map[string]string{}
+	for key, value := range storageCluster.GetAnnotations() {
+		if strings.Contains(key, "ocs.openshift.io") {
```
nb-ohad, line 142:
```suggestion
		if strings.HasPrefix(key, "ocs.openshift.io/") {
```

controllers/storagecluster/odfinfoconfig.go
```diff
@@ -137,6 +137,13 @@ func getOdfInfoData(r *StorageClusterReconciler, storageCluster *ocsv1.StorageCl
 		return "", err
 	}
 
+	annotation := map[string]string{}
```
nb-ohad, line 140:
```suggestion
	annotations := map[string]string{}
```

nb-ohad (changes requested):

controllers/storagecluster/odfinfoconfig.go
```diff
@@ -137,6 +137,14 @@ func getOdfInfoData(r *StorageClusterReconciler, storageCluster *ocsv1.StorageCl
 		return "", err
 	}
 
+	annotations := map[string]string{}
+	for key, value := range storageCluster.GetAnnotations() {
+		parts := strings.Split(key, "/")
+		if len(parts) == 2 && strings.HasSuffix(parts[0], "ocs.openshift.io") {
```
umangachapagain, line 143: Just a note: As a side effect this will also include other annotations like "multicluster.ocs.openshift.io", if any.

PR #3006, opened 2025-02-04
Fix CephFS volumes failing to mount after upgrade to 4.18 in Provider Mode

malayparida2000:
Until 4.18 Provider mode was using v1(6789) port as default, so v1 ports were present in the rook-ceph-mon-endpoints CM. Rook doesn’t update this CM to v2 3300 port until the mons are failed over, even after requireMsgr2 is set to true. Provider sends the mon endpoints from the same rook-ceph-mon-endpoints CM to the client, so the client uses the v1 (6789) port address it received in it’s ceph-csi-config CM.

But client receives the cephFS kernel mount option from provider as ‘prefer-crc’ as requireMsgr2 is true. When mounting new cephFS volume on client side it tries to use the v1 6789 port with the ‘prefer-crc’ kernel mount option. Which can't work,thus cephFS volumes fail to mount.

As since 4.18 we are using v2 port always, so the provider should send the v2 port address to the client by modifying the mon IPs. Similar implementation can be seen in rook.

### DIFF
--- a/services/provider/server/server.go
+++ b/services/provider/server/server.go
@@ -1077,7 +1077,28 @@ func extractMonitorIps(data string) ([]string, error) {
 	}
 	// sorting here removes any positional change which reduces spurious reconciles
 	slices.Sort(ips)
+
+	// Rook does not update the rook-ceph-mon-endpoint ConfigMap until mons failover
+	// Starting from 4.18, RequireMsgr2 is always enabled, and encryption in transit is allowed on existing clusters.
+	// So, we need to replace the msgr1 port with msgr2 port.
+	replaceMsgr1PortWithMsgr2(ips)
+
 	return ips, nil
+}
+
+func replaceMsgr1PortWithMsgr2(ips []string) {
+	const (
+		// msgr2port is the listening port of the messenger v2 protocol
+		msgr2port = "3300"
+		// msgr1port is the listening port of the messenger v1 protocol
+		msgr1port = "6789"
+	)
+
+	for i, ip := range ips {
+		if strings.HasSuffix(ip, msgr1port) {
+			ips[i] = strings.TrimSuffix(ip, msgr1port) + msgr2port
+		}
+	}
 }
 
 func (s *OCSProviderServer) PeerStorageCluster(ctx context.Context, req *pb.PeerStorageClusterRequest) (*pb.PeerStorageClusterResponse, error) {
--- a/services/provider/server/server_test.go
+++ b/services/provider/server/server_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"reflect"
 	"strconv"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


leelavg:
> until the mons are failed over,
- is pod restart enough for mons failover?

> cephFS volumes fail to mount.
- so current PR is a proactive fix and requires backport? how is this found out?

> As since 4.18 we are using v2 port always
- I suppose as a product we are defaulting to v2 from 4.18?

> Similar implementation can be seen in rook.
- possible to link a PR if possible? asking because based on my first question if restart of mons trigger a failover and rook updates the configmap then why should provider server need to deal w/ rook/ceph particulars?

malayparida2000:
According to the discussion with @travisn . Similar mechanism is present in rook here
https://github.com/rook/rook/blob/c0a3db95a25f55ff4b68570dc380a223e97dd27c/pkg/operator/ceph/csi/cluster_config.go#L88-L104

leelavg:
> Similar mechanism is present in rook here
- ack, that is internal to rook and maybe along some code paths shouldn't rook update the mon endpoints in the configmap?

malayparida2000:
> > Similar mechanism is present in rook here
> 
> * ack, that seems to be internal to rook and maybe along some code paths shouldn't rook update the mon endpoints in the configmap?

That was my expectation as well. But @travisn suggested that rook doesn't update that configmap until mons are failed over due to some other considerdations.

@travisn can add more once he is availble today.

malayparida2000:
> is pod restart enough for mons failover?
No, Mon failover is a lengthy process. We have to scale each mon down, Wait for around 5 mins, then rook brings another new mon. Have to do this for all availble mons one after another.

> so current PR is a proactive fix and requires backport? how is this found out?
yes, Due to severe lack of provider mode setups QE was unable to proivde a proivder mode setup to me for testing. I got hold of a setup after much request yesterday and found this.

> I suppose as a product we are defaulting to v2 from 4.18?
Right, from 4.18 we will be always using v2 port for all modes.

> possible to link a PR if possible?
https://github.com/rook/rook/blob/c0a3db95a25f55ff4b68570dc380a223e97dd27c/pkg/operator/ceph/csi/cluster_config.go#L88-L104

malayparida2000:
The fix works, Tested this just now on a provider mode setup from QE.

travisn:
@leelavg What @malayparida2000 said is correct, the port is not expected to update in that rook-ceph-mon-endpoints configmap. Rook only converts them immediately for the CSI driver. That configmap is really not intended to be an API, so it may require special implementations like this.

travisn (commented):
/lgtm

leelavg:
> That configmap is really not intended to be an API, so it may require special implementations like this.
- Ack, as rook-ceph-csi-config can be considered as an interface for CSI does the ports get updated in that? iirc, it also has endpoints, isn't it?

travisn:
> > That configmap is really not intended to be an API, so it may require special implementations like this.
> 
> * Ack, as rook-ceph-csi-config can be considered as an interface for CSI does the ports get updated in that? iirc, it also has endpoints, isn't it?

Yes, CSI gets the ports in that configmap. But soon that configmap will not be created anymore as we convert to the CSI operator. Instead, the CephConnection CR would have the info needed for CSI. But with convergence, anyway Rook will not be generating that either, so I don't see another option that can be used besides the rook-ceph-mon-endpoints.

PR #3007, opened 2025-02-04
[release-4.18] Fix CephFS volumes failing to mount after upgrade to 4.18

malayparida2000:
Until 4.18 Provider mode was using v1(6789) port as default, so v1 ports were present in the rook-ceph-mon-endpoints CM. Rook doesn’t update this CM to v2 3300 port until the mons are failed over, even after requireMsgr2 is set to true. Provider sends the mon endpoints from the same rook-ceph-mon-endpoints CM to the client, so the client uses the v1 (6789) port address it received in it’s ceph-csi-config CM.

But client receives the cephFS kernel mount option from provider as ‘prefer-crc’ as requireMsgr2 is true. When mounting new cephFS volume on client side it tries to use the v1 6789 port with the ‘prefer-crc’ kernel mount option. Which can't work,thus cephFS volumes fail to mount.

As since 4.18 we are using v2 port always, so the provider should send the v2 port address to the client by modifying the mon IPs. Similar implementation can be seen in rook.


travisn (commented):
/lgtm

PR #3008, opened 2025-02-05
provider: update the storageID for ramen

rewantsoni:
update ramen storageID to be:
for rbd: hash(cephCluster_FSID,blockpool_id,rns_name)
for cephfs: hash(cephCluster_FSID,filesystem_name,svg_name)

This is align both provider and internal mode to have the same storageID and to make the storageID future proof for Metro DR as well


controllers/util/k8sutil.go
```diff
@@ -203,6 +204,20 @@ func GetStorageClusterInNamespace(ctx context.Context, cl client.Client, namespa
...
+	if err != nil {
+		return nil, fmt.Errorf("unable to list cephCluster(s) in namespace %s: %v", namespace, err)
+	}
+
+	if len(cephClusterList.Items) == 0 {
+		return nil, fmt.Errorf("no cephCluster found in namespace %s", namespace)
+	}
+
+	return &cephClusterList.Items[0], nil
+}
```
nb-ohad, line 219: What happens when you have both internal and external Ceph clusters?

services/provider/server/server.go
```diff
@@ -799,6 +803,19 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
...
+			if err != nil {
+				return nil, status.Errorf(codes.Internal, "failed to get %s CephBlockPool. %v", blockPool.Name, err)
+			}
+
+			// SID for RamenDR
+			storageID := util.CalculateMD5Hash(fmt.Sprintf("%v-%v-%v",
+				cephCluster.Status.CephStatus.FSID,
+				blockPool.Status.PoolID,
+				rns.Name),
+			)
```
nb-ohad, line 817: Don't use string concatenation for producing ids. 
Use a struct (embedded struct) or array to represent the tuple - preferably a struct because you don't need to worry about ordering. Then serialize it to a well-know serialisation (either JSON or YAML). And lastly calculate the hash for the serialized string

services/provider/server/server_test.go
nb-ohad, line 1: Why are there changes to this file?

nb-ohad (changes requested):

controllers/util/k8sutil.go
rewantsoni: Since cephCluster deployment is controlled by storageCluster, We will not have more than one cephcluster in one namespace

services/provider/server/server_test.go
rewantsoni: Since the storageID changes, we need to update the tests to adapt to the new storageID

services/provider/server/server.go
rewantsoni: used a struct for producing the ids. The CalculateMD5Hash handle the serialisation to json and calculates the hash for the serialized string

services/provider/server/server.go
```diff
@@ -915,6 +936,17 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
 				kernelMountOptions[parts[0]] = parts[1]
 			}
 
+			// SID for RamenDR
+			storageID := util.CalculateMD5Hash(struct {
```
umangachapagain, line 940: storageID has to be within allowed string length. How is that ensured?

rewantsoni: The CalculateMD5Hash returns a hash that is 32 characters long

services/provider/server/server.go
```diff
@@ -799,6 +803,23 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
...
+				CephClusterID      string `json:"cephClusterID"`
+				BlockPoolID        string `json:"blockPoolID"`
+				RadosNamespaceName string `json:"radosNamespaceName"`
+			}{
+				cephCluster.Status.CephStatus.FSID,
+				fmt.Sprintf("%v", blockPool.Status.PoolID),
+				rns.Name,
+			})
```
nb-ohad, line 821: Can we have CalculateStorageId as its own function/util? If not then the usage of an anonymous struct is problamatic as it require the redefinition of the struct on each call site

nb-ohad (changes requested):

PR #2276, opened 2023-11-20
Bug 2248449: Fusion HCI integration changes for OCS operator [4.14 Backport]

nb-ohad:
This PR contains backports of multiple fixes and enhancements to allow integration of the OCS Operator with Fusion HCI deployment and management model

PR #3012, opened 2025-02-05
remove the storageprofile apis

iamniting:
storageprofile was depricated and it is not used
    
remove changes of below commits:
8809ca0b601720aec7e7cbf68442ae30a57da801 #2276
c29859e63dd90b8ad53104ad3685ab42aa48d29e #2288

Bug: https://issues.redhat.com/browse/DFBUGS-1531


controllers/storagecluster/initialization_reconciler_test.go
```diff
@@ -149,87 +149,6 @@ func createUpdateRuntimeObjects(t *testing.T) []client.Object {
...
-		result, err := reconciler.Reconcile(context.TODO(), request)
-		assert.NoError(t, err)
-		assert.Equal(t, reconcile.Result{}, result)
-		err = os.Setenv("WATCH_NAMESPACE", request.Namespace)
-		assert.NoError(t, err)
-	}
-
-	return t, reconciler, cr, requestOCSInit, requestsStorageProfiles
-}
-
```
nb-ohad, line 232: Why are we removing all of this and not just the storage profile parts?

controllers/storagecluster/cephfilesystem_test.go
nb-ohad, line 1: What is the reason to have the deletions in this file?

nb-ohad (changes requested):

controllers/storagecluster/initialization_reconciler_test.go
iamniting: This is not being used anywhere else and it was created for storageprofile tests itself. you can refer to the 2nd commit in the description for the same.

controllers/storagecluster/cephfilesystem_test.go
iamniting: This func was also altered while adding tests for the storage profile, but it was not changed properly while removing the tests.

PR #3019, opened 2025-02-06
deprecate and remove the allowRemoteStorageConsumer spec from storageCluster 

rewantsoni:


controllers/storagecluster/noobaa_system_reconciler.go
```diff
@@ -139,9 +139,7 @@ func (r *StorageClusterReconciler) setNooBaaDesiredState(nb *nbv1.NooBaa, sc *oc
 		"app": "noobaa",
 	}
 
-	if sc.Spec.AllowRemoteStorageConsumers {
-		util.AddAnnotation(nb, "MulticloudObjectGatewayProviderMode", "true")
-	}
+	util.AddAnnotation(nb, "MulticloudObjectGatewayProviderMode", "true")
```
leelavg, line 142: does any change to this needs a touch in Noobaa? Let's note such instances directly on corresponding Jira for posterity.

can you get more info on this? is removing this check makes MCG behave differently than internal mode, which we don't want.

controllers/storagecluster/storageclasses.go
```diff
@@ -461,12 +461,10 @@ func (r *StorageClusterReconciler) newStorageClassConfigurations(initData *ocsv1
 		newCephBlockPoolStorageClassConfiguration(initData),
 	}
 
-	// when allowing consumers, creation of storage classes should only be done via storagerequests
-	if !initData.Spec.AllowRemoteStorageConsumers {
-		// If kubevirt crd is present, we create a specialized rbd storageclass for virtualization environment
-		if r.AvailableCrds[VirtualMachineCrdName] {
-			ret = append(ret, newCephBlockPoolVirtualizationStorageClassConfiguration(initData))
-		}
+	//TODO: will be removed when we start handling kubevirt class through provider-client communication
```
leelavg, line 464: as mentioned above, pls add these notes to corresponding Jira as well, preferably as a single comment (after PR merge)?

controllers/storagecluster/cephblockpools.go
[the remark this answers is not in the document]
rewantsoni: The code paths that are removed in this profile were ensuring that mirroring of cephblockpool is established for DR. The replacement for that Mirroring controller, that handles mirroring of blockpools as well as radosnamespace.
There would need to be adjustments in MCO op as well such that they follow the new method of setting up RDR

controllers/storagecluster/storageclasses.go
rewantsoni: ack, I will add these to jira

controllers/storagecluster/cephblockpools.go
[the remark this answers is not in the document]
leelavg: > There would need to be adjustments in MCO
- Ack, this what I was looking for, thanks.

controllers/storagecluster/noobaa_system_reconciler.go
rewantsoni: Adding this annotation sets the "RESTRICT_RESOURCE_DELETION" env variable in noobaa. This ensures that the bucket can be deleted only by the owner of the bucket and not by any other noobaa account.
cc: @ezio-auditore

catalog/index.yaml
```diff
@@ -9,6 +9,17 @@ name: alpha
 entries:
   - name: ocs-operator.v4.19.0
 
+---
+defaultChannel: alpha
```
Madhu-1, line 13: can you please move this to a different commit with details for this change?

controllers/storagecluster/cephblockpools.go
```diff
@@ -18,29 +18,6 @@ import (
 
 type ocsCephBlockPools struct{}
 
-// ensures that peer cluster secret exists and adds it to CephBlockPool
-func (o *ocsCephBlockPools) addPeerSecretsToCephBlockPool(r *StorageClusterReconciler, storageCluster *ocsv1.StorageCluster, poolName, poolNamespace string) *cephv1.MirroringPeerSpec {
```
Madhu-1, line 22: This commit is having 3 different changes? can we move specific changes to a commit (with details) so that it helps in reviewing and captures the history on why its changed

controllers/storagecluster/cephfilesystem.go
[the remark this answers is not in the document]
Madhu-1: Why, shouldnt we need to have storageconsumer as the source of truth for the SVG and RNS? we should not be creating it in multiple classes

controllers/storagecluster/volumegroupsnapshotterclasses.go
```diff
@@ -56,15 +56,19 @@ func newCephFilesystemGroupSnapshotClassConfiguration(instance *ocsv1.StorageClu
 	return GroupSnapshotClassConfiguration{
 		groupSnapshotClass: newVolumeGroupSnapshotClass(instance, cephfsGroupSnapshotter),
 		reconcileStrategy:  ReconcileStrategy(instance.Spec.ManagedResources.CephFilesystems.ReconcileStrategy),
-		disable:            instance.Spec.AllowRemoteStorageConsumers,
+		//TODO: This resource will be managed by the provider-client flow, replacing it with same field used to disable
```
Madhu-1, line 59: open a jira task for the specific TODO so that we dont miss it

pkg/deploy-manager/subscription.go
```diff
@@ -186,6 +186,20 @@ func (t *DeployManager) generateClusterObjects(ocsCatalogImage string, subscript
 	}
 	ocsSubscription.SetGroupVersionKind(schema.GroupVersionKind{Group: v1alpha1.SchemeGroupVersion.Group, Kind: "Subscription", Version: v1alpha1.SchemeGroupVersion.Version})
 
+	ocsClientSubscription := v1alpha1.Subscription{
```
Madhu-1, line 189: can you please add a comment for this change?

controllers/storagecluster/cephfilesystem_test.go
```diff
@@ -2,12 +2,12 @@ package storagecluster
 
 import (
 	"context"
+	"k8s.io/apimachinery/pkg/types"
```
nb-ohad, line 5: This should not have been moved here

controllers/storagecluster/noobaa_system_reconciler.go
nb-ohad: How would that affect users who deletes buckets using S3? or using the new s3 bowser experience?

nb-ohad (changes requested):

controllers/storagecluster/volumegroupsnapshotterclasses.go
rewantsoni: There is a jira already present and Malay is working on it: https://github.com/red-hat-storage/ocs-operator/pull/3037

controllers/storagecluster/noobaa_system_reconciler.go
rewantsoni: Correction: It is meant to restrict the deletion of a resource (backingstore\namespacestore) and only allow it to be deleted by the resources creator or system admin.

leelavg (approved):

rewantsoni:
Updated the commit message

rewantsoni:
Rebased it against main

PR #3021, opened 2025-02-07
added missing alert from upstream

rchikatw:
Jira: https://issues.redhat.com/browse/DFBUGS-1547
- Runbook will be added as part of different PR and different repo
- Test results are added as comment

### DIFF
--- a/controllers/storagecluster/prometheus/localcephrules.yaml
+++ b/controllers/storagecluster/prometheus/localcephrules.yaml
@@ -187,6 +187,18 @@ spec:
       for: 1m
       labels:
         severity: critical
+    - alert: CephOSDDown
+      annotations:
+        description: Ceph OSD(s) is/are down, please check the ceph-osd daemons and take corrective measures. Please see runbook URL for more details
+        message: An OSD has been marked down
+        severity_level: warning
+        storage_type: ceph
+        runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/openshift-container-storage-operator/CephOSDDown.md
+      expr: |
+        ceph_health_detail{name="OSD_DOWN"} == 1
+      for: 5m
+      labels:
+        severity: warning
     - alert: CephOSDSlowOps
       annotations:
         description: '{{ $value }} Ceph OSD requests are taking too long to process. Please check ceph status to find out the cause.'


rchikatw:
Test results:
Screenshot from 2025-02-07 16-27-19
Screenshot from 2025-02-07 16-27-45

rchikatw:
Adding screenshot for test result
Screenshot from 2025-02-07 16-55-23
Screenshot from 2025-02-07 16-55-39

rchikatw:
@travisn please review

controllers/storagecluster/prometheus/localcephrules.yaml
```diff
@@ -187,6 +187,18 @@ spec:
...
         severity: critical
+    - alert: CephOSDDown
+      annotations:
+        description: Ceph OSD(s) is/are down, please check the ceph-osd daemons and take corrective measures. Please see runbook URL for more details
+        message: An OSD has been marked down
+        severity_level: warning
+        storage_type: ceph
+        runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/openshift-container-storage-operator/CephOSDDown.md
+      expr: |
+        ceph_health_detail{name="OSD_DOWN"} == 1
```
travisn, line 198: If there is more than one OSD down, this alert will still fire, right? Checking for `1` is just that there is one health warning regardless of the number of OSDs down, correct?

controllers/storagecluster/prometheus/localcephrules.yaml
```diff
@@ -187,6 +187,18 @@ spec:
       for: 1m
       labels:
         severity: critical
+    - alert: CephOSDDown
```
travisn, line 190: An OSD being down isn't really a health warning, if Ceph has since repaired the PGs. This alert would be better if it were only raised if PGs were not clean. But perhaps that's better as a follow-up PR with more time to review the behavior.

rchikatw: Agree.

controllers/storagecluster/prometheus/localcephrules.yaml
rchikatw: yes true.

aruniiird (approved):
LGTM
Make sure we merge the runbook PR soon

iamniting (approved):

PR #3022, opened 2025-02-10
shim layer for sending claim resources referring to storageconsumers

leelavg:
until client-op is on same version as provider running convergence code provider should continue to support older clients by sending already claimed resources.

new claim creation and deletion of existing claims are no-ops but fulfill storageclaim continues to send already created resources before upgrade until client-op also upgraded.


leelavg:
Part of RHSTOR-6918

skipped unit tests will be fixed in a new commit.

services/provider/server/consumer.go
```diff
@@ -55,6 +58,43 @@ func newConsumerManager(ctx context.Context, cl client.Client, namespace string)
...
+		panic("failed to marshal storage class request name")
+	}
+	md5Sum := md5.Sum(requestName)
+	return hex.EncodeToString(md5Sum[:16])
+}
+
+// getStorageRequestName generates a name for a StorageRequest resource.
+func getStorageRequestName(consumerUUID, storageClaimName string) string {
+	return fmt.Sprintf("storagerequest-%s", getStorageRequestHash(consumerUUID, storageClaimName))
+}
```
nb-ohad, line 84: Do we really need these 2 functions as standalone functions? I see the logic belong for a single flow which is the `GetRequest` logic. Can we fold everything there? 
It will make it easier to move around and potentially remove in the future.

services/provider/server/consumer.go
```diff
@@ -55,6 +58,43 @@ func newConsumerManager(ctx context.Context, cl client.Client, namespace string)
...
+
+func (c *ocsConsumerManager) GetRequest(ctx context.Context, consumerUUID, storageClaimName string) (*ocsv1alpha1.StorageRequest, error) {
```
nb-ohad, line 86: Please add a "TODO"  comment explaining that this function is to be remove when we do the jump to 4.20

services/provider/server/consumer.go
```diff
@@ -55,6 +58,43 @@ func newConsumerManager(ctx context.Context, cl client.Client, namespace string)
...
+	generatedRequestName := getStorageRequestName(consumerUUID, storageClaimName)
+	storageRequestObj := &ocsv1alpha1.StorageRequest{}
+	err := c.client.Get(ctx, types.NamespacedName{Name: generatedRequestName, Namespace: c.namespace}, storageRequestObj)
+	if err != nil {
+		klog.Errorf("failed to get a StorageRequest named %q for consumer %q and request %q. %v", generatedRequestName, consumerUUID, storageClaimName, err)
+		return nil, err
+	}
```
nb-ohad, line 93: Isn't the point to remove the storage request API with this shim. If so it does not make sense to load the resource

services/provider/server/server.go
```diff
@@ -658,30 +651,13 @@ func decodeAndValidateTicket(ticket string, pubKey *rsa.PublicKey) (*services.On
...
-	case pb.FulfillStorageClaimRequest_SHAREDFILE:
-		storageType = "sharedfile"
-	default:
-		return nil, status.Errorf(codes.InvalidArgument, "encountered an unknown storage type, %s", storageType)
-	}
-
-	err = s.storageRequestManager.Create(ctx, consumerObj, req.StorageClaimName, storageType, req.EncryptionMethod, req.StorageProfile)
+	// this is only here for backward compatibility as we expect no existing older client
+	// wants to fulfill claims as current provider deprecated storagerequests
+	_, err := s.consumerManager.GetRequest(ctx, req.StorageConsumerUUID, req.StorageClaimName)
```
nb-ohad, line 656: I think the better approach is to always return internal error (or some other error) for this operation

services/provider/server/server.go
```diff
@@ -692,13 +668,8 @@ func (s *OCSProviderServer) FulfillStorageClaim(ctx context.Context, req *pb.Ful
...
 func (s *OCSProviderServer) RevokeStorageClaim(ctx context.Context, req *pb.RevokeStorageClaimRequest) (*pb.RevokeStorageClaimResponse, error) {
-	err := s.storageRequestManager.Delete(ctx, req.StorageConsumerUUID, req.StorageClaimName)
-	if err != nil {
-		errMsg := fmt.Sprintf("failed to revoke storage class claim %q for %q. %v", req.StorageClaimName, req.StorageConsumerUUID, err)
-		klog.Error(errMsg)
-		return nil, status.Error(codes.Internal, errMsg)
-	}
-
+	// client cluster want to stop using storageclasses and data on the backend can be removed
+	// TODO: this should be transferred into removal of storageclasses and corresponding data
```
nb-ohad, line 672: I think the better approach is to always return internal error (or some other error) for this operation

nb-ohad (changes requested):

services/provider/server/consumer.go
leelavg: > remove the storage request API with this shim
- not the intention w/ this specific PR, I'm trying to remove build time dependency on StorageRequest and when all the required resources (via different PRs but in 4.19 itself) are updated on StorageConsumer API we'll directly send the necessary details to older clients from there itself

leelavg:
/hold

until storageconsumer API is fully developed.

leelavg:
closing in favor of #3145

PR #3026, opened 2025-02-10
provider: update the replicationID for ramen

rewantsoni:
Ramen's replication ID is combination of storageIDs.

storageID is a combination of cephfsid,cephblockpool id and the rns name
To generate the peerStorageID, 
- use the blockpool's mirroring token to get the peer cephfsid.
- to the cephblockpool id from the peer cluster, use the annotation on the blockpool
- use the rns.Spec.Mirroring.RemoteNamespace to fetch the remote namespace

### DIFF
--- a/controllers/mirroring/mirroring_controller.go
+++ b/controllers/mirroring/mirroring_controller.go
@@ -355,7 +355,7 @@ func (r *MirroringReconciler) reconcileBlockPoolMirroring(
 				cephBlockPool := blockPoolByName[blockPoolName]
 
 				mirroringToken := response.BlockPoolsInfo[i].MirroringToken
-				secretName := fmt.Sprintf("rbd-mirroring-token-%s", util.CalculateMD5Hash(blockPoolName))
+				secretName := GetMirroringSecretName(blockPoolName)
 				if mirroringToken != "" {
 					mirroringSecret := &corev1.Secret{}
 					mirroringSecret.Name = secretName
@@ -569,4 +569,8 @@ func (r *MirroringReconciler) delete(obj client.Object, opts ...client.DeleteOpt
 
 func (r *MirroringReconciler) own(owner *corev1.ConfigMap, obj client.Object) error {
 	return controllerutil.SetControllerReference(owner, obj, r.Scheme)
+}
+
+func GetMirroringSecretName(blockPoolName string) string {
+	return fmt.Sprintf("rbd-mirroring-token-%s", util.CalculateMD5Hash(blockPoolName))
 }
--- a/services/provider/server/server.go
+++ b/services/provider/server/server.go
@@ -1,6 +1,7 @@
 package server
 
 import (
+	"cmp"
 	"context"
 	"crypto"
 	"crypto/md5"
@@ -22,6 +23,7 @@ import (
 	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
 	ocsv1alpha1 "github.com/red-hat-storage/ocs-operator/api/v4/v1alpha1"
 	pb "github.com/red-hat-storage/ocs-operator/services/provider/api/v4"
+	"github.com/red-hat-storage/ocs-operator/v4/controllers/mirroring"
 	controllers "github.com/red-hat-storage/ocs-operator/v4/controllers/storageconsumer"
 	"github.com/red-hat-storage/ocs-operator/v4/controllers/util"
 	"github.com/red-hat-storage/ocs-operator/v4/services"
@@ -747,8 +749,6 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
 		return nil, err
 	}
 
-	replicationID := req.StorageClaimName
-
 	for _, cephRes := range storageRequest.Status.CephResources {
 		switch cephRes.Kind {
 		case "CephClient":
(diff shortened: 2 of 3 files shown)


services/provider/server/server.go
```diff
@@ -810,11 +808,70 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
...
 
 			// SID for RamenDR
-			storageID := calculateCephRbdStorageId(
+			storageID := calculateCephRbdStorageID(
 				cephCluster.Status.CephStatus.FSID,
 				strconv.Itoa(blockPool.Status.PoolID),
 				rns.Name)
 
+			if blockPool.GetAnnotations()[util.BlockPoolMirroringTargetIDAnnotation] != "" {
+				peerCephfsid, err := s.getPeerCephFSID(blockPool.Spec.Mirroring.Peers.SecretNames[0])
```
nb-ohad, line 817: 1. You should not assume SecretNames is not empty
2. How do we know that we need to take the first item and not the second or the third?
3. Secret is located in the namespace of the block pool so you need to send the block pool namespace as well and don't assume it is the same namespace where the server lives
4. There is no reason the have `gerPeerCephFSID` as a method on the server if the only thing it needs from the server is the client. You should just send the client to the function as well

services/provider/server/server.go
```diff
@@ -810,11 +808,70 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
...
+				if err != nil {
+					klog.Errorf("failed to get peer Ceph FSIS. %v", err)
+				}
+
+				peerStorageID := calculateCephRbdStorageID(
+					peerCephfsid,
+					blockPool.GetAnnotations()[util.BlockPoolMirroringTargetIDAnnotation],
+					*rns.Spec.Mirroring.RemoteNamespace)
```
nb-ohad, line 825: We should not be assuming that `RemoteNamespace` is not nil

services/provider/server/server.go
```diff
@@ -810,11 +808,70 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
 			}
 
 			// SID for RamenDR
-			storageID := calculateCephRbdStorageId(
+			storageID := calculateCephRbdStorageID(
 				cephCluster.Status.CephStatus.FSID,
 				strconv.Itoa(blockPool.Status.PoolID),
 				rns.Name)
 
+			if blockPool.GetAnnotations()[util.BlockPoolMirroringTargetIDAnnotation] != "" {
```
nb-ohad, line 816: If you are already loading the annotation here, you should not do it again later in the func

services/provider/server/server.go
```diff
@@ -1332,10 +1346,38 @@ func (s *OCSProviderServer) isConsumerMirrorEnabled(ctx context.Context, consume
...
+		return "", err
+	}
+	return token.FSID, nil
+}
+
+func generateReplicationID(storageIDs ...string) string {
+	slices.Sort(storageIDs)
+	return util.CalculateMD5Hash(storageIDs)
+
+}
```
nb-ohad, line 1383: Do we really need a util for that if it is only used once?

nb-ohad (changes requested):

services/provider/server/server.go
rewantsoni: Updated since we already know the secret name. If secret is not present we would log error and not create the VRCs

services/provider/server/server.go
```diff
@@ -810,11 +808,78 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
...
 
 			// SID for RamenDR
-			storageID := calculateCephRbdStorageId(
+			storageID := calculateCephRbdStorageID(
 				cephCluster.Status.CephStatus.FSID,
 				strconv.Itoa(blockPool.Status.PoolID),
 				rns.Name)
 
+			if peerPoolID := blockPool.GetAnnotations()[util.BlockPoolMirroringTargetIDAnnotation]; peerPoolID != "" &&
+				rns.Spec.Mirroring.RemoteNamespace != nil {
```
nb-ohad, line 817: That is a very long line. Let's make it more readable
```suggestion
			peerPoolID := blockPool.GetAnnotations()[util.BlockPoolMirroringTargetIDAnnotation]
			if peerPoolID != "" && rns.Spec.Mirroring.RemoteNamespace != nil {
```

services/provider/server/server.go
```diff
@@ -810,11 +808,78 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
...
+					fmt.Sprintf("rbd-mirroring-token-%s", util.CalculateMD5Hash(blockPool.Name)),
+					blockPool.Namespace)
+				if err != nil {
+					klog.Errorf("failed to get peer Ceph FSIS. %v", err)
+				}
+
+				peerStorageID := calculateCephRbdStorageID(
+					peerCephfsid,
+					peerPoolID,
+					*rns.Spec.Mirroring.RemoteNamespace)
```
nb-ohad, line 830:
```suggestion
					*rns.Spec.Mirroring.RemoteNamespace,
				)
```

nb-ohad (changes requested):

controllers/mirroring/mirroring_controller.go
```diff
@@ -355,7 +356,7 @@ func (r *MirroringReconciler) reconcileBlockPoolMirroring(
 				cephBlockPool := blockPoolByName[blockPoolName]
 
 				mirroringToken := response.BlockPoolsInfo[i].MirroringToken
-				secretName := fmt.Sprintf("rbd-mirroring-token-%s", util.CalculateMD5Hash(blockPoolName))
+				secretName := fmt.Sprintf("%s-%s", MirroringSecretPrefix, util.CalculateMD5Hash(blockPoolName))
```
nb-ohad, line 359: If we need to export the functionality of creating a mirroring secret name then we should export it as a function and not as a prefix. This mean the change will become:
```suggestion
				secretName := GetMirroringSecretName(blockPoolName)
```

services/provider/server/server.go
```diff
@@ -810,11 +809,79 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
...
 				strconv.Itoa(blockPool.Status.PoolID),
 				rns.Name)
 
+			peerPoolID := blockPool.GetAnnotations()[util.BlockPoolMirroringTargetIDAnnotation]
+			if peerPoolID != "" && rns.Spec.Mirroring.RemoteNamespace != nil {
+				peerCephfsid, err := getPeerCephFSID(
+					ctx,
+					s.client,
+					mirroring.GetMirroringSecretName(blockPool.Name),
+					blockPool.Namespace)
```
nb-ohad, line 823:
```suggestion
					blockPool.Namespace,
				)
```

services/provider/server/server.go
```diff
@@ -810,11 +809,79 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
 			}
 
 			// SID for RamenDR
-			storageID := calculateCephRbdStorageId(
+			storageID := calculateCephRbdStorageID(
 				cephCluster.Status.CephStatus.FSID,
 				strconv.Itoa(blockPool.Status.PoolID),
 				rns.Name)
```
nb-ohad, line 815:
```suggestion
				rns.Name,
			)
```

services/provider/server/server.go
```diff
@@ -810,11 +809,79 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
...
 
+			peerPoolID := blockPool.GetAnnotations()[util.BlockPoolMirroringTargetIDAnnotation]
+			if peerPoolID != "" && rns.Spec.Mirroring.RemoteNamespace != nil {
```
nb-ohad, line 818:
```suggestion
			remoteNamespace := rns.Spec.Mirroring.RemoteNamespace
			if peerPoolID != "" && remoteNamespace != nil {
```

services/provider/server/server.go
```diff
@@ -810,11 +809,79 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
...
+					mirroring.GetMirroringSecretName(blockPool.Name),
+					blockPool.Namespace)
+				if err != nil {
+					klog.Errorf("failed to get peer Ceph FSIS. %v", err)
+				}
+
+				peerStorageID := calculateCephRbdStorageID(
+					peerCephfsid,
+					peerPoolID,
+					*rns.Spec.Mirroring.RemoteNamespace,
```
nb-ohad, line 831:
```suggestion
					*remoteNamespace,
```

nb-ohad (changes requested):

services/provider/server/server.go
rewantsoni: Mirroring struct is nullable it might throw an error

PR #3033, opened 2025-02-11
Stop adding duplicate default tolerations for the toolbox pod

malayparida2000:
The default toleration was being added twice ones as a base toleration and again from the getPlacement function. We do not need the base toleration as it is already being added by the getPlacement function.


malayparida2000:
This is not a bug as confirmed by me from testing.

PR #3044, opened 2025-02-18
autoscaling: add new crd for automatic storage scaling

parth-gr:
added a new crd for automatic  storage scaling


parth-gr:
Please review the new CRD 
/assign @BlaineEXE  @malayparida2000  @iamniting  @nb-ohad @malayparida2000

iamniting (changes requested):
Can you pls add all generated changes? Pls look at the failed github action for more details.

config/crd/bases/ocs.openshift.io_autostoragescalings.yaml
```diff
@@ -0,0 +1,127 @@
...
+                  Used for vertical scaling of OSDs.
+                pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
+                x-kubernetes-int-or-string: true
+              lastRunTimeStamp:
+                description: LastRunTimeStamp is the time stamp of the last run of
+                  the storage scaling
+                format: date-time
+                type: string
+              phase:
+                description: Phase describes the Phase of AutoStorageScaling
```
alfonsomthd, line 120: @parth-gr Shouldn't we describe the `enum` allowed values: `waiting,inProgress,completed,failed` ?

parth-gr: Its the controller that would be updating the status values, so I don't think we need enum here, wdyt?

api/v1/autostoragescale.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="Age",type=date,JSONPath=.metadata.creationTimestamp, description="Age"
+// +kubebuilder:printcolumn:name="Phase",type=string,JSONPath=.status.phase,description="Current Phase"
+// +kubebuilder:printcolumn:name="LastRunTimeStamp",type=date,JSONPath=.status.lastRunTimeStamp,description="Last Run Time Stamp"
+// +operator-sdk:csv:customresourcedefinitions:displayName="Auto Storage Scaling"
+
+// AutoStorageScaling represents the automatic storage scaling for odf cluster.
```
Madhu-1, line 68: how about calling it as `StorageAutoScaling` or `StorageAutoScaler` ?  because keeping Storage in the front aligns with the idea that we are trying to scale storage.

api/v1/autostoragescale.go
```diff
@@ -0,0 +1,75 @@
...
+)
+
+// AutoStorageScalingSpec defines the desired state of AutoStorageScaling
+type AutoStorageScalingSpec struct {
+	// StorageClusterName is the name of the storage cluster for which the storage scaling is to be done.
+	// +kubebuilder:validation:Required
+	StorageClusterName string `json:"storageClusterName,omitempty"`
+	// DeviceClassName is the name of the device class for which the storage scaling is to be done.
+	// +kubebuilder:validation:Required
+	DeviceClassName string `json:"deviceClassName,omitempty"`
```
Madhu-1, line 31: why deviceClass is required? AFAIK its optional in the storagecluster creation

api/v1/autostoragescale.go
```diff
@@ -0,0 +1,75 @@
...
+// AutoStorageScalingStatus defines the observed state of AutoStorageScaling
+type AutoStorageScalingStatus struct {
+	// Phase describes the Phase of AutoStorageScaling
+	Phase string `json:"phase,omitempty"`
+	// LastRunTimeStamp is the time stamp of the last run of the storage scaling
+	LastRunTimeStamp metav1.Time `json:"lastRunTimeStamp,omitempty"`
+	// The known size is the original size of the OSDs before the expansion in progress is completed.
+	// After the expansion is completed, this would be updated to the expected OSD size.
+	// Used for vertical scaling of OSDs.
+	KnownOsdSize resource.Quantity `json:"knownOsdSize,omitempty"`
```
Madhu-1, line 46: instead of `KnownOsdISize` does it make sense to keep this as `currentOsdSize`?

api/v1/autostoragescale.go
```diff
@@ -0,0 +1,75 @@
...
+	DeviceClassName string `json:"deviceClassName,omitempty"`
+	// CapacityLimit is the capacity limit for the storage scaling for the specific deviceClass and storagecluster.
+	// +kubebuilder:validation:Required
+	CapacityLimit resource.Quantity `json:"capacityLimit,omitempty"`
+}
+
+// AutoStorageScalingStatus defines the observed state of AutoStorageScaling
+type AutoStorageScalingStatus struct {
+	// Phase describes the Phase of AutoStorageScaling
+	Phase string `json:"phase,omitempty"`
```
Madhu-1, line 40: can we please define a known Phase const of it?

config/crd/bases/ocs.openshift.io_autostoragescalings.yaml
alfonsomthd: Right, it's not strictly necessary. My suggestion intended to have a more descriptive explanation for users just like the `StorageCluster` CRD does for `deviceType`:
```
        deviceType:
          description: |-
            DeviceType is the value of device type in
            this StorageDeviceSet. It can have one of the
            three values (SSD, HDD, NVMe)
          enum:
          - SSD
          - ssd
          - HDD
          - hdd
          - NVMe
          - NVME
          - nvme
          type: string
```

api/v1/autostoragescale.go
```diff
@@ -0,0 +1,75 @@
...
+	// Used for vertical scaling of OSDs.
+	KnownOsdSize resource.Quantity `json:"knownOsdSize,omitempty"`
+	// The ExpectedOsdSize is the size that the auto-expansion has decided to set.
+	// This will be set on the storageCLuster CR as the desired size of the OSDs.
+	// Used for vertical scaling of OSDs.
+	ExpectedOsdSize resource.Quantity `json:"expectedOsdSize,omitempty"`
+	// The known OSD count is the original count of the OSDs before the expansion in progress is completed.
+	// After the expansion is completed, this would be updated to the expected OSD count.
+	// Used for horizontal scaling of OSDs.
+	KnownOsdCount int32 `json:"knownOsdCount,omitempty"`
```
Madhu-1, line 54: same here as well. and also do we need it to be int32? or just use uint16?

api/v1/autostoragescale.go
```diff
@@ -0,0 +1,75 @@
...
+	// The Expected OSD count is the count that the auto-expansion has decided to set.
+	// This will be set on the storageCLuster CR as the desired count of the OSDs.
+	// Used for horizontal scaling of OSDs.
```
Madhu-1, line 57: `storageCLuster` to `storageCluster`

api/v1/autostoragescale.go
```diff
@@ -0,0 +1,75 @@
+/*
+Copyright 2020 Red Hat OpenShift Container Storage.
```
Madhu-1, line 2: 2025?

api/v1/autostoragescale.go
Madhu-1, line 68: `odf cluster.` to `storage cluster`

Madhu-1 (changes requested):
CRD need to be added to https://github.com/red-hat-storage/ocs-operator/blob/main/PROJECT as well.

api/v1/autostoragescale.go
parth-gr: its optionally but ocs operator alwyas add `ssd` if not provided by storagecluster

config/crd/bases/ocs.openshift.io_autostoragescalings.yaml
parth-gr: done, with constant

api/v1/autostoragescale.go
alfonsomthd: +1

parth-gr: Updated

api/v1/autostoragescale.go
BlaineEXE: I'm still confused though. The autoscaler should be autoscaling existing storageclassdevicesets, and the device class should be able to be read from the scDeviceSet being scaled, right?

alfonsomthd: > I'm still confused though. The autoscaler should be autoscaling existing storageclassdevicesets, and the device class should be able to be read from the scDeviceSet being scaled, right?

AFAIK autoscaling will apply to the devicesets of a particular deviceclass, so we need to know the deviceclass in order to filter devicesets by deviceclass.

Madhu-1: During creation its an optional it should be kept optional as well, if its not specified we should treat it the same way we do during creation.

alfonsomthd: > During creation its an optional it should be kept optional as well, if its not specified we should treat it the same way we do during creation.

You mean that if not specified it should default to `ssd` like `StorageCluster` CR, right?

Madhu-1: yes correct whatever ocs-operator chooses during storagecluster creation.

api/v1/storage_autoscaler.go
```diff
@@ -0,0 +1,85 @@
...
+)
+
+// StorageAutoScalerSpec defines the desired state of StorageAutoScaler
+type StorageAutoScalerSpec struct {
+	// StorageClusterName is the name of the storage cluster for which the storage scaling is to be done.
+	// +kubebuilder:validation:Required
+	StorageClusterName string `json:"storageClusterName,omitempty"`
+	// DeviceClassName is the name of the device class for which the storage scaling is to be done.
+	// +kubebuilder:validation:Required
+	DeviceClassName string `json:"deviceClassName,omitempty"`
```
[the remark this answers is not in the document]
parth-gr: Yaa maybe we can apply the default value as `ssd` if not set,

api/v1/autostoragescale.go
parth-gr: Will set a default value to ssd

api/v1/storage_autoscaler.go
```diff
@@ -0,0 +1,101 @@
...
+	// Used for horizontal scaling of OSDs.
+	ExpectedOsdCount uint16 `json:"expectedOsdCount,omitempty"`
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="Age",type=date,JSONPath=.metadata.creationTimestamp, description="Age"
+// +kubebuilder:printcolumn:name="Phase",type=string,JSONPath=.status.phase,description="Current Phase"
+// +kubebuilder:printcolumn:name="LastRunTimeStamp",type=date,JSONPath=.status.lastRunTimeStamp,description="Last Run Time Stamp"
+// +operator-sdk:csv:customresourcedefinitions:displayName="Auto Storage Scaling"
```
BlaineEXE, line 92: ?

```suggestion
// +operator-sdk:csv:customresourcedefinitions:displayName="Storage Auto Scaling"
```

api/v1/storage_autoscaler.go
```diff
@@ -0,0 +1,101 @@
...
+	// StorageClusterName is the name of the storage cluster for which the storage scaling is to be done.
+	// +kubebuilder:validation:Required
+	StorageCluster corev1.LocalObjectReference `json:"storageCluster,omitempty"`
+	// DeviceClassName is the name of the device class for which the storage scaling is to be done.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default="ssd"
+	DeviceClassName string `json:"deviceClassName,omitempty"`
+	// DeviceClassStorageCapacityLimit is the capacity limit for the storage scaling for the specific deviceClass and storagecluster.
+	// +kubebuilder:validation:Required
+	DeviceClassStorageCapacityLimit resource.Quantity `json:"deviceClassStorageCapacityLimit,omitempty"`
```
BlaineEXE, line 36:
```suggestion
	// DeviceClassStorageCapacityLimit is the total aggregate capacity limit for the storage scaling for the specific deviceClass and storagecluster.
	// +kubebuilder:validation:Required
	DeviceClassStorageCapacityLimit resource.Quantity `json:"deviceClassStorageCapacityLimit,omitempty"`
```

api/v1/storage_autoscaler.go
```diff
@@ -0,0 +1,101 @@
...
+	TimeoutSeconds int `json:"timeoutSeconds,omitempty"`
+}
+
+type StorageAutoScalerPhase string
+
+const (
+	StorageAutoScalerPhaseNotStarted StorageAutoScalerPhase = "NotStarted"
+	StorageAutoScalerPhaseInProgress StorageAutoScalerPhase = "InProgress"
+	StorageAutoScalerPhaseSucceeded  StorageAutoScalerPhase = "Succeeded"
+)
```
BlaineEXE, line 58:
```suggestion
const (
	StorageAutoScalerPhaseNotStarted StorageAutoScalerPhase = "NotStarted"
	StorageAutoScalerPhaseInProgress StorageAutoScalerPhase = "InProgress"
	StorageAutoScalerPhaseSucceeded  StorageAutoScalerPhase = "Succeeded"
	StorageAutoScalerPhaseFailed     StorageAutoScalerPhase = "Failed"
)
```

api/v1/storage_autoscaler.go
```diff
@@ -0,0 +1,101 @@
...
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default="ssd"
+	DeviceClassName string `json:"deviceClassName,omitempty"`
+	// DeviceClassStorageCapacityLimit is the capacity limit for the storage scaling for the specific deviceClass and storagecluster.
+	// +kubebuilder:validation:Required
+	DeviceClassStorageCapacityLimit resource.Quantity `json:"deviceClassStorageCapacityLimit,omitempty"`
+	// MaxOSDSize is the maximum size that OSD disk can be expanded to.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default="8Ti"
+	MaxOSDSize resource.Quantity `json:"maxOSDSize,omitempty"`
```
BlaineEXE, line 40: Consistency nit

```suggestion
	// MaxOsdSize is the maximum size that OSD disk can be expanded to.
	// +kubebuilder:validation:Optional
	// +kubebuilder:default="8Ti"
	MaxOsdSize resource.Quantity `json:"maxOsdSize,omitempty"`
```

config/crd/bases/ocs.openshift.io_autostoragescalings.yaml
```diff
@@ -0,0 +1,127 @@
...
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.16.1
+  name: autostoragescalings.ocs.openshift.io
+spec:
+  group: ocs.openshift.io
+  names:
+    kind: AutoStorageScaling
```
travisn, line 11: This file (and a few others) look like they need to be removed from the commit, generated from a previous type name.

api/v1/storage_autoscaler.go
```diff
@@ -0,0 +1,105 @@
...
+
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="Age",type=date,JSONPath=.metadata.creationTimestamp, description="Age"
+// +kubebuilder:printcolumn:name="Phase",type=string,JSONPath=.status.phase,description="Current Phase"
+// +kubebuilder:printcolumn:name="LastRunTimeStamp",type=date,JSONPath=.status.lastRunTimeStamp,description="Last Run Time Stamp"
+// +operator-sdk:csv:customresourcedefinitions:displayName="Storage Auto Scaling"
+
+// StorageAutoScaler represents the automatic storage scaling for storage cluster.
+type StorageAutoScaler struct {
```
travisn, line 99: IMO the type name would be better without the "r" suffix, in other words search and replace `StorageAutoScaler` --> `StorageAutoScale`. But it's not important to debate this detail, if you prefer the "r".
```suggestion
type StorageAutoScale struct {
```

config/crd/bases/ocs.openshift.io_autostoragescalings.yaml
parth-gr: thanks for catching it out

api/v1/storage_autoscaler.go
parth-gr: I think StorageAutoscaler is more generic in terms of  referring a tool,

And its not a hard change anyway to revert. but most of code its already been used like in the UI and alert team codebase, so better to keep this

travisn (commented):
/lgtm

api/v1/storage_autoscaler.go
```diff
@@ -0,0 +1,105 @@
...
+)
+
+// StorageAutoScalerStatus defines the observed state of StorageAutoScaler
+type StorageAutoScalerStatus struct {
+	// Phase describes the Phase of StorageAutoScaler
+	// +kubebuilder:validation:Enum=NotStarted;InProgress;Failed;Succeeded
+	Phase StorageAutoScalerPhase `json:"phase,omitempty"`
+	// Conditions represents the latest available observations of an object's state
+	// +kubebuilder:validation:Optional
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
```
BlaineEXE, line 68: I find that Conditions are hard to manage well from controllers without a really high level of attention to detail. I'd suggest not using these. And in general, I think we should be able to enumerate in this design doc each individual condition we are tracking and how it will be used. If there is some OCP/ODF requirement for them, it's fine, but again I'd caution that we should probably enumerate them here if we are going to manage conditions.

BlaineEXE: In my experience, `Phase` is usually enough to convey state to the user and doesn't have the complexity and bug risk that conditions carry. It might be best if we remove Conditions from the current plan and add them later if we decide they are needed.

parth-gr: As we have different `reasons` for failure, so in  that case `phase` wont be just sufficient,

I looked a little bit on how to add reasons with phases, I had seen we are using `conditions` but I would be fine in using any other way if recommend, maybe just a `message string`?

BlaineEXE: Conditions aren't supposed to be used as a dumping ground for errors. They are part of the API and should be carefully treated as such if they are used. Kubernetes Node conditions are a prime example of how Conditions are intended to be used -- they convey a persistent and common set of details about the node. 

Personally, I have found that explicitly-typed status items are much easier to manage in code (and deal with as a user) than conditions. Some developers/projects still really like them, however.

I suggested a while back to follow CSI Volume Snapshotter's design and use a timestamped error for tracking the latest reason for failure. This will be much easier for us to manage and will still provide clear information for users.

```go
type XyzStatus struct {
    /* ... */
    
    Error *TimestampedError
}

type TimestampedError struct {
    Time metav1.Time

    // NOTE: message may be logged, and it should not contain sensitive information.
    Message string 
}
```

api/v1/storage_autoscaler.go
```diff
@@ -0,0 +1,118 @@
...
+
+const (
+	StorageAutoScalerPhaseNotStarted StorageAutoScalerPhase = "NotStarted"
+	StorageAutoScalerPhaseInProgress StorageAutoScalerPhase = "InProgress"
+	StorageAutoScalerPhaseSucceeded  StorageAutoScalerPhase = "Succeeded"
+	StorageAutoScalerPhaseFailed     StorageAutoScalerPhase = "Failed"
+)
+
+// TimeStampedError is the error message with the time stamp.
+type TimeStampedError struct {
```
BlaineEXE, line 62: "timestamp" is one word. No need to capitalize the "S"

```suggestion
type TimestampedError struct {
```

api/v1/storage_autoscaler.go
```diff
@@ -0,0 +1,128 @@
...
+
+// StorageAutoScalerStatus defines the observed state of StorageAutoScaler
+type StorageAutoScalerStatus struct {
+	// Phase describes the Phase of StorageAutoScaler
+	// +kubebuilder:validation:Enum=NotStarted;InProgress;Failed;Succeeded
+	Phase StorageAutoScalerPhase `json:"phase,omitempty"`
+	// Error is the error message in case the storage scaling operation has failed.
+	// +optional
+	// +nullable
+	Error TimestampedError `json:"error,omitempty"`
```
BlaineEXE, line 77:
```suggestion
	// Error is the error message in case the storage scaling operation has failed.
	// +optional
	// +nullable
	Error *TimestampedError `json:"error,omitempty"`
```

parth-gr: as we set `// +nullable` we need a pointer,
	I am trying to understanding when we would be explicitly setting this to null

api/v1/storage_autoscaler.go
```diff
@@ -0,0 +1,128 @@
...
+	Message string `json:"message,omitempty"`
+	// Timestamp is the time stamp when the error occurred.
+	Timestamp metav1.Time `json:"timestamp,omitempty"`
+}
+
+// StorageAutoScalerStatus defines the observed state of StorageAutoScaler
+type StorageAutoScalerStatus struct {
+	// Phase describes the Phase of StorageAutoScaler
+	// +kubebuilder:validation:Enum=NotStarted;InProgress;Failed;Succeeded
+	Phase StorageAutoScalerPhase `json:"phase,omitempty"`
```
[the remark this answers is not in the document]
BlaineEXE: Actually, this enum should probably be removed. StorageClusterStatus doesn't have an enum for its phase.

https://github.com/red-hat-storage/ocs-operator/blob/be77c56795c42217cbd4c55b1d3b41039cf500be/api/v1/storagecluster_types.go#L545-L548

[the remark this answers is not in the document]
parth-gr: Then I think we dont need `optional`

api/v1/storage_autoscaler.go
```diff
@@ -0,0 +1,128 @@
...
+	// Error is the error message in case the storage scaling operation has failed.
+	// +optional
+	// +nullable
+	Error *TimestampedError `json:"error,omitempty"`
+	// LastExpansionStartTime is the time stamp of the last run start of the storage scaling
+	// +optional
+	LastExpansionStartTime metav1.Time `json:"lastExpansionStartTime,omitempty"`
+	// LastExpansionCompletionTime is the time stamp of the last run completion of the storage scaling
+	// +optional
+	LastExpansionCompletionTime metav1.Time `json:"lastExpansionCompletionTime,omitempty"`
```
BlaineEXE, line 83: I'd suggest asking @nb-ohad and @iamniting whether they think these time fields should be pointers/nullable so that the output doesn't include the epoch date-time when time is equal to 0.

```suggestion
	// LastExpansionStartTime is the time stamp of the last run start of the storage scaling
	// +optional
	// +nullable
	LastExpansionStartTime *metav1.Time `json:"lastExpansionStartTime,omitempty"`
	// LastExpansionCompletionTime is the time stamp of the last run completion of the storage scaling
	// +optional
	// +nullable
	LastExpansionCompletionTime *metav1.Time `json:"lastExpansionCompletionTime,omitempty"`
```

api/v1/storage_autoscaler.go
```diff
@@ -0,0 +1,128 @@
...
+	Message string `json:"message,omitempty"`
+	// Timestamp is the time stamp when the error occurred.
+	Timestamp metav1.Time `json:"timestamp,omitempty"`
+}
+
+// StorageAutoScalerStatus defines the observed state of StorageAutoScaler
+type StorageAutoScalerStatus struct {
+	// Phase describes the Phase of StorageAutoScaler
+	// +kubebuilder:validation:Enum="";'NotStarted;InProgress;Failed;Succeeded
+	Phase StorageAutoScalerPhase `json:"phase,omitempty"`
```
BlaineEXE, line 73: I think a previous note I added as a reply got missed. It'd be best to not enum the phase. This can only introduce errors creating the CRD, and I don't know of any resources that enum status items like this.

For comparison the StorageCluster status phase isn't enum-ed

https://github.com/red-hat-storage/ocs-operator/blob/be77c56795c42217cbd4c55b1d3b41039cf500be/api/v1/storagecluster_types.go#L545-L548

api/v1/storage_autoscaler.go
```diff
@@ -0,0 +1,128 @@
...
+	// +optional
+	ExpectedOsdCount uint16 `json:"expectedOsdCount,omitempty"`
+	// The start storage capacity is the original storage capacity of the OSDs before the expansion in progress is completed.
+	// After the expansion is completed, this would be updated to the expected storage capacity.
+	// +optional
+	StartStorageCapacity resource.Quantity `json:"startStorageCapacity,omitempty"`
+	// The Expected storage capacity is the storage capacity that the auto-expansion has decided to set.
+	// This will be set on the storageCluster CR as the desired storage capacity of the OSDs.
+	// +optional
+	ExpectedStorageCapacity resource.Quantity `json:"expectedStorageCapacity,omitempty"`
```
BlaineEXE, line 111: I'm curious what benefit this gives us over `StartOsdSize` and `ExpectedOsdSize`. I think those are the things that actually control whether expansion will continue or not right?

Doesn't this have room for calculation errors between Ceph's reported total capacity and the PV/PVC capacity?

api/v1/storage_autoscaler.go
parth-gr: why cant we use enum with `+ nullable`?

api/v1/storage_autoscaler.go
parth-gr: we have to use this values  the alerts, https://github.com/red-hat-storage/ocs-operator/pull/3045#discussion_r1985442114

api/v1/storage_autoscaler.go
BlaineEXE: It would be null by default when the status is instantiated. Thus it also wouldn't display anything in the json or yaml output. When time is `0`, that might show up as the epoch date, which is usually `Wednesday, December 31, 1969 5:00:00 PM GMT-07:00`. We don't want an irrelevant date showing up in output. Some users might see it and think there is an error that they need to contact support about.

api/v1/storage_autoscaler.go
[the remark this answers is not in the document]
BlaineEXE: Likely not. I don't think the validator counts `""` as unset

api/v1/storage_autoscaler.go
BlaineEXE: I'm not sure if we can use enum with nullable, since nullable implies a type that is a pointer. It might be valid or might not. But in general, I don't think enums in status items are very helpful. In my experience, they are unlikely to help and likely to cause a problem, so don't bother. They won't help users, since users don't set the status themselves. Also, they can result in the operator being unable to set otherwise valid statuses, resulting in unexpected runtime errors that we need to bug fix.

parth-gr: Actually I removed +nullable, and used a default value,

The main reason here is to use enum is that coming from alert team, as alert are based on the phase they had to make sure operator is setting up correctly

api/v1/storage_autoscaler.go
iamniting: It depends on the use case. If someone else is going to read the time, I would like to have it as a pointer so the other operator or code understands whether it is set or not yet.

parth-gr: I dont think so any one else is reading it, its for the ui

parth-gr: okay then having it  nullable  make sense, updated it to pointer

api/v1/storage_autoscaler.go
parth-gr: Updated

api/v1/storage_autoscaler.go
```diff
@@ -0,0 +1,137 @@
...
+	// Message is the error message in case the storage scaling operation has failed.
+	Message string `json:"message,omitempty"`
+	// Timestamp is the time stamp when the error occurred.
+	Timestamp metav1.Time `json:"timestamp,omitempty"`
+}
+
+// StorageAutoScalerStatus defines the observed state of StorageAutoScaler
+type StorageAutoScalerStatus struct {
+	// Phase describes the Phase of StorageAutoScaler
+	// +kubebuilder:default="NotStarted"
```
BlaineEXE, line 72: I'm not sure about a default here. Personally, I think `""` is a useful status for indicating when the controller hasn't acknowledged the resource yet. This is probably not huring anything, but it does subtly change an shared design pattern.

BlaineEXE (approved):
Approving, with a suggestion not to default the phase

api/v1/storage_autoscaler.go
parth-gr: As it is against the pattern and it should be ack by the controller, removed the default,

Made the status.spec optional, hope it is fine?

api/v1/storage_autoscaler.go
```diff
@@ -0,0 +1,137 @@
...
+	StorageCapacityLimitReached bool `json:"storageCapacityLimitReached,omitempty"`
+	// StartStorageCapacity is the original storage capacity of the storage cluster before the expansion in progress is completed.
+	StartStorageCapacity resource.Quantity `json:"startStorageCapacity,omitempty"`
+	// ExpectedStorageCapacity is the expected storage capacity of the storage cluster after the expansion in progress is completed.
+	ExpectedStorageCapacity resource.Quantity `json:"expectedStorageCapacity,omitempty"`
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="Age",type=date,JSONPath=.metadata.creationTimestamp, description="Age"
```
travisn, line 116: How about a printcolumn for the deviceclass?

BlaineEXE (approved):

api/v1/storage_autoscaler.go
```diff
@@ -0,0 +1,137 @@
...
+	LastExpansionStartTime *metav1.Time `json:"lastExpansionStartTime,omitempty"`
+	// LastExpansionCompletionTime is the time stamp of the last run completion of the storage scaling
+	// +optional
+	// +nullable
+	LastExpansionCompletionTime *metav1.Time `json:"lastExpansionCompletionTime,omitempty"`
+	// The start OSD size is the original size of the OSDs before the expansion in progress is completed.
+	// After the expansion is completed, this would be updated to the expected OSD size.
+	// Used for vertical scaling of OSDs.
+	// +optional
+	StartOsdSize resource.Quantity `json:"startOsdSize,omitempty"`
```
travisn, line 90: You'll need to make all of these status fields nullable for the same reason as the last expansion times. If the phase is `NotStarted`, all of these non-nullable status fields will have default values, which will be confusing. 

An alternative approach would be to change move all the status fields under a `lastExpansion` field, and put them all in a struct. Then only the struct would need to be nullable, such as this:
```go
type StorageAutoScalerStatus struct {
	Phase StorageAutoScalerPhase `json:"phase,omitempty"`
	Error *TimestampedError `json:"error,omitempty"`
	LastExpansion *LastExpansionStatus `json:"lastExpansion,omitempty"`
}

type LastExpansionStatus struct {
	StartTime metav1.Time `json:"startTime,omitempty"`
	CompletionTime metav1.Time `json:"completionTime,omitempty"`
	StartOsdSize resource.Quantity `json:"startOsdSize,omitempty"`
	...
}
```

parth-gr: So in this case we don't need individual `metav1.Time` as pointer so removed pointers on individual level

api/v1/storage_autoscaler.go
```diff
@@ -0,0 +1,142 @@
...
+	// Message is the error message in case the storage scaling operation has failed.
+	Message string `json:"message,omitempty"`
+	// Timestamp is the time stamp when the error occurred.
+	Timestamp metav1.Time `json:"timestamp,omitempty"`
+}
+
+type LastExpansionStatus struct {
+	// LastExpansionStartTime is the time stamp of the last run start of the storage scaling
+	// +optional
+	LastExpansionStartTime metav1.Time `json:"lastExpansionStartTime,omitempty"`
```
travisn, line 72: No need for the "lastExpansion" prefix on the fields since the parent struct has the name
```suggestion
	StartTime metav1.Time `json:"startTime,omitempty"`
```

api/v1/storage_autoscaler.go
```diff
@@ -0,0 +1,142 @@
...
+	// LastExpansionCompletionTime is the time stamp of the last run completion of the storage scaling
+	// +optional
+	LastExpansionCompletionTime metav1.Time `json:"lastExpansionCompletionTime,omitempty"`
```
travisn, line 75:
```suggestion
	CompletionTime metav1.Time `json:"completionTime,omitempty"`
```

travisn (commented):
Looks great with the `lastExpansion` struct, will be nice not to have so many individual status fields nullable.

api/v1/storage_autoscaler.go
parth-gr: correct

api/v1/storage_autoscaler.go
```diff
@@ -0,0 +1,142 @@
...
+	// Used for horizontal scaling of OSDs.
+	// +optional
+	StartOsdCount uint16 `json:"startOsdCount,omitempty"`
+	// The Expected OSD count is the count that the auto-expansion has decided to set.
+	// This will be set on the storageCluster CR as the desired count of the OSDs.
+	// Used for horizontal scaling of OSDs.
+	// +optional
+	ExpectedOsdCount uint16 `json:"expectedOsdCount,omitempty"`
+	// StorageCapacityLimitReached is the flag that indicates if the storage capacity limit has been reached.
+	StorageCapacityLimitReached bool `json:"storageCapacityLimitReached,omitempty"`
```
travisn, line 97: Seems like this bool should go on the `StorageAutoScalerStatus` struct, rather than the last expansion status struct

parth-gr:
@travisn it should be /lgtm?

PROJECT
```diff
@@ -25,6 +25,15 @@ resources:
...
   path: github.com/red-hat-storage/ocs-operator/api/v4/v1
   version: v1
+- api:
+    crdVersion: v1
+    namespaced: true
+  controller: true
+  domain: openshift.io
+  group: ocs
+  kind: StorageAutoScaler
+  path: github.com/red-hat-storage/ocs-operator/api/v4/v1
```
alfonsomthd, line 35: @parth-gr Just to confirm: is `/api/v4` intended?

parth-gr: yaa as the repo is github.com/red-hat-storage/ocs-operator/v4

iamniting:
> > @travisn it should be /lgtm?
> 
> I'd like to give @Madhu-1 @iamniting a chance for final review before it merges

I had a mtg with @parth-gr and we made the changes as required.

iamniting (approved):

PROJECT
iamniting: This is because it is defined this way in go.mod

docs/design/storage-auto-scale.md
```diff
@@ -32,13 +32,17 @@ Implement auto storage scale CR per device class of a specific Storagecluster.
         phase: (NotStarted | InProgress | Succeeded | Failed)
         error:
             message: ""
-        lastExpansionStartTime: (always set after a new expansion is started)
-        lastExpansionCompletionTime: (only set on success)
```
alfonsomthd, line 36: @parth-gr shouldn't we also set it on `Failed` so the user knows when the failure occured?

PR #3045, opened 2025-02-18
autoscale: add a design doc for auto storage scaling

parth-gr:
added a design doc for auto storage scaling

### DIFF
--- a/docs/design/storage-auto-scale.md
+++ b/docs/design/storage-auto-scale.md
@@ -1,0 +1,128 @@
+# Storage Automatic Scaler
+
+## Summary
+
+This feature would help users to get rid of maintenance task of keeping the system with enough available free space.
+In virtualized environments, we can scale the storage devices Osds so this feature helps in automatic scaling. 
+
+## Proposal
+
+Implement auto storage scale CR per device class of a specific Storagecluster.
+
+## High-Level Design
+
+- Create a StorageAutoScale CR per Storagecluster device class.
+- UI will create the StorageAutoScale CR with a unique name i.e. <storageclusterName>-<deviceClassName>.
+  
+    ```
+    apiVersion: ocs.openshift.io/v1
+    kind: StorageAutoScaler
+    metadata:
+        name: <name>
+        namespace: openshift-storage
+    spec:
+        storageCluster: 
+            name: <storageclusterName>
+        deviceClass: ssd
+        storageCapacityLimit: <capacity>
+        storageScalingThresholdPercent: 70  # should be less than osd nearfull percentage 
+        maxOsdSize: 8Ti
+        timeoutSeconds: 1800   
+    status:
+        phase: (NotStarted | InProgress | Succeeded | Failed)
+        error:
+            message: ""
+        lastExpansionStartTime: (always set after a new expansion is started)
+        lastExpansionCompletionTime: (only set on success)
+        startOsdSize: 2TiB
+        expectedOsdSize: 4TiB
+        startOsdCount: 3​
+        expectedOsdCount: 3
+        storageCapacityLimitReached: false
+    ```
+
+- Watch prometheus periodically every 10 mins and depending on the Osd fill percentage, reconcile the StorageAutoScaling CR.
+  
+- If the Osd size is less than maxOsdSize(default:8Tib), do vertical scaling by doubling the each osd sizes for that device class.
+
+- If the Osd size is equal to maxOsdSize(default:8Tib), do a horizontal scaling, by adding 1 osd of 8Tib on each storage node, by simply increasing the (count+1) in the device set associated with a device class.   
+
+## Low-Level Design 
+
+Go-routine Scraper:
+
+1) Add a single go-routine to watch periodically(10mins) all the OSDs sizes using Prometheus metrics in the storagecluster namespace.
+    Query: (ceph_osd_metadata * on (ceph_daemon, namespace, managedBy) group_right(device_class,hostname) (ceph_osd_stat_bytes_used / ceph_osd_stat_bytes))
+
+2) With every run calculate different device class highest osd filled percentage.
+   
+        OsdPercentage:
+        
+        ssd               69%
+        nvme              71%
+        ...   
+
+3) Create a sync map with `OsdPercentage`, Send a event to the go channel.
+
+Controller:
+
+1) Create a named controller that watches for [channel generic event](https://book-v1.book.kubebuilder.io/beyond_basics/controller_watches) per device class.
+
+2) If the expansion is not in progress set the status `phase` to `NotStarted` 
+   
+3) If an expansion is in progress(expectedOsdSize!=startOsdSize || expectedOsdCount!=startOsdCount), check the progress and then requeue each 1 minute until the expansion is completed successfully(jump to step 10).
+   
+4) If the highest osd percentage reported in the sync map is more than osdScalingThresholdPercent(70%) means reaching osd nearfull, scaling is needed.
+   
+5) If scaling is needed calculate the `expectedOsdSize` and `expectedOsdCount`.
+       
+    1)  If the Osd size is less than maxOsdSize(default:8Tib), do vertical scaling by doubling the each osd sizes for that device class.
+            
+    2)  If the Osd size is equal to maxOsdSize(default:8Tib), do a horizontal scaling, by adding 1 osd of maxOsdSize(default:8Tib) on each `storageDeviceSet`.
+   
+6) Calculate the `expectedStorageCapacity` based on expected size and count.
+   
+7) Check if the `storageCapacityLimit` > `expectedStorageCapacity`.
+        
+    1) If yes, Update `phase` to `InProgress`  and `lastExpansionStartTime` as `current-time` on the `StorageAutoScaling` CR which need scaling.
+        
+    2) If no, don't reconcile further and the set the `storageCapacityLimitReached` as true in the status.
+
+8)  Update the status, set the `expectedOsdSize` and `expectedOsdCount` to reflect the new expected value and also set `startOsdSize` and `startOsdCount` with current storagecluster values.
+   
+9)  Scale by patching the Storagecluster, with all the device sets update needed at the same time.
+
+10) Verify and Alert:
+   
+   1) Verify the Storagecluster whether the new osds are added or scaled in size, for all the device sets.
+ 
+   2) If the scaling is successful will update the status of the `StorageAutoScaling` CR with `lastExpansionCompletionTime` and `phase` and also osd count and size.
+   
+   3) If the auto scale is not completed, it will do a requeue every 1 min and, change the phase to `failed` if scaling not `Succeeded` with in timeoutSeconds(default:1800) interval.
+    
+   4) Alert the user if the phase changes to `Succeeded` or `Failed`, alerting will be implemented with ocs-metrics-exporter. 
+
+### InProgress state:
+
+Based on the above algorithm there would be two conditions where in-progress is set, elaborating those conditions,
+
+1) If scaling is just started:
+    1) Set `phase` to `InProgress`.
+    2) Verify is the scaling is successful. 
+    3) If the scaling is successful set the `phase` to `Succeeded`.
+    4) Alert the user if the phase changes to `Succeeded`, alerting will be implemented with ocs-metrics-exporter.
+    5) If the scaling is not yet completed requeue every 1 min, we have the 2nd case. 
+
+2) If the scaling has already started and its requeue
+    1) Now the requeue will happen every 1 min.
+    2) At the start of reconcile will match that `startOsdSize` and `expectedOsdSize` is not equal and similar for osd count.
+    3) And another validation will do is equating storagecluster spec with prometheus response.
+    4) Will requeue till the scaling is in-progress.
+    5) If the scaling is in-progress with more than timeoutSeconds(default:1800) interval we set the phase to `failed`.
+    6) Alert the user if the phase changes to `Failed`, alerting will be implemented with ocs-metrics-exporter. 
+    7) If there as a failure alert, provide a mitigation guide for the user. 
+
+## Failure case
+
+There might cases where system resources are exhausted and new Osds cannot be started.
+


docs/design/auto-storage-scale.md
```diff
@@ -0,0 +1,49 @@
...
+- Create a AutoStorageScale CR per Storagecluster device class
+    ```
+    apiVersion: ocs.openshift.io/v1
+    Kind: AutoStorageScaling
+    Metadata:
+	    Name: <storageclusterName>-<deviceClassName>
+	    namespace: <namespace>
+    Spec: 
+	    storageClusterName: <storageclusterName>
+        deviceClassName: <deviceClassName>
```
travisn, line 23: nit: indentation issue

docs/design/auto-storage-scale.md
```diff
@@ -0,0 +1,49 @@
...
+
+- Create a AutoStorageScale CR per Storagecluster device class
+    ```
+    apiVersion: ocs.openshift.io/v1
+    Kind: AutoStorageScaling
+    Metadata:
+	    Name: <storageclusterName>-<deviceClassName>
+	    namespace: <namespace>
+    Spec: 
+	    storageClusterName: <storageclusterName>
```
travisn, line 22: Only one storage cluster CR is supported, right? If so, we don't need the name

docs/design/auto-storage-scale.md
```diff
@@ -0,0 +1,49 @@
...
+Implement auto storage scale CR per device class of a specific Storagecluster.
+
+# High-Level Design
+
+- Create a AutoStorageScale CR per Storagecluster device class
+    ```
+    apiVersion: ocs.openshift.io/v1
+    Kind: AutoStorageScaling
+    Metadata:
+	    Name: <storageclusterName>-<deviceClassName>
```
travisn, line 19: The name is not important, right? It can be anything. The spec fields would have the important settings.

docs/design/auto-storage-scale.md
```diff
@@ -0,0 +1,49 @@
...
+	    namespace: <namespace>
```
travisn, line 20:
```suggestion
	    namespace: openshift-storage
```

docs/design/auto-storage-scale.md
```diff
@@ -0,0 +1,49 @@
...
+        expectedOSDSize: 4TiB
+        knownOSDCount: 3​ 
+        expectedOSDCount: 3
+    ```
+
+-  Watch prometheus periodically after 10 mins and depending on the OSD fill percentage, recocnile the CR. 
+
+# Low-Level Design 
+
+1) Add a go-routine to watch for Prometheus metrics
```
travisn, line 38: Is there a new go routine for every `AutoStorageScaling` CR? Or one go routine that iterates over all of the CRs? Commonly there is only one CR anyway, but we need to be clear in the design. I'm thinking one go-routine to handle all of them will be best.

docs/design/auto-storage-scale.md
```diff
@@ -0,0 +1,49 @@
...
+2) Go routine updates the channel and the sync-map with osd percentage
```
travisn, line 39: What channel and sync-map is this referring to? I thought it would just update the status on the AutoStorageScaling CR?

docs/design/auto-storage-scale.md
```diff
@@ -0,0 +1,49 @@
...
+and also evaluating flag whether resizing is needed
+3) Create a named controller that watcher for channel generic event
```
travisn, line 41: This is the controller to watch for AutoStorageScaling CRs?

docs/design/auto-storage-scale.md
```diff
@@ -0,0 +1,49 @@
...
+4) If there is an update on the channel reconcile the controller 
+5) Controller will patch the storagecluster with all the device class device set update needed at the same time
```
travisn, line 43: Doesn't the go-routine patch the storagecluster CR when it determines the scale is needed? I'm not understanding the difference between the go-routine and the controller.

docs/design/auto-storage-scale.md
```diff
@@ -0,0 +1,49 @@
...
+2) Go routine updates the channel and the sync-map with osd percentage
+and also evaluating flag whether resizing is needed
+3) Create a named controller that watcher for channel generic event
+4) If there is an update on the channel reconcile the controller 
+5) Controller will patch the storagecluster with all the device class device set update needed at the same time 
+6) Sync map will act as the desired state and source of truth for the controller.
+
+# Risk
+
+There might cases where system reaches out of memory.
```
travisn, line 48: Do you mean the OSDs will stay in pending state because there is not enough CPU/memory during horizontal scale?

docs/design/auto-storage-scale.md
alfonsomthd: @travisn From my previous conversations with @parth-gr, `AutoStorageScale` CR will be associated to a `StorageCluster` and a deviceclass, so it's required as the UI needs to retrieve the correct CR looking filtering by storage cluster name and deviceclass.

docs/design/auto-storage-scale.md
alfonsomthd: @travisn this is a suggestion I left in the Design doc just to identify at-a-glance the storagecluster and deviceclass which this CR is associated to.

docs/design/auto-storage-scale.md
parth-gr: Yaa lets add the support, even not needed right now, to make it generic

docs/design/auto-storage-scale.md
parth-gr: Correct name is not important, it has to be unique

docs/design/auto-storage-scale.md
parth-gr: yes there a single go-routine running for all the CR's

docs/design/auto-storage-scale.md
parth-gr: We need a map that will act as a source of truth instead of status

docs/design/auto-storage-scale.md
parth-gr: Right, in that case we will report failure

docs/design/auto-storage-scale.md
travisn: @alfonsomthd Sounds good to add the storageClusterName just to be clear, but does ODF even support multiple storageCluster CRs? I can't think where that would be needed, except perhaps when there is both an external and internal cluster, but even those are in different namespaces.

docs/design/auto-storage-scale.md
travisn: Sounds good as long as it's just a convention, and we aren't using that to extract details

travisn (changes requested):

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,94 @@
...
+   6) If the scaling is successful will raise a `success` alert and update the status of all the `StorageAutoScaling` CR.
+   7) If it is fails will do a re-reconcile and, if 3 recent reconcile fails(track by observed generation) will raise a `failure` alert.   
+ 
+5) Sync map will act as the desired state and source of truth for the controller.
+
+Note:
+    The above prometheus query work per namespace, so the algorithm will be per storagecluster operation, later in future if we need the auto scaling for different storagecluster, either run multiple go-routine and multiple sync-maps, 
+    Or update the current prometheus query and sync-map to accommodate different storagecluster values.
+
+# Risk
```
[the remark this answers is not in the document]
parth-gr: just  failure case

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,94 @@
...
+            expectedOSDSize: 4
+            expectedOSDCount: 9
+        ...
+        ```
+
+2) If the sync map is created or updated then go routine will send a signal to the go channel.
+
+3) Create a named controller that watches for [channel generic event](https://book-v1.book.kubebuilder.io/beyond_basics/controller_watches).
+
+4) If there is an update on the channel, watcher will reconcile the new named controller.
```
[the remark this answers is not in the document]
travisn: To follow up from our discussion, another pattern could be to keep a single goroutine approach with the following:
1. If there is a scaling in progress, every 1 minute check the status
2. Every 10 minutes check if scaling is needed. 

The point is that we don't need to sync across multiple goroutines, it keeps things simpler if we can approach with a single goroutine.

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,94 @@
...
+   1) Named Controller will list all the StorageAutoScaling CRs.
+   2) Depending on the `OsdPercentage sync-map` it will set the `expectedOSDSize` and `expectedOSDCount` and on the status and update `phase` to `in-progress` on the `StorageAutoScaling` CR which need scaling.
```
BlaineEXE, line 78:
```suggestion
   2) Depending on the `OsdPercentage sync-map` it will set the `expectedOSDSize` and `expectedOSDCount` and on the status and update `phase` to `InProgress` on the `StorageAutoScaling` CR which need scaling.
```

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,94 @@
...
+
+- Watch prometheus periodically after 10 mins and depending on the OSD fill percentage, reconcile the StorageAutoScaling CR.
+  
+- If the OSD size is less than 8Tib, do vertical scaling by doubling the each osd sizes for that device class.
+
+- If the OSD size is equal to 8Tib, do a horizontal scaling, by adding 1 osd of 8Tib on each storage node, by simply increasing the (count+1) in the device set associated with a device class.   
+
+# Low-Level Design 
+
+1) Add a single go-routine to watch periodically(10mins) all the OSDs sizes using Prometheus metrics in the storagecluster namespace.
```
BlaineEXE, line 42: Alternative suggestion: don't use a goroutine at all. Instead, use the controller-runtime `requeueAfter` feature to re-run the reconcile loop every 10 minutes.

https://github.com/kubernetes-sigs/controller-runtime/blob/6f611112490242f9e8d5b0ccdc0f3bb8c86c85ef/pkg/reconcile/reconcile.go#L30-L37

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,94 @@
...
+        ```
+        ssd               69%
+        nvme               71%
+        ...
+        ```
+    2) If the highest osd percentage reported is more than 73-74%(reaching osd nearfull), scaling is needed.
+    3) If scaling is needed calculate the `expectedOSDSize` and `expectedOSDCount` per device class.
+       1)  Calculate the total number osds and osd size per device class from the prometheus response.
+       2)  Calculate the total number osds and osd size per device class from the storageCluster.
+       3)  If the above values are equal(needed to avoid double re-sizing), create or update the sync-map mapping:
```
BlaineEXE, line 54: `Go routine would update the sync-map` is the step below. Isn't this redundant?

BlaineEXE, line 54:
```suggestion
    3) If scaling is needed calculate the `expectedOSDSize` and `expectedOSDCount` per device class.
       1)  Calculate the total number osds and osd size per device class from the prometheus response.
       2)  Calculate the total number osds and osd size per device class from the storageCluster.
       3)  Update the StorageAutoScaling status to reflect the new expected value. This allows the controller to start (or resume) autoscaling in the event of controller failures after this point
```

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,94 @@
...
+            expectedOSDSize: 4
+            expectedOSDCount: 9
+        ...
+        ```
+
+2) If the sync map is created or updated then go routine will send a signal to the go channel.
+
+3) Create a named controller that watches for [channel generic event](https://book-v1.book.kubebuilder.io/beyond_basics/controller_watches).
+
+4) If there is an update on the channel, watcher will reconcile the new named controller.
```
BlaineEXE, line 76: This logic is complicated (read: bug risk) compared to just using controller-runtime's `requeueAfter` option

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,94 @@
...
+   1) Named Controller will list all the StorageAutoScaling CRs.
+   2) Depending on the `OsdPercentage sync-map` it will set the `expectedOSDSize` and `expectedOSDCount` and on the status and update `phase` to `in-progress` on the `StorageAutoScaling` CR which need scaling.
+   3) Check if the totalCapacity allows to scale it or not.
+   4) Patch the Storagecluster, with all the device class device sets update needed at the same time.
+   5) Verify the Storagecluster whether the new osds are added or scaled in size, for all the device class device sets. 
+   6) If the scaling is successful will raise a `success` alert and update the status of all the `StorageAutoScaling` CR.
+   7) If it is fails will do a re-reconcile and, if 3 recent reconcile fails(track by observed generation) will raise a `failure` alert.   
+ 
+5) Sync map will act as the desired state and source of truth for the controller.
```
BlaineEXE, line 85: This statement is a little concerning. The desired state source of truth should be the CRD spec and status. The map is just an internal mechanism to keep track of info. I would suggest not keeping the map and instead rebuilding it with each reconcile (we will need to look at prometheus metrics each reconcile no matter what anyway). That will help prevent against bugs that result from caching stale data that never gets corrected.

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,94 @@
+# Storage Automatic Scaler
+
+# Summary
```
BlaineEXE, line 3: Recommendation: Don't use L1 headers for everything. In this doc, we can use L2 headers (`##`) for everything but the doc title. 

```suggestion
## Summary
```

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,94 @@
...
+
+1) Add a single go-routine to watch periodically(10mins) all the OSDs sizes using Prometheus metrics in the storagecluster namespace.
+   1) With every run calculate different device class highest osd filled percentage.
+       `OsdPercentage`:
+        ```
+        ssd               69%
+        nvme               71%
+        ...
+        ```
+    2) If the highest osd percentage reported is more than 73-74%(reaching osd nearfull), scaling is needed.
```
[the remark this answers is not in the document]
travisn: How about if we make this a setting in the new CRD? If not set, how about a default of 70%, just before hitting the 75% nearfull alert.

docs/design/storage-auto-scale.md
travisn: Good idea. With a new controller for the new CRD, and it could just constantly requeue every 10 minutes when not expanding, and every 1 minute to check status when expansion is in progress.

docs/design/auto-storage-scale.md
parth-gr: https://github.com/red-hat-storage/ocs-operator/pull/3044?new_mergebox=false#discussion_r1967449209

docs/design/storage-auto-scale.md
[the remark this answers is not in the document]
parth-gr: Closing in preference of https://github.com/red-hat-storage/ocs-operator/pull/3045#discussion_r1962065170

docs/design/storage-auto-scale.md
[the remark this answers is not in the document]
parth-gr: I don't think so it is a direct user setting, and this should be governed/set by user.
Can come under advanced settings, but doesn't seem a good idea to expose to user, as we are thinking on per device class CRD, so user might configure different percentages for different cr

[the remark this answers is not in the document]
parth-gr: Will keep the hardcode default as 70%

docs/design/storage-auto-scale.md
parth-gr: That's a good idea,

But with this approach we need to query Prometheus with every CR,
And we cant use the same approach we were doing to query the prometheus once for all the cr's

docs/design/storage-auto-scale.md
parth-gr: correct, merge it

parth-gr: we need to control the events of double re-sizing, that's why we need the above check, lets discuss

docs/design/storage-auto-scale.md
parth-gr: Lets discuss again the only con with requeue is,

> We need to run Prometheus query with every cr
> If multiple device class need update, we need to update storagecluster multiple times

docs/design/storage-auto-scale.md
parth-gr: Yes we would be rebuilding the map every 10 min, 

But not on the cases where we updated the storagecluster and the storagcluster is in pending state to resize the storage

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,95 @@
...
+Implement auto storage scale CR per device class of a specific Storagecluster.
+
+## High-Level Design
+
+- Create a StorageAutoScale CR per Storagecluster device class
+    ```
+    apiVersion: ocs.openshift.io/v1
+    Kind: StorageAutoScaling
+    Metadata:
+        name: <storageclusterName>-<deviceClassName>
```
nb-ohad, line 19: As a user facing API,  you should not dictate rules the name the user gives to the CR. I would suggest leaving the name be and have the spec point to the storage cluster and device class names

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,95 @@
...
+        namespace: openshift-storage
+    Spec:
+        storageClusterName: <storageclusterName>
```
nb-ohad, line 22: Please use a LocalObjectReference structure for relationships within a single namespace

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,95 @@
...
+        deviceClassName: <deviceClassName>
+        capacityLimit: <capacity>
```
nb-ohad, line 24: Capacity is a loaded word. It might mean available capacity provided by the device set, or storage capacity of the ODS. I would prefer us using a more specific term.

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,95 @@
...
+        maxOSDSize: 8Ti
```
nb-ohad, line 25: Should we also support an upper limit on OSD count?

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,95 @@
...
+        maxScalingRetry: 3   
+    Status:
+        phase: (Waiting/InProgress/Completed/Failed)
```
nb-ohad, line 28: What would `waiting` mean in this context?
Same question for `in progress`?  

Also consider the fact that completed is never completed because utilization changes all of the time.

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,95 @@
...
+        LastRunTimeStamp: (only on completed/failure)
```
nb-ohad, line 29: Timestamp is a single word
```suggestion
        LastRunTimestamp: (only on completed/failure)
```

docs/design/storage-auto-scale.md
nb-ohad: We should not do a `requestAfter` and run the entire reconcile just for a querying prom. It will basically make this controller into an infinite loop which is a bad practice. The only thing that needs iteration is the querying of prometheus and  my point of view is that we should isolate this as its own thing that runs in a loop and awakens the reconciler whenever it detect a change

docs/design/storage-auto-scale.md
nb-ohad: Having a controller run in an infinite loop has its own complications. I would like us to avoid that

docs/design/storage-auto-scale.md
nb-ohad: @BlaineEXE There are 2 sources of truth here, each one is of its own domain (one is configuration, the other one is that state of the system) and we need to trigger the reconcile whenever either one changes.

nb-ohad (changes requested):

docs/design/storage-auto-scale.md
BlaineEXE: Agreed. This could be total usable Ceph capacity, aggregate raw disk capacity, or per-disk capacity. We should make sure our API makes it clear what scenario it is.

docs/design/storage-auto-scale.md
BlaineEXE, line 25: The latest best-practice I'm aware of is to use CamelCase even for acronyms like OSD.

```suggestion
        maxOsdSize: 8Ti
```

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,95 @@
...
+    Kind: StorageAutoScaling
+    Metadata:
+        name: <storageclusterName>-<deviceClassName>
+        namespace: openshift-storage
+    Spec:
+        storageClusterName: <storageclusterName>
+        deviceClassName: <deviceClassName>
+        capacityLimit: <capacity>
+        maxOSDSize: 8Ti
+        maxScalingRetry: 3
```
BlaineEXE, line 26:
```suggestion
        maxScalingRetryCount: 3   
```

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,95 @@
...
+        capacityLimit: <capacity>
+        maxOSDSize: 8Ti
+        maxScalingRetry: 3   
+    Status:
+        phase: (Waiting/InProgress/Completed/Failed)
+        LastRunTimeStamp: (only on completed/failure)
+        currentOSDSize: 2TiB
+        expectedOSDSize: 4TiB
+        currentOSDCount: 3​
+        expectedOSDCount: 3
```
BlaineEXE, line 33:
```suggestion
        currentOsdSize: 2TiB
        expectedOsdSize: 4TiB
        currentOsdCount: 3​
        expectedOsdCount: 3
```

docs/design/storage-auto-scale.md
BlaineEXE: We are going to reconcile in an infinite loop with 10 minute retry period regardless of which implementation is used. I don't see why we should pick the more complicated implementation.

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,95 @@
...
+
+3) Create a named controller that watches for [channel generic event](https://book-v1.book.kubebuilder.io/beyond_basics/controller_watches).
+
+4) If there is an update on the channel, watcher will reconcile the new named controller.
+   1) Named Controller will list all the StorageAutoScaling CRs.
+   2) Depending on the `OsdPercentage sync-map` it will set the `expectedOSDSize` and `expectedOSDCount` and on the status and update `phase` to `InProgress` on the `StorageAutoScaling` CR which need scaling.
+   3) Check if the totalCapacity allows to scale it or not.
+   4) Patch the Storagecluster, with all the device class device sets update needed at the same time.
+   5) Verify the Storagecluster whether the new osds are added or scaled in size, for all the device class device sets. 
+   6) If the scaling is successful will raise a `success` alert and update the status of all the `StorageAutoScaling` CR.
```
BlaineEXE, line 83:
```suggestion
   6) If the scaling is successful will raise a `success` alert and update the status of the `StorageAutoScaling` CR.
```

docs/design/storage-auto-scale.md
travisn: This is inferred by the capacity limit. It may be confusing to have both of those settings since they may conflict. I thought the discussion in the design review was that customers understand max capacity better than osd counts.

docs/design/storage-auto-scale.md
travisn: Agreed, we need clarity about the status. I also think we need to keep status for every expansion, not just the latest. What about something like this?
```yaml
status:
  latestExpansion:
    phase: NotStarted | InProgress | Failed | Succeeded
    startTime: <timestamp>
    startOsdSize: 8TiB
    desiredOsdSize: 8TiB
    startOsdCount: 3
    desiredOsdCount: 6
  completedExpansions:
    - startTime: <timestamp>
      completionTime: <timestamp> 
      startOsdSize: 4TiB
      endOsdSize: 8TiB
      startOsdCount: 3
      endOsdCount: 3
```
If an expansion has never been started, the phase would be `notStarted`. After an expansion is started, the phase would always be one of the others (inprogress, failed, or succeeded).

docs/design/storage-auto-scale.md
travisn: Once design is updated from the latest discussions, I don't think we need a syncmap, as no async communication is needed. Each 10 min, the reconcile just rebuilds the latest state.

docs/design/storage-auto-scale.md
parth-gr: Let's not complicate the design, as we are achieving the same goal from both perspectives.

docs/design/storage-auto-scale.md
parth-gr: We already have spec for it, we need a systematic unique naming of CRs so its just for that

docs/design/storage-auto-scale.md
parth-gr: Correct we just want to have a single user facing spec that is clusterCapacityLImit, 
And maxOSD is for internal logic, and not directly user facing in the UI

docs/design/storage-auto-scale.md
parth-gr: As UI would be reading it we don't want a lot of complication here,
As per what we had decided and with the above suggestion
We can make it like, 

```
        phase: (NotStarted | InProgress | Failed | Succeeded)
        LastRunTimeStamp: (only on completed/failure)
        currentOSDSize: 2TiB
        expectedOSDSize: 4TiB
        currentOSDCount: 3​
        expectedOSDCount: 3
```

`LastRunTimeStamp` would be sufficient to say when the last re-sized happen

docs/design/storage-auto-scale.md
parth-gr: Updated to use `clusterCapacityLimit`

docs/design/storage-auto-scale.md
parth-gr: moved to timeoutSecond instead of counting number of  re-tries

docs/design/storage-auto-scale.md
BlaineEXE: IMO, the phases that really matter are `InProgress` and `Succeeded`. If there is a failure, we will try to reconcile again, so the `Failed` condition would only exist until `InProgress` is set again. `NotStarted` is mostly equivalent to an empty phase, unless we have a specific need for it.

In lieu of an error phase, the Volume Snapshot system uses an error field in the status to record errors with timestamp. I think this is a good way of indicating errors without using a phase for it with unclear lifetime.

```
Status {

	// ...

	// error is the last observed error during snapshot creation, if any.
	// Upon success after retry, this error field will be cleared.
	// +optional
	Error *VolumeSnapshotError `json:"error,omitempty" protobuf:"bytes,5,opt,name=error,casttype=VolumeSnapshotError"
		
	// ...
}


// VolumeSnapshotError describes an error encountered during snapshot creation.
type VolumeSnapshotError struct {
	// time is the timestamp when the error was encountered.
	// +optional
	Time *metav1.Time `json:"time,omitempty" protobuf:"bytes,1,opt,name=time"`

	// message is a string detailing the encountered error during snapshot
	// creation if specified.
	// NOTE: message may be logged, and it should not contain sensitive
	// information.
	// +optional
	Message *string `json:"message,omitempty" protobuf:"bytes,2,opt,name=message"`
}
```

alfonsomthd: > What would `waiting` mean in this context? Same question for `in progress`?
> 
> Also consider the fact that completed is never completed because utilization changes all of the time.

@nb-ohad @BlaineEXE just to give more context: as per the UX designs, we need to inform the user about the current status:
- `Waiting` (or `NotStarted`, `Ready` or `whatever-we-choose`): Autoscaling is enabled but no expansion has been triggered yet.
- `InProgress`: an expansion is ongoing. This is important as after a conversation with @parth-gr we agreed the UI will prevent the user from disabling the smart scaling feature while an expansion is going on (wait until completion (success/error) to prevent unexpected behaviors).

BlaineEXE: Thanks, Alfonso. I see how `NotStarted` is a good indicator that the system is on but no expansion is needed. Sounds like we should keep that.

I'm glad to have more info about `InProgress` also. I think it would be good to eliminatethe the `Failed` condition so we don't have a system that flaps between `InProgress` and `Failed` if the reconcile encounters errors.

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,83 @@
...
+    spec:
+        storageCluster: 
+            name: <storageclusterName>
+        deviceClassName: ssd
+        clusterCapacityLimit: <capacity>
+        maxOsdSize: 8Ti
+        timeoutSeconds: 1800   
+    status:
+        phase: (NotStarted | InProgress | Failed | Succeeded)
+        LastRunTimestamp: (only on Failed | Succeeded)
```
travisn, line 30: This timestamp is confusing. If it's failed, that just means the most recent expansion hasn't succeeded yet. It could still succeed if there is some issue resolved in the cluster (add more nodes, add more mem/cpu, etc). What about splitting to these two times? As a naming convention, I also recommend "Time" instead of "Timestamp"
```yaml
    lastExpansionStartTime: (always set after a new expansion is started)
    lastExpansionCompletionTime: (only set on success)

File: docs/design/storage-auto-scale.md [hunk 29]
```diff
@@ -0,0 +1,83 @@
...
+
+- Watch prometheus periodically every 10 mins and depending on the Osd fill percentage, reconcile the StorageAutoScaling CR.
+  
+- If the Osd size is less than maxOsdSize(default:8Tib), do vertical scaling by doubling the each osd sizes for that device class.
+
+- If the Osd size is equal to maxOsdSize(default:8Tib), do a horizontal scaling, by adding 1 osd of 8Tib on each storage node, by simply increasing the (count+1) in the device set associated with a device class.   
+
+## Low-Level Design 
+
+1) Create a controller that requeueAfter(https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#Result) every 10 mins.
```
travisn, line 45: format as a link
```suggestion
1) Create a controller that [requeueAfter](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#Result) every 10 mins.
```

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,83 @@
...
+   
+2) List all the device class Osds used percentage using Prometheus metrics.
+    Query: (ceph_osd_metadata * on (ceph_daemon, namespace, managedBy) group_right(device_class,hostname) (ceph_osd_stat_bytes_used / ceph_osd_stat_bytes))
+       `OsdPercentage`:
+        ```
+        osd1-ssd               69%
+        osd2-ssd               72%
+        ```
+
+3) If the highest osd percentage reported is more than 70%(reaching osd nearfull), scaling is needed.
```
travisn, line 55: Let's make this a CR setting that they can override the default if desired

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,83 @@
...
+- Watch prometheus periodically every 10 mins and depending on the Osd fill percentage, reconcile the StorageAutoScaling CR.
+  
+- If the Osd size is less than maxOsdSize(default:8Tib), do vertical scaling by doubling the each osd sizes for that device class.
+
+- If the Osd size is equal to maxOsdSize(default:8Tib), do a horizontal scaling, by adding 1 osd of 8Tib on each storage node, by simply increasing the (count+1) in the device set associated with a device class.   
+
+## Low-Level Design 
+
+1) Create a controller that requeueAfter(https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#Result) every 10 mins.
+
```
travisn, line 46: 1a) If an expansion is in progress, check the progress and then requeue each 1 minute until the expansion is completed successfully.

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,83 @@
...
+        ```
+
+3) If the highest osd percentage reported is more than 70%(reaching osd nearfull), scaling is needed.
+
+4) If scaling is needed calculate the `expectedOsdSize` and `expectedOsdCount`.
+    1)  Calculate the total number osds and osd size from the prometheus response.
+    2)  Calculate the total number osds and osd size from the storageCluster.
+    3)  If the above values are equal(needed to avoid double re-sizing), create or update the `OsdPercentage map`
+        1)  If the Osd size is less than maxOsdSize(default:8Tib), do vertical scaling by doubling the each osd sizes for that device class.
+        2)  If the Osd size is equal to maxOsdSize(default:8Tib), do a horizontal scaling, by adding 1 osd of maxOsdSize(default:8Tib) on each storage node.
```
travisn, line 62: The OSDs will be added for each device set, right? It might be per-zone, rack, or host.
```suggestion
        2)  If the Osd size is equal to maxOsdSize(default:8Tib), do a horizontal scaling, by adding 1 osd of maxOsdSize(default:8Tib) on each `storageDeviceSet`
```

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,83 @@
...
+                expectedOsdSize: 4
+                expectedOsdCount: 6
+            ```
+
+5) Depending on the `OsdPercentage map` set the `expectedOsdSize` and `expectedOsdCount` and on the StorageAutoScaling status to reflect the new expected value. And update `phase` to `InProgress` on the `StorageAutoScaling` CR which need scaling.
+
+6) Scale with the following algorithm:
+   1) Check if the totalCapacity allows to scale it or not.
+   2) Patch the Storagecluster, with all the device sets update needed at the same time.
+   3) Verify the Storagecluster whether the new osds are added or scaled in size, for all the device sets.
```
travisn, line 76: Seems like no need for verification, if the patch in step 2 successfully updated?

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,83 @@
...
+                expectedOsdCount: 6
+            ```
+
+5) Depending on the `OsdPercentage map` set the `expectedOsdSize` and `expectedOsdCount` and on the StorageAutoScaling status to reflect the new expected value. And update `phase` to `InProgress` on the `StorageAutoScaling` CR which need scaling.
+
+6) Scale with the following algorithm:
+   1) Check if the totalCapacity allows to scale it or not.
+   2) Patch the Storagecluster, with all the device sets update needed at the same time.
+   3) Verify the Storagecluster whether the new osds are added or scaled in size, for all the device sets. 
+   4) If the scaling is successful will raise a `success` alert and update the status of the `StorageAutoScaling` CR.
```
travisn, line 77: For updating the status of success or failure, it will come during a requeue, right? Let's have a separate section in the doc to be clear about what happens during the status check for in-progress expansions, compared to what is done at the time the expansion is started.

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,83 @@
...
+   5) If it is fails will do a re-reconcile and, will raise a `failure` alert if scaling not `succeed` with in timeoutSeconds(default:1800) interval.   
+
```
travisn, line 79: 7) Requeue for checking the expansion status each 1 minute

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,83 @@
...
+        storageCluster: 
+            name: <storageclusterName>
+        deviceClassName: ssd
+        clusterCapacityLimit: <capacity>
+        maxOsdSize: 8Ti
+        timeoutSeconds: 1800   
+    status:
+        phase: (NotStarted | InProgress | Failed | Succeeded)
+        LastRunTimestamp: (only on Failed | Succeeded)
+        currentOsdSize: 2TiB
```
travisn, line 31: After the expansion is completed (successful), will these "currentOsdSize" values remain in the status? If so, it would be confusing. But if we rename them to "start" osd size and count, then the status still makes sense when it is seen in the status after completion.
```suggestion
        startOsdSize: 2TiB
```

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,83 @@
...
+        expectedOsdSize: 4TiB
+        currentOsdCount: 3​
```
travisn, line 33:
```suggestion
        startOsdCount: 3​
```

travisn (changes requested):

docs/design/storage-auto-scale.md
parth-gr: @BlaineEXE we wont be switching from InProgress to failed directly, we would be setting it to In-progress with a timeout, if the timeout occurs then we will set the status to failed and raise a user alert.

parth-gr: okay then we can just keep in-progress

docs/design/storage-auto-scale.md
parth-gr: Can you take a look https://github.com/red-hat-storage/ocs-operator/pull/3045#discussion_r1969786669
And lets discuss if its really that important? then we can have a storagecluster level setting

docs/design/storage-auto-scale.md
parth-gr: correct

docs/design/storage-auto-scale.md
parth-gr: we need verification wether the scaling is done successfully or not, as the actual scaling is dependent on 3rd party apis

docs/design/storage-auto-scale.md
parth-gr: Both will set it to in-progress, based on the status values we can understand if its in-progress or needs to start

parth-gr: So if it in the end of the reconcile if its in-progress we will requeue 1 min , otherwise we will requeue every 10 min

docs/design/storage-auto-scale.md
travisn: Customers can change the near-full level from 75% to some other percentage with the odf cli tool (see [here](https://github.com/red-hat-storage/odf-cli/blob/main/docs/set.md#nearfull)). If they change that warning level, then they likely want to have a different threshold for autoscaling. So I would vote to have this as an autoscaler CR setting. 

And why not allow different thresholds for different CRs? I don't know why different thresholds would be desired for different device classes, but I don't see a problem if they were different.

docs/design/storage-auto-scale.md
BlaineEXE: Am I understanding that this part of the design is intended to explain how ocs-operator or the UI will name the resource when scaling is turned on? If so, it would be worth explaining that in the design so we are clear on what the plan is.

docs/design/storage-auto-scale.md
BlaineEXE: I would disagree that a LocalObjectReference is necessary. Consider that PersistentVolumeClaims reference their bound PersistentVolume simply by the name. This allows for a more portable architecture that isn't possible when objects are bound strictly by their UID. An object spec copied from one cluster to another is portable using `storageClusterName` but isn't portable using a full, weighty LocalObjectReference

docs/design/storage-auto-scale.md
BlaineEXE: This is still somewhat unclear to me. Is this total capacity of all disks, raw? Is this the Ceph cluster capacity after factoring in replication/erasure-coding?

docs/design/storage-auto-scale.md
BlaineEXE: (note: deleted prior comment where I had confused LocalObjectReference with the heavier-weight ObjectReference)

docs/design/storage-auto-scale.md
parth-gr: @BlaineEXE  It is the total capacity of all disks, what the final size disk can grow horizontally and vertically,
 
 size of all  OSD per device class* max size <= clusterCapacityLimit

As it is already a CR specific to device class, it was fine to not to add a prefix to it `deviceClassStorageCapacityLimit`
But I can change it to `deviceClassStorageCapacityLimit` unless any other  suggestions

docs/design/storage-auto-scale.md
parth-gr: okay,

docs/design/storage-auto-scale.md
travisn: Ok, I see you're just referring to checking for the completion so we can update the status. Let's just be clear that checking the status happens after we requeue, rather than waiting for completion here.

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,107 @@
...
+  
+- If the Osd size is less than maxOsdSize(default:8Tib), do vertical scaling by doubling the each osd sizes for that device class.
+
+- If the Osd size is equal to maxOsdSize(default:8Tib), do a horizontal scaling, by adding 1 osd of 8Tib on each storage node, by simply increasing the (count+1) in the device set associated with a device class.   
+
+## Low-Level Design 
+
+1) Create a controller that [requeueAfter](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#Result) every 10 mins.
+    a) If an expansion is in progress, check the progress and then requeue each 1 minute until the expansion is completed successfully.
+    b) If the scaling is not in progress set the `lastExpansionStartTime` as `current-time`.
```
BlaineEXE, line 51: I'm curious about this point. It seems counterintuitive to me to update a field called "last expansion start time" when there aren't expansions ongoing.

docs/design/storage-auto-scale.md
parth-gr: as discussed offline, we need this as we need to update the startOSD size and count in the stauts

docs/design/storage-auto-scale.md
parth-gr: if there is no scaling in progress, it means scaling needs to be started that is why setting it as current time, we can move this later too as you mentioned when the actually scaling will take place

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,107 @@
...
+    b) Verify is the scaling is successful. 
+    c) If the scaling is successful set the `phase` to `completed`.
+    b) If the scaling is un-successful requeue every 1 min, we have the 2nd case. 
+
+2) If the scaling has already started and its requeue
+    a) Now the requeue will happen every 1 min.
+    b) At the start of reconcile will match that `startOsdSize` and `expectedOsdSize` is not equal and similar for osd count.
+    c) And another validation will do is equating storagecluster spec with prometheus response.
+    d) Will requeue till the scaling is in-progress.
+    e) If the scaling is in-progress with more than timeoutSeconds(default:1800) interval we will raise a user alert.
```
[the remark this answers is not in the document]
parth-gr: it's about UI alerts, UI will see if the scaling is progress more than timeoutSeconds then raise a alert,

@aruniiird  that can be possible?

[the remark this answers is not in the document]
alfonsomthd: The UI just receives the backend alerts, we don't raise any other alerts. Thta alerts should be triggered by the backend.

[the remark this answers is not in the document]
aruniiird: That is right (what @alfonsomthd mentioned in the above comment), we don't have a separate alerts which are raised from UI, it only receives alerts from backend.
@alfonsomthd , is it possible, for the UI, to add a logic in the UI code itself such that if UI detects that the scaling is in `in-progress` state for `timeout` period of time (that is to say for a long time), then raise some kind of UI warning/visual-indication that the `scaling` process is taking a very long time than expected (something of that sort, we can discuss the UX part later if this is possible)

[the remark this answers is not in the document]
parth-gr: We need a user alert, so why not include it in the alert manager ocs metrics pod as we do for other crs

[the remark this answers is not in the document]
alfonsomthd: No, we don't have any implementation for background alerting neither the Toast notification system yet.
As this specific alert is currently out of scope (not part of the feature request), we can explore its implementation in the next feature iteration.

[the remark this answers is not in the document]
alfonsomthd: > We need a user alert, so why not include it in the alert manager ocs metrics pod as we do for other crs

You're referring to add it on the backend side, right? @aruniiird to provide more insights on this.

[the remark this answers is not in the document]
aruniiird: Had a discussion with @alfonsomthd, as we have an alert part we may not require a separate UI logic.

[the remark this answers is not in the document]
aruniiird: > You're referring to add it on the backend side, right? @aruniiird to provide more insights on this.

Yes @alfonsomthd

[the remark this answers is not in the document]
Madhu-1: This sounds good, As the design should work for both UI and the CLI commands, we should not consider UI as the only user.

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,111 @@
...
+  
+    ```
+    apiVersion: ocs.openshift.io/v1
+    kind: StorageAutoScaling
+    metadata:
+        name: <storageclusterName>-<deviceClassName>
+        namespace: openshift-storage
+    spec:
+        storageCluster: 
+            name: <storageclusterName>
```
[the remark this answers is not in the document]
parth-gr: its `LocalObjectReference` to reference object within same namespace

docs/design/storage-auto-scale.md
parth-gr: We want the CR names to be unique that is created by UI, so its just a UI requirement flow and user can have any names

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,111 @@
...
+        deviceClassName: ssd
+        deviceClassStorageCapacityLimit: <capacity>
+        osdScalingThresholdPercent: 70%  # should be less than osd nearfull percentage 
+        maxOsdSize: 8Ti
```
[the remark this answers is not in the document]
parth-gr: Yes that should be a good way for the CRD, will take care of it during the implementation. 
But anyway the size still cant be decrease as there is a check at rook operator

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,111 @@
...
+    a) Now the requeue will happen every 1 min.
+    b) At the start of reconcile will match that `startOsdSize` and `expectedOsdSize` is not equal and similar for osd count.
+    c) And another validation will do is equating storagecluster spec with prometheus response.
+    d) Will requeue till the scaling is in-progress.
+    e) If the scaling is in-progress with more than timeoutSeconds(default:1800) interval we will raise a user `failure` alert.
+    f) If there as a failure alert, provide a mitigation guide for the user. 
+
+## Failure case
+
+There might cases where system resources are exhausted and new Osds cannot be started.
```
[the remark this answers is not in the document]
parth-gr: That's a separate document that will be added to runbooks, no need to keep it here, once that is ready I will link it here.

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,111 @@
...
+    kind: StorageAutoScaling
+    metadata:
+        name: <storageclusterName>-<deviceClassName>
+        namespace: openshift-storage
+    spec:
+        storageCluster: 
+            name: <storageclusterName>
+        deviceClassName: ssd
+        deviceClassStorageCapacityLimit: <capacity>
+        osdScalingThresholdPercent: 70%  # should be less than osd nearfull percentage
```
BlaineEXE, line 28: Nit: With percent in name, this can be an int type IMO

```suggestion
        osdScalingThresholdPercent: 70  # should be less than osd nearfull percentage 
```

docs/design/storage-auto-scale.md
travisn: For clarity, how about we just change the design doc to this? If it's just a UI convention, we don't need to declare it in this design doc.
```suggestion
        name: <name>
```

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,113 @@
...
+    ```
+    apiVersion: ocs.openshift.io/v1
+    kind: StorageAutoScaler
+    metadata:
+        name: <storageclusterName>-<deviceClassName>
+        namespace: openshift-storage
+    spec:
+        storageCluster: 
+            name: <storageclusterName>
+        deviceClassName: ssd
```
travisn, line 26:
```suggestion
        deviceClass: ssd
```

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,113 @@
...
+        deviceClassStorageCapacityLimit: <capacity>
```
travisn, line 27: All settings are implicit to the device class, so we don't need the `deviceClass` as a prefix to the settings
```suggestion
        storageCapacityLimit: <capacity>
```

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,113 @@
...
+        osdScalingThresholdPercent: 70  # should be less than osd nearfull percentage
```
travisn, line 28: What about this name? But if you still like `osdScalingThresholdPercent` better, I'm fine with that.
```suggestion
        storageScalingThresholdPercent: 70  # should be less than osd nearfull percentage 
```

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,113 @@
...
+Based on the above algorithm there would be two conditions where in-progress is set, elaborating those conditions,
+
+1) If scaling is just started:
+    1) Set `phase` to `InProgress`.
+    2) Verify is the scaling is successful. 
+    3) If the scaling is successful set the `phase` to `completed`.
+    4) If the scaling is un-successful requeue every 1 min, we have the 2nd case. 
+
+2) If the scaling has already started and its requeue
+    1) Now the requeue will happen every 1 min.
```
travisn, line 103: @nb-ohad No concern with the re-queue pattern for checking status of an in-progress scaling, right?

nb-ohad:
> After Parth's and Travis's reports about the meeting earlier today, I started to think in earnest about how to design the system with a prometheus goroutine, and the more I think into it, the more I think that there is too much complexity and risk there.

> Error handling becomes challenging with controller-runtime when goroutines are involved. This is something we've seen in Rook. The prometheus goroutine isn't going to have a resource associated with it to report errors to, so where do those errors go? Its error reporting options will be to either fail the whole ocs-operator routine (risking CLBO) or it will have to just log an error to the log, where no one may notice it.

> The routine could report events, but events are also often missed. I don't think it's appropriate to have the goroutine changing the status of an autoscale CR. Which one will it update? A random one? All of them?

> I also don't think giving the goroutine Kube API access will solve the error reporting issue. If there is an error with error, no one will know the system isn't operating as intended, and because it exists outside of controller-runtime, the system may just happily carry on oblivious to the fact.

> I can understand the worry that there will be bugs introduced over time in the retryAfter approach, but that risk exists for all software regardless. There are other ways we can mitigate future issues in the retryAfter approach without resorting to using a groutine that has us trying to work around controller-runtime rather than working within it.

> As for the issue of querying prometheus too often, that issue is easily solved without a goroutine. A monad, lazy-loading-type prometheus cache object that updates its cache when called and the cache is older than 10 minutes is simple. And any errors encountered could be reported from the reconcile as normal. The error would be applied to a seemingly-random autoscale CR, but at least it is reported somewhere.

@BlaineEXE error handling even when we have everything in a single controller is tricky. Mainly because you cannot just error out of your reconcile whenever you see the first error as that might result in blocking update on any other device class in the list of device classes. 

I don't thing that this is going to change with goroutine. Go is known for its concurrency model and provide building block that can address all of our concerns. As a matter of fact, we support a full operator that took the approach of running all resource reconciling in parallel, instead of doing it serially, (which is a wider goal then the one we are describing in this design)  and the resulted code complexity is not much higher while still getting all of the benefits.  (https://github.com/ceph/ceph-csi-operator/blob/main/internal/controller/driver_controller.go#L232-L242) 

My experience showed me that there are more hurdle when it comes to having controller running within an infinite loops,  and not adhere to separation of responsibility within a solution

nb-ohad:
On the topic of error handling:

### Controller:
For controllers, If an iteration fails, we usually report an error, update the status then reschedule (happens automatically because we returned an error) -  For the controller this stays exactly the same

### Scrapper Goroutine:
We should expect similar behavior. The nature of the routine is to run an infinite loop, which by its nature gives you a "rescheduling" mechanism. Now one might argue that waiting a fixed X min for failure is good enough, that is easily solved by have the sleep at the end of the iteration depend on the result the current iteration of the loop (we could take a backoff approach, the facilities are already available today as util functions)

On the subject of reporting the error, there is no issue writing to the log from the goroutine. the logger we use is thread safe.
But I do agree we need to surface the following information on a CR status section, which can be done using a Status patch command. 

The  main issue is selecting which CR to update as the scraping operation is not targeting a specific device class, but all of them at once and it does not make sense to go and update the status of all autoscaling CRs when this one fails. 
But we do need to understand that this is a problem that we will have to solve even if we don't use a goroutine and stay within controller bounderies! The scraping happen once for all CRs, we already rules out a design where each device class issue its own prom scraping call

docs/design/storage-auto-scale.md
nb-ohad: 1. I have no issue with that, mainly because if we consider scaling in progress as an invalid state for acting on new configuration changes then this is just a plain error that will result in a re-queuing. 

2. Putting the above aside, I would like to understand what is that issue of acting on a change on the desired state while we are in progress?  k8s itself can handle it, so why can't we?

docs/design/storage-auto-scale.md
parth-gr: after discussion we came to this, I started with `storageCapacityLimit` itself, I think no harm in putting prefix

docs/design/storage-auto-scale.md
parth-gr: what would be the source of truth for the desired state, we don't have desired state in the cr spec, and the sync map we even don't have the desired count, its moreover in the cr status that is been updated

docs/design/storage-auto-scale.md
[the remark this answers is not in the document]
alfonsomthd: @parth-gr @BlaineEXE So we're raising a `Failure` alert but not transitioning to `Failed` state?
How we'll prevent the alerting system from continuously raising alerts (provided the same conditions are still the same over time)?
cc @aruniiird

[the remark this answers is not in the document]
BlaineEXE: My understanding from a convo earlier today is that we will transition to a `Failed` state if a scale event exceeds the timeout. 

The important detail is that the state should not transition from `Failed` back to `InProgress` to avoid flapping-state issues. `Failed`->`Successful` is valid, however. This mirrors the `CrashLoopBackoff` state for Pods. Pods don't stop being in CLBO until they recover.

docs/design/storage-auto-scale.md
travisn: By the same argument, seems like then we need to name them `deviceClassMaxOsdSizeLimit`, or `deviceClassTimeoutSeconds`, and so on. So I would still vote to remove the prefix for consistency.

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,120 @@
...
+   
+8) Scale by patching the Storagecluster, with all the device sets update needed at the same time.
+
+9)  Verify and Alert:
+   
+   1) Verify the Storagecluster whether the new osds are added or scaled in size, for all the device sets.
+ 
+   2) If the scaling is successful will raise a `Succeeded` alert and update the status of the `StorageAutoScaling` CR with `lastExpansionCompletionTime` and `phase` and also osd count and size.
+   
+   3) If it is fails will do a requeue every 1 min and, will raise a `failure` alert and change the phase to `failed` if scaling not `Succeeded` with in timeoutSeconds(default:1800) interval.
```
travisn, line 97:
```suggestion
   3) If the auto scale is not completed, it will do a requeue every 1 min and, will raise a `failure` alert and change the phase to `failed` if scaling not `Succeeded` with in timeoutSeconds(default:1800) interval.
```

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,120 @@
...
+   
+7) Update the status, set the `expectedOsdSize` and `expectedOsdCount` to reflect the new expected value and also set `startOsdSize` and `startOsdCount` with current storagecluster values.
+   
+8) Scale by patching the Storagecluster, with all the device sets update needed at the same time.
+
+9)  Verify and Alert:
+   
+   1) Verify the Storagecluster whether the new osds are added or scaled in size, for all the device sets.
+ 
+   2) If the scaling is successful will raise a `Succeeded` alert and update the status of the `StorageAutoScaling` CR with `lastExpansionCompletionTime` and `phase` and also osd count and size.
```
travisn, line 95: Alerting is based on the phase, right? Let's be clear in this design that we update the phase, and that monitoring will raise the alert based on the phase.
```suggestion
   2) If the scaling is successful will update the status of the `StorageAutoScaling` CR with `lastExpansionCompletionTime` and `phase` and also osd count and size.
```

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,120 @@
...
+   3) If it is fails will do a requeue every 1 min and, will raise a `failure` alert and change the phase to `failed` if scaling not `Succeeded` with in timeoutSeconds(default:1800) interval.
+
+### InProgress state:
+
+Based on the above algorithm there would be two conditions where in-progress is set, elaborating those conditions,
+
+1) If scaling is just started:
+    1) Set `phase` to `InProgress`.
+    2) Verify is the scaling is successful. 
+    3) If the scaling is successful set the `phase` to `completed`.
```
travisn, line 106: I thought we didn't have a `completed` phase?
```suggestion
    3) If the scaling is successful set the `phase` to `succeeded`.
```

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,120 @@
...
+    4) If the scaling is un-successful requeue every 1 min, we have the 2nd case.
```
travisn, line 107:
```suggestion
    4) If the scaling is not yet completed requeue every 1 min, we have the 2nd case. 
```

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,120 @@
...
+    2) Verify is the scaling is successful. 
+    3) If the scaling is successful set the `phase` to `completed`.
+    4) If the scaling is un-successful requeue every 1 min, we have the 2nd case. 
+
+2) If the scaling has already started and its requeue
+    1) Now the requeue will happen every 1 min.
+    2) At the start of reconcile will match that `startOsdSize` and `expectedOsdSize` is not equal and similar for osd count.
+    3) And another validation will do is equating storagecluster spec with prometheus response.
+    4) Will requeue till the scaling is in-progress.
+    5) If the scaling is in-progress with more than timeoutSeconds(default:1800) interval we will raise a user `failure` alert.
```
travisn, line 114: Separately, let's be clear about how alerts are raised. OCS operator doesn't do that, right?
```suggestion
    5) If the scaling is in-progress with more than timeoutSeconds(default:1800) interval we set the phase to `failed`.
```

travisn (commented):
LGTM, just a few minor clarifications.

docs/design/storage-auto-scale.md
```diff
@@ -0,0 +1,120 @@
...
+   
+3) If an expansion is in progress(expectedOsdSize!=startOsdSize || expectedOsdCount!=startOsdCount), check the progress and then requeue each 1 minute until the expansion is completed successfully(jump to step 9).
+   
+4) If the highest osd percentage reported in the sync map is more than osdScalingThresholdPercent(70%) means reaching osd nearfull, scaling is needed.
+   
+5) Check if the totalCapacity allows to scale it or not.
+        
+    1) If yes, Update `phase` to `InProgress`  and `lastExpansionStartTime` as `current-time` on the `StorageAutoScaling` CR which need scaling.
+        
+    2) If no, don't reconcile further and change the `phase` to `Failed`
```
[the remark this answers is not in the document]
parth-gr: @travisn then you have to raise an alert in this case to the user, who has configured auto-scaling, 
We don't want users to wait till they arrive at a near-full state.

As per Eran,
> fail due to datastore exhaustion or exceeding quota in the cloud.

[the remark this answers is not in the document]
travisn: This is different from datastore exhaustion or exceeding cloud quota. The user should already be aware of this maximum storage threshold since it's in the autoscale settings. So maybe there is an alert when we have reached that threshold, but it's just informational. If they hit the nearfull alert later, they can adjust the autoscale settings if they want.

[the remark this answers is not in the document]
parth-gr: >So maybe there is an alert when we have reached that threshold,

How would that logic implemented, alert can only look at `phase`, do you also want  alert  when phase: `succeeded` and to match `currentosdState` and `threshold`, if they are almost equal raise an alert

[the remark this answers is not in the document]
parth-gr: can we add `expectedTotalCapacity` in the status, and the alert can read if the `expectedTotalCapacity` >= `storageCapacityLimit`?
And if we alert this user will keep getting this alert until they fixes the limits
cc @travisn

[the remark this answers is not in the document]
travisn: I like the idea of adding `storageCapacity` to the status. And to be consistent with the other status, perhaps we have both `startStorageCapacity` and `expectedStorageCapacity`. Then some alert could watch for exceeding the `storageCapacityLimit` as you suggested.

docs/design/storage-auto-scale.md
parth-gr: > would like to understand what is the issue of acting on a change on the desired state while we are in progress? k8s itself can handle it, so why can't we? 

The desired state is internally calculated by the controller so let's finish the in-progress state

docs/design/storage-auto-scale.md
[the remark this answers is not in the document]
BlaineEXE: I don't understand why we would design a system that sets an `expectedStorageCapacity` larger than is limited by the input spec. We've said that any time `expectedXyz` is set, it's is an indication that the system is beginning an expansion. Shouldn't we return an error when we can no longer expand instead?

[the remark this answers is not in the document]
parth-gr: >We've said that any time expectedXyz is set, it's is an indication that the system is beginning an expansion and that the user can expect the system to arrive at expectedXyz at some point in the future. 

`expectedStorageCapacity` is not a indication of expansion, but `expectedOSDCount` and size,
I feel your point generally if expected is set it turns out that expansion is needed, but we need a system that notifies users on the basis status.


> Shouldn't we return an error when we can no longer expand instead?

if the phase is not failing, we cant error it, correct?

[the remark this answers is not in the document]
BlaineEXE: We could also consider a separate failure condition for this named `CapacityLimitReached`. I'm not sure it's necessary, but it is an error condition that the system won't exit from unless the user changes the `spec.totalCapacityLimit`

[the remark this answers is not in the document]
parth-gr: As we have moved away from `conditions` and using `TimeStampedError` not sure how this would be possible,
Moreover, I like to see @travisn views on it, I vote for the current flow but open for any better approach.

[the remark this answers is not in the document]
BlaineEXE: Apologies. I used the word "condition" when I should have said "phase". 

I still think that most users, without looking at documentation, will assume that any `expectedXyz` field different from `currentXyz` will imply to them that the system is operating normally and attempting to expand the cluster. That would be especially true if we leave the status at `NotStarted` or `Succeeded`. They will likely not notice that the expected is in excess of the limit, and I suspect a user in this position will contact support.

If we want to communicate this condition to users clearly, I believe it will be better to enter `Failed` state (with TimestampedError set) or to add a new `CapacityLimitReached` state to communicate this edge case.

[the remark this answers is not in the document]
BlaineEXE: FWIW,

> I like the idea of adding storageCapacity to the status.

I do agree with this. Doing a little math for users seems beneficial.

> And to be consistent with the other status, perhaps we have both startStorageCapacity and expectedStorageCapacity. 

I'm mostly ambivalent about this detail. It could be `totalCapacity` or separated capacities.

> Then some alert could watch for exceeding the storageCapacityLimit as you suggested.

I have disagreement here. While OCS **can** alert on this, I still think that the custom resource should reasonably be able to alert on it since it knows all of the information. That way, OCS doesn't have to implement specialized alert logic for that corner case because it can just surface the CR's status, which is also what we expect to happen when the timeout is reached.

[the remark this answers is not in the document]
parth-gr: Understood, setting a  new status spec `CapacityLimitReached` would be a good idea, As we really dont want to fail as per the above concern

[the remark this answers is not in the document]
parth-gr: I would like to call it as `StorageCapacityLimitReached `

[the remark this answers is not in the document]
travisn: Perfect to have a `storageCapacityLimitReached` so it's easy to see if any more expansions could be made. 

But I'm not clear about the discussion around storage capacity status... It seem very consistent to include it next to the osd size and osd count, to show what the user should expect after the expansion is completed. 
```
    status:
        ...
        startOsdSize: 2TiB
        expectedOsdSize: 4TiB
        startOsdCount: 3​
        expectedOsdCount: 3
        startStorageCapacity: 6TiB
        expectedStorageCapacity: 12TiB
```
Or perhaps what would help clarify is to rename the prefix. In previous comments, we already changed the prefix, but it still seems confusing. Perhaps they would be better as `expected` --> `expanded`:
```
    status:
        ...
        startOsdSize: 2TiB
        expandedOsdSize: 4TiB
        startOsdCount: 3​
        expandedOsdCount: 3
        startStorageCapacity: 6TiB
        expandedStorageCapacity: 12TiB
```
However, the storage capacity status is not at all necessary to close on the design doc, just seems nice to have.

travisn (commented):
/lgtm
This design looks in a stable state to merge. Any changes later could be made with the implementation PR.

travisn:
hmm, I did not expect the auto merge after just adding the lgtm label, while there were other open requests for changes!

docs/design/storage-auto-scale.md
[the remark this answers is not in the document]
parth-gr: `expected` lets users know to what size it will be expected to expand, and expand doesn't give a good idea, I think we can keep as it is,

`startStorageCapacity` is not a necessary and not needed by reconcile, so we can add it later as improvement if needed in the UI

parth-gr:
> hmm, I did not expect the auto merge after just adding the lgtm label, while there were other open requests for changes!

let's discuss anything else is needed, hope this is the only thread https://github.com/red-hat-storage/ocs-operator/pull/3045#discussion_r1987071323,

PR #3047, opened 2025-02-19
provider: Add S3Endpoint Service Proxy

ezio-auditore:


services/provider/server/server.go
```diff
@@ -537,6 +543,36 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
+	if err = s.client.Get(ctx, client.ObjectKeyFromObject(nb), nb); err != nil {
+		return nil, fmt.Errorf("failed to get noobaa %v", err)
+	}
+	if nb.Status.Phase != nbv1.SystemPhaseReady {
+		// Noobaa system is not yet ready
+		return extR, nil
+	}
+	nbS3Endpoint := nb.Status.Services.ServiceS3.NodePorts[0]
+	extR = append(extR, &pb.ExternalResource{
+		Name: "S3EndpointProxy",
```
alfonsomthd, line 560: @ezio-auditore As per the current console-proxy service declaration, the service name in the consumer cluster must be `s3`:
```
oc -n openshift-storage get consoleplugin odf-console -oyaml
[...]
spec:
  [...]
  proxy:
  [...]
  - alias: s3
    authorization: None
    endpoint:
      service:
        name: s3    <=================================
        namespace: openshift-storage
        port: 443
      type: Service
```

alfonsomthd (changes requested):

services/provider/server/server.go
alfonsomthd: @ezio-auditore sorry, in fact we'll be adding a new declaration for the `odf-client-console` plugin, so we can go with your suggested name but with this format:`s3-endpoint-proxy`.

ezio-auditore: @alfonsomthd have updated in the latest commit

ezio-auditore:
https://github.com/noobaa/noobaa-operator/pull/1559

services/provider/server/server.go
```diff
@@ -523,20 +538,84 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
+	nb.Namespace = s.namespace
+
+	if err = s.client.Get(ctx, client.ObjectKeyFromObject(nb), nb); err != nil {
+		return nil, fmt.Errorf("failed to get noobaa %v", err)
+	}
+	if nb.Status.Phase != nbv1.SystemPhaseReady {
+		// Noobaa system is not yet ready
+		return extR, nil
+	}
+	nbS3Endpoint := nb.Status.Services.ServiceS3.NodePorts[0]
```
alfonsomthd, line 569: @ezio-auditore Provided this is for internal usage and not to be exposed, can we also add the http endpoint (no SSL) if there is no security concern?

services/provider/server/server.go
```diff
@@ -519,20 +534,84 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
+		}),
+	})
+
+	extR = append(extR, &pb.ExternalResource{
+		Name: "s3-endpoint-proxy",
+		Kind: "Endpoints",
+		Data: mustMarshal(&v1.Endpoints{
+			Subsets: []v1.EndpointSubset{
+				{
+					Addresses: []v1.EndpointAddress{
```
leelavg, line 599: Is there any reason to use Endpoints rather than the newer recommended EndpointSlice?

services/provider/server/server.go
ezio-auditore: Added  http port 80

services/provider/server/server.go
ezio-auditore: EndpointSlices introduce extra API objects, leading to more Kubernetes API traffic. Since Noobaa service only uses 1 backend pod using a single Endpoints object reduces unnecessary API operations.
Easier to debug and troubleshoot

leelavg: Fair enough, thanks.

ezio-auditore:
possible intermitent issue

ezio-auditore:
Unrelated test fail
```Github rate-limiter failed the request. Either authenticate or wait a couple of minutes.```

PR #2609, opened 2024-05-14
remove CSI_DISABLE_HOLDER_PODS key for all clusters 4.18 and onwards

iamniting:
Customers have had two releases (4.16, 4.17) for migration, As planned we are removing this setting.

Signed-off-by: Nitin Goyal <nigoyal@redhat.com>

PR #3049, opened 2025-02-20
core: add back the CSI_DISABLE_HOLDER_PODS key on the cm

parth-gr:
If the OCS operator is upgraded to 4.18 before the rook operator,
It will go and remove the `CSI_DISABLE_HOLDER_PODS` key from the ocs-configmap,
Removing the key from the ocs-cm will re-start the rook operator pods,

As Rook operator pod has not been upgraded and is still on 4.17, it still looks for the env variable `CSI_DISABLE_HOLDER_PODS` key, And as the key is deleted already by ocs operator 4.18 pod, rook operator got stuck and upgrade fails.

Revert of the pr https://github.com/red-hat-storage/ocs-operator/pull/2609/ 

And Revert the fix in 4.19.


hack/install-ocs-operator.sh
```diff
@@ -28,6 +28,7 @@ data:
   CSI_ENABLE_TOPOLOGY: "test"
   CSI_TOPOLOGY_DOMAIN_LABELS: "test"
   ROOK_CSI_ENABLE_NFS: "false"
+  CSI_DISABLE_HOLDER_PODS: "true"
```
Madhu-1, line 31: @parth-gr can you please update the PR description on how this will solve the problem? i assume this file is only for dev testing not used in any released deployment.

parth-gr: Correct added the changes

PR #3050, opened 2025-02-20
DFBUGS-1656: [release-4.18] core: add back the CSI_DISABLE_HOLDER_PODS key on the cm

parth-gr:
If the OCS operator is upgraded before the rook operator, It will go and remove the CSI_DISABLE_HOLDER_PODS key from the ocs-configmap, Removing the key from the cm will re-start the rook operator pods,

As Rook has not been upgraded and is still on 4.17, it looks for the env variable value which was coming from ocs-configmap key, And as the key is deleted rook operator got stuck.

Revert of the pr #2609

And Revert the fix in 4.19.

Signed-off-by: parth-gr <partharora1010@gmail.com>
(cherry picked from commit 843bd4c94cf3d46e6e76ad30731990ac44b4b12a)


malayparida2000 (approved):
/lgtm

agarwal-mudit:
@parth-gr 
Did you test the fresh install also or only upgrade?
Also, it will not impact the holder pod functionality?

malayparida2000:
> @parth-gr Also, it will not impact the holder pod functionality?

No as in 4.18 rook CSV doesn’t reference to this Key after upgrade, so it just stays in the ocs-operator-config CM as a placeholder with no suage after upgrade to 4.18 is complete.

parth-gr:
>Did you test the fresh install also or only upgrade?
Also, it will not impact the holder pod functionality?

@agarwal-mudit the holder pod is already deprecated https://github.com/rook/rook/pull/14819 that's why this key was removed,


For testing:

1) On a 4.18 cluster -> try adding this empty key -> it worked
2) Tried a 4.17 to 4.18 upgrade


>changes looks good to me but please test with upgrade case

We cant control the OLM to upgrade the ocs operator first, so did 2 upgrades with @malayparida2000 but in those cases rook operator only upgraded first

malayparida2000:
> Did you test the fresh install also or only upgrade? 
We have tested the upgrade scenarior by replacing our fix image in a affected cluster.
And we have not directly tested the fresh install but we did some testing by scaling down deployments & editing the CM to see how it would behave upon fresh install

malayparida2000:
> We cant control the OLM to upgrade the ocs operator first, so did 2 upgrades with @malayparida2000 but in those cases rook operator only upgraded first

Yeah it's very difficult to reproduce this issue we had 2 extra clusters and in both rook operator upgraded first so we couldn't hit the issue. We did test our image in the affected cluster we had from QE & that worked there.

parth-gr:
Upgrade testing:

1) We scale down the ocs -operator 
2) changed the ocs cm to set `CSI_DISABLE_HOLDER_PODS: ""`
3) Restarted the rook operator, 4.17 Rook operator was still running fine, with reconcile failures
4) UPgraded the cluster, csv was successfully updated to 4.18 as the 4.17 rook operator was up and running.

PR #3054, opened 2025-02-24
DFBUGS-1667: [release-4.17] Fix conflict between SSD and HDD device classes causing CRUSH rule issue

malayparida2000:
In clusters where the devices were initially of type hdd and later replaced with ssd, Ceph continues to list hdd in the available device classes under ceph osd crush class ls. This occurs even after all devices have been replaced with ssd. As rook populates the available device classes from this list, the hdd device class persists in the CephCluster status.

Starting from OCS 4.16, OCS operator determines and assigns a defaultCephDeviceClass to all pools by selecting the first non-replica-1 entry in the CephCluster status. Due to the persistence of the hdd device class, the OCS operator incorrectly selects hdd as the default device class, even though all underlying device sets have been updated to ssd.

To address this issue, we are introducing a hardcoded check in the OCS operator: if both hdd and ssd device classes are present in the CephCluster status, the operator will prioritize and select the ssd device class as the default. Ideally Ceph should automatically update the device class list when devices are replaced, so may be a fix is required on ceph side as well.

### DIFF
--- a/controllers/storagecluster/cephcluster.go
+++ b/controllers/storagecluster/cephcluster.go
@@ -1405,25 +1405,35 @@ func getOsdCount(sc *ocsv1.StorageCluster) int {
 }
 
 func determineDefaultCephDeviceClass(foundDeviceClasses []rookCephv1.DeviceClasses, isReplica1 bool, replica1DeviceClasses []string) string {
+	determinedDeviceClass := DeviceTypeSSD
 	// If device classes are found in status
 	if len(foundDeviceClasses) != 0 {
-		// If replica-1 is not enabled return the first device class
+		// If replica-1 is not enabled choose the first device class
 		if !isReplica1 {
-			return foundDeviceClasses[0].Name
-		}
-		// If replica-1 is enabled return the first one that is not a replica1 device class
-		if len(foundDeviceClasses) != 0 {
+			determinedDeviceClass = foundDeviceClasses[0].Name
+		} else { // If replica-1 is enabled choose the first one that is not a replica1 device class
 			replica1DeviceClassMap := make(map[string]bool)
 			for _, deviceClass := range replica1DeviceClasses {
 				replica1DeviceClassMap[deviceClass] = true
 			}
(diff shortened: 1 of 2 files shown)


iamniting (approved):

malayparida2000:
After the repeated CI failures decided to peek into the run env found that ocs-metrics-exporter pod has some issues

```
~ $ oc get pods ocs-metrics-exporter-57b996798f-8v465        
NAME                                    READY   STATUS             RESTARTS      AGE
ocs-metrics-exporter-57b996798f-8v465   0/1     CrashLoopBackOff   5 (59s ago)   4m4s
mparida ~ $ oc logs ocs-metrics-exporter-57b996798f-8v465 
flag provided but not defined: -namespaces
Usage of /usr/local/bin/metrics-exporter:
  -enable-leader-election
    	Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.
  -health-probe-bind-address string
    	The address the probe endpoint binds to. (default ":8081")
  -kubeconfig string
    	Paths to a kubeconfig. Only required if out-of-cluster.
  -metrics-addr string
    	The address the metric endpoint binds to. (default ":8080")
  -zap-devel
    	Development Mode defaults(encoder=consoleEncoder,logLevel=Debug,stackTraceLevel=Warn). Production Mode defaults(encoder=jsonEncoder,logLevel=Info,stackTraceLevel=Error)
  -zap-encoder value
    	Zap log encoding (one of 'json' or 'console')
  -zap-log-level value
    	Zap Level to configure the verbosity of logging. Can be one of 'debug', 'info', 'error', or any integer value > 0 which corresponds to custom debug levels of increasing verbosity
  -zap-stacktrace-level value
    	Zap Level at and above which stacktraces are captured (one of 'info', 'error', 'panic').
  -zap-time-encoding value
    	Zap time encoding (one of 'epoch', 'millis', 'nano', 'iso8601', 'rfc3339' or 'rfc3339nano'). Defaults to 'epoch'.

```

PR #3062, opened 2025-02-26
update storageconsumer api for centralising distribution and making it user facing

leelavg:
new fields with *Classes are intended to be filled by storage admin for distributing to connected client cluster that the storageconsumer represents. Only `name` is specified and later will be expanded if more config is required.

configmapref name if provided will be owned by corresponding consumer and reconciled by the consumer controller or else will be created by controller and same will be reflected in the consumer status.


api/v1alpha1/storageconsumer_types.go
```diff
@@ -42,7 +42,18 @@ type StorageConsumerSpec struct {
 	Enable bool `json:"enable,omitempty"`
 	// StorageQuotaInGiB describes quota for the consumer
 	// +optional
-	StorageQuotaInGiB int `json:"storageQuotaInGiB,omitempty"`
+	StorageQuotaInGiB          int         `json:"storageQuotaInGiB,omitempty"`
+	RadosNamespace             string      `json:"radosNamespace,omitempty"`
+	SubVolumeGroup             string      `json:"subVolumeGroup,omitempty"`
```
nb-ohad, line 47: Should we call this `SubVolumeGroup` or `CephFsSubVolumeGroup`?

api/v1alpha1/storageconsumer_types.go
```diff
@@ -42,7 +42,18 @@ type StorageConsumerSpec struct {
...
+	StorageClasses             []ClassSpec `json:"storageClasses,omitempty"`
+	VolumeSnapshotClasses      []ClassSpec `json:"volumeSnapshotClasses,omitempty"`
+	VolumeGroupSnapshotClasses []ClassSpec `json:"volumeGroupSnapshotClasses,omitempty"`
+	ClientProfiles             []string    `json:"clientProfiles,omitempty"`
```
nb-ohad, line 51: @leelavg I believe we agreed that ClientProfiles are not admin configurable. Why do we have this as part of the consumer spec?

api/v1alpha1/storageconsumer_types.go
```diff
@@ -42,7 +42,18 @@ type StorageConsumerSpec struct {
...
+}
+
+// Represents the struct to represent a class
+type ClassSpec struct {
+	Name string `json:"name,omitempty"`
```
nb-ohad, line 56: Please have different types for the different types of classes

nb-ohad (changes requested):

api/v1alpha1/storageconsumer_types.go
leelavg: I picked these names to be consistent with clientprofile fields https://github.com/ceph/ceph-csi-operator/blob/main/api/v1alpha1/clientprofile_types.go

api/v1alpha1/storageconsumer_types.go
leelavg: I'm not sure, we decided to remove the secret (provisioner, node) aliases from spec but not clientProfiles.

Nevertheless, if we already know the storageconsumer upgrade mode we can figure these out, will update.

api/v1alpha1/storageconsumer_types.go
leelavg: can be done.

api/v1alpha1/storageconsumer_types.go
rewantsoni: In one of the meetings, we were talking about having the clientProfile name something similar to storageID (ceph_FSID,radosnamespace) and (ceph_FSID,subvolumegroupName). If we are going with that shouldn't we have two different clientProfiles? one for rbd and another for cephfs?

leelavg: This was in context when @Madhu-1 raised a point wrt MDR but it was decided that StorageConsumerUID will always be unique (for all intents & purposes) and could be used as clusterID and subsequently as clientProfile, isn't it?

Madhu-1: This is all depends on the clientProfile naming scheme. Do we have a PR in place where we are calculating the clientProfile names?

leelavg: no, didn't we settle w/ storageconsumer uid as they are unique? I don't remember we are calculating anything.

api/v1alpha1/storageconsumer_types.go
```diff
@@ -42,7 +42,24 @@ type StorageConsumerSpec struct {
 	Enable bool `json:"enable,omitempty"`
 	// StorageQuotaInGiB describes quota for the consumer
 	// +optional
-	StorageQuotaInGiB int `json:"storageQuotaInGiB,omitempty"`
+	StorageQuotaInGiB          int                            `json:"storageQuotaInGiB,omitempty"`
+	RadosNamespace             string                         `json:"radosNamespace,omitempty"`
```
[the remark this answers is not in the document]
leelavg: done.

api/v1alpha1/storageconsumer_types.go
```diff
@@ -42,7 +42,24 @@ type StorageConsumerSpec struct {
...
+	SubVolumeGroup             string                         `json:"subVolumeGroup,omitempty"`
+	StorageClasses             []StorageClassSpec             `json:"storageClasses,omitempty"`
+	VolumeSnapshotClasses      []VolumeSnapshotClassSpec      `json:"volumeSnapshotClasses,omitempty"`
+	VolumeGroupSnapshotClasses []VolumeGroupSnapshotClassSpec `json:"volumeGroupSnapshotClasses,omitempty"`
+}
+
+type StorageClassSpec struct {
+	Name string `json:"name,omitempty"`
```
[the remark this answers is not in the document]
leelavg: done by removing `omitempty`

leelavg:
/hold

Will squash after final review

api/v1alpha1/storageconsumer_types.go
nb-ohad: @leelavg @Madhu-1 @rewantsoni 
Going forward we have a single scheme as discussed and agreed and captured on the weekly call meeting notes.
We will support the older schemes based on the indicator that marks an updated internal or updated provider.
In all cases, this is not an admin nob

api/v1alpha1/storageconsumer_types.go
```diff
@@ -37,12 +37,37 @@ const (
...
 	// +optional
 	StorageQuotaInGiB int `json:"storageQuotaInGiB,omitempty"`
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	// +kubebuilder:validation:MaxLength=512
+	RadosNamespace string `json:"radosNamespace"`
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	// +kubebuilder:validation:MaxLength=512
+	SubVolumeGroup             string                         `json:"subVolumeGroup"`
```
nb-ohad, line 55: The validation here does not make sense. This need to be configurable all the way until the Consumer is enabled. At that point we should freeze it. 

But I question the need of adding these fields as part of this PR. Why do we need to couple these changes with the rest of the changes in this PR?

api/v1alpha1/storageconsumer_types.go
```diff
@@ -100,6 +129,18 @@ type ClientStatus struct {
...
+}
+
+type SubVolumeGroupSpec struct {
+	Name string `json:"name"`
+}
+
+type ClientProfileSpec struct {
+	Name string `json:"name"`
+}
+
```
nb-ohad, line 143: I think you meant to have these with a Status safix

nb-ohad (changes requested):

api/v1alpha1/storageconsumer_types.go
```diff
@@ -42,7 +42,24 @@ type StorageConsumerSpec struct {
 	Enable bool `json:"enable,omitempty"`
 	// StorageQuotaInGiB describes quota for the consumer
 	// +optional
-	StorageQuotaInGiB int `json:"storageQuotaInGiB,omitempty"`
+	StorageQuotaInGiB          int                            `json:"storageQuotaInGiB,omitempty"`
+	RadosNamespace             string                         `json:"radosNamespace,omitempty"`
+	SubVolumeGroup             string                         `json:"subVolumeGroup,omitempty"`
```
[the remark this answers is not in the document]
nb-ohad: I don't think so. The idea is to point particular consumer to a specific entity by identifying it.  
Are you predicting that we need more then a name to uniquely identify these Ceph entities?

api/v1alpha1/storageconsumer_types.go
umangachapagain, line 55:
```suggestion
	SubVolumeGroup             string                         `json:"subVolumeGroup,omitempty"`
```
since it is Optional.

umangachapagain (changes requested):

api/v1alpha1/storageconsumer_types.go
leelavg: `omitempty == // +kubebuilder:validation:Optional`

leelavg: removed.

api/v1alpha1/storageconsumer_types.go
leelavg: yes, done.

api/v1alpha1/storageconsumer_types.go
```diff
@@ -62,11 +79,15 @@ type StorageConsumerStatus struct {
...
+	// +kubebuilder:deprecatedversion:warning="This field has been deprecated and will be removed in future."
 	CephResources []*CephResourcesSpec `json:"cephResources,omitempty"`
 	// Timestamp of last heartbeat received from consumer
 	LastHeartbeat metav1.Time `json:"lastHeartbeat,omitempty"`
 	// Information of storage client received from consumer
-	Client ClientStatus `json:"client,omitempty"`
+	Client                ClientStatus          `json:"client,omitempty"`
+	RadosNamespace        RadosNamespaceStatus  `json:"radosNamespace,omitempty"`
+	CephFsSubVolumeGroup  SubVolumeGroupStatus  `json:"cephFsSubVolumeGroup,omitempty"`
+	CephCsiClientProfiles []ClientProfileStatus `json:"cephCsiClientProfile,omitempty"`
```
rewantsoni, line 90: Will it benefit us to have different sections for rbd and cephfs, or club the clientProfile name in the radosnamespaceStatus or subVolumeGroupStatus?

leelavg: This is as per agreed design, clientProfiles data will just be a copy of the info generated during api response and here we just have the alias names.

Madhu-1 (commented):
LGTM

api/v1alpha1/storageconsumer_types.go
```diff
@@ -42,7 +43,23 @@ type StorageConsumerSpec struct {
 	Enable bool `json:"enable,omitempty"`
 	// StorageQuotaInGiB describes quota for the consumer
 	// +optional
-	StorageQuotaInGiB int `json:"storageQuotaInGiB,omitempty"`
+	StorageQuotaInGiB          int                            `json:"storageQuotaInGiB,omitempty"`
+	CephResourcesConfigMapRef  corev1.LocalObjectReference    `json:"cephResourcesConfigMapRef,omitempty"`
+	StorageClasses             []StorageClassSpec             `json:"storageClasses,omitempty"`
+	VolumeSnapshotClasses      []VolumeSnapshotClassSpec      `json:"volumeSnapshotClasses,omitempty"`
+	VolumeGroupSnapshotClasses []VolumeGroupSnapshotClassSpec `json:"volumeGroupSnapshotClasses,omitempty"`
```
umangachapagain, line 50: It'd be helpful if these fields

- supported patch operations, so that users can patch individual fields rather than updating the entire spec.
- supported ServerSideApply to build automation.
- included validations to ensure that array items are unique.
- provided labelSelector instead of name, to distribute a group of resources.

These could be discussed and added in a separate PR, but I think these would be useful features that would enable many usecases. @nb-ohad thoughts?

leelavg: > supported patch operations
- if you mean by SMP I don't think CRDs can do that (only Aggregated APIs) iirc. Merge & JsonPatch is supported by default anyways

> ServerSideApply to build automation.
- good one, we could certainly do that

> included validations
- yes, we want to and mostly in a different PR

> provided labelSelector instead of name,
- iirc, this was discussed in one of the meetings and wasn't chosen as this creates a two step operations and the number of SCs would be in single digits, open for any discussions though

thanks for the suggestions! will await Ohad's reply as well.

api/v1alpha1/storageconsumer_types.go
nb-ohad: @leelavg 
`omitempty` is about serialization of default values. `Optional` is a validation rule

api/v1alpha1/storageconsumer_types.go
nb-ohad: @umangachapagain Can you please create an RFE issue on this repo for each feature you are interested in with justification and a use case?


> provided labelSelector instead of name,

This is undesired, we don't want to group distribution of multiple entities under a single key. We intentionally decided to use a name as it can only identify a single CR at a single point in time (as opposed to UUID which identify a fixed single CR, or a selector which can identify a group of CRs at a single point in time)

api/v1alpha1/storageconsumer_types.go
```diff
@@ -42,7 +43,23 @@ type StorageConsumerSpec struct {
 	Enable bool `json:"enable,omitempty"`
 	// StorageQuotaInGiB describes quota for the consumer
 	// +optional
-	StorageQuotaInGiB int `json:"storageQuotaInGiB,omitempty"`
+	StorageQuotaInGiB          int                            `json:"storageQuotaInGiB,omitempty"`
+	CephResourcesConfigMapRef  corev1.LocalObjectReference    `json:"cephResourcesConfigMapRef,omitempty"`
```
nb-ohad, line 47: The name here does not fit the intent of the configmap.
The content of the config map is not directly related to ceph resources but more to reconcile ownership.
Also, this field is optional be definition and we should add an +optional comment above it

api/v1alpha1/storageconsumer_types.go
umangachapagain: "we don't want to group distribution of multiple entities under a single key". Why not?

I am not requesting for grouping of multiple types under same label set. I am asking for grouping of resources of same type that needs to be distributed to same client. Why should I be forced to name each one explicitly?
"Distribute all StorageClasses with LabelA to ClientA. Distribute all StorageClasses with LabelB to ClientB. " seems like a reasonable request.

api/v1alpha1/storageconsumer_types.go
nb-ohad, line 90: There is no point in having this here anymore, what we want is to reflect the configmap we are using

api/v1alpha1/storageconsumer_types.go
```diff
@@ -100,6 +121,18 @@ type ClientStatus struct {
...
+type RadosNamespaceStatus struct {
+	Name string `json:"name"`
+}
+
+type SubVolumeGroupStatus struct {
+	Name string `json:"name"`
+}
+
+type ClientProfileStatus struct {
+	Name string `json:"name"`
```
nb-ohad, line 133: See last comment about us dropping this

nb-ohad (changes requested):

api/v1alpha1/storageconsumer_types.go
nb-ohad: I am not saying it will not be handy, but at the same time it does not serve any of agreed UX designs. The current UX is build upon selecting of individual entities upon creation of the entity or upon creation of the consumer (we support both)

I am not saying we should not consider that as a future enhancement, but I do want a justification that is more then just "it be cool to have such flexibility".  You need to remember that we are talking about a resource space where we anticipate a single-digit count of entities (of each type), not hundreds or thousands of entities (which is the case where labels become an important tool)

api/v1alpha1/storageconsumer_types.go
leelavg: removed.

api/v1alpha1/storageconsumer_types.go
leelavg: changed to `OwningResourcesConfigMapRef`

api/v1alpha1/storageconsumer_types.go
leelavg: removed.

api/v1alpha1/storageconsumer_types.go
leelavg: removed.

PR #3056, opened 2025-02-24
DFBUGS-1654: [release-4.18] Delete the succeeded pods with duplicate tolerations to avoid alert

malayparida2000:
PrometheusDuplicateTimestamps alert is generated due to the presence of duplicate tolerations on the osd-prepare job pods & osd-key-rotation cronjob pods. Although the root of the issue has been fixed with another fix, the existing succeeded pods need to be cleaned up to stop the alert.

PR #3065, opened 2025-02-26
DFBUGS-1702: [release-4.16] Delete the succeeded pods with duplicate tolerations to avoid alert

malayparida2000:
PrometheusDuplicateTimestamps alert is generated due to the presence of duplicate tolerations on the osd-prepare job pods & osd-key-rotation cronjob pods. Although the root of the issue has been fixed with another fix, the existing succeeded pods need to be cleaned up to stop the alert.
Manual backport of https://github.com/red-hat-storage/ocs-operator/pull/3056

### DIFF
--- a/controllers/storagecluster/cephcluster.go
+++ b/controllers/storagecluster/cephcluster.go
@@ -94,6 +94,13 @@ var (
 	testSkipPrometheusRules = false
 )
 
+var deletedSucceededPodsWithDuplicateTolerations bool
+
+const (
+	osdPrepareLabelSelector     = "rook-ceph-osd-prepare"
+	osdKeyRotationLabelSelector = "rook-ceph-osd-key-rotation"
+)
+
 func arbiterEnabled(sc *ocsv1.StorageCluster) bool {
 	return sc.Spec.Arbiter.Enable
 }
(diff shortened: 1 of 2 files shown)


controllers/storagecluster/cephcluster.go
```diff
@@ -114,6 +121,24 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
...
 
+	// The deletion function needs to run only once successfully on the cluster
+	if !deletedSucceededPodsWithDuplicateTolerations && !sc.Spec.ExternalStorage.Enable {
+		// delete the osd-prepare job completed pods
+		err = r.deleteSucceededPodsWithDuplicateTolerations(map[string]string{"app": osdPrepareLabelSelector}, sc.Namespace)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		// if cluster wide or deviceSet encryption is true, delete the osd-key-rotation cronjob completed pods
+		if sc.Spec.Encryption.ClusterWide {
```
iamniting, line 132: Don't we need to check for the other encryption as we are doing in the original PR?

iamniting (changes requested):

controllers/storagecluster/cephcluster.go
malayparida2000: Deviceset wise encryption was not possible pre 4.18, it was only added in 4.18.

malayparida2000: I will add another field to this, thanks for pointing it out. Although the sc.Spec.Encryption.Enable has been deprecated since long but would be safe to add that here.

malayparida2000: Updated now.

iamniting (approved):

PR #3066, opened 2025-02-26
Allow specifying poolSpec for metadata pools of cephFS and cephObjectStore

malayparida2000:
This change enables configuring poolSpec for metadata pool of default
cephFS & cephObjectStore through the StorageCluster CR.
The poolSpec is passed directly to the Ceph CRs, with only necessary
fields validated. Other fields remain untouched.
Ref-https://issues.redhat.com/browse/RHSTOR-7068


api/v1/storagecluster_types.go
```diff
@@ -289,6 +291,8 @@ type ManageCephObjectStores struct {
 	// +kubebuilder:validation:MaxLength=253
 	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$
 	StorageClassName string `json:"storageClassName,omitempty"`
+	// MetadataPoolSpec specifies the pool specification for the default cephFS metadata pool
```
travisn, line 294:
```suggestion
	// MetadataPoolSpec specifies the pool specification for the default cephObjectStore metadata pool
```

controllers/storagecluster/cephobjectstores.go
```diff
@@ -168,13 +168,8 @@ func (r *StorageClusterReconciler) newCephObjectStoreInstances(initData *ocsv1.S
...
 				PreservePoolsOnDelete: false,
-				DataPool:              initData.Spec.ManagedResources.CephObjectStores.DataPoolSpec, // Pass the poolSpec from the storageCluster CR
-				MetadataPool: cephv1.PoolSpec{
-					DeviceClass:        initData.Status.DefaultCephDeviceClass,
-					EnableCrushUpdates: true,
-					FailureDomain:      initData.Status.FailureDomain,
-					Replicated:         generateCephReplicatedSpec(initData, poolTypeMetadata),
-				},
+				DataPool:              initData.Spec.ManagedResources.CephObjectStores.DataPoolSpec,     // Pass the poolSpec from the storageCluster CR
+				MetadataPool:          initData.Spec.ManagedResources.CephObjectStores.MetadataPoolSpec, // Pass the poolSpec from the storageCluster CR
```
travisn, line 172: We don't need to call `setDefaultMetadataPoolSpec()` to add the failure domain and other needed settings to the spec? Or where does that happen for this pool and the filesystem metadata pool?

controllers/storagecluster/cephcluster.go
```diff
@@ -1483,6 +1483,26 @@ func isEncrptionSettingUpdated(clusterWideEncrytion bool, existingDeviceSet []ro
 	return false
 }
 
+// setDefaultMetadataPoolSpec sets the common pool spec for all metadata pools as necessary
+func setDefaultMetadataPoolSpec(poolSpec *rookCephv1.PoolSpec, sc *ocsv1.StorageCluster) {
```
travisn, line 1487: I only see this method being called for the `.nfs` and `.mgr` pools, which don't have any spec exposed in the storageCluster CR, right? If that's the intended case (and not for the filesystem and object store), seems like this method can be simplified. For example, device class and failure domain would always be `""`, so no need for those checks on lines 1489 and 1492.

api/v1/storagecluster_types.go
malayparida2000: Updated

controllers/storagecluster/cephobjectstores.go
malayparida2000: Now added

controllers/storagecluster/cephcluster.go
malayparida2000: We will use setDefaultMetadataPoolSpec() for all the metadata pools

malayparida2000:
@travisn The PR was actually half cooked to be honest & sorry for that.
Now updated the changes correctly.

malayparida2000:
> LGTM. Can you just confirm if testing is already done?
> 
> btw is #2980 waiting for this to merge first? It might be convenient to test these changes together.

I have not tested this PR individually yet. But we had merged https://github.com/red-hat-storage/ocs-operator/pull/2979 on the similar pattern and I have tested that one. as the CI passes we can take that as good enough for now. I plan to test complete testing after this merged and work on #2980 is complete.

travisn:
/lgtm
sounds good to test them together soon

PR #3067, opened 2025-02-26
Allow specifying poolSpec parameters for non-resilient cephBlockPools

malayparida2000:
Ref-https://issues.redhat.com/browse/RHSTOR-7069

This change enables configuring the parameters of the non-resilient
cephBlockPools through the StorageCluster CR.
The parameters is passed directly to the pool's poolSpec, with
only necessary fields validated. Other fields remain untouched.


travisn (commented):
/lgtm

PR #3074, opened 2025-03-10
provider-server: enable csi drivers for ocs-client-op

rohan47:
Sends the drivers to enable based on Storage Cluster spec in the `GetStorageConfig` response (`SystemAttributes`) and in `StatusReporterResponse` (as the required Hash)


rohan47:
Fixing tests

rewantsoni:
LGTM

services/provider/server/server.go
```diff
@@ -252,14 +252,22 @@ func (s *OCSProviderServer) GetStorageConfig(ctx context.Context, req *pb.Storag
...
 
+		systemAtteributes := &pb.SystemAttributes{
+			SystemInMaintenanceMode: inMaintenanceMode,
+			MirrorEnabled:           isConsumerMirrorEnabled,
+			DesiredRbdDriver:        &pb.RbdDriverRequirements{},
+			DesiredCephFsDriver:     &pb.CephFsDriverRequirements{},
+		}
+
+		if storageCluster.Spec.NFS != nil && storageCluster.Spec.NFS.Enable {
+			systemAtteributes.DesiredNfsDriver = &pb.NfsDriverRequirements{}
```
leelavg, line 263: what is the de-serialized form for w/ & w/o the message? Is it going to be compared to `nil` at client side? possible to check it as I forgot how it'll look like.

services/provider/server/server.go
```diff
@@ -1076,12 +1084,15 @@ func (s *OCSProviderServer) ReportStatus(ctx context.Context, req *pb.ReportStat
 		return nil, status.Errorf(codes.Internal, "Failed to get mirroring status for consumer.")
 	}
 
+	isNfsEnabled := storageCluster.Spec.NFS.Enable
```
leelavg, line 1087: iirc Ohad mentioned we can always deploy NFS, possible to reconfirm?

services/provider/server/server.go
rohan47: Without the message it will be de-serialized as `nil` on the client side

services/provider/server/server.go
rohan47: You mean enable NFS and deploy NFS CSI driver always or just NFS CSI drivers always?

rohan47:
Closing this PR in favour of https://github.com/red-hat-storage/ocs-operator/pull/3195

PR #3080, opened 2025-03-17
Set mon_target_pg_per_osd to 400 for better performance from higher PG count

malayparida2000:
The mon_target_pg_per_osd parameter determines the recommended total PG count and PG-per-OSD ratio. The default value is 100, but setting it to 400 improves performance by increasing parallelism through a higher PG count. 
On existing clusters this can trigger data movement and rebalancing.
https://issues.redhat.com/browse/DFBUGS-2249

### DIFF
--- a/controllers/storagecluster/cephcluster.go
+++ b/controllers/storagecluster/cephcluster.go
@@ -570,6 +570,11 @@ func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, kmsConfigMap *co
 			UpgradeOSDRequiresHealthyPGs: sc.Spec.ManagedResources.CephCluster.UpgradeOSDRequiresHealthyPGs,
 			// if resource profile change is in progress, then set this flag to false
 			ContinueUpgradeAfterChecksEvenIfNotHealthy: sc.Spec.ResourceProfile == sc.Status.LastAppliedResourceProfile,
+			CephConfig: map[string]map[string]string{
+				"global": {
+					"mon_target_pg_per_osd": "400",
+				},
+			},
 		},
 	}
 


controllers/storagecluster/cephconfig.go
```diff
@@ -66,6 +67,30 @@ func (obj *ocsCephConfig) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.S
 		Data: rookConfigOverrideData,
 	}
 	_, err := ctrl.CreateOrUpdate(context.Background(), r.Client, rookConfigOverrideCM, func() error {
+		// mon_target_pg_per_osd=200 added only during new CM creation, as setting it on existing clusters can cause data movement.
```
travisn, line 70: To keep this simple, what if the new setting is only added to the `defaultRookConfig`? It's not critical for upgraded clusters, right?

Alternatively, instead of using the configmap at all, this setting could be the first one that we add to the CephCluster spec under the `CephConfig` setting. I'd suggest this for the preferred approach for all the ceph settings that don't need to be overridable by the user, and also are not settings that affect the mons at first cluster creation. This setting is just for the PG auto scaler, so it's fine to set later.

malayparida2000:
> Nice and simple! As in the other thread, I'm just concerned about data movement if we always apply to upgraded clusters.

Yeah, My suggestion is let's apply it to the upgraded cluster for now. This will help me & the QE test the upgrade scenario to find out the consquences & results. If we find too many problems we will skip out on upgraded clusters at the end. WDYT?

travisn:
> > Nice and simple! As in the other thread, I'm just concerned about data movement if we always apply to upgraded clusters.
> 
> Yeah, My suggestion is let's apply it to the upgraded cluster for now. This will help me & the QE test the upgrade scenario to find out the consquences & results. If we find too many problems we will skip out on upgraded clusters at the end. WDYT?

Sounds fine to go ahead and merge if it simplifies the upgrade testing, but let's open an issue and have a fix ready for not setting it by default during upgrades, after we confirm the upgrade behavior, particularly for clusters that have a lot of data.

malayparida2000:
> > > Nice and simple! As in the other thread, I'm just concerned about data movement if we always apply to upgraded clusters.
> > 
> > 
> > Yeah, My suggestion is let's apply it to the upgraded cluster for now. This will help me & the QE test the upgrade scenario to find out the consquences & results. If we find too many problems we will skip out on upgraded clusters at the end. WDYT?
> 
> Sounds fine to go ahead and merge if it simplifies the upgrade testing, but let's open an issue and have a fix ready for not setting it by default during upgrades, after we confirm the upgrade behavior, particularly for clusters that have a lot of data.

Ack, I will make a fix ready for enabling it only for new clusters, skipping upgraded clusters. We can just reutilize the same Bug which we will use to backport this fix. But I will keep the fix ready just in case we need to apply it quickly.

travisn (approved):
/lgtm

IceManny2:
Hello @BlaineEXE and @travisn ,

I think exceeding 250 PGs per OSD is a bad idea.

Please see these tickets I have opened today:
https://issues.redhat.com/browse/DFBUGS-2657
https://issues.redhat.com/browse/DFBUGS-2665
https://issues.redhat.com/browse/ODFRFE-10

The Docs around setting and changing ratios is extremely poor, we just had a customer create a off-named Pool with 1 PG.  When this Pool grew to the size = to any OSD, the system went into a Health Error.  Even spamming the system with 450 PGs per OSD would have helped this as it was an additional off-named Pool.

It would be better if our Docs socialized the need for a target ratio instead of spamming the system with PGs.

Given that Rook is outside of Ceph, it would better if Rook looked at capacity and adjusted the ratios to match reality and not require the Admin to do the Math and tweak the ratios.

/Manny

travisn:
> The Docs around setting and changing ratios is extremely poor, we just had a customer create a off-named Pool with 1 PG. When this Pool grew to the size = to any OSD, the system went into a Health Error. Even spamming the system with 450 PGs per OSD would have helped this as it was an additional off-named Pool.

Why was there a pool with 1 PG? I've only seen that in case the device class is not set, and the PG autoscaler refuses to scale it. Obviously 1 PG is a bad idea, not sure what that has to do with raising the PG count.

> It would be better if our Docs socialized the need for a target ratio instead of spamming the system with PGs.

Can you elaborate? I don't see how increasing the PGs is "spamming the system". I've only seen feedback that customers benefited from increasing their PGs beyond our defaults. Now we are only increasing the defaults to get the increased performance by default.

> Given that Rook is outside of Ceph, it would better if Rook looked at capacity and adjusted the ratios to match reality and not require the Admin to do the Math and tweak the ratios.

This is the job of the PG autoscaler, it has much more context than Rook. How could Rook calculate it better than the PG autoscaler?

IceManny2:
@travisn ,  The Auto Scaler was never designed to set the ratio based on capacity.  The Auto Scaler expects something external to assert the ratios or assert the "bulk" flag.

/MC

travisn:
Ultimately it depends on the expected usage of the cluster, which involves user input. But unless there is something broken with the current approach for 4.19, I'd suggest let's open a new issue and discuss for 4.20.

PR #3081, opened 2025-03-17
provider: get the radosnamespace name from consumer configmap

rewantsoni:
Get the radosnamespace for a consumer from the consumer configmap 

Review only the last commit

Depends On: https://github.com/red-hat-storage/ocs-operator/pull/3062


umangachapagain (approved):

leelavg (approved):
if the naming of the key changes, can fix in separate PR.

PR #3082, opened 2025-03-17
cephblockpool: stop reconciling mirroring field

rewantsoni:
Remove the deployment mode annotation ref from cephblockpool reconcile as blockpool mirroring will be handled by the mirroring controller.


umangachapagain (changes requested):
We prefer keeping the unit-test changes in the same commit as the change that breaks it. Please squash the commits and we can merge.

umangachapagain (approved):

PR #3083, opened 2025-03-18
Specify the poolSpec API field as a pointer to avoid unneceesary empty fields in storageCluster yaml output

malayparida2000:


iamniting (approved):
Changes looks good to me. @travisn Can you also pls take a look?

/hold
/lgtm

travisn:
The changes look good, but I'm not clear about the behavior during upgrades when changing this to a pointer. I'm guessing that upgraded clusters will just retain all the empty fields and never have the nil pools, but no failures during upgrade. Only new clusters will have the nil pool specs by default. Is that what you observed?

malayparida2000:
@travisn I am unable to properly stimulate a upgrade for this to test out the scenario. Will check the upgrade ones it's part of the build.

travisn:
> @travisn I am unable to properly stimulate a upgrade for this to test out the scenario. Will check the upgrade ones it's part of the build.

Sounds good thanks

malayparida2000:
I also raised https://github.com/red-hat-storage/ocs-operator/pull/3085, Do take a look. Thanks

PR #3085, opened 2025-03-18
Update ManageCephNonResilientPools & ManageCephRBDMirror API fields so they are not in the CR output if not specified

malayparida2000:


iamniting (approved):
Changes looks good to me. @travisn Can you also pls take a look?

/hold
/lgtm

travisn (commented):
Just the same question as #3083 about upgrades
/lgtm
/unhold

malayparida2000:
> Just the same question as #3083 about upgrades /lgtm /unhold

Will test both once they are available in the build, by doing 4.18 to 4.19 upgrade.

PR #3087, opened 2025-03-18
provider: enhance GetStorageConfig to replace GetStorageClaimInfo

rewantsoni:
PR does the following:

1. Removes unused code and Refactors the GetStorageConfig RPC
2. Send ClientProfile from the GetStorageConfig RPC
3. Send StorageClass from the GetStorageConfig RPC
4. Send Volume[Group]SnapshotClass from the GetStorageConfig RPC
5. Send CephClients from the GetStorageConfig RPC

Day 2 StorageClass and Day 2 [Group]SnapshotClass will be a part of a different PR


controllers/storagecluster/provider_server.go
```diff
@@ -2,6 +2,7 @@ package storagecluster
 
 import (
 	"fmt"
+	//"github.com/red-hat-storage/ocs-operator/v4/services/provider/server"
```
leelavg, line 5: remove commented out imports?

leelavg (commented):
will this be reworked after the merge of other 2 PRs?

services/provider/server/server.go
```diff
@@ -341,77 +345,314 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
+			labels = map[string]string{
+				ramenDRStorageIDLabelKey: cephFsStorageID,
+			}
+			parameters = map[string]string{
+				"clusterID": cephfsClientProfileName,
+				"csi.storage.k8s.io/snapshotter-secret-name":      cephFsProvisionerSecretName,
+				"csi.storage.k8s.io/snapshotter-secret-namespace": consumerResource.Status.Client.OperatorNamespace,
+			}
+		}
+		//TODO: Day-2 snapshotclass and nfsSnapshot
```
Madhu-1, line 589: comment is not correct

services/provider/server/server.go
```diff
@@ -341,77 +345,314 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
+				ramenDRStorageIDLabelKey: cephFsStorageID,
+			}
+			parameters = map[string]string{
+				"clusterID": cephfsClientProfileName,
+				"csi.storage.k8s.io/group-snapshotter-secret-name":      cephFsProvisionerSecretName,
+				"csi.storage.k8s.io/group-snapshotter-secret-namespace": consumerResource.Status.Client.OperatorNamespace,
+				"fsName": storagecluster.GenerateNameForCephFilesystem(storageCluster),
+			}
+		}
+		//TODO: Day-2 groupSnapshotclass and nfsSnapshot
```
Madhu-1, line 638: comment is not correct

controllers/storagecluster/storageclasses.go
nb-ohad, line 1: Does it even make sense to keep this file?

controllers/storagecluster/volumegroupsnapshotterclasses.go
nb-ohad, line 1: Does it even make sense to keep this file?

controllers/storagecluster/volumesnapshotterclasses.go
nb-ohad, line 1: Does it even make sense to keep this file?

services/provider/server/server.go
```diff
@@ -56,7 +60,6 @@ import (
 
 const (
 	TicketAnnotation          = "ocs.openshift.io/provider-onboarding-ticket"
-	ProviderCertsMountPoint   = "/mnt/cert"
```
nb-ohad, line 59: Why are er moving this from here?

services/provider/server/server.go
```diff
@@ -341,77 +345,314 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
-		Name: "monitor-endpoints",
+		Name: consumerResource.Status.Client.Name,
 		Data: mustMarshal(&csiopv1a1.CephConnectionSpec{Monitors: monIps}),
 	})
 
-	scMon := &v1.Secret{}
-	// Secret storing cluster mon.admin key, fsid and name
-	err = s.client.Get(ctx, types.NamespacedName{Name: monSecret, Namespace: s.namespace}, scMon)
+	consumerConfigMap := &v1.ConfigMap{}
+	consumerConfigMap.Name = consumerResource.Status.ResourceNameMappingConfigMap.Name
```
nb-ohad, line 360: What happens if the name is an empty string?

services/provider/server/server.go
```diff
@@ -341,77 +345,314 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
+		if kernelMountOptions == nil {
+			kernelMountOptions = map[string]string{}
+		}
+		parts := strings.Split(option, "=")
+		kernelMountOptions[parts[0]] = parts[1]
+	}
+	rbdClientProfileName := consumerConfigMap.Data["rbd-client-profile"]
+	cephfsClientProfileName := consumerConfigMap.Data["cephfs-client-profile"]
+	radosNamespace := consumerConfigMap.Data["rados-namespace-name"]
+	svgName := consumerConfigMap.Data["svg-name"]
```
nb-ohad, line 383: Please use the wrapper that provide typed-const access to keys. 
This comment also applies to all other lines below that directly access the config's data map using string keys

services/provider/server/server.go
```diff
@@ -341,77 +345,314 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
+				&csiopv1a1.ClientProfileSpec{
+					CephFs: &csiopv1a1.CephFsConfigSpec{
+						SubVolumeGroup:     svgName,
+						KernelMountOptions: kernelMountOptions,
+						RadosNamespace:     ptr.To(svgName),
+					},
+					CephConnectionRef: corev1.LocalObjectReference{Name: consumerResource.Status.Client.Name},
+				},
+			),
+		},
```
nb-ohad, line 410: 1. This will always produce too different client profiles, which does not align with our design where we will only produce a single client profile if the name are the same. 
2. This code does not address the NFS client profile handling

services/provider/server/server.go
```diff
@@ -341,77 +345,314 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
+	for i := 0; i < len(consumerResource.Spec.StorageClasses); i++ {
+		storageClassName := consumerResource.Spec.StorageClasses[i].Name
+
+		provisioner := ""
+		annotations := map[string]string{}
+		labels := map[string]string{}
+		parameters := map[string]string{}
+
+		//TODO: keyRotation and defaultStorageClass annotation
+		if storageClassName == storagecluster.GenerateNameForCephBlockPoolSC(storageCluster) {
```
nb-ohad, line 443: This if-else approach is unreasonable for so many built in *classes. Can we move to a template based approach where the template is chosen based on the name?

services/provider/server/server.go
```diff
@@ -341,77 +345,314 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
+					Provisioner:          provisioner,
+					Parameters:           parameters,
+				}),
+			},
+		)
 	}
 
-	if port < 0 {
-		return nil, fmt.Errorf("mgr pod port is empty")
+	for i := 0; i < len(consumerResource.Spec.VolumeSnapshotClasses); i++ {
```
nb-ohad, line 561: This if-else approach is unreasonable for so many built in *classes. Can we move to a template based approach where the template is chosen based on the name?

services/provider/server/server.go
```diff
@@ -341,77 +345,314 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
 	}
 
-	extR = append(extR, &pb.ExternalResource{
-		Name: "monitoring-endpoint",
-		Kind: "CephCluster",
-		Data: mustMarshal(map[string]string{
-			"MonitoringEndpoint": mgrPod.Status.HostIP,
-			"MonitoringPort":     strconv.Itoa(int(port)),
-		})})
+	for i := 0; i < len(consumerResource.Spec.VolumeGroupSnapshotClasses); i++ {
```
nb-ohad, line 608: This if-else approach is unreasonable for so many built in *classes. Can we move to a template based approach where the template is chosen based on the name?

controllers/util/provider.go
```diff
@@ -19,6 +19,10 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 )
 
+const (
+	ProviderCertsMountPoint = "/mnt/cert"
+)
```
nb-ohad, line 24: What is the reason for moving this here? I believe this is only used from the provide server package so it should live there

nb-ohad (changes requested):

nb-ohad:
@rewantsoni This PR need to keep on serving GetStorageClaimInfo but at the same time not duplicate the implementation

controllers/storagecluster/storageclasses.go
rewantsoni: external mode uses it to create Classes

controllers/storagecluster/volumegroupsnapshotterclasses.go
rewantsoni: external mode uses it to create Classes

controllers/storagecluster/volumesnapshotterclasses.go
rewantsoni: external mode uses it to create Classes

services/provider/server/server.go
```diff
@@ -341,77 +346,283 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
-	for i := range mgrPod.Spec.Containers {
-		container := &mgrPod.Spec.Containers[i]
-		if container.Name == "mgr" {
-			for j := range container.Ports {
-				if container.Ports[j].Name == "http-metrics" {
-					port = container.Ports[j].ContainerPort
-				}
-			}
+	rbdClientProfileName := consumerConfig.GetRbdClientProfileName()
+	cephFsClientProfileName := consumerConfig.GetCephFsClientProfileName()
```
nb-ohad, line 394: Please move this to line 395 just before it is used

services/provider/server/server.go
nb-ohad, line 1: The `getExternalResources` became so bloated it is almost impossible to keep the entire context in mind while reading it. Can we have a function for each resource (or resource type) that need to be added and then have this func just call all these other funcs to add their resources to the list?

services/provider/server/server.go
```diff
@@ -341,77 +346,283 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
+	scMap[util.GenerateNameForCephFilesystemSC(storageCluster)] = func() *storagev1.StorageClass {
+		return util.NewDefaultCephFsStorageClass(
+			consumerConfig.GetRbdClientProfileName(),
+			util.GenerateNameForCephFilesystem(storageCluster.Name),
+			consumerConfig.GetCsiCephFsProvisionerSecretName(),
+			consumerConfig.GetCsiCephFsNodeSecretName(),
+			consumerResource.Status.Client.OperatorNamespace,
+			drCephFsId,
+		)
+	}
```
nb-ohad, line 532: Lose all of the spaces. They don't add anything beside beefing the line count

nb-ohad (changes requested):

controllers/util/k8sutil.go
```diff
@@ -179,17 +180,91 @@ func AnnotationIndexFieldFunc(obj client.Object) []string {
 	return maps.Keys(obj.GetAnnotations())
 }
 
-func GenerateNameForNonResilientCephBlockPoolSC(initData *ocsv1.StorageCluster) string {
-	if initData.Spec.ManagedResources.CephNonResilientPools.StorageClassName != "" {
-		return initData.Spec.ManagedResources.CephNonResilientPools.StorageClassName
+func GetTopologyConstrainedPools(storageCluster *ocsv1.StorageCluster) string {
```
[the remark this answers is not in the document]
rewantsoni: No that is for external mode this is different

controllers/storagecluster/volumesnapshotterclasses.go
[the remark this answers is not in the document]
rewantsoni: We cannot do that as the volumesnapshotterclasses struct is added to ensureCreated for external mode

controllers/storagecluster/volumegroupsnapshotterclasses.go
```diff
@@ -2,67 +2,52 @@ package storagecluster
...
+			instance.Namespace,
+			"rook-csi-rbd-provisioner",
+			instance.Namespace,
+			util.GenerateNameForCephBlockPool(instance.Name),
+			"",
+		),
+		reconcileStrategy: ReconcileStrategy(instance.Spec.ManagedResources.CephBlockPools.ReconcileStrategy),
 	}
+	gsc.groupSnapshotClass.Name = util.GenerateNameForGroupSnapshotClass(instance, util.RbdGroupSnapshotter)
+	return gsc
```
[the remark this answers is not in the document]
rewantsoni: We cannot do that as the volumesnapshotterclasses struct is added to ensureCreated for external mode

controllers/storagecluster/generate.go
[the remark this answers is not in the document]
rewantsoni: moved most of it at right places, but the problem is some shouldn't have been in this file in the first place, so moved them to near the call site

controllers/storagecluster/external_resources.go
```diff
@@ -346,22 +358,79 @@ func (r *StorageClusterReconciler) createExternalStorageClusterResources(instanc
...
+					storageClass: util.NewDefaultCephFsStorageClass(
+						instance.Namespace,
+						util.GenerateNameForCephFilesystem(instance.Name),
+						"rook-csi-cephfs-provisioner",
+						"rook-csi-cephfs-node",
+						instance.Namespace,
+						"",
+					),
+					reconcileStrategy: ReconcileStrategy(instance.Spec.ManagedResources.CephFilesystems.ReconcileStrategy),
+					isClusterExternal: instance.Spec.ExternalStorage.Enable,
```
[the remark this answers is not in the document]
rewantsoni: Yes it is true, but createStorageClasses is using it, I didn't want to touch that code

services/provider/server/server.go
```diff
@@ -330,104 +334,421 @@ func newScheme() (*runtime.Scheme, error) {
...
+	for i := 0; i < len(consumer.Spec.StorageClasses); i++ {
+		storageClassName := consumer.Spec.StorageClasses[i].Name
+		scGen := scMap[storageClassName]
+		if scGen != nil {
+			extR = append(extR, &pb.ExternalResource{
+				Kind: "StorageClass",
+				Name: storageClassName,
+				Data: mustMarshal(scGen()),
+			})
+		}
```
nb-ohad, line 637: Please add a log message in case we encountered an unexpected sc name

services/provider/server/server.go
```diff
@@ -330,104 +334,421 @@ func newScheme() (*runtime.Scheme, error) {
...
+	for i := 0; i < len(consumer.Spec.VolumeSnapshotClasses); i++ {
+		snapshotClassName := consumer.Spec.VolumeSnapshotClasses[i].Name
+		vscGen := vscMap[snapshotClassName]
+		if vscGen != nil {
+			extR = append(extR, &pb.ExternalResource{
+				Kind: "VolumeSnapshotClass",
+				Name: snapshotClassName,
+				Data: mustMarshal(vscGen()),
+			})
+		}
```
nb-ohad, line 685: Please add a log message in case we encountered an unexpected sc name

services/provider/server/server.go
```diff
@@ -330,104 +334,421 @@ func newScheme() (*runtime.Scheme, error) {
...
+	for i := 0; i < len(consumer.Spec.VolumeGroupSnapshotClasses); i++ {
+		groupSnapshotClassName := consumer.Spec.VolumeGroupSnapshotClasses[i].Name
+		vgscGen := vgscMap[groupSnapshotClassName]
+		if vgscGen != nil {
+			extR = append(extR, &pb.ExternalResource{
+				Kind: "VolumeGroupSnapshotClass",
+				Name: groupSnapshotClassName,
+				Data: mustMarshal(vgscGen()),
+			})
+		}
```
nb-ohad, line 727: Please add a log message in case we encountered an unexpected sc name

services/provider/server/server.go
```diff
@@ -479,61 +805,63 @@ func (s *OCSProviderServer) getExternalResources(ctx context.Context, consumerRe
...
 	noobaaOperatorSecret.Namespace = s.namespace
 
-	if err := s.client.Get(ctx, client.ObjectKeyFromObject(noobaaOperatorSecret), noobaaOperatorSecret); err != nil {
-		if kerrors.IsNotFound(err) {
-			// ignoring because it is a provider cluster and the noobaa secret does not exist
-			return extR, nil
-
+	if err := s.client.Get(ctx, client.ObjectKeyFromObject(noobaaOperatorSecret), noobaaOperatorSecret); client.IgnoreNotFound(err) != nil {
+		return fmt.Errorf("failed to get %s secret. %v", noobaaOperatorSecret.Name, err)
+	} else if !kerrors.IsNotFound(err) {
```
nb-ohad, line 824: There is no need in moving the entire body into an else
```suggestion
	if err := s.client.Get(ctx, client.ObjectKeyFromObject(noobaaOperatorSecret), noobaaOperatorSecret); kerrors.IsNotFound(err) != nil {
		return nil
} else err != nil {
		return fmt.Errorf("failed to get %s secret. %v", noobaaOperatorSecret.Name, err)
	} 
```

controllers/storagecluster/external_resources.go
```diff
@@ -345,23 +357,81 @@ func (r *StorageClusterReconciler) createExternalStorageClusterResources(instanc
...
+				scc = StorageClassConfiguration{
+					storageClass: util.NewDefaultNonResilientRbdStorageClass(
+						instance.Namespace,
+						topologyConstrainedPools,
+						"rook-csi-rbd-provisioner",
+						"rook-csi-rbd-node",
+						instance.Namespace,
+						"",
+						instance.GetAnnotations()[defaults.KeyRotationEnableAnnotation] == "false",
+					),
```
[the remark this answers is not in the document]
rewantsoni: it was missing added it

leelavg:
pls use latest client-op bundle quay.io/ocs-dev/ocs-client-operator-bundle:51f7ce3

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -300,8 +300,8 @@ func (r *StorageConsumerReconciler) reconcileEnabledPhases() (reconcile.Result,
 		// A provider cluster already has a NooBaa system and does not require a NooBaa account
 		// to connect to a remote cluster, unlike client clusters.
 		// A NooBaa account only needs to be created if the storage consumer is for a client cluster.
-		clusterID := util.GetClusterID(r.ctx, r.Client, &r.Log)
-		if clusterID != "" && !strings.Contains(r.storageConsumer.Name, clusterID) {
+		//TODO: (rewant) This is a hack that we are using and this needs to go away, we shouldn't rely on consumer name
+		if r.storageConsumer.Name != defaults.LocalStorageConsumerName {
```
nb-ohad, line 304: I would prefer if we compare the local cluster id with the one on .Status.Client.ClusterID

services/provider/server/server.go
```diff
@@ -320,104 +324,465 @@ func newScheme() (*runtime.Scheme, error) {
...
+		}
+
+		if err := s.client.Get(ctx, client.ObjectKeyFromObject(cephUserSecret), cephUserSecret); err != nil {
+			return extR, fmt.Errorf("failed to get %s secret. %v", cephUserSecret, err)
 		}
+
+		secretContent := map[string]string{}
+		for key, value := range cephUserSecret.Data {
+			secretContent[key] = string(value)
+		}
```
nb-ohad, line 588: We should be sending `map[string]bytes` and not  `map[string]string`. We should fix the receiver side to handle `map[string]byte`
```suggestion
```

nb-ohad (changes requested):

PR #3091, opened 2025-03-21
api: add vrc to the consumer spec

rewantsoni:
update the consumer api to add VRC for distribution to clients


leelavg (approved):
approving based on earlier finalised discussions that storageconsumer will have the details of all the classes that needs to be distributed.

Madhu-1 (approved):
LGTM

iamniting (approved):

PR #3092, opened 2025-03-21
remove the storagesystem from the odfinfo configmap

iamniting:
storagesystem is depricated

Bug: https://issues.redhat.com/browse/DFBUGS-1880


rewantsoni (changes requested):
There is also storageSystem ref in storagecluster_controller_test.go and a few other places can you remove that as well?

PR #3093, opened 2025-03-21
mgr: disable mgr liveness and startup probe

parth-gr:
there is a bug in ceph https://tracker.ceph.com/issues/70571 until that is fixed diabling the probes

### DIFF
--- a/controllers/storagecluster/cephcluster.go
+++ b/controllers/storagecluster/cephcluster.go
@@ -546,6 +546,18 @@ func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, kmsConfigMap *co
 				rookCephv1.KeyMonitoring:   getCephClusterMonitoringLabels(*sc),
 				rookCephv1.KeyCephExporter: getCephClusterMonitoringLabels(*sc),
 			},
+			HealthCheck: rookCephv1.CephClusterHealthCheckSpec{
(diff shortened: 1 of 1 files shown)


iamniting (approved):
/hold
for @travisn review

travisn (commented):
/approve
/lgtm
This is clearly just a workaround for a ceph issue. Let's merge this to unblock the builds, and we need to follow up with the ceph fix and reverting this later...

travisn:
/hold
Shall we just merge this to the 4.19 branch instead of main?

agarwal-mudit:
@parth-gr why this is not backported to 4.19?

parth-gr:
> @parth-gr why this is not backported to 4.19?

We didn't  have a successful build, so the build automation didn't create it.

Should we create a manual backport as its a build blocker cc @agarwal-mudit @nik-redhat

Nikhil-Ladha:
We are still using the master branch, IIRC. So, how come this issue is still present in a new build?
Are we sure we had a build post this PR was merged?

parth-gr:
@nik-redhat  can you check with Elad, he mentioned it should be specifically present in release 4.19

parth-gr:
@agarwal-mudit this should already be in 4.19, by sync prs as it was created before ff

PR #3094, opened 2025-03-24
rgw: add readAffinity

sp98:
Add RgwReadAffinity settings to the cephobjectstore CR. Localize read mode is used in case stretch cluster or "zone" failure domain.


controllers/storagecluster/cephobjectstores.go
```diff
@@ -272,3 +283,12 @@ func getCephObjectStoreGatewayInstances(sc *ocsv1.StorageCluster) int {
 	}
 	return defaults.CephObjectStoreGatewayInstances
 }
+
+func isPortableOSD(sc *ocsv1.StorageCluster) bool {
```
malayparida2000, line 287: The helper func checks if any of the OSDs are portable, So instead of isPortableOSD I would suggest hasPortableOSD, Also as the result of the func is utilized for a -ve check we can optimize this function to be 
hasNonPortable OSD and return immidiately when we see a non-portable OSD

sp98: done.

malayparida2000 (approved):
/lgtm
/cc @iamniting

iamniting (approved):

PR #3096, opened 2025-03-24
unless explicitly mentioned infer providerApiServerService type from hostNetwork

leelavg:
only serving storage to external clients requires provider api server reach-ability which can be inferred from storagecluster hostNetwork field and in other scenarios clusterIP is used as default to serve local client.

### DIFF
--- a/api/v1/storagecluster_types.go
+++ b/api/v1/storagecluster_types.go
@@ -97,7 +97,8 @@ type StorageClusterSpec struct {
 	AllowRemoteStorageConsumers bool `json:"allowRemoteStorageConsumers,omitempty"`
 
 	// ProviderAPIServerServiceType Indicates the ServiceType for OCS Provider API Server Service.
-	// The supported values are NodePort or LoadBalancer. The default ServiceType is NodePort if the value is empty.
+	// The default ServiceType is derived from hostNetwork field.
+	// +kubebuilder:validation:Enum=ClusterIP;NodePort;LoadBalancer
 	ProviderAPIServerServiceType corev1.ServiceType `json:"providerAPIServerServiceType,omitempty"`
 
 	// EnableCephTools toggles on whether or not the ceph tools pod
--- a/controllers/storagecluster/provider_server.go
+++ b/controllers/storagecluster/provider_server.go
@@ -64,8 +64,11 @@ func (o *ocsProviderServer) ensureDeleted(r *StorageClusterReconciler, instance 
 
 	var finalErr error
 
+	apiServerService := &corev1.Service{}
+	apiServerService.Name = ocsProviderServerName
+	apiServerService.Namespace = instance.Namespace
 	for _, resource := range []client.Object{
-		GetProviderAPIServerService(instance),
+		apiServerService,
 		GetProviderAPIServerDeployment(instance),
 	} {
 		err := r.Client.Delete(r.ctx, resource)
@@ -128,18 +131,6 @@ func (o *ocsProviderServer) createDeployment(r *StorageClusterReconciler, instan
 
 func (o *ocsProviderServer) createService(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) (reconcile.Result, error) {
 
-	if instance.Spec.ProviderAPIServerServiceType != "" {
-		switch instance.Spec.ProviderAPIServerServiceType {
-		case corev1.ServiceTypeClusterIP, corev1.ServiceTypeLoadBalancer, corev1.ServiceTypeNodePort:
-		default:
-			err := fmt.Errorf("providerAPIServer only supports service of type %s, %s and %s",
-				corev1.ServiceTypeNodePort, corev1.ServiceTypeLoadBalancer, corev1.ServiceTypeClusterIP)
-			r.Log.Error(err, "Failed to create/update service, Requested ServiceType is", "ServiceType", instance.Spec.ProviderAPIServerServiceType)
-			return reconcile.Result{}, err
-		}
-
-	}
-
 	desiredService := GetProviderAPIServerService(instance)
 	actualService := &corev1.Service{
 		ObjectMeta: metav1.ObjectMeta{
@@ -170,7 +161,7 @@ func (o *ocsProviderServer) createService(r *StorageClusterReconciler, instance 
 
 	r.Log.Info("Service create/update succeeded")
 
-	switch instance.Spec.ProviderAPIServerServiceType {
+	switch desiredService.Spec.Type {
 	case corev1.ServiceTypeLoadBalancer:
 		endpoint := o.getLoadBalancerServiceEndpoint(actualService)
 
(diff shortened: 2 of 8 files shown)


controllers/storagecluster/provider_server.go
```diff
@@ -62,10 +62,16 @@ func (o *ocsProviderServer) ensureDeleted(r *StorageClusterReconciler, instance
...
+		serviceType = *instance.Spec.ProviderAPIServerServiceType
+	} else if instance.Spec.HostNetwork {
+		serviceType = corev1.ServiceTypeNodePort
+	}
 	var finalErr error
 
 	for _, resource := range []client.Object{
-		GetProviderAPIServerService(instance),
+		GetProviderAPIServerService(instance.Namespace, serviceType),
 		GetProviderAPIServerDeployment(instance),
```
rewantsoni, line 75: We don't need to explicitly delete deployment and services as storageCluster owns both. Should we remove this logic?

leelavg: can be done, I'm keeping the changes to the minimum until we test the whole flow atleast once and then start cleaning up.

rewantsoni:
LGTM

api/v1/storagecluster_types.go
```diff
@@ -97,8 +97,8 @@ type StorageClusterSpec struct {
 	AllowRemoteStorageConsumers bool `json:"allowRemoteStorageConsumers,omitempty"`
 
 	// ProviderAPIServerServiceType Indicates the ServiceType for OCS Provider API Server Service.
-	// The supported values are NodePort or LoadBalancer. The default ServiceType is NodePort if the value is empty.
-	ProviderAPIServerServiceType corev1.ServiceType `json:"providerAPIServerServiceType,omitempty"`
+	// The default ServiceType is derived from hostNetwork field.
+	ProviderAPIServerServiceType *corev1.ServiceType `json:"providerAPIServerServiceType,omitempty"`
```
nb-ohad, line 101: Why do we need a pointer when service type can be set to an empty string?

controllers/storagecluster/provider_server.go
```diff
@@ -62,10 +62,16 @@ func (o *ocsProviderServer) ensureDeleted(r *StorageClusterReconciler, instance
 		return reconcile.Result{}, err
 	}
 
+	serviceType := corev1.ServiceTypeClusterIP
+	if instance.Spec.ProviderAPIServerServiceType != nil {
+		serviceType = *instance.Spec.ProviderAPIServerServiceType
+	} else if instance.Spec.HostNetwork {
+		serviceType = corev1.ServiceTypeNodePort
+	}
```
nb-ohad, line 70: Why do we need this logic inside ensure deleted?

controllers/storagecluster/provider_server.go
```diff
@@ -62,10 +62,16 @@ func (o *ocsProviderServer) ensureDeleted(r *StorageClusterReconciler, instance
...
 	var finalErr error
 
 	for _, resource := range []client.Object{
-		GetProviderAPIServerService(instance),
+		GetProviderAPIServerService(instance.Namespace, serviceType),
```
nb-ohad, line 74: Why do we need to get a full service spec just to delete something? can't we just use the name and namespace?

controllers/storagecluster/provider_server.go
```diff
@@ -127,20 +133,23 @@ func (o *ocsProviderServer) createDeployment(r *StorageClusterReconciler, instan
...
+	case corev1.ServiceTypeClusterIP, corev1.ServiceTypeLoadBalancer, corev1.ServiceTypeNodePort:
+	default:
+		err := fmt.Errorf("providerAPIServer only supports service of type %s, %s and %s",
+			corev1.ServiceTypeNodePort, corev1.ServiceTypeLoadBalancer, corev1.ServiceTypeClusterIP)
+		r.Log.Error(err, "Failed to create/update service, Requested ServiceType is", "ServiceType", instance.Spec.ProviderAPIServerServiceType)
+		return reconcile.Result{}, err
 	}
 
-	desiredService := GetProviderAPIServerService(instance)
+	desiredService := GetProviderAPIServerService(instance.Namespace, serviceType)
```
nb-ohad, line 152: The logic to decide the spec of the service should reside inside the `GetProviderAPIServerService` func. There is no reason to calc outside then send it in

nb-ohad (changes requested):

controllers/storagecluster/provider_server.go
leelavg: we are reusing the calculated serviceType after this func as well.

api/v1/storagecluster_types.go
leelavg: changed to empty string, used a pointer as it was mentioned in one of the meetings.

controllers/storagecluster/provider_server.go
leelavg: removed.

controllers/storagecluster/provider_server.go
leelavg: changed.

controllers/storagecluster/provider_server.go
nb-ohad: You can read it from the desired service

controllers/storagecluster/provider_server.go
```diff
@@ -127,20 +130,23 @@ func (o *ocsProviderServer) createDeployment(r *StorageClusterReconciler, instan
...
+	}
 
+	switch serviceType {
+	case corev1.ServiceTypeClusterIP, corev1.ServiceTypeLoadBalancer, corev1.ServiceTypeNodePort:
+	default:
+		err := fmt.Errorf("providerAPIServer only supports service of type %s, %s and %s",
+			corev1.ServiceTypeNodePort, corev1.ServiceTypeLoadBalancer, corev1.ServiceTypeClusterIP)
+		r.Log.Error(err, "Failed to create/update service, Requested ServiceType is", "ServiceType", instance.Spec.ProviderAPIServerServiceType)
+		return reconcile.Result{}, err
 	}
```
nb-ohad, line 147: Don't do error handling with a switch case. A jump table for empty jump does not make sense

controllers/storagecluster/provider_server.go
```diff
@@ -127,20 +130,23 @@ func (o *ocsProviderServer) createDeployment(r *StorageClusterReconciler, instan
...
-		default:
-			err := fmt.Errorf("providerAPIServer only supports service of type %s, %s and %s",
-				corev1.ServiceTypeNodePort, corev1.ServiceTypeLoadBalancer, corev1.ServiceTypeClusterIP)
-			r.Log.Error(err, "Failed to create/update service, Requested ServiceType is", "ServiceType", instance.Spec.ProviderAPIServerServiceType)
-			return reconcile.Result{}, err
-		}
+		serviceType = instance.Spec.ProviderAPIServerServiceType
+	} else if instance.Spec.HostNetwork {
+		serviceType = corev1.ServiceTypeNodePort
+	}
```
nb-ohad, line 138: This code need to move inside `GetProviderAPIServerService`

nb-ohad (changes requested):

controllers/storagecluster/provider_server.go
nb-ohad: Closing this as I have also mentioned on a newer review

leelavg:
after the approval, unit test was failing, removed the failing test as it was expecting a failure if service is set as `External` and now it isn't possible at API layer itself due to allowing enums.

leelavg:
rebased.

PR #3101, opened 2025-03-25
autoscale: avoid querying prometheus if autosclaer cr is not present

parth-gr:
StorageautoScaler CR should be present to query prometheus, othrwise we can avoid calling to prometheus
It reduces the api calls if the cluster doesnt have auto scaling enabled


parth-gr:
cc @iamniting @travisn

travisn (commented):
/lgtm

PR #3105, opened 2025-03-26
autoscaler: update storageautoscaler design doc

parth-gr:
Currrently the controller was dependednt on reading cr status, update the design so controller to not depend on cr status

### DIFF
--- a/docs/design/storage-auto-scale.md
+++ b/docs/design/storage-auto-scale.md
@@ -29,7 +29,7 @@ Implement auto storage scale CR per device class of a specific Storagecluster.
         maxOsdSize: 8Ti
         timeoutSeconds: 1800   
     status:
-        phase: (NotStarted | InProgress | Succeeded | Failed)
+        phase: ("" | NotStarted | InProgress | Succeeded | Failed)
         error:
             message: ""
             timestamp: <time>
@@ -59,7 +59,7 @@ Go-routine Scraper:
     Query: (ceph_osd_metadata * on (ceph_daemon, namespace, managedBy) group_right(device_class,hostname) (ceph_osd_stat_bytes_used / ceph_osd_stat_bytes))
 
 2) With every run calculate different device class highest osd filled percentage.
-   
+
         OsdPercentage:
         
         ssd               69%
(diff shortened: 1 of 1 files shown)


docs/design/storage-auto-scale.md
```diff
@@ -58,23 +58,41 @@ Go-routine Scraper:
...
-2) If the expansion is not in progress set the status `phase` to `NotStarted` 
+2) Set the status `phase` to `NotStarted`, if no expansion has triggered(status.phase==""). 
+   
+3) Check if the expansion is in progress:
+   
+   1) Load the actualOsdCount and actualOsdSize from the syncMap.
+   
+   2) Load the desiredOsdCount and desiredOsdSize from the Storagecluster.
+   
+   3) If the (actualOsdSize!=desiredOsdSize && actualOsdCount!=desiredOsdCount), expansion is in progress.
```
travisn, line 91:
```suggestion
   3) If the (actualOsdSize<=desiredOsdSize || actualOsdCount<=desiredOsdCount), expansion is in progress. 
```

docs/design/storage-auto-scale.md
```diff
@@ -58,23 +58,41 @@ Go-routine Scraper:
...
 1) Create a named controller that watches for [channel generic event](https://book-v1.book.kubebuilder.io/beyond_basics/controller_watches) per device class.
 
-2) If the expansion is not in progress set the status `phase` to `NotStarted` 
+2) Set the status `phase` to `NotStarted`, if no expansion has triggered(status.phase==""). 
+   
+3) Check if the expansion is in progress:
+   
+   1) Load the actualOsdCount and actualOsdSize from the syncMap.
+   
+   2) Load the desiredOsdCount and desiredOsdSize from the Storagecluster.
```
travisn, line 89: The desired size is simple to load directly, but the desired osd count needs to be calculated for each storageDeviceSet settings.
```suggestion
   2) Calculate the desiredOsdCount and desiredOsdSize from the Storagecluster.
```

docs/design/storage-auto-scale.md
```diff
@@ -102,6 +120,10 @@ Controller:
    
     1) Verify the Storagecluster whether the new osds are added or scaled in size, for all the device sets.
     
+       1) For vertical scaling, Query osd size from Prometheus and match it with storagecluster.spec..size.
```
travisn, line 123: Will we use the same prometheus query from the goroutine to get the osd size and count? Hopefully we can at least share some code there to make sure we are getting consistent values.

travisn: And if the values change for osd size/count, the syncmap needs to be updated here.

travisn (changes requested):

docs/design/storage-auto-scale.md
parth-gr: Yes are already querying the prometheus query to get osd size and count, we can use the same code with go-routine.

> And if the values change for osd size/count, the syncmap needs to be updated here.

Should we just give the scrapper the responsibility to update the sync map (every 10 mins),
 As if we do it here, all individual crs will update the syncmap in parallel.

BlaineEXE: I would assume we read from the cache for this?

docs/design/storage-auto-scale.md
```diff
@@ -58,23 +58,41 @@ Go-routine Scraper:
...
 Controller:
 
 1) Create a named controller that watches for [channel generic event](https://book-v1.book.kubebuilder.io/beyond_basics/controller_watches) per device class.
 
-2) If the expansion is not in progress set the status `phase` to `NotStarted` 
+2) Set the status `phase` to `NotStarted`, if no expansion has triggered(status.phase==""). 
+   
+3) Check if the expansion is in progress:
+   
+   1) Load the actualOsdCount and actualOsdSize from the syncMap.
```
parth-gr, line 87: We can even calculate the `actualOsdCount and actualOsdSize` in the controller reconcile by querying prometehus, 
And then we don't want to update actual values in the syncMap.  

cc @travisn @nb-ohad @leelavg for more thoughts.

travisn: Agreed, seems like the scraper has no reason to query the actual osd count and size. All the scraper is concerned with, is whether it appears the cluster is getting full. 

So it seems to me the controller should just directly query prometheus for the osd count and size.

docs/design/storage-auto-scale.md
```diff
@@ -72,36 +72,54 @@ Controller:
...
    
-10) Scale by patching the Storagecluster, with all the device sets update needed at the same time.
+12) Scale by patching the Storagecluster, with all the device sets update needed at the same time.
 
-11) Verify and Alert:
+13) Verify and Alert:
    
     1) Verify the Storagecluster whether the new osds are added or scaled in size, for all the device sets.
     
+       1) For vertical scaling, Query osd size from Prometheus and match it with storagecluster.spec..size.
```
parth-gr, line 119: @travisn should we requery prometheues for osd size and count during verify?
Or we should query at the start only

travisn: If we just triggered the autoscale, I expect it will take at least a few minutes for the OSDs to scale up/out. So I would just expect we can requeue immediately here instead of checking for completion.

docs/design/storage-auto-scale.md
```diff
@@ -72,35 +72,53 @@ Controller:
...
+   3) If the (actualOsdSize < desiredOsdSize || actualOsdCount < desiredOsdCount), expansion is in progress. 
    
-3) If an expansion is in progress(expectedOsdSize!=startOsdSize || expectedOsdCount!=startOsdCount), check the progress and then requeue each 1 minute until the expansion is completed successfully(jump to step 11).
+4) If the expansion is in progress, check the progress and then requeue each 1 minute until the expansion is completed successfully(jump to step 13).
    
-4) If the LSO storageclass is detected in the storageClassDeviceSet, raise a warning and do not recocnile further.
+5) If no-expansion is in progress,
    
-5) If the highest osd percentage reported in the sync map is more than osdScalingThresholdPercent(70%) means reaching osd nearfull, scaling is needed.
+   1) Check the (status.phase == "InProgress"), if yes move the status phase to succeeded.
```
travisn, line 89: If the phase is still in progress, shouldn't we just to step 13? It seems odd that it's still in progress, but we will evaluate whether to again expand it in the same reconcile.

parth-gr: I think we don't even need step 13, as we are doing same at step 3)

What if we remove step 13) and keep the step 5) as it is

docs/design/storage-auto-scale.md
```diff
@@ -72,60 +72,80 @@ Controller:
...
+   
+   3) If the (actualOsdSize < desiredOsdSize || actualOsdCount < desiredOsdCount), expansion is in progress. 
+   
+4) If the expansion is in progress, check the progress and then requeue each 1 minute until the expansion is completed successfully, change the phase to `failed` if scaling not `Succeeded` with in timeoutSeconds(default:1800) interval.
    
-3) If an expansion is in progress(expectedOsdSize!=startOsdSize || expectedOsdCount!=startOsdCount), check the progress and then requeue each 1 minute until the expansion is completed successfully(jump to step 11).
+5) If no-expansion is in progress,
    
-4) If the LSO storageclass is detected in the storageClassDeviceSet, raise a warning and do not recocnile further.
+   1) Check the (status.phase == "InProgress"), if yes scaling is successful, update the status of the `StorageAutoScaling` CR with `lastExpansionCompletionTime` and `phase` and also osd count and size.
```
travisn, line 89: It's confusing that it would be InProgress if it's detected that it's not in progress. But if this code path is just part of the status check, then I'm sure it's fine.

travisn (approved):
/approve

docs/design/storage-auto-scale.md
parth-gr: It's just for updating the status to success from in progress, when we have already detected it was succeeded

PR #3107, opened 2025-03-26
Add autoscaler unit test

parth-gr:
test: add unit test for verify functions

### DIFF
--- a/controllers/storageautoscaler/verify_test.go
+++ b/controllers/storageautoscaler/verify_test.go
@@ -1,10 +1,15 @@
 package storageautoscaler
 
 import (
+	"context"
 	"testing"
 	"time"
 
+	"github.com/go-logr/logr"
+	"github.com/prometheus/common/model"
+	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
 	"github.com/stretchr/testify/assert"
+	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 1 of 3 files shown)


controllers/storageautoscaler/verify_test.go
```diff
@@ -25,3 +30,30 @@ func TestTimeoutHasElapsed(t *testing.T) {
...
+	// mock getOsdSize() function for test
+	mockGetOsdSize = func(ctx context.Context, namespace string, log logr.Logger) (model.Vector, error) {
+		return model.Vector{
+			&model.Sample{
+				Metric: model.Metric{
+					"device_class": "hdd",
+				},
+				Value: model.SampleValue(10),
+			},
+		}, nil
```
BlaineEXE, line 44: Verifying that scaling has/hasn't worked is one of the core pieces of logic that this feature depends on. Do you think that a single unit test that covers only the basic happy path is sufficient to prevent regressions?

parth-gr: Still need to add more test cases to it, for sure this is the crucial part of the logic, waiting on the new design to finalsied so we can act on this accordingly.
https://github.com/red-hat-storage/ocs-operator/pull/3105

parth-gr:
closing as unit test will be added in https://github.com/red-hat-storage/ocs-operator/pull/3179

PR #3113, opened 2025-03-31
set noobaa account availability explicitly for reconciling the resource

leelavg:
after consumer is enabled, client doesn't report status instantaneously
and if it's empty we reconcileNoobaaAccount once, which'll get into a
deadlock as noobaa will not be ready until we send storageconfig for
storageclass creation which will check for consumer readiness and it'll
never be ready.

to fix above and introduce configurability we introduce a new option
for explicity mentioning the need of noobaa account while creating
storageconsumer.

### DIFF
--- a/controllers/storagecluster/storageconsumer.go
+++ b/controllers/storagecluster/storageconsumer.go
@@ -2,6 +2,8 @@ package storagecluster
 
 import (
 	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
 	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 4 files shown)


controllers/util/storageconsumer.go
```diff
@@ -107,6 +112,10 @@ func (wrapper storageConsumerResourceMapWrapper) GetNfsClientProfileName() strin
 	return wrapper.data[nfsClientProfileKey]
 }
 
+func (wrapper storageConsumerResourceMapWrapper) GetNoobaaAccountApplicability() string {
+	return wrapper.data[noobaaAccountKey]
+}
+
```
nb-ohad, line 118: I am not sure what do we need this for? And I am also failing to find where `GetNoobaaAccountName`/`SetNoobaaAccountName` are located?

controllers/util/storageconsumer.go
```diff
@@ -48,6 +52,7 @@ type StorageConsumerResources interface {
 	SetRbdClientProfileName(string)
 	SetCephFsClientProfileName(string)
 	SetNfsClientProfileName(string)
+	MarkNoobaaAccountNotApplicable()
```
nb-ohad, line 55:
```suggestion

        // Mark as not applicable
	MarkNoobaaAccountNotApplicable()
```

controllers/storagecluster/storageconsumer.go
```diff
@@ -24,6 +26,20 @@ type storageConsumer struct{}
...
+	if _, err := controllerutil.CreateOrUpdate(r.ctx, r.Client, consumerConfigMap, func() error {
+		if consumerConfigMap.Data == nil {
+			consumerConfigMap.Data = map[string]string{}
+		}
+		resourceMap := util.WrapStorageConsumerResourceMap(consumerConfigMap.Data)
+		resourceMap.MarkNoobaaAccountNotApplicable()
+		return nil
+	}); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to create/update storageconsumer configmap %s: %v", localStorageConsumerConfigMapName, err)
+	}
```
nb-ohad, line 41: This code need to move to the same place where we are already creating the CM for the internal client

nb-ohad (changes requested):

PR #3114, opened 2025-03-31
consumer: reconcile rbd/cephfs/nfs if service required

rewantsoni:
We should reconcile the rbd/cephfs/nfs related resources when available. This PR addresses this concern.

The caps for nfs are similar to that we have for cephFs


controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -296,6 +296,23 @@ func (r *StorageConsumerReconciler) reconcileEnabledPhases() (reconcile.Result,
 				return reconcile.Result{}, err
 			}
 
+			//TODO: Verify the caps
```
nb-ohad, line 299: Why do we need this comment?

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -296,6 +296,23 @@ func (r *StorageConsumerReconciler) reconcileEnabledPhases() (reconcile.Result,
...
+				return reconcile.Result{}, err
+			}
+
+			if err := r.reconcileCephClientNfsNode(
+				consumerResources.GetCsiNfsNodeSecretName(),
+				consumerResources.GetSubVolumeGroupName(),
+				consumerConfigMap,
+			); err != nil {
+				return reconcile.Result{}, err
+			}
```
nb-ohad, line 314: Shouldn't this be behind a condition?

nb-ohad (changes requested):

controllers/storageconsumer/storageconsumer_controller.go
rewantsoni: We need to verify the caps, as these caps are similar to cephFs and might provide access to multiple radosnamesapces

controllers/storageconsumer/storageconsumer_controller.go
rewantsoni: Right, we should put it behind if we have NFS enabled on storageCluster spec

controllers/storageconsumer/storageconsumer_controller.go
nb-ohad: But why a comment in the code? if we don't need to change then we need to create another PR just to remove the comment. Please capture this somewhere else? (an issue, an internal todo list, etc.)

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -240,13 +241,21 @@ func (r *StorageConsumerReconciler) reconcileEnabledPhases() (reconcile.Result,
...
 				return reconcile.Result{}, fmt.Errorf("failed to find StorgeCluster owning the StorageConsumer")
 			}
+			storageCluster := &ocsv1.StorageCluster{}
+			storageCluster.Name = ownerRef.Name
+			storageCluster.Namespace = r.namespace
+
+			err := r.Client.Get(r.ctx, client.ObjectKeyFromObject(storageCluster), storageCluster)
+			if err != nil {
+				return reconcile.Result{}, err
+			}
```
nb-ohad, line 258:
```suggestion
			if err := r.Client.Get(r.ctx, client.ObjectKeyFromObject(storageCluster), storageCluster); err != nil {
				return reconcile.Result{}, err
			}
```

nb-ohad (changes requested):

controllers/storageconsumer/storageconsumer_controller.go
rewantsoni: done

controllers/storageconsumer/storageconsumer_controller.go
rewantsoni: ack, updating it

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -155,6 +155,35 @@ func (r *StorageConsumerReconciler) reconcileEnabledPhases() (reconcile.Result,
...
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+
+		// isCephFsAvailable - get the default filesystem, isRbdAvailable - list blockpools CR with count 1, isNfsAvailable -
+		blockPoolList := &rookCephv1.CephBlockPoolList{}
+		if err := r.Client.List(r.ctx, blockPoolList, client.InNamespace(r.namespace), client.Limit(1)); err != nil {
+			return reconcile.Result{}, err
+		}
+		isRbdAvailable := len(blockPoolList.Items) > 0
```
rewantsoni, line 169: Checking based on blockpool list may not be a very good idea as the we have a blockpool created for NFS, one builtin for mgr

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -155,6 +155,35 @@ func (r *StorageConsumerReconciler) reconcileEnabledPhases() (reconcile.Result,
...
+		isCephFsAvailable := cephFs.UID != ""
+
+		cephNFS := &rookCephv1.CephNFS{}
+		cephNFS.Name = util.GenerateNameForCephNFS(storageCluster)
+		cephNFS.Namespace = r.namespace
+		if err := r.get(cephNFS); client.IgnoreNotFound(err) != nil {
+			return reconcile.Result{}, fmt.Errorf("failed to get CephNFS: %v", err)
+		}
+		isNfsAvailable := cephNFS.UID != ""
+
```
nb-ohad, line 186: Can we move this code into a function? one that returns available services?
RBD, CephFS, NFS, MCG

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -155,6 +155,35 @@ func (r *StorageConsumerReconciler) reconcileEnabledPhases() (reconcile.Result,
 		if err := r.deleteOnboardingSecret(); err != nil {
 			return reconcile.Result{}, nil
 		}
+
+		storageCluster, err := util.GetStorageClusterInNamespace(r.ctx, r.Client, r.namespace)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+
+		// isCephFsAvailable - get the default filesystem, isRbdAvailable - list blockpools CR with count 1, isNfsAvailable -
```
nb-ohad, line 164: If you want to mentioned the criteria in a comment, please do it per service above the section of code that deal with that service

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -165,57 +194,10 @@ func (r *StorageConsumerReconciler) reconcileEnabledPhases() (reconcile.Result,
...
-					r.storageConsumer,
-					consumerConfigMap,
-					r.Scheme,
-				); err != nil {
-					return err
-				}
-			}
-			return nil
-		}); err != nil {
+		if err := r.reconcileConsumerConfigMap(consumerConfigMap, isRbdAvailable, isCephFsAvailable, isNfsAvailable); err != nil {
```
nb-ohad, line 197: You already have 3 variables for  3 services but we also need to take into account noobaa, and maybe other services in the future. To make sure the code is not cluttered i would suggest using a new struct `AvailableServices`

controllers/storageconsumer/storageconsumer_controller.go
nb-ohad, line 1: - missing code the for conditional client profiles base on service availablity
- Bug where you decide if you are primary based on an in memory change

nb-ohad (changes requested):

controllers/storageconsumer/storageconsumer_controller.go
rewantsoni: creating a funtion and moving it to the utils

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -378,9 +352,78 @@ func (r *StorageConsumerReconciler) deleteOnboardingSecret() error {
 	return nil
 }
 
+func (r *StorageConsumerReconciler) reconcileConsumerConfigMap(
+	availableServices *util.AvailableServices,
```
nb-ohad, line 356: Do we really need a new line for a func with a single param?

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -391,7 +434,7 @@ func (r *StorageConsumerReconciler) reconcileCephRadosNamespace(
 	var combinedErr error
 	for idx := range blockPools.Items {
 		bp := &blockPools.Items[idx]
-		if bp.Name == "builtin-mgr" {
+		if bp.Name == "builtin-mgr" || bp.Name != util.GenerateNameForCephNFSBlockPool(storageCluster) {
```
nb-ohad, line 437: What is this change about?

controllers/util/storagecluster.go
```diff
@@ -0,0 +1,68 @@
...
+	CephFs bool
+	Nfs    bool
+	Mcg    bool
+}
+
+func GetAvailableServices(
+	ctx context.Context,
+	cl client.Client,
+	storageCluster *ocsv1.StorageCluster,
+) (*AvailableServices, error) {
```
nb-ohad, line 25: These are only relevant for the storageconsumer controller? why are we pushing them to the util package?

controllers/util/storagecluster.go
```diff
@@ -0,0 +1,68 @@
...
+	availableServices := &AvailableServices{}
+
+	// rbd - list blockpools and filter out the builtin ones, isNfsAvailable -
+	blockPoolList := &rookCephv1.CephBlockPoolList{}
+	if err := cl.List(ctx, blockPoolList, client.InNamespace(storageCluster.Namespace)); err != nil {
+		return nil, err
+	}
+	for i := range blockPoolList.Items {
+		poolName := blockPoolList.Items[i].Name
+		if poolName != "builtin-mgr" && poolName != GenerateNameForCephNFSBlockPool(storageCluster) {
```
nb-ohad, line 35: Doesn't it make more sense to just check if the RBD default block pool exists?

controllers/util/storagecluster.go
```diff
@@ -0,0 +1,68 @@
...
+			availableServices.Rbd = true
+		}
```
nb-ohad, line 37:
```suggestion
		availableServices.Rbd = poolName != "builtin-mgr" && poolName != GenerateNameForCephNFSBlockPool(storageCluster) {
```

services/provider/server/server.go
```diff
@@ -500,42 +516,48 @@ func (s *OCSProviderServer) appendClientProfileExternalResources(
...
 	profileMap := make(map[string]*csiopv1a1.ClientProfileSpec)
 
-	rbdClientProfileName := consumerConfig.GetRbdClientProfileName()
-	rbdClientProfile := profileMap[rbdClientProfileName]
-	if rbdClientProfile == nil {
-		rbdClientProfile = &csiopv1a1.ClientProfileSpec{
-			CephConnectionRef: corev1.LocalObjectReference{Name: consumer.Status.Client.Name},
-		}
-		profileMap[rbdClientProfileName] = rbdClientProfile
+	if availableServices.Rbd {
```
nb-ohad, line 519: In this context you don't need the available services map, you already decided what is available based on the keys in the configmap

services/provider/server/server.go
```diff
@@ -500,42 +516,48 @@ func (s *OCSProviderServer) appendClientProfileExternalResources(
...
+			RadosNamespace: consumerConfig.GetRbdRadosNamespaceName(),
+		}
 	}
 
-	cephFsClientProfileName := consumerConfig.GetCephFsClientProfileName()
-	cephFsClientProfile := profileMap[cephFsClientProfileName]
-	if cephFsClientProfile == nil {
-		cephFsClientProfile = &csiopv1a1.ClientProfileSpec{
-			CephConnectionRef: corev1.LocalObjectReference{Name: consumer.Status.Client.Name},
+	if availableServices.CephFs {
```
nb-ohad, line 534: Same as last comment

services/provider/server/server.go
```diff
@@ -500,42 +516,48 @@ func (s *OCSProviderServer) appendClientProfileExternalResources(
...
-		KernelMountOptions: kernelMountOptions,
-		RadosNamespace:     ptr.To(consumerConfig.GetSubVolumeGroupRadosNamespaceName()),
 	}
 
-	nfsClientProfileName := consumerConfig.GetCephFsClientProfileName()
-	nfsClientProfile := profileMap[nfsClientProfileName]
-	if nfsClientProfile == nil {
-		nfsClientProfile = &csiopv1a1.ClientProfileSpec{
-			CephConnectionRef: corev1.LocalObjectReference{Name: consumer.Status.Client.Name},
+	if availableServices.Nfs {
```
nb-ohad, line 550: Same as last comment

services/provider/server/server.go
```diff
@@ -552,13 +574,32 @@ func (s *OCSProviderServer) appendCephClientSecretExternalResources(
 	extR []*pb.ExternalResource,
 	consumer *ocsv1alpha1.StorageConsumer,
 	consumerConfig util.StorageConsumerResources,
+	storageCluster *ocsv1.StorageCluster,
+	availableServices *util.AvailableServices,
```
nb-ohad, line 578: Same as last comment

services/provider/server/server.go
```diff
@@ -552,13 +574,32 @@ func (s *OCSProviderServer) appendCephClientSecretExternalResources(
...
 ) ([]*pb.ExternalResource, error) {
 
-	cephClients := []string{
-		consumerConfig.GetCsiRbdProvisionerSecretName(),
-		consumerConfig.GetCsiRbdNodeSecretName(),
-		consumerConfig.GetCsiCephFsProvisionerSecretName(),
-		consumerConfig.GetCsiCephFsNodeSecretName(),
+	cephClients := []string{}
+
+	if availableServices.Rbd {
```
nb-ohad, line 583: Same as last comment

nb-ohad (changes requested):

controllers/storageconsumer/storageconsumer_controller.go
rewantsoni: The radosnamespace that we create should not be created for `builtin-mgr`, and the blockpool that we have for nfs.

controllers/util/storagecluster.go
rewantsoni: No, as the default blockpool might be disabled and a custom one might be created

controllers/util/storageconsumer.go
```diff
@@ -20,6 +20,13 @@ const (
 	nfsClientProfileKey             = "csiop-nfs-client-profile"
 )
 
+type AvailableServices struct {
+	Rbd    bool
+	CephFs bool
+	Nfs    bool
+	Mcg    bool
+}
+
```
rewantsoni, line 29: The reason to have this here instead of in storageconsumer controller is because the GetStorageConsumerDefaultResourceNames uses this struct other option would be to pass three bool for each service type

controllers/util/storagecluster.go
nb-ohad: This is an edge case. But regardless I don't think there is a way cheph without RBD as it is needed for other services to function. in that case let's remove the conditions for rbd here

controllers/util/storageconsumer.go
nb-ohad: @rewantsoni I am not sure I understand this comment. can you please provide more context?

PR #3115, opened 2025-03-31
consumer: update calculation of the default clientprofile name

rewantsoni:
calculate the default client profile name based on storageconsumer name and uid

### DIFF
--- a/controllers/util/storageconsumer.go
+++ b/controllers/util/storageconsumer.go
@@ -156,28 +156,20 @@ func (wrapper storageConsumerResourceMapWrapper) SetNfsClientProfileName(name st
 	wrapper.data[nfsClientProfileKey] = name
 }
 
-func GetStorageConsumerDefaultResourceNames(storageConsumerName, fsid string) map[string]string {
-
-	//TODO: Revisit this to use a different method to calculate the name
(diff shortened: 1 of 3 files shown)


controllers/util/storageconsumer.go
```diff
@@ -156,10 +156,10 @@ func (wrapper storageConsumerResourceMapWrapper) SetNfsClientProfileName(name st
 	wrapper.data[nfsClientProfileKey] = name
 }
 
-func GetStorageConsumerDefaultResourceNames(storageConsumerName, fsid string) map[string]string {
+func GetStorageConsumerDefaultResourceNames(storageConsumerName, storageConsumerUid string) map[string]string {
 
-	//TODO: Revisit this to use a different method to calculate the name
-	clientProfileName := CalculateMD5Hash([2]string{fsid, storageConsumerName})
+	//TODO: Revisit this to use a different method to calculate the hash
+	clientProfileName := CalculateMD5Hash([2]string{storageConsumerUid, storageConsumerName})
```
nb-ohad, line 162: Why do we use bot name and uid? you don't even need the variable, just use the uid in place

nb-ohad (changes requested):

PR #3116, opened 2025-03-31
rdr: update the controller/server for rdr after convergence

rewantsoni:
This PR does the following:
- update the blockpool mirroring mode to init-only - ceph requires the blockpools to be enabled for mirroring before we enable mirroring for the rns. The init-only mode specifies that a particular blockpool's rns are ready to be mirrored 
- send consumers/client info in odf-info if they are enabled - If a consumer is not enabled, we shouldn't let ACM read the information about the consumer
- use StorageConsumerResources in GetStorageClientsInfo to fetch the right rns - Till 4.18 we had one rns per consumer hence the assumption that we had to send the rns held. But with 4.19 we are moving to multiple rns with the same rns name and which is stored in a configmap, hence moving to fetch the configmap and using the rns name in the configmap
- fetch the rns using ownerRef instead of labels - till 4.18, we didn't have an ownerRef of consumer on the rns as it was created by StorageRequest controller. So we relied on labels. From 4.19 we are directly creating the rns CR and hence we can add the ownerRef

### DIFF
--- a/controllers/mirroring/mirroring_controller.go
+++ b/controllers/mirroring/mirroring_controller.go
@@ -23,7 +23,6 @@ import (
 	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
 	ocsv1alpha1 "github.com/red-hat-storage/ocs-operator/api/v4/v1alpha1"
 	providerClient "github.com/red-hat-storage/ocs-operator/services/provider/api/v4/client"
-	controllers "github.com/red-hat-storage/ocs-operator/v4/controllers/storageconsumer"
 	"github.com/red-hat-storage/ocs-operator/v4/controllers/util"
 
 	"github.com/go-logr/logr"
@@ -31,6 +30,7 @@ import (
 	"golang.org/x/exp/maps"
 	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/ptr"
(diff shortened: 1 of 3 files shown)


umangachapagain (approved):

rewantsoni:
Discussed it with Ohad offline, PR will be updated as per the fix.

services/provider/server/server.go
```diff
@@ -1662,10 +1664,11 @@ func (s *OCSProviderServer) GetStorageClientsInfo(ctx context.Context, req *pb.S
 			klog.Error(err)
 			continue
 		}
-		clientInfo := &pb.ClientInfo{ClientID: req.ClientIDs[i]}
+		consumerConfig := util.WrapStorageConsumerResourceMap(consumerConfigMap.Data)
```
nb-ohad, line 1671: Please move this just before line 1670 where it is first used

controllers/mirroring/mirroring_controller.go
```diff
@@ -394,7 +392,7 @@ func (r *MirroringReconciler) reconcileBlockPoolMirroring(
 					)
 
 					cephBlockPool.Spec.Mirroring.Enabled = true
-					cephBlockPool.Spec.Mirroring.Mode = "image"
+					cephBlockPool.Spec.Mirroring.Mode = "init-only"
```
nb-ohad, line 395: Can we do this change as a post-upgrade operation? Do we know how would it effect existing RDR deployments?

nb-ohad (changes requested):

controllers/mirroring/mirroring_controller.go
rewantsoni: I am not sure how this change is going to affect upgrades. We would need to check that. 
@subhamkrai, do you know what will happen if we have blockpool mirroring enabled with mode as `image` and  then we change the mode to `init-only` and create the rns cr for implicit namespace to enable mirroring?

subhamkrai: I think changing mirroring mode is supported so we'll good

PR #3117, opened 2025-03-31
check consumer status for clusterid before reconciling noobaaaccount

leelavg:
after consumer is enabled, client doesn't report status instantaneously and if it's empty we reconcileNoobaaAccount once, which'll get into a deadlock as noobaa will not be ready until we send storageconfig for storageclass creation which will check for consumer readiness and it'll never be ready.


leelavg:
Tested, w/ fix the state is moved to Ready after ceph resources are created rather than waiting for first heartbeat removing the noobaaacount deadlock scenario.
```
  status:
    client: {}
    onboardingTicketSecret:
      name: ""
    resourceNameMappingConfigMap:
      name: storageconsumer-internal
    state: Ready
```

PR #3121, opened 2025-04-01
remove namespaced owner reference from cluster-scoped StorageClass

umangachapagain:


leelavg:
@rewantsoni possible to do a quick check if there are more of these when the code is refactored?

rewantsoni:
@leelavg Only obcstorageClass was moved to a new file, other storageclasses for internal mode are created via client op so we should be good

umangachapagain:
/hold

I see more issues during testing

PR #3124, opened 2025-04-02
prefix svg name with cephfs name and remove old owner

leelavg:
we should prefix svg with cephfs name and StorageCluster was the owner of svg pre4.19 which this PR removes & transfers the ownership to configmap & StorageConsumer.

### DIFF
--- a/controllers/storageconsumer/storageconsumer_controller.go
+++ b/controllers/storageconsumer/storageconsumer_controller.go
@@ -482,10 +482,18 @@ func (r *StorageConsumerReconciler) reconcileCephFilesystemSubVolumeGroup(
 	}
 
 	svg := &rookCephv1.CephFilesystemSubVolumeGroup{}
-	svg.Name = subVolumeGroupName
+	svg.Name = fmt.Sprintf("%s-%s", cephFs.Name, subVolumeGroupName)
 	svg.Namespace = r.namespace
 
 	if _, err := ctrl.CreateOrUpdate(r.ctx, r.Client, svg, func() error {
+		index := slices.IndexFunc(svg.OwnerReferences, func(owner metav1.OwnerReference) bool {
+			return owner.Kind == "StorageCluster"
+		})
+		if index > -1 {
+			length := len(svg.OwnerReferences)
+			svg.OwnerReferences[index] = svg.OwnerReferences[length-1]
+			svg.OwnerReferences = svg.OwnerReferences[:length-1]
+		}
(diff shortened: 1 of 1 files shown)


controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -476,10 +476,11 @@ func (r *StorageConsumerReconciler) reconcileCephFilesystemSubVolumeGroup(
 	}
 
 	svg := &rookCephv1.CephFilesystemSubVolumeGroup{}
-	svg.Name = subVolumeGroupName
+	svg.Name = fmt.Sprintf("%s-%s", cephFs.Name, subVolumeGroupName)
```
nb-ohad, line 479: This is not the agreed design. There is only one subvolumegroup per consumer and its name should be by default the consumer name unless it got override by a name from an pre configured configmap name

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -822,3 +823,12 @@ func (r *StorageConsumerReconciler) getAvailableServices(storageCluster *ocsv1.S
...
 }
+
+func removeStorageClusterAsOwner(obj client.Object) {
+	refs := obj.GetOwnerReferences()
+	if idx := slices.IndexFunc(refs, func(owner metav1.OwnerReference) bool {
+		return owner.Kind == "StorageCluster"
+	}); idx != -1 {
+		obj.SetOwnerReferences(slices.Delete(refs, idx, idx+1))
+	}
+}
```
nb-ohad, line 834: This is only called once, please embed it in the call site

nb-ohad (changes requested):

controllers/storageconsumer/storageconsumer_controller.go
leelavg: _**its name should be by default the consumer name**_ isn't this statement for the Ceph layer? we are also giving similar treatment to rns, the one mentioned in the config is at ceph and we prepend blockpool name at rook.

controllers/storageconsumer/storageconsumer_controller.go
leelavg: done.

leelavg: done.

leelavg:
closing due to:
- two rook CRs can refer to same Ceph layer (rns/svg)
- currently rook doesn't allow removing a duplicate CR created in above fashion
- rook will provide a feature which allows removing such a CR in future
combining w/ #3142 this PR is obsolete.

PR #3125, opened 2025-04-02
controllers: fix obc storageclass creation

rewantsoni:
The storageClass is a cluster scoped resources, it doesn't need a namespace. 
Also, a storageClass should follow createOrReplace logic instead of createOrUpdate as the params are immutable

### DIFF
--- a/controllers/util/k8sutil.go
+++ b/controllers/util/k8sutil.go
@@ -4,12 +4,12 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/red-hat-storage/ocs-operator/v4/controllers/defaults"
 	"os"
 	"strings"
 	"time"
 
 	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
+	"github.com/red-hat-storage/ocs-operator/v4/controllers/defaults"
 
 	"github.com/go-logr/logr"
 	configv1 "github.com/openshift/api/config/v1"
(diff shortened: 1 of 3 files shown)


controllers/util/k8sutil.go
```diff
@@ -289,3 +292,51 @@ func NewK8sClient(scheme *runtime.Scheme) (client.Client, error) {
...
+		}
+		return nil
+	}
+
+	existing := obj.DeepCopyObject()
+	if err := mutate(f, key, obj); err != nil {
+		return err
+	}
+
+	if equality.Semantic.DeepEqual(existing, obj) {
```
iamniting, line 316: Will not this always return false as we are also checking the metadata and objectmeta which will differ?

controllers/util/k8sutil.go
```diff
@@ -289,3 +292,51 @@ func NewK8sClient(scheme *runtime.Scheme) (client.Client, error) {
...
+	}
+	return nil
+}
+
+// mutate wraps a MutateFn and applies validation to its result.
+func mutate(f controllerutil.MutateFn, key client.ObjectKey, obj client.Object) error {
+	if err := f(); err != nil {
+		return err
+	}
+	if newKey := client.ObjectKeyFromObject(obj); key != newKey {
```
iamniting, line 338: Should we not check this before calling a func?

iamniting (changes requested):

controllers/util/k8sutil.go
rewantsoni: We have been using this in the client op from 4.18 and it's working as expected

controllers/util/k8sutil.go
rewantsoni: We are applying the mutate function after getting the current object. This is make sure that the fields we want are preserved and additional labels/annotations added by users are not removed

iamniting (approved):

PR #3126, opened 2025-04-03
storageautoscaler: match the storagecluster uid with storageutoscaler

parth-gr:
there can be a storageautocaler instance left in the system
that was created for older storagecluster,
So match the uid for which storagecluster instance it was
created and then only reconcile.

### DIFF
--- a/api/v1/storageautoscaler_types.go
+++ b/api/v1/storageautoscaler_types.go
@@ -26,7 +26,7 @@ import (
 type StorageAutoScalerSpec struct {
 	// StorageCluster is the name of the storage cluster for which the storage scaling is to be done.
 	// +kubebuilder:validation:Required
-	StorageCluster corev1.LocalObjectReference `json:"storageCluster,omitempty"`
+	StorageCluster corev1.ObjectReference `json:"storageCluster,omitempty"`
 	// DeviceClass is the name of the device class for which the storage scaling is to be done.
 	// +kubebuilder:validation:Optional
 	// +kubebuilder:default="ssd"
--- a/controllers/storageautoscaler/storageautoscaler_controller.go
+++ b/controllers/storageautoscaler/storageautoscaler_controller.go
@@ -94,11 +94,25 @@ func (r *StorageAutoscalerReconciler) Reconcile(ctx context.Context, request rec
 
 	// list the storagecluster
 	storageCluster := &ocsv1.StorageCluster{}
+	if storageAutoScaler.Spec.StorageCluster.Name == "" {
+		err = fmt.Errorf("storage cluster name is empty")
+		r.Log.Error(err, "scaling cannot be performed")
+		return reconcile.Result{}, err
+	}
 	storageClusterName := storageAutoScaler.Spec.StorageCluster.Name
 	err = r.Get(ctx, types.NamespacedName{Name: storageClusterName, Namespace: request.Namespace}, storageCluster)
 	if err != nil {
 		r.Log.Error(err, "failed to get storage cluster")
 		return reconcile.Result{}, err
+	}
+
+	// check if the storage cluster uid matches the storageautoscaler, storagecluster spec uid
+	if storageAutoScaler.Spec.StorageCluster.UID != "" {
+		if storageCluster.ObjectMeta.UID != storageAutoScaler.Spec.StorageCluster.UID {
+			err = fmt.Errorf("storage cluster uid does not match with storage autoscaler, storagecluster spec uid")
+			r.Log.Error(err, "uid did not match", "storage cluster uid", storageCluster.UID, "storage autoscaler storagecluster spec uid", storageAutoScaler.Spec.StorageCluster.UID)
+			return reconcile.Result{}, err
+		}
 	}
 
 	// list the storagecluster deviceset storageclass
--- a/docs/design/storage-auto-scale.md
+++ b/docs/design/storage-auto-scale.md
@@ -23,6 +23,8 @@ Implement auto storage scale CR per device class of a specific Storagecluster.
     spec:
         storageCluster: 
             name: <storageclusterName>
+            namespace: <storageclusterNamespace>
+            uid: <storageclusterUid>
         deviceClass: ssd
         storageCapacityLimit: <capacity>
         storageScalingThresholdPercent: 70  # should be less than osd nearfull percentage 
--- a/config/crd/bases/ocs.openshift.io_storageautoscalers.yaml
+++ b/config/crd/bases/ocs.openshift.io_storageautoscalers.yaml
@@ -87,14 +87,43 @@ spec:
                 description: StorageCluster is the name of the storage cluster for
                   which the storage scaling is to be done.
                 properties:
+                  apiVersion:
+                    description: API version of the referent.
+                    type: string
+                  fieldPath:
+                    description: |-
+                      If referring to a piece of an object instead of an entire object, this string
+                      should contain a valid JSON/Go field access statement, such as desiredState.manifest.containers[2].
+                      For example, if the object reference is to a container within a pod, this would take on a value like:
+                      "spec.containers{name}" (where "name" refers to the name of the container that triggered
+                      the event) or if no container name is specified "spec.containers[2]" (container with
+                      index 2 in this pod). This syntax is chosen only to have some well-defined way of
+                      referencing a part of an object.
+                    type: string
+                  kind:
+                    description: |-
+                      Kind of the referent.
+                      More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds
+                    type: string
                   name:
-                    default: ""
                     description: |-
                       Name of the referent.
-                      This field is effectively required, but due to backwards compatibility is
-                      allowed to be empty. Instances of this type with an empty value here are
-                      almost certainly wrong.
                       More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+                    type: string
+                  namespace:
+                    description: |-
+                      Namespace of the referent.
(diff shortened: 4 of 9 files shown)


alfonsomthd:
Fixes: https://issues.redhat.com/browse/DFBUGS-2047

parth-gr:
> This is not gonna work if the UID is not set by the customer. We should set the UID ourself in the status and read from there if matches reconcile else skip.

If we set the status from the controller, users can not just change the UID if needed, as they can now change in the yaml setting,
@ShyamsundarR do you want to add your thoughts?

iamniting:
> > This is not gonna work if the UID is not set by the customer. We should set the UID ourself in the status and read from there if matches reconcile else skip.
> 
> If we set the status from the controller, users can not just change the UID if needed, as they can now change in the yaml setting, @ShyamsundarR do you want to add your thoughts?

In that case we should make the uid as a required field.

api/v1/storageautoscaler_types.go
```diff
@@ -26,7 +26,7 @@ import (
 type StorageAutoScalerSpec struct {
 	// StorageCluster is the name of the storage cluster for which the storage scaling is to be done.
 	// +kubebuilder:validation:Required
-	StorageCluster corev1.LocalObjectReference `json:"storageCluster,omitempty"`
+	StorageCluster corev1.ObjectReference `json:"storageCluster,omitempty"`
```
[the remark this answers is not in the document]
parth-gr: Using the genric k8s reference helps a lot, `corev1.ObjectReference`

In future, we would be using finalizers for protecting deletion when the phase is `in-progress`, and we can also use generation field, etc with future usecases

controllers/storageautoscaler/storageautoscaler_controller.go
```diff
@@ -94,13 +94,27 @@ func (r *StorageAutoscalerReconciler) Reconcile(ctx context.Context, request rec
...
+	}
 	storageClusterName := storageAutoScaler.Spec.StorageCluster.Name
 	err = r.Get(ctx, types.NamespacedName{Name: storageClusterName, Namespace: request.Namespace}, storageCluster)
 	if err != nil {
 		r.Log.Error(err, "failed to get storage cluster")
 		return reconcile.Result{}, err
 	}
 
+	// check if the storage cluster uid matches the storageautoscaler, storagecluster spec uid
+	if storageAutoScaler.Spec.StorageCluster.UID != "" {
```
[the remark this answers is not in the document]
parth-gr: According to me, we can still progress if it's not set,
 
If we really want it I can add a check to stop further reconciliation if its not set. cc @ShyamsundarR

controllers/storageautoscaler/storageautoscaler_controller.go
```diff
@@ -94,13 +94,27 @@ func (r *StorageAutoscalerReconciler) Reconcile(ctx context.Context, request rec
...
+		if storageCluster.ObjectMeta.UID != storageAutoScaler.Spec.StorageCluster.UID {
+			err = fmt.Errorf("storage cluster uid does not match with storage autoscaler, storagecluster spec uid")
+			r.Log.Error(err, "uid did not match", "storage cluster uid", storageCluster.UID, "storage autoscaler storagecluster spec uid", storageAutoScaler.Spec.StorageCluster.UID)
+			return reconcile.Result{}, err
+		}
```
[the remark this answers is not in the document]
parth-gr: @alfonsomthd if we add a new state can UI accommodate that? For having a new state in the CR, we need changes int he UI?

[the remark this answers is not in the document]
alfonsomthd: > @alfonsomthd if we add a new state can UI accommodate that? For having a new state in the CR, we need changes in the UI?

We can map it: Right now, for any state not recognized by the UI state mapping, we show `Current status: Unknown`. If we want to properly show a new one (e.g. `Invalid`) then we have to map it.

api/v1/storageautoscaler_types.go
BlaineEXE, line 29: I would propose that we consider having the StorageAutoscaler controller apply the UID to the resource so that users/UI don't have to apply it themselves. This matches the design of Kubernetes resource metadata, where things like creation timestamp and UID are applied automatically at creation time. This would result in a system where the autoscale resource is 'bound' to the StorageCluster automatically, ensuring the UID is always set. When we can rely on the value always being present, we can be more confident that the system will prevent users from accidentally reusing old StorageAutoscaler configs.

One of the challenges with the approach I suggested above is that we should consider not applying UID to the resource `spec` in that case. The next options would be to put it in the status or in the metadata (i.e., as an annotation). IMO, either is fine, but I think putting it in the `status` would be more visible for users to troubleshoot when the controller rejects an old config having a non-matching UID.

My suggested approach doesn't prevent us from adding UID as an input in the `spec` as in the current changeset, but it would be optional. I would assume the spec's UID would be copied to the status/annotation in that case.

@ShyamsundarR @umangachapagain thoughts?

docs/design/storage-auto-scale.md
```diff
@@ -23,6 +23,8 @@ Implement auto storage scale CR per device class of a specific Storagecluster.
     spec:
         storageCluster: 
             name: <storageclusterName>
+            namespace: <storageclusterNamespace>
+            uid: <storageclusterUid>
```
travisn, line 27: I have a higher level question about this whole PR. Why is it needed? What is the precedence for adding such extensive checks on the reference? The `namespace` can be implied to be the same namespace as where this autoscaler CR is found, and why do we have to be so strict about the `uid`? 

Other CRs are not strict about referencing other CRs. For example, the cephblockpool, cephfilesystem, and all the other Rook CRs don't have a reference to the cephcluster CR. They just assume they apply to the single cephcluster CR in the same namespace. 

Let's consider the scenarios:
- New cluster install: Just specify the `name` of the storage cluster CR. (IMO even this is not necessary because we only support one storageCluster CR per namespace, right? But let's set this question aside for now.)
- Reinstall the cluster: When the cluster is reinstalled, if the autoscaler CR remains from the previous install, what is the harm? Probably they will even want the same autoscaler settings as the previous install. If they don't, they can just disable it when they recognize that it's still enabled. 

If we cannot define a clear scenario where the uid is needed, I recommend we abandon this PR.

travisn (changes requested):

docs/design/storage-auto-scale.md
parth-gr: @alfonsomthd can you add your thoughts

controllers/storageautoscaler/storageautoscaler_controller.go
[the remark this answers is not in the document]
umangachapagain: Other places can be changed later. This can be changed now. So why keep it for a follow-up?
Additionally, the suggestion to mark duplicate CR with "phase: error" makes it even simpler.

controllers/storageautoscaler/storageautoscaler_controller.go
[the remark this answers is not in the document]
umangachapagain: The whole idea of this PR is to match UID. How will you progress if UID is not set? What am I missing?

api/v1/storageautoscaler_types.go
umangachapagain: I believe the intent here was to be explicit (about which StorageCluster to scale) rather than implicit.

If we are ok being implicit, then what you suggested would work. There are many examples in K8s that follows this. I believe @travisn had raised a question on why we need to be explicit here. So maybe we can be implicit and use this approach.

docs/design/storage-auto-scale.md
umangachapagain: Agree with @travisn here.

Alternatively, https://github.com/red-hat-storage/ocs-operator/pull/3126#discussion_r2027593630 has some suggestions to solve it in a different way.

api/v1/storageautoscaler_types.go
parth-gr: @umangachapagain do you mean to only set the UID from the controller?

docs/design/storage-auto-scale.md
alfonsomthd: @travisn 
>When the cluster is reinstalled, if the autoscaler CR remains from the previous install, what is the harm? Probably they will even want the same autoscaler settings as the previous install. If they don't, they can just disable it when they recognize that it's still enabled.

IMHO that is unexpected behavior (is against the [principle of least surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment)): If I delete a cluster, and I re-create it WITHOUT enabling autoscaling, I don't expect to find out that it is enabled against my will.

To prevent this, currently the UI is cleaning up manually the `ui-convention-name` autoscaler CR during cluster creation (and recreating it if necessary) so the user can get what he/she selected on Day-1.

That said, if the current behavior is OK for everyone and nobody sees value in this effort, let's keep the current behavior.

travisn: My main concern is that we should not require the uid as part of the spec, as this prevents the CRs from being statically defined and deployed on multiple clusters (through automation, not UI install). So let's please not add the uid to the spec. 

> Alternatively, https://github.com/red-hat-storage/ocs-operator/pull/3126#discussion_r2027593630 has some suggestions to solve it in a different way.

While I'm not against this approach, IMO it's not worth the effort. Reinstall for customers is really not a common scenario. We need to support it, but reinstall is honestly more of a test scenario. And the side effects of autoscaling being enabled accidentally is very low severity, since most clusters will never even hit the 70% threshold for autoscaling anyway.

parth-gr: Okay closing this as per latest comments

PR #3128, opened 2025-04-03
send older subscription channel for older clients

leelavg:
when a 4.18 client asks for subscription change from 4.19 provider we send 4.18 susbscription till the storageconsumer is processed to be compatible with 4.19 resources.

### DIFF
--- a/services/provider/server/server.go
+++ b/services/provider/server/server.go
@@ -1471,9 +1471,25 @@ func (s *OCSProviderServer) ReportStatus(ctx context.Context, req *pb.ReportStat
 		return nil, status.Errorf(codes.Internal, "Failed to get storageConsumer resource: %v", err)
 	}
 
-	channelName, err := s.getOCSSubscriptionChannel(ctx)
+	clientOperatorVersion, err := semver.Parse(req.ClientOperatorVersion)
 	if err != nil {
-		return nil, status.Errorf(codes.Internal, "Failed to construct status response: %v", err)
+		return nil, status.Errorf(codes.InvalidArgument, "Malformed ClientOperatorVersion: %v", err)
+	}
+
+	channelName := ""
+	adjusted := strings.EqualFold(storageConsumer.GetAnnotations()[util.Is419AdjustedAnnotationKey], "true")
+	if !adjusted && clientOperatorVersion.Major == 4 && clientOperatorVersion.Minor == 18 {
+		// TODO (leelavg): need to be removed in 4.20
+		// We have a new controller which maps the resources from 4.18 to 4.19 way of management,
+		// until the resources are mapped we don't want connected client to be upgrading, we'll
+		// relax the condition on knowing the resources are mapped in a separate PR
+		channelName = "stable-4.18"
+	} else {
+		channel, err := s.getOCSSubscriptionChannel(ctx)
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "Failed to construct status response: %v", err)
+		}
+		channelName = channel
 	}
 
 	storageCluster, err := util.GetStorageClusterInNamespace(ctx, s.client, s.namespace)


controllers/util/storageconsumer.go
```diff
@@ -13,7 +13,9 @@ import (
 
 const (
 	// Reserved RadosNamespaceName for internal use and their representation at different layes
-	ImplicitRbdRadosNamespaceName = "<implicit>"
+	ImplicitRbdRadosNamespaceName                = "<implicit>"
+	StorageConsumer418ProcessedAnnotationKey     = "ocs.openshift.io/storageconsumer-4-18-processed"
+	StorageConsumer418ProcessedAnnotationPending = "pending"
```
nb-ohad, line 18: Please remove any annotation logic from this PR. We will introduce similar logic based on feedback coming from the upgrade controller implementation

services/provider/server/server.go
```diff
@@ -1468,6 +1468,17 @@ func (s *OCSProviderServer) ReportStatus(ctx context.Context, req *pb.ReportStat
...
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "Malformed ClientOperatorVersion: %v", err)
+	}
+	consumerProcessed, exist := storageConsumer.GetAnnotations()[util.StorageConsumer418ProcessedAnnotationKey]
+	if clientOperatorVersion.Major == 4 &&
+		clientOperatorVersion.Minor == 18 &&
+		(!exist || consumerProcessed == util.StorageConsumer418ProcessedAnnotationPending) {
+		channelName = strings.Replace(channelName, "4.19", "4.18", 1)
+	}
+
```
nb-ohad, line 1481: Please remove any annotation logic from this PR. We will introduce similar logic based on feedback coming from the upgrade controller implementation

services/provider/server/server.go
```diff
@@ -1468,6 +1468,17 @@ func (s *OCSProviderServer) ReportStatus(ctx context.Context, req *pb.ReportStat
...
 
+	clientOperatorVersion, err := semver.Parse(req.ClientOperatorVersion)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "Malformed ClientOperatorVersion: %v", err)
+	}
+	consumerProcessed, exist := storageConsumer.GetAnnotations()[util.StorageConsumer418ProcessedAnnotationKey]
+	if clientOperatorVersion.Major == 4 &&
+		clientOperatorVersion.Minor == 18 &&
+		(!exist || consumerProcessed == util.StorageConsumer418ProcessedAnnotationPending) {
+		channelName = strings.Replace(channelName, "4.19", "4.18", 1)
```
nb-ohad, line 1479: There is no reason to replace anything. just return the same value you read from the consumer status

nb-ohad (changes requested):

services/provider/server/server.go
leelavg: removed.

services/provider/server/server.go
leelavg: using hardcoded value `stable-4.18`

services/provider/server/server.go
```diff
@@ -1468,6 +1468,15 @@ func (s *OCSProviderServer) ReportStatus(ctx context.Context, req *pb.ReportStat
...
 
+	clientOperatorVersion, err := semver.Parse(req.ClientOperatorVersion)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "Malformed ClientOperatorVersion: %v", err)
+	}
+	if clientOperatorVersion.Major == 4 &&
+		clientOperatorVersion.Minor == 18 {
+		channelName = "stable-4.18"
+	}
+
```
nb-ohad, line 1479: - Let's not override the existing value captured in line 1466, but instead have an if else approach to selecting the channel name. 
- Add a TODO comment for this to be removed post 4.19

nb-ohad (changes requested):

services/provider/server/server.go
```diff
@@ -1463,8 +1463,16 @@ func (s *OCSProviderServer) ReportStatus(ctx context.Context, req *pb.ReportStat
...
+	channelName := ""
+	if clientOperatorVersion, err := semver.Parse(req.ClientOperatorVersion); err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "Malformed ClientOperatorVersion: %v", err)
+	} else if clientOperatorVersion.Major == 4 && clientOperatorVersion.Minor == 18 {
+		// TODO (leelavg): need to be removed in 4.20
+		// We have a new controller which maps the resources from 4.18 to 4.19 way of management,
+		// until the resources are mapped we don't want connected client to be upgrading, we'll
+		// relax the condition on knowing the resources are mapped in a separate PR
+		channelName = "stable-4.18"
+	} else if channelName, err = s.getOCSSubscriptionChannel(ctx); err != nil {
```
nb-ohad, line 1475: This if else is very convoluted, it mixes error handling and the 2 different cases we have. In some if statement you are relaying on filling in the desired channel name via an inline return value and the other branch via direct assignment. You are also using `else` after `if`s that do a direct return.

I propose a different way to structure the code.:

```

	if clientOperatorVersion, err := semver.Parse(req.ClientOperatorVersion); err != nil {
		return nil, status.Errorf(codes.InvalidArgument, "Malformed ClientOperatorVersion: %v", err)
	} 

	channelName := ""	
	if clientOperatorVersion.Major == 4 && clientOperatorVersion.Minor == 18 {
		// TODO (leelavg): need to be removed in 4.20
		// We have a new controller which maps the resources from 4.18 to 4.19 way of management,
		// until the resources are mapped we don't want connected client to be upgrading, we'll
		// relax the condition on knowing the resources are mapped in a separate PR
		channelName = "stable-4.18"
	} else {
		if channel, err := s.getOCSSubscriptionChannel(ctx); err != nil {
			return nil, status.Errorf(codes.Internal, "Failed to construct status response: %v",
		}
		channelName = channel
	}
```

this way the code is clearly separated into 2 different cases, one for 4.18 and another for all other

nb-ohad (changes requested):

nb-ohad:
We are moving desired channel reporting to GetStorageInfo in 4.19. This mean it does not make sense to merge code that return anything that is not 4.18 channel from within this code

controllers/util/storageconsumer.go
```diff
@@ -14,6 +14,7 @@ import (
 const (
 	// Reserved RadosNamespaceName for internal use and their representation at different layes
 	ImplicitRbdRadosNamespaceName = "<implicit>"
+	Is419AdjustedAnnotationKey    = "ocs.openshift.io/4_19-adjusted"
```
nb-ohad, line 17: Isn't this const already defined somewhere else in the codebase?

services/provider/server/server.go
```diff
@@ -1471,9 +1471,26 @@ func (s *OCSProviderServer) ReportStatus(ctx context.Context, req *pb.ReportStat
...
+		// We have a new controller which maps the resources from 4.18 to 4.19 way of management,
+		// until the resources are mapped we don't want connected client to be upgrading, we'll
+		// relax the condition on knowing the resources are mapped in a separate PR
+		channelName = "stable-4.18"
+	} else {
+		if channel, err := s.getOCSSubscriptionChannel(ctx); err != nil {
+			return nil, status.Errorf(codes.Internal, "Failed to construct status response: %v", err)
+		} else {
+			channelName = channel
+		}
```
nb-ohad, line 1493:
```suggestion
		} 
		
		channelName = channel
```

nb-ohad (changes requested):

controllers/util/storageconsumer.go
leelavg: yes, in another PR #3131

services/provider/server/server.go
leelavg: done. used that as earlier channel & err were being defined & used in the condition itself.

controllers/util/storageconsumer.go
leelavg: rebased.

leelavg:
rebased.

services/provider/server/server.go
```diff
@@ -1471,9 +1471,27 @@ func (s *OCSProviderServer) ReportStatus(ctx context.Context, req *pb.ReportStat
...
 	if err != nil {
-		return nil, status.Errorf(codes.Internal, "Failed to construct status response: %v", err)
+		return nil, status.Errorf(codes.InvalidArgument, "Malformed ClientOperatorVersion: %v", err)
+	}
+
+	channelName := ""
+	adjusted := strings.EqualFold(storageConsumer.GetAnnotations()[util.Is419AdjustedAnnotationKey], "true")
+	if !adjusted &&
+		clientOperatorVersion.Major == 4 &&
+		clientOperatorVersion.Minor == 18 {
```
nb-ohad, line 1483:
```suggestion
	if !adjusted && clientOperatorVersion.Major == 4 && clientOperatorVersion.Minor == 18 {
```

PR #3129, opened 2025-04-03
do not reconcile older storageconsumer

leelavg:
a separate controller will process older storageconsumer owned resources and until then storageconsumer controller should not proceed with reconcililation.

### DIFF
--- a/controllers/storageconsumer/storageconsumer_controller.go
+++ b/controllers/storageconsumer/storageconsumer_controller.go
@@ -32,6 +32,7 @@ import (
 	"github.com/red-hat-storage/ocs-operator/v4/controllers/defaults"
 	"github.com/red-hat-storage/ocs-operator/v4/controllers/util"
 
+	"github.com/blang/semver/v4"
 	"github.com/go-logr/logr"
 	nbv1 "github.com/noobaa/noobaa-operator/v5/pkg/apis/noobaa/v1alpha1"
 	rookCephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
@@ -159,6 +160,23 @@ func (r *StorageConsumerReconciler) reconcileEnabledPhases() (reconcile.Result, 
 		storageCluster, err := util.GetStorageClusterInNamespace(r.ctx, r.Client, r.namespace)
 		if err != nil {
 			return reconcile.Result{}, err
+		}
+
+		if r.storageConsumer.Status.Client.OperatorVersion == "" {
+			return reconcile.Result{}, fmt.Errorf("client does not expose operator version in its status")
+		}
+
+		clientOperatorVersion, err := semver.Parse(r.storageConsumer.Status.Client.OperatorVersion)
+		if err != nil {
+			return reconcile.Result{}, fmt.Errorf("malformed ClientOperatorVersion: %v", err)
+		}
+		if clientOperatorVersion.Major == 4 &&
+			clientOperatorVersion.Minor == 18 {
+			// TODO (leelavg): need to be removed in 4.20
+			// We have a new controller which maps the resources from 4.18 to 4.19 way of management,
+			// until the resources are mapped we don't want to reconcile consumer belonging to connected client
+			// relax the condition on knowing the resources are mapped in a separate PR
+			return reconcile.Result{}, nil
 		}
 
 		availableServices, err := util.GetAvailableServices(r.ctx, r.Client, storageCluster)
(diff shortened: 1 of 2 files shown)


controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -161,6 +162,19 @@ func (r *StorageConsumerReconciler) reconcileEnabledPhases() (reconcile.Result,
 			return reconcile.Result{}, err
 		}
 
+		if storageCluster.Spec.AllowRemoteStorageConsumers {
```
nb-ohad, line 165: Why do we need this check here. It is implicit that if we have a pre 4.19 consumer then we don't proceed

nb-ohad (changes requested):

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -161,6 +162,19 @@ func (r *StorageConsumerReconciler) reconcileEnabledPhases() (reconcile.Result,
...
+			if err != nil {
+				return reconcile.Result{}, fmt.Errorf("malformed ClientOperatorVersion: %v", err)
+			}
+			consumerProcessed, exist := r.storageConsumer.GetAnnotations()[util.StorageConsumer418ProcessedAnnotationKey]
+			if clientOperatorVersion.Major == 4 &&
+				clientOperatorVersion.Minor == 18 &&
+				(!exist || consumerProcessed == util.StorageConsumer418ProcessedAnnotationPending) {
+				return reconcile.Result{}, nil
+			}
+		}
```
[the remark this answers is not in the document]
leelavg: removed.

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -756,7 +770,12 @@ func (r *StorageConsumerReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	)
 
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&ocsv1alpha1.StorageConsumer{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		For(&ocsv1alpha1.StorageConsumer{}, builder.WithPredicates(
+			predicate.Or(
+				predicate.GenerationChangedPredicate{},
+				predicate.AnnotationChangedPredicate{},
```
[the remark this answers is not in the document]
leelavg: removed.

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -161,6 +162,15 @@ func (r *StorageConsumerReconciler) reconcileEnabledPhases() (reconcile.Result,
 			return reconcile.Result{}, err
 		}
 
+		clientOperatorVersion, err := semver.Parse(r.storageConsumer.Status.Client.OperatorVersion)
+		if err != nil {
+			return reconcile.Result{}, fmt.Errorf("malformed ClientOperatorVersion: %v", err)
+		}
```
nb-ohad, line 168: Do we have a guarantee that his exists from the second the consumer is enabled? Or can it be that this value will be empty until the first heartbeat?

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -161,6 +162,15 @@ func (r *StorageConsumerReconciler) reconcileEnabledPhases() (reconcile.Result,
...
+		if clientOperatorVersion.Major == 4 &&
```
nb-ohad, line 169: Wee need a TODO comment that say that this code should be removed post 4.19

nb-ohad (changes requested):

controllers/storageconsumer/storageconsumer_controller.go
leelavg: empty until first hearbeat, in earlier revision I was guarding this w/ `allowRemoteStorageConsumers` as older clients would've this set and version is always available.

leelavg: added check for emptiness.

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -161,6 +162,23 @@ func (r *StorageConsumerReconciler) reconcileEnabledPhases() (reconcile.Result,
 			return reconcile.Result{}, err
 		}
 
+		if r.storageConsumer.Status.Client.OperatorVersion == "" {
+			return reconcile.Result{}, fmt.Errorf("awaiting connected client to report heartbeat for it's operatorversion")
```
nb-ohad, line 166:
```suggestion
			return reconcile.Result{}, fmt.Errorf("client does not expose operator version in its status")
```

nb-ohad (changes requested):

PR #3130, opened 2025-04-03
storageautoscaler: add a invalid phase in cr

parth-gr:
Add a new invalid phase:

1)  if more than one storage autoscaler present with same
device class and same storage cluster, do not reconcile

2) If lso storageclass is detected

3) if lean profile is detected

### DIFF
--- a/api/v1/storageautoscaler_types.go
+++ b/api/v1/storageautoscaler_types.go
@@ -52,6 +52,7 @@ type StorageAutoScalerSpec struct {
 type StorageAutoScalerPhase string
 
 const (
+	StorageAutoScalerPhaseInvalid    StorageAutoScalerPhase = "Invalid"
 	StorageAutoScalerPhaseNotStarted StorageAutoScalerPhase = "NotStarted"
 	StorageAutoScalerPhaseInProgress StorageAutoScalerPhase = "InProgress"
 	StorageAutoScalerPhaseSucceeded  StorageAutoScalerPhase = "Succeeded"
--- a/controllers/storageautoscaler/storageautoscaler_controller.go
+++ b/controllers/storageautoscaler/storageautoscaler_controller.go
@@ -69,15 +69,18 @@ func (r *StorageAutoscalerReconciler) Reconcile(ctx context.Context, request rec
 
 	// if more than one storage autoscaler is present in the cluster with same device class and same storage cluster
 	// return error
-	err = r.detectDuplicateStorageAutoScaler(ctx, storageAutoScaler, request.Namespace)
+	duplicateCr, err := r.detectDuplicateStorageAutoScaler(ctx, storageAutoScaler, request.Namespace)
 	if err != nil {
 		return reconcile.Result{}, err
+	}
+	if duplicateCr {
+		return reconcile.Result{}, nil
 	}
 
 	// if no expansion is in-progress update the phase to "not-started"
 	if storageAutoScaler.Status.Phase == "" {
 		originalStorageAutoScaler := storageAutoScaler.DeepCopy()
-		storageAutoScaler.Status.Phase = "NotStarted"
+		storageAutoScaler.Status.Phase = ocsv1.StorageAutoScalerPhaseNotStarted
 		err := r.updateStatus(ctx, storageAutoScaler, originalStorageAutoScaler)
 		if err != nil {
 			return reconcile.Result{}, err
@@ -85,7 +88,7 @@ func (r *StorageAutoscalerReconciler) Reconcile(ctx context.Context, request rec
 	}
 
 	// if the expansion is InProgress or Failed, verify the scaling
-	if storageAutoScaler.Status.Phase == "InProgress" || storageAutoScaler.Status.Phase == "Failed" {
+	if storageAutoScaler.Status.Phase == ocsv1.StorageAutoScalerPhaseInProgress || storageAutoScaler.Status.Phase == ocsv1.StorageAutoScalerPhaseFailed {
 		return r.verifyScaling(ctx, storageAutoScaler)
 	}
 
@@ -109,9 +112,12 @@ func (r *StorageAutoscalerReconciler) Reconcile(ctx context.Context, request rec
 	}
 
 	// list the storagecluster deviceset storageclass
-	err = r.detectLsoStorageclass(ctx, storageCluster)
+	duplicateClass, err := r.detectLsoStorageclass(ctx, storageCluster)
 	if err != nil {
 		return reconcile.Result{}, err
+	}
+	if duplicateClass {
+		return reconcile.Result{}, nil
 	}
 
 	// check if the ceph cluster is healthy
@@ -170,7 +176,7 @@ func (r *StorageAutoscalerReconciler) Reconcile(ctx context.Context, request rec
 
 	// update the status
 	originalStorageAutoScaler := storageAutoScaler.DeepCopy()
-	storageAutoScaler.Status.Phase = "InProgress"
+	storageAutoScaler.Status.Phase = ocsv1.StorageAutoScalerPhaseInProgress
 	storageAutoScaler.Status.Error = nil
 	storageAutoScaler.Status.LastExpansion = &ocsv1.LastExpansionStatus{
 		StartOsdCount:           uint16(startOsdCount),
@@ -209,7 +215,7 @@ func (r *StorageAutoscalerReconciler) Reconcile(ctx context.Context, request rec
 	return r.verifyScaling(ctx, storageAutoScaler)
 }
 
-func (r *StorageAutoscalerReconciler) detectLsoStorageclass(ctx context.Context, storageCluster *ocsv1.StorageCluster) error {
+func (r *StorageAutoscalerReconciler) detectLsoStorageclass(ctx context.Context, storageCluster *ocsv1.StorageCluster) (bool, error) {
 	for _, deviceSet := range storageCluster.Spec.StorageDeviceSets {
 		storageclassName := deviceSet.DataPVCTemplate.Spec.StorageClassName
 		// list the storageclass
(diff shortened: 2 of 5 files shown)


BlaineEXE:
From discussion in huddle, I feel like the work to 'bind' an autoscale resource to a StorageCluster in https://github.com/red-hat-storage/ocs-operator/pull/3126 will do quite a bit to help make this situation less likely to occur. I'd propose revisiting the design of this work afterwards, once we understand the narrowed scope a bit more

travisn (commented):
From our discussion last week, I still don't understand why we need to check for this corner case? We don't do it for other CRs such as CephBlockPool. What is the severe side effect for this CR that requires checking?

parth-gr:
> From our discussion last week, I still don't understand why we need to check for this corner case? We don't do it for other CRs such as CephBlockPool. What is the severe side effect for this CR that requires checking?

I went ahead as the fix was simple and direct. We can consider not doing it, but it didn't add any extra API cost as it queries controller runtime cache

api/v1/storageautoscaler_types.go
```diff
@@ -115,6 +115,8 @@ type StorageAutoScalerStatus struct {
 	// +optional
 	// +nullable
 	StorageCapacityLimitReached *bool `json:"storageCapacityLimitReached,omitempty"`
+	// CrsWithDuplicateSpecDetected is the flag that indicates if the CRs with duplicate spec have been detected.
+	CrsWithDuplicateSpecDetected *bool `json:"crsWithDuplicateSpecDetected,omitempty"`
```
travisn, line 119: No need for a new field, just set the phase to error, and set the error message.

travisn (changes requested):
> I went ahead as the fix was simple and direct. We can consider not doing it, but it didn't add any extra API cost as it queries controller runtime cache

Let's just keep it simple and only set it as an error, no need for a new field every time there is a different type of error.

api/v1/storageautoscaler_types.go
parth-gr: We need separate metrics for the alert, as decided with @umangachapagain, 
As the error string can't be passed while raising alerts

controllers/storageautoscaler/storageautoscaler_controller.go
```diff
@@ -67,6 +67,30 @@ func (r *StorageAutoscalerReconciler) Reconcile(ctx context.Context, request rec
...
+	if len(storageAutoScalerList.Items) > 1 {
+		err = fmt.Errorf("more than one storage autoscaler present with same device class and same storage cluster")
+		r.Log.Error(err, "duplicate cr detected", "device class", storageAutoScaler.Spec.DeviceClass, "storage cluster", storageAutoScaler.Spec.StorageCluster.Name)
+
+		// update the status
+		originalStorageAutoScaler := storageAutoScaler.DeepCopy()
+		duplicateCrsDetected := true
+		storageAutoScaler.Status.CrsWithDuplicateSpecDetected = &duplicateCrsDetected
+		err := r.updateStatus(ctx, storageAutoScaler, originalStorageAutoScaler)
+		return reconcile.Result{}, err
```
aruniiird, line 91: Which error should be returned , `error` in line#83 or the update status `error` in line#90?

api/v1/storageautoscaler_types.go
parth-gr: Okay we can just change the error message

umangachapagain: My suggestion was to have a new phase `unknown/invalid` with some error message to distinguish between scaling errors and user errors. But, setting phase `error` is fine as well. We do not need a new boolean field.
Ideally these cases could be filtered out by admission webhooks but we don't have that in ocs-operator.

docs/design/storage-auto-scale.md
```diff
@@ -34,6 +34,7 @@ Implement auto storage scale CR per device class of a specific Storagecluster.
             message: ""
             timestamp: <time>
         storageCapacityLimitReached: false
+        CrsWithDuplicateSpecDetected: false
```
travisn, line 37: revert?

controllers/storageautoscaler/storageautoscaler_controller.go
```diff
@@ -67,6 +67,31 @@ func (r *StorageAutoscalerReconciler) Reconcile(ctx context.Context, request rec
...
+	)
+	if err != nil {
+		r.Log.Error(err, "failed to list storage autoscaler")
+		return reconcile.Result{}, err
+	}
+
+	if len(storageAutoScalerList.Items) > 1 {
+		// update the status
+		originalStorageAutoScaler := storageAutoScaler.DeepCopy()
+		err := r.updateStatus(ctx, storageAutoScaler, originalStorageAutoScaler)
```
travisn, line 85: Don't we need to update the CR status after applying the error state?

controllers/storageautoscaler/storageautoscaler_controller.go
```diff
@@ -67,6 +67,31 @@ func (r *StorageAutoscalerReconciler) Reconcile(ctx context.Context, request rec
...
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+
+		err = fmt.Errorf("more than one storage autoscaler present with same device class and same storage cluster")
```
travisn, line 90: Need to format in more details to this error message so it's returned in the reconcile failure?

travisn (changes requested):

controllers/storageautoscaler/storageautoscaler_controller.go
```diff
@@ -67,6 +67,31 @@ func (r *StorageAutoscalerReconciler) Reconcile(ctx context.Context, request rec
 		return reconcile.Result{}, err
 	}
 
+	// if more than one storage autoscaler is present in the cluster with same device class and same storage cluster
+	// return error
+	storageAutoScalerList := &ocsv1.StorageAutoScalerList{}
+	err = r.List(ctx, storageAutoScalerList,
+		client.InNamespace(request.Namespace),
+		client.MatchingFields{"spec.storageCluster.name": storageAutoScaler.Spec.StorageCluster.Name, "spec.deviceClass": storageAutoScaler.Spec.DeviceClass},
```
umangachapagain, line 75: Does this work? Our CustomerResources don't support field selection.

parth-gr: Yes it didnt worked we need to ` indexed fields in the cache` maybe I can list all the crs and query it

controllers/storageautoscaler/storageautoscaler_controller.go
parth-gr: Yes, but we need to introduce a new `Invalid`  phase will do it as a follow up pr

controllers/storageautoscaler/storageautoscaler_controller.go
parth-gr: We decided to not return the error in case of invalid phase.

parth-gr:
Testing: 
```
{"level":"error","ts":"2025-04-21T12:03:42Z","logger":"controllers.StorageAutoScaler","msg":"duplicate cr detected","device class":"ssd","storage cluster":"ocs-storagecluster","error":"more than one storage autoscaler present with same device class and same storage cluster name","stacktrace":"github.com/red-hat-storage/ocs-operator/v4/controllers/storageautoscaler.(*StorageAutoscalerReconciler).Reconcile\n\t/workspace/controllers/storageautoscaler/storageautoscaler_controller.go:91\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Reconcile\n\t/workspace/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:118\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler\n\t/workspace/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:328\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem\n\t/workspace/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:288\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func2.2\n\t/workspace/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:249"}

```

controllers/storageautoscaler/storageautoscaler_controller.go
```diff
@@ -67,6 +67,32 @@ func (r *StorageAutoscalerReconciler) Reconcile(ctx context.Context, request rec
 		return reconcile.Result{}, err
 	}
 
+	// if more than one storage autoscaler is present in the cluster with same device class and same storage cluster
```
travisn, line 70: Could factor this error check to a helper method? Really long methods are more difficult to maintain in the long term.

parth-gr: DOne

controllers/storageautoscaler/storageautoscaler_controller.go
```diff
@@ -300,3 +307,31 @@ func (r *StorageAutoscalerReconciler) updateStatus(ctx context.Context, storageA
...
+		client.InNamespace(namespace),
+	)
+	if err != nil {
+		r.Log.Error(err, "failed to list storage autoscaler")
+		return err
+	}
+
+	for _, storageAutoScalerItem := range storageAutoScalerList.Items {
+		if storageAutoScalerItem.Name != storageAutoScaler.Name && storageAutoScalerItem.Spec.DeviceClass == storageAutoScaler.Spec.DeviceClass && storageAutoScalerItem.Spec.StorageCluster.Name == storageAutoScaler.Spec.StorageCluster.Name {
+			// update the status
```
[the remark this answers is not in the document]
travisn: I don't see the phase updated to either one. What do you propose? Either it's a failed phase, or an invalid phase if you want to differentiate misconfiguration from a failed autoscale.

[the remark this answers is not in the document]
parth-gr: OKay lets go with  `invalid phase` but thats has to be done as a follow up as its needs crd changes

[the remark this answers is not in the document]
travisn: How about completing the other PR first? No need to merge this PR before it's fully implemented.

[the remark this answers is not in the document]
parth-gr: DOne with in this PR

travisn (commented):
Could unit tests be added to check for these invalid states? I'll approve in case you need to add the tests in a follow-up PR to unblock other items. 
/lgtm

travisn:
/hold
Remove the hold if unit tests will be done separately.

parth-gr:
/unhold 

I didn't thought of adding unit test, but as requested I can add unit test with https://github.com/red-hat-storage/ocs-operator/pull/3179

PR #3133, opened 2025-04-04
provider: implement a new rpc GetStorageInfo for the new hub & spoke to be used from 4.19 onwards

rewantsoni:
Introduce a new rpc GetStorageInfo that will be used by the new hub & spoke from 4.19 onwards the old rpc will act as a shim layer for clients on version 4.18


services/provider/proto/provider.proto
```diff
@@ -15,6 +15,9 @@ service OCSProvider {
   // GetStorageConfig RPC call to generate the json config for connecting to storage provider cluster
   rpc GetStorageConfig(StorageConfigRequest)
   returns (StorageConfigResponse){}
+  // GetStorageInfo RPC call to generate the json config for connecting to storage provider cluster
+  rpc GetStorageInfo(StorageInfoRequest)
```
nb-ohad, line 19: As this is a new RPC function, can we call it getClientDesiredState? This will also eliminate the need to keep on adding a desired prefix to members

Of course all other names need to align with this rename

services/provider/proto/provider.proto
```diff
@@ -109,6 +118,21 @@ message StorageConfigResponse{
...
+  string desiredSubscriptionChannel = 3;
+}
+
+// StorageInfoResponse holds the response for the GetStorageInfo API request
+message StorageInfoResponse{
+  // ExternalResource holds the configuration data of external storage cluster
+  repeated bytes kubeResources = 1;
+  // Contains hash of desired config
+  string desiredConfigHash = 2;
+  ClientAttributes clientAttributes = 3;
```
nb-ohad, line 133: Now that everything is typed, is there really a need for this indirection?

services/provider/proto/provider.proto
```diff
@@ -109,6 +118,21 @@ message StorageConfigResponse{
     SystemAttributes systemAttributes = 3;
 }
 
+message ClientAttributes {
+  bool inMaintenanceMode = 1;
+  bool mirrorEnabled = 2;
+  string desiredSubscriptionChannel = 3;
```
nb-ohad, line 124:
```suggestion
  string subscriptionChannel = 3;
```

services/provider/proto/provider.proto
```diff
@@ -109,6 +118,21 @@ message StorageConfigResponse{
     SystemAttributes systemAttributes = 3;
 }
 
+message ClientAttributes {
+  bool inMaintenanceMode = 1;
```
nb-ohad, line 122:
```suggestion
  bool maintenanceMode = 1;
```

nb-ohad (changes requested):

services/provider/proto/provider.proto
rewantsoni: Do we need the clientAttributes to be directly embedded into the GetDesiredClientStateResponse?

services/provider/proto/provider.proto
```diff
@@ -109,6 +118,15 @@ message StorageConfigResponse{
     SystemAttributes systemAttributes = 3;
 }
 
+// GetDesiredClientStateResponse holds the response for the GetDesiredClientState API request
+message GetDesiredClientStateResponse{
+  repeated bytes kubeResources = 1;
+  string desiredConfigHash = 2;
```
nb-ohad, line 124:
```suggestion
  string desiredStateHash = 2;
```

services/provider/proto/provider.proto
```diff
@@ -109,6 +118,15 @@ message StorageConfigResponse{
...
+  string subscriptionChannel = 3;
```
nb-ohad, line 125:
```suggestion
  clientOperatorChannel
```

nb-ohad (changes requested):

nb-ohad:
Overriding e2e as the only change this PR contains is the introduction of a new gRPC call that is not yet in use

PR #3136, opened 2025-04-08
Adding an alert to detect StorageAutoScaler CRs in invalid state

aruniiird:
https://issues.redhat.com/browse/DFBUGS-2048

### DIFF
--- a/metrics/deploy/prometheus-ocs-rules.yaml
+++ b/metrics/deploy/prometheus-ocs-rules.yaml
@@ -449,3 +449,15 @@ spec:
       for: 5m
       labels:
         severity: warning
+    - alert: StorageAutoScalerCRIsInvalid
+      annotations:
+        description: |-
+          Invalid phase detected for StorageAutoScaler {{ $labels.name }} attached with device class {{ $labels.device_class }} under storage cluster {{ $labels.storagecluster }} in the namespace {{ $labels.namespace }}.
+        message: StorageAutoScaler is in an invalid state.
+        runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/openshift-container-storage-operator/StorageAutoScalerCRIsInvalid.md
+        severity_level: warning
+      expr: |
+        ocs_storageautoscaler_phase_status == 5
+      for: 30s
+      labels:
+        severity: warning
--- a/metrics/internal/collectors/storage-auto-scaler.go
+++ b/metrics/internal/collectors/storage-auto-scaler.go
@@ -25,6 +25,7 @@ const (
 	InProgressPhaseStatus
 	SucceededPhaseStatus
 	FailedPhaseStatus
+	InvalidPhaseStatus
 )
 
 type StorageAutoScalerCollector struct {
@@ -76,6 +77,8 @@ func (s *StorageAutoScalerCollector) collectStorageAutoScalerPhaseStatuses(ch ch
 			phaseStatus = succeededPhaseValidation(storageAutoScaler.Status.LastExpansion.CompletionTime)
 		case v1.StorageAutoScalerPhaseInProgress:
 			phaseStatus = InProgressPhaseStatus
+		case v1.StorageAutoScalerPhaseInvalid:
+			phaseStatus = InvalidPhaseStatus
 		default:
 			phaseStatus = NotStartedPhaseStatus
 		}
@@ -139,7 +142,8 @@ func NewStorageAutoScalerCollector(opts *options.Options) *StorageAutoScalerColl
 	return &StorageAutoScalerCollector{
 		AutoScalerPhaseStatus: prometheus.NewDesc(
 			prometheus.BuildFQName(namespace, "storageautoscaler", "phase_status"),
-			fmt.Sprintf("Phases are; Not Started: %d, In Progress: %d, Succeessfull: %d, Failed: %d", NotStartedPhaseStatus, InProgressPhaseStatus, SucceededPhaseStatus, FailedPhaseStatus),
+			fmt.Sprintf("Phases are; Not Started: %d, In Progress: %d, Succeessfull: %d, Failed: %d, Invalid: %d",
+				NotStartedPhaseStatus, InProgressPhaseStatus, SucceededPhaseStatus, FailedPhaseStatus, InvalidPhaseStatus),
 			[]string{"name", "namespace", "storagecluster", "device_class"},
 			nil,
 		),


metrics/deploy/prometheus-ocs-rules.yaml
```diff
@@ -449,3 +449,15 @@ spec:
       for: 5m
       labels:
         severity: warning
+    - alert: DuplicateStorageAutoScalerCRsFound
+      annotations:
+        description: |-
+          Duplicate StorageAutoScaler CRs found in the namespace {{ $labels.namespace }} trying to scale device class {{ $labels.device_class }} under storage cluster {{ $labels.storagecluster }}.
+        message: Duplicate StorageAutoScaler CRs found.
+        runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/openshift-container-storage-operator/DuplicateStorageAutoScalerCRsFound.md
```
aruniiird, line 457: Run books URL yet to be created.
@parth-gr , @ShravaniVangur , please take a look

metrics/deploy/prometheus-ocs-rules.yaml
```diff
@@ -449,3 +449,15 @@ spec:
...
+        severity_level: warning
+      expr: |
+        count by (namespace, managedBy, storagecluster, device_class) (ocs_storageautoscaler_capacity_reached) > 1
```
parth-gr, line 460: I thought we would be creating separate metrics for duplicate CR by a new `duplicate: true` status spec field, @umangachapagain, what are your thoughts?

aruniiird: IMHO, we don't really require a separate backend implementation (like this [PR#3130](https://github.com/red-hat-storage/ocs-operator/pull/3130)). This is a common thing to do through prometheus query.

If we are going with the [PR#3130](https://github.com/red-hat-storage/ocs-operator/pull/3130) backend approach, we really don't require a separate alert. [When we detect duplicate CRs in the backend] We just have to update the CR with a failure status and an alert will be automatically fired (as we have alert for the failure cases).

parth-gr: Okay lets go with the current approach

metrics/deploy/prometheus-ocs-rules.yaml
parth-gr: @ShravaniVangur can you work upon the runbooks

metrics/deploy/prometheus-ocs-rules.yaml
umangachapagain: This approach is better.
If backend sets the error phase, we'll probably have 2 alerts. Do we want that?
Or do you want to handle scaling error vs duplicate CRs in runbook?

parth-gr: From the backend we wont set it as failed state, but will introduce a new state as Invalid

metrics/deploy/prometheus-ocs-rules.yaml
```diff
@@ -449,3 +449,15 @@ spec:
...
+      for: 1m
+      labels:
+        severity: warning
```
parth-gr, line 463: @aruniiird  @umangachapagain  wont this be critical? As we have two crs that are scaling at the same time

parth-gr (changes requested):

metrics/deploy/prometheus-ocs-rules.yaml
umangachapagain: No since you'd be marking duplicates as invalid and not scaling those.

parth-gr: I see, got your point

parth-gr (approved):
/lgtm

aruniiird:
@umangachapagain , if everything is good, can we tag this

metrics/deploy/prometheus-ocs-rules.yaml
```diff
@@ -449,3 +449,15 @@ spec:
       for: 5m
       labels:
         severity: warning
+    - alert: DuplicateStorageAutoScalerCRsFound
+      annotations:
+        description: |-
+          Duplicate StorageAutoScaler CRs found in the namespace {{ $labels.namespace }} trying to scale device class {{ $labels.device_class }} under storage cluster {{ $labels.storagecluster }}.
```
umangachapagain, line 455:
```suggestion
          Multiple StorageAutoScaler resources found in the namespace {{ $labels.namespace }} trying to scale device class {{ $labels.device_class }} under storage cluster {{ $labels.storagecluster }}.
```

metrics/deploy/prometheus-ocs-rules.yaml
```diff
@@ -449,3 +449,15 @@ spec:
...
+        message: Duplicate StorageAutoScaler CRs found.
```
umangachapagain, line 456:
```suggestion
        message: Multiple StorageAutoScaler resources found for the same target.
```

umangachapagain (commented):
@aruniiird I have some nitpicks.

travisn (commented):
Do we have similar duplicate alerts for other CRs? This seems like a corner case that wouldn't require an alert. IMO alerts are best for cluster health issues, rather than misconfigurations. Instead of a specific alert for this corner case, seems like we should just look at the error phase and message on the CR and treat it as a generic error. Can we raise an error when there is any error phase/message on the CR? Maybe the UI is already doing this, and no need for an alert?

parth-gr:
> Do we have similar duplicate alerts for other CRs? This seems like a corner case that wouldn't require an alert. IMO alerts are best for cluster health issues, rather than misconfigurations. Instead of a specific alert for this corner case, seems like we should just look at the error phase and message on the CR and treat it as a generic error. Can we raise an error when there is any error phase/message on the CR? Maybe the UI is already doing this, and no need for an alert?

@travisn we are not changing to cr to error/failed state, but we are changing the cr state to `invalid state`, that's a new phase which will be set when there a not valid state,

Invalid state can happen:

1) Duplicate CRS
2) LSO sc detected

travisn:
> > Do we have similar duplicate alerts for other CRs? This seems like a corner case that wouldn't require an alert. IMO alerts are best for cluster health issues, rather than misconfigurations. Instead of a specific alert for this corner case, seems like we should just look at the error phase and message on the CR and treat it as a generic error. Can we raise an error when there is any error phase/message on the CR? Maybe the UI is already doing this, and no need for an alert?
> 
> @travisn we are not changing to cr to error/failed state, but we are changing the cr state to `invalid state`, that's a new phase which will be set when there a not valid state,
> 
> Invalid state can happen:
> 
> 1. Duplicate CRS
> 2. LSO sc detected
> 3. Lean mode

Ok, so the `phase` will be set to `Invalid`? To differentiate from the phase of `Failed`, which means an auto scale was attempted? That seems reasonable to me, though I still don't see the need for an alert for such a misconfiguration. Let's discuss.

parth-gr:
@travisn as we have the separate invalid phase https://github.com/red-hat-storage/ocs-operator/pull/3130, 
Lets go ahead with a `info` level alert for this, wdyt?

travisn:
Agreed, sounds better to have a single alert for any invalid config.

parth-gr:
@aruniiird  please go ahead and make the changes

@ShravaniVangur  please adopt to it accordingly, we have a invalid alert,

Which can be caused by :

1) Duplicate cr

2) lean mode detected

3) Lso storageclass detected

parth-gr:
cc @aruniiird ^^

parth-gr:
@aruniiird we have a new status field, `Invalid` https://github.com/red-hat-storage/ocs-operator/blob/main/api/v1/storageautoscaler_types.go#L55

Try use this for raising this alert

metrics/deploy/prometheus-ocs-rules.yaml
aruniiird: Done

metrics/deploy/prometheus-ocs-rules.yaml
aruniiird: Done

aruniiird:
> @aruniiird we have a new status field, `Invalid` https://github.com/red-hat-storage/ocs-operator/blob/main/api/v1/storageautoscaler_types.go#L55
> 
> Try use this for raising this alert

@parth-gr , does that mean , we need to rename the alert to `InvalidStorageAutoScalerCRsFound` ?
Another check,
When a duplicate AutoScaler CR is created, which one will be marked invalid? The newly created second CR only or both the CRs (first and second both)?

parth-gr:
> > @aruniiird we have a new status field, `Invalid` https://github.com/red-hat-storage/ocs-operator/blob/main/api/v1/storageautoscaler_types.go#L55
> > Try use this for raising this alert
> 
> @parth-gr , does that mean , we need to rename the alert to `InvalidStorageAutoScalerCRsFound` ? Another check, When a duplicate AutoScaler CR is created, which one will be marked invalid? The newly created second CR only or both the CRs (first and second both)?

When the duplicate cr is created it will be marked as invalid, and later if there is any reconcile for the older cr, then it will be also marked as invalid

aruniiird:
Made the changes. @parth-gr , please take a look.

metrics/deploy/prometheus-ocs-rules.yaml
```diff
@@ -449,3 +449,15 @@ spec:
...
+        description: |-
+          Invalid phase detected for StorageAutoScaler {{ $labels.name }} attached with device class {{ $labels.device_class }} under storage cluster {{ $labels.storagecluster }} in the namespace {{ $labels.namespace }}.
+        message: StorageAutoScaler is in an invalid state.
+        runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/openshift-container-storage-operator/InvalidStorageAutoScalerCRsFound.md
+        severity_level: warning
+      expr: |
+        ocs_storageautoscaler_phase_status == 5
+      for: 1m
+      labels:
+        severity: warning
```
aruniiird, line 463: @parth-gr , currently severity/severity-level is in `warning` level. (As per one of your previous comments) Do you want to change it to an `info` level alert? Please let me know

metrics/deploy/prometheus-ocs-rules.yaml
```diff
@@ -449,3 +449,15 @@ spec:
       for: 5m
       labels:
         severity: warning
+    - alert: InvalidStorageAutoScalerCRsFound
+      annotations:
+        description: |-
+          Invalid phase detected for StorageAutoScaler {{ $labels.name }} attached with device class {{ $labels.device_class }} under storage cluster {{ $labels.storagecluster }} in the namespace {{ $labels.namespace }}.
+        message: StorageAutoScaler is in an invalid state.
+        runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/openshift-container-storage-operator/InvalidStorageAutoScalerCRsFound.md
```
aruniiird, line 457: @parth-gr , @ShravaniVangur , please note that I've changed the previous name (DuplicateStorageAutoScalerCRsFound.md) to `InvalidStorageAutoScalerCRsFound.md`. We need to make corresponding change to the runbook side.
Or let me know if we have decided on any other name for the file, will make the changes here.

metrics/deploy/prometheus-ocs-rules.yaml
parth-gr: I am fine with Info level, 
cc @travisn if you have any other opinion for it

metrics/deploy/prometheus-ocs-rules.yaml
parth-gr: Okay thanks

metrics/deploy/prometheus-ocs-rules.yaml
```diff
@@ -449,3 +449,15 @@ spec:
       for: 5m
       labels:
         severity: warning
+    - alert: InvalidStorageAutoScalerCRsFound
```
parth-gr, line 452: @aruniiird how about `StorageAutoScalerCRIsInvalid`?

aruniiird: Ok, made the change...
PS: was trying to avoid `Is` / `Was` and other verbs (=

metrics/deploy/prometheus-ocs-rules.yaml
aruniiird: Just an FYI: made the change to the alert name, as @parth-gr suggested, `StorageAutoScalerCRIsInvalid` ... So made the same change to the MD file name as well

metrics/deploy/prometheus-ocs-rules.yaml
```diff
@@ -449,3 +449,15 @@ spec:
...
+        description: |-
+          Invalid phase detected for StorageAutoScaler {{ $labels.name }} attached with device class {{ $labels.device_class }} under storage cluster {{ $labels.storagecluster }} in the namespace {{ $labels.namespace }}.
+        message: StorageAutoScaler is in an invalid state.
+        runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/openshift-container-storage-operator/StorageAutoScalerCRIsInvalid.md
+        severity_level: info
+      expr: |
+        ocs_storageautoscaler_phase_status == 5
+      for: 1m
+      labels:
+        severity: info
```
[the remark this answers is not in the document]
parth-gr: cc @aruniiird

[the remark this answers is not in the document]
aruniiird: Done

parth-gr (approved):
/lgtm

aruniiird:
@umangachapagain , @iamniting , can you please take a look? If no more changes needed, can we merge this PR?

metrics/deploy/prometheus-ocs-rules.yaml
```diff
@@ -449,3 +449,15 @@ spec:
...
+    - alert: StorageAutoScalerCRIsInvalid
+      annotations:
+        description: |-
+          Invalid phase detected for StorageAutoScaler {{ $labels.name }} attached with device class {{ $labels.device_class }} under storage cluster {{ $labels.storagecluster }} in the namespace {{ $labels.namespace }}.
+        message: StorageAutoScaler is in an invalid state.
+        runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/openshift-container-storage-operator/StorageAutoScalerCRIsInvalid.md
+        severity_level: warning
+      expr: |
+        ocs_storageautoscaler_phase_status == 5
+      for: 1m
```
umangachapagain, line 461: This can be lowered to 30s.

aruniiird: Done

umangachapagain (approved):

PR #3140, opened 2025-04-08
skip creation of local storageconsumer on upgraded provider clusters

leelavg:
If the cluster was configured in provider mode pre4.19 a storageconsumer would already exist for serving storage to local services. We use allowRemoteStorageConsumers for finding the older provider mode as this field is made immutable from 4.19.

### DIFF
--- a/controllers/storagecluster/storageconsumer.go
+++ b/controllers/storagecluster/storageconsumer.go
@@ -25,6 +25,20 @@ type storageConsumer struct{}
 var _ resourceManager = &storageConsumer{}
 
 func (s *storageConsumer) ensureCreated(r *StorageClusterReconciler, storageCluster *ocsv1.StorageCluster) (ctrl.Result, error) {
+	// TODO (leelavg): revisit in 4.20 whether this can be retired or not
+	// If the cluster was configured in provider mode pre4.19 a storageconsumer would already exist for serving storage to local services.
+	// We use allowRemoteStorageConsumers for finding the older provider mode as this field is made immutable from 4.19.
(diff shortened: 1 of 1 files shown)


controllers/storagecluster/storageconsumer.go
```diff
@@ -25,6 +25,13 @@ type storageConsumer struct{}
 var _ resourceManager = &storageConsumer{}
 
 func (s *storageConsumer) ensureCreated(r *StorageClusterReconciler, storageCluster *ocsv1.StorageCluster) (ctrl.Result, error) {
+	// TODO (leelavg): revisit in 4.20 whether this can be retired or not
+	// If the cluster was configured in provider mode pre4.19 a storageconsumer would already exist for serving storage to local services.
+	// We use allowRemoteStorageConsumers for finding the older provider mode as this field is made immutable from 4.19.
+	if storageCluster.Spec.AllowRemoteStorageConsumers {
```
nb-ohad, line 31: Don't we want to at least verify that a consumer exists in addition to the flag?

nb-ohad (changes requested):

controllers/storagecluster/storageconsumer.go
leelavg: What should the plan if one doesn't exist? Create internal storageconsumer?

nb-ohad: For now if it does not exists then return a reconcile error. There should not be any deployments of provider mode where there is no local consumer 

We might want to address creation of such consumer after doing multiple tests in the lab

PR #3142, opened 2025-04-08
controllers: implement upgrade controller for provider mode

rewantsoni:
This commit fixes small issues with the upgrade controller and removesthe  ownerRef of the StorageRequest from the SVG CR

### DIFF
--- a/controllers/storageconsumer/storageconsumer_upgrade_controller.go
+++ b/controllers/storageconsumer/storageconsumer_upgrade_controller.go
@@ -22,14 +22,18 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
+	"slices"
 	"strconv"
 	"strings"
 
+	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
 	ocsv1alpha1 "github.com/red-hat-storage/ocs-operator/api/v4/v1alpha1"
 	"github.com/red-hat-storage/ocs-operator/v4/controllers/util"
 
+	rookCephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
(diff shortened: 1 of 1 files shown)


controllers/storageconsumer/storageconsumer_upgrade_controller.go
```diff
@@ -143,12 +147,34 @@ func (r *StorageConsumerUpgradeReconciler) Reconcile(ctx context.Context, reques
...
+			return reconcile.Result{}, err
+		}
+		if svg.UID != "" {
+			refs := svg.GetOwnerReferences()
+			if idx := slices.IndexFunc(refs, func(owner metav1.OwnerReference) bool {
+				return owner.Kind == "StorageRequest"
+			}); idx != -1 {
+				svg.SetOwnerReferences(slices.Delete(refs, idx, idx+1))
+			}
+		}
```
nb-ohad, line 171: - Why is this code located here? it has nothing to do with filling in the CM 
- I don't see any code that applies the change. As the code stands now this is only an in memory change.

nb-ohad (changes requested):

controllers/storageconsumer/storageconsumer_upgrade_controller.go
```diff
@@ -106,35 +110,63 @@ func (r *StorageConsumerUpgradeReconciler) Reconcile(ctx context.Context, reques
...
+
+	return reconcile.Result{}, nil
+}
+
+func (r *StorageConsumerUpgradeReconciler) createConsumerConfigMap(
+	ctx context.Context,
+	storageCluster *ocsv1.StorageCluster,
+	storageConsumer *ocsv1alpha1.StorageConsumer,
+	configMapName string,
+) (util.StorageConsumerResources, error) {
```
nb-ohad, line 136: The wrapping scope is this function, this mean you should not be returning this type

controllers/storageconsumer/storageconsumer_upgrade_controller.go
```diff
@@ -162,12 +196,78 @@ func (r *StorageConsumerUpgradeReconciler) Reconcile(ctx context.Context, reques
...
+	// Removing the onwerRef from rns to preserve it after deletion the StorageRequest
+	rns := &metav1.PartialObjectMetadata{}
+	rns.SetGroupVersionKind(rookCephv1.SchemeGroupVersion.WithKind("CephBlockPoolRadosNamespace"))
+	rns.Name = rnsName
+	rns.Namespace = storageConsumer.Namespace
+	if err := r.Client.Get(ctx, client.ObjectKeyFromObject(rns), rns); client.IgnoreNotFound(err) != nil {
+		return err
+	}
+	if rns.UID != "" && removeStorageRequestAsOwner(rns) {
+		if err := r.Client.Update(ctx, rns); client.IgnoreNotFound(err) != nil {
```
leelavg, line 225: either use a full object or do a patch here as you need entire view for performing update.

rewantsoni: Updated to use patch

controllers/storageconsumer/storageconsumer_upgrade_controller.go
```diff
@@ -162,12 +195,42 @@ func (r *StorageConsumerUpgradeReconciler) Reconcile(ctx context.Context, reques
...
+		return err
+	}
+	if svg.UID != "" {
+		refs := svg.GetOwnerReferences()
+		if idx := slices.IndexFunc(refs, func(owner metav1.OwnerReference) bool {
+			return owner.Kind == "StorageRequest"
+		}); idx != -1 {
+			svg.SetOwnerReferences(slices.Delete(refs, idx, idx+1))
 		}
 	}
```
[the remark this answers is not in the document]
rewantsoni: Discussed offline we need to remove ownerRef on svg and rns

PR #3143, opened 2025-04-08
add support for distributing day2 storageclasses

leelavg:
when the storageclass mentioned in storageconsumer doesn't match one of the known names we consider that being unmanaged day2 created sc and we update the dynamic params corresponding to the driver and then send as a resource to be created & owned by storageclient in client-op.


rewantsoni:
We should also add the drStorageID label to the day2 storageClasses

nb-ohad:
This latest revision just makes the code complexity grow without a need. 
I would suggest focusing on treating the second day storage classes as additional templates (ones that might or might not exists)

With that strategy you should have a compile time function that given an SC name (template name) returns a generator for that name or an error.

leelavg:
> SC name
- user chooses this during runtime, we can have template by provisioner?

services/provider/server/server.go
```diff
@@ -2790,3 +2810,55 @@ func zeroFieldByName(obj any, fieldName string) {
...
+	clientProfileName := ""
+	provisionerSecretName := ""
+	nodeSecretName := ""
+	operatorNamespace := consumer.Status.Client.OperatorNamespace
+	switch storageClass.Provisioner {
+	case util.RbdDriverName:
+		clientProfileName = consumerConfig.GetRbdClientProfileName()
+		provisionerSecretName = consumerConfig.GetCsiRbdProvisionerSecretName()
+		nodeSecretName = consumerConfig.GetCsiRbdNodeSecretName()
+		util.AddLabel(storageClass, ramenDRStorageIDLabelKey, drRbdStorageId)
```
nb-ohad, line 2837:
```suggestion
```

services/provider/server/server.go
```diff
@@ -2790,3 +2810,55 @@ func zeroFieldByName(obj any, fieldName string) {
...
+	case util.CephFSDriverName:
+		clientProfileName = consumerConfig.GetCephFsClientProfileName()
+		provisionerSecretName = consumerConfig.GetCsiCephFsProvisionerSecretName()
+		nodeSecretName = consumerConfig.GetCsiCephFsNodeSecretName()
+		util.AddLabel(storageClass, ramenDRStorageIDLabelKey, drCephFsId)
```
nb-ohad, line 2842:
```suggestion
```

services/provider/server/server.go
```diff
@@ -2790,3 +2810,55 @@ func zeroFieldByName(obj any, fieldName string) {
...
+		params = map[string]string{}
+		storageClass.Parameters = params
+	}
+	params["clusterID"] = clientProfileName
+	params["csi.storage.k8s.io/provisioner-secret-name"] = provisionerSecretName
+	params["csi.storage.k8s.io/provisioner-secret-namespace"] = operatorNamespace
+	params["csi.storage.k8s.io/node-stage-secret-name"] = nodeSecretName
+	params["csi.storage.k8s.io/node-stage-secret-namespace"] = operatorNamespace
+	params["csi.storage.k8s.io/controller-expand-secret-name"] = provisionerSecretName
+	params["csi.storage.k8s.io/controller-expand-secret-namespace"] = operatorNamespace
```
nb-ohad, line 2862:
```suggestion
	params["csi.storage.k8s.io/controller-expand-secret-namespace"] = operatorNamespace
	util.AddLabel(storageClass, storageIdLabelKey, storageId)
```

services/provider/server/server.go
```diff
@@ -2790,3 +2810,55 @@ func zeroFieldByName(obj any, fieldName string) {
...
 	}
 }
+
+func storageClassFromExisting(
+	ctx context.Context,
+	kubeClient client.Client,
+	storageClassName string,
+	consumer *ocsv1alpha1.StorageConsumer,
+	consumerConfig util.StorageConsumerResources,
+	drRbdStorageId,
```
nb-ohad, line 2820:
```suggestion
	storageId,
```

services/provider/server/server.go
```diff
@@ -2790,3 +2810,55 @@ func zeroFieldByName(obj any, fieldName string) {
...
+	drCephFsId string,
```
nb-ohad, line 2821:
```suggestion
     cephFsId string,
```

services/provider/server/server.go
```diff
@@ -2432,15 +2436,31 @@ func (s *OCSProviderServer) appendStorageClassKubeResources(
...
-			klog.Warningf("encountered an unexpected storage class: %s", storageClassName)
+			var err error
+			storageClass, err = storageClassFromExisting(
+				ctx,
+				s.client,
+				storageClassName,
+				consumer,
+				consumerConfig,
+				drRbdStorageId,
+				drCephFsId,
```
nb-ohad, line 2454: I am not sure where these are defined. But they should not have the `dr` prefix in the name

nb-ohad (changes requested):

services/provider/server/server.go
leelavg: any reason on why this should be removed?

services/provider/server/server.go
leelavg: removed.

services/provider/server/server.go
leelavg: removed.

services/provider/server/server.go

services/provider/server/server.go

leelavg:
tested and rebased.

PR #3145, opened 2025-04-09
server: shim for GetStorageClaimConfig

rohan47:
Shim for GetStorageClaimConfig


nb-ohad:
@rohan47 The content of this PR re-invent the way we serialize storageclasses and other resources. The intent was to re-implement GetStorageClaim config base on data coming from our shared code that is already being utilized by `GetDesiredClinetState` and the re-implementation of  `GetStorageConfig`.

leelavg:
this PR should also return empty response for FulfillStorageClaim and a failure for RevokeStorageClaim RPCs.

services/provider/server/server.go
```diff
@@ -610,358 +608,172 @@ func decodeAndValidateTicket(ticket string, pubKey *rsa.PublicKey) (*services.On
...
-		errMsg := fmt.Sprintf("failed to fulfill storage class claim for %q. %v", req.StorageConsumerUUID, err)
-		klog.Error(errMsg)
-		if kerrors.IsAlreadyExists(err) {
-			return nil, status.Error(codes.AlreadyExists, errMsg)
-		}
-		return nil, status.Error(codes.Internal, errMsg)
-	}
-
-	return &pb.FulfillStorageClaimResponse{}, nil
+	return nil, nil
```
leelavg, line 613:
```suggestion
	return &pb.FulfillStorageClaimResponse{}, nil
```
I'm a bit unsure how nil is interpreted on the other side or over the wire.

services/provider/server/server.go
```diff
@@ -610,358 +608,172 @@ func decodeAndValidateTicket(ticket string, pubKey *rsa.PublicKey) (*services.On
...
-	}
-
-	return &pb.RevokeStorageClaimResponse{}, nil
-}
-
-func storageClaimCephCsiSecretName(secretType, suffix string) string {
-	return fmt.Sprintf("ceph-client-%s-%s", secretType, suffix)
+	errMsg := fmt.Sprintf("failed to revoke storage class claim %q for %q.", req.StorageClaimName, req.StorageConsumerUUID)
+	klog.Error(errMsg)
+	return nil, status.Error(codes.Internal, errMsg)
```
leelavg, line 621: you are not actually failing to revoke but not proceeding with revoke which basically is an unsupported operation in 4.19, pls make the status code & message reflect the same.

services/provider/server/server_test.go
```diff
@@ -656,629 +652,3 @@ func TestOCSProviderServerRevokeStorageClaim(t *testing.T) {
 	_, err = server.RevokeStorageClaim(ctx, req)
 	assert.NoError(t, err)
 }
-
-func TestOCSProviderServerGetStorageClaimConfig(t *testing.T) {
```
leelavg, line 660: is there a way to adjust this test rather than removing it? this test effectively verifies that shim is working, isn't it?

services/provider/server/server.go
```diff
@@ -610,358 +608,172 @@ func decodeAndValidateTicket(ticket string, pubKey *rsa.PublicKey) (*services.On
...
+						Data:        mustMarshal(storageClass.Parameters),
+					})
 			}
 
-			storageCluster, err := util.GetStorageClusterInNamespace(ctx, s.client, s.namespace)
-			if err != nil {
-				return nil, err
+		case csiopv1a1.GroupVersion.WithKind("ClientProfile"):
+			clientProfile := resource.(*csiopv1a1.ClientProfile)
+			var clientProfileName string
```
leelavg, line 758: this func was called more than a couple of times, either store the value in a var and use it multiple times or you directly compare to `ocs-storagecluster-ceph-rbd` or `ocs-storagecluster-cephfs` since those were the only supported claims.

services/provider/server/server.go
```diff
@@ -610,358 +608,172 @@ func decodeAndValidateTicket(ticket string, pubKey *rsa.PublicKey) (*services.On
...
-
-func storageClaimCephCsiSecretName(secretType, suffix string) string {
-	return fmt.Sprintf("ceph-client-%s-%s", secretType, suffix)
+	errMsg := fmt.Sprintf("failed to revoke storage class claim %q for %q.", req.StorageClaimName, req.StorageConsumerUUID)
+	klog.Error(errMsg)
+	return nil, status.Error(codes.Internal, errMsg)
 }
 
 // GetStorageClaim RPC call to get the ceph resources for the StorageClaim.
 func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.StorageClaimConfigRequest) (*pb.StorageClaimConfigResponse, error) {
```
leelavg, line 625: leaving this func as is, is it possible to create a proxy function and just call that accepting a consumer uid?

unlike the other 2 funcs where we just collapse the body into a single line, this func makes a lot of changes hard to reason/write/review. wdyt?

services/provider/server/server.go
```diff
@@ -610,358 +608,172 @@ func decodeAndValidateTicket(ticket string, pubKey *rsa.PublicKey) (*services.On
...
-			return nil, status.Errorf(codes.Unavailable, "status is not set for storage class claim %q for %q", req.StorageClaimName, req.StorageConsumerUUID)
-		default:
-			return nil, status.Error(codes.Internal, msg)
-		}
+		return nil, err
+	}
+	storageCluster, err := util.GetStorageClusterInNamespace(ctx, s.client, s.namespace)
+	if err != nil {
+		return nil, err
 	}
```
rewantsoni, line 633: Are we using the storageCluster anywhere?

services/provider/server/server.go
```diff
@@ -610,358 +608,172 @@ func decodeAndValidateTicket(ticket string, pubKey *rsa.PublicKey) (*services.On
...
-			subVolumeGroup := &rookCephv1.CephFilesystemSubVolumeGroup{}
-			err := s.client.Get(ctx, types.NamespacedName{Name: cephRes.Name, Namespace: s.namespace}, subVolumeGroup)
-			if err != nil {
-				return nil, status.Errorf(codes.Internal, "failed to get %s cephFilesystemSubVolumeGroup. %v", cephRes.Name, err)
+		case storagev1.SchemeGroupVersion.WithKind("StorageClass"):
+			storageClass := resource.(*storagev1.StorageClass)
+			if req.StorageClaimName == util.GenerateNameForCephBlockPoolStorageClass(storageCluster) {
+				storageClass.Parameters["radosnamespace"] = consumerConfig.GetRbdRadosNamespaceName()
+			} else if req.StorageClaimName == util.GenerateNameForCephFilesystemStorageClass(storageCluster) {
+				storageClass.Parameters["subvolumegroup"] = consumerConfig.GetSubVolumeGroupName()
```
rewantsoni, line 743: I don't think we need to add the radosnamespace and subvolumegroup field here, we are deleting them on the client side in 4.18 clusters. https://github.com/red-hat-storage/ocs-client-operator/blob/release-4.18/internal/controller/storageclaim_controller.go#L426-L427

rewantsoni (changes requested):

services/provider/server/server.go
rohan47: Yes, `req.StorageClaimName == util.GenerateNameForCephBlockPoolStorageClass(storageCluster)`

services/provider/server/server.go
rohan47: Its being deleted after the value is consumed right?

rewantsoni: I don't see the value consumed

services/provider/server/server_test.go
rohan47: Its a hassle to make the units succeed with small tweaks

services/provider/server/server.go
```diff
@@ -662,6 +629,58 @@ func storageClaimCephCsiSecretName(secretType, suffix string) string {
...
+	var extR []*pb.ExternalResource
+
+	if req.StorageConsumerUUID != "" {
+		if req.StorageClaimName == util.GenerateNameForCephBlockPoolStorageClass(storageCluster) {
+			extR, err = s.proxyGetStorageClaimConfigBlock(req, storageCluster, consumerConfig, kubeResources)
+			if err != nil {
+				klog.Errorf("failed to get storage claim config. %v", err)
+				return nil, status.Errorf(codes.Internal, "failed to get storage claim config. %v", err)
+			}
+		} else if req.StorageClaimName == util.GenerateNameForCephFilesystemStorageClass(storageCluster) {
```
nb-ohad, line 672: I am not sure you are comparing to the correct name. The `GenerateNameFor...` function takes rename input from the storage cluster while the old code didn't

services/provider/server/server.go
```diff
@@ -662,6 +629,58 @@ func storageClaimCephCsiSecretName(secretType, suffix string) string {
...
+			if err != nil {
+				klog.Errorf("failed to get storage claim config. %v", err)
+				return nil, status.Errorf(codes.Internal, "failed to get storage claim config. %v", err)
+			}
+		}
+		return &pb.StorageClaimConfigResponse{ExternalResource: extR}, nil
+	}
+
+	// TODO:
+	// The following code will be removed in the future
```
nb-ohad, line 683: If this PR introduces logic that replaces the code below, why are we still having it here?

services/provider/server/server.go
```diff
@@ -969,6 +986,223 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
 
 }
 
+// SHIM for GetStorageClaimConfig to return response to 4.18 clients
+func (s *OCSProviderServer) proxyGetStorageClaimConfigBlock(
```
nb-ohad, line 990: - What is exactly "proxied" here? I don't see a reason to have proxy in the name of this function
- This is not a function for any block storage claim. It is for exactly the storage claim that was the default storage claim in 4.18. Please rename the function accordingly 

```suggestion
func (s *OCSProviderServer) get4_18DefaultBlockStorageClaimConfig(
```

services/provider/server/server.go
```diff
@@ -969,6 +986,223 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
...
+	req *pb.StorageClaimConfigRequest,
+	storageCluster *ocsv1.StorageCluster,
```
nb-ohad, line 992: Why do we need the storage cluster in this function?

services/provider/server/server.go
```diff
@@ -969,6 +986,223 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
...
+	consumerConfig util.StorageConsumerResources,
```
nb-ohad, line 993: Don't send the wrapper between function, sand the raw map/config map and wrap it only for local use

services/provider/server/server.go
```diff
@@ -969,6 +986,223 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
...
+		gvk, err := apiutil.GVKForObject(resource, s.scheme)
+		if err != nil {
+			return nil, err
+		}
+
+		switch gvk {
+		case v1.SchemeGroupVersion.WithKind("Secret"):
+			secret := resource.(*corev1.Secret)
+			provSecName := consumerConfig.GetCsiRbdProvisionerSecretName()
+			nodeSecName := consumerConfig.GetCsiRbdNodeSecretName()
```
nb-ohad, line 1009: We already sent the entire kube resource list to this function, why do we need to look at the consume config again, or even send it to this function in the first place?


This comment is relevant for all of the other cases in this switch as well

services/provider/server/server.go
```diff
@@ -969,6 +986,223 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
...
+			if secret.Name == provSecName || secret.Name == nodeSecName {
+				oldSecretFormat := map[string]string{}
+				for k, v := range secret.Data {
+					oldSecretFormat[k] = string(v)
+				}
```
nb-ohad, line 1014: `secret.Data` is already a `map[string]string`. What is the reason to copy it into an new `map[string]string`?

services/provider/server/server.go
```diff
@@ -969,6 +986,223 @@ func (s *OCSProviderServer) GetStorageClaimConfig(ctx context.Context, req *pb.S
...
+	return extR, nil
+
+}
+
+// SHIM for GetStorageClaimConfig to return response to 4.18 clients
+func (s *OCSProviderServer) proxyGetStorageClaimConfigCephFs(
+	req *pb.StorageClaimConfigRequest,
+	storageCluster *ocsv1.StorageCluster,
+	consumerConfig util.StorageConsumerResources,
+	kubeResources []client.Object,
```
nb-ohad, line 1109: All of the comments raised for the `proxyGetStorageClaimConfigBlock` should also apply to this function and the implementation in its body

services/provider/server/server.go
```diff
@@ -577,7 +577,292 @@ func (s *OCSProviderServer) RevokeStorageClaim(ctx context.Context, req *pb.Revo
...
+	case ocsv1alpha1.StorageConsumerStateReady:
+		consumerConfigMap := &v1.ConfigMap{}
+		if storageConsumer.Status.ResourceNameMappingConfigMap.Name == "" {
+			return nil, fmt.Errorf("waiting for ResourceNameMappingConfig to be generated")
+		}
+		consumerConfigMap.Name = storageConsumer.Status.ResourceNameMappingConfigMap.Name
+		consumerConfigMap.Namespace = storageConsumer.Namespace
+		err = s.client.Get(ctx, client.ObjectKeyFromObject(consumerConfigMap), consumerConfigMap)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get %s configMap. %v", consumerConfigMap.Name, err)
```
rewantsoni, line 609: Let's use inline errors
```suggestion
		if err = s.client.Get(ctx, client.ObjectKeyFromObject(consumerConfigMap), consumerConfigMap); err != nil {
			return nil, fmt.Errorf("failed to get %s configMap. %v", consumerConfigMap.Name, err)
```

services/provider/server/server.go
```diff
@@ -577,7 +577,292 @@ func (s *OCSProviderServer) RevokeStorageClaim(ctx context.Context, req *pb.Revo
...
+			return nil, err
+		}
+
+		if req.StorageClaimName == fmt.Sprintf("%s-ceph-rbd", storageCluster.Name) {
+			extR, err = s.getStorageClaimConfigBlock(req, kubeResources,
+				storageCluster.Name,
+				consumerConfig.GetCsiRbdProvisionerCephUserName(),
+				consumerConfig.GetCsiRbdNodeCephUserName(),
+				consumerConfig.GetRbdClientProfileName(),
+			)
```
rewantsoni, line 629:
```suggestion
			extR, err = s.getStorageClaimConfigBlock(
			        req, 
			        kubeResources,
				storageCluster.Name,
				consumerConfig.GetCsiRbdProvisionerCephUserName(),
				consumerConfig.GetCsiRbdNodeCephUserName(),
				consumerConfig.GetRbdClientProfileName(),
			)
```

services/provider/server/server.go
```diff
@@ -577,7 +577,292 @@ func (s *OCSProviderServer) RevokeStorageClaim(ctx context.Context, req *pb.Revo
...
+				klog.Errorf("failed to get storage claim config. %v", err)
+				return nil, status.Errorf(codes.Internal, "failed to get storage claim config. %v", err)
+			}
+		} else if req.StorageClaimName == fmt.Sprintf("%s-cephfs", storageCluster.Name) {
+			extR, err = s.getStorageClaimConfigCephFs(req, kubeResources,
+				storageCluster.Name,
+				consumerConfig.GetCsiCephFsProvisionerCephUserName(),
+				consumerConfig.GetCsiRbdNodeCephUserName(),
+				consumerConfig.GetCephFsClientProfileName(),
+			)
```
rewantsoni, line 640:
```suggestion
			extR, err = s.getStorageClaimConfigCephFs(
			        req, 
			        kubeResources,
				storageCluster.Name,
				consumerConfig.GetCsiCephFsProvisionerCephUserName(),
				consumerConfig.GetCsiRbdNodeCephUserName(),
				consumerConfig.GetCephFsClientProfileName(),
			)
```

rewantsoni: Make sure to use GetCsiCephFsNodeCephUserName instead of GetCsiRbdNodeCephUserName

services/provider/server/server.go
```diff
@@ -577,7 +577,292 @@ func (s *OCSProviderServer) RevokeStorageClaim(ctx context.Context, req *pb.Revo
...
+		if err != nil {
+			return nil, err
+		}
+
+		switch gvk {
+		case v1.SchemeGroupVersion.WithKind("Secret"):
+			secret := resource.(*corev1.Secret)
+			provSecName := provSecretName
+			nodeSecName := nodeSecretName
+			if secret.Name == provSecName || secret.Name == nodeSecName {
```
rewantsoni, line 676: Why do we need new variables here?

services/provider/server/server.go
```diff
@@ -577,7 +577,292 @@ func (s *OCSProviderServer) RevokeStorageClaim(ctx context.Context, req *pb.Revo
...
+						Kind:        "VolumeReplicationClass",
+						Labels:      vrc.Labels,
+						Annotations: vrc.Annotations,
+						Data:        mustMarshal(vrc.Spec),
+					})
+			}
+
+		case snapapi.SchemeGroupVersion.WithKind("VolumeSnapshotClass"):
+			vsc := resource.(*snapapi.VolumeSnapshotClass)
+			vscName := fmt.Sprintf("%s-%splugin-snapclass", storageClusterName, util.RbdSnapshotter)
```
rewantsoni, line 703: We can use util.GenerateSnapshotClassName here

services/provider/server/server.go
```diff
@@ -577,7 +577,292 @@ func (s *OCSProviderServer) RevokeStorageClaim(ctx context.Context, req *pb.Revo
...
+						Labels:      vsc.Labels,
+						Annotations: vsc.Annotations,
+						Data:        mustMarshal(vsc.Parameters),
+					})
+			}
+
+		case groupsnapapi.SchemeGroupVersion.WithKind("VolumeGroupSnapshotClass"):
+			vgsc := resource.(*groupsnapapi.VolumeGroupSnapshotClass)
+			var vgscName string
+			vgscName = fmt.Sprintf("%s-ceph-%s-groupsnapclass", storageClusterName, util.RbdGroupSnapshotter)
```
rewantsoni, line 718: We can use util.GenerateGroupSnapshotClassName here

services/provider/server/server.go
```diff
@@ -577,7 +577,292 @@ func (s *OCSProviderServer) RevokeStorageClaim(ctx context.Context, req *pb.Revo
...
+		if err != nil {
+			return nil, err
+		}
+
+		switch gvk {
+		case v1.SchemeGroupVersion.WithKind("Secret"):
+			secret := resource.(*corev1.Secret)
+			provSecName := provSecretName
+			nodeSecName := nodeSecretName
+			if secret.Name == provSecName || secret.Name == nodeSecName {
```
rewantsoni, line 789: Why do we need new variables here?

services/provider/server/server.go
```diff
@@ -577,7 +577,292 @@ func (s *OCSProviderServer) RevokeStorageClaim(ctx context.Context, req *pb.Revo
...
+				extR = append(extR, &pb.ExternalResource{
+					Name:        secret.Name,
+					Kind:        gvk.Kind,
+					Labels:      resource.GetLabels(),
+					Annotations: resource.GetAnnotations(),
+					Data:        mustMarshal(secret.Data),
+				})
```
rewantsoni, line 683: Old secret format is missing here

services/provider/server/server.go
```diff
@@ -577,7 +577,292 @@ func (s *OCSProviderServer) RevokeStorageClaim(ctx context.Context, req *pb.Revo
...
+						Kind:        "Secret",
+						Annotations: secret.Annotations,
+						Labels:      secret.Labels,
+						Data:        mustMarshal(oldSecretFormat),
+					})
+			}
+
+		case snapapi.SchemeGroupVersion.WithKind("VolumeSnapshotClass"):
+			vsc := resource.(*snapapi.VolumeSnapshotClass)
+			vscName := fmt.Sprintf("%s-%splugin-snapclass", storageClusterName, util.CephfsSnapshotter)
```
rewantsoni, line 806: Let's use the util that generates the snapshot name

services/provider/server/server.go
```diff
@@ -577,7 +577,292 @@ func (s *OCSProviderServer) RevokeStorageClaim(ctx context.Context, req *pb.Revo
...
+						Kind:        "VolumeSnapshotClass",
+						Labels:      vsc.Labels,
+						Annotations: vsc.Annotations,
+						Data:        mustMarshal(vsc.Parameters),
+					})
+			}
+
+		case groupsnapapi.SchemeGroupVersion.WithKind("VolumeGroupSnapshotClass"):
+			vgsc := resource.(*groupsnapapi.VolumeGroupSnapshotClass)
+			vgscName := fmt.Sprintf("%s-ceph-%s-groupsnapclass", storageClusterName, util.RbdGroupSnapshotter)
```
rewantsoni, line 820: Let's use the util that generates the GroupSnapshot name

rewantsoni (changes requested):

services/provider/server/server.go
rohan47: Why is the old secret format required here?
https://github.com/red-hat-storage/ocs-operator/pull/3145#discussion_r2049446914

services/provider/server/server.go
rewantsoni: `secret.data` is a `map[string][]byte` `configmap.data` is `map[string]string`

rohan47:
> @rohan47 @rewantsoni Is this PR still relevant?

Discussed with @rewantsoni,
if the client cluster ocp is still at 4.18. It won't have the new catalog source but we still need to support this client and also when install plan for client op is using manual approval strategy

leelavg (commented):
Rest lg.

PR #3147, opened 2025-04-09
server: write shim for GetStorageConfig 

rewantsoni:
Write a shim layer for older clients on 4.18 which will be using the old rpc calls

### DIFF
--- a/services/provider/server/server.go
+++ b/services/provider/server/server.go
@@ -214,31 +214,101 @@ func (s *OCSProviderServer) GetStorageConfig(ctx context.Context, req *pb.Storag
 	case ocsv1alpha1.StorageConsumerStateDeleting:
 		return nil, status.Errorf(codes.NotFound, "storageConsumer is already in deleting phase")
 	case ocsv1alpha1.StorageConsumerStateReady:
-		conString, err := s.getExternalResources(ctx, consumerObj)
+		kubeResources, err := s.getKubeResources(ctx, consumerObj)
 		if err != nil {
-			return nil, status.Errorf(codes.Internal, "failed to get external resources. %v", err)
+			klog.Errorf("failed to get kube resources: %v", err)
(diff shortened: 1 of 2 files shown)


services/provider/server/server.go
```diff
@@ -213,32 +213,103 @@ func (s *OCSProviderServer) GetStorageConfig(ctx context.Context, req *pb.Storag
...
-			return nil, status.Errorf(codes.Internal, "Failed to construct status response: %v", err)
-		}
+		response := &pb.StorageConfigResponse{}
+
+		for _, kubeResource := range kubeResources {
+			gvk, err := apiutil.GVKForObject(kubeResource, s.scheme)
+			if err != nil {
+				return nil, err
+			}
+			switch gvk.Kind {
```
nb-ohad, line 228: Why are we only comparing to the Kind and not the entire gvk?

services/provider/server/server.go
```diff
@@ -213,32 +213,103 @@ func (s *OCSProviderServer) GetStorageConfig(ctx context.Context, req *pb.Storag
...
+		storageCluster, err := util.GetStorageClusterInNamespace(ctx, s.client, s.namespace)
+		if err != nil {
+			return nil, err
+		}
+
+		channelName, err := s.getOCSSubscriptionChannel(ctx)
+		if err != nil {
+			klog.Errorf("failed to get channel name for Client Operator: %v", err)
+			return nil, status.Errorf(codes.Internal, "failed to get channel name for Client Operator")
+		}
```
nb-ohad, line 312: Is there a reason to add these. the 4.18 client don't do anything with these

nb-ohad (changes requested):

services/provider/server/server.go
rewantsoni: This is required for the hash we send from GetStorageInfo

rewantsoni:
fixed panic in GetStorageConfig

PR #3149, opened 2025-04-09
osd: add a new ceph config to prevent slow ops

parth-gr:
a new ceph config bluestore_slow_ops_warn_lifetime:0 needed to be set to avoid slow ops warning
coming from latest 19.2.1.x release

On 4.19 build we didn't see the issue after deployment of ODF, but after triggering the regression test suit we saw the issue.


iamniting (approved):

iamniting:
pls remove the hold once it is tested.

Arpanchak:
I test it in 4.19 new set-up to set this flag bluestore_slow_ops_warn_lifetime = 0 and i also run regression test suit ..the **ceph health is okay** and i not face any issue ......
**Build use - > 4.19.0-75**

PR #3157, opened 2025-04-14
controller: handle internal client in provider mode upgrade 

rewantsoni:
The internal Client/consumer in provider mode needs to be handled differently.
We need to onboard the client to the internal consumer. Hence, from the upgrade controller, we are creating a consumer/configMap, updating the client status to point to this new consumer, and in the end, deleting the old storage consumer.

Fixes: https://issues.redhat.com/browse/DFBUGS-2290

### DIFF
--- a/controllers/storagecluster/storageconsumer.go
+++ b/controllers/storagecluster/storageconsumer.go
@@ -2,6 +2,7 @@ package storagecluster
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"maps"
 	"slices"
@@ -20,14 +21,14 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
+	"k8s.io/utils/ptr"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
 const (
-	localStorageConsumerConfigMapName = "storageconsumer-internal"
-	subVolumeGroupName                = "csi"
+	subVolumeGroupName = "csi"
 )
 
 var (
@@ -44,6 +45,12 @@ var _ resourceManager = &storageConsumer{}
 
 func (s *storageConsumer) ensureCreated(r *StorageClusterReconciler, storageCluster *ocsv1.StorageCluster) (ctrl.Result, error) {
 
+	//TODO: This code is to be removed in 4.20 when AllowRemoteStorageConsumers is removed. This is for the Storage Cluster
+	// reconciler to wait so that it doesn't create StorageConsumer with wrong names
+	if storageCluster.Spec.AllowRemoteStorageConsumers && len(storageCluster.GetAnnotations()[util.BackwardCompatabilityInfoAnnotationKey]) <= 0 {
+		return ctrl.Result{}, fmt.Errorf("invalid Configuration, allowRemoteStorageConsumers is set while missing a compatibility mode annotation on the storage cluster")
+	}
+
 	storageClassesSpec, err := getLocalStorageClassNames(r.ctx, r.Client, storageCluster)
 	if err != nil {
 		return ctrl.Result{}, fmt.Errorf("failed to generate storageclasses list for distribution: %v", err)
@@ -57,6 +64,13 @@ func (s *storageConsumer) ensureCreated(r *StorageClusterReconciler, storageClus
 		return ctrl.Result{}, fmt.Errorf("failed to generate volumegroupsnapshotclasses list for distribution: %v", err)
 	}
 
+	backwardCompatibleInfo := &util.BackwardCompatabilityInfo{}
+	if backwardCompatibleValue, ok := storageCluster.GetAnnotations()[util.BackwardCompatabilityInfoAnnotationKey]; ok {
+		if err := json.Unmarshal([]byte(backwardCompatibleValue), backwardCompatibleInfo); err != nil {
+			return ctrl.Result{}, fmt.Errorf("invalid backwardCompatibleInfo annotation, value is not a json: %v", err)
+		}
+	}
+
 	storageConsumer := &ocsv1a1.StorageConsumer{}
 	storageConsumer.Name = defaults.LocalStorageConsumerName
 	storageConsumer.Namespace = storageCluster.Namespace
@@ -65,7 +79,7 @@ func (s *storageConsumer) ensureCreated(r *StorageClusterReconciler, storageClus
 			return err
 		}
 		spec := &storageConsumer.Spec
-		spec.ResourceNameMappingConfigMap.Name = localStorageConsumerConfigMapName
+		spec.ResourceNameMappingConfigMap.Name = defaults.LocalStorageConsumerConfigMapName
 		spec.StorageClasses = storageClassesSpec
 		spec.VolumeSnapshotClasses = volumeSnapshotClassesSpec
 		spec.VolumeGroupSnapshotClasses = volumeGroupSnapshotClassesSpec
@@ -98,29 +112,45 @@ func (s *storageConsumer) ensureCreated(r *StorageClusterReconciler, storageClus
 		return ctrl.Result{}, fmt.Errorf("failed to get available services configured in StorageCluster: %v", err)
 	}
 	consumerConfigMap := &corev1.ConfigMap{}
-	consumerConfigMap.Name = localStorageConsumerConfigMapName
+	consumerConfigMap.Name = defaults.LocalStorageConsumerConfigMapName
 	consumerConfigMap.Namespace = storageCluster.Namespace
 	if _, err := controllerutil.CreateOrUpdate(r.ctx, r.Client, consumerConfigMap, func() error {
+		owners := consumerConfigMap.OwnerReferences
+		controllerIndex := slices.IndexFunc(
+			owners,
+			func(ref metav1.OwnerReference) bool {
+				return ref.UID != storageCluster.UID && ptr.Deref(ref.Controller, false)
+			},
+		)
+		if controllerIndex != -1 {
+			owners[controllerIndex].Controller = nil
+		}
+
 		if err := controllerutil.SetControllerReference(storageCluster, consumerConfigMap, r.Scheme); err != nil {
 			return err
 		}
+
 		data := util.GetStorageConsumerDefaultResourceNames(
 			defaults.LocalStorageConsumerName,
 			string(storageConsumer.UID),
 			availableServices,
 		)
 		resourceMap := util.WrapStorageConsumerResourceMap(data)
-		resourceMap.ReplaceRbdRadosNamespaceName(util.ImplicitRbdRadosNamespaceName)
-		resourceMap.ReplaceSubVolumeGroupName(subVolumeGroupName)
-		resourceMap.ReplaceSubVolumeGroupRadosNamespaceName(subVolumeGroupName)
-		resourceMap.ReplaceRbdClientProfileName("openshift-storage")
-		resourceMap.ReplaceCephFsClientProfileName("openshift-storage")
-		resourceMap.ReplaceNfsClientProfileName("openshift-storage")
-		// NB: Do we need to allow user changing/overwriting any values in this configmap?
+		if backwardCompatibleInfo.Pre4_19InternalConsumer == "" {
+			resourceMap.ReplaceRbdRadosNamespaceName(util.ImplicitRbdRadosNamespaceName)
+			resourceMap.ReplaceSubVolumeGroupName(subVolumeGroupName)
+			resourceMap.ReplaceSubVolumeGroupRadosNamespaceName(subVolumeGroupName)
+			resourceMap.ReplaceRbdClientProfileName("openshift-storage")
+			resourceMap.ReplaceCephFsClientProfileName("openshift-storage")
+			resourceMap.ReplaceNfsClientProfileName("openshift-storage")
+		} else {
+			util.FillBackwardCompatibleConsumerConfigValues(storageCluster, backwardCompatibleInfo.Pre4_19InternalConsumer, resourceMap)
+		}
+
 		consumerConfigMap.Data = data
 		return nil
 	}); err != nil {
-		return ctrl.Result{}, fmt.Errorf("failed to create/update storageconsumer configmap %s: %v", localStorageConsumerConfigMapName, err)
+		return ctrl.Result{}, fmt.Errorf("failed to create/update storageconsumer configmap %s: %v", consumerConfigMap.Name, err)
 	}
 
 	svgPre4_19 := &rookCephv1.CephFilesystemSubVolumeGroup{}
--- a/controllers/storageconsumer/storageconsumer_upgrade_controller.go
+++ b/controllers/storageconsumer/storageconsumer_upgrade_controller.go
@@ -20,22 +20,25 @@ import (
 	"context"
 	"crypto/md5"
 	"encoding/hex"
-	"encoding/json"
 	"fmt"
 	"slices"
 
 	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
 	ocsv1alpha1 "github.com/red-hat-storage/ocs-operator/api/v4/v1alpha1"
+	"github.com/red-hat-storage/ocs-operator/v4/controllers/defaults"
 	"github.com/red-hat-storage/ocs-operator/v4/controllers/util"
 
+	ocsclientv1a1 "github.com/red-hat-storage/ocs-client-operator/api/v1alpha1"
 	rookCephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
@@ -77,6 +80,8 @@ func (r *StorageConsumerUpgradeReconciler) SetupWithManager(mgr ctrl.Manager) er
 // +kubebuilder:rbac:groups=ocs.openshift.io,resources=storagerequests/finalizers,verbs=update
 // +kubebuilder:rbac:groups=ceph.rook.io,resources=cephfilesystemsubvolumegroups,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=ceph.rook.io,resources=cephblockpoolradosnamespaces,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=ocs.openshift.io,resources=storageclients,verbs=get
+// +kubebuilder:rbac:groups=ocs.openshift.io,resources=storageclients/status,verbs=patch
 
 func (r *StorageConsumerUpgradeReconciler) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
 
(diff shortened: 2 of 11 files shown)


controllers/storagecluster/storageconsumer.go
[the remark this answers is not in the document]
rewantsoni: Right, will do

controllers/storageconsumer/storageconsumer_upgrade_controller.go
```diff
@@ -106,35 +116,81 @@ func (r *StorageConsumerUpgradeReconciler) Reconcile(ctx context.Context, reques
 		return reconcile.Result{}, err
 	}
 
+	consumerConfigMapName := fmt.Sprintf("storageconsumer-%v", util.FnvHash(storageConsumer.Name))
+	consumerName := storageConsumer.Name
+
+	clusterID := util.GetClusterID(ctx, r.Client, &log)
+
+	internalConsumer := clusterID != "" && clusterID == storageConsumer.Status.Client.ClusterID
```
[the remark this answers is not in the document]
rewantsoni: I agree

controllers/storageconsumer/storageconsumer_upgrade_controller.go
```diff
@@ -266,24 +292,61 @@ func (r *StorageConsumerUpgradeReconciler) removeStorageRequestOwner(ctx context
 	return nil
 }
 
-func (r *StorageConsumerUpgradeReconciler) reconcileStorageConsumer(ctx context.Context, storageCluster *ocsv1.StorageCluster, storageConsumer *ocsv1alpha1.StorageConsumer, consumerConfigMapName string) error {
-	spec := &storageConsumer.Spec
-	spec.ResourceNameMappingConfigMap.Name = consumerConfigMapName
-	spec.StorageClasses = []ocsv1alpha1.StorageClassSpec{
-		{Name: util.GenerateNameForCephBlockPoolStorageClass(storageCluster)},
-		{Name: util.GenerateNameForCephFilesystemStorageClass(storageCluster)},
+func (r *StorageConsumerUpgradeReconciler) reconcileStorageConsumer(
```
leelavg, line 295: is there any reason duplicating it here, the minute provider internal consumer is marked as internal storagecluster controller will reconcile it and fills the necessary spec, moreover this is overwritten as well.

let me know if I'm misunderstanding anything.

rewantsoni: This will be called for both internal consumers and external consumers that were in provider mode

leelavg: ok, that's the missing piece, thanks.

leelavg:
reopening as #3274 only replaces
```
util.RestartPod(ctx, r.Client, &log, "ocs-provider-server", storageCluster.Namespace)
```
	}

controllers/storageconsumer/storageconsumer_upgrade_controller.go
```diff
@@ -121,8 +135,29 @@ func (r *StorageConsumerUpgradeReconciler) Reconcile(ctx context.Context, reques
...
+		newConsumer := &ocsv1alpha1.StorageConsumer{}
+		newConsumer.Name = defaults.LocalStorageConsumerName
+		newConsumer.Namespace = storageConsumer.Namespace
+		if err := r.Client.Get(ctx, client.ObjectKeyFromObject(newConsumer), newConsumer); err != nil {
+			return reconcile.Result{}, err
+		}
+		if err = r.reconcileStorageClient(ctx, storageCluster.Name, string(newConsumer.UID)); err != nil {
+			return reconcile.Result{}, err
+		}
+		newConsumer.Spec.Enable = true
```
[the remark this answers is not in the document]
rewantsoni: Yes, it assumes the configmap is created. We are creating/updating the configmap in the upgrade controller as well. So we don't need to wait here, right?

controllers/storagecluster/storageconsumer.go
```diff
@@ -113,11 +112,14 @@ func (s *storageConsumer) ensureCreated(r *StorageClusterReconciler, storageClus
 		resourceMap.ReplaceRbdClientProfileName("openshift-storage")
 		resourceMap.ReplaceCephFsClientProfileName("openshift-storage")
 		resourceMap.ReplaceNfsClientProfileName("openshift-storage")
-		// NB: Do we need to allow user changing/overwriting any values in this configmap?
+
+		if oldConsumerUID := storageConsumer.GetAnnotations()[util.ProviderModeUpgradedConsumerUID]; oldConsumerUID != "" {
```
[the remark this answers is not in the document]
rewantsoni: I thought about it, and If we are adding a new set of keys in the configmap in future, we would need to update both the if and else block for the keys. Hence chose to keep it this way

rewantsoni:
>seems like a good approach, we'll depend on internal storageconsumer from provider mode having an annotation?
once the upgrade controller is removed, no artifact will exist such an operation was done, better to add a comment in cluster controller that to never overwrite annotations during consumer reconcile? wdyt?

Yes we need the storageConsumer uid to figure out the names, hence we need to have this annotation. Yes, I can add a comment to never overwrite this annotation.

controllers/storagecluster/storageconsumer.go
```diff
@@ -98,26 +116,30 @@ func (s *storageConsumer) ensureCreated(r *StorageClusterReconciler, storageClus
 		return ctrl.Result{}, fmt.Errorf("failed to get available services configured in StorageCluster: %v", err)
 	}
 	consumerConfigMap := &corev1.ConfigMap{}
-	consumerConfigMap.Name = localStorageConsumerConfigMapName
+	consumerConfigMap.Name = storageConsumer.Spec.ResourceNameMappingConfigMap.Name
 	consumerConfigMap.Namespace = storageCluster.Namespace
 	if _, err := controllerutil.CreateOrUpdate(r.ctx, r.Client, consumerConfigMap, func() error {
 		data := util.GetStorageConsumerDefaultResourceNames(
-			defaults.LocalStorageConsumerName,
+			storageConsumer.Name,
```
nb-ohad, line 123: Drop the change, this should be the old value

controllers/storagecluster/storageconsumer.go
```diff
@@ -98,26 +116,30 @@ func (s *storageConsumer) ensureCreated(r *StorageClusterReconciler, storageClus
...
 		)
 		resourceMap := util.WrapStorageConsumerResourceMap(data)
-		resourceMap.ReplaceRbdRadosNamespaceName(util.ImplicitRbdRadosNamespaceName)
-		resourceMap.ReplaceSubVolumeGroupName(subVolumeGroupName)
-		resourceMap.ReplaceSubVolumeGroupRadosNamespaceName(subVolumeGroupName)
-		resourceMap.ReplaceRbdClientProfileName("openshift-storage")
-		resourceMap.ReplaceCephFsClientProfileName("openshift-storage")
-		resourceMap.ReplaceNfsClientProfileName("openshift-storage")
-		// NB: Do we need to allow user changing/overwriting any values in this configmap?
+		if consumerUID != "" {
```
leelavg, line 128:
```suggestion
		if consumerUidFromStorageCluster != "" {
```

since consumerUid is generic and uid will only be empty when object doesn't exist which is used in different scenario?

controllers/storagecluster/storageconsumer.go
```diff
@@ -136,8 +158,26 @@ func (s *storageConsumer) ensureCreated(r *StorageClusterReconciler, storageClus
...
+			client.InNamespace(storageCluster.Namespace),
+		); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to get the internal storageconsumer: %v", err)
+		}
+		if len(storageConsumers.Items) == 1 {
+			consumerName = storageConsumers.Items[0].Name
+		} else if len(storageConsumers.Items) > 1 {
+			return ctrl.Result{}, fmt.Errorf("more than one storage consumer is available for the storage client")
+		}
+	}
```
leelavg, line 177: why is this required?

controllers/storagecluster/storageconsumer.go
nb-ohad, line 128: The code should not know that this is a consumer UID, just a backward compatible magic value

controllers/storagecluster/storageconsumer.go
nb-ohad, line 177: No need, the consumer name will always be the same

controllers/storageconsumer/storageconsumer_upgrade_controller.go
```diff
@@ -113,6 +116,17 @@ func (r *StorageConsumerUpgradeReconciler) Reconcile(ctx context.Context, reques
 
 	consumerConfigMapName := fmt.Sprintf("storageconsumer-%v", util.FnvHash(storageConsumer.Name))
 
+	clusterID := util.GetClusterID(ctx, r.Client, &log)
+	if clusterID == "" {
+		return reconcile.Result{}, fmt.Errorf("failed to get openshift cluster ID")
+	}
+
+	internalConsumer := clusterID == storageConsumer.Status.Client.ClusterID && storageConsumer.Status.Client.Name == storageCluster.Name
```
nb-ohad, line 124: - This is a boolean, 
- Client could be nil
```suggestion
	IsInternalConsumer := clusterID == storageConsumer.Status.Client.ClusterID && storageConsumer.Status.Client.Name == storageCluster.Name
```

controllers/storageconsumer/storageconsumer_upgrade_controller.go
```diff
@@ -113,6 +116,17 @@ func (r *StorageConsumerUpgradeReconciler) Reconcile(ctx context.Context, reques
...
+
+	if internalConsumer {
+		consumerConfigMapName = defaults.LocalStorageConsumerConfigMapName
+	}
```
nb-ohad, line 128: Don't override an existing value, make a decision in an `if - else` manner when the name need to be selected

controllers/storageconsumer/storageconsumer_upgrade_controller.go
```diff
@@ -121,8 +135,30 @@ func (r *StorageConsumerUpgradeReconciler) Reconcile(ctx context.Context, reques
...
+			return reconcile.Result{}, fmt.Errorf("failed to add storage sluster compatibility mode annotation: %w", err)
+		}
+		// Delete the old consumer
+		if err = r.Client.Delete(ctx, storageConsumer); err != nil {
+			return reconcile.Result{}, err
+		}
+	} else {
+		if err = r.reconcileStorageConsumer(ctx, storageCluster, storageConsumer, consumerConfigMapName); err != nil {
+			return reconcile.Result{}, err
+		}
```
nb-ohad, line 161: We are already doing a create or update on the every consumer. why do we need a special path here? If you select the correct name there should be no 'other path' for creation of the new consumer

controllers/storageconsumer/storageconsumer_upgrade_controller.go
```diff
@@ -208,7 +209,7 @@ func (r *StorageConsumerUpgradeReconciler) reconcileStorageRequest(
 
 	rbdStorageRequest := &metav1.PartialObjectMetadata{}
 	rbdStorageRequest.SetGroupVersionKind(ocsv1alpha1.GroupVersion.WithKind("StorageRequest"))
-	rbdStorageRequest.Name = getStorageRequestName(
+	rbdStorageRequest.Name = util.GetStorageRequestName(
```
nb-ohad, line 212: Where do we need this outside of the current package?

controllers/storageconsumer/storageconsumer_upgrade_controller.go
```diff
@@ -305,30 +311,21 @@ func (r *StorageConsumerUpgradeReconciler) reconcileStorageConsumer(ctx context.
...
+		return err
 	}
 
-	requestName, err := json.Marshal(s)
-	if err != nil {
-		panic("failed to marshal storage class request name")
+	storageClient.Status.ConsumerID = storageConsumerUID
+	if err := r.Client.Status().Update(ctx, storageClient); err != nil {
+		return err
 	}
```
nb-ohad, line 329: The decision need to be made from the outside. 
If you are already inside a reconcile function you should reconcile

controllers/util/k8sutil.go
```diff
@@ -186,6 +187,10 @@ func OwnersIndexFieldFunc(obj client.Object) []string {
 	return owners
 }
 
+func UIDIndexFieldFunc(obj client.Object) []string {
+	return []string{string(obj.GetUID())}
+}
```
nb-ohad, line 192: There is no need for this please remove

controllers/util/storageconsumer.go
```diff
@@ -305,3 +308,77 @@ func GetStorageConsumerDefaultResourceNames(
 	}
 	return defaults
 }
+
+// These methods are for backward compatibility for provider mode upgraded cluster
+
+func FillBackwardCompatibleConsumerConfigValues(
```
nb-ohad, line 314: We need a better name here. Also why does this need to move outside of the controllers package?

controllers/storageconsumer/storageconsumer_upgrade_controller.go
```diff
@@ -121,8 +135,30 @@ func (r *StorageConsumerUpgradeReconciler) Reconcile(ctx context.Context, reques
 		return reconcile.Result{}, err
 	}
 
-	if err = r.reconcileStorageConsumer(ctx, storageCluster, storageConsumer, consumerConfigMapName); err != nil {
-		return reconcile.Result{}, err
+	if internalConsumer {
+		newConsumer := &ocsv1alpha1.StorageConsumer{}
+		newConsumer.Name = defaults.LocalStorageConsumerName
+		newConsumer.Namespace = storageConsumer.Namespace
+		newConsumer.Spec.Enable = true
```
leelavg, line 142: this'll create issues afaics, enable will trigger consumer controller since no CM is created the defaults values will be propagated and consumer will be made as controlling owner.

controllers/storageconsumer/storageconsumer_upgrade_controller.go
```diff
@@ -121,8 +135,30 @@ func (r *StorageConsumerUpgradeReconciler) Reconcile(ctx context.Context, reques
...
+		}
+		if err = r.reconcileStorageClient(ctx, storageCluster.Name, string(newConsumer.UID)); err != nil {
+			return reconcile.Result{}, err
+		}
+		util.AddAnnotation(storageCluster, util.ProviderModeCompatibleInternalConsumerUID, string(storageConsumer.UID))
+		if err := r.Client.Update(ctx, storageCluster); err != nil {
+			return reconcile.Result{}, fmt.Errorf("failed to add storage sluster compatibility mode annotation: %w", err)
+		}
+		// Delete the old consumer
+		if err = r.Client.Delete(ctx, storageConsumer); err != nil {
```
leelavg, line 155: due to a create at the start, if the step fails here this will result in an error loop.

controllers/storageconsumer/storageconsumer_upgrade_controller.go
rewantsoni: The consumer configmap is created above in the upgrade controller

controllers/storageconsumer/storageconsumer_upgrade_controller.go
rewantsoni: Since I am moving the filling of the method to util, I need to export this or create a copy of the code

controllers/storagecluster/storageconsumer.go
```diff
@@ -57,15 +56,34 @@ func (s *storageConsumer) ensureCreated(r *StorageClusterReconciler, storageClus
...
```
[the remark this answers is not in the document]
rewantsoni: removed this

controllers/storagecluster/storageconsumer.go
rewantsoni: Updated it to backwardCompatibleValue as per https://github.com/red-hat-storage/ocs-operator/pull/3157#discussion_r2166569267

rewantsoni:
/retest-required
failed due to infra issue

nb-ohad (changes requested):

controllers/util/storageconsumer.go
```diff
@@ -13,9 +15,10 @@ import (
 
 const (
 	// Reserved RadosNamespaceName for internal use and their representation at different layes
-	ImplicitRbdRadosNamespaceName          = "<implicit>"
-	TicketAnnotation                       = "ocs.openshift.io/provider-onboarding-ticket"
-	AnnotationNonResilientPoolsTopologyKey = "ocs.openshift.io/non-resilient-pools-topology-key"
+	ImplicitRbdRadosNamespaceName             = "<implicit>"
+	TicketAnnotation                          = "ocs.openshift.io/provider-onboarding-ticket"
+	AnnotationNonResilientPoolsTopologyKey    = "ocs.openshift.io/non-resilient-pools-topology-key"
+	ProviderModeCompatibleInternalConsumerUID = "ocs.openshift.io/provider-mode-compatible-consumer-uid"
```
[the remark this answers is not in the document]
rewantsoni: In ocs-op there is a package for each controller and the storageconsumer package has the name "controller". We are using it in the storageCluster package and for storageconsumer hence I needed to move it to a different package.

controllers/util/storageconsumer.go
rewantsoni: In ocs-op there is a package for each controller and the storageconsumer package has the name "controller". We are using it in the storageCluster package and for storageconsumer hence I needed to move it to a util.

controllers/storageconsumer/storageconsumer_upgrade_controller.go
```diff
@@ -147,52 +176,26 @@ func (r *StorageConsumerUpgradeReconciler) reconcileConsumerConfigMap(
...
-		resourceMap.ReplaceCsiCephFsNodeCephUserName(cephFsNodeSecretName)
-		resourceMap.ReplaceCsiCephFsProvisionerCephUserName(cephFsProvisionerSecretName)
+		util.FillBackwardCompatibleConsumerConfigValues(storageCluster, string(storageConsumer.UID), resourceMap)
 		consumerConfigMap.Data = data
-
+		if isInternalConsumer {
+			if err := controllerutil.SetControllerReference(storageCluster, consumerConfigMap, r.Scheme); err != nil {
+				return err
+			}
+		}
```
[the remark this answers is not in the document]
rewantsoni: The problem is that is we are enbable the storageconsumer here the consumer controller will get triggers and it will take ownership of the consumer configmap.

controllers/storageconsumer/storageconsumer_upgrade_controller.go
```diff
@@ -282,53 +285,63 @@ func (r *StorageConsumerUpgradeReconciler) removeStorageRequestOwner(ctx context
...
-	delete(storageConsumer.GetAnnotations(), util.TicketAnnotation)
-
-	if err := r.Client.Update(ctx, storageConsumer); client.IgnoreNotFound(err) != nil {
+	storageConsumer.Namespace = storageCluster.Namespace
+
+	if _, err := controllerutil.CreateOrUpdate(ctx, r.Client, storageConsumer, func() error {
+		spec := &storageConsumer.Spec
+		if isInternalConsumer {
+			spec.ResourceNameMappingConfigMap.Name = defaults.LocalStorageConsumerConfigMapName
+			spec.Enable = true
```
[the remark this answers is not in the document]
rewantsoni: Since we are changing the consumer ID on the status of the storage client we should be the one enabling the consumer as well.
If storagecosnumer enables the consumer, it would need different logic to fill the consumer spec.
The consumer is enabled during onboarding. Are you suggesting that we remove the ID from the storage client status, which would make the internal storage client onboard to the internal storage consumer?

controllers/storageconsumer/storageconsumer_upgrade_controller.go
```diff
@@ -282,53 +283,92 @@ func (r *StorageConsumerUpgradeReconciler) removeStorageRequestOwner(ctx context
...
+	storageClient := &ocsclientv1a1.StorageClient{}
+	storageClient.Name = storageClientName
+	if err := r.Client.Get(ctx, client.ObjectKeyFromObject(storageClient), storageClient); err != nil {
+		return err
+	}
 
-func storageClaimCephCsiSecretName(secretType, suffix string) string {
-	return fmt.Sprintf("ceph-client-%s-%s", secretType, suffix)
+	if storageClient.Status.ConsumerID != string(localStorageConsumer.UID) {
+		onboardingTicketPatch := client.RawPatch(types.JSONPatchType, util.JsonMustMarshal(
```
rewantsoni, line 351: The onboardingTicket will be populated from the storagecluster/storageclient.go. If we add the onboardingToken here, it will lead to the consumer being enabled before storagecluster controller can put an controller ownerRef on the consumer configmap. Which would mean that consumer will be the controller owner of the configmap which would lead to errors

controllers/storageconsumer/storageconsumer_upgrade_controller.go
```diff
@@ -282,53 +290,75 @@ func (r *StorageConsumerUpgradeReconciler) removeStorageRequestOwner(ctx context
...
-	md5Sum := md5.Sum(requestName)
-	return hex.EncodeToString(md5Sum[:16])
+	return nil
 }
 
-// getStorageRequestName generates a name for a StorageRequest resource.
-func getStorageRequestName(consumerUUID, storageClaimName string) string {
-	return fmt.Sprintf("storagerequest-%s", getStorageRequestHash(consumerUUID, storageClaimName))
-}
+func (r *StorageConsumerUpgradeReconciler) reconcileStorageClient(
```
nb-ohad, line 339:
```suggestion
func (r *StorageConsumerUpgradeReconciler) reconcileLocalStorageClient(
```

controllers/storageconsumer/storageconsumer_upgrade_controller.go
```diff
@@ -282,53 +290,75 @@ func (r *StorageConsumerUpgradeReconciler) removeStorageRequestOwner(ctx context
...
+	ctx context.Context,
+	storageClientName string,
+	namespace string,
+) error {
+	localStorageConsumer := &ocsv1alpha1.StorageConsumer{}
+	localStorageConsumer.Name = defaults.LocalStorageConsumerName
+	localStorageConsumer.Namespace = namespace
+	if err := r.Client.Get(ctx, client.ObjectKeyFromObject(localStorageConsumer), localStorageConsumer); err != nil {
+		return err
+	}
```
nb-ohad, line 349: Is there a reason to load it again here? We have the desired state already at the call site (the UID), why not send it into this function?

controllers/util/k8sutil.go
```diff
@@ -75,8 +75,13 @@ const (
 	StorageConsumerMirroringInfoAnnotation = "ocs.openshift.io/consumer-mirroring-info"
 	ForceDeletionAnnotationKey             = "ocs.openshift.io/force-deletion"
 	RookForceDeletionAnnotationKey         = "rook.io/force-deletion"
+	BackwardCompatabilityInfoAnnotationKey = "ocs.openshift.io/backward-compatability-info"
 )
 
+type BackwardCompatabilityInfo struct {
+	ConsumerUID string `json:"uid"`
```
nb-ohad, line 82:
```suggestion
	InternalConsumerUID string `json:"uid"`
```

controllers/util/storageconsumer.go
```diff
@@ -305,3 +307,77 @@ func GetStorageConsumerDefaultResourceNames(
...
+	if err != nil {
+		panic("failed to marshal storage class request name")
+	}
+	md5Sum := md5.Sum(requestName)
+	return hex.EncodeToString(md5Sum[:16])
+}
+
+func storageClaimCephCsiSecretName(secretType, suffix string) string {
+	return fmt.Sprintf("ceph-client-%s-%s", secretType, suffix)
+}
```
nb-ohad, line 383: These does not seems like they belong here

rewantsoni: These functions are used for configmap value generation. The reason I chose this place was that this file contains the interface for consumer configmap. I can move it to k8sutils or a new file. What do you suggest

controllers/util/k8sutil.go
```diff
@@ -75,8 +75,13 @@ const (
 	StorageConsumerMirroringInfoAnnotation = "ocs.openshift.io/consumer-mirroring-info"
 	ForceDeletionAnnotationKey             = "ocs.openshift.io/force-deletion"
 	RookForceDeletionAnnotationKey         = "rook.io/force-deletion"
+	BackwardCompatabilityInfoAnnotationKey = "ocs.openshift.io/backward-compatability-info"
 )
 
+type BackwardCompatabilityInfo struct {
+	InternalConsumerUID string `json:"uid"`
```
nb-ohad, line 82: 1. I don't think the name is descriptive enough. we need something that state that this is a pre 4.19 internal consumer UID

2. the json name should reflect that as well

```suggestion
	Pre4_19InternalConsumer string `json:"pre4_19InternalConsumer"`
```

nb-ohad (changes requested):

PR #3174, opened 2025-04-21
cluster: update exporter and crashcollector resources

parth-gr:
update the exporter and crashcollector with limits and priority class, so it wont get evicted
during memeory pressure testing

Used the default limits as of `mon`

### DIFF
--- a/controllers/defaults/resources.go
+++ b/controllers/defaults/resources.go
@@ -70,6 +70,26 @@ var (
 				"cpu":    resource.MustParse("50m"),
 			},
 		},
+		"crashcollector": {
+			Requests: corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("0.1"),
+				corev1.ResourceMemory: resource.MustParse("60Mi"),
+			},
+			Limits: corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("0.1"),
+				corev1.ResourceMemory: resource.MustParse("60Mi"),
+			},
+		},
+		"exporter": {
+			Requests: corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("0.05"),
+				corev1.ResourceMemory: resource.MustParse("50Mi"),
+			},
+			Limits: corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("0.05"),
+				corev1.ResourceMemory: resource.MustParse("128Mi"),
+			},
+		},
 	}
 
 	LeanDaemonResources = map[string]corev1.ResourceRequirements{
--- a/controllers/storagecluster/cephcluster.go
+++ b/controllers/storagecluster/cephcluster.go
@@ -535,9 +535,11 @@ func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, kmsConfigMap *co
 				"arbiter": getPlacement(sc, "arbiter"),
 			},
 			PriorityClassNames: rookCephv1.PriorityClassNamesSpec{
-				rookCephv1.KeyMgr: systemNodeCritical,
-				rookCephv1.KeyMon: systemNodeCritical,
-				rookCephv1.KeyOSD: systemNodeCritical,
+				rookCephv1.KeyMgr:            systemNodeCritical,
+				rookCephv1.KeyMon:            systemNodeCritical,
+				rookCephv1.KeyOSD:            systemNodeCritical,
+				rookCephv1.KeyCephExporter:   systemNodeCritical,
+				rookCephv1.KeyCrashCollector: systemNodeCritical,
 			},
 			Resources:    newCephDaemonResources(sc),
 			LogCollector: logCollector,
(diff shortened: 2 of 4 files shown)


parth-gr:
```
    priorityClassNames:
      crashcollector: system-node-critical
      exporter: system-node-critical
      mgr: system-node-critical
      mon: system-node-critical
      osd: system-node-critical

```

controllers/defaults/resources.go
```diff
@@ -93,6 +93,26 @@ var (
...
+			Requests: corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("0.5"),
+				corev1.ResourceMemory: resource.MustParse("1Gi"),
+			},
+			Limits: corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("0.5"),
+				corev1.ResourceMemory: resource.MustParse("1Gi"),
+			},
+		},
+		"exporter": {
```
travisn, line 106: See [this example](https://github.com/rook/rook/blob/master/deploy/charts/rook-ceph-cluster/values.yaml#L344-L349) for exporter resources

controllers/defaults/resources.go
```diff
@@ -93,6 +93,26 @@ var (
 				corev1.ResourceMemory: resource.MustParse("1Gi"),
 			},
 		},
+		"crashcollector": {
```
travisn, line 96: See [this example](https://github.com/rook/rook/blob/master/deploy/charts/rook-ceph-cluster/values.yaml#L326-L331) for crash collector resources.

controllers/defaults/resources.go
```diff
@@ -146,6 +166,26 @@ var (
 				corev1.ResourceMemory: resource.MustParse("2Gi"),
 			},
 		},
+		"crashcollector": {
```
travisn, line 169: Let's keep the resources the same across all the profiles, no need to change them since it's such a simple background daemon that doesn't impact performance of the cluster.

travisn (changes requested):

controllers/defaults/resources.go
```diff
@@ -199,6 +239,26 @@ var (
 				corev1.ResourceMemory: resource.MustParse("3Gi"),
 			},
 		},
+		"crashcollector": {
```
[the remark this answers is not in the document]
parth-gr: if we do that way this will break the pattern from where it is called https://github.com/red-hat-storage/ocs-operator/blob/22fe7f7962ee1f7261bd91ed7a095fd88aed7367/controllers/defaults/utils.go#L20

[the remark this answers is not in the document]
parth-gr: Done

travisn (commented):
LGTM, if you could just confirm when you could test the latest changes

malayparida2000:
@parth-gr For the eviction to not happen the priority class alone would be enough, Why are we specifying the limits/requests for them as well? that should not be necessary right?

parth-gr:
> @parth-gr For the eviction to not happen the priority class alone would be enough, Why are we specifying the limits/requests for them as well? that should not be necessary right?

Pods are selected for eviction based on the pod priority as well as the resource usage and requests, look on https://issues.redhat.com/browse/DFBUGS-1917# for more info

parth-gr:
@travisn Testing looks good

```
    priorityClassNames:
      crashcollector: system-node-critical
      exporter: system-node-critical
      mgr: system-node-critical
      mon: system-node-critical
      osd: system-node-critical
    resources:
      crashcollector:
        limits:
          cpu: 100m
          memory: 60Mi
        requests:
          cpu: 100m
          memory: 60Mi
      exporter:
        limits:
          cpu: 50m
          memory: 128Mi
        requests:
          cpu: 50m
          memory: 50Mi
      mgr:
        limits:
          cpu: "2"
          memory: 3Gi
        requests:
          cpu: "1"
          memory: 1536Mi
      mon:
        limits:
          cpu: "1"
          memory: 2Gi
        requests:
          cpu: "1"
          memory: 2Gi

```

travisn (approved):
/lgtm
Before considering backporting, let's get verification from QE that the crash collector and exporter are not ever crashing due to memory limits.

parth-gr:
> /lgtm Before considering backporting, let's get verification from QE that the crash collector and exporter are not ever crashing due to memory limits.

the build is using 4.19 branch, not sure how they will test it, do you want to test by custom image??
I myself has tested the pods are running fine, do we want to test with memory stress testing??

travisn:
> > /lgtm Before considering backporting, let's get verification from QE that the crash collector and exporter are not ever crashing due to memory limits.
> 
> the build is using 4.19 branch, not sure how they will test it, do you want to test by custom image?? I myself has tested the pods are running fine, do we want to test with memory stress testing??

QE can validate on the 4.19 builds, I don't think we need a custom build. 

>  I myself has tested the pods are running fine, do we want to test with memory stress testing??

We don't need stress testing, just to confirm that under normal cluster operation the daemons are running as expected. These two components wouldn't be affected by stress testing anyway.

parth-gr:
>QE can validate on the 4.19 builds, I don't think we need a custom build.

So we need a backport to be merge if we need the code in 4.19 build

PR #3177, opened 2025-04-22
controllers: make client information under consumer status a pointer 

rewantsoni:
1. Update the client information under storage consumer status to be a pointer.
2. Add checks to see if the client is nil before checking client Information

fixes: https://issues.redhat.com/browse/DFBUGS-2269

### DIFF
--- a/controllers/mirroring/mirroring_controller.go
+++ b/controllers/mirroring/mirroring_controller.go
@@ -79,7 +79,7 @@ func (r *MirroringReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		&ocsv1alpha1.StorageConsumer{},
 		clientIDIndexName,
 		func(obj client.Object) []string {
-			if storageConsumer, ok := obj.(*ocsv1alpha1.StorageConsumer); ok {
+			if storageConsumer, ok := obj.(*ocsv1alpha1.StorageConsumer); ok && storageConsumer.Status.Client != nil {
 				return []string{storageConsumer.Status.Client.ID}
 			}
 			return nil
@@ -531,7 +531,7 @@ func (r *MirroringReconciler) reconcileRadosNamespaceMirroring(
 			}
 			consumerOwner := &rns.OwnerReferences[consumerIndex]
 			consumer := storageConsumerByName[consumerOwner.Name]
-			if consumer == nil || consumer.Status.Client.ID == "" {
+			if consumer == nil || consumer.Status.Client == nil {
 				continue
 			}
 			remoteClientID := clientMappingConfig.Data[consumer.Status.Client.ID]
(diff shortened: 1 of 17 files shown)


DanielOsypenko (approved):

controllers/storagecluster/reconcile.go
```diff
@@ -953,15 +953,17 @@ func getUnsupportedClientsCount(r *StorageClusterReconciler, namespace string) (
 	var count int
 	providerVersion, _ := semver.Make(version.Version)
 	for idx := range scList.Items {
-		clientVersion, err := semver.Make(scList.Items[idx].Status.Client.OperatorVersion)
-		if err == nil {
-			// provider operator and client operator should be on same version for full compatibility
-			if providerVersion.Major != clientVersion.Major || providerVersion.Minor != clientVersion.Minor {
+		if scList.Items[idx].Spec.Enable {
```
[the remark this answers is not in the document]
rewantsoni: I have updated the PR to check the version only if it is not empty

controllers/mirroring/mirroring_controller.go
```diff
@@ -531,7 +531,7 @@ func (r *MirroringReconciler) reconcileRadosNamespaceMirroring(
 			}
 			consumerOwner := &rns.OwnerReferences[consumerIndex]
 			consumer := storageConsumerByName[consumerOwner.Name]
-			if consumer == nil || consumer.Status.Client.ID == "" {
+			if consumer == nil || consumer.Status.Client == nil || consumer.Status.Client.ID == "" {
```
nb-ohad, line 534: What would be the case where we have the client struct but we don't have an id?

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -276,15 +276,15 @@ func (r *StorageConsumerReconciler) reconcileEnabledPhases() (reconcile.Result,
...
 		}
 
-		if availableServices.Mcg {
+		if availableServices.Mcg && r.storageConsumer.Status.Client != nil {
 			// A provider cluster already has a NooBaa system and does not require a NooBaa account
 			// to connect to a remote cluster, unlike client clusters.
 			// A NooBaa account only needs to be created if the storage consumer is for a client cluster.
 			clusterID := util.GetClusterID(r.ctx, r.Client, &r.Log)
-			clientStatus := &r.storageConsumer.Status.Client
+			clientClusterID := r.storageConsumer.Status.Client.ClusterID
```
nb-ohad, line 284: This makes assumptions that we have client on the status. that might not be the case.

nb-ohad (changes requested):

controllers/mirroring/mirroring_controller.go
rewantsoni: I agree that it is not possible, removing it

PR #3179, opened 2025-04-23
storageautoscaler: refactor the resizing verification logic

parth-gr:
1) Dont look for status to decide the scaling

2) Query prometheus to get current osd size

3) Query storagecluster to get the desired osd size

PR #3186, opened 2025-04-24
storageautoscaler: add a watcher and predicate for storagecluster

parth-gr:
storagecluster provide the desired state,
if there is user change in desired state while scaling, we need to watch that change and make that as new desired state

with the changes, auto scaler will watch for osd changes whether its the total osd count or osd size.

Currently we will fan out all the auto scaler with in the same namespace of storageclass and reconcile them.

PS: there is no filter on reconciling the specific device class autoscaler for which the storageclass device set is updated

### DIFF
--- a/controllers/storageautoscaler/storageautoscaler_controller.go
+++ b/controllers/storageautoscaler/storageautoscaler_controller.go
@@ -38,13 +38,76 @@ type StorageAutoscalerReconciler struct {
 
 // SetupWithManager sets up the reconciler with the manager
 func (r *StorageAutoscalerReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	enqueueAutoScaler := handler.EnqueueRequestsFromMapFunc(
+		func(context context.Context, obj client.Object) []reconcile.Request {
+			storageCluster, ok := obj.(*ocsv1.StorageCluster)
+			if !ok {
+				return nil
+			}
+
+			// list the storage autoscaler objects
+			storageAutoScalerList := &ocsv1.StorageAutoScalerList{}
+			err := r.List(context, storageAutoScalerList,
+				client.InNamespace(storageCluster.Namespace),
+			)
+			if err != nil {
+				r.Log.Error(err, "failed to list storage autoscaler")
+				return nil
+			}
+			// create a reconcile request for each storage autoscaler object
+			requests := make([]reconcile.Request, 0, len(storageAutoScalerList.Items))
+			for _, storageAutoScaler := range storageAutoScalerList.Items {
+				requests = append(requests, reconcile.Request{
+					NamespacedName: types.NamespacedName{
+						Namespace: storageAutoScaler.Namespace,
+						Name:      storageAutoScaler.Name,
+					},
+				})
+			}
+			return requests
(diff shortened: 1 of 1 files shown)


controllers/storageautoscaler/storageautoscaler_controller.go
```diff
@@ -38,13 +38,76 @@ type StorageAutoscalerReconciler struct {
...
+				return false
+			}
+
+			// check if the osd size or count or replica has changed
+			for i, deviceSet := range oldStorageCluster.Spec.StorageDeviceSets {
+				if deviceSet.Count != newStorageCluster.Spec.StorageDeviceSets[i].Count {
+					return true
+				}
+				oldOsdSize := deviceSet.DataPVCTemplate.Spec.Resources.Requests["storage"]
+				newOsdSize := newStorageCluster.Spec.StorageDeviceSets[i].DataPVCTemplate.Spec.Resources.Requests["storage"]
```
BlaineEXE, line 88: One of the inherent assumptions of this code is that the list of sets in the old storage cluster is the same in the new storage cluster.

If the new cluster has a deviceset removed or a new deviceset inserted in the middle, this logic may get unexpected results.

travisn (commented):
Questions:
1. The PR title and description mentions `storageclass`, but the PR seems to be watching `storageCluster`. Was that a typo?
2. This seems like a rabbit hole where it will be very difficult to try and support the user making changes to the storageCluster while we are in the middle of a resize. What happens if we close this PR and do nothing? IMO it's ok if the autoscaler CR gets into a failure state if they cause it with some unexpected change.

parth-gr:
>     1. The PR title and description mentions `storageclass`, but the PR seems to be watching `storageCluster`. Was that a typo?

thanks that was a typo

>     2. This seems like a rabbit hole where it will be very difficult to try and support the user making changes to the storageCluster while we are in the middle of a resize. 

if the customer is making any changes to the storagecluster cr(size or count), it will still work, we can get the new desired state which is bigger then the previous one.

> What happens if we close this PR and do nothing? IMO it's ok if the autoscaler CR gets into a failure state if they cause it with some unexpected change.

The autoscaler has no impact either ways,  its important to take in consideration user specified desired state while scaling.

controllers/storageautoscaler/storageautoscaler_controller.go
parth-gr: okay thanks,  thats a good point.

travisn:
> > ```
> > 2. This seems like a rabbit hole where it will be very difficult to try and support the user making changes to the storageCluster while we are in the middle of a resize. 
> > ```
> 
> if the customer is making any changes to the storagecluster cr(size or count), it will still work, we can get the new desired state which is bigger then the previous one.
> 
> > What happens if we close this PR and do nothing? IMO it's ok if the autoscaler CR gets into a failure state if they cause it with some unexpected change.
> 
> The autoscaler has no impact either ways, its important to take in consideration user specified desired state while scaling.

Ok, so I would think we don't need a watcher and we can close this PR? I wonder if #3179 will already help check a more reliable status to handle a changed desired size.

PR #3192, opened 2025-04-28
Offboarding a client removes client info from Storage Consumer status instead of deleting storageconsumer

rohan47:
Current behaviour of offboarding a client deletes the Storage Consumer but with this PR offboarding will remove the client information from the storageConsumer status instead of deleting the storageConsumer.
Also removed the RBAC for deleting storage consumer


controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -73,7 +73,7 @@ type StorageConsumerReconciler struct {
 	namespace       string
 }
 
-//+kubebuilder:rbac:groups=ocs.openshift.io,resources=storageconsumers,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=ocs.openshift.io,resources=storageconsumers,verbs=get;list;watch;create;update;patch
```
rewantsoni, line 76: We shouldn't remove the delete permission from here. It should be from ocs-provider-server rbac. Look for rbac/provider-role

services/provider/server/server.go
```diff
@@ -418,10 +418,17 @@ func (s *OCSProviderServer) GetDesiredClientState(ctx context.Context, req *pb.G
 
 // OffboardConsumer RPC call to delete the StorageConsumer CR
 func (s *OCSProviderServer) OffboardConsumer(ctx context.Context, req *pb.OffboardConsumerRequest) (*pb.OffboardConsumerResponse, error) {
-	err := s.consumerManager.Delete(ctx, req.StorageConsumerUUID)
+	storageConsumer, err := s.consumerManager.Get(ctx, req.StorageConsumerUUID)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to get storageConsumer resource with the provided UUID. %v", err)
+	}
+	storageConsumer.Status.Client = &ocsv1alpha1.ClientStatus{}
```
rewantsoni, line 425:
```suggestion
	storageConsumer.Status.Client = nil
```

services/provider/server/server.go
```diff
@@ -418,10 +418,17 @@ func (s *OCSProviderServer) GetDesiredClientState(ctx context.Context, req *pb.G
...
+	err = s.consumerManager.client.Update(ctx, storageConsumer)
```
rewantsoni, line 426:
```suggestion
	err = s.client.Status.Update(ctx, storageConsumer)
```
and make the error inline

services/provider/server/server.go
```diff
@@ -418,10 +418,17 @@ func (s *OCSProviderServer) GetDesiredClientState(ctx context.Context, req *pb.G
...
 	if err != nil {
-		return nil, status.Errorf(codes.Internal, "failed to delete storageConsumer resource with the provided UUID. %v", err)
+		return nil, status.Errorf(codes.Internal, "failed to update storageConsumer resource status with the provided UUID. %v", err)
```
rewantsoni, line 428: log the err and return a message
```suggestion
                klog.Errorf(err)
		return nil, status.Errorf(codes.Internal, "failed to offboard from storgeconsumer")
```

services/provider/server/server.go
```diff
@@ -418,10 +418,17 @@ func (s *OCSProviderServer) GetDesiredClientState(ctx context.Context, req *pb.G
...
 	}
+
+	klog.Infof("Removed Client status from StorageConsumer with the provided UUID %q", req.StorageConsumerUUID)
```
rewantsoni, line 431:
```suggestion
	klog.Infof("Successfully Offboarded Client from StorageConsumer %v",consumerName)
```

rewantsoni (changes requested):

rohan47:
Addressed the requested changes and tested the PR

rewantsoni:
This PR looks good, @rohan47 have you made the client-side changes not to send the client status when the client is marked for deletion?

services/provider/server/server.go
```diff
@@ -418,10 +418,18 @@ func (s *OCSProviderServer) GetDesiredClientState(ctx context.Context, req *pb.G
...
 	if err != nil {
-		return nil, status.Errorf(codes.Internal, "failed to delete storageConsumer resource with the provided UUID. %v", err)
+		return nil, status.Errorf(codes.Internal, "failed to get storageConsumer resource with the provided UUID. %v", err)
 	}
+	storageConsumer.Status.Client = nil
+	err = s.client.Status().Update(ctx, storageConsumer)
+	if err != nil {
+		klog.Error(err)
+		return nil, status.Errorf(codes.Internal, "failed to offboard from storgeconsumer")
+	}
```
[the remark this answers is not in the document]
rohan47: Updated the client side PR to return error instead of silently fail

services/provider/server/server.go
```diff
@@ -406,10 +406,18 @@ func (s *OCSProviderServer) GetDesiredClientState(ctx context.Context, req *pb.G
...
 func (s *OCSProviderServer) OffboardConsumer(ctx context.Context, req *pb.OffboardConsumerRequest) (*pb.OffboardConsumerResponse, error) {
-	err := s.consumerManager.Delete(ctx, req.StorageConsumerUUID)
+	storageConsumer, err := s.consumerManager.Get(ctx, req.StorageConsumerUUID)
 	if err != nil {
-		return nil, status.Errorf(codes.Internal, "failed to delete storageConsumer resource with the provided UUID. %v", err)
+		return nil, status.Errorf(codes.Internal, "failed to get storageConsumer resource with the provided UUID. %v", err)
 	}
+	storageConsumer.Status.Client = nil
+	err = s.client.Status().Update(ctx, storageConsumer)
+	if err != nil {
```
rewantsoni, line 415:
```suggestion
	if err = s.client.Status().Update(ctx, storageConsumer); err != nil {
```

rewantsoni (changes requested):

services/provider/server/consumer.go
```diff
@@ -227,3 +227,16 @@ func (c *ocsConsumerManager) GetByClientID(ctx context.Context, clientID string)
...
+func (c *ocsConsumerManager) ClearClientInformation(ctx context.Context, consumerID string) error {
+	storageConsumer, err := c.Get(ctx, consumerID)
+	if err != nil {
+		return fmt.Errorf("failed to get storageConsumer resource with the provided UUID. %v", err)
+	}
+	storageConsumer.Status.Client = nil
+	err = c.client.Status().Update(ctx, storageConsumer)
+	if err != nil {
+		return fmt.Errorf("failed to remove client information from storageConsumer status %v: %v", storageConsumer.Name, err)
+	}
```
nb-ohad, line 240: The server does not use a cached client, this mean that the get goes directly into the API server. Can we use a patch instead to remove the need for the get operation?

nb-ohad (changes requested):

PR #3193, opened 2025-04-29
server: add monips as a part of desired state hash

rewantsoni:
1. Add monips to the desired state hash for clients
2. Update the hash to include only the consumer spec since status changes every min

Fixes: https://issues.redhat.com/browse/DFBUGS-2359

### DIFF
--- a/services/provider/server/server.go
+++ b/services/provider/server/server.go
@@ -387,9 +387,16 @@ func (s *OCSProviderServer) GetDesiredClientState(ctx context.Context, req *pb.G
 			return nil, err
 		}
 
+		cephConnection, err := s.getDesiredCephConnection(ctx, consumer)
+		if err != nil {
+			klog.Error(err)
+			return nil, status.Errorf(codes.Internal, "failed to produce client state hash")
+		}
+
 		desiredClientConfigHash := getDesiredClientConfigHash(
 			channelName,
 			consumer,
+			cephConnection.Spec,
 			isEncryptionInTransitEnabled(storageCluster.Spec.Network),
 			inMaintenanceMode,
 			isConsumerMirrorEnabled,
@@ -651,9 +658,16 @@ func (s *OCSProviderServer) ReportStatus(ctx context.Context, req *pb.ReportStat
 		return nil, status.Errorf(codes.Internal, "Failed to get mirroring status for consumer.")
 	}
 
+	cephConnection, err := s.getDesiredCephConnection(ctx, storageConsumer)
+	if err != nil {
+		klog.Error(err)
+		return nil, status.Errorf(codes.Internal, "failed to produce client state hash")
+	}
+
 	desiredClientConfigHash := getDesiredClientConfigHash(
 		channelName,
 		storageConsumer,
+		cephConnection.Spec,
 		isEncryptionInTransitEnabled(storageCluster.Spec.Network),
 		inMaintenanceMode,
 		isConsumerMirrorEnabled,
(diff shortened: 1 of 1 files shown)


services/provider/server/server.go
```diff
@@ -1018,31 +1025,40 @@ func (s *OCSProviderServer) ReportStatus(ctx context.Context, req *pb.ReportStat
...
 	if err != nil {
 		klog.Error(err)
-		return nil, status.Errorf(codes.Internal, "Failed to get mirroring status for consumer.")
+		return nil, status.Errorf(codes.Internal, "failed to produce client state hash")
+	}
+
+	monIps, err := getMonIps(ctx, s.client, storageConsumer.Namespace)
+	if err != nil {
+		klog.Errorf("failed to extract monitor IPs from configmap %s: %v", monConfigMap, err)
+		return nil, status.Errorf(codes.Internal, "failed to produce client state")
```
nb-ohad, line 1055: Aren't the mon IPs already retrieved in order to create the CephConnection Spec? if so why do we need to call getMonIps again here? and also what about any other change to the desired CephConnection spec?

nb-ohad (changes requested):

services/provider/server/server.go
rewantsoni: The hash is constructed in two places: ReportStatus and GetDesiredClientState. If I take a hash of the cephConnection spec, I would need to reconstruct the object in ReportStatus
I could get the monIPs before the kubeResource call and pass them to the function as well as to the hash. This way I won't have to call the getMonIps twice

nb-ohad: I understand, but this does not answer the question: " and also what about any other change to the desired CephConnection spec?"

rewantsoni: I was looking at the cephConnection CR spec for the fields that we have. Currently, there are two fields, `ReadAffinity` and `RbdMirrorDaemonCount`. They would need to be added separately to the hash as and when required

nb-ohad: @rewantsoni That does not make a lot sense, just use the same code to build the ceph connection CR itself then use  a that data to build the hash.

services/provider/server/server.go
```diff
@@ -387,9 +387,16 @@ func (s *OCSProviderServer) GetDesiredClientState(ctx context.Context, req *pb.G
...
+		cephConnection, err := s.getDesiredCephConnection(ctx, consumer)
+		if err != nil {
+			klog.Error(err)
+			return nil, status.Errorf(codes.Internal, "failed to produce client state hash")
+		}
+
 		desiredClientConfigHash := getDesiredClientConfigHash(
 			channelName,
 			consumer,
+			cephConnection,
```
nb-ohad, line 399:
```suggestion
			cephConnection.Spec,
```

services/provider/server/server.go
```diff
@@ -1052,10 +1066,11 @@ func (s *OCSProviderServer) getKubeResources(ctx context.Context, consumer *ocsv
 	}
 
 	kubeResources := []client.Object{}
-	kubeResources, err = s.appendCephConnectionKubeResources(ctx, kubeResources, consumer)
+	cephConnection, err := s.getDesiredCephConnection(ctx, consumer)
 	if err != nil {
 		return nil, err
 	}
+	kubeResources = append(kubeResources, cephConnection)
```
nb-ohad, line 1073: More structured and limits the scope of the ceph connection object
```suggestion
	if cephConnection, err := s.getDesiredCephConnection(ctx, consumer); err == nil {
		kubeResources = append(kubeResources, cephConnection)
	} else {
		return nil, err
	}

```

PR #3194, opened 2025-04-29
Always apply rbd-mirror debug logging ceph configs

malayparida2000:
Irrespective of whether rbd-mirror is deployed or not, the debug logging configs are always applied. This maakes the code simpler & clearr. As without rbd-mirror present, the debug logging configs are anyways not used so it is not a problem.

Ref-https://issues.redhat.com/browse/DFBUGS-2087

### DIFF
--- a/controllers/storagecluster/cephconfig.go
+++ b/controllers/storagecluster/cephconfig.go
@@ -42,6 +42,16 @@ bluestore_prefer_deferred_size_hdd = 0
 bluestore_slow_ops_warn_lifetime = 0
 [osd]
 osd_memory_target_cgroup_limit_ratio = 0.8
+[client.rbd-mirror.a]
+debug_ms = 1
+debug_rbd = 15
+debug_rbd_mirror = 30
+log_file = /var/log/ceph/\$cluster-\$name.log
+[client.rbd-mirror-peer]
+debug_ms = 1
+debug_rbd = 15
+debug_rbd_mirror = 30
+log_file = /var/log/ceph/\$cluster-\$name.log
 `
 )
 
--- a/controllers/storagecluster/cephrbdmirrors.go
+++ b/controllers/storagecluster/cephrbdmirrors.go
@@ -3,30 +3,21 @@ package storagecluster
 import (
 	"context"
 	"fmt"
-	"os"
-	"time"
 
 	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
 	"github.com/red-hat-storage/ocs-operator/v4/controllers/defaults"
 	"github.com/red-hat-storage/ocs-operator/v4/controllers/util"
 
 	cephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
-	batchv1 "k8s.io/api/batch/v1"
-	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/klog/v2"
-	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
 type ocsCephRbdMirrors struct{}
-
-var (
-	rbdMirrorDebugLoggingEnabled = false
-)
 
 func (r *StorageClusterReconciler) fetchCephRbdMirrorInstance(cephRbdMirror *cephv1.CephRBDMirror) (cephv1.CephRBDMirror, error) {
 	existing := cephv1.CephRBDMirror{}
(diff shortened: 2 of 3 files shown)


controllers/mirroring/mirroring_controller.go
```diff
@@ -286,11 +292,78 @@ func (r *MirroringReconciler) reconcileRbdMirror(clientMappingConfig *corev1.Con
...
+			err = r.Client.Create(r.ctx, rbdMirrorDebugLoggingJob)
+			if err != nil && !k8serrors.IsAlreadyExists(err) {
+				return true
+			}
+			// Check if the job has succeeded
+			job := &batchv1.Job{}
+			err = r.Client.Get(r.ctx, types.NamespacedName{Name: util.EnableRbdMirrorDebugLoggingJobName, Namespace: clientMappingConfig.Namespace}, job)
+			if err != nil {
+				return true
+			}
```
rewantsoni, line 313:
```suggestion
			if err = r.Client.Get(r.ctx, client.ObjectKeyFromObject(rbdMirrorDebugLoggingJob), job); err != nil {
				return true
			}
```

controllers/mirroring/mirroring_controller.go
```diff
@@ -286,11 +292,78 @@ func (r *MirroringReconciler) reconcileRbdMirror(clientMappingConfig *corev1.Con
...
+			rbdMirrorDebugLoggingJob, err := util.GetRbdMirrorDebugLoggingJob(util.EnableRbdMirrorDebugLoggingJobName, clientMappingConfig.Namespace, util.EnableRbdMirrorDebugLoggingCommands, clientMappingConfig)
+			if err != nil {
+				r.log.Error(err, "Failed to create the enable rbd mirror debug logging job")
+				return true
+			}
+
+			err = r.Client.Create(r.ctx, rbdMirrorDebugLoggingJob)
+			if err != nil && !k8serrors.IsAlreadyExists(err) {
+				return true
+			}
```
rewantsoni, line 307:
```suggestion
			if err = r.Client.Create(r.ctx, rbdMirrorDebugLoggingJob); client.IgnoreNotFound(err) != nil {
				return true
			}
```

controllers/mirroring/mirroring_controller.go
```diff
@@ -286,11 +292,78 @@ func (r *MirroringReconciler) reconcileRbdMirror(clientMappingConfig *corev1.Con
...
+			rbdMirrorDebugLoggingJob, err := util.GetRbdMirrorDebugLoggingJob(util.DisableRbdMirrorDebugLoggingJobName, clientMappingConfig.Namespace, util.DisableRbdMirrorDebugLoggingCommands, clientMappingConfig)
+			if err != nil {
+				r.log.Error(err, "failed to get disable rbd mirror debug logging job")
+				return true
+			}
+
+			err = r.Client.Create(r.ctx, rbdMirrorDebugLoggingJob)
+			if err != nil && !k8serrors.IsAlreadyExists(err) {
+				return true
+			}
```
rewantsoni, line 338:
```suggestion
			if err = r.Client.Create(r.ctx, rbdMirrorDebugLoggingJob); client.IgnoreNotFound(err) != nil {
				return true
			}
```

controllers/mirroring/mirroring_controller.go
```diff
@@ -286,11 +292,78 @@ func (r *MirroringReconciler) reconcileRbdMirror(clientMappingConfig *corev1.Con
...
+			// Check if the job has succeeded
+			job := &batchv1.Job{}
+			err = r.Client.Get(r.ctx, types.NamespacedName{Name: util.DisableRbdMirrorDebugLoggingJobName, Namespace: clientMappingConfig.Namespace}, job)
+			if err != nil {
+				return true
+			}
```
rewantsoni, line 344:
```suggestion
			if err = r.Client.Get(r.ctx, types.NamespacedName{Name: util.DisableRbdMirrorDebugLoggingJobName, Namespace: clientMappingConfig.Namespace}, job); err != nil {
				return true
			}
```

controllers/mirroring/mirroring_controller.go
```diff
@@ -286,11 +292,78 @@ func (r *MirroringReconciler) reconcileRbdMirror(clientMappingConfig *corev1.Con
...
+
+			// Delete both the enable & disable jobs
+			err = r.Client.Delete(r.ctx, &batchv1.Job{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      util.EnableRbdMirrorDebugLoggingJobName,
+					Namespace: clientMappingConfig.Namespace,
+				}})
+			if err != nil && !k8serrors.IsNotFound(err) {
+				return true
+			}
```
rewantsoni, line 354: Please use inline error

controllers/mirroring/mirroring_controller.go
```diff
@@ -286,11 +292,78 @@ func (r *MirroringReconciler) reconcileRbdMirror(clientMappingConfig *corev1.Con
...
+			err = r.Client.Delete(r.ctx, &batchv1.Job{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      util.DisableRbdMirrorDebugLoggingJobName,
+					Namespace: clientMappingConfig.Namespace,
+				}})
+			if err != nil && !k8serrors.IsNotFound(err) {
+				return true
+			}
```
rewantsoni, line 362: Please use inline error

rewantsoni (changes requested):
I have one question, since we know that we need this for around 6 releases because of volumeGroupReplication why not update the ceph Config directly to have this? this way we won't need to deploy the job to update the config. Since it's related to rbd mirror it won't do any harm if we are not enabling rbd mirror

controllers/mirroring/mirroring_controller.go
malayparida2000: I had raised the same question back in 4.12, At that point it was said to not pollute the ceph config with unrelated settings on non-rdr clusters. I don't know if that argument still holds true.

controllers/mirroring/mirroring_controller.go
malayparida2000: Addressed

controllers/mirroring/mirroring_controller.go
malayparida2000: Addressed

controllers/mirroring/mirroring_controller.go
malayparida2000: Addressed

controllers/mirroring/mirroring_controller.go
malayparida2000: Addressed

controllers/mirroring/mirroring_controller.go
malayparida2000: Addressed

travisn (commented):
Nice simplification!

travisn (commented):
/lgtm

PR #3195, opened 2025-04-29
provider-server: enable csi drivers for ocs-client-op

rohan47:
Send information of desired csi drivers to enable to ocs-client-operator


services/provider/proto/provider.proto
```diff
@@ -118,13 +118,30 @@ message StorageConfigResponse{
     SystemAttributes systemAttributes = 3;
 }
 
+message RbdDriverRequirements {
+  bool generateOmapInfo = 1;
```
nb-ohad, line 122: You are surfacing requirements so it does not make sense to have something named as a verb (generate)
```suggestion
  bool omapInfo = 1;
```

services/provider/proto/provider.proto
```diff
@@ -118,13 +118,30 @@ message StorageConfigResponse{
...
+}
+
+message CephFsDriverRequirements {
+}
+
+message NfsDriverRequirements {
+}
+
+message CsiDrivers {
+  RbdDriverRequirements DesiredRbdDriver = 1;
```
nb-ohad, line 132: Let make sure we are naming the field in accordance to our terminology. As we decided to call these "requirements" I would argue the name of the field should be `RbdDriverRequirments`

services/provider/proto/provider.proto
```diff
@@ -118,13 +118,30 @@ message StorageConfigResponse{
...
+}
+
 // GetDesiredClientStateResponse holds the response for the GetDesiredClientState API request
 message GetDesiredClientStateResponse{
   repeated bytes kubeResources = 1;
   string desiredStateHash = 2;
   string clientOperatorChannel = 3;
   bool maintenanceMode = 4;
   bool mirrorEnabled = 5;
+  CsiDrivers CsiDrivers = 6;
```
nb-ohad, line 144: Why do we need a substruct here? Can we just have the requirement sections on the root struct?

services/provider/server/server.go
```diff
@@ -1040,12 +1049,15 @@ func (s *OCSProviderServer) ReportStatus(ctx context.Context, req *pb.ReportStat
...
 
+	isNfsDriverEnabled := storageCluster.Spec.NFS != nil && storageCluster.Spec.NFS.Enable
+
 	desiredClientConfigHash := getDesiredClientConfigHash(
 		channelName,
 		storageConsumer,
 		isEncryptionInTransitEnabled(storageCluster.Spec.Network),
 		inMaintenanceMode,
 		isConsumerMirrorEnabled,
+		isNfsDriverEnabled,
```
nb-ohad, line 1060: Why only NFS? what about the other drivers?

services/provider/server/server.go
```diff
@@ -399,12 +399,21 @@ func (s *OCSProviderServer) GetDesiredClientState(ctx context.Context, req *pb.G
 			return nil, err
 		}
 
+		response.CsiDrivers.DesiredRbdDriver = &pb.RbdDriverRequirements{}
+		response.CsiDrivers.DesiredRbdDriver.GenerateOmapInfo = isConsumerMirrorEnabled
+		response.CsiDrivers.DesiredCephFsDriver = &pb.CephFsDriverRequirements{}
```
nb-ohad, line 404: Why are we assuming these are always enabled?

nb-ohad (changes requested):

services/provider/proto/provider.proto
```diff
@@ -118,13 +118,26 @@ message StorageConfigResponse{
     SystemAttributes systemAttributes = 3;
 }
 
+message RbdDriverRequirements {
+  bool omapInfo = 1;
```
rewantsoni, line 122: @nb-ohad, do we want to move this to operatorConfig Spec instead of Driver Spec and always deploy it?

services/provider/proto/provider.proto
```diff
@@ -129,6 +129,13 @@ message KubeObject {
 message RbdDriverRequirements {
     // topologyDomainLables are the list of topology that needs to be set in nodeplugin
     repeated string topologyDomainLables = 1;
+    bool omapInfo = 2;
```
rewantsoni, line 132: omap Info has been moved from driver spec to OperatorConfig in ocs-client-op

services/provider/server/server.go
```diff
@@ -404,6 +404,29 @@ func (s *OCSProviderServer) GetDesiredClientState(ctx context.Context, req *pb.G
...
+		consumerConfigMap.Namespace = consumer.Namespace
+		if err := s.client.Get(ctx, client.ObjectKeyFromObject(consumerConfigMap), consumerConfigMap); err != nil {
+			return nil, status.Errorf(codes.Internal, "failed to get resource name mapping config map. %v", err)
+		}
+		consumerConfig := util.WrapStorageConsumerResourceMap(consumerConfigMap.Data)
+
+		isRbdDriverEnabled := consumerConfig.GetRbdClientProfileName() != ""
+		if isRbdDriverEnabled {
+			response.RbdDriverRequirements = &pb.RbdDriverRequirements{}
+			response.RbdDriverRequirements.OmapInfo = isConsumerMirrorEnabled
```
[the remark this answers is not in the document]
rohan47: So, omap info is not required in rbdDriverRequirements?

rohan47:
Tested the changes end to end with client operator changes.
First this PR needs to go in and then the client operator PR needs to pull in ocs API changes

services/provider/server/server.go
```diff
@@ -404,6 +404,28 @@ func (s *OCSProviderServer) GetDesiredClientState(ctx context.Context, req *pb.G
 			klog.Error(err)
 			return nil, status.Errorf(codes.Internal, "failed to produce client state hash")
 		}
+		consumerConfigMap := &corev1.ConfigMap{}
+		consumerConfigMap.Name = consumer.Status.ResourceNameMappingConfigMap.Name
+		consumerConfigMap.Namespace = consumer.Namespace
+		if err := s.client.Get(ctx, client.ObjectKeyFromObject(consumerConfigMap), consumerConfigMap); err != nil {
+			return nil, status.Errorf(codes.Internal, "failed to get resource name mapping config map. %v", err)
```
rewantsoni, line 411:
```suggestion
			klog.Error(err)
			return nil, status.Errorf(codes.Internal, "failed to produce client state hash")
```

services/provider/server/server.go
```diff
@@ -404,6 +404,28 @@ func (s *OCSProviderServer) GetDesiredClientState(ctx context.Context, req *pb.G
 			klog.Error(err)
 			return nil, status.Errorf(codes.Internal, "failed to produce client state hash")
 		}
+		consumerConfigMap := &corev1.ConfigMap{}
+		consumerConfigMap.Name = consumer.Status.ResourceNameMappingConfigMap.Name
+		consumerConfigMap.Namespace = consumer.Namespace
```
rewantsoni, line 409: We are also loading the consumer configMap inside the getKubeResources, maybe we should load it once outside and pass it around

services/provider/server/server.go
```diff
@@ -681,6 +706,17 @@ func (s *OCSProviderServer) ReportStatus(ctx context.Context, req *pb.ReportStat
 	}
 
 	topologyKey := storageConsumer.GetAnnotations()[util.AnnotationNonResilientPoolsTopologyKey]
+	consumerConfigMap := &corev1.ConfigMap{}
+	consumerConfigMap.Name = storageConsumer.Status.ResourceNameMappingConfigMap.Name
+	consumerConfigMap.Namespace = storageConsumer.Namespace
+	if err := s.client.Get(ctx, client.ObjectKeyFromObject(consumerConfigMap), consumerConfigMap); err != nil {
+		return nil, status.Errorf(codes.Internal, "Failed to get storage consumer config map: %v", err)
```
rewantsoni, line 713:
```suggestion
                 klog.Error(err)
		return nil, status.Errorf(codes.Internal, "failed to produce client state hash")
```

rewantsoni (changes requested):

services/provider/server/server.go
rohan47: Moved consumerConfig generation to a helper function

services/provider/server/server.go
```diff
@@ -681,6 +713,14 @@ func (s *OCSProviderServer) ReportStatus(ctx context.Context, req *pb.ReportStat
 	}
 
 	topologyKey := storageConsumer.GetAnnotations()[util.AnnotationNonResilientPoolsTopologyKey]
+	consumerConfig, err := s.getConsumerConfig(ctx, storageConsumer.Status.ResourceNameMappingConfigMap.Name, storageConsumer.Namespace)
+	if err != nil {
+		klog.Error(err)
+		return nil, status.Errorf(codes.Internal, "failed to storage consumer config")
```
rewantsoni, line 719:
```suggestion
		klog.Errorf("failed to get consumer config: %v", err)
		return nil, status.Errorf(codes.Internal, "failed to produce client state hash")
```

rewantsoni (changes requested):
one small comment, other than that the PR looks good

services/provider/server/server.go
```diff
@@ -420,6 +449,9 @@ func (s *OCSProviderServer) GetDesiredClientState(ctx context.Context, req *pb.G
 			inMaintenanceMode,
 			isConsumerMirrorEnabled,
 			topologyKey,
+			isRbdDriverEnabled,
```
parth-gr, line 452: have you tested for disabling of any driver at runtime?

services/provider/server/server.go
```diff
@@ -349,7 +356,14 @@ func (s *OCSProviderServer) GetDesiredClientState(ctx context.Context, req *pb.G
 	case ocsv1alpha1.StorageConsumerStateDeleting:
 		return nil, status.Errorf(codes.NotFound, "StorageConsumer is in deleting phase")
 	case ocsv1alpha1.StorageConsumerStateReady:
-		kubeResources, err := s.getKubeResources(ctx, consumer)
+
+		consumerConfig, err := s.getConsumerConfig(ctx, consumer.Status.ResourceNameMappingConfigMap.Name, consumer.Namespace)
```
parth-gr, line 360: Is it a user editable cm?
Or is it controlled by storageconsumer?

services/provider/server/server.go
rohan47: We were only enabling drivers

services/provider/server/server.go
rohan47: its controlled by storage consumer

services/provider/server/server.go
```diff
@@ -405,6 +419,21 @@ func (s *OCSProviderServer) GetDesiredClientState(ctx context.Context, req *pb.G
...
+
+		isCephFsDriverEnabled := consumerConfig.GetCephFsClientProfileName() != ""
+		if isCephFsDriverEnabled {
+			response.CephFsDriverRequirements = &pb.CephFsDriverRequirements{}
+		}
+
+		isNfsDriverEnabled := consumerConfig.GetNfsClientProfileName() != ""
+		if isNfsDriverEnabled {
+			response.NfsDriverRequirements = &pb.NfsDriverRequirements{}
+		}
```
nb-ohad, line 435: @rohan47 I am not sure this is the correct source of truth for the decision you are trying to make? What is the thinking behind basing the decision on profile name existence in the consumers configmap?

services/provider/server/server.go
```diff
@@ -212,7 +212,14 @@ func (s *OCSProviderServer) GetStorageConfig(ctx context.Context, req *pb.Storag
...
 	case ocsv1alpha1.StorageConsumerStateReady:
-		kubeResources, err := s.getKubeResources(ctx, consumerObj)
+
+		consumerConfig, err := s.getConsumerConfig(ctx, consumerObj.Status.ResourceNameMappingConfigMap.Name, consumerObj.Namespace)
+		if err != nil {
+			klog.Errorf("failed to get consumer config: %v", err)
+			return nil, status.Errorf(codes.Internal, "failed to produce client state")
+		}
+
+		kubeResources, err := s.getKubeResources(ctx, consumerObj, consumerConfig)
```
nb-ohad, line 222: Why are we touching this function?

nb-ohad (changes requested):

services/provider/server/server.go
rohan47: I was using the Profile names because we were enabling drivers for a client for a particular storageconsumer, and consumerConfig only gives us profilenames.

Changed the implementation to user `util.GetAvailableServices()`

rewantsoni:
@rohan47 What will happen in the case of external mode? From what I recall driver CRs are created by ocs-client-op and the CephConnection/ClientProfile is created by rook. Since there are no storageClients we might not deploy the driver CRs and there would be no drivers for external mode

services/provider/server/server.go
rohan47: @nb-ohad using `util.GetAvailableServices()` gets all the running services based on the StorageCluster and not StorageConsumer, and also requires ocs-provider-server to have permissions to get cephfilesystems, cephnfses and noobaas.

It will make sense to use distributed storageClasses as the source of truth for enabling drivers.
If you agree with using storageClasses I will make the changes to this PR.

cc @leelavg

services/provider/server/server.go
```diff
@@ -310,6 +310,21 @@ func (s *OCSProviderServer) GetDesiredClientState(ctx context.Context, req *pb.G
 			return nil, status.Errorf(codes.Internal, "failed to produce client state hash")
 		}
 
+		availableServices, err := util.GetAvailableServices(ctx, s.client, storageCluster)
+		if err != nil {
+			logger.Error(err, "failed to get available services")
+			return nil, status.Errorf(codes.Internal, "failed to produce client state")
```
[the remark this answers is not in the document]
rohan47: as per the error logging format, it should be failed to produce client state hash
https://github.com/red-hat-storage/ocs-operator/pull/3195#discussion_r2251080628

Madhu-1 (approved):

PR #3200, opened 2025-05-02
storageautoscaler: scale only when more than threshold minus 10% percent of cluster is used

parth-gr:
add a another check for scaling to scale only when more than threshold minus 10% of cluster is full

in some cases ceph takes time to balance the osd,
so its important to calculate the total cluster size for a device class

### DIFF
--- a/controllers/storageautoscaler/scraper.go
+++ b/controllers/storageautoscaler/scraper.go
@@ -26,6 +26,12 @@ type StorageAutoscalerScraper struct {
 	EventCh           chan event.GenericEvent
 }
 
+type OsdUsage struct {
+	TotalUsage   float64
+	HighestUsage float64
+	TotalOsd     float64
+}
+
 func (r *StorageAutoscalerScraper) ScrapeMetricsPeriodically(ctx context.Context) {
 	for {
 		select {
@@ -59,7 +65,7 @@ func (r *StorageAutoscalerScraper) enqueueGenericEventsFromMetrics(ctx context.C
 
 	// refresh the sync map
 	r.SyncMap.Clear()
-	// update the sync map with the highest value of osd usage per device class
+	// update the sync map with the highest value of osd usage and total osd usage per device class
 	updateSyncMap(metrics, r.SyncMap)
 
 	filteredObjectList := filterObjectsForScaling(objectList, r.SyncMap, r.Log)
@@ -71,10 +77,14 @@ func (r *StorageAutoscalerScraper) enqueueGenericEventsFromMetrics(ctx context.C
 // update the sync map with the highest value of osd usage per device class
 func updateSyncMap(metrics model.Vector, syncMap *sync.Map) {
 	// create a map of device class to highest osd usage
-	deviceClassUsage := make(map[string]float64)
+	deviceClassUsage := make(map[string]OsdUsage)
 	for _, osd := range metrics {
 		deviceClass := string(osd.Metric["device_class"])
-		deviceClassUsage[deviceClass] = math.Max(deviceClassUsage[deviceClass], float64(osd.Value))
+		deviceClassUsage[deviceClass] = OsdUsage{
+			HighestUsage: math.Max(deviceClassUsage[deviceClass].HighestUsage, float64(osd.Value)),
+			TotalUsage:   deviceClassUsage[deviceClass].TotalUsage + float64(osd.Value),
+			TotalOsd:     deviceClassUsage[deviceClass].TotalOsd + 1,
+		}
 	}
 
 	// update the sync map
@@ -99,22 +109,45 @@ func filterObjectsForScaling(storageAutoScalerList *ocsv1.StorageAutoScalerList,
 	for _, storageAutoScaler := range storageAutoScalerList.Items {
 		deviceClass := storageAutoScaler.Spec.DeviceClass
 		deviceClassThreshold := storageAutoScaler.Spec.StorageScalingThresholdPercent
-		// get the osd usage from the sync map
-		usage, ok := syncMap.Load(deviceClass)
-		if !ok {
-			log.Error(fmt.Errorf("osd usage not found for device class"), "osd usage not found for device class", "device class", deviceClass)
-		} else if usage == nil {
-			err := fmt.Errorf("osd usage is nil for device class, device class: %s", storageAutoScaler.Spec.DeviceClass)
-			log.Error(err, "device class not found in sync map")
-		} else {
-			deviceClassUsage := usage.(float64)
-			if (deviceClassUsage * 100) >= float64(deviceClassThreshold) {
-				objectList = append(objectList, types.NamespacedName{Namespace: storageAutoScaler.Namespace, Name: storageAutoScaler.Name})
-			}
+		if isScalingRequired(syncMap, deviceClassThreshold, log, deviceClass) {
+			objectList = append(objectList, types.NamespacedName{Namespace: storageAutoScaler.Namespace, Name: storageAutoScaler.Name})
 		}
 	}
 
 	return objectList
+}
+
+func isScalingRequired(syncMap *sync.Map, deviceClassThreshold int, log logr.Logger, deviceClass string) bool {
+	// get the osd usage from the sync map
+	usage, ok := syncMap.Load(deviceClass)
+	if !ok {
+		log.Error(fmt.Errorf("osd usage not found for device class"), "osd usage not found for device class", "device class", deviceClass)
+	} else if usage == nil {
+		err := fmt.Errorf("osd usage is nil for device class, device class: %s", deviceClass)
+		log.Error(err, "device class not found in sync map")
+	} else {
+		deviceClassUsage := usage.(OsdUsage)
+		if (deviceClassUsage.HighestUsage * 100) >= float64(deviceClassThreshold) {
+			if totalUsageGreaterThanAdjustedThreshold(deviceClassUsage, deviceClassThreshold, log, deviceClass) {
+				return true
+			}
+			log.Info("highest osd usage is greater than the threshold, but total cluster usage is less than the (threshold-10)%, skipping scaling", "device class", deviceClass, "highest osd usage", deviceClassUsage.HighestUsage*100, "total cluster usage", averageOsdUsage(deviceClassUsage, log, deviceClass))
+		}
+	}
+
+	return false
+}
+
+func totalUsageGreaterThanAdjustedThreshold(deviceClassUsage OsdUsage, deviceClassThreshold int, log logr.Logger, deviceClass string) bool {
+	return averageOsdUsage(deviceClassUsage, log, deviceClass) > math.Max(float64(deviceClassThreshold)-10, float64(0))
+}
+
+func averageOsdUsage(deviceClassUsage OsdUsage, log logr.Logger, deviceClass string) float64 {
+	if deviceClassUsage.TotalOsd == 0 {
+		log.Error(fmt.Errorf("total osd is 0"), "highest osd usage is loaded incorrectly", "device class", deviceClass)
+		return 0
+	}
+	return (deviceClassUsage.TotalUsage * 100) / deviceClassUsage.TotalOsd
 }
 
 func sendGenericEvent(objectList []types.NamespacedName, eventCh chan event.GenericEvent, log logr.Logger) {
--- a/controllers/storageautoscaler/scraper_test.go
+++ b/controllers/storageautoscaler/scraper_test.go
@@ -27,7 +27,11 @@ func TestFilterObjectsForScaling(t *testing.T) {
 	}
 
 	syncMap := &sync.Map{}
-	syncMap.Store("deviceClass", 0.60)
+	syncMap.Store("deviceClass", OsdUsage{
+		TotalUsage:   1.80,
+		TotalOsd:     3,
+		HighestUsage: 0.60,
+	})
 
 	log := logr.Logger{}
 	objectList := filterObjectsForScaling(&scalerList, syncMap, log)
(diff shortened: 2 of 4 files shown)


controllers/storageautoscaler/scraper.go
```diff
@@ -107,9 +112,11 @@ func filterObjectsForScaling(storageAutoScalerList *ocsv1.StorageAutoScalerList,
 			err := fmt.Errorf("osd usage is nil for device class, device class: %s", storageAutoScaler.Spec.DeviceClass)
 			log.Error(err, "device class not found in sync map")
 		} else {
-			deviceClassUsage := usage.(float64)
-			if (deviceClassUsage * 100) >= float64(deviceClassThreshold) {
-				objectList = append(objectList, types.NamespacedName{Namespace: storageAutoScaler.Namespace, Name: storageAutoScaler.Name})
+			deviceClassUsage := usage.(OsdUsage)
+			if deviceClassUsage.TotalUsage*100 >= 50 {
```
travisn, line 116: Let's make this more dependent on the configured threshold. After some quick calculations I would suggest `deviceClassThreshold - 10%`. By default, this would be a 60% threshold, which gives some room for OSDs to have different filled ratios, but also not too low of a threshold. If the threshold is too low, it may not work as well when we are horizontally scaling and adding 3 OSDs (instead of doubling OSD sizes in vertical expansion). Proportionally, we won't be adding as much per horizontal expansion as the cluster gets bigger.

parth-gr: Correct I did some calculations too,

if we are scaling horizontally after 8tb osds,

Total capacity we increase like, 
24, 48, 72, 96......, 216, 240, 268...

248 ---> 268 => 10% increase only....

Which will fail the above algorithm if someone if going above 240Tb.....

And the max supported we can go to is 768Tb

parth-gr: (768-744)/744 = 3.2% approx

So we can have a limit saying atleast (threshod-3%) is full??

travisn: The OSD filled percentage will always have some variability. 3% is too small. As discussed offline, horizontal scaling a small percentage such as 3% is really the issue. I'd suggest:
1. Use `threshold - 10%` for this secondary decision point (in this PR)
2. Horizontal scaling should increase the storage capacity at least 10% (in a separate PR). Instead of adding 3 OSDs, we will need to add some multiple of 3 OSDs to make it expand at least 10%. This means larger clusters will expand more proportionally.

parth-gr: We can open a separate discussion of second point.

controllers/storageautoscaler/scraper.go
```diff
@@ -107,9 +112,11 @@ func filterObjectsForScaling(storageAutoScalerList *ocsv1.StorageAutoScalerList,
 			err := fmt.Errorf("osd usage is nil for device class, device class: %s", storageAutoScaler.Spec.DeviceClass)
 			log.Error(err, "device class not found in sync map")
 		} else {
-			deviceClassUsage := usage.(float64)
-			if (deviceClassUsage * 100) >= float64(deviceClassThreshold) {
-				objectList = append(objectList, types.NamespacedName{Namespace: storageAutoScaler.Namespace, Name: storageAutoScaler.Name})
+			deviceClassUsage := usage.(OsdUsage)
+			if deviceClassUsage.TotalUsage*100 >= math.Max(float64(deviceClassThreshold)-10, 0) {
+				if (deviceClassUsage.HighestUsage * 100) >= float64(deviceClassThreshold) {
```
travisn, line 117: nit: It may be helpful to log if the highest usage OSD is above the threshold, but we are skipping the expansion. So you may want to switch the order of these checks, and log in that case?

controllers/storageautoscaler/scraper.go
alfonsomthd: @parth-gr @travisn are these new calculations affecting the way the UI calculates all the discrete values for the cluster capacity?

travisn: @alfonsomthd I expect those discrete values will all still be valid. 
@parth-gr One consideration is that when we hit the max capacity, we only scale to that capacity (and not horizontally scale too many OSDs past that). Is that already in the calculation?

parth-gr: If we are scaling horizontally, we add 3 osds, and if we cant add 3 more osd we do not scale, https://github.com/red-hat-storage/ocs-operator/blob/main/controllers/storageautoscaler/storageautoscaler_controller.go#L165

>I expect those discrete values will all still be valid.

It will change as we will be scaling to specific numbers

travisn: > If we are scaling horizontally, we add 3 osds, and if we cant add 3 more osd we do not scale,

Understood, though let me give an example just to be clear... Let's say we have a cluster with max capacity 400TiB
and current capacity is 360TiB...
When we reach the 70% threshold:
   - Expansion of 3 OSDs (8TiB each) would take us to 384TiB capacity. 
   - This is only a 6.7% expansion and we want a minimum of 10% expansion if possible. 
   - So with the new proposal we would consider adding 6 OSDs to get at least 10% expansion. But that would give us 408TiB, which is over the 400TiB capacity, which is too much. 
   - So we just fall back to 3 OSDs expansion in this case.

parth-gr: Agree, we can do that, but that also means, 10% is the min. increase?

as we have 3 osd with 8 tb and we add three more, we would be increasing 50% of the cluster.

travisn: Yes, let's make 10% the min increase. For large clusters, horizontal scale of 3 OSDs is not enough of a jump.

alfonsomthd: @parth-gr Can you update in the design doc the new algorithm? We need to update the calculations also on UI.

parth-gr: @alfonsomthd as planned we will be doing it as part of 4.20

controllers/storageautoscaler/scraper.go
```diff
@@ -71,10 +77,10 @@ func (r *StorageAutoscalerScraper) enqueueGenericEventsFromMetrics(ctx context.C
 // update the sync map with the highest value of osd usage per device class
 func updateSyncMap(metrics model.Vector, syncMap *sync.Map) {
 	// create a map of device class to highest osd usage
-	deviceClassUsage := make(map[string]float64)
+	deviceClassUsage := make(map[string]OsdUsage)
 	for _, osd := range metrics {
 		deviceClass := string(osd.Metric["device_class"])
-		deviceClassUsage[deviceClass] = math.Max(deviceClassUsage[deviceClass], float64(osd.Value))
+		deviceClassUsage[deviceClass] = OsdUsage{HighestUsage: math.Max(deviceClassUsage[deviceClass].HighestUsage, float64(osd.Value)), TotalUsage: deviceClassUsage[deviceClass].TotalUsage + float64(osd.Value), TotalOsd: deviceClassUsage[deviceClass].TotalOsd + 1}
```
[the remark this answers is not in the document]
travisn: Or even better, have a helper method create this struct and unit test it

controllers/storageautoscaler/scraper.go
```diff
@@ -107,16 +113,23 @@ func filterObjectsForScaling(storageAutoScalerList *ocsv1.StorageAutoScalerList,
...
+				}
+				log.Info("highest osd usage is greater than the threshold, but total cluster usage is less than the (threshold-10)%, skipping scaling", "device class", deviceClass, "highest osd usage", deviceClassUsage.HighestUsage*100, "total osd usage", deviceClassUsage.TotalUsage*100)
 			}
 		}
 	}
 
 	return objectList
 }
 
+func totalUsageGreaterThanThresholdMinusTenPercent(deviceClassUsage OsdUsage, deviceClassThreshold int) bool {
```
travisn, line 129: nit: don't bake the 10% threshold into the method name
```suggestion
func totalUsageGreaterThanAdjustedThreshold(deviceClassUsage OsdUsage, deviceClassThreshold int) bool {
```

controllers/storageautoscaler/scraper.go
```diff
@@ -107,16 +113,23 @@ func filterObjectsForScaling(storageAutoScalerList *ocsv1.StorageAutoScalerList,
...
+	return (deviceClassUsage.TotalUsage/deviceClassUsage.TotalOsd)*100 >= math.Max(float64(deviceClassThreshold)-1, 0)
```
travisn, line 130: Check if `TotalOsd` is 0, otherwise you could have divide by 0.

travisn: Wait, we need a different computation. My comment [here](https://github.com/red-hat-storage/ocs-operator/pull/3200#discussion_r2080388652) must not be clear. The 10% difference to check is that if we expand horizontally, that we are expanding at least 10% of the capacity. If adding 3 OSDs will result in less than 10% expansion, then we need to have some multiple of 3 OSDs until it is at least 10%.

An exception to this rule would be if adding more than 3 OSDs will exceed the total autoscale storage limit, then we can just expand 3.

controllers/storageautoscaler/scraper.go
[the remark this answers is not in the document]
parth-gr: Its a straight forward input, i dont see a value to unit test it

controllers/storageautoscaler/scraper.go
parth-gr: we will do that process in the next release, currently sticking to add only 3 osds at a time

parth-gr:
Testing looks good,

for testing I did (threshod-1) (8-1=7) as the cluster was less filled, but worked fine,

After first resize, it didnt did a second resize, and the log was like 

```
```
{"level":"info","ts":"2025-05-14T14:21:34Z","logger":"scraper.StorageAutoscalerScraper","msg":"highest osd usage is greater than the threshold, but total cluster usage is less than the (threshold-10)%, skipping scaling","device class":"ssd","highest osd usage":9.021980285644531,"total osd usage":4.669428507486979}
```

```

controllers/storageautoscaler/scraper_test.go
```diff
@@ -156,6 +164,27 @@ func TestUpdateSyncMap(t *testing.T) {
...
+		deviceClassUsage := OsdUsage{
+			TotalUsage:   1.80,
+			TotalOsd:     3,
+			HighestUsage: 0.60,
+		}
+		result := totalUsageGreaterThanAdjustedThreshold(deviceClassUsage, 50)
+		assert.True(t, result)
+	})
+	t.Run("after resize test total usage less than threshold minus 10 percent", func(t *testing.T) {
+		deviceClassUsage := OsdUsage{
```
travisn, line 182: How about adding a comment here to explain more? I had to think about this too long. :) 
For example, that the average is 32% so it's less than the 40% adjusted threshold.

controllers/storageautoscaler/scraper.go
travisn: Oh right, that will be a future PR. The calculation for this purpose looks good now.

controllers/storageautoscaler/scraper_test.go
```diff
@@ -156,6 +164,27 @@ func TestUpdateSyncMap(t *testing.T) {
...
+			TotalUsage:   1.90,
+			TotalOsd:     6,
+			HighestUsage: 0.55,
+		}
```
travisn, line 186: How about adding a test case that is barely over 40% adjusted threshold? Boundary checking is helpful for computations like this.

controllers/storageautoscaler/scraper.go
```diff
@@ -107,16 +117,30 @@ func filterObjectsForScaling(storageAutoScalerList *ocsv1.StorageAutoScalerList,
 			err := fmt.Errorf("osd usage is nil for device class, device class: %s", storageAutoScaler.Spec.DeviceClass)
 			log.Error(err, "device class not found in sync map")
 		} else {
-			deviceClassUsage := usage.(float64)
-			if (deviceClassUsage * 100) >= float64(deviceClassThreshold) {
-				objectList = append(objectList, types.NamespacedName{Namespace: storageAutoScaler.Namespace, Name: storageAutoScaler.Name})
+			deviceClassUsage := usage.(OsdUsage)
+			if (deviceClassUsage.HighestUsage * 100) >= float64(deviceClassThreshold) {
+				if deviceClassUsage.TotalOsd == 0 {
```
travisn, line 122: It's best to check for 0 in the same place of the potential divide by zero method. So I'd suggest to move this check into the `totalUsageGreaterThanAdjustedThreshold()` method, log the message, and then just return false in that case. Then you can unit test that case too.

travisn (changes requested):

controllers/storageautoscaler/scraper.go
parth-gr: we have a another log message at line 132, `highest osd usage is greater than the threshold, but t` 

where we again need to divide it by osd_count, so thats why had a big if with it,

Not sure how can we move that log, but will think again

travisn: How about add a helper method for `averageOsdUsage()`, and then it can check for the divide by 0. If number of OSDs is 0, it could just log the error and return 0. That helper can be called from both the log message and `totalUsageGreaterThanAdjustedThreshold()`

travisn (commented):
LGTM, perhaps just another small unit test

travisn (commented):
/lgtm

controllers/storageautoscaler/scraper.go
```diff
@@ -99,24 +109,47 @@ func filterObjectsForScaling(storageAutoScalerList *ocsv1.StorageAutoScalerList,
...
-			}
+		if checkIfScalingRequired(syncMap, deviceClassThreshold, log, deviceClass) {
+			objectList = append(objectList, types.NamespacedName{Namespace: storageAutoScaler.Namespace, Name: storageAutoScaler.Name})
 		}
 	}
 
 	return objectList
 }
 
+func checkIfScalingRequired(syncMap *sync.Map, deviceClassThreshold int, log logr.Logger, deviceClass string) bool {
```
alfonsomthd, line 120: Just a simpler name suggestion:
```suggestion
func isScalingRequired(syncMap *sync.Map, deviceClassThreshold int, log logr.Logger, deviceClass string) bool {
```

parth-gr: Done

alfonsomthd: Tagged!

PR #3201, opened 2025-05-05
Set mon_max_pg_per_osd to 1000 to avoid too many PGs warning

malayparida2000:
Ref-https://issues.redhat.com/browse/DFBUGS-2372
After mon_target_pg_per_osd is set to 400, sometimes we see a ceph health warning about too many PGs per OSD as the osd count reaches 600-700 sometimes. Setting the mon_max_pg_per_osd to 1000 will help avoid this warning.


malayparida2000:
Depends on https://github.com/red-hat-storage/ocs-operator/pull/3170

travisn (commented):
/lgtm

PR #3203, opened 2025-05-05
Add target size ratio value to custom cephBlockPool during creation

malayparida2000:
Custom cephBlockPool are not reconciled by ocs-operator, so if the user does not set the target size ratio value, it will be not set. This would affect the scaling of the pool.

So we need to add the target size ratio default value to the custom cephBlockPool during creation.

Part Of https://github.com/red-hat-storage/odf-console/pull/2067

https://issues.redhat.com/browse/DFBUGS-2059


SanjalKatiyar:
LGTM

PR #2213, opened 2023-10-15
Add default virtualization storage class annotation

akalenyu:
This signals to kubevirt that requests not explicitly specifying any storage class should use this one.
https://github.com/kubevirt/containerized-data-importer/pull/2913

PR #3208, opened 2025-05-06
do not mark virt SC as default for kubevirt workloads

leelavg:
we were partially reconciling virt storageclass after it's creation and this allowed external entities to remove the annotation which shouldn't have been allowed.

with updated design ocs-operator holds full control over the storageclass and this is disallowing user not to remove the annotation. since we allow the addition of annotations we create the SC w/o the annotation and user can mark it as default.

we are considering this as an interim fix to unblock CNV.

revert of #2213

### DIFF
--- a/controllers/util/storageclasses.go
+++ b/controllers/util/storageclasses.go
@@ -132,8 +132,13 @@ func NewDefaultVirtRbdStorageClass(
 		ObjectMeta: metav1.ObjectMeta{
 			Annotations: map[string]string{
 				"description": "Provides RWO and RWX Block volumes suitable for Virtual Machine disks",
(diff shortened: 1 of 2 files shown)


leelavg:
/cherrypick release-4.19
via DFBUGS-2342

leelavg:
@nb-ohad possible to override this PR? The CI started due to a merge to main (#3202) which doesn't effect current PR.

The previous run was [successful](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/red-hat-storage_ocs-operator/3208/pull-ci-red-hat-storage-ocs-operator-main-ocs-operator-bundle-e2e-aws/1919704636734836736). Asking because I want to backport and make it to today's nightly.

PR #3210, opened 2025-05-06
Set Maintenance Mode only if RDR is configured & rbd-mirror CR is absent

malayparida2000:
Until now we were setting maintenance mode based on only the absence of the rbd-mirror CR only, due to which maintenance mode was getting set always even in non rdr cluster. To solve this we are now checking if the cluster is configured for RDR & if the rbd-mirror CR is absent.
Ref-https://issues.redhat.com/browse/DFBUGS-2315

### DIFF
--- a/controllers/util/k8sutil.go
+++ b/controllers/util/k8sutil.go
@@ -359,3 +359,10 @@ func mutate(f controllerutil.MutateFn, key client.ObjectKey, obj client.Object) 
 	}
 	return nil
 }
+
+// If the storage-client-mapping configMap is present, then the cluster is configured for RDR otherwise not
+func IsClusterConfiguredForRDR(ctx context.Context, kubeClient client.Client, namespace string) bool {
+	cm := &corev1.ConfigMap{}
+	err := kubeClient.Get(ctx, types.NamespacedName{Name: StorageClientMappingConfigName, Namespace: namespace}, cm)
+	return err == nil
+}
--- a/services/provider/server/server.go
+++ b/services/provider/server/server.go
@@ -968,15 +968,18 @@ func (s *OCSProviderServer) GetBlockPoolsInfo(ctx context.Context, req *pb.Block
 }
 
 func (s *OCSProviderServer) isSystemInMaintenanceMode(ctx context.Context) (bool, error) {
-	// found - false, not found - true
-	cephRBDMirrors := &rookCephv1.CephRBDMirror{}
-	cephRBDMirrors.Name = util.CephRBDMirrorName
-	cephRBDMirrors.Namespace = s.namespace
-	err := s.client.Get(ctx, client.ObjectKeyFromObject(cephRBDMirrors), cephRBDMirrors)
-	if client.IgnoreNotFound(err) != nil {
-		return false, err
+	// if cluster is configured for RDR & cephRBDMirror CR is not found, then system is in maintenance mode
+	if util.IsClusterConfiguredForRDR(ctx, s.client, s.namespace) {
+		cephRBDMirrors := &rookCephv1.CephRBDMirror{}
+		cephRBDMirrors.Name = util.CephRBDMirrorName
+		cephRBDMirrors.Namespace = s.namespace
+		err := s.client.Get(ctx, client.ObjectKeyFromObject(cephRBDMirrors), cephRBDMirrors)
+		if client.IgnoreNotFound(err) != nil {
+			return false, err
+		}
+		return kerrors.IsNotFound(err), nil
 	}
-	return kerrors.IsNotFound(err), nil
+	return false, nil
 }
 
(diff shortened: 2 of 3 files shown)


rewantsoni:
@malayparida2000 A system is in Maintenance Mode when rbd mirror daemon is deleted/removed. Presence of rbd mirror doesn't mean maintenance mode is activated

malayparida2000:
> @malayparida2000 A system is in Maintenance Mode when rbd mirror daemon is deleted/removed. Presence of rbd mirror doesn't mean maintenance mode is activated

Okay, seems like there was some confusion on my side based on the bug description

```
Actual results:
Maintenance Mode is set even when DR is not enabled

Expected results:
 Maintenance Mode should be set only when mirroring/RDR is enabled
```

Can you specify the exact condition when maintenanceMode should be on?

rewantsoni:
> Can you specify the exact condition when maintenanceMode should be on?

Since Maintenance Mode is a cluster-wide operation. It should be on when mirroring is enabled for at least one consumer and the cephRBDMirror is not deployed.

services/provider/server/server.go
```diff
@@ -907,15 +907,44 @@ func (s *OCSProviderServer) GetBlockPoolsInfo(ctx context.Context, req *pb.Block
...
+			return false, err
+		}
+		if enabled {
+			atLeastOneConsumerMirrorEnabled = true
+			break
+		}
+	}
+	if !atLeastOneConsumerMirrorEnabled {
+		return false, nil
+	}
```
rewantsoni, line 930: We could check if the data field in storage-client-mapping configMap is empty or not.

rewantsoni (changes requested):

services/provider/server/server.go
malayparida2000: Isn't that what the existing isConsumerMirrorEnabled func here does https://github.com/red-hat-storage/ocs-operator/blob/3ea6c860f7885b1e678b8242dc4e8c4036956481/services/provider/server/server.go#L921?

rewantsoni: isConsumerMirrorEnabled will load the configMap again and again from etdc, and since we are not using a cached client for the worst case scenario, it will hit the API server as many times we have the storageConsumer.

malayparida2000: Thanks. Updated now to avoid fetching the configmap again n again.

nb-ohad (changes requested):

services/provider/server/server.go
```diff
@@ -907,15 +907,31 @@ func (s *OCSProviderServer) GetBlockPoolsInfo(ctx context.Context, req *pb.Block
...
+		cephRBDMirror := &rookCephv1.CephRBDMirror{}
+		err := s.client.Get(ctx, client.ObjectKey{Name: util.CephRBDMirrorName, Namespace: s.namespace}, cephRBDMirror)
+		// if rbd-mirror cr is not present, system is in maintenance mode
+		if kerrors.IsNotFound(err) {
+			return true, nil
+		}
+		return false, err
+	}
+
+	return false, nil
```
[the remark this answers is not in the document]
rewantsoni: Maintenance Mode is desired when there is an annotation on the storageConsumer CR
Maintenance Mode is activated when it is desired by any one of the consumers and the cephRBDMirror CR is not present.

Absence of cephRBDMirror cannot be the only condition to conclude that the system is in maintenance mode. As for non-RDR setups, the cephRBDMirror is always absent. 

With that being said, maybe we need an indicator on the system to identify if it is in maintenance mode or not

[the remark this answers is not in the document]
malayparida2000: > Maintenance Mode is desired when there is an annotation on the storageConsumer CR
> Maintenance Mode is activated when it is desired by any one of the consumers and the cephRBDMirror CR is not present.
I have used this logic for now in this fix PR

nb-ohad (changes requested):

controllers/util/k8sutil.go
```diff
@@ -359,3 +359,10 @@ func mutate(f controllerutil.MutateFn, key client.ObjectKey, obj client.Object)
 	}
 	return nil
 }
+
+// If the storage-client-mapping configMap is present, then the cluster is configured for RDR otherwise not
+func IsClusterConfiguredForRDR(ctx context.Context, kubeClient client.Client, namespace string) bool {
+	cm := &corev1.ConfigMap{}
+	err := kubeClient.Get(ctx, types.NamespacedName{Name: StorageClientMappingConfigName, Namespace: namespace}, cm)
+	return err == nil
+}
```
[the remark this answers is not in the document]
malayparida2000: Do we today have the usage of MM without the context of RDR?
@rewantsoni ?

iamniting:
This is obsolete by https://github.com/red-hat-storage/ocs-operator/pull/3723

PR #3213, opened 2025-05-07
improved logs in the provider server for debugging

mrudraia1:
This PR improves the logs in the ocs-provider-server pod.

The earlier logs were not fully explainable, partial information was provided and hence it was confusing to debug based on log events.

I have made changes with respect to  ocs-provider-server pod, since a bug was raised for  log issues.

### DIFF
--- a/services/provider/server/server.go
+++ b/services/provider/server/server.go
@@ -18,6 +18,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/go-logr/logr"
 	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
 	ocsv1alpha1 "github.com/red-hat-storage/ocs-operator/api/v4/v1alpha1"
 	pb "github.com/red-hat-storage/ocs-operator/services/provider/api/v4"
@@ -30,6 +31,7 @@ import (
 	"github.com/blang/semver/v4"
 	csiopv1a1 "github.com/ceph/ceph-csi-operator/api/v1alpha1"
 	replicationv1alpha1 "github.com/csi-addons/kubernetes-csi-addons/api/replication.storage/v1alpha1"
+	"github.com/google/uuid"
 	groupsnapapi "github.com/kubernetes-csi/external-snapshotter/client/v8/apis/volumegroupsnapshot/v1beta1"
 	snapapi "github.com/kubernetes-csi/external-snapshotter/client/v8/apis/volumesnapshot/v1"
 	nbapis "github.com/noobaa/noobaa-operator/v5/pkg/apis"
@@ -78,12 +80,13 @@ type OCSProviderServer struct {
 	pb.UnimplementedOCSProviderServer
 	client                    client.Client
 	scheme                    *runtime.Scheme
+	log                       logr.Logger
 	consumerManager           *ocsConsumerManager
 	storageClusterPeerManager *storageClusterPeerManager
 	namespace                 string
 }
 
-func NewOCSProviderServer(ctx context.Context, namespace string) (*OCSProviderServer, error) {
+func NewOCSProviderServer(ctx context.Context, namespace string, log logr.Logger) (*OCSProviderServer, error) {
 	scheme, err := newScheme()
 	if err != nil {
 		return nil, fmt.Errorf("failed to create new scheme. %v", err)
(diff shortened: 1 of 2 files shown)


services/provider/server/server.go
```diff
@@ -408,7 +408,7 @@ func (s *OCSProviderServer) GetDesiredClientState(ctx context.Context, req *pb.G
 		)
 		response.DesiredStateHash = desiredClientConfigHash
 
-		klog.Infof("successfully returned the config details to the consumer.")
+		klog.Infof("successfully returned the config details of the consumer.")
```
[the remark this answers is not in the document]
mrudraia1: ok, reverted the earlier changes

services/provider/server/server.go
```diff
@@ -338,7 +338,7 @@ func (s *OCSProviderServer) GetDesiredClientState(ctx context.Context, req *pb.G
 		return nil, status.Errorf(codes.Internal, "failed to get StorageConsumer")
 	}
 
-	klog.Infof("Found StorageConsumer for GetDesiredClientState")
+	klog.Infof("storageConsumer found with id %v", req.StorageConsumerUUID)
```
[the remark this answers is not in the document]
mrudraia1: done

rewantsoni:
@mrudraia1 The expectation with the bug that was raised was to improve logging to all the RPC's that we have

mrudraia1:
> @mrudraia1 The expectation with the bug that was raised was to improve logging to all the RPC's that we have

Added logs for all the rpc calls

services/provider/server/server.go
```diff
@@ -566,17 +567,20 @@ func decodeAndValidateTicket(ticket string, pubKey *rsa.PublicKey) (*services.On
 // FulfillStorageClaim RPC call to create the StorageClaim CR on
 // provider cluster.
 func (s *OCSProviderServer) FulfillStorageClaim(ctx context.Context, req *pb.FulfillStorageClaimRequest) (*pb.FulfillStorageClaimResponse, error) {
+	klog.Infof("FullfillStorageClaim to create Storageclaim %v", req.StorageClaimName)
```
rohan47, line 570: Since the function returns with error that code is not implemented, does this info make sense?

services/provider/server/server.go
```diff
@@ -566,17 +567,20 @@ func decodeAndValidateTicket(ticket string, pubKey *rsa.PublicKey) (*services.On
...
 	return nil, status.Error(codes.Unimplemented, "not implemented")
 }
 
 // RevokeStorageClaim RPC call to delete the StorageClaim CR on
 // provider cluster.
 func (s *OCSProviderServer) RevokeStorageClaim(ctx context.Context, req *pb.RevokeStorageClaimRequest) (*pb.RevokeStorageClaimResponse, error) {
+	klog.Infof("RevokeStorageClaim to delete Storageclaim %v", req.StorageClaimName)
```
rohan47, line 577: Since the function returns with error that code is not implemented, does this info make sense?

rohan47 (changes requested):

services/provider/server/server.go
```diff
@@ -176,19 +176,19 @@ func (s *OCSProviderServer) OnboardConsumer(ctx context.Context, req *pb.Onboard
 
 // AcknowledgeOnboarding acknowledge the onboarding is complete
 func (s *OCSProviderServer) AcknowledgeOnboarding(ctx context.Context, req *pb.AcknowledgeOnboardingRequest) (*pb.AcknowledgeOnboardingResponse, error) {
+	klog.Infof("AcknowledgeOnboarding to consumer with id %v the onboarding is complete", req.StorageConsumerUUID)
 	return nil, status.Errorf(codes.Unimplemented, "Not expecting a two step onboarding process")
 }
```
leelavg, line 181: as you can see from the status code, this RPC is no more being used and no point in adding a log.

leelavg (commented):
not all RPCs are called equally and debugging onboard or offboard is straightforward if they fail in any step, looking at the Jira description we are having a bit of issue for RPCs which are called frequently.

I suggest to work w/ the bug reporter and expand the description before proceeding w/ this PR.

services/provider/server/server.go
mrudraia1: acknowledge

nb-ohad (commented):
The API server might execute multiple endpoint implementation in parallel. This means the logging of "Starting ..." and "Ending ..." will not solve the issue described.

Instead, I would suggest the for each endpoint we will created a subset logger that holds context information about the calls, which include the endpoint/operation name and the identity of the key identity param/s this call is working on. 

This way every log inside the specific call could be traced to the call and the key entity the call is working on

PR #3217, opened 2025-05-08
Make mon_target_pg_per_osd = 400 always both for new & upgraded clusters

malayparida2000:
Following discussion on https://issues.redhat.com/browse/DFBUGS-2391 & ODF Weekly call it was decided to make mon_target_pg_per_osd = 400 always, both for new & upgraded clusters. This was decided based on factors like
1. New & Upgraded clusters should not diverge in terms of behavior
2. Old customers are able to utilize the performance gain from this

This will trigger a data movement upon upgrade from 4.18 to 4.19. We have to add upgrade notes, release notes notifying customers of the data movement & the health warning that would appear upon upgrade.

### DIFF
--- a/controllers/storagecluster/cephcluster.go
+++ b/controllers/storagecluster/cephcluster.go
@@ -1563,7 +1563,8 @@ func generateCephReplicatedSpec(initData *ocsv1.StorageCluster, poolType string)
 func getCephClusterCephConfig(sc *ocsv1.StorageCluster) map[string]map[string]string {
 	cephConfig := map[string]map[string]string{
 		"global": {
-			"mon_max_pg_per_osd": "1000",
+			"mon_target_pg_per_osd": "400",
+			"mon_max_pg_per_osd":    "1000",
 		},
 	}
 	if sc.Spec.ManagedResources.CephCluster.CephConfig != nil {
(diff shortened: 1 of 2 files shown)


controllers/storagecluster/cephcluster.go
```diff
@@ -1564,37 +1560,6 @@ func generateCephReplicatedSpec(initData *ocsv1.StorageCluster, poolType string)
...
 }
 
-// setMonTargetPgPerOsd sets the mon_target_pg_per_osd value to 400 if not already set
-func setMonTargetPgPerOsd(cephConfig *map[string]map[string]string) {
-	if *cephConfig == nil {
-		*cephConfig = make(map[string]map[string]string)
-	}
-	if global, exists := (*cephConfig)["global"]; exists {
-		if _, exists := global["mon_target_pg_per_osd"]; !exists {
-			global["mon_target_pg_per_osd"] = "400"
```
travisn, line 1574: Where is this set now? I only see deleted lines in the PR, so maybe there is another place it's set by default that just isn't in the PR?

malayparida2000: Thanks for pointing it out, I had prepared another commit which I forgot to push. Now updated.

PR #3218, opened 2025-05-09
delete rns cr only if there are no rbd images

leelavg:
a delete on cbp cr can't directly be transferred as rns cr deletion or else the rbd images for an rns can't be restored.

we need to look at cbp conditions when it is marked for deletion to proceed with rns deletion or not.


leelavg:
/hold

for testing.

leelavg:
/cherrypick release-4.19
via DFBUGS-2139

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -464,18 +458,44 @@ func (r *StorageConsumerReconciler) reconcileCephRadosNamespace(
 		}
 		rns.Namespace = r.namespace
 
-		if _, err := ctrl.CreateOrUpdate(r.ctx, r.Client, rns, func() error {
-			if err := controllerutil.SetControllerReference(r.storageConsumer, rns, r.Scheme); err != nil {
-				return err
+		deleteRns := false
+		if !bp.DeletionTimestamp.IsZero() && bp.Status != nil {
```
nb-ohad, line 462: Please handle delete "after" normal reconcile code

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -464,18 +458,44 @@ func (r *StorageConsumerReconciler) reconcileCephRadosNamespace(
...
+			for idx := range bp.Status.Conditions {
+				condition := &bp.Status.Conditions[idx]
+				// block pool waiting for rns deletion
+				if condition.Type == rookCephv1.ConditionDeletionIsBlocked &&
+					condition.Reason == rookCephv1.ObjectHasDependentsReason &&
+					condition.Status == corev1.ConditionTrue {
+					deleteRns = true
+				}
```
nb-ohad, line 470: Can you please explain why do we need this? This condition is not a blocker for our RNS deletion

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -464,18 +458,44 @@ func (r *StorageConsumerReconciler) reconcileCephRadosNamespace(
 		}
 		rns.Namespace = r.namespace
 
-		if _, err := ctrl.CreateOrUpdate(r.ctx, r.Client, rns, func() error {
-			if err := controllerutil.SetControllerReference(r.storageConsumer, rns, r.Scheme); err != nil {
-				return err
+		deleteRns := false
```
nb-ohad, line 461: There is no point in tracking a `deleteRNS` flag. There is only a single point in the code where the delete code should run.

nb-ohad (changes requested):

controllers/storageconsumer/storageconsumer_controller.go
leelavg: possible to expand, we either create/update rns or delete rns, if handling after normal, it'll be we reconcile update and then delete.

controllers/storageconsumer/storageconsumer_controller.go
leelavg: yes, not a blocker and at the same time, we do need to know when to delete the rns, ie, no blocked on images but only on rns cr.

controllers/storageconsumer/storageconsumer_controller.go
leelavg: sry, possible to elaborate, we need to decide b/n update or delete.

leelavg:
/unhold
tested, using the feature https://github.com/rook/rook/pull/15817
rns deletion is blocked if there are any rbd images or else proceed w/ deletion.

controllers/storageconsumer/storageconsumer_controller.go
leelavg: done.

controllers/storageconsumer/storageconsumer_controller.go
leelavg: done.

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -464,18 +458,36 @@ func (r *StorageConsumerReconciler) reconcileCephRadosNamespace(
...
+				rns.Spec.BlockPoolName = bp.Name
+				return nil
+			}); err != nil {
+				multierr.AppendInto(&combinedErr, err)
 			}
-			if err := controllerutil.SetOwnerReference(additionalOwner, rns, r.Scheme); err != nil {
-				return err
+		} else {
+			if bp.Status != nil {
+				for idx := range bp.Status.Conditions {
```
nb-ohad, line 477: I think it will make the code clearer if we are to search for the condition using `slices.Index` or `slices.IndexFunc`

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -464,18 +458,36 @@ func (r *StorageConsumerReconciler) reconcileCephRadosNamespace(
...
+					condition := &bp.Status.Conditions[idx]
+					// block pool is not waiting for rbd images
+					if condition.Type == rookCephv1.ConditionPoolDeletionIsBlocked &&
+						condition.Status == corev1.ConditionFalse {
+						if err := r.Delete(r.ctx, rns); client.IgnoreNotFound(err) != nil {
+							multierr.AppendInto(&combinedErr, err)
+						}
+						break
```
nb-ohad, line 485: The loop is a search loop and should only concern itself with finding the item in the list. Handling of the item should be handled after the loop

nb-ohad (changes requested):

controllers/storageconsumer/storageconsumer_controller.go
leelavg: done.

controllers/storageconsumer/storageconsumer_controller.go
leelavg: done.

leelavg:
retested w/ the change of logic.

PR #3221, opened 2025-05-12
controllers: add finalizers for internal client/consumer

ShravaniVangur:
Deletion of internal client/consumer shouldn't be allowed, and they should be removed only when storageCluster is deleted.
To ensure the same, finalizers would be added to storageConsumer and storageClient in such a way that they can only be deleted when the storageCluster is deleted. 

Resolves: [DFBUGS-2333](https://issues.redhat.com/browse/DFBUGS-2333)

### DIFF
--- a/controllers/storagecluster/reconcile.go
+++ b/controllers/storagecluster/reconcile.go
@@ -80,13 +80,17 @@ const (
 	StorageClientCrdName = "storageclients.ocs.openshift.io"
 
 	VolumeGroupSnapshotClassCrdName = "volumegroupsnapshotclasses.groupsnapshot.storage.k8s.io"
+
+	internalComponentFinalizer = "ocs.openshift.io/internal-component"
+
+	labelZoneRegionWithoutBeta = "failure-domain.kubernetes.io/region"
+
+	labelZoneFailureDomainWithoutBeta = "failure-domain.kubernetes.io/zone"
+
+	labelRookPrefix = "topology.rook.io"
 )
 
 var storageClusterFinalizer = "storagecluster.ocs.openshift.io"
-
-const labelZoneRegionWithoutBeta = "failure-domain.kubernetes.io/region"
-const labelZoneFailureDomainWithoutBeta = "failure-domain.kubernetes.io/zone"
-const labelRookPrefix = "topology.rook.io"
 
 var validTopologyLabelKeys = []string{
 	// This is the most preferred key as kubernetes recommends zone and region
--- a/controllers/storagecluster/storageclient.go
+++ b/controllers/storagecluster/storageclient.go
@@ -64,6 +64,9 @@ func (s *storageClient) ensureCreated(r *StorageClusterReconciler, storagecluste
 		// we could just use svcName:port however in-cluster traffic from "*.svc" is generally not proxied and
 		// we using qualified name upto ".svc" makes connection not go through any proxies.
 		storageClient.Spec.StorageProviderEndpoint = fmt.Sprintf("%s.%s.svc:%d", ocsProviderServerName, storagecluster.Namespace, ocsProviderServicePort)
+
+		controllerutil.AddFinalizer(storageClient, internalComponentFinalizer)
+
 		return nil
 	})
 	if err != nil {
@@ -81,9 +84,18 @@ func (s *storageClient) ensureDeleted(r *StorageClusterReconciler, storagecluste
 	}
 	storageClient := &ocsclientv1a1.StorageClient{}
 	storageClient.Name = storagecluster.Name
+
 	if err := r.Delete(r.ctx, storageClient); err != nil && !kerrors.IsNotFound(err) {
 		r.Log.Error(err, "Failed to initiate deletion of local StorageClient CR")
 		return reconcile.Result{}, err
+	}
+
+	if controllerutil.RemoveFinalizer(storageClient, internalComponentFinalizer) {
+		r.Log.Info("Removing finalizer from StorageClient.", "StorageClient:", storageClient.Name, " Finalizer:", internalComponentFinalizer)
+		if err := r.Update(r.ctx, storageClient); err != nil {
+			r.Log.Info("Failed to remove finalizer from StorageClient.", "StorageClient:", storageClient.Name, " Finalizer:", internalComponentFinalizer)
+			return reconcile.Result{}, fmt.Errorf("failed to remove finalizer from StorageClient: %v", err)
+		}
 	}
 	return reconcile.Result{}, nil
 }
--- a/controllers/storagecluster/storageconsumer.go
+++ b/controllers/storagecluster/storageconsumer.go
@@ -66,6 +66,8 @@ func (s *storageConsumer) ensureCreated(r *StorageClusterReconciler, storageClus
 		spec.VolumeSnapshotClasses = volumeSnapshotClassesSpec
 		spec.VolumeGroupSnapshotClasses = volumeGroupSnapshotClassesSpec
 
+		controllerutil.AddFinalizer(storageConsumer, internalComponentFinalizer)
+
 		return nil
 	}); err != nil {
 		return ctrl.Result{}, fmt.Errorf("failed to create/update storageconsumer %s: %v", storageConsumer.Name, err)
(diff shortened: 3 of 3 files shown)


controllers/storagecluster/storageclient.go
```diff
@@ -64,6 +64,15 @@ func (s *storageClient) ensureCreated(r *StorageClusterReconciler, storagecluste
...
 		// we using qualified name upto ".svc" makes connection not go through any proxies.
 		storageClient.Spec.StorageProviderEndpoint = fmt.Sprintf("%s.%s.svc:%d", ocsProviderServerName, storagecluster.Namespace, ocsProviderServicePort)
+
+		if controllerutil.AddFinalizer(storageClient, storageClusterFinalizer) {
+			r.Log.Info("Finalizer not found for StorageClient. Adding finalizer.", "StorageCLient", storageClient.Name)
+			if err := r.Client.Update(r.ctx, storageClient); err != nil {
+				r.Log.Error(err, "Failed to add finalizer to StorageClient")
+				return err
+			}
+		}
```
leelavg, line 74: you are already inside createorupdate, what is the need for another api call?

controllers/storagecluster/storageconsumer.go
```diff
@@ -104,8 +105,29 @@ func (s *storageConsumer) ensureCreated(r *StorageClusterReconciler, storageClus
...
+		r.Log.Info("removing finalizer from StorageConsumer.", "StorageConsumer", storageConsumer.Name)
+		if err := r.Update(r.ctx, storageConsumer); err != nil {
+			r.Log.Info("Failed to remove finalizer from StorageConsumer", "StorageConsumer", storageConsumer.Name)
+			return ctrl.Result{}, fmt.Errorf("failed to remove finalizer from StorageConsumer: %v", err)
+		}
+	}
+
+	if err := r.Client.Delete(r.ctx, storageConsumer); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to delete storageconsumer %s: %v", storageConsumer.Name, err)
+	}
```
leelavg, line 129: be consistent on how you are using embedded struct, `r.Update/r.Delete/r.Get` vs `r.Client.Update/r.Client.Delete/r.Client.Get`

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -107,6 +108,14 @@ func (r *StorageConsumerReconciler) Reconcile(ctx context.Context, request recon
 		return reconcile.Result{}, err
 	}
 
+	if controllerutil.AddFinalizer(r.storageConsumer, StorageConsumerFinalizer) {
```
leelavg, line 111: who is going to remove this finalizer? isn't the PR addressing only internal client & consumer, adding here should be out of scope as the controller deals w/ both internal & external consumers.

controllers/storagecluster/storageclient.go
```diff
@@ -64,6 +64,11 @@ func (s *storageClient) ensureCreated(r *StorageClusterReconciler, storagecluste
 		// we could just use svcName:port however in-cluster traffic from "*.svc" is generally not proxied and
 		// we using qualified name upto ".svc" makes connection not go through any proxies.
 		storageClient.Spec.StorageProviderEndpoint = fmt.Sprintf("%s.%s.svc:%d", ocsProviderServerName, storagecluster.Namespace, ocsProviderServicePort)
+
+		if controllerutil.AddFinalizer(storageClient, storageClusterFinalizer) {
```
nb-ohad, line 68: I am not sure that adding a finalizer called `storagecluster.ocs.openshift.io` makes sense here. A finalizer name should describe the intent behind the finalizer. One good example is the finalizer: `kubernetes.io/pv-protection`.
The name contain the domain (in our case it is `ocs.openshift.io`) and the intent `pv-protection`.

I would suggest to use something like: `ocs.openshift.io/locked-by-owner` or `ocs.openshift.io/internal-component` or something else of the same mindset

nb-ohad (changes requested):

controllers/storagecluster/storageclient.go
ShravaniVangur: Sure. Added `ocs.openshift.io/internal-component` as the finalizer.

controllers/storagecluster/reconcile.go
```diff
@@ -84,6 +84,8 @@ const (
 
 var storageClusterFinalizer = "storagecluster.ocs.openshift.io"
 
+var internalComponentFinalizer = "ocs.openshift.io/internal-component"
```
rewantsoni, line 87: Why a var and not a const?

controllers/storagecluster/storageclient.go
```diff
@@ -64,6 +64,11 @@ func (s *storageClient) ensureCreated(r *StorageClusterReconciler, storagecluste
 		// we could just use svcName:port however in-cluster traffic from "*.svc" is generally not proxied and
 		// we using qualified name upto ".svc" makes connection not go through any proxies.
 		storageClient.Spec.StorageProviderEndpoint = fmt.Sprintf("%s.%s.svc:%d", ocsProviderServerName, storagecluster.Namespace, ocsProviderServicePort)
+
+		if controllerutil.AddFinalizer(storageClient, internalComponentFinalizer) {
+			r.Log.Info("Finalizer not found for StorageClient. Adding finalizer.", "StorageClient", storageClient.Name)
```
rewantsoni, line 69: Should we also add the finalizer name to the logs?

controllers/storagecluster/storageconsumer.go
```diff
@@ -66,6 +66,10 @@ func (s *storageConsumer) ensureCreated(r *StorageClusterReconciler, storageClus
 		spec.VolumeSnapshotClasses = volumeSnapshotClassesSpec
 		spec.VolumeGroupSnapshotClasses = volumeGroupSnapshotClassesSpec
 
+		if controllerutil.AddFinalizer(storageConsumer, internalComponentFinalizer) {
+			r.Log.Info("Finalizer not found for StorageConsumer. Adding finalizer.", "StorageConsumer", storageConsumer.Name)
+		}
```
rewantsoni, line 71: Should we also add the finalizer name to the logs?

controllers/storagecluster/reconcile.go
ShravaniVangur: Should be a const. Updated the same.

controllers/storagecluster/storageclient.go
ShravaniVangur: Added it for `AddFInalizer` and `RemoveFinalizer` logs.

controllers/storagecluster/reconcile.go
rewantsoni: We already have a const block just above this, maybe add it there? Also shouldn't storageClusterFinalizer also be a const instead of var?

controllers/storagecluster/storageclient.go
```diff
@@ -64,6 +64,11 @@ func (s *storageClient) ensureCreated(r *StorageClusterReconciler, storagecluste
 		// we could just use svcName:port however in-cluster traffic from "*.svc" is generally not proxied and
 		// we using qualified name upto ".svc" makes connection not go through any proxies.
 		storageClient.Spec.StorageProviderEndpoint = fmt.Sprintf("%s.%s.svc:%d", ocsProviderServerName, storagecluster.Namespace, ocsProviderServicePort)
+
+		if controllerutil.AddFinalizer(storageClient, internalComponentFinalizer) {
+			r.Log.Info("Finalizer not found for StorageClient. Adding finalizer:", internalComponentFinalizer, ". StorageClient:", storageClient.Name)
```
rewantsoni, line 69: log follows `Info(msg string, keysAndValues ...any)` construct

controllers/storagecluster/storageclient.go
```diff
@@ -64,6 +64,11 @@ func (s *storageClient) ensureCreated(r *StorageClusterReconciler, storagecluste
 		// we could just use svcName:port however in-cluster traffic from "*.svc" is generally not proxied and
 		// we using qualified name upto ".svc" makes connection not go through any proxies.
 		storageClient.Spec.StorageProviderEndpoint = fmt.Sprintf("%s.%s.svc:%d", ocsProviderServerName, storagecluster.Namespace, ocsProviderServicePort)
+
+		if controllerutil.AddFinalizer(storageClient, internalComponentFinalizer) {
+			r.Log.Info("Finalizer not found for StorageClient, adding finalizer.", "StorageClient:", storageClient.Name, "Finalizer:", internalComponentFinalizer)
```
nb-ohad, line 69: Please use a namespace name (which include the namespace as well)

controllers/storagecluster/storageclient.go
```diff
@@ -81,6 +86,15 @@ func (s *storageClient) ensureDeleted(r *StorageClusterReconciler, storagecluste
...
 	storageClient := &ocsclientv1a1.StorageClient{}
 	storageClient.Name = storagecluster.Name
+
+	if controllerutil.RemoveFinalizer(storageClient, internalComponentFinalizer) {
+		r.Log.Info("Removing finalizer from StorageClient.", "StorageClient:", storageClient.Name, " Finalizer:", internalComponentFinalizer)
+		if err := r.Update(r.ctx, storageClient); err != nil {
+			r.Log.Info("Failed to remove finalizer from StorageClient.", "StorageClient:", storageClient.Name, " Finalizer:", internalComponentFinalizer)
+			return reconcile.Result{}, fmt.Errorf("failed to remove finalizer from StorageClient: %v", err)
+		}
+	}
```
nb-ohad, line 96: Finalizer should be removed after you call the Delete operation not before

controllers/storagecluster/storageconsumer.go
```diff
@@ -66,6 +66,10 @@ func (s *storageConsumer) ensureCreated(r *StorageClusterReconciler, storageClus
 		spec.VolumeSnapshotClasses = volumeSnapshotClassesSpec
 		spec.VolumeGroupSnapshotClasses = volumeGroupSnapshotClassesSpec
 
+		if controllerutil.AddFinalizer(storageConsumer, internalComponentFinalizer) {
+			r.Log.Info("Finalizer not found for StorageConsumer, adding finalizer.", "StorageConsumer:", storageConsumer.Name, "Finalizer:", internalComponentFinalizer)
```
nb-ohad, line 70: - Please use a namespace name and not just a name
- if we cannot add a finalizer it should be treated as an error, not as an info in the log. Please remove the log entry and return an error instead

nb-ohad (changes requested):

controllers/storagecluster/storageclient.go
ShravaniVangur: For `storageclient` the namespace is not explicitly set. It only takes the `storagecluster`'s name.

controllers/storagecluster/storageconsumer.go
ShravaniVangur: - Added the namespace
- The condition checks if the finalizer is already present. If not it adds the finalizer and logs it as well. The only issue here would have been if the finalizer was a variable. Since it is a constant these might be the only possible conditions?

controllers/storagecluster/storageclient.go
nb-ohad: You are right, the stoargeclient is a cluster scope CR

controllers/storagecluster/storageclient.go
nb-ohad: The issue with your code is that you might get into a situation where the finalizer is gone but the object is not deleted. Please issue the delete command first then remove the finalizer.

controllers/storagecluster/storageconsumer.go
nb-ohad: you are right regarding the return value of the AddFinalizer function. But in that case, there is not point in logging a message because the finalizer will only get saved at the exist of createOrUpdate and only if successful.

controllers/storagecluster/storageclient.go
ShravaniVangur: Understood. Changed the workflow in the function.

PR #3222, opened 2025-05-12
external: add external mode label to storageclasses

parth-gr:
1) Adding labels to external storageclass and external volume snapshot and volume group snapshot storageclaaes, so distinguish between internal and external storageclass

2) Add a filter to deafult storageconsumer to not use the external classes 

3) Adding filter in provider reconcile to not distribute external storageclasses to the clients

### DIFF
--- a/controllers/storagecluster/external_resources.go
+++ b/controllers/storagecluster/external_resources.go
@@ -2,7 +2,6 @@ package storagecluster
 
 import (
 	"context"
-	"crypto/sha512"
 	"encoding/json"
 	"fmt"
 	"net"
@@ -78,14 +77,6 @@ func checkEndpointReachable(endpoint string, timeout time.Duration) error {
 	return nil
 }
 
-func sha512sum(tobeHashed []byte) (string, error) {
-	h := sha512.New()
-	if _, err := h.Write(tobeHashed); err != nil {
-		return "", err
-	}
-	return fmt.Sprintf("%x", h.Sum(nil)), nil
-}
-
 func parseMonitoringIPs(monIP string) []string {
 	return strings.Fields(strings.ReplaceAll(monIP, ",", " "))
 }
@@ -98,14 +89,6 @@ func findNamedResourceFromArray(extArr []ExternalResource, name string) (Externa
 		}
 	}
 	return ExternalResource{}, fmt.Errorf("Unable to retrieve %q external resource", name)
-}
-
-func (r *StorageClusterReconciler) getExternalSecretDataChecksum(instance *ocsv1.StorageCluster) (string, error) {
-	found, err := r.retrieveSecret(externalClusterDetailsSecret, instance)
-	if err != nil {
-		return "", err
-	}
-	return sha512sum(found.Data[externalClusterDetailsKey])
 }
 
 // retrieveSecret function retrieves the secret object with the specified name
@@ -225,26 +208,12 @@ func (r *StorageClusterReconciler) newExternalCephObjectStoreInstances(
 // being created
 func (obj *ocsExternalResources) ensureCreated(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) (reconcile.Result, error) {
 
-	// rhcs external mode
-	extSecretChecksum, err := r.getExternalSecretDataChecksum(instance)
-	if err != nil {
-		r.Log.Error(err, "Failed to get checksum of external secret data.")
-		return reconcile.Result{}, err
-	}
-
-	// if the 'ExternalSecretHash' and fetched hash are same, then return
-	if instance.Status.ExternalSecretHash == extSecretChecksum {
-		return reconcile.Result{}, nil
-	}
-
-	err = r.createExternalStorageClusterResources(instance)
+	err := r.createExternalStorageClusterResources(instance)
 	if err != nil {
 		r.Log.Error(err, "Could not create ExternalStorageClusterResource.", "StorageCluster", klog.KRef(instance.Namespace, instance.Name))
 		return reconcile.Result{}, err
 	}
 
-	// external resources are successfully created, update the checksums in the status
-	instance.Status.ExternalSecretHash = extSecretChecksum
 	return reconcile.Result{}, nil
 }
 
@@ -460,6 +429,8 @@ func (r *StorageClusterReconciler) createExternalStorageClusterResources(instanc
 			for k, v := range d.Data {
 				scc.storageClass.Parameters[k] = v
 			}
+			// add external mode label to storageclass
+			util.AddLabel(scc.storageClass, util.ExternalClassLabelKey, strconv.FormatBool(true))
 			availableSCCs = append(availableSCCs, scc)
 		}
 	}
@@ -552,7 +523,7 @@ func (r *StorageClusterReconciler) createExternalModeStorageClasses(sccs []Stora
 			if existing.DeletionTimestamp != nil {
 				return fmt.Errorf("failed to restore StorageClass  %s because it is marked for deletion", existing.Name)
 			}
-			if !reflect.DeepEqual(sc.Parameters, existing.Parameters) {
+			if !reflect.DeepEqual(sc.Parameters, existing.Parameters) || existing.Labels[util.ExternalClassLabelKey] != sc.Labels[util.ExternalClassLabelKey] {
 				// Since we have to update the existing StorageClass
 				// So, we will delete the existing storageclass and create a new one
 				r.Log.Info("StorageClass needs to be updated, deleting it.", "StorageClass", klog.KRef(sc.Namespace, existing.Name))
--- a/controllers/storagecluster/storageconsumer.go
+++ b/controllers/storagecluster/storageconsumer.go
@@ -17,6 +17,8 @@ import (
 	storagev1 "k8s.io/api/storage/v1"
 	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
@@ -109,6 +111,19 @@ func (s *storageConsumer) ensureDeleted(_ *StorageClusterReconciler, _ *ocsv1.St
 	return ctrl.Result{}, nil
 }
 
+func getExternalClassesBlaclistSelector() labels.Selector {
+	blackListRequirement, err := labels.NewRequirement(
+		util.ExternalClassLabelKey,
+		selection.NotEquals,
+		[]string{"true"},
+	)
+	if err != nil {
+		panic(fmt.Sprintf("Error in external class label selector definition: %v", err))
+	}
+
+	return labels.NewSelector().Add(*blackListRequirement)
+}
+
 func getLocalStorageClassNames(ctx context.Context, kubeClient client.Client, storageCluster *ocsv1.StorageCluster) (
 	[]ocsv1a1.StorageClassSpec, error) {
 
@@ -140,11 +155,13 @@ func getLocalStorageClassNames(ctx context.Context, kubeClient client.Client, st
 
 	// for day2 storageclasses
 	storageClassesInCluster := &storagev1.StorageClassList{}
-	if err := kubeClient.List(ctx, storageClassesInCluster); err != nil {
+	if err := kubeClient.List(ctx, storageClassesInCluster, &client.MatchingLabelsSelector{
+		// not select storageclass with external labels
+		Selector: getExternalClassesBlaclistSelector(),
+	}); err != nil {
 		return nil, err
 	}
 	for idx := range storageClassesInCluster.Items {
-		// TODO: skip storageclasses that are from external mode if both internal & external mode is enabled
 		sc := &storageClassesInCluster.Items[idx]
 		if slices.Contains(supportedCsiDrivers, sc.Provisioner) {
 			storageClassNames[sc.Name] = true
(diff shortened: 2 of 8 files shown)


services/provider/server/server.go
```diff
@@ -77,6 +77,7 @@ const (
 	monSecret                        = "rook-ceph-mon"
 	volumeReplicationClass5mSchedule = "5m"
 	mirroringTokenKey                = "rbdMirrorBootstrapPeerSecretName"
+	ExternalClassLabelKey            = "storageclass.kubernetes.io/is-external"
```
rewantsoni, line 80: Let's move it to storagecluster/utils?

a.yaml
rewantsoni, line 1: created by mistake?

rewantsoni (changes requested):

services/provider/server/server.go
parth-gr: i cant see such file, can you be specific on which file among this?
https://github.com/red-hat-storage/ocs-operator/tree/main/controllers/util

rewantsoni: https://github.com/red-hat-storage/ocs-operator/blob/main/controllers/util/k8sutil.go
Should be a good place

controllers/storagecluster/external_resources.go
```diff
@@ -552,7 +527,7 @@ func (r *StorageClusterReconciler) createExternalModeStorageClasses(sccs []Stora
 			if existing.DeletionTimestamp != nil {
 				return fmt.Errorf("failed to restore StorageClass  %s because it is marked for deletion", existing.Name)
 			}
-			if !reflect.DeepEqual(sc.Parameters, existing.Parameters) {
+			if !reflect.DeepEqual(sc.Parameters, existing.Parameters) || !reflect.DeepEqual(sc.Labels, existing.Labels) {
```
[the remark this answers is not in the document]
parth-gr: Maybe I should add to the current list of labels at line 435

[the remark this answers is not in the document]
rewantsoni: But in that case, if the user removes the label, there is no check to validate and add it back, right?

controllers/storagecluster/storageconsumer.go
```diff
@@ -144,9 +145,8 @@ func getLocalStorageClassNames(ctx context.Context, kubeClient client.Client, st
 		return nil, err
 	}
 	for idx := range storageClassesInCluster.Items {
-		// TODO: skip storageclasses that are from external mode if both internal & external mode is enabled
 		sc := &storageClassesInCluster.Items[idx]
-		if slices.Contains(supportedCsiDrivers, sc.Provisioner) {
+		if slices.Contains(supportedCsiDrivers, sc.Provisioner) && sc.Labels[server.ExternalClassLabelKey] != "true" {
```
nb-ohad, line 149: There is no point in filtering on a able when you can just list to exclude anything with the label

controllers/storagecluster/volumegroupsnapshotterclasses.go
```diff
@@ -111,6 +112,12 @@ func (obj *ocsGroupSnapshotClass) ensureCreated(r *StorageClusterReconciler, ins
 		cephFsGroupSnapshotClass,
 	}
 
+	if instance.Spec.ExternalStorage.Enable {
```
nb-ohad, line 115: Why is there a need for a check. I believe that this code only runs in external mode

controllers/storagecluster/volumesnapshotterclasses.go
```diff
@@ -143,6 +144,12 @@ func (obj *ocsSnapshotClass) ensureCreated(r *StorageClusterReconciler, instance
 	cephFsSnapClass.snapshotClass.Name = util.GenerateNameForSnapshotClass(instance.Name, util.CephfsSnapshotter)
 	volumeSnapshotClasses := []SnapshotClassConfiguration{rbdSnapClass, cephFsSnapClass}
 
+	if instance.Spec.ExternalStorage.Enable {
```
nb-ohad, line 147: Why is there a need for a check. I believe that this code only runs in external mode

services/provider/server/server.go
```diff
@@ -1834,6 +1835,8 @@ func (s *OCSProviderServer) appendStorageClassKubeResources(
 			klog.Warningf("Encountered unsupported provisioner in storage class %s", storageClassName)
 		} else if storageClass == nil {
 			klog.Warningf("The name %s does not points to a builtin or an existing storage class, skipping", storageClassName)
+		} else if storageClass.Labels[ExternalClassLabelKey] == "true" {
```
nb-ohad, line 1838:
```suggestion
		} else if strconv.ParseBool(storageClass.Labels[ExternalClassLabelKey]) {
```

parth-gr: it returns `value of type (bool, error)`

controllers/storagecluster/storageconsumer.go
```diff
@@ -140,11 +142,15 @@ func getLocalStorageClassNames(ctx context.Context, kubeClient client.Client, st
 
 	// for day2 storageclasses
 	storageClassesInCluster := &storagev1.StorageClassList{}
-	if err := kubeClient.List(ctx, storageClassesInCluster); err != nil {
+	if err := kubeClient.List(ctx, storageClassesInCluster, &client.ListOptions{
+		// not select storageclass with external labels
+		LabelSelector: labels.SelectorFromSet(map[string]string{
```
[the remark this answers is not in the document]
rewantsoni: This is what I have used before to filter out blockpools that doesn't have the mirroring label
```
	blackListRequirement, err := labels.NewRequirement("ocs.openshift.io/forbid-mirroring", selection.NotEquals, []string{"true"})
	if err != nil {
		fmt.Println(err)
		return
	}

	selector = selector.Add(*blackListRequirement)

	blockpools := &rookCephv1.CephBlockPoolList{}
	if err := cl.List(context.Background(), podList,
		client.InNamespace(namespace),
		client.MatchingLabelsSelector{Selector: selector}); err != nil {
		fmt.Println(err)
		return
	}

	for _, pod := range blockpools.Items {
		fmt.Println(pod.Name)
	}
```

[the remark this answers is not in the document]
parth-gr: thanks!

controllers/storagecluster/external_resources.go
```diff
@@ -552,7 +523,7 @@ func (r *StorageClusterReconciler) createExternalModeStorageClasses(sccs []Stora
 			if existing.DeletionTimestamp != nil {
 				return fmt.Errorf("failed to restore StorageClass  %s because it is marked for deletion", existing.Name)
 			}
-			if !reflect.DeepEqual(sc.Parameters, existing.Parameters) {
+			if !reflect.DeepEqual(sc.Parameters, existing.Parameters) || existing.Labels[util.ExternalClassLabelKey] == "" {
```
rewantsoni, line 526: shouldn't we check if the existing label value is != what we desire?

controllers/storagecluster/storageconsumer.go
```diff
@@ -109,6 +111,16 @@ func (s *storageConsumer) ensureDeleted(_ *StorageClusterReconciler, _ *ocsv1.St
 	return ctrl.Result{}, nil
 }
 
+func getExternalStorageClassBlaclistSelector() labels.Selector {
```
rewantsoni, line 114: This is used for snapshotClasses and groupSnapshotClasses as well. Can we use a better name for this?

controllers/storagecluster/storageconsumer.go
```diff
@@ -109,6 +111,16 @@ func (s *storageConsumer) ensureDeleted(_ *StorageClusterReconciler, _ *ocsv1.St
...
+	blackListRequirement, err := labels.NewRequirement(util.ExternalClassLabelKey, selection.NotEquals, []string{"true"})
+	if err != nil {
+		fmt.Println(err)
```
rewantsoni, line 117: If you want to print, use a log instead of fmt

controllers/storagecluster/storageconsumer.go
```diff
@@ -173,7 +187,10 @@ func getLocalVolumeSnapshotClassNames(ctx context.Context, kubeClient client.Cli
 
 	// for day2 volumesnapshotclasses
 	volumeSnapshotClassesInCluster := &snapapi.VolumeSnapshotClassList{}
-	if err := kubeClient.List(ctx, volumeSnapshotClassesInCluster); err != nil {
+	if err := kubeClient.List(ctx, volumeSnapshotClassesInCluster, &client.MatchingLabelsSelector{
+		// not select storageclass with external labels
```
rewantsoni, line 191:
```suggestion
		// not select snapshotclasses with external labels
```

controllers/storagecluster/storageconsumer.go
```diff
@@ -207,7 +224,10 @@ func getLocalVolumeGroupSnapshotClassNames(ctx context.Context, kubeClient clien
 	if crd.UID != "" {
 		// for day2 volumegroupsnapshotclasses
 		volumeGroupSnapshotClassesInCluster := &groupsnapapi.VolumeGroupSnapshotClassList{}
-		if err := kubeClient.List(ctx, volumeGroupSnapshotClassesInCluster); err != nil {
+		if err := kubeClient.List(ctx, volumeGroupSnapshotClassesInCluster, &client.MatchingLabelsSelector{
+			// not select storageclass with external labels
```
rewantsoni, line 228:
```suggestion
			// not select groupsnapshotclasses with external labels
```

controllers/util/k8sutil.go
```diff
@@ -71,6 +71,7 @@ const (
 	CephRBDMirrorName                    = "cephrbdmirror"
 	OcsClientTimeout                     = 10 * time.Second
 	StorageClientMappingConfigName       = "storage-client-mapping"
+	ExternalClassLabelKey                = "storageclass.kubernetes.io/is-external"
```
rewantsoni, line 74: instead of kubernetes.io can we use ocs.openshift.io? maybe something like `ocs.openshift.io/external-ceph-class` or if you can think of something better

services/provider/server/server.go
```diff
@@ -1524,6 +1526,8 @@ func (s *OCSProviderServer) appendVolumeSnapshotClassKubeResources(
 			klog.Warningf("Encountered unsupported driver in volume snapshot class %s", snapshotClassName)
 		} else if snapshotClass == nil {
 			klog.Warningf("The name %s does not points to a builtin or an existing volumesnapshot class, skipping", snapshotClassName)
+		} else if snapshotClass.Labels[util.ExternalClassLabelKey] == "true" {
+			klog.Warningf("The snapshot storage class %s is an external storage class, skipping", snapshotClassName)
```
rewantsoni, line 1530:
```suggestion
			klog.Warningf("The snapshot class %s is an external class, skipping", snapshotClassName)
```

services/provider/server/server.go
```diff
@@ -1589,6 +1593,8 @@ func (s *OCSProviderServer) appendVolumeGroupSnapshotClassKubeResources(
 			klog.Warningf("Encountered unsupported driver in volume group snapshot class %s", groupSnapshotClassName)
 		} else if groupSnapshotClass == nil {
 			klog.Warningf("The name %s does not points to a builtin or an existing volume group snapshot class, skipping", groupSnapshotClassName)
+		} else if groupSnapshotClass.Labels[util.ExternalClassLabelKey] == "true" {
+			klog.Warningf("The groupSnapshot storage class %s is an external storage class, skipping", groupSnapshotClassName)
```
rewantsoni, line 1597:
```suggestion
			klog.Warningf("The groupSnapshot class %s is an external class, skipping", groupSnapshotClassName)
```

rewantsoni (changes requested):

controllers/storagecluster/external_resources.go
parth-gr: We desire to update the storageclass when the existing CR doesn't have the desired external labels,
So if `existing.Labels[util.ExternalClassLabelKey] == ""` means we need to update it

controllers/storagecluster/storageconsumer.go
parth-gr: getExternalStorageClassesBlaclistSelector? or getExternalClassesBlaclistSelector?

controllers/storagecluster/storageconsumer.go
parth-gr: I think we need to error it, use fmt.error

controllers/util/k8sutil.go
parth-gr: how about storageclass.ocs.openshift.io/is-external?

controllers/storagecluster/external_resources.go
rewantsoni: What if the customer updates the label in external storageClass to existing.Labels[util.ExternalClassLabelKey] = "false"? This code will see that we have a label and it won't update the StorageClass and consumer controller will pick this up for distribution as it won't be filtered out

parth-gr: If the customer update it manually, code will again set it to `true`.

nb-ohad, line 526: Why are you checking for an empty value? What if it was not an empty value in the prev reconcile?
I think you need to check the diff in value between the actual state (existing) and desired state (sc)

controllers/storagecluster/storageconsumer.go
```diff
@@ -109,7 +112,17 @@ func (s *storageConsumer) ensureDeleted(_ *StorageClusterReconciler, _ *ocsv1.St
 	return ctrl.Result{}, nil
 }
 
-func getLocalStorageClassNames(ctx context.Context, kubeClient client.Client, storageCluster *ocsv1.StorageCluster) (
+func getExternalClassesBlaclistSelector(log logr.Logger) labels.Selector {
+	blackListRequirement, err := labels.NewRequirement(util.ExternalClassLabelKey, selection.NotEquals, []string{"true"})
+	if err != nil {
+		log.Error(err, "failed to create label requirement for external classes")
```
nb-ohad, line 118: - Error shouldn't be logged without returning them and effecting the result of the reconcile.
- What kind of error can even be returned here? wouldn't it be safe to ignore it all of the requirement information is hardcoded?

controllers/storagecluster/storageconsumer.go
```diff
@@ -109,7 +112,17 @@ func (s *storageConsumer) ensureDeleted(_ *StorageClusterReconciler, _ *ocsv1.St
 	return ctrl.Result{}, nil
 }
 
-func getLocalStorageClassNames(ctx context.Context, kubeClient client.Client, storageCluster *ocsv1.StorageCluster) (
+func getExternalClassesBlaclistSelector(log logr.Logger) labels.Selector {
+	blackListRequirement, err := labels.NewRequirement(util.ExternalClassLabelKey, selection.NotEquals, []string{"true"})
```
nb-ohad, line 116:
```suggestion
	blackListRequirement, err := labels.NewRequirement(
		util.ExternalClassLabelKey, 
		selection.NotEquals, 
		[]string{strconv.FormatBool(true)},
	)
```

nb-ohad (changes requested):

controllers/storagecluster/storageconsumer.go
parth-gr: Made it info

parth-gr: added a panic as discussed offline, as it will be always a static error

nb-ohad (changes requested):

PR #3223, opened 2025-05-13
server: FullFillStorageClaim returns empty response and no error

rohan47:
Instead of returning nil and error, `FullFillStorageClaim` will return empty `FullFillStorageClaimResponse` and no error

### DIFF
--- a/services/provider/server/server.go
+++ b/services/provider/server/server.go
@@ -573,7 +573,10 @@ func decodeAndValidateTicket(ticket string, pubKey *rsa.PublicKey) (*services.On
 // FulfillStorageClaim RPC call to create the StorageClaim CR on
 // provider cluster.
 func (s *OCSProviderServer) FulfillStorageClaim(ctx context.Context, req *pb.FulfillStorageClaimRequest) (*pb.FulfillStorageClaimResponse, error) {
-	return nil, status.Error(codes.Unimplemented, "not implemented")
+	// returning empty response in case a upgrade scenario where provider is upgraded to 4.19 and an onboarded 4.18 client is calling FulFillStorageClaim,
+	// the client storageClaim reconciler will get stuck in FulFillStorageClaim if this function is returning an error,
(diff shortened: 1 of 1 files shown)


leelavg:
pls update the description around `why` we are doing it and `what/how` should be covered by code comments & code resp.

services/provider/server/server.go
```diff
@@ -566,7 +566,7 @@ func decodeAndValidateTicket(ticket string, pubKey *rsa.PublicKey) (*services.On
 // FulfillStorageClaim RPC call to create the StorageClaim CR on
 // provider cluster.
 func (s *OCSProviderServer) FulfillStorageClaim(ctx context.Context, req *pb.FulfillStorageClaimRequest) (*pb.FulfillStorageClaimResponse, error) {
-	return nil, status.Error(codes.Unimplemented, "not implemented")
+	return &pb.FulfillStorageClaimResponse{}, nil
```
nb-ohad, line 569: Please add a comment that explain why we are not returning an error

nb-ohad (changes requested):

services/provider/server/server.go
```diff
@@ -573,7 +573,8 @@ func decodeAndValidateTicket(ticket string, pubKey *rsa.PublicKey) (*services.On
 // FulfillStorageClaim RPC call to create the StorageClaim CR on
 // provider cluster.
 func (s *OCSProviderServer) FulfillStorageClaim(ctx context.Context, req *pb.FulfillStorageClaimRequest) (*pb.FulfillStorageClaimResponse, error) {
-	return nil, status.Error(codes.Unimplemented, "not implemented")
+	// returning empty response else the old client will always error out
```
nb-ohad, line 576: This Comment does not say anything meaningful. 
- What is the "old client"?
- How would it error out? What would be the behaviour? What would be the end result?

nb-ohad (changes requested):

PR #3226, opened 2025-05-14
controllers: implement consumer deletion

rewantsoni:
The PR implements consumer deletion and force deletion.

For force Deletion, we add an annotation to the rook rns/svg so that they are also deleted.

Fixes: https://issues.redhat.com/browse/DFBUGS-2332

### DIFF
--- a/controllers/storageconsumer/storageconsumer_controller.go
+++ b/controllers/storageconsumer/storageconsumer_controller.go
@@ -24,6 +24,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"slices"
+	"strconv"
 	"strings"
 	"time"
 
@@ -40,12 +41,14 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/klog/v2"
 	"k8s.io/utils/ptr"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
@@ -58,6 +61,7 @@ const (
 	StorageProfileLabel           = "ocs.openshift.io/storageprofile"
 	ConsumerUUIDLabel             = "ocs.openshift.io/storageconsumer-uuid"
 	StorageConsumerNameLabel      = "ocs.openshift.io/storageconsumer-name"
+	storageConsumerFinalizer      = "ocs.openshift.io/storageconsumer-protection"
 )
 
 // StorageConsumerReconciler reconciles a StorageConsumer object
@@ -132,6 +136,11 @@ func (r *StorageConsumerReconciler) Reconcile(ctx context.Context, request recon
 }
 
 func (r *StorageConsumerReconciler) reconcilePhases() (reconcile.Result, error) {
+	if _, notAdjusted := r.storageConsumer.GetAnnotations()[util.TicketAnnotation]; notAdjusted {
+		r.Log.Info("waiting for StorageConsumer to be adjusted for 4.19")
+		return reconcile.Result{}, nil
+	}
+
 	r.storageConsumer.Status.CephResources = []*ocsv1alpha1.CephResourcesSpec{}
 	if r.storageConsumer.Spec.Enable {
 		return r.reconcileEnabledPhases()
@@ -153,7 +162,25 @@ func (r *StorageConsumerReconciler) reconcileNotEnabledPhases() (reconcile.Resul
 }
 
 func (r *StorageConsumerReconciler) reconcileEnabledPhases() (reconcile.Result, error) {
+
+	consumerConfigMap := &corev1.ConfigMap{}
+	consumerConfigMap.Namespace = r.namespace
+	consumerConfigMap.Name = cmp.Or(
+		r.storageConsumer.Spec.ResourceNameMappingConfigMap.Name,
+		fmt.Sprintf("storageconsumer-%v", util.FnvHash(r.storageConsumer.Name)),
+	)
+	if err := r.get(consumerConfigMap); client.IgnoreNotFound(err) != nil {
+		return reconcile.Result{}, err
+	}
+
 	if r.storageConsumer.GetDeletionTimestamp().IsZero() {
+		if controllerutil.AddFinalizer(r.storageConsumer, storageConsumerFinalizer) {
+			r.Log.Info("Finalizer not found for StorageConsumer. Adding finalizer.")
+			if err := r.Client.Update(r.ctx, r.storageConsumer); err != nil {
+				return reconcile.Result{}, fmt.Errorf("failed to update StorageConsumer: %v", err)
+			}
+		}
+
 		r.storageConsumer.Status.State = v1alpha1.StorageConsumerStateConfiguring
 		if err := r.deleteOnboardingSecret(); err != nil {
 			return reconcile.Result{}, nil
@@ -164,30 +191,12 @@ func (r *StorageConsumerReconciler) reconcileEnabledPhases() (reconcile.Result, 
 			return reconcile.Result{}, err
 		}
 
-		if _, notAdjusted := r.storageConsumer.GetAnnotations()[util.TicketAnnotation]; notAdjusted {
-			r.Log.Error(
-				fmt.Errorf("upgraded 4.18 StorageConsumer"),
-				"waiting for StorageConsumer to be adjusted for 4.19",
-			)
-			return reconcile.Result{}, nil
-		}
-
 		availableServices, err := util.GetAvailableServices(r.ctx, r.Client, storageCluster)
 		if err != nil {
 			return reconcile.Result{}, err
 		}
 
-		if err := r.reconcileConsumerConfigMap(availableServices); err != nil {
-			return reconcile.Result{}, err
-		}
-
-		consumerConfigMap := &corev1.ConfigMap{}
-		consumerConfigMap.Namespace = r.namespace
-		consumerConfigMap.Name = cmp.Or(
-			r.storageConsumer.Spec.ResourceNameMappingConfigMap.Name,
-			fmt.Sprintf("storageconsumer-%v", util.FnvHash(r.storageConsumer.Name)),
-		)
-		if err := r.get(consumerConfigMap); client.IgnoreNotFound(err) != nil {
+		if err := r.reconcileConsumerConfigMap(availableServices, consumerConfigMap); err != nil {
 			return reconcile.Result{}, err
 		}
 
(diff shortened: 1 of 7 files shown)


controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -303,7 +311,7 @@ func (r *StorageConsumerReconciler) reconcileEnabledPhases() (reconcile.Result,
 		}
 
 	} else {
-		r.storageConsumer.Status.State = v1alpha1.StorageConsumerStateDeleting
+		return r.deletionPhase()
```
nb-ohad, line 314: The fact that the client was marked for deletion does not mean we can ignore reconciles going forward. You need to know if the resource is allowed to be deleted or not before deciding if to reconcile or to delete

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -721,6 +729,95 @@ func (r *StorageConsumerReconciler) reconcileNoobaaAccount() error {
...
 
+func (r *StorageConsumerReconciler) deletionPhase() (reconcile.Result, error) {
+
+	consumerConfigMap := &corev1.ConfigMap{}
+	consumerConfigMap.Namespace = r.namespace
+	consumerConfigMap.Name = cmp.Or(
+		r.storageConsumer.Spec.ResourceNameMappingConfigMap.Name,
+		fmt.Sprintf("storageconsumer-%v", util.FnvHash(r.storageConsumer.Name)),
+	)
+	if err := r.get(consumerConfigMap); client.IgnoreNotFound(err) != nil {
```
nb-ohad, line 740: How can you continue if you didn't find the configmap? Also, what does it mean in relation to deletion if we cannot find a config map?

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -721,6 +729,95 @@ func (r *StorageConsumerReconciler) reconcileNoobaaAccount() error {
...
+
+	// Get config map's controller reference
+	controllerIndex := slices.IndexFunc(
+		consumerConfigMap.OwnerReferences,
+		func(ref metav1.OwnerReference) bool { return ptr.Deref(ref.Controller, false) },
+	)
+	var controllerRef *metav1.OwnerReference
+	if controllerIndex != -1 {
+		controllerRef = &consumerConfigMap.OwnerReferences[controllerIndex]
+	}
```
nb-ohad, line 752: If the configmap has no controller ref, how does that effect the deletion process? should we continue with the procedure? should we assume we are the controller? should we fail the reconcile?

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -721,6 +729,95 @@ func (r *StorageConsumerReconciler) reconcileNoobaaAccount() error {
...
+						)
+					}
+					rns.Namespace = r.namespace
+					if err := r.get(rns); client.IgnoreNotFound(err) != nil {
+						return reconcile.Result{}, fmt.Errorf("failed to get CephBlockPoolRadosNamespace: %v", err)
+					} else if rns.UID != "" && util.AddAnnotation(rns, util.RookForceDeletionAnnotationKey, "true") {
+						if err := r.Client.Update(r.ctx, rns); err != nil {
+							return reconcile.Result{}, fmt.Errorf("failed to annotate CephBlockPoolRadosNamespace: %v", err)
+						}
+					}
```
nb-ohad, line 788: What is the point in having this here and not in normal reconciliation?

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -721,6 +729,95 @@ func (r *StorageConsumerReconciler) reconcileNoobaaAccount() error {
...
+				svg := &rookCephv1.CephFilesystemSubVolumeGroup{}
+				svg.Name = consumerResources.GetSubVolumeGroupName()
+				svg.Namespace = r.namespace
+				if err := r.get(svg); client.IgnoreNotFound(err) != nil {
+					return reconcile.Result{}, fmt.Errorf("failed to get CephFileSystemSubVolumeGroup: %v", err)
+				} else if svg.UID != "" && util.AddAnnotation(svg, util.RookForceDeletionAnnotationKey, "true") {
+					if err := r.Client.Update(r.ctx, svg); err != nil {
+						return reconcile.Result{}, fmt.Errorf("failed to annotate CephFileSystemSubVolumeGroup: %v", err)
+					}
+				}
```
nb-ohad, line 800: What is the point in having this here and not in normal reconciliation?

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -721,6 +729,95 @@ func (r *StorageConsumerReconciler) reconcileNoobaaAccount() error {
 	return nil
 }
 
+func (r *StorageConsumerReconciler) deletionPhase() (reconcile.Result, error) {
```
nb-ohad, line 732: The structure of the function does not make a lot of sense. 
A better approach would be to have a set of checks for allowing the deletion as a separate function which then is invoked outside of this function to decide if we need to delete or to reconcile. And this function should assume safe for deletion

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -768,8 +865,13 @@ func (r *StorageConsumerReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		},
 	)
 
+	consumerPredicate := util.ComposePredicates(
+		predicate.GenerationChangedPredicate{},
+		predicate.AnnotationChangedPredicate{},
+	)
```
nb-ohad, line 871: This is used once, can you please inline?

nb-ohad (changes requested):

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -306,8 +323,19 @@ func (r *StorageConsumerReconciler) reconcileEnabledPhases() (reconcile.Result,
...
 
-	} else {
+	}
+
+	if markedForDeletion {
 		r.storageConsumer.Status.State = v1alpha1.StorageConsumerStateDeleting
+		if r.storageConsumer.Status.Client == nil || hasForceDeleteAnnotation {
+			if controllerutil.RemoveFinalizer(r.storageConsumer, storageConsumerFinalizer) {
+				r.Log.Info("removing finalizer from StorageConsumer.")
+				if err := r.Client.Update(r.ctx, r.storageConsumer); err != nil {
```
nb-ohad, line 333: This client update will also update the state. Is that what we want to do?

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -306,8 +323,19 @@ func (r *StorageConsumerReconciler) reconcileEnabledPhases() (reconcile.Result,
 			r.storageConsumer.Status.State = v1alpha1.StorageConsumerStateReady
 		}
 
-	} else {
+	}
+
+	if markedForDeletion {
 		r.storageConsumer.Status.State = v1alpha1.StorageConsumerStateDeleting
+		if r.storageConsumer.Status.Client == nil || hasForceDeleteAnnotation {
```
nb-ohad, line 330: Why do we care about the client's status at this point?

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -479,6 +508,9 @@ func (r *StorageConsumerReconciler) reconcileCephRadosNamespace(
 				}
 				rns.Spec.Name = radosNamespaceName
 				rns.Spec.BlockPoolName = bp.Name
+				if forceDelete {
+					util.AddAnnotation(rns, util.RookForceDeletionAnnotationKey, strconv.FormatBool(true))
+				}
```
nb-ohad, line 513: Please remove this directly below the owner ref because this is metadata and not spec

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -437,6 +465,7 @@ func (r *StorageConsumerReconciler) reconcileCephRadosNamespace(
 	radosNamespaceName string,
 	additionalOwner client.Object,
 	builtinBlockPools []string,
+	forceDelete bool,
```
nb-ohad, line 468: Params should describe desired state, as such the nam of the param should be something like: 
```suggestion
	hasForceDeleteAnnotaion bool,
```

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -495,6 +527,7 @@ func (r *StorageConsumerReconciler) reconcileCephFilesystemSubVolumeGroup(
 	cephFileSystemName string,
 	subVolumeGroupName string,
 	additionalOwner client.Object,
+	forceDelete bool,
```
nb-ohad, line 530: Params should describe desired state, as such the nam of the param should be something like: 
```suggestion
	hasForceDeleteAnnotaion bool,
```

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -516,6 +549,9 @@ func (r *StorageConsumerReconciler) reconcileCephFilesystemSubVolumeGroup(
 		}
 		svg.Spec.FilesystemName = cephFs.Name
 		svg.Spec.Name = subVolumeGroupName
+		if forceDelete {
+			util.AddAnnotation(svg, util.RookForceDeletionAnnotationKey, strconv.FormatBool(true))
+		}
```
nb-ohad, line 554: Please remove this directly below the owner ref because this is metadata and not spec

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -777,8 +813,34 @@ func (r *StorageConsumerReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+			objNew := e.ObjectNew.(*ocsv1alpha1.StorageConsumer)
+			return objOld.Status.Client != nil && objNew.Status.Client == nil
+		},
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			return false
+		},
+		GenericFunc: func(e event.GenericEvent) bool {
+			return false
+		},
+	}
```
nb-ohad, line 831: Why do we care about this? This is the storage consumer controller you should be acting to ant storage consumer change

nb-ohad (changes requested):

controllers/storageconsumer/storageconsumer_controller.go
rewantsoni: If the status is not nil that means that we have a client onboarded and we shouldn't be deleting the consumer until we have the force delete annotation

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -308,6 +316,58 @@ func (r *StorageConsumerReconciler) reconcileEnabledPhases() (reconcile.Result,
 
 	} else {
 		r.storageConsumer.Status.State = v1alpha1.StorageConsumerStateDeleting
+		_, hasForceDeleteAnnotation := r.storageConsumer.GetAnnotations()[util.ForceDeletionAnnotationKey]
+
+		if len(consumerConfigMap.OwnerReferences) <= 1 && hasForceDeleteAnnotation {
+			consumerResources := util.WrapStorageConsumerResourceMap(consumerConfigMap.Data)
```
nb-ohad, line 322: Please move this line to just above where you start using the return value

controllers/storageconsumer/storageconsumer_controller.go
nb-ohad: That does not make sense at all, the storage consumer controller should react to any change to the storage consumer.  I think the reason you are not getting notification is because you applied `GeneraionChangePredicate` which is not something that make sense if you want to watch for status changes

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -308,6 +316,60 @@ func (r *StorageConsumerReconciler) reconcileEnabledPhases() (reconcile.Result,
...
+		if consumerOwners <= 1 && hasForceDeleteAnnotation {
+			blockPools := &rookCephv1.CephBlockPoolList{}
+			if err := r.List(r.ctx, blockPools, client.InNamespace(r.namespace)); err != nil {
+				return reconcile.Result{}, err
+			}
+			consumerResources := util.WrapStorageConsumerResourceMap(consumerConfigMap.Data)
+			radosNamespaceName := consumerResources.GetRbdRadosNamespaceName()
+
+			patchPayload := fmt.Sprintf(`{"metadata":{"annotations":{"%s":"true"}}}`, util.RookForceDeletionAnnotationKey)
+			annotationPatch := client.RawPatch(types.MergePatchType, []byte(patchPayload))
```
nb-ohad, line 337: I would prefer to not use hardcoded json payload. You can do something like the following which is typed and when coupled with panic, can guarantee a well formatted result
```	
annotationPatch := client.RawPatch(types.MergePatchType,  util.JsonMustMarshal(
	map[string]any{
		"metadata": map[string]any{
			"annotations": map[string]string{
				util.RookForceDeletionAnnotationKey: strconv.FormatBool(true),
			},
		},
	}),
)
``` 

where `JsonMustMarshal` is a wrapper around `json.Marshal` with a panic for error handling. I beleive we already implemented such a util somewhere

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -308,6 +316,60 @@ func (r *StorageConsumerReconciler) reconcileEnabledPhases() (reconcile.Result,
...
+
+			for idx := range blockPools.Items {
+				bp := &blockPools.Items[idx]
+				rns := &rookCephv1.CephBlockPoolRadosNamespace{}
+				rns.Name = fmt.Sprintf("%s-%s", bp.Name, radosNamespaceName)
+				if radosNamespaceName == util.ImplicitRbdRadosNamespaceName {
+					rns.Name = fmt.Sprintf(
+						"%s-builtin-%s",
+						bp.Name,
+						util.ImplicitRbdRadosNamespaceName[1:len(util.ImplicitRbdRadosNamespaceName)-1],
```
nb-ohad, line 347: Why aren't we reading the name from the configmap. What is this `buildin` prefix doing here?

nb-ohad (changes requested):

controllers/storageconsumer/storageconsumer_controller.go
rewantsoni: This is the special case for the internal client, where the radosnamespacename is <implicit>
The CR name cannot have '<' '>' hence the name for the implicit raodsnamespace for each blockpool is blockpoolname-builtin-implicit

This is what we were doing while creating the radosnamespace https://github.com/red-hat-storage/ocs-operator/blob/main/controllers/storageconsumer/storageconsumer_controller.go#L454-L463

nb-ohad: This implementations defeats the purpose of comment you were referring to but I do agree that a change for this need to be provided in a different PR. Resolving

PR #3227, opened 2025-05-14
Send OCS sub channel only if the current CSV of the sub has succeeded

malayparida2000:
When the channel of the ocs-operator subscription is changed (e.g., from stable-4.18 to stable-4.19), this update is immediately propagated to the client, triggering its upgrade. However, if the client upgrades before the provider has completed its own CSV upgrade, the provider may detect that the client is ahead and mark itself as NotUpgradeable, blocking the upgrade process.

This commit ensures that the updated channel information is only sent to the client after the provider's CSV has reached the Succeeded phase. This prevents premature client upgrades and avoids upgrade deadlocks caused by version skew between provider and client.

Ref- https://issues.redhat.com/browse/DFBUGS-2502

### DIFF
--- a/services/provider/server/server.go
+++ b/services/provider/server/server.go
@@ -679,7 +679,22 @@ func (s *OCSProviderServer) getOCSSubscriptionChannel(ctx context.Context) (stri
 		return sub.Spec.Package == "ocs-operator"
 	})
 	if subscription == nil {
-		return "", fmt.Errorf("unable to find ocs-operator subscription")
+		klog.Info("unable to find ocs-operator subscription")
+		return "", nil
+	}
+	if subscription.Status.InstalledCSV == "" {
+		klog.Infof("Subscription %q does not have an installed CSV", subscription.Name)
+		return "", nil
+	}
+	csv := &opv1a1.ClusterServiceVersion{}
+	err = s.client.Get(ctx, client.ObjectKey{Name: subscription.Status.InstalledCSV, Namespace: s.namespace}, csv)
+	if err != nil {
+		klog.Errorf("failed to get installed CSV %q: %v", subscription.Status.InstalledCSV, err)
+		return "", err
+	}
+	if csv.Status.Phase != opv1a1.CSVPhaseSucceeded {
+		klog.Infof("installed CSV %q is in phase %q, not Succeeded", csv.Name, csv.Status.Phase)
+		return "", nil
 	}
 	return subscription.Spec.Channel, nil
 }
--- a/deploy/ocs-operator/manifests/provider-role.yaml
+++ b/deploy/ocs-operator/manifests/provider-role.yaml
@@ -52,6 +52,7 @@ rules:
       - operators.coreos.com
     resources:
       - subscriptions
+      - clusterserviceversions
     verbs:
       - get
       - list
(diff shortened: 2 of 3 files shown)


malayparida2000:
The function getOCSSubscriptionChannel is used by 3 function RPC Calls
GetStorageConfig, GetDesiredClientState, ReportStatus

GetDesiredClientState() https://github.com/red-hat-storage/ocs-client-operator/blob/7e2024d5fbb763c9fd878e4e8497c6de92f96b75/internal/controller/storageclient_controller.go#L364-L367

Simply returns a reconcile error, 

GetStorageConfig() https://github.com/red-hat-storage/ocs-client-operator/blob/7e2024d5fbb763c9fd878e4e8497c6de92f96b75/service/status-report/main.go#L145-L148

ReportStatus() https://github.com/red-hat-storage/ocs-client-operator/blob/7e2024d5fbb763c9fd878e4e8497c6de92f96b75/service/status-report/main.go#L124-L127

services/provider/server/server.go
leelavg, line 1: can you make a change in `ReportStatus` as not to send the err returned by this func to client and just log the error?

services/provider/server/server.go
```diff
@@ -681,6 +681,17 @@ func (s *OCSProviderServer) getOCSSubscriptionChannel(ctx context.Context) (stri
...
 	}
+	if subscription.Status.InstalledCSV == "" {
+		return "", fmt.Errorf("subscription %s has no installedCSV", subscription.Name)
+	}
+	csv := &opv1a1.ClusterServiceVersion{}
+	err = s.client.Get(ctx, client.ObjectKey{Name: subscription.Status.InstalledCSV, Namespace: s.namespace}, csv)
+	if err != nil {
+		return "", fmt.Errorf("failed to get installed CSV %q: %w", subscription.Status.InstalledCSV, err)
+	}
+	if csv.Status.Phase != opv1a1.CSVPhaseSucceeded {
```
leelavg, line 692: a fair solution, I believe it'll be helpful not to block sending the channel if we are having issues w/ csv upgrade in patch versions.

rationale being, we aren't introducing any api change in patch versions and a failure here gets surfaced to client when the desiredstate is being processed.

malayparida2000: That will be tricky to detect.
As a customer might not always upgrade to 4.19.0.
They might for example upgrade from 4.18.4 to 4.19.2. And Sub doesn't store the previous CSV version for us to make a comparision.

services/provider/server/server.go
leelavg: ah! when I re-read I get why you took it differently, I meant not to directly send the error that got raised in here as that is the internal detail. 

so mentioned, log the internal detail and only send the error that we faile to construct the response w/o any specifics to the other side.

services/provider/server/server.go
leelavg: I mean we could just compare major & minor versions.

services/provider/server/server.go
malayparida2000: Thanks for the clarification, updated now

leelavg:
This is introducing a regression based on CI, did you test this PR?

nb-ohad (changes requested):
This change adds many more cases where we return an actual failure to the reporter job. This is undesirable, because the reporter job has other responsibilities.

Can we instead return the old value when we don't recognize a new value? or at least an "unknown" value so the reporter could resonate about its existing value?

PR #3228, opened 2025-05-14
storageautoscaler: fix the invalid state phase

parth-gr:
if we moved the cr to invalid state, there was
no way to make the phase back to no-started,
If the invalid state is not detected anymore
move the phase to not-started

### DIFF
--- a/controllers/storageautoscaler/storageautoscaler_controller.go
+++ b/controllers/storageautoscaler/storageautoscaler_controller.go
@@ -86,6 +86,17 @@ func (r *StorageAutoscalerReconciler) Reconcile(ctx context.Context, request rec
 		return reconcile.Result{}, nil
 	}
 
+	// if previously it was in invalid state, update the status to not started
+	if storageAutoScaler.Status.Phase == ocsv1.StorageAutoScalerPhaseInvalid {
+		storageAutoScaler.Status.Phase = ocsv1.StorageAutoScalerPhaseNotStarted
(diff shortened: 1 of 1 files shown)


travisn (commented):
/lgtm

parth-gr:
@travisn is it possible in the scaling is in-progress, and then someone introduce a invalid state,

1) While scaling up no invalid state detected
2) If its in progress

3) Someone introduce a invalid state:
     1) By creating a another cr : possible
      2) By updating the storagecluster with lso storageclass: less likely
      3) changing the profile to lean: not possible.
      
    So can we just ignore the 1st case??

travisn:
> Otherwise currently will move the state from `in-progress -> invalid -> not started`

If case 1 happens, then on the next reconcile it will move back to in-progress, right? It seems like a corner case that we don't need to worry about it, since the in-progress or future autoscales would not be stuck.

parth-gr:
> > Otherwise currently will move the state from `in-progress -> invalid -> not started`
> 
> If case 1 happens, then on the next reconcile it will move back to in-progress, right? It seems like a corner case that we don't need to worry about it, since the in-progress or future autoscales would not be stuck.

The only problem is on how we update the status https://github.com/red-hat-storage/ocs-operator/blob/5bd8d35a3cf5dbc642bcc42f82e3b34bf2d4bb37/controllers/storageautoscaler/storageautoscaler_controller.go#L126
we might lost the struct

PR #3235, opened 2025-05-16
DFBUGS-949: consider the worst latency between osds for system latency

weirdwiz:
currently the latency is tightly coupled with the number of osds, which leads to misleading numbers. changing it to only account the worse latency


weirdwiz:
@umangachapagain can you ptal this PR?

umangachapagain (commented):
@weirdwiz Can you run some tests and document the difference in behavior here?

yati1998 (commented):
looks good to me

weirdwiz:
Closing stale PR. Will send a fresh one if needed.

PR #3237, opened 2025-05-19
Follow the replicate size of the default CephBlockPool for the .mgr pool

malayparida2000:
In a cluster if customer has specified the replication size for the default CephBlockPool for example 2, then the .mgr pool should also have the same replication size. This will majorly help in prototyping ODF for 2 node cluster as for all other pools we have ways to specify the replication size but not for the .mgr pool.

Ref-https://issues.redhat.com/browse/DFBUGS-2568


travisn (commented):
/lgtm

malayparida2000:
@travisn this is the PR for the main branch, Once it merges we will create a cherry-pick for 4.19 & link the bug there

PR #3240, opened 2025-05-19
Revert "mgr: disable mgr liveness and startup probe"

parth-gr:
This reverts commit 5ae16668fe3d355783d5c4f60aee03c3f211ff0f.

As we have the fix of the mgr liveneess probe enabling it back https://bugzilla.redhat.com/show_bug.cgi?id=2354903


travisn:
@parth-gr Just for the record, we have confirmed we have the ceph build with the fix in the 4.19 builds? If we merge before that is confirmed, it would break the build.

PR #3242, opened 2025-05-19
Set mon_target_pg_per_osd based on the resource profile

malayparida2000:
There were concerns of additional memory load on the OSDs with the increased mon_target_pg_per_osd value of 400. To mitigate this, we are setting the mon_target_pg_per_osd value based on the resource profile. By default for the balanced profile, the value is set to 200. For the lean profile, the value is set to 100. For the performance profile the value is set to 400.

https://issues.redhat.com/browse/DFBUGS-2569


travisn (commented):
Can the user override this default? Or perhaps that could be a follow up issue if we decide it's needed.

malayparida2000:
> Can the user override this default? Or perhaps that could be a follow up issue if we decide it's needed.

It should be overiridable as the override logic block is kept after this logic block

travisn (commented):
/lgtm

PR #3253, opened 2025-05-20
replica1: fix replica1 for internal client

parth-gr:
if the replica1 is enabled from the storagecluster update the rbd csi driver with topological labels

### DIFF
--- a/controllers/ocsinitialization/ocsinitialization_controller.go
+++ b/controllers/ocsinitialization/ocsinitialization_controller.go
@@ -7,6 +7,7 @@ import (
 	"strconv"
 	"strings"
 
+	csiopv1a1 "github.com/ceph/ceph-csi-operator/api/v1alpha1"
 	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
 	"github.com/red-hat-storage/ocs-operator/v4/controllers/defaults"
 	"github.com/red-hat-storage/ocs-operator/v4/controllers/platform"
@@ -323,6 +324,16 @@ func (r *OCSInitializationReconciler) SetupWithManager(mgr ctrl.Manager) error {
 				},
 			),
 		).
+		// Watcher for csiDriver
+		Watches(
+			&csiopv1a1.Driver{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      util.RbdDriverName,
+					Namespace: r.OperatorNamespace,
+				},
+			},
+			enqueueOCSInit,
+		).
 		// Watcher for rook-ceph-operator-config cm
 		Watches(
 			&corev1.ConfigMap{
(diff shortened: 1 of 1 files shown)


controllers/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -323,6 +324,16 @@ func (r *OCSInitializationReconciler) SetupWithManager(mgr ctrl.Manager) error {
 				},
 			),
 		).
+		// Watcher for csiDriver
+		Watches(
+			&csiopv1a1.Driver{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      util.RbdDriverName,
+					Namespace: r.OperatorNamespace,
+				},
```
nb-ohad, line 333: This does not do what you want it do to. Watch uses the passed struct only for getting the GVK out of it using the schema mappings. All of the information you pass here will just get ignored and the watch will be for all Driver CRs.

What you need is a predicate that select on an object key. I think we already implements such predicate in the predicate utils file

controllers/ocsinitialization/ocsinitialization_controller.go
nb-ohad, line 1: Why are you adding the code in the initialization controller and not in the storage cluster controller where we have all of the other code that relates to setting up the client and consumer?

controllers/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -517,6 +551,14 @@ func (r *OCSInitializationReconciler) ensureOcsOperatorConfigExists(initialData
 		util.DisableCSIDriverKey:         strconv.FormatBool(true),
 	}
 
+	if util.EnableTopologyKey == "true" {
```
nb-ohad, line 554: If you want to compare to boolean as a string, please use parse bool instead of comparing to a hardcoded string value
```suggestion
	if val. err := strconv.ParseBool(util.EnableTopologyKey; err == nil && val  {
```

controllers/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -493,6 +504,29 @@ func (r *OCSInitializationReconciler) getCsiTolerations(csiTolerationKey string)
 	return tolerations
 }
 
+func patchCsiDriver(ctx context.Context, cl client.Client, log *logr.Logger, topologyDomainLabelsKey, namespace string) error {
```
nb-ohad, line 507: This is a reconcile function, please call it reconcileCsiDriver and not patchCSIDriver.
Also there is no reason to not have with the reconciler as a target

nb-ohad (changes requested):

controllers/ocsinitialization/ocsinitialization_controller.go
rewantsoni: util.EnableTopologyKey is a constant with the value of "CSI_ENABLE_TOPOLOGY" I think you need to call the function `getEnableTopologyKeyValue` or read it from the configmap data

controllers/ocsinitialization/ocsinitialization_controller.go
```diff
@@ -517,6 +551,14 @@ func (r *OCSInitializationReconciler) ensureOcsOperatorConfigExists(initialData
 		util.DisableCSIDriverKey:         strconv.FormatBool(true),
 	}
 
+	if util.EnableTopologyKey == "true" {
+		err = patchCsiDriver(r.ctx, r.Client, &r.Log, util.TopologyDomainLabelsKey, initialData.Namespace)
```
rewantsoni, line 555: util.TopologyDomainLabelsKey is again a string it doesn't contain the desired topology domain labels

rewantsoni (changes requested):

PR #3255, opened 2025-05-20
Add disable key rotation annotation only for the internal client

malayparida2000:
We are adding the disable key rotation annotation from the storageCluster to the internal storage consumer. And based on that annotation, we will add the key rotation annotation to the storageClass.

Ref-https://issues.redhat.com/browse/DFBUGS-2288

### DIFF
--- a/services/provider/server/server.go
+++ b/services/provider/server/server.go
@@ -1421,7 +1421,7 @@ func (s *OCSProviderServer) appendStorageClassKubeResources(
 					consumerConfig.GetCsiRbdNodeCephUserName(),
 					consumer.Status.Client.OperatorNamespace,
 					kmsServiceName,
-					storageCluster.GetAnnotations()[defaults.KeyRotationEnableAnnotation] == "false",
+					consumer.GetAnnotations()[defaults.KeyRotationEnableAnnotation],
 				)
 			}
 		}
--- a/controllers/storagecluster/storageconsumer.go
+++ b/controllers/storagecluster/storageconsumer.go
@@ -72,6 +72,10 @@ func (s *storageConsumer) ensureCreated(r *StorageClusterReconciler, storageClus
 
 		controllerutil.AddFinalizer(storageConsumer, internalComponentFinalizer)
 
+		if val, ok := storageCluster.GetAnnotations()[defaults.KeyRotationEnableAnnotation]; ok {
+			util.AddAnnotation(storageConsumer, defaults.KeyRotationEnableAnnotation, val)
+		}
+
 		return nil
 	}); err != nil {
 		return ctrl.Result{}, fmt.Errorf("failed to create/update storageconsumer %s: %v", storageConsumer.Name, err)
--- a/controllers/util/storageclasses.go
+++ b/controllers/util/storageclasses.go
@@ -175,7 +175,7 @@ func NewDefaultEncryptedRbdStorageClass(
 	nodeSecret,
 	namespace,
 	encryptionServiceName string,
-	disableKeyRotation bool,
+	KeyRotationAnnotationValue string,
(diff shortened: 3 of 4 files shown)


rewantsoni:
Key rotation is not a valid scenario for the clients that are external to the cluster, should we send in desired state for external clients the keyrotation annotation?

malayparida2000:
> Key rotation is not a valid scenario for the clients that are external to the cluster, should we send in desired state for external clients the keyrotation annotation?

Any explanation for that? cc @Madhu-1 ?

Madhu-1:
We had a dicussion about this one

Today the internal ODF supports the PV encryption and the key rotation and we dont have this feature for remote clients. This PR is making generic assumption and working with  both clients. Make sure we will add this annotation for the built-in RBD encryption enabled storageclass only as if the encryption is enabled it wont get distributed to the remote clients.

@nb-ohad ^^

nb-ohad:
@malayparida2000 We should change the code to set the annotation only if the returned class has already the encryption param on

rewantsoni:
LGTM

services/provider/server/server.go
```diff
@@ -1421,7 +1421,7 @@ func (s *OCSProviderServer) appendStorageClassKubeResources(
 					consumerConfig.GetCsiRbdNodeCephUserName(),
 					consumer.Status.Client.OperatorNamespace,
 					kmsServiceName,
-					storageCluster.GetAnnotations()[defaults.KeyRotationEnableAnnotation] == "false",
+					consumer.GetAnnotations()[defaults.KeyRotationEnableAnnotation] == "false",
```
nb-ohad, line 1424: why are we comparing to a false value? Is the lack of annotation means true? that does not make common sense.

nb-ohad, line 1424: If there is a need to evaluate a boolean value that is provided via text, please use `strconv.ParseBool` instead of comparing to a constant string

nb-ohad (changes requested):

services/provider/server/server.go
malayparida2000: Yes without the disable annotation by default key rotation is true on the CSI side, ref- https://github.com/red-hat-storage/ocs-operator/pull/2817

malayparida2000: We are just checking the value of an annotation here, Today it's false tomorrow might be something else might not be restricted to true & false only. So using ParseBool doesn't make sense according to me.

malayparida2000: What exactly are we gaining by calling strconv.ParseBool here? It's a bit puzzling why we’d introduce unnecessary overhead and complicate the logic when a simple string comparison achieves the same outcome. The function does precisely what a basic == "false" check already handles — no magic there. For reference: [Go source](https://cs.opensource.google/go/go/+/refs/tags/go1.23.9:src/strconv/atob.go;l=10-19).

Also, regarding consistency — the current codebase already uses direct comparisons == false rather than converting to bool and then comparing ref https://github.com/red-hat-storage/ocs-operator/blob/6dec5ed6becfd5b9297541bdcd73c0ed65034c42/controllers/storagecluster/external_resources.go#L395. So unless there’s a strong reason to deviate (which hasn’t been provided), it might be better to stick with the existing style.

malayparida2000: It's a design decision by csi not ocs operator. And as of now it's already documented in this way, If we intend to change that behaviour there has to be a separate discussion on that, that is out of scope for this bug fix.

rewantsoni:
@malayparida2000 the sanity checks are failing

malayparida2000:
> @malayparida2000 the sanity checks are failing

Added the vendor changes now

PR #3260, opened 2025-05-22
server: add rbd storageid and clientProfile name to the GetStorageClientInfo rpc

rewantsoni:
ClientProfileMapping creation requires blockpoolIDs (local and remote) and clientProfile (local and remote). We are sending the remote blockpool IDs via the getBlockPoolsInfo rpc call. For 4.18 provider mode RDR, we made an assumption as the clientProfile name was a hash of the storageClaim name. This doesn't hold true for 4.19 onwards. We need to move away from such an assumption and fetch the remote ClientProfile from the other side. Hence, adding the clientProfile name to the GetStorageClientsInfo rpc.

There is another hack that we use to construct the rbdStorageID from the mirroring token and remote radosnamesape to build the replicationID. We should stop relying on this hack a well and send the rbdstorageID in the GetStorageClientsInfo rpc.

add the rbd clientprofile to the response of GetStorageClientInfo rpc
Send the rbd storageID and clientProfile from the GetStorageClientInfo rpc
Fixes a part of https://issues.redhat.com/browse/DFBUGS-2307 , more PRs will follow for the mirroring controller and the provider server (update the ClientProfileMapping and remove the hack for peer rbd storageID)


controllers/util/storageconsumer.go
```diff
@@ -79,6 +79,20 @@ type StorageConsumerResources interface {
...
+	GetSubVolumeGroupRadosNamespaceNameKey() string
+	GetCsiRbdProvisionerCephUserNameKey() string
+	GetCsiRbdNodeCephUserNameKey() string
+	GetCsiCephFsProvisionerCephUserNameKey() string
+	GetCsiCephFsNodeCephUserNameKey() string
+	GetCsiNfsProvisionerCephUserNameKey() string
+	GetCsiNfsNodeCephUserNameKey() string
+	GetRbdClientProfileNameKey() string
+	GetCephFsClientProfileNameKey() string
+	GetNfsClientProfileNameKey() string
```
nb-ohad, line 95: Keys are constant and do not change, and the entire point of this helper struct is to make sure you don't need to deal of care about the keys. Adding "keys accessor" is against the design of this util

controllers/util/storageconsumer.go
```diff
@@ -89,6 +103,54 @@ func WrapStorageConsumerResourceMap(data map[string]string) StorageConsumerResou
...
+	return rbdClientProfileKey
+}
+
+func (wrapper storageConsumerResourceMapWrapper) GetCephFsClientProfileNameKey() string {
+	return cephFsClientProfileKey
+}
+
+func (wrapper storageConsumerResourceMapWrapper) GetNfsClientProfileNameKey() string {
+	return nfsClientProfileKey
+}
```
nb-ohad, line 152:
```suggestion
```

services/provider/server/server.go
```diff
@@ -361,9 +361,9 @@ func (s *OCSProviderServer) GetDesiredClientState(ctx context.Context, req *pb.G
 			kubeResourceBytes := mustMarshal(kubeResource)
 			response.KubeResources = append(response.KubeResources, kubeResourceBytes)
 			response.KubeObjects = append(response.KubeObjects, &pb.KubeObject{
-				Bytes: kubeResourceBytes,
+				Bytes:           kubeResourceBytes,
 				ReconcilePolicy: pb.KubeObjectReconcilePolicy_CreateOrUpdate,
-				UpdateStrategy: pb.KubeObjectUpdateStrategy_UpdateWithReplaceAsFallback,
+				UpdateStrategy:  pb.KubeObjectUpdateStrategy_UpdateWithReplaceAsFallback,
```
nb-ohad, line 366: Please don't do that. Our build system enforces code linting and formatting. If it didn't raise an error then you should not be changing this in an unrelated PR

services/provider/server/server.go
```diff
@@ -874,13 +874,31 @@ func (s *OCSProviderServer) GetStorageClientsInfo(ctx context.Context, req *pb.S
...
+			var fsid string
+			if cephCluster, err := util.GetCephClusterInNamespace(ctx, s.client, s.namespace); err != nil {
+				klog.Errorf("failed to get cephCluster in namespace %s: %v", s.namespace, err)
+				continue
+			} else if cephCluster.Status.CephStatus == nil || cephCluster.Status.CephStatus.FSID == "" {
+				klog.Infof("waiting for Ceph FSID")
+				continue
+			} else {
+				fsid = cephCluster.Status.CephStatus.FSID
+			}
```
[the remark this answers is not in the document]
rewantsoni: No, we don't have any util for this we have an util to GetCephClusterInNamespace

services/provider/server/server.go
```diff
@@ -874,13 +874,38 @@ func (s *OCSProviderServer) GetStorageClientsInfo(ctx context.Context, req *pb.S
...
+						Message:  "failed loading client information",
+					},
+				)
+				continue
+			} else if cephCluster.Status.CephStatus == nil || cephCluster.Status.CephStatus.FSID == "" {
+				klog.Infof("waiting for Ceph FSID")
+				continue
+			} else {
+				fsid = cephCluster.Status.CephStatus.FSID
+			}
```
nb-ohad, line 898: Is there a reason we are repeating this operation for each client?

services/provider/server/server.go
```diff
@@ -874,13 +874,38 @@ func (s *OCSProviderServer) GetStorageClientsInfo(ctx context.Context, req *pb.S
...
 			clientInfo.RadosNamespace = consumerConfig.GetRbdRadosNamespaceName()
+			var fsid string
+			if cephCluster, err := util.GetCephClusterInNamespace(ctx, s.client, s.namespace); err != nil {
+				klog.Errorf("failed to get cephCluster in namespace %s: %v", s.namespace, err)
+				response.Errors = append(response.Errors,
+					&pb.StorageClientInfoError{
+						ClientID: req.ClientIDs[i],
+						Code:     pb.ErrorCode_Internal,
+						Message:  "failed loading client information",
+					},
```
nb-ohad, line 890: Is there a reason to add individual error for each client? If we cannot get an FSID then we will get an error for all clients

services/provider/proto/provider.proto
```diff
@@ -318,7 +318,10 @@ message ClientInfo{
   // For each client we are going to have the same radosnamespace,
   // assumption based on existing supported feature set and plan for converged mode
   string radosNamespace = 2;
+  // rbdStorageID is the storageID for the block required for DR
   string rbdStorageID = 3;
+  // clientProfiles hold the clientProfile names
```
nb-ohad, line 323:
```suggestion
  // clientProfiles hold the CSI operator ClientProfile names utilized by the client
```

services/provider/proto/provider.proto
```diff
@@ -318,7 +318,10 @@ message ClientInfo{
   // For each client we are going to have the same radosnamespace,
   // assumption based on existing supported feature set and plan for converged mode
   string radosNamespace = 2;
+  // rbdStorageID is the storageID for the block required for DR
```
nb-ohad, line 321:
```suggestion
  // rbdStorageID is a unique identifier for the RBD storage that is used by the client
```

nb-ohad (changes requested):

nb-ohad (changes requested):

rewantsoni:
```
{  openshift cluster install failed overall} 
```
ci keeps failing with

PR #3262, opened 2025-05-22
Fix extensibility issue with GetDesiredClientState

nb-ohad:
API Server endpoint GetDesiredClientState returns the information about the KubeResources to be reconciled on the client side as a list of bytes. This prevent us from a clean future extension of the info we send per kube resource when the time comes. 

In order to fix the issue I introduced a new field called KubeObjects which is a list of objects with a single field called bytes. This new field will supersede the existing KubeResources field and will solve the problem. 

The old field was kept in place to make sure CI passes and will be removed in a future PR, after we will update the ocs-client-operator code to start using the new field


services/provider/proto/provider.proto
```diff
@@ -118,13 +118,32 @@ message StorageConfigResponse{
     SystemAttributes systemAttributes = 3;
 }
 
+// KubeObject holds information about a kube object needed in order to reconcile it 
+// but the client operator
```
rewantsoni, line 122:
```suggestion
// KubeObject holds information about a kube object needed in order to be reconciled 
// by the client operator
```

services/provider/proto/provider.proto
```diff
@@ -118,13 +118,32 @@ message StorageConfigResponse{
...
+
 // GetDesiredClientStateResponse holds the response for the GetDesiredClientState API request
 message GetDesiredClientStateResponse{
-  repeated bytes kubeResources = 1;
-  string desiredStateHash = 2;
-  string clientOperatorChannel = 3;
-  bool maintenanceMode = 4;
-  bool mirrorEnabled = 5;
+    // kubeResources holds the definition of all the kube object needed to reconcile it by 
+    // the client operator
```
rewantsoni, line 131:
```suggestion
    // kubeResources holds the definition of all the kube object needed to be reconciled by 
    // the client operator 
```

rewantsoni (commented):
The PR looks good, just 2 small nits

services/provider/server/server.go
```diff
@@ -358,7 +358,9 @@ func (s *OCSProviderServer) GetDesiredClientState(ctx context.Context, req *pb.G
 			}
 			kubeResource.GetObjectKind().SetGroupVersionKind(gvk)
 			sanitizeKubeResource(kubeResource)
-			response.KubeResources = append(response.KubeResources, mustMarshal(kubeResource))
+			bytes := mustMarshal(kubeResources)
```
leelavg, line 361: There is a possibility of `bytes` var shadowing `bytes` module.

services/provider/proto/provider.proto
nb-ohad: Did some changes to the comment

services/provider/proto/provider.proto
nb-ohad: Did some changes to the comment

services/provider/server/server.go
nb-ohad: renamed

leelavg (approved):

services/provider/server/server.go
```diff
@@ -358,7 +358,9 @@ func (s *OCSProviderServer) GetDesiredClientState(ctx context.Context, req *pb.G
 			}
 			kubeResource.GetObjectKind().SetGroupVersionKind(gvk)
 			sanitizeKubeResource(kubeResource)
-			response.KubeResources = append(response.KubeResources, mustMarshal(kubeResource))
+			kubeResourceBytes := mustMarshal(kubeResources)
```
leelavg, line 361:
```suggestion
			kubeResourceBytes := mustMarshal(kubeResource)
```
Taking a closer look, CI could be failing due to this.

PR #3264, opened 2025-05-26
Expend KubeObject message to return expected client reconcile behavior

nb-ohad:
Add a `ReconcilePolicy` and `UpdateStrategy` knobs for each KubeObject send in the response of `GetDesiredClientState` to allow the hub to specify the reconcile behavior  expected for each object


nb-ohad:
@leelavg Why was this LGTMed if we still had offline discussions?

PR #3265, opened 2025-05-26
rdr: update peer storageID and clientProfile to be fetched from the peerClientInfo annotation

rewantsoni:
This PR does the following:

1. Removes the hack we had to fetch the rbd StorageID to fetch it from the GetStorageClientsInfo rpc
2. Generate clientProfileMapping with the right local and remote clientProfiles

This is a followup PR to https://github.com/red-hat-storage/ocs-operator/pull/3260

Review only the last two commits

Fixes: https://issues.redhat.com/browse/DFBUGS-2307

### DIFF
--- a/controllers/mirroring/mirroring_controller.go
+++ b/controllers/mirroring/mirroring_controller.go
@@ -15,6 +15,7 @@ package mirroring
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"slices"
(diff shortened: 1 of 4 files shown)


controllers/mirroring/mirroring_controller.go
```diff
@@ -471,38 +472,28 @@ func (r *MirroringReconciler) reconcileRadosNamespaceMirroring(
...
-				fmt.Sprintf("expected 1 StorageConsumer but got %v", len(storageConsumers.Items)),
-			)
-			errorOccurred = true
-			continue
+	storageConsumerByRemoteClientID := map[string]*ocsv1alpha1.StorageConsumer{}
+
+	storageConsumerList := &ocsv1alpha1.StorageConsumerList{}
+	if err := r.List(r.ctx, storageConsumerList); err != nil {
+		r.log.Error(err, "failed to list storage consumers")
+		return true
```
nb-ohad, line 481: We need to return the error here, not true

services/provider/server/server.go
```diff
@@ -1663,9 +1609,7 @@ func (s *OCSProviderServer) appendVolumeReplicationClassKubeResources(
 		return kubeResources, nil
 	}
 
-	//TODO: this is a ugly hack to generate the peerStorageID for generating replicationID
-	// peerStorageID should come from GetStorageClientInfo
-	peerStorageID, err := s.getPeerStorageID(ctx, storageCluster, consumer.Namespace, consumerConfig.GetRbdRadosNamespaceName())
+	peerStorageID, err := s.getPeerStorageID(consumer)
```
nb-ohad, line 1612: Why do we need to call into a supporting function if we only do one unmarshal? you might need to use other information for different resources. I would say, unmarshel before calling any of the appends and have the entire struct available for use

services/provider/server/server.go
```diff
@@ -1774,29 +1719,33 @@ func (s *OCSProviderServer) appendClientProfileMappingKubeResources(
...
-					Mappings: []csiopv1a1.MappingsSpec{
-						{
-							LocalClientProfile:  clientProfileName,
-							RemoteClientProfile: clientProfileName,
-							BlockPoolIdMapping:  blockPoolMapping,
+		if peerClientInfoInBytes := consumer.GetAnnotations()[util.StorageConsumerMirroringInfoAnnotation]; peerClientInfoInBytes != "" {
+			peerClientInfo := &pb.ClientInfo{}
+			if err := json.Unmarshal([]byte(peerClientInfoInBytes), &peerClientInfo); err != nil {
+				return kubeResources, err
+			}
```
nb-ohad, line 1726: You already unmarshled, you should not be unmarshling again

nb-ohad (changes requested):

services/provider/server/server.go
```diff
@@ -1655,6 +1600,7 @@ func (s *OCSProviderServer) appendVolumeReplicationClassKubeResources(
 	consumerConfig util.StorageConsumerResources,
 	storageCluster *ocsv1.StorageCluster,
 	rbdStorageId string,
+	peerRbdStorageId string,
```
nb-ohad, line 1603: Please don't use the term Peer in this context

services/provider/server/server.go
```diff
@@ -1774,29 +1715,26 @@ func (s *OCSProviderServer) appendClientProfileMappingKubeResources(
...
-					Name:      consumer.Status.Client.Name,
-					Namespace: consumer.Status.Client.OperatorNamespace,
-				},
-				Spec: csiopv1a1.ClientProfileMappingSpec{
-					Mappings: []csiopv1a1.MappingsSpec{
-						{
-							LocalClientProfile:  clientProfileName,
-							RemoteClientProfile: clientProfileName,
-							BlockPoolIdMapping:  blockPoolMapping,
+		if remoteClientProfileName := peerClientInfo.ClientProfiles[clientInfoRbdClientProfileKey]; remoteClientProfileName != "" {
```
nb-ohad, line 1718: There is no need for nested `if` here. You should do something like

```
remoteClientProfileName := peerClientInfo.ClientProfiles[clientInfoRbdClientProfileKey]
if len(blockPoolMapping) > 0 && remoteClientProfileName != "" {

**Review by nb-ohad — CHANGES_REQUESTED** [review]

**Comment by rewantsoni** [comment]
ci failed due to infra issue
/retest-required

**Comment by rewantsoni** [comment]
```
{  openshift cluster install failed overall}
```
/retest-required

PR #3271, opened 2025-05-28
controllers: add nw annotation to stoarageclient when multus enabled

rohan47:
For a storage cluster using multus
For internal mode csi controller plugins were managed by rook in 4.18, but in upgraded version csi controller plugins are managed by ceph-csi-operator managed by ocs-client-operator.
The csi controller plugin pods are missing network spec required for connecting to multus network.
[DFBUGS-2636](https://issues.redhat.com//browse/DFBUGS-2636)

### DIFF
--- a/controllers/storagecluster/storageclient.go
+++ b/controllers/storagecluster/storageclient.go
@@ -1,13 +1,17 @@
 package storagecluster
 
 import (
+	"encoding/json"
 	"fmt"
 	"maps"
 	"strconv"
 
+	nadv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
 	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
 	ocsv1a1 "github.com/red-hat-storage/ocs-operator/api/v4/v1alpha1"
 	"github.com/red-hat-storage/ocs-operator/v4/controllers/defaults"
+	"github.com/red-hat-storage/ocs-operator/v4/controllers/util"
+	rookCephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
 
 	ocsclientv1a1 "github.com/red-hat-storage/ocs-client-operator/api/v1alpha1"
 	corev1 "k8s.io/api/core/v1"
@@ -21,6 +25,8 @@ const (
 	ocsClientConfigMapName             = "ocs-client-operator-config"
 	manageNoobaaSubKey                 = "manageNoobaaSubscription"
 	useHostNetworkForCsiControllersKey = "useHostNetworkForCsiControllers"
+	// cephNetworkAnnotationKey is the annotation key used to store network details used by ceph
+	cniNetworksAnnotationKey = "k8s.v1.cni.cncf.io/networks"
 )
 
 type storageClient struct{}
@@ -64,6 +70,14 @@ func (s *storageClient) ensureCreated(r *StorageClusterReconciler, storagecluste
 		// we could just use svcName:port however in-cluster traffic from "*.svc" is generally not proxied and
 		// we using qualified name upto ".svc" makes connection not go through any proxies.
 		storageClient.Spec.StorageProviderEndpoint = fmt.Sprintf("%s.%s.svc:%d", ocsProviderServerName, storagecluster.Namespace, ocsProviderServicePort)
+
+		if storagecluster.Spec.Network != nil && storagecluster.Spec.Network.IsMultus() {
+			cephNWAnnotationValue, err := getCephNetworkAnnotationValue(storagecluster.Spec.Network, storagecluster.Namespace)
+			if err != nil {
+				return fmt.Errorf("failed to get Ceph network annotation value: %v", err)
+			}
+			util.AddAnnotation(storageClient, cniNetworksAnnotationKey, cephNWAnnotationValue)
+		}
 
 		controllerutil.AddFinalizer(storageClient, internalComponentFinalizer)
 
@@ -125,4 +139,20 @@ func (s *storageClient) updateClientConfigMap(r *StorageClusterReconciler, names
 	}
 
 	return nil
+}
+
+// getCephNetworkAnnotationValue returns the network annotation value for the given StorageCluster NetworkSpec.
+func getCephNetworkAnnotationValue(cephNetworkSpec *rookCephv1.NetworkSpec, scNamespace string) (string, error) {
+	if len(cephNetworkSpec.Selectors) == 0 {
+		return "", fmt.Errorf("invalid ceph network spec")
+	}
+	networkSelectionElement, err := cephNetworkSpec.GetNetworkSelection(scNamespace, rookCephv1.CephNetworkType("public"))
+	if err != nil || networkSelectionElement == nil {
+		return "", fmt.Errorf("failed to get network selection element: %v", err)
+	}
+	nwAnnotation, err := json.Marshal([]*nadv1.NetworkSelectionElement{networkSelectionElement})
+	if err != nil {
+		return "", fmt.Errorf("failed to marshal network selection elements: %v", err)
+	}
+	return string(nwAnnotation), nil
 }


controllers/storagecluster/storageclient.go
```diff
@@ -67,6 +75,14 @@ func (s *storageClient) ensureCreated(r *StorageClusterReconciler, storagecluste
 
 		controllerutil.AddFinalizer(storageClient, internalComponentFinalizer)
 
+		if storagecluster.Spec.Network != nil && storagecluster.Spec.Network.IsMultus() {
+			cephNWAnnotationValue, err := getCephNetworkAnnotationValue(storagecluster.Spec.Network, storagecluster.Namespace)
+			if err != nil || cephNWAnnotationValue == "" {
```
leelavg, line 80: the function should know to send an error if empty value is not valid and you should check only for err or not.

controllers/storagecluster/storageclient.go
```diff
@@ -21,6 +25,10 @@ const (
 	ocsClientConfigMapName             = "ocs-client-operator-config"
 	manageNoobaaSubKey                 = "manageNoobaaSubscription"
 	useHostNetworkForCsiControllersKey = "useHostNetworkForCsiControllers"
+	// cephNetworkAnnotationKey is the annotation key used to store network details used by ceph
+	cephNetworkAnnotationKey = "k8s.v1.cni.cncf.io/networks"
```
leelavg, line 29: I don't think this has anything to do w/ ceph, this is a cniNetworksAnnotationKey

controllers/storagecluster/storageclient.go
```diff
@@ -126,3 +142,37 @@ func (s *storageClient) updateClientConfigMap(r *StorageClusterReconciler, names
...
+func nwSelectionsToAnnotationValue(selections ...*nadv1.NetworkSelectionElement) (string, error) {
+	reduced := []*nadv1.NetworkSelectionElement{}
+	for _, s := range selections {
+		if s != nil {
+			reduced = append(reduced, s)
+		}
+	}
+	if len(reduced) == 0 {
+		return "", nil
+	}
```
leelavg, line 172: what does it mean to hit this condition?

controllers/storagecluster/storageclient.go
```diff
@@ -126,3 +142,37 @@ func (s *storageClient) updateClientConfigMap(r *StorageClusterReconciler, names
...
+		nwAnnotation, err := nwSelectionsToAnnotationValue(networkSelectionElement)
+		if err != nil || nwAnnotation == "" {
+			return "", fmt.Errorf("failed to convert network selection elements to annotation value: %v", err)
+		}
+		return nwAnnotation, nil
+	}
+	return "", fmt.Errorf("invalid ceph network spec")
+}
+
+// nwSelectionsToAnnotationValue removes nil values and converts a slice of NetworkSelectionElement to annotation value string
```
leelavg, line 162: I don't see any other data transformation here, doesn't json marshal already filter out `nil` values?

controllers/storagecluster/storageclient.go
```diff
@@ -126,3 +142,37 @@ func (s *storageClient) updateClientConfigMap(r *StorageClusterReconciler, names
...
 }
+
+// getCephNetworkAnnotationValue returns the network annotation value for the given StorageCluster NetworkSpec.
+func getCephNetworkAnnotationValue(cephNetworkSpec *rookCephv1.NetworkSpec, scNamespace string) (string, error) {
+	if len(cephNetworkSpec.Selectors) != 0 {
+		networkSelectionElement, err := cephNetworkSpec.GetNetworkSelection(scNamespace, rookCephv1.CephNetworkType(cephPublicNetworkType))
+		if err != nil {
+			return "", fmt.Errorf("failed to get network selection element: %v", err)
+		}
+		nwAnnotation, err := nwSelectionsToAnnotationValue(networkSelectionElement)
```
leelavg, line 153: maybe reflect the param var as plural as the func args is expecting a slice?

controllers/storagecluster/storageclient.go
leelavg, line 1: no new rbac is required as I'm not sure what the external funcs do?

rohan47: No new RBAC is required
https://github.com/rook/rook/blob/d2dfc29ed6fb2458d8680740e60e6e11ac2fd9fa/pkg/apis/ceph.rook.io/v1/network.go#L113

controllers/storagecluster/storageclient.go
rohan47: updated

controllers/storagecluster/storageclient.go
```diff
@@ -126,3 +142,19 @@ func (s *storageClient) updateClientConfigMap(r *StorageClusterReconciler, names
 
 	return nil
 }
+
+// getCephNetworkAnnotationValue returns the network annotation value for the given StorageCluster NetworkSpec.
+func getCephNetworkAnnotationValue(cephNetworkSpec *rookCephv1.NetworkSpec, scNamespace string) (string, error) {
+	if len(cephNetworkSpec.Selectors) != 0 {
+		networkSelectionElement, err := cephNetworkSpec.GetNetworkSelection(scNamespace, rookCephv1.CephNetworkType(cephPublicNetworkType))
+		if err != nil || networkSelectionElement == nil {
```
leelavg, line 150: looking at the `GetNetworkSelection`, `nil` seems to be a valid value for `networkSelectionElement` as well, in that scenario if we requeue via error are we expecting user to update the value?

rohan47: yes, if multus is enabled then network selections are must to be present

controllers/storagecluster/storageclient.go
```diff
@@ -21,6 +25,10 @@ const (
 	ocsClientConfigMapName             = "ocs-client-operator-config"
 	manageNoobaaSubKey                 = "manageNoobaaSubscription"
 	useHostNetworkForCsiControllersKey = "useHostNetworkForCsiControllers"
+	// cephNetworkAnnotationKey is the annotation key used to store network details used by ceph
+	cniNetworksAnnotationKey = "k8s.v1.cni.cncf.io/networks"
+	// cephPublicNetworkType is the type of network used by Ceph for public communication
+	cephPublicNetworkType = "public"
```
nb-ohad, line 31: We we get rid of this const? it is being used in a very specific context and the value is self explanatory in that context. Better to just inline this

controllers/storagecluster/storageclient.go
```diff
@@ -67,6 +75,14 @@ func (s *storageClient) ensureCreated(r *StorageClusterReconciler, storagecluste
...
 		controllerutil.AddFinalizer(storageClient, internalComponentFinalizer)
 
+		if storagecluster.Spec.Network != nil && storagecluster.Spec.Network.IsMultus() {
+			cephNWAnnotationValue, err := getCephNetworkAnnotationValue(storagecluster.Spec.Network, storagecluster.Namespace)
+			if err != nil {
+				return fmt.Errorf("failed to get Ceph network annotation value: %v", err)
+			}
+			util.AddAnnotation(storageClient, cniNetworksAnnotationKey, cephNWAnnotationValue)
+		}
+
```
nb-ohad, line 85: Can this please be moved above the AddFinalizer so we keep the add finalizer as the last change we do within the create or update?

controllers/storagecluster/storageclient.go
```diff
@@ -126,3 +142,19 @@ func (s *storageClient) updateClientConfigMap(r *StorageClusterReconciler, names
...
+		if err != nil || networkSelectionElement == nil {
+			return "", fmt.Errorf("failed to get network selection element: %v", err)
+		}
+		nwAnnotation, err := json.Marshal([]*nadv1.NetworkSelectionElement{networkSelectionElement})
+		if err != nil {
+			return "", fmt.Errorf("failed to marshal network selection elements: %v", err)
+		}
+		return string(nwAnnotation), nil
+	}
+	return "", fmt.Errorf("invalid ceph network spec")
```
nb-ohad, line 159: Can you please flip the condition so the last return will be the "happy" return?

nb-ohad (changes requested):

controllers/storagecluster/storageclient.go
rohan47: moved above AddFinalizer

leelavg:
@rohan47 possible to test the ci locally once, this wasn't passed even once afair.

PR #1254, opened 2021-07-02
Bug 1978578: [release-4.6] fixed cephTools update when spec is changed

agarwal-mudit:
Signed-off-by: Santosh Pillai <sapillai@redhat.com>
(cherry picked from commit e524fc1ce339429ed4263796ea4a858e81678b0b)

PR #3273, opened 2025-05-30
Enable ReadAffinity with default CrushLocationLabels in CephConnection

OdedViner:
This PR addresses the bug where read affinity was not configured in the CephConnection resource.

Adds ReadAffinity section with default CrushLocationLabels:
```
"topology.rook.io/chassis", "topology.rook.io/rack", "topology.rook.io/row", "topology.rook.io/pdu", "topology.rook.io/pod", "topology.rook.io/room", "topology.rook.io/datacenter", "topology.kubernetes.io/region", "topology.kubernetes.io/zone", "kubernetes.io/hostname"
```
This change ensures that CSI read affinity is enabled by default.

### DIFF
--- a/services/provider/server/server.go
+++ b/services/provider/server/server.go
@@ -389,7 +389,7 @@ func (s *OCSProviderServer) GetDesiredClientState(ctx context.Context, req *pb.G
 			return nil, err
 		}
 
-		cephConnection, err := s.getDesiredCephConnection(ctx, consumer)
+		cephConnection, err := s.getDesiredCephConnection(ctx, consumer, storageCluster)
 		if err != nil {
 			klog.Error(err)
 			return nil, status.Errorf(codes.Internal, "failed to produce client state hash")
@@ -656,7 +656,7 @@ func (s *OCSProviderServer) ReportStatus(ctx context.Context, req *pb.ReportStat
 		return nil, status.Errorf(codes.Internal, "Failed to get mirroring status for consumer.")
 	}
 
-	cephConnection, err := s.getDesiredCephConnection(ctx, storageConsumer)
+	cephConnection, err := s.getDesiredCephConnection(ctx, storageConsumer, storageCluster)
 	if err != nil {
 		klog.Error(err)
 		return nil, status.Errorf(codes.Internal, "failed to produce client state hash")
@@ -1036,7 +1036,7 @@ func (s *OCSProviderServer) getKubeResources(ctx context.Context, consumer *ocsv
 	}
 
 	kubeResources := []client.Object{}
-	if cephConnection, err := s.getDesiredCephConnection(ctx, consumer); err == nil {
+	if cephConnection, err := s.getDesiredCephConnection(ctx, consumer, storageCluster); err == nil {
 		kubeResources = append(kubeResources, cephConnection)
 	} else {
 		return nil, err
@@ -1148,6 +1148,7 @@ func (s *OCSProviderServer) getKubeResources(ctx context.Context, consumer *ocsv
 func (s *OCSProviderServer) getDesiredCephConnection(
 	ctx context.Context,
 	consumer *ocsv1alpha1.StorageConsumer,
+	storageCluster *ocsv1.StorageCluster,
 ) (*csiopv1a1.CephConnection, error) {
 
 	configmap := &v1.ConfigMap{}
@@ -1166,13 +1167,39 @@ func (s *OCSProviderServer) getDesiredCephConnection(
 		return nil, fmt.Errorf("failed to extract monitor IPs from configmap %s: %v", monConfigMap, err)
 	}
 
+	readAffinityOptions := util.GetReadAffinityOptions(storageCluster)
+	readAffinityDisabledAnnotation, _ := strconv.ParseBool(consumer.Annotations["ocs.openshift.io/disable-read-affinity"])
+
+	var readAffinity *csiopv1a1.ReadAffinitySpec = nil
+	if readAffinityOptions.Enabled && !readAffinityDisabledAnnotation {
+		labels := readAffinityOptions.CrushLocationLabels
+		if len(labels) == 0 {
+			labels = []string{
+				"kubernetes.io/hostname",
+				"topology.kubernetes.io/region",
+				"topology.kubernetes.io/zone",
+				"topology.rook.io/datacenter",
+				"topology.rook.io/room",
+				"topology.rook.io/pod",
+				"topology.rook.io/pdu",
+				"topology.rook.io/row",
+				"topology.rook.io/rack",
+				"topology.rook.io/chassis",
+			}
+		}
+		readAffinity = &csiopv1a1.ReadAffinitySpec{
+			CrushLocationLabels: labels,
+		}
+	}
+
 	return &csiopv1a1.CephConnection{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      consumer.Status.Client.Name,
 			Namespace: consumer.Status.Client.OperatorNamespace,
 		},
 		Spec: csiopv1a1.CephConnectionSpec{
-			Monitors: monIps,
+			Monitors:     monIps,
+			ReadAffinity: readAffinity,
 		},
 	}, nil
 }


services/provider/server/server.go
```diff
@@ -1236,6 +1236,20 @@ func (s *OCSProviderServer) getDesiredCephConnection(
 		},
 		Spec: csiopv1a1.CephConnectionSpec{
 			Monitors: monIps,
+			ReadAffinity: &csiopv1a1.ReadAffinitySpec{
```
leelavg, line 1239: @OdedViner having this here makes every connected client have readaffinity but only a corresponding storageclass will use this and no side effect for other SCs?

parth-gr: @leelavg readaffinity will be enabled for all the storageclasses.

parth-gr, line 1239: Changes:
Only add `ReadAffinity` when there is no `ocs.openshift.io/disable-replica1` annotation on the storageconsumer

parth-gr (changes requested):

services/provider/server/server.go
parth-gr: Looks good, we can check for its value to, if its true

parth-gr: Also we need to check whats set in the storagecluster CR, if nothing is set by default turn it on. 
Like this
https://github.com/red-hat-storage/ocs-operator/blob/9b583bc5acd101af990c90da1ce450eb7fcf0531/controllers/util/csi_options.go#L21

After that test your changes and remove WIP

OdedViner: Thanks @parth-gr 
I’ve updated the code to check if the annotation `ocs.openshift.io/disable-replica1` is set to `true`. If it’s not, I merge the default labels with `sc.Spec.CSI.ReadAffinity.CrushLocationLabels` from the StorageCluster using `GetReadAffinityOptions()`

Just to confirm — is it correct to merge the default labels with the ones from the StorageCluster, or should we only use the list defined in `sc.Spec.CSI.ReadAffinity.CrushLocationLabels` if it's set?

services/provider/server/server.go
```diff
@@ -1229,13 +1229,50 @@ func (s *OCSProviderServer) getDesiredCephConnection(
...
+			"topology.kubernetes.io/region",
+			"topology.kubernetes.io/zone",
+			"kubernetes.io/hostname",
+		}
+
+		labelSet := make(map[string]struct{})
+		for _, label := range defaultLabels {
+			labelSet[label] = struct{}{}
+		}
+		storageCluster, err := util.GetStorageClusterInNamespace(ctx, s.client, s.namespace)
```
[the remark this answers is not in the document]
parth-gr: I still see issue, can follow this if else block

services/provider/server/server.go
```diff
@@ -1229,13 +1229,45 @@ func (s *OCSProviderServer) getDesiredCephConnection(
...
 
+	var readAffinity *csiopv1a1.ReadAffinitySpec = nil
+	storageCluster, err := util.GetStorageClusterInNamespace(ctx, s.client, s.namespace)
+	if err != nil {
+		return nil, err
+	}
+	readAffinityOptions := util.GetReadAffinityOptions(storageCluster)
+
+	replica1Disabled := consumer.Annotations["ocs.openshift.io/disable-replica1"] == "true"
+
```
parth-gr, line 1240: Only this condition is needed `readAffinityOptions.Enabled && !replica1Disabled`
 and merge the labels always.

OdedViner: @parth-gr done. Can you review?

services/provider/server/server.go
```diff
@@ -1229,13 +1229,57 @@ func (s *OCSProviderServer) getDesiredCephConnection(
 		return nil, fmt.Errorf("failed to extract monitor IPs from configmap %s: %v", monConfigMap, err)
 	}
 
+	var readAffinity *csiopv1a1.ReadAffinitySpec = nil
```
parth-gr, line 1232: Looks good now, can you do the testing,
and remove WIP

services/provider/server/server.go
```diff
@@ -1229,13 +1229,57 @@ func (s *OCSProviderServer) getDesiredCephConnection(
...
+	storageCluster, err := util.GetStorageClusterInNamespace(ctx, s.client, s.namespace)
+	if err != nil {
+		return nil, err
+	}
+	readAffinityOptions := util.GetReadAffinityOptions(storageCluster)
+
+	replica1Disabled := consumer.Annotations["ocs.openshift.io/disable-replica1"] == "true"
```
parth-gr, line 1239: ocs.openshift.io/disable-replica1 -> ocs.openshift.io/disable-read-affinity

OdedViner: done

services/provider/server/server.go
```diff
@@ -1229,13 +1231,41 @@ func (s *OCSProviderServer) getDesiredCephConnection(
...
+		"topology.rook.io/datacenter",
+		"topology.kubernetes.io/region",
+		"topology.kubernetes.io/zone",
+		"kubernetes.io/hostname",
+	}
+	var mergedLabels []string
+
+	if readAffinityOptions.Enabled && !replica1Disabled {
+		mergedLabels = slices.Concat(defaultCrushLocationLabels, readAffinityOptions.CrushLocationLabels)
+		sort.Strings(mergedLabels)
```
parth-gr, line 1254: #1254 and 1255 lines are not needed 
Cc @Rakshith-R

services/provider/server/server.go
```diff
@@ -1229,13 +1229,57 @@ func (s *OCSProviderServer) getDesiredCephConnection(
 		return nil, fmt.Errorf("failed to extract monitor IPs from configmap %s: %v", monConfigMap, err)
 	}
 
+	var readAffinity *csiopv1a1.ReadAffinitySpec = nil
+	storageCluster, err := util.GetStorageClusterInNamespace(ctx, s.client, s.namespace)
```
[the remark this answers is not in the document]
leelavg: That's not required.

services/provider/server/server.go
```diff
@@ -1229,13 +1231,41 @@ func (s *OCSProviderServer) getDesiredCephConnection(
 		return nil, fmt.Errorf("failed to extract monitor IPs from configmap %s: %v", monConfigMap, err)
 	}
 
+	var readAffinity *csiopv1a1.ReadAffinitySpec = nil
+	readAffinityOptions := util.GetReadAffinityOptions(storageCluster)
```
parth-gr, line 1235: readAffinityOptions -> readAffinity

services/provider/server/server.go
```diff
@@ -1229,13 +1231,41 @@ func (s *OCSProviderServer) getDesiredCephConnection(
...
+
+	replica1Disabled := consumer.Annotations["ocs.openshift.io/disable-read-affinity"] == "true"
```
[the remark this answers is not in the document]
OdedViner: done

services/provider/server/server.go
OdedViner: I'm using the readAffinity variable name for another variable just above, so I kept readAffinityOptions here to avoid conflict

services/provider/server/server.go
```diff
@@ -1229,13 +1230,39 @@ func (s *OCSProviderServer) getDesiredCephConnection(
 		return nil, fmt.Errorf("failed to extract monitor IPs from configmap %s: %v", monConfigMap, err)
 	}
 
+	var readAffinity *csiopv1a1.ReadAffinitySpec = nil
```
nb-ohad, line 1233: Please move the definition of the var the nearest place above the place where it is being used

services/provider/server/server.go
```diff
@@ -1229,13 +1230,39 @@ func (s *OCSProviderServer) getDesiredCephConnection(
...
+		"kubernetes.io/hostname",
+	}
+	var mergedLabels []string
+
+	if readAffinityOptions.Enabled && !readaffinityDisabledAnnotation {
+		mergedLabels = slices.Concat(defaultCrushLocationLabels, readAffinityOptions.CrushLocationLabels)
+		readAffinity = &csiopv1a1.ReadAffinitySpec{
+			CrushLocationLabels: mergedLabels,
+		}
+	}
```
nb-ohad, line 1256: This entire code does not take into account duplications. 
Please use a set (`map[string]struct{}`) for the default values, add the additional values to the set, then use `maps.keys` to get back the list without duplicates

services/provider/server/server.go
```diff
@@ -1229,13 +1230,39 @@ func (s *OCSProviderServer) getDesiredCephConnection(
...
+		"topology.rook.io/rack",
+		"topology.rook.io/row",
+		"topology.rook.io/pdu",
+		"topology.rook.io/pod",
+		"topology.rook.io/room",
+		"topology.rook.io/datacenter",
+		"topology.kubernetes.io/region",
+		"topology.kubernetes.io/zone",
+		"kubernetes.io/hostname",
+	}
```
nb-ohad, line 1248: There is no need to define the slice out side of the if below, please move inside the if statement

nb-ohad (changes requested):

services/provider/server/server.go
parth-gr: @Rakshith-R I think we can send duplicates, but lets be sure,

I think we can use slice.compact(s1, s2) too

Rakshith-R: > @Rakshith-R I think we can send duplicates, but lets be sure,

It is handled at cephcsi here 
https://github.com/ceph/ceph-csi/blob/dd93de0bfc8165b8676c6e5045477290077fbf85/internal/util/crushlocation.go#L43-L46

services/provider/server/server.go
OdedViner: done

services/provider/server/server.go
OdedViner: done

parth-gr (approved):

services/provider/server/server.go
```diff
@@ -1229,13 +1230,39 @@ func (s *OCSProviderServer) getDesiredCephConnection(
 		return nil, fmt.Errorf("failed to extract monitor IPs from configmap %s: %v", monConfigMap, err)
 	}
 
+	readAffinityOptions := util.GetReadAffinityOptions(storageCluster)
+	readaffinityDisabledAnnotation := consumer.Annotations["ocs.openshift.io/disable-read-affinity"] == "true"
```
nb-ohad, line 1234: I would prefer if we would not hard code boolean values as strings. There are 2 alternatives:
- format an actual boolean value into the string that we care about

```suggestion
	readAffinityDisabledAnnotation := consumer.Annotations["ocs.openshift.io/disable-read-affinity"] == strconv.FormatBool(true) {
```

Or to parse the value of the annotation as a bool
```suggestion
	readAffinityDisabledAnnotation, _ := strconv.ParseBool(consumer.Annotations["ocs.openshift.io/disable-read-affinity"])
```

The second approach automatically supports multiple variation of a truth value: "true", "True", "TRUE", "1", etc.

nb-ohad (changes requested):

services/provider/server/server.go
OdedViner: done

services/provider/server/server.go
```diff
@@ -1229,13 +1230,39 @@ func (s *OCSProviderServer) getDesiredCephConnection(
...
+
+	var readAffinity *csiopv1a1.ReadAffinitySpec = nil
+	if readAffinityOptions.Enabled && !readaffinityDisabledAnnotation {
+		labels := readAffinityOptions.CrushLocationLabels
+		if len(labels) == 0 {
+			labels = []string{
+				"topology.rook.io/chassis",
```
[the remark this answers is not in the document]
OdedViner: @parth-gr  I've reviewed the list at the link and matched the label priorities accordingly. The following labels from the code were already added:
```
corev1.LabelZoneRegionStable= "topology.kubernetes.io/region"
corev1.LabelZoneFailureDomainStable= "topology.kubernetes.io/zone"
LabelHostname = "kubernetes.io/hostname"
```
Let me know if any specific ordering or exclusion is needed.

services/provider/server/server.go
```diff
@@ -1166,13 +1167,39 @@ func (s *OCSProviderServer) getDesiredCephConnection(
...
+				"topology.kubernetes.io/region",
+				"topology.kubernetes.io/zone",
+				"topology.rook.io/datacenter",
+				"topology.rook.io/room",
+				"topology.rook.io/pod",
+				"topology.rook.io/pdu",
+				"topology.rook.io/row",
+				"topology.rook.io/rack",
+				"topology.rook.io/chassis",
+				"kubernetes.io/hostname",
```
parth-gr, line 1187: Hostname 
Zone
Region
Chassis 
Rack
....


This is the correct order try to Match with Rook code
/hold

nb-ohad: @OdedViner The order in the code is still not aligned with the order @parth-gr provided above. Can you please fix it?

parth-gr: Cc @Rakshith-R ^^

OdedViner: @parth-gr  can you check now? https://github.com/rook/rook/blob/cd2d887c928821e95b6d008981bc2826a6a32083/pkg/operator/ceph/csi/cluster_config_test.go#L434

services/provider/server/server.go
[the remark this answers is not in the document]
subhamkrai: These are the labels
```
topology.kubernetes.io/region
topology.kubernetes.io/zone
topology.rook.io/datacenter
topology.rook.io/room
topology.rook.io/pod
topology.rook.io/pdu
topology.rook.io/row
topology.rook.io/rack
topology.rook.io/chassis
```

and label `kubernetes.io/hostname` is added by default in Rook if the env `ROOK_CUSTOM_HOSTNAME_LABEL` is not set otherwise the custom label is added according to the env var.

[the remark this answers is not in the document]
OdedViner: @subhamkrai Did you take this data from the following doc?
https://rook.io/docs/rook/latest-release/CRDs/Cluster/ceph-cluster-crd/?h=topology.rook.io%2Fchassis#osd-topology
Just want to confirm.

parth-gr:
@nb-ohad @OdedViner 
In the code https://github.com/rook/rook/blob/d013b479b07ffbbe419b404f976b166e1130936b/pkg/operator/ceph/cluster/osd/topology/topology.go#L144
The code has list as:
```
Hostname
Region
Zone
Chassis
Rack
....
```


And in docs it has
```
topology.kubernetes.io/region
topology.kubernetes.io/zone
topology.rook.io/datacenter
topology.rook.io/room
topology.rook.io/pod
```


So we need to confirm which is correct, cc @Rakshith-R 
@subhamkrai @sp98 @travisn ^^

PR #3274, opened 2025-06-02
server: add a event handler to update uid store of consumer

rewantsoni:
Create a new Informer that watches the storageConsumer and updates the uid store based on the creation/deletion of the storageConsumer in provider server

Fixes restart issue with https://github.com/red-hat-storage/ocs-operator/pull/3157

Part of Fixes: https://issues.redhat.com/browse/DFBUGS-2290


services/provider/server/consumer.go
```diff
@@ -229,22 +215,15 @@ func (c *ocsConsumerManager) GetByClientID(ctx context.Context, clientID string)
 }
 
 func (c *ocsConsumerManager) ClearClientInformation(ctx context.Context, consumerID string) error {
-	uid := types.UID(consumerID)
-	c.mutex.RLock()
-	consumerName, ok := c.nameByUID[uid]
-	c.mutex.RUnlock()
-	if !ok {
-		return fmt.Errorf("no storageConsumer found with the UID %q", consumerID)
+	consumer, err := c.Get(ctx, consumerID)
```
rewantsoni, line 218: Using get here to make sure that the consumer we get from cache is not stale

services/provider/server/consumer.go
```diff
@@ -23,68 +25,56 @@ type ocsConsumerManager struct {
...
+		},
+	}); err != nil {
+		return nil, fmt.Errorf("failed to create informer: %w", err)
 	}
 
-	c.mutex.Lock()
-	delete(c.nameByUID, uid)
-	c.mutex.Unlock()
+	go func() {
+		if err := cache.Start(ctx); err != nil {
```
leelavg, line 68: what is the lifetime of this `ctx`, this shouldn't be less than server lifetime.

rewantsoni: the ctx here is the same as used for creating the NewProviderServer so it's lifetime shouldn't be lesser than server lifetime

PR #3275, opened 2025-06-02
send vgrc from provider to client

rewantsoni:
Send VGRC from the provider to the client.

Fixes: https://issues.redhat.com/browse/DFBUGS-2672


controllers/util/groupreplicationclasses.go
```diff
@@ -0,0 +1,72 @@
+package util
+
+import (
+	"context"
```
leelavg, line 4: pls configure your editor to group the imports properly,
```
import (
    stdlib

    own-module

    third-party
)
```

services/provider/server/server.go
```diff
@@ -1813,6 +1826,42 @@ func (s *OCSProviderServer) appendVolumeReplicationClassKubeResources(
...
+	if mirrorEnabled, err := s.isConsumerMirrorEnabled(ctx, consumer); err != nil {
+		return kubeResources, err
+	} else if !mirrorEnabled {
+		klog.Infof("skipping distribution of VolumeGroupReplicationClass as mirroring is not enabled for the consumer")
+		return kubeResources, nil
+	}
+
+	resources := getKubeResourcesForClass(
+		consumer.Spec.VolumeGroupReplicationClasses,
+		"VolumeGroupSnapshotClass",
```
leelavg, line 1847:
```suggestion
		"VolumeGroupReplicationClass",
```

services/provider/server/server.go
```diff
@@ -1813,6 +1826,42 @@ func (s *OCSProviderServer) appendVolumeReplicationClassKubeResources(
...
+	ctx context.Context,
+	kubeResources []client.Object,
+	consumer *ocsv1alpha1.StorageConsumer,
+	consumerConfig util.StorageConsumerResources,
+	storageCluster *ocsv1.StorageCluster,
+	rbdStorageId string,
+	remoteRbdStorageId string,
+) ([]client.Object, error) {
+	if mirrorEnabled, err := s.isConsumerMirrorEnabled(ctx, consumer); err != nil {
+		return kubeResources, err
```
leelavg, line 1839:
```suggestion
		return nil, err
```
no need for caller to second guess

services/provider/server/server.go
```diff
@@ -1813,6 +1826,42 @@ func (s *OCSProviderServer) appendVolumeReplicationClassKubeResources(
...
+	storageCluster *ocsv1.StorageCluster,
+	rbdStorageId string,
+	remoteRbdStorageId string,
+) ([]client.Object, error) {
+	if mirrorEnabled, err := s.isConsumerMirrorEnabled(ctx, consumer); err != nil {
+		return nil, err
+	} else if !mirrorEnabled {
+		klog.Infof("skipping distribution of VolumeGroupReplicationClass as mirroring is not enabled for the consumer")
+		return kubeResources, nil
+	}
```
nb-ohad, line 1843: Why does the server condition the sending of the information to the client if mirroring is not enabled? I would say that this is more a confusing behavior then anything else. If the admin configured something to be sent it should be sent, if he didn't first enabled mirroring then it is his mistake but the system should continue to be consistent.

controllers/util/groupreplicationclasses.go
```diff
@@ -0,0 +1,74 @@
...
+		return nil, fmt.Errorf("volume group replication class name mismatch: %s, %v", replicationClassName, vgrc.Name)
+	}
+
+	switch vgrc.Spec.Provisioner {
+	case RbdDriverName:
+		// For VGRC the replicationID will be a combination of RBDStorageID, RemoteRBDStorageID and the poolName
+		// pool name is added to the VGRC's template
+		poolName := vgrc.Spec.Parameters["pool"]
+		storageIDs := []string{rbdStorageId, remoteRbdStorageId, poolName}
+		slices.Sort(storageIDs)
```
nb-ohad, line 62: This is wrong! it will create an assortment of potential orders between the storageId and the poolName. 
You should sort only the storage ID and the name should be added last. Because we are only talking about 2 strings you don't need even apply a slice sort functionality, just a string compare functionality

controllers/util/groupreplicationclasses.go
```diff
@@ -0,0 +1,74 @@
...
+		replicationID := CalculateMD5Hash(storageIDs)
+		vgrc.Spec.Parameters["replication.storage.openshift.io/group-replication-secret-name"] = consumerConfig.GetCsiRbdProvisionerCephUserName()
```
nb-ohad, line 64: Is there a reason the replication class uses the same user as the storage class?

controllers/util/groupreplicationclasses.go
```diff
@@ -0,0 +1,74 @@
...
+		vgrc.Spec.Parameters["replication.storage.openshift.io/group-replication-secret-namespace"] = consumer.Status.Client.OperatorNamespace
+		vgrc.Spec.Parameters["clusterID"] = consumerConfig.GetRbdClientProfileName()
+		AddLabel(vgrc, ramenDRStorageIDLabelKey, rbdStorageId)
+		AddLabel(vgrc, ramenMaintenanceModeLabelKey, "Failover")
```
nb-ohad, line 68: IF this is a constant in need to be part of the template

nb-ohad (changes requested):

PR #3301, opened 2025-06-17
controllers: update the correct values for ConsumerConfigMap

ShravaniVangur:
This PR prevents `storageconsumer_controller` from setting default values in `consumerConfigMap` for `cephfs-subvolumegroup`, to avoid inconsistent CR creation.

Resolves: [DFBUGS-2765](https://issues.redhat.com/browse/DFBUGS-2765)

### DIFF
--- a/controllers/storageconsumer/storageconsumer_controller.go
+++ b/controllers/storageconsumer/storageconsumer_controller.go
@@ -62,6 +62,7 @@ const (
 	ConsumerUUIDLabel             = "ocs.openshift.io/storageconsumer-uuid"
 	StorageConsumerNameLabel      = "ocs.openshift.io/storageconsumer-name"
 	storageConsumerFinalizer      = "ocs.openshift.io/storageconsumer-protection"
+	primaryConsumerUIDAnnotation  = "ocs.openshift.io/primary-consumer-uid"
 )
 
 // StorageConsumerReconciler reconciles a StorageConsumer object
@@ -200,17 +201,9 @@ func (r *StorageConsumerReconciler) reconcileEnabledPhases() (reconcile.Result, 
 			return reconcile.Result{}, err
 		}
 
-		// Get config map's controller reference
-		controllerIndex := slices.IndexFunc(
-			consumerConfigMap.OwnerReferences,
-			func(ref metav1.OwnerReference) bool { return ptr.Deref(ref.Controller, false) },
-		)
-		var controllerRef *metav1.OwnerReference
-		if controllerIndex != -1 {
-			controllerRef = &consumerConfigMap.OwnerReferences[controllerIndex]
-		}
+		primaryConsumerUID := consumerConfigMap.GetAnnotations()[primaryConsumerUIDAnnotation]
 
-		isPrimaryConsumer := controllerRef != nil && controllerRef.UID == r.storageConsumer.UID
+		isPrimaryConsumer := primaryConsumerUID == string(r.storageConsumer.UID)
 
 		if isPrimaryConsumer {
 			consumerResources := util.WrapStorageConsumerResourceMap(consumerConfigMap.Data)
@@ -445,21 +438,6 @@ func (r *StorageConsumerReconciler) reconcileConsumerConfigMap(
 ) error {
 
 	if _, err := controllerutil.CreateOrUpdate(r.ctx, r.Client, consumerConfigMap, func() error {
-		if consumerConfigMap.Data == nil {
-			consumerConfigMap.Data = map[string]string{}
-		}
-
-		defaultConsumerResourceNames := util.GetStorageConsumerDefaultResourceNames(
-			r.storageConsumer.Name,
-			string(r.storageConsumer.UID),
-			availableServices,
-		)
-		for key := range defaultConsumerResourceNames {
-			consumerConfigMap.Data[key] = cmp.Or(
-				strings.Trim(consumerConfigMap.Data[key], " "),
-				defaultConsumerResourceNames[key],
-			)
-		}
 
 		// Get config map's controller reference
 		controllerIndex := slices.IndexFunc(
@@ -470,8 +448,8 @@ func (r *StorageConsumerReconciler) reconcileConsumerConfigMap(
 		if controllerIndex != -1 {
 			controllerRef = &consumerConfigMap.OwnerReferences[controllerIndex]
 		}
-		// If there is no controller ref, take control over the config map
-		if controllerRef == nil {
+
+		if controllerRef == nil || controllerRef.UID == r.storageConsumer.UID {
 			if err := controllerutil.SetControllerReference(
 				r.storageConsumer,
 				consumerConfigMap,
@@ -479,14 +457,40 @@ func (r *StorageConsumerReconciler) reconcileConsumerConfigMap(
 			); err != nil {
 				return err
 			}
-			// If I am not the config map controller add me as an owner
-		} else if controllerRef.UID != r.storageConsumer.UID {
+
+			if consumerConfigMap.Data == nil {
+				consumerConfigMap.Data = map[string]string{}
+			}
+			defaultConsumerResourceNames := util.GetStorageConsumerDefaultResourceNames(
+				r.storageConsumer.Name,
+				string(r.storageConsumer.UID),
+				availableServices,
+			)
+			for key := range defaultConsumerResourceNames {
+				consumerConfigMap.Data[key] = cmp.Or(
+					strings.Trim(consumerConfigMap.Data[key], " "),
+					defaultConsumerResourceNames[key],
+				)
+			}
+
+			util.AddAnnotation(consumerConfigMap, primaryConsumerUIDAnnotation, string(r.storageConsumer.UID))
+		} else {
 			if err := controllerutil.SetOwnerReference(
 				r.storageConsumer,
 				consumerConfigMap,
 				r.Scheme,
 			); err != nil {
 				return err
+			}
+
+			primaryRefIndex := slices.IndexFunc(
+				consumerConfigMap.OwnerReferences,
+				func(ref metav1.OwnerReference) bool {
+					return string(ref.UID) == consumerConfigMap.GetAnnotations()[primaryConsumerUIDAnnotation]
+				},
+			)
+			if primaryRefIndex == -1 {
+				util.AddAnnotation(consumerConfigMap, primaryConsumerUIDAnnotation, string(r.storageConsumer.UID))
 			}
 		}
 		return nil
--- a/controllers/storagecluster/storageconsumer.go
+++ b/controllers/storagecluster/storageconsumer.go
@@ -101,6 +101,9 @@ func (s *storageConsumer) ensureCreated(r *StorageClusterReconciler, storageClus
 	consumerConfigMap.Name = localStorageConsumerConfigMapName
 	consumerConfigMap.Namespace = storageCluster.Namespace
 	if _, err := controllerutil.CreateOrUpdate(r.ctx, r.Client, consumerConfigMap, func() error {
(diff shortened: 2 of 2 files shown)


controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -479,6 +457,23 @@ func (r *StorageConsumerReconciler) reconcileConsumerConfigMap(
...
+				r.storageConsumer.Name,
+				string(r.storageConsumer.UID),
+				availableServices,
+			)
+			for key := range defaultConsumerResourceNames {
+				consumerConfigMap.Data[key] = cmp.Or(
+					strings.Trim(consumerConfigMap.Data[key], " "),
+					defaultConsumerResourceNames[key],
+				)
+			}
```
rewantsoni, line 476: This works for the first reconciliation if there is no controllerRef. What if, as a day-2 operation, I enable the NFS service, in this case the NFS keys/values won't exist in the configmap as the controllerRef is not nil. 

We need to move this below the OwnerRef/ControllerRef and check if the current consumer if the controllerRef we update the values

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -489,6 +484,13 @@ func (r *StorageConsumerReconciler) reconcileConsumerConfigMap(
 				return err
 			}
 		}
+
+		annotations := consumerConfigMap.GetAnnotations()
+		primaryConsumerUID, hasAnnotation := annotations[StorageConsumerUIDAnnotation]
+		if !hasAnnotation || primaryConsumerUID != string(r.storageConsumer.UID) {
+			util.AddAnnotation(consumerConfigMap, StorageConsumerUIDAnnotation, string(r.storageConsumer.UID))
+		}
```
rewantsoni, line 492: The check here needs to be broken down:
1. If the annotation is not present add the annotation
2. If it is present:
   2.1 If a consumer with the UID specified in the annotation doesn't exist, then we update the annotation to point to the current consumer uid

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -62,6 +62,7 @@ const (
 	ConsumerUUIDLabel             = "ocs.openshift.io/storageconsumer-uuid"
 	StorageConsumerNameLabel      = "ocs.openshift.io/storageconsumer-name"
 	storageConsumerFinalizer      = "ocs.openshift.io/storageconsumer-protection"
+	StorageConsumerUIDAnnotation  = "ocs.openshift.io/primary-consumer-uid"
```
rewantsoni, line 65:
```suggestion
	PrimaryConsumerUIDAnnotation  = "ocs.openshift.io/primary-consumer-uid"
```

rewantsoni (changes requested):

controllers/storageconsumer/storageconsumer_controller.go
ShravaniVangur: Updated the logic to reflect the above, but my question here is that in any case even if primaryConsumerUID is nil the condition would evaluate to true and hence the Annotations would still be added?

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -62,6 +62,7 @@ const (
 	ConsumerUUIDLabel             = "ocs.openshift.io/storageconsumer-uuid"
 	StorageConsumerNameLabel      = "ocs.openshift.io/storageconsumer-name"
 	storageConsumerFinalizer      = "ocs.openshift.io/storageconsumer-protection"
+	PrimaryConsumerUIDAnnotation  = "ocs.openshift.io/primary-consumer-uid"
```
leelavg, line 65: Lowercase the var, I don't think we are exporting it.

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -489,6 +468,46 @@ func (r *StorageConsumerReconciler) reconcileConsumerConfigMap(
...
+			consumerList := &ocsv1alpha1.StorageConsumerList{}
+			err := r.Client.List(r.ctx, consumerList, client.InNamespace(r.namespace))
+			var consumer *ocsv1alpha1.StorageConsumer
+			if err == nil {
+				for i := range consumerList.Items {
+					if string(consumerList.Items[i].UID) == primaryConsumerUID {
+						consumer = &consumerList.Items[i]
+						break
+					}
+				}
```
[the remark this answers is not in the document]
rewantsoni: It does work but using an Indexer might be a better approach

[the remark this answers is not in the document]
ShravaniVangur: Updated it.

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -489,6 +468,42 @@ func (r *StorageConsumerReconciler) reconcileConsumerConfigMap(
...
+			}
+		}
+
+		annotations := consumerConfigMap.GetAnnotations()
+		primaryConsumerUID, hasAnnotation := annotations[PrimaryConsumerUIDAnnotation]
+		if !hasAnnotation {
+			util.AddAnnotation(consumerConfigMap, PrimaryConsumerUIDAnnotation, string(r.storageConsumer.UID))
+		} else {
+			// Check if the consumer with the UID in the annotation exists; if not, update the annotation to the current consumer UID
+			consumerList := &ocsv1alpha1.StorageConsumerList{}
```
leelavg, line 495: You don't need to list, the idea here (as per Ohad) is, when an object is removed k8s will remove the owner and you just need to check in current owner does the annotation UID exists or not.

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -489,6 +468,30 @@ func (r *StorageConsumerReconciler) reconcileConsumerConfigMap(
...
+				consumerConfigMap.Data[key] = cmp.Or(
+					strings.Trim(consumerConfigMap.Data[key], " "),
+					defaultConsumerResourceNames[key],
+				)
+			}
+		}
+
+		annotations := consumerConfigMap.GetAnnotations()
+		primaryConsumerUID, hasAnnotation := annotations[primaryConsumerUIDAnnotation]
+		if !hasAnnotation || (hasAnnotation && primaryConsumerUID != string(controllerRef.UID)) {
```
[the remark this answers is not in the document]
rewantsoni: we can use something like:

```
if !hasAnnotation || !(slices.ContainsFunc(cm1.GetOwnerReferences(), func(ownerRef metav1.OwnerReference) bool{
		return ownerRef.UID == primaryConsumerUID
	})){
	//add annotation
```

leelavg (commented):
LGTM.

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -489,6 +468,38 @@ func (r *StorageConsumerReconciler) reconcileConsumerConfigMap(
 				return err
 			}
 		}
+
+		if controllerRef != nil && controllerRef.UID == r.storageConsumer.UID {
```
nb-ohad, line 472: This is basically an else to the 2 cases above. 
```suggestion
		else {
```

I would also suggest to reorder the cases
1. `if controllerRef == nil`
2 `else if controllerRef.UID == r.storageConsumer.UID`
3. `else` (here we know that controllerRef is not consumer UID)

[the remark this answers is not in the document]
ShravaniVangur: `Owns(&corev1.ConfigMap{}, builder.MatchEveryOwner)` exists currently

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -489,6 +468,38 @@ func (r *StorageConsumerReconciler) reconcileConsumerConfigMap(
...
+				r.storageConsumer.Name,
+				string(r.storageConsumer.UID),
+				availableServices,
+			)
+			for key := range defaultConsumerResourceNames {
+				consumerConfigMap.Data[key] = cmp.Or(
+					strings.Trim(consumerConfigMap.Data[key], " "),
+					defaultConsumerResourceNames[key],
+				)
+			}
```
nb-ohad, line 486: This code should also run in case there is no controller ref. You should move it into the first if case and argument the condition to 
`if controllerRef == nil || controllerRef == r.storageConsumer.UID`

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -489,6 +468,38 @@ func (r *StorageConsumerReconciler) reconcileConsumerConfigMap(
...
+			for _, ref := range consumerConfigMap.OwnerReferences {
+				if string(ref.UID) == primaryConsumerUID {
+					hasMatchingOwner = true
+					break
+				}
+			}
+		}
+		if !hasAnnotation || !hasMatchingOwner {
+			util.AddAnnotation(consumerConfigMap, primaryConsumerUIDAnnotation, string(r.storageConsumer.UID))
+		}
```
nb-ohad, line 502: This entire code section can lose a lot of assumption and can be collapsed into:

```suggestion
		primaryConsumerUID := consumerConfigMap.GetAnnotations()[primaryConsumerUIDAnnotation]
		primaryRefIndex := slices.IndexFunc(
			consumerConfigMap.OwnerReferences,
			func(ref metav1.OwnerReference) bool { return ref.UID == primaryConsumerUID) },
		)
		if  primaryRefIndex == -1 {
			util.AddAnnotation(consumerConfigMap, primaryConsumerUIDAnnotation, string(r.storageConsumer.UID))
		}
```

controllers/storageconsumer/storageconsumer_controller.go
nb-ohad, line 1: There is still an issue here of not having a guarantee that if a consumer is the controller owner it is also the primary consumer. This guarantee is needed

nb-ohad (changes requested):

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -471,16 +449,30 @@ func (r *StorageConsumerReconciler) reconcileConsumerConfigMap(
 			controllerRef = &consumerConfigMap.OwnerReferences[controllerIndex]
 		}
 		// If there is no controller ref, take control over the config map
-		if controllerRef == nil {
+		if controllerRef == nil || controllerRef.UID == r.storageConsumer.UID {
```
nb-ohad, line 452: I think you need to remove/change the comment above this line

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -489,6 +481,14 @@ func (r *StorageConsumerReconciler) reconcileConsumerConfigMap(
 				return err
 			}
 		}
+
+		primaryConsumerUID := consumerConfigMap.GetAnnotations()[primaryConsumerUIDAnnotation]
+		primaryRefIndex := slices.IndexFunc(consumerConfigMap.OwnerReferences, func(ref metav1.OwnerReference) bool {
+			return string(ref.UID) == primaryConsumerUID
+		})
```
nb-ohad, line 488: this is almost unreadable, please use our established multiline convention where each arg gets its own line and the closing parentheses get the last line with the same indentation as the function call
```suggestion
		primaryRefIndex := slices.IndexFunc(
			consumerConfigMap.OwnerReferences, 
			func(ref metav1.OwnerReference) bool {
				return string(ref.UID) == primaryConsumerUID
			},
		)
```

controllers/storageconsumer/storageconsumer_controller.go
```diff
@@ -489,6 +481,14 @@ func (r *StorageConsumerReconciler) reconcileConsumerConfigMap(
 				return err
 			}
 		}
+
```
nb-ohad, line 484: The code below still have the issue of not guaranteeing that if there is a consumer controller it is the primary consumer

PR #3302, opened 2025-06-17
storageclasses: automate key rotation for encrypted RBD storageclasses

black-dragon74:
This patch sets the key rotation annotation by default on encrypted RBD storageclasses.


black-dragon74:
> @black-dragon74 What about the day2 storageClass created by the user, will they be responsible for adding the key rotation annotation?

Yes. This is only for the storageclasses that we create.

black-dragon74:
@leelavg @rewantsoni PTAL. Requires one more ✅

leelavg (approved):

Madhu-1:
@iamniting can you please review/approve it?

PR #3303, opened 2025-06-17
provider: donot send the topology resposne if replica-1 is disabled

parth-gr:
if the topology is empty we were still sending the response, and by this the client is treating the empty value as of slice length 1.
By which we were adding a empty annotation in the storageclient


parth-gr:
> @parth-gr
> 
> ```
> Merging is blocked
> Commits must have verified signatures.
> At least 2 approving reviews are required by reviewers with write access.
> ```

It should auto merge once the test get pass

PR #3305, opened 2025-06-17
odf-info: add blockpools to the odf-info configmap

rewantsoni:
Add the cephBlockPools list to the odf-info configmap

Creation of VGRC needs a pool parameter. We are sending the cephblockpool available in the cluster for MCO to take a look at and decide VGRC should be created for which blockpools


api/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -24,6 +24,21 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
```
leelavg, line 24: small doubt, do we actually use the deepcopy?

rewantsoni: I don't think we do

rewantsoni:
@Nikhil-Ladha @umangachapagain We are adding all non-built-in blockpools created by ocs-op to the configmap. Does MCO also need to know if the blockpools present in the configmap are enabled for mirroring or not?

Nikhil-Ladha:
> @Nikhil-Ladha @umangachapagain We are adding all non-built-in blockpools created by ocs-op to the configmap. Does MCO also need to know if the blockpools present in the configmap are enabled for mirroring or not?

That would be logical.
Because VGRClass with pools that don't have mirroring enabled doesn't make sense. If a VGRClass contains the pool it is obvious that the pool has mirroring enabled, otherwise we will start seeing day 2 issues like someone created a VGR using a VGRClass that doesn't have mirroring enabled and DR doesn't work for it.
Therefore, we should have only those pools added to the cm which have mirroring enabled.

controllers/storagecluster/cephblockpools.go
```diff
@@ -126,6 +126,7 @@ func (o *ocsCephBlockPools) reconcileMgrCephBlockPool(r *StorageClusterReconcile
 
 		setDefaultMetadataPoolSpec(&cephBlockPool.Spec.PoolSpec, storageCluster)
 		util.AddLabel(cephBlockPool, util.ForbidMirroringLabel, "true")
+		util.AddLabel(cephBlockPool, util.BuiltInLabelKey, "true")
```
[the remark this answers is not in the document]
rewantsoni: No, we could remove it

PR #3308, opened 2025-06-18
Removed unused unmaintained files

malayparida2000:


deploy/single-node-storagecluster-provider.yaml
```diff
@@ -1,119 +0,0 @@
----
```
iamniting, line 1: I believe this was added by @umangachapagain as an example. @umangachapagain Can you check if we do need this example?

hack/ticketgen/README.md
```diff
@@ -1,22 +0,0 @@
-# ticketgen
```
iamniting, line 1: @rewantsoni Can you confirm that we dont need this?

iamniting (changes requested):

hack/ticketgen/README.md
rewantsoni: ticketgen was used in previous versions and is no longer used, we can remove it now

PR #3312, opened 2025-06-19
Filter packageManifests with the catalogSource Name & Namespace

malayparida2000:
We just discovered there can be multiple catalogSources with the same name & namespace. So we need to filter the list with the catSrc Name & Namespace of the subscription to get the correct packageManifest.

### DIFF
--- a/services/provider/server/server.go
+++ b/services/provider/server/server.go
@@ -723,6 +723,7 @@ func (s *OCSProviderServer) getOCSSubscriptionChannel(ctx context.Context) (stri
 func (s *OCSProviderServer) isSubscriptionChannelValid(ctx context.Context, subscription *opv1a1.Subscription) (bool, error) {
 	installedCSVName := subscription.Status.InstalledCSV
 	channelName := subscription.Spec.Channel
+	packageName := subscription.Spec.Package
 
 	// Check if the installed CSV of the subscription is in Succeeded phase
 	if installedCSVName == "" {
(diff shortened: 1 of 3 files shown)


services/provider/server/server.go
```diff
@@ -741,17 +742,46 @@ func (s *OCSProviderServer) isSubscriptionChannelValid(ctx context.Context, subs
...
+		client.MatchingFields{"metadata.name": packageName},
+	)
 	if err != nil {
-		klog.Errorf("failed to get PackageManifest %q: %v", subscription.Spec.Package, err)
+		klog.Errorf("failed to list packageManifests for package %q: %v", packageName, err)
 		return false, err
 	}
+	// Find the packageManifest that matches the subscription's catalogSource and catalogSourceNamespace
+	var packageManifest *unstructured.Unstructured
+	for i := range packageManifestList.Items {
```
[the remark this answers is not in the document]
malayparida2000: Did this as such as these are custom labels, they might not be indexed on the k8s server. Let me try it & see the results.

services/provider/server/server.go
```diff
@@ -741,17 +742,43 @@ func (s *OCSProviderServer) isSubscriptionChannelValid(ctx context.Context, subs
...
+		if !catalogExists || catalogLabel != subscription.Spec.CatalogSource {
+			continue
+		}
+		catalogNamespaceLabel, catalogNamespaceExists := labels["catalog-namespace"]
+		if !catalogNamespaceExists || catalogNamespaceLabel != subscription.Spec.CatalogSourceNamespace {
+			continue
+		}
+		packageManifest = pm
+		break
+	}
```
nb-ohad, line 776: Suggestion only:
Given all of the other comments above
This entire body can be written more compactly without the need for a `continue` jump
 ```suggestion
	for i := range packageManifestList.Items {
		pm := &packageManifestList.Items[i]
		labels := pm.GetLabels()
		
		if pm.GetName() != packageName &&
		   	labels["catalog"] != subscription.Spec.CatalogSource && 
		    	labels["catalog-namespace"] != subscription.Spec.CatalogSourceNamespace {
		    	
		    	packageManifest = pm
			break
	        }
	}
```

File: services/provider/server/server.go [hunk 3]
```diff
@@ -741,17 +742,36 @@ func (s *OCSProviderServer) isSubscriptionChannelValid(ctx context.Context, subs
...
 		Group:   "packages.operators.coreos.com",
 		Version: "v1",
 		Kind:    "PackageManifest",
 	})
-	err = s.client.Get(ctx, client.ObjectKey{Name: subscription.Spec.Package, Namespace: s.namespace}, packageManifest)
+	err = s.client.List(ctx, packageManifestList,
+		client.InNamespace(s.namespace),
+		// Field match prone to future change, but better than processing hundreds of resources, consuming more CPU/memory
+		client.MatchingFields{"metadata.name": packageName},
+	)
```
nb-ohad, line 755: A small nit pick about formatting. If we are to split a function call into multiple lines we put each argument on its own line for reading consistency 
```suggestion
	err = s.client.List(
		ctx, 
		packageManifestList,
		client.InNamespace(s.namespace),
		// Field match prone to future change, but better than processing hundreds of resources, consuming more CPU/memory
		client.MatchingFields{"metadata.name": packageName},
	)
```

nb-ohad (changes requested):

malayparida2000:
We had a succesful run of the CI here https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/red-hat-storage_ocs-operator/3312/pull-ci-red-hat-storage-ocs-operator-main-ocs-operator-bundle-e2e-aws/1935670137050894336, Then it retriggered. Due to lack of time overriding.

PR #3313, opened 2025-06-19
controllers: handle update of provider mode internal consumer

rewantsoni:
This PR stops deploying the consumer with a name "internal" if the local client is onboarded to a consumer with a different name. We will reconcile this old consumer instead of creating a new internal consumer.

We are also updating all the Setters to set the values in the configmap if the key doesn't exist


rewantsoni:
Tested the changes and it works

controllers/storagecluster/storageconsumer.go
```diff
@@ -160,6 +189,40 @@ func (s *storageConsumer) ensureDeleted(r *StorageClusterReconciler, storageClus
...
+			if err := r.Client.List(
+				r.ctx,
+				storageConsumers,
+				client.MatchingFields{util.UIDIndexName: storageClientObj.Status.ConsumerID},
+			); err != nil {
+				return "", fmt.Errorf("failed to list the internal storageconsumers: %v", err)
+			}
+			if len(storageConsumers.Items) > 1 {
+				return "", fmt.Errorf("more than one storage consumer is available for the storage client")
+			}
```
leelavg, line 217: this'll not happen and we are just being cautious, isn't it?

controllers/storagecluster/storageconsumer.go
```diff
@@ -160,6 +189,40 @@ func (s *storageConsumer) ensureDeleted(r *StorageClusterReconciler, storageClus
...
+			if len(storageConsumers.Items) == 1 {
```
leelavg, line 218: can you club the condition and move the high chance to first if?
`if len == 1 { return first } else if len > 1 { return err }`

controllers/storagecluster/storageconsumer.go
```diff
@@ -57,39 +60,47 @@ func (s *storageConsumer) ensureCreated(r *StorageClusterReconciler, storageClus
...
+	storageConsumerObj.Namespace = storageCluster.Namespace
+
+	if _, err := controllerutil.CreateOrUpdate(r.ctx, r.Client, storageConsumerObj, func() error {
+		if err := controllerutil.SetControllerReference(storageCluster, storageConsumerObj, r.Scheme); err != nil {
 			return err
 		}
-		spec := &storageConsumer.Spec
-		spec.ResourceNameMappingConfigMap.Name = localStorageConsumerConfigMapName
+		spec := &storageConsumerObj.Spec
+		if spec.ResourceNameMappingConfigMap.Name == "" {
```
leelavg, line 77: is this check for upgrade controller could set this? if yes, both of the controllers can run in parallel, isn't it?

controllers/storagecluster/storageconsumer.go
```diff
@@ -57,39 +60,47 @@ func (s *storageConsumer) ensureCreated(r *StorageClusterReconciler, storageClus
...
-	storageConsumer.Name = defaults.LocalStorageConsumerName
-	storageConsumer.Namespace = storageCluster.Namespace
-	if _, err := controllerutil.CreateOrUpdate(r.ctx, r.Client, storageConsumer, func() error {
-		if err := controllerutil.SetControllerReference(storageCluster, storageConsumer, r.Scheme); err != nil {
+	consumerName, err := getInternalStorageConsumerName(r, storageCluster.Name, storageCluster.Namespace)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to get internal storage consumer name: %v", err)
+	}
+
+	storageConsumerObj := &ocsv1a1.StorageConsumer{}
```
leelavg, line 68: why the rename? I don't see a reason to call this out which also introduces not mandatory diff.

controllers/storagecluster/storageconsumer.go
```diff
@@ -98,26 +109,40 @@ func (s *storageConsumer) ensureCreated(r *StorageClusterReconciler, storageClus
...
+		}
+		if availableServices.CephFs {
+			resourceMap.SetSubVolumeGroupName(subVolumeGroupName)
+			resourceMap.SetSubVolumeGroupRadosNamespaceName(subVolumeGroupName)
+			resourceMap.SetCephFsClientProfileName("openshift-storage")
+		}
+		if availableServices.Nfs {
+			resourceMap.SetNfsClientProfileName("openshift-storage")
+		}
+		defaultConsumerResourceNames := util.GetStorageConsumerDefaultResourceNames(
```
leelavg, line 131: is this correct? this overwrites the desired state of upgraded internal consumer, isn't it?

leelavg (commented):
so the decision is to stop trying to rename the upgraded internal consumer?

controllers/storagecluster/storageconsumer.go
rewantsoni: It was creating confusion as we have a struct with the same name, I could preserve the original name

leelavg: ouch, you are fixing an actual bug of overshadowing type. fine.

rewantsoni:
/retest-required
Failed due to infra issue

controllers/storagecluster/storagecluster_controller.go
```diff
@@ -251,7 +262,16 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		).
 		Watches(&storagev1.StorageClass{}, enqueueStorageClusterRequest).
 		Watches(&volumesnapshotv1.VolumeSnapshotClass{}, enqueueStorageClusterRequest).
-		Watches(&ocsv1alpha1.StorageConsumer{}, enqueueStorageClusterRequest, builder.WithPredicates(storageConsumerStatusPredicate)).
+		Watches(
```
leelavg, line 265: We don't want to trigger for every consumer, in the reconcile you are placing owner and change this to own.

controllers/util/k8sutil.go
```diff
@@ -6,13 +6,12 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
+	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
```
leelavg, line 9: Observed a couple of times, pls have your formatter (just assuming) not to adjust imports incorrectly.

controllers/storagecluster/storagecluster_controller.go
rewantsoni: Every consumer is owned by the storageCluster we would still trigger reconcile for every cosumer

leelavg (commented):
LGTM.

PR #3314, opened 2025-06-19
Fix the check for provider not upgradable check

malayparida2000:
We want to not block provider upgrade when the client is by anyway ahead of the provider. But our earlier fix broke the check as even provider & client on the same version are now not upgradable. Now to simplify the logic we do a negation of the allowed conditions (provider should be equal to or less than client) for deciding if the provider is not upgradable.

### DIFF
--- a/controllers/storagecluster/reconcile.go
+++ b/controllers/storagecluster/reconcile.go
@@ -950,7 +950,7 @@ func getUnsupportedClientsCount(r *StorageClusterReconciler, namespace string) (
 		if scList.Items[idx].Status.Client != nil {
 			clientVersion, err := semver.Make(scList.Items[idx].Status.Client.OperatorVersion)
 			if err == nil {
-				if providerVersion.Major != clientVersion.Major || providerVersion.Minor <= clientVersion.Minor {
+				if providerVersion.Major != clientVersion.Major || providerVersion.Minor > clientVersion.Minor {
 					count++
 				}
 			} else {


controllers/storagecluster/reconcile.go
```diff
@@ -950,7 +950,7 @@ func getUnsupportedClientsCount(r *StorageClusterReconciler, namespace string) (
 		if scList.Items[idx].Status.Client != nil {
 			clientVersion, err := semver.Make(scList.Items[idx].Status.Client.OperatorVersion)
 			if err == nil {
-				if providerVersion.Major != clientVersion.Major || providerVersion.Minor <= clientVersion.Minor {
+				if !(providerVersion.Major == clientVersion.Major || providerVersion.Minor <= clientVersion.Minor) {
```
nb-ohad, line 953: There is a bug here, If you are to move the not (!) outside the || need to become a &&
But why not just change <= to >? It will be logically equivalent and a smaller change as
!(A && B) = !A || !B

nb-ohad (changes requested):

controllers/storagecluster/reconcile.go
leelavg, line 953:
```
!(p18 <= c17) -> !(false) -> true  -> stop  provider upgrade -> when clubbed w/ #3312 -> stop consumer upgrade as well (isn't this wrong?)
!(p18 <= c18) -> !(true)  -> false -> allow provider upgrade
!(p18 <= c19) -> !(true)  -> true  -> allow provider upgrade
```
?

PR #3332, opened 2025-06-30
controllers: update the consumer configmap default values

rewantsoni:
The PR does the following:
1. Skip default csi user creation by rook
2. Fill the internal and fresh deployment configmap with the correct values for the csi secret names

### DIFF
--- a/controllers/storagecluster/reconcile.go
+++ b/controllers/storagecluster/reconcile.go
@@ -7,8 +7,6 @@ import (
 	"strings"
 	"time"
 
-	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
-
 	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
 	ocsv1alpha1 "github.com/red-hat-storage/ocs-operator/api/v4/v1alpha1"
 	"github.com/red-hat-storage/ocs-operator/v4/controllers/util"
@@ -25,6 +23,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
--- a/controllers/storagecluster/storageconsumer.go
+++ b/controllers/storagecluster/storageconsumer.go
@@ -11,10 +11,10 @@ import (
 	ocsv1a1 "github.com/red-hat-storage/ocs-operator/api/v4/v1alpha1"
 	"github.com/red-hat-storage/ocs-operator/v4/controllers/defaults"
 	"github.com/red-hat-storage/ocs-operator/v4/controllers/util"
-	rookCephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
 
 	groupsnapapi "github.com/kubernetes-csi/external-snapshotter/client/v8/apis/volumegroupsnapshot/v1beta1"
 	snapapi "github.com/kubernetes-csi/external-snapshotter/client/v8/apis/volumesnapshot/v1"
+	rookCephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
 	corev1 "k8s.io/api/core/v1"
 	storagev1 "k8s.io/api/storage/v1"
 	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
@@ -62,13 +62,6 @@ func (s *storageConsumer) ensureCreated(r *StorageClusterReconciler, storageClus
 	volumeGroupSnapshotClassesSpec, err := getLocalVolumeGroupSnapshotClassNames(r.ctx, r.Client, storageCluster)
 	if err != nil {
 		return ctrl.Result{}, fmt.Errorf("failed to generate volumegroupsnapshotclasses list for distribution: %v", err)
-	}
-
-	backwardCompatibleInfo := &util.BackwardCompatabilityInfo{}
-	if backwardCompatibleValue, ok := storageCluster.GetAnnotations()[util.BackwardCompatabilityInfoAnnotationKey]; ok {
-		if err := json.Unmarshal([]byte(backwardCompatibleValue), backwardCompatibleInfo); err != nil {
-			return ctrl.Result{}, fmt.Errorf("invalid backwardCompatibleInfo annotation, value is not a json: %v", err)
-		}
 	}
 
 	storageConsumer := &ocsv1a1.StorageConsumer{}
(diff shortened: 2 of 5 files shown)


controllers/storagecluster/storageconsumer.go
```diff
@@ -136,15 +136,34 @@ func (s *storageConsumer) ensureCreated(r *StorageClusterReconciler, storageClus
 			availableServices,
 		)
 		resourceMap := util.WrapStorageConsumerResourceMap(data)
-		if backwardCompatibleInfo.Pre4_19InternalConsumer == "" {
+		switch backwardCompatibleInfo.Pre4_19DeploymentMode {
+		case "":
```
nb-ohad, line 140: Don't handle both compatibility mode and non compatibility mode using the same switch. One decision need to be made before the other on a separate context.
```
if !compatabilityMode {
} else {
     switch compatability mode type
}
```

controllers/util/k8sutil.go
```diff
@@ -76,10 +76,14 @@ const (
 	ForceDeletionAnnotationKey             = "ocs.openshift.io/force-deletion"
 	RookForceDeletionAnnotationKey         = "rook.io/force-deletion"
 	BackwardCompatabilityInfoAnnotationKey = "ocs.openshift.io/backward-compatability-info"
+	InternalDeploymentMode                 = "Internal"
+	ExternalDeploymentMode                 = "External"
+	ProviderDeploymentMode                 = "Provider"
```
nb-ohad, line 81: Please a separate string type for an enum

nb-ohad (changes requested):

controllers/storagecluster/external_resources_test.go
```diff
@@ -114,8 +114,9 @@ func createExternalClusterReconcilerFromCustomResources(
 		ObjectMeta: metav1.ObjectMeta{
 			Name: "ocsinit",
 			Annotations: map[string]string{
-				UninstallModeAnnotation: string(UninstallModeGraceful),
-				CleanupPolicyAnnotation: string(CleanupPolicyDelete),
+				UninstallModeAnnotation:                     string(UninstallModeGraceful),
+				CleanupPolicyAnnotation:                     string(CleanupPolicyDelete),
+				util.BackwardCompatabilityInfoAnnotationKey: string(util.JsonMustMarshal(util.BackwardCompatabilityInfo{Pre4_19DeploymentMode: util.ExternalDeploymentMode})),
```
nb-ohad, line 119: This is too long to be a single line.
```suggestion
				util.BackwardCompatabilityInfoAnnotationKey: string(
					util.JsonMustMarshal(
						util.BackwardCompatabilityInfo{
							Pre4_19DeploymentMode: util.ExternalDeploymentMode
						}
					)
				),
```

controllers/storagecluster/reconcile.go
```diff
@@ -355,9 +356,42 @@ func (r *StorageClusterReconciler) reconcilePhases(
...
+				return ctrl.Result{}, fmt.Errorf("failed to get pre4_19 subvolumegroup cr: %v", err)
+			}
+			if svgPre4_19.UID != "" {
+				updateRequired = util.AddAnnotation(
+					instance,
+					util.BackwardCompatabilityInfoAnnotationKey,
+					string(util.JsonMustMarshal(util.BackwardCompatabilityInfo{Pre4_19DeploymentMode: util.InternalDeploymentMode})),
+				)
+			}
+		}
```
nb-ohad, line 392: Can you please separate the calculation of the backward compatibility struct into a separate function?

controllers/storagecluster/storageconsumer.go
```diff
@@ -136,15 +136,36 @@ func (s *storageConsumer) ensureCreated(r *StorageClusterReconciler, storageClus
 			availableServices,
 		)
 		resourceMap := util.WrapStorageConsumerResourceMap(data)
-		if backwardCompatibleInfo.Pre4_19InternalConsumer == "" {
+		if backwardCompatibleInfo.Pre4_19DeploymentMode == "" {
```
nb-ohad, line 139: This is wrong, if you are not in backward compatibility mode then you should not even look at this struct.

controllers/util/k8sutil.go
```diff
@@ -65,21 +65,27 @@ const (
...
+	OcsClientTimeout                                      = 10 * time.Second
+	StorageClientMappingConfigName                        = "storage-client-mapping"
+	ExternalClassLabelKey                                 = "storageclass.ocs.openshift.io/is-external"
+	StorageConsumerMirroringInfoAnnotation                = "ocs.openshift.io/consumer-mirroring-info"
+	ForceDeletionAnnotationKey                            = "ocs.openshift.io/force-deletion"
+	RookForceDeletionAnnotationKey                        = "rook.io/force-deletion"
+	BackwardCompatabilityInfoAnnotationKey                = "ocs.openshift.io/backward-compatability-info"
+	InternalDeploymentMode                 DeploymentType = "Internal"
+	ExternalDeploymentMode                 DeploymentType = "External"
+	ProviderDeploymentMode                 DeploymentType = "Provider"
```
nb-ohad, line 81: This should go into its own const section. This is the convention that is used for enum like string in golang

nb-ohad (changes requested):

controllers/storagecluster/reconcile.go
```diff
@@ -838,6 +853,25 @@ func (r *StorageClusterReconciler) ownStorageConsumersInNamespace(instance *ocsv
...
+func (r *StorageClusterReconciler) calculateBackwardCompatibilityMode(storageCluster *ocsv1.StorageCluster) (*util.BackwardCompatabilityInfo, error) {
+	if storageCluster.Spec.AllowRemoteStorageConsumers {
+		return &util.BackwardCompatabilityInfo{Pre4_19DeploymentMode: util.ProviderDeploymentMode}, nil
+	} else if storageCluster.Spec.ExternalStorage.Enable {
+		return &util.BackwardCompatabilityInfo{Pre4_19DeploymentMode: util.ExternalDeploymentMode}, nil
+	} else {
+		svgPre4_19 := &rookCephv1.CephFilesystemSubVolumeGroup{}
+		svgPre4_19.Name = fmt.Sprintf("%s-%s", util.GenerateNameForCephFilesystem(storageCluster.Name), subVolumeGroupName)
+		svgPre4_19.Namespace = storageCluster.Namespace
+		if err := r.Get(r.ctx, client.ObjectKeyFromObject(svgPre4_19), svgPre4_19); client.IgnoreNotFound(err) != nil {
```
leelavg, line 865: this heuristic is time dependent and when storageconsumer unit in cluster controller is reconciled, pre4.19 svg will be marked for deletion then in an upgraded internal cluster we always send `nil, nil`, is that what we want?

controllers/storagecluster/reconcile.go
```diff
@@ -355,9 +355,24 @@ func (r *StorageClusterReconciler) reconcilePhases(
...
+			updateRequired = true
+		}
+
+		if backwardCompatibilityMode, err := r.calculateBackwardCompatibilityMode(instance); err != nil {
+			return reconcile.Result{}, err
+		} else if backwardCompatibilityMode != nil {
+			updateRequired = util.AddAnnotation(
+				instance,
+				util.BackwardCompatabilityInfoAnnotationKey,
+				string(util.JsonMustMarshal(&backwardCompatibilityMode)),
```
leelavg, line 371: a bit confused, weren't we already adding the same key w/ a different value in upgrade controller for tracking internal consumer id in provider mode?

PR #3336, opened 2025-06-30
get the cephclient based on generation during desiredstate rpc

leelavg:
consumer configmap now stores name of the secret that needs to be sent
to client rather than the name of cephclient.

### DIFF
--- a/services/provider/server/server.go
+++ b/services/provider/server/server.go
@@ -1410,29 +1410,26 @@ func (s *OCSProviderServer) appendCephClientSecretKubeResources(
 	consumerConfig util.StorageConsumerResources,
 ) ([]client.Object, error) {
 
-	cephClients := []string{}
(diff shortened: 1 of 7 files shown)


services/provider/server/server.go
```diff
@@ -1410,29 +1410,26 @@ func (s *OCSProviderServer) appendCephClientSecretKubeResources(
...
 	}
 
-	for i := range cephClients {
+	for cephClientName, secretName := range cephClientsToSecrets {
+		if secretName == "" {
+			continue
+		}
 		cephClient := &rookCephv1.CephClient{}
-		cephClient.Name = cephClients[i]
+		cephClient.Name = cephClientName
```
rewantsoni, line 1432: Rook is generating the secret with the same name that we have for the cephClients. Do we need to get the ceph clients to get the secret?

nb-ohad (changes requested):

PR #3339, opened 2025-07-01
Add default requests and limits for ocs-operator pods

mrudraia1:
This PR adds the default requests and limits for the ocs-operator pods.
The default metrics are set for the following pods
- ocs-operator
- ocs-metircs-exporter
- ux-backend
- crash-collector
- log-collector

### DIFF
--- a/controllers/defaults/resources.go
+++ b/controllers/defaults/resources.go
@@ -66,18 +66,32 @@ var (
 		},
 		"ocs-metrics-exporter": {
 			Requests: corev1.ResourceList{
-				"memory": resource.MustParse("50Mi"),
-				"cpu":    resource.MustParse("50m"),
+				"memory": resource.MustParse("512Mi"),
+				"cpu":    resource.MustParse("250m"),
+			},
+			Limits: corev1.ResourceList{
+				"memory": resource.MustParse("1.5Gi"),
+				"cpu":    resource.MustParse("1"),
 			},
 		},
 		"crashcollector": {
 			Requests: corev1.ResourceList{
-				corev1.ResourceCPU:    resource.MustParse("0.1"),
-				corev1.ResourceMemory: resource.MustParse("60Mi"),
+				corev1.ResourceCPU:    resource.MustParse("50m"),
+				corev1.ResourceMemory: resource.MustParse("50Mi"),
 			},
 			Limits: corev1.ResourceList{
-				corev1.ResourceCPU:    resource.MustParse("0.1"),
-				corev1.ResourceMemory: resource.MustParse("60Mi"),
+				corev1.ResourceCPU:    resource.MustParse("100m"),
+				corev1.ResourceMemory: resource.MustParse("100Mi"),
+			},
+		},
+		"logcollector": {
+			Requests: corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("50m"),
+				corev1.ResourceMemory: resource.MustParse("50Mi"),
+			},
+			Limits: corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("100m"),
+				corev1.ResourceMemory: resource.MustParse("250Mi"),
 			},
 		},
 		"exporter": {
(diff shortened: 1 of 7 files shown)


controllers/defaults/resources.go
```diff
@@ -66,18 +66,32 @@ var (
...
 		"ocs-metrics-exporter": {
 			Requests: corev1.ResourceList{
-				"memory": resource.MustParse("50Mi"),
-				"cpu":    resource.MustParse("50m"),
+				"memory": resource.MustParse("512Mi"),
+				"cpu":    resource.MustParse("250m"),
+			},
+			Limits: corev1.ResourceList{
+				"memory": resource.MustParse("1.5Gi"),
+				"cpu":    resource.MustParse("1m"),
```
leelavg, line 74: 1m?

controllers/defaults/resources.go
```diff
@@ -66,18 +66,32 @@ var (
 		},
 		"ocs-metrics-exporter": {
 			Requests: corev1.ResourceList{
-				"memory": resource.MustParse("50Mi"),
-				"cpu":    resource.MustParse("50m"),
+				"memory": resource.MustParse("512Mi"),
+				"cpu":    resource.MustParse("250m"),
+			},
+			Limits: corev1.ResourceList{
+				"memory": resource.MustParse("1.5Gi"),
```
leelavg, line 73: Why does the exporter go till 1.5Gi, seems a lot.

Nikhil-Ladha: We have seen memory consumption reaching till `1.25Gi` for Scale tests ran during testing, therefore setting a safe limit of `1.5Gi`

leelavg: Need a bit more investigation, I understand even w/o this PR the exporter has no bounds but we better take this moment to re-look at the exporter and come to a conclusion for bumping the values.

Nikhil-Ladha: I would suggest to take this PR in and work on parallel to identify any memory leaks or such issues that could lead to a memory spike in exporter.

controllers/defaults/resources.go
mrudraia1: changed to "1"

metrics/vendor/github.com/red-hat-storage/ocs-operator/v4/controllers/defaults/resources.go
```diff
@@ -66,18 +66,32 @@ var (
 		},
 		"ocs-metrics-exporter": {
 			Requests: corev1.ResourceList{
-				"memory": resource.MustParse("50Mi"),
-				"cpu":    resource.MustParse("50m"),
+				"memory": resource.MustParse("512Mi"),
+				"cpu":    resource.MustParse("250m"),
+			},
+			Limits: corev1.ResourceList{
+				"memory": resource.MustParse("1.5Gi"),
```
Madhu-1, line 73: Are there any ways to modify these resource when customer gets into problem? If we dont have a way setting the limits can cause problem in some customer(where they have more number of pods/pvc/nodes etc)

Nikhil-Ladha: Yes, there are ways to modify the resources for such pods via the [storagecluster.spec.resources](https://github.com/red-hat-storage/ocs-operator/blob/main/api/v1/storagecluster_types.go#L47) field.
So, if customer runs into issues with these values they can modify/update as per their needs.

controllers/defaults/resources.go
```diff
@@ -66,18 +66,32 @@ var (
...
+		"logcollector": {
+			Requests: corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("50m"),
+				corev1.ResourceMemory: resource.MustParse("50Mi"),
 			},
 			Limits: corev1.ResourceList{
-				corev1.ResourceCPU:    resource.MustParse("0.1"),
-				corev1.ResourceMemory: resource.MustParse("60Mi"),
+				corev1.ResourceCPU:    resource.MustParse("100m"),
+				corev1.ResourceMemory: resource.MustParse("250Gi"),
```
Madhu-1, line 94: why do we need 250Gi of the limit?  isnt this too much?

PR #3343, opened 2025-07-01
Add cache folder to gitignore & remove the folder

malayparida2000:


leelavg:
Can you also create .dockerignore file and start w/ ignoring bin dir etc..?

malayparida2000:
> Can you also create .dockerignore file and start w/ ignoring bin dir etc..?

Nice idea. Will raise another PR for it.

leelavg:
Then, why do we even need the CI for this PR? it's just adding more pressure to other PRs competing for cloud resources.

malayparida2000:
The cache folder was coming when operator-sdk bundle run the CI command was being used.  So just want to make sure the removal of the folder doesn't break CI flow.

leelavg:
Even then you don't need an OCP server from CI to test the hypothesis. Anyways, will leave you to it and at the same time pls don't override now as there are PRs required for next RC are running in CI.

malayparida2000:
> Even then you don't need an OCP server from CI to test the hypothesis. Anyways, will leave you to it and at the same time pls don't override now as there are PRs required for next RC are running in CI.

If PRs for RC are pending I will wait for few days before taking in this PR as this is not urgent

iamniting (approved):

PR #3344, opened 2025-07-02
util: add kernel version checks libraries

iamniting:
These funcs allow user to check the kernel support on ocp cluster

https://issues.redhat.com/browse/RHSTOR-7293

### DIFF
--- a/metrics/vendor/github.com/red-hat-storage/ocs-operator/v4/controllers/util/kernel_version.go
+++ b/metrics/vendor/github.com/red-hat-storage/ocs-operator/v4/controllers/util/kernel_version.go
@@ -1,0 +1,104 @@
+package util
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	semver "github.com/blang/semver/v4"
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type kernelSupportVersion struct {
+	// Latest is the kernel version in which the issue was fixed downstream.
+	// Any version greater than this is considered to have the fix.
+	Latest string
+
+	// Backports contains base kernel version prefixes that received backported fixes.
+	// Matching logic: match the first 4 version segments, and the 5th segment should be greater
+	// than or equal to the backported entry to consider the fix as present.
+	Backports []string
+}
+
+var (
+	CephxKeyRotaionKernelSupportVersion = kernelSupportVersion{
+		Latest:    "5.14.0-570.22.1",
+		Backports: []string{"5.14.0-570.22.1"},
+	}
+)
+
+func GetKernelVersionFromAllNodes(ctx context.Context, cli client.Client) ([]string, error) {
+
+	var nodeList corev1.NodeList
+	if err := cli.List(ctx, &nodeList); err != nil {
+		return nil, fmt.Errorf("listing nodes: %w", err)
+	}
+
+	var kernelVersions []string
+	var kernelVersionsMap map[string]bool = make(map[string]bool)
+	for _, node := range nodeList.Items {
+		// Check if the node has a kernel version
+		if node.Status.NodeInfo.KernelVersion == "" {
+			return nil, fmt.Errorf("node %s has empty kernel version", node.Name)
+		}
+
+		if _, ok := kernelVersionsMap[node.Status.NodeInfo.KernelVersion]; !ok {
+			kernelVersionsMap[node.Status.NodeInfo.KernelVersion] = true
+			// Append the kernel version to the slice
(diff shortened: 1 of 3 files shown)


controllers/util/version.go
```diff
@@ -0,0 +1,125 @@
...
+	semver "github.com/blang/semver/v4"
+	configv1 "github.com/openshift/api/config/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	// TODO: update the fixed in kernel versions
+	// supported kernel version corresponding to the OCP version
```
nixpanic, line 17: Why would you need to know/check/use the OCP version?

controllers/util/version.go
```diff
@@ -0,0 +1,125 @@
...
+	CephxKeyRotaionKernelSupportMatrix = map[string]string{
+		"4.14": "5.14.0-570.22.1.el9_6.x86_64",
```
nixpanic, line 19: The last parts of these versions are not interesting, `.el9_6` and `.x86_64` can be dropped from the version comparing check. RHEL kernels always have those last two parts.

`570.22.1` is not guaranteed to be only numeric. Hotfix kernels often refer to a particular bug by having `.bz123456` or similar in it.

nixpanic (commented):
A unit test that shows how these functions work together would be very welcome.

controllers/util/version.go
iamniting: I will drop this `el9_6.x86_64`. We should be able to compare even in the case of a hotfix and all.

controllers/util/version.go
```diff
@@ -0,0 +1,104 @@
...
+
+	// Backports contains base kernel version prefixes that received backported fixes.
+	// Matching logic: match the first 4 version segments, and the 5th segment should be greater
+	// than or equal to the backported entry to consider the fix as present.
+	Backports []string
+}
+
+var (
+	CephxKeyRotaionKernelSupportVersion = kernelSupportVersion{
+		Latest:    "5.14.0-570.22.1",
```
nixpanic, line 26: The comment for `Latest` refers to _upstream_, this does not look like an upstream version. Does `semver.Make()` work with a version like `6.16.0`?

nixpanic (commented):
This version checking is pretty much RHEL specific, this is fine but could be noted somewhere.

Added one question, otherwise it looks good to me, thanks.

iamniting:
> This version checking is pretty much RHEL specific, this is fine but could be noted somewhere.
> 
> Added one question, otherwise it looks good to me, thanks.

As far as we are concerned with rhel and coreos. As far as I remember, we discussed that it is going to be the same for both. Pls correct me if I am wrong.

controllers/util/version.go
iamniting: Can we resolve this?

controllers/util/version.go
nixpanic: not any more, looks good now.

controllers/util/version.go
leelavg, line 1: better be explicit in the filename that we are dealing w/ kernel version as we already have a version folder at the root and during imports it helps to distinguish?

controllers/util/version.go
```diff
@@ -0,0 +1,104 @@
+package util
```
leelavg, line 1: can you add a sample unit test on how you intend that one should use these util funcs? isn't the intention just to export a single func whether I can rotate the ceph key or not?

rohan47:
@iamniting @leelavg ocs-client-operator will also require this functionality to send the kernel version to the provider,
will it make sense to move this changes to provider api? 
else this will have to be copied to ocs-client-operator as well

iamniting:
> @iamniting @leelavg ocs-client-operator will also require this functionality to send the kernel version of nodes to the provider, will it make sense to move this changes to provider api? else this will have to be copied to ocs-client-operator as well

You mean under `services/provider/api` as a new directory

rohan47:
> > @iamniting @leelavg ocs-client-operator will also require this functionality to send the kernel version of nodes to the provider, will it make sense to move this changes to provider api? else this will have to be copied to ocs-client-operator as well
> 
> You mean under `services/provider/api` as a new directory

yes

metrics/vendor/github.com/red-hat-storage/ocs-operator/v4/controllers/util/kernel_version.go
rohan47, line 1: there is no vendor for metrics, so this file should be removed right?

rohan47 (changes requested):

nixpanic:
Note that there is also https://pkg.go.dev/github.com/ceph/ceph-csi/pkg/util/kernel which you could consider using.

PR #3349, opened 2025-07-02
Add API support to forcefully cleanup OSD disks

rohan47:
Add API support to forcefully cleanup OSD disks belonging to other clusters.
This is usefull in scenarios where storagecluster was reinstalled and osd disks still contain the metadata from the previous install.


api/v1/storagecluster_types.go
```diff
@@ -211,6 +211,11 @@ type ManageCephCluster struct {
 
 	// Ceph Config options
 	CephConfig map[string]map[string]string `json:"cephConfig,omitempty"`
+
+	// WipeDevicesFromOtherClusters wipes the OSD disks belonging to other clusters. This is useful in scenarios where storage cluster
```
malayparida2000, line 215: > OSD disks belonging to other clusters

Did you mean the previous install here?

api/v1/storagecluster_types.go
```diff
@@ -211,6 +211,11 @@ type ManageCephCluster struct {
...
+	// was reinstalled but OSD disk still contains the metadata from previous storage cluster.
+	// +optional
+	WipeDevicesFromOtherClusters bool `json:"wipeDevicesFromOtherClusters,omitempty"`
```
malayparida2000, line 218: This setting would be better to keep under a subsection CleanUpPolicy, As the CleanUpPolicy section contains few other values as well. A better thing would be don't expose just this one field instead Put the CleanUpPolicy spec here and pass it directly to the cephCluster if defined. This would help in future expansion.

controllers/storagecluster/cephcluster.go
```diff
@@ -654,6 +654,12 @@ func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, kmsConfigMap *co
 	cephCluster.Spec.Security.KeyRotation.Enabled = isEnabled
 	cephCluster.Spec.Security.KeyRotation.Schedule = rotationSchedule
 
+	if sc.Spec.ManagedResources.CephCluster.WipeDevicesFromOtherClusters {
+		cephCluster.Spec.CleanupPolicy = rookCephv1.CleanupPolicySpec{
+			WipeDevicesFromOtherClusters: sc.Spec.ManagedResources.CephCluster.WipeDevicesFromOtherClusters,
```
malayparida2000, line 659: As suggested before it might be better to expose & pass the whole CleanUpPolicy.

malayparida2000 (changes requested):
The name WipeDevicesFromOtherClusters is coming from rook it seems, but the naming looks a bit confusing to me as it's not other cluster but previous installation actually. Other cluster can have confusing meaning as we have multiple storageCluster support too.

PR #3352, opened 2025-07-03
controller: consumer should own all cephclients

rewantsoni:
storage consumer should own all the Ceph clients, new as well as the ones that are being rotated. 
Use the owner ref to find the ceph clients being owned by the storage request in upgrade controller to own the cephClients from the storage consumer

Fixes: https://issues.redhat.com/browse/DFBUGS-2924

### DIFF
--- a/controllers/storageconsumer/storageconsumer_upgrade_controller.go
+++ b/controllers/storageconsumer/storageconsumer_upgrade_controller.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"crypto/md5"
 	"encoding/hex"
-	"encoding/json"
 	"fmt"
 	"slices"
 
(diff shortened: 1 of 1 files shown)


leelavg (approved):

controllers/storageconsumer/storageconsumer_upgrade_controller.go
```diff
@@ -244,6 +234,10 @@ func (r *StorageConsumerUpgradeReconciler) reconcileStorageRequest(
 	if err := r.Client.Get(ctx, client.ObjectKeyFromObject(rbdStorageRequest), rbdStorageRequest); client.IgnoreNotFound(err) != nil {
 		return err
 	} else if rbdStorageRequest.UID != "" {
+		if err := r.reconcilePre4_19CephClients(ctx, rbdStorageRequest.UID, storageConsumer); err != nil {
+			return err
+		}
+
```
nb-ohad, line 240: Why did we move the code around? wouldn't this call make more sense in the same position of the original code?

controllers/storageconsumer/storageconsumer_upgrade_controller.go
```diff
@@ -265,32 +259,17 @@ func (r *StorageConsumerUpgradeReconciler) reconcileStorageRequest(
...
 	cephFsStorageRequest := &metav1.PartialObjectMetadata{}
 	cephFsStorageRequest.SetGroupVersionKind(ocsv1alpha1.GroupVersion.WithKind("StorageRequest"))
 	cephFsStorageRequest.Name = cephFsStorageRequestName
 	cephFsStorageRequest.Namespace = storageConsumer.Namespace
 	if err := r.Client.Get(ctx, client.ObjectKeyFromObject(cephFsStorageRequest), cephFsStorageRequest); client.IgnoreNotFound(err) != nil {
 		return err
 	} else if cephFsStorageRequest.UID != "" {
+		if err := r.reconcilePre4_19CephClients(ctx, cephFsStorageRequest.UID, storageConsumer); err != nil {
+			return err
+		}
```
nb-ohad, line 271: Why did we move the code around? wouldn't this call make more sense in the same position of the original code?

controllers/storageconsumer/storageconsumer_upgrade_controller.go
rewantsoni: I had to move the code around as I need the storageRequest UID to list the ceph clients owned by the storage request

PR #3360, opened 2025-07-04
Don't set replication spec on pools if erasure coded spec are present

malayparida2000:
Previously ocs-operator enforced all pools to have a replication spec, which was preventing erasure coded pools from being created. Now if erasure coded spec is present, it will not set the replication spec.

### DIFF
--- a/controllers/storagecluster/cephcluster.go
+++ b/controllers/storagecluster/cephcluster.go
@@ -1505,14 +1505,17 @@ func setDefaultMetadataPoolSpec(poolSpec *rookCephv1.PoolSpec, sc *ocsv1.Storage
 	if poolSpec.FailureDomain == "" {
 		poolSpec.FailureDomain = getFailureDomain(sc)
 	}
-	// Set default replication settings if necessary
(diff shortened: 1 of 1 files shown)


malayparida2000:
> The changes look good. What about adding unit tests to confirm the ec or replication settings are preserved?

Right, Better to write unit tests for all the conditions these setPools are handling.
One question, In the erasurecoded spec, do we need to supply any defaults with this function?
What happens if someone has defined only coding chunks & no data chunks & vice versa as well.

travisn:
> > The changes look good. What about adding unit tests to confirm the ec or replication settings are preserved?
> 
> Right, Better to write unit tests for all the conditions these setPools are handling. One question, In the erasurecoded spec, do we need to supply any defaults with this function? What happens if someone has defined only coding chunks & no data chunks & vice versa as well.

Let's not attempt to set defaults. If they set the erasure coding, they will need to be specific about the coding and data chunks. If one of them is 0, it would be an error. In Rook we need to add a CEL rule for this case, but I would suggest no checking needed for that in OCS operator.

malayparida2000:
@travisn Added the unit tests, PTAL

PR #3374, opened 2025-07-15
Add controller publish secrets to default RBD/CephFS StorageClasses

iPraveenParihar:
Ceph-CSI will be implementing `ControllerPublishVolume()`/`ControllerUnpublishVolume()` which requires
controller-publish-secret from StorageClass. The secret is needed to access the Ceph cluster for [add|remove] metadata/blocklist operations.

The following parameters to be added in default RBD/CephFS StorageClasses created by ocs-operator.

```
csi.storage.k8s.io/controller-publish-secret-name
csi.storage.k8s.io/controller-publish-secret-namespace
```

Story: https://issues.redhat.com/browse/RHSTOR-7598

### DIFF
--- a/controllers/util/storageclasses.go
+++ b/controllers/util/storageclasses.go
@@ -84,7 +84,6 @@ func NewDefaultRbdStorageClass(
 	storageId string,
 	isDefaultStorageClass bool,
 ) *storagev1.StorageClass {
-
 	sc := &storagev1.StorageClass{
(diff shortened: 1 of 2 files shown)


Madhu-1 (commented):
No harm in setting this one. 

@iPraveenParihar if we set it in the Clientprofile CR, we dont need this change right?  what is the benefit of setting it in both places?

@nb-ohad WDYT?

PR #3376, opened 2025-07-17
set ceph daemons key rotation policy

iamniting:

### DIFF
--- a/controllers/storagecluster/cephcluster.go
+++ b/controllers/storagecluster/cephcluster.go
@@ -55,6 +55,10 @@ const (
 	poolTypeMetadata = "metadata"
 )
 
+const (
+	desiredCephxKeyGenEnvVarName = "DESIRED_CEPHX_KEY_GEN"
+)
+
 type knownDiskType struct {
 	speed            diskSpeed
 	provisioner      StorageClassProvisionerType
@@ -259,9 +263,15 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
 					return reconcile.Result{}, err
 				}
 			}
-			cephCluster = newCephCluster(sc, r.images.Ceph, kmsConfigMap, r.Log)
+			cephCluster, err = newCephCluster(sc, r.images.Ceph, kmsConfigMap, r.Log)
+			if err != nil {
+				return reconcile.Result{}, err
+			}
 		} else {
-			cephCluster = newCephCluster(sc, r.images.Ceph, nil, r.Log)
+			cephCluster, err = newCephCluster(sc, r.images.Ceph, nil, r.Log)
+			if err != nil {
+				return reconcile.Result{}, err
+			}
 		}
 	}
 
@@ -479,7 +489,7 @@ func getCephClusterMonitoringLabels(sc ocsv1.StorageCluster) map[string]string {
 }
 
 // newCephCluster returns a CephCluster object.
-func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, kmsConfigMap *corev1.ConfigMap, reqLogger logr.Logger) *rookCephv1.CephCluster {
+func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, kmsConfigMap *corev1.ConfigMap, reqLogger logr.Logger) (*rookCephv1.CephCluster, error) {
 	labels := map[string]string{
 		"app": sc.Name,
 	}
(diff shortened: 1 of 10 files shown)


controllers/storagecluster/cephcluster.go
```diff
@@ -53,6 +54,8 @@ const (
 const (
 	poolTypeData     = "data"
 	poolTypeMetadata = "metadata"
+
+	keyRotationCounterEnvVarName = "KEY_ROTATION_COUNTER"
```
rewantsoni, line 58: Maybe a better name like: CEPH_USER_CURR_GEN

controllers/storagecluster/cephcluster.go
```diff
@@ -562,9 +565,28 @@ func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, kmsConfigMap *co
...
 
+	KeyGenerationValue, ok := os.LookupEnv(keyRotationCounterEnvVarName)
+	if !ok {
+		panic("KEY_ROTATION_COUNTER env variable is not set")
+	}
+
+	KeyGeneration, err := strconv.Atoi(KeyGenerationValue)
+	if err != nil {
+		panic("can not convert KEY_ROTATION_COUNTER value to a integer")
+	}
```
rewantsoni, line 586: We could have a util for this as this is going to be used in multiple places

controllers/storagecluster/cephcluster_test.go
```diff
@@ -26,6 +27,10 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 )
 
+func init() {
+	os.Setenv(keyRotationCounterEnvVarName, "2")
```
rewantsoni, line 31: From what I remember, there are various other env variables we are setting for tests. Any reason to set this differently from the others?

rewantsoni (changes requested):

controllers/storagecluster/cephcluster.go
iamniting: Done

controllers/storagecluster/cephcluster.go
iamniting: Lets hear from others, I believe we can come up with something better.

nb-ohad: @iamniting This is less about rotation and about identifying the desired generation (the same term we use everywhere, even in rook APIs). So I agree with @rewantsoni but will offer a small change:
`DESIRED_CEPH_USER_GEN`

controllers/storagecluster/cephcluster.go
```diff
@@ -562,6 +566,14 @@ func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, kmsConfigMap *co
 			// if resource profile change is in progress, then set this flag to false
 			ContinueUpgradeAfterChecksEvenIfNotHealthy: sc.Spec.ResourceProfile == sc.Status.LastAppliedResourceProfile,
 			CephConfig: getCephClusterCephConfig(sc),
+			Security: rookCephv1.ClusterSecuritySpec{
+				CephX: rookCephv1.ClusterCephxConfig{
+					Daemon: rookCephv1.CephxConfig{
+						KeyRotationPolicy: rookCephv1.KeyGenerationCephxKeyRotationPolicy,
+						KeyGeneration:     uint32(util.MustGetEnvInt(keyRotationCounterEnvVarName)),
```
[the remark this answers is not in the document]
iamniting: The external source is nothing but set by us only. To catch such build errors we should panic.

controllers/storagecluster/initialization_reconciler_test.go
```diff
@@ -270,6 +270,8 @@ func initStorageClusterResourceCreateUpdateTest(t *testing.T, runtimeObjs []clie
 
 	err := os.Setenv("OPERATOR_NAMESPACE", cr.Namespace)
 	assert.NoError(t, err)
+	err = os.Setenv(keyRotationCounterEnvVarName, "2")
```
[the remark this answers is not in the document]
iamniting: Changes in the test files are not reflected in the actual binary. However, I made the change.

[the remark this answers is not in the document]
Madhu-1: if you still prefer to use os.setenv there should be a cleanup as well, That is the reason to have t.setenv exists today for testing.

controllers/storagecluster/cephcluster.go
```diff
@@ -55,6 +55,10 @@ const (
 	poolTypeMetadata = "metadata"
 )
 
+const (
+	desiredCephxKeyGenEnvVarName = "DESIRED_CEPH_X_KEY_GEN"
```
nb-ohad, line 59: Just a nit pick about naming conventions: `X` is either part of the Ceph X or its own word. So you need to change to one of the followings:
```suggestion
	desiredCephXKeyGenEnvVarName = "DESIRED_CEPH_X_KEY_GEN"
```
or
```suggestion
	desiredCephxKeyGenEnvVarName = "DESIRED_CEPHX_KEY_GEN"
```

I prefer the first but will accept either

controllers/storagecluster/cephcluster.go
```diff
@@ -492,6 +496,21 @@ func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, kmsConfigMap *co
 		MaxLogSize:  &maxLogSize,
 	}
 
+	strVal := util.MustGetEnv(desiredCephxKeyGenEnvVarName)
+	intVal, err := strconv.Atoi(strVal)
```
nb-ohad, line 500: These vars are in function scope so we need to give them proper names
```suggestion
	desiredCephxKeyGenAsString := util.MustGetEnv(desiredCephxKeyGenEnvVarName)
	desiredCephxKeyGen, err := strconv.Atoi(strVal)
```

controllers/storagecluster/cephcluster.go
```diff
@@ -492,6 +496,21 @@ func newCephCluster(sc *ocsv1.StorageCluster, cephImage string, kmsConfigMap *co
...
+	if err != nil {
+		panic(fmt.Sprintf("could not convert the value %q of env var %q", strVal, desiredCephxKeyGenEnvVarName))
```
nb-ohad, line 502: As stated in the prev PR please return an error as this does not reflect a coding error (a build time assertion) but a run time assertion

controllers/storagecluster/initialization_reconciler_test.go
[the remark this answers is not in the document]
iamniting: I made the changes to all the tests where the func was being called.

PR #3378, opened 2025-07-18
metrics: add tini to prevent zombie processes in metrics-exporter

weirdwiz:
The ocs-metrics-exporter creates zombie processes when executing `ceph` and `rbd` commands to collect metrics. While we use `CombinedOutput()` which should wait for process completion, zombies are still accumulating in some customer environments. Adding tini as a minimal init system (via the ENTRYPOINT) ensures all zombie processes are properly reaped regardless of how they're created.

https://issues.redhat.com/browse/DFBUGS-2875

### DIFF
--- a/metrics/Dockerfile
+++ b/metrics/Dockerfile
@@ -14,10 +14,15 @@ RUN go build -ldflags "$LDFLAGS" -tags netgo,osusergo -o metrics-exporter main.g
 
 FROM quay.io/ceph/ceph:v18
 
+# Install tini to reap zombie processes created by ceph/rbd commands
+USER root
+RUN dnf install -y tini && \
+    dnf clean all
+
 COPY --from=builder workspace/metrics-exporter /usr/local/bin/metrics-exporter
 
 RUN chmod +x /usr/local/bin/metrics-exporter
 
 USER 2024
 
-ENTRYPOINT ["/usr/local/bin/metrics-exporter"]
+ENTRYPOINT ["/usr/bin/tini", "--", "/usr/local/bin/metrics-exporter"]
--- a/controllers/storagecluster/exporter.go
+++ b/controllers/storagecluster/exporter.go
@@ -457,7 +457,7 @@ func deployMetricsExporter(ctx context.Context, r *StorageClusterReconciler, ins
 							"--namespaces", instance.Namespace,
 							"--ceph-auth-namespace", r.OperatorNamespace,
 						},
-						Command: []string{"/usr/local/bin/metrics-exporter"},
+						Command: []string{"/usr/bin/tini", "--", "/usr/local/bin/metrics-exporter"},
 						Image:   r.images.OCSMetricsExporter,
 						Name:    metricsExporterName,
 						LivenessProbe: &corev1.Probe{
--- a/metrics/deploy/deployment.yaml
+++ b/metrics/deploy/deployment.yaml
@@ -110,6 +110,8 @@ spec:
         image: quay.io/ocs-dev/ocs-operator
         name: ocs-metrics-exporter
         command:
+        - /usr/bin/tini
+        - --
         - /usr/local/bin/metrics-exporter
         args:
           - '--namespaces'


weirdwiz:
ptal @umangachapagain @aruniiird

weirdwiz:
docker's `--init` flag also uses tini as an init process to handle zombies

idryomov (commented):
> While we use `CombinedOutput()` which should wait for process completion, zombies are still accumulating in some customer environments.

Do you have any details as to what those customer environments look like?  I noticed that metrics-exporter sets a 30 second timeout for all `execCommand()` invocations -- is there evidence of those timeouts expiring?
```
func execCommand(command string, args []string, timeout int) ([]byte, error) {
	var cancel context.CancelFunc
	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(timeout)*time.Second)
	defer cancel()
	cmd := exec.CommandContext(ctx, command, args...)

	output, err := cmd.CombinedOutput()
        ...
```
I'm asking because when `WithTimeout` is applied to `CommandContext`,  SIGKILL is sent only to that particular process.  If `command` happens to spawn child processes, they normally remain unaffected.

metrics/Dockerfile
```diff
@@ -14,10 +14,15 @@ RUN go build -ldflags "$LDFLAGS" -tags netgo,osusergo -o metrics-exporter main.g
 
 FROM quay.io/ceph/ceph:v18
 
+# Install tini to reap zombie processes created by ceph/rbd commands
+USER root
+RUN dnf install -y tini && \
+    dnf clean all
```
BlaineEXE, line 20: The ceph containers built by RHCS don't have `dnf` available. They use `microdnf` instead, and `tini` isn't an installable package. This might break when using RHCS images unless the build team has a different dockerfile for this

BlaineEXE: As an alternative, I would suggest installing `tini-static` (tini built as a static binary) in a new builder stage and then copying `/usr/bin/tini-static` to the final container from the build stage

idryomov: Before going through the motions with tini, I'd suggest taking a harder look at whether there is a need to have a process reaper in place at all.  If the issue occurs only when the timeout expires, it would likely be easier to fix the respective code in `execCommand()`.

leelavg:
@weirdwiz how did you reproduce this issue? If the subprocess is becoming zombie and you are using tini only to reap them when they get re-parented, it maybe possible to just do `cmd.Wait()` on it (we can also disable sigchld but it may effect our whole parent process)?

As Ilya mentioned, if the subprocess is re-forking you may end up w/ orphan's. Not sure how this can be done in Go, usual way would be during exec do a `setsid` and kill the group (ie, negative pid) and wait, when done/timedout which will result in no zombie and no orphan.

So, if you can provide the steps on how you reproduced the issue & the leaks observed, it'll be very helpful.

thanks!

weirdwiz:
> @weirdwiz how did you reproduce this issue? If the subprocess is becoming zombie and you are using tini only to reap them when they get re-parented, it maybe possible to just do cmd.Wait() on it (we can also disable sigchld but it may effect our whole parent process)?

I was not able to reproduce, it completely in a dev env. And tested out the reap with manually creating zombies.

I also suspect like @idryomov that sub process is re-forking and we're exiting out, but there was not enough hard evidence in the logs (the process exited out with -1), so I thought that the safest way to make sure we don't run into it is to setup `tini`

>As Ilya mentioned, if the subprocess is re-forking you may end up w/ orphan's. Not sure how this can be done in Go, usual way would be during exec do a setsid and kill the group (ie, negative pid) and wait, when done/timedout which will result in no zombie and no orphan.

i think we can set `cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}` to also send SIGKILL to the reforked processes

idryomov:
@weirdwiz Have you reproduced the timeout expiry scenario specifically?

weirdwiz:
> @weirdwiz Have you reproduced the timeout expiry scenario specifically?

yes, i've reproduced that in a cluster, but that didn't lead to zombies getting created

weirdwiz:
closing as the changes are not required

PR #3379, opened 2025-07-21
mirroring: disable mirroring when secondary cluster is down

rewantsoni:
Refactor the mirroring controller to disable mirroring on cephblockpool and rns when the secondary cluster is down.

1. Create a desired state from the entries in the storage-client-mapping. This includes fetching the information from the peer cluster and creating a state out of it.
2. Reconcile the storagecosumer, blockpools, radosnamespaces, and rbdmirror.

Fixes: [DFBUGS-2957](https://issues.redhat.com/browse/DFBUGS-2957), [DFBUGS-3638](https://issues.redhat.com/browse/DFBUGS-3638)

### DIFF
--- a/internal/controller/mirroring/mirroring_controller.go
+++ b/internal/controller/mirroring/mirroring_controller.go
@@ -21,17 +21,16 @@ import (
 	"reflect"
 	"slices"
 	"strconv"
-	"time"
 
 	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
 	ocsv1alpha1 "github.com/red-hat-storage/ocs-operator/api/v4/v1alpha1"
+	pb "github.com/red-hat-storage/ocs-operator/services/provider/api/v4"
 	providerClient "github.com/red-hat-storage/ocs-operator/services/provider/api/v4/client"
+	storageclusterctrl "github.com/red-hat-storage/ocs-operator/v4/internal/controller/storagecluster"
 	"github.com/red-hat-storage/ocs-operator/v4/pkg/util"
 
 	"github.com/go-logr/logr"
-	storageclusterctrl "github.com/red-hat-storage/ocs-operator/v4/internal/controller/storagecluster"
 	rookCephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
-	"golang.org/x/exp/maps"
 	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -51,7 +50,7 @@ import (
 
 const (
 	mirroringFinalizer              = "ocs.openshift.io/mirroring"
-	clientIDIndexName               = "clientID"
+	clientIDIndexName               = "index:clientID"
 	storageClusterPeerAnnotationKey = "ocs.openshift.io/storage-cluster-peer"
 )
 
@@ -92,7 +91,6 @@ func (r *MirroringReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		return fmt.Errorf("failed to set up FieldIndexer on StorageConsumer for clientID: %v", err)
 	}
 
-	// Reconcile the OperatorConfigMap object when the cluster's version object is updated
 	enqueueConfigMapRequest := handler.EnqueueRequestsFromMapFunc(
 		func(_ context.Context, obj client.Object) []reconcile.Request {
 			return []reconcile.Request{{
@@ -125,6 +123,16 @@ func (r *MirroringReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		},
 	}
 	generationChangePredicate := builder.WithPredicates(predicate.GenerationChangedPredicate{})
+	storageClusterPeerStatusPredicate := predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			if e.ObjectOld == nil || e.ObjectNew == nil {
+				return false
+			}
+			oldObj := e.ObjectOld.(*ocsv1.StorageClusterPeer)
+			newObj := e.ObjectNew.(*ocsv1.StorageClusterPeer)
+			return !reflect.DeepEqual(oldObj.Status, newObj.Status)
+		},
+	}
 
 	return ctrl.NewControllerManagedBy(mgr).
 		Named("MirroringController").
@@ -139,7 +147,10 @@ func (r *MirroringReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Watches(
 			&ocsv1.StorageClusterPeer{},
 			enqueueConfigMapRequest,
-			generationChangePredicate,
+			builder.WithPredicates(
+				predicate.GenerationChangedPredicate{},
+				storageClusterPeerStatusPredicate,
+			),
 		).
 		Watches(
 			&ocsv1alpha1.StorageConsumer{},
@@ -214,9 +225,33 @@ func (r *MirroringReconciler) reconcilePhases(clientMappingConfig *corev1.Config
 		return ctrl.Result{}, err
 	}
 
-	shouldMirror := clientMappingConfig.DeletionTimestamp.IsZero() &&
-		clientMappingConfig.Data != nil &&
-		len(clientMappingConfig.Data) > 0
+	storageConsumerList := &ocsv1alpha1.StorageConsumerList{}
+	if err := r.list(storageConsumerList, client.InNamespace(clientMappingConfig.Namespace)); err != nil {
+		r.log.Error(err, "Failed to list StorageConsumers")
+		return ctrl.Result{}, err
+	}
+
+	remoteClientIds := []string{}
+	storageConsumerByName := map[string]*ocsv1alpha1.StorageConsumer{}
+	for i := range storageConsumerList.Items {
+		consumer := &storageConsumerList.Items[i]
+		storageConsumerByName[consumer.Name] = consumer
+		if cl := consumer.Status.Client; cl != nil && clientMappingConfig.Data[cl.ID] != "" {
+			remoteClientIds = append(remoteClientIds, clientMappingConfig.Data[cl.ID])
+		}
+	}
+
+	cephBlockPoolsList := &rookCephv1.CephBlockPoolList{}
+	if err := r.list(cephBlockPoolsList, client.InNamespace(clientMappingConfig.Namespace)); err != nil {
+		r.log.Error(err, "Failed to list CephBlockPools")
+		return ctrl.Result{}, err
+	}
+
+	remoteClientInfoById := map[string]*pb.ClientInfo{}
+	remoteBlockPoolInfoByName := map[string]*pb.BlockPoolInfo{}
+	errorOccurred := false
+
+	shouldMirror := clientMappingConfig.DeletionTimestamp.IsZero() && len(clientMappingConfig.Data) > 0
 
 	if shouldMirror {
 		if controllerutil.AddFinalizer(clientMappingConfig, mirroringFinalizer) {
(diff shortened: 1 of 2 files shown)


controllers/mirroring/mirroring_controller.go
```diff
@@ -218,13 +219,17 @@ func (r *MirroringReconciler) reconcilePhases(clientMappingConfig *corev1.Config
...
+	var ocsClient *providerClient.OCSProviderClient
+	var err error
 
-	ocsClient, err := providerClient.NewProviderClient(r.ctx, storageClusterPeer.Spec.ApiEndpoint, util.OcsClientTimeout)
-	if err != nil {
-		r.log.Error(err, "failed to create a new provider client")
-		errorOccurred = true
-	} else if ocsClient != nil {
-		defer ocsClient.Close()
+	if shouldMirror {
```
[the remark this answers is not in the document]
rewantsoni: Even in other functions the usage of ocsClient is guarded by shouldMirror

nb-ohad:
@rewantsoni This refactor introduce a lot of noise into the code base. Although the shouldMirror is only calculated once, the decision on how to act on it is spread all over the reconcile function and even being passed downward into other sub functions. Preferably the decision need to be made as early as possible in the reconcile flow then a reconcile path is being choosed based on the result. Doing it this way will benefit the code base, the readability and the ability to reason about the flow.

internal/controller/mirroring/mirroring_controller.go
leelavg, line 74: the value is just "annotation" please make it a bit unique? "index:annotation"?

leelavg:
```suggestion
	clientIDIndexName               = "index:clientID"
```

leelavg: incorrect comment? this isn't OperatorConfigMap.

internal/controller/mirroring/mirroring_controller.go
```diff
@@ -214,6 +214,32 @@ func (r *MirroringReconciler) reconcilePhases(clientMappingConfig *corev1.Config
...
+		return ctrl.Result{}, err
+	}
+
+	remoteClientInfoById := map[string]*pb.ClientInfo{}
+	remoteBlockPoolInfoByName := map[string]*pb.BlockPoolInfo{}
+	errorOccurred := false
+
 	shouldMirror := clientMappingConfig.DeletionTimestamp.IsZero() &&
 		clientMappingConfig.Data != nil &&
 		len(clientMappingConfig.Data) > 0
```
leelavg:
```suggestion
len(clientMappingConfig.Data) > 0
```

internal/controller/mirroring/mirroring_controller.go
```diff
@@ -231,63 +257,59 @@ func (r *MirroringReconciler) reconcilePhases(clientMappingConfig *corev1.Config
...
-			"StorageClusterPeer",
-			storageClusterPeer.Name,
-		)
-		return ctrl.Result{RequeueAfter: 3 * time.Second}, nil
-	}
+		if storageClusterPeer.Status.State != ocsv1.StorageClusterPeerStatePeered ||
+			storageClusterPeer.Status.PeerInfo == nil ||
+			storageClusterPeer.Status.PeerInfo.StorageClusterUid == "" {
+			r.log.Info("Waiting for StorageClusterPeer to be in Peered state", "StorageClusterPeer", storageClusterPeer.Name)
+			return ctrl.Result{RequeueAfter: 3 * time.Second}, nil
```
leelavg: any reasoning for 3 sec, why can't this be requeue on error or add extra predicate for peer?

internal/controller/mirroring/mirroring_controller.go
```diff
@@ -231,63 +257,59 @@ func (r *MirroringReconciler) reconcilePhases(clientMappingConfig *corev1.Config
...
+		}
 
-	errorOccurred := false
+		ocsClient, err := providerClient.NewProviderClient(r.ctx, storageClusterPeer.Spec.ApiEndpoint, util.OcsClientTimeout)
+		if err != nil {
+			r.log.Error(err, "failed to create a new provider client")
+			return ctrl.Result{}, fmt.Errorf("failed to reconcile StorageClientMapping")
+		} else if ocsClient != nil {
+			defer ocsClient.Close()
+		}
```
leelavg:
```suggestion
		defer ocsClient.Close()
```
if it's not err then the conn is established and more over the condition says if the conn is nil proceed? which is wrong.

internal/controller/mirroring/mirroring_controller.go
```diff
@@ -300,6 +322,94 @@ func (r *MirroringReconciler) reconcilePhases(clientMappingConfig *corev1.Config
...
+				resp.BlockPoolName,
+			)
+		}
+		errorOccurred = true
+	}
+
+	for i := range response.BlockPoolsInfo {
+		remoteBlockPoolInfo := response.BlockPoolsInfo[i]
+		remoteBlockPoolInfoByName[remoteBlockPoolInfo.BlockPoolName] = remoteBlockPoolInfo
+	}
```
leelavg, line 379: shouldn't this be a single condition, either you get error or you get info, so

```suggestion
	if response.Errors != nil {
		for i := range response.Errors {
			resp := response.Errors[i]
			r.log.Error(
				errors.New(resp.Message),
				"failed to get BlockPoolsInfo",
				"CephBlockPool",
				resp.BlockPoolName,
			)
		}
		return true
	} else {
  	    for i := range response.BlockPoolsInfo {
  		    remoteBlockPoolInfo := response.BlockPoolsInfo[i]
  		    remoteBlockPoolInfoByName[remoteBlockPoolInfo.BlockPoolName] = remoteBlockPoolInfo
     	}
  	}
```

internal/controller/mirroring/mirroring_controller.go
```diff
@@ -300,6 +322,94 @@ func (r *MirroringReconciler) reconcilePhases(clientMappingConfig *corev1.Config
...
 }
 
+func (r *MirroringReconciler) getBlockPoolsInfo(
+	ocsClient *providerClient.OCSProviderClient,
+	storageClusterPeer *ocsv1.StorageClusterPeer,
+	cephBlockPoolsList *rookCephv1.CephBlockPoolList,
+	remoteBlockPoolInfoByName map[string]*pb.BlockPoolInfo,
+) bool {
+	blockPoolNames := []string{}
+	errorOccurred := false
```
leelavg: not explicitly required as a single condition can handle this.

internal/controller/mirroring/mirroring_controller.go
```diff
@@ -300,6 +322,94 @@ func (r *MirroringReconciler) reconcilePhases(clientMappingConfig *corev1.Config
...
+				resp.ClientID,
+			)
+		}
+		errorOccurred = true
+	}
+
+	for i := range response.ClientsInfo {
+		remoteClientIdByClientInfo[response.ClientsInfo[i].ClientID] = response.ClientsInfo[i]
+	}
+
```
leelavg, line 417: similar to other func.

internal/controller/mirroring/mirroring_controller.go
leelavg:
```suggestion
} else if len(rbdMirrorList.Items) == 1 && rbdMirrorList.Items[0].Name != util.CephRBDMirrorName {
```

internal/controller/mirroring/mirroring_controller.go
```diff
@@ -383,223 +493,192 @@ func (r *MirroringReconciler) reconcileRbdMirror(clientMappingConfig *corev1.Con
 }
 
 func (r *MirroringReconciler) reconcileBlockPoolMirroring(
-	ocsClient *providerClient.OCSProviderClient,
 	clientMappingConfig *corev1.ConfigMap,
-	storageClusterPeer *ocsv1.StorageClusterPeer,
-	shouldMirror bool,
+	cephBlockPoolsList *rookCephv1.CephBlockPoolList,
+	remoteBlockPoolInfoByName map[string]*pb.BlockPoolInfo,
```
leelavg, line 504: are we expecting that we were able to retrieve info for some block pools?

internal/controller/mirroring/mirroring_controller.go
```diff
@@ -383,223 +493,192 @@ func (r *MirroringReconciler) reconcileRbdMirror(clientMappingConfig *corev1.Con
...
 				r.log.Error(
-					errors.New(resp.Message),
-					"failed to get BlockPoolsInfo",
+					err,
+					"failed to disable mirroring for CephBlockPool",
 					"CephBlockPool",
-					resp.BlockPoolName,
+					cephBlockPool.Name,
 				)
+				errorOccurred = true
```
leelavg: there is nothing after the next else, why not just return here and left shift the else block?

internal/controller/mirroring/mirroring_controller.go
```diff
@@ -383,223 +493,192 @@ func (r *MirroringReconciler) reconcileRbdMirror(clientMappingConfig *corev1.Con
...
-					r.log.Error(
-						err,
-						"failed to disable mirroring for CephBlockPool",
-						"CephBlockPool",
-						cephBlockPool.Name,
-					)
-					errorOccurred = true
+
+			// We need to enable mirroring for the blockPool, else the mirroring secret will not be generated
+			_, err := controllerutil.CreateOrUpdate(r.ctx, r.Client, cephBlockPool, func() error {
```
leelavg: a bit confused here, why can't this be in above else block, if you have token create a secret or reconcile bp to have the secret created to be used in token?

internal/controller/mirroring/mirroring_controller.go
```diff
@@ -383,223 +493,192 @@ func (r *MirroringReconciler) reconcileRbdMirror(clientMappingConfig *corev1.Con
...
+		if consumer != nil && consumer.Status.Client != nil {
+			remoteClientId := clientMappingConfig.Data[consumer.Status.Client.ID]
+			if remoteClientInfo := remoteClientInfoById[remoteClientId]; remoteClientInfo != nil {
+				remoteNamespace = remoteClientInfo.RadosNamespace
+			}
+		}
+		_, err := controllerutil.CreateOrUpdate(r.ctx, r.Client, rns, func() error {
+			if remoteNamespace != "" {
+				rns.Spec.Mirroring = &rookCephv1.RadosNamespaceMirroring{
+					RemoteNamespace: ptr.To(remoteNamespace),
```
leelavg, line 612: when we aren't mirroring, the remoteNamespace would be empty, is that still fine or do we consider it as the implicit namespace if it's empty?

internal/controller/mirroring/mirroring_controller.go
rewantsoni: Yes, we are expecting that info for some blockpool was retrived

internal/controller/mirroring/mirroring_controller.go
rewantsoni: In case of multiple blockpools, if there was an issue with one, we still want to continue to reconcile the other blockpools

internal/controller/mirroring/mirroring_controller.go
rewantsoni: Yes that is fine, for impicit rns we add the <implicit> keyword in remoteNamespace

internal/controller/mirroring/mirroring_controller.go
rewantsoni: moved the setting of fields to above block and repaced createOrUpdate with update

internal/controller/mirroring/mirroring_controller.go
rewantsoni: Yes, I agree that this would be better and it won't flood the logs

internal/controller/mirroring/mirroring_controller.go
leelavg: this will contradict https://github.com/red-hat-storage/ocs-operator/pull/3379#discussion_r3193421983

internal/controller/mirroring/mirroring_controller.go
```diff
@@ -18,20 +18,18 @@ import (
...
-	"time"
-
 	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
 	ocsv1alpha1 "github.com/red-hat-storage/ocs-operator/api/v4/v1alpha1"
+	pb "github.com/red-hat-storage/ocs-operator/services/provider/api/v4"
 	providerClient "github.com/red-hat-storage/ocs-operator/services/provider/api/v4/client"
 	"github.com/red-hat-storage/ocs-operator/v4/pkg/util"
+	"reflect"
+	"slices"
+	"strconv"
```
leelavg: recheck.

pkg/util/k8sutil.go
```diff
@@ -57,7 +57,7 @@ const (
 
 	// This is the name for the FieldIndex
 	OwnerUIDIndexName   = "ownerUID"
-	AnnotationIndexName = "annotation"
+	AnnotationIndexName = "index:annotation"
 	ObjectUidIndexName  = "objectUID"
```
leelavg: better change for all these?

rewantsoni: I don't want to make this change in this PR, will send a different PR for this

leelavg (approved):

rewantsoni:
Tested the changes on a cluster and it works.
/unhold

PR #3383, opened 2025-07-22
DFBUGS-3690:[release-4.16]: provider: do not send mon secret to client

leelavg:
there is no requirement to send mon-secret to client and due to the change in content client.admin keyring is being overwritten at client end when both rook-op and client-op runs in same namespace.


agarwal-mudit:
Closed because of https://github.com/red-hat-storage/ocs-client-operator/pull/435

PR #3384, opened 2025-07-23
controller: fix deletion phase

rewantsoni:
This PR makes sure to delete the storageclient and the storageconsumer internal configmap, so that deletion of storagecluster is not blocked

1. Get the storage client before removing the finalizer
2. Delete the local storage consumer configmap

Fixes: [DFBUGS-2674](https://issues.redhat.com/browse/DFBUGS-2674)

### DIFF
--- a/controllers/storagecluster/storageclient.go
+++ b/controllers/storagecluster/storageclient.go
@@ -9,11 +9,11 @@ import (
 	ocsv1a1 "github.com/red-hat-storage/ocs-operator/api/v4/v1alpha1"
 	"github.com/red-hat-storage/ocs-operator/v4/controllers/defaults"
 	"github.com/red-hat-storage/ocs-operator/v4/controllers/util"
-	rookCephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
 
 	ocsclientv1a1 "github.com/red-hat-storage/ocs-client-operator/api/v1alpha1"
+	rookCephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
 	corev1 "k8s.io/api/core/v1"
-	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
@@ -97,10 +97,14 @@ func (s *storageClient) ensureDeleted(r *StorageClusterReconciler, storagecluste
 	}
 	storageClient := &ocsclientv1a1.StorageClient{}
 	storageClient.Name = storagecluster.Name
+	if err := r.Get(r.ctx, client.ObjectKeyFromObject(storageClient), storageClient); client.IgnoreNotFound(err) != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to get storageclient %s: %v", storageClient.Name, err)
+	} else if storageClient.UID == "" {
+		return reconcile.Result{}, nil
+	}
 
-	if err := r.Delete(r.ctx, storageClient); err != nil && !kerrors.IsNotFound(err) {
-		r.Log.Error(err, "Failed to initiate deletion of local StorageClient CR")
-		return reconcile.Result{}, err
+	if err := r.Delete(r.ctx, storageClient); err != nil {
+		return reconcile.Result{}, fmt.Errorf("failed to delete storageclient %s: %v", storageClient.Name, err)
 	}
 
 	if controllerutil.RemoveFinalizer(storageClient, internalComponentFinalizer) {
--- a/controllers/storagecluster/storageconsumer.go
+++ b/controllers/storagecluster/storageconsumer.go
@@ -190,24 +190,39 @@ func (s *storageConsumer) ensureDeleted(r *StorageClusterReconciler, storageClus
 	storageConsumer := &ocsv1a1.StorageConsumer{}
 	storageConsumer.Name = defaults.LocalStorageConsumerName
 	storageConsumer.Namespace = storageCluster.Namespace
-	if err := r.Get(r.ctx, client.ObjectKeyFromObject(storageConsumer), storageConsumer); err != nil {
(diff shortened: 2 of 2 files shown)


controllers/storagecluster/storageclient.go
```diff
@@ -98,10 +98,15 @@ func (s *storageClient) ensureDeleted(r *StorageClusterReconciler, storagecluste
 	}
 	storageClient := &ocsclientv1a1.StorageClient{}
 	storageClient.Name = storagecluster.Name
+	if err := r.Get(r.ctx, client.ObjectKeyFromObject(storageClient), storageClient); err != nil {
+		if client.IgnoreNotFound(err) != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to get storageclient %s: %v", storageClient.Name, err)
+		}
+		return ctrl.Result{}, nil
+	}
```
leelavg, line 106: this err handling is quite confusing, what bug is being solved?

controllers/storagecluster/storageconsumer.go
```diff
@@ -208,6 +208,15 @@ func (s *storageConsumer) ensureDeleted(r *StorageClusterReconciler, storageClus
...
 		}
 	}
+
+	consumerConfigMap := &corev1.ConfigMap{}
+	consumerConfigMap.Name = defaults.LocalStorageConsumerConfigMapName
+	consumerConfigMap.Namespace = storageCluster.Namespace
+
+	if err := r.Delete(r.ctx, consumerConfigMap); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to delete local consumerConfigMap %s: %v", consumerConfigMap.Name, err)
+	}
```
leelavg, line 218: why should this be removed explicitly? either deletion of storagecluster or storageconsumer removes this, isn't it?

controllers/storagecluster/storageclient.go
rewantsoni: We are removing the finalizer from storageclient after deleting it. If we don't load the storageclient, the finalizer won't be present in the local object and it won't go through the update operation to remove the finalizer on the object

controllers/storagecluster/storageconsumer.go
malayparida2000: I believe the deletion will automatically get triggered by the consumer deletion, We need to remove the finalizer from the configmap which blocks it's deletion.

controllers/storagecluster/storageclient.go
leelavg: maybe this can be restructured as
```
if err := r.get(); client.IgnoreNotFound(err) != nil { handle api-error }
else if storageClient.UID == "" { nothing to do here }
// rest as usual + no check for not found in delete
```
wdyt?

controllers/storagecluster/storageconsumer.go
leelavg: you response has a lot of info and some of it needs to be added as a comment, specifically the cyclic relationship which is the main reason for us not waiting for storagecluster deletion.

btw, don't we need any check that this CM is not being owned by 2 consumers? better safe than sorry even though MDR is for external client only?

leelavg: iirc there is some effort around reinstating unintentional storagecluster deletion and it could help if we can have this CM around, that means not deleting this at all.

will let other reviewers or Ohad to decide.

controllers/storagecluster/storageconsumer.go
```diff
@@ -190,24 +190,39 @@ func (s *storageConsumer) ensureDeleted(r *StorageClusterReconciler, storageClus
...
+		if err := r.Delete(r.ctx, storageConsumer); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to delete storageconsumer %s: %v", storageConsumer.Name, err)
+		}
 
-	if controllerutil.RemoveFinalizer(storageConsumer, internalComponentFinalizer) {
-		r.Log.Info("Removing finalizer from StorageConsumer.", "StorageConsumer:", storageConsumer.Name, " StorageConsumer Namespace:", storageConsumer.Namespace, " Finalizer:", internalComponentFinalizer)
-		if err := r.Update(r.ctx, storageConsumer); err != nil {
-			r.Log.Info("Failed to remove finalizer from StorageConsumer.", "StorageConsumer:", storageConsumer.Name, "Finalizer:", internalComponentFinalizer)
-			return ctrl.Result{}, fmt.Errorf("failed to remove finalizer from StorageConsumer: %v", err)
+		if controllerutil.RemoveFinalizer(storageConsumer, internalComponentFinalizer) {
```
malayparida2000, line 204: Shouldn't the finalizer removal block be before the Delete call?

controllers/storagecluster/storageclient.go
```diff
@@ -97,10 +97,14 @@ func (s *storageClient) ensureDeleted(r *StorageClusterReconciler, storagecluste
...
+	if err := r.Get(r.ctx, client.ObjectKeyFromObject(storageClient), storageClient); client.IgnoreNotFound(err) != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to get storageclient %s: %v", storageClient.Name, err)
+	} else if storageClient.UID == "" {
+		return reconcile.Result{}, nil
+	}
 
-	if err := r.Delete(r.ctx, storageClient); err != nil && !kerrors.IsNotFound(err) {
-		r.Log.Error(err, "Failed to initiate deletion of local StorageClient CR")
-		return reconcile.Result{}, err
+	if err := r.Delete(r.ctx, storageClient); err != nil {
```
malayparida2000, line 106: Same here, Shouldn't the finalizer removal be done before Delete() is called on a object?

controllers/storagecluster/storageconsumer.go
rewantsoni: This is the internal component finalizer that is added from storagecluster to the storageconsumer we can remove it after the deletion call

malayparida2000:
@rewantsoni Did you test the fix? 
My expectation would be that if everything else works.
oc delete storagecluster <name> should succeed.

rewantsoni:
@malayparida2000 I have tested this and it works. The expectation is that there are no pv/pvc/snapshots from the rbd/cephfs storageclass (expect Noobaa since we create the Noobaa instance), and storageCluster deletion is triggered, it will succeed.

PR #3387, opened 2025-07-23
Remove deprecated fields from storagecluster & ocsinitialization types

malayparida2000:


rewantsoni:
We should probably move the NewToolsDeployment from storagecluster types to cephtoolbox.go

PR #3392, opened 2025-07-29
Halve the CPU limits of daemons for IBM Z platform to maintain QOS class

malayparida2000:
For pods to get the guaranteed QOS class, they need to have the same CPU limit & request. Not getting the guaranteed QOS class will lead to eviction of the pods early when the node is under pressure.
Ref-https://issues.redhat.com/browse/RHSTOR-7611


travisn (commented):
/lgtm

PR #3396, opened 2025-07-30
Update ocs-operator bundle dependencies

malayparida2000:


black-dragon74:
@malayparida2000 this commit needs to be a part of #3391 else now CI will fail for a different reason "where ocs-op is sending v1a1 and ocs-client is expecting v1".

PR #3400, opened 2025-07-31
fix annotation error in multus cluster-only environs

BlaineEXE:
**VERY IMPORTANTLY:** I haven't had time to test this fix in a deployed dev environment. I am quite confident that the issue is resolved, especially by new unit tests. Depending on priorities, this might be merged and then validated by QE without further dev testing. Otherwise, a dev will have to pick up the testing portion of this change. For now, I am leaving this as a draft so that it won't be merged.

-------------

The existing `getCephNetworkAnnotationValue()` function and the immediate caller behavior made the assumption that any StorageCluster/CephCluster deployed with Multus enabled was required to include a `public` network selection.

In reality, `public` and `cluster` networks can be selected independently and individually. Update the function and its caller to support cluster-only Multus installation environments.

Add unit tests to prevent regression within the function itself.

### DIFF
--- a/controllers/storagecluster/storageclient.go
+++ b/controllers/storagecluster/storageclient.go
@@ -1,12 +1,10 @@
 package storagecluster
 
 import (
-	"encoding/json"
 	"fmt"
 	"maps"
 	"strconv"
 
-	nadv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
 	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
 	ocsv1a1 "github.com/red-hat-storage/ocs-operator/api/v4/v1alpha1"
 	"github.com/red-hat-storage/ocs-operator/v4/controllers/defaults"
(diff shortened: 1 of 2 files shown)


controllers/storagecluster/storageclient.go
```diff
@@ -145,16 +143,27 @@ func (s *storageClient) updateClientConfigMap(r *StorageClusterReconciler, names
...
-	if err != nil || networkSelectionElement == nil {
+	if err != nil {
 		return "", fmt.Errorf("failed to get network selection element: %v", err)
 	}
-	nwAnnotation, err := json.Marshal([]*nadv1.NetworkSelectionElement{networkSelectionElement})
+	if networkSelectionElement == nil {
+		return "", nil // annotation only needed for clusters w/ multus public net selected
+	}
+
+	nwAnnotation, err := rookCephv1.NetworkSelectionsToAnnotationValue(networkSelectionElement)
```
BlaineEXE, line 164: One of the other changes I made was to begin using Rook's helper function designed to convert network selections to annotation values. While marshalling to the struct is essentially all that is needed, usage of Rook's helper function allows ocs-operator to benefit from Rook's unit and CI testing of the functionality.

QE may need to note that the format is expected to be in this format: `[{"name":"odf-public","namespace":"ns"}]` including the surrounding square brackets `[]`.

If anyone disagrees with the benefit, feel free to revert this to JSON marshalling as desired.

leelavg:
@rohan47 fyi...

nb-ohad (changes requested):

leelavg:
/cherrypick release-4.19-hotfix.dfbugs-3726
local cherrypick was clean and doesn't have any conflicts.

PR #3414, opened 2025-08-05
Disallow the creation of OSDs using size greater than 16Ti

malayparida2000:
OSDS can be deployed in any size and put customers at risk of unsupported cluster without raising any flag. It's too easy to make such a mistake and reversing it is not easy at all.
Ref-https://issues.redhat.com/browse/DFBUGS-2528


travisn (approved):
/lgtm

PR #3415, opened 2025-08-05
Remove long deprecated topology label keys, only process valid keys.

malayparida2000:
We were considering few long deprecated topology label keys about region & zone etc then later filtering duplicate labels from our node topology map. This juggling between old & new labels was making the building of the topology map unnecessarily complex and confusing. We did encounter few bugs from customer too on this regard. This commit removes those deprecated keys and only processes the valid topology label keys for region, zone, host and rack.
Also Remove unnecessary log during node topology map build in each reconcile
ref-https://issues.redhat.com/browse/DFBUGS-2904

### DIFF
--- a/controllers/storagecluster/topology.go
+++ b/controllers/storagecluster/topology.go
@@ -301,7 +301,6 @@ func (r *StorageClusterReconciler) reconcileNodeTopologyMap(sc *ocsv1.StorageClu
 			for _, key := range validTopologyLabelKeys {
 				if strings.Contains(label, key) {
 					if !topologyMap.Contains(label, value) {
-						r.Log.Info("Adding topology label from Node.", "Node", node.Name, "Label", label, "Value", value)
 						topologyMap.Add(label, value)
 					}
 				}


controllers/storagecluster/topology.go
```diff
@@ -23,6 +23,13 @@ import (
 
 type ocsTopologyMap struct{}
 
+var validTopologyLabelKeys = []string{
+	corev1.LabelTopologyRegion,
+	corev1.LabelTopologyZone,
+	corev1.LabelHostname,
+	defaults.RackTopologyKey,
```
travisn, line 30: If other [Rook topology labels](https://rook.io/docs/rook/latest-release/CRDs/Cluster/ceph-cluster-crd/#osd-topology) (e.g. room, row, pdu) are found on the nodes, they will just be ignored by ODF, right?

travisn (commented):
Nice cleanup! just one question...

controllers/storagecluster/topology.go
malayparida2000: Yes, ODF only supports 3 failure domains zone, host, rack. So any other labels found on the Nodes would be ignored.
https://github.com/red-hat-storage/ocs-operator/blob/164f7c4c3b997e5e2c66cb3fdebcb91a8d55d6d2/api/v1/topologymap.go#L69-L74

travisn (commented):
/lgtm

PR #3418, opened 2025-08-06
Move ceph configs from rook-ceph-config-override CM to cephcluster CR spec.cephConfig

malayparida2000:
The cephCluster CR is a better place to specify the ceph configs ocs-operator needs to apply. It's easier to maintain & its the recommended way by rook.
Ref-https://issues.redhat.com/browse/DFBUGS-2409


controllers/storagecluster/cephcluster.go
```diff
@@ -1589,24 +1589,89 @@ func generateCephReplicatedSpec(initData *ocsv1.StorageCluster, poolType string)
...
+	partialList := &metav1.PartialObjectMetadataList{}
+	partialList.SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   "ceph.rook.io",
+		Version: "v1",
+		Kind:    "CephRBDMirrorList",
+	})
+	err := r.Client.List(context.TODO(), partialList,
+		client.InNamespace(sc.Namespace),
+		client.Limit(1),
+	)
```
malayparida2000, line 1653: Have used partialObjectMetadata to list as I just need to be aware of the CRs presence and I don't need it's contents. So have tried to make the call as minimal as possible.

malayparida2000:
> mon_osd_backfillfull_ratio: ".80"
>   mon_osd_full_ratio: ".85"
>   mon_osd_nearfull_ratio: ".75"

Okay got the answer,

```
sh-5.1$ ceph config get mon mon_osd_full_ratio
Error EINVAL: mon_osd_full_ratio is special and cannot be stored by the mon

sh-5.1$ ceph osd dump | grep full_ratio
full_ratio 0.85
backfillfull_ratio 0.8
nearfull_ratio 0.75
```

travisn:
Questions:
- If the user wants to override these settings or add additional settings to the list, how can they do it? 
- How did the user previously override these settings? Would those be preserved during upgrade?

PR #3422, opened 2025-08-07
Remove csi toleration specification & Reserve the rook-ceph-operator-config configmap for overrides

malayparida2000:
Till 4.18, tolerations for csi pods were being added to the storageCluster CR, and those were being passed to rook ceph operator via the rook-ceph-operator-config configMap. From 4.19 onwards, rook no longer deploys the csi, it's now done by ceph-csi-operator.
To add tolerations to csi pods user has to now directly edit the csi driver CR, ref-https://issues.redhat.com/browse/DFBUGS-2764. So to avoid confusion, the code relating to adding the csi tolerations via ocs-operator should be removed.

Then We can just create the configmap and not reconcile it.
Ref-https://issues.redhat.com/browse/DFBUGS-3770


travisn (commented):
/lgtm

PR #2975, opened 2025-01-23
set ROOK_CSI_ENABLE_CEPHFS in the ocs-operator-config configmap

iamniting:
Signed-off-by: Nitin Goyal <nigoyal@redhat.com>

Related PRs:
#2970
https://github.com/red-hat-storage/rook/pull/803

https://issues.redhat.com/browse/DFBUGS-1357

PR #3423, opened 2025-08-07
enable host network for rook operator

rohan47:
when using host network with specified public address range rook operator should run on host network to avoid connection issue between pod network and second host network

### DIFF
--- a/controllers/storagecluster/reconcile.go
+++ b/controllers/storagecluster/reconcile.go
@@ -405,6 +405,7 @@ func (r *StorageClusterReconciler) reconcilePhases(
 			// list of default ensure functions
 			// preserve list order
 			objs = []resourceManager{
+				&rookCephCsvHostNetwork{},
 				&ocsProviderServer{},
 				&backingStorageClasses{},
 				&ocsTopologyMap{},
--- a/controllers/storagecluster/reconcile.go
+++ b/controllers/storagecluster/reconcile.go
@@ -108,7 +108,7 @@ var storageClusterFinalizer = "storagecluster.ocs.openshift.io"
 // +kubebuilder:rbac:groups=operators.coreos.com,resources=operatorconditions,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=quota.openshift.io,resources=clusterresourcequotas,verbs=get;list;watch;create;update;delete
 // +kubebuilder:rbac:groups=batch,resources=cronjobs;jobs,verbs=get;list;create;update;watch;delete
-// +kubebuilder:rbac:groups=operators.coreos.com,resources=clusterserviceversions,verbs=get;list;watch
+// +kubebuilder:rbac:groups=operators.coreos.com,resources=clusterserviceversions,verbs=get;list;watch;update
 // +kubebuilder:rbac:groups=k8s.cni.cncf.io,resources=network-attachment-definitions,verbs=get;list;watch
 // +kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=clusterrolebindings;rolebindings;clusterroles;roles,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=objectbucket.io,resources=objectbuckets;objectbucketclaims,verbs=get;list;watch
--- a/controllers/storagecluster/rookoperatorhostnetwork.go
+++ b/controllers/storagecluster/rookoperatorhostnetwork.go
@@ -1,0 +1,84 @@
+package storagecluster
+
+import (
+	"fmt"
+
+	opv1a1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
+	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+type rookCephCsvHostNetwork struct{}
+
+// if the StorageCluster is configured to run on non-default host network, rook-ceph-operator needs to run on host network as well
+// this is validated by checking if the AddressRanges.Public field is set in the Network spec
+// since pod network cannot always communicate with non-default host network rook-ceph-operator needs to run on host network
+func (obj *rookCephCsvHostNetwork) ensureCreated(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) (reconcile.Result, error) {
+
+	csvList := &opv1a1.ClusterServiceVersionList{}
+	labelSelector := labels.Set{
+		fmt.Sprintf("operators.coreos.com/rook-ceph-operator.%s", instance.Namespace): "",
+	}.AsSelector()
+	listOpts := &client.ListOptions{
+		Namespace:     instance.Namespace,
+		LabelSelector: labelSelector,
+	}
+
+	// List the ClusterServiceVersion for rook-ceph-operator
+	err := r.Client.List(r.ctx, csvList, listOpts)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	var rookCSV *opv1a1.ClusterServiceVersion
+	for i := range csvList.Items {
+		csv := &csvList.Items[i]
+		if csv.Status.Phase == opv1a1.CSVPhaseSucceeded {
+			rookCSV = csv
+			break
+		}
+	}
+
+	if rookCSV == nil {
+		return reconcile.Result{}, fmt.Errorf("no rook-ceph-operator CSV found in Succeeded phase")
+	}
+
+	// Update the rook-ceph-operator deployment in CSV to enable host network
+	for i := range rookCSV.Spec.InstallStrategy.StrategySpec.DeploymentSpecs {
+		deployment := &rookCSV.Spec.InstallStrategy.StrategySpec.DeploymentSpecs[i]
+		if deployment.Name == "rook-ceph-operator" {
+			if deployment.Spec.Template.Spec.HostNetwork == shouldUseHostNetworking(instance) {
+				return reconcile.Result{}, nil
+			}
+			// Enable host network for rook-ceph-operator
+			deployment.Spec.Template.Spec.HostNetwork = shouldUseHostNetworking(instance)
+			if err := r.Client.Update(r.ctx, rookCSV); err != nil {
+				r.Log.Error(err, "Failed to update rook-ceph-operator CSV.", "StorageCluster", klog.KRef(instance.Namespace, instance.Name))
+				return reconcile.Result{}, err
+			}
+			break
+		}
+	}
+
+	return reconcile.Result{}, nil
+}
+
+func (obj *rookCephCsvHostNetwork) ensureDeleted(_ *StorageClusterReconciler, _ *ocsv1.StorageCluster) (reconcile.Result, error) {
+	return reconcile.Result{}, nil
+}
+
+// shouldUseHostNetworking checks if the StorageCluster is configured to run on non-default host network
+// this is validated by checking if the AddressRanges.Public field is set in the Network spec
+// since pod network cannot always communicate with non-default host network pods needs to run on host net
+func shouldUseHostNetworking(instance *ocsv1.StorageCluster) bool {
+	if !isMultus(instance.Spec.Network) &&
+		instance.Spec.Network != nil &&
+		instance.Spec.Network.AddressRanges != nil &&
+		instance.Spec.Network.AddressRanges.Public != nil {
+		return true
+	}
+	return false
+}
--- a/controllers/storagecluster/initialization_reconciler_test.go
+++ b/controllers/storagecluster/initialization_reconciler_test.go
@@ -144,6 +144,31 @@ var (
 			},
 		}
 	}
+	createRookCephOperatorCSV = func(namespace string) *opv1a1.ClusterServiceVersion {
+		return &opv1a1.ClusterServiceVersion{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "rook-ceph-operator-csv",
+				Namespace: namespace,
+				Labels: map[string]string{
+					fmt.Sprintf("operators.coreos.com/rook-ceph-operator.%s", namespace): "",
+				},
+			},
+			Spec: opv1a1.ClusterServiceVersionSpec{
+				InstallStrategy: opv1a1.NamedInstallStrategy{
+					StrategySpec: opv1a1.StrategyDetailsDeployment{
+						DeploymentSpecs: []opv1a1.StrategyDeploymentSpec{
+							{
+								Name: "rook-ceph-operator",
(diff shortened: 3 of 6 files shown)


rohan47:
/hold 
need to update RBAC and test changes

controllers/storagecluster/cephcluster.go
```diff
@@ -208,6 +209,16 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
 		}
 	}
 
+	if sc.Spec.HostNetwork && sc.Spec.Network != nil {
+		if sc.Spec.Network.AddressRanges != nil && sc.Spec.Network.AddressRanges.Public != nil {
```
Madhu-1, line 213: Please add a comment about the change  and combine all thecks into one if and do we need `sc.Spec.HostNetwork` network check as well?

controllers/storagecluster/cephcluster.go
```diff
@@ -1707,3 +1718,52 @@ func getFullRatio(specifiedRatio *float64) *float64 {
...
+	err := c.List(ctx, csvList, listOpts)
+	if err != nil {
+		return err
+	}
+
+	// Check if we found exactly one ClusterServiceVersion for rook-ceph-operator
+	if len(csvList.Items) == 0 {
+		return fmt.Errorf("no ClusterServiceVersion found for rook-ceph-operator in namespace %s", opNamespace)
+	}
+	if len(csvList.Items) > 1 {
```
Madhu-1, line 1743: This happens during upgrade?

controllers/storagecluster/cephcluster.go
```diff
@@ -1707,3 +1718,52 @@ func getFullRatio(specifiedRatio *float64) *float64 {
...
+		return fmt.Errorf("multiple ClusterServiceVersions found for rook-ceph-operator, skipping update during upgrade")
+	}
+	rookCSV := &csvList.Items[0]
```
Madhu-1, line 1746: use DeepCopy instead of taking it as pointer and updating it directly?

controllers/storagecluster/cephcluster.go
```diff
@@ -1707,3 +1718,52 @@ func getFullRatio(specifiedRatio *float64) *float64 {
...
+
+	// Update the rook-ceph-operator deployment in CSV to enable host network
+	updated := false
+	for i := range rookCSV.Spec.InstallStrategy.StrategySpec.DeploymentSpecs {
+		deployment := &rookCSV.Spec.InstallStrategy.StrategySpec.DeploymentSpecs[i]
+		if deployment.Name == "rook-ceph-operator" {
+			// Enable host network
+			deployment.Spec.Template.Spec.HostNetwork = true
```
Madhu-1, line 1754: return early if the HostNetwork is already set?

controllers/storagecluster/cephcluster.go
```diff
@@ -1707,3 +1718,52 @@ func getFullRatio(specifiedRatio *float64) *float64 {
...
+			updated = true
```
Madhu-1, line 1755: instead of new variable issue `c.Update(ctx, rookCSV)` here only

controllers/storagecluster/exporter.go
```diff
@@ -83,6 +84,14 @@ func (r *StorageClusterReconciler) enableMetricsExporter(
 		return err
 	}
 
+	// bind the exporter to hostnetwork-v2 SCC if needed
+	if getRookCephHostNetwork(instance) {
```
Madhu-1, line 88: Please add comment for this one and separate this change to a new commit with more details for this change

controllers/storagecluster/exporter.go
```diff
@@ -1107,3 +1117,39 @@ func (r *StorageClusterReconciler) deleteMetricsExporterCephClient(namespace str
 
 	return nil
 }
+
+func getRookCephHostNetwork(instance *ocsv1.StorageCluster) bool {
```
Madhu-1, line 1121: i think this is not a better function name it should be `runOnHostNetwork()`?

controllers/storagecluster/exporter.go
```diff
@@ -1107,3 +1117,39 @@ func (r *StorageClusterReconciler) deleteMetricsExporterCephClient(namespace str
...
+	if instance.Spec.HostNetwork && instance.Spec.Network != nil {
+		if instance.Spec.Network.AddressRanges != nil && instance.Spec.Network.AddressRanges.Public != nil {
+			return true
+		}
+	}
+	return false
+}
+
+func bindExporterToHostNetworkV2SCC(ctx context.Context, r *StorageClusterReconciler, ns string) error {
+	crb := &rbacv1.ClusterRoleBinding{
```
Madhu-1, line 1131: can you please add details why we need to create new clusterRoleBinding is this for metrics exporter to talk to prometheus?

controllers/storagecluster/exporter.go
```diff
@@ -83,6 +84,14 @@ func (r *StorageClusterReconciler) enableMetricsExporter(
 		return err
 	}
 
+	// bind the exporter to hostnetwork-v2 SCC if needed
+	if getRookCephHostNetwork(instance) {
+		if err := bindExporterToHostNetworkV2SCC(ctx, r, instance.Namespace); err != nil {
```
Madhu-1, line 89: deletion code is missing for this one?

controllers/storagecluster/cephcluster.go
```diff
@@ -1707,3 +1718,52 @@ func getFullRatio(specifiedRatio *float64) *float64 {
 	defaultRatio := 0.85
 	return &defaultRatio
 }
+
+// updateRookCephOperatorHostNetwork updates the rook-ceph-operator CSV to enable host network
+func updateRookCephOperatorHostNetwork(ctx context.Context, c client.Client, opNamespace string) error {
+	csvList := &opv1a1.ClusterServiceVersionList{}
+	labelSelector := labels.SelectorFromSet(map[string]string{
+		fmt.Sprintf("operators.coreos.com/rook-ceph-operator.%v", opNamespace): "",
```
Madhu-1, line 1726: dont we need to use our own labels (Rook provided one) instead of depending on the label set by OLM?

controllers/storagecluster/cephcluster.go
malayparida2000, line 1754: Why are we using this approach of getting the CSV & patching the deployment spec inside it, For setting values from ocs-operator to rook, we already have the ocs-operator-config CM,
You can add the value based on your conditions to that CM & that CM's values can be fed to rook operator via CSV env vars. See example here https://github.com/red-hat-storage/ocs-operator/pull/2975

malayparida2000 (changes requested):

controllers/storagecluster/cephcluster.go
rohan47: yes sc.Spec.HostNetwork check is required as AddressRanges can also be used in case of multus

controllers/storagecluster/cephcluster.go
nb-ohad: @rohan47 Changing the CSV is problematic as when upgrading the new CSV will not include the changes.

controllers/storagecluster/cephcluster.go
```diff
@@ -1707,3 +1718,52 @@ func getFullRatio(specifiedRatio *float64) *float64 {
 	defaultRatio := 0.85
 	return &defaultRatio
 }
+
+// updateRookCephOperatorHostNetwork updates the rook-ceph-operator CSV to enable host network
+func updateRookCephOperatorHostNetwork(ctx context.Context, c client.Client, opNamespace string) error {
+	csvList := &opv1a1.ClusterServiceVersionList{}
+	labelSelector := labels.SelectorFromSet(map[string]string{
```
nb-ohad, line 1725: I don't thin you need to use SelectorFromSet here.  You only need to match a single label.

controllers/storagecluster/cephcluster.go
```diff
@@ -1707,3 +1718,52 @@ func getFullRatio(specifiedRatio *float64) *float64 {
...
+	// List the ClusterServiceVersion for rook-ceph-operator
+	err := c.List(ctx, csvList, listOpts)
+	if err != nil {
+		return err
+	}
+
+	// Check if we found exactly one ClusterServiceVersion for rook-ceph-operator
+	if len(csvList.Items) == 0 {
+		return fmt.Errorf("no ClusterServiceVersion found for rook-ceph-operator in namespace %s", opNamespace)
+	}
```
nb-ohad, line 1742: This is wrong, you need to check if the CSV is the current CSV. During upgrades, even stale upgrades you will always find more then one csv

controllers/storagecluster/cephcluster.go
nb-ohad, line 1754: This will get lost upon upgrade, we need a different approach

controllers/storagecluster/exporter.go
nb-ohad, line 1121: This function returns a bool, it should not be called `get` as it does not get anything. Please consider renaming it to something like `shouldUseHostNetworking`

controllers/storagecluster/exporter.go
nb-ohad, line 1131: All role and role binding should be part of the olm package, our operator should not be able to create new roles or role bindings for security reasons

nb-ohad (changes requested):

controllers/storagecluster/cephcluster.go
```diff
@@ -208,6 +210,19 @@ func (obj *ocsCephCluster) ensureCreated(r *StorageClusterReconciler, sc *ocsv1.
...
+	// since pod network cannot always communicate with non-default host network rook-ceph-operator needs to run on host network
+	if !isMultus(sc.Spec.Network) &&
+		sc.Spec.Network != nil &&
+		sc.Spec.Network.AddressRanges != nil &&
+		sc.Spec.Network.AddressRanges.Public != nil {
+		if err := updateRookCephOperatorHostNetwork(r.ctx, r.Client, r.OperatorNamespace); err != nil {
+			r.Log.Error(err, "Failed to update Rook Ceph Operator HostNetwork.", "StorageCluster", klog.KRef(sc.Namespace, sc.Name))
+			return reconcile.Result{}, err
+		}
+	}
```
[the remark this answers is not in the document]
rohan47: Moved the changes from cephCluster ensureCreated to reconcile.go before the the list of ensure functions https://github.com/red-hat-storage/ocs-operator/blob/main/controllers/storagecluster/reconcile.go#L397

controllers/storagecluster/reconcile.go
```diff
@@ -394,6 +396,19 @@ func (r *StorageClusterReconciler) reconcilePhases(
...
+	// since pod network cannot always communicate with non-default host network rook-ceph-operator needs to run on host network
+	if !instance.Spec.ExternalStorage.Enable &&
+		!r.IsNoobaaStandalone &&
+		shouldUseHostNetworking(instance) {
+		err := updateOpHostNetworkInRookCephCSV(ctx, r.Client, instance.Namespace)
+		if err != nil {
+			r.Log.Error(err, "Failed to update Rook Ceph Operator HostNetwork.", "StorageCluster", klog.KRef(instance.Namespace, instance.Name))
+			return reconcile.Result{}, err
+		}
+	}
```
nb-ohad, line 410: Why is this here and does not follow the usual ensureCreated/ensureDeleted standard this repo standardised on?

nb-ohad (changes requested):

controllers/storagecluster/rookoperatorhostnetwork.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+func (obj *rookCephCsvHostNetwork) ensureDeleted(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) (reconcile.Result, error) {
+	err := updateOpHostNetworkInRookCephCSV(r.ctx, r.Client, instance.Namespace, false)
+	if err != nil {
+		r.Log.Error(err, "Failed to update Rook Ceph Operator HostNetwork.", "StorageCluster", klog.KRef(instance.Namespace, instance.Name))
+		return reconcile.Result{}, err
+	}
+	return reconcile.Result{}, nil
+}
```
nb-ohad, line 39: Why do we need to revert anything here when the storage cluster is getting deleted? 
Wouldn't a new storage cluster will just re-set (using ensure created) to the proper value when it will get created?

controllers/storagecluster/rookoperatorhostnetwork.go
```diff
@@ -0,0 +1,104 @@
...
+func (obj *rookCephCsvHostNetwork) ensureCreated(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) (reconcile.Result, error) {
+	if shouldUseHostNetworking(instance) {
+		err := updateOpHostNetworkInRookCephCSV(r.ctx, r.Client, instance.Namespace, true)
+		if err != nil {
+			r.Log.Error(err, "Failed to update Rook Ceph Operator HostNetwork.", "StorageCluster", klog.KRef(instance.Namespace, instance.Name))
+			return reconcile.Result{}, err
+		}
+	}
+	return reconcile.Result{}, nil
+}
```
nb-ohad, line 30: What is going to happen in mixed mode, when we have more then one storagecluster/cephcluster?

controllers/storagecluster/rookoperatorhostnetwork.go
```diff
@@ -0,0 +1,104 @@
...
+		instance.Spec.Network != nil &&
+		instance.Spec.Network.AddressRanges != nil &&
+		instance.Spec.Network.AddressRanges.Public != nil {
+		return true
+	}
+	return false
+}
+
+// updateOpHostNetworkInRookCephCSV updates the rook-ceph-operator CSV to enable host network
+func updateOpHostNetworkInRookCephCSV(ctx context.Context, c client.Client, opNamespace string, enableHostNet bool) error {
```
nb-ohad, line 55: Isn't this function just the implementation of ensureCreated? why is there a reason to have a util function for this?

nb-ohad (changes requested):

controllers/storagecluster/rookoperatorhostnetwork.go
rohan47: If there is another storageCluster which is not using dedicated hostNetwork will break the current scenario.
Is it possible to restrict only one storage cluster creation, or we set the hostNetwork for rook-op once but other storageCluster may error out if not using host-net

controllers/storagecluster/rookoperatorhostnetwork.go
rohan47: CSV changes will not be removed once ensure created changes it (unless ensureCreated checks the rook-csv every time and set appropriate host nw value)

controllers/storagecluster/reconcile.go
rohan47: was not hitting test case errors because the code was trying to check a specific use case, 
but now with using ensureCreated, a lot of unit tests are failing because the code fails on finding rook-csv and deployment.

Do we want to just log a warning instead of throwing errors?

PR #3426, opened 2025-08-08
DFBUGS-3771: [release-4.19] Remove csi toleration specification & Reserve the rook-ceph-operator-config configmap for overrides

malayparida2000:
Manual Cherry-pick of https://github.com/red-hat-storage/ocs-operator/pull/3422
Till 4.18, tolerations for csi pods were being added to the storageCluster CR, and those were being passed to rook ceph operator via the rook-ceph-operator-config configMap. From 4.19 onwards, rook no longer deploys the csi, it's now done by ceph-csi-operator.
To add tolerations to csi pods user has to now directly edit the csi driver CR, ref-https://issues.redhat.com/browse/DFBUGS-2764. So to avoid confusion, the code relating to adding the csi tolerations via ocs-operator should be removed.

Then We can just create the configmap and not reconcile it.

### DIFF
--- a/api/v1/ocsinitialization_types.go
+++ b/api/v1/ocsinitialization_types.go
@@ -54,17 +54,17 @@ type OCSInitializationStatus struct {
 	// operator. Object references will be added to this list after they have
 	// been created AND found in the cluster.
 	// +optional
-	RelatedObjects                []corev1.ObjectReference     `json:"relatedObjects,omitempty"`
-	ErrorMessage                  string                       `json:"errorMessage,omitempty"`
-	SCCsCreated                   bool                         `json:"sCCsCreated,omitempty"`
-	RookCephOperatorConfigCreated bool                         `json:"rookCephOperatorConfigCreated,omitempty"`
-	RookCephOperatorConfig        RookCephOperatorConfigStatus `json:"rookCephOperatorConfig,omitempty"`
+	RelatedObjects []corev1.ObjectReference `json:"relatedObjects,omitempty"`
+	ErrorMessage   string                   `json:"errorMessage,omitempty"`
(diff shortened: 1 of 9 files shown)


Madhu-1:
@malayparida2000 i think this contains API changes. do we do API changes in the patch release?

malayparida2000:
> i think this contains API changes. do we do API changes in the patch release?

We do not have any strict standard about it, But we normally avoid API changes in patch releases. But if you notice the actual changes the fields that are removed are status fields. And status fields are always set by the owner operator(i.e us), So I don't forsee any risk with this. @nb-ohad WDYT?

Madhu-1:
> > i think this contains API changes. do we do API changes in the patch release?
> 
> We do not have any strict standard about it, But we normally avoid API changes in patch releases. But if you notice the actual changes the fields that are removed are status fields. And status fields are always set by the owner operator(i.e us), So I don't forsee any risk with this. @nb-ohad WDYT?

okay, i thought this is spec changes, i hope no customer build any CI tools for it (any way it wont work if its set here)

nb-ohad:
> > i think this contains API changes. do we do API changes in the patch release?
> 
> We do not have any strict standard about it, But we normally avoid API changes in patch releases. But if you notice the actual changes the fields that are removed are status fields. And status fields are always set by the owner operator(i.e us), So I don't forsee any risk with this. @nb-ohad WDYT?

We should not be removing in a z stream. The reason is that we don't know what customer are expecting. They might have automations that relay on these status fields. The proper way to do it is to mark them as deprecated on the next y release then remove them completely in the y release the follows

iamniting (approved):

PR #3432, opened 2025-08-12
Use Topology Spread Constraints instead of Pod anti affinity for MDS daemon placement

malayparida2000:
We have no reliable way to avoid scheduling 2 active MDS pods on the same node. So instead of pod anti-affinity we can switch to TSCs similar to how we use them for OSDs, with a max skew of 1. We can use the rook_file_system: <fs-name> label selector so that all MDS daemon (doesn't matter active or passive) of a file system are distributed
equally over the cluster. 
Also TSCs are a better way to distribute pods across the cluster then pod anti-affinity. Pod anti-affinity can't handle the case where more than one pod need to be scheduled on the same node for example when there are more than 1 active MDS. Tscs are easier to manage and understand. osd & osd-prepare are already using TSCs.
Ref-https://issues.redhat.com/browse/DFBUGS-1612

### DIFF
--- a/controllers/storagecluster/placement_test.go
+++ b/controllers/storagecluster/placement_test.go
@@ -181,9 +181,9 @@ func TestGetPlacement(t *testing.T) {
 					PodAntiAffinity: defaults.DaemonPlacements["mon"].PodAntiAffinity,
 				},
 				"mds": {
-					NodeAffinity:    defaults.DefaultNodeAffinity,
-					PodAntiAffinity: defaults.DaemonPlacements["mds"].PodAntiAffinity,
-					Tolerations:     defaults.DaemonPlacements["mds"].Tolerations,
+					NodeAffinity:              defaults.DefaultNodeAffinity,
+					TopologySpreadConstraints: defaults.DaemonPlacements["mds"].TopologySpreadConstraints,
+					Tolerations:               defaults.DaemonPlacements["mds"].Tolerations,
 				},
 			},
 		},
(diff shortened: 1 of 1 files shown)


controllers/defaults/placements.go
```diff
@@ -66,10 +66,9 @@ var (
 			Tolerations: []corev1.Toleration{
 				getOcsToleration(),
 			},
-			PodAntiAffinity: &corev1.PodAntiAffinity{
-				RequiredDuringSchedulingIgnoredDuringExecution: []corev1.PodAffinityTerm{
-					// left the selector value empty as it will be updated later in the getPlacement()
-				},
+			TopologySpreadConstraints: []corev1.TopologySpreadConstraint{
+				// leave the label value empty as it will be updated with the filesystem name later in getPlacement()
+				getTopologySpreadConstraint(1, "rook_file_system", []string{}, corev1.DoNotSchedule),
```
travisn, line 71: We always are strict for `DoNotSchedule`? I guess it's not that strict though since it allows a difference of one per node if needed.

travisn (commented):
/lgtm

travisn:
/hold
pending test confirmation

malayparida2000:
Test results
Verified the TSC's presence on the mds pods.
Now when 4 mds pods are there they get distributed properly.
```
~ $ oc get pods -o wide | grep mds
rook-ceph-mds-ocs-storagecluster-cephfilesystem-a-7d89975b5tlvc   2/2     Running     0          15m     10.129.2.37    mparida-1-8n9hf-worker-1-4gwm2   <none>           <none>
rook-ceph-mds-ocs-storagecluster-cephfilesystem-b-78cd56dcpmqhb   2/2     Running     0          15m     10.131.0.79    mparida-1-8n9hf-worker-2-hxvdz   <none>           <none>
rook-ceph-mds-ocs-storagecluster-cephfilesystem-c-7cc555656fnr4   2/2     Running     0          8m42s   10.128.2.52    mparida-1-8n9hf-worker-3-qk8q7   <none>           <none>
rook-ceph-mds-ocs-storagecluster-cephfilesystem-d-66c4774dz6lr8   2/2     Running     0          8m40s   10.129.2.44    mparida-1-8n9hf-worker-1-4gwm2   <none>           <none>
```

PR #3437, opened 2025-08-12
Always collect ocs must-gather irrespective of test result

malayparida2000:
After some discussion & thoughts, we decided to always collect the ocs must-gather irrespective of test result. This would help a lot in verifying the results of any fix whether or not that fix actually results in a test failure. We acknowledge that this would result in sight increase in the time taken to run the tests. But this would help a lot in looking at the impact of any PR readily.

### DIFF
--- a/functests/setup_teardown.go
+++ b/functests/setup_teardown.go
@@ -37,14 +37,12 @@ func AfterTestSuiteCleanup() {
 	debug("\n------------------------------\n")
 
 	// collect debug log before deleting namespace & cluster
-	if SuiteFailed {
-		debug("AfterTestSuite: collecting debug information\n")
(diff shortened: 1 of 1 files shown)


malayparida2000:
Must gather can be seen collected here https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/red-hat-storage_ocs-operator/3437/pull-ci-red-hat-storage-ocs-operator-main-ocs-operator-bundle-e2e-aws/1955349295553056768/artifacts/ocs-operator-bundle-e2e-aws/e2e-test/artifacts/ocs-must-gather/

PR #3441, opened 2025-08-13
Make possible to mark EnableCrushUpdates as false

malayparida2000:
The enableCrushUpdates field is a bool field, and after we get the storageCluster as a structured object it's impossible to know if the field is set to false because it was not specified or because it was explicitly set to false. Hence till now we were setting the field to true always. 

Now we are getting the storageCluster as unstructured, and we can check if the field exists in the unstructured object. If it does, it means the user has explicitly set the field. If it does not, it means the field is not specified by the user. This way the user can now make the field enableCrushUpdates as false.

Ref-https://issues.redhat.com/browse/DFBUGS-3813

### DIFF
--- a/controllers/storagecluster/cephcluster.go
+++ b/controllers/storagecluster/cephcluster.go
@@ -28,6 +28,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
@@ -1529,8 +1530,13 @@ func isEncrptionSettingUpdated(clusterWideEncrytion bool, existingDeviceSet []ro
 }
 
 // setDefaultMetadataPoolSpec sets the common pool spec for all metadata pools as necessary
-func setDefaultMetadataPoolSpec(poolSpec *rookCephv1.PoolSpec, sc *ocsv1.StorageCluster) {
-	poolSpec.EnableCrushUpdates = true
+func (r *StorageClusterReconciler) setDefaultMetadataPoolSpec(poolSpec *rookCephv1.PoolSpec, sc *ocsv1.StorageCluster, specifiedPoolSpecPath []string) {
+	// Check if enableCrushUpdates is explicitly set in the unstructured obj else set to true
+	if val, found := getValueAtPath(r.unstructuredSC, append(specifiedPoolSpecPath, "enableCrushUpdates")); found {
+		poolSpec.EnableCrushUpdates = val.(bool)
+	} else {
+		poolSpec.EnableCrushUpdates = true
+	}
 	if poolSpec.DeviceClass == "" {
 		poolSpec.DeviceClass = sc.Status.DefaultCephDeviceClass
 	}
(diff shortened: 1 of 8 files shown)


controllers/storagecluster/cephcluster.go
```diff
@@ -1529,9 +1530,17 @@ func isEncrptionSettingUpdated(clusterWideEncrytion bool, existingDeviceSet []ro
...
+	// Check if enableCrushUpdates is explicitly set in the unstructured obj else set to true
+	if val, found := getValueAtPath(r.unstructuredSC, append(specifiedPoolSpecPath, "enableCrushUpdates")); found {
+		poolSpec.EnableCrushUpdates = val.(bool)
+	} else {
+		poolSpec.EnableCrushUpdates = true
+	}
+	// Check if DeviceClass is explicitly set in the unstructured obj else set to the default device class
+	if val, found := getValueAtPath(r.unstructuredSC, append(specifiedPoolSpecPath, "deviceClass")); found {
+		poolSpec.DeviceClass = val.(string)
+	} else if poolSpec.DeviceClass == "" {
```
travisn, line 1543: Is it possible for the user to explicitly set blank? Seems like we need to still enforce the default device class in that case.
```suggestion
	}
	if poolSpec.DeviceClass == "" {
```

controllers/storagecluster/cephcluster.go
```diff
@@ -1549,12 +1558,21 @@ func setDefaultMetadataPoolSpec(poolSpec *rookCephv1.PoolSpec, sc *ocsv1.Storage
...
```
travisn, line 1575: Seems like lines 1566-1575 is duplicated from the method above. Could this be factored out to a helper method?

controllers/storagecluster/cephcluster.go
```diff
@@ -1577,6 +1595,33 @@ func setDefaultDataPoolSpec(poolSpec *rookCephv1.PoolSpec, sc *ocsv1.StorageClus
 	}
 }
 
+func getValueAtPath(unstructuredSC *unstructured.Unstructured, path []string) (any, bool) {
```
travisn, line 1598: How about unit tests for this method?

travisn (changes requested):
While I am not excited about the additional complexity of this approach, it is nice that a new user setting does not need to be added. It does seem worth it with added testing.

leelavg:
jira description explained the golang mechanics but not the business justification, I see the statement `who are upgrading from pre-4.16 scenarios are facing some issues in their setups if they have multiple deviceClasses.` and wondering what `some issues` mean here 🤔.

I believe we all agree this PR is introducing a complexity to fill the gap of incorrect api, to think of alternatives I want to know what `some issues` mean in above para, wrt to provided solution I see below are some of the cons:
1. opens a way to retrofit all the fields which we designed incorrectly
2. unstructured is inherently not type check-able and a change to api need to consider the usage of unstructured code during upgrades, upgrade can get stuck if it's in the path of operator conditions
3. unstructured is afaik not cached (maybe there is an explicit opt in 🤔), so every reconcile does a live lookup and for an object like StorageCluster (which is relatively big) and a reconciler like StorageCluster (which has relatively many watches) this introduces time & memory latencies/burden.

the previous use of unstructured (that I'm aware of) is in provider-server which is laser focused and only do a get this was done as not to import the whole olm repo 😅, so I guess this PR needs more justification.

open to be corrected.

controllers/storagecluster/cephcluster.go
malayparida2000: Okay, Thanks for the clarification removed the check for deviceClass now

malayparida2000:
> jira description explained the golang mechanics but not the business justification, I see the statement `who are upgrading from pre-4.16 scenarios are facing some issues in their setups if they have multiple deviceClasses.` and wondering what `some issues` mean here 🤔.
> 
> I believe we all agree this PR is introducing a complexity to fill the gap of incorrect api, to think of alternatives I want to know what `some issues` mean in above para, wrt to provided solution I see below are some of the cons:
> 
> 1. opens a way to retrofit all the fields which we designed incorrectly
> 2. unstructured is inherently not type check-able and a change to api need to consider the usage of unstructured code during upgrades, upgrade can get stuck if it's in the path of operator conditions
> 3. unstructured is afaik not cached (maybe there is an explicit opt in 🤔), so every reconcile does a live lookup and for an object like StorageCluster (which is relatively big) and a reconciler like StorageCluster (which has relatively many watches) this introduces time & memory latencies/burden.
> 
> the previous use of unstructured (that I'm aware of) is in provider-server which is laser focused and only do a get this was done as not to import the whole olm repo 😅, so I guess this PR needs more justification.
> 
> open to be corrected.

The issues were a long chain of things, starting with the partner setting up clusters with multiple device classes.
The end solution we reached was that they need to disable makeEnableCrushUpdates: false.

Keeping aside this specific customer's issue, we have made all other fields of the poolSpec configurable, except this field, which sounds wrong. So we do need some way for someone to be able to mark EnableCrushUpdates to false.

> 1. opens a way to retrofit all the fields that we designed incorrectly

If rook were to change the type of the api field to a pointer to bool may be that would the cleanest solution. But what I heard was this issue does not affect rook upstream as for them the default is false anyways. So they are unwilling to make an API change for that. And Changing API type introduces risk during upgrade. But open to discuss this more.

>2. upgrade can get stuck if it's in the path of operator conditions

For now the function I added does not return any error, So the reconcile is not blocked due to any problem in parsing the paths in the unstructured object.

>3. unstructured is afaik not cached (maybe there is an explicit opt in 🤔)

You are right. Thanks for letting me know that. I was not aware of it but after looking up the controller runtime package I have now added a client option for caching the unstructured object. Although I have to read more if that would spike the memory usage of the operator pod

PR #3442, opened 2025-08-14
external: do not create metrics cephclient CR for external mode

parth-gr:
in external mode we dont have admin privileges
and so we should avoid creating cephclient from the ocs operator,
If needed in future we can create the cephclient
from the python script.

Currently external mode will rely on the fall back mechanism to use the csi rbd secret for getting the metrics exporter run.


parth-gr:
cc @malayparida2000

PR #3448, opened 2025-08-19
ocs-metrics-exporter and toolbox needs to run on host network for data replication separation

rohan47:
ocs-metrics-exporter and toolbox needs to run on host network for data replication separation as pod network will not have access to secondary host network and will not be able to execute ceph commands

### DIFF
--- a/controllers/storagecluster/cephtoolbox.go
+++ b/controllers/storagecluster/cephtoolbox.go
@@ -71,6 +71,9 @@ func (r *StorageClusterReconciler) ensureToolsDeployment(sc *ocsv1.StorageCluste
 				toolsDeployment.Spec.Template.Spec.HostNetwork = false
 			}
 		}
+		if shouldUseHostNetworking(sc) {
+			toolsDeployment.Spec.Template.Spec.HostNetwork = true
+		}
 
 		if !isFound {
 			return r.Client.Create(context.TODO(), toolsDeployment)
--- a/controllers/storagecluster/exporter.go
+++ b/controllers/storagecluster/exporter.go
@@ -377,6 +377,7 @@ func deployMetricsExporter(ctx context.Context, r *StorageClusterReconciler, ins
 				SecurityContext: &corev1.PodSecurityContext{
 					RunAsNonRoot: ptr.To(true),
 				},
+				HostNetwork: shouldUseHostNetworking(instance),
 				Containers: []corev1.Container{
 					{
 						Resources: defaults.MonitoringResources["kube-rbac-proxy"],
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -350,6 +350,14 @@ rules:
 - apiGroups:
   - security.openshift.io
   resourceNames:
+  - hostnetwork-v2
+  resources:
+  - securitycontextconstraints
+  verbs:
+  - use
+- apiGroups:
+  - security.openshift.io
+  resourceNames:
   - privileged
   resources:
   - securitycontextconstraints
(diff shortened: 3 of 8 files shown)


malayparida2000:
> > I see work was going on in #3423, Can you briefly mention what was the output of that PR & the need of this different PR
> 
> The other PR is only for rook csv changes and this PR is for metrics-exporter and toolbox the result of the other PR is that rook-operator pod will run on hostNetwork and this PR enables ocs-metrics-exporter and toolbox pod to run on hostNetwork

But that PR name & desc still mentions about ocs-metrics-exporter, hence I got confused.

controllers/storagecluster/cephtoolbox.go
```diff
@@ -81,6 +81,9 @@ func (r *StorageClusterReconciler) ensureToolsDeployment(sc *ocsv1.StorageCluste
 				toolsDeployment.Spec.Template.Spec.HostNetwork = false
 			}
 		}
+		if runOnHostNetwork(sc) {
```
leelavg, line 84: I see SCC for ocs-metrics but not for toolbox, is it misisng?

rohan47: toolbox pod uses rook-ceph-default service account and it’s able to use host network, so did not add additional scc

controllers/storagecluster/exporter.go
```diff
@@ -759,6 +760,12 @@ func updateMetricsExporterClusterRoles(ctx context.Context, r *StorageClusterRec
 				Resources: []string{"objectbuckets"},
 				Verbs:     []string{"get", "list"},
 			},
+			{
+				APIGroups:     []string{"security.openshift.io"},
+				Resources:     []string{"securitycontextconstraints"},
+				ResourceNames: []string{"hostnetwork-v2"},
+				Verbs:         []string{"use"},
+			},
```
nb-ohad, line 768: We shouldn't be touching (creating/updating) RBAC from code. Why is this needed? why can't we just add these statically via olm?

nb-ohad (changes requested):

controllers/storagecluster/exporter.go
rohan47: The changes are here to align with the already existing flow for (creating/updating) RBAC for metrics exporter,
If the changes here doesn't make sense then will add static changes

nb-ohad: @rohan47 The fact that we have some code that stray form the boundaries of what the operator should be able to do or shouldn't be able to do, which we also consider a mistake and a problem. Does not mean we should add to that problem be adding more

Please use a build-time/static approach for the additional rbac

controllers/storagecluster/exporter.go
```diff
@@ -1114,3 +1115,16 @@ func (r *StorageClusterReconciler) deleteMetricsExporterCephClient(namespace str
 
 	return nil
 }
+
+// shouldUseHostNetworking checks if the StorageCluster is configured to run on non-default host network
+// this is validated by checking if the AddressRanges.Public field is set in the Network spec
+// since pod network cannot always communicate with non-default host network metrics exporter needs to run on host net
+func shouldUseHostNetworking(instance *ocsv1.StorageCluster) bool {
```
rohan47, line 1122: this part should be removed when the following code is in
https://github.com/red-hat-storage/ocs-operator/pull/3423/files#diff-4160c186e4e6b63623e530892b0aa7f6385e7f60f45a5f3da944492ea6211787R969

controllers/storagecluster/exporter.go
rohan47: Made changes

nb-ohad:
@rohan47 
looks good. But according to the comment you made here https://github.com/red-hat-storage/ocs-operator/pull/3448#discussion_r2309724248,  this one should only be merged after #3448

rohan47:
https://github.com/red-hat-storage/ocs-operator/pull/3423 is merged and this PR is rebased

PR #3464, opened 2025-08-28
store client info during onboard process

leelavg:
when the provider server ack's that a client is onboarded provider doesn't allow onboard again even if the original client lost the info that it got onboarded.

to overcome this and make the onboarding idempotent we store the client info on the consumer during onboard and check against it if the same call is made by the client.


leelavg:
/cherrypick release-4.20
via DFBUGS-3791

nb-ohad:
/lgtm
/hold

Adding a hold to wait for the removal of the unused `SetPlatformVersion` and `SetOperatorVersion` functions

leelavg:
tested below scenarios:
1. fresh onboard
2. spec is enabled but client missing the notification

PR #3466, opened 2025-08-28
sever: move to a cached client for provider server

rewantsoni:
Move from an uncached client in the provider server to a cached client

This fix is a part of [DFBUGS-2341](https://issues.redhat.com/browse/DFBUGS-2341). More PRs will follow.

### DIFF
--- a/controllers/util/k8sutil.go
+++ b/controllers/util/k8sutil.go
@@ -60,6 +60,7 @@ const (
 	// This is the name for the FieldIndex
 	OwnerUIDIndexName   = "ownerUID"
 	AnnotationIndexName = "annotation"
+	ObjectUidIndexName  = "objectUID"
 
 	OdfInfoNamespacedNameClaimName      = "odfinfo.odf.openshift.io"
 	ExitCodeThatShouldRestartTheProcess = 42
@@ -196,6 +197,10 @@ func OwnersIndexFieldFunc(obj client.Object) []string {
 
 func AnnotationIndexFieldFunc(obj client.Object) []string {
 	return maps.Keys(obj.GetAnnotations())
+}
+
+func ObjectUidIndexFieldFunc(obj client.Object) []string {
+	return []string{string(obj.GetUID())}
 }
 
 func GetTopologyConstrainedPools(storageCluster *ocsv1.StorageCluster) string {
--- a/services/provider/server/consumer.go
+++ b/services/provider/server/consumer.go
@@ -3,85 +3,37 @@ package server
 import (
 	"context"
 	"fmt"
-	"sync"
 
 	ocsv1alpha1 "github.com/red-hat-storage/ocs-operator/api/v4/v1alpha1"
 	ifaces "github.com/red-hat-storage/ocs-operator/services/provider/api/v4/interfaces"
 	"github.com/red-hat-storage/ocs-operator/v4/controllers/util"
 
-	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
-	clientgocache "k8s.io/client-go/tools/cache"
 	ctrlcache "sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/client/config"
 )
 
 type ocsConsumerManager struct {
 	client    client.Client
 	namespace string
-	nameByUID map[types.UID]string
-	mutex     sync.RWMutex
 }
 
-func newConsumerManager(ctx context.Context, cl client.Client, namespace string) (*ocsConsumerManager, error) {
+func newConsumerManager(ctx context.Context, cl client.Client, cache ctrlcache.Cache, namespace string) (*ocsConsumerManager, error) {
 
-	consumerManager := &ocsConsumerManager{
+	if err := cache.IndexField(
+		ctx,
+		&ocsv1alpha1.StorageConsumer{},
+		util.ObjectUidIndexName,
+		util.ObjectUidIndexFieldFunc,
+	); err != nil {
+		return nil, fmt.Errorf("failed to set up FieldIndexer on StorageConsumer for uid: %v", err)
+	}
+
+	return &ocsConsumerManager{
 		client:    cl,
 		namespace: namespace,
-		nameByUID: map[types.UID]string{},
-	}
-
-	cache, err := newStorageConsumerCache(namespace)
-	if err != nil {
-		return nil, fmt.Errorf("failed to create cache: %w", err)
-	}
-
-	consumer := &metav1.PartialObjectMetadata{}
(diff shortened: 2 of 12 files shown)


services/provider/server/server.go
```diff
@@ -107,23 +111,47 @@ func NewOCSProviderServer(ctx context.Context, namespace string) (*OCSProviderSe
...
+			DefaultNamespaces: map[string]cache.Config{
+				namespace: {},
+			},
+		},
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	consumerManager, err := newConsumerManager(ctx, mgr, namespace)
```
leelavg, line 131: same, rather than sending manager, could do `mgr.GetClient()`

leelavg (commented):
since we are creating informers on the fly pls check all the places we use a get/list and if the resources of that kind will be more in number then we need to restrict them during cache init.

ex: if we do a get on crd then the runtime will start an informer and blocks till we get the whole list of CRDs in the cluster.

services/provider/server/server.go
rewantsoni: I am creating the informer to populate the nameByUID map in newConsumerManager, and it requires a cache. I am using the cache from mgr instead of creating a new one, hence passing the mgr instead of the client

services/provider/server/server.go
```diff
@@ -107,23 +111,47 @@ func NewOCSProviderServer(ctx context.Context, namespace string) (*OCSProviderSe
...
 	if err != nil {
 		return nil, fmt.Errorf("failed to create new OCSConumer instance. %v", err)
 	}
 
-	storageClusterPeerManager, err := newStorageClusterPeerManager(client, namespace)
+	storageClusterPeerManager, err := newStorageClusterPeerManager(mgr, namespace)
```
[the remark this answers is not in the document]
rewantsoni: I wanted to align both newConsumerManager and newStorageClusterPeerManager, hence passing the mgr to it as well. Should I switch to only passing the client?

services/provider/server/server.go
leelavg: ok, missed it. thanks.

services/provider/server/consumer.go
```diff
@@ -26,20 +24,17 @@ type ocsConsumerManager struct {
...
+		client:    mgr.GetClient(),
 		namespace: namespace,
 		nameByUID: map[types.UID]string{},
 	}
 
-	cache, err := newStorageConsumerCache(namespace)
-	if err != nil {
-		return nil, fmt.Errorf("failed to create cache: %w", err)
-	}
+	cache := mgr.GetCache()
```
nb-ohad, line 35: Why do we need to handle cache if we are using a cached client for all of the resources?

services/provider/server/server.go
```diff
@@ -105,23 +108,47 @@ func NewOCSProviderServer(ctx context.Context, namespace string) (*OCSProviderSe
...
+				namespace: {},
+			},
+		},
+	})
+	if err != nil {
+		return nil, err
 	}
 
-	consumerManager, err := newConsumerManager(ctx, client, namespace)
+	consumerManager, err := newConsumerManager(ctx, mgr, namespace)
```
nb-ohad, line 128: Is there a reason we need to pass the entire manager and not just the client?

services/provider/server/server.go
```diff
@@ -105,23 +108,47 @@ func NewOCSProviderServer(ctx context.Context, namespace string) (*OCSProviderSe
...
+	}
+
+	mgr, err := manager.New(config, manager.Options{
+		Scheme: scheme,
+		Cache: cache.Options{
+			DefaultNamespaces: map[string]cache.Config{
+				namespace: {},
+			},
+		},
+	})
```
nb-ohad, line 123: Can you please why do we need to use a controller manager in the server where the only thing we want is a cached backed client?

nb-ohad (changes requested):

controllers/util/k8sutil.go
```diff
@@ -60,6 +60,7 @@ const (
 	// This is the name for the FieldIndex
 	OwnerUIDIndexName   = "ownerUID"
 	AnnotationIndexName = "annotation"
+	UIDIndexName        = "uid"
```
leelavg, line 63: let's have a suffix to the index as plain uid is actually a field in every resource.

services/provider/server/consumer.go
```diff
@@ -3,85 +3,26 @@ package server
...
 	client    client.Client
 	namespace string
-	nameByUID map[types.UID]string
-	mutex     sync.RWMutex
 }
 
-func newConsumerManager(ctx context.Context, cl client.Client, namespace string) (*ocsConsumerManager, error) {
-
-	consumerManager := &ocsConsumerManager{
+func newConsumerManager(cl client.Client, namespace string) (*ocsConsumerManager, error) {
```
leelavg, line 21: I don't see any error being returned in this func?

rewantsoni: Yes, removed the error from here as well as from storageclusterpeer manager

services/provider/server/consumer.go
```diff
@@ -120,29 +61,21 @@ func (c *ocsConsumerManager) GetByName(ctx context.Context, name string) (*ocsv1
...
-		if kerrors.IsNotFound(err) {
-			return nil, fmt.Errorf("storageConsumer resource %q not found. %v", consumerName, err)
-		}
-		return nil, fmt.Errorf("failed to get storageConsumer resource with name %q. %v", consumerName, err)
+	storageConsumers := &ocsv1alpha1.StorageConsumerList{}
+	if err := c.client.List(
+		ctx,
+		storageConsumers,
+		client.MatchingFields{util.ObjectUidIndexName: id},
+	); err != nil {
```
nb-ohad, line 70: Please cap your list operation to 1

services/provider/server/consumer.go
```diff
@@ -3,85 +3,26 @@ package server
...
 	client    client.Client
 	namespace string
-	nameByUID map[types.UID]string
-	mutex     sync.RWMutex
 }
 
-func newConsumerManager(ctx context.Context, cl client.Client, namespace string) (*ocsConsumerManager, error) {
-
-	consumerManager := &ocsConsumerManager{
+func newConsumerManager(cl client.Client, namespace string) *ocsConsumerManager {
```
nb-ohad, line 21: The consumer manager now requires a cached client and an indexer for the UID. 
This code need to make sure that the passed client actually support both before allowing usage of the client.

Or at least, be the one to set the index itself

services/provider/server/server.go
```diff
@@ -105,26 +107,57 @@ func NewOCSProviderServer(ctx context.Context, namespace string) (*OCSProviderSe
...
+	}
+
+	if err := cache.IndexField(
+		ctx,
+		&ocsv1alpha1.StorageConsumer{},
+		util.ObjectUidIndexName,
+		util.ObjectUidIndexFieldFunc,
+	); err != nil {
+		return nil, fmt.Errorf("failed to set up FieldIndexer on StorageConsumer for uid: %v", err)
 	}
```
nb-ohad, line 132: The index itself is an assumption of the consumer manager and is only consumed (and required) by it. It does not make sense to set it from the outside.

nb-ohad (changes requested):

leelavg (approved):

leelavg (approved):

PR #3467, opened 2025-08-28
snapshotclasses: add params required for ListSnapshots

black-dragon74:
This patch adds the params required by the `external-snapshotter` for `ListSnapshots` RPC.

Without this the status updates for the VolumeSnapshots will fail with an error stating: `the provided secrets are empty`


controllers/util/snapshotclasses.go
```diff
@@ -47,6 +47,9 @@ func NewDefaultRbdSnapshotClass(
 			"clusterID": clusterID,
 			"csi.storage.k8s.io/snapshotter-secret-name":      provisionerSecret,
 			"csi.storage.k8s.io/snapshotter-secret-namespace": namespace,
+
```
leelavg, line 50: there is no need to explicitly separate these by a line as all these are just fields in a map?

this isn't required for groupsc?

black-dragon74: That extra space is to keep git blame cleaner (i can remove it if you want). This change will be reverted in a future release due to upcoming changes in external-snapshotter.

No, this is only for volsnaps.

leelavg: Can remove, the PR isn't huge that warrants a distinction or as it's slated for removal, pls comment the same and you'll get clear diff for free 😅.

black-dragon74:
@leelavg this requires `/approve` as well.

PR #3469, opened 2025-09-01
Fix workaround for CSV merger

SupremeMortal:
This removes the workaround that replaces descriptions containing '---' with proper parsing of multiple YAML documents in a single file, therefore fixing the markdown in the parsed CRD.


iamniting (changes requested):
Can you pls fix the commit msg?

PR #3477, opened 2025-09-02
make onboarding from same client resistant against expired ticket

leelavg:
if the onboarding is already done and being retried by the same client we only need to validate the ticket signature for it's data and skip the timestamp thereby completing the onboard.


leelavg:
/cherrypick release-4.20
via DFBUGS-3791

PR #3502, opened 2025-09-10
enable hostNetwork and change ports for rgw for non default hostNetwork while using data replication seperation

rohan47:
ceph object stores need to run on hostNetwork when using data replication seperation on non default host network and the ports need to be changes to avoid port collision.


controllers/storagecluster/cephobjectstores.go
```diff
@@ -283,3 +287,19 @@ func hasNonPortableOSD(sc *ocsv1.StorageCluster) bool {
...
+	if shouldUseHostNetworking(sc) {
+		return 31080
+	}
+	return 80
+}
+
+// when using non default hostNetwork Object Store should run on hostNetwork and use different secure port to avoid port collision
+func getSecurePort(sc *ocsv1.StorageCluster) int32 {
+	if shouldUseHostNetworking(sc) {
+		return 31044
```
rohan47, line 302: @nb-ohad @sp98  @BlaineEXE @parth-gr @leelavg 
Currently ocs-operator hard codes the port and securePort for CephObjectStores, 
Do we want to enable ocs-operator to provide option to set port and secure port for ObjectStore or we hard code a different port for Data replication use case?

rohan47: Updated, to have option in StorageCluster CR and a fallback if not set

rohan47:
/unhold
Tested the changes

BlaineEXE (commented):
Code seems fine. If this were Rook, I would require unit tests, but I think ocs-operator is a bit more lax on that. Someone from the ocs-op team can approve assuming this meets code maintenance/quality

controllers/storagecluster/cephcluster.go
```diff
@@ -1572,6 +1572,12 @@ func getCephClusterCephConfig(r *StorageClusterReconciler, sc *ocsv1.StorageClus
 		r.Log.Error(err, "could not get network config details", "CephCluster", klog.KRef(sc.Namespace, sc.Name))
 	}
 
+	if shouldUseHostNetworking(sc) {
+		cephConfig["client.rgw.*"] = map[string]string{
+			"rgw_frontends": fmt.Sprintf("beast endpoint=0.0.0.0:%d secure_port=%d", getRGWPort(sc), getRGWSecurePort(sc)),
```
BlaineEXE, line 1577: Have you verified that this actually changes RGW's behavior? This is given to RGW as a CLI flag which usually has a higher precedence than Ceph configs. I would expect this change to have no effect on RGW behavior.

Also, have youtried to set this on the CephObjectStore directly using `.gateway.rgwConfig` spec? Those are a better option since they are applied to only the specific object store and not all possible object stores.

rohan47: Config takes effect
```
$ ceph config dump | grep rgw_front
client.rgw.*                                           basic     rgw_frontends                          beast endpoint=0.0.0.0:50080 secure_port=50443  *
 ```

and RGW is listining on the correct port
```
 netstat -tlnp | grep rados
tcp        0      0 0.0.0.0:50080           0.0.0.0:*               LISTEN      187974/radosgw
tcp        0      0 0.0.0.0:50443           0.0.0.0:*               LISTEN      187974/radosgw
```

Comment (BlaineEXE) [reply controllers/storagecluster/cephcluster.go:1577]: `ceph config dump` doesn't guarantee that RGW is actually **using** the config. It only dumps the config database. You need to look at the RGW log or and/or use `ceph config show` to actually see the running config.

Comment (BlaineEXE) [reply controllers/storagecluster/cephcluster.go:1577]: > Applying `.gateway.rgwConfig` got removed

I'm not sure what you mean by that. Can you clarify?

If it's not being stored (or even if it is), I'd recommend overriding any frontend configs using `RgwCommandFlags`. And I'd recommend not overriding the `rgw_frontends` config unless it's needed. Overriding this makes the config unique among Rook object configs in our ODF product and could make debugging and support harder.

Comment (rohan47) [reply controllers/storagecluster/cephcluster.go]: I was looking for `.gateway.rgwConfig` in ceph config wich was missing but was working, but it turns out the issue was because of rook operator not using dnsPolicy `ClusterFirstWithHostNet`

Removed the frontend config change

File: controllers/storagecluster/rookoperatorhostnetwork.go [hunk 3]
```diff
@@ -55,6 +56,7 @@ func (obj *rookCephCsvHostNetwork) ensureCreated(r *StorageClusterReconciler, in
 			}
 			// Enable host network for rook-ceph-operator
 			deployment.Spec.Template.Spec.HostNetwork = shouldUseHostNetworking(instance)
+			deployment.Spec.Template.Spec.DNSPolicy = corev1.DNSClusterFirstWithHostNet
```
nb-ohad, line 59: @rohan47 Can you please explain whey this is now set everytime?

api/v1/storagecluster_types.go
```diff
@@ -291,6 +291,8 @@ type ManageCephObjectStores struct {
 	GatewayInstances  int    `json:"gatewayInstances,omitempty"`
 	DisableRoute      bool   `json:"disableRoute,omitempty"`
 	HostNetwork       *bool  `json:"hostNetwork,omitempty"`
+	Port              int    `json:"port,omitempty"`
+	SecurePort        int    `json:"securePort,omitempty"`
```
nb-ohad, line 295:
```suggestion
	GatewayPort              int    `json:"port,omitempty"`
	GatewaySecurePort        int    `json:"securePort,omitempty"`
```

controllers/storagecluster/rookoperatorhostnetwork.go
rohan47: This is required for DNS resolution for openshift service when hostNetwork is set to true

BlaineEXE (commented):
From the ocs-operator/rook integration, I think things look good to me

PR #3512, opened 2025-09-16
Skip adding default node affinity for NooBaa standalone or external mode

malayparida2000:
In noobaa-standalone mode or external mode the default ocs labels are not added to the nodes. So we should not add the default ocs node affinity for these modes.


malayparida2000:
As tested in https://ibm-systems-storage.slack.com/archives/C06DWD30QTZ/p1758083612460409?thread_ts=1757423665.444559&cid=C06DWD30QTZ.
With both mcs-only mode & external mode where the cluster had scheduling issues, after replacing the ocs-operator image, the issue was resolved.

iamniting (approved):

PR #3533, opened 2025-09-29
add default resource requests/limits for ocs-provider-server and other sidecars

Nikhil-Ladha:
add default resource requests/limits for ocs-provider-server, mgr-sidecar, oauth-proxy.
And, also add capability to modify toolbox's resource requirements.

### DIFF
--- a/controllers/defaults/resources.go
+++ b/controllers/defaults/resources.go
@@ -124,6 +124,36 @@ var (
 				corev1.ResourceMemory: resource.MustParse("128Mi"),
 			},
 		},
+		"ocs-provider-server": {
+			Requests: corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("10m"),
+				corev1.ResourceMemory: resource.MustParse("128Mi"),
+			},
+			Limits: corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("50m"),
+				corev1.ResourceMemory: resource.MustParse("512Mi"),
+			},
+		},
+		"mgr-sidecar": {
+			Requests: corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("50m"),
+				corev1.ResourceMemory: resource.MustParse("75Mi"),
+			},
+			Limits: corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("200m"),
+				corev1.ResourceMemory: resource.MustParse("256Mi"),
+			},
+		},
+		"rook-ceph-tools": {
+			Requests: corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("50m"),
+				corev1.ResourceMemory: resource.MustParse("128Mi"),
+			},
+			Limits: corev1.ResourceList{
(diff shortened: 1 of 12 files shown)


controllers/storagecluster/cephtoolbox.go
```diff
@@ -48,6 +48,13 @@ func (r *StorageClusterReconciler) ensureToolsDeployment(sc *ocsv1.StorageCluste
 		return err
 	}
 
+	specifiedResourceRequirements, specified := sc.Spec.Resources["rook-ceph-tools"]
+	// if specified resource requirements is present but empty, the intention is to have no resource requirements
+	if specified && isResourceNonEmpty(specifiedResourceRequirements) {
```
malayparida2000, line 53: Can't we call getDaemonResources("toolbox", sc)
This merge logic is already present inside getDaemonResources

controllers/storagecluster/provider_server.go
```diff
@@ -252,6 +252,10 @@ func (o *ocsProviderServer) ensureDeploymentReplica(actual, desired *appsv1.Depl
 
 func GetProviderAPIServerDeployment(instance *ocsv1.StorageCluster) *appsv1.Deployment {
 
+	// We need to copy the values from getDaemonResources into a varirable as the Pod spec expects
+	// a pointer to the ResourceRequirements api.
+	ocsProviderServerResource := getDaemonResources(ocsProviderServerName, instance)
```
malayparida2000, line 257: Instead of specifying the resources at Pod level try specifying them at the container level. That's the recommended way by k8s, and it does not require a pointer as well.

controllers/defaults/resources.go
```diff
@@ -124,6 +124,36 @@ var (
...
+			Requests: corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("10m"),
+				corev1.ResourceMemory: resource.MustParse("128Mi"),
+			},
+			Limits: corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("50m"),
+				corev1.ResourceMemory: resource.MustParse("512Mi"),
+			},
+		},
+		"mgr-sidecar": {
```
malayparida2000, line 137: Hope these numbers of backed by some kind of recommendation or testing

controllers/defaults/resources.go
```diff
@@ -124,6 +124,36 @@ var (
 				corev1.ResourceMemory: resource.MustParse("128Mi"),
 			},
 		},
+		"ocs-provider-server": {
```
malayparida2000, line 127: Hope these numbers of backed by some kind of recommendation or testing

controllers/defaults/resources.go
malayparida2000: You need to add the resources under the cephCluster spec as well

malayparida2000 (changes requested):

controllers/defaults/resources.go
Nikhil-Ladha: Yes, these numbers are derived from testing and checked with Leela on this.

malayparida2000 (approved):

PR #3536, opened 2025-10-03
Enable caching + add new non-admin endpoints to UX backend server

SanjalKatiyar:
https://issues.redhat.com/browse/RHSTOR-7965
https://issues.redhat.com/browse/RHSTOR-7970
https://issues.redhat.com/browse/RHSTOR-8003

### Introduced Changes:

- Segregated 2 paths for OAuth Proxy:
   - **"/":** (Default) Only authenticated + authorized users allowed. This path is needed for admin user APIs.
   - **"/info/":** Only authenticated users allowed (no authorization). This path is needed for any OpenShift user (admin or non-admin) APIs.

- Enabled cached client:
   - `DefaultNamespaces` set to watch `openshift-storage` and `openshift-storage-extended` only.
   - Indexing (cluster-scoped resources).
 
- Added 3 new endpoints to "/info/" path:
   - **"/info/featureflags?flags=noobaa,rgw":**
     - Enable/Disable (or Show/Hide) features on UI based on flags.
     - Will be used globally by UI (irrespective of user type or which page user navigate to).
     - Faster & lightweight (less response size, we don't need all the info, we only need to know the subset).
   - **"/info/bucket/\<bucket-name\>":**
     - Determine whether a S3 bucket (NooBaa/RGW) is created via OBC or directly via S3 API.
     - Specific non-admin only use case for now.
   - **"/info/storages":**
     - Comprehensive namespace-wise cluster state.
     - Specific non-admin only use case for now. Also, only called when/if user go to a specific UI page (local usage), so only needed sometimes optionally.
     - **Comparatively** slower and comprehensive (we need all the information for UI to make any decision, so response is **relatively** bigger).

### Testing (endpoint):
```
$ curl -k -H "Authorization: Bearer <ADMIN_USER>" https://ux-backend-proxy.openshift
-storage.svc.cluster.local:8888/info/featureflags?flags=noobaa,rgw,fakeFlag

>> {"fakeFlag":{"value":false,"error":"unsupported flag"},"noobaa":{"value":true},"rgw":{"value":false}}
```

```
$ curl -k -H "Authorization: Bearer <NON_ADMIN_USER>" https://ux-backend-proxy.openshift
-storage.svc.cluster.local:8888/info/featureflags?flags=noobaa,rgw,fakeFlag

>> {"fakeFlag":{"value":false,"error":"unsupported flag"},"noobaa":{"value":true},"rgw":{"value":false}}
```

```
$ curl -k -H "Authorization: Bearer <ADMIN_USER>" https://ux-backend-proxy.openshift
-storage.svc.cluster.local:8888/info/bucket/test-bucket

>> {"createdVia":"obc"}
```

```
$ curl -k -H "Authorization: Bearer <ADMIN_USER>" https://ux-backend-proxy.openshift
-storage.svc.cluster.local:8888/info/bucket/test-e60c8369-b6d9-4e84-bf2d-89cae86cbafd

>> {"createdVia":"s3"}
```

```
$ curl -k -H "Authorization: Bearer <NON_ADMIN_USER>" https://ux-backend-proxy.openshift
-storage.svc.cluster.local:8888/info/bucket/test-e60c8369-b6d9-4e84-bf2d-89cae86cbafd

>> {"createdVia":"s3"}
```

```
$ curl -v -k -H "Authorization: Bearer <FAKE_USER>" https://ux-backend-proxy.o
penshift-storage.svc.cluster.local:8888/info/featureflags?flags=noobaa,rgw,fakeFlag

>> HTTP/1.1 403 Forbidden
```

```
$ curl -v -k -H "Authorization: Bearer <FAKE_USER>" https://ux-backend-proxy.o
penshift-storage.svc.cluster.local:8888/info/bucket/test-bucket

>> HTTP/1.1 403 Forbidden
```

```
$ curl -v -k -H "Authorization: Bearer <NON_ADMIN_USER>" https://ux-backend-proxy.o
penshift-storage.svc.cluster.local:8888/info/storages

>> {"operatorNamespace":"openshift-storage","clusterNamespaces":{"openshift-storage":{"storageClusterName":"test","deploymentType":"internal"}}}
```

```
$ curl -v -k -H "Authorization: Bearer <NON_ADMIN_USER>" https://ux-backend-proxy.o
penshift-storage.svc.cluster.local:8888/info/storages

>> {"operatorNamespace":"openshift-storage","clusterNamespaces":{"openshift-storage":{"storageClusterName":"test","deploymentType":"internal"}}}
```

### Testing (cache):

First request - slower (populates cache)
```
{"operatorNamespace":"openshift-storage","clusterNamespaces":{"openshift-storage":{"storageClusterName":"test","deploymentType":"internal"}}}

real    0m0.459s
user    0m0.009s
sys     0m0.006s
```

Subsequent requests - faster (cache exists)
```
{"operatorNamespace":"openshift-storage","clusterNamespaces":{"openshift-storage":{"storageClusterName":"test-new","deploymentType":"internal"}}}

real    0m0.023s
user    0m0.010s
sys     0m0.004s
```

### DIFF
--- a/config/rbac/oauth_proxy_role.yaml
+++ b/config/rbac/oauth_proxy_role.yaml
@@ -10,6 +10,7 @@ rules:
 - apiGroups: ["authorization.k8s.io"]
   resources:
   - subjectaccessreviews
+  - selfsubjectaccessreviews
   verbs: ["create"]
 - apiGroups:
   - storage.k8s.io
@@ -17,3 +18,30 @@ rules:
   - storageclasses
   verbs:
   - create
+  - get
+  - list
+  - watch
+- apiGroups:
+  - objectbucket.io
+  resources:
+  - objectbuckets
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - ceph.rook.io
+  resources:
+  - cephobjectstores
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - ocs.openshift.io
+  resources:
+  - storageclusters
+  verbs:
+  - get
+  - list
+  - watch
--- a/rbac/ux_backend_role.yaml
+++ b/rbac/ux_backend_role.yaml
@@ -28,3 +28,11 @@ rules:
   - cephblockpools
   verbs:
   - create
+- apiGroups:
+  - noobaa.io
+  resources:
+  - noobaas
+  verbs:
+  - get
+  - list
+  - watch
--- a/services/ux-backend/handlers/bucket/handler.go
+++ b/services/ux-backend/handlers/bucket/handler.go
@@ -1,0 +1,91 @@
+package bucket
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	"github.com/red-hat-storage/ocs-operator/v4/services/ux-backend/handlers"
+
+	nbv1a1 "github.com/noobaa/noobaa-operator/v5/pkg/apis/noobaa/v1alpha1"
+	uxutil "github.com/red-hat-storage/ocs-operator/v4/services/ux-backend/util"
+	"k8s.io/klog/v2"
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	creationMethodOBC = "obc"
+	creationMethodS3  = "s3"
+)
+
+func HandleMessage(w http.ResponseWriter, r *http.Request, client ctrlclient.Client) {
+	switch r.Method {
+	case "GET":
+		handleGet(w, r, client)
+	default:
+		handleUnsupportedMethod(w, r)
+	}
+}
+
+// request format: /info/bucket/<bucket-name>
+func handleGet(w http.ResponseWriter, r *http.Request, client ctrlclient.Client) {
+	bucketName := r.PathValue("name")
+	if bucketName == "" {
+		klog.Errorf("bucket name is required in path")
+		http.Error(w, "bucket name is required in path", http.StatusBadRequest)
+		return
+	}
+
+	creationMethod, err := getBucketCreationMethod(r.Context(), client, bucketName)
+	if err != nil {
+		klog.Errorf("failed to get bucket creation method: %v", err)
+		http.Error(w, fmt.Sprintf("Failed to get bucket creation method: %v", err), http.StatusInternalServerError)
+		return
+	}
+
+	result := map[string]string{
+		"createdVia": creationMethod,
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
+
+	if err := json.NewEncoder(w).Encode(result); err != nil {
+		klog.Errorf("failed to encode response: %v", err)
+		http.Error(w, "Internal server error", http.StatusInternalServerError)
+		return
+	}
+}
+
+func getBucketCreationMethod(ctx context.Context, client ctrlclient.Client, bucketName string) (string, error) {
+	klog.Infof("Getting creation method for bucket '%s'", bucketName)
+	
+	objectBucketList := &nbv1a1.ObjectBucketList{}
+	if err := client.List(ctx, objectBucketList, ctrlclient.MatchingFields{uxutil.IndexBucketName: bucketName}, ctrlclient.Limit(1)); err != nil {
+		return "", fmt.Errorf("failed to list ObjectBuckets by bucketName index: %w", err)
+	}
+	
+	if len(objectBucketList.Items) > 0 {
+		klog.Infof("Found ObjectBucket '%s' with bucket name '%s'", objectBucketList.Items[0].Name, bucketName)
+		return creationMethodOBC, nil
+	}
+	
+	// There are two ways to create a bucket:
+	// 1. Either via OBC (ObjectBucketClaim) CR
+	// 2. Or via S3 endpoint directly
+	// This endpoint is only called on existing buckets, so we assume S3 creation if no OBC is found.
+	klog.Infof("No ObjectBucket found with bucket name '%s', assuming S3 creation", bucketName)
+	return creationMethodS3, nil
+}
+
+func handleUnsupportedMethod(w http.ResponseWriter, r *http.Request) {
+	klog.Infof("Only GET method should be used for this endpoint %s", r.URL.Path)
+	w.Header().Set("Content-Type", handlers.ContentTypeTextPlain)
+	w.Header().Set("Allow", "GET")
+	w.WriteHeader(http.StatusMethodNotAllowed)
+
+	if _, err := w.Write([]byte(fmt.Sprintf("Unsupported method: %s", r.Method))); err != nil {
+		klog.Errorf("failed to write data to response writer: %v", err)
+	}
+}
--- a/services/ux-backend/handlers/featureflags/handler.go
+++ b/services/ux-backend/handlers/featureflags/handler.go
@@ -1,0 +1,170 @@
+package featureflags
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"strings"
+
+	"github.com/red-hat-storage/ocs-operator/v4/controllers/util"
+	"github.com/red-hat-storage/ocs-operator/v4/services/ux-backend/handlers"
+
+	nbv1a1 "github.com/noobaa/noobaa-operator/v5/pkg/apis/noobaa/v1alpha1"
+	cephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
+	storagev1 "k8s.io/api/storage/v1"
+	"k8s.io/klog/v2"
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	flagNoobaa string = "noobaa"
+	flagRGW    string = "rgw"
+)
+
+func HandleMessage(w http.ResponseWriter, r *http.Request, client ctrlclient.Client, namespace string) {
+	switch r.Method {
+	case "GET":
+		handleGet(w, r, client, namespace)
+	default:
+		handleUnsupportedMethod(w, r)
+	}
+}
+
+// request format (comma-separated flags): /info/featureflags?flags=noobaa,rgw
+func handleGet(w http.ResponseWriter, r *http.Request, client ctrlclient.Client, namespace string) {
+	flagsQuery := r.URL.Query().Get("flags")
+	if flagsQuery == "" {
+		klog.Errorf("flags parameter is required")
+		http.Error(w, "flags parameter is required", http.StatusBadRequest)
+		return
+	}
+
+	flagNames := strings.Split(flagsQuery, ",")
+	for i, flag := range flagNames {
+		flagNames[i] = strings.TrimSpace(flag)
+	}
+
+	type flagResult struct {
+		Value bool   `json:"value"`
+		Error string `json:"error,omitempty"`
+	}
+
+	result := make(map[string]flagResult)
+	for _, flagName := range flagNames {
+		switch flagName {
+		case flagNoobaa:
+			value, err := checkNoobaaFlag(r.Context(), client, namespace)
+			if err != nil {
+				klog.Errorf("failed to check noobaa flag: %v", err)
+				result[flagName] = flagResult{Value: false, Error: err.Error()}
+			} else {
+				result[flagName] = flagResult{Value: value}
+			}
+		case flagRGW:
+			value, err := checkRGWFlag(r.Context(), client, namespace)
+			if err != nil {
+				klog.Errorf("failed to check rgw flag: %v", err)
+				result[flagName] = flagResult{Value: false, Error: err.Error()}
+			} else {
+				result[flagName] = flagResult{Value: value}
+			}
+		default:
+			result[flagName] = flagResult{Value: false, Error: "unsupported flag"}
+		}
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+
+	allFlagsFailed := true
+	for _, res := range result {
+		if res.Error == "" {
+			allFlagsFailed = false
+			break
+		}
+	}
+
+	// if all flags failed, returning "500" as response
+	// if some flags succeeded, returning "200" with partial results and error details
+	if allFlagsFailed && len(result) > 0 {
+		w.WriteHeader(http.StatusInternalServerError)
+	} else {
+		w.WriteHeader(http.StatusOK)
+	}
+
+	if err := json.NewEncoder(w).Encode(result); err != nil {
+		klog.Errorf("failed to encode response: %v", err)
+		http.Error(w, "Internal server error", http.StatusInternalServerError)
+		return
+	}
+}
+
+func checkNoobaaFlag(ctx context.Context, client ctrlclient.Client, namespace string) (bool, error) {
+	klog.Info("Checking NooBaa CR existence in namespace:", namespace)
+	noobaa := &nbv1a1.NooBaa{}
+	if err := client.Get(ctx, ctrlclient.ObjectKey{Name: util.NoobaaResourceName, Namespace: namespace}, noobaa); err != nil {
+		if ctrlclient.IgnoreNotFound(err) == nil {
+			klog.Info("No NooBaa CR found in namespace:", namespace)
+			return false, nil
+		}
+		return false, fmt.Errorf("failed to get NooBaa CR: %w", err)
+	}
+	klog.Info("Found NooBaa CR in namespace:", namespace)
+
+	klog.Info("Checking for NooBaa StorageClass existence")
+	storageClassList := &storagev1.StorageClassList{}
+	if err := client.List(ctx, storageClassList); err != nil {
+		return false, fmt.Errorf("failed to list StorageClasses: %w", err)
+	}
+
+	for _, sc := range storageClassList.Items {
+		if strings.HasSuffix(sc.Provisioner, util.NoobaaDriverNameSuffix) {
+			klog.Info("Found NooBaa StorageClass:", sc.Name, "with provisioner:", sc.Provisioner)
+			return true, nil
+		}
+	}
+
+	klog.Info("No NooBaa StorageClasses found")
+	return false, nil
+}
+
+func checkRGWFlag(ctx context.Context, client ctrlclient.Client, namespace string) (bool, error) {
+	klog.Info("Checking CephObjectStore existence")
+	cephObjectStoreList := &cephv1.CephObjectStoreList{}
+	if err := client.List(ctx, cephObjectStoreList, ctrlclient.Limit(1)); err != nil {
+		return false, fmt.Errorf("failed to list CephObjectStores: %w", err)
+	}
+
+	if len(cephObjectStoreList.Items) == 0 {
+		klog.Info("No CephObjectStores found")
+		return false, nil
+	}
+	klog.Info("Found CephObjectStore")
+
+	klog.Info("Checking for RGW StorageClass existence")
+	storageClassList := &storagev1.StorageClassList{}
+	if err := client.List(ctx, storageClassList); err != nil {
+		return false, fmt.Errorf("failed to list StorageClasses: %w", err)
+	}
+
+	for _, sc := range storageClassList.Items {
+		if strings.HasSuffix(sc.Provisioner, util.ObcDriverNameSuffix) {
+			klog.Info("Found RGW StorageClass:", sc.Name, "with provisioner:", sc.Provisioner)
+			return true, nil
+		}
+	}
+
+	klog.Info("No RGW StorageClasses found")
+	return false, nil
+}
+
+func handleUnsupportedMethod(w http.ResponseWriter, r *http.Request) {
+	klog.Infof("Only GET method should be used for this endpoint %s", r.URL.Path)
+	w.Header().Set("Content-Type", handlers.ContentTypeTextPlain)
+	w.Header().Set("Allow", "GET")
+	w.WriteHeader(http.StatusMethodNotAllowed)
+
+	if _, err := w.Write([]byte(fmt.Sprintf("Unsupported method: %s", r.Method))); err != nil {
+		klog.Errorf("failed to write data to response writer: %v", err)
+	}
+}
--- a/services/ux-backend/handlers/storages/handler.go
+++ b/services/ux-backend/handlers/storages/handler.go
@@ -1,0 +1,147 @@
+package storages
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	"github.com/red-hat-storage/ocs-operator/v4/controllers/util"
+	"github.com/red-hat-storage/ocs-operator/v4/services/ux-backend/handlers"
+	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
+
+	cephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
+	"k8s.io/klog/v2"
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	deploymentTypeInternal = "internal"
+	deploymentTypeExternal = "external"
+)
+
+func HandleMessage(w http.ResponseWriter, r *http.Request, client ctrlclient.Client, namespace string) {
+	switch r.Method {
+	case "GET":
+		handleGet(w, r, client, namespace)
+	default:
+		handleUnsupportedMethod(w, r)
+	}
+}
+
+// request format: /info/storages
+func handleGet(w http.ResponseWriter, r *http.Request, client ctrlclient.Client, namespace string) {
+	result, err := getStorageInfo(r.Context(), client, namespace)
+	if err != nil {
+		klog.Errorf("failed to get storage info: %v", err)
+		http.Error(w, fmt.Sprintf("Failed to get storage info: %v", err), http.StatusInternalServerError)
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
+
+	if err := json.NewEncoder(w).Encode(result); err != nil {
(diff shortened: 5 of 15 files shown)


leelavg:
@SanjalKatiyar I didn't look at code yet, looking at PR description alone I'm not sure that should be the final, is there a finalized design doc for the API? Some of my questions are:

1. grouping 
- `/` vs `/auth` -> a client doesn't need to know this distinction unless the grouping is in terms of resources but not in terms of privilege levels
- I believe you chose this for ease of representation at oauth-proxy, does that not accept a mounted configmap?
- your api should define the endpoints and what operations possible on that and the middleware (proxy) need to take a decision whether a client should be allowed or not.

2. featureflags
- why should the pathparam (`/featureflags`) and query (`flags`) has redundant information?
- I would only use queries on a list for filtering the resources and not support the splitting of values etc, a simple flat structure could be good
- why can't we follow k8s design, like
`GET /flags` list all the currently available flags with their state
`PUT /flags/noobaa {"enable": true}` activate the flag or some more operations to be done
`PUT /flags/noobaa {"enable": false}` deactivate the flag
`PUT /flags/SOMETHING {"enable": true}` we don't support this flag
- a single user updating flag reflects the UI to all other logged in users, is that the intention or the UI should only be different to that user (I suppose the former)

3. bucket-origin
- better not make the API answer a question, this isn't something like a RPC function call 😅
- return the metadata for a bucket as the response against the name of the bucket? like
`GET /buckets` list all the buckets we are aware of, maybe you can limit to only the names?
`GET /buckets/{name}` respond with the metadata that you have for this bucket and let the client do the processing for any fields if it has to

Please excuse, I don't have all background information of the requirement and just provided general guidelines, open to be corrected!

SanjalKatiyar:
> however, grouping also has merits: 1. proxy normalizes the path, like /auth/../../ would be normalized to / and that review is checked, so it's kinda safe, 2. it is convenient to have all the prefixes checked in a single instance

Yes, instead of multiple redundant entries under `-openshift-delegate-urls` (one per endpoint), idea is to have just one entry and all endpoints which needs similar auth + authr can be defined using that prefix.

> I suggest to have a different name rather than auth, like tenant?

I am not sure, intention behind "/auth/" was that all endpoints under it will just authenticate (hence the prefix).
"/tenant/" could ideally have different access level, so that is not 100% accurate as well.
There could be other suggestions as well like using "/user/" or "/info/", but unless there is a strong objection against using "/auth/", IMHO let's stick to it.

> the other option is doing a POST w/ a body to check for all the required flags since the requirement is not to have multiple calls, unless we'll have more flags and the query value keeps on increasing hitting limits we are fine w/ GET /flags=a,b

We can/will definitely support more flags in future, but I don't think it will ever exceed URL limit (we don't need that many).
Since this endpoint doesn't need to support any "create" operations ("read" only retrievals is the intention here), let's stick to "GET" method only for now. We can always enhance if/when needed, UI is the only consumer of this endpoint.

> agreed to use params GET /bucket/{name} to return {"createdVia": "s3/obc"}
 
Updated @leelavg, PTAL.

services/ux-backend/handlers/bucket/handler.go
```diff
@@ -0,0 +1,97 @@
...
+
+	"github.com/red-hat-storage/ocs-operator/v4/services/ux-backend/handlers"
+	nbv1alpha1 "github.com/noobaa/noobaa-operator/v5/pkg/apis/noobaa/v1alpha1"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	CreationMethodOBC = "obc"
+	CreationMethodS3  = "s3"
```
leelavg: are these being exported?

SanjalKatiyar: no...
should I use camelCase format for the variable name ??

SanjalKatiyar:
> * to me it reads as something to do with authentication functionality like, some client trying to authenticate or get a token etc.

I could think of these and corresponding issues:
`/auth/` - As u mentioned, sounds like "something to do with authentication functionality".
`/tenant/` - AFAIK "tenant" in OpenShift is usually a namespace/project. Also, each tenant can theoretically have diff access levels (authorisation). So, this is misleading as well.
`/public/` or `/open/`- Since we are still authenticating this is not a public/open API.
`/user/` - Sound like user-specific API, in reality it's opposite.
`/info/` - Very generic term, any API can be used for information purpose. 
`/console/` - This is wrong as well. UI is the main/major consumer of entire ux-backend-server, not just single path under it.

That's why I just stuck with `/auth/`. But, let me know if any other suggestion ?? Or, if u still think `/tenant/` is the best prefix. I will update in that case @leelavg.

services/ux-backend/handlers/bucket/handler.go
SanjalKatiyar: updated...

services/ux-backend/handlers/featureflags/handler.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	if err := json.NewEncoder(w).Encode(result); err != nil {
+		klog.Errorf("failed to encode response: %v", err)
+		http.Error(w, "Internal server error", http.StatusInternalServerError)
+		return
+	}
+}
+
+func checkNoobaaFlag(client ctrlclient.Client, namespace string, ctx context.Context) (bool, error) {
```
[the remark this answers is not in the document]
rewantsoni: >If u r suggesting importing this utility into the "/featureflag" handler then: that won't work, since the StorageCluster instance (which this utility depends on) won't cover all possible/potential future use cases for this endpoint.

The method depends on storageCluster to generate the names for CephFilesystem and cephNFS based on storageCluster name and namespace. I think we might be able to reuse it since GetAvailableServices is a util designed to get the services available

If we don't want to reuse, then we know the name of the cephObjectStore and noobaa instance that we create, we don't need to list them we can go ahead and get it

services/ux-backend/handlers/bucket/handler.go
```diff
@@ -0,0 +1,94 @@
...
+	return false, nil
+}
+
+func handleUnsupportedMethod(w http.ResponseWriter, r *http.Request) {
+	klog.Infof("Only GET method should be used for this endpoint %s", r.URL.Path)
+	w.Header().Set("Content-Type", handlers.ContentTypeTextPlain)
+	w.Header().Set("Allow", "GET")
+	w.WriteHeader(http.StatusMethodNotAllowed)
+
+	if _, err := w.Write([]byte(fmt.Sprintf("Unsupported method: %s", r.Method))); err != nil {
```
leelavg: I believe there is a method in fmt package which outputs to io.Reader, like
fmt.Fprintf(w, "msg", ...args)?

services/ux-backend/handlers/bucket/handler.go
```diff
@@ -0,0 +1,94 @@
...
+	} else {
+		creationMethod = creationMethodS3
+	}
+
+	result := map[string]string{
+		"createdVia": creationMethod,
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
```
leelavg: I believe the first write automatically adds the OK header.

services/ux-backend/handlers/bucket/handler.go
```diff
@@ -0,0 +1,97 @@
...
+		http.Error(w, "Internal server error", http.StatusInternalServerError)
+		return
+	}
+}
+
+func checkBucketCreatedViaOBC(client client.Client, bucketName string, ctx context.Context) (bool, error) {
+	klog.Infof("Checking if bucket '%s' was created via OBC", bucketName)
+	
+	objectBucketList := &nbv1alpha1.ObjectBucketList{}
+	if err := client.List(ctx, objectBucketList); err != nil {
```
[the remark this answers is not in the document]
leelavg: > UI won't request data for all buckets at once.
- that's not my point, even if UI is requesting for a single bucket, backend is do a list of all buckets to find that single bucket. so w/o cache + index every request pulls total number of buckets * sizeof each object.

services/ux-backend/handlers/featureflags/handler.go
```diff
@@ -0,0 +1,215 @@
...
+		w.WriteHeader(http.StatusOK)
+	}
+
+	if err := json.NewEncoder(w).Encode(result); err != nil {
+		klog.Errorf("failed to encode response: %v", err)
+		http.Error(w, "Internal server error", http.StatusInternalServerError)
+		return
+	}
+}
+
```
leelavg: too much chaining of functions increasing the call depth. below is my suggestion:
```
func isNoobaaAvailable() {
// do a get on NoobaaCR, it's not about hardcoding it's about not using the info we already know
// do a get on StorageClass with MatchingField on Provisioner (yes, we need cache + index), again we know the name of Provisioner
}

func isRGWAvailable() {
// ditto as above
}
```
ux-backend is an extension of ocs-op and ocs-op is the one creating or dictating the values for the fields that we are interested in and not some other component, not using them is a missed opportunity.

services/ux-backend/handlers/bucket/handler.go
```diff
@@ -0,0 +1,94 @@
...
+func HandleMessage(w http.ResponseWriter, r *http.Request, client client.Client) {
+	switch r.Method {
+	case "GET":
+		handleGet(w, r, client)
+	default:
+		handleUnsupportedMethod(w, r)
+	}
+}
+
+// request format: /auth/bucket/<bucket-name>
```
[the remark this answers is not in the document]
leelavg: ack, you can keep it then.

services/ux-backend/handlers/bucket/handler.go
SanjalKatiyar: leaving it as it is for now, same as other (existing) handlers we have...

services/ux-backend/handlers/featureflags/handler.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+	
+	if len(storageClassList.Items) > 0 {
+		klog.Info("Found RGW StorageClass:", storageClassList.Items[0].Name, "with provisioner:", storageClassList.Items[0].Provisioner)
+		return true, nil
+	}
+	
+	klog.Info("No RGW StorageClasses found")
+	return false, nil
+}
```
rewantsoni: I think this comment was not addressed. I still see chaining in the functions that is increasing the call depth. I think you should let the reviewer resolve the comments https://github.com/red-hat-storage/ocs-operator/pull/3536#discussion_r2468005832

services/ux-backend/handlers/storages/handler.go
```diff
@@ -0,0 +1,143 @@
...
+	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
+
+	cephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
+	"k8s.io/klog/v2"
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	deploymentTypeInternal = "internal"
+	deploymentTypeExternal = "external"
```
rewantsoni: We already have deploymentType defined in [storagecluster/odfInfo](https://github.com/red-hat-storage/ocs-operator/blob/main/controllers/storagecluster/odfinfoconfig.go#L26-L27), let's move it to the utils package and then use it instead of having new constants

services/ux-backend/handlers/storages/handler.go
```diff
@@ -0,0 +1,143 @@
...
+
+	if cos.Status.Info != nil {
+		if endpoint, exists := cos.Status.Info["secureEndpoint"]; exists && endpoint != "" {
+			return endpoint
+		}
+	}
+
+	if len(cos.Status.Endpoints.Secure) > 0 {
+		return cos.Status.Endpoints.Secure[0]
+	}
```
rewantsoni:
```suggestion
	if cos.Status == nil {
		return ""
	}

	if endpoint := cos.Status.Info["secureEndpoint"]; endpoint != "" {
		return endpoint
	}

	if len(cos.Status.Endpoints.Secure) > 0 {
		return cos.Status.Endpoints.Secure[0]
	}
```
Will there be different values at cos.Status.Endpoints.Secure and cos.Status.Info["secureEndpoint"]?

services/ux-backend/handlers/featureflags/handler.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+
+	klog.Info("No NooBaa CRs found in namespace:", namespace)
+	return false, nil
+}
+
+func checkNoobaaStorageClassExists(ctx context.Context, client ctrlclient.Client, namespace string) (bool, error) {
+	klog.Info("Checking for NooBaa StorageClass existence")
+	
+	provisioner := fmt.Sprintf("%s.noobaa.io/obc", namespace)
```
rewantsoni: provisioner name won't depend on namespace it's always hardcoded https://github.com/red-hat-storage/ocs-operator/blob/main/controllers/util/csi.go#L4-L8

services/ux-backend/handlers/featureflags/handler.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+
+	klog.Info("No CephObjectStores found")
+	return false, nil
+}
+
+func checkRGWStorageClassExists(ctx context.Context, client ctrlclient.Client, namespace string) (bool, error) {
+	klog.Info("Checking for RGW StorageClass existence")
+	
+	provisioner := fmt.Sprintf("%s.ceph.rook.io/bucket", namespace)
```
rewantsoni: Provisioner name is hardcoded and doesn't depend on namespace https://github.com/red-hat-storage/ocs-operator/blob/main/controllers/util/csi.go#L4-L8

rewantsoni (changes requested):

services/ux-backend/handlers/featureflags/handler.go
SanjalKatiyar: https://github.com/red-hat-storage/ocs-operator/pull/3536/files#r2476652372

services/ux-backend/handlers/storages/handler.go
SanjalKatiyar: another example of nit-pickyness :) but I will update it because it's cleaner...

> Will there be different values at cos.Status.Endpoints.Secure and cos.Status.Info["secureEndpoint"]

No, it's just a fallback.

services/ux-backend/handlers/featureflags/handler.go
SanjalKatiyar: Yes, I noticed this file.
But, I prefer not to hardcode, unless there is absolutely no other way out.

SanjalKatiyar: Okay, I think there is one case where it can break. What happens in ROSA ?? It is still "openshift-storage", right ?? If so, then probably hardcoding is the only way out.

SanjalKatiyar: Also, [csi.go#L4-L8](https://github.com/red-hat-storage/ocs-operator/blob/main/controllers/util/csi.go#L4-L8) is for Ceph, what about NooBaa, does that use "openshift-storage" prefix in ROSA as well ??

SanjalKatiyar: Another thing, can these provisioner prefix be overridden by users via some spec or env variable or something similar, does that happen in downstream ??

services/ux-backend/handlers/storages/handler.go
rewantsoni: Can there be a case where we can have one field populated and the other one isn't?

services/ux-backend/handlers/storages/handler.go
rewantsoni: `deploymentType` Is not something that's going to change even if they do, they have to be changed in both places. They operate independently, but share the same codebase. I think it would be good not to define them again and move them to the utils package

services/ux-backend/handlers/featureflags/handler.go
SanjalKatiyar: Or, I keep indexing and if a StorageClass is not found using the index, I fallback to list and filter, just to be sure. But, that is not a clean code, rather a mid-way between 2 approaches which is guaranteed to always work in case of edge cases.

services/ux-backend/handlers/featureflags/handler.go
rewantsoni: We shouldn't create unnecessary functions if they aren't required. In this case, you are concerned about whether Noobaa/CephRGW is available or not. I don't think we need to split the logic for CR and storageClass into two different functions. I would argue that we aren't benefiting from it; instead, it's making the code less readable and making me jump from one line to another to see what the code is trying to achieve.

services/ux-backend/handlers/storages/handler.go
rewantsoni: It would be good to know if there is a case, if there can be a scenario like this, and if yes, then which endpoint should we give a priority to? If it cannot be different, then let's use one of the fields. Why write more code than what is required?

services/ux-backend/main.go
```diff
@@ -73,12 +83,68 @@ func main() {
...
+	if err != nil {
+		klog.Exitf("failed to create cache: %v", err)
+	}
+
+	if err := cache.IndexField(ctx, &storagev1.StorageClass{}, "provisioner", func(obj client.Object) []string {
+		sc := obj.(*storagev1.StorageClass)
+		return []string{sc.Provisioner}
+	}); err != nil {
+		klog.Exitf("failed to add provisioner index for StorageClasses: %v", err)
+	}
```
leelavg: I agree an index for provisioner isn't mandatory and can simply loop over at call site.

services/ux-backend/main.go
```diff
@@ -73,12 +83,68 @@ func main() {
...
 
-	cl, err := util.NewK8sClient(scheme)
+	if err := cache.IndexField(ctx, &nbv1alpha1.ObjectBucket{}, "bucketName", func(obj client.Object) []string {
+		ob := obj.(*nbv1alpha1.ObjectBucket)
+		if ob.Spec.Endpoint != nil {
+			return []string{ob.Spec.Endpoint.BucketName}
+		}
+		return []string{}
+	}); err != nil {
```
leelavg: you've mentioned there could be 100 buckets as well, may I known what is the timeout that UI tolerates for the bucket/{name} endpoint?

reason is, by creating a cache we get fast lookup at the cost of increased memory consumption or can UI tolerate the latency of looping through bucket list for each call + num of concurrent users. Since you know the calling pattern better from UI, I want you to decide:

Suppose there are 100 buckets and as each bucket endpoint is unique the indexing overhead is 100 * X for each index value.

Are there any pieces in the bucket name which we can factor out and use a single key as the index, as an example:
b1 - xyz.some.123
b2 - abc.some.345
b3 - asd.some.789

so in all these endpoints `some` is common and we can index on that which doesn't create much overhead.

services/ux-backend/main.go
```diff
@@ -73,12 +83,68 @@ func main() {
...
+
+	if err := cache.IndexField(ctx, &storagev1.StorageClass{}, "provisioner", func(obj client.Object) []string {
+		sc := obj.(*storagev1.StorageClass)
+		return []string{sc.Provisioner}
+	}); err != nil {
+		klog.Exitf("failed to add provisioner index for StorageClasses: %v", err)
+	}
 
-	cl, err := util.NewK8sClient(scheme)
+	if err := cache.IndexField(ctx, &nbv1alpha1.ObjectBucket{}, "bucketName", func(obj client.Object) []string {
```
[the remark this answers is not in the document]
SanjalKatiyar: make sense... will update...

services/ux-backend/main.go
SanjalKatiyar: ack, I will remove indexing on StorageClass then. Thanks for confirming.

services/ux-backend/handlers/bucket/handler.go
```diff
@@ -0,0 +1,94 @@
...
+		http.Error(w, "Internal server error", http.StatusInternalServerError)
+		return
+	}
+}
+
+func checkBucketCreatedViaOBC(client client.Client, bucketName string, ctx context.Context) (bool, error) {
+	klog.Infof("Checking if bucket '%s' was created via OBC", bucketName)
+	
+	objectBucketList := &nbv1alpha1.ObjectBucketList{}
+	if err := client.List(ctx, objectBucketList); err != nil {
```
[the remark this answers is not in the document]
SanjalKatiyar: If by "bucketname" name u mean ObjectBucket CR name, then yes. Underlying actual s3 bucket (in spec) has different name than the CR name.

services/ux-backend/handlers/featureflags/handler.go
SanjalKatiyar: https://github.com/red-hat-storage/ocs-operator/pull/3536#discussion_r2476692314

services/ux-backend/main.go
SanjalKatiyar: Default timeout should be 60s.

UI should tolerate IMO. Still, as per my latest/updated understanding cache is lazily loaded (plz correct me if I am wrong), and we will only need to call this endpoint in those clusters where users have non-admin use case. For users with admin only use cases, those clusters will never call and hence populate the cache with ObjectBuckets.

IIUC, your main concern here is overhead of the index itself ?? Sadly, I don't see any other way for indexing, bucket name is the only information that UI has and it can send in the request to uniquely identify the bucket. But, if u prefer then just like StorageClasses case, we can remove indexing here as well, let me know ??

leelavg: > understanding cache is lazily loaded
- since we don't have watches when a request is made to list bucket only then cache is initially synced with api server for the first time, pls do note this'll incure additional time during first call based on number of buckets

last question:
does the UI use this data in a list mode or individual item mode, like, when a page is displayed with all the buckets does that have a icon/info representing the creation method or it gets displayed only when users drills into a specific bucket.

let's go w/ current implementation and later on (after FF) we can do some simple test to see the memory growth?

services/ux-backend/main.go
leelavg: no, it is not a prefix, that's my point in https://github.com/red-hat-storage/ocs-operator/pull/3536#discussion_r2477182388 as mentioned in the linked comment, did you ask either RGW or MCG folks to confirm? I believe they are not changing it, however better confirm once.

services/ux-backend/handlers/bucket/handler.go
```diff
@@ -0,0 +1,86 @@
...
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	"github.com/red-hat-storage/ocs-operator/v4/services/ux-backend/handlers"
+
+	nbv1alpha1 "github.com/noobaa/noobaa-operator/v5/pkg/apis/noobaa/v1alpha1"
```
leelavg:
```suggestion
	nbv1a1 "github.com/noobaa/noobaa-operator/v5/pkg/apis/noobaa/v1alpha1"
```
nit, change if you are fine & refreshing the PR.

services/ux-backend/handlers/featureflags/handler.go
leelavg: I don't want to force, you are nack'ng the change. let it be, I don't have anymore supporting arguments.

config/rbac/oauth_proxy_role.yaml
```diff
@@ -10,10 +10,34 @@ rules:
...
+  verbs:
+  - get
+  - list
+- apiGroups:
+  - ocs.openshift.io
+  resources:
+  - storageclusters
+  verbs:
+  - get
+  - list
```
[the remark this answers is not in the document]
SanjalKatiyar: Yup, I have yet to update the PR. Thanks for pointing it out, will update and test soon.

services/ux-backend/handlers/featureflags/handler.go
SanjalKatiyar: Resolving this comment earlier was a mistake on my end.

services/ux-backend/handlers/bucket/handler.go
```diff
@@ -0,0 +1,86 @@
...
+	if err := client.List(ctx, objectBucketList, ctrlclient.MatchingFields{"bucketName": bucketName}, ctrlclient.Limit(1)); err != nil {
+		return "", fmt.Errorf("failed to list ObjectBuckets by bucketName index: %w", err)
+	}
+	
+	if len(objectBucketList.Items) > 0 {
+		klog.Infof("Found ObjectBucket '%s' with bucket name '%s'", objectBucketList.Items[0].Name, bucketName)
+		return creationMethodOBC, nil
+	}
+	
+	klog.Infof("No ObjectBucket found with bucket name '%s', assuming S3 creation", bucketName)
```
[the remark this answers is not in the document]
leelavg: the comment isn't helpful. it should mention why we consider the absence of a obc resource as created by s3, it's due to UI calling this api against an existing bucket.

leelavg (commented):
only 2 points remain for me:
1. you only have one noobaa cr with a specific name and list wouldn't capture that intent, sometimes it's erroneous if indeed you have 1 noobaa cr but with a different name and mcg woudl refuse to reconcile it
2. matching the provisioner name by prefix, it doesn't matter what the environment is, it's always `openshift-storage.<our-driver>` pls note this is not a prefix, I just wrote it as there are many drivers. if there is `<something>.<our-driver>` you'd consider that such a storageclass exists which is wrong. (any comment on this statement @Madhu-1? thanks).

rbac/ux_backend_role.yaml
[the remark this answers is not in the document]
SanjalKatiyar: I doubt so. I will test and add if I find any missing RBACs.

`storageclusters` already has cluster-wide list/watch privilege, this file here only adds extra `update` in `openshift-storage` namespace (no need for cluster-wide `update` yet).

`cephblockpools` are also only allowed `create` for now, list/watch not needed yet.

`secrets` are not fetched via endpoint's cached client afaik, used only by the service account.

SanjalKatiyar:
/hold
(need to test once)

SanjalKatiyar:
@leelavg @rewantsoni addressed all comments and did the final testing (updated the PR description with relevant testing details).
PTAL and tag (if it looks good). Thanks !!

PR #3539, opened 2025-10-07
Only create but do not update the rook-ceph-mon-endpoints configMap

malayparida2000:
Right now both ocs-operator and rook are in race updating the configMap. This is not required as rook will update the configMap with the latest mon endpoints which it's capable of discovering. So ocs-operator should only create the configMap if it doesn't exist, then do not update it.


parth-gr:
cc @travisn

PR #3586, opened 2025-11-18
Mark operator as not upgradeable if LSO PVs are using bad symlink paths

malayparida2000:
With a bad symlink path which does not point to a valid UUID, the ODF upgrade will most probably fail & require manual intervention to fix. Hence, we mark the operator as not upgradeable in this case.

### DIFF
--- a/controllers/storagecluster/reconcile.go
+++ b/controllers/storagecluster/reconcile.go
@@ -88,6 +88,9 @@ const (
 	OdfVolumeGroupSnapshotClassCrdName = "volumegroupsnapshotclasses.groupsnapshot.storage.openshift.io"
 
 	internalComponentFinalizer = "ocs.openshift.io/internal-component"
+
+	// LSO namespace
+	LSONamespace = "openshift-local-storage"
 )
 
 var storageClusterFinalizer = "storagecluster.ocs.openshift.io"
@@ -604,6 +607,11 @@ func (r *StorageClusterReconciler) reconcilePhases(
 				notUpgradeableMessages = append(notUpgradeableMessages, fmt.Sprintf("%d connected ODF Client Operators are not up to date", count))
 			}
 
+			if err := checkLSOPVSymlinks(r, instance); err != nil {
+				notUpgradeableReasons = append(notUpgradeableReasons, "BadLSOPVSymlinks")
+				notUpgradeableMessages = append(notUpgradeableMessages, err.Error())
+			}
+
 			if len(notUpgradeableMessages) > 0 {
 				// we are not upgradeable
 				returnErr = r.SetOperatorConditions(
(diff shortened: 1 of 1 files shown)


controllers/storagecluster/reconcile.go
```diff
@@ -1011,3 +1017,33 @@ func getUnsupportedClientsCount(r *StorageClusterReconciler, namespace string) (
 
 	return count, nil
 }
+
+// checkLSOPVSymlinks checks if any LSO provisioned PVs are using bad symlink paths
+func checkLSOPVSymlinks(r *StorageClusterReconciler) error {
+	pvList := &corev1.PersistentVolumeList{}
+	if err := r.Client.List(r.ctx, pvList); err != nil {
```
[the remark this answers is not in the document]
malayparida2000: @Madhu-1 
> Will there be any event or some thing which we can watch from LSO or somewhere if this happens

No event but there are logs inside LSO operator pod, But I don't think that would be possible to watch
https://github.com/openshift/local-storage-operator/blob/release-4.20/pkg/diskmaker/controllers/lvset/reconcile.go#L189

> If the ocs-operator misses this Reconcile or this reconcile wont get triggered, what is the expectation?

We will miss making the operator not upgradable. Nothing apart from that. If the PV has bad path the upgrade would get stuck.

controllers/storagecluster/reconcile.go
```diff
@@ -1010,3 +1018,67 @@ func getUnsupportedClientsCount(r *StorageClusterReconciler, namespace string) (
 
 	return count, nil
 }
+
+// checkLSOPVSymlinks checks if any LSO provisioned PVs used by ODF are using bad paths with device names instead of UUIDs
+func checkLSOPVSymlinks(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) error {
```
Madhu-1: can we please add a unit testing for this function?

controllers/storagecluster/reconcile.go
```diff
@@ -1010,3 +1018,67 @@ func getUnsupportedClientsCount(r *StorageClusterReconciler, namespace string) (
...
+		// Check if it has a local volume source with a bad symlink path
+		if pv.Spec.Local != nil && pv.Spec.Local.Path != "" {
+			// Check if the path's last component matches the device-name annotation
+			// Bad paths end with device names like "sdb" or "sdc"
+			if deviceName, exists := pv.Annotations["storage.openshift.com/device-name"]; exists {
+				pathParts := strings.Split(pv.Spec.Local.Path, "/")
+				lastPathComponent := pathParts[len(pathParts)-1]
+				if lastPathComponent == deviceName {
+					r.Log.Info("Found LSO PV with bad symlink path", "PV", pv.Name, "PVC", pvc.Name, "Path", pv.Spec.Local.Path, "DeviceName", deviceName)
+					return fmt.Errorf("PV %s (used by OSD PVC %s) uses bad symlink path %s (ends with device name %s) instead of UUID-based path. This must be fixed before upgrade", pv.Name, pvc.Name, pv.Spec.Local.Path, deviceName)
```
Madhu-1: we just need to return the first found error, dont we need to return all the problematic PVC's?

malayparida2000:
Closing as per https://issues.redhat.com/browse/RHSTOR-7671?focusedId=28546489&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-28546489

PR #3588, opened 2025-11-18
DFBUGS-4202: [release-4.19] Allow EnableCrushUpdates field to be false by making it a pointer to bool instead of bool

malayparida2000:
manual cherry-pick of https://github.com/red-hat-storage/ocs-operator/pull/3509

The enableCrushUpdates field was a bool field, and after we get the storageCluster, it's impossible to know if the field is set to false because it was not specified or because it was explicitly set to false. Hence, until now, we have always set the field to true.
With this change, the field has now become a pointer to a bool. Hence, the user can set it to false to disable crush updates. If the field is not set, it will be set to true by default by ocs-operator.


travisn (approved):
/lgtm

PR #3697, opened 2026-02-17
DFBUGS-5630: replica1: fix name of the storageclass for external mode

parth-gr:
for replica-1 fix the name of the storageclass for the external mode creation


travisn:
This is intentionally not a cherry-pick of #3625 since we don't need the remaining changes in that PR in 4.20, right? If so, LGTM

subhamkrai:
> cc @travisn @subhamkrai @malayparida2000 please review it and we will get it in once we have the qe ack

LGTM

malayparida2000 (approved):

PR #3700, opened 2026-02-18
zero the nonhashable fields during calculating of desired client state

leelavg:
for every heartbeat, consumer status gets updated with latest timestamp changing the resourceVersion of the resource effectively rendering the heartbeat mechanism for the reconciliation of resources at client less useful and we fix that by zeroing nonhashable fields before calculating hash.

### DIFF
--- a/services/provider/server/server.go
+++ b/services/provider/server/server.go
@@ -393,7 +393,7 @@ func (s *OCSProviderServer) GetDesiredClientState(ctx context.Context, req *pb.G
 
 		desiredClientConfigHash := getDesiredClientConfigHash(
 			channelName,
-			consumer,
+			zerodNonHashableFields(consumer),
 			cephConnection.Spec,
 			isEncryptionInTransitEnabled(storageCluster.Spec.Network),
 			inMaintenanceMode,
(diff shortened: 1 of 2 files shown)


nb-ohad:
After talking with @leelavg offline we decided that the correct approach should be to just ignore the `Status.LastHeartbeatTimestamp` when calculating the hash. This can easily be done by setting the value to 0 before hashing.

services/provider/server/server.go
```diff
@@ -391,10 +391,11 @@ func (s *OCSProviderServer) GetDesiredClientState(ctx context.Context, req *pb.G
 			}
 		}
 
+		zeroNonHashableFields(consumer)
 		desiredClientConfigHash := getDesiredClientConfigHash(
 			channelName,
 			consumer,
-			cephConnection.Spec,
+			cephConnection.GetGeneration(),
```
nb-ohad: As discussed before, we should not be using `generation` as an indicator.

nb-ohad (changes requested):

services/provider/server/server.go
leelavg: this resource has spec and status, using generation in place of spec only reduces the data for calculating the hash with no change in outcome.

services/provider/server/server.go
```diff
@@ -708,6 +710,12 @@ func (s *OCSProviderServer) ReportStatus(ctx context.Context, req *pb.ReportStat
 	}, nil
 }
 
+func zeroNonHashableFields(consumer *ocsv1alpha1.StorageConsumer) {
+	consumer.ManagedFields = nil
+	consumer.ResourceVersion = ""
+	consumer.Status.LastHeartbeat.Time = time.Time{}
+}
```
nb-ohad: By changing the original object we might get into a situation where we break other code that uses the same object (even if the code is not written yet). The safer approach is to make a copy, change the copy, then return the copy

nb-ohad (changes requested):

PR #3731, opened 2026-03-09
external: fix creation of rbd storage classes

ShravaniVangur:
During the creation of CephRBD StorageClasses in external mode:

- If there is a radosnamespace in externalResources, then we create all rbd classes on top of that radosnamespace client profile.
- If there is no radosnamespace, then we should create the classes on top of the default client profile.

Ref: https://github.com/red-hat-storage/ocs-operator/pull/3703#discussion_r2882105705

### DIFF
--- a/internal/controller/storagecluster/external_resources.go
+++ b/internal/controller/storagecluster/external_resources.go
@@ -353,6 +353,11 @@ func (r *StorageClusterReconciler) createExternalStorageClusterResources(instanc
 	rbdStorageID := util.CalculateCephRbdStorageID(fsid, getExternalModeRadosNamespaceName(instance))
 	cephFsStorageID := util.CalculateCephFsStorageID(fsid, "csi")
 
+	rbdClientProfileName, err := r.getExternalModeRbdClientProfileName(instance)
+	if err != nil {
+		return err
+	}
+
 	for _, d := range data {
 		if d.Kind != "StorageClass" {
 			continue
@@ -377,7 +382,7 @@ func (r *StorageClusterReconciler) createExternalStorageClusterResources(instanc
 		} else if d.Name == cephRbdStorageClassName {
 			scc = StorageClassConfiguration{
 				storageClass: util.NewDefaultRbdStorageClass(
-					instance.Namespace,
+					rbdClientProfileName,
 					util.If(
 						util.IsDefaultPoolErasureCodingEnabled(instance.Spec.ManagedResources.CephBlockPools),
 						instance.Spec.ManagedResources.CephBlockPools.ErasureCodedMetadataPool,
@@ -402,7 +407,7 @@ func (r *StorageClusterReconciler) createExternalStorageClusterResources(instanc
 		} else if strings.HasPrefix(d.Name, cephRbdRadosNamespaceStorageClassNamePrefix) { // ceph-rbd-rados-namespace-<radosNamespaceName>
 			scc = StorageClassConfiguration{
 				storageClass: util.NewDefaultRbdStorageClass(
-					instance.Namespace,
+					rbdClientProfileName,
 					util.GenerateNameForCephBlockPool(instance.Name),
 					"rook-csi-rbd-provisioner",
 					"rook-csi-rbd-node",
@@ -430,7 +435,7 @@ func (r *StorageClusterReconciler) createExternalStorageClusterResources(instanc
 			}
 			scc = StorageClassConfiguration{
 				storageClass: util.NewDefaultNonResilientRbdStorageClass(
-					instance.Namespace,
+					rbdClientProfileName,
 					topologyConstrainedPools,
 					"rook-csi-rbd-provisioner",
 					"rook-csi-rbd-node",
@@ -499,7 +504,7 @@ func (r *StorageClusterReconciler) createExternalStorageClusterResources(instanc
 	if crd.UID != "" {
 		scc = StorageClassConfiguration{
 			storageClass: util.NewDefaultVirtRbdStorageClass(
-				instance.Namespace,
+				rbdClientProfileName,
 				util.If(
 					util.IsDefaultPoolErasureCodingEnabled(instance.Spec.ManagedResources.CephBlockPools),
 					instance.Spec.ManagedResources.CephBlockPools.ErasureCodedMetadataPool,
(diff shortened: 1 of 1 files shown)


controllers/storagecluster/external_resources.go
```diff
@@ -262,6 +262,35 @@ func (r *StorageClusterReconciler) createExternalStorageClusterResources(instanc
...
+			OwnerReferences: []metav1.OwnerReference{ownerRef},
+		}
+		radosNamespace := &cephv1.CephBlockPoolRadosNamespace{ObjectMeta: objectMeta}
+		mutateFn := func() error {
+			radosNamespace.Spec = cephv1.CephBlockPoolRadosNamespaceSpec{
+				BlockPoolName: rbdPool,
+			}
+			return nil
+		}
+		_, err := ctrl.CreateOrUpdate(context.TODO(), r.Client, radosNamespace, mutateFn)
```
rewantsoni: Please use the context in the storageClusterReconciler

rewantsoni (changes requested):

internal/controller/storagecluster/external_resources.go
```diff
@@ -476,6 +489,27 @@ func (r *StorageClusterReconciler) createExternalStorageClusterResources(instanc
...
+	if err != nil ||
+		radosNamespace.Status == nil ||
+		radosNamespace.Status.Phase != cephv1.ConditionReady ||
+		radosNamespace.Status.Info["clusterID"] == "" {
+		r.Log.Info("Waiting for radosNamespace to be Ready",
+			"radosNamespace", klog.KRef(namespace, radosNamespaceName),
+		)
+		return namespace
+	}
+	return radosNamespace.Status.Info["clusterID"]
```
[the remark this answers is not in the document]
ShravaniVangur: Previously, considering the example of virt storage class. 
If the cephblookpoolnamespace was created, it was never applied to the virt storage class since it does not have the prefix `ceph-rbd-rados-namespace`. It would have the incorrect namespace. 
With the fix, for each of the rbd storage classes, it is checked if the `cephBlockPoolRadosNamespace` is available or not.

ShravaniVangur:
@rewantsoni @parth-gr could you please review this?

internal/controller/storagecluster/external_resources.go
```diff
@@ -553,6 +566,27 @@ func (r *StorageClusterReconciler) createExternalStorageClusterResources(instanc
 	return nil
 }
 
+// getExternalModeRbdClientProfileName returns the client profile name for RBD storage classes in external mode.
+// If a CephBlockPoolRadosNamespace exists and is ready, its client profile name is returned.
+// If no rados namespace is configured, the default client profile name (namespace) is returned.
+func (r *StorageClusterReconciler) getExternalModeRbdClientProfileName(namespace string) string {
+	if radosNamespaceName == "" {
+		return namespace
+	}
```
rewantsoni, line 574: we have a getExternalModeRadosNamespaceName somewhere in this file that can help us get the radosNamespaceeName for externalMode. It returns "" for default rns.

internal/controller/storagecluster/external_resources.go
```diff
@@ -265,6 +265,35 @@ func (r *StorageClusterReconciler) createExternalStorageClusterResources(instanc
...
+			radosNamespace.Spec = cephv1.CephBlockPoolRadosNamespaceSpec{
+				BlockPoolName: rbdPool,
+			}
+			return nil
+		}
+		_, err := ctrl.CreateOrUpdate(r.ctx, r.Client, radosNamespace, mutateFn)
+		if err != nil {
+			r.Log.Error(err, "Could not create CephBlockPoolRadosNamespace.", "CephBlockPoolRadosNamespace", klog.KRef(radosNamespace.Namespace, radosNamespace.Name))
+			return err
+		}
```
rewantsoni: with https://github.com/red-hat-storage/ocs-operator/pull/3771, we already moved the storageClass creation to a second loop, I don't think we need to separate it more. We could get the radosNamespaceName in the first loop

rewantsoni (changes requested):

internal/controller/storagecluster/external_resources.go
parth-gr: yes try to rebase with this

internal/controller/storagecluster/external_resources.go
```diff
@@ -553,6 +553,25 @@ func (r *StorageClusterReconciler) createExternalStorageClusterResources(instanc
 	return nil
 }
 
+// getExternalModeRbdClientProfileName returns the client profile name for RBD storage classes in external mode.
+// If a CephBlockPoolRadosNamespace exists and is ready, its client profile name is returned.
+// If no rados namespace is configured, the default client profile name (namespace) is returned.
+func (r *StorageClusterReconciler) getExternalModeRbdClientProfileName(instance *ocsv1.StorageCluster) string {
+	radosNamespaceName := getExternalModeRadosNamespaceName(instance)
```
rewantsoni, line 565: getExternalModeRadosNamespaceName can return "" string for default rns

internal/controller/storagecluster/external_resources.go
```diff
@@ -553,6 +553,25 @@ func (r *StorageClusterReconciler) createExternalStorageClusterResources(instanc
...
+	radosNamespace := cephv1.CephBlockPoolRadosNamespace{}
+	err := r.Get(r.ctx, types.NamespacedName{Name: radosNamespaceName, Namespace: instance.Namespace}, &radosNamespace)
+	if err != nil ||
```
rewantsoni:
```suggestion
	radosNamespace := &cephv1.CephBlockPoolRadosNamespace{}
	radosNamespace.Name = radosNamespaceName
	radosNamespace.Namespace = instance.Namespace
	if err := r.Get(r.ctx, client.ObjectKeyFromObject(radosNamespace), radosNamespace); err != nil {
```

rewantsoni (changes requested):

internal/controller/storagecluster/external_resources.go
ShravaniVangur: Isn't this the current behaviour?

ShravaniVangur:
@iamniting @malayparida2000 @leelavg  Could you please review and approve?
Thank you.

internal/controller/storagecluster/external_resources.go
```diff
@@ -561,28 +588,8 @@ func (r *StorageClusterReconciler) createExternalModeStorageClasses(sccs []Stora
...
-					"StorageClass", klog.KRef("", sc.Name),
-				)
-				skippedSC = append(skippedSC, sc.Name)
-				continue
-			}
-			sc.Parameters["clusterID"] = radosNamespace.Status.Info["clusterID"]
-		case strings.Contains(sc.Name, "-nfs") || strings.Contains(sc.Provisioner, util.NfsDriverName):
-			// wait for CephNFS to be ready
+		// wait for CephNFS to be ready
+		if strings.Contains(sc.Name, "-nfs") || strings.Contains(sc.Provisioner, util.NfsDriverName) {
```
leelavg: why can't this also be pushed one level up, the pr is cleanup up some tweaks required based on naming but nfs is the only thing left behind in this func but should be supplied as args and have this func only create the classes? wdyt?

ShravaniVangur: Looking into the creation of storage classes, in external mode, StorageClasses are only created for CephFS and CephRBD-related  and is not supported for NFS storage classes. Hence this part of the code is unreachable and is removed. 
Context for this change is here: https://rook.io/docs/rook/latest-release/Storage-Configuration/NFS/nfs-csi-driver/#consuming-nfs-from-an-external-source

ShravaniVangur: PTAL @parth-gr @rewantsoni @leelavg

leelavg (approved):

internal/controller/storagecluster/external_resources.go
```diff
@@ -553,6 +558,28 @@ func (r *StorageClusterReconciler) createExternalStorageClusterResources(instanc
 	return nil
 }
 
+// getExternalModeRbdClientProfileName returns the client profile name for RBD storage classes in external mode.
+// If a CephBlockPoolRadosNamespace exists and is ready, its client profile name is returned.
+// If no rados namespace is configured, the default client profile name (namespace) is returned.
+func (r *StorageClusterReconciler) getExternalModeRbdClientProfileName(instance *ocsv1.StorageCluster) (string, error) {
```
parth-gr, line 564: is the function name okay?
cc @rewantsoni 

we are actually re-turning the cluster-id

rewantsoni: Aren't both clientProfile name and clusterId name the same thing?

parth-gr: Not sure, can we please re-check,
How the client profile is calaculted for internal mode

rewantsoni: For internal mode, we have a configmap per consumer which contains the clientProfile name
For external mode, rook creates the clientProfile with this name

parth-gr: ok

PR #3767, opened 2026-04-02
transfer ocs-tls config to noobaa

leelavg:
overall flow:
user creates a TLSProfile CR in the operator namespace
storagecluster watches the CR and transfers the config to Noobaa CR during reconciliation
if either the selector or the CR is not found the config is removed from the operands


nb-ohad:
@BlaineEXE The OCP APIs are too limiting to our needs. We defined our own DF layer apis that allow the configurability range and agility we need. The expected behavior is that DF layer will act as a translator to apply the desired state in this new API to the different DF/OCS component through APIs that they defined for their upstream project. If a component want to support the DF API directly I will not oppose it as some components (e.g. console plugin) are DF specific anyways

api/go.mod
```diff
@@ -122,6 +125,8 @@ require (
 	sigs.k8s.io/yaml v1.6.0 // indirect
 )
 
+replace github.com/noobaa/noobaa-operator/v5 v5.21.0 => github.com/noobaa/noobaa-operator/v5 v5.0.0-20260409182054-d38322829145
```
nb-ohad, line 128: Is there a reason we need to have this replace where it was not there before?

internal/controller/storagecluster/noobaa_system_reconciler.go
```diff
@@ -272,6 +276,65 @@ func (r *StorageClusterReconciler) setNooBaaDesiredState(nb *nbv1.NooBaa, sc *oc
 	isEnabled, rotationSchedule := util.GetKeyRotationSpec(sc)
 	nb.Spec.Security.KeyManagementService.EnableKeyRotation = isEnabled
 	nb.Spec.Security.KeyManagementService.Schedule = rotationSchedule
+
+	tlsProfile := &ocstlsv1.TLSProfile{}
+	tlsProfile.Name = defaults.TLSProfileName
+	tlsProfile.Namespace = r.OperatorNamespace
+	if err := r.Get(r.ctx, client.ObjectKeyFromObject(tlsProfile), tlsProfile); client.IgnoreNotFound(err) != nil {
```
nb-ohad: It make zero sense to load the profile here. This same profile is also needed for other parts of the storage cluster reconcile, as such i would suggest to load it in the main reconcile code (where we load the storage cluster) then pass it into the specific reconcilers that needs it

internal/controller/storagecluster/noobaa_system_reconciler.go
```diff
@@ -272,6 +276,65 @@ func (r *StorageClusterReconciler) setNooBaaDesiredState(nb *nbv1.NooBaa, sc *oc
...
+		return err
+	} else if tlsProfile.UID == "" {
+		nb.Spec.Security.APIServerSecurity = nil
+		return nil
+	}
+	if cfg, found := ocstlsv1.GetConfigForServer(tlsProfile, "noobaa.io", ""); found {
+		gotls, err := ocstlsv1.ValidateAndGetGoTLSConfig(cfg)
```
nb-ohad: Why do we do the validate and the get as one operation? I would expect that we will have 2 different operations: ValidateTLSConfig(cfg) then error handling then GetGoTLSConfig(cfg)

internal/controller/storagecluster/storagecluster_controller.go
```diff
@@ -332,7 +333,8 @@ func (r *StorageClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		).
 		Watches(&ocsclientv1a1.StorageClient{}, enqueueStorageClusterRequest).
 		// Watch Nodes with OCS label to trigger Blackbox Probe updates
-		Watches(&corev1.Node{}, enqueueStorageClusterRequest, builder.WithPredicates(ocsNodeLabelPredicate))
+		Watches(&corev1.Node{}, enqueueStorageClusterRequest, builder.WithPredicates(ocsNodeLabelPredicate)).
+		Watches(&ocstlsv1.TLSProfile{}, enqueueStorageClusterRequest, builder.WithPredicates(predicate.GenerationChangedPredicate{}))
```
nb-ohad: If the name is known ahead of time, wouldn't it make sense to have a Name predicate?

services/provider/api/go.mod
```diff
@@ -98,5 +99,7 @@ require (
 	sigs.k8s.io/yaml v1.6.0 // indirect
 )
 
+replace github.com/noobaa/noobaa-operator/v5 v5.21.0 => github.com/noobaa/noobaa-operator/v5 v5.0.0-20260409182054-d38322829145
```
nb-ohad, line 102: Same as a previous comment. We didn't need to have a replace in the past, why do we need it now

nb-ohad (changes requested):

api/go.mod
leelavg: yes, the tagged release doesn't contain the api change.

services/provider/api/go.mod
leelavg: yes, the tagged release doesn't contain the api change.

romayalon:
@leelavg 
I created this PR https://github.com/noobaa/noobaa-operator/pull/1891/
inorder to use your new repo as @nb-ohad mentioned it can be useful for us on slack - it should be reviewed and merged today, please take our changes after

internal/controller/storagecluster/noobaa_system_reconciler.go
leelavg: https://github.com/red-hat-storage/ocs-tls-profiles/pull/9

api/go.mod
nb-ohad: Ok, but isn't that a little bit problematic, as we relay on a hush instead of a tagged release?

leelavg: no, out of coincidence we got onto a hash or else we usually pick at a specific commit. the replace here is due to tagged release being 5.21 series and the api on main branch is only tagged at 5.20, either this or I need to follow up with noobaa to get this api backported or tag the main.

nb-ohad: ok, resolving

PR #3784, opened 2026-04-09
 rgw: Support for SSE-S3 with vault Agent

sp98:
Adds support for RGW SSE-S3 encryption with Vault by deploying a standalone Vault Agent as a Deployment and Service managed by ODF. RGW pods connect to it over the in-cluster service.

Testing: 

- Added env variable to the ocs operator CSV file. 

```
 oc get pods | grep vault                                                                                                                                                                                        ─╯
vault-agent-rgw-5c94cc4885-27pvw                                  1/1     Running     0          3h37m
vault-agent-rgw-5c94cc4885-ftxqr                                  1/1     Running     0          3h37m
```

```
oc get service | grep agent                                                                                                                                                                                     ─╯
vault-agent-rgw                                    ClusterIP      172.30.243.9     <none>                                                                    8100/TCP                                                   4h22m
```

```
oc get cm vault-agent-rgw-config -o yaml                                                                                                                                                                        ─╯
apiVersion: v1
data:
  vault-agent-config.hcl: |
    disable_mlock = true

    vault {
      address = "https://vault.qe.rh-ocs.com:8200"
      tls_config {
        ca_cert     = "/vault/tls/ca/cert"
        client_cert = "/vault/tls/client-cert/cert"
        client_key  = "/vault/tls/client-key/key"
      }
    }

    auto_auth {
      method "kubernetes" {
        mount_path = "auth/kubernetes"
        config = {
          role = "rook-ceph-rgw"
        }
      }
      sink "file" {
        config = {
          path = "/vault/token/.vault-token"
        }
      }
    }

    cache {
      use_auto_auth_token = true
    }

    listener "tcp" {
      address     = "0.0.0.0:8100"
      tls_disable = true
    }
kind: ConfigMap
metadata:
  creationTimestamp: "2026-04-09T09:00:48Z"
  labels:
    app: vault-agent-rgw
  name: vault-agent-rgw-config
  namespace: openshift-storage
  ownerReferences:
    blockOwnerDeletion: true
    controller: true
    kind: StorageCluster
    name: ocs-storagecluster
    uid: 7d96bf81-0977-4f96-9123-ff76fa5ec619
  resourceVersion: "173980"
  uid: ceee11e9-86d2-43c0-adbc-d4a7e9b5a7fc
```

cephobjectstore CR updated by the OCS operator to add S3 credentials: 
```
....
security:
      keyRotation:
        enabled: false
      kms: {}
      s3:
        connectionDetails:
          KMS_PROVIDER: vault
          VAULT_ADDR: http://vault-agent-rgw.openshift-storage.svc:8100
          VAULT_AUTH_METHOD: agent
          VAULT_SECRET_ENGINE: transit
```

RGW daemon pods were updated (by rook)  to add the new crypt args while running rgw daemon: 
```
 - --rgw-crypt-sse-s3-backend=vault
    - --rgw-crypt-sse-s3-vault-addr=http://vault-agent-rgw.openshift-storage.svc:8100
    - --rgw-crypt-sse-s3-vault-auth=agent
    - --rgw-crypt-sse-s3-vault-prefix=/v1/transit
    - --rgw-crypt-sse-s3-vault-secret-engine=transit
 ```

Upload encrypted object: 
```
 aws s3api put-object --bucket sse-s3-test --key test-object --server-side-encryption AES256
{
    "ETag": "\"eba30a8547b6193c006a46e98b73c092\"",
    "ChecksumCRC64NVME": "0yjPg1yyVZE=",
    "ChecksumType": "FULL_OBJECT",
    "ServerSideEncryption": "AES256"
}
```

 Verifying encryption metadata (HEAD)
```
aws s3api head-object --bucket sse-s3-test --key test-object
{
    "AcceptRanges": "bytes",
    "LastModified": "2026-04-09T13:15:36+00:00",
    "ContentLength": 73,
    "ETag": "\"eba30a8547b6193c006a46e98b73c092\"",
    "ContentType": "binary/octet-stream",
    "ServerSideEncryption": "AES256",
    "Metadata": {}
}
```

Download encrypted object: 
```
aws s3api get-object --bucket sse-s3-test --key test-object
{
    "AcceptRanges": "bytes",
    "LastModified": "2026-04-09T13:15:36+00:00",
    "ContentLength": 73,
    "ETag": "\"eba30a8547b6193c006a46e98b73c092\"",
    "ChecksumCRC64NVME": "0yjPg1yyVZE=",
    "ChecksumType": "FULL_OBJECT",
    "ContentType": "binary/octet-stream",
    "ServerSideEncryption": "AES256",
    "Metadata": {}
}```

### DIFF
--- a/internal/controller/storagecluster/cephobjectstores.go
+++ b/internal/controller/storagecluster/cephobjectstores.go
@@ -234,37 +234,63 @@ func (r *StorageClusterReconciler) newCephObjectStoreInstances(initData *ocsv1.S
 		// Set default values in the metadata pool spec as necessary
 		setDefaultMetadataPoolSpec(&obj.Spec.MetadataPool, initData)
 
-		// if kmsConfig is not 'nil', add the KMS details to ObjectStore spec
+		// if kmsConfig is not 'nil', add encryption details to ObjectStore spec.
+		// SSE-KMS and SSE-S3 are mutually exclusive so configure one or the other.
 		if kmsConfigMap != nil {
-
-			// skip if KMS_PROVIDER is ibmkeyprotect or VAULT_AUTH_METHOD is kubernetes, not supported for RGW
-			if kmsConfigMap.Data["KMS_PROVIDER"] == IbmKeyProtectKMSProvider ||
-				kmsConfigMap.Data["KMS_PROVIDER"] == ThalesKMSProvider ||
-				kmsConfigMap.Data["VAULT_AUTH_METHOD"] == VaultSAAuthMethod {
-				r.Log.Info("IBMKeyProtect/Thales as KMS provider or Vault authentication via Service Account is unsupported configuration for RGW KMS, hence skipping")
+			// IBMKeyProtect and Thales are unsupported KMS providers for RGW, skip entirely
+			if kmsConfigMap.Data[KMSProviderKey] == IbmKeyProtectKMSProvider ||
+				kmsConfigMap.Data[KMSProviderKey] == ThalesKMSProvider {
+				r.Log.Info("IBMKeyProtect/Thales as KMS provider is unsupported for RGW, skipping")
 				continue
 			}
-			// Set default KMS_PROVIDER and VAULT_SECRET_ENGINE values, refer https://issues.redhat.com/browse/RHSTOR-1963
-			if _, ok := kmsConfigMap.Data["KMS_PROVIDER"]; !ok {
-				kmsConfigMap.Data["KMS_PROVIDER"] = VaultKMSProvider
+
+			// Set default KMS_PROVIDER, refer https://issues.redhat.com/browse/RHSTOR-1963
+			if _, ok := kmsConfigMap.Data[KMSProviderKey]; !ok {
+				kmsConfigMap.Data[KMSProviderKey] = VaultKMSProvider
 			}
-			rgwConnDetails := make(map[string]string)
-			for key, value := range kmsConfigMap.Data {
-				// ignore kv engine related options
-				if key == "VAULT_BACKEND_PATH" || key == "VAULT_BACKEND" {
+
+			switch kmsConfigMap.Data[VaultRGWAuthMethodKey] {
+			case "":
+				// VAULT_RGW_AUTH_METHOD not set, skip RGW encryption
+				r.Log.Info("VAULT_RGW_AUTH_METHOD not set, skipping RGW encryption configuration")
+			case VaultAgentAuthMethod:
+				if r.images.VaultAgent == "" {
+					r.Log.Info("VAULT_AGENT_IMAGE not set, skipping SSE-S3 Vault Agent configuration for RGW")
 					continue
 				}
-				rgwConnDetails[key] = value
-			}
-			// overwrite SecretEngine value to transit
-			rgwConnDetails["VAULT_SECRET_ENGINE"] = "transit"
-			obj.Spec.Security = &cephv1.ObjectStoreSecuritySpec{
-				SecuritySpec: cephv1.SecuritySpec{
-					KeyManagementService: cephv1.KeyManagementServiceSpec{
-						ConnectionDetails: rgwConnDetails,
-						TokenSecretName:   KMSTokenSecretName,
+				// Configure SSE-S3 with Vault Agent auth.
+				// RGW connects to the ODF-managed Vault Agent service instead of directly to Vault.
+				obj.Spec.Security = &cephv1.ObjectStoreSecuritySpec{
+					ServerSideEncryptionS3: cephv1.KeyManagementServiceSpec{
+						ConnectionDetails: map[string]string{
+							"KMS_PROVIDER":        VaultKMSProvider,
+							"VAULT_ADDR":          VaultAgentServiceURL(initData.Namespace),
+							"VAULT_AUTH_METHOD":   VaultAgentAuthMethod,
+							"VAULT_SECRET_ENGINE": "transit",
+						},
 					},
-				},
+				}
+			case VaultTokenAuthMethod:
+				// Configure SSE-KMS with token auth
+				rgwConnDetails := make(map[string]string)
+				for key, value := range kmsConfigMap.Data {
+					if key == "VAULT_BACKEND_PATH" || key == "VAULT_BACKEND" {
+						continue
+					}
+					rgwConnDetails[key] = value
+				}
+				rgwConnDetails["VAULT_SECRET_ENGINE"] = "transit"
+				obj.Spec.Security = &cephv1.ObjectStoreSecuritySpec{
+					SecuritySpec: cephv1.SecuritySpec{
+						KeyManagementService: cephv1.KeyManagementServiceSpec{
+							ConnectionDetails: rgwConnDetails,
+							TokenSecretName:   KMSTokenSecretName,
+						},
+					},
+				}
+			default:
+				return nil, fmt.Errorf("unsupported %s value %q, must be %q or %q",
+					VaultRGWAuthMethodKey, kmsConfigMap.Data[VaultRGWAuthMethodKey], VaultAgentAuthMethod, VaultTokenAuthMethod)
 			}
 		}
 
--- a/internal/controller/storagecluster/storagecluster_controller.go
+++ b/internal/controller/storagecluster/storagecluster_controller.go
@@ -76,6 +76,15 @@ func (r *StorageClusterReconciler) initializeImageVars() error {
 		err := fmt.Errorf("BLACKBOX_EXPORTER_IMAGE environment variable not found")
 		r.Log.Error(err, "BLACKBOX_EXPORTER_IMAGE environment variable not set; will use default image")
 	}
+
+	// TODO(GA): VAULT_AGENT_IMAGE is not set by default in the CSV (dev preview).
+	// For GA, add a downstream vault image to the CSV env vars and relatedImages
+	// so that SSE-S3 with Vault Agent works out of the box.
+	r.images.VaultAgent = os.Getenv("VAULT_AGENT_IMAGE")
+	if r.images.VaultAgent == "" {
+		r.Log.Info("VAULT_AGENT_IMAGE environment variable not set; Vault Agent deployment for RGW SSE-S3 will not be available")
+	}
+
 	return nil
 }
 
@@ -87,6 +96,7 @@ type ImageMap struct {
 	OCSMetricsExporter string
 	BlackboxExporter   string
 	KubeRBACProxy      string
+	VaultAgent         string
 }
 
 // StorageClusterReconciler reconciles a StorageCluster object
--- a/internal/controller/storagecluster/vault_agent.go
+++ b/internal/controller/storagecluster/vault_agent.go
@@ -1,0 +1,495 @@
+package storagecluster
+
+import (
+	"fmt"
+	"strings"
+
+	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
+	"github.com/red-hat-storage/ocs-operator/v4/pkg/defaults"
+	"github.com/red-hat-storage/ocs-operator/v4/pkg/util"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+const (
+	vaultAgentDefaultRole          = "rook-ceph-rgw"
+	vaultAgentDefaultAuthMountPath = "auth/kubernetes"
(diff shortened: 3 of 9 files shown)


hack/generate-latest-csv.sh
```diff
@@ -17,6 +17,7 @@ export OCS_METRICS_EXPORTER_IMAGE=${OCS_METRICS_EXPORTER_IMAGE:-${METRICS_EXPORT
 export BLACKBOX_EXPORTER_IMAGE=${BLACKBOX_EXPORTER_IMAGE:-${LATEST_BLACKBOX_EXPORTER_IMAGE}}
 export KUBE_RBAC_PROXY_IMAGE=${KUBE_RBAC_PROXY_IMAGE:-${KUBE_RBAC_PROXY_FULL_IMAGE_NAME}}
 export OCS_MUST_GATHER_IMAGE=${OCS_MUST_GATHER_IMAGE:-${LATEST_MUST_GATHER_IMAGE}}
+export VAULT_AGENT_IMAGE=${VAULT_AGENT_IMAGE:-${LATEST_VAULT_AGENT_IMAGE}}
```
BlaineEXE: I think we might benefit from an "off switch" based on this config. 

Let's imagine the case where a user with an unexpected corner case has issues with this new feature being enabled in their cluster. With the code as-is, we might have to make a hotfix for them to turn off the feature and unblock their cluster.

However, if we have an implicit off switch that disables this feature when this value is set to the empty string, we can tell the user to set the empty string to unblock their issue, which they could use to become unblocked in much less time.

The likelihood that this feature breaks a customer's cluster is probably low, but I still think it worth considering.

internal/controller/storagecluster/cephobjectstores.go
```diff
@@ -234,37 +234,59 @@ func (r *StorageClusterReconciler) newCephObjectStoreInstances(initData *ocsv1.S
...
-				// ignore kv engine related options
-				if key == "VAULT_BACKEND_PATH" || key == "VAULT_BACKEND" {
-					continue
+
+			switch kmsConfigMap.Data[VaultRGWAuthMethodKey] {
+			case "":
+				// VAULT_RGW_AUTH_METHOD not set, skip RGW encryption
+				r.Log.Info("VAULT_RGW_AUTH_METHOD not set, skipping RGW encryption configuration")
+			case VaultAgentAuthMethod:
+				// Configure SSE-S3 with Vault Agent auth.
```
BlaineEXE, line 261: Continuing from my previous thought, we could allow a semi-hidden off switch like this

```suggestion
			case VaultAgentAuthMethod:
				if kmsConfigMap.Data["VAULT_AGENT_IMAGE"] == "" {
				    r.Log.Info("skipping SSE-S3 setup because agent image is unset")
				    continue
				}
				// Configure SSE-S3 with Vault Agent auth.
```

internal/controller/storagecluster/cephobjectstores.go
```diff
@@ -234,37 +234,59 @@ func (r *StorageClusterReconciler) newCephObjectStoreInstances(initData *ocsv1.S
...
-			}
-			// overwrite SecretEngine value to transit
-			rgwConnDetails["VAULT_SECRET_ENGINE"] = "transit"
-			obj.Spec.Security = &cephv1.ObjectStoreSecuritySpec{
-				SecuritySpec: cephv1.SecuritySpec{
-					KeyManagementService: cephv1.KeyManagementServiceSpec{
-						ConnectionDetails: rgwConnDetails,
-						TokenSecretName:   KMSTokenSecretName,
+			case VaultTokenAuthMethod:
+				// Configure SSE-KMS with token auth
```
BlaineEXE, line 274: Off switch?

internal/controller/storagecluster/vault_agent.go
```diff
@@ -0,0 +1,491 @@
...
+							},
+							LivenessProbe: &corev1.Probe{
+								ProbeHandler: corev1.ProbeHandler{
+									TCPSocket: &corev1.TCPSocketAction{
+										Port: intstr.FromInt(VaultAgentPort),
+									},
+								},
+								InitialDelaySeconds: 10,
+								PeriodSeconds:       15,
+							},
```
BlaineEXE, line 300: Just a thought that sometimes users have unexpectedly slow networks or other issues that cause grief with liveness/readiness probes that we have to fix in z-streams. I would suggest making these more lenient than we currently think is necessary, especially allowing for a number of failures before rebooting a pod.

internal/controller/storagecluster/vault_agent.go
```diff
@@ -0,0 +1,491 @@
...
+			Selector: &metav1.LabelSelector{
+				MatchLabels: vaultAgentLabels,
+			},
+			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: vaultAgentLabels,
+				},
+				Spec: corev1.PodSpec{
+					ServiceAccountName: VaultAgentSAName,
+					Containers: []corev1.Container{
```
BlaineEXE, line 252: I don't believe DF is required to provide resource requests/limits to all pod containers, but I want to double check.

@travisn or @iamniting do you remember if we have to specify requests?

BlaineEXE (commented):
Some thoughts for consideration. I don't think anything I flagged is a hard requirement.

internal/controller/storagecluster/vault_agent.go
travisn: I don't believe there is a hard requirement, but we are generally expected to apply resource requests/limits, and make them configurable similar to other components.

hack/generate-latest-csv.sh
sp98: If a user sets VAULT_AGENT_IMAGE to empty string on the operator deployment, the feature is cleanly disabled at every layer. 

- Controller init (storagecluster_controller.go): we are just logging if the VAULT_AGENT_IMAGE env is empty and not throwing any error. So essentially treating this as non essential image. 
 
- Reconciler (vault_agent.go): We don't reconcile the vault agent deployment if the VAULT_AGENT_IMAGE is empty. So users can delete the deployment after removing the VAULT_AGENT_IMAGE env. 
 
- CSV merger (csv-merger.go): Only includes vault-agent in relatedImages if the image is non-empty

- For this review [comment](https://github.com/red-hat-storage/ocs-operator/pull/3784#discussion_r3059480586), added changes to not configure SSE-S3 spec in the CephObjectStoreCR when `VAULT_AGENT_IMAGE` is removed.

internal/controller/storagecluster/vault_agent.go
sp98: That's a good point. We have seen such issues a number of times.  Updated to make the liveness/readiness probes a bit more lenient.

internal/controller/storagecluster/cephobjectstores.go
sp98: This is a token based auth. Does not rely of the VAULT_AGENT_IMAGE variable. We might to think of another way to handle the off switch for this option. But maybe in a separate PR because token based auth is an old behavior.

deploy/ocs-operator/manifests/ocs-operator.clusterserviceversion.yaml
```diff
@@ -798,4 +800,6 @@ spec:
     name: blackbox-exporter
   - image: gcr.io/kubebuilder/kube-rbac-proxy:v0.13.1
     name: kube-rbac-proxy
+  - image: hashicorp/vault:latest
```
travisn: As discussed, sounds good to remove this image from the csv, and the customer can specify the vault image

sp98: Done. The VAULT_AGENT_IMAGE env var is not set in ODF by default. Customer who want SSE-S3 with vault can set it themselves in their CSV YAML. The controller code is handling the empty case  where it just logs a message and skips vault agent configuration. This should be good for dev preview.

internal/controller/storagecluster/vault_agent.go
```diff
@@ -0,0 +1,494 @@
...
+			return reconcile.Result{}, err
+		}
+	}
+	return reconcile.Result{}, nil
+}
+
+// isVaultAgentRequired checks whether the Vault Agent deployment is needed
+func (obj *ocsVaultAgent) isVaultAgentRequired(r *StorageClusterReconciler, instance *ocsv1.StorageCluster) bool {
+	if !util.IsClusterOrDeviceSetEncrypted(instance) {
+		return false
```
iamniting:
```suggestion
	if util.IsClusterOrDeviceSetEncrypted(instance) {
		return true
```

internal/controller/storagecluster/vault_agent.go
```diff
@@ -0,0 +1,494 @@
...
+	}
+	if !instance.Spec.Encryption.KeyManagementService.Enable {
+		return false
```
iamniting:
```suggestion
	if instance.Spec.Encryption.KeyManagementService.Enable {
		return true
```

internal/controller/storagecluster/vault_agent.go
```diff
@@ -0,0 +1,494 @@
...
+	}
+
+	kmsProvider := kmsConfigMap.Data[KMSProviderKey]
+	if kmsProvider == "" {
+		kmsProvider = VaultKMSProvider
+	}
+	if kmsProvider != VaultKMSProvider {
+		return false
+	}
```
iamniting:
```suggestion
	kmsProvider := kmsConfigMap.Data[KMSProviderKey]
	if kmsProvider == "" {
		return true
	}
```

internal/controller/storagecluster/vault_agent.go
```diff
@@ -0,0 +1,494 @@
...
+	kmsConfigMap, err := util.GetKMSConfigMap(defaults.KMSConfigMapName, instance, r.Client)
+	if err != nil {
+		r.Log.Error(err, "Failed to get KMS ConfigMap for Vault Agent")
+		return reconcile.Result{}, err
+	}
+	if kmsConfigMap == nil {
+		return reconcile.Result{}, nil
+	}
+
+	ctx := context.TODO()
```
iamniting: Can you pls use the ctx from `r`

iamniting (changes requested):

internal/controller/storagecluster/vault_agent.go
iamniting: Yes, there is no hard requirement, but it is good to have that so we can calculate the resources required based on the requests/limits set in the pods.

internal/controller/storagecluster/vault_agent.go
sp98: @iamniting The suggestions for this method won't work. The current logic is **AND** . It requires all the conditions (encryption in spec + KMS in spec + vault provider + agent auth method) to satisfied before we can deploy the vault agent daemon. 

The suggested approach would deploy vault agent if ANY of encryption/KMS is enabled, regardless of the other conditions. We don't want that.

internal/controller/storagecluster/vault_agent.go
sp98: same as https://github.com/red-hat-storage/ocs-operator/pull/3784#discussion_r3093637182

internal/controller/storagecluster/vault_agent.go
sp98: same as https://github.com/red-hat-storage/ocs-operator/pull/3784#discussion_r3093637182

internal/controller/storagecluster/vault_agent.go
sp98: done.

internal/controller/storagecluster/vault_agent.go
sp98: updated the function to remove the negative checks. Also add function comment.

internal/controller/storagecluster/cephobjectstores.go
BlaineEXE: My bad. I thought I was making a note on the agent auth method here.

internal/controller/storagecluster/storagecluster_controller.go
```diff
@@ -76,6 +76,15 @@ func (r *StorageClusterReconciler) initializeImageVars() error {
...
 		r.Log.Error(err, "BLACKBOX_EXPORTER_IMAGE environment variable not set; will use default image")
 	}
+
+	// TODO(GA): VAULT_AGENT_IMAGE is not set by default in the CSV (dev preview).
+	// For GA, add a downstream vault image to the CSV env vars and relatedImages
+	// so that SSE-S3 with Vault Agent works out of the box.
+	r.images.VaultAgent = os.Getenv("VAULT_AGENT_IMAGE")
+	if r.images.VaultAgent == "" {
+		r.Log.Info("VAULT_AGENT_IMAGE environment variable not set; Vault Agent deployment for RGW SSE-S3 will not be available")
+	}
```
BlaineEXE, line 86: I think the plan is that a KCS will doc for the user how to enable this. If not, let's plan for that. But this doesn't affect merge-ability of this pr.

BlaineEXE (approved):
Looks good to me. I added a note to make sure our plans for KCS and enabling the feature are recorded for other reviewers.

internal/controller/storagecluster/storagecluster_controller.go
sp98: Yes. I'll work on the KCS for dev preview.

malayparida2000:
@sp98 from this PR this particular code block
https://github.com/red-hat-storage/ocs-operator/blob/main/internal/controller/storagecluster/vault_agent.go#L45-L47
if !required {
		return obj.ensureDeleted(r, instance)
	}
	
Has introduced an issue. It keeps calling the deletion function continuosly forever. 
And that generates 404 not found errors in the audit logs.

See here https://redhat.atlassian.net/browse/DFBUGS-784?focusedCommentId=17341404

Can you please provide a follow up PR to solve this 404 error issue

PR #3791, opened 2026-04-14
fix missing toleration in rook-ceph-rbd-mirror pod

AnikaYadav:
Fix rbd-mirror pods failing to schedule on tainted nodes by using GetPlacement to configure placement from          StorageCluster. Add watch on StorageCluster to keep placement in sync.

### DIFF
--- a/internal/controller/mirroring/mirroring_controller.go
+++ b/internal/controller/mirroring/mirroring_controller.go
@@ -29,6 +29,7 @@ import (
 	"github.com/red-hat-storage/ocs-operator/v4/pkg/util"
 
 	"github.com/go-logr/logr"
+	storageclusterctrl "github.com/red-hat-storage/ocs-operator/v4/internal/controller/storagecluster"
 	rookCephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
 	"golang.org/x/exp/maps"
 	corev1 "k8s.io/api/core/v1"
@@ -306,6 +307,12 @@ func (r *MirroringReconciler) reconcileRbdMirror(clientMappingConfig *corev1.Con
 		return true
 	}
 
+	storageCluster, err := util.GetStorageClusterInNamespace(r.ctx, r.Client, clientMappingConfig.Namespace)
+	if err != nil {
+		r.log.Error(err, "failed to get StorageCluster")
+		return true
+	}
+
 	maintenanceModeRequested := len(storageConsumers.Items) >= 1
 
 	if shouldMirror && !maintenanceModeRequested {
@@ -314,6 +321,7 @@ func (r *MirroringReconciler) reconcileRbdMirror(clientMappingConfig *corev1.Con
 				return err
 			}
 			rbdMirror.Spec.Count = 1
+			rbdMirror.Spec.Placement = storageclusterctrl.GetPlacement(storageCluster, "rbd-mirror")
 			return nil
 		})
 		if err != nil {
--- a/internal/controller/mirroring/mirroring_controller.go
+++ b/internal/controller/mirroring/mirroring_controller.go
@@ -42,6 +42,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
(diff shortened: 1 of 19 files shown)


internal/controller/mirroring/mirroring_controller.go
```diff
@@ -314,6 +315,7 @@ func (r *MirroringReconciler) reconcileRbdMirror(clientMappingConfig *corev1.Con
 				return err
 			}
 			rbdMirror.Spec.Count = 1
+			rbdMirror.Spec.Placement = defaults.DaemonPlacements["rbd-mirror"]
```
malayparida2000: Call getPlacement func instead of directly using the placement struct
Like
https://github.com/red-hat-storage/ocs-operator/blob/8668ee98fd97e432bfe01a6b6922c621ffb0c8b8/internal/controller/storagecluster/cephnfs.go#L36

internal/controller/mirroring/mirroring_controller_test.go
```diff
@@ -0,0 +1,79 @@
+package mirroring
```
malayparida2000: The unit test addition can be done here https://github.com/red-hat-storage/ocs-operator/blob/8668ee98fd97e432bfe01a6b6922c621ffb0c8b8/internal/controller/storagecluster/placement_test.go#L26, instead of creating the new file here

malayparida2000 (changes requested):

internal/controller/mirroring/mirroring_controller.go
```diff
@@ -315,7 +316,13 @@ func (r *MirroringReconciler) reconcileRbdMirror(clientMappingConfig *corev1.Con
 				return err
 			}
 			rbdMirror.Spec.Count = 1
-			rbdMirror.Spec.Placement = defaults.DaemonPlacements["rbd-mirror"]
+			storageCluster, err := util.GetStorageClusterInNamespace(r.ctx, r.Client, clientMappingConfig.Namespace)
+			if err != nil {
+				r.log.Error(err, "failed to get StorageCluster")
+				return err
```
malayparida2000, line 322: Get the storageCluster CR outside this loop may be at Line 310 , so in future, if required by any other funcs it can be utilized without movement.

internal/controller/mirroring/mirroring_controller.go
```diff
@@ -141,6 +190,11 @@ func (r *MirroringReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			enqueueConfigMapRequest,
 			generationChangePredicate,
 		).
+		Watches(
+			&ocsv1.StorageCluster{},
+			enqueueConfigMapForStorageClusterPlacement,
```
malayparida2000, line 195: Not sure, couldn;t we have used just the enqueueConfigMapRequest why do we need to define a separate enqueueConfigMapForStorageClusterPlacement?

malayparida2000 (changes requested):
Changes are looking fine but the commit structure is a mess.
The first commit .
Seems to be unnecessary [fix missing toleration in rook-ceph-rbd-mirror pod](https://github.com/red-hat-storage/ocs-operator/pull/3791/changes/f57518d0300089bf56279039fb9960471c2db19b)

In the last commit we are removing the uni test file, which I think we should keep.

The commit structure can be of a total of 3 commits.

1. Export getPlacement & its related renaming only.
2. Use GetPlacement for rbd-mirror placement along with the unit test file
3. About adding the watch & predicate to the mirroring controller

internal/controller/mirroring/mirroring_controller.go
```diff
@@ -101,7 +102,26 @@ func (r *MirroringReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			}}
 		},
 	)
-
```
malayparida2000, line 104: keep the space here

internal/controller/mirroring/mirroring_controller.go
```diff
@@ -330,7 +331,11 @@ func (r *MirroringReconciler) reconcileRbdMirror(clientMappingConfig *corev1.Con
 		r.log.Error(err, "failed to list StorageConsumer(s)")
 		return true
 	}
-
```
malayparida2000, line 308: Keep the space here as well

malayparida2000 (changes requested):
@AnikaYadav The commit structure should be progressive in order meaning one change should go on the top of the other.

We can't have [Use GetPlacement for rbd-mirror placement](https://github.com/red-hat-storage/ocs-operator/pull/3791/changes/80b06a4f847c65aadb37d00f981221f33ea875f5)
before [Export getPlacement function to GetPlacement](https://github.com/red-hat-storage/ocs-operator/pull/3791/changes/ef73f3245c5047351871b92695925a7ad48733e9)

As without the export how can the function be used inside a diff package.

The logical ordering of the commits would be
1. Export the func
2. Use the func
3. Add watch & predicate

And along with that in each commit can you please add a bit more details on what's actually being changes or what's the context of the change.

For example in the export commit you could put that it needs to be exported so that it can used by other packages. And other packages need to use it as storageCluster is the only way we allow placement specifications.
Similarly you can add bit more context to each of the commits & the PR desc as a whole also

The changes look fine to me. Do get in touch with the QE member for the original bug & try to get a setup where this fix can be tested.

malayparida2000 (approved):

malayparida2000:
Fixes-https://redhat.atlassian.net/browse/DFBUGS-6129

PR #3792, opened 2026-04-15
util: disable reclaimspace for rbd virt storageclass

black-dragon74:
This patch disables (automated) reclaimspace for virt storageclass to prevent sparsify operations from blacklisting the kRBD clients (VMs).

### DIFF
--- a/pkg/util/storageclasses.go
+++ b/pkg/util/storageclasses.go
@@ -159,7 +159,6 @@ func NewDefaultVirtRbdStorageClass(
 		ObjectMeta: metav1.ObjectMeta{
 			Annotations: map[string]string{
 				"description": "Provides RWO and RWX Block volumes suitable for Virtual Machine disks",
-				"reclaimspace.csiaddons.openshift.io/schedule": "@weekly",
 			},
 			Labels: map[string]string{},
 		},


Rakshith-R (approved):
Virt storageclass is used for RWX block mode PVCs for VMs.

Only RBD sparsify runs for such PVC, which should only be run when PVC is unmounted.

For a RWX PVC, multiple client access is allowed and therefore, there is a risk that a sparsify operation while PVC was unmounted but continues when it get mounted. This has lead to scenario where clients are being randomly blocklisted.

Sparsify releases 4MB of zeroed extent and has little advantage of being run frequently.

Reclaimspace Op will be documented to be run manually as one time job for RWX RBD PVCs while the user making sure it is not mounted at that time.

iamniting (changes requested):
Do you want to remove this annotation from the existing storage classes as well?

if yes we need to make changes here as well. https://github.com/red-hat-storage/ocs-operator/blob/main/internal/controller/storagecluster/external_resources.go#L605

black-dragon74:
>> Do you want to remove this annotation from the existing storage classes as well?

> Nope, this is for the virt StorageClass only :)

I mean the existing virt storage class in the older versions?

black-dragon74:
> I mean the existing virt storage class in the older versions?

@iamniting Ah, my bad. Fixed. Do we need changes in client-op as well or the changes will be auto-reconciled by it due to diff annotations?

iamniting:
> > I mean the existing virt storage class in the older versions?
> 
> @iamniting Ah, my bad. Fixed. Do we need changes in client-op as well or the changes will be auto-reconciled by it due to diff annotations?

These changes are for external cluster. @leelavg Can tell us better about the client.

leelavg:
not sure of the fix @black-dragon74 

external:
you see there is an init section for which the created sc wouldn't be reconciled and the solution isn't full.

internal & remote:
metadata can't be removed to allow users manipulating non-managd meta, so existing annotation would continue to exist.

isn't it possible to have an oc command in the release note to remove this annotation on the cluster plus remove it from code such that they won't get readded by external mode script or client-op.

black-dragon74:
> metadata can't be removed to allow users manipulating non-managd meta, so existing annotation would continue to exist.

An ugly hack would be to set the annotation value as `""`

> isn't it possible to have an oc command in the release note to remove this annotation on the cluster plus remove it from code such that they won't get readded by external mode script or client-op.

Much cleaner, +1. 

I'll update the PR to remove the annotation from code. The release note would include a command to remove the annotation from existing SC.

leelavg:
> An ugly hack would be to set the annotation value as ""
- there was a long thread (around readaffinity) to change the behavior such that removals are actually enforced be it any part of the resources and when that's available we could probably allow removals, however that could involve changing protos, considering upgrades etc and so no ETA yet.

iamniting (approved):

iamniting:
@black-dragon74 pls raise a doc bug.

black-dragon74:
> @black-dragon74 pls raise a doc bug.

I'll create a new bug for 4.22 and 4.21. The plan is to have the removal command in release notes. We don't need a doc bug for that, right?

black-dragon74:
The CI failures are not related to this PR. Triggering it once more.

/test ocs-operator-bundle-e2e-aws

PR #3795, opened 2026-04-16
DFBUGS-6406: [release-4.18] Send OCS Sub channel to clients after validating with PackageManifest

malayparida2000:
A subscription channel change is pushed to the client and can start its upgrade. If the client updates before the provider’s CSV is ready, the provider may treat the client as ahead and set NotUpgradeable, deadlocking the path. Even requiring the provider CSV in Succeeded phase first, leaves a brief window where spec.channel is already updated but installedCSV lags the new graph; we would still advertise a channel the cluster has not fully moved to.
After the CSV is Succeeded, validate the channel by loading the ocs-operator PackageManifest (unstructured) and matching status.installedCSV to an entry in status.channels[].entries[].name for spec.channel—not to currentCSV alone, since the catalog head does not represent every valid patch on the line. That rejects the inconsistent window while still allowing non-tip patch levels. Serialize cached state with a mutex on concurrent provider RPCs;
add RBAC to get ClusterServiceVersions and PackageManifests in the operator namespace.

Backport of https://github.com/red-hat-storage/ocs-operator/pull/3227, https://github.com/red-hat-storage/ocs-operator/pull/3296, https://github.com/red-hat-storage/ocs-operator/pull/3796

### DIFF
--- a/deploy/ocs-operator/manifests/provider-role.yaml
+++ b/deploy/ocs-operator/manifests/provider-role.yaml
@@ -61,6 +61,14 @@ rules:
       - operators.coreos.com
     resources:
       - subscriptions
+      - clusterserviceversions
+    verbs:
+      - get
+      - list
+  - apiGroups:
+      - packages.operators.coreos.com
+    resources:
+      - packagemanifests
     verbs:
       - get
       - list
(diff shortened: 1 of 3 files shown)


malayparida2000:
The e2e failure is due to
`Found ClusterServiceVersion \"openshift-storage/cephcsi-operator.v4.17.0\" phase: Failed"`

It has been there since some time & is unrelated to this code change

leelavg (approved):

PR #3796, opened 2026-04-16
validate OCS channel using PackageManifest entries instead of currentCSV

malayparida2000:
Compare subscription.status.installedCSV to
status.channels[].entries[].name for the subscription channel instead of currentCSV alone. currentCSV is the catalog head; entries include older patches on the same channel, so we no longer require the cluster to be on the channel tip while still rejecting the case where spec.channel has moved but installedCSV is not yet on the new channel graph.

Fixes-https://redhat.atlassian.net/browse/DFBUGS-6516

### DIFF
--- a/services/provider/server/server.go
+++ b/services/provider/server/server.go
@@ -955,28 +955,36 @@ func (s *OCSProviderServer) isSubscriptionChannelValid(ctx context.Context, subs
 		klog.Infof("channels not found in PackageManifest %q", packageManifest.GetName())
 		return false, nil
 	}
-	var expectedCSVName string
 	for _, ch := range channels {
 		chMap, ok := ch.(map[string]any)
 		if !ok {
 			continue
 		}
-		if chMap["name"].(string) == channelName {
-			expectedCSVName, ok = chMap["currentCSV"].(string)
-			if !ok {
-				klog.Infof("currentCSV not found or invalid in channel %s", channelName)
-				return false, nil
-			}
-			break
+		if chn, ok := chMap["name"].(string); !ok || chn != channelName {
+			continue
 		}
-	}
-	if installedCSVName != expectedCSVName {
-		klog.Infof("installed CSV %q does not match expected %q", installedCSVName, expectedCSVName)
+		entries, ok := chMap["entries"].([]any)
+		if !ok {
+			klog.Infof("channel %q in PackageManifest %q has no entries", channelName, packageManifest.GetName())
+			return false, nil
+		}
+		for _, entry := range entries {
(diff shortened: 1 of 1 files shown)


services/provider/server/server.go
```diff
@@ -955,28 +955,48 @@ func (s *OCSProviderServer) isSubscriptionChannelValid(ctx context.Context, subs
...
+		for _, entry := range entries {
+			entryMap, ok := entry.(map[string]any)
 			if !ok {
-				klog.Infof("currentCSV not found or invalid in channel %s", channelName)
-				return false, nil
+				continue
+			}
+			entryCSVName, ok := entryMap["name"].(string)
+			if ok && entryCSVName != "" {
+				expectedCSVNames = append(expectedCSVNames, entryCSVName)
```
leelavg, line 981: why not `if ok && entryCSVName != "" && entryCSVName == installedCSVName { return true, nil }`?

leelavg (commented):
thanks for disclosing AI assisted and at the same time I suggest to scrutinize the code more before review, there is good amount of redundancy in this code.

malayparida2000:
@leelavg Have updated the PR now, Claude didn't really do a fine job when left unattended 😅.

malayparida2000:
basic validation from the e2e run logs

`2026-04-23T08:45:01.473697226Z I0423 08:45:01.473661       1 server.go:987] subscription for "ocs-operator" with channel "operator-sdk-run-bundle" is valid with installed CSV "ocs-operator.v4.22.0"`

services/provider/server/server.go
```diff
@@ -961,28 +955,38 @@ func (s *OCSProviderServer) isSubscriptionChannelValid(ctx context.Context, subs
...
-				return false, nil
+				continue
+			}
+			entryCSVName, ok := entryMap["name"].(string)
+			if !ok || entryCSVName == "" {
+				continue
+			}
+			if entryCSVName == installedCSVName {
+				klog.Infof("subscription for %q with channel %q is valid with installed CSV %q", subscription.Spec.Package, channelName, installedCSVName)
+				return true, nil
```
leelavg: already too many `continue`, can you do `ok && entryCSVName == installedCSVName`? one nit if can be solved, the log statement at 985 referencing the detail inferred at line 963, too much distant with multiple exist points in between.

malayparida2000: Updated the continue logic.
Can't think of any clever way to solve the log thing without restructuring.
In the current format, only after we traverse the whole list of channels can we conclude we did not find our channelName.

leelavg (approved):

PR #3797, opened 2026-04-16
Update Go version & direct dependecy package versions of ocs-operator

malayparida2000:


services/provider/api/go.mod
```diff
@@ -5,98 +5,100 @@ go 1.25.8
...
-	sigs.k8s.io/structured-merge-diff/v6 v6.3.2 // indirect
+	sigs.k8s.io/structured-merge-diff/v6 v6.4.0 // indirect
 	sigs.k8s.io/yaml v1.6.0 // indirect
 )
 
 require (
-	github.com/noobaa/noobaa-operator/v5 v5.0.0-20251118072940-a392e524a776
-	k8s.io/apimachinery v0.35.3
+	github.com/noobaa/noobaa-operator/v5 v5.21.0
+	k8s.io/apimachinery v0.35.4
```
iamniting: As these are direct dependencies can you move them in the first require section

iamniting (changes requested):

services/provider/api/go.mod
malayparida2000: Yeah, right, seems like this separate blob was left out previously. Updated now

iamniting (approved):

PR #3798, opened 2026-04-16
Skip OBC related resources if not found when OBC is in pending state

karthikps97:
### Describe the issue: 
When the OBC is in `Pending` state, its related resources like configMap, secret, OB would not be created. Current behavior is to return an error in case these are not found.

### Explain the change:
- If any supporting resources are not found, return only the OBC (pending state) in `GetDesiredClientState` RPC.
- Use only the OBC ResourceVersion for hash calculation in `ReportStatus` RPC.

### DIFF
--- a/services/provider/server/server.go
+++ b/services/provider/server/server.go
@@ -845,17 +845,23 @@ func (s *OCSProviderServer) getOBCResourceVersions(ctx context.Context, logger l
 		obc := &obcList.Items[i]
 
 		ob, configMap, secret, err := s.getOBCRelatedResources(ctx, obc.Name, consumer)
-		if err != nil {
+		if client.IgnoreNotFound(err) != nil {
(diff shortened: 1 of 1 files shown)


karthikps97:
@leelavg, @rewantsoni Could you please take a look.

services/provider/server/server.go
leelavg, line 1: possible to do this without changing the `getOBC...` a lot? feels a bit off that you've to validate both the error and values to take a decision. could you try embedding the error "%w" in the `getOBC...` func and at call site if it's `kerrors.IsNotFound(err)` skip reading the resources?

and ob, cm and secret form a set and even if one is missing consumption of storage can't happen so above should be fine?

karthikps97: Yes, Its possible to chain error. I will update this.

leelavg (approved):

PR #3803, opened 2026-04-21
Trim RBAC from ocs-operator serviceaccount

AnikaYadav:
Fix privilege escalation


rbac/ocs-metrics-exporter-clusterrole.yaml
```diff
@@ -0,0 +1,74 @@
...
+  - get
+  - create
+- apiGroups:
+  - csiaddons.openshift.io
+  resources:
+  - csiaddonsnodes
+  verbs:
+  - get
+  - list
+
```
iamniting, line 74: Can you remove this extra line at the end of file.

rbac/ocs-metrics-reader-clusterrole.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: ocs-metrics-reader
+rules:
+- nonResourceURLs:
+  - /metrics
+  - /healthz
+  - /readyz
+  verbs:
+  - get
+
```
iamniting, line 12: Can you remove this extra line at the end of file.

rbac/ocs-metrics-reader-clusterrolebinding.yaml
```diff
@@ -0,0 +1,13 @@
...
+  name: ocs-metrics-reader
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: ocs-metrics-reader
+subjects:
+- kind: ServiceAccount
+  name: prometheus-k8s
+  namespace: openshift-monitoring
+
```
iamniting, line 13: Can you remove this extra line at the end of file.

rbac/ocs-metrics-exporter-clusterrolebinding.yaml
```diff
@@ -0,0 +1,13 @@
...
+  name: ocs-metrics-exporter
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: ocs-metrics-exporter
+subjects:
+- kind: ServiceAccount
+  name: ocs-metrics-exporter
+  namespace: openshift-storage
+
```
iamniting, line 13: As this is at a cluster level, We may require an extra subject here for the `openshift-storage-extended` namespace

rohan47 (changes requested):
Please run `make gen-latest-csv` and commit the changes

Also all the changes to static RBAC can be in a single commit something like "convert metrics exporter and rook-ceph RBAC to static definitions"

rest looks good

leelavg:
@AnikaYadav here are some of the comments I have, rephrased with AI to expand in depth:

1. Missing ServiceAccount static manifest

`createMetricsExporterServiceAccount` in exporter.go was not removed. The ocs-metrics-exporter ServiceAccount should be added to --extra-service-accounts in hack/source-manifests.sh line 47 alongside ocs-provider-server, which lets operator-sdk bundle it into the CSV and OLM manage its lifecycle. The function and the serviceaccounts create/update/delete permissions in role.yaml can then be dropped.

2. ocs-metrics-svc Role and RoleBinding should be dropped, not added

rbac/exporter-service-role.yaml and rbac/exporter-service-role_binding.yaml already define an identical ocs-metrics-svc Role and RoleBinding. The new ocs-metrics-svc-role.yaml and ocs-metrics-svc-rolebinding.yaml added by this PR are duplicates and will conflict when the CSV generator processes them. Beyond the duplication, prometheus-k8s already has get/list/watch on services/endpoints/pods cluster-wide via the existing odf-prometheus and k8s-metrics-sm-prometheus-k8s ClusterRoles - the namespace-scoped Role is redundant regardless.

3. Dead constants

prometheusRoleName and metricsExporterRoleName in exporter.go lines 29-30 become unused after the RBAC functions are removed. Should be cleaned up.

4. Document the openshift-storage-extended case

The old ClusterRoleBinding used r.clusters.GetNamespaces() to add subjects for every StorageCluster namespace. The new static manifest hardcodes only openshift-storage. For an external StorageCluster in openshift-storage-extended, an admin would need to manually apply: ServiceAccount/ocs-metrics-exporter, Role/ocs-metrics-exporter, RoleBinding/ocs-metrics-exporter in that namespace, plus a separate ClusterRoleBinding binding ocs-metrics-exporter ClusterRole to the SA in openshift-storage-extended (the OLM-owned one cannot be patched). The rook bindings do not need to be included since internal StorageClusters are enforced to only run in the OperatorNamespace.

----

@travisn 

Background
Initially, the metrics exporter was a static component deployed only in openshift-storage. Commit 49d9846c19 ("Changes to ocs-operator code for multicluster mode") introduced support for multiple StorageClusters where external StorageClusters could run in additional namespaces (e.g. openshift-storage-extended), each requiring its own metrics exporter deployment with its own ServiceAccount and namespace-scoped RBAC. Since OLM only manages resources in the operator's install namespace, the operator had to dynamically create these resources at reconcile time - in whichever namespace the StorageCluster CR lived. The ClusterRoleBinding further needed to aggregate subjects from all StorageCluster namespaces at runtime using r.clusters.GetNamespaces().

Why this is a problem
To do all of this, the operator was granted write permissions on serviceaccounts, roles, rolebindings, clusterroles and clusterrolebindings. An operator with these combined permissions can create a ServiceAccount in any namespace and bind an existing ClusterRole to it, effectively granting that ServiceAccount the same level of access as the ocs-operator.

What PR #3803 should achieve
Move all RBAC resources for the primary namespace (openshift-storage) to static OLM-managed manifests, removing the operator's need for write permissions on RBAC resources entirely. The trade-off is that a second namespace (openshift-storage-extended) can no longer be handled automatically by the operator - instead the required resources are documented as a one-time manual step for the admin, which is an acceptable trade given that external StorageClusters in a second namespace are a deliberate administrative decision anyway.

The ask is, can we move this specific rbac creation for the openshift-storage-extended namespace to odf-cli or this whole rbac be documented as is?

thanks!

travisn:
> The ask is, can we move this specific rbac creation for the openshift-storage-extended namespace to odf-cli or this whole rbac be documented as is?

What RBAC is needed for the `openshift-storage-extended` namespace? 
For any RBAC not included in the csv, IMO it should just be documented with a yaml, and the user could `oc create` the RBAC. This is much more transparent than putting it in the odf cli.

leelavg:
> > The ask is, can we move this specific rbac creation for the openshift-storage-extended namespace to odf-cli or this whole rbac be documented as is?
> 
> What RBAC is needed for the `openshift-storage-extended` namespace? For any RBAC not included in the csv, IMO it should just be documented with a yaml, and the user could `oc create` the RBAC. This is much more transparent than putting it in the odf cli.

ServiceAccount/ocs-metrics-exporter
Role/ocs-metrics-exporter
RoleBinding/ocs-metrics-exporter in that namespace

plus a separate ClusterRoleBinding binding ocs-metrics-exporter ClusterRole to the SA in openshift-storage-extended (the OLM-owned one cannot be patched). The rook bindings do not need to be included since internal StorageClusters are enforced to only run in the OperatorNamespace.

so a total of 4 small to medium sized yamls.

iamniting:
FYi @SanjalKatiyar @yati1998 @Madhu-1

iamniting (commented):
Making the openshift-storage namespace hardcoded in the bindings will stop us to use the operator in any namespace. Which was requirement for ROSA if i am not mistaken.

PR #3810, opened 2026-04-23
support Key Rotation for RBD Mirror Peer

ShravaniVangur:
This PR aims to support Key Rotation for RBD Mirror. This includes introducing a new RPC (RotateMirroringKey) to support sharing of Key Generation for the specific cluster.


controllers/mirroring/mirroring_controller.go
```diff
@@ -50,6 +52,7 @@ const (
 	mirroringFinalizer              = "ocs.openshift.io/mirroring"
 	clientIDIndexName               = "clientID"
 	storageClusterPeerAnnotationKey = "ocs.openshift.io/storage-cluster-peer"
+	desiredCephxKeyGenEnvVarName    = "DESIRED_CEPHX_KEY_GEN"
```
rewantsoni: We are using this in multiple pkgs, let's move this to a util?

controllers/mirroring/mirroring_controller.go
```diff
@@ -231,6 +234,10 @@ func (r *MirroringReconciler) reconcilePhases(clientMappingConfig *corev1.Config
 		errorOccurred = true
 	}
 
+	if errored := r.reconcileRotateMirroringKey(ocsClient, storageClusterPeer); errored {
```
rewantsoni: we can have a better name for the method, maybe just rotatePeerMirroringKey or something similar? Since we are going to rotate the mirroring key for the peer

ShravaniVangur:
@iamniting Can you please review and approve this pr? 
Thank you.

iamniting (approved):

PR #3817, opened 2026-04-27
make client aware of the state corresponding to last known version

leelavg:
we already have checks in place to deny the request when the updated client calls the state from older server however the server only knows the client version based on last heartbeat which could be stale while the server is servicing newer client thinking the hearbeat was correct.

in best case the delays are about a minute, however even a single state exchange can delete resources in the client end if the state doesn't contain refreshed resources.

this was a dormant bug and surfaced when the kubeObjects structure is stated as part of OBC feature.

caveat:
the solution here stops the client reconciles when a mismatch occurs even at patch version.


leelavg:
should fix DFBUGS-6475. Bug was dormant till #3747.

leelavg:
will handle at client end alone.

PR #3822, opened 2026-04-27
prevent ODF auto-upgrade without manual approval in external mode

AnikaYadav:
Set disableInstallPlanAutoApproval=true to prevent auto-upgrading without manual approval. 

ocs-client-operator auto-approves InstallPlans if this key is not set, causing ODF to upgrade without user approval in external clusters. This fix ensures the key is set for external clusters.

https://redhat.atlassian.net/browse/DFBUGS-6043

### DIFF
--- a/internal/controller/storagecluster/external_resources.go
+++ b/internal/controller/storagecluster/external_resources.go
@@ -482,7 +482,7 @@ func (r *StorageClusterReconciler) createExternalStorageClusterResources(instanc
 		availableSCCs = append(availableSCCs, scc)
 	}
 
-	err = r.configureCsiDrivers(availableSCCs, instance)
+	err = r.configureCsiClient(availableSCCs, instance)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


iamniting (changes requested):
@AnikaYadav Can you pls update the proper commit msg and description stating Why this change is required?

iamniting (approved):

iamniting (changes requested):
This will never update the existing configmap as data wont be nil. Can you change this in such a way that it will also update the existing configmaps.

iamniting (commented):
@leelavg Can you pls take a look? I believe we should also change the func name.

leelavg:
> @leelavg Can you pls take a look? I believe we should also change the func name.
- `configureCsiClient` comes to my mind if we want to change, can't think of anything else.

iamniting (approved):

PR #3828, opened 2026-04-28
Remove Enum `NotifyReason`

shirady:
### Describe the problem
Part of [RHSTOR-6230](https://issues.redhat.com/browse/RHSTOR-6230)
Tech debt that was mentioned in the comment https://github.com/red-hat-storage/ocs-operator/pull/3690#issuecomment-3920110756:

> a follow up PR should move the Enum from grpc code and into go code as of the reason that we wanted to make sure that adding additional event types should not require a protobuf change (`services/provider/api/provider.pb.go`).

### Explain the changes
1. Remove the `NotifyReason` and add the type `NotifyReason` with const using `iota`.
Note: the new type and const were added in file `services/provider/api/interfaces/interfaces.go` (see https://github.com/red-hat-storage/ocs-operator/pull/3828#discussion_r3160042909) as they are used in the file `services/provider/server/server.go` (`package server`) and `services/provider/api/client/client.go` (`package client`) and I want to avoid duplicate code.
2. gRPC auto-generated files after running `make gen-protobuf`.
3. Edited the files in the server and client and tests that used this enum.
4. Remove unused `mockNotify`.

#### Unit Tests
Please run: `go test ./services/provider/server -run TestNotify count=1`
Output:
```bash
ok  	github.com/red-hat-storage/ocs-operator/v4/services/provider/server	1.062s
```


shirady:
@rewantsoni @leelavg, please take a look

shirady:
@nb-ohad @leelavg @rewantsoni I just wanted to make sure that the comment https://github.com/red-hat-storage/ocs-operator/pull/3690#issuecomment-3920110756 is still relevant because in PR #3747 @karthikps97 made the changes in enum.

services/provider/api/api_types/api_types.go
nb-ohad, line 1: I don't think that we need another package introduced. This enum and the type associated with it are part of the server API (even if not formalized within the protobuf definition). I would ask to move this to an existing package that deal with the server API. One such a place can be https://github.com/red-hat-storage/ocs-operator/blob/main/services/provider/api/interfaces/interfaces.go 

@leelavg Do you agree?

nb-ohad (changes requested):

services/provider/api/api_types/api_types.go
leelavg: yes, rechecked, that'd be a good spot as it's imported into both server & client. The pkg being named `interfaces` was a bit off when it got first created but didn't revisit.

leelavg:
> PR https://github.com/red-hat-storage/ocs-operator/pull/3747 @karthikps97 made the changes in enum.
- those are for type of op to perform, unrelated to this PR or are you suggesting we are anyways using enum in other places and it may not be mandatory to change for `NotifyReason`. I don't have a clear cut answer for this.

services/provider/api/api_types/api_types.go
shirady: I will move it to this file, thanks.

nb-ohad:
> > PR #3747 @karthikps97 made the changes in enum.
> 
> * those are for type of op to perform, unrelated to this PR or are you suggesting we are anyways using enum in other places and it may not be mandatory to change for `NotifyReason`. I don't have a clear cut answer for this.

@leelavg @karthikps97 Please note that there is nothing wrong with having enums in protobuf. The reason I requested to move this enum out of protobuf because we are talking here potentially about an open set not a close one, as opposed to the close set that is the kube object op. The bottom line is that kube object op has a theoretical limit (even if we are not implementing all of possible operations today) while events type is an open set.

PR #3845, opened 2026-05-06
DFBUGS-6704: [release-4.19]: fix isEncrptionSettingUpdated method

sp98:
isEncrptionSettingUpdated only checks the
spec.encryption.clusterWide option before
deciding to set the `yes-really-migrate-osds`
flag to cephCluster.

For legacy cluster, < 4.10, the storagecluster
only has spec.encryption.enable for encryption.

So this PR updates the isEncrptionSettingUpdated
function to check for both `spec.encryption.enable` and ``spec.encryption.clusterWide` before setting
the "yes-really-migration-osds` flag in the cephcluster.

(cherry picked from commit 48fbd0338ab4c1a29e7c7e2c053137ca410f1184)


subhamkrai (commented):
/lgtm

malayparida2000 (approved):

PR #3846, opened 2026-05-07
tls: transfer TLSProfile configuration to rook rgw

leelavg:
Wire TLSProfile into CephObjectStore (RGW) reconciliation using the "rook.io" selector, translating the profile config into OpenSSL cipher and group names in Spec.Security. When no profile matches, TlsGroups defaults to "DEFAULT" enabling X25519MLKEM768 hybrid PQ key exchange. Includes unit tests.

See individual commits for details.

Description and tests are generated by AI.


leelavg:
@subhamkrai ptal, thanks!

leelavg:
/cherrypick release-4.22
via [DFBUGS-6819](https://redhat.atlassian.net/browse/DFBUGS-6819)

rewantsoni:
LGTM

hack/common.sh
```diff
@@ -22,7 +22,7 @@ OCS_FINAL_DIR="deploy/ocs-operator/manifests"
 BUNDLEMANIFESTS_DIR="rbac"
 
 # External images used in CSV generation
-LATEST_ROOK_IMAGE="quay.io/ocs-dev/rook-ceph:vmaster-b0d17ebba" # Using downstream rook image as it contains downstream-only changes
+LATEST_ROOK_IMAGE="quay.io/ocs-dev/rook-ceph:vrelease-4.22-f49f22505" # Using downstream rook image as it contains downstream-only changes
```
malayparida2000, line 25: This change should be part of the commit about bundle image changes.

internal/controller/storagecluster/noobaa_system_reconciler.go
```diff
@@ -288,20 +287,22 @@ func (r *StorageClusterReconciler) setNooBaaDesiredState(nb *nbv1.NooBaa, sc *oc
...
+		}
 		gotls := ocstlsv1.GetGoTLSConfig(cfg)
-		apiSecurity := cmp.Or(nb.Spec.Security.APIServerSecurity, &nbv1.TLSSecuritySpec{})
 		switch gotls.MinVersion {
 		case tls.VersionTLS12:
-			apiSecurity.TLSMinVersion = ptr.To(nbv1.VersionTLS12)
+			nb.Spec.Security.APIServerSecurity.TLSMinVersion = ptr.To(nbv1.VersionTLS12)
 		case tls.VersionTLS13:
-			apiSecurity.TLSMinVersion = ptr.To(nbv1.VersionTLS13)
+			nb.Spec.Security.APIServerSecurity.TLSMinVersion = ptr.To(nbv1.VersionTLS13)
```
malayparida2000, line 298:
```
Also replaces cmp.Or pointer aliasing with explicit nil-checks in both
cephobjectstores.go and noobaa_system_reconciler.go for clarity.

```
Please make this a separate commit

internal/controller/storagecluster/cephobjectstores.go
```diff
@@ -307,6 +310,35 @@ func (r *StorageClusterReconciler) newCephObjectStoreInstances(initData *ocsv1.S
...
+				return nil, err
+			}
+
+			osstls := ocstlsv1.OpenSSLConfigFrom(ocstlsv1.GetGoTLSConfig(cfg))
+			switch osstls.Protocol {
+			case "TLSv1.3":
+				if obj.Spec.Security.SslOptions == nil {
+					obj.Spec.Security.SslOptions = &cephv1.SslOptionsSpec{}
+				}
+				// rook by default disables older versions except 1.2 and we explicitly disable it
```
malayparida2000, line 327: I could not understand this block clearly. Does SSLv2 transalates to 1.2?

malayparida2000:
Seems like the PR was merged before I was able to post the review

hack/common.sh
leelavg: it wasn't because the bundle commit only contains generated changes, it's intentional to show this image contains the new rook api.

internal/controller/storagecluster/noobaa_system_reconciler.go
leelavg: yes, thought of it but that would also need a backport, it's not a bug but a refactor in similar code based on learned mistake the cmp.Or isn't being using correctly.

internal/controller/storagecluster/cephobjectstores.go
leelavg: no, it's tls1.2, sslv2 is by default disabled and we are explicitly disabling tls1.2 when the config is to only support tls1.3

leelavg: oh! you are correct, will make the change.

PR #3849, opened 2026-05-08
tls: fix wrong ssl option used to disable tls 1.2 for rgw

leelavg:
SSLv2 is already disabled by default in beast; setting it to false was a no-op. Use TLSv1_2=false instead, which is what beast enables by default and must be explicitly disabled to make TLS 1.3 the minimum.


leelavg:
fixes https://github.com/red-hat-storage/ocs-operator/pull/3846#discussion_r3206570035

malayparida2000 (approved):

PR #3859, opened 2026-05-13
autoscaler: add atleast 10% of storage with scaling

parth-gr:
during horizontal scaling after sometime,
we were just adding 3 more osd,
but in real use case we atleast need to scale the cluster by 10%, so atleast add 10% storage in the horizontal scaling

### DIFF
--- a/internal/controller/storageautoscaler/storageautoscaler_controller.go
+++ b/internal/controller/storageautoscaler/storageautoscaler_controller.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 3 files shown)


internal/controller/storageautoscaler/storageautoscaler_controller.go
```diff
@@ -550,14 +552,56 @@ func calculateExpectedOsdSizeAndCount(storageCluster *ocsv1.StorageCluster, stor
 		// vertical scaling
 		// double the osd size
 		expectedOsdSize = *resource.NewQuantity(startOsdSize.Value()*2, startOsdSize.Format)
+
```
travisn, line 556: This method has a lot of calculations. How about adding more unit tests for all the edge cases in the calculations to make sure we get the expected number of OSDs or detect limit is reached? The unit tests will give more confidence in the calculations than trying to code review it.

travisn (changes requested):
How about updating the storage-auto-scale.md design doc?

parth-gr:
Updated ready for final review

deploy/storageautoscaler.yaml
```diff
@@ -8,6 +8,6 @@ spec:
     name: ocs-storagecluster
   deviceClass: ssd
   storageCapacityLimit: 100Ti
-  maxOsdSize: 8Ti
-  storageScalingThresholdPercent: 70
+  maxOsdSize: 100Gi
```
travisn: Revert this file from testing

internal/controller/storageautoscaler/storageautoscaler_controller.go
```diff
@@ -550,14 +553,56 @@ func calculateExpectedOsdSizeAndCount(storageCluster *ocsv1.StorageCluster, stor
...
+			// calculate the maximum device sets we can add without exceeding the limit
+			remainingCapacity := storageAutoScaler.Spec.StorageCapacityLimit.Value() - startStorageCapacity.Value()
+			capacityPerDeviceSet := int64(replicaCount*deviceSetCount) * startOsdSize.Value()
+
+			maxDeviceSetsToAdd := int(remainingCapacity / capacityPerDeviceSet)
+
+			if maxDeviceSetsToAdd < 1 {
+				// cannot add even one device set, limit reached
+				limitReached = true
+				devicesToAdd = 1
```
travisn, line 593: Will this cause an expansion even after reaching the limit in a previous expansion? It appears to always allowing expansion by 1 device set even at the max.

travisn (changes requested):

internal/controller/storageautoscaler/storageautoscaler_controller.go
parth-gr: No, this is for the capcity limit alert message, where we need the next atleast available capacity to scale it.

```
			Message:   fmt.Sprintf("storage capacity limit reached for storageAutoScaler %q device class %q least expected storage capacity is %q", storageAutoScaler.Name, storageAutoScaler.Spec.DeviceClass, expectedStorageCapacity.String()),

```

travisn (approved):
/lgtm

PR #3861, opened 2026-05-14
logcollector: make resource request equal to limits

subhamkrai:
To satisfy the Guaranteed QoS criteria let's make
logcollector resource Requests==Limits.

### DIFF
--- a/pkg/defaults/resources.go
+++ b/pkg/defaults/resources.go
@@ -116,8 +116,8 @@ var (
 		},
 		"crashcollector": {
 			Requests: corev1.ResourceList{
-				corev1.ResourceCPU:    resource.MustParse("50m"),
-				corev1.ResourceMemory: resource.MustParse("50Mi"),
+				corev1.ResourceCPU:    resource.MustParse("100m"),
+				corev1.ResourceMemory: resource.MustParse("100Mi"),
 			},
 			Limits: corev1.ResourceList{
 				corev1.ResourceCPU:    resource.MustParse("100m"),
@@ -126,8 +126,8 @@ var (
 		},
 		"logcollector": {
 			Requests: corev1.ResourceList{
-				corev1.ResourceCPU:    resource.MustParse("50m"),
-				corev1.ResourceMemory: resource.MustParse("50Mi"),
+				corev1.ResourceCPU:    resource.MustParse("100m"),
+				corev1.ResourceMemory: resource.MustParse("250Mi"),
 			},
 			Limits: corev1.ResourceList{
 				corev1.ResourceCPU:    resource.MustParse("100m"),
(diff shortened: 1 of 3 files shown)


subhamkrai:
@travisn @malayparida2000 ^^

Also, @travisn I noticed for crashcollector the Requests!=Limits, do we need to change that as well?

https://github.com/red-hat-storage/ocs-operator/blob/99ddcb2941813f77aa49a2d50f7bca798dd3b351/pkg/defaults/resources.go#L117-L125

malayparida2000:
> @travisn @malayparida2000 ^^
> 
> Also, @travisn I noticed for crashcollector the Requests!=Limits, do we need to change that as well?
> 
> https://github.com/red-hat-storage/ocs-operator/blob/99ddcb2941813f77aa49a2d50f7bca798dd3b351/pkg/defaults/resources.go#L117-L125

Yeah we should make the change for that as well

malayparida2000:
lgtm but not sure why there is a change in go.work.sum

subhamkrai:
> lgtm but not sure why there is a change in go.work.sum

I just ran, make ocs-operator and that added the changes

malayparida2000 (approved):

pkg/defaults/resources.go
```diff
@@ -116,8 +116,8 @@ var (
 		},
 		"crashcollector": {
 			Requests: corev1.ResourceList{
-				corev1.ResourceCPU:    resource.MustParse("50m"),
-				corev1.ResourceMemory: resource.MustParse("50Mi"),
+				corev1.ResourceCPU:    resource.MustParse("100m"),
```
travisn, line 119: Let's actually revert this change to the crash collector. It is a separate pod from the ceph daemons, and does not require the highest Guaranteed QoS. We can leave it as burstable.

pkg/defaults/resources.go
```diff
@@ -126,8 +126,8 @@ var (
 		},
 		"logcollector": {
 			Requests: corev1.ResourceList{
-				corev1.ResourceCPU:    resource.MustParse("50m"),
-				corev1.ResourceMemory: resource.MustParse("50Mi"),
+				corev1.ResourceCPU:    resource.MustParse("100m"),
```
travisn, line 129: Let's revert this CPU change, and instead reduce the CPU limits to `50m`. It's ok if the log collector runs slower.

pkg/defaults/resources.go
```diff
@@ -126,8 +126,8 @@ var (
...
+				corev1.ResourceMemory: resource.MustParse("250Mi"),
```
travisn, line 130: This will effectively increase the ODF memory requirements for every Ceph daemon. We need to be sensitive to increasing those resources requirements unless absolutely necessary. Instead of increasing to match the limits, let's consider a change somewhere in the middle where we reduce the limits and increase the requests. Perhaps `75Mi` or `100Mi` is sufficient. As I recall, the log collector is normally only using 50Mi, so a small bump hopefully is all that is needed, instead of increasing all the way to 250Mi. Let's do a bit of research to see if we can get a minimal memory limit that will be sufficient.

travisn (commented):
I know the suggestion in the bug was to just increase the requests to match the limits, but that was overly simplistic. We really need to consider how to avoid increasing ODF resources as much as possible since the log collector affects all ceph daemons.

pkg/defaults/resources.go
subhamkrai: I'm chatting with Lijo and @Nikhil-Ladha to understand the resource request/limit we should set.

PR #3888, opened 2026-06-01
exclude alerts for probe failed nodes

yati1998:
Currently we are calculating node latency for all
ips from the cluster, but in case of multus
enabled cluster where we have more multiple ips
one in default network path and other in different path than blackbox-exporter.
In this case the alert is raised for the ips whose probe failed which is not our requirement.
This commit modifies the node latency to raise
alerts for the nodes or ips for which the
probes_success is 1

### DIFF
--- a/metrics/deploy/prometheus-ocs-rules.yaml
+++ b/metrics/deploy/prometheus-ocs-rules.yaml
@@ -674,6 +674,11 @@ spec:
               daemon_type="osd"
               }[24h:]
           ) > 0.010
+          and on(instance, job, daemon_type, namespace)
+          probe_success{
+            job="odf-blackbox-exporter",
+            daemon_type="osd"
+          } == 1
         ) > 0
       for: 10m
       labels:
@@ -693,6 +698,11 @@ spec:
               daemon_type="mon"
               }[24h:]
           ) > 0.100
+          and on(instance, job, daemon_type, namespace)
+          probe_success{
+            job="odf-blackbox-exporter",
+            daemon_type="mon"
+          } == 1
         ) > 0
       for: 10m
       labels:


yati1998:
oc get pods for osd nodes output: 
                            "name": "cilium",
                            "interface": "eth0",
                            "ips": [
                                "172.28.121.231"
                            ],
                            "mac": "32:39:63:14:7a:09",
                            "default": true,
                            "dns": {},
                            "gateway": [
                                "172.28.123.78"
                            ]
                        },{
                            "name": "openshift-storage/odf-cluster-net",
                            "interface": "net1",
                            "ips": [
                                "172.16.0.56"
                            ],
                            "mac": "be:26:a8:06:83:36",
                            "dns": {}
                        }]
blackbox-exporter-logs

```
2026-05-28T12:45:22.364143679Z time=2026-05-28T12:45:22.364Z level=WARN source=icmp.go:345 msg="Timeout reading from socket" module=icmp_internal target=172.16.0.15 err="read udp 0.0.0.0:52417: raw-read udp 0.0.0.0:52417: i/o timeout"
2026-05-28T12:45:22.364164037Z time=2026-05-28T12:45:22.364Z level=ERROR source=handler.go:130 msg="Probe failed" module=icmp_internal target=172.16.0.15 duration_seconds=5.000900954
```

Since there are multiple networks present, the one that is not in the path of blackbox-exporter fails the probes giving false alerts. 

@weirdwiz @aruniiird please do review

weirdwiz (approved):
looks good to me mostly, but is there a way we can just probe the nodes we want to? instead of filtering it out? 

i guess works both ways for now

aruniiird (approved):
Changes looks good

yati1998:
> looks good to me mostly, but is there a way we can just probe the nodes we want to? instead of filtering it out?
> 
> i guess works both ways for now

No, I initially tries that approach only, but in different clusters we might have different default network paths, initially when I tested in multus cluster we didn't get this error, but with this customer they use celium as default network path and hence we are getiing this error. So it would be good to scrape all ips and trigger alerts only for required nodes

malayparida2000 (approved):
Approving based on Divansh's & Arun's acks.

